Issue #280, opened 2017-10-31
.getItem from a local/session storage throws when 3rd party cookies/storage is disabled - chrome mobile & other similar behaving browsers


AVVS:
### Describe your environment

  * Operating System version: OS X 10.13
  * Firebase SDK version: 4.5.2
  * Firebase Product: auth, database & storage

### Describe the problem

Accessing window.localStorage && localStorage.getItem may throw security error based on settings of chrome, code is minified so cant really point to where it happens, but search via .getItem in the auth dist and it will be there

bojeil-google:
Can you provide exact steps how to recreate this? Right now, I am not sure what you are experiencing.
We currently persist the Auth state in web storage (localStorage). Before we do so, we try and catch setting and getting an item from localStorage. 
3rd party cookies/data needs to be enabled for the state to persist, but if they are not enabled, we fallback to in memory state persistence.

AVVS:
This is the best I dug out of the logs, you can see auth service is called, and then it goes on from that. I assume somewhere its not checking properly for localstorage availability - needs to be wrapped in the try/catch even for simple window.localStorage calls - since that throws in certain modes

<img width="1487" alt="screen shot 2017-10-31 at 12 50 33 am" src="https://user-images.githubusercontent.com/1713617/32212848-f1f28c8a-bdd5-11e7-9b23-a56d293d0160.png">

AVVS:
Dont have unminified version since it still wasnt open sourced per say - dist was just a minified version, used `@firebase/auth` package

AVVS:
Another log - might be helpful

<img width="1469" alt="screen shot 2017-10-31 at 12 56 41 am" src="https://user-images.githubusercontent.com/1713617/32212975-8b60e3c6-bdd6-11e7-9e83-c9053728e267.png">

bojeil-google:
Tried testing it with the Chrome emulator but can't reproduce it as you report it.
Are you running some operation like `signInWithPopup` or does this trigger on `firebase.initializeApp`? Are you changing Auth state persistence?
Some snippets would make my life easier as I am still forced to speculate.

AVVS:
I do this in 2 cases:

Initializing redux store:

```
export const firebaseApp = firebase.initializeApp(firebaseConfig);
const auth = firebaseApp.auth();

// make sure we only persist session and nothing else
auth.setPersistence(firebase.auth.Auth.Persistence.NONE);

...

// enable listener for id token and update http auth header accordingly
[...]

Issue #245, opened 2017-10-20
Signing in using a disabled account, using signInWithPopup, returns an 'Internal Error'


pmullot:
### Version info

**Angular:** 4.4.3

**Firebase:** 

**AngularFire:** 4.0.0-rc.2

**Other (e.g. Ionic/Cordova, Node, browser, operating system):** no

### How to reproduce these conditions
When you try to sign in using a Disabled account (via the Firebase Console), using signInWithPopup, it returns an 'Internal Error'.

**Steps to set up and reproduce**
With a typical login function, as described in all examples and tutorials

```
login() {
	const provider = new firebase.auth.GoogleAuthProvider();
	this.afAuth.auth.signInWithPopup(provider)
		.then(credential =>  {
	        // use credential observable to get user data ...
		// with a disabled user, this is not reached
		})
		.catch(error => {
		  console.log(error);
		  // error =  {code: "auth/internal-error", message: "An internal error has occurred."}
		});
}
```

### Expected behavior

Returned error should specify "User disabled" or something similar

### Actual behavior

Returned error isn't specific enough. It's also misguiding since it refers to an internal error
`{code: "auth/internal-error", message: "An internal error has occurred."}`

rommelpe:
@pmullot, I believed this issue was fixed in the tagged commit. I was able to confirm it with [my sample repro](https://stackblitz.com/edit/angular-hxsaia) using Angular:4.4.3 with @angular/fire:5.0.0 (the old AngularFire was [depreciated](https://www.npmjs.com/package/angularfire2) so I’ve used the older version of new @angular/fire). Let us know if you are still having issues.

PR #287, opened 2017-11-02
fix(auth): fixes various Auth bugs.

bojeil-google:
Fixes verifyAssertion unrecoverable errors when returnIdpCredential is set to true.
In this case, the error code is returned along with the credential in the errorMessage without any
ID token/refresh token.

Catch, suppress/handle when localStorage is null or when `localStorage.getItem`
throws a security error due to access being disabled by the browser for whatever reason.


bojeil-google:
This should fix the following 2 issues:
https://github.com/firebase/firebase-js-sdk/issues/280
https://github.com/firebase/firebase-js-sdk/issues/245

PR #296, opened 2017-11-06
b/68058918: Async Queue error handling testing / improvement.

mikelehen:
The bug reported in b/68058918 did not actually exist, but I've added testing for it (1) and made a minor improvement (2).

1. Adds testing to ensure that the original error is preserved and propagated in
  the face of multiple failures in the queue (this already worked).
2. Adds an asynchronous exception on queue failures so that they can be caught by
  crash reporting libraries (otherwise, only subsequent failures with
  "AsyncQueue is already failed: ..." would be caught).

### DIFF
--- a/packages/firestore/test/unit/util/async_queue.test.ts
+++ b/packages/firestore/test/unit/util/async_queue.test.ts
@@ -82,7 +82,7 @@ describe('AsyncQueue', () => {
       })
       .then(
         () => {
-          expect.fail('expected an error');
+          expect.fail('expected op1 to fail');
         },
         (err: AnyJs) => {
           expect(err).to.equal(expected);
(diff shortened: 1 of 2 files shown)


mikelehen:
Travis failure was addressed by https://github.com/firebase/firebase-js-sdk/pull/298.  I think this is good to go now.

packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -90,10 +90,38 @@ describe('AsyncQueue', () => {
...
+    // executed and failed). It should not be run.
+    const op2Promise = queue
+      .schedule(() => {
+        return defer(() => {
+          expect.fail('op2 should not be executed.');
+        });
+      })
+      .then(
+        () => {
+          expect.fail('expected an error');
```
wilhuff: Maybe localize these failures to which operation they apply to? e.g. 'op2 expected an error' here and 'op1 ...' above?

mikelehen: Done.

PR #297, opened 2017-11-06
Sending an empty write request before tearing down the WriteStream.

schmidt-sebastian:
This PR makes sure that we sent an empty write request before the write steam closes. Most of the changes in this PR are geared towards making this testable via the Spec tests (which includes adding support for enable/disableNetwork and request counting).

### DIFF
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -323,16 +323,16 @@ export abstract class PersistentStream<
       this.backoff.resetToMax();
     }
 
-    // This state must be assigned before calling onClose() to allow the callback to
-    // inhibit backoff or otherwise manipulate the state in its non-started state.
-    this.state = finalState;
-
     // Clean up the underlying stream because we are no longer interested in events.
     if (this.stream !== null) {
+      this.tearDown();
       this.stream.close();
       this.stream = null;
     }
 
+    // This state must be assigned before calling onClose() to allow the callback to
+    // inhibit backoff or otherwise manipulate the state in its non-started state.
+    this.state = finalState;
     const listener = this.listener!;
 
     // Clear the listener to avoid bleeding of events from the underlying streams.
@@ -346,6 +346,12 @@ export abstract class PersistentStream<
       return Promise.resolve();
     }
   }
+
+  /**
+   * Can be overridden to perform additional cleanup before the stream is closed.
+   * Calling super.tearDowm() is not required.
+   */
+  protected tearDown(): void {}
 
   /**
    * Used by subclasses to start the concrete RPC and return the underlying
@@ -644,6 +650,12 @@ export class PersistentWriteStream extends PersistentStream<
   start(listener: WriteStreamListener): void {
     this.handshakeComplete_ = false;
     super.start(listener);
+  }
+
+  protected tearDown() {
+    if (this.handshakeComplete_) {
+      this.writeMutations([]);
+    }
   }
 
   protected startRpc(
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -349,7 +349,7 @@ export abstract class PersistentStream<
 
   /**
    * Can be overridden to perform additional cleanup before the stream is closed.
-   * Calling super.tearDowm() is not required.
+   * Calling super.tearDown() is not required.
    */
   protected tearDown(): void {}
 
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -237,4 +237,21 @@ describeSpec('Listens:', [], () => {
         .expectEvents(allQuery, { fromCache: false })
     );
   });
+
+  specTest('Listens are reestablished after network disconnect', [], () => {
+    const query = Query.atPath(path('collection'));
+    const docA = doc('collection/a', 1000, { key: 'a' });
+    const docB = doc('collection/b', 2000, { key: 'b' });
+    return spec()
+      .userListens(query)
+      .expectNumWatchStreamRequests(1)
+      .watchAcksFull(query, 1000, docA)
+      .expectEvents(query, { added: [docA] })
+      .disableNetwork()
+      .enableNetwork()
+      .restoreListen(query, 'resume-token-1000')
+      .expectNumWatchStreamRequests(2)
+      .watchAcksFull(query, 2000, docB)
+      .expectEvents(query, { added: [docB] });
+  });
 });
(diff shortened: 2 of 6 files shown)


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -420,6 +429,7 @@ abstract class TestRunner {
 
   run(steps: SpecStep[]): Promise<void> {
     console.log('Running spec: ' + this.name);
+    this.connection.reset();
```
wilhuff: Why put this in start() along with all the rest of the setup logic?

packages/firestore/test/util/equality_matcher.ts
```diff
@@ -71,8 +71,8 @@ export function addEqualityMatcher() {
 
           this.assert(
             customDeepEqual(left, right),
-            'expected #{this} to roughly deeply equal #{exp}',
-            'expected #{this} to not roughly deeply equal #{exp}',
+            'expected #{act} to roughly deeply equal #{exp}',
```
wilhuff: What are `act` and `exp` here? Could you add a comment if this is actually correct?

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -348,6 +348,12 @@ export abstract class PersistentStream<
   }
 
   /**
+   * Can be overridden to perform additional cleanup before the stream is closed.
+   * Calling super.tearDowm() is not required.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -237,4 +237,21 @@ describeSpec('Listens:', [], () => {
...
+    const docB = doc('collection/b', 2000, { key: 'b' });
+    return spec()
+      .userListens(query)
+      .expectNumWatchStreamRequests(1)
+      .watchAcksFull(query, 1000, docA)
+      .expectEvents(query, { added: [docA] })
+      .disableNetwork()
+      .enableNetwork()
+      .restoreListen(query, 'resume-token-1000')
+      .expectNumWatchStreamRequests(2)
```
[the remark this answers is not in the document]
schmidt-sebastian: I added a comment to `expectWriteStreamRequestCount` and `expectWatchStreamRequestCount`.

I also changed the test to use the same format as the write stream test, which uses named parameters.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -129,6 +129,27 @@ export class SpecBuilder {
     return this;
   }
 
+  /** Registers a previously active target after a stream disconnect. */
+  restoreListen(query: Query, resumeToken: string): SpecBuilder {
```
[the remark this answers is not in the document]
schmidt-sebastian: It does, but the internal state in the spec test doesn't know about this (all state gets cleared by the spec tests when the stream closes). This bring the spec test state in line with what remote store should have.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -493,6 +534,22 @@ export class SpecBuilder {
     return this;
   }
 
+  expectNumWriteStreamRequests(num: number): SpecBuilder {
```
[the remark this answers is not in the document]
schmidt-sebastian: Comment added.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Moved to start().

packages/firestore/test/util/equality_matcher.ts
schmidt-sebastian: Without this change, these assertions provide little value. `this` is the object that is asserted on, and not the asserted value. See how it's done here: https://github.com/chaijs/chai/blob/3ace4a0bbd14cb3fad45067c29fed9880936e3b7/lib/chai/core/assertions.js#L1170

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -493,6 +537,29 @@ export class SpecBuilder {
     return this;
   }
 
+  /**
+   * Verifies fhe total number of requests sent to the write backend since test
```
wilhuff: s/fhe/the/ here and below

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -493,6 +537,29 @@ export class SpecBuilder {
...
+   * initialization.
+   */
+  expectWriteStreamRequestCount(num: number): SpecBuilder {
+    this.assertStep('Expectations require previous step');
+    const currentStep = this.currentStep!;
+    currentStep.stateExpect = currentStep.stateExpect || {};
+    currentStep.stateExpect.writeStreamRequestCount = num;
+    return this;
+  }
+  /**
```
wilhuff: Newline between functions.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -89,6 +89,8 @@ import {
   version
 } from '../../util/helpers';
 
+type MutationBatch = api.Write[];
```
wilhuff: FWIW I'm not super wild about this name. It's an array of `Writes` not "mutations". It also makes things more confusing since we have members like `earlyWrites` that now point to mutation batches.

Note that `api.Write` is the proto while `Mutation` is our internal type. It's worth it not to confuse these.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -120,15 +128,22 @@ class MockConnection implements Connection {
...
+    this.earlyWrites = [];
+    this.activeTargets = [];
+  }
+
   invoke(rpcName: string, request: any): Promise<any> {
     throw new Error('Not implemented!');
   }
 
-  waitForWriteSend(): Promise<api.Write> {
+  waitForMutationsSend(): Promise<MutationBatch> {
```
wilhuff: This should be `waitForWritesSend`

packages/firestore/test/unit/specs/spec_builder.ts
schmidt-sebastian: Done

packages/firestore/test/unit/specs/spec_builder.ts
schmidt-sebastian: Done

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: I struggled to name this, and the best I could come up with is either MutationBatch or WriteRequest (since I don't want earlyWrites to be a WriteList[]). I ended up going with WriteRequest - and removing the type altogether since we can just use api.WriteRequest.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: With the rename above, this is now waitForWriteRequest.

schmidt-sebastian:
Comments addressed.

PR #304, opened 2017-11-08
Upgrade closure-builder to fix #183.

### LINKED ISSUE #183: "Image is not defined" error when using Cloud Firestore from React Native app
[REQUIRED] Describe your environment

react-native: 0.47.2,
Firebase SDK version: 4.5.0
Firebase Product: cloud firestore_ (auth, database, storage, etc)

[REQUIRED] Describe the problem

Around 2 - 3 mins after I query firestore I'm getting a Image reference error 

Code to reproduce:

 db.collection("users").get().then((snap)=>{
   snap.forEach((doc)=>{
     console.log(doc.data())
   })
 })

simulator screen shot - iphone 6 - 2017-10-04 at 16 46 02

console outputs 
screen shot 2017-10-04 at 7 03 12 pm
when signing out of auth

mikelehen:
Josh, this should fix https://github.com/firebase/firebase-js-sdk/issues/183. Thoughts about including it in this week's release?

Also, I don't really understand why extra stuff in yarn.lock changed. Let me know if you think it's problematic.

### DIFF
--- a/packages/webchannel-wrapper/package.json
+++ b/packages/webchannel-wrapper/package.json
@@ -10,7 +10,7 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "closure-builder": "^2.2.23",
+    "closure-builder": "^2.2.29",
     "watch": "^1.0.2"
   },
   "repository": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -939,7 +939,7 @@ browserify@14.4.0:
     vm-browserify "~0.0.1"
     xtend "^4.0.0"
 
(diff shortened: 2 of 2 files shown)


jshcrowthe:
Sounds good! I'll stage the release for testing after merging this.

The `yarn.lock` is updated as it fixed the version of closure builder to the specific patch. It being updated here is to be expected.

mikelehen:
Yeah, I expected closure-builder to change in yarn.lock, but there were several other diffs too.  Perhaps those were dependencies of closure-builder or something though. I'm not sure.

PR #305, opened 2017-11-08
b/68783609: Ensure test cleanup runs after failed tests.

mikelehen:
Note that I couldn't actually reproduce the 'already open in another tab' error, I think because we create a new FirebaseApp instance (with a unique app name) for every test...  But I did confirm that cleanup now runs after failed tests, whereas it didn't before.

### DIFF
--- a/packages/firestore/test/integration/util/helpers.ts
+++ b/packages/firestore/test/integration/util/helpers.ts
@@ -191,18 +191,25 @@ export function withTestDbsSettings(
   }
 
   return Promise.all(promises).then((dbs: firestore.Firestore[]) => {
-    return fn(dbs)
-      .then(wipeDb.bind(null, dbs[0]), error => {
-        return wipeDb(dbs[0]).then(() => {
-          throw error;
-        });
-      })
-      .then(() => {
-        return dbs.reduce(
+    const cleanup = () => {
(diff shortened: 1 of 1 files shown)


packages/firestore/test/integration/util/helpers.ts
```diff
@@ -191,18 +191,25 @@ export function withTestDbsSettings(
...
-    return fn(dbs)
-      .then(wipeDb.bind(null, dbs[0]), error => {
-        return wipeDb(dbs[0]).then(() => {
-          throw error;
-        });
-      })
-      .then(() => {
-        return dbs.reduce(
+    const cleanup = () => {
+      return wipeDb(dbs[0]).then(() =>
```
schmidt-sebastian: Why don't we call wipeDb for every database as part of dbs.reduce? The extra overhead of calling a function that doesn't do anything should be fairly small :)

mikelehen: wipeDb is meant to delete the contents of the database.  Since all of the db objects are just separate Firestore instances that point to the same underlying database, there's no point in doing the wipe multiple times. The most correct thing would probably actually be to create a new Firestore instance (rather than use dbs[0] which may be in a broken / unknown state after the test runs), but given wipeDb() isn't implemented I don't care. :-)

PR #310, opened 2017-11-09
Firestore: in tests, fix SpecWatchFilter type to work on upcoming Typescript 2.7

### LINKED ISSUE #309: SpecWatchFilter type in Firestore tests will not compile with upcoming Typescript 2.7

  * Operating System version: n/a
  * Firebase SDK version: building from master
  * Firebase Product: firestore

Previously, the type SpecWatchFilter in spec_test_runner was specified as a tuple, but it was later used with `push`. It's actually an array with a first element that is guaranteed to be present, and of type TargetId[].

In the upcoming Typescript 2.7, [tuples will be fixed-length](https://github.com/Microsoft/TypeScript/pull/17765) and this usage will fail.  The fix is to change the definition of SpecWatchFilter to an interface that extends Array<TargetId[] | string> and whose required '0' property is a TargetId[]. I'll have a PR up shortly that does this.

#### Steps to reproduce:

Build with typescript@next.

#### Relevant Code:

```ts
// this will be an error in Typescript 2.7, because [targetIds].length === 1 
// but  [SpecFilterId[], string].length: 2
const filter: [SpecFilterId[], string] = [targetIds] as [SpecFilterId[], string];
```

sandersn:
Fixes #309

Previously, the type SpecWatchFilter in spec_test_runner was specified as a tuple, but it was later used with `push` and is actually an array with a first element that is guaranteed to be present, and of type TargetId[].

In Typescript 2.7, tuples will be fixed-length and this usage will fail.  This changes the definition of SpecWatchFilter to an interface that extends Array<TargetId[] | string> and whose required '0' property is a TargetId[].

### DIFF
--- a/packages/firestore/test/unit/specs/spec_test_runner.ts
+++ b/packages/firestore/test/unit/specs/spec_test_runner.ts
@@ -1190,7 +1190,10 @@ export interface SpecWatchEntity {
  * Note that the last parameter is really of type ...string (spread operator)
  * The filter is based of a list of keys to match in the existence filter
  */
-export type SpecWatchFilter = [TargetId[], string];
+export interface SpecWatchFilter extends Array<TargetId[] | string> {
+    '0': TargetId[];
+    '1'?: string;
+}
 
 /**
  * [field, op, value]
--- a/packages/firestore/test/unit/specs/spec_test_runner.ts
+++ b/packages/firestore/test/unit/specs/spec_test_runner.ts
@@ -1191,8 +1191,8 @@ export interface SpecWatchEntity {
  * The filter is based of a list of keys to match in the existence filter
  */
 export interface SpecWatchFilter extends Array<TargetId[] | string> {
-    '0': TargetId[];
-    '1'?: string;
+  '0': TargetId[];
+  '1'?: string;
 }
 
 /**


sandersn:
Looks like CI broke because of incorrect credentials. Not sure what is going on there.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1190,7 +1190,10 @@ export interface SpecWatchEntity {
  * Note that the last parameter is really of type ...string (spread operator)
  * The filter is based of a list of keys to match in the existence filter
  */
-export type SpecWatchFilter = [TargetId[], string];
+export interface SpecWatchFilter extends Array<TargetId[] | string> {
+  '0': TargetId[];
```
wilhuff: This is surprising that this works. Thanks for digging into it.

sandersn: I found it when I was testing stricter tuples for Typescript 2.7. The new definition is basically equivalent to the old meaning of 'tuple' in Typescript.

jshcrowthe:
Hey @sandersn, if you'll update this branch, you should pick up a change that will run the tests properly 👍.

Thanks for the ping @wilhuff

sandersn:
All right, looks like CI passed this time.

PR #313, opened 2017-11-09
Update changelog for upcoming release.

mikelehen:


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
+- [changed] Network streams are automatically closed after 60 seconds of
+  idleness.
+- [changed] We no longer log 'RPC failed' messages for expected failures.
 
 # v0.1.2
 - [fixed] Fixed validation of nested arrays to allow indirect nesting
```
schmidt-sebastian: - [changed] We now support field deletes for set() with `{merge:true}`.

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Unreleased
```
jshcrowthe: If we can add a `# v0.1.4` here that'd be great

mikelehen: done

packages/firestore/CHANGELOG.md
mikelehen: Added:
- [changed] We now support `FieldValue.delete()` sentinels in `set()` calls
  with `{merge:true}`.

PR #319, opened 2017-11-16
Revive Node.JS Support for Cloud Firestore (fixes #221).

### LINKED ISSUE #221: Add node.js support for Cloud Firestore (to "firebase" SDK)
Currently to use Cloud Firestore from node.js you can use the "firebase-admin" SDK (see node.js getting started snippets at https://firebase.google.com/docs/firestore/quickstart).

But if you want to use the "firebase" SDK instead (e.g. so you can use Firebase Authentication), you are currently out-of-luck. The "firebase" npm module will not work from node.js and you may get errors like:

```
(node:15795) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: k.setTimeout is not a function
```

This is a known issue and something we intend to address in the future.

mikelehen:
* Fix miscellaneous node / GRPC bit rot.
  * Re-add grpc / protobufjs dependencies.
  * Remove crazy grpc stream error handling nonsense that is no longer
    necessary after https://github.com/grpc/grpc/pull/9101
  * Clean up grpc_connection logging (and consistently use util.inspect()).
* Fix WebChannel / GRPC Connection compatibility issues.
  * Add an explicit mapping from "RPC name" (e.g. "BatchGetDocuments") to the
    REST url path (e.g. "batchGet") for WebChannel, and for GRPC just assume
    the first letter should be lowercased (e.g. "batchGetDocuments").
  * Split Connection.invoke() into invokeRPC() and invokeStreamingRPC(), with
    the latter accepting a stream of results and aggregating them into an
    array (needed to support BatchGetDocuments RPC).
* Fix serializer issues
  * Query limits are an 'Int32Value' but we were serializing them as a normal
    int which GRPC / protobufjs didn't like.
  * Several proto "oneof tags" were outdated.
* Add build steps to copy protos into npm package.
* Run integration tests for Node.JS
  * Added to 'test:node' script in package.json and in .vscode/launch.json
  * Include index.ts for browser and index.node.ts for node so the appropriate
    PlatformSupport gets registered.
* Misc cleanup
  * Remove unused MockPlatform since we use the normal NodePlatform now.
  * Remove 'google-auth-library' CredentialsProvider that we used to use for
    node.js (before we were integrated with FirebaseAuth).
  * Fixed several tests that were hitting node.js warnings about unhandled
    promise failures.
  * mocha commmand-line args:
    * "--compilers ts:ts-node/register" was deprecated in favor of
      "--require ts-node/register"
    * Remove "--retries 5" when running mocha tests from VS Code.
    * Consistently use "--require" instead of "-r"
    * Add "--exit" when running from VS Code.


packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -20,6 +20,11 @@ import { DatabaseId, DatabaseInfo } from '../../../src/core/database_info';
 import * as utilHelpers from '../util/helpers';
 
 describe('WebChannel', () => {
```
schmidt-sebastian: Can we do something like: `(typedef window | describe ? xdescribe)('WebChannel', ...`

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -148,35 +140,79 @@ export class GrpcConnection implements Connection {
     return this.cachedStub.stub;
   }
 
-  invoke(rpcName: string, request: any, token: Token | null): Promise<any> {
+  private getRpc(rpcName: string, token: Token | null): any {
     const stub = this.getStub(token);
+
+    // RPC Methods have the first character lower-cased
+    // (e.g. Listen => listen(), BatchGetDocuments => batchGetDocuments()).
+    const rpcMethod = rpcName.charAt(0).toLowerCase() + rpcName.slice(1);
```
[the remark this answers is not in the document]
schmidt-sebastian: Sorry, I wasn't more clear. I meant the mapping technique (a hardcoded map instead of this on-the-fly conversion).

packages/firestore/src/platform_node/load_protos.ts
```diff
@@ -14,48 +14,23 @@
...
+ * @returns The GrpcObject representing our protos.
  */
-export function loadProtos(
-  callback?: NodeCallback<ProtobufProtoBuilder>
-): ProtobufProtoBuilder | undefined {
-  const builder = protobufjs.newBuilder({
+export function loadProtos(): grpc.GrpcObject {
+  const options = {
     // Beware that converting fields to camel case does not convert the tag
     // fields in oneof groups (!!!).
```
schmidt-sebastian: Does this refer to valueType? ProtobufJS 6.x+ is now doing this.

packages/firestore/src/platform_node/grpc_connection.ts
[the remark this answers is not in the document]
mikelehen: I like the on-the-fly conversion since it means we don't have to hardcode all of the RPCs we use.  With WebChannel we can't avoid the hardcoding since the mapping isn't deterministic (whereas I /think/ it is for GRPC).

packages/firestore/src/platform_node/load_protos.ts
mikelehen: Interesting.  I tried to go down the route of depending directly on protobufjs again and upgrading to 6.x, but it did not seem straightforward (the entire way we're consuming protobufjs has changed an doesn't seem to be documented in the release notes, etc.).  So I'm punting on this for now. I did update the comment to mention that upgrading to protobufjs 6.x would resolve this though.

kylehotchkiss:
Hello! When will this be launched in the firebase package on NPM?

mikelehen:
Likely this Thursday.  You can keep an eye on https://firebase.google.com/support/release-notes/js to see when it goes out.

jmensch1:
hey thanks for the update! I think there's a problem though. I upgraded to 4.7.0 with npm, and now when I initialize firestore I'm getting the error 'firebase.firestore is not a function'

My imports are the same they were before:
```
import * as firebase from 'firebase'
import 'firebase/firestore'
```

And I initialize firestore by calling `firebase.firestore()`

jshcrowthe:
@jmensch1 feel free to [open an issue]() with more context (hopefully a sample repro) of the behavior you are seeing. I was unable to reproduce (just verified across a node build and a browser build).

PR #322, opened 2017-11-16
b/68251551: Send headers via URL Param to avoid CORS preflight round-trip.

mikelehen:
I verified that our Authorization header is sent via the the '$httpHeaders' URL parameter now, the OPTIONS network round-trip no longer occurs, and our authentication token is still correctly respected by the backend.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Unreleased
 
+# v0.1.4
+- [changed] Network streams are automatically closed after 60 seconds of
+  idleness.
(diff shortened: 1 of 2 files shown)


mikelehen:
Thanks @wenbozhu I added that in a comment (it's not convenient to use the actual constant from closure from here, and since it can't ever change, I don't think it's worth the trouble).

wilhuff:
Please add a note to the changelog as a part of this PR.

mikelehen:
Agh!  Good call, but I already pushed the merge button. :-(  I'll send another PR...

PR #326, opened 2017-11-20
Publish helper

jshcrowthe:
Some scripts to help automate the publishing of this project


scripts/prepublish
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
hiranya911: Suggest adding `.sh` extension to all bash scripts.

scripts/publish
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+# Get the current file directory
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Add our `npm bin` to the PATH
+export PATH="$(npm bin):$PATH";
+
+# Run the publish script in each directory
+lerna exec --scope @firebase/* --scope firebase -- $DIR/try-publish
```
hiranya911: Some of these arguments were not used in the old script. Hope that's ok.

scripts/prepublish
jshcrowthe: done

PR #327, opened 2017-11-20
Add Node.JS support to changelog.

mikelehen:
I forgot to include this in my previous PR...


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- [feature] Added Node.JS support for Cloud Firestore.
```
wilhuff: It's spelled "Node.js".

Also should we note that this support is limited (i.e. only with persistence disabled)?

mikelehen: Good points. Fixed.

PR #331, opened 2017-11-23
Add Query overload typing for `Transaction.get`

adamduren:


mikelehen:
Hi!  Thanks for the pull request, but the web/mobile Firestore SDKs don't support transaction.get(query) so I don't think this change is correct.  Unfortunately they cannot support get(query) as they use optimistic concurrency (optimistically assuming that none of the read documents will change, and then re-running the transaction if they do) rather than pessimistic locking (acquiring exclusive locks on the read documents to prevent them from changing) and there isn't currently a way to support reading queries while doing optimistic concurrency.

adamduren:
Thank you for the reply. The differences between what's possibly in the admin vs client apis can be confusing. I see now the admin api docs mention pessimistic locking and accept a query. It might worth be nothing on the client side docs that it uses optimistic locking and as such doesn't support fetching a query. Thanks again!

PR #345, opened 2017-12-01
 Add deleteToken call to FCM and add unit tests for iid-model.deleteToken

pinarx:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/messaging/src/controllers/controller-interface.ts
+++ b/packages/messaging/src/controllers/controller-interface.ts
@@ -113,14 +113,22 @@ export default class ControllerInterface {
   }
 
   /**
-   * This method deletes tokens that the token manager looks after and then
-   * unregisters the push subscription if it exists.
+   * This method deletes tokens that the token manager looks after,
+   * unsubscribes the token from FCM  and then unregisters the push
+   * subscription if it exists. It returns a promise that indicates
+   * whether or not the unsubscribe request was processed successfully.
    * @export
-   * @param {string} token
-   * @return {Promise<void>}
    */
-  deleteToken(token) {
-    return this.tokenDetailsModel_.deleteToken(token).then(() => {
+  deleteToken(token: string): Promise<Boolean> {
+    return this.tokenDetailsModel_.deleteToken(token)
+      .then(details => {
+        return this.iidModel_.deleteToken(
+          details['fcmSenderId'],
+          details['fcmToken'],
+          details['fcmPushSet']
+        );
+      })
+      .then(() => {
       return this.getSWRegistration_()
         .then(registration => {
           if (registration) {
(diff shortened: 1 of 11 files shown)


packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -113,26 +113,35 @@ export default class ControllerInterface {
...
-          if (subscription) {
-            return subscription.unsubscribe();
-          }
-        });
-    });
+  deleteToken(token: string): Promise<Boolean> {
+    return this.tokenDetailsModel_
+      .deleteToken(token)
+      .then(details => {
+        return this.iidModel_.deleteToken(
```
gauntface: Will the application server key be needed to delete this as well, or is that more a piece of meta data that goes along with this info?

packages/messaging/test/iid-model-get-token.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  let stubedFetch;
+  let globalIIDModel;
+
+  const cleanUp = () => {
+    sandbox.restore();
+    globalIIDModel = null;
+  };
+
+  before(function() {
+    stubedFetch = sinon.stub(window, 'fetch');
```
gauntface: Please move this into `beforeEach()` after cleanup and use the sandbox.stub().

Generally, if you stub out a method in the global scopes, its trivial to slip up and cause all future tests to fail because a stub isn't cleared up properly.

gauntface: I'm not a fan of global stubbing outside of the sandbox (too many examples where it doesn't get restored and it breaks)

packages/messaging/test/iid-model-get-token.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    return cleanUp();
+  });
+
+  it('should getToken on valid request', async function() {
+    globalIIDModel = new IIDModel();
+    const mockResponse = {
+      token: fcmToken,
+      pushSet: fcmPushSet
+    };
+    (window as any).fetch.returns(
```
gauntface: This is pretty hard to reason with.

Could you move the stubbing into the tests and then chain the .returns on that.

packages/messaging/src/controllers/controller-interface.ts
pinarx: nope, it shouldn't be needed - app pub key is treated the same way as the PushSubscription in this case.

packages/messaging/test/iid-model-get-token.test.ts
pinarx: done through next comment

packages/messaging/test/iid-model-get-token.test.ts
pinarx: done

gauntface:
cc @jshcrowthe (the test failing in Travis is firebase database test - just an FYI in case it becomes flakey)

PR #352, opened 2017-12-05
Firestore log offline warning

zxu123:
Firestore log offline warning

### Discussion
To avoid surprise by adding a warning. Right now, when offline, empty result is returned with no log nor error, see https://groups.google.com/forum/#!topic/google-cloud-firestore-discuss/puFl9HVU57I/discussion

### Testing
Tests pass. No new test required.

### API Changes
API review approved by all platform owners.

### DIFF
--- a/packages/firestore/src/remote/remote_store.ts
+++ b/packages/firestore/src/remote/remote_store.ts
@@ -185,6 +185,9 @@ export class RemoteStore {
     const didChange = this.watchStreamOnlineState !== onlineState;
     this.watchStreamOnlineState = onlineState;
     if (didChange) {
+      if (onlineState == OnlineState.Failed) {
+        log.debug(LOG_TAG, 'Could not reach Firestore backend');
+      }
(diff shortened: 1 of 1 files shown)


packages/firestore/src/remote/remote_store.ts
```diff
@@ -185,6 +185,9 @@ export class RemoteStore {
     const didChange = this.watchStreamOnlineState !== onlineState;
     this.watchStreamOnlineState = onlineState;
     if (didChange) {
+      if (onlineState == OnlineState.Failed) {
```
mikelehen: Per my comment on the Android version, I think we should only log this once, and only if we've never succeeded.

zxu123: Done. Added a variable to track the logic: shouldWarnOffline, which is initialized to true and changed to false if when ever online and when ever offline is logged.

zxu123:
Updated. PTAL.

PR #357, opened 2017-12-07
Add database invalidation every 7 days.

pinarx:
Add DB invalidation every 7 days, meaning:
- 7 days after token creation, if getToken is called, we will update the token and make an HTTP call to FCM backends. Token is likely to not change.
- Add a updateToken call from IIDModel, as the request and response from the API will be different than what's expected in getToken call.

@mmermerkaya for FYI+typescript


packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -61,7 +63,7 @@ export default class ControllerInterface {
    * @return {Promise<string> | Promise<null>} Returns a promise that
    * resolves to an FCM token.
    */
-  getToken() {
+  async getToken() {
```
gauntface: @jshcrowthe does Firebase polyfill async and await?

jshcrowthe: Since this is in a typescript file, the build process compiles down to ES5 which will result in the inclusion of the regenerator runtime.

Firebase itself doesn't have to polyfill these, but your runtime cost will go up.

pinarx: so i'm guessing the suggested way going forward is not to use async/await in the the source, but tests are ok?

gauntface: yes @pinarx that's right

pinarx:
@gauntface PTAL, couple of additions:
- Change async/await calls in the source to promise chains.
- Add integration tests for deleteToken

Next PR: Integration tests for updateToken

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -74,42 +73,94 @@ export default class ControllerInterface {
...
+        );
+      })
+      .catch(err => {
+        return this.deleteToken(tokenDetails['fcmToken']).then(() => {
+          throw err;
         });
-    });
+      })
+      .then(token => {
+        updatedToken = token;
```
gauntface: Does this mean that every 7 days the developer will have a new token for the user? I thought this would simply check if the token should be deleted and if so it would be reset at that point, otherwise it would continue being used.

I'm just wondering if this will grow a long list of tokens in databases that are now no longer relevant.

packages/messaging/src/models/iid-model.ts
```diff
@@ -123,19 +172,23 @@ export default class IIDModel {
...
-      const message = fcmTokenResponse['error']['message'];
-      throw this.errorFactory_.create(Errors.codes.TOKEN_UNSUBSCRIBE_FAILED, {
-        message: message
-      });
-    }
-
-    return;
+    ).then(fetchResponse => {
+      if (!fetchResponse.ok) {
+        return fetchResponse.json().then(fcmTokenResponse => {
```
gauntface: This should probably have a catch added to it - sometime FCM returns HTML and `.json()` will fail on that case.

Alternative would be to parse as text, then attempt parse as JSON and if the parsing to JSON throws, we can just return the text.

packages/messaging/src/controllers/controller-interface.ts
pinarx: currently the backend is wired to always return the existing token. So technically I can just do what you have suggested. What I had in mind:
- This gives us a way from the backend to recover clients if somehow the tokens issued from FCM are problematic, ie we can roll out new non-problematic tokens.
- I was assuming that since the tokenDetails like senderId, swScope etc didn't change we'd just be overwriting the existing entry in the DB. Is that not the case?

packages/messaging/src/models/iid-model.ts
pinarx: added error handling for all such cases in iid-model + test coverage.

packages/messaging/src/controllers/controller-interface.ts
gauntface: It'll overwrite any of the details based on the service worker scope - which won't change :)

PR #358, opened 2017-12-07
b/68276665: Raise isFromCache=true events when offline

mikelehen:
* Plumbs OnlineState changes through to views.
* View sets this.current to false on OnlineState.Failed, triggering
  isFromCache=true events. this.current will automatically be returned to true
  once the listen is reestablished and we get a new CURRENT message.
* Updated tests (and added new ones) to verify behavior.

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -350,13 +350,13 @@ export class SyncEngine implements RemoteSyncer {
   applyOnlineStateChange(onlineState: OnlineState) {
     const newViewSnapshots = [] as ViewSnapshot[];
     this.queryViewsByQuery.forEach((_, queryView) => {
-      const viewSnap = queryView.view.applyOnlineStateChange(onlineState);
+      const viewChange = queryView.view.applyOnlineStateChange(onlineState);
       assert(
-        viewSnap.limboChanges.length === 0,
+        viewChange.limboChanges.length === 0,
         'OnlineState should not affect limbo documents.'
       );
-      if (viewSnap.snapshot) {
-        newViewSnapshots.push(viewSnap.snapshot);
+      if (viewChange.snapshot) {
+        newViewSnapshots.push(viewChange.snapshot);
       }
     });
     this.viewHandler(newViewSnapshots);
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -42,7 +42,7 @@ import { Query } from './query';
 import { SnapshotVersion } from './snapshot_version';
 import { TargetIdGenerator } from './target_id_generator';
 import { Transaction } from './transaction';
-import { BatchId, ProtoByteString, TargetId } from './types';
+import { BatchId, ProtoByteString, TargetId, OnlineState } from './types';
 import {
   AddedLimboDocument,
   LimboDocumentChange,
@@ -341,6 +341,25 @@ export class SyncEngine implements RemoteSyncer {
     return this.localStore.applyRemoteEvent(remoteEvent).then(changes => {
       return this.emitNewSnapsAndNotifyLocalStore(changes, remoteEvent);
     });
+  }
+
+  /**
+   * Applies an OnlineState change to the sync engine and notifies any views of
+   * the change.
+   */
+  applyOnlineStateChange(onlineState: OnlineState) {
+    const newViewSnapshots = [] as ViewSnapshot[];
+    this.queryViewsByQuery.forEach((_, queryView) => {
+      const viewSnap = queryView.view.applyOnlineStateChange(onlineState);
+      assert(
+        viewSnap.limboChanges.length === 0,
+        'OnlineState should not affect limbo documents.'
+      );
+      if (viewSnap.snapshot) {
+        newViewSnapshots.push(viewSnap.snapshot);
+      }
+    });
+    this.viewHandler(newViewSnapshots);
   }
 
   rejectListen(targetId: TargetId, err: FirestoreError): Promise<void> {
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -42,7 +42,7 @@ import { Query } from './query';
 import { SnapshotVersion } from './snapshot_version';
 import { TargetIdGenerator } from './target_id_generator';
 import { Transaction } from './transaction';
-import { BatchId, ProtoByteString, TargetId, OnlineState } from './types';
+import { BatchId, OnlineState, ProtoByteString, TargetId } from './types';
 import {
   AddedLimboDocument,
   LimboDocumentChange,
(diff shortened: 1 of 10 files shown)


packages/firestore/src/remote/remote_store.ts
```diff
@@ -228,15 +228,26 @@ export class RemoteStore {
...
-    this.updateAndBroadcastOnlineState(OnlineState.Failed);
+  /**
+   * Temporarily disables the network. The network can be re-enabled using
+   * enableNetwork(). By default this will also update (and broadcast) our
+   * onlineState to OnlineState.Failed. If this should not happen (e.g. because
+   * this is being called as part of shutdown logic or because we're going to
+   * immediately re-enable the network), pass `true` for
+   * suppressOnlineStateFailure.
+   */
+  disableNetwork(suppressOnlineStateFailure?: boolean): Promise<void> {
```
schmidt-sebastian: `suppressOnlineStateNotification` ?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -343,6 +343,25 @@ export class SyncEngine implements RemoteSyncer {
     });
   }
 
+  /**
+   * Applies an OnlineState change to the sync engine and notifies any views of
+   * the change.
+   */
+  applyOnlineStateChange(onlineState: OnlineState) {
+    const newViewSnapshots = [] as ViewSnapshot[];
+    this.queryViewsByQuery.forEach((_, queryView) => {
```
schmidt-sebastian: I feel like we should keep track of the online state somewhere (and this would be a good place). Then we could either assert that is has indeed changed or only raise events if it did.

packages/firestore/src/remote/remote_store.ts
mikelehen: We only suppress failure (if you pass false, we still trigger the Unknown state).

packages/firestore/src/core/sync_engine.ts
mikelehen: Can you elaborate?  I'm not sure what the benefit of tracking the onlineState here is just to assert that it changes.  This code doesn't actually rely on it having changed and I'd rather not add unnecessary state (as a general rule).

FWIW- We already track the current onlineState in RemoteStore, EventManager, and QueryListener, which is too many places in my opinion. :-)

schmidt-sebastian: I just want to make sure that subsequent calls to disableNetwork() only trigger one snapshot event.

mikelehen: I think we're pretty safe since we are guaranteed to only raise an OnlineState event if it actually changed. https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/remote_store.ts#L195

And the View code I added is also robust against duplicate OnlineState.Failed events since this.current will already be false here: https://github.com/firebase/firebase-js-sdk/pull/358/files#diff-59002bb6a4e41ae4ef48ff413744e9d3R277

schmidt-sebastian: Good point. Guess we won't be adding a third layer of protection then. LGTM :)

packages/firestore/src/core/sync_engine.ts
```diff
@@ -42,7 +42,7 @@ import { Query } from './query';
 import { SnapshotVersion } from './snapshot_version';
 import { TargetIdGenerator } from './target_id_generator';
 import { Transaction } from './transaction';
-import { BatchId, ProtoByteString, TargetId } from './types';
+import { BatchId, ProtoByteString, TargetId, OnlineState } from './types';
```
wilhuff: Did you `prettier` this? (I'm surprised OnlineState sorts after ProtoByteString.)

packages/firestore/src/core/sync_engine.ts
wilhuff: Google's typescript style disallows `_` as an identifier. I don't know if we're still holding ourselves to it, but for better or worse it's a way to pre-settle arguments about these kinds of low-level things. The suggestion on typescript-users was to name unused parameters with their regular name, or if the body is long enough to call out that they're unused with an "unused" prefix or suffix.

packages/firestore/src/core/view.ts
```diff
@@ -37,6 +37,7 @@ import {
   SyncState,
   ViewSnapshot
 } from './view_snapshot';
+import { OnlineState } from './types';
```
wilhuff: Does prettier not sort imports (by location in `from`)?

packages/firestore/src/remote/remote_store.ts
wilhuff: Is there any reason why any external caller would ever want to pass `true` for this? Is there any reason to allow it?

If not, consider rearranging this to keep the API as narrow as possible, i.e. by making a private method that takes the intended `OnlineState` as a parameter. This boolean in the public surface is a bit of a wart.

packages/firestore/src/core/sync_engine.ts
mikelehen: Yes, I did, but it sucks at imports I guess. I fixed it manually, but going forward we should probably just not worry about it if it isn't automated. :-/

packages/firestore/src/core/sync_engine.ts
mikelehen: Oh! I didn't realize that. Fixed, thanks.

packages/firestore/src/core/view.ts
mikelehen: manually fixed.

packages/firestore/src/remote/remote_store.ts
mikelehen: I reworked this per your comment here and your out-of-band feedback, see https://github.com/firebase/firebase-js-sdk/pull/358/commits/a9e62e34fe730664869a7a402d76b347a2695f46

Let me know what you think.

packages/firestore/src/core/sync_engine.ts
wilhuff: Yuck.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -228,26 +233,29 @@ export class RemoteStore {
...
+  }
+
+  /**
+   * Disables the network, setting the OnlineState to the specified
+   * targetOnlineState.
+   */
+  private disableNetworkInternal(
+    targetOnlineState: OnlineState
+  ): Promise<void> {
+    this.updateOnlineState(targetOnlineState);
```
wilhuff: This makes me uncomfortable because the online state handler gets called essentially before any of the actual work to disable the network has completed. While I don't think it currently does this, that handler could call back into the remote store and it would be inconsistent.

My advice would be do to make this update last.

Alternatively (but less attractively) force the update onto the async queue so that it can't run until this function returns.

Note that if we ever add a second state that can broadcast this pattern of update and broadcast if necessary stops working. It could also be worth biting the bullet and splitting the state change from the notification. I'm not sure this is really needed though and YAGNI so probably not worth it.

packages/firestore/src/remote/remote_store.ts
wilhuff: Looks much better to me. Thank you for taking the time.

packages/firestore/src/remote/remote_store.ts
mikelehen: Good point.  Moved the update to be last.

PR #362, opened 2017-12-09
Encode strings as UTF-8 before/after converting to base64

ryanpbrewster:
SDK behavior when reconnecting changed between v4.5.0 and v4.5.1. This is currently causing incorrect data to be saved whenever a user:
 - disconnects
 - tries to perform a write while disconnected
 - reconnects

During the reconnect phase, it looks like we're pushing the queued writes by base64-encoding the data and making a long-polling request. Our base64 encoding is invalid for multi-byte characters (see https://github.com/google/closure-library/issues/78).

This PR changes the byte array conversions to use a UTF-8 encoding, which fixes the issue.

I used the utf8 conversion functions from here:
https://github.com/google/closure-library/blob/e877b1eac410c0d842bcda118689759512e0e26f/closure/goog/crypt/crypt.js#L115


ryanpbrewster:
I do not know how or why the behavior changed from v4.5.0 to v4.5.1. The SDK code looks the same to me, but I have a repro and bisected the change to there. Happy to discuss further.

packages/util/test/base64.test.ts
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { assert } from 'chai';
+import { base64Encode, base64Decode } from '../src/crypt';
+
+describe('base64', () => {
+  it('bijective', () => {
+    let str = '…';
+    assert.strictEqual(base64Decode(base64Encode(str)), str);
```
rockwotj: Can we test each of the variants? So a 2-byte UTF-8 char, a 3-byte char and a 4 byte char? (ascii too).

ryanpbrewster:
My understanding is that `stringToByteArray` and `byteArrayToString` are doing UTF-8 encoding/decoding respectively.

rockwotj:
You're right, nvm

packages/util/test/base64.test.ts
ryanpbrewster: Added cases for 1--4 bytes

PR #364, opened 2017-12-11
Making DocumentSnaphot.data() nullable

schmidt-sebastian:
This changes DocumentSnaphot.data() to be nullable and introduce a non-nullable QueryDocumentSnapshot.

And by nullable I mean undefined.


packages/firestore/src/api/database.ts
```diff
@@ -1079,6 +1072,24 @@ export class DocumentSnapshot implements firestore.DocumentSnapshot {
...
+    key: DocumentKey,
+    document: Document,
+    fromCache: boolean
+  ) {
+    super(firestore, key, document, fromCache);
+  }
+
+  data(): firestore.DocumentData {
+    const data = super.data();
+    assert(!!data, 'Document in a QueryDocumentSnapshot should exist');
```
mikelehen: nit: why not assert typeof data === 'object' ?  That's more precise and explicit than doing the truthy-cast-to-boolean thing.

packages/firestore/src/typings/firestore.d.ts
```diff
@@ -544,9 +544,13 @@ declare namespace firestore {
    * A `DocumentSnapshot` contains data read from a document in your Firestore
    * database. The data can be extracted with `.data()` or `.get(<field>)` to
    * get a specific field.
+   *
+   * For a `DocumentSnapshot` that points to a non-existing document, any data
+   * access will return `undefined`. You can use the `exists` property to
+   * explicitly verify a documents existence.
```
mikelehen: "documents existence" => "document's existence" or just simply "document exists"

packages/firestore/src/typings/firestore.d.ts
```diff
@@ -563,14 +567,17 @@ declare namespace firestore {
     readonly metadata: SnapshotMetadata;
 
     /**
-     * Retrieves all fields in the document as an Object.
+     * Retrieves all fields in the document as an Object. Returns `undefined` if
+     * the document doesn't exist
```
mikelehen: End sentence with period?

packages/firestore/src/typings/firestore.d.ts
```diff
@@ -544,9 +544,13 @@ declare namespace firestore {
    * A `DocumentSnapshot` contains data read from a document in your Firestore
    * database. The data can be extracted with `.data()` or `.get(<field>)` to
    * get a specific field.
+   *
```
mikelehen: So in firestore-api-changes, I believe our .d.ts types are repeated in 3 places for good measure (packages/firebase/app/index.d.ts, packages/firebase/index.d.ts, packages/firestore/src/typings/firestore.d.ts), but Josh just merged his typings refactor into master which unifies them to a single place (packages/firestore-types/index.d.ts).  So we probably should merge master into firsetore-api-changes and then rework your change on that.

packages/firestore/src/typings/firestore.d.ts
```diff
@@ -563,14 +567,17 @@ declare namespace firestore {
...
+     * @return An Object containing all fields in the document or `undefined` if
+     * the document doesn't exist.
      */
-    data(): DocumentData;
+    data(): DocumentData | undefined;
 
     /**
-     * Retrieves the field specified by `fieldPath`.
+     * Retrieves the field specified by `fieldPath`. Returns `undefined` if the
+     * document or field doesn't exist
```
mikelehen: End sentence with period?

packages/firestore/src/typings/firestore.d.ts
```diff
@@ -579,6 +586,29 @@ declare namespace firestore {
...
   }
 
+  /**
+   * A `QueryDocumentSnapshot` contains data read from a document in your
+   * Firestore database as part of a query. The document is guaranteed to exist
+   * and its data can be extracted with `.data()` or `.get(<field>)` to get a
+   * specific field.
+   *
+   * A `QueryDocumentSnapshot` offers the same API surface as a
+   * `DocumentSnapshot`. As deleted documents are not returned from queries,
```
mikelehen: I think "deleted documents are not returned from queries" may invite unnecessary confusion.  Perhaps "Since query results are all existing documents, ..." or even just omit the justification and say, "offers the same API surface as a `DocumentSnapshot` but its exists property ..."

packages/firestore/src/api/database.ts
```diff
@@ -990,31 +990,24 @@ export class DocumentSnapshot implements firestore.DocumentSnapshot {
     private _fromCache: boolean
   ) {}
 
-  data(): firestore.DocumentData {
+  data(): firestore.DocumentData | undefined {
```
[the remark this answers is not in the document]
schmidt-sebastian: I basically picked `undefined` based on a two step process:
- I threw a coin to decide between `undefined` and `null`.
- I then disregarded the coin toss and went with `undefined` since it is the only solution that works with our `get()` call. I wouldn't otherwise know what to return from get() for non-existing documents.

I'm glad you came to the same conclusion.

packages/firestore/src/api/database.ts
schmidt-sebastian: Fixed.

packages/firestore/src/typings/firestore.d.ts
schmidt-sebastian: I merged Josh's PR and changes this in just one place. Note that we will have to update firestore-externs.js when we decide to merge this branch into master.

packages/firestore/src/typings/firestore.d.ts
schmidt-sebastian: Done

packages/firestore/src/typings/firestore.d.ts
schmidt-sebastian: Done

packages/firestore/src/typings/firestore.d.ts
schmidt-sebastian: Done

packages/firestore/src/typings/firestore.d.ts
schmidt-sebastian: Fixed with the same texts that I used in Android.

PR #366, opened 2017-12-11
Refactor @firebase/admin entrypoint 

jshcrowthe:
Refactored the `@firebase/database` entrypoint for the firebase-admin-node SDK


packages/database/admin.js
```diff
@@ -0,0 +1,52 @@
...
+exports.initStandalone = function(app, url, version) {
+  const instance = RepoManager.getInstance().databaseFromApp(app, url);
+
+  if (version) {
+    /**
+     * We are patching the version info in @firebase/app as
+     * the database SDK sends this version back to our backend
+     */
+    firebaseApp.SDK_VERSION = version;
+    CONSTANTS.NODE_ADMIN = true;
```
hiranya911: Perhaps `NODE_ADMIN` should be set outside the `if` block?

jshcrowthe: Yes it should. Great catch.

Issue #114, opened 2017-07-27
Wrong error fired on useServiceWorker() called second time 


zezic:
* Operating System version: Ubuntu 16.04 (don't matter)
  * Firebase SDK version: 4.1.5
  * Firebase Product: messaging

#### Steps to reproduce:

Call `firebase.messaging().useServiceWorker(sw_registration)` second time and get wrong error message: `USE_SW_BEFORE_GET_TOKEN`
There should be some warning about repeated assigning of Service Worker because we don't try to get token here, only assign Service Worker.

#### Relevant Code:

```javascript
let firebase = require('firebase/app')
require('firebase/messaging')
firebase.initializeApp({
  messagingSenderId: FCM_SENDER_ID
})
firebase.messaging().useServiceWorker(sw_registration)
firebase.messaging().useServiceWorker(sw_registration) // wrong error here
```

gauntface:
@zezic thanks for raising the issue, will look into this when I can.

rejhgadellaa:
FYI, looks like it's also present in 4.2.0

gauntface:
The error message does need changing. I hadn't intended that developers would call this method twice, out of interest what is the use case?

The reason for enforcing a single call is to avoid any risk of the library being called with two different service workers.

So would people prefer the ability to just allow any number of calls and block after getToken is called or would they prefer only being able to call it once and just have the error corrected to something along the lines of "useServiceWorker can only be called once and must be called before getToken()" ?

PR #368, opened 2017-12-12
Fix  issue #114  useServiceWorker using registration param

tenosiswono:
Fix checking undefined for registration parameter instead of this.registrationToUse_ 
Fix issue #114


gauntface:
This isn't the correct fix for that issue - will add explanation to the bug.

PR #374, opened 2017-12-12
Firestore Console Build

jshcrowthe:
This creates a build target for the Firestore console

### DIFF
--- a/packages/firestore/console.ts
+++ b/packages/firestore/console.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 import './src/platform_browser/browser_init';
 
 export {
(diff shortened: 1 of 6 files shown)


packages/firestore/console.ts
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import './src/platform_browser/browser_init';
+
+export {
+  Firestore,
+  FirestoreDatabase,
+  PublicCollectionReference as CollectionReference,
```
mikelehen: Hrm.  Could we export CollectionReference from firestore-types instead (same for the below types)? It seems like we should only need to export the actual classes for things with constructors / static methods.  In general I'd prefer stick to exporting curated types rather than the raw implementations (as per my usual reasons), but since the console is internal I'm a little more willing to play fast-and-loose.

packages/firestore/gulpfile.js
```diff
@@ -36,6 +39,40 @@ const setupWatcher = () => {
...
+ * behind its existence
+ */
+function buildConsole() {
+  return gulp
+    .src('console.ts')
+    .pipe(gulpWebpack(require('./webpack.config'), webpack))
+    .pipe(
+      through.obj(function(file, encoding, cb) {
+        if (file.isBuffer()) {
+          file.contents = Buffer.concat([
```
tstirrat: Turns out the export name is super important. With any other name, jscompiler renames several important things (when in super-optimized mode).

I suspect we are pulling in this TS definition through some transitive dependency, and it prevents renaming:

```ts
declare let __firestore_exports__: {[key: string]: {}};
```

The following wrapper will get it all working, until we can gradually remove some of the incredible layers of jank 🤠:

```js
goog.module('firestore');
exports =
    (function() {
      var sdk_exports = eval(
          'var __firestore_exports__ = { firestore: {}};' +
          '__firestore_exports__.firestore=**CONTENT_HERE**\n' + 
          '__firestore_exports__');
      return sdk_exports.firestore;
    }).call(window);
```

PR #375, opened 2017-12-13
Always log resource exhausted errors to console ...

rsgowman:
... when receiving webchannel messages. Even without debug mode one
(i.e. even when the caller hasn't called
firebase.firestore.setLogLevel('debug')).

Without this, the developer would have no idea why firestore was not
returning any data, as it would just silently not work.

b/69638501

### DIFF
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -301,19 +301,10 @@ export class WebChannelConnection implements Connection {
         // can be removed once the fix has been rolled out.
         const error = msgData.error || (msgData[0] && msgData[0].error);
         if (error) {
+          log.debug(LOG_TAG, 'WebChannel received error:', error);
           // error.status will be a string like 'OK' or 'NOT_FOUND'.
           const status: string = error.status;
           let code = mapCodeFromRpcStatus(status);
-          if (code === Code.RESOURCE_EXHAUSTED) {
-            log.error(
-              LOG_TAG,
-              'WebChannel received resource exhausted error:',
-              error
-            );
-          } else {
-            log.debug(LOG_TAG, 'WebChannel received error:', error);
-          }
-
           let message = error.message;
           if (code === undefined) {
             code = Code.INTERNAL;
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -301,10 +301,15 @@ export class WebChannelConnection implements Connection {
         // can be removed once the fix has been rolled out.
         const error = msgData.error || (msgData[0] && msgData[0].error);
         if (error) {
-          log.debug(LOG_TAG, 'WebChannel received error:', error);
           // error.status will be a string like 'OK' or 'NOT_FOUND'.
           const status: string = error.status;
           let code = mapCodeFromRpcStatus(status);
+          if (code === Code.RESOURCE_EXHAUSTED) {
+            log.error(LOG_TAG, 'WebChannel received resource exhausted error:', error);
+          } else {
+            log.debug(LOG_TAG, 'WebChannel received error:', error);
+          }
+
           let message = error.message;
           if (code === undefined) {
             code = Code.INTERNAL;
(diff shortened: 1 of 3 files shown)


packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -301,10 +301,19 @@ export class WebChannelConnection implements Connection {
         // can be removed once the fix has been rolled out.
         const error = msgData.error || (msgData[0] && msgData[0].error);
         if (error) {
-          log.debug(LOG_TAG, 'WebChannel received error:', error);
           // error.status will be a string like 'OK' or 'NOT_FOUND'.
           const status: string = error.status;
           let code = mapCodeFromRpcStatus(status);
+          if (code === Code.RESOURCE_EXHAUSTED) {
```
rsgowman: @mikelehen: or should we just log.error this regardless (excepting Code.OK)?

mikelehen: Some errors are expected (e.g. when a transaction fails due to concurrent writes, we get a FAILED_PRECONDITION error or similar) so we don't want to log all errors.

Also, we have multiple network transports (webchannel_connection.ts for the browser, grpc_connection.ts for node).  So it may make sense to tweak the log message here instead: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/persistent_stream.ts#L321  (that code is used by both our write stream and watch stream, regardless of transport)

rsgowman: Done.

Completely unrelated to this change, I just noticed that log.debug() takes a tag parameter, but log.error() does not. Is that on purpose? (A quick grep through the code base suggests we always supply a tag when using log.debug. However, we seem to also supply a tag when using log.error() roughly a third of the time.) Given that log.error() takes a va_list anyway, it'll obviously still work; it just seems a little odd.

mikelehen: Hah...  Uh, I'm not 100% sure. My guess is the goal with the tag is to be able to theoretically enable/disable the debug logs on a per-tag basis (this isn't actually possible currently).  But error messages are always-on, much fewer, and are more user-facing, so the tag isn't needed.  We should probably clean up the usages of log.error() that include a tag in that case though.  Perhaps open a P3 / internal cleanup bug (or just make a PR :-)) if you feel like it.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -318,8 +318,9 @@ export abstract class PersistentStream<
       // If this is an intentional close ensure we don't delay our next connection attempt.
       this.backoff.reset();
     } else if (error && error.code === Code.RESOURCE_EXHAUSTED) {
-      log.debug(
-        LOG_TAG,
+      // Log the error. (Probably either 'quota exceeded' or 'max queue length reached'.)
```
mikelehen: Oh. Good call logging the actual error (and commenting the justification). Thanks!  Were you able to verify that this logs sanely? I think our logger just outputs the JSON representation for objects, which may be wonky. It may be better to do error.toString() ?

rsgowman: Yeah, it looks reasonable to me:

```
Firestore (4.8.0) 2017-12-13T16:41:36.587Z: FirebaseError: [code=resource-exhausted]: Stream reached its maximum queuing limit, and aborted.
Firestore (4.8.0) 2017-12-13T16:41:36.590Z: Using maximum backoff delay to prevent overloading the backend.
```

mikelehen: 👍

rsgowman: Turns out that .toString() is needed anyways due to the TS tests. Logs still look the same.

packages/firestore/src/platform_browser/webchannel_connection.ts
rsgowman: https://github.com/firebase/firebase-js-sdk/pull/376

rsgowman:
@jshcrowthe: prettier changed packages/util/test/base64.test.ts, so I think I need your lgtm.

PR #378, opened 2017-12-13
Expose network management

gsoltis:
Add enableNetwork() and disableNetwork() to the public interface for FirebaseFirestore.

b/70348588


mikelehen:
If this is ready for review, can you set the "assignee" ? Else, either nobody will review it or Gil, Sebastian, and I will all review it. :p

packages/firestore-types/index.d.ts
```diff
@@ -123,6 +123,22 @@ export class FirebaseFirestore {
    */
   app: any;
 
+  /**
+   * Re-enables use of the network for this Firestore instance after a prior
+   * call to disableNetwork().
+   * @return A promise that is resolved once the network has been enabled.
```
mikelehen: Can you add a newline between the description and the jsdoc annotations for consistency with the rest of this file?

[and also sync this with whatever we end up with for Android]

packages/firestore/src/remote/remote_store.ts
```diff
@@ -242,6 +237,9 @@ export class RemoteStore {
    * enableNetwork().
    */
   disableNetwork(): Promise<void> {
+    if (!this.isNetworkEnabled()) {
+      return Promise.resolve();
+    }
```
mikelehen: [The beauty of multiple ports is you get to think about the change multiple times and catch bugs in your 2nd or 3rd review... :-)]

Currently disableNetworkInternal() is not idempotent (in particular, it'll crash on subsequent calls). And we have 2 places we call disableNetworkInternal() besides here, and I think both are problematic.  shutdown() should not call disableNetworkInternal() if it's already disabled, and handleUserChange() should not call disableNetworkInternal() /or/ enableNetwork() if it's currently disabled.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -669,4 +665,17 @@ apiDescribe('Database', persistence => {
         .then(() => awaitOnlineSnapshot());
     });
   });
+
+  it('can enable and disable networking', () => {
```
mikelehen: I don't think I saw this test in the Android version.  Can you either add it there, or remove it here (I don't care strongly which)?

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -435,10 +435,6 @@ apiDescribe('Queries', persistence => {
 
   it('can query while reconnecting to network', () => {
     return withTestCollection(persistence, /* docs= */ {}, coll => {
-      // TODO(mikelehen): Find better way to expose this to tests.
-      // tslint:disable-next-line:no-any enableNetwork isn't exposed via d.ts
-      const firestoreInternal = coll.firestore.INTERNAL as any;
```
mikelehen: There's another one of these in the next test (which you picked up when you rebased on firestore-api-changes... sorry :-))

gsoltis: Done

packages/firestore-types/index.d.ts
gsoltis: Spaces added, will defer to android review before updating comment.

packages/firestore/src/remote/remote_store.ts
gsoltis: Ok, I think I've addressed this now. In one case I've added an assert, which I think is safe, but take a look.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -814,10 +818,20 @@ export class RemoteStore {
...
+      this.disableNetworkInternal(OnlineState.Unknown);
+      return this.enableNetwork();
+    } else {
+      // The network is disabled, so we don't want to reconnect. Remain in
+      // 'failed' state. This assert *should* only ever trigger if shutdown() is
+      // called without first removing the user change listener.
+      assert(
+        this.watchStreamOnlineState == OnlineState.Failed,
+        'Since the network is disabled, expected to be in failed state'
+      );
```
mikelehen: I think this assert should hold, but it feels a bit out-of-place.  I don't think it's necessary for this method to function correctly, so it's kind of a "By the way, this should be true..." assert.  We could consider verifying that methods aren't called after shutdown(), but to do that I think we should add an `assertNotShutdown()` method and call it from every method... though I'm not sure that's necessary (and if we do it, we should do it across all of our shutdown-able components).

packages/firestore/test/integration/api/database.test.ts
gsoltis: Yes, I haven't done that yet. I will address other clients once I get this one in an acceptable state.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -268,10 +266,16 @@ export class RemoteStore {
...
-    return Promise.resolve(undefined);
+    if (this.isNetworkEnabled()) {
+      // Set the OnlineState to Unknown (rather than Failed) to avoid potentially
+      // triggering spurious listener events with cached data, etc.
+      this.disableNetworkInternal(OnlineState.Unknown);
+    } else {
+      // The network is disabled, but we also no longer want events, so
+      // transition to 'Unknown' state.
+      this.updateOnlineState(OnlineState.Unknown);
+    }
```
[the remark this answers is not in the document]
gsoltis: I started looking into this and discovered that updateOnlineState is only *mostly* idempotent. So, I think fixing that, and then making disableNetworkInternal idempotent could work.

packages/firestore/src/remote/remote_store.ts
gsoltis: I could replace it with a now-idempotent call to disableInternal w/ state == failed, or leave it out entirely. I think it's at least worth keeping a comment there, since it might not be immediately obvious why we're only doing something if the network is enabled.

packages/firestore/src/remote/remote_store.ts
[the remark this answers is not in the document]
mikelehen: Maybe just change disableNetworkInternal() to not call updateOnlineState() and instead call updateOnlineState() explicitly in the places we call disableNetworkInternal()?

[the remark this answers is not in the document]
gsoltis: I like that better.

packages/firestore/src/remote/remote_store.ts
mikelehen: FWIW- I would just do:

```
// Ensure we don't start the network if it's currently disabled.
if (this.isNetworkEnabled()) {
   this.disableNetworkInternal(OnlineState.Unknown);
   return this.enableNetwork();
} else {
  return Promise.resolve();
}
```

packages/firestore/src/remote/remote_store.ts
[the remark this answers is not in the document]
gsoltis: This is done.

packages/firestore/src/remote/remote_store.ts
gsoltis: Ok, this is done now.

gsoltis:
Ok, I think the main sticking points are now updating the public-facing comments to be in line with whatever we decide on the java CL, and porting the enable/disable test to other SDKs.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -155,25 +155,26 @@ export class RemoteStore {
...
-      // The state is set to unknown when a healthy stream is closed (e.g. due to
-      // a token timeout) or when we have no active listens and therefore there's
-      // no need to start the stream. Assuming there is (possibly in the future)
-      // an active listen, then we will eventually move to state Online or Failed,
-      // but we always want to make at least ONLINE_ATTEMPTS_BEFORE_FAILURE
-      // attempts before failing, so we reset the count here.
-      this.watchStreamFailures = 0;
     }
 
     // Update and broadcast the new state.
```
mikelehen: Can you move this comment closer to the actual assignment / broadcast?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -155,25 +155,26 @@ export class RemoteStore {
 
   /**
    * Updates our OnlineState to the new state, updating local state
-   * and notifying the onlineStateHandler as appropriate.
+   * and notifying the onlineStateHandler as appropriate. Idempotent.
    */
-  private updateOnlineState(newState: OnlineState) {
+  private updateOnlineState(newState: OnlineState): void {
     if (newState === OnlineState.Healthy) {
```
mikelehen: FWIW, it is safe (and makes the "Idempotent." part of the method comment more correct) to move this inside the `if (newState !== this.watchStreamOnlineState) { ... }` check, so that would be my preference.  

And at that point, if you felt like it, you could flip the condition and make it an early return if you felt like it (note though that we generally avoid early return except for these sorts of "there's nothing to do, so just return immediately" cases).

packages/firestore/src/remote/remote_store.ts
```diff
@@ -242,36 +238,38 @@ export class RemoteStore {
...
     // Set the OnlineState to failed so get()'s return from cache, etc.
-    this.disableNetworkInternal(OnlineState.Failed);
+    this.updateOnlineState(OnlineState.Failed);
     return Promise.resolve();
   }
 
   /**
    * Disables the network, setting the OnlineState to the specified
-   * targetOnlineState.
+   * targetOnlineState. Idempotent.
```
mikelehen: Comment needs updating.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -814,10 +812,15 @@ export class RemoteStore {
   handleUserChange(user: User): Promise<void> {
     log.debug(LOG_TAG, 'RemoteStore changing users: uid=', user.uid);
 
-    // Tear down and re-create our network streams. This will ensure we get a fresh auth token
-    // for the new user and re-fill the write pipeline with new mutations from the LocalStore
-    // (since mutations are per-user).
-    this.disableNetworkInternal(OnlineState.Unknown);
-    return this.enableNetwork();
+    // If the network has been explicitly disabled, make sure we don't
+    // accidentally re-enabled it.
```
mikelehen: "re-enabled" => "re-enable" (sorry if I had that typo in my original suggestion)

packages/firestore/src/remote/remote_store.ts
gsoltis: Done.

packages/firestore/src/remote/remote_store.ts
gsoltis: Done.

packages/firestore/src/remote/remote_store.ts
gsoltis: Ended up dropping this comment. I think it was just describing what the two lines were doing, and not why and therefore is not super helpful.

packages/firestore/src/remote/remote_store.ts
gsoltis: Moved it inside.

PR #379, opened 2017-12-13
Make offline warning an `error` instead of `debug`.

zxu123:
Make offline warning an `error` instead of `debug`. JS client SDK does not have `warning`. To make warning default behavior, we actually want `error` here instead of `debug`.

### DIFF
--- a/packages/firestore/src/remote/remote_store.ts
+++ b/packages/firestore/src/remote/remote_store.ts
@@ -192,7 +192,7 @@ export class RemoteStore {
       this.watchStreamFailures++;
       if (this.watchStreamFailures >= ONLINE_ATTEMPTS_BEFORE_FAILURE) {
         if (this.shouldWarnOffline) {
-          log.error(LOG_TAG, 'Could not reach Firestore backend.');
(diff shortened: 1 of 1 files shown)


mikelehen:
Er, actually, can you remove the LOG_TAG, too?  @rsgowman just cleaned up our existing error logs in master. :-)

packages/firestore/src/remote/remote_store.ts
```diff
@@ -192,7 +192,7 @@ export class RemoteStore {
       this.watchStreamFailures++;
       if (this.watchStreamFailures >= ONLINE_ATTEMPTS_BEFORE_FAILURE) {
         if (this.shouldWarnOffline) {
-          log.debug(LOG_TAG, 'Could not reach Firestore backend.');
+          log.error(LOG_TAG, 'Could not reach Firestore backend.');
```
wilhuff: Pull #376 is removing tags from `log.error` calls. Let's please not add another one :-).

zxu123:
tag is removed from `log.error` call.

PR #268, opened 2017-10-26
Instrument Repo for Code Coverage

jshcrowthe:
This is going to probably take a little extra time (we have to get the extra CI integration approved) but having this instrumented will give us a path to improving our coverage.

### Browser

All of our browser builds running through `karma-webpack` are instrumented via [`istanbul-instrumenter-loader`](https://www.npmjs.com/package/istanbul-instrumenter-loader). We then dump this info for each component into the component `coverage/browser` dir by browser.

### Node

All of our node builds are instrumented via `nyc` which is a tool provided by the istanbul folks. Unfortunately, karma is not supported in the same way so we needed two tools. Coverage for this tool is dumped in the `coverage/node` directory.

## Merging `lcov.info` Files

After collecting `lcov.info` files (i.e. the coverage reports) for each file we leverage [`lcov-result-merger`](https://github.com/mweibel/lcov-result-merger) to consolidate to a single file that we can then pipe to CI.

I am going to leave the CI integration (i.e. `coveralls`) in as @vikrum is currently evaluating if we can get access here.

PR #380, opened 2017-12-14
Make disable internal sync and update notes about OnlineState

gsoltis:
Tried to clarify OnlineState comments a little. Also removed the `Promise<void>` return from `disableNetworkInternal()`, since it operates synchronously and was routinely being ignored anyways. Public API of `RemoteStore` remains unchanged.

### DIFF
--- a/packages/database/test/order_by.test.ts
+++ b/packages/database/test/order_by.test.ts
@@ -363,7 +363,9 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(done) {
+  it('Removing default listener removes non-default listener that loads all data', function(
+    done
+  ) {
     const ref = getRandomNode() as Reference;
 
     const initial = { key: 'value' };
--- a/packages/database/test/order_by.test.ts
+++ b/packages/database/test/order_by.test.ts
@@ -363,9 +363,7 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(
-    done
-  ) {
+  it('Removing default listener removes non-default listener that loads all data', function(done) {
     const ref = getRandomNode() as Reference;
 
     const initial = { key: 'value' };
--- a/packages/firestore/src/core/types.ts
+++ b/packages/firestore/src/core/types.ts
@@ -30,12 +30,17 @@ export type TargetId = number;
 // they're strings. We should probably (de-)serialize to a common internal type.
 export type ProtoByteString = Uint8Array | string;
 
-/** Describes the online state of the Firestore client */
+/**
+ * Describes the online state of the Firestore client. Note that this does not
+ * indicate whether or not the remote store is trying to connect or not. This is
+ * primarily used for views determining whether or not raise events from cache.
+ */
 export enum OnlineState {
   /**
    * The Firestore client is in an unknown online state. This means the client
    * is either not actively trying to establish a connection or it is currently
    * trying to establish a connection, but it has not succeeded or failed yet.
+   * Events are not raised from cache.
    */
   Unknown,
 
@@ -49,7 +54,7 @@ export enum OnlineState {
   /**
    * The client considers itself offline. It is either trying to establish a
    * connection but failing, or it has been explicitly marked offline via a call
-   * to disableNetwork().
+   * to disableNetwork(). Events will be raised from cache.
    */
   Failed
 }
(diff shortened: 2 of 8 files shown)


packages/database/test/order_by.test.ts
```diff
@@ -363,9 +363,7 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(
-    done
-  ) {
+  it('Removing default listener removes non-default listener that loads all data', function(done) {
```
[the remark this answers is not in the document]
gsoltis: I've got version 1.8.2, which is what's specified in `yarn.lock`. Looks like the version in `yarn.lock` was updated most recently here: https://github.com/firebase/firebase-js-sdk/pull/268

packages/firestore/src/core/types.ts
```diff
@@ -30,12 +30,17 @@ export type TargetId = number;
 // they're strings. We should probably (de-)serialize to a common internal type.
 export type ProtoByteString = Uint8Array | string;
 
-/** Describes the online state of the Firestore client */
+/**
+ * Describes the online state of the Firestore client. Note that this does not
+ * indicate whether or not the remote store is trying to connect or not. This is
+ * primarily used for views determining whether or not raise events from cache.
```
mikelehen: So it's more nuanced than "whether or not to raise events from cache"...  Maybe:

This is primarily used by the View / EventManager code to change their behavior while offline (e.g. get() calls shouldn't wait for data from the server and snapshot events should set metadata.isFromCache=true).

packages/firestore/src/core/types.ts
```diff
@@ -30,12 +30,17 @@ export type TargetId = number;
...
+ */
 export enum OnlineState {
   /**
    * The Firestore client is in an unknown online state. This means the client
    * is either not actively trying to establish a connection or it is currently
    * trying to establish a connection, but it has not succeeded or failed yet.
+   * Events are not raised from cache.
```
mikelehen: This isn't really true because snapshot listeners always get events from cache (since we'll raise a new event when we get up-to-date server data).

packages/firestore/src/core/types.ts
```diff
@@ -49,7 +54,7 @@ export enum OnlineState {
   /**
    * The client considers itself offline. It is either trying to establish a
    * connection but failing, or it has been explicitly marked offline via a call
-   * to disableNetwork().
+   * to disableNetwork(). Events will be raised from cache.
```
mikelehen: Per my previous comments, it's more nuanced than this so we should either drop this or rephrase it somehow (like maybe "Higher-level components should operate in offline mode.", but that's not that different than "The client considers itself offline.")

packages/database/test/order_by.test.ts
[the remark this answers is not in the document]
mikelehen: When I do a "yarn prepush" from the root of the repo, it makes a commit that undoes all of these changes...  Can you give that a try?

packages/firestore/src/core/types.ts
gsoltis: Ok, I've updated the comments. Take a look.

packages/firestore/src/core/types.ts
gsoltis: I rephrased this a little, let me know if you think it's correct now.

packages/firestore/src/core/types.ts
```diff
@@ -47,9 +54,9 @@ export enum OnlineState {
   Healthy,
 
   /**
-   * The client considers itself offline. It is either trying to establish a
-   * connection but failing, or it has been explicitly marked offline via a call
-   * to disableNetwork().
+   * The client is either trying to establish a connection but failing, or it
+   * has been explicitly marked offline via a call to disableNetwork().
+   * Higher-level components should not operate in offline mode.
```
mikelehen: I think this is the opposite of correct now. :-P

packages/firestore/src/remote/remote_store.ts
```diff
@@ -243,16 +243,15 @@ export class RemoteStore {
...
   }
 
   /**
    * Disables the network, setting the OnlineState to the specified
    * targetOnlineState.
    */
-  private disableNetworkInternal(
-    targetOnlineState: OnlineState
-  ): Promise<void> {
+  private disableNetworkInternal(targetOnlineState: OnlineState) {
```
mikelehen: Can you explicitly declare the return type to be void now?

packages/firestore/src/core/types.ts
gsoltis: Oops, yeah, cut-and-paste fail.

packages/firestore/src/remote/remote_store.ts
gsoltis: Done.

mikelehen:
:+1:

PR #382, opened 2017-12-14
Add changelog entry for new warning log messages.

mikelehen:
Cherry-pick https://github.com/firebase/firebase-js-sdk/commit/2fa51d3b45705e9f27f500ab4def078905cbed4e (as the original change https://github.com/firebase/firebase-js-sdk/commit/72536c468d2375794a3b18c20fd032d6f0adf178 is already in master) and update changelog based on the two log messages we've added.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Unreleased
-- [changed] A "Could not reach Firestore backend." warning message will be
+- [changed] A "Could not reach Firestore backend." message will be
   logged when the initial connection to the Firestore backend fails.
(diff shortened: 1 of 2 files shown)


mikelehen:
Oops, I'm going to amend this. I realized that:

1. This log isn't on in master since https://github.com/firebase/firebase-js-sdk/commit/2fa51d3b45705e9f27f500ab4def078905cbed4e is only in firestore-api-changes.
2. The "resource exhausted" log message is in master too, and should be included in the changelog too.

So I think I'll cherry-pick 1 and update the changelog for 2.

mikelehen:
Re-assigning to @wilhuff since apparently I need a code owner _in addition to myself_ to approve. :-/

Issue #202, opened 2017-10-10
FR: Provide way to leverage structural sharing in DataSnapshot


dtinth:
### [REQUIRED] Describe your environment

  * Operating System version: Mac OS X 10.12
  * Firebase SDK version: 4.1.1
  * Firebase Product: database

### [REQUIRED] Describe the problem

In React-based apps, we optimize the performance of the app with “structural sharing.”

Since a DataSnapshot is immutable, I would expect a `DataSnapshot` of an unchanged child to be the same object in previous snapshot.

#### Steps to reproduce:

Consider the state of the realtime database where `task2.title` changed:

```
s1: { task1: { title: 'A' }, task2: { title: 'B' } }
s2: { task1: { title: 'A' }, task2: { title: 'Z' } }
```

Since between these 2 snapshots, the data of `task1` child is unchanged. Therefore, I would expect that:

```
s1.child('task1') === s2.child('task1')
```

However, a [new DataSnapshot is created every time you call `.child()`](https://github.com/firebase/firebase-js-sdk/blob/3f827b82aa83cb6b5538bb7881f3253e9af0d15f/packages/database/src/api/DataSnapshot.ts#L87-L100), hence, the above code yields `false`.

A workaround is to compare the underlying node:

```
s1.child('task1').node_ === s2.child('task1').node_
```

But it does not work with the hosted Firebase SDK as it gets closure-compiled, and `node_` was renamed to `A`.

Having structural sharing makes it much easier to [optimize a React application](https://reactjs.org/docs/optimizing-performance.html#avoid-reconciliation), by leveraging `PureComponent` and `shouldComponentUpdate`.

#### Relevant Code:

The workaround I am currently using is to parse the source code of `dataSnapshot.val()` at runtime to obtain access to the underlying node object.

```javascript
function getUnderlyingNode (dataSnapshot) {
  // ULITAME HACK:
  //     PARSE THE FIREBASE SDK SOURCE CODE TO FIGURE OUT HOW
  //     TO OBTAIN THE UNDERLYING NODE.
  const match = dataSnapshot.val.toString().match(/return\s+this\.(\w+)/)
  return dataSnapshot[match[1]]
}
```

By comparing the underlying `node_`, I was able to improve the performance of my React app by ~100x. (It works with a lot of data.)

#### API suggestions:

To avoid such ugly hack, here are some ways that the DataSnapshot API can be improved so that userland code could leverage structural sharing:

[...]

PR #396, opened 2017-12-22
Firestore `isEqual()` on Public Types

zxu123:
Discussion
*   See b/62143881
*   Also see inspiration on http://github.com/firebase/firebase-js-sdk/issues/202

Testing
*   Adds unit test for all relevant isEqual(), either already implemented in master or in this branch
*   All unit tests and integration tests pass.

API Changes
*   See proposal go/firestoreisequal

### DIFF
--- a/packages/database/test/order_by.test.ts
+++ b/packages/database/test/order_by.test.ts
@@ -363,7 +363,9 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(done) {
+  it('Removing default listener removes non-default listener that loads all data', function(
+    done
+  ) {
     const ref = getRandomNode() as Reference;
 
     const initial = { key: 'value' };
--- a/packages/database/test/order_by.test.ts
+++ b/packages/database/test/order_by.test.ts
@@ -363,9 +363,7 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(
-    done
-  ) {
+  it('Removing default listener removes non-default listener that loads all data', function(done) {
     const ref = getRandomNode() as Reference;
 
     const initial = { key: 'value' };
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -157,7 +157,7 @@ class FirestoreSettings {
     this.credentials = settings.credentials;
   }
 
-  equals(other: FirestoreSettings): boolean {
+  isEqual(other: FirestoreSettings): boolean {
     return (
       this.host === other.host &&
       this.ssl === other.ssl &&
@@ -237,7 +237,7 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     }
 
     const newSettings = new FirestoreSettings(settingsLiteral);
-    if (this._firestoreClient && !this._config.settings.equals(newSettings)) {
+    if (this._firestoreClient && !this._config.settings.isEqual(newSettings)) {
       throw new FirestoreError(
         Code.FAILED_PRECONDITION,
         'Firestore has already been started and its settings can no longer ' +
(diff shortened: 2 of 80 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -1031,11 +1045,24 @@ export class DocumentSnapshot implements firestore.DocumentSnapshot {
...
+    if (!(other instanceof DocumentSnapshot)) {
+      throw invalidClassError('isEqual', 'DocumentSnapshot', 1, other);
+    }
+    return (
+      this.firestore === other.firestore &&
+      this._key.isEqual(other._key) &&
+      (this._document === null
+        ? other._document === null
+        : this._document.isEqual(other._document)) &&
+      this._fromCache === other._fromCache
```
schmidt-sebastian: Nit: I would move the _fromCache comparison up since it's cheaper.

packages/firestore/src/api/database.ts
```diff
@@ -1618,6 +1645,45 @@ export class QuerySnapshot implements firestore.QuerySnapshot {
...
+      this.firestore !== other.firestore ||
+      !this._originalQuery.isEqual(other._originalQuery)
+    ) {
+      return false;
+    }
+    let snapshot: ViewSnapshsot = this._snapshot;
+    let otherSnapshot: ViewSnapshot = other._snapshot;
+    if (
+      !snapshot.query.isEqual(otherSnapshot.query) ||
+      !snapshot.docs.isEqual(otherSnapshot.docs) ||
```
schmidt-sebastian: I'm kind of scared of this as this comparison is super expensive (it materializes the docs array). Unless we add `readTime`, I don't have a better solution to offer.

packages/firestore/src/api/database.ts
```diff
@@ -1618,6 +1645,45 @@ export class QuerySnapshot implements firestore.QuerySnapshot {
...
+      !snapshot.oldDocs.isEqual(otherSnapshot.oldDocs) ||
+      snapshot.fromCache !== otherSnapshot.fromCache ||
```
schmidt-sebastian: Please move the cheaper comparisons up.

zxu123: done

packages/firestore/src/api/database.ts
zxu123: done

packages/firestore/src/api/database.ts
zxu123: Let me add a comment to warn user on the expensiveness.

zxu123:
Node test is fixed; was confused the internal `FieldPath` w/ api `FieldPath`. Adding missing isEqual into api `FieldPath` fixed all node test failures. Please take another look.

mikelehen:
I'll take a look at this PR, but can you see why the build / tests are still failing in travis?

zxu123:
Fixed. Please take another look.

packages/database/test/order_by.test.ts
```diff
@@ -363,9 +363,7 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(
-    done
-  ) {
+  it('Removing default listener removes non-default listener that loads all data', function(done) {
```
mikelehen: So when I run "yarn prepush", it undoes these changes...  can you run "yarn" from the root of the repo to make sure you have the correct prettier installed and then run "yarn prepush" to see if it creates a commit that undoes these changes for you too?

packages/firestore/test/unit/api/database.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  collectionReference,
+  documentReference,
+  documentSnapshot,
+  query,
+  querySnapshot
+} from '../../util/api_helpers';
+import { expectEqual, expectNotEqual } from '../../util/helpers';
+
+describe('CollectionReference', () => {
+  it('CollectionReference equality checks', () => {
```
mikelehen: The way the describe() and it() work is they get concatenated together and are supposed to essentially form a sentence, e.g.:

```
describe('CollectionReferences', () => {
  it('support equality checking with isEqual()', () => {
    ...
  })
})
```

and then when the tests run you'll see

CollectionReferences support equality testing with isEqual() PASSED

We're not super rigorous about it forming a real sentence, but you should at least avoid duplicating 'CollectionReference' in the describe() and the it().

packages/firestore/src/api/database.ts
mikelehen: I don't think users should use QuerySnapshot.isEqual() very often (if ever), so I'm okay with it being slow.  We could add a TODO to improve the performance, but I think DocumentSnapshot.isEqual() is much more important (e.g. so devs can compare documents from two different QuerySnapshots to see what changed).

zxu123:
Updated. PTAL.

PR #411, opened 2018-01-04
Add integration tests for updateToken

pinarx:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/integration/messaging/test/test-updateToken.js
+++ b/integration/messaging/test/test-updateToken.js
@@ -19,15 +19,16 @@ const expect = require('chai').expect;
 
 const setupNotificationPermission = require('./utils/setupNotificationPermission');
 const testServer = require('./utils/test-server');
-const getErrors = require('./utils/getErrors');
 const retrieveFCMToken = require('./utils/retrieveFCMToken');
-const deleteFCMToken = require('./utils/deleteFCMToken');
+const timeForward = require('./utils/timeForward');
+const getFCMToken = require('./utils/getFCMToken');
+const getErrors = require('./utils/getErrors');
 const demoSetup = require('./utils/getDemoSetup');
 
 const ENDPOINT = 'https://fcm.googleapis.com';
 const DEMOS = demoSetup.DEMOS;
 
-describe('Firebase Messaging Integration Tests > get and delete token', function() {
+describe('Firebase Messaging Integration Tests > update a token', function() {
   this.timeout(60 * 1000);
   if (process.env.TRAVIS) {
     this.retries(3);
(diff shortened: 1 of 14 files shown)


integration/messaging/test/utils/getFCMToken.js
```diff
@@ -15,15 +15,13 @@
...
     return webdriver.executeScript(() => {
-      return !!window.__test && !!window.__test.token;
+      return !!window.__test;
     });
   });
-
-  console.log('Found token.');
   return webdriver.executeScript(() => {
-    return window.__test.token;
+    return window.__test.triggerGetToken();
```
gauntface: Does this work?

`triggerGetToken()` returns a promise and I'm not sure `executeScript` can pass the browsers promise to node.

integration/messaging/test/test-updateToken.js
```diff
@@ -0,0 +1,102 @@
...
+            `${testServer.serverAddress}/${demoInfo.name}/`
+          );
+
+          const token = await retrieveFCMToken(globalWebDriver);
+          expect(token).to.exist;
+
+          // roll the clock forward > 7days
+          await timeForward(globalWebDriver, EIGHT_DAYS_IN_MS);
+          try {
+            await getFCMToken(globalWebDriver);
```
gauntface: Related to the above question - this function is getting a toke, but it's not checking for anything so it's unlikely to be testing the behavior you think it is (i.e. `executeScript()` returns a promise, which I don't think is returning the promise you return from within the promise).

packages/database/test/order_by.test.ts
```diff
@@ -363,9 +363,7 @@ describe('.orderBy tests', function() {
     expect(addedPrevNames).to.deep.equal(expectedPrevNames);
   });
 
-  it('Removing default listener removes non-default listener that loads all data', function(
-    done
-  ) {
+  it('Removing default listener removes non-default listener that loads all data', function(done) {
```
gauntface: Did you make these changes or is this a prettier thing? cc @jshcrowthe

integration/messaging/test/utils/getFCMToken.js
pinarx: it works in the sense that the network request is made, and it doesn't fail - which is mostly what I wanted to cover, but agreed that the logic doesn't make much sense... Changing to your suggestion.

integration/messaging/test/test-updateToken.js
pinarx: I actually want to test the right network call is made, but not sure if there's a way to do that. Changed it to check that the token exists.

packages/database/test/order_by.test.ts
pinarx: prettier.

gauntface: @jshcrowthe Is there any chance you could look into a way of using the git changes to detect which packages have been changed and only run prettier over that directory.

pinarx:
@jshcrowthe, @mikelehen, @schmidt-sebastian @wilhuff: PTAL for automated prettier changes.
@gauntface: It looks like I can't pass parameters into executeScript so moving the time constant into app.js for integration tests.

schmidt-sebastian:
LGTM for Prettier changes in RTDB and Firestore.

packages/database/test/order_by.test.ts
jshcrowthe: Yeah I can do that. The react team has done it so I'll see what I can reuse.

mikelehen: I think this may actually be an issue with using the wrong prettier version.  Can you run "yarn" from the root of the repo and then "yarn prepush" and see if it makes a commit that undoes this change?

pinarx: tried,  doesn't seem to help.

gauntface:
@jshcrowthe whats the protocol here? Just merge away? Expect reviews on generated code? Help please.

jshcrowthe:
@gauntface as this isn't a PR to master, this PR can be merged whenever you are happy w/ it. It isn't  until a PR is targeting master that the reviewers and checks enforcement is required.

jshcrowthe:
Additionally @pinarx , much of the churn here and in https://github.com/firebase/firebase-js-sdk/pull/423 seems to be due to use of `npm` instead of `yarn` (i.e. multiple accidental inclusions of `package-lock.json` files, manual editing of the `yarn.lock` file, mismatching prettier versions, etc). I'd recommend following the SDK Dev Workflow instructions in the README, you'll save yourself quite a few headaches if you do.

pinarx:
@gauntface PTAL

PR #412, opened 2018-01-05
Add timeout to OnlineState tracking.

mikelehen:
Add 10 second timeout to OnlineState tracking.

* Added a timer that triggers after 10 seconds without a successful stream attempt, reverting to "offline" behavior, potentially triggering data from cache.
* Cleanup:
  * Renamed OnlineState enum values to Offline, Online, and Unknown.
  * Refactored "online state tracking" into its own OnlineStateTracker class.
  * Added a dummy .catch() handler to the CancelablePromises returned by
    AsyncQueue.enqueueAfterDelay() to avoid UnhandledPromiseRejection log spam
    due to canceled timers.
* Added a SpecTest to verify new timeout behavior.
  * Added ability to run timers from spec tests (including assigning string
    names to TimerId enum values)
  * Added TimerId.All to match iOS and make it easier to run all timers from spec
    tests.


packages/firestore/src/remote/remote_store.ts
```diff
@@ -51,19 +51,14 @@ import {
   WatchTargetChange,
   WatchTargetChangeState
 } from './watch_change';
+import { clearTimeout } from 'timers';
```
wilhuff: This is no longer necessary, right?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -139,8 +124,10 @@ export class RemoteStore {
     private localStore: LocalStore,
     /** The client-side proxy for interacting with the backend. */
     private datastore: Datastore,
-    private onlineStateHandler: (onlineState: OnlineState) => void
-  ) {}
+    onlineStateHandler: (onlineState: OnlineState) => void
+  ) {
+    this.onlineStateTracker = new OnlineStateTracker(onlineStateHandler);
```
wilhuff: Might as well pass the `OnlineStateTracker` in so the RemoteStore doesn't even have to know about the onlineStateHandler.

packages/firestore/src/remote/online_state_tracker.ts
```diff
@@ -0,0 +1,158 @@
...
+ * whether or not the client as a whole should be considered to be online or
+ * offline), implementing the appropriate heuristics.
+ *
+ * In particular, when the client is trying to connect to the backend, we
+ * allow up to MAX_WATCH_STREAM_FAILURES within MAX_WATCH_STREAM_TIMEOUT_MS for
+ * a connection to succeed. If we have too many failures or the timeout elapses,
+ * then we set the OnlineState to Offline, and the client will behave as if
+ * it is offline (get()s will return cached data, etc.).
+ */
+export class OnlineStateTracker {
```
wilhuff: I can't think of any way it would happen but just in case: are there any interactions between this class and the idle timeout timers?

packages/firestore/src/remote/online_state_tracker.ts
```diff
@@ -0,0 +1,158 @@
...
+   * Called by RemoteStore when a watch stream is started.
+   *
+   * It sets the OnlineState to Unknown and starts a MAX_WATCH_STREAM_TIMEOUT_MS
+   * timer if necessary.
+   */
+  handleWatchStreamStart(): void {
+    this.setAndBroadcast(OnlineState.Unknown);
+
+    if (this.watchStreamTimer === null) {
+      this.watchStreamTimer = setTimeout(() => {
```
wilhuff: Given all the timer gunk we have now I wonder if it would make sense to either add methods to the async queue to just do this for us or add a special clock/delay type that you could pass in to the constructor to allow us to test these kinds of things directly.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -236,8 +178,8 @@ export class RemoteStore {
    */
   disableNetwork(): Promise<void> {
     this.disableNetworkInternal();
-    // Set the OnlineState to failed so get()'s return from cache, etc.
-    this.updateOnlineState(OnlineState.Failed);
+    // Set the OnlineState to Offline so get()'s return from cache, etc.
```
wilhuff: Don't use apostrophes for making plurals. (it should be "get()s" or just "gets").

packages/firestore/src/remote/remote_store.ts
```diff
@@ -340,18 +282,24 @@ export class RemoteStore {
...
+    this.onlineStateTracker.handleWatchStreamStart();
   }
 
   /**
-   * Returns whether the watch stream should be started because there are
-   * active targets trying to be listened too
+   * Returns whether a watch stream is currently necessary (network is enabled
+   * and we have active watch targets).
+   */
+  private isWatchStreamNecessary(): boolean {
```
wilhuff: I'm not a fan of this extra concept that's confusingly similar to shouldStartWatchStream, especially since it's only used in the implementation of it.

wilhuff:
Despite the nitpicking in here I really like the direction this is heading.

packages/firestore/src/remote/online_state_tracker.ts
mikelehen: There shouldn't be since this timer goes away once you're connected. And the idle timers start once you're connected...  And with the improved testing, if both timers were active at once, runDelayedOperationsEarly() would run both so we'd be exercising them together, at least in theory...

packages/firestore/src/remote/online_state_tracker.ts
```diff
@@ -0,0 +1,166 @@
...
+   *
+   * It sets the OnlineState to Unknown and starts a MAX_WATCH_STREAM_TIMEOUT_MS
+   * timer if necessary.
+   */
+  handleWatchStreamStart(): void {
+    this.setAndBroadcast(OnlineState.Unknown);
+
+    if (this.watchStreamTimerPromise === null) {
+      this.watchStreamTimerPromise = this.asyncQueue.enqueueAfterDelay(
+        TimerId.OnlineStateTimeout,
```
wilhuff: I find this cluster of names confusingly dissimlar. Why is this an "OnlineStateTimeout", the promise is a "watchStreamTimerPromise", and the timeout a "MAX_WATCH_STREAM_TIMEOUT_MS"?

These all seem related to the handling the initial connection but:
  * OnlineStateTimeout seems like it might be applicable to other transitions
  * watchStreamTimer seems wholly unrelated to that purpose
  * MAX_WATCH_STREAM_TIMEOUT_MS seems just as likely to apply to an idle timeout. Also, why "MAX"? Isn't this just the timeout duration?

packages/firestore/src/remote/online_state_tracker.ts
```diff
@@ -0,0 +1,166 @@
...
+            'Timer should be canceled if we transitioned to a different state.'
+          );
+          log.debug(
+            LOG_TAG,
+            `Watch stream didn't reach online or offline within ` +
+              `${MAX_WATCH_STREAM_TIMEOUT_MS}ms. Considering client offline.`
+          );
+          this.logClientOfflineWarningIfNecessary();
+          this.setAndBroadcast(OnlineState.Offline);
+          return Promise.resolve();
```
wilhuff: Should this failure to connect in time bump the number of watch stream failures? If not could you please add comments why not?

packages/firestore/src/remote/remote_store.ts
wilhuff: I don't think there's a strong case either way. The general principle I was following was to invert dependencies where possible, the general rule being to pass your dependencies rather than new them up.

This makes it possible to construct a testing OnlineStateTracker if we ever wanted to do so and pushes all the object construction into one place. While it's true that FirestoreClient doesn't need to know about OnlineStateTracker, FirestoreClient is the assembly point for the rest of our constellation of objects so it doesn't seem like a bad place.

packages/firestore/src/util/async_queue.ts
```diff
@@ -27,12 +27,17 @@ type TimerHandle = any;
...
-  ListenStreamConnection,
-  WriteStreamIdle,
-  WriteStreamConnection
+  // All can be used with runDelayedOperationsEarly() to run all timers.
+  All = 'all',
+  ListenStreamIdle = 'listen_stream_idle',
+  ListenStreamConnection = 'listen_stream_connection',
+  WriteStreamIdle = 'write_stream_idle',
+  WriteStreamConnection = 'write_stream_connection',
+  OnlineStateTimeout = 'online_state_timeout'
```
wilhuff: I don't understand how OnlineStateTimeout is different from either the connection or idle timeouts. I'm not sure if I'm asking for it to be deduplicated against those others, comments on the individual timer ids, or some kind of theory of operation primer on how these all fit together.

mikelehen: I added comments that hopefully makes things somewhat clearer.  FWIW, coming up with names for these is a bit tricky because they're meant to be high-level concepts (since they'll be triggered from high-level tests, etc.) and so e.g. "Connection" was chosen over "Backoff"

packages/firestore/src/remote/online_state_tracker.ts
wilhuff: This is a marked improvement and LGTM as-is. Thanks. Agree that leaning on comments can be the way to go in cases like these.

I'm still torn about this because the connection timer seems very much related to what we're doing here: namely if we're ever connecting our state starts at unknown. If we logically fail to connect (i.e. fail to transition out of unknown) then this timer fires. It seems like, for all intents and purposes, OnlineStateUnknownTimeout is the real connection timeout. Could that other thing we have be some lower-level concept?

Or maybe the name of OnlineStateUnknownTimeout itself is caught in my craw. Our other timeouts are named in terms of the end state, not the starting state. IOW: how long until some state, not how long in some state. How long until idle, how long until connected, etc.

This comment isn't super actionable but I'm still uneasy, and if history has shown anything: this code is fiddly, keeps changing, and we keep misunderstanding it over time, so if you have the energy I'd like to try to find a way to make it more immediately obvious and less fuzzy.

mikelehen: I agree that our timers have been fiddly / confusing and getting this right is worthwhile. :-)

So right now our timers are: ListenStreamIdle / ListenStreamConnection, WriteStreamIdle / WriteStreamConnection, OnlineStateTimeout

The Connection timers are used to schedule delayed stream attempts subject to exponential backoff...  Not all stream attempts are delayed in which case the "Connection" timer doesn't come into play (e.g the first attempt once a stream becomes necessary, or the first attempt after an established stream encounters an error, etc.). 

But the OnlineStateTimeout is still relevant when the stream attempt isn't delayed (e.g. it'll fire if the first stream attempt doesn't succeed / fail within 10 seconds, triggering us to OnlineState.Offline while we wait).

And the Connection timers will continue to be used even once the OnlineStateTimeout has triggered (because we'll continue scheduling stream attempts with backoff even after we've entered OnlineState.Offline).  

So I think they're not actually that related, and we probably have a naming problem that makes them seem related. We could try to be more verbose, e.g. ListenStreamDelayedRestart instead of ListenStreamConnection.

In naming these, it's worth thinking about their usage from tests, e.g.: runDelayedOperationsUntil(TimerId.ListenStreamDelayedRestart) to make a test skip the exponential backoff delay.

wilhuff: OK could the connection timer ids really be backoff timer ids? This is essentially where we use them:

```js
    this.backoff = new ExponentialBackoff(
      queue,
      connectionTimerId,
      BACKOFF_INITIAL_DELAY_MS,
      BACKOFF_FACTOR,
      BACKOFF_MAX_DELAY_MS
    );
```

`TimerId.ListenStreamBackoff` => `runDelayedOperationsUntil(TimerId.ListenStreamBackoff)`?

I don't think the fact that the online state timeout is for succeeding or failing much matters. Connection timeouts pretty much always work this way: they introduce a client-generated error if the connection hasn't already succeeded or failed on its own.

mikelehen: Yes, that's exactly what they are, except "backoff" describes an implementation detail about how the timers are scheduled rather than what they actually are for. You could hypothetically imagine that the stream could have multiple timers, for different purposes, that happen to use exponential backoff scheduling, at which point "ListenStreamBackoff" would be meaningless (as opposed to "ListenStreamDelayedRestart" or whatever).

That said, we probably conflate "backoff" with the act of reconnecting already, and so I'd be okay with ListenStreamBackoff if you think that's clearer.

mikelehen: Also, I still do think there's a meaningful distinction between our "online state timeout" and a "connection timeout" because we explicitly don't cancel the existing stream attempt and start a new one...  Instead we're relying on grpc's internal timeout / retry connection logic.

Issue #408, opened 2018-01-03
Meaningless return statements


merlinnot:
Hi,

I’ve noticed a meaningless return statement like this:
```ts
async () => {
  return Promise.resolve()
}
```

I’d like to create a PR to remove this statement, but you’ve asked in CONTRIBUTING.md to create an issue first. So here I am :)

jshcrowthe:
Hey @merlinnot 😄 

Thanks for reading the `CONTRIBUTING.md` if you can point to the specific issue(s) then let's talk!

merlinnot:
Here you go:

https://github.com/firebase/firebase-js-sdk/blob/1a1bd233b0f459932f8f970836fb7ec2c13d1ee5/packages/firestore/test/integration/api/database.test.ts#L678

jshcrowthe:
Feel free to submit a PR. I'll leave it to those closer to the code to discuss if there are any issues.

PR #420, opened 2018-01-09
Remove meaningless return statement

merlinnot:
Simplifies one of the integrations tests in a firestore module.

See https://github.com/firebase/firebase-js-sdk/issues/408


mikelehen:
Thanks for the PR! These used to be required for the code to compile (we may have been using a different Promise type definition or something), and so we have these `return` statements all over.  Would you be interested in removing all of them, or at least the rest of the ones in this file?  Else I can merge this as-is, but it would be best not to do this cleanup 1 line at a time. :-)

merlinnot:
Y, I’ll do that.

merlinnot:
Closing in favor of #422 which is made against `master`

Issue #116, opened 2017-07-27
Unable to Proxy Firebase Url


therynamo:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: macOs Sierra 10.12.5
  * Firebase SDK version: shouldn't make much of a difference in this issue
  * Library version: ^5.1.0
  * Firebase Product: `database`

### [REQUIRED] Step 3: Describe the problem

When attempting to use `firebase-admin`s `initializeApp` function, the URL is strictly enforced. I'm sure this is intended, however it is keeping me from proxying the firebase URL to a local server for testing. Example:
```js
firebase.initializeApp({
  credential: firebase.credential.cert(FIREBASE_CREDENTIALS),
  databaseURL: 'http://localhost:4001/https://my.firebaseio.com'
});
```

Resulting in the following error:

```bash
Error: FIREBASE FATAL ERROR: Cannot parse Firebase url. Please use https://<YOUR FIREBASE>.firebaseio.com
```

I use [mockyeah](https://github.com/ryanricard/mockyeah/wiki/Service-Snapshot-CLI) for integration testing, which is a nice solution for being able to record snapshots for outgoing requests and their responses. The only caveat is that in order to record the response you have to proxy the URL so that the server can make the call itself. 

The issue isn't necessarily with implementation on either side. My overall question is, would supporting a proxied URL be feasible?

#### Steps to reproduce:
Listed above.
#### Relevant Code:

Listed above. 

[Related Issue From `firebase-admin`](https://github.com/firebase/firebase-admin-node/issues/63#issuecomment-318475037).

therynamo:
@jshcrowthe thanks for adding the label. I thought I'd add a little more information now that I've explored around looking for more options. 

Since `mockyeah` would not directly work in this situation due to the strict url enforcement on the firbase side, I moved over to `nock` to try and utilize their direct http intercept functionality. Effectively it should have blocked the call from ever being made and just returning `json`. I quickly found out that google oath was the first call being made by `firebase-admin` and that you couldn't directly return the desired response from there. 

[...]

PR #426, opened 2018-01-11
Allow localhost database URL

rynobax:
This change was discussed in #116.

### DIFF
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -138,7 +138,6 @@ export const parseURL = function(
       domain = parts[0];
     } else if (parts[0].slice(0, colonInd).toLowerCase() === 'localhost') {
       domain = 'localhost';
-      subdomain = 'localhost';
     }
 
     // If we have a port, use scheme for determining if it's secure.
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -57,7 +57,7 @@ export const parseRepoInfo = function(
   }
 
   // Catch common error of uninitialized namespace value.
-  if (!namespace || namespace == 'undefined') {
+  if ((!namespace || namespace == 'undefined') && parsedUrl.domain !== 'localhost') {
     fatal(
       'Cannot parse Firebase url. Please use https://<YOUR FIREBASE>.firebaseio.com'
     );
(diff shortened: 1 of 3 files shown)


rynobax:
I signed it!

packages/database/src/core/util/libs/parser.ts
```diff
@@ -131,6 +131,8 @@ export const parseURL = function(
       subdomain = parts[0].toLowerCase();
     } else if (parts.length === 2) {
       domain = parts[0];
+    } else if (parts[0].split(':')[0].toLowerCase() === 'localhost') {
+      subdomain = 'localhost';
```
schmidt-sebastian: Can you move the port extraction logic from line 136 up so that it happens before this if-statement? Then we won't have to double parse the port number.

Then I would also suggest that instead of assigning 'localhost' to `subdomain` you should assign it to `domain`. A subdomain would be something like 'foo' in 'foo.localhost'.

rynobax:
I believe I made the change you requested about parsing the port number, let me know if that is not what you had in mind.

I also changed it to set the domain to localhost.  However, the subdomain also needs to be set to something, as it is used for the 'namespace' variable, which cannot be empty.  For now it also sets the subdomain to 'localhost', as I am not entirely sure of what the namespace is used for.

schmidt-sebastian:
> I also changed it to set the domain to localhost. However, the subdomain also needs to be set to something, as it is used for the 'namespace' variable, which cannot be empty. For now it also sets the subdomain to 'localhost', as I am not entirely sure of what the namespace is used for.

Thanks, the changes look mostly good. The namespace is used by our backend if you are not connecting the intended host. For mocking, this should not be needed. It seems like you changed the assert to no longer fire AND you set the subdomain to localhost? Can you explain why we need to do both - otherwise I would just remove the setting of the subdomain and we should be good to submit this PR.

rynobax:
If the subdomain is not set, the test fails with the error `Error: Invalid Firebase Database URL failed: First argument must be a valid firebase URL and the path can't contain ".", "#", "$", "[", or "]".`

The error seems to be triggered by this line, because the namespace is an empty string:
https://github.com/firebase/firebase-js-sdk/blob/73a586c92afe3f39a844b2be86086fddb6877bb7/packages/database/src/core/util/validation.ts#L420

If I remove that line all the tests pass.

The two options that come to my mind are:

1. Do not set the subdomain on localhost domains
2. In validation.ts, skip the namepace validity check if the host is `localhost`

Or:

1. Set the subdomain on localhost domains
2. Remove the extra check I added to the assert

schmidt-sebastian:
> Do not set the subdomain on localhost domains
> In validation.ts, skip the namespace validity check if the host is localhost

Please go ahead and do this. It seems cleaner.

rynobax:
I made the changes.

schmidt-sebastian:
LGTM. Note that this won't work for localhost URLs that don't have a port set. I will send you a follow up PR to add support for this.

PR #427, opened 2018-01-11
Fix a bunch of lint warnings in Firestore

schmidt-sebastian:
This fixes all Lint warnings in Firestore that are not related to the `any` type.


schmidt-sebastian:
FYI: The node tests are passing for me. I am having some trouble with the browser tests:

```
11 01 2018 00:00:10.507:ERROR [reporter]: Can not load reporter "coverage-istanbul", it is not registered!
  Perhaps you are missing some plugin?
```

@jshcrowthe Do you have a suggestion for this?

PR #433, opened 2018-01-12
Add a new package template

jshcrowthe:
For those users looking to create new packages in the SDK I figured it'd be good to have a template.

This is just some bare bones files that can be copied to create a working base for a new package.

### DIFF
--- a/packages/template/.npmignore
+++ b/packages/template/.npmignore
@@ -1,0 +1,1 @@
+# This file is left intentionally blank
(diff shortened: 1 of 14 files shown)


packages/template/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../config/tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "dist"
+  },
+  "exclude": [
+    "dist/**/*"
+  ]
+}
```
hiranya911: Nit: Add new line at eof

packages/template/.npmignore
```diff
@@ -0,0 +1 @@
+# This file is left intentionally blank
```
hiranya911: New line at eof

PR #440, opened 2018-01-16
[Multi-Tab] Persisting foreground state for Firestore tabs

schmidt-sebastian:
This is the first PR for Firestore Multi-Tab. It adds a new WebStorage class that is used to persist the foreground state of a tab. This state gets refreshed every 4s.

b/71753634

This also pulls in the License Header change in https://github.com/firebase/firebase-js-sdk/pull/439 (which I will merge first).


mikelehen:
I want to read your doc before reviewing this so that we're on the same page about the implementation goals, etc. But from my brief look, my first high-level feedback is that I'm questioning whether SyncEngine and EventManager should know about browser-specific concepts (WebStorage, visibility events). Ideally that would be encapsulated inside the persistence components, I think? Today, SyncEngine and EventManager have identical functionality across platforms. We may not be able to maintain that while introducing multi-tab, but we should try to, and be explicit where we start deviating (and keep it as isolated and self-contained as possible).

To be clear though, you probably shouldn't take any action on this feedback until I read your doc and review this for real. So ignore me for now. :-)

packages/firestore/src/local/web_storage.ts
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * WebStorage of the Firestore client. Firestore uses WebStorage for cross-tab
+ * notifications and to persist the metadata state of each tab. WebStorage is
+ * used to perform leader election and to inform other tabs about changes in the
+ * IndexedDB-backed persistence layer.
+ *
+ * WebStorage is only used in Persistence-enabled Firestore instances. If you
+ * are not using Persistence, consider using the `NoOpWebStorage` class.
+ */
+export interface WebStorage {
```
wilhuff: I'm not wild about this name--it doesn't really describe what this thing is for. Also, we already have LocalStore and RemoteStore, so if this thing really is going to be named after storing things we should probably go for a name ending in "Store".

However, the first sentence of the comment indicates this is cross tab notifications. How about `TabNotificationChannel`?

packages/firestore/src/local/web_storage.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    assert(!this.started, 'PersistedWebStorage already started');
+
+    this.localStorage = window.localStorage;
+    this.started = true;
+    this.persistState();
+    this.scheduleRefresh();
+  }
+
+  shutdown(): void {
```
wilhuff: Doesn't this need to cancel the scheduled task?

packages/firestore/src/local/web_storage.ts
```diff
@@ -0,0 +1,146 @@
...
+      return Promise.resolve();
+    }, WEB_STORAGE_REFRESH_INTERVAL_MS);
+  }
+
+  /** Persists the entire known state. */
+  private persistState(): void {
+    assert(this.started, 'PersistedWebStorage not started');
+    debug(LOG_TAG, 'Persisting state in LocalStorage');
+    this.localStorage[
+      this.buildKey(VISIBILITY_PREFIX, this.persistenceKey, this.instanceId)
```
wilhuff: This key seems like something you could build once in the constructor and not even retain the arguments used to construct it.

packages/firestore/src/local/web_storage.ts
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * `PersistedWebStorage` uses Local Storage as the backing store for the
+ * WebStorage class.
+ *
+ * Once started, PersistedWebStorage will rewrite its contents to Local Storage
+ * every four seconds. Other clients may disregard its state after five seconds
+ * of inactivity.
+ */
+export class PersistedWebStorage implements WebStorage {
```
wilhuff: If IndexedDB listeners became available someday and we wanted to use either those or this one maybe it would be better to name this after how it's implemented?

Is LocalStorageTabNotificationChannel too much?

packages/firestore/src/local/web_storage.ts
```diff
@@ -0,0 +1,146 @@
...
+    return JSON.stringify(persistedData);
+  }
+}
+
+/**
+ * `NoOpWebStorage` implements the WebStorage API but neither persist data nor
+ * delivers notifications. This call should be used for non-persistence enabled
+ * clients.
+ */
+export class NoOpWebStorage implements WebStorage {
```
wilhuff: I wonder if we even need this at all.

If persistence is disabled we don't need to create a window visibility listener and we don't need to create a tab notification channel either, maybe we can just not create them?

It's a little hard to know, I guess, and maybe it will be clearer once you write the code to actually consume this value, but something to keep in mind.

packages/firestore/src/core/event_manager.ts
```diff
@@ -55,11 +56,34 @@ export class EventManager {
...
+  constructor(private asyncQueue: AsyncQueue, private syncEngine: SyncEngine) {
     this.syncEngine.subscribe(
       this.onChange.bind(this),
       this.onError.bind(this)
     );
+    this.registerVisibilityListener();
+  }
+
+  private registerVisibilityListener(): boolean {
+    if (typeof window !== 'undefined') {
```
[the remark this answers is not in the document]
schmidt-sebastian: I opted to put this into the EventManager since it is at least partially responsible for handling events from the user (well, only for 'listen' and 'unlisten') and I consider tab visibility user events. I agree that it is not necessarily the best fit, and since it introduces a dependency on the asyncQueue, I updated the PR to add a seperate class (platform_browser/window_event_listener).

packages/firestore/src/core/firestore_client.ts
```diff
@@ -65,7 +71,9 @@ export class FirestoreClient {
   private persistence: Persistence;
   private localStore: LocalStore;
   private remoteStore: RemoteStore;
+  private webStorage: WebStorage;
   private syncEngine: SyncEngine;
+  private ownerId: string = this.generateOwnerId();
```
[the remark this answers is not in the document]
schmidt-sebastian: That works. Thanks.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -379,4 +401,9 @@ export class FirestoreClient {
         return this.syncEngine.runTransaction(updateFunction);
       });
   }
+
+  private generateOwnerId(): string {
```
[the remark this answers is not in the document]
schmidt-sebastian: I removed it, but note that this was just moved from `indexeddb_persistence.ts`

packages/firestore/src/core/sync_engine.ts
```diff
@@ -616,4 +626,9 @@ export class SyncEngine implements RemoteSyncer {
         return this.remoteStore.handleUserChange(user);
       });
   }
+
+  applyVisibilityChange(visibilityState: VisibilityState): void {
```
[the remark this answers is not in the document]
schmidt-sebastian: SyncEngine will eventually need to plumb through all events that lead to changes in the persistence layer. For now, I have removed the dependency, but I feel that I will have to add it back soon.

packages/firestore/src/local/web_storage.ts
schmidt-sebastian: I explicitly avoided naming anything LocalStorage in this code since this will create mighty confusion with LocalStore. `TabNotificationChannel` works for me and PR has been updated.

packages/firestore/src/local/web_storage.ts
schmidt-sebastian: I'll get back to you 2025, but for now I named it`LocalStorageNotificationChannel`.

packages/firestore/src/local/web_storage.ts
schmidt-sebastian: It could, but in order to do this, I have to return the handle from AsyncQueue. We currently don't do this for any of the tasks. Instead, I rely on the fact that AsyncQueue.drain() gets called at the same time as shutdown().

packages/firestore/src/local/web_storage.ts
schmidt-sebastian: Sounds good.

packages/firestore/src/local/web_storage.ts
schmidt-sebastian: I removed this for now, since I no longer call into WebStorage from SyncEngine.

schmidt-sebastian:
> I want to read your doc before reviewing this so that we're on the same page about the implementation goals, etc. But from my brief look, my first high-level feedback is that I'm questioning whether SyncEngine and EventManager should know about browser-specific concepts (WebStorage, visibility events). Ideally that would be encapsulated inside the persistence components, I think? Today, SyncEngine and EventManager have identical functionality across platforms. We may not be able to maintain that while introducing multi-tab, but we should try to, and be explicit where we start deviating (and keep it as isolated and self-contained as possible).
> 
> To be clear though, you probably shouldn't take any action on this feedback until I read your doc and review this for real. So ignore me for now. :-)

I think that we will eventually need to modify SyncEngine to make this work, but this is not necessarily needed at this stage yet. I created a new class for the code in EventManager (as Gil suggested) and will try to keep as much out of SyncEngine as possible.

PR #441, opened 2018-01-16
Adding generics to the Connection API

schmidt-sebastian:
This removes all 'any' violations in the grpc_connection.ts, webchannel_connection.ts, stream_bridge.ts and connection.ts.


packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -135,26 +129,57 @@ export class GrpcConnection implements Connection {
...
-    const stub = this.getStub(token);
-
+  private getRpcCallable<Req, Resp>(
+    rpcName: string
+  ): UnaryRpc<Req, Resp> | ReadableRpc<Req> | DuplexRpc {
     // RPC Methods have the first character lower-cased
     // (e.g. Listen => listen(), BatchGetDocuments => batchGetDocuments()).
     const rpcMethod = rpcName.charAt(0).toLowerCase() + rpcName.slice(1);
-    const rpc = stub[rpcMethod];
+    const rpc = this.cachedStub.stub[rpcMethod];
```
mikelehen: Are you intentionally not calling ensureActiveStub() here? As-is this code isn't obviously correct, since I have to go figure out that some other code makes sure to call ensureActiveStub() before this code is called (I hope!).

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -135,26 +129,57 @@ export class GrpcConnection implements Connection {
...
+    const rpc = this.getRpcCallable<Req, Resp>(rpcName) as DuplexRpc;
+    return rpc.bind(stub);
+  }
 
+  private getStreamingRpc<Req, Resp>(
+    rpcName: string,
+    token: Token | null
+  ): ReadableRpc<Req> {
+    const stub = this.ensureActiveStub(token);
+    const rpc = this.getRpcCallable<Req, Resp>(rpcName) as ReadableRpc<Req>;
```
mikelehen: So part of me likes these typed helper methods, but another part of me sees that:
1) They're exactly the same except for an "as ..." call,
2) They're each only called exactly once.

So I think I'd be inclined to save 20 lines and repetition and just put the "as ..." at each call-site.

packages/firestore/src/remote/stream_bridge.ts
```diff
@@ -63,25 +63,25 @@ export class StreamBridge<I, O> implements Stream<I, O> {
 
   callOnOpen(): void {
     assert(
-      this.wrappedOnOpen !== null,
-      'Cannot call onOpen because no callback ' + 'was set'
+      this.wrappedOnOpen !== undefined,
+      'Cannot call onOpen because no callback was set'
     );
-    this.wrappedOnOpen();
+    this.wrappedOnOpen!();
```
mikelehen: FWIW, I always feel bad about myself when I use `!` and so I write something like:
```
if (this.wrappedOnOpen === undefined) {
  fail('Cannot call onOpen because no callback was set');
} else {
  this.wrappedOnOpen();
}
```

You could also do `return fail(...)` and drop the else block.

If you don't feel bad about yourself though, that's cool.

packages/firestore/src/platform_node/grpc_connection.ts
schmidt-sebastian: I should have called it here as well, but all callers of this method already called it. This was just a helper method of the helper methods, so it seemed safe.

packages/firestore/src/platform_node/grpc_connection.ts
schmidt-sebastian: I think that's a good point - I do have to do a type conversion anyways, so I might as well do it in the callers.

packages/firestore/src/remote/stream_bridge.ts
schmidt-sebastian: This is very confined and follows the existing code, so I am inclined to leave it as is.

packages/firestore/src/platform_node/grpc_connection.ts
mikelehen: Safe yes, but obviously correct, no. :-)

PR #442, opened 2018-01-16
Removing any from LLRBTree

schmidt-sebastian:
Removing `any` from LLRBTree

### DIFF
--- a/packages/firestore/src/util/sorted_map.ts
+++ b/packages/firestore/src/util/sorted_map.ts
@@ -44,7 +44,7 @@ export class SortedMap<K, V> {
   // visible for testing
   root: LLRBNode<K, V> | LLRBEmptyNode<K, V>;
 
-  constructor(public comparator: Comparator<K>, root?: LLRBNode<K, V>) {
+  constructor(public comparator: Comparator<K>, root?: LLRBNode<K, V> | LLRBEmptyNode<K, V>) {
     this.root = root ? root : LLRBNode.EMPTY;
   }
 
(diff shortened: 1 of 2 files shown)


packages/firestore/src/util/sorted_map.ts
```diff
@@ -298,14 +301,14 @@ export class SortedMapIterator<K, V, T> {
     }
   }
 
-  getNext(): T {
+  getNext(): T | Entry<K, V> {
```
mikelehen: This is a pretty weird return type and I was surprised it didn't cause you to have to cast in all of the usages of it, but I _think_ T always ends up being Entry<K, V> (because we don't provide a resultGenerator) and so the type simplifies to just Entry<K, V> and no cast is necessary.

But I think I would just remove resultGenerator entirely and then the return type here is just Entry<K, V>. Android and iOS don't seem to have a resultGenerator anyway...

packages/firestore/src/util/sorted_map.ts
```diff
@@ -440,7 +444,7 @@ export class LLRBNode<K, V> {
 
   // Returns new tree, with the key/value added.
   insert(key: K, value: V, comparator: Comparator<K>): LLRBNode<K, V> {
-    let n = this;
+    let n = this as LLRBNode<K, V>;
```
mikelehen: This tripped me up (I didn't realize that 'this' was a special type)...  I wonder if 
```
let n: LLRBNode<K, V> = this;
```
would be less confusing? Eh, don't care strongly.

packages/firestore/src/util/sorted_map.ts
schmidt-sebastian: I was debating getting rid of resultGenerator, but wasn't certain where this code was coming from and whether we wanted to keep it closer to the original implementation. But I am going to remove it since it will make the types much cleaner.

packages/firestore/src/util/sorted_map.ts
schmidt-sebastian: Oh, yeah, I do think that's less confusing. Thanks!

PR #450, opened 2018-01-18
Implement SWController.getPublicVAPIDKey

pinarx:
Implement getPublicVAPIDKey method for SWController, and complete the TODOs in the unit tests. @gauntface for review.

Again: Firestore tests are failing.

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -17,6 +17,7 @@
 
 import ControllerInterface from './controller-interface';
 import Errors from '../models/errors';
+import FCMDetails from '../models/fcm-details';
 import WorkerPageMessage from '../models/worker-page-message';
 
 const FCM_MSG = 'FCM_MSG';
@@ -359,5 +360,20 @@ export default class SWController extends ControllerInterface {
    */
   getSWRegistration_() {
     return Promise.resolve((self as any).registration);
+  }
+
+  /**
+   * This will return the default VAPID key or the uint8array version of the
+   * public VAPID key provided by the developer.
+   */
+  getPublicVapidKey_(): Promise<Uint8Array> {
+    return this.getSWRegistration_().then(swReg => {
+      return swReg.pushManager.getSubscription().then(subscription => {
+        if (subscription !== null && subscription.options !== null) {
+          return subscription.options.applicationServerKey;
+        }
+        return FCMDetails.DEFAULT_PUBLIC_VAPID_KEY;
+      });
+    });
   }
 }
--- a/packages/messaging/src/controllers/controller-interface.ts
+++ b/packages/messaging/src/controllers/controller-interface.ts
@@ -97,10 +97,14 @@ export default class ControllerInterface {
     tokenDetails: Object,
     swReg: ServiceWorkerRegistration
   ): Promise<string> {
-    const publicVapidKey = this.getPublicVapidKey_();
+    let publicVapidKey: Uint8Array;
     let updatedToken: string;
     let subscription: PushSubscription;
-    return this.getPushSubscription_(swReg, publicVapidKey)
+    return this.getPublicVapidKey_()
+      .then(publicKey => {
+        publicVapidKey = publicKey;
+        return this.getPushSubscription_(swReg, publicVapidKey);
+      })
       .then(pushSubscription => {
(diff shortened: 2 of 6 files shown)


packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -360,4 +361,19 @@ export default class SWController extends ControllerInterface {
...
+   * This will return the default VAPID key or the uint8array version of the
+   * public VAPID key provided by the developer.
+   */
+  getPublicVapidKey_(): Promise<Uint8Array> {
+    return this.getSWRegistration_().then(swReg => {
+      return swReg.pushManager.getSubscription().then(subscription => {
+        if (subscription !== null && subscription.options !== null) {
+          return subscription.options.applicationServerKey;
+        }
+        return FCMDetails.DEFAULT_PUBLIC_VAPID_KEY;
```
gauntface: This feels very risky to me.

I originally intended to use: the vapid model to handle this:

https://github.com/firebase/firebase-js-sdk/blob/messaging-model/packages/messaging/src/models/vapid-details-model.ts#L83

gauntface: If pushsubscriptionchange is triggered, the subscription *might* not be defined after (firefox triggers when permission state changes and there may be a fire when the new subscription can't be renewed). *if* the subscription doesn't exist - you'll use the default key meaning developers vapid keys will no longer be used.

3 options:

1.Use the model
2. Remove the API from the controller interface and only enable it in the window (i.e. we enforce no use of the vapid key in the service worker - this assumes we'll never have to subscribe which I believe should be the case for most of the time.
3. We require developers to set the vapid key in the window and the service worker.

Using the model:
1. Dev uses key, gets a token, it saved to the token model, and then saves to the vapid details.
2. Removes the key, gets a token, update token model and update the vapid details.

pinarx: OK, mostly as a clarification to myself, the issue seems to be that: Currently we *only* allow usePublicVapidKey method in the window, which in itself makes sense - but leads to different behaviour when getToken is called which requires us to retrieve the VAPID public key.

**1) Use the model/DB to store VAPID values:**
- **Keeping getPublicVAPIDKey consistent across SW and Window controller:**
  - If a custom VAPID public key is specified in the WindowController, but we haven't subscribed yet (didn't call get token yet), SWController.getPublicVAPIDKey will return an old or default VAPID key - even if a custom one is specified. So a new VAPID key can only take affect, if we subscribe through the windowController.
  - If we store the VAPID key when useVapidPublicKey is called, we will hit the same issue when the developer decides not to specify the custom VAPID key. Removing useVapidPublicKey will not trigger any calls before subscribing from WindowController.

**2) Removing getPublicVapidKey from SW Interface:** I am not sure about the implications on the getToken flow:
  - onpushsubscriptionchange needs to have access to updateToken (which uses getPublicVapidKey) in the longer run, right? If the underlying pushsubscription changes, we would like to call updateToken. I guess we can trust on the browser to return the correct VAPID key in the push subscription, but sounds a bit risky to not be able to flag this in case of browser issues.
  - In what case do we expect getToken to be called from SWController?

**2) Have the dev set VAPID key in both SW and Window:** This actually sounds like the most consistent behaviour across SW and Window interfaces, but puts the responsibility on the developer. Better if we can avoid it...

pinarx: And the VAPID model version is implemented in here for reference: https://github.com/pinarx/firebase-js-sdk/tree/swGetVAPIDModel

gauntface: @pinarx the vapid model is implemented in the messaging-model branch. https://github.com/firebase/firebase-js-sdk/blob/messaging-model/packages/messaging/src/models/vapid-details-model.ts

pinarx: yes I know=) I based the above branch on that - will create a PR so that it's easier to see the changes...

pinarx: See: https://github.com/firebase/firebase-js-sdk/pull/465. If you prefer to go that way and use VAPID model, I'll close this PR.

gauntface:
@pinarx are you happy to close this now that the other PR is merged?

PR #452, opened 2018-01-19
[Multi-Tab] PSEUDOCODE

schmidt-sebastian:


packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -102,6 +171,81 @@ export class LocalStorageNotificationChannel implements TabNotificationChannel {
...
+    this.rejectedTargetIds[targetId] = { date: new Date(), error };
+  }
+
+  updateQuery(updatedTargetIds: TargetId[]): void {
+    for (const targetId of updatedTargetIds) {
+      this.updatedTargetIds[targetId] = new Date();
+    }
+  }
+
+  private onUpdate(key: string, value: string) {
```
mikelehen: Your pseudocode needs more comments and less code. :-) Who calls this? when?

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -102,6 +171,81 @@ export class LocalStorageNotificationChannel implements TabNotificationChannel {
     this.persistState();
   }
 
+  addMutation(batchId: BatchId): void {
+    this.pendingBatchIds[batchId] = this.instanceId;
```
mikelehen: I assume the intention is that these write to local storage? Again, more comments and less code would be more helpful.

packages/firestore/src/local/tab_notification_channel.ts
schmidt-sebastian: All these methods are supposed to call this.persistState(). Comment added.

schmidt-sebastian: This is the event listener hook called when a key in LocalStorage changes.

PR #454, opened 2018-01-20
Refactor integration/api/ tests to have no src/ dependencies.

mikelehen:
This lets us run these tests (via firebase-js-sdk/integration/firestore/)
against the minified build without risk of accidentally mixing in non-minified
classes (leading to tests misbehaving or skipping testing the minified code
that we were intending to test).

Fixes b/71721842 and b/66946692.

Changes include:
* Split integration/api into api/ and api_internal/
  * Move Idle Timeout tests (that drain the async queue) to api_internal/ and
    remove semi-public db.INTERNAL.drainAsyncQueue() method.
* Remove src/ dependencies from integration/api tests:
  * Split integration/util/helpers.ts into helpers.ts (with no src/
    dependencies) and internal_helpers.ts (with src/ dependencies).
  * Create test/util/promise.ts with Deferred<> implementation to avoid depending
    on src/util/promise.ts.
  * Move EventsAccumulator into its own
    test/integration/util/events_accumulator.ts file.
* Change minified test build (/integration/firestore/) to only include the
  desired test/ code without pulling over any src/ code (protect us from 
  accidental src/ dependencies).

### DIFF
--- a/packages/firestore/test/integration/api/README.md
+++ b/packages/firestore/test/integration/api/README.md
@@ -1,0 +1,4 @@
+All tests in this directory are also run against our (minified)
+firebase-firestore.js build (via 'yarn test:manual' under
+firebase-js-sdk/integration/firestore/) and therefore must not have any
+dependencies on src/ files.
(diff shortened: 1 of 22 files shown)


packages/firestore/test/integration/api/README.md
```diff
@@ -0,0 +1,4 @@
+All tests in this directory are also run against our (minified)
+firebase-firestore.js build (via 'yarn test:manual' under
+firebase-js-sdk/integration/firestore/) and therefore must not have any
+dependencies on src/ files.
```
wilhuff: Add a note here indicating that tests run against the API but otherwise need internal hooks (i.e. can't be written purely against the API) should go in `test/integration/api_internal`?

PR #463, opened 2018-01-23
 GetOptions for controlling offline behaviour

rsgowman:
Add option to allow the user to control where DocumentReference.get()
and Query.get() fetches from. By default, it fetches from the server (if
possible) and falls back to the local cache.  It's now possible to
alternatively fetch from the local cache only, or to fetch from the
server only (though in the server only case, latency compensation is
still enabled).

ts port of https://github.com/firebase/firebase-ios-sdk/pull/655


packages/firestore/src/api/database.ts
```diff
@@ -1566,10 +1574,19 @@ export class Query implements firestore.Query {
...
+  get(options?: firestore.GetOptions): Promise<firestore.QuerySnapshot> {
+    validateBetweenNumberOfArgs('Query.get', arguments, 0, 1);
     return new Promise(
       (resolve: Resolver<firestore.QuerySnapshot>, reject: Rejecter) => {
+        if (options && options.source === 'cache') {
+          this.firestore
+            .ensureClientConfigured()
+            .getDocumentsFromLocalCache(this, {
+              next: (snap: firestore.QuerySnapshot) => resolve(snap),
+              error: reject
```
mikelehen: I think we want to resolve with an empty snapshot here?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -15,6 +15,12 @@
  */
 
 import { CredentialsProvider } from '../api/credentials';
+import {
+  DocumentReference,
+  DocumentSnapshot,
+  Query,
+  QuerySnapshot
+} from '../api/database';
```
mikelehen: I think we want to avoid api/ dependencies here (CredentialsProvider is kind of an exception). The api/ layer should be responsible to converting to/from model types.

FWIW, I think as-written this may be impossible to port to Android, since you can't have circular dependencies in java.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -357,6 +367,57 @@ export class FirestoreClient {
...
+          observer.next(
+            new DocumentSnapshot(doc.firestore, doc._key, maybeDoc, true)
+          );
+        } else {
+          observer.error(
+            new Error(
+              'Failed to get document from cache. (However, this document may ' +
+                "exist on the server. Run again without setting 'source' in " +
+                'the GetOptions to attempt to retrieve the document from the ' +
+                'server.)'
```
mikelehen: We have a FirestoreError class we use for all of our public-facing errors.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -357,6 +367,57 @@ export class FirestoreClient {
     });
   }
 
+  getDocumentFromLocalCache(
+    doc: DocumentReference,
```
mikelehen: Per my comment elsewhere, this should probably take a DocumentKey.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -357,6 +367,57 @@ export class FirestoreClient {
...
+    observer: Observer<DocumentSnapshot>
+  ): void {
+    this.asyncQueue.schedule(() => {
+      return this.localStore.readDocument(doc._key).then(maybeDoc => {
+        if (maybeDoc instanceof Document) {
+          observer.next(
+            new DocumentSnapshot(doc.firestore, doc._key, maybeDoc, true)
```
mikelehen: Per my comments elsewhere, this should return Document instead of creating a DocumentSnapshot.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -357,6 +367,57 @@ export class FirestoreClient {
...
+
+        const view = new View(query._query, remoteKeys);
+        const viewDocChanges: ViewDocumentChanges = view.computeDocChanges(
+          docs
+        );
+        const viewChange: ViewChange = view.applyChanges(viewDocChanges);
+        assert(
+          viewChange.limboChanges.length === 0,
+          'View returned limbo documents during local-only query execution.'
+        );
```
mikelehen: So I was actually surprised that this assert held because remoteKeys is empty, and so all the documents should be in limbo (they're in the local cache but not in the remote results). But it turns out we don't calculate limbo documents until you call applyChanges() with a TargetChange.

So this assert _holds_ but not for the reason I would have expected. And if there _were_ limbo documents, that wouldn't be a problem for us.  So I'd argue this assert is confusing and not protecting us from anything, so I'd just remove it.

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -0,0 +1,612 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect } from 'chai';
+import { apiDescribe, withTestDoc, withTestCollection } from '../util/helpers';
+
+apiDescribe('GetOptions', persistence => {
```
mikelehen: Thanks for all the tests. I love good test coverage! One thing I don't love though is boilerplate and lots of code to port though...  So you might consider if:
1) Any of these tests are redundant
2) If you could simplify them (and reduce boilerplate) with helper method(s).
3) If you could get equivalent coverage but with less boilerplate by testing at a lower-level. [this probably doesn't apply here, but in general we try to do our more exhaustive testing as "unit" tests which run much more quickly and then only sanity-checking at the API layer]

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -0,0 +1,612 @@
...
+    };
+    return withTestCollection(persistence, initialDocs, colRef => {
+      return colRef.get().then(qrySnap => {
+        expect(qrySnap.metadata.fromCache).to.be.false;
+        expect(qrySnap.metadata.hasPendingWrites).to.be.false;
+        expect(qrySnap.docChanges.length).to.equal(3);
+        let docsData = {};
+        qrySnap.forEach(doc => {
+          docsData[doc.id] = doc.data();
+        });
```
mikelehen: We have a toDataArray() helper that does something similar already. Since it stores in an array, it drops the .id, but preserves ordering (which you lose here).

I guess it is convenient to be able to directly `expect(...).to.deep.equal(initialDocs)` though.  So maybe keep this as-is, but consider creating a .toDataMap() helper since you do this in multiple tests?

packages/firestore/src/api/database.ts
mikelehen: Actually, it looks like as-implemented, error can't be called so this is fine. I think this will be clearer if we implement it with a Promise return rather than Observer in any case...

packages/firestore/src/core/firestore_client.ts
mikelehen: I am a big fan of commenting non-obvious literals in function calls, e.g.:
```
new DocumentSnapshot(doc.firestore, doc._key, maybeDoc, /*fromCache=*/true)
```
We do this in most places (though probably not quite 100%).

Also: It occurs to me that it's possible that the user has an ongoing listener for this document already and so it's guaranteed to be up-to-date, and technically isFromCache should be false in that case. But to make that work, we'd need to route this into the SyncEngine and detect that there's an active View for this document query already, and use the snapshot from that View. We could do this but I'm not sure it's worth the trouble, so I'm fine with keeping this as-is for now. We could consider doing it later.

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -0,0 +1,612 @@
...
+      doc2: { key2: 'value2' },
+      doc3: { key3: 'value3' }
+    };
+    return withTestCollection(persistence, initialDocs, colRef => {
+      // force local cache of these
+      return (
+        colRef
+          .get()
+          // now go offine. Note that if persistence is disabled, this will cause
+          // the initialDocs to be garbage collected.
```
mikelehen: I see that this (and the above test) combat the GC'ing by doing local .set()s instead...  But this means that:
1. The initialDocs don't really matter (the test would still pass without them I believe).
2. This is really testing latency compensation instead of the cache.

Instead I think we should either:
1. Disable the test when persistence is not enabled (`if (!persistence) { return; }` or whatever)
2. Register an onSnapshot listener that includes the data. This will hold them in the cache for the duration of the listener.

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -0,0 +1,612 @@
...
+  });
+
+  it('get document while online cache only', () => {
+    const initialData = { key: 'value' };
+    return withTestDoc(persistence, docRef => {
+      return docRef
+        .set(initialData)
+        .then(() => docRef.get({ source: 'cache' }))
+        .then(doc => {
+          if (persistence) {
```
mikelehen: As mentioned earlier, you could register a listener to control whether the document is in the cache, rather than having conditional test logic.

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -0,0 +1,612 @@
...
+        .then(doc => {
+          expect(doc.exists).to.be.true;
+          expect(doc.metadata.fromCache).to.be.true;
+          expect(doc.metadata.hasPendingWrites).to.be.true;
+          expect(doc.data()).to.deep.equal(newData);
+        });
+    });
+  });
+
+  it('get collection while offline cache only', () => {
```
mikelehen: FWIW, to make it extra obvious what's being tested, perhaps say "with source=cache" instead of "cache only"

packages/firestore/src/api/database.ts
```diff
@@ -956,10 +956,18 @@ export class DocumentReference implements firestore.DocumentReference {
...
     return new Promise(
       (resolve: Resolver<firestore.DocumentSnapshot>, reject: Rejecter) => {
+        if (options && options.source === 'cache') {
+          this._firestoreClient.getDocumentFromLocalCache(this, {
+            next: (snap: firestore.DocumentSnapshot) => resolve(snap),
+            error: reject
+          });
+          return;
+        }
+
```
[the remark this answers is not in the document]
mikelehen: And if I'm not confused, please figure out why the tests didn't catch this. :-)

packages/firestore/src/core/firestore_client.ts
wilhuff: It's actually the EventManager that does the initial snapshot delivery if you relisten:

https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/event_manager.ts#L79

packages/firestore/src/core/firestore_client.ts
```diff
@@ -357,6 +367,57 @@ export class FirestoreClient {
...
+
+  getDocumentsFromLocalCache(
+    query: Query,
+    observer: Observer<QuerySnapshot>
+  ): void {
+    this.asyncQueue.schedule(() => {
+      return this.localStore.executeQuery(query._query).then(docs => {
+        const remoteKeys: DocumentKeySet = documentKeySet();
+
+        const view = new View(query._query, remoteKeys);
```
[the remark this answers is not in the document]
wilhuff: I was the one that recommended we use the view to compute the snapshot we're sending because they're hard to get right, considering the changes and metadata bits that need to be just so.

I still think that's a good idea: though your idea of defering to the sync engine (or event manager based on which is actually the right place to do this) to pick up the existing snapshot would be worthwhile if it's not too crazy.

It would also have the effect of keeping the FirestoreClient mostly logic free, which is important because it's fiendishly hard to test (today).

mohshraim:
any chance to have this feature in near beta?
i am watching this pull daily :)
this will give us a full control for better user experience..
thanks

rsgowman:
Hi @mohshraim,

This feature is currently scheduled for an internal api review. (We want to make sure we get the api right before committing ourselves to it.) Once that's done, we should be able to get this out fairly quickly.

In the meantime, depending on your use case, you may be able to get similar behaviour by issuing a listen request rather than a get via `DocumentReference.onSnapshot()` (or `CollectionReference.onSnapshot()`) If you set the `includeMetadataChanges` option to true, your callback should be executed twice: the first time with the local data (if any), and the second time with the server data. (`includeMetadataChanges` ensures that your callback will be called for the server data, even if the data is the same in the local cache and the server.)

https://firebase.google.com/docs/reference/js/firebase.firestore.DocumentReference#onSnapshot

For instance, lets assume you suspect the local cached data is stale, and you *only* want to display the data from the server (rather than displaying stale data briefly while waiting for server data). Then you might do something like this (untested):

```
docref.onSnapshot( {includeMetadataChanges:true}, function(doc) {
  if (doc.metadata.fromCache) {
    // possibly stale cache data that we want to ignore
    return;
  } else {
    // up-to-date server data that we should display
    displayResultsToUser(doc);
  }
});
```

mohshraim:
Hi @rsgowman

Thanks a lot for all the work and detailed response.

Almost read the firestore documentation 20 times. it's my favourite..
My case i have many users will login to system many times every day(20-50 time); they have shared item collection that changed once daily and contain around 10K doc. so my target is minimize the read count and bandwidth consumption from server to read it from cached data. and get fresh data a according to some conditions 2-3 times daily per user.. 
This feature according to my scenario calculation will save +90% bandwidth daily.. we was looking to save the data on another IndexedDB and read from it but we don't want to reinvent the wheel!

Waiting this amazing feature.
Thanks

mikelehen:
Please assign back to me if this is ready for another pass. :-)

mikelehen:
(oh, you have merge conflicts though, so please resolve them first)

packages/firestore/src/api/database.ts
```diff
@@ -1000,43 +1000,91 @@ export class DocumentReference implements firestore.DocumentReference {
...
+              );
+            }, reject);
+        } else {
+          this.getViaSnapshotListener_(resolve, reject, options);
+        }
+      }
+    );
+  }
+
+  private getViaSnapshotListener_(
```
mikelehen: I don't think we usually add `_` to private methods.

We would ideally not use `_` as a naming convention at all and just rely on private/public modifiers, etc but with our current build setup, any properties or public methods on classes end up showing up on the object at runtime and so end users could potentially discover them.  So for any properties / public methods we don't want end users to see / use (because they're not part of our official API), we add a `_` suffix which triggers the minification step in our build to obfuscate the name: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/webpack.config.js#L48

Anyway, private methods are completely hidden at runtime based on how TypeScript implements private methods and so we don't need / want to use the `_` suffix (here and on the other getViaSnapshotListener_ method).

packages/firestore/src/core/firestore_client.ts
```diff
@@ -41,7 +49,7 @@ import { debug } from '../util/log';
 import { Deferred } from '../util/promise';
 
 import { DatabaseId, DatabaseInfo } from './database_info';
-import { Query } from './query';
+import { Query as InternalQuery } from './query';
```
mikelehen: I think this type alias is no longer necessary? You can just import it as Query?

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -197,6 +207,28 @@ export function withTestDoc(
   });
 }
 
+// TODO(rsgowman): Modify withTestDoc to take in (an optional) initialData and
+// fix existing usages of it. Then delete this function. This makes withTestDoc
+// more analogous to withTestCollection and eliminates the pattern of
+// `withTestDoc(..., docRef => { docRef.set(initialData) ...});` that otherwise is
+// quite common.
```
mikelehen: Why is this a TODO rather than just doing it? :-)

packages/firestore-types/index.d.ts
```diff
@@ -976,9 +1009,15 @@ export class Query {
...
+   * Note: By default, get() attempts to provide up-to-date data when possible
+   * by waiting for data from the server, but it may return cached data or fail
+   * if you are offline and the server cannot be reached. This behavior can be
+   * altered via the `GetOptions` parameter.
+   *
+   * @param options An object to configure the get behavior.
    * @return A Promise that will be resolved with the results of the Query.
    */
-  get(): Promise<QuerySnapshot>;
+  get(options?: GetOptions): Promise<QuerySnapshot>;
```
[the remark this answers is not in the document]
mikelehen: Oh!  Actually one of those may have just gone away (yay): https://github.com/firebase/firebase-js-sdk/pull/653/commits/c198f9a62474eb90853b33f59098c08e5daf3074

packages/firestore/test/integration/util/helpers.ts
rsgowman: Because fixing existing uses of it seems big(ish) and otherwise orthogonal to this PR. But since you asked so nicely: https://github.com/firebase/firebase-js-sdk/pull/675

mohshraim:
Hi @rsgowman

I would like to ask if this will be available in next beta? 
or still need more internal API review?
 
thanks for great work, amazing feature. 
and thanks @mikelehen for super fast reviews..

packages/firestore/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   `FirestoreSettings` to `true`. Note that the current behavior
   (`DocumentSnapshot`s returning JS Date objects) will be removed in a future
   release. `Timestamp` supports higher precision than JS Date.
+- [feature] Added ability to control whether DocumentReference.get() and
+  Query.get() should fetch from server only, cache only, or attempt server and
+  fall back to the cache (which was the only option previously, and is now the
+  default.)
```
mikelehen: super nit: period should go after paren, I think?

packages/firestore/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   `FirestoreSettings` to `true`. Note that the current behavior
   (`DocumentSnapshot`s returning JS Date objects) will be removed in a future
   release. `Timestamp` supports higher precision than JS Date.
+- [feature] Added ability to control whether DocumentReference.get() and
+  Query.get() should fetch from server only, cache only, or attempt server and
```
mikelehen: To make this maximally helpful (especially since there's generally no statement completion in JS and docs will probably not be updated immediately), I'd probably say:

... should fetch from server only (by passing `{ source: 'server' }`, cache only (by passing `{ source: 'cache' }`), or attempt ...

rsgowman: Done.

packages/firestore/CHANGELOG.md
rsgowman: Done.

packages/firestore/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
   `FirestoreSettings` to `true`. Note that the current behavior
   (`DocumentSnapshot`s returning JS Date objects) will be removed in a future
   release. `Timestamp` supports higher precision than JS Date.
+- [feature] Added ability to control whether DocumentReference.get() and
+  Query.get() should fetch from server only, (by passing { source: 'server' },
```
mikelehen: missing close paren after "{ source: 'server' }" :)

rsgowman: Done. (sigh).

rsgowman:
@mohshraim: I think this is now done. :)

If you want to give it a try immediately, you should be able to check out the code from master. Otherwise, it should be available in the next release. If you have any feedback on this feature, either positive or negative, please send it in; thanks!

PR #465, opened 2018-01-24
Implement SWController getPublicVAPIDKey using VAPID model

pinarx:


gauntface:
Failures are database not messaging so going to merge this.

PR #468, opened 2018-01-25
Add VAPID key validation when getToken is called.

pinarx:


pinarx:
@jshcrowthe reverted yarn-lock changes, seems to have escaped during rebasing.

pinarx:
@gauntface failures are DB related as usual - FYI.

pinarx:
PTAL @gauntface @jshcrowthe

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -84,18 +85,56 @@ export default class ControllerInterface {
...
       .then(tokenDetails => {
         if (tokenDetails) {
-          // TODO Validate the details are still accurate
-          const now = Date.now();
-          if (now < tokenDetails['createTime'] + TOKEN_EXPIRATION_MILLIS) {
-            return tokenDetails['fcmToken'];
-          } else {
-            return this.updateToken(tokenDetails, swReg);
-          }
+          return this.manageExistingToken(tokenDetails, swReg);
```
gauntface: Would it make more sense to rename this to `this.validateExistingToken()` and have the function return the token details OR return null and then call getNewToken() in the next if statement?

pinarx: it's a bit more complicated as there are 2 flows in manageExistingTokenDetails:
1) If token details have changed (VAPID key change): Delete existing details and get a new token.
2) If the token details have not changed, but it's been more than 7 days since token creation, call updated token.

I can break up manageExistingToken into 2 methods and call from getToken (isTokenStillValid, hasTokenExpired) - but imho that makes it harder to understand the flow.

pinarx:
@gauntface I still don't have write access as we're working on a branch off of master, and the .github file is not synced=) Would you mind merging?

PR #473, opened 2018-01-29
@firebase/logger

jshcrowthe:
This PR has two parts, 1) the `@firebase/logger` module and 2) the integrations with `@firebase/database` and `@firebase/firestore`. 

## `@firebase/logger`

This module exports a class called `Logger` which supports the following 5 log methods:

- `debug`
- `log`
- `info`
- `warn`
- `error`

### Log Handlers

Each instance of logger, will, by default, log to the corresponding `console` method (i.e. `console.debug`, `console.log`, `console.info`, `console.warn`, `console.error`), however the logHandling itself is customizable by passing a new `logHandler` function to a `Logger` instance.

Each log handler has the following signature:

```typescript
export type LogHandler = (
  logType: LogLevel,
  currentLogLevel: LogLevel,
  ...args: any[]
) => void;
```

The handler is passed: 
- the intended log type (i.e. did I call the `log` method, or the `error` method of my Logger instance?)
- the current log level (i.e. `LogLevel.VERBOSE`, or `LogLevel.ERROR`, etc. This will be discussed in more detail below)
- An array of arguments to actually log

The default handler, respects the passed log level, and if it is a valid log, passes it on to the appropriate `console` stream.

You can define other handlers on an as needed basis.

### Log Levels

We support 5 "active" levels and a "silent" mode for our logs. The full log set we support will be:

```
export enum LogLevel {
  DEBUG,
  VERBOSE,
  INFO,
  WARN,
  ERROR,
  SILENT
}
```

All of the log levels besides `LogLevel.SILENT` map to a console value and should be used instead of the `console` functions (to allow us to centralize logging for the SDK). `LogLevel.SILENT`, by default, disables the logs (however each `LogHandler` can decide what it wants to do with this level).

### `@firebase/database`

_See https://github.com/firebase/firebase-js-sdk/commit/f5ad4922693a8f490d0f8bf88f148bde6d116173 for detailed diff_

I replaced the logging utilities found in https://github.com/firebase/firebase-js-sdk/blob/logger/packages/database/src/core/util/util.ts, these are used throughout the SDK which quickly propagated the change to the rest of the database SDK. We have not yet exposed the log levels here.

All of the `enableLogging` functionality remained unchanged.

### `@firebase/firestore`

_See https://github.com/firebase/firebase-js-sdk/commit/4e59a63f1d73a20da67d48d48a60b9efd3b14848 for detailed diff_

I replaced the logging utilities found in https://github.com/firebase/firebase-js-sdk/blob/logger/packages/firestore/src/util/log.ts. Firestore uses a subset of the log types we intend to support and as such a small map was required.


packages/database/src/core/PersistentConnection.ts
```diff
@@ -935,8 +935,8 @@ export class PersistentConnection extends ServerActions {
     if (this.securityDebugCallback_) {
       this.securityDebugCallback_(body);
     } else {
-      if ('msg' in body && typeof console !== 'undefined') {
-        console.log('FIREBASE: ' + body['msg'].replace('\n', '\nFIREBASE: '));
+      if ('msg' in body) {
+        log('FIREBASE: ' + body['msg'].replace('\n', '\nFIREBASE: '));
```
mikelehen: I believe this was explicitly meant to be logged without turning on logging.  Essentially you can use a special auth token and the backend will send back debug messages that get automatically logged to the console.

packages/database/src/core/Repo.ts
```diff
@@ -653,7 +653,7 @@ export class Repo {
     forEach(stats, (stat: string, value: any) => {
       // pad stat names to be the same length (plus 2 extra spaces).
       for (let i = stat.length; i < longestName + 2; i++) stat += ' ';
-      console.log(stat + value);
+      log(stat + value);
```
mikelehen: Again, this is actually meant to be logged to the console, without enabling any logging settings.

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
 import { SDK_VERSION } from '../core/version';
 import { AnyJs } from './misc';
 import { PlatformSupport } from '../platform/platform';
+import { Logger, LogLevel as _LogLevel } from '@firebase/logger';
+
+const client = new Logger();
```
mikelehen: client => firebaseLogger? (or at least logClient or similar)

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
 import { SDK_VERSION } from '../core/version';
 import { AnyJs } from './misc';
 import { PlatformSupport } from '../platform/platform';
+import { Logger, LogLevel as _LogLevel } from '@firebase/logger';
```
mikelehen: _LogLevel => FirebaseLogLevel ?  (maybe Logger => FirebaseLogger too for symmetry)

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
...
+
+  if (client.logLevel === _LogLevel.SILENT) {
+    return LogLevel.SILENT;
+  }
+
+  return LogLevel.ERROR;
 }
 export function setLogLevel(newLevel: LogLevel): void {
-  logLevel = newLevel;
+  client.logLevel = newLevel;
```
mikelehen: Shouldn't this do the inverse LogLevel => _LogLevel mapping as getLogLevel()?

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
...
 }
 
 export function debug(tag: string, msg: string, ...obj: AnyJs[]): void {
-  if (logLevel <= LogLevel.DEBUG) {
-    const time = new Date().toISOString();
-    const args = obj.map(argToString);
-    console.log(`Firestore (${SDK_VERSION}) ${time} [${tag}]: ${msg}`, ...args);
-  }
+  const time = new Date().toISOString();
+  const args = obj.map(argToString);
```
mikelehen: This is semi-expensive and should be avoided if logging isn't enabled.

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
...
+  client.log(`Firestore (${SDK_VERSION}) ${time} [${tag}]: ${msg}`, ...args);
```
mikelehen: Isn't time included by the logger client? (It was in your proposal, I believe)

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
...
-let logLevel = LogLevel.ERROR;
-
 // Helper methods are needed because variables can't be exported as read/write
 export function getLogLevel(): LogLevel {
-  return logLevel;
+  if (client.logLevel === _LogLevel.DEBUG) {
+    return LogLevel.DEBUG;
+  }
+
+  if (client.logLevel === _LogLevel.SILENT) {
```
mikelehen: How about else if or a switch statement?

packages/firestore/src/util/log.ts
```diff
@@ -19,37 +19,42 @@
...
+    return LogLevel.SILENT;
+  }
+
+  return LogLevel.ERROR;
```
mikelehen: If you do if / else if, can you put this in an "else"? We avoid early-return.

packages/logger/README.md
```diff
@@ -0,0 +1,13 @@
...
+Firebase JS SDK. It will give you a couple things OOTB:
+
+- **Typescript Support:** Your code should be written in TS to be consistent
+  with the rest of the SDK.
+- **Isomorphic Testing/Coverage:** Your tests will be run in both Node.js and
+  Browser environments and coverage from both, collected.
+- **Links to all of the other packages:** Should your new package need to take
+  a dependency on any of the other packages in this monorepo (e.g.
+  `@firebase/app`, `@firebase/util`, etc), all those dependencies are already
+  set up, you can just remove the ones you don't need.
```
mikelehen: Did you intend to leave this template as-is?

packages/logger/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is the file that people using Node.js will actually import. You should
+ * only include this file if you have something specific about your
+ * implementation that mandates having a separate entrypoint. Otherwise you can
+ * just use index.ts
+ */
```
mikelehen: I don't understand this comment. Why will "people using Node.js" import this?

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is the file that people using Node.js will actually import. You should
+ * only include this file if you have something specific about your
+ * implementation that mandates having a separate entrypoint. Otherwise you can
+ * just use index.ts
+ */
```
mikelehen: Confusing comment again.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+ * A container for all of the Logger instances
+ */
+export const instances: Logger[] = [];
+
+/**
+ * The JS SDK supports 5 log levels and also allows a user the ability to
+ * silence the logs altogether.
+ *
+ * The order is a follows:
+ * DEBUG < VERBOSE < INFO < WARN < ERROR < SILENT
```
mikelehen: I think VERBOSE is supposed to be more verbose than DEBUG?  Honestly, I would just drop VERBOSE unless we find ourselves needing it for some reason.  If the meaning wasn't intuitive for you, it won't be intuitive for users either.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * A container for the default log level
+ */
+let defaultLogLevel: LogLevel = LogLevel.WARN;
+
+/**
+ * A function to set the default log level externally
+ */
+export function setDefaultLogLevel(val: LogLevel) {
```
mikelehen: What is this function for?  Setting the default seems nonintuitive (the default is what you get if you _don't_ set the log level).

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+  if (!(val in LogLevel)) {
+    throw new TypeError('Attempted to Invalid value assigned to `logLevel`');
```
mikelehen: grammar malfunction (here and below)

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+      console.info(...args);
+      break;
+    case LogLevel.WARN:
+      console.warn(...args);
+      break;
+    case LogLevel.ERROR:
+      console.error(...args);
+      break;
+    default:
+      console.debug(...args);
```
mikelehen: Per https://developer.mozilla.org/en-US/docs/Web/API/Console: Starting with Chromium 58 this method only appears in Chromium browser consoles when level "Verbose" is selected.

Which means we don't want to do it (developers shouldn't have to opt into logs twice to get them).

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+  logType: LogLevel,
+  currentLevel: LogLevel,
+  ...args: any[]
+) => {
+  if (logType < currentLevel) return;
+  switch (logType) {
+    case LogLevel.SILENT:
+      return;
+    case LogLevel.VERBOSE:
+      console.log(...args);
```
mikelehen: The format in your proposal was:

[<timestamp>]  <name>@<version>: <message>

Are we not doing that?

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+let defaultLogLevel: LogLevel = LogLevel.WARN;
+
+/**
+ * A function to set the default log level externally
+ */
+export function setDefaultLogLevel(val: LogLevel) {
+  if (!(val in LogLevel)) {
+    throw new TypeError('Attempted to Invalid value assigned to `logLevel`');
+  }
+  defaultLogLevel = val;
```
mikelehen: Hrm...  So if you call setDefaultLogLevel(), which I assume will be the public API we eventually expose to users, shouldn't we notify all of the existing instances?  As a user I should be able to enable / disable logging at will.

I'm fuzzy on how the global and per-component log levels are meant to interact. E.g. what happens if I set the default one, then set a per-component one, then set the default one again?

Honestly I would just simplify and only have the default / global log level for now.  But if you keep both, please make sure they interact in a sane way (taking into consideration how you expect to expose the API to the end user eventually).

packages/firestore/src/util/log.ts
jshcrowthe: Technically we don't have to because the 3 log types that Firestore supports are supported by the Logger package itself. This just works.

packages/logger/README.md
jshcrowthe: Nope, thank you!

packages/logger/index.ts
jshcrowthe: This is a leftover from the template, I've pulled this comment from both places as it's not relevant. In principle tho, you can ship a different binary to Node.js users than you do to Browser users, I have separate entrypoints specified for that.

packages/logger/src/logger.ts
jshcrowthe: Removed.

packages/logger/src/logger.ts
jshcrowthe: Great catch on this, I forgot it entirely. I'll rev this appropriately

packages/logger/src/logger.ts
jshcrowthe: I need to get like a grammar checker for my code haha.

packages/firestore/src/util/log.ts
jshcrowthe: This was accidentally omitted. I'll get it fixed up!

packages/firestore/src/util/log.ts
jshcrowthe: I'll refactor this to an `if-else-if-else`, I like to use early returns.

packages/firestore/src/util/log.ts
mikelehen: Ahh, and TypeScript is validating that I guess (if you added a different entry to Firestore's LogLevel, typescript would complain?).

If so, cool though I'd still probably add a comment to that effect: 

```
// Since Firestore's LogLevel enum is a pure subset of Firebase's LogLevel, we can return it directly.
```

packages/firestore/src/util/log.ts
schmidt-sebastian: `logClient` is the name used in `packages/database/src/core/util/util.ts`.

packages/database/src/core/util/util.ts
```diff
@@ -179,14 +167,8 @@ export const fatal = function(...var_args: string[]) {
  * @param {...*} var_args
```
schmidt-sebastian: This comment is for the hidden block just above (line 173): I feel that we should also log 'fatal' messages to the user-specified log functions.

packages/logger/src/logger.ts
schmidt-sebastian: Also would be nice to include the value here.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * We allow users the ability to pass their own log handler. We will pass the
+ * type of log, the current log level, and any other arguments passed (i.e. the
+ * messages that the user wants to log) to this function.
+ */
+export type LogHandler = (
+  logType: LogLevel,
+  currentLogLevel: LogLevel,
```
schmidt-sebastian: Can you make sure these names match?

packages/logger/src/logger.ts
schmidt-sebastian: I would go even further and suggest that we only need to set the default log level for the Firebase logger. If a user specifies their own logger, they can strip out log messages as they deem necessary.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+     */
+    instances.push(this);
+  }
+
+  /**
+   * The log level of the given logger. Though all of the log levels can be
+   * centrally set, each logger can be set individually if it desires.
+   */
+  private _logLevel = defaultLogLevel;
+  get logLevel() {
```
schmidt-sebastian: Please add typings here and below.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+    return this._logLevel;
+  }
+  set logLevel(val) {
+    if (!(val in LogLevel)) {
+      throw new TypeError('Attempted to Invalid value assigned to `logLevel`');
```
schmidt-sebastian: Make sure this message is also updated.

packages/logger/test/logger.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  const spies = {
+    debugSpy: null,
+    logSpy: null,
+    infoSpy: null,
+    warnSpy: null,
+    errorSpy: null
+  };
+  /**
+   * Before each test, instantiate a new instance of Logger and set the log
+   * level so that all
```
schmidt-sebastian: Please finish your thought :)

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+    this._logLevel = val;
+  }
+
+  /**
+   * The log handler for the current logger instance. This can be set to any
+   * function value, though this should not be needed the vast majority of the
+   * time
```
schmidt-sebastian: I would remove the second sentence.

packages/logger/src/logger.ts
jshcrowthe: I'll refactor this to a `console.log`

packages/logger/src/logger.ts
jshcrowthe: @salqadri has a doc internally on the different levels and their use cases. There is a clear delineation between these

packages/database/src/core/PersistentConnection.ts
jshcrowthe: SGTM

packages/database/src/core/Repo.ts
jshcrowthe: SGTM

packages/firestore/src/util/log.ts
jshcrowthe: SGTM

packages/firestore/src/util/log.ts
jshcrowthe: SGTM

packages/firestore/src/util/log.ts
jshcrowthe: Done

packages/logger/src/logger.ts
mikelehen: AFAICT, his doc (and the iOS implementation [here](https://github.com/firebase/firebase-ios-sdk/blob/14ea068f5fd03a658017f8472a4078a727fabc3a/Firebase/Core/Public/FIRLoggerLevel.h#L20) has: DEBUG < INFO < NOTICE < WARNING < ERROR.  So to match that we would get rid of VERBOSE and add NOTICE.  Personally I don't think NOTICE adds any value though and just makes it harder to decide what level to use (both for us and our users), so I'd still be inclined to just drop VERBOSE and have: DEBUG < INFO < WARN < ERROR < SILENT.

jshcrowthe: I've renamed the `VERBOSE` level to `LOG` (to match the console function) and have updated the doc as well. I think there is still clear use cases that we miss if we don't have `LOG` (e.g. The use case where it is **not** debug information, but the user should know, and there isn't a concrete action they can take). I'm going to leave as is.

packages/logger/src/logger.ts
jshcrowthe: It was originally to help with some testing and I thought it might be nice to allow people to set the default and have it persist. Per feedback here, I've removed it.

packages/logger/src/logger.ts
jshcrowthe: I've tried to make it so that each log client can set their own log levels (e.g. Database and Firestore, currently, can't rely on a single log level as they are handled separately). Though there are other ways you could solve this, I've opted to just keep the LogLevel, local to the Logger instance.

I've removed the global default thing, so that is gone.

There is still a top level "set-all" type of functionality. This will allow us to do global level sets across all `Logger` instances which we will need going forward.

packages/logger/src/logger.ts
jshcrowthe: I've refactored this a bit PTAL

packages/logger/src/logger.ts
jshcrowthe: Done

packages/logger/test/logger.test.ts
jshcrowthe: Done :)

packages/database/src/core/util/util.ts
jshcrowthe: Done

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * The default log handler will forward DEBUG, LOG, INFO, WARN, and ERROR
+ * messages on to their corresponding console counterparts (if the log method
+ * is supported by the current log level)
+ */
+const defaultLogHandler: LogHandler = (instance, logType, ...args) => {
+  if (logType < instance.logLevel) return;
+  const now = new Date();
+  switch (logType) {
+    case LogLevel.SILENT:
```
mikelehen: We should never generate log messages with level SILENT, right?

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,147 @@
...
+    case LogLevel.INFO:
+      console.info(`[${now}]  ${instance.name}:`, ...args);
+      break;
+    case LogLevel.WARN:
+      console.warn(`[${now}]  ${instance.name}:`, ...args);
+      break;
+    case LogLevel.ERROR:
+      console.error(`[${now}]  ${instance.name}:`, ...args);
+      break;
+    default:
```
mikelehen: 1) You left out DEBUG
2) Rather than default to console.log(), I'd throw an exception. That way you'll notice if you leave out a log level.  If for some reason you were intentionally handling DEBUG via the `default:` case, then a comment is warranted to explain your intent.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,147 @@
...
+  DEBUG,
+  LOG,
+  INFO,
+  WARN,
+  ERROR,
+  SILENT
+}
+
+/**
+ * A container for the default log level
```
mikelehen: Should this just be "The Default Log Level" (I'm not sure what 'container' refers to)

packages/logger/test/logger.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  afterEach(() => {
+    spies.logSpy.restore();
+    spies.infoSpy.restore();
+    spies.warnSpy.restore();
+    spies.errorSpy.restore();
+  });
+
+  function testLog(message, channel, shouldLog) {
+    /**
+     * `debug` logs also go through log channel
```
mikelehen: This suggests you are currently intentionally merging log and debug messages, but I didn't see this explained anywhere... Did I miss it?

packages/logger/test/logger.test.ts
```diff
@@ -0,0 +1,94 @@
...
+import { spy as Spy } from 'sinon';
+import { Logger, LogLevel } from '../src/logger';
+import { setLogLevel } from '../index';
+import { debug } from 'util';
+
+describe('@firebase/logger', () => {
+  const message = 'Hello there!';
+  let client: Logger;
+  const spies = {
+    debugSpy: null,
```
mikelehen: unused

packages/logger/test/logger.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      setLogLevel(LogLevel.DEBUG);
+    });
+    testLog(message, 'debug', true);
+    testLog(message, 'log', true);
+    testLog(message, 'info', true);
+    testLog(message, 'warn', true);
+    testLog(message, 'error', true);
+  });
+
+  describe('Defaults', () => {
```
mikelehen: How is this test different than the above one other than it's using WARN instead of DEBUG as the log level?  Perhaps it should omit the beforeEach() so that we actually get the default log level automatically?  And then the describe should be something like "Defaults to LogLevel.WARN" ?

packages/firestore/src/util/log.ts
```diff
@@ -19,36 +19,50 @@
...
 
 export function debug(tag: string, msg: string, ...obj: AnyJs[]): void {
-  if (logLevel <= LogLevel.DEBUG) {
+  if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
     const time = new Date().toISOString();
     const args = obj.map(argToString);
-    console.log(`Firestore (${SDK_VERSION}) ${time} [${tag}]: ${msg}`, ...args);
+    logClient.log(
+      `Firestore (${SDK_VERSION}) ${time} [${tag}]: ${msg}`,
+      ...args
```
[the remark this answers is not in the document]
jshcrowthe: Good point, I have pulled this from these logs.

packages/firestore/src/util/log.ts
```diff
@@ -19,36 +19,50 @@
...
+    return LogLevel.SILENT;
+  } else {
+    return LogLevel.ERROR;
+  }
 }
 export function setLogLevel(newLevel: LogLevel): void {
-  logLevel = newLevel;
+  /**
+   * Since Firestore's LogLevel enum is a pure subset of Firebase's LogLevel, we
+   * can return it directly.
```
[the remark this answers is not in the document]
jshcrowthe: I've refactored this to do a manual set. PTAL.

packages/firestore/src/util/log.ts
```diff
@@ -19,36 +19,50 @@
...
+   */
+  logClient.logLevel = newLevel;
 }
 
 export function debug(tag: string, msg: string, ...obj: AnyJs[]): void {
-  if (logLevel <= LogLevel.DEBUG) {
+  if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
```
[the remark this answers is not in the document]
jshcrowthe: Great catch here, refactored the logging call to `logClient.debug`

packages/logger/README.md
```diff
@@ -0,0 +1,39 @@
...
+import { Logger } from '@firebase/logger';
+
+const logClient = new Logger(`@firebase/<COMPONENT>`);
+```
+
+Each `Logger` instance supports 5 log functions each to be used in a specific
+instance:
+
+- `debug`: Internal logs; use this to allow developers to send us their debug
+  logs for us to be able to diagnose an issue.
```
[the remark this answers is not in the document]
jshcrowthe: I've left a comment in `logger.ts` to explain this and I am now intentionally routing `debug` logs to `console.log`

packages/logger/src/logger.ts
jshcrowthe: Done

packages/logger/README.md
```diff
@@ -0,0 +1,39 @@
...
+- `warn`: Use when a product feature may stop functioning correctly; unexpected
+  scenario. Majority of user-facing logs.
+- `error`: Only use when user App would stop functioning correctly - super rare!
+
+## Log Format
+
+Each log will be formatted in the following manner:
+
+```typescript
+`[${new Date()}]  ${this.name}: ${...args}`
```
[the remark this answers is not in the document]
jshcrowthe: Done

packages/logger/src/logger.ts
jshcrowthe: I've added a comment here to explain the rationale. There is value for allowing custom log handlers to handle `SILENT` logs in their own way which is why this is included.

packages/logger/src/logger.ts
jshcrowthe: Reworded.

packages/logger/test/logger.test.ts
jshcrowthe: I pulled the `beforeEach`. Things should be working as expected.

packages/logger/test/logger.test.ts
jshcrowthe: Thanks!

packages/logger/test/logger.test.ts
jshcrowthe: Renamed.

packages/logger/README.md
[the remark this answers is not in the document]
mikelehen: If there's an action item for the user, wouldn't we want the log on by default?  INFO logs are off by default, so that would not be appropriate.

But my bigger problem is that "LOG" is meaningless since these are _all_ logs and I've never come across another logging system that has "LOG" as its own log level so nobody will know what this means or where it fits in the overall ordering. Even other JavaScript loggers (where `log` as a loglevel might almost make sense because of the `console.log` legacy) don't have it...

[winston](https://github.com/winstonjs/winston) [10k stars] - error, warn, info, verbose, debug
[bunyan](https://github.com/trentm/node-bunyan) [5k stars] - fatal, error, warn, info, debug, trace
[loglevel](https://github.com/pimterry/loglevel) [1k stars] - error, warn, info, debug, trace

So I appreciate that you've made up your own level and have strawmen examples to back it up, but I think it would be better to stick with well-known ones.

I think in practice:
1. 90% of customers will stick with the default (warn and error) all the time.
2. 10% of customers may enable debug logging because support asked them to or they found a StackOverflow / forum post that mentions it.
3. 0% of customers will care to distinguish between info/log/debug.
4. Only Firestore and RTDB use these logs and they only have error, warn, and debug. So there's currently no distinction to make anyway.

All that said, I don't really care. I think we should avoid LOG. If you think we need 5 log levels instead of 4 or 3, then please pick from the existing set of log levels that other systems use (notice, verbose, trace, etc.) for your 5th.  Keep in mind we could always add more log levels later if we need them.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+ * messages that the user wants to log) to this function.
+ */
+export type LogHandler = (
+  loggerInstance: Logger,
+  logType: LogLevel,
+  ...args: any[]
+) => void;
+
+/**
+ * The default log handler will forward DEBUG, LOG, INFO, WARN, and ERROR
```
mikelehen: LOG => VERBOSE

packages/logger/README.md
```diff
@@ -0,0 +1,39 @@
+# @firebase/logger
+
+This package serves as the base of all logging in the JS SDK. Any logging that
+is intended to be for firebase end developers, should go through this module.
```
schmidt-sebastian: Remove comma and maybe reword as:

This package serves as the base of all logging in the JS SDK. Any logging that
is intended to be visible to Firebase end developers should go through this module.

packages/logger/README.md
```diff
@@ -0,0 +1,39 @@
...
+Each `Logger` instance supports 5 log functions each to be used in a specific
+instance:
+
+- `debug`: Internal logs; use this to allow developers to send us their debug
+  logs for us to be able to diagnose an issue.
+- `log`: Use to inform your user about things they may need to know.
+- `info`: Use if you have to inform the user about something that they need to
+  take a concrete action on. Once they take that action, the log should go away.
+- `warn`: Use when a product feature may stop functioning correctly; unexpected
+  scenario. Majority of user-facing logs.
```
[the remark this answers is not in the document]
schmidt-sebastian: It'd take out "Majority of user-facing logs.". This is likely only true when the user doesn't enable debug logging.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+
+/**
+ * The JS SDK supports 5 log levels and also allows a user the ability to
+ * silence the logs altogether.
+ *
+ * The order is a follows:
+ * DEBUG < VERBOSE < INFO < WARN < ERROR
+ *
+ * All of the log types above the current log level will be captured (i.e. if
+ * I set the log level to `INFO`, errors will still be logged, but `DEBUG` and
```
schmidt-sebastian: ... "if YOU set the log level" ...

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+     * Capture the current instance for later use
+     */
+    instances.push(this);
+  }
+
+  /**
+   * The log level of the given Logger instance.
+   */
+  private _logLevel = defaultLogLevel;
+  get logLevel() {
```
schmidt-sebastian: Add return type. 

Do we need to add add proper JSDoc to each of these methods? Or will you do this as part of firebase-types?

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+    return this._logLevel;
+  }
+  set logLevel(val) {
```
schmidt-sebastian: Add argument type.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+      throw new TypeError('Invalid value assigned to `logLevel`');
+    }
+    this._logLevel = val;
+  }
+
+  /**
+   * The log handler for the Logger instance.
+   */
+  private _logHandler: LogHandler = defaultLogHandler;
+  get logHandler() {
```
schmidt-sebastian: Add return type.

packages/logger/src/logger.ts
```diff
@@ -0,0 +1,156 @@
...
+    return this._logHandler;
+  }
+  set logHandler(val) {
```
schmidt-sebastian: Add argument type.

packages/logger/README.md
jshcrowthe: Done

packages/logger/src/logger.ts
jshcrowthe: Done

packages/logger/src/logger.ts
jshcrowthe: Done.

No, this module is only internally used for the time being. When we actually surface it to end developers, we'll take care of the typings.

packages/logger/src/logger.ts
jshcrowthe: Done

packages/logger/src/logger.ts
jshcrowthe: Done

packages/logger/src/logger.ts
jshcrowthe: Done

PR #482, opened 2018-02-02
Adds usePublicVapidKey()

gauntface:
This implements a new API that has been discussed before `usePublicVapidKey()` and fixes up the integration tests.

@jshcrowthe I'm not sure if you want the integration enabled those as they take some time to complete.

cc @pinarx


config/karma.base.js
```diff
@@ -107,7 +107,7 @@ const config = {
 
   client: {
     mocha: {
-      timeout: 20000,
+      timeout: 3000,
```
jshcrowthe: I believe this line exists for firestore as well, if we can revert this timeout that'd be great.

gauntface:
@jshcrowthe the karma config has been reset to it's much longer time :)

gauntface:
@jshcrowthe it looks like the failing travis is auth - but it's hard to tell from all the logs.

PR #483, opened 2018-02-02
[Multi-Tab] Adding basic functionality for TabNotificationChannel

schmidt-sebastian:
This PR adds the InstanceState manipulation to the TabNotificationChannel. This is the basic functionality needed for secondary tabs as per go/multi-tab.

Still missing are:
- Notifications for newly added instances.
- Persisting of mutation batch and query target state (as performed by the primary)


packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -27,7 +38,7 @@ import { JsonProtoSerializer } from '../../../src/remote/serializer';
 export async function testIndexedDbPersistence(): Promise<
   IndexedDbPersistence
 > {
-  const prefix = 'PersistenceTestHelpers/';
+  const prefix = '${testPersistencePrefix}/';
```
mikelehen: string interpolation fail?

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
   await persistence.start();
   return persistence;
 }
+
+/**
+ * Creates LocalStorageNotificationChannel instance for testing, destroying any
+ * previous contents if they existed.
+ */
+export async function testLocalStorageNotificationChannel(
+  ownerId: string,
```
mikelehen: instanceId?

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+  existingMutationBatches: BatchId[],
+  existingQueryTargets: TargetId[]
```
mikelehen: BatchIds, TargetIds

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+): Promise<TabNotificationChannel> {
+  window.localStorage.clear();
```
mikelehen: I wonder if this is dangerous to do...  You could imagine the test runner using local storage or something...  Consider iterating localStorage and only deleting things with the firestore prefix...  just in case?

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+  const instances = [];
+
+  if (existingMutationBatches.length + existingQueryTargets.length != 0) {
```
mikelehen: Would `existingMutatitonBatches.length > 0 || existingQueryTargets.length > 0` be a more natural way to express this?

(for a brief moment I actually thought this was checking if the two arrays cancelled each other out or something... but that's silly since length can't be negative)

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+    const secondaryChannel = new LocalStorageNotificationChannel(
```
mikelehen: This feels a bit dirty, I'd add a comment.

// HACK: Create a secondary channel to seed the data into LocalStorage. 
// NOTE: We don't call shutdown() on it because that would delete the data.

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +55,42 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+ * Creates LocalStorageNotificationChannel instance for testing, destroying any
+ * previous contents if they existed.
+ */
+export async function testLocalStorageNotificationChannel(
+  ownerId: string,
+  existingMutationBatches: BatchId[],
+  existingQueryTargets: TargetId[]
+): Promise<TabNotificationChannel> {
+  window.localStorage.clear();
+  const instances = [];
```
mikelehen: knownInstances ?

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+ *
+ * Firestore uses WebStorage for cross-tab notifications to propagate the
+ * state of pending mutation batches and query targets. All Firestore tabs are
+ * responsible for persisting their instance state as their in-memory state
+ * changes. Primary tabs will further update the state of each mutation batch
+ * and query target.
+ */
+export interface TabNotificationChannel {
+  /** Associates a new Mutation Batch with the current Firestore instance. */
+  addPendingMutation(batchId: BatchId): void;
```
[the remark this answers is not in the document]
schmidt-sebastian: These names are getting a tad long, but I am not opposed. Done (used some slightly different names).

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  /** Gets the minimum mutation batch for all active instances. */
+  getMinPendingMutation(): BatchId | null;
+
+  /** Associates a new Query Target with the current Firestore instance. */
+  addActiveQueryTarget(targetId: TargetId): void;
+
+  /** Removes a new Query Target from the current Firestore instance. */
+  removeActiveQueryTarget(targetId: TargetId): void;
+
+  /** Gets the active query targets for all active instances. */
```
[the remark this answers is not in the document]
schmidt-sebastian: I thought about this and don't think it's going to have a noticeable performance impact. This will only get called when a new primary tab gets selected.

I added a comment here and to `getMinimumGlobalPendingMutation` about their runtime complexity.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+   * `knownInstances` and registers listeners for newly added instances.
+   */
+  start(knownInstances: InstanceKey[]): void;
+
+  /** Shuts down the `TabNotificationChannel` and its listeners. */
+  shutdown(): void;
+}
+
+// Visible for testing
+export interface InstanceStateSchema {
```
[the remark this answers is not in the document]
schmidt-sebastian: Comment added.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Metadata state of an instance. Includes query targets and minumum and
+ * maximum pending mutation batch ids,
+ */
+class InstanceState {
+  constructor(
+    readonly instanceKey: InstanceKey,
+    public updateTime: Date,
+    readonly activeTargets: Set<TargetId>,
```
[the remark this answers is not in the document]
schmidt-sebastian: The main reason I went with Set over SortedSet is to match the mutability of the `Map` is used elsewhere in the code. It felt really wonky to mix and match data structures that behave so differently. This is more of a style issue though and I can easily be convinced that backwards compatibility is a lot more important.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    );
+    const updateTime = new Date(
+      validateNamedType(
+        'InstanceState.fromJSON',
+        'number',
+        'updateTime',
+        instanceData.updateTime
+      )
+    );
+    const activeTargets = new Set<TargetId>(
```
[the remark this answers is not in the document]
schmidt-sebastian: Yeah, good catch. I removed this.

packages/firestore/src/core/types.ts
```diff
@@ -26,6 +26,11 @@ export type BatchId = number;
  */
 export type TargetId = number;
 
+/**
+ * A randomly-generated key assigned to each Firestore instance at startup.
+ */
+export type InstanceKey = string;
```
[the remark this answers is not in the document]
schmidt-sebastian: It's part of the public interface in this class and will eventually be used to create values of this type. I don't think it will ever be used without pulling in the multi-tab code, so I I'm ok with moving it there.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+import { SortedSet } from '../util/sorted_set';
+import {
+  validateNamedArray,
+  validateNamedOptionalType,
+  validateNamedType
+} from '../util/input_validation';
+
+// Prefix keys used in WebStorage.
+const FIRESTORE_PREFIX = 'fs';
+const INSTANCE_ROW_PREFIX = 'instances';
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (named it INSTANCE_KEY_NAMESPACE).

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+/**
+ * WebStorage of the Firestore client.
+ *
+ * Firestore uses WebStorage for cross-tab notifications to propagate the
+ * state of pending mutation batches and query targets. All Firestore tabs are
+ * responsible for persisting their instance state as their in-memory state
+ * changes. Primary tabs will further update the state of each mutation batch
+ * and query target.
+ */
+export interface TabNotificationChannel {
```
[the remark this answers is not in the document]
schmidt-sebastian: Updated the comment.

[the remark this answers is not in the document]
schmidt-sebastian: `TabNotificationChannel` is a name that Gil suggested in the earlier PR. My original name for this was just `WebStorage`, which is admittedly strictly worse. 

I actually prefer the term 'instance'. In general, I have been using it in this code, partly because names like `instance_key` read better, and partly because the implementation doesn't really have anything tab specific. It's about instance communication, and could work just as well on desktop for C++. I think a good name for the data in this class would be `InstanceMetadata`.

So, that brings us back to naming. Here are some more suggestions:
- InstanceMetadataNotifier
- InstanceMetadataChannel
- InstanceMetadataBroadcast
- InstanceMetadataRelay
- SynchronizedInstanceState

For now, I picked InstanceMetdataChannel. Despite your argument against it, it reads a little less clunky to me.

packages/firestore/src/local/tab_notification_channel.ts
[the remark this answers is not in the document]
schmidt-sebastian: In an earlier version of go/multi-tab, I had a section on versioning. The idea was to add a global key in Local Storage (`fs_schema_version`) and clients could only participate if their schema version matches.  I removed this from the doc for the sake of brevity and since it is something that we could easily add later by treating its non-existence as version 0 (kind of like we are doing now in iOS for LevelDB persistence).

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  start(knownInstances: InstanceKey[]): void;
+
+  /** Shuts down the `TabNotificationChannel` and its listeners. */
+  shutdown(): void;
+}
+
+// Visible for testing
+export interface InstanceStateSchema {
+  instanceKey?: string;
+  updateTime: number;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  activeTargets: number[];
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  minMutationBatch?: number;
+  maxMutationBatch?: number;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+}
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, since this should translate cleanly into JSON, and using the actual types here makes that more obvious.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+
+/**
+ * Metadata state of an instance. Includes query targets and minumum and
+ * maximum pending mutation batch ids,
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  constructor(
+    readonly instanceKey: InstanceKey,
+    public updateTime: Date,
+    readonly activeTargets: Set<TargetId>,
+    private readonly _minMutationBatch?: BatchId,
+    private readonly _maxMutationBatch?: BatchId
+  ) {}
+
+  get minMutationBatch(): BatchId | null {
+    return this._minMutationBatch !== undefined ? this._minMutationBatch : null;
```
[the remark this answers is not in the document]
schmidt-sebastian: To me, `undefined` seemed like the better fit in the constructor arguments, but it doesn't match the semantics of SortedSet.min(), so I did the translation here. I updated the code to use `null` everywhere.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  get maxMutationBatch(): BatchId | null {
+    return this._maxMutationBatch !== undefined ? this._maxMutationBatch : null;
+  }
+
+  /**
+   * Parses an InstanceState from JSON. Throws if the value is not valid.
+   */
+  static fromJSON(json: string): InstanceState {
```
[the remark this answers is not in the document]
schmidt-sebastian: I hinted at this in the updated class comments. InstanceState is only ever read, and LocalInstanceState is only ever written.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    const instanceData = JSON.parse(json) as InstanceStateSchema;
+    const instanceId: InstanceKey = validateNamedType(
```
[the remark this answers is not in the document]
schmidt-sebastian: I abandoned this part of the PR and added my own validation.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+      'InstanceState.fromJSON',
+      'string',
+      'instanceKey',
+      instanceData.instanceKey
```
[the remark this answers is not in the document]
schmidt-sebastian: Good catch.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+      'InstanceState.fromJSON',
+      'number',
+      'minMutationBatch',
+      instanceData.minMutationBatch
+    );
+    const maxMutationBatch: number | undefined = validateNamedOptionalType(
+      'InstanceState.fromJSON',
+      'number',
+      'maxMutationBatch',
+      instanceData.minMutationBatch
```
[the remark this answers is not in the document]
schmidt-sebastian: It's really only one method call... but I reworked this to do this validation locally.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+      updateTime,
+      activeTargets,
+      minMutationBatch,
+      maxMutationBatch
+    );
+  }
+}
+
+/**
+ * Instance state of the local instance. Keeps track of all pending mutations.
```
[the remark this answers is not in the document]
schmidt-sebastian: The main reason for this separation is that an InstanceState read from LocalStorage only knows about the range of mutation batch IDs, whereas a the LocalInstanceState needs to know each batch ID to keep the range accurate.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+ */
+class LocalInstanceState extends InstanceState {
+  private pendingBatches = new SortedSet<BatchId>(primitiveComparator);
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  removeQueryTarget(targetId: TargetId): void {
+    assert(
+      this.activeTargets.has(targetId),
+      `Active Target ID '${targetId}' not found.`
+    );
+    this.activeTargets.delete(targetId);
+  }
+
+  /** Sets the update time to the current time. */
+  markCurrent(): void {
```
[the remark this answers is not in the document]
schmidt-sebastian: Makes sense.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    this.updateTime = new Date();
+  }
+
+  toJSON(): string {
+    const activeTargets = [];
```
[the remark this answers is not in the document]
schmidt-sebastian: Now using it.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    this.activeTargets.forEach(value => activeTargets.push(value));
+
+    const data: InstanceStateSchema = {
+      instanceKey: this.instanceKey,
+      updateTime: this.updateTime.getTime(),
+      activeTargets: Array.from(this.activeTargets)
```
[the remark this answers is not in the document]
schmidt-sebastian: That's what `activeTargets` was for. Thanks for catching this.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    return JSON.stringify(data);
+  }
+}
+
+/**
+ * `LocalStorageNotificationChannel` uses LocalStorage as the backing store for
+ * the TabNotificationChannel.
+ */
+export class LocalStorageNotificationChannel implements TabNotificationChannel {
+  private activeInstances = new Map<InstanceKey, InstanceState>();
```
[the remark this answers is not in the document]
schmidt-sebastian: I was under the impression that the TS compiler compiles these down to Objects for string:string maps. But it looks like it doesn't actually do that, so I updated it to an Object.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  private localStorage: Storage;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  private started = false;
+  private currentInstanceKey: string;
```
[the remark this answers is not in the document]
schmidt-sebastian: Changed to localInstanceKey and marked readonly.

BTW, I changed the name of this class to WebStorageMetadataChannel to not conflict with my other usages of 'local'.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  private storageListener = event => {
```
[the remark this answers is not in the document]
schmidt-sebastian: I added the private helper method to clean this up, but kept the local instance variable to allow for removal. I did not test whether `this.handleStorageEvent.bind(this)` constitutes a named function, which would allow me not to do this. Future me is going to verify this in all browsers.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    if (!this.started) {
+      return;
+    }
+    if (event.newValue == null) {
```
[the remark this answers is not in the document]
schmidt-sebastian: This is now asserted when we parse the instance row.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+      const oldInstance = this.parseInstanceRow(event.oldValue);
+      if (oldInstance) {
+        this.activeInstances.delete(oldInstance.instanceKey);
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, we can do that. I split the parsing up into two phases, and removed the redundant key.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+      }
+    } else {
+      const newInstance = this.parseInstanceRow(event.newValue);
+      if (newInstance) {
+        this.activeInstances.set(newInstance.instanceKey, newInstance);
+      }
+    }
+  };
+
+  constructor(private persistenceKey: string, private instanceId: string) {
```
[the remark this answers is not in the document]
schmidt-sebastian: This matches the behavior in IndexedDbPersistence and only throws when we call start().

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+    try {
+      return InstanceState.fromJSON(json);
+    } catch (err) {
+      error(`Unable to parse LocalStorage row: ${json}`, err.message);
+      return null;
+    }
+  }
+
+  private get localInstanceState(): LocalInstanceState {
+    return this.activeInstances.get(this.instanceId) as LocalInstanceState;
```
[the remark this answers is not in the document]
schmidt-sebastian: I added an assert here, but I think that we will have to revisit this as we add garbage collection.

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+  }
+
+  getMinPendingMutation(): BatchId | null {
+    let minMutationBatch = null;
+    this.activeInstances.forEach(instanceRow => {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
```diff
@@ -0,0 +1,393 @@
...
+
+    this.activeInstances.forEach(instanceRow => {
+      instanceRow.activeTargets.forEach(value => activeTargets.add(value));
+    });
+
+    return activeTargets;
+  }
+
+  start(knownInstances: InstanceKey[]): void {
+    if (!LocalStorageNotificationChannel.isAvailable()) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Matches the style of IndexedDbPersistence.

packages/firestore/src/util/input_validation.ts
```diff
@@ -117,56 +135,63 @@ export function validateNamedArrayAtLeastNumberOfElements<T>(
  * Validates the provided positional argument has the native JavaScript type
  * using typeof checks.
  */
-export function validateArgType(
+export function validateArgType<T>(
```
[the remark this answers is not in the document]
schmidt-sebastian: I went back on this and reverted the change in input_validation. It's true that this is type-checking by appearance only and basically accepts any type as we please. I don't think it's particularly unsafe as these method manually validate the input type.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Changed to instanceKey.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Updated, but I don't think this is more of a nice-to-have since this is typed to begin with.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Hm. Hrmm. Okay.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Certainly :) Updated.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Done

packages/firestore/src/local/tab_notification_channel.ts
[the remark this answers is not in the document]
mikelehen: Thanks. I hate to nit further, but I think you should use 'Global' and 'Local' instead of 'Globally' and 'Locally'.  'Global' and 'Local' are adjectives so they apply to the noun ('Mutation', 'QueryTarget', etc.) whereas 'Locally' and 'Globally' are adverbs so they apply to an adjective, and I don't know what "Locally active" vs "Globally active" means. Plus, I was hoping to use the same convention for all the methods.

That said, I'm feeling less gung-ho on the local vs global naming convention today, so if you want to drop it again, that's fine too (sorry).

packages/firestore/src/local/tab_notification_channel.ts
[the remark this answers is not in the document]
mikelehen: [irrelevant now] You were validating the same value twice by accident. I should have probably been less silly and more explicit. :-)

packages/firestore/test/unit/local/persistence_test_helpers.ts
mikelehen: Yeah.  But since we also have actual MutationBatch and Target classes in the client, it's confusing to mismatch the names and types (for humans, that is...  the compiler can keep it straight :-))

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+   */
+  toLocalStorageJSON(): string {
+    const data: InstanceStateSchema = {
+      lastUpdateTime: this.lastUpdateTime.getTime(),
+      activeTargetIds: []
+    };
+
+    this.activeTargetIds.forEach(targetId => {
+      data.activeTargetIds.push(targetId);
+    });
```
mikelehen: If you feel like it, I'd be a fan of adding SortedSet.toArray() method to simplify this code.

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+      }
+    }
+
+    this.started = true;
+    this.persistState();
+  }
+
+  getMinimumGloballyPendingMutation(): BatchId | null {
+    let minMutationBatch = null;
+    Object.keys(this.activeInstances).forEach(key => {
```
mikelehen: We have a `obj.forEach()` helper that'd probably be a bit cleaner?

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+      }
+    });
+
+    return minMutationBatch;
+  }
+
+  getGloballyActiveQueryTargets(): SortedSet<TargetId> {
+    let activeTargets = new SortedSet<TargetId>(primitiveComparator);
+
+    Object.keys(this.activeInstances).forEach(key => {
```
mikelehen: Again, consider using `obj.forEach()`

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+        }
+      }
+    }
+  }
+
+  private get localInstanceState(): LocalInstanceState {
+    return this.activeInstances[this.instanceKey] as LocalInstanceState;
+  }
+
+  private persistState(): void {
```
mikelehen: I think we should consider aggregating state changes to avoid too much thrashing if the user does a bunch of operations back-to-back (writing to local storage tons of times and other tabs parsing the new entries tons of times). So rather than write to localStorage immediately, we could e.g. delay it by 100ms and batch up all the persistState() calls within that window into 1.  This would actually alleviate a lot of my performance concerns.

Consider adding a TODO(multitab)?

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+
+  removeLocallyActiveQueryTarget(targetId: TargetId): void {
+    this.localInstanceState.removeQueryTarget(targetId);
+    this.persistState();
+  }
+
+  shutdown(): void {
+    assert(
+      this.started,
+      'LocalStorageMetadataNotifier.shutdown() called when not started'
```
mikelehen: wrong class name.

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +59,53 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
   await persistence.start();
   return persistence;
 }
+
+/**
+ * Creates and starts a LocalStorageMetadataNotifier instance for testing,
```
mikelehen: wrong class name

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+import {
+  InstanceStateSchema,
+  WebStorageMetadataChannel,
+  InstanceMetadataChannel
+} from '../../../src/local/instance_metadata_channel';
+import { BatchId, TargetId } from '../../../src/core/types';
+import { AutoId } from '../../../src/util/misc';
+import { expect } from 'chai';
+
+const GRACE_INTERVAL_MS = 100;
```
mikelehen: Nice constant. Can you comment it? :-)

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +59,53 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+ * destroying any previous contents if they existed.
+ */
+export async function testLocalStorageNotificationChannel(
```
mikelehen: wrong class name

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+describe('WebStorageMetadataChannel', () => {
+  if (!WebStorageMetadataChannel.isAvailable()) {
+    console.warn(
+      'No LocalStorage. Skipping LocalStorageNotificationChannelTests tests.'
```
mikelehen: wrong class name

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    );
+    return;
+  }
+
+  const localStorage = window.localStorage;
+
+  let notificationChannel: InstanceMetadataChannel;
```
mikelehen: fix var name

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      )
+    );
+
+    let expectedKeyCount = /* lastUpdateTime + activeTargetIds */ 2;
+
+    expect(actual.lastUpdateTime).to.be.a('number');
+    expect(actual.lastUpdateTime).to.be.greaterThan(
+      Date.now() - GRACE_INTERVAL_MS
+    );
+    expect(actual.lastUpdateTime).to.be.at.most(Date.now());
```
mikelehen: Mixing "greaterThan" and "at.most" tripped me up a bit. Could you use .at.least?  And maybe combine all of this?  Can you do something like:

expect(actual.lastUpdateTime).to.be.a('number').at.least(Date.now() - GRACE_PERIOD).and.at.most(Date.now())

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    expect(actual.activeTargetIds).to.be.an('array');
+    expect(actual.activeTargetIds).to.be.deep.equal(activeTargetIds);
```
mikelehen: I don't think you normally use `be` and `equal` together (here and below).

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    expect(Object.keys(actual).length).to.be.equal(expectedKeyCount);
+  }
+
+  describe('persists mutation batches', () => {
+    beforeEach(() => {
+      return persistenceHelpers
+        .testLocalStorageNotificationChannel(ownerId, [], [])
+        .then(nc => {
+          notificationChannel = nc;
```
mikelehen: wrong names...

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        });
+    });
+
+    afterEach(() => {
+      notificationChannel.shutdown();
+    });
+
+    it('with empty batch', () => {
```
mikelehen: "Persists mutation batches with empty batch" is a strange name.  Perhaps change this to 'when empty' or something?

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      notificationChannel.addLocalPendingMutation(0);
+      notificationChannel.addLocalPendingMutation(1);
+      assertInstanceState([], 0, 1);
+
+      notificationChannel.addLocalPendingMutation(2);
+      notificationChannel.addLocalPendingMutation(3);
+      assertInstanceState([], 0, 3);
+
+      notificationChannel.removeLocalPendingMutation(0);
+      notificationChannel.removeLocalPendingMutation(2);
```
mikelehen: FWIW I think this is an invalid operation (mutation batches must always be removed in order). I was actually tempted to have you assert it in the remove implementation, but I think it's asserted elsewhere so it seemed overkill.

I'd be inclined to either change this to 1 or add a comment that this is technically invalid.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      assertInstanceState([], 1, 3);
+    });
+  });
+
+  describe('persists query targets', () => {
+    beforeEach(() => {
+      return persistenceHelpers
+        .testLocalStorageNotificationChannel(ownerId, [], [])
+        .then(nc => {
+          notificationChannel = nc;
```
mikelehen: fix names

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+  describe('combines instance data', () => {
+    let previousAddEventListener;
+    let storageCallback: (StorageEvent) => void;
+
+    beforeEach(() => {
+      previousAddEventListener = window.addEventListener;
+      window.addEventListener = (type, callback) => {
+        expect(type).to.be.equal('storage');
+        storageCallback = callback;
+      };
```
mikelehen: Not a big fan of this, but since we won't need to port it, maybe it's okay...  Would it be terrible to actually go through LocalStorage and rely on the real events?  That seems less hacky and provide better validation anyway (that local storage events actually work the way we think they do).

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+          expect(storageCallback).to.not.be.undefined;
+        });
+    });
+
+    afterEach(() => {
+      notificationChannel.shutdown();
+      window.addEventListener = previousAddEventListener;
+    });
+
+    const verifyState = (minBatchId: BatchId, expectedTargets: TargetId[]) => {
```
mikelehen: make this a normal function like you did for assertInstanceState() ?

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+      const actualTargets = notificationChannel.getGloballyActiveQueryTargets();
+      expect(actualTargets.size).to.equal(expectedTargets.length);
+
+      let missingTargets = actualTargets;
+      for (const targetId of expectedTargets) {
+        missingTargets = missingTargets.delete(targetId);
+      }
+      expect(missingTargets.size).to.equal(0);
```
mikelehen: If SortedSet had a .toArray() I think that would clean this up too, right?  You could just `expect(actualTargets.toArray()).to.deep.equal(expectedTargets.slice().sort())`?

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+      expect(
+        notificationChannel.getMinimumGloballyPendingMutation()
+      ).to.be.equal(minBatchId);
```
mikelehen: more unnecessary `be`s.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    it('with existing instance', () => {
+      // The prior instance has one pending mutation and two active query targets
+      verifyState(1, [3, 4]);
+
+      notificationChannel.addLocalPendingMutation(3);
+      notificationChannel.addLocallyActiveQueryTarget(4);
+      verifyState(1, [3, 4]);
+
+      notificationChannel.addLocalPendingMutation(0);
```
mikelehen: this is technically invalid usage too... It should be >= 3

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        missingTargets = missingTargets.delete(targetId);
+      }
+      expect(missingTargets.size).to.equal(0);
+
+      expect(
+        notificationChannel.getMinimumGloballyPendingMutation()
+      ).to.be.equal(minBatchId);
+    };
+
+    it('with existing instance', () => {
```
mikelehen: should this be 'from local instance'?

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
```diff
@@ -0,0 +1,271 @@
...
+        minMutationBatchId: 0,
+        maxMutationBatchId: 0
+      };
+
+      const updatedState: InstanceStateSchema = {
+        lastUpdateTime: Date.now(),
+        activeTargetIds: [5, 6],
+        minMutationBatchId: 0,
+        maxMutationBatchId: 0
+      };
```
mikelehen: I think I'd be much happier to see us actually creating multiple channels and testing that modifications via one are surfaced correctly in others.  As-is, we're skipping a big chunk of the implementation (LocalStorage) and making our tests dependent on implementation details (the LocalStorage schema) unnecessarily.

That would also mean that if we implemented persistence on Node one day and used a communication channel other than LocalStorage, this testing would cover it, so it's more valuable.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Done

packages/firestore/src/local/instance_metadata_channel.ts
schmidt-sebastian: Added todo

packages/firestore/src/local/instance_metadata_channel.ts
```diff
@@ -0,0 +1,455 @@
...
+  expectedMatch: RegExp,
+  dataSegement: number
+): string | null {
+  if (!expectedMatch.test(key)) {
+    console.log('key didnt match ' + key);
+    return null;
+  }
+
+  return key.split('_')[dataSegement];
+}
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (methods are called toLocalStorageClientKey and fromLocalStorageClientKey to allow for me to add new formats later).

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Sounds good

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Removed

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Sounds good

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: I added a comment.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Fixed here and everywhere

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: This doesn't work for us since LocalStorage doesn't fire for local updates. Added comment.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Done. Ended up using `to.have.members` here.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Removed

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Renamed to 'with data from existing client'. It's actually what I refer to as a remote instance.

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Added comment

packages/firestore/test/unit/local/web_storage_metadata_channel.test.ts
schmidt-sebastian: Some of this testing exists hidden in `testWebStorageSharedClientState`. I updated the test here to rely on the schema embedded in RemoteClientState, but as pointed out before, I can't actually test the notifications.

schmidt-sebastian:
This PR is ready for review again.

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -44,3 +59,53 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
+
+  const knownInstances = [];
+
+  if (
+    existingMutationBatchIds.length > 0 ||
+    existingQueryTargetIds.length > 0
+  ) {
+    // HACK: Create a secondary client state to seed data into LocalStorage.
+    // NOTE: We don't call shutdown() on it because that would delete the data.
+    const secondaryChannel = new WebStorageSharedClientState(
```
mikelehen: secondaryClientState

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * This class represents the immutable ClientState for a client read from
+ * LocalStorage. It contains the list of its active query targets and the range
+ * of its pending mutation batch IDs.
+ */
+class RemoteClientState implements ClientState {
+  constructor(
+    readonly clientKey: ClientKey,
+    public lastUpdateTime: Date,
+    public activeTargetIds: SortedSet<TargetId>,
```
mikelehen: lastUpdateTime and activeTargetIds can be readonly too, I think?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+  readonly minMutationBatchId: BatchId | null;
+}
+
+/**
+ * This class represents the immutable ClientState for a client read from
+ * LocalStorage. It contains the list of its active query targets and the range
+ * of its pending mutation batch IDs.
+ */
+class RemoteClientState implements ClientState {
+  constructor(
```
mikelehen: Can the constructor be private?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+    readonly clientKey: ClientKey,
+    public lastUpdateTime: Date,
+    public activeTargetIds: SortedSet<TargetId>,
+    readonly minMutationBatchId: BatchId | null,
+    readonly maxMutationBatchId: BatchId | null
+  ) {}
+
+  /**
+   * Parses a ClientState from the JSON representation in LocalStorage.
```
mikelehen: ClientState => RemoteClientState ?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+    assert(
+      this.started,
+      'WebStorageSharedClientState.shutdown() called when not started'
+    );
+    window.removeEventListener('storage', this.storageListener);
+    this.storage.removeItem(this.storageKey);
+    this.started = false;
+  }
+
+  private handleStorageEvent(event: StorageEvent): void {
```
mikelehen: Just FYI- It occurred to me that once you start triggering events / callbacks to SyncEngine on changes we're going to need to make sure we are on the AsyncQueue, so you'll probably need to make WebStorageSharedClientState have an AsyncQueue reference, etc.

Since JS is single-threaded and we're not calling back into any other client code here, it doesn't matter for now.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+ */
+export class WebStorageSharedClientState implements SharedClientState {
+  private readonly storage: Storage;
+  private readonly storageKey: string;
+  private readonly activeClients: { [key: string]: ClientState } = {};
+  private readonly storageListener = this.handleStorageEvent.bind(this);
+  private readonly clientStateKeyRe: RegExp;
+  private started = false;
+
+  constructor(private persistenceKey: string, private clientKey: string) {
```
mikelehen: clientKey => localClientKey (to avoid confusion with the generic 'clientKey' variables you use in this code)?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+    if (!WebStorageSharedClientState.isAvailable()) {
+      throw new FirestoreError(
+        Code.UNIMPLEMENTED,
+        'LocalStorage is not available on this platform.'
+      );
+    }
+    this.storage = window.localStorage;
+    this.storageKey = this.toLocalStorageClientKey(this.clientKey);
```
mikelehen: storageKey => localClientStorageKey ?  (or remove this and just generate it on-demand from clientKey).

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -0,0 +1,422 @@
...
+    if (event.storageArea === this.storage) {
+      // TODO(multitab): This assert will likely become invalid as we add garbage
+      // collection.
+      assert(
+        event.key !== this.storageKey,
+        'Received LocalStorage notification for local change.'
+      );
+      const clientKey = this.fromLocalStorageClientKey(event.key);
+      if (clientKey) {
+        if (event.newValue == null) {
```
mikelehen: Consider adding logging for these LocalStorage notifications (can do in a later PR, e.g. when adding callbacks to SyncEngine if you prefer).

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Done

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Yes, fixed.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Done

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Done. This also works nicely with the method `localClientState().`

Also made this readonly.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Gil suggested this variable in the first PR, which I think makes sense. Renamed.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: I think we would want to log the actual notifications that we deliver to SyncEngine, so I am going to defer on this until we get to that part.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: Done

PR #485, opened 2018-02-02
[Multi-Tab] Adding Schema Migration

schmidt-sebastian:
This performs the IndexedDB schema migration required for MultiTab:

- It adds a new IndexedDB store to store instance metadata.
- It adds a new IndexedDB store to store a query changelog.
- It adds a field to DbMutationQueue, which we can do silently without changing the schema.

The functionality still allows for creation of schema version 1, so that we can make testing easier.

It's likely that we will want to coordinate this migration with our Garbage Collection efforts.


packages/firestore/src/local/indexeddb_migrations.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Runs any migrations needed to bring the given database up to the current
+ * schema version.
+ */
+export function createOrUpgradeDb(
+  db: IDBDatabase,
+  oldVersion: number,
+  newVersion: number
+): void {
+  assert(
+    oldVersion >= 0 || oldVersion <= 1,
```
mikelehen: && ?

packages/firestore/src/local/indexeddb_migrations.ts
```diff
@@ -0,0 +1,123 @@
...
+    'Unexpected upgrade from version ' + oldVersion
+  );
+  assert(
+    newVersion >= 1 || newVersion <= 2,
```
mikelehen: && ?

packages/firestore/src/local/indexeddb_migrations.ts
```diff
@@ -0,0 +1,123 @@
...
+    'Unexpected upgrade to version ' + newVersion
+  );
+
+  const createV1 = newVersion >= 1 && oldVersion <= 1;
+  const dropV1 = oldVersion >= 1;
```
mikelehen: Should this be <= ?  FWIW, I'm not sure these intermediate variables are helping here. I keep looking back and forth between the if-checks...

packages/firestore/src/local/indexeddb_migrations.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Runs any migrations needed to bring the given database up to the current
+ * schema version.
+ */
+export function createOrUpgradeDb(
+  db: IDBDatabase,
+  oldVersion: number,
+  newVersion: number
```
mikelehen: Having newVersion be a parameter is super confusing to me. Are you intending to support downgrades (I think we should try hard to avoid that!)?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -131,7 +87,14 @@ export class DbMutationQueue {
      * After sending this token, earlier tokens may not be used anymore so
      * only a single stream token is retained.
      */
-    public lastStreamToken: string
+    public lastStreamToken: string,
+    /**
```
mikelehen: newline

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -131,7 +87,14 @@ export class DbMutationQueue {
...
+     * An identifier for the highest numbered batch that has been acknowledged
+     * by the server. All MutationBatches in this queue with batchIds less
+     * than or equal to this value are considered to have been acknowledged by
+     * the server.
+     */
+    public highestPendingBatchId: number
```
mikelehen: This only exists so that we can leave acknowledged mutations in the queue for the sake of other tabs being able to read them, right?  I think it's worth calling that out here. Also, are we going to port this to other platforms?  If not, please add a "// PORTING NOTE:" comment.

packages/firestore/src/local/indexeddb_migrations.ts
```diff
@@ -0,0 +1,123 @@
...
+  DbTargetDocument,
+  DbTargetGlobal
+} from './indexeddb_schema';
+import { assert, fail } from '../util/assert';
+
+// TODO(mikelehen): Get rid of "as any" if/when TypeScript fixes their types.
+// https://github.com/Microsoft/TypeScript/issues/14322
+type KeyPath = any; // tslint:disable-line:no-any
+
+function createCache(db: IDBDatabase): void {
```
mikelehen: FWIW, "cache" is non-obvious, so I would at least add a comment.

// Create IndexedDB stores for our query / remote document caches.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -404,6 +367,7 @@ export class DbTargetDocument {
      * The targetId identifying a target.
      */
     public targetId: TargetId,
+    public snapshotVersion: number,
```
mikelehen: comment? (especially call out if we are doing any cheating with fake snapshot versions or whatever...)

Also I think we want to do some design review (with Gil at least) before moving forward with this (unless it's living in a non-master branch or something).

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -455,6 +419,20 @@ export class DbTargetGlobal {
   ) {}
 }
 
+export type DbInstanceKey = [string, string];
```
mikelehen: comments?

Also, "Instance" feels too generic. Perhaps "ClientInstance" (and comments should explain the multi-tab scenario)

packages/firestore/src/local/indexeddb_migrations.ts
schmidt-sebastian: I renamed this and split this up to match the names in Android:  createQueryCache() and createRemoteDocumentCache()

packages/firestore/src/local/indexeddb_migrations.ts
schmidt-sebastian: I am not at all planning on adding  the ability to downgrade. I renamed the arguments to `fromVersion` and `toVersion` and added an assert that one must be lower than the other.

packages/firestore/src/local/indexeddb_migrations.ts
schmidt-sebastian: Fixed (combined the asserts into one)

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: I added newlines here and everywhere else where they were missing. I also tried to make this semi-valid JSDoc along with it (I still don't think it's valid as we would probably have to move the JSDoc comments to the function header).

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: I updated the outdated comment. This is to reduce the number of scans we have to do in the mutation batch. Right now, to get the highest batch ID, we do an full scan at startup. I don't want to do this every time we add a mutation from a secondary tab.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: I talked to Gil and Greg about this yesterday. Gil favors a separate ChangeLog, as primary key updates are pretty expensive. I updated the Design Doc with this yesterday and added a new table to store this changelog. The nice thing about this is that now all changes in the schema are additive and we don't need to drop the Query & Remote Document Cache anymore.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: No longer used.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Note that prettier stripped the newlines again.

schmidt-sebastian:
> I didn't review in detail but looked over this and flagged a few things that jumped out.
> 
> Have you looked at the SQLitePersistence runMigrations() code for Android to see if we can follow a similar pattern? In particular:
> 
> I like the switch / case (with fall-through) approach it uses.
> In general we should try to make the upgrade mechanism as similar as possible across platforms.
> It may make sense to copy the INDEXING_SUPPORT_ENABLED flag mechanism we used there. I think we should be careful not to migrate anybody's schema until we're fully ready to launch multi-tab.
> Also, Greg is working on schema upgrade infrastructure for iOS right now too, so it may be worth peeking at what he's doing.

This PR is now ready for review.

I looked at the iOS PR, which originally inspired the creation of a separate file to run the migrations. I like the Android way better to do this in place and merged my change into indexeddb_schema.

Both Android and iOS use the switch-statement with fallthrough. This doesn't allow us to specify a range of versions to update (from 0 to 1 for example) and I would like to use that for testing of the actual migration process.

I did not make this flag controlled as this PR is against the `firestore-multi-tab` branch. We should decide later on whether this feature will be optional and how we want to deal with optional features in the schema. Even if multi-tab is optional, it might make sense to add it to the schema but not write any data into the multi-tab specific stores.

schmidt-sebastian:
BTW, I updated this PR to remove the "userId" from the InstanceMetadata store. The rest of the multi-tab work assumes that only one user is logged in at a time, and there is no reason to keep state around for users that are no longer active. I haven't updated the doc yet and wanted have this sanity checked first.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -432,40 +454,144 @@ export class DbTargetGlobal {
...
+}
+
+function createTargetChangeStore(db: IDBDatabase): void {
+  db.createObjectStore(DbTargetChange.store, {
+    keyPath: DbTargetChange.keyPath as KeyPath
+  });
+}
+
+/**
+ * A record of the metadata state of each instance.
```
mikelehen: I think 'instance' is a pretty ambiguous term and so it would be good to at least occasionally in comments clarify, e.g. "of each instance of the Firestore client (i.e. running in different browser tabs)."

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -432,40 +454,144 @@ export class DbTargetGlobal {
...
+ *
+ * PORTING NOTE: This is used for primary-tab selection to support multi-tab
+ * persistence and does not need to be ported to iOS or Android.
+ */
+export class DbInstanceMetadata {
+  /** Name of the IndexedDb object store. */
+  static store = 'instanceMetadata';
+
+  /** Keys are automatically assigned via the userId and instanceKey properties. */
+  static keyPath = ['userId', 'instanceKey'];
```
mikelehen: Why is userId here?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -432,40 +454,144 @@ export class DbTargetGlobal {
...
+
+  constructor(
+    /** @param userId - The normalized user ID to which this batch belongs. */
```
mikelehen: what batch?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -432,40 +454,144 @@ export class DbTargetGlobal {
...
-  DbTarget.store,
   DbOwner.store,
+  DbRemoteDocument.store,
+  DbTargetDocument.store,
   DbTargetGlobal.store,
-  DbTargetDocument.store
+  DbTarget.store
 ];
+
+const SCHEMA_V2_STORES = [DbInstanceMetadata.store, DbTargetChange.store];
```
mikelehen: V1_STORES, SCHEMA_V2_STORES, ALL_STORES

One of these is not like the others.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -432,40 +454,144 @@ export class DbTargetGlobal {
...
+
+/**
+ * The list of all IndexedDB stored used by the SDK. This is used when creating
```
mikelehen: stored=>stores

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  const request = window.indexedDB.deleteDatabase(INDEXEDDB_TEST_DATABASE);
+  request.onsuccess = (event: Event) => {
+    deferred.resolve();
+  };
+  request.onerror = (event: ErrorEvent) => {
+    deferred.reject((event.target as IDBOpenDBRequest).error);
+  };
+
+  return deferred;
+}
```
mikelehen: Can you use SimpleDb.delete() instead?

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import {
+  ALL_STORES,
+  createOrUpgradeDb,
+  V1_STORES
+} from '../../../src/local/indexeddb_schema';
+import { Deferred } from '../../../src/util/promise';
+
+const INDEXEDDB_TEST_DATABASE = 'schemaTest';
+
+function deleteDb() {
```
mikelehen: If you add your return type, you'll likely find out you have a bug. :-)

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    deferred.resolve();
+  };
+  request.onerror = (event: ErrorEvent) => {
+    deferred.reject((event.target as IDBOpenDBRequest).error);
+  };
+
+  return deferred;
+}
+
+function initDb(targetVersion) {
```
mikelehen: return type?

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  };
+  request.onerror = (event: ErrorEvent) => {
+    deferred.reject((event.target as IDBOpenDBRequest).error);
+  };
+  request.onupgradeneeded = (event: IDBVersionChangeEvent) => {
+    const db = (event.target as IDBOpenDBRequest).result;
+    createOrUpgradeDb(db, event.oldVersion, targetVersion);
+  };
+
+  return deferred.promise;
```
mikelehen: Could you use SimpleDb.openOrCreate() ?  [I'm not 100% sure]

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+function getAllObjectStores(db: IDBDatabase): String[] {
+  const objectStores: String[] = [];
+  for (let i = 0; i < db.objectStoreNames.length; ++i) {
+    objectStores.push(db.objectStoreNames.item(i));
+  }
+  objectStores.sort();
+  return objectStores;
+}
+
+// Sorting these arrays directly should not affect the functionality of the SDK.
```
mikelehen: But it is ugly!  Why not just do `V1_STORES.slice().sort()` where you need them sorted?

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  beforeEach(() => {
+    return deleteDb();
+  });
+
+  it('can install schema version 1', () => {
+    return initDb(1).then(db => {
+      expect(db.version).to.be.equal(1);
+      expect(getAllObjectStores(db)).to.deep.equal(V1_STORES);
```
mikelehen: can you use `to.have.members(...)` and not need to ensure they're sorted the same?

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+  objectStores.sort();
+  return objectStores;
+}
+
+// Sorting these arrays directly should not affect the functionality of the SDK.
+V1_STORES.sort();
+ALL_STORES.sort();
+
+describe('SchemaMigration', () => {
```
mikelehen: What is `SchemaMigration`? :-)  Maybe 'Schema Migration: createOrUpgradeDb()' (since createOrUpgradeDb is the method you're testing here).

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -22,53 +22,38 @@ import { assert } from '../util/assert';
...
+/**
+ * Performs database creation and schema migrations up to schema version 2.
+ */
+export function createOrUpgradeDb(
+  db: IDBDatabase,
+  fromVersion: number,
+  toVersion: number
+): void {
+  assert(
+    fromVersion < toVersion && fromVersion >= 0 && toVersion <= 2,
```
mikelehen: 2 => SCHEMA_VERSION ?

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: No longer used.

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: SimpleDb doesn't expose the version and the list of object stores.

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: Done, and then replaced with SimpleDb.delete()

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Fixed

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Fixed

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Removed.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: This used to be user centric, but I removed the userId from this. Clients for older users that are no longer active should become outdated based on their updateTime.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Changed the naming here and elsewhere from `instance` to `client` as per our discussion.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -183,6 +182,20 @@ export class DbMutationBatch {
...
+function createMutationQueue(db: IDBDatabase): void {
+  db.createObjectStore(DbMutationQueue.store, {
+    keyPath: DbMutationQueue.keyPath
+  });
+
+  db.createObjectStore(DbMutationBatch.store, {
+    keyPath: DbMutationBatch.keyPath as KeyPath
+  });
+
+  // NOTE: keys for these stores are specified explicitly rather than using a
```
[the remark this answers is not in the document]
schmidt-sebastian: Removed this comment.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -112,26 +101,35 @@ export class DbMutationQueue {
...
-    public lastStreamToken: string
+    public lastStreamToken: string,
+    /**
+     * @param highestPendingBatchId - An identifier for the highest numbered
+     * batch in the mutation queue. This allows for efficient insert of new
+     * batches without relying on in-memory state.
+     *
+     * PORTING NOTE: iOS and Android clients keep this value in-memory.
+     */
+    public highestPendingBatchId?: number
```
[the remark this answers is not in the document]
schmidt-sebastian: It was optional because it would not have been set in schema version 1.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -112,26 +101,35 @@ export class DbMutationQueue {
...
      * only a single stream token is retained.
      */
-    public lastStreamToken: string
+    public lastStreamToken: string,
+    /**
+     * @param highestPendingBatchId - An identifier for the highest numbered
+     * batch in the mutation queue. This allows for efficient insert of new
+     * batches without relying on in-memory state.
+     *
+     * PORTING NOTE: iOS and Android clients keep this value in-memory.
```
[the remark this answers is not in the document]
schmidt-sebastian: I changed this to `nextBatchId` (which is the internal name used in the clients), but then ultimately removed it as pointed out above.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -112,26 +101,35 @@ export class DbMutationQueue {
...
-     * The normalized user ID to which this queue belongs.
+     * @param userId - The normalized user ID to which this queue belongs.
      */
     public userId: string,
     /**
-     * An identifier for the highest numbered batch that has been acknowledged
-     * by the server. All MutationBatches in this queue with batchIds less
-     * than or equal to this value are considered to have been acknowledged by
-     * the server.
+     * @param lastAcknowledgedBatchId - An identifier for the highest numbered
```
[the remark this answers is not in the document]
schmidt-sebastian: It only took seconds to undo, so it's undone.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -22,53 +22,38 @@ import { assert } from '../util/assert';
...
-  db.createObjectStore(DbDocumentMutation.store);
-  db.createObjectStore(DbRemoteDocument.store);
-  db.createObjectStore(DbOwner.store);
-  db.createObjectStore(DbTargetGlobal.store);
+  if (fromVersion === 0 && toVersion >= 1) {
+    createOwnerStore(db);
+    createMutationQueue(db);
+    createQueryCache(db);
+    createRemoteDocumentCache(db);
+  }
```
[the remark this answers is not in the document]
schmidt-sebastian: I used code snippet version 1.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -22,53 +22,38 @@ import { assert } from '../util/assert';
...
-  });
-  // NOTE: This is unique only because the TargetId is the suffix.
-  targetStore.createIndex(
-    DbTarget.queryTargetsIndexName,
-    DbTarget.queryTargetsKeyPath,
-    { unique: true }
-  );
+  if (toVersion === 2) {
+    createInstanceMetadataStore(db);
+    createTargetChangeStore(db);
```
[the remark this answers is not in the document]
schmidt-sebastian: I talked to Gil about storing the `nextBatchId` and whether it needs to be globally or locally unique. We currently compute a globally unique batch ID at startup by looking at all pending mutations for all users. This requires one read per user, and is likely something we don't want to do for every mutation. 
The multi-tab proposal already deviates from this and only guarantees unique Batch IDs per user (as they are stored along with each user's mutation queue metadata). Unlike the existing `loadNextBatchIdFromDb`, this only requires one read from the mutationsStore. While this index scan is likely a little more expensive than reading from the tiny mutationQueuesStore, using the existing data in mutationsStore requires 1 fewer writes per mutation.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: I actually changed `SCHEMA_VERSION` to `DEFAULT_SCHEMA_VERSION` and reset the value back to 1. This way, I can merge this code to master, and Greg can build the GC migration on top of this.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -22,53 +22,38 @@ import { assert } from '../util/assert';
...
-    DbTargetDocument.store,
-    // tslint:disable-next-line:no-any
-    { keyPath: DbTargetDocument.keyPath as any }
-  );
-  targetDocumentsStore.createIndex(
-    DbTargetDocument.documentTargetsIndex,
-    DbTargetDocument.documentTargetsKeyPath,
-    { unique: true }
+/**
+ * Performs database creation and schema migrations up to schema version 2.
```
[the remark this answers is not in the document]
schmidt-sebastian: I changed the comment as suggested. The schema version specific comment is not just above the assert, which has to be changed as we add new schema version. I thought about adding named constant for each schema version, but I am not sure how much value that adds.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Even if it is not for external documentation, it could show up in code completion when done properly - whereas right not it just shows up as a warning due do the `/** JSDoc */` marks.

I agree that this is mostly a non-issue though and happily pressed some buttons to revert this.

schmidt-sebastian:
This is ready for review again. I have updated the base branch to 'master`, but used the SCHEMA_VERSION constant to make sure that we don't run this upgrade yet.

schmidt-sebastian:
I want to keep the multi-tab work in a separate branch to avoid unnecessary churn on the master branch in case I need to revisit some of my design decisions. The reason I want(ed) to merge this into Master is that Greg is going to write schema migration code pretty soon and it would be nice if he could built it on top of this PR (which might bump multi-tab to version 3). 

There are many other ways to guarantee that @gsoltis  doesn't do duplicate work, and I'll just ask him to pull in this PR and merge it to Master with multi-tab removed and GC added.

TL/DR: Rebased to firestore-multi-tab and made Schema Version 2 the default in this branch.

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -26,23 +26,30 @@ import { SimpleDb } from '../../../src/local/simple_db';
...
+    request.onsuccess = (event: Event) => {
+      resolve((event.target as IDBOpenDBRequest).result);
+    };
+    request.onerror = (event: ErrorEvent) => {
+      reject((event.target as IDBOpenDBRequest).error);
+    };
+  })
+    .then(db => {
+      fn(db);
+      return db;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -581,10 +582,4 @@ const V2_STORES = [DbClientMetadata.store, DbTargetChange.store];
...
  * atomically.
  */
-export const DEFAULT_STORES = V1_STORES;
-
-/**
- * The list of all IndexedDb stores. This is only available in multi-tab enabled
- * clients with the schema version 2.
- */
-export const ALL_STORES = [...V1_STORES, ...V2_STORES];
+export const DEFAULT_STORES = [...V1_STORES, ...V2_STORES];
```
mikelehen: Put back to ALL_STORES?

PR #489, opened 2018-02-05
Refactor AsyncQueue's "delayed scheduling" support and add cancellation.

mikelehen:
* Introduces a DelayedOperation helper class in AsyncQueue to encapsulate
  delayed op logic.
* Adds cancellation support which I want to use in
  https://github.com/firebase/firebase-js-sdk/pull/412
* Remove delayedOperationsCount in favor of keeping delayedOperations populated
  correctly at all times.
* Fixes a preexisting issue in AsyncQueue.schedule() where the returned promise
  would always be resolved with undefined, instead of the result of your op.
  Also remove an AnyDuringMigration usage.

### DIFF
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -256,6 +256,8 @@ export abstract class PersistentStream<
    * states imply pending network operations.
    */
   markIdle(): void {
+    // Starts the idle time if we are in state 'Open' and are not yet already
+    // running a timer (in which case the previous idle timeout still applies).
     if (this.isOpen() && this.inactivityTimerPromise === null) {
       this.inactivityTimerPromise = this.queue.scheduleWithDelay(
         () => this.handleIdleCloseTimer(),
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -31,6 +31,7 @@ import { Connection, Stream } from './connection';
 import { JsonProtoSerializer } from './serializer';
 import { WatchChange } from './watch_change';
 import { isNullOrUndefined } from '../util/types';
+import { CancelablePromise } from '../util/promise';
 
 const LOG_TAG = 'PersistentStream';
 
@@ -154,7 +155,7 @@ export abstract class PersistentStream<
   ListenerType extends PersistentStreamListener
 > {
   private state: PersistentStreamState;
-  private idle = false;
+  private inactivityTimerPromise: CancelablePromise<void> | null = null;
   private stream: Stream<SendType, ReceiveType> | null = null;
 
   protected backoff: ExponentialBackoff;
@@ -245,16 +246,23 @@ export abstract class PersistentStream<
   }
 
   /**
-   * Initializes the idle timer. If no write takes place within one minute, the
-   * WebChannel stream will be closed.
+   * Marks this stream as idle. If no further actions are performed on the
+   * stream for one minute, the stream will automatically close itself and
+   * notify the stream's onClose() handler with Status.OK. The stream will then
+   * be in a !isStarted() state, requiring the caller to start the stream again
+   * before further use.
+   *
+   * Only streams that are in state 'Open' can be marked idle, as all other
+   * states imply pending network operations.
    */
   markIdle(): void {
-    this.idle = true;
-    this.queue
-      .scheduleWithDelay(() => {
-        return this.handleIdleCloseTimer();
-      }, IDLE_TIMEOUT_MS)
-      .promise.catch((err: FirestoreError) => {
+    if (this.isOpen() && this.inactivityTimerPromise === null) {
+      this.inactivityTimerPromise = this.queue.scheduleWithDelay(
+        () => this.handleIdleCloseTimer(),
+        IDLE_TIMEOUT_MS
+      );
+
+      this.inactivityTimerPromise.catch((err: FirestoreError) => {
         // When the AsyncQueue gets drained during testing, pending Promises
         // (including these idle checks) will get rejected. We special-case
         // these cancelled idle checks to make sure that these specific Promise
@@ -266,6 +274,7 @@ export abstract class PersistentStream<
           }`
         );
       });
+    }
   }
 
   /** Sends a message to the underlying stream. */
@@ -276,7 +285,7 @@ export abstract class PersistentStream<
 
   /** Called by the idle timer when the stream should close due to inactivity. */
   private handleIdleCloseTimer(): Promise<void> {
-    if (this.isOpen() && this.idle) {
+    if (this.isOpen()) {
       // When timing out an idle stream there's no reason to force the stream into backoff when
       // it restarts so set the stream state to Initial instead of Error.
       return this.close(PersistentStreamState.Initial);
@@ -286,7 +295,10 @@ export abstract class PersistentStream<
 
   /** Marks the stream as active again. */
   private cancelIdleCheck() {
-    this.idle = false;
+    if (this.inactivityTimerPromise) {
+      this.inactivityTimerPromise.cancel();
+      this.inactivityTimerPromise = null;
+    }
   }
 
   /**
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -251,10 +251,10 @@ export abstract class PersistentStream<
   markIdle(): void {
     this.idle = true;
     this.queue
-      .schedule(() => {
+      .scheduleWithDelay(() => {
         return this.handleIdleCloseTimer();
       }, IDLE_TIMEOUT_MS)
-      .catch((err: FirestoreError) => {
+      .promise.catch((err: FirestoreError) => {
         // When the AsyncQueue gets drained during testing, pending Promises
         // (including these idle checks) will get rejected. We special-case
         // these cancelled idle checks to make sure that these specific Promise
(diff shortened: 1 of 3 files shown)


packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -251,10 +251,10 @@ export abstract class PersistentStream<
   markIdle(): void {
     this.idle = true;
     this.queue
-      .schedule(() => {
+      .scheduleWithDelay(() => {
         return this.handleIdleCloseTimer();
```
schmidt-sebastian: `handleIdleCloseTimer` has its own logic to prevent itself from running. Do you think you could migrate this code to use this new functionality. We already cancel the timer in Android: https://cs.corp.google.com/piper///depot/google3/java/com/google/android/gmscore/dev/thick_client/firebase-firestore/src/com/google/firebase/firestore/remote/AbstractStream.java?rcl=184463491&l=362

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -251,10 +251,10 @@ export abstract class PersistentStream<
...
       }, IDLE_TIMEOUT_MS)
-      .catch((err: FirestoreError) => {
+      .promise.catch((err: FirestoreError) => {
```
schmidt-sebastian: Would it be possible to do something along the lines of https://github.com/alkemics/CancelablePromise/blob/master/CancelablePromise.js?

We might just want to expose .catch/.then, and keep .promise to make more advanced operations possible.

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
 import { Deferred } from './promise';
 import { Code, FirestoreError } from './error';
 
-type DelayedOperation<T> = {
+/** External Result of scheduling a delayed operation. */
+export interface DelayedOperationResult<T> {
+  /** A promise that will resolve once the operation has been run. */
```
schmidt-sebastian: ... and will get rejected when it is cancelled.

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
 import { Deferred } from './promise';
 import { Code, FirestoreError } from './error';
 
-type DelayedOperation<T> = {
+/** External Result of scheduling a delayed operation. */
+export interface DelayedOperationResult<T> {
```
wilhuff: Is there any reason to split the interface and the class?

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
...
+class DelayedOperation<T> implements DelayedOperationResult<T> {
   // tslint:disable-next-line:no-any Accept any return type from setTimeout().
-  handle: any;
-  op: () => Promise<T>;
-  deferred: Deferred<T>;
-};
+  private timerHandle: any;
+  private readonly deferred = new Deferred<T>();
+  /** true if the operation has not been executed or cancelled yet. */
+  private pending = true;
```
schmidt-sebastian: The separation between `pending` and `timerHandle` reads nicer (because of the variable names), but isn't strictly necessary. We could just rely on timerHandle and check whether it is null or not.

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
...
+      this.deferred.reject(
+        new FirestoreError(
+          Code.CANCELLED,
+          'Operation cancelled' + (reason ? ': ' + reason : '')
+        )
+      );
+    }
+  }
+
+  private runIfNecessary(): Promise<void> {
```
schmidt-sebastian: Suggest to use `runIfPending` to match the names in the rest of this class.

packages/firestore/src/util/async_queue.ts
```diff
@@ -118,11 +154,42 @@ export class AsyncQueue {
...
+    op: () => Promise<T>,
+    delayMs: number
+  ): DelayedOperationResult<T> {
+    this.verifyNotFailed();
+
+    const delayedOp = DelayedOperation.createAndSchedule(this, op, delayMs);
+    const index = this.delayedOperations.push(delayedOp);
+
+    delayedOp.promise.catch(err => {}).then(() => {
+      this.delayedOperations.slice(index, 1);
```
schmidt-sebastian: This doesn't seem right. The index might have been changed if another promise got fulfilled before this one.

packages/firestore/src/util/async_queue.ts
```diff
@@ -143,26 +210,14 @@ export class AsyncQueue {
    * scheduled with a delay can be rejected or queued for immediate execution.
    */
   drain(executeDelayedTasks: boolean): Promise<void> {
```
schmidt-sebastian: As a future cleanup, it seems like we could now get rid of this method. Each caller could invalidate its tasks when we call .shutdown().

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
...
-  handle: any;
-  op: () => Promise<T>;
-  deferred: Deferred<T>;
-};
+  private timerHandle: any;
+  private readonly deferred = new Deferred<T>();
+  /** true if the operation has not been executed or cancelled yet. */
+  private pending = true;
+
+  static createAndSchedule<T>(
```
wilhuff: This seems unnecessarily circuitous:

  * scheduleWithDelay calls
  * createAndSchedule
  * which after a while calls scheduleNow(asyncQueue)
  * which calls asyncQueue.schedule

It mixes state in closures and in the classes in ways that are pretty confusing.
  
Relatedly you have `createAndSchedule` and `scheduleNow` as two vastly different operations which just from their names I would surmise as having similar behavior.

I'd suggest the following changes:
  * Make the asyncQueue explicitly a member of DelayedOperation
  * Rename scheduleNow to conflict less with createAndSchedule. Maybe submit or resume?
  * I'd consider just splitting creation of the op from the deferred scheduling so that you can get rid of createAndSchedule altogether. New up your DelayedOperation and then call delay on it.

packages/firestore/src/util/async_queue.ts
```diff
@@ -20,30 +20,103 @@ import { AnyDuringMigration, AnyJs } from './misc';
...
+  get promise(): Promise<T> {
+    return this.deferred.promise;
+  }
+
+  /**
+   * Schedules the operation to run on the provided AsyncQueue if it has not
+   * already been run or cancelled.
+   */
+  scheduleNow(asyncQueue: AsyncQueue): void {
+    this.clearTimeout();
```
wilhuff: Would it be harmful to call `clearTimeout` inside `runIfNecessary`?

As it stands you have three different callbacks here:
  * The lambda that fires when `setTimeout` calls you back
  * Which calls `scheduleNow`
  * Which calls `schedule` with `runIfNecessary.bind(this)`

packages/firestore/src/util/async_queue.ts
mikelehen: * Made `asyncQueue` a member.
* Renamed `scheduleNow()` to a private `handleDelayElapsed()`, and added a new `skipDelay()` method.
* I'm inclined to keep `createAndSchedule()` as it prevents the DelayedOperation from being in a created-but-not-yet-scheduled state.  As-is, `timerHandle` is initially defined and then becomes null once the op has been executed or canceled.  I'd rather not have it go null=>defined=>null with null meaning two different things.  The other option is to have the constructor automatically schedule the op, but I don't really like constructors with side-effects.

By the way, part of the original naming problem is that `AsyncQueue.schedule()` is a bad name. It should be `AsyncQueue.enqueue()`, and then `scheduleNow()` could be `enqueueNow()`. But I don't want to do that in this PR.

packages/firestore/src/util/async_queue.ts
mikelehen: Done. And I merged runIfNecessary into this method.

packages/firestore/src/util/async_queue.ts
mikelehen: Merged this method into scheduleNow (which got renamed to handleDelayElapesd)

packages/firestore/src/util/async_queue.ts
mikelehen: Not sure how that works with the executeDelayedTasks codepath...  which I think we'll actually want to extend to be able to test timeouts better...

packages/firestore/src/util/async_queue.ts
mikelehen: Yeesh. :-( Good catch, thank you!

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -245,16 +246,23 @@ export abstract class PersistentStream<
...
+   * states imply pending network operations.
    */
   markIdle(): void {
-    this.idle = true;
-    this.queue
-      .schedule(() => {
-        return this.handleIdleCloseTimer();
-      }, IDLE_TIMEOUT_MS)
-      .catch((err: FirestoreError) => {
+    if (this.isOpen() && this.inactivityTimerPromise === null) {
```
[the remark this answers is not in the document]
mikelehen: Woops, thanks!

packages/firestore/src/util/async_queue.ts
```diff
@@ -17,33 +17,112 @@
...
+
+  /**
+   * Creates and returns a DelayedOperation that has been scheduled to be
+   * executed on the provided asyncQueue after the provided delayMs.
+   */
+  static createAndSchedule<T>(
+    asyncQueue: AsyncQueue,
+    op: () => Promise<T>,
+    delayMs: number
+  ) {
```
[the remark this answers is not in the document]
mikelehen: Oops, thanks. Done. And opened a bug to make tslint check this for us: b/72988931

PR #493, opened 2018-02-06
Adding no-unused-vars to TSLint

schmidt-sebastian:
This PR adds type checking to tslint and enables "no-unused-vars".

I also added a new auto-fix option to the default scripts (`yarn lint:fix`), which is responsible for some of the changes in here.

Tests are all passing locally for me and `yarn prepare` works in the repo root as well as in packages/firestore

### DIFF
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -514,8 +514,8 @@ export class IndexedDbMutationQueue implements MutationQueue {
     const startRange = IDBKeyRange.lowerBound(indexKey);
     let containsKey = false;
     return documentMutationsStore(txn)
-      .iterate({ range: startRange, keysOnly: true }, (key, _, control) => {
-        const [userID, keyPath] = key;
+      .iterate({ range: startRange, keysOnly: true }, (key, value, control) => {
+        const [userID, keyPath, /*batchID*/ _] = key;
         if (userID === this.userId && keyPath === encodedPath) {
           containsKey = true;
         }
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -222,7 +222,6 @@ export class IndexedDbMutationQueue implements MutationQueue {
       .next(() => {
         const promises: Array<PersistencePromise<void>> = [];
         for (const mutation of mutations) {
-          const encodedPath = EncodedResourcePath.encode(mutation.key.path);
           const indexKey = DbDocumentMutation.key(
             this.userId,
             mutation.key.path,
@@ -379,7 +378,6 @@ export class IndexedDbMutationQueue implements MutationQueue {
       this.userId,
       queryPath
     );
-    const encodedQueryPath = indexPrefix[1];
     const indexStart = IDBKeyRange.lowerBound(indexPrefix);
 
     // Collect up unique batchIDs encountered during a scan of the index. Use a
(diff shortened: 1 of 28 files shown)


packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -638,7 +636,6 @@ export class PersistentWriteStream extends PersistentStream<
   private handshakeComplete_ = false;
 
   constructor(
-    private databaseInfo: DatabaseInfo,
```
mikelehen: FWIW, removing databaseInfo everywhere is probably creating a porting difference (web doesn't need it because of the way connection is plumbed through vs. the way grpc works on iOS / Android), but it's probably fine.

packages/firestore/test/util/equality_matcher.ts
```diff
@@ -85,7 +85,7 @@ export function addEqualityMatcher() {
   afterEach(() => {
     if (originalFunction) {
       use(chai => {
-        const wrappedDefault = _super => {
+        return _super => {
```
mikelehen: How confident are you that this is right?  It doesn't seem right to me...  Seems like this should be calling Assertion.overwriteMethod() to undo what beforeEach() did...  Maybe ping Josh or log a bug?

packages/firestore/test/util/helpers.ts
```diff
@@ -233,7 +232,7 @@ export function bound(
 ): Bound {
   const components: FieldValue[] = [];
   for (const value of values) {
-    const [field, dataValue] = value;
+    const dataValue = value[1];
```
mikelehen: This is less readable. :-/  Is there any other option?  a placeholder instead of `field`?  What if you bump this into the for loop? (i.e. `for (const [field, dataValue] of values) { ... }`

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -517,7 +515,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
     let containsKey = false;
     return documentMutationsStore(txn)
       .iterate({ range: startRange, keysOnly: true }, (key, _, control) => {
-        const [userID, keyPath, batchID] = key;
+        const [userID, keyPath] = key;
```
mikelehen: I don't love this change. Perhaps we should use the ignore-pattern option of  no-unused-variable to exclude _blah variables, and then name this _batchId or something?

packages/firestore/test/util/equality_matcher.ts
```diff
@@ -54,26 +54,30 @@ function customDeepEqual(left, right) {
...
+              customDeepEqual(left, right),
+              'expected #{act} to roughly deeply equal #{exp}',
+              'expected #{act} to not roughly deeply equal #{exp}',
+              left,
+              right,
+              true
+            );
+          } else {
+            _super.apply(this, args);
+          }
```
mikelehen: FWIW, I suspect this code works but is a bit bonkers.  We'll end up overwriting the method multiple times and I think each _super will point to the previous method, so as this code executes repeatedly you'll end up with a chain of _super's pointing to each other and so _super() will call _super() will call _super() until you finally get back to the original function which won't call _super()...  To make things extra bonkers, there will be multiple instances of the isActive variable (one for each time addEqualityMatcher() has been called since the tests started)...

But I don't really care. :-)

schmidt-sebastian: I mostly uploaded this to give Josh a chance to look at this... but now that I have your attention:

- The existing code never actually performed any cleanup and even if it had, this cleanup would only have worked if deepEqual had ever been called. 
- With the new cleanup, some tests started failing that compared Map values directly. I added the equality helper to these tests.
- I removed some of the 'bonkersness' by keepin the original Function around. This should reduce the chain of _super._super._super calls.

PR #496, opened 2018-02-07
Support ns queryparam

tonymeng:
This PR adds support for the `?ns=` query param. This will be used for testing against non-production endpoints.

### DIFF
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -38,6 +38,25 @@ function decodePath(pathString: string): string {
 }
 
 /**
+ * @param {!string} queryString
+ * @return {!Map<string, string>} key value hash
+ */
+function decodeQuery(queryString: string): Map<string, string> {
+  let results = new Map<string, string>();
+  const pieces = queryString.split('&');
+  for (let i = 0; i < pieces.length; i++) {
+    if (pieces[i].length > 0) {
+      try {
+        const piece = decodeURIComponent(pieces[i]);
+        const kv = piece.split('=');
+        results.set(kv[0], kv[1]);
+      } catch (e) {}
+    }
+  }
+  return results;
+}
+
+/**
  *
  * @param {!string} dataURL
  * @return {{repoInfo: !RepoInfo, path: !Path}}
@@ -119,13 +138,21 @@ export const parseURL = function(
       dataURL = dataURL.substring(colonInd + 2);
     }
 
-    // Parse host and path.
+    // Parse host, path, and query string.
     let slashInd = dataURL.indexOf('/');
     if (slashInd === -1) {
       slashInd = dataURL.length;
     }
-    host = dataURL.substring(0, slashInd);
-    pathString = decodePath(dataURL.substring(slashInd));
+    let questionInd = dataURL.indexOf('?');
+    if (questionInd === -1) {
+      questionInd = dataURL.length;
+    }
+    host = dataURL.substring(0, Math.min(slashInd, questionInd));
+    if (slashInd < questionInd) {
+      // For pathString, questionInd will always come after slashInd
+      pathString = decodePath(dataURL.substring(slashInd, questionInd));
+    }
+    let queryParams = decodeQuery(dataURL.substring(Math.min(dataURL.length, questionInd + 1)));
 
     // If we have a port, use scheme for determining if it's secure.
     colonInd = host.indexOf(':');
@@ -147,17 +174,9 @@ export const parseURL = function(
       domain = 'localhost';
     }
     // Support `ns` query param if subdomain not already set
-    const queryStartIndex = host.indexOf('?');
-    if (queryStartIndex > 0) {
-      if (subdomain === '') {
-        const queryString = host.substring(queryStartIndex, host.length);
-        const match = queryString.match(/ns=([a-zA-Z0-9-]+)/);
-        if (match.length > 0) {
-          subdomain = match[1];
-        }
-      }
-      // Always remove all query params from the host
-      host = host.substring(0, queryStartIndex);
+    if (subdomain === '' && queryParams.has('ns')) {
+      subdomain = queryParams.get('ns');
+      console.log("here", queryParams, subdomain);
     }
   }
 
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -153,8 +153,8 @@ export const parseURL = function(
         const queryString = host.substring(queryStartIndex, host.length);
         const match = queryString.match(/ns=([a-zA-Z0-9-]+)/);
         if (match.length > 0) {
-            subdomain = match[1];
-            host = host.substring(0, queryStartIndex);
+          subdomain = match[1];
+          host = host.substring(0, queryStartIndex);
         }
       }
     }
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -146,6 +146,18 @@ export const parseURL = function(
     } else if (parts[0].slice(0, colonInd).toLowerCase() === 'localhost') {
       domain = 'localhost';
     }
+    // Support `ns` query param if subdomain not already set
+    if (subdomain === '') {
+      const queryStartIndex = host.indexOf('?');
+      if (queryStartIndex != -1) {
+        const queryString = host.substring(queryStartIndex, host.length);
+        const match = queryString.match(/ns=([a-zA-Z0-9-]+)/);
+        if (match.length > 0) {
+            subdomain = match[1];
+            host = host.substring(0, queryStartIndex);
+        }
+      }
+    }
   }
 
   return {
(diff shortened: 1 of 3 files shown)


packages/database/test/database.test.ts
```diff
@@ -76,6 +76,20 @@ describe('Database Tests', function() {
...
 
+  it('Can read ns query param', function() {
+    var db = defaultApp.database('http://localhost:80?ns=foo');
+    expect(db).to.be.ok;
+    expect(db.repo_.repoInfo_.namespace).to.equal('foo');
+    expect(db.ref().toString()).to.equal('http://localhost:80/');
+  });
+
+  it('Only reads ns query param when subdomain not set', function() {
+    var db = defaultApp.database('http://bar.firebaseio.com?ns=foo');
```
schmidt-sebastian: Does this code work for `http://bar.firebaseio.com/?ns=foo`? I assume our rock-solid URL scheme parsing will then put this in the `path` instead.

packages/database/src/core/util/libs/parser.ts
```diff
@@ -146,6 +146,19 @@ export const parseURL = function(
     } else if (parts[0].slice(0, colonInd).toLowerCase() === 'localhost') {
       domain = 'localhost';
     }
+    // Support `ns` query param if subdomain not already set
```
schmidt-sebastian: Can we add a `queryString` similar to the `pathString` above?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -146,6 +146,19 @@ export const parseURL = function(
...
+    if (queryStartIndex > 0) {
+      if (subdomain === '') {
+        const queryString = host.substring(queryStartIndex, host.length);
+        const match = queryString.match(/ns=([a-zA-Z0-9-]+)/);
+        if (match.length > 0) {
+          subdomain = match[1];
+        }
+      }
+      // Always remove all query params from the host
+      host = host.substring(0, queryStartIndex);
```
schmidt-sebastian: This should go in line 127.

packages/database/test/database.test.ts
```diff
@@ -76,6 +76,20 @@ describe('Database Tests', function() {
     expect(db.ref().toString()).to.equal('https://localhost/');
   });
 
+  it('Can read ns query param', function() {
+    var db = defaultApp.database('http://localhost:80?ns=foo');
+    expect(db).to.be.ok;
+    expect(db.repo_.repoInfo_.namespace).to.equal('foo');
+    expect(db.ref().toString()).to.equal('http://localhost:80/');
```
schmidt-sebastian: Don't you need to send this namespace to the backend as well?

packages/database/src/core/util/libs/parser.ts
tonymeng: done

packages/database/src/core/util/libs/parser.ts
tonymeng: done

packages/database/test/database.test.ts
tonymeng: it is sent to the database as a query param in the ws connection. It's not visible in `db.ref()`, but the ws connection looks like
`ws://s-local.fblocal.com:9000/.ws?v=5&ns=foo`

packages/database/test/database.test.ts
tonymeng: fixed

packages/database/src/core/util/libs/parser.ts
```diff
@@ -119,13 +138,21 @@ export const parseURL = function(
...
 
-    // Parse host and path.
+    // Parse host, path, and query string.
     let slashInd = dataURL.indexOf('/');
     if (slashInd === -1) {
       slashInd = dataURL.length;
     }
-    host = dataURL.substring(0, slashInd);
-    pathString = decodePath(dataURL.substring(slashInd));
+    let questionInd = dataURL.indexOf('?');
```
schmidt-sebastian: Nit: Can you make this `questionMarkInd`?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -37,6 +37,25 @@ function decodePath(pathString: string): string {
   return pathStringDecoded;
 }
 
+/**
+ * @param {!string} queryString
+ * @return {!Map<string, string>} key value hash
+ */
+function decodeQuery(queryString: string): Map<string, string> {
```
schmidt-sebastian: As I just learned in a Firestore PR, IE 11's Map support is kind of lackluster and so it would be better if you used a normal Object here: `{ [ key: string] : string }`

packages/database/src/core/util/libs/parser.ts
```diff
@@ -37,6 +37,25 @@ function decodePath(pathString: string): string {
...
+  let results = new Map<string, string>();
+  const pieces = queryString.split('&');
+  for (let i = 0; i < pieces.length; i++) {
+    if (pieces[i].length > 0) {
+      try {
+        const piece = decodeURIComponent(pieces[i]);
+        const kv = piece.split('=');
+        results.set(kv[0], kv[1]);
```
schmidt-sebastian: Don't you need to split first and then decode?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -37,6 +37,25 @@ function decodePath(pathString: string): string {
...
 }
 
+/**
+ * @param {!string} queryString
+ * @return {!Map<string, string>} key value hash
+ */
+function decodeQuery(queryString: string): Map<string, string> {
+  let results = new Map<string, string>();
+  const pieces = queryString.split('&');
+  for (let i = 0; i < pieces.length; i++) {
```
schmidt-sebastian: How about:

```
for (const segment of queryString.split('&')) {
   const kv = segment.split('=');

   if (kv.length === 2) {
      results.set(decode(kv[0]), decode(kv[1]));
   }  else {
     ... log warning
   }
}

File: packages/database/src/core/util/libs/parser.ts [hunk 9]
```diff
@@ -119,13 +138,21 @@ export const parseURL = function(
...
       slashInd = dataURL.length;
     }
-    host = dataURL.substring(0, slashInd);
-    pathString = decodePath(dataURL.substring(slashInd));
+    let questionInd = dataURL.indexOf('?');
+    if (questionInd === -1) {
+      questionInd = dataURL.length;
+    }
+    host = dataURL.substring(0, Math.min(slashInd, questionInd));
+    if (slashInd < questionInd) {
```
schmidt-sebastian: Since you are setting `questionInd` to dataUrl.length, can't you always to the substring operation below?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -119,13 +138,21 @@ export const parseURL = function(
...
+      // For pathString, questionInd will always come after slashInd
+      pathString = decodePath(dataURL.substring(slashInd, questionInd));
+    }
+    let queryParams = decodeQuery(dataURL.substring(Math.min(dataURL.length, questionInd + 1)));
```
schmidt-sebastian: Can you let decodeQuery deal with the leading `?` the same way that decodePath does?

packages/database/src/core/util/libs/parser.ts
tonymeng: done

packages/database/src/core/util/libs/parser.ts
tonymeng: yep

packages/database/src/core/util/libs/parser.ts
tonymeng: done

packages/database/src/core/util/libs/parser.ts
tonymeng: no because the SDK appends a trailing `/` to everything prior to getting here so sometimes we get host = `localhost:9000?ns=foo/`. When we try to figure out the pathString for this, it'll give us the queryString instead.

```javascript
'localhost:9000?ns=foo/'.indexOf('?')
14
'localhost:9000?ns=foo/'.indexOf('/')
21
'localhost:9000?ns=foo/'.substring(21, 14);
"?ns=foo"
```

packages/database/src/core/util/libs/parser.ts
tonymeng: done

packages/database/test/database.test.ts
```diff
@@ -76,6 +76,20 @@ describe('Database Tests', function() {
     expect(db.ref().toString()).to.equal('https://localhost/');
   });
 
+  it('Can read ns query param', function() {
+    var db = defaultApp.database('http://localhost:80/?ns=foo');
```
schmidt-sebastian: You can improve your test coverage if you add another param here.

jc275:
https://github.com/firebase/firebase-js-sdk/issues/517

vincentwoo:
Can we have some docs about this? Right now, it seems like the subdomain overrides the explicit `ns` field which is kind of strange. I want to be able to use `ns` to proxy Firebase connection in production, and right now I just have a ton of subdomain rules.

PR #499, opened 2018-02-08
Improve usage and testing of delayed operations.

mikelehen:
Core changes:
* Moves ExponentialBackoff to the AsyncQueue (matches iOS / Android).
* Adds a TimerId enum for identifying delayed operations on the queue and uses it to identify our existing backoff and idle timers.
* Added AsyncQueue.hasDelayedOperation(id) and .runDelayedOperationsEarly(id) which can be used from tests to check for the presence of an operation and to schedule them to run early.
  * Idle tests now use these mechanisms.
  * Spec tests now use this rather than setting initalBackoffDelay to 1ms.
* Reworked mechanism by which DelayedOperation objects get removed from AsyncQueue's delayedOperations list to make sure it happens synchronously.

Cleanup:
* Renamed schedule() to enqueue() and scheduleWithDelay() to enqueueAfterDelay().
* Reorders AsyncQueue.enqueueAfterDelay() arguments to put operation last.

This PR doesn't yet expose timer ids and async queue manipulation to the spec tests, but I suspect that's the direction we'll want to go.

### DIFF
--- a/packages/firestore/src/remote/backoff.ts
+++ b/packages/firestore/src/remote/backoff.ts
@@ -15,8 +15,8 @@
  */
 
 import * as log from '../util/log';
-import { Deferred } from '../util/promise';
-
+import { CancelablePromise } from '../util/promise';
+import { AsyncQueue, TimerId } from '../util/async_queue';
 const LOG_TAG = 'ExponentialBackoff';
 
 /**
@@ -30,8 +30,17 @@ const LOG_TAG = 'ExponentialBackoff';
  */
 export class ExponentialBackoff {
   private currentBaseMs: number;
+  private timerPromise: CancelablePromise<void> | null = null;
 
   constructor(
+    /**
+     * The AsyncQueue to run backoff operations on.
+     */
+    private readonly queue: AsyncQueue,
+    /**
+     * The ID to use when scheduling backoff operations on the AsyncQueue.
+     */
+    private readonly timerId: TimerId,
     /**
      * The initial delay (used as the base delay on the first retry attempt).
      * Note that jitter will still be applied, so the actual delay could be as
@@ -74,10 +83,13 @@ export class ExponentialBackoff {
 
   /**
    * Returns a promise that resolves after currentDelayMs, and increases the
-   * delay for any subsequent attempts.
+   * delay for any subsequent attempts. If there was a pending backoff operation
+   * already, it will be canceled.
    */
-  backoffAndWait(): Promise<void> {
-    const def = new Deferred<void>();
+  backoffAndRun(op: () => Promise<void>): void {
+    if (this.timerPromise !== null) {
+      this.timerPromise.cancel();
+    }
 
     // First schedule using the current base (which may be 0 and should be
     // honored as such).
@@ -89,9 +101,11 @@ export class ExponentialBackoff {
           `(base delay: ${this.currentBaseMs} ms)`
       );
     }
-    setTimeout(() => {
-      def.resolve();
-    }, delayWithJitterMs);
+    this.timerPromise = this.queue.enqueueAfterDelay(
+      this.timerId,
+      delayWithJitterMs,
+      op
+    );
 
     // Apply backoff factor to determine next delay and ensure it is within
     // bounds.
@@ -102,8 +116,6 @@ export class ExponentialBackoff {
     if (this.currentBaseMs > this.maxDelayMs) {
       this.currentBaseMs = this.maxDelayMs;
     }
-
-    return def.promise;
   }
 
   /** Returns a random value in the range [-currentBaseMs/2, currentBaseMs/2] */
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -163,14 +163,14 @@ export abstract class PersistentStream<
 
   constructor(
     private queue: AsyncQueue,
-    backoffTimerId: TimerId,
+    connectionTimerId: TimerId,
     private idleTimerId: TimerId,
     protected connection: Connection,
     private credentialsProvider: CredentialsProvider
   ) {
     this.backoff = new ExponentialBackoff(
       queue,
-      backoffTimerId,
+      connectionTimerId,
       BACKOFF_INITIAL_DELAY_MS,
       BACKOFF_FACTOR,
       BACKOFF_MAX_DELAY_MS
@@ -539,7 +539,7 @@ export class PersistentListenStream extends PersistentStream<
   ) {
     super(
       queue,
-      TimerId.ListenStreamBackoff,
+      TimerId.ListenStreamConnection,
       TimerId.ListenStreamIdle,
       connection,
       credentials
@@ -647,7 +647,7 @@ export class PersistentWriteStream extends PersistentStream<
   ) {
     super(
       queue,
-      TimerId.WriteStreamBackoff,
+      TimerId.WriteStreamConnection,
       TimerId.WriteStreamIdle,
       connection,
       credentials
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -21,7 +21,7 @@ import { ProtoByteString, TargetId } from '../core/types';
 import { QueryData } from '../local/query_data';
 import { Mutation, MutationResult } from '../model/mutation';
 import { assert } from '../util/assert';
-import { AsyncQueue } from '../util/async_queue';
+import { AsyncQueue, TimerId } from '../util/async_queue';
 import { Code, FirestoreError } from '../util/error';
 import * as log from '../util/log';
 
@@ -163,13 +163,15 @@ export abstract class PersistentStream<
 
   constructor(
     private queue: AsyncQueue,
+    backoffTimerId: TimerId,
+    private idleTimerId: TimerId,
     protected connection: Connection,
-    private credentialsProvider: CredentialsProvider,
-    // Used for faster retries in testing
-    initialBackoffDelay?: number
+    private credentialsProvider: CredentialsProvider
   ) {
     this.backoff = new ExponentialBackoff(
-      initialBackoffDelay ? initialBackoffDelay : BACKOFF_INITIAL_DELAY_MS,
+      queue,
+      backoffTimerId,
+      BACKOFF_INITIAL_DELAY_MS,
       BACKOFF_FACTOR,
       BACKOFF_MAX_DELAY_MS
     );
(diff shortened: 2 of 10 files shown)


packages/firestore/src/util/async_queue.ts
```diff
@@ -23,6 +23,18 @@ import { Code, FirestoreError } from './error';
 // tslint:disable-next-line:no-any Accept any return type from setTimeout().
 type TimerHandle = any;
 
+/**
+ * Wellknown "timer" IDs used when scheduling delayed operations on the
+ * AsyncQueue. These IDs can then be used from tests to check for the presence
+ * of operations or to run them early.
+ */
+export enum TimerId {
```
schmidt-sebastian: I'm not sure I am a big fan of this Enum. Can we use well-defined strings kind of like we for the action strings in `Persistence.runTransaction`?  It seems strange that we would have to edit this type everytime we add a new f task.

mikelehen: I started out with a string union (`type TimerId = 'listen_stream_idle'|...`), but it was feeling awkward to have long string constants that had to match between our tests and implementation code.  And since iOS / Java don't have string unions I assume we'd want to use an enum there? So I just went with an enum.

I don't follow "It seems strange that we would have to edit this type everytime we add a new task."  Wouldn't the same be true if we used a string union? Or did you have something else in mind?

FWIW, I think the runTransaction strings are just used for logging, so they don't have the same constraint of being used from multiple places.

packages/firestore/src/remote/backoff.ts
```diff
@@ -30,8 +30,17 @@ const LOG_TAG = 'ExponentialBackoff';
...
 
   constructor(
+    /**
+     * The AsyncQueue to run backoff operations on.
+     */
+    private readonly queue: AsyncQueue,
+    /**
+     * The ID to use when scheduling backoff operations on the AsyncQueue.
+     */
+    private readonly timerId: TimerId,
```
schmidt-sebastian: It would make more sense to me if this was the ID of the underlying operation. The fact that is is being backed-off seems secondary to me.

packages/firestore/src/remote/backoff.ts
```diff
@@ -74,10 +83,13 @@ export class ExponentialBackoff {
...
   /**
    * Returns a promise that resolves after currentDelayMs, and increases the
-   * delay for any subsequent attempts.
+   * delay for any subsequent attempts. If there was a pending backoff operation
+   * already, it will be canceled.
    */
-  backoffAndWait(): Promise<void> {
-    const def = new Deferred<void>();
+  backoffAndRun(op: () => Promise<void>): void {
+    if (this.timerPromise !== null) {
```
schmidt-sebastian: With the old API, we had a way to tell the caller that its operation got cancelled. Right now, if I follow this correctly, we will just silently drop the previous invocation/

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -163,13 +163,15 @@ export abstract class PersistentStream<
 
   constructor(
     private queue: AsyncQueue,
+    backoffTimerId: TimerId,
```
schmidt-sebastian: This name makes it sound like you want to use different IDs for operations that get backed off. I don't think we should do that, and instead use the same ID regardless of how it gets scheduled. 
Do you mind changing this to `connectionTimerId` (or something similar) to indicate the operation type rather than the fact that it is can be run with backoff?

packages/firestore/src/remote/backoff.ts
```diff
@@ -74,10 +83,13 @@ export class ExponentialBackoff {
 
   /**
    * Returns a promise that resolves after currentDelayMs, and increases the
-   * delay for any subsequent attempts.
+   * delay for any subsequent attempts. If there was a pending backoff operation
+   * already, it will be canceled.
    */
-  backoffAndWait(): Promise<void> {
-    const def = new Deferred<void>();
+  backoffAndRun(op: () => Promise<void>): void {
```
schmidt-sebastian: `backoffAndRun` lets you schedule any type of operation here, but the ID already got assigned in the constructor. Should we move the assignment of the ID to here as well?

packages/firestore/src/util/async_queue.ts
```diff
@@ -38,20 +50,42 @@ class DelayedOperation<T> implements CancelablePromise<T> {
   private readonly deferred = new Deferred<T>();
 
   private constructor(
-    private asyncQueue: AsyncQueue,
-    private op: () => Promise<T>
+    private readonly asyncQueue: AsyncQueue,
+    readonly timerId: TimerId,
+    readonly targetTime: number,
```
schmidt-sebastian: TargetTime is pretty unspecific. Should me make this a Date or call this `targetTimeMs`?

packages/firestore/src/util/async_queue.ts
```diff
@@ -38,20 +50,42 @@ class DelayedOperation<T> implements CancelablePromise<T> {
   private readonly deferred = new Deferred<T>();
 
   private constructor(
-    private asyncQueue: AsyncQueue,
-    private op: () => Promise<T>
+    private readonly asyncQueue: AsyncQueue,
+    readonly timerId: TimerId,
```
schmidt-sebastian: Should this name actually be `OperationId` or something similar?

packages/firestore/src/util/async_queue.ts
```diff
@@ -178,25 +206,35 @@ export class AsyncQueue {
...
+    const delayedOp = DelayedOperation.createAndSchedule(
+      this,
+      timerId,
+      delayMs,
+      op,
+      op => this.removeDelayedOperation(op)
+    );
     this.delayedOperations.push(delayedOp);
 
-    delayedOp.catch(err => {}).then(() => {
```
schmidt-sebastian: I think this way of dealing with the dequeuing was much more straightforward (albeit you need to re-throw the error)

packages/firestore/src/util/async_queue.ts
```diff
@@ -223,17 +261,55 @@ export class AsyncQueue {
...
+    // Note that draining may generate more delayed ops, so we do that first.
+    return this.drain().then(() => {
+      assert(
+        lastTimerId === undefined || this.hasDelayedOperation(lastTimerId),
+        `Attempted to drain to missing operation ${lastTimerId}`
+      );
+
+      // Run ops in the same order they'd run if they ran naturally.
+      this.delayedOperations.sort((a, b) => a.targetTime - b.targetTime);
+
```
schmidt-sebastian: This seems like a lot of hoops to go through that (at least for now) I don't cleary see the immediate benefit of. For unrelated operations scheduled with delay, it seems fine to a) run them out of order (removing the need for target time) and b) only run them when calling `runDelayedOperationsEarly`.

packages/firestore/test/integration/api_internal/idle_timeout.test.ts
```diff
@@ -25,7 +26,9 @@ apiDescribe('Idle Timeout', persistence => {
       return docRef
         .set({ foo: 'bar' })
         .then(() => {
-          return drainAsyncQueue(db);
+          return asyncQueue(db).runDelayedOperationsEarly(
+            TimerId.WriteStreamIdle
```
schmidt-sebastian: This would also read nicer to me if this was a constant on WriteStream.

packages/firestore/src/util/async_queue.ts
```diff
@@ -223,17 +261,55 @@ export class AsyncQueue {
...
-      } else {
-        delayedOp.cancel('shutdown');
+  drain(): Promise<void> {
+    return this.enqueue(() => Promise.resolve());
+  }
+
+  /**
+   * For Tests: Determine if a particular delayed operation exists.
+   */
+  hasDelayedOperation(timerId: TimerId): boolean {
```
schmidt-sebastian: Don't feel strongly, but maybe use ContainsDelayedOperation here since has indicates a 1:1 mapping?

packages/firestore/test/integration/api_internal/idle_timeout.test.ts
schmidt-sebastian: I was going to suggest a String but I think you already changed my mind of that.

packages/firestore/src/util/async_queue.ts
```diff
@@ -23,6 +23,18 @@ import { Code, FirestoreError } from './error';
...
+/**
+ * Wellknown "timer" IDs used when scheduling delayed operations on the
+ * AsyncQueue. These IDs can then be used from tests to check for the presence
+ * of operations or to run them early.
+ */
+export enum TimerId {
+  ListenStreamIdle,
+  ListenStreamConnection,
+  WriteStreamIdle,
+  WriteStreamConnection
```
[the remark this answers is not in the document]
mikelehen: If we do that then we have to allow duplicates on the queue and it becomes ambiguous which one tests are manipulating, so I'm inclined to keep them separate. But I agree the stream constructor ends up awkward. :-/  I considered having them be abstract properties on the stream instead (so each subclass would need to define them appropriately), but I'm not sure that's actually better, and abstract doesn't port super great to Obj-C.

packages/firestore/src/util/async_queue.ts
mikelehen: Sure, sounds good.

packages/firestore/src/util/async_queue.ts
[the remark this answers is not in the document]
wilhuff: This strikes me as pretty ugly too but I don't think reducing the number of states is worthwhile.

What if we fully constructed the exponential backoff and passed that in to the remote store instead of constructing it here?

Similarly can we encapsulate the idle id and duration into a "delayed runner"? That way the streams just have these dumb interfaces to delay tasks without having to distinguish how long to delay or which id to use where.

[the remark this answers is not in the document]
mikelehen: I'm not 100% sure what you're proposing.  We could do something like:

1. Rename `TimerId` enum to `TimeoutId`
2. Add a `Timeout` tuple class (encapsulating "idle id" and "duration" as you suggested):
```
class Timeout {
  constructor(public delayMs: number, public timeoutId: TimoutId) { }
}
```
3. Change 
`asyncQueue.enqueueAfterDelay(timerId: TimerId, delayMs: number, op: () => Promise<T>)` 
to: `asyncQueue.enqueueAfterTimeout(timeout: Timeout, op: () => Promise<T>)`

4. Change PersistentStream constructor to accept an ExponentialBackoff object and Timeout object:
```
  constructor(
    private queue: AsyncQueue,
    connectionTimeout: Timeout,
    // or instead: connectionBackoff: ExponentialBackoff,
    private idleTimeout: Timeout,
    protected connection: Connection,
    private credentialsProvider: CredentialsProvider
  ) { ... }
```

I don't have strong feelings about any of this, but it doesn't reduce the number of arguments to PersistentStream's constructor.

If you're proposing we actually have a DelayedRunner class/interface that encapsulates an AsyncQueue + Timeout so we can pass it into the stream and it can do `idleTimeoutRunner.run(op)`, we _could_... but going the next step and trying to make ExponentialBackoff implement DelayedRunner as well would require surgery to move all the .reset() and .resetToMax() calls out of PersistentStream and I'm not sure it is feasible or makes sense.

And at the end of the day we're still passing in an "idleBLAH" and "connectionBLAH" to the PersistentStream constructor (where "BLAH" could be "TimerId" (today) or "Timeout" or "DelayedRunner").

So based on my understanding of your suggestion, I am not really excited...  but let me know if my understanding missed.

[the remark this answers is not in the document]
wilhuff: Nope, you understood what I meant. And yeah, it's not much of an improvement.

Issue #498, opened 2018-02-08
firebase/index.d.ts contains wrong type for DocumentSnapshot data()


wereHamster:
DocumentSnapshot data() return type should be nullable (or rather, undefined-able). The same underlying issue has already been fixed in #364, but in a different file. There seems to be duplication of code between packages/firestore-types/index.d.ts and packages/firebase/index.d.ts.

```javascript
const ds: DocumentSnapshot = undefined as any
ds.data().foo // <- should be a type error
```

samtstern:
@schmidt-sebastian can you look into this?

PR #500, opened 2018-02-08
Updating firebase/index.d.ts

schmidt-sebastian:
Updating the types in firebase/index.d.ts to pull in the recent API changes.

Fixes: https://github.com/firebase/firebase-js-sdk/issues/498

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -699,7 +699,7 @@ declare namespace firebase.firestore {
     /**
      * The `FirebaseApp` associated with this `Firestore` instance.
      */
-    app: any;
+    app: firebase.app.App;
 
     /**
      * Re-enables use of the network for this Firestore instance after a prior
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -599,14 +599,14 @@ declare namespace firebase.firestore {
    * mapped to values.
    */
   export type DocumentData = { [field: string]: any };
-
+  
   /**
    * Update data (for use with `DocumentReference.update()`) consists of field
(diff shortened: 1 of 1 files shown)


packages/firebase/index.d.ts
```diff
@@ -617,15 +617,14 @@ declare namespace firebase.firestore {
 
   export type LogLevel = 'debug' | 'error' | 'silent';
 
-  function setLogLevel(logLevel: LogLevel): void;
+  export function setLogLevel(logLevel: LogLevel): void;
```
jshcrowthe: This is not needed as it is part of the `firebase` namespace which is exported at the end.

packages/firebase/index.d.ts
```diff
@@ -698,9 +697,27 @@ declare namespace firebase.firestore {
     batch(): WriteBatch;
 
     /**
-     * The `firebase.app.App` associated with this `Firestore` instance.
+     * The `FirebaseApp` associated with this `Firestore` instance.
      */
-    app: firebase.app.App;
+    app: any;
```
jshcrowthe: Why did we remove the type?

schmidt-sebastian: Sorry, that was unintended.

packages/firebase/index.d.ts
schmidt-sebastian: Before I change this, do you know if this is causing a problem?

I would really like if these typings matched the ones in the firestore.d.ts file and having this extra 'export' seems like a step in that direction.

jshcrowthe: This is fine. I looked and it seems to have no impact.

PR #501, opened 2018-02-08
[Multi-Tab] Master Election

schmidt-sebastian:
This PR adds "master election" to Firestore (which isn't quite as complicated as it sounds thanks to IndexedDb).

The PR lays the groundwork for two different persistence modes: Some operations can require the primary lease, and some operations can run without it. Hence, the `runTransaction` function now takes in a parameter to indicate whether it needs the primary lease. A lot of the code in this PR is formatting changes due to this additional argument, and if deemed reasonable, I can split this up and move it to a different PR.

I am also adding Spec test for multi-client workflows. I added a new spec builder that deals with multi-client tests. There's a lot of code in there that doesn't do anything but change the return value of the super class, and if we combine the MultiClientSpecBuilder and the SpecBuilder, this would go away. The main benefit of the separation is that we can easily hide this from Android and iOS.

I also added a `MockPlatform` to override some of the `document` functionality to use it in testing. This allows me to alter its behavior per client.


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -89,17 +99,25 @@ const UNSUPPORTED_PLATFORM_ERROR_MSG =
  * owner lease immediately regardless of the current lease timestamp.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -265,13 +392,14 @@ export class IndexedDbPersistence implements Persistence {
...
    * against losing the owner lease.
    */
-  private ensureOwnerLease(txn: SimpleDbTransaction): PersistencePromise<void> {
-    const store = txn.store<DbOwnerKey, DbOwner>(DbOwner.store);
-    return store.get('owner').next(dbOwner => {
-      if (dbOwner === null || dbOwner.ownerId !== this.ownerId) {
+  private ensurePrimaryLease(
+    txn: SimpleDbTransaction
+  ): PersistencePromise<void> {
+    return this.tryAcquirePrimaryLease(txn).next(owner => {
```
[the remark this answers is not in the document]
schmidt-sebastian: I rewrote this entire section based on our HipChat discussion. I only acquire the lease at the end, but verify that I can obtain it at the beginning.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -280,63 +408,60 @@ export class IndexedDbPersistence implements Persistence {
     });
   }
 
+  /** Verifies that that `updateTimeMs` is within CLIENT_STATE_MAX_AGE_MS. */
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

packages/firestore/test/unit/specs/persistence_spec.test.ts
```diff
@@ -182,4 +182,50 @@ describeSpec('Persistence:', ['persistence'], () => {
         })
     );
   });
+
+  specTest('Single tab acquires primary lease', ['multi-client'], () => {
+    // This test simulates primary state handoff between two background tabs.
+    // With all instances are in the background, the first active tab acquires
+    // ownership.
+    return client(0)
+      .becomeHidden()
```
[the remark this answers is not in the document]
schmidt-sebastian: This both necessary and the default. We need to schedule a step before we can verify expectations.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +170,143 @@ export class IndexedDbPersistence implements Persistence {
...
+            }
+            return electedPrimaryCandidate;
+          }
+        );
+      })
+      .then(electedPrimaryCandidate => {
+        const isPrimary = this.isLocalClient(electedPrimaryCandidate);
+        if (this.isPrimary !== isPrimary) {
+          this.isPrimary = isPrimary;
+          this.primaryStateListener(this.isPrimary);
```
mikelehen: I'm glad this is now happening on the AsyncQueue but I'm still concerned about component reentrancy. That is, SyncEngine could call into LocalStore which calls into IndexedDbPersistence.runTransaction() which hits this code which calls back into SyncEngine and now SyncEngine starts doing something "new" even though it was already in the middle of doing something else...  Generally we have not had components call back and forth into each other like that and allowing it will lead to complexity and bugs.

Perhaps we could solve it by just throwing a `queue.enqueue(() => ... )` around the notification (perhaps create a notifyPrimaryStateListener() helper method that does that).

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -61,6 +68,16 @@ export class MemoryPersistence implements Persistence {
...
   }
 
+  tryBecomePrimary(): Promise<boolean> {
+    // All clients using memory persistence act as primary.
+    return Promise.resolve(true);
+  }
+
+  setPrimaryStateListener(primaryStateListener: PrimaryStateListener) {
+    // All clients using memory persistence act as primary.
+    this.queue.enqueue(() => primaryStateListener(true));
```
mikelehen: If you keep the IndexedDb code that calls this with false on shutdown I'd suggest that you do the same for MemoryPersistence.  But really, I think we can probably just treat shutdown as a special case and not do the notification.

packages/firestore/src/local/persistence.ts
```diff
@@ -31,6 +31,18 @@ import { RemoteDocumentCache } from './remote_document_cache';
...
 
+/**
+ * Callback type for primary state notifications. This callback can be
+ * registered with the persistence layer to get notified when we transition from
+ * primary to secondary state and vice versa.
+ *
+ * Note: Instances can only toggle between Primary and Secondary state if
+ * IndexedDB persistence is enabled and multiple clients are active. If this
+ * listener is registered with MemoryPersistence, the callback will be called
+ * exactly once marking the current instance as Primary.
```
mikelehen: If we keep the call in shutdown(), then we should update this comment.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -407,12 +575,32 @@ export class IndexedDbPersistence implements Persistence {
...
   }
 
-  private zombiedOwnerLocalStorageKey(): string {
-    return this.localStoragePrefix + ZOMBIE_OWNER_LOCALSTORAGE_SUFFIX;
+  private zombiedClientLocalStorageKey(): string {
+    return this.localStoragePrefix + ZOMBIE_PRIMARY_LOCALSTORAGE_SUFFIX;
   }
 
-  private generateOwnerId(): string {
+  private generateClientKey(): ClientKey {
```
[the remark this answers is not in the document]
mikelehen: Oh. I found a todo about renaming ClientKey.  I'm fine with "finishing" these renames in a separate PR... let's just make sure it's all-encompassing.

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -41,14 +41,12 @@ const KNOWN_TAGS = [
...
-  tags: string[],
-  persistenceEnabled: boolean
-) => tags.indexOf(MULTI_CLIENT_TAG) !== -1 && persistenceEnabled;
-
-const WEB_SPEC_TEST_FILTER = (tags: string[]) =>
-  tags.indexOf(NO_WEB_TAG) === -1;
+const WEB_SPEC_TEST_FILTER = (tags: string[], persistenceEnabled: boolean) => {
+  return (
+    tags.indexOf(NO_WEB_TAG) === -1 &&
+    (tags.indexOf(MULTI_CLIENT_TAG) === -1 || persistenceEnabled)
```
mikelehen: Seeing this code now it looked incorrect that we don't check for PERSISTENCE_TAG too, and in fact we don't check it anywhere, and it doesn't exist on iOS or Android either. I think it was a temporary measure that we used before all the platforms had persistence.  Can you remove the tag here (and its usage as 'persistence' in persistence_spec.test.ts)?  Else I can do it against master, but it'll probably cause a conflict with this PR.

packages/firestore/test/unit/specs/persistence_spec.test.ts
```diff
@@ -197,7 +197,7 @@ describeSpec('Persistence:', ['persistence'], () => {
       .shutdown()
       .expectPrimaryState(false)
       .client(1)
-      .restart()
+      .tryAcquirePrimaryLease()
```
mikelehen: Once https://github.com/firebase/firebase-js-sdk/pull/412 lands, you can merge it and replace this with

```
.runTimer(TimerId.ClientStateRefresh)
```

Can you add a TODO somewhere?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -788,6 +792,12 @@ abstract class TestRunner {
     await this.remoteStore.enableNetwork();
   }
 
+  private async doAcquirePrimaryLease(): Promise<void> {
+    expect(this.queue.containsDelayedOperation(TimerId.ClientStateRefresh)).to
+      .be.true;
```
mikelehen: Doesn't matter since this code should go away, but FYI- runDelayedOperationsEarly() will throw if the timer doesn't exist, so this check should be unnecessary.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1236,16 +1245,18 @@ export interface SpecStep {
...
    * components. This allows you to queue writes, get documents into cache,
    * etc. and then simulate an app restart.
    */
-  restart?: boolean;
+  restart?: true;
 
   /** Shut down the client and close it network connection. */
-  shutdown?: boolean;
-
+  shutdown?: true;
```
mikelehen: nice. :-)

schmidt-sebastian: Ack.

packages/firestore/test/unit/specs/persistence_spec.test.ts
schmidt-sebastian: Added TODO to spec_builder.ts's tryAcquirePrimaryLease.

packages/firestore/src/local/indexeddb_persistence.ts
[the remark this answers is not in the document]
schmidt-sebastian: Yep, will do. :)

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -208,135 +402,104 @@ export class IndexedDbPersistence implements Persistence {
...
-        // refresh the lease.
-        log.error(reason);
-        this.stopOwnerLeaseRefreshes();
+  private attachVisibilityHandler(): void {
+    this.documentVisibilityHandler = () => {
+      this.queue.enqueue<DbOwner | void>(() => {
+        const inForeground = this.document.visibilityState === 'visible';
+        if (inForeground !== this.inForeground) {
+          this.inForeground = inForeground;
+          return this.tryBecomePrimary();
```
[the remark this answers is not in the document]
schmidt-sebastian: Method was renamed.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +170,143 @@ export class IndexedDbPersistence implements Persistence {
...
+          this.schedulePrimaryLeaseRefreshes(nextDelayMs);
+        });
+      }
+    );
+  }
+
+  /**
+   * Attempts to obtain the primary lease for the local client. If successful,
+   * returns a `DbOwner` with the local client id. Otherwise, returns the
+   * current leaseholder or 'null' (if the primary lease should be obtained
```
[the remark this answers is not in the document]
schmidt-sebastian: This comment has been moved to `canActAsPrimary` and now reads:
```
* Evaluate the state of all active instances and determine whether the local
* client is or can act as the holder of the primary lease. Returns whether
* the client is eligible for the lease, but does not actually acquire it.
* May return 'false' even if there is no active leaseholder and another
* (foreground) client should become leaseholder instead.
```

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +170,143 @@ export class IndexedDbPersistence implements Persistence {
...
+  setPrimaryStateListener(primaryStateListener: PrimaryStateListener) {
+    this.primaryStateListener = primaryStateListener;
+    primaryStateListener(this.isPrimary);
+  }
+
+  /**
+   * Schedules a recurring timer to refresh or acquire the primary lease.
+   */
+  private schedulePrimaryLeaseRefreshes(initialDelayMs: number): Promise<void> {
+    return this.queue.enqueueAfterDelay(
```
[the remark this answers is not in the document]
schmidt-sebastian: The method now returns void. `updateClientMetadataAndTryBecomePrimary` is the only method that returns a Promise now, and the return value has been documented.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +170,143 @@ export class IndexedDbPersistence implements Persistence {
...
+          let nextDelayMs;
+
+          if (this.isLocalClient(currentPrimary)) {
+            nextDelayMs = PRIMARY_LEASE_MAX_REFRESH_INTERVAL_MS;
+          } else if (currentPrimary !== null) {
+            nextDelayMs = Math.max(
+              PRIMARY_LEASE_MIN_REFRESH_INTERVAL_MS,
+              Date.now() -
+                (currentPrimary.leaseTimestampMs + CLIENT_STATE_MAX_AGE_MS) +
+                1
```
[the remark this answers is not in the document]
schmidt-sebastian: I added comments in the code to clarify this a bit.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1167,11 +1169,19 @@ export async function runSpec(
...
-    let stepcnt = 0;
     await sequence(steps, async step => {
+      ++count;
+      lastStep = step;
       return runners[step.clientIndex || 0].run(step);
     });
+  } catch (err) {
+    console.warn(
+      `Spec test failed at step ${count}: ${JSON.stringify(lastStep)}`
+    );
```
mikelehen: You are my hero. :-)

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +174,187 @@ export class IndexedDbPersistence implements Persistence {
...
+   */
+  private updateClientMetadataAndTryBecomePrimary(): Promise<number> {
+    return this.simpleDb.runTransaction('readwrite', ALL_STORES, txn => {
+      const metadataStore = clientMetadataStore(txn);
+      metadataStore.put(
+        new DbClientMetadata(this.clientKey, Date.now(), this.inForeground)
+      );
+
+      return this.getCurrentPrimary(txn).next(currentPrimary => {
+        return this.canActAsPrimary(currentPrimary, txn).next(
```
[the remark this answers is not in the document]
schmidt-sebastian: I kept the name since `canActAsPrimary` no longer takes an argument (I inlined getCurrentPrimary as suggested).

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +174,187 @@ export class IndexedDbPersistence implements Persistence {
...
       .then(db => {
         this.simpleDb = db;
       })
-      .then(() => this.tryAcquireOwnerLease())
       .then(() => {
-        this.scheduleOwnerLeaseRefreshes();
+        this.attachVisibilityHandler();
         this.attachWindowUnloadHook();
+        return this.updateClientMetadataAndTryBecomePrimary().then(leaseTtl =>
+          this.scheduleClientMetadataAndPrimaryLeaseRefreshes(leaseTtl)
```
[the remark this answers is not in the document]
schmidt-sebastian: I kept the name. The comment regarding `leaseTtl` is no longer applicable.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +174,187 @@ export class IndexedDbPersistence implements Persistence {
...
+              return this.acquireOrExtendPrimaryLease(txn).next(
+                () => CLIENT_METADATA_REFRESH_INTERVAL_MS
+              );
+            } else if (currentPrimary != null) {
+              // If another client currently holds the lease, we use a custom
+              // refresh interval and schedule a lease refresh immediately after
+              // the current lease is expected to expire.
+              const remainingLifetimeMs =
+                Date.now() -
+                (currentPrimary.leaseTimestampMs + CLIENT_METADATA_MAX_AGE_MS);
```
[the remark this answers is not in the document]
schmidt-sebastian: Oh, yes, that's true. I did go though and remove this logic altogether though (bringing this aspect back the very first iteration of this PR). If we do need faster lease handover, we can add this in the feature and then also figure out how to properly test it.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +174,187 @@ export class IndexedDbPersistence implements Persistence {
...
+              return Math.min(
+                MINIMUM_REFRESH_INTERVAL_MS,
+                remainingLifetimeMs + 1
+              );
```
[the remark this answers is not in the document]
schmidt-sebastian: As stated above, I went ahead with your suggestion and removed the synchronization aspect of this PR. This simplifies most of the handling that I have not quite been able to get right (with all of it untested). This change removes `leaseTtl` and `getCurrentOwner()`.

I also added the assert.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -178,11 +387,45 @@ export class IndexedDbPersistence implements Persistence {
...
-      return this.ensureOwnerLease(txn).next(() => operation(txn));
+      if (requirePrimaryLease) {
+        // While we merely verify that we have (or can acquire) the lease
+        // immediately, we wait to extend the primary lease until after
+        // executing transactionOperation(). This ensures that even if the
+        // transactionOperation takes a long time, we'll use a recent
+        // leaseTimestampMs in the extended (or newly acquired) lease.
+        return this.getCurrentPrimary(txn)
+          .next(currentPrimary => this.canActAsPrimary(currentPrimary, txn))
+          .next(isPrimary => {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

schmidt-sebastian:
Sorry, I pushed the code changes before I got the chance to answer inline. All comments addressed.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +167,146 @@ export class IndexedDbPersistence implements Persistence {
...
+  }
+
+  /**
+   * Updates the client metadata in IndexedDb and attempts to either obtain or
+   * extend the primary lease for the local client. Asynchronously notifies the
+   * primary state listener if the client either newly obtained or released its
+   * primary lease.
+   *
+   * @return {Promise<number>} A Promise that resolves with the interval in ms
+   * after which the metadata and primary lease needs to be refreshed.
```
mikelehen: stale return comment

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -132,23 +167,146 @@ export class IndexedDbPersistence implements Persistence {
...
+    );
+  }
+
+  /** Checks whether `client` is the local client. */
+  private isLocalClient(client: DbOwner | null): boolean {
+    return client ? client.ownerId === this.clientKey : false;
+  }
+
+  /**
+   * Evaluate the state of all active instances and determine whether the local
```
mikelehen: instances => clients

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -646,6 +699,248 @@ export class SpecBuilder {
   }
 }
 
-export function spec() {
+/**
+ * SpecBuilder that supports serialized interactions between different clients.
+ *
+ * Use `client(clientIndex)` to switch between clients.
+ */
+export class MultiClientSpecBuilder extends SpecBuilder {
```
mikelehen: Sorry, I never looked at this until now, but why do we need an inherited class for this?  I am not wild about needing to override every method...  that kinda' defeats the purpose of inheritance.

In an effort to avoid spinning on this PR forever, I'd be okay leaving this as-is with a TODO(multitab) to revisit (ideally in the near-term).

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1037,31 +1158,58 @@ export async function runSpec(
...
+    throw err;
   } finally {
-    await runner.shutdown();
+    for (const runner of runners) {
+      await runner.shutdown();
+    }
+    if (usePersistence) {
+      await IndexedDbTestRunner.destroyPersistence();
+    } else {
+      await MemoryTestRunner.destroyPersistence();
```
mikelehen: Given we're no longer relying on TestRunner.destroyPersistence() as a virtual method implemented by both runners, you can probably just delete MemoryTestRunner.destroyPersistence().

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1037,31 +1158,58 @@ export async function runSpec(
...
+  /** The number of active clients for this test run. */
+  numClients: number;
 }
 
 /**
  * Union type for each step. The step consists of (mostly) exactly one `field`
  * set and optionally expected events in the `expect` field.
  */
 export interface SpecStep {
+  /** The index of the current client for multi-client spec tests. */
```
mikelehen: Can you add "PORTING NOTE: Only used by web multi-tab tests." or similar to the new steps you added (except maybe shutdown).

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1190,6 +1346,11 @@ export interface SpecWatchEntity {
   removedTargets?: TargetId[];
 }
 
+export type SpecClientState = {
```
mikelehen: Can you add a "PORTING NOTE: Only used by web multi-tab tests." ?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1251,6 +1412,10 @@ export interface StateExpectation {
   watchStreamRequestCount?: number;
   /** Current documents in limbo. Verified in each step until overwritten. */
   limboDocs?: string[];
+  /**
+   * Whether the instance holds the primary lease. Used in multi-client tests.
```
mikelehen: Can you make this a PORTING NOTE ?

packages/firestore/test/unit/specs/spec_builder.ts
schmidt-sebastian: I was torn on this as well (and still am). I did this because
- It makes it obvious what functionality does not need to be ported.
- The whole builder aspect makes much more sense for a single client. `clientA().build().build(`) and `clientB().build().build()` should be independent. We could do this with an improved version MultiSpecBuilder and swap out the underlying instance of SpecBuilder every time `client(...)` gets called.

That being said, I am open to remove it for now (in a follow up).

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Added here and in a bunch of other places.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Added here and in a bunch of other places.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Added here and in a bunch of other places.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Removed.

mikelehen:
🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢🚢

PR #510, opened 2018-02-13
Keep track of number of queries in the query cache

gsoltis:
- Adds a `targetCount` to the global target metadata
- Splits `QueryCache` `add` and `update` methods so that we can update the count appropriately.
- Adds `count()` method to `QueryCache` to report the number of targets currently in the cache
- Adds some schema migration infrastructure, partially courtesy of @schmidt-sebastian 
- Adds a schema migration to ensure that the global target metadata exists, and then updates it with the count of any targets that might already exist in the persistence layer.
- Changes `LocalStore` to call `updateQueryData` instead of `addQueryData` in the appropriate place.

iOS counterpart: https://github.com/firebase/firebase-ios-sdk/pull/776


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
+ * local feature development.
+ */
+export function createOrUpgradeDb(
+  db: IDBDatabase,
+  txn: IDBTransaction,
+  fromVersion: number,
+  toVersion: number
+): PersistencePromise<void> {
+  // This function currently supports migrating to schema version 1 (Mutation
+  // Queue, Query and Remote Document Cache) and schema version 2 (Multi-Tab).
```
schmidt-sebastian: This comment should be more GC specific and less Multi-Tab specific.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,73 @@ export class IndexedDbQueryCache implements QueryCache {
...
-      this.metadata.highestTargetId = targetId;
-      return addedQueryPromise.next(() =>
-        globalTargetStore(transaction).put(DbTargetGlobal.key, this.metadata)
-      );
-    } else {
-      return addedQueryPromise;
+    return targetsStore(transaction).put(this.serializer.toDbTarget(queryData));
+  }
+
+  private updateMetadataForQueryData(queryData: QueryData): boolean {
```
schmidt-sebastian: I am not a big fan of this method name.

1) It seems very iOSy.
2) The name suggests that it also writes the metadata to IndexedDb.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,73 @@ export class IndexedDbQueryCache implements QueryCache {
...
+    );
+  }
+
+  addQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
+    return this.saveQueryData(transaction, queryData).next(() => {
+      this.metadata.targetCount += 1;
+      this.updateMetadataForQueryData(queryData);
```
schmidt-sebastian: I feel like 'updateMetadataForQueryData' should update all the metadata, including the target count.... and at that point it might as well persist to IndexedDB as well.

config/karma.base.js
```diff
@@ -101,7 +101,7 @@ const config = {
   singleRun: false,
 
   client: {
-    mocha: {timeout: 20000, retries: 3},
+    mocha: { timeout: 20000, retries: 3 },
```
schmidt-sebastian: Can you a) make sure you are in sync with master and b) run `yarn` from the repo root?

This might get rid of this style update.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
-  // TODO(mikelehen): Get rid of "as any" if/when TypeScript fixes their
-  // types. https://github.com/Microsoft/TypeScript/issues/14322
-  db.createObjectStore(
-    DbMutationBatch.store,
-    // tslint:disable-next-line:no-any
-    { keyPath: DbMutationBatch.keyPath as any }
-  );
+/**
+ * Schema Version for the Web client (containing the Mutation Queue, the Query
+ * and the Remote Document Cache) and Multi-Tab Support.
```
schmidt-sebastian: Please update comment.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
-  // NOTE: keys for these stores are specified explicitly rather than using a
-  // keyPath.
-  db.createObjectStore(DbDocumentMutation.store);
-  db.createObjectStore(DbRemoteDocument.store);
-  db.createObjectStore(DbOwner.store);
-  db.createObjectStore(DbTargetGlobal.store);
+  let p = PersistencePromise.resolve();
+  if (fromVersion < 2 && toVersion >= 2) {
+    p = ensureTargetGlobal(p, txn);
+    p = addTargetCount(p, txn);
```
schmidt-sebastian: This feels really weird. Can you not just do:

```
if (...)
return ensureTargetGlobal(txn).next(addTargetCount(txn))
else
return .resolve()
```

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,100 @@ export class DbTargetGlobal {
      * until the backend has caught up to this snapshot version again. This
      * prevents our cache from ever going backwards in time.
      */
-    public lastRemoteSnapshotVersion: DbTimestamp
+    public lastRemoteSnapshotVersion: DbTimestamp,
+    /**
+     * A cache of the number of targets persisted.
```
schmidt-sebastian: Nit: The number of targets persisted.

This doesn't strike me as a cache (especially since we have use the word Cache extensively in this file).

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,100 @@ export class DbTargetGlobal {
...
- * The list of all IndexedDB stored used by the SDK. This is used when creating
- * transactions so that access across all stores is done atomically.
+ * Counts the number of targets persisted and adds that value to the target
+ * global singleton.
+ * @param {PersistencePromise<void>} p The promise used to keep the transaction open
+ * @param {IDBTransaction} txn The version upgrade transaction for indexeddb
+ * @return {PersistencePromise<void>} A promise of the operations to add the count to the metadata row.
  */
-export const ALL_STORES = [
+function addTargetCount(
```
schmidt-sebastian: `saveTargetCount`/`persistTargetCount` ?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,100 @@ export class DbTargetGlobal {
...
+    .next((count: number) =>
+      wrapRequest<DbTargetGlobal>(globalStore.get(DbTargetGlobal.key)).next(
+        (metadata: DbTargetGlobal) => {
+          metadata.targetCount = count;
+          return metadata;
+        }
+      )
+    )
+    .next((metadata: DbTargetGlobal) =>
+      wrapRequest<void>(globalStore.put(metadata, DbTargetGlobal.key))
```
schmidt-sebastian: You can combine these two Promise branches.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,100 @@ export class DbTargetGlobal {
...
+    );
+}
+
+/**
+ * Ensures that the target global singleton row exists by adding it if it's missing.
+ * @param {PersistencePromise<void>} p The promise used to keep the transaction open
+ * @param {IDBTransaction} txn The version upgrade transaction for indexeddb
+ * @return {PersistencePromise<void>} A promise of operations to ensure that the metadata row exists
+ */
+function ensureTargetGlobal(
```
schmidt-sebastian: This function could also return a targetGlobal, which you could then use in addTargetCount.

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    }).then(() =>
+      withDb(2, db => {
+        expect(db.version).to.be.equal(2);
+        expect(getAllObjectStores(db)).to.have.members(ALL_STORES);
+        const txn = db.transaction([DbTargetGlobal.store], 'readonly');
+        const req = txn
+          .objectStore(DbTargetGlobal.store)
+          .get(DbTargetGlobal.key);
+        return wrapRequest<DbTargetGlobal>(req)
+          .toPromise()
```
schmidt-sebastian: Consider wrapping this in SimpleDb, which would remove some of the complexity here.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Combined a little.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: I restructured it a little, but I think it's a good idea to keep the promise around, as subsequent migrations that are added will want to chain onto it.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: We can't update the target count and share the logic between `addQueryTarget` and `updateQueryTarget`. One updates, the other adds, so the change to `targetCount` is different. We also can't tell which is which without doing a read or passing in a parameter, which I think would defeat the point.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: I left saving in a separate method since `removeQueryData` needs it as well, but does not need the same metadata update logic. 

I renamed `updateMetadataForQueryData` to `updateMetadata` and added a comment that it updates the in-memory copy and does not save.

packages/firestore/test/unit/local/schema_migration.test.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Sure, `saveTargetCount` sounds fine.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
 
-  // NOTE: keys for these stores are specified explicitly rather than using a
-  // keyPath.
-  db.createObjectStore(DbDocumentMutation.store);
-  db.createObjectStore(DbRemoteDocument.store);
-  db.createObjectStore(DbOwner.store);
-  db.createObjectStore(DbTargetGlobal.store);
+  let p = PersistencePromise.resolve();
+  if (fromVersion < 2 && toVersion >= 2) {
+    p = p.next(() => ensureTargetGlobal(txn)).next(() => saveTargetCount(txn));
```
mikelehen: Minor but I'd rather combine these two methods and have saveTargetCount just gracefully handle if the row is missing. This 1) Avoids unnecessarily doing a get-put-get-put, and 2) Makes it clearer what is actually new in schema version 2 (the target count).  If you want, you can keep ensureTargetGlobal() but make it return the TargetGlobal row, and then call it from within saveTargetCount().

Also, I suggest renaming saveTargetCount() to populateTargetCount() to make it clearer that it needs to do work to figure out what the target count needs to be.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -469,3 +566,10 @@ export const ALL_STORES = [
   DbTargetGlobal.store,
   DbTargetDocument.store
 ];
+
+/**
+ * The list of all default IndexedDB stores used throughout the SDK. This is
+ * used when creating transactions so that access across all stores is done
+ * atomically.
+ */
+export const ALL_STORES = [...V1_STORES];
```
mikelehen: This is weird...  I wonder if we should get rid of the V1_STORES  list or else do:

```
// No new stores added in V2.
const V2_STORES = V1_STORES;

...
export const ALL_STORES = V2_STORES;
```
(maybe V2_STORES is overkill, but I think we should at least have the comment explaining why we're using V1_STORES even though SCHEMA_VERSION is 2)

packages/firestore/src/local/simple_db.ts
```diff
@@ -494,7 +510,7 @@ export class SimpleDbStore<KeyType extends IDBValidKey, ValueType> {
  * Wraps an IDBRequest in a PersistencePromise, using the onsuccess / onerror
  * handlers to resolve / reject the PersistencePromise as appropriate.
  */
-function wrapRequest<R>(request: IDBRequest): PersistencePromise<R> {
+export function wrapRequest<R>(request: IDBRequest): PersistencePromise<R> {
```
mikelehen: As mentioned elsewhere I'd like to keep this abstracted away from consumers of SimpleDb and not export it.

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -65,15 +67,29 @@ export class MemoryQueryCache implements QueryCache {
...
-    transaction: PersistenceTransaction,
-    queryData: QueryData
-  ): PersistencePromise<void> {
+  private saveQueryData(queryData: QueryData) {
     this.queries.set(queryData.query, queryData);
     const targetId = queryData.targetId;
     if (targetId > this.highestTargetId) {
       this.highestTargetId = targetId;
     }
+    // TODO(gsoltis): track sequence number
```
mikelehen: Consider doing TODO(GC) or similar instead so that we can grep for all of these once GC is "done" and verify nothing was missed. This is what we've done for other features in the past (and Sebastian is doing TODO(multitab) for multi-tab, etc).

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -65,15 +67,29 @@ export class MemoryQueryCache implements QueryCache {
...
+  }
+
+  addQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
```
mikelehen: Can we assert that it already exists?  That a) protects us from bugs, and b) makes this code more obviously correct [the reader doesn't have to wonder if the targetCount accounting needs to take into account prior existence].

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -65,15 +67,29 @@ export class MemoryQueryCache implements QueryCache {
...
+    this.saveQueryData(queryData);
+    this.targetCount += 1;
+    return PersistencePromise.resolve();
+  }
+
+  updateQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
```
mikelehen: Can we assert that it already exists?

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
-      return addedQueryPromise.next(() =>
-        globalTargetStore(transaction).put(DbTargetGlobal.key, this.metadata)
-      );
-    } else {
-      return addedQueryPromise;
+    return targetsStore(transaction).put(this.serializer.toDbTarget(queryData));
+  }
+
+  // Updates the in-memory version of the metadata. Saving is done separately.
+  // Returns true if there were any changes to the metadata.
```
mikelehen: nit: We usually use /** ... */ for method comments like this.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
     );
   }
 
-  addQueryData(
+  private saveQueryData(
```
mikelehen: nit: We usually put private helpers after the public methods that use them.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
-        globalTargetStore(transaction).put(DbTargetGlobal.key, this.metadata)
-      );
-    } else {
-      return addedQueryPromise;
+    return targetsStore(transaction).put(this.serializer.toDbTarget(queryData));
+  }
+
+  // Updates the in-memory version of the metadata. Saving is done separately.
+  // Returns true if there were any changes to the metadata.
+  private updateMetadata(queryData: QueryData): boolean {
```
mikelehen: I'm not sure this helper pays for itself since although it's used twice, we only use the return value once, so the other usage could just be "this.metadata.highestTargetId = Math.Max(...)"

If we keep it, can you rename to updateHighestTargetId() to better capture what it's doing and when it'll return true?

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
+    );
+  }
+
+  addQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
+    return this.saveQueryData(transaction, queryData).next(() => {
+      this.metadata.targetCount += 1;
+      this.updateMetadata(queryData);
```
mikelehen: As mentioned previously, it looks strange to call an updateMetadata() method after we've just manually updated metadata. It's not very clear what it actually does.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
+  // Updates the in-memory version of the metadata. Saving is done separately.
+  // Returns true if there were any changes to the metadata.
+  private updateMetadata(queryData: QueryData): boolean {
+    let needsUpdate = false;
+    if (queryData.targetId > this.metadata.highestTargetId) {
+      this.metadata.highestTargetId = queryData.targetId;
+      needsUpdate = true;
     }
+
+    // TODO(gsoltis): add sequence number check
```
mikelehen: Ahh. I guess maybe this is what motivates having this helper?  I guess I'd still advocate renaming (eventually to updateTargetIdAndSequenceNumberGlobals or whatever)

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
+    return globalTargetStore(transaction).put(
+      DbTargetGlobal.key,
+      this.metadata
+    );
+  }
+
+  addQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
```
mikelehen: If it were me I'd add logic to verify that the query data doesn't already exist (asserting otherwise) as I suggested for the in-memory implementation. This is a perf hit, so I'd add a TODO to revisit for performance.

That said, having just the memory implementation enforce the contract is _probably_ sufficient, so I'm okay with leaving it out here. I'd perhaps add a comment to make the expectation clear though, e.g.:

// Although we require that the target does not already exist, we don't enforce it here for performance (though MemoryQueryCache does).

[and similar for the updateQueryData() method]

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,75 @@ export class IndexedDbQueryCache implements QueryCache {
...
   }
 
+  removeQueryData(
+    transaction: PersistenceTransaction,
+    queryData: QueryData
+  ): PersistencePromise<void> {
+    return this.removeMatchingKeysForTargetId(transaction, queryData.targetId)
+      .next(() => targetsStore(transaction).delete(queryData.targetId))
+      .next(() => {
+        this.metadata.targetCount -= 1;
```
mikelehen: The comment for removeQueryData() says "no-op if no entry exists".  We need to either change the comment (and add appropriate assertions / comments about existence requirement), or change the implementation to check for existence first...

As a sanity check, we may want to add a targetCount >= 0 assertion here.

packages/firestore/src/local/query_cache.ts
```diff
@@ -88,6 +100,11 @@ export interface QueryCache extends GarbageSource {
     queryData: QueryData
   ): PersistencePromise<void>;
 
+  /**
+   * The number of targets currently in the cache.
+   */
+  count(): number;
```
mikelehen: FWIW, this could be a property since it's synchronous and cheap...

packages/firestore/src/local/simple_db.ts
```diff
@@ -70,7 +76,17 @@ export class SimpleDb {
         // cheating and just passing the raw IndexedDB in, since
         // createObjectStore(), etc. are synchronous.
         const db = (event.target as IDBOpenDBRequest).result;
-        runUpgrade(db, event.oldVersion);
+        runUpgrade(
+          db,
+          request.transaction,
```
mikelehen: As mentioned elsewhere, I think we should wrap this in a PersistenceTransaction.  

And please remove / update the above TODO as appropriate.

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,127 @@
...
+function getAllObjectStores(db: IDBDatabase): String[] {
+  const objectStores: String[] = [];
+  for (let i = 0; i < db.objectStoreNames.length; ++i) {
+    objectStores.push(db.objectStoreNames.item(i));
+  }
+  objectStores.sort();
+  return objectStores;
+}
+
+describe('IndexedDbSchema: createOrUpgradeDb', () => {
```
mikelehen: Perhaps this file should be named indexeddb_schema.test.ts now ?

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import { PersistencePromise } from '../../../src/local/persistence_promise';
+
+const INDEXEDDB_TEST_DATABASE = 'schemaTest';
+
+function withDb(
+  schemaVersion,
+  fn: (db: IDBDatabase) => Promise<void>
+): Promise<void> {
+  return new Promise<IDBDatabase>((resolve, reject) => {
+    const request = window.indexedDB.open(
```
mikelehen: I can't remember why Sebastian said we couldn't use SimpleDb.openOrCreate, but you might want to revisit, especially if we need it to create the PersistenceTransaction that gets passed to createOrUpgradeDb().

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (!IndexedDbPersistence.isAvailable()) {
+    console.warn('No IndexedDB. Skipping createOrUpgradeDb() tests.');
+    return;
+  }
+
+  beforeEach(() => SimpleDb.delete(INDEXEDDB_TEST_DATABASE));
+
+  it('can install schema version 1', () => {
+    return withDb(1, db => {
+      expect(db.version).to.be.equal(1);
```
mikelehen: I thought Sebastian fixed this in his version, but generally we use ".to.equal(...)" not ".to.be.equal(...)"  (you use "be" when you do things like "to.be.true")

packages/firestore/test/unit/local/schema_migration.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      expect(getAllObjectStores(db)).to.have.members(V1_STORES);
+      return Promise.resolve();
+    });
+  });
+
+  it('can install schema version 2', () => {
+    return withDb(2, db => {
+      expect(db.version).to.be.equal(2);
+      expect(getAllObjectStores(db)).to.have.members(ALL_STORES);
```
mikelehen: Should we validate that the TargetGlobal entry has 0 targetCount?

packages/firestore/test/unit/local/schema_migration.test.ts
schmidt-sebastian: It doesn't expose the version or the list of available object stores.

mikelehen: Ahh, thanks (sorry I was being lazy and not looking back at that PR).  I think it would be fine to add that functionality if it'd be useful to us.

packages/firestore/test/unit/local/schema_migration.test.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,91 @@ export class DbTargetGlobal {
...
+    DbTarget.queryTargetsIndexName,
+    DbTarget.queryTargetsKeyPath,
+    { unique: true }
+  );
+  db.createObjectStore(DbTargetGlobal.store);
+}
+
+/**
+ * Counts the number of targets persisted and adds that value to the target
+ * global singleton.
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +472,91 @@ export class DbTargetGlobal {
...
+ * @param {IDBTransaction} txn The version upgrade transaction for indexeddb
+ * @return {PersistencePromise<void>} A promise of the operations to add the
+ *                                    count to the metadata row.
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
-  );
-  targetDocumentsStore.createIndex(
-    DbTargetDocument.documentTargetsIndex,
-    DbTargetDocument.documentTargetsKeyPath,
-    { unique: true }
+/**
+ * Performs database creation and schema upgrades.
+ *
+ * Note that in production, this method is only ever used to upgrade the schema
+ * to SCHEMA_VERSION. Different versions are only used for testing and
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
-  // TODO(mikelehen): Get rid of "as any" if/when TypeScript fixes their
-  // types. https://github.com/Microsoft/TypeScript/issues/14322
-  db.createObjectStore(
-    DbMutationBatch.store,
-    // tslint:disable-next-line:no-any
-    { keyPath: DbMutationBatch.keyPath as any }
-  );
+/**
+ * Schema Version for the Web client (containing the Mutation Queue, the Query
+ * and the Remote Document Cache) and target count added to the target global.
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/test/unit/local/schema_migration.test.ts
gsoltis: Done.

gsoltis:
Ok, I think I've addressed everything except the `SimpleDb` and `PersistenceTransaction` stuff. I'll see what needs to be done to make that work.

packages/firestore/src/local/simple_db.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,55 @@ import { ResourcePath } from '../model/path';
...
+/**
+ * Performs database creation and schema upgrades.
+ *
+ * Note that in production, this method is only ever used to upgrade the schema
+ * to SCHEMA_VERSION. Different versions are only used for testing and
+ * local feature development.
+ */
+export function createOrUpgradeDb(
+  db: IDBDatabase,
+  txn: IDBTransaction,
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/test/unit/local/schema_migration.test.ts
gsoltis: I did not add this, as this test is the only place where the functionality would be required, and this is testing a very db-specific mechanism. If we end up with more tests needing this, we can revisit. Instead, I allowed creating a `SimpleDbTransaction` from an existing `IDBTransaction`, which gets us the API we want to use.

gsoltis:
Ok, `SimpleDb` usage is now included as well.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: Sure.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: Asserts added to memory version.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: yup.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Comment changed.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: It was originally `updateMetadataForQueryData`, but it was requested that I change it. I'd rather not spell out exactly what is getting updated in the method name, since it's conceivable that we might someday track other metadata as well. `updateMetadataFromQueryData`?

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: I'd rather not add the read, as in practice we actually do the read before calling this anyways. I am intending on encapsulating this behavior though in the rethink of the interaction between `LocalStore` and the persistence layer. My current prototype calls for `getOrCreateQueryData` and various `updateQuery` methods that take a target ID and do a `get` first. The raw `addQueryData` and `updateQueryData` would no longer be exposed to the `LocalStore`.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,76 @@ export class IndexedDbQueryCache implements QueryCache {
     );
   }
 
+  private saveMetadata(
```
schmidt-sebastian: I was gonna claim ownership of the term 'metadata' (but Mike already shut me down). Can you make this more query-specific (updateQueryMetadata) ?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,54 +21,57 @@ import { ResourcePath } from '../model/path';
...
-  db.createObjectStore(
-    DbMutationBatch.store,
-    // tslint:disable-next-line:no-any
-    { keyPath: DbMutationBatch.keyPath as any }
-  );
+/**
+ * Schema Version for the Web client:
+ * 1. Initial version including Mutation Queue, Query Cache, and Remote Document
+ *    Cache
+ * 2. Added targetCount to targetGlobal row.
```
schmidt-sebastian: Add note that this is needed for GC?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,15 +474,87 @@ export class DbTargetGlobal {
...
+  });
+}
+
+/**
+ * Ensures that the target global singleton row exists by adding it if it's
+ * missing.
+ *
+ * @param {IDBTransaction} txn The version upgrade transaction for indexeddb
+ */
+function ensureTargetGlobal(
```
schmidt-sebastian: `ensureTargetGlobalEntryExists` ? The current name raises more questions than it answers :)

packages/firestore/src/local/simple_db.ts
```diff
@@ -63,14 +70,21 @@ export class SimpleDb {
...
-        // TODO(mikelehen): If/when we need to do an actual data
-        // migration, we'll want to wrap db in a SimpleDb and have the
-        // runUpgrade function return a PersistencePromise, since we'll
-        // likely need to do async reads and writes. For now we're
-        // cheating and just passing the raw IndexedDB in, since
-        // createObjectStore(), etc. are synchronous.
+        // We are provided a version upgrade transaction from the request, so
+        // we wrap that in a SimpleDbTransaction to allow use of our friendlier
+        // API for schema migration operations.
         const db = (event.target as IDBOpenDBRequest).result;
```
schmidt-sebastian: Can you do the wrapping right here? That way, your comment is immediately related to the next line.

packages/firestore/src/local/simple_db.ts
```diff
@@ -342,6 +360,14 @@ export class SimpleDbStore<KeyType extends IDBValidKey, ValueType> {
     return wrapRequest<void>(request);
   }
 
+  // if we ever need more of the count variants, we can add overloads. For now,
```
schmidt-sebastian: Suggestion: /** Counts all objects in the store. Runs in O(n) */

In my opinion, The part about the overloads doesn't seem to warrant a comment.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -47,6 +47,17 @@ describe('MemoryQueryCache', () => {
   });
 
   genericQueryCacheTests();
+
+  it('cannot remove nonexistent query', async () => {
+    // We don't check on indexeddb query cache, but we do for memory.
```
schmidt-sebastian: I don't understand this comment.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -47,6 +47,17 @@ describe('MemoryQueryCache', () => {
...
+    let gotError = false;
+    try {
+      await cache.removeQueryData(testQueryData(QUERY_ROOMS, 1, 1));
+    } catch {
+      gotError = true;
+    }
+    expect(gotError).to.be.true;
```
schmidt-sebastian: So this should only be true for MemoryPersistence?

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -47,6 +47,17 @@ describe('MemoryQueryCache', () => {
   });
 
   genericQueryCacheTests();
+
+  it('cannot remove nonexistent query', async () => {
+    // We don't check on indexeddb query cache, but we do for memory.
+    let gotError = false;
+    try {
+      await cache.removeQueryData(testQueryData(QUERY_ROOMS, 1, 1));
```
schmidt-sebastian: You can just fail() here.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: I think `metadata` is fine, since it's a private method referring specifically to the instance-scoped field called `metadata`. I don't think it lays claim to an sdk-wide notion of `metadata`.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: I'm not sure this is actually a good idea. It's the sort of comment that could easily become outdated by accident, since whoever starts using it for something else would have no reason to check and update this location. I think it's better to stick to just describing what the schema updates are.

packages/firestore/src/local/simple_db.ts
gsoltis: I left the part about overloads because we are explicitly not supporting part of the underlying API, whereas other methods in this class provide overloads to handle all of the variants of the API calls that they are wrapping. As this is a deviation from what we are currently do, I thought it warranted a comment.

I added the `O(n)` note though.

packages/firestore/src/local/simple_db.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Updated.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: I updated the comment, but left the helper. It's being ported from here, which includes the sequence number bit: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Local/FSTLevelDBQueryCache.mm#L139

packages/firestore/src/local/query_cache.ts
gsoltis: Ah, that's neat. I didn't realize typescript could do that.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Done.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Yes, I updated the comment.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Updated.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Changed to `ensureTargetGlobalExists`. The comment should also clarify any questions raised.

packages/firestore/src/local/simple_db.ts
```diff
@@ -251,11 +249,21 @@ export class SimpleDbTransaction {
...
+      db.transaction(objectStoresNames, mode as AnyDuringMigration)
+    );
+  }
+
+  constructor(private transaction: IDBTransaction) {
+    /*this.transaction = db.transaction(
       objectStoresNames,
       mode as AnyDuringMigration
-    );
+    );*/
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/simple_db.ts
```diff
@@ -240,7 +238,7 @@ export interface IterateOptions {
  * specific object store.
  */
 export class SimpleDbTransaction {
-  private transaction: IDBTransaction;
+  //private transaction: IDBTransaction;
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/simple_db.ts
```diff
@@ -251,11 +249,21 @@ export class SimpleDbTransaction {
    */
   readonly completionPromise: Promise<void>;
 
-  constructor(db: IDBDatabase, mode: string, objectStoresNames: string[]) {
-    this.transaction = db.transaction(
+  static open(
+    db: IDBDatabase,
+    mode: string,
+    objectStoresNames: string[]
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: I went with `updateMetadataFromQueryData` to make it clear the updates are coming from the passed in `QueryData`. I'm open to other suggestions though.

packages/firestore/src/local/simple_db.ts
gsoltis: I reverted the export.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: I don't feel super strongly about this, but I do have a slight preference for keeping schema changes independent. Even though these two are in the same version, the actions aren't really strongly related, and the reason for the return value would be convenience.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -21,29 +21,31 @@ import { ResourcePath } from '../model/path';
...
+ * and local feature development.
  */
 export function createOrUpgradeDb(
   db: IDBDatabase,
-  txn: IDBTransaction,
+  txn: SimpleDbTransaction,
   fromVersion: number,
   toVersion: number
-): PersistencePromise<void> {
+): Promise<void> {
```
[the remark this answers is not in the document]
gsoltis: Done.

config/karma.base.js
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Renamed to `populateTargetCount`. I was initially going to keep the return value of `ensureTargetGlobal` as `void`, since they are independent steps (making sure the global exists could be its own schema migration), but since both you and @schmidt-sebastian asked for it and I don't feel too strongly about it, I'll make the change.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Ok, @mikelehen asked for this change too, so I guess I'm outvoted.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: I got rid of `V1_STORES`, since there are not [yet] any new stores. Whoever adds new stores can decide how to split them up. Probably not with `V2_STORES`, since that won't be v2.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Renamed the method.

gsoltis:
Ok, I think I have addressed everything now.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -97,36 +97,79 @@ export class IndexedDbQueryCache implements QueryCache {
...
```
mikelehen: Move this down with the other private helper methods? (probably next to saveQueryData)

packages/firestore/src/local/indexeddb_query_cache.ts
mikelehen: Thanks, I think this is much clearer!

packages/firestore/src/local/indexeddb_schema.ts
mikelehen: I don't see the rename...  I also still don't think there's a need to have two steps here.  TargetGlobal was an existing part of the schema in v1. The only thing you added was a new field to it.

Another way to deal with this would be to get rid of ensureTargetGlobal() and have populateTargetCount() just do nothing if the TargetGlobal row doesn't exist yet (since it should get populated correctly as 0 when the TargetGlobal row gets created as a normal part of the client operating).  And actually, I think I like this better since 1) it avoids us duplicating the default `new DbTargetGlobal( ... )` code.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,13 +474,87 @@ export class DbTargetGlobal {
...
+    { unique: true }
+  );
+  db.createObjectStore(DbTargetGlobal.store);
+}
+
+/**
+ * Counts the number of targets persisted and adds that value to the target
+ * global singleton.
+ *
+ * @param {IDBTransaction} txn The version upgrade transaction for indexeddb
```
mikelehen: jsdoc type annotation is unnecessary and no longer correct. :-)

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,13 +474,87 @@ export class DbTargetGlobal {
...
+ */
+function saveTargetCount(txn: SimpleDbTransaction): PersistencePromise<void> {
+  const globalStore = txn.store<DbTargetGlobalKey, DbTargetGlobal>(
+    DbTargetGlobal.store
+  );
+  const targetStore = txn.store<DbTargetKey, DbTarget>(DbTarget.store);
+  return targetStore.count().next(count => {
+    return globalStore.get(DbTargetGlobal.key).next(metadata => {
```
mikelehen: `metadata` is nullable here and so you should be getting an error for not handling it, but unfortunately we are not compiling with strictNullChecks (this was an accident when we migrated from google3 to github; b/73018483 tracks turning it on).  In an effort to avoid adding additional debt, can you handle null here?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -451,13 +474,87 @@ export class DbTargetGlobal {
...
+      );
+      return globalStore.put(DbTargetGlobal.key, metadata).next(() => metadata);
+    }
+  });
+}
+
 /**
- * The list of all IndexedDB stored used by the SDK. This is used when creating
- * transactions so that access across all stores is done atomically.
+ * The list of all default IndexedDB stores used throughout the SDK. This is
```
mikelehen: Drop "default"

packages/firestore/src/local/simple_db.ts
```diff
@@ -235,12 +244,17 @@ export class SimpleDbTransaction {
...
+    db: IDBDatabase,
+    mode: string,
+    objectStoreNames: string[]
+  ): SimpleDbTransaction {
+    return new SimpleDbTransaction(
+      db.transaction(objectStoreNames, mode as AnyDuringMigration)
     );
+  }
 
+  constructor(private transaction: IDBTransaction) {
```
mikelehen: `private readonly` if you want...

packages/firestore/src/local/simple_db.ts
```diff
@@ -342,6 +356,17 @@ export class SimpleDbStore<KeyType extends IDBValidKey, ValueType> {
     return wrapRequest<void>(request);
   }
 
+  // If we ever need more of the count variants, we can add overloads. For now,
+  // all we need is to count everything in a store.
+  //
+  // Returns the number of rows in the store. Execution time, off of the event
+  // loop, is O(# rows).
```
mikelehen: Use `/* ... */` style comment?  And I'm not sure what "off of the event loop" is referring to. I would probably drop it unless there's something special about this method compared to other methods on this class...

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -47,6 +48,17 @@ describe('MemoryQueryCache', () => {
   });
 
   genericQueryCacheTests();
+
+  it('cannot remove nonexistent query', async () => {
+    // We don't check for existence in the indexeddb query cache, but we do for
+    // memory. So, this test is specific to the memory-backed query cache.
+    try {
+      await cache.removeQueryData(testQueryData(QUERY_ROOMS, 1, 1));
+      fail('should not have been able to remove nonexistent query');
```
mikelehen: This `fail` method just throws an exception, which you'll then swallow...  and so I think this test can't fail.  You probably want `expect.fail()` instead, which hooks into the test runner to register the failure rather than relying on an exception.

You could also consider writing it without await, since .then() lets you attach both a success and rejection handler at once:

```
return cache.removeQueryData(...).then(
  () => fail(' ... '),
  (err) => { /* expected. */ });
```

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -68,35 +80,35 @@ describe('IndexedDbQueryCache', () => {
   genericQueryCacheTests();
 });
 
+const QUERY_ROOMS = Query.atPath(path('rooms'));
+const QUERY_HALLS = Query.atPath(path('halls'));
+const QUERY_GARAGES = Query.atPath(path('garages'));
```
mikelehen: Now that these are file-level globals can you move them to the top of the file (above `persistence` and `cache`)?

packages/firestore/test/unit/local/query_cache.test.ts
mikelehen: Alternatively I don't feel strongly that we need tests to cover assertions and since you didn't add tests for updateQueryData() or addQueryData(), I'd be somewhat inclined to just drop this test.  If you do that, please undo the QUERY_* / testQueryData() refactor.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -131,25 +143,32 @@ function genericQueryCacheTests() {
...
+    expect(cache.count()).to.equal(0);
   });
 
   it('can set query to new value', async () => {
     await cache.addQueryData(testQueryData(QUERY_ROOMS, 1, 1));
-    await setAndReadQuery(testQueryData(QUERY_ROOMS, 1, 2));
+    const updated = testQueryData(QUERY_ROOMS, 1, 2);
+    await cache.updateQueryData(updated);
+    const retrieved = await cache.getQueryData(updated.query);
+    expect(retrieved).to.deep.equal(updated);
```
mikelehen: With this change, setAndReadQuery() is only used once and therefore doesn't justify itself as a helper method. Can you inline the single usage?

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: I fixed it to actually thread the metadata through, so it should not be null anymore.

gsoltis:
Ok, I can't seem to reply to the comment about not seeing the `ensureTargetGlobalExists` rename, because github now thinks it's outdated. Anyways, I have renamed it, and it remains necessary because schema 0 does create the object store, but it does not create the row. We need the row to exist before updating the target count, although in that case we are guaranteed that the count is 0. I would rather handle ensuring the row exists in schema migrations. Then, the rest of the code can treat it as guaranteed-to-exist. I've added an assert that it exists in the query cache, since the query cache code gets called after the migration code. This also matches what we do for iOS: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Local/FSTLevelDBQueryCache.mm#L96

Otherwise, I think it's weird to do some db initialization in migrations, and some in constructors / start methods.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Done.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Undid this change.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Sounds good.

packages/firestore/src/local/simple_db.ts
gsoltis: The execution time is O(n), but the execution occurs on a different thread, so it does not block the event loop. It just means the amount of time until the you get the callback is proportional to the number of items that need to be counted. It was previously requested that I note that it was O(n). I don't feel too strongly either way...

For now I switched the comment style and reworded it slightly.

packages/firestore/src/local/simple_db.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Deleted.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -56,7 +55,8 @@ export class IndexedDbQueryCache implements QueryCache {
   private metadata = new DbTargetGlobal(
     /*highestTargetId=*/ 0,
     /*lastListenSequenceNumber=*/ 0,
-    SnapshotVersion.MIN.toTimestamp()
+    SnapshotVersion.MIN.toTimestamp(),
+    /*targetCount=*/ 0
   );
```
mikelehen: Since you've reworked this so that the DBTargetGlobal always exists, I think we don't need to initialize metadata here, and I'd prefer not have these defaults duplicated unnecessarily between here and indexeddb_schema.ts.  So perhaps change this to just:

private metadata: DbTargetGlobal = null;

packages/firestore/src/local/simple_db.ts
mikelehen: There are no threads in JavaScript [well, excluding web workers] and everything runs on the event loop.  So if you want to say something, you could say, "Execution time is O(# rows), although asynchronous." But since it returns a PersistencePromise, it's probably not worth calling out that it's asynchronous.

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: Done, and fixed the bug in the tests where we weren't calling `start()` but didn't know because we were initializing this to a dummy value.

So, good suggestion!

packages/firestore/src/local/simple_db.ts
gsoltis: Well, yes, there are not multiple threads running js code, but the VM has multiple threads, and it appears that at least in chrome, indexeddb gets its own thread.

Regardless, I dropped the comment about running time. If anyone wants it back, they can send a PR.

gsoltis:
I'm OOO Friday and Monday, so I'm going to wait until Tuesday to merge this, on the off chance that something breaks.

PR #514, opened 2018-02-15
Add testing module

tonymeng:
The goal of this PR is to add a new module `@firebase/testing` that is a relatively stand alone module that can be included to test js apps locally.


packages/testing/src/api/index.ts
```diff
@@ -0,0 +1,101 @@
...
+export function initializeAdminApp(options: any): admin.app.App {
+  if (!('databaseName' in options)) {
+    throw new Error('databaseName not specified');
+  }
+  return admin.initializeApp(
+    {
+      credential: new FakeCredentials(),
+      databaseURL: DBURL + '?ns=' + options.databaseName
+    },
+    'app-' + (new Date().getTime() + Math.random())
```
ryanpbrewster: I think this `Math.random()` is actually increasing the chance of a name collision. Can we just use the timestamp or an autoincrement id?

packages/testing/src/api/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    throw new Error('databaseName not specified');
+  }
+  // if options.auth is not present, we will construct an app with auth == null
+  return admin.initializeApp(
+    {
+      credential: new FakeCredentials(),
+      databaseURL: DBURL + '?ns=' + options.databaseName,
+      databaseAuthVariableOverride: options.auth || null
+    },
+    'app-' + (new Date().getTime() + Math.random())
```
ryanpbrewster: maybe call this `admin-app-{}`?

packages/testing/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@firebase/testing",
+  "version": "0.1.0",
```
jshcrowthe: Lets make sure we have the `private` value set to `true` for this module. I don't think we want to publish this quite yet?

packages/testing/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "description": "",
+  "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
+  "main": "dist/cjs/index.js",
+  "browser": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "scripts": {
+    "dev": "gulp dev",
+    "test": "run-p test:node",
```
jshcrowthe: If you don't have any browser tests you can omit the `npm-run-all` dependency, and replace this `test` script with the one in `test:node`

packages/testing/package.json
tonymeng: done

packages/testing/package.json
tonymeng: done

jshcrowthe:
Currently, there is no line in the `.github/CODEOWNERS` file for this package, as such you'll always need someone above that scope to approve PRs.

Want to add the team working on this as a line there? Something like:

```
packages/testing  @tonymeng @ryanpbrewster # Others if there are more
```

PR #523, opened 2018-02-20
Messaging - IDB Cleanup

gauntface:
A few bugs have been highlighted from 4.9x => 4.10.0:

- There was a bug in <= 4.9.X builds of the messaging SDK that meant the SDK was using an IDB name of 'undefined'. My hunch is something went wonky in closure -> typescript.
- If the name wasn't `undefined` the changes in 4.10 should have been fine, but because there is now a difference IDB name, the SDK is creating a new token, meaning end users can get two notifications - one for the old token and one for the new token.

This PR will bump the version of the IDB database in the next release, and clean up any old IDB instances if they exist in the version upgrade step.

### DIFF
--- a/packages/messaging/src/models/clean-v1-undefined.ts
+++ b/packages/messaging/src/models/clean-v1-undefined.ts
@@ -1,0 +1,87 @@
+/**
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * There seems to have been a bug in the messaging SDK versions <= 4.9.x
+ * where the IndexedDB model was using a database name of 'undefined'.
+ *
+ * In 4.10.x we changed the model implementation, but kept the database
+ * name as it should have been. This however introduced an issue where
+ * two tokens were pointing to the same underlying PushSubscription.
+ *
+ * This code will look for the undefined database and delete any of the
+ * underlying tokens.
+ */
+
+import IIDModel from '../models/iid-model';
+
+const OLD_DB_NAME = 'undefined';
+const OLD_OBJECT_STORE_NAME = 'fcm_token_object_Store';
+
+function handleDb(db: IDBDatabase) {
+  if (!db.objectStoreNames.contains(OLD_OBJECT_STORE_NAME)) {
+    // We found a database with the right name, but our expected object
+    // store isn't defined.
+    return;
+  }
+
+  const transaction = db.transaction([OLD_OBJECT_STORE_NAME]);
+  const objectStore = transaction.objectStore(OLD_OBJECT_STORE_NAME);
+
+  const iidModel = new IIDModel();
+
+  const openCursorRequest = objectStore.openCursor();
+  openCursorRequest.onerror = event => {
+    // NOOP - Nothing we can do.
+  };
+
+  openCursorRequest.onsuccess = event => {
+    var cursor = event.target.result;
+    if (cursor) {
+      // cursor.value contains the current record being iterated through
+      // this is where you'd do something with the result
+      const tokenDetails = cursor.value;
+
+      iidModel.deleteToken(
+        tokenDetails.fcmSenderId,
+        tokenDetails.fcmToken,
+        tokenDetails.fcmPushSet
+      );
+
+      cursor.continue();
+    } else {
+      // No more results - delete database if we are the only
+      // SDK using it
+      if (db.objectStoreNames.length === 1) {
+        indexedDB.deleteDatabase(OLD_DB_NAME);
+      }
+      db.close();
+    }
+  };
+}
+
+function cleanV1() {
+  const request: IDBOpenDBRequest = indexedDB.open(OLD_DB_NAME);
+  request.onerror = function(event) {
+    // NOOP - Nothing we can do.
+  };
+  request.onsuccess = function(event) {
+    const db = request.result;
+    handleDb(db);
+  };
+}
+
+export {cleanV1};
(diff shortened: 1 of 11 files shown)


gauntface:
@jshcrowthe I can't add @pinarx as a reviewer to this PR for some reason.

packages/messaging/src/models/token-details-model.ts
```diff
@@ -39,29 +41,31 @@ ValidateInput.prototype.fcmPushSet;
...
   constructor() {
-    super(TokenDetailsModel.DB_NAME, DB_VERSION);
+    super(DB_NAME, DB_VERSION);
   }
 
-  static get DB_NAME() {
-    return 'fcm_token_details_db';
-  }
+  onDBUpgrade(db: IDBDatabase, evt: IDBVersionChangeEvent) {
+    if (evt.oldVersion < 1) {
```
pinarx: this means it's brand new, right? can you add a quick comment?

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+import IIDModel from '../models/iid-model';
+
+const OLD_DB_NAME = 'undefined';
+const OLD_OBJECT_STORE_NAME = 'fcm_token_object_Store';
+
+function handleDb(db: IDBDatabase) {
+  if (!db.objectStoreNames.contains(OLD_OBJECT_STORE_NAME)) {
+    // We found a database with the right name, but our expected object
+    // store isn't defined.
+    return;
```
pinarx: this comment is confusing. So the v1 object store doesn't exist, and the new one is not defined yet? does this imply the old object store was dropped (maybe manually), and a new one isn't created yet?

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  const transaction = db.transaction([OLD_OBJECT_STORE_NAME]);
```
pinarx: can probably drop the brackets: 

the following lines are equivalent:
var transaction = db.transaction(['my-store-name']); 
var transaction = db.transaction('my-store-name');

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+ * two tokens were pointing to the same underlying PushSubscription.
+ *
+ * This code will look for the undefined database and delete any of the
+ * underlying tokens.
+ */
+
+import IIDModel from '../models/iid-model';
+
+const OLD_DB_NAME = 'undefined';
+const OLD_OBJECT_STORE_NAME = 'fcm_token_object_Store';
```
pinarx: OBJECT_STORE_NAME_V1 and DB_NAME_V1?

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  const transaction = db.transaction([OLD_OBJECT_STORE_NAME]);
+  const objectStore = transaction.objectStore(OLD_OBJECT_STORE_NAME);
+
+  const iidModel = new IIDModel();
+
+  const openCursorRequest: IDBRequest = objectStore.openCursor();
+  openCursorRequest.onerror = event => {
+    // NOOP - Nothing we can do.
```
pinarx: would it be helpful to log something to the console?

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (cursor) {
+      // cursor.value contains the current record being iterated through
+      // this is where you'd do something with the result
+      const tokenDetails = cursor.value;
+
+      iidModel.deleteToken(
+        tokenDetails.fcmSenderId,
+        tokenDetails.fcmToken,
+        tokenDetails.fcmPushSet
+      );
```
pinarx: hmm this deletes the push subscription as well, which is a bit of an overkill, but we don't have a good way of only deleting the FCM token from the client I guess...

But assume I've already switched to 4.10.x and have 2 DBs with 2 tokens pointing to the same push subscription. When this happens, we will end up with a FCM token with a PushSubscription that is not registered. Would we ever get into that situation? 

I am guessing not since they would already have created the second DB so there wouldn't be an update event. But how do they get out of that situation in that case?

packages/messaging/src/models/clean-v1-undefined.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      cursor.continue();
+    } else {
+      // No more results - delete database if we are the only
+      // SDK using it
+      if (db.objectStoreNames.length === 1) {
```
pinarx: can we just check if the name matches in that case? rather than checking the length?

packages/messaging/src/models/clean-v1-undefined.ts
gauntface: This is looking for the database named 'undefined', which *could* be used be created anyone, which if it is - we perhaps may not be the owner and therefore the expected object store wouldnt exist.

packages/messaging/src/models/clean-v1-undefined.ts
gauntface: This code is purely cleaning up the old model (which shouldn't have been named 'undefined' but it was). So this, while related to the token-details-model, they are separate databases and I am just using the v1 -> v2 change to perform this one off clean up. So the version shouldn't every be included or needed here.

packages/messaging/src/models/clean-v1-undefined.ts
gauntface: I don't think this will unsubscribe the push subscription - can you point me in the direction of where that is the case?

packages/messaging/src/models/clean-v1-undefined.ts
gauntface: Clear the database.

packages/messaging/src/models/clean-v1-undefined.ts
pinarx: you're right, i've mistaken it for controller.deleteToken. Disregard my comment.

packages/messaging/src/models/clean-v1-undefined.ts
gauntface: I added a warning log.

packages/messaging/src/models/token-details-model.ts
gauntface: Added comment

CTassisF:
Is it intentional that after this PR firebase.js is creating a new, empty IDB named "undefined" even when it doesn't exist, like when a new visitor comes or firebase.js 4.10.1 is deployed for the first time in a website?

PR #524, opened 2018-02-20
Version Patch for Firebase Admin

jshcrowthe:
Refactor the `initStandalone` function to optionally patch the firebase version.
Also sets the `NODE_ADMIN` constant to `true`

### DIFF
--- a/packages/database/index.node.ts
+++ b/packages/database/index.node.ts
@@ -15,6 +15,7 @@
  */
 
 import firebase from '@firebase/app';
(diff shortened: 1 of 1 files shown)


jshcrowthe:
@hiranya911 you should be able to pull down this change and run `yarn link:packages` this will set up a link to all the packages.

You can then go to your firebase-admin dev env and run `yarn link @firebase/database` and that should let you test locally.

PR #527, opened 2018-02-21
Updating API surface

gauntface:
@pinarx pointed out that the types file was out of sync with the API surface.

@jshcrowthe we can do away with the types files if we get the "typescript-y-ness" of the source inline and up-to-date right, or is there something I'm missing? I.e. move away from closure and opt for typescript typing


jshcrowthe:
This PR as it is looks fine. Really the larger issue here is that the `@firebase/messaging` and `@firebase/messaging-types` packages are entirely separate. In an ideal world you would have all of your implementation code, linked to your interfaces via `implements` statements. Going forward there will also be cases were components will need to be able to take dependencies on the interfaces provided by other components, w/o actually importing them directly.

gauntface:
Yeah we need to figure something out here to ensure it doesn't happen again.

PR #534, opened 2018-02-27
Database API: Ref from Ref

schmidt-sebastian:
This implements go/firebase-ref-from-ref:

// Before:
// The developer must know of the little-used refFromUrl and must also know that a 
// ref's toString is the full URL of the ref
let ref2 = admin.database().refFromURL(ref1.toString());

// After:
// more intuitive
let ref2 = admin.database().ref(ref1);


gsoltis:
Late to the party, but LGTM to me too.

PR #539, opened 2018-03-01
Mark @firebase/testing as non-private

ryanpbrewster:


tonymeng:
i'm a bit hesitant to do this as we have no official documentation and no product that supports this feature. I assume we'll get a lot of questions about it if we push it to npm. If the goal is to allow specific folks to use this, can we distribute this some other way? maybe @jshcrowthe knows

ryanpbrewster:
Will re-open when there is a better story around this.

PR #542, opened 2018-03-02
feat: Add offset to query

atomita:
Hi there,

Firestore 's StructuredQuery has `offset`, but this package has no method, so I added it.

[StructuredQuery  |  Cloud Firestore  |  Google Cloud Platform](https://cloud.google.com/firestore/docs/reference/rest/v1beta1/StructuredQuery)

Many thanks,
atomita

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -1319,6 +1319,19 @@ export class Query implements firestore.Query {
     return new Query(this._query.withEndAt(bound), this.firestore);
   }
 
+  offset(n: number): firestore.Query {
+    validateExactNumberOfArgs('Query.offset', arguments, 1);
+    validateArgType('Query.offset', 'number', 1, n);
+    if (n < 0) {
+      throw new FirestoreError(
+        Code.INVALID_ARGUMENT,
+        `Invalid Query. Query offset (${n}) is invalid. Offset must be ` +
+          'positive.'
+      );
+    }
+    return new Query(this._query.withOffset(n), this.firestore);
+  }
+
   isEqual(other: firestore.Query): boolean {
     if (!(other instanceof Query)) {
       throw invalidClassError('isEqual', 'Query', 1, other);
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -41,7 +41,8 @@ export class Query {
     readonly filters: Filter[] = [],
     readonly limit: number | null = null,
     readonly startAt: Bound | null = null,
-    readonly endAt: Bound | null = null
+    readonly endAt: Bound | null = null,
+    readonly offset: number | null = null
   ) {
     if (this.startAt) {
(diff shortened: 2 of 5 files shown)


wilhuff:
I'm so sorry to say this after you've put this together, but we intentionally did not include `offset` in the mobile and web SDKs. We can't accept this PR as is. The reason is that `offset` breaks latency compensation: this implementation will show flickers of incorrect results and just won't work offline.

Latency compensation is a general technique where we try to hide the fact that we need to talk to a server to get query results. We use it to keep applications feeling responsive despite the network. We run queries locally and serve any results we can in the first snapshot of a listen. This includes values that have been written on the client but not yet committed to the server. Subsequent snapshots merge in results from the server. This works _even when you have persistence disabled_ because we can use the results of a query that's already running (or pending writes) to show refinements or overlapping results.

Let's ignore pending writes and just consider concurrently running queries. For example, let's say you had a collection of books and started a query for the whole collection:
```js
db.collection("books");
```
If you just left that running and then refined the query to show unread books like so:
```js
db.collection("books").where("read_state", "==", "unread");
```
... the client can show you all the results immediately before even talking to the server. This works even when the client has an incomplete view:
```js
db.collection("books").limit(10);
db.collection("books").where("read_state", "==", "unread").limit(10);
```
If half your books are unread, the client will
  * Immediately deliver an initial latency compensated snapshot containing 5 unread books that locally match the first query
  * Later merge the results from the server and deliver a second snapshot containing all 10 unread books

If you enter a subway tunnel before the server returns its results your app continues to feel responsive despite the bad or nonexistent connection.

So the problem with `offset` is that there's no way for the client to evaluate that query locally. What if this were the standing query:
```js
db.collection("books").offset(10);
```
And this were the refinement:
```
db.collection("books").where("read_state", "==", "unread").limit(10).offset(10);
```
Now we're stuck: do any of the books we're currently listening to match? It's not clear because the client never saw the first 10 books, so it doesn't how many of them are unread. That means that it can't know how many of the books it does have actually match the query.

Things get even worse when persistence is enabled. If the cache currently has results from these queries already populated:
```js
db.collection("books").where("publication_date", "==", 2018);
db.collection("books").where("authors", "==", ["Kernighan, Brian W.", "Ritchie, Dennis M."]);
```
Without `offset` we can locally merge these results into any other query on books if they happen to match. If there's an `offset` it's not clear how many of the local results need to be discarded.

The implementation you've made here does not add the local evaluation component of the query. This means that when starting a listen it will initially deliver the wrong results and then supply the correct result once the server corrects it. For example, if you have this query running:
```js
db.collection("books");
```
And then you run:
```js
db.collection("books").offset(10);
```
This implementation will initially deliver a snapshot containing all the books and then deliver a second snapshot removing the first ten. This kind of flicker can be very noticeable on mobile networks.

atomita:
I did not notice that you did not dare implement it.
Thank you.

PR #544, opened 2018-03-03
Make Timestamp a part of the public Firestore API.

var-const:


packages/firestore-types/index.d.ts
```diff
@@ -172,6 +196,66 @@ export class GeoPoint {
...
+  static now(): Timestamp;
+
+  /** Creates a new timestamp from the given date. */
+  static fromDate(date: Date): Timestamp;
+
+  /**
+   * Creates a new timestamp from the given number of milliseconds since Unix
+   * epoch 1970-01-01T00:00:00Z.
+   */
+  static fromMillis(milliseconds: number): Timestamp;
```
var-const: I changed the name to match the API review doc, let me know if you think we should try to push back on this.

packages/firestore/src/api/database.ts
```diff
@@ -296,6 +321,32 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
       'FirestoreSettings.host cannot be falsey'
     );
 
+    if (!this._config.settings.timestampsInSnapshotsEnabled) {
+      log.error(`
+The default behavior for Date objects stored in Firestore is going to change
```
var-const: 1. After seeing this message in the logs, wrapping it to 80 columns seemed like a good idea.
2. This message gets logged about a thousand times during tests. It's super-annoying, do you have any suggestions on how to disable it during a test run?

packages/firestore/src/api/database.ts
```diff
@@ -387,7 +438,9 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
       if (this._firestoreClient) {
         return this._firestoreClient.shutdown();
       }
-    }
+    },
```
var-const: I'm not sure if this should be in INTERNAL. My reasoning was that it seems that in TypeScript, current settings aren't exposed (or I just missed it).

packages/firestore/src/api/timestamp.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  static fromMillis(milliseconds: number): Timestamp {
+    const seconds = Math.floor(milliseconds / 1000);
+    const nanos = (milliseconds - seconds * 1000) * 1e6;
+    return new Timestamp(seconds, nanos);
+  }
+
+  constructor(readonly seconds: number, readonly nanoseconds: number) {
+    if (nanoseconds < 0) {
+      throw new FirestoreError(
```
var-const: Use `throw` instead of `assert`.

packages/firestore/src/model/field_value.ts
```diff
@@ -69,21 +69,32 @@ export enum ServerTimestampBehavior {
 
 /** Holds properties that define field value deserialization options. */
 export class FieldValueOptions {
-  static readonly defaultOptions = new FieldValueOptions(
```
var-const: Removed `defaultOptions` like elsewhere.

packages/firestore/src/remote/serializer.ts
```diff
@@ -83,6 +83,9 @@ const OPERATORS = (() => {
   return ops;
 })();
 
+// A RegExp matching ISO 8601 UTC timestamps with optional fraction.
+const ISO_REG_EXP = new RegExp(/^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d(?:\.(\d+))?Z$/);
```
var-const: I originally moved everything connected to parsing ISO strings here to avoid exposing this functionality in the public API. I guess it was possible to leave it on the Timestamp and not declare it in `index.d.ts` file (at least I assume so), but on the other hand, that functionality is only used in the serializer. Feel free to push back on this.

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -86,6 +86,11 @@ export function toDataArray(
   return docSet.docs.map(d => d.data());
 }
 
+/** Converts a DocumentSet to an array with the id of each document */
+export function toIds(docSet: firestore.QuerySnapshot): String[] {
```
var-const: There is a similar helper function on iOS and Android (I hope I didn't miss it here).

packages/firestore/test/unit/api/timestamp.test.ts
```diff
@@ -24,38 +24,12 @@ describe('Timestamp', () => {
...
+      nanoseconds: 916000000
     });
 
     expect(Timestamp.fromDate(new Date(-1250))).to.deep.equal({
       seconds: -2,
-      nanos: 750000000
-    });
-  });
-
-  it('fromISOString', () => {
```
var-const: Moved to serializer tests.

packages/firestore/src/api/database.ts
mikelehen: INTERNAL is semi-public. So as-is users of the SDK could access this via firebase.firestore().INTERNAL.areTimestampsInSnapshotsEnabled() which we don't really want.  So I think instead I'd make this a property on Firestore:

get _areTimestampsInSnapshotsEnabled(): boolean {
  return this._config.settings.timestampsInSnapshotsEnabled;
}

Note that we use '_' for any members of our public API classes (like Firestore) that we don't want to be visible in our public API. There's a minification step as part of our build that will rename '_blah' properties to obfuscated names.

packages/firestore/src/api/timestamp.ts
mikelehen: I'm not sure if this comment is asking a question, but I think throw is appropriate since this is a user-facing error rather than an internal assertion.

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -544,7 +554,8 @@ export class UserDataConverter {
               'of your update data'
           );
         } else {
-          // We shouldn't encounter delete sentinels for queries or non-merge set() calls.
+          // We shouldn't encounter delete sentinels for queries or non-merge
+          // set() calls.
```
mikelehen: FYI- You seem to have re-wrapped lines in a few places.  We used to strictly adhere to 80 characters, but I think that enforcement got lost when we went to github and so we've gotten lax...  We probably should have an enforced line length, but until then I'd _slightly_ prefer avoiding reformatting otherwise un-touched lines in PRs. :-)  This PR only has a few and it's not a big deal (so definitely don't bother reverting them or anything).

It seems like we should be able to make prettier enforce a line length, but I haven't looked into it.

packages/firestore/src/remote/serializer.ts
mikelehen: Yeah, it would be fine to have it in Timestamp but not expose it in the .d.ts. Though we would also want to prefix the APIs with '_' to ensure they're minified.  That said, it seems like ISO timestamp conversion is specific to the web serializer implementation (we don't have iso timestamp conversion in the iOS / Android Timestamp.java implementation) and so keeping it in the serializer sounds good to me.

packages/firestore/test/integration/util/helpers.ts
mikelehen: It's possible it was omitted because map() is so easy in JS, but I'm not opposed to having it.

packages/firestore/test/integration/api/cursor.test.ts
```diff
@@ -233,4 +237,85 @@ apiDescribe('Cursors', persistence => {
         });
     });
   });
+
+  const makeTimestamp = (seconds, micros) =>
+    new Timestamp(seconds, micros * 1000);
+
+  it('can accept Timestamps as limits', () => {
```
mikelehen: Not sure if this was copied from elsewhere, but since we have a limit() API, this would be clearer as "can accept Timestamps as bounds" or similar.

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -214,12 +214,7 @@ describe('Serializer', () => {
     it('converts GeoPointValue', () => {
       const example = new GeoPoint(1.23, 4.56);
 
-      const expected = {
-        geoPointValue: {
-          latitude: 1.23,
-          longitude: 4.56
-        }
-      };
+      const expected = { geoPointValue: { latitude: 1.23, longitude: 4.56 } };
```
mikelehen: I'm curious if you reformatted these or prettier did it for some reason...

packages/firestore/src/api/timestamp.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  toDate(): Date {
+    return new Date(this.toMillis());
+  }
+
+  toMillis(): number {
+    return this.seconds * 1000 + this.nanoseconds / 1e6;
+  }
+
+  compareTo(other: Timestamp): number {
```
mikelehen: Assuming we don't make this public, can you rename to _compareTo()?

packages/firestore-types/index.d.ts
```diff
@@ -172,6 +196,66 @@ export class GeoPoint {
...
+  /**
+   * Returns the number of milliseconds since Unix epoch 1970-01-01T00:00:00Z.
+   */
+  toMillis(): number;
+
+  compareTo(other: Timestamp): number;
+
+  isEqual(other: Timestamp): boolean;
+
+  toString(): string;
```
mikelehen: toString() exists on any Object in JavaScript and doesn't need to be explicitly listed here.

packages/firestore/src/api/database.ts
var-const: Done, thanks for the explanation.

packages/firestore/src/api/database.ts
```diff
@@ -157,12 +165,29 @@ class FirestoreSettings {
...
+        'settings',
+        'boolean',
+        'timestampsInSnapshotsEnabled',
+        settings.timestampsInSnapshotsEnabled
+      );
+      this.timestampsInSnapshotsEnabled = objUtils.defaulted(
+        settings.timestampsInSnapshotsEnabled,
+        DEFAULT_TIMESTAMPS_IN_SNAPSHOTS_ENABLED
+      );
+      this.timestampsInSnapshotsEnabled = settings.timestampsInSnapshotsEnabled;
```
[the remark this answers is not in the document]
var-const: Done, thanks for spotting it.

packages/firestore-types/index.d.ts
```diff
@@ -20,21 +20,45 @@ import { FirebaseApp, FirebaseNamespace } from '@firebase/app-types';
...
   host?: string;
   /** Whether to use SSL when connecting. */
   ssl?: boolean;
+
+  /**
+   * Enables the use of {@link Timestamp}s for timestamp fields in {@link
+   * DocumentSnapshot}s.
+   *
+   * Currently, Firestore returns timestamp fields as {@link Date} but {@link
+   * Date} only supports millisecond precision, which leads to truncation and
```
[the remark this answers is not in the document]
mikelehen: Yeah, that seems fine and consistent with what we've been doing.  Using `@link` might be a good idea, but we can tackle that holistically at a later time.

packages/firestore-types/index.d.ts
```diff
@@ -20,21 +20,45 @@ import { FirebaseApp, FirebaseNamespace } from '@firebase/app-types';
  * Document data (for use with `DocumentReference.set()`) consists of fields
  * mapped to values.
  */
-export type DocumentData = { [field: string]: any };
+export type DocumentData = {
+  [field: string]: any;
```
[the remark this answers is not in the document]
mikelehen: Oh, I see. It looks like if I remove the semicolon, prettier adds it back, but if I completely undo this change (remove the semicolon _and_ put this back to being one line) then prettier leaves it as-is...  Maybe you ended up formatting with a "wrong" version and then the "right" version, which had a different net effect than just formatting with the "right" version or something?  I don't know...  Anyway, I'm not inclined to worry about it for this PR...

packages/firestore/src/api/database.ts
mikelehen: 1. Hrm.  I suspect wrapping to 80 columns would be friendly on the other platforms, though I don't remember if Xcode and Android Studio wrap automatically in their consoles or not.  So when you revisit the log messages pending API review feedback, etc., maybe wrap them, but I am not _too_ concerned about it.
2. I think the 11 tests should be changed to expect the new behavior, but I'd make sure we have at least one test that verifies server timestamps return Dates (for ServerTimestampBehavior 'estimate') if you have the enableTimestampsInSnapshots setting turned off.

packages/firestore/src/api/timestamp.ts
mikelehen: isEqual() is considered part of our public API and so we don't want it minified and so it should stay isEqual(), not _isEqual().

packages/firestore/src/api/database.ts
```diff
@@ -157,12 +165,29 @@ class FirestoreSettings {
       settings.credentials
     );
     this.credentials = settings.credentials;
+
+    if (settings.timestampsInSnapshotsEnabled === undefined) {
+      this.timestampsInSnapshotsEnabled = DEFAULT_TIMESTAMPS_IN_SNAPSHOTS_ENABLED;
```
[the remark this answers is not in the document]
var-const: Done.

packages/firestore/src/api/database.ts
var-const: Changed the tests, including one test for server timestamps that previously checked for Timestamps if the setting is true, but now checks for Dates if setting is false.

packages/firestore-types/index.d.ts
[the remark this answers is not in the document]
var-const: Hmm, this sounds like it could be the cause. I first formatted using `clang-format`, which resulted in a pretty different formatting, and I guess `prettier` couldn't restore it successfully.

mikelehen:
Hrm. My guess is you accidentally created the package-lock.json file while the prepush hook was running and it looks like (https://github.com/firebase/firebase-js-sdk/blob/69c9a971fa4e48bf965369744559e4c9a2fab12b/tools/gitHooks/prettier.js#L50) it just does a "git add ." at the end and so it inadvertently included it...  cc/ @jshcrowthe We should probably do "git add -u" or skip "git add" and just do a "git commit -a"

I'm not 100% sure what you would have done to create package-lock.json though...  perhaps you did an "npm" command instead of "yarn" or something?  FYI- I think I see an unintentional files.txt in your PR too...

As for the test configuration change, yes please!  Other than things that are actually platform-specific we want to keep the platforms as in-sync as possible.  It's a bit painful but it makes for healthy and consistent codebases. So the effort is appreciated. :-)

var-const:
@mikelehen Michael, could you please take another look? The only change since the last time is slight tweaking of warning due to feedback on the doc. The idea was to phrase it clearer that the current behavior is not going to be supported in the future. I'll be happy to change if you'd like.

packages/firestore-types/index.d.ts
```diff
@@ -35,6 +35,26 @@ export interface Settings {
   host?: string;
   /** Whether to use SSL when connecting. */
   ssl?: boolean;
+
```
mikelehen: So...  it turns out our types exist in 3 .d.ts files. 😦 @jshcrowthe has plans to fix this, but for now can you update packages/firebase/index.d.ts and packages/firebase/app/index.d.ts too?

packages/firestore/src/api/database.ts
```diff
@@ -296,6 +316,32 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
     );
 
+    if (!this._config.settings.timestampsInSnapshots) {
+      log.error(`
+The behavior for Date objects stored in Firestore is going to change
+AND YOUR APP MAY BREAK.
+To hide this warning and ensure your app does not break, you need to add the
+following code to your app before calling any other Cloud Firestore methods:
+
+  const firestore = firebase.firestore(firebase.app());
```
mikelehen: I think `firebase.app()` is not necessary?  Just doing `firebase.firestore()` should use the default app...

mikelehen: So it's a bit convoluted.  firebase.firestore() is actually a function that FirebaseApp creates when we (Firestore) call firebase.INTERNAL.registerService() to register ourselves.  So:

1. We call firebase.INTERNAL.registerService() here: https://github.com/firebase/firebase-js-sdk/blob/b5aa5b8335cd13ce8c30c188843723f3a766ac53/packages/firestore/src/platform/config.ts#L58
2. It creates this `serviceNamespace` which is the firebase.firestore() function [see line 369] except it also has properties like firebase.firestore.FieldValue hence it's called a "namespace": https://github.com/firebase/firebase-js-sdk/blob/b5aa5b8335cd13ce8c30c188843723f3a766ac53/packages/app/src/firebaseApp.ts#L353
3. That serviceNamespace function defaults to using the default app (`appArg: FirebaseApp = app()`).

So app is optional...

var-const: Thanks! I tried finding it in the code, realized it's something dynamically created and gave up.

PR #559, opened 2018-03-12
Fix MutationQueue issue resulting in re-sending acknowledged writes.

mikelehen:
getNextMutationBatchAfterBatchId() was not respecting highestAcknowledgedBatchId and therefore we were resending writes if they had been acknowledged but not removed (aka the held write acks case). This showed up when a user enabled / disabled the network as reported here and I've included a spec test to emulate that case: https://github.com/firebase/firebase-ios-sdk/issues/772  (cc/ @gsoltis FYI).

This is the JS port which I'll port to iOS (where the issue was reported) once we're happy with the fix here.

### DIFF
--- a/packages/firestore/test/unit/specs/write_spec.test.ts
+++ b/packages/firestore/test/unit/specs/write_spec.test.ts
@@ -395,6 +395,56 @@ describeSpec('Writes:', [], () => {
   });
 
   specTest(
+    'Held writes are not re-sent after disable/enable network.',
+    [],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const docALocal = doc(
+        'collection/a',
+        0,
+        { v: 1 },
+        { hasLocalMutations: true }
+      );
+      const docA = doc('collection/a', 1000, { v: 1 });
+
+      return (
+        spec()
+          .userListens(query)
+          .watchAcksFull(query, 500)
+          .expectEvents(query, {})
+          .userSets('collection/a', { v: 1 })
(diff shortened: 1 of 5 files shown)


packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -394,6 +394,56 @@ describeSpec('Writes:', [], () => {
...
+          // ack write but without a watch event.
+          .writeAcks(1000)
+
+          .disableNetwork()
+          .expectEvents(query, {
+            hasPendingWrites: true,
+            fromCache: true
+          })
+
+          // handshake + write + close = 3 requests
```
wilhuff: This comment is a bit too terse. Is this happening as a result of disabling the network or something else?

If so it's weird because I would have thought when we disable the network nothing happens.

This could be a little clearer if you expected the write stream request count was 2 above and then this would show that all we did after that was close.

mikelehen: Good point.  Added the requestCount==2 expectation prior to close.  FWIW, disabling the network sends an additional no-op write (https://github.com/firebase/firebase-js-sdk/blob/69c9a971fa4e48bf965369744559e4c9a2fab12b/packages/firestore/src/remote/persistent_stream.ts#L662). I don't know the full context of that, but I assume it's intentional (so the backend knows we got all the acks and can GC state probably).

wilhuff:
Changes LGTM too. Thanks for the changelog entry.

PR #564, opened 2018-03-14
Fix for b/74749605: Cancel pending backoff operations when closing streams.

mikelehen:


packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -316,6 +316,9 @@ export abstract class PersistentStream<
 
     this.cancelIdleCheck();
 
+    // Ensure we don't leave a pending backoff operation queued.
```
wilhuff: This comment can move above the idle cancel and cover both:

```
// Ensure we don't leave pending timer-based operations queued.
```

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -316,6 +316,9 @@ export abstract class PersistentStream<
...
+    this.backoff.cancel();
```
wilhuff: It's a little weird that cancelling the idle and backoff timers look different. Is there any way to make these uniform? I suppose this happens because the backoff timer is held within the backoff object but the idle timer is held directly?

I think this may be non-actionable, but mentioning it anyway.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -461,10 +464,10 @@ export abstract class PersistentStream<
     this.state = PersistentStreamState.Backoff;
 
     this.backoff.backoffAndRun(async () => {
-      if (this.state === PersistentStreamState.Stopped) {
-        // Stream can be stopped while waiting for backoff to complete.
-        return;
-      }
+      assert(
```
wilhuff: Man am I gun-shy about this assertion. I'm somewhat in favor of just leaving the old behavior in place to keep this consistent with the other callbacks.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: Hrm.  I had trouble creating a unified comment that I didn't feel was overly general or misleading, so instead I just commented both of them separately.

FWIW- the idle timer will only be active if the stream _is_ currently connected when close() is called, and the backoff timer will only be active if the stream _is not_ currently connected when close() is called.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: acknowledged.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: Yeah...  I am torn between being as strict as possible with our state transitions vs being defensive. I mostly changed it to an assert, since as-is it's basically dead code.

I put it back though and just tweaked the comment.

PR #579, opened 2018-03-21
Move getPushSubscription method to ControllerInterface. 

pinarx:
Both SWController and WindowController need to implement getPushSubscription method to use in the getToken flow. For SwController, the parameters of the method is a bit redundant: we pass a swRegistration in the method, even though SWController has internal access to it.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -154,10 +154,10 @@ declare namespace firebase.auth {
     applyActionCode(code: string): Promise<any>;
     checkActionCode(code: string): Promise<any>;
     confirmPasswordReset(code: string, newPassword: string): Promise<any>;
-    createUserAndRetrieveDataWithEmailAndPassword( 
-      email: string, 
-      password: string 
-    ): Promise<any>;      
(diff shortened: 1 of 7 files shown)


packages/firebase/index.d.ts
```diff
@@ -154,10 +154,10 @@ declare namespace firebase.auth {
     applyActionCode(code: string): Promise<any>;
     checkActionCode(code: string): Promise<any>;
     confirmPasswordReset(code: string, newPassword: string): Promise<any>;
-    createUserAndRetrieveDataWithEmailAndPassword( 
-      email: string, 
-      password: string 
-    ): Promise<any>;      
+    createUserAndRetrieveDataWithEmailAndPassword(
```
gauntface: @jshcrowthe this prettier auto stuff really drags on PRs :(

jshcrowthe: Agreed,  @schmidt-sebastian and I have both proposed solutions to handle this (#585, #580). Both scope the change down to only your package so other components changes shouldn't be included as much.

Issue #531, opened 2018-02-23
Prettier pre-push Improvement Suggestions


mikelehen:
I would love to see a couple improvements to the prettier / pre-push hook:

1. Verify that `prettier -v` matches the pinned version in package.json (and log a message telling you to run `yarn` if it does not).  This should avoid PRs that accidentally undo prettier formatting changes because the author hasn't updated to the correct version.
2. Run prettier only on the "changed" files for performance.  This is a little tricky, but something like `git diff --name-only upstream/master || git diff --name-only origin/master` might be a good enough approximation.  Just to make sure nothing gets missed, we could have travis do a prettier run over the entire codebase and verify there are no diffs.

mikelehen:
After @hsubox76's improvements, I think we're in good shape!

PR #580, opened 2018-03-21
Adding Prettier on Commit

schmidt-sebastian:
I should have discussed this per email first, but writing this in form a PR was a lot quicker. 

This PR addresses two of my issues that I see with the JS SDK development:

- It is hard to run formatting on the code on the fly. I have to commit my code, then run the githooks manually from the repo root. I now have a commit in my commit history that I didn't necessarily want.

- Formatting all source files on `git push` takes too long. We should speed this up by only formatting the files that changed.

The common way to do this seems to be via `lint-staged`, which is actually so fast that we can run in on commit. Commit times for me with this change are < 1.5 seconds for small sets of changes.


schmidt-sebastian:
BTW, this adds 86 (!) new indirect dependencies. The size of node_modules increases from 615 MB to 626 MB (we still fit on a CD-R...).

package.json
```diff
@@ -17,6 +17,7 @@
   ],
   "scripts": {
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase dev",
+    "precommit": "lint-staged",
```
mikelehen: Oh boy...  I don't know if I trust prettier enough to let it silently modify my commits.  Then I can't tell what it did and I might not notice problems.  Today I always run "yarn prepush" manually, check what it did, and then if I'm satisfied, I'll clean up the git history via "git rebase -i master" or whatever.

package.json
```diff
@@ -62,5 +64,15 @@
     "semver": "^5.5.0",
     "simple-git": "^1.91.0",
     "yargs": "^11.0.0"
+  },
+  "lint-staged": {
+  	"*.ts": [
+      "prettier --write",
```
mikelehen: prettier.js references repo-root/.prettierrc.  Are we losing that?

tools/gitHooks/prettier.js
```diff
@@ -1,60 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-const { resolve } = require('path');
-const { spawn } = require('child-process-promise');
-const fs = require('mz/fs');
-const glob = require('glob');
-const simpleGit = require('simple-git/promise');
-const ora = require('ora');
```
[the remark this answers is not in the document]
jshcrowthe: We won't need it necessarily no. It can always be run manually by just calling `./node_modules/.bin/prettier` and I'd imagine we do so fairly rarely.

jshcrowthe:
So it's interesting that this was proposed as it was actually the original way this was done.

We had originally changed to the `pre-push` hook as `pre-commit` didn't handle a couple use cases very well:

- Code migrations (we were migrating Firestore at the time and didn't want to mix styling changes w/ actual structural changes)
- Patch commits, because lint-staged runs on files (and not diffs) it messes up files with partial commits

Now, with those things on the table, I'm largely in favor of this change (like I said it was how it was done originally). It is fairly common amongst folks using prettier and I'd love to see us follow that convention.

package.json
schmidt-sebastian: Using the staged files makes it really easy to figure out what files to re-format. A while ago, I tried something similar to what you suggested in https://github.com/firebase/firebase-js-sdk/issues/531 (diff against origin & upstream) but wasn't able to get it to work properly in between my other work. 

I personally don't like the Prettier commits but I also don't want to take them away from you (since I can just rebase if need be). Maybe there is some way we can still stage them in a separate commit (by not passing `--write`, or not doing `git add`?). Do you have suggestions for this, @jshcrowthe?

I do want to ask though if it is even possible to disagree with prettier. If you manually re-format prettier's changes, doesn't it just revert these changes on the next run?

jshcrowthe: I don't know what this would entail, but we could do a `post-commit` hook and just try and format the last diff? I dunno what it would require/look like, but it seems doable.

mikelehen: I sometimes choose to wrap lines differently (or restructure code to avoid wrapping, etc.).  And I just like to know what prettier is doing. We've run into cases where it just does wonky reformatting (either due to bugs or poor design) and I like to notice those when they happen.

tools/gitHooks/prettier.js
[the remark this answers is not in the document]
jshcrowthe: Agreed.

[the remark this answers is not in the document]
mikelehen: Yeah...  I more meant "Do we still need a script that runs prettier on the whole repo?" since the commit hooks aren't going to catch everything (rebases, amend commits, post-prettier-upgrade reformats, etc.)?

Maybe it's simple enough to just run it from the command-line manually though. I'm not sure what all this script was doing...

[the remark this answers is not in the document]
jshcrowthe: This script just runs prettier on the whole SDK and gives you progress updates. So we could probably just make an NPM script in the repo root to take care of the global style hammer.

schmidt-sebastian:
Closing this in favor of https://github.com/firebase/firebase-js-sdk/pull/585. Thanks Josh.

PR #584, opened 2018-03-21
Persisting mutation batch state in LocalStorage

schmidt-sebastian:
This PR allows pending mutation batches to be persisted in LocalStorage. In order to keep things focussed, this only adds the serialization and deserialization in SharedClientState. The next PR will add the calls from LocalStorage that invoke SharedClientState, and in a further follow up I will add the wiring between SyncEngine and SharedClientState.

This PR also paves the way for SyncEngine to gain two pieces of functionality:
- Apply pending batches by Batch ID to all Local Views 
- Apply successful mutations by Batch ID

Note: The changes in `fromLocalStorageEntry` are formatting only  as I wrapped the logic in a `if (value !== null)` block.

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -36,6 +36,9 @@ const CLIENT_STATE_KEY_PREFIX = 'fs_clients';
 // The format of the LocalStorage key that stores the mutation state is:
 //     fs_mutations_<persistence_prefix>_<batch_id> (for unauthenticated users)
 // or: fs_mutations_<persistence_prefix>_<batch_id>_<user_uid>
+//
+// 'user_uid' is last to avoid needing to escape '_' characters that it might
+// contain.
 const MUTATION_BATCH_KEY_PREFIX = 'fs_mutations';
 
 /**
@@ -53,11 +56,11 @@ export type ClientKey = string;
  * `SharedClientState` is primarily used for synchronization in Multi-Tab
  * environments. Each tab is responsible for registering its active query
  * targets and mutations. `SharedClientState` will then notify the listener
- * passed to `subscribe()` for updates to mutations and queries that originated
- * in other clients.
+ * assigned to `.syncEngine` for updates to mutations and queries that
+ * originated in other clients.
  *
- * To receive notifications, both `subscribe()` and `start()` have to be called
- * in order.
+ * To receive notifications, `.syncEngine` has to be assigned before calling
+ * `start()`.
  *
  * TODO(multitab): Add callbacks to SyncEngine
  */
@@ -115,7 +118,7 @@ interface MutationMetadataSchema {
 }
 
 /**
- * Holds the state of a mutation batch, including its user ID, batch ID annd
+ * Holds the state of a mutation batch, including its user ID, batch ID and
  * whether the batch is 'pending', 'acknowledged' or 'rejected'.
  */
 // Visible for testing
@@ -181,15 +184,18 @@ export class MutationMetadata {
   }
 
   toLocalStorageJSON(): string {
-    const batchState: MutationMetadataSchema = {
+    const batchMetadata: MutationMetadataSchema = {
       state: this.state
     };
 
     if (this.error) {
-      batchState.error = { code: this.error.code, message: this.error.message };
+      batchMetadata.error = {
+        code: this.error.code,
+        message: this.error.message
+      };
     }
 
-    return JSON.stringify(batchState);
+    return JSON.stringify(batchMetadata);
   }
 }
 
@@ -366,13 +372,14 @@ export class LocalClientState implements ClientState {
  */
 // TODO(multitab): Rename all usages of LocalStorage to WebStorage to better differentiate from LocalClient.
 export class WebStorageSharedClientState implements SharedClientState {
+  syncEngine: SharedClientStateSyncer | null;
+
   private readonly storage: Storage;
   private readonly localClientStorageKey: string;
   private readonly activeClients: { [key: string]: ClientState } = {};
   private readonly storageListener = this.handleLocalStorageEvent.bind(this);
   private readonly clientStateKeyRe: RegExp;
   private readonly mutationBatchKeyRe: RegExp;
-  private syncEngine: SharedClientStateSyncer | null;
   private user: User;
   private started = false;
 
@@ -395,7 +402,7 @@ export class WebStorageSharedClientState implements SharedClientState {
       `^${CLIENT_STATE_KEY_PREFIX}_${persistenceKey}_([^_]*)$`
     );
     this.mutationBatchKeyRe = new RegExp(
-      `^${MUTATION_BATCH_KEY_PREFIX}_${persistenceKey}_(\\d*)(?:_(.*))$`
+      `^${MUTATION_BATCH_KEY_PREFIX}_${persistenceKey}_(\\d+)(?:_(.*))?$`
     );
   }
 
@@ -404,17 +411,9 @@ export class WebStorageSharedClientState implements SharedClientState {
     return typeof window !== 'undefined' && window.localStorage != null;
   }
 
-  subscribe(syncEngine: SharedClientStateSyncer) {
-    this.syncEngine = syncEngine;
-  }
-
   // TODO(multitab): Handle user changes.
   start(initialUser: User, knownClients: ClientKey[]): void {
     assert(!this.started, 'WebStorageSharedClientState already started');
-    assert(
-      this.syncEngine !== null,
-      'Start() called before subscribing to events'
-    );
     window.addEventListener('storage', this.storageListener);
 
     for (const clientKey of knownClients) {
@@ -497,13 +496,8 @@ export class WebStorageSharedClientState implements SharedClientState {
         'Received LocalStorage notification for local change.'
       );
 
-      if (event.newValue == null) {
-        if (this.clientStateKeyRe.test(event.key)) {
-          const clientId = this.fromLocalStorageClientStateKey(event.key);
-          delete this.activeClients[clientId];
-        }
-      } else {
-        if (this.clientStateKeyRe.test(event.key)) {
+      if (this.clientStateKeyRe.test(event.key)) {
+        if (event.newValue != null) {
           const clientState = this.fromLocalStorageClientState(
             event.key,
             event.newValue
@@ -511,7 +505,12 @@ export class WebStorageSharedClientState implements SharedClientState {
           if (clientState) {
             this.activeClients[clientState.clientId] = clientState;
           }
-        } else if (this.mutationBatchKeyRe.test(event.key)) {
+        } else {
+          const clientId = this.fromLocalStorageClientStateKey(event.key);
+          delete this.activeClients[clientId];
+        }
+      } else if (this.mutationBatchKeyRe.test(event.key)) {
+        if (event.newValue !== null) {
           const mutationMetadata = this.fromLocalStorageMutationMetadata(
             event.key,
             event.newValue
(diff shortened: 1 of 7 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -376,18 +495,28 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+          this.activeClients[clientState.clientId] = clientState;
+          return;
+        }
+        const mutationBatch = this.fromLocalStorageMutation(
+          event.key,
+          event.newValue
+        );
+        if (mutationBatch) {
+          this.handleMutationBatchEvent(mutationBatch);
+          return;
```
mikelehen: I'm not a big fan of this code structure, calling fromBlah() indiscriminately and using nulls to decide whether the thing you passed to fromBlah() was actually a Blah, and using early-returns...  Bunch of minor code smells that I don't enjoy. :-)

I'd be much happier doing a bit of extra work (e.g. running the regex twice) but making the logic more explicit...  something like:

```
if (this.clientStateKeyRe.test(event.key)) {
  ...
} else if (this.mutationBatchKeyRe.test(event.key)) {
  ...
}
```

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -409,8 +538,27 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+    const mutationState = new RemoteMutationBatch(
+      this.user,
+      batchId,
+      state,
+      error
+    );
+
+    const mutationKey = `${MUTATION_BATCH_KEY_PREFIX}_${this.persistenceKey}_${
+      this.user.uid
+    }_${batchId}`;
```
mikelehen: FWIW- This is the kind of prettier formatting where I'd be tempted to undo prettier's change and instead do:

```
    const mutationKey = `${MUTATION_BATCH_KEY_PREFIX}_${this.persistenceKey}_` +
        `${this.user.uid}_${batchId}`;
```

Not asking you to change it... just calling it out for context on the prettier stuff. :-)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -409,8 +538,27 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+
+    this.storage.setItem(mutationKey, mutationState.toLocalStorageJSON());
+  }
+
   /** Assembles the key for a client state in LocalStorage */
-  private toLocalStorageClientKey(clientKey: string): string {
+  private toLocalStorageClientStateKey(clientKey: string): string {
```
mikelehen: Should the arg be `clientId: ClientKey` ?  This code is going to be so much less confusing when we change ClientKey to ClientId. :-)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -423,8 +571,81 @@ export class WebStorageSharedClientState implements SharedClientState {
...
    */
-  private fromLocalStorageClientKey(key: string): string | null {
+  private fromLocalStorageClientStateKey(key: string): ClientKey | null {
     const match = this.clientStateKeyRe.exec(key);
     return match ? match[1] : null;
   }
+
+  /**
+   * Parses a client state in LocalStorage. Returns null if the key does not
+   * match the expected key format.
```
mikelehen: Per my comment above, I'm not a fan of these methods. If we keep them I'd suggest we put `try` in the method name to make it clearer that they're expected to be called with "incorrect" values and will just return null.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -423,8 +571,81 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+    key: string,
+    value: string | null
+  ): RemoteMutationBatch | null {
+    const match = this.mutationBatchKeyRe.exec(key);
+
+    if (match) {
+      const isCurrentUser =
+        match[1] === 'null'
+          ? this.user.uid === null
+          : match[1] === this.user.uid;
```
mikelehen: Hrm...  'null' could technically be a valid uid I think.  FWIW we just use an empty string for unauthenticated users in the mutation queue: https://github.com/firebase/firebase-js-sdk/blob/84896ec489ef162555c5b48881822982281d8921/packages/firestore/src/local/indexeddb_mutation_queue.ts#L83

Though having local storage keys with `__` to indicate unauthenticated is perhaps a bit awkward.

Oh, but on that note, I think `_` is allowed in uids and so our regex is going to be broken. :-/

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -70,14 +73,23 @@ export async function testMemoryPersistence(): Promise<MemoryPersistence> {
...
 }
 
+class NoOpSharedClientDelegate implements SharedClientDelegate {
+  async loadPendingBatch(batchId: BatchId): Promise<void> {}
+  async applySuccessfulWrite(batchId: BatchId): Promise<void> {}
+  async rejectFailedWrite(
+    batchId: BatchId,
+    err: FirestoreError
+  ): Promise<void> {}
+}
```
mikelehen: Hrm.  Will prettier let you put a newline after this?

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -206,22 +275,23 @@ describe('WebStorageSharedClientState', () => {
...
         storageArea: window.localStorage,
         newValue: oldState.toLocalStorageJSON()
       });
-      verifyState(0, [3, 4, 5]);
+      verifyState(1, [3, 4, 5]);
+
+      const updatedState = new LocalClientState();
+      updatedState.addQueryTarget(5);
+      updatedState.addQueryTarget(6);
+      updatedState.addPendingMutation(0);
```
mikelehen: I don't fully understand why you added this `addPendingMutation(0)` call to this test...  It seems like the old code was broken (it was returning 0 for getMinimumGlobalPendingMutation(), but it probably should have been returning -1 (BATCHID_UNKNOWN)... or maybe null).

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -261,4 +331,89 @@ describe('WebStorageSharedClientState', () => {
...
+    let clientDelegate;
+
+    beforeEach(() => {
+      clientDelegate = new TestClientDelegate();
+
+      return persistenceHelpers
+        .testWebStorageSharedClientState(ownerId, clientDelegate)
+        .then(nc => {
+          sharedClientState = nc;
+          expect(storageCallback).to.not.be.undefined;
```
mikelehen: Is `.to.be.defined` not a thing?

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -261,4 +331,89 @@ describe('WebStorageSharedClientState', () => {
...
+
+  describe('processes mutation updates', () => {
+    let clientDelegate;
+
+    beforeEach(() => {
+      clientDelegate = new TestClientDelegate();
+
+      return persistenceHelpers
+        .testWebStorageSharedClientState(ownerId, clientDelegate)
+        .then(nc => {
```
mikelehen: what does `nc` mean?

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -261,4 +331,89 @@ describe('WebStorageSharedClientState', () => {
...
+      storageCallback({
+        key: mutationKey(1),
+        storageArea: window.localStorage,
+        newValue: new RemoteMutationBatch(
+          TEST_USER,
+          1,
+          'rejected',
+          new FirestoreError('internal', 'Test Error')
+        ).toLocalStorageJSON()
+      });
```
mikelehen: consider a helper method to reduce test boilerplate...

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -399,7 +395,7 @@ export class WebStorageSharedClientState implements SharedClientState {
       `^${CLIENT_STATE_KEY_PREFIX}_${persistenceKey}_([^_]*)$`
     );
     this.mutationBatchKeyRe = new RegExp(
-      `^${MUTATION_BATCH_KEY_PREFIX}_${persistenceKey}_([^_]*)_(\\d*)$`
+      `^${MUTATION_BATCH_KEY_PREFIX}_${persistenceKey}_(\\d*)(?:_(.*))$`
```
mikelehen: I'm trying to figure out two things:

1. Why `\\d*` instead of `\\d+`?
2. Why `(?:_(.*))$` instead of just `_(.*)$`?

Am I missing something or would it be more precise / simple to do:

    `^${MUTATION_BATCH_KEY_PREFIX}_${persistenceKey}_(\\d+)_(.*)$`

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -408,37 +404,42 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+    this.syncEngine = syncEngine;
   }
 
+  // TODO(multitab): Handle user changes.
   start(initialUser: User, knownClients: ClientKey[]): void {
     assert(!this.started, 'WebStorageSharedClientState already started');
     assert(
-      this.sharedClientDelegate !== null,
+      this.syncEngine !== null,
       'Start() called before subscribing to events'
```
mikelehen: "subscribing to events" is no longer accurate (especially with the getActiveClients change)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -497,26 +498,27 @@ export class WebStorageSharedClientState implements SharedClientState {
       );
 
       if (event.newValue == null) {
```
mikelehen: minor, but I think this code would read better if you branch based on the key before inspecting the value.  This probably means you'll need to add a newValue != null check in your mutationBatchKeyRe branch, but I think that will actually be clearer / more explicit.

packages/firestore/src/local/shared_client_state.ts
mikelehen: Oh...  It looks like you only add _blah for authenticated users.  But then for this regex to work you'd need a `?` after the non-capturing group.  i.e. `...(?:_(.*))?$` right?  

Should there be a test for unauthenticated mutation batches?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -577,50 +583,40 @@ export class WebStorageSharedClientState implements SharedClientState {
   }
 
   /**
-   * Parses a client state in LocalStorage. Returns null if the key does not
-   * match the expected key format.
+   * Parses a client state in LocalStorage. Returns 'null' if the value could
+   * not be parsed.
```
mikelehen: It looks like this never returns null. Update comment and return type?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -577,50 +583,40 @@ export class WebStorageSharedClientState implements SharedClientState {
...
-      : null;
+    assert(clientId !== null, `Cannot parse client state key '${key}'`);
+    return RemoteClientState.fromLocalStorageEntry(clientId, value);
   }
 
   /**
-   * Parses a mutation batch state in LocalStorage. Returns null if the key does
-   * not match the expected key format.
+   * Parses a mutation batch state in LocalStorage. Returns 'null' if the value
+   * could not be parsed.
```
mikelehen: Ditto. Never returns null now.

packages/firestore/src/local/shared_client_state_syncer.ts
```diff
@@ -18,24 +18,22 @@ import { BatchId } from '../core/types';
 import { FirestoreError } from '../util/error';
 
 /**
- * Callback methods invoked by SharedClientState for notifications received
- * from Local Storage.
+ * A interface that describes the actions the SharedClientState class needs to
```
mikelehen: "An interface"

PR #585, opened 2018-03-22
Refactor Prettier Implementation (Post Commit Style)

jshcrowthe:
This is a follow up to #580 per a conversation with @schmidt-sebastian. This changes the behavior of the prettier implementation to do the following:

- Functions only on changed files (`.ts`/`.js`/`.json`)
- Functions post-commit and generates individual styling/licensing commits

Limitations w/ this approach:

- No support for doing patch commits (we could hard disable them)
- Code migrations will be stylistically nuked in this approach. (This doesn't seem to be a huge issue ATM)


schmidt-sebastian:
Can we work around the limitations if we also run `prettier` in CI and enforce that it does not produce any diff?

tools/postCommit.js
```diff
@@ -0,0 +1,146 @@
...
+const root = resolve(__dirname, '../');
+const Listr = require('listr');
+const fs = require('mz/fs');
+const { existsSync } = require('fs');
+
+/**
+ * Checks if the files in question have associated diffs
+ */
+async function hasDiff(files) {
+  let { stdout } = await exec(`git diff ${files.join(' ')}`, { cwd: root });
```
schmidt-sebastian: You can pass `--quiet` and then check the exit code: https://git-scm.com/docs/git-diff.

You can then likely skip passing in the filenames, which will prevent me from breaking this in the future by maliciously using filenames that contain single quotes.

jshcrowthe:
I'ma close this for now. Will revisit in the future.

PR #586, opened 2018-03-22
 Adding getActiveClients()

schmidt-sebastian:
This will be used by SharedClientState to retrieve the list of active clients during its startup. Using this list, the SharedClientState can then read all relevant rows in LocalStorage.

### DIFF
--- a/packages/firestore/test/unit/specs/spec_test_runner.ts
+++ b/packages/firestore/test/unit/specs/spec_test_runner.ts
@@ -320,6 +320,8 @@ interface OutstandingWrite {
 abstract class TestRunner {
   protected queue: AsyncQueue;
 
+  private started = false;
+
   private connection: MockConnection;
   private eventManager: EventManager;
   private syncEngine: SyncEngine;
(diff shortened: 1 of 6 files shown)


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -419,15 +421,19 @@ abstract class TestRunner {
     serializer: JsonProtoSerializer
   ): Persistence;
 
-  async start(): Promise<void> {
-    this.connection.reset();
-    await this.persistence.start();
-    await this.localStore.start();
-    await this.remoteStore.start();
+  async startIfNeeded(): Promise<void> {
```
mikelehen: Why does this need to be idempotent now?  I'm not exactly following what changed.

schmidt-sebastian: The way this was implemented before called `start()` for all clients before execution any of the steps in the tests. That meant by the time client(0).expectNumActiveClients(1) was called, we had already initialized both clients, causing this test to fail.

mikelehen: Oh...  I see.  We're creating clients at the start, but not starting them until later...  Would it be possible / make sense to just not create them until needed?

I'm a bit worried about making SpecTestRunner deviate too much from other platforms.  If nothing else, it seems like we should have some "PORTING NOTE" comments if we're going to change how start() works here...

schmidt-sebastian: I added a PORTING NOTE comment and moved this logic to an inner function (ensureRunner).

PR #587, opened 2018-03-22
Unify pushSubscription/publicVapidKey retrievals

pinarx:
cc @gauntface

Should be a no-op in terms of functionality:
1) Use async/await in the ControllerInterface.
2) Merge the retrieval of pushSubscription and publicVapidKey, so we don't call getPushSubscription and getPublicVapidKey multiple times in the flow.

### DIFF
--- a/packages/messaging/src/controllers/controller-interface.ts
+++ b/packages/messaging/src/controllers/controller-interface.ts
@@ -64,7 +64,7 @@ export default class ControllerInterface {
   /**
    * @export
    */
-  getToken(): Promise<string | null> {
+  async getToken(): Promise<string | null> {
     // Check with permissions
     const currentPermission = this.getNotificationPermission_();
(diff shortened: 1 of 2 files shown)


packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -100,26 +108,31 @@ export default class ControllerInterface {
...
+          swReg,
+          pushSubscription,
+          publicVapidKey,
+          tokenDetails
+        );
       }
-    });
+    }
+
+    // If the VAPID details are updated, delete the existing token,
```
gauntface: May reword this to as the logic suggests the delete and create steps are undertaken if the token is not longer, like when VAPID Details are changed.

pinarx: done

PR #588, opened 2018-03-23
Fixing Storage Types

schmidt-sebastian:
Removing 'any' from some of the Storage APIs. I copied this from the Storage externs definition in Google3.


sphippen:
The Storage type changes look good to me. Is there any particular reason for the formatting differences in `firebase/app/index.d.ts`?

schmidt-sebastian:
> The Storage type changes look good to me. Is there any particular reason for the formatting differences in firebase/app/index.d.ts?

These show up for me on 'master' too. I think we should go ahead and commit them (they remove spaces at the line end for example). Otherwise, the next PR will pick them up too.

sphippen:
Okay, sounds fine then. Thanks!

PR #603, opened 2018-03-28
Small cleanup to FieldValue sentinel code to make it easier to add more sentinels in the future.

mikelehen:
This technically includes a small breaking change. Before we would allow you to do:

`doc.set({a: [0, FieldValue.delete(), 2] }, {merge: true})`

Which ended up equivalent to:

`doc.set({a: [0, null, 2] }, {merge: true})`

But with my change it will now result in an error:

`FieldValue.delete() is not currently supported inside arrays.`

Let me know if you think that's an issue. I can rework it but I don't think the existing behavior was particularly useful or intentional.

### DIFF
--- a/packages/firestore/src/api/user_data_converter.ts
+++ b/packages/firestore/src/api/user_data_converter.ts
@@ -584,7 +584,7 @@ export class UserDataConverter {
         new FieldTransform(context.path, ServerTimestampTransform.instance)
       );
     } else {
-      return fail('Unknown FieldValue type: ' + value);
+      fail('Unknown FieldValue type: ' + value);
     }
   }
 }
--- a/packages/firestore/src/api/user_data_converter.ts
+++ b/packages/firestore/src/api/user_data_converter.ts
@@ -441,28 +441,31 @@ export class UserDataConverter {
    */
   private parseData(input: AnyJs, context: ParseContext): FieldValue | null {
     input = this.runPreConverter(input, context);
-    if (input instanceof Array) {
-      // TODO(b/34871131): Include the path containing the array in the error
-      // message.
-      if (context.arrayElement) {
-        throw context.createError('Nested arrays are not supported');
-      }
-      // If context.path is null we are already inside an array and we don't
-      // support field mask paths more granular than the top-level array.
-      if (context.path) {
-        context.fieldMask.push(context.path);
-      }
-      return this.parseArray(input as AnyJs[], context);
-    } else if (looksLikeJsonObject(input)) {
+    if (looksLikeJsonObject(input)) {
       validatePlainObject('Unsupported field value:', context, input);
       return this.parseObject(input as Dict<AnyJs>, context);
     } else {
-      // If context.path is null, we are inside an array and we should have
-      // already added the root of the array to the field mask.
+      // If context.path is null we are inside an array and we don't support
+      // field mask paths more granular than the top-level array.
       if (context.path) {
         context.fieldMask.push(context.path);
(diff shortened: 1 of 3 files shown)


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -530,51 +530,60 @@ export class UserDataConverter {
     } else if (value instanceof DocumentKeyReference) {
       return new RefValue(value.databaseId, value.key);
     } else if (value instanceof FieldValueImpl) {
```
schmidt-sebastian: Just a suggestion - we could move this block up to `parseData`. Then we would have a single place in the code that decides between our parseX implementations.

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -530,51 +530,60 @@ export class UserDataConverter {
...
+  }
+
+  private parseSentinelFieldValue(
+    value: FieldValueImpl,
+    context: ParseContext
+  ): FieldValue | null {
+    // Sentinels are only supported with writes, and not within arrays.
+    if (!isWrite(context.dataSource)) {
+      throw context.createError(
+        `${value.methodName} can only be used with update() and set().`
```
schmidt-sebastian: Nit: Remove period to match other errors.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -437,7 +437,7 @@ apiDescribe('Validation:', persistence => {
         return expectSetToFail(
           db,
           { foo: firebase.firestore.FieldValue.delete() },
-          'FieldValue.delete() can only be used with update() and set() with {merge:true} (found in field foo)'
+          'FieldValue.delete() cannot be used with set() unless you pass {merge:true}'
```
schmidt-sebastian: Why did the field name disappear?

packages/firestore/src/api/user_data_converter.ts
mikelehen: Good idea, done, though that led to a bit more refactoring to uniformly handle context.path being null (sorry).

packages/firestore/src/api/user_data_converter.ts
mikelehen: Done.

packages/firestore/test/integration/api/validation.test.ts
mikelehen: Uh, turns out it didn't...  But I accidentally removed it from the test and chai's .to.throw() expectation just verifies that the error _contains_ the specified string, so it still passed.

We should probably look into making the expectations more strict, but I think we'd need to switch to using a regex and we'd need to do escaping and stuff, so it's a bit of a pain.  _contains_ is probably good enough for now.  

Fixed the test in any case.

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -529,54 +532,61 @@ export class UserDataConverter {
...
+          'FieldValue.delete() cannot be used with set() unless you pass ' +
+            '{merge:true}'
+        );
+      }
+    } else if (value instanceof ServerTimestampFieldValueImpl) {
+      context.fieldTransforms.push(
+        new FieldTransform(context.path, ServerTimestampTransform.instance)
+      );
+    } else {
+      return fail('Unknown FieldValue type: ' + value);
```
schmidt-sebastian: You should be able to remove the `return` since the function now returns `void`.

Issue #556, opened 2018-03-08
Incorrect Typings for database.Reference.path


hiranya911:
The `path` field of the `database.Reference` type is defined to be a string in our type defs:

```
export interface Reference extends Query {
  child(path: string): Reference;
  key: string | null;
  onDisconnect(): OnDisconnect;
  parent: Reference | null;
  path: string;
  ...
```

But in reality this takes on an object value:

```
Path { pieces_: [ 'foo' ], pieceNum_: 0 }
```

mikelehen:
Where are you seeing that?  It should definitely be a string in our public API (but we do have an internal Path class used for our internal implementation)

schmidt-sebastian:
The `path` property is internal and should not have been exposed in the public types. Since this is technically breaking, I will remove it in the next major release.

PR #613, opened 2018-04-02
Removing .path from Database Typings

schmidt-sebastian:
Fixes: https://github.com/firebase/firebase-js-sdk/issues/556

.path is not an exported member of the Database API (the property is named .key): https://firebase.google.com/docs/reference/js/firebase.database.Reference


mikelehen:
I think it's okay.  Any change to a public type in JavaScript is technically a breaking change since there's no (great) way to separate private / public methods. This particular property was never documented or used in any public samples (since it wasn't meant to be exposed). Our TypeScript typings files are provided as a curtesy but are not the most authoritative  source of our public APIs. So I'd consider this a bugfix to our typings within the realm of acceptable changes.

schmidt-sebastian:
As per our recent policy change, rolling this back so that we can merge it into the v5-master branch instead.

PR #614, opened 2018-04-02
 Fix a bug that was introduced in PR: #591

pinarx:
Basically when the permissions are revoked and restored, do not unsubscribe from the PushSubscription, as we're using a brand new one.

Just remove the token from DB and from FCM server DB.

### DIFF
--- a/packages/messaging/src/controllers/controller-interface.ts
+++ b/packages/messaging/src/controllers/controller-interface.ts
@@ -80,6 +80,8 @@ export default class ControllerInterface {
 
     const swReg = await this.getSWRegistration_();
     const publicVapidKey = await this.getPublicVapidKey_();
+    // If there didn't exist one, this call will subscribe a new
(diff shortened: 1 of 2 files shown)


pinarx:
Firestore tests are failing - FYI.

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -80,6 +80,8 @@ export default class ControllerInterface {
 
     const swReg = await this.getSWRegistration_();
     const publicVapidKey = await this.getPublicVapidKey_();
+    // If there didn't exist one, this call will subscribe a new
+    // push subscription.
```
gauntface: Re-word to:

If a PushSubscription exists it's returned, otherwise a new subscription is generated and returned.

pinarx: done

PR #615, opened 2018-04-02
Adding .skip and .only for validation tests

schmidt-sebastian:
This add mocha-like .skip and .only functions to validationIt. I am not sure if this can be done without the 'any' keyword, as validationIt is now both a function and an object.

As always, this also contains unrelated prettier changes :)


schmidt-sebastian:
I updated this PR to remove the usage of `any`. As per https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es6.d.ts#L4439, `Object.assign` can be used to build up a type at initialization time.

PR #616, opened 2018-04-03
Round-tripping pending mutations

schmidt-sebastian:
This PR combines the remaining changes to share pending mutations between primary and secondary clients.

- It implements SyncEngine.applyPendingBatch.
- It changes the network state of RemoteStore to be only online when requested by the user & when the client is primary. Remote Store will still always contain all active targets.
- It adds SyncEngine.start()/shutdown() so that we can now enable/disable the network only from RemoteStore.
- It adds lots of Spec Test infrastructure:
  - The tracked state is now specific to each client.
  - The SpecTests now handle LocalStorage notifications via mocks (since the browser doesn't forward it's own writes).


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
...
+
+  start(eventListener: EventListener): void {
+    MockStorage.activeInstances.set(eventListener, this);
+    this.eventListener = eventListener;
+  }
+
+  shutdown(): void {
+    MockStorage.activeInstances.delete(this.eventListener);
+    this.eventListener = null;
+  }
```
mikelehen: If you keep start / shutdown, move them to the top?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
...
 }
 
+/**
+ * `WebStorage` mock that implements the WebStorage API used by Firestore.
+ */
+class MockStorage {
+  private static sharedData = new Map<string, string>();
+  private static activeInstances = new Map<EventListener, MockStorage>();
+
+  eventListener: EventListener | null = null;
```
mikelehen: So the data is static but the listener is not?  But there's a static map of listeners to MockStorage instances (where the only non-static member is the eventListener)?  This seems funky to me.  I don't know how I would describe this class. It's like a handle to a shared Storage instance but with a non-shared listener or something.

Would it make sense to drop `activeInstances` and just have a static array of `listeners`? and change start() / shutdown() into addListener() / removeListener() or similar?

That said I kinda' expect test mocks not to have any static state since that could bleed over between tests.  So maybe nothing should be static and you should just make sure that the same instance is used throughout the test (which may already be happening?  I'm not sure.)

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
...
+    this.raiseStorageEvent(key, oldValue, null);
+  }
+
+  setItem(key: string, value: string): void {
+    const oldValue = MockStorage.sharedData.get(key);
+    MockStorage.sharedData.set(key, value);
+    this.raiseStorageEvent(key, oldValue, value);
+  }
+
+  raiseStorageEvent(
```
mikelehen: private?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
...
+
+/**
+ * `Window` mock that implements the event and storage API that is used by
+ * Firestore.
+ */
+class MockWindow {
+  private webStorage = new MockStorage();
+
+  get localStorage(): Storage {
+    return this.webStorage as any;
```
mikelehen: Doesn't tslint complain?  FWIW it's probably not hard to make MockStorage fully implement Storage but I don't care either way.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
...
+
 /**
  * Implementation of `Platform` that allows mocking of `document` and `window`.
  */
 class TestPlatform implements Platform {
-  private mockDocument = new MockDocument();
-
-  constructor(private readonly basePlatform: Platform) {}
+  mockDocument: MockDocument | null = null;
+  mockWindow: MockWindow | null = null;
```
mikelehen: readonly?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1089,21 +1111,122 @@ class MockDocument {
   }
 }
 
+/**
+ * `WebStorage` mock that implements the WebStorage API used by Firestore.
+ */
+class MockStorage {
+  private static sharedData = new Map<string, string>();
+  private static activeInstances = new Map<EventListener, MockStorage>();
```
mikelehen: readonly?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -57,26 +97,34 @@ export class SpecBuilder {
...
-   * Automatically keeping track of the active targets makes writing tests
-   * much simpler and the tests much easier to follow.
-   *
-   * Whenever the map changes, the expected state is automatically encoded in
-   * the tests.
-   */
-  private activeTargets: {
-    [targetId: number]: { query: SpecQuery; resumeToken: string };
-  } = {};
+  private currentMemoryState: SpecMemoryState = new SpecMemoryState();
```
mikelehen: readonly?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -653,11 +660,37 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+
+  async enableNetwork(): Promise<void> {
+    this.networkEnabled = true;
+    if (this.isPrimary) {
+      return this.remoteStore.enableNetwork();
+    }
+  }
+
+  async disableNetwork(): Promise<void> {
+    // TODO(multi-tab): Release primary lease
```
mikelehen: Is it multi-tab or multitab?  Not sure if the code is already inconsistent, but you're inconsistent in this PR...

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -637,4 +637,74 @@ describeSpec('Writes:', [], () => {
...
+        .expectEvents(query, {
+          hasPendingWrites: true,
+          added: [docV1]
+        })
+        .client(1)
+        .userListens(query)
+        .expectEvents(query, {
+          hasPendingWrites: true,
+          // TODO(multitab): `fromCache` should reflect the primary client's
+          // networking state (and be false).
```
mikelehen: I'm glad you flagged this.  Do you have a plan to fix it?  I think fromCache=false means the primary tab has received a CURRENT message for the particular query since the last time the tab was in OnlineState.Offline.  Do we need to persist this (on a per-query basis) in IndexedDB or WebStorage somehow?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -333,12 +334,14 @@ export class FirestoreClient {
...
         this.eventMgr = new EventManager(this.syncEngine);
 
-        // NOTE: RemoteStore and SharedClientState depend on LocalStore (for
-        // persisting stream tokens, refilling mutation queue, etc.) so they must
-        // be started after LocalStore.
+        // NOTE: SyncEngine depends on both LocalStore and SharedClientState
+        // (for persisting stream tokens, refilling mutation queue, retrieving
+        // the list of active targets, etc.) so it must be started last.
+
+        // PORTING NOTE: RemoteStore does not need an explicit start on web.
```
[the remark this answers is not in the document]
schmidt-sebastian: This porting note should go away. Changing the way the network state works is likely the only thing that should be ported.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -410,12 +412,15 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
   }
 
   // PORTING NOTE: Multi-tab only
-  applyPendingBatch(batchId: BatchId): Promise<void> {
-    // TODO(multitab): Implement applyPendingBatch
-    throw new FirestoreError(
-      Code.UNIMPLEMENTED,
-      'applyPendingBatch not implemented'
+  async applyPendingBatch(batchId: BatchId): Promise<void> {
```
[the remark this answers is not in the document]
mikelehen: I think I like applyBatchFromPersistence().  But maybe just add a TODO on SharedClientStateSyncer for now.  It may make sense to rename all of those methods together (I wonder if adding "FromPersistence" suffixes on all of the methods might possibly make sense).

packages/firestore/src/core/sync_engine.ts
```diff
@@ -625,8 +630,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
         this.viewHandler!(newSnaps);
         return this.localStore.notifyLocalViewChanges(docChangesInAllViews);
       })
-      .then(() => {
-        return this.localStore.collectGarbage();
+      .then(async () => {
+        if (this.isPrimary) {
```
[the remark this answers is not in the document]
mikelehen: Cool. FWIW- if it simplifies our lives to only support LRU garbage collection with persistence, that's probably 100% fine.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -653,11 +660,37 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
   // PORTING NOTE: Multi-tab only
   applyPrimaryState(isPrimary: boolean): Promise<void> {
     this.isPrimary = isPrimary;
-    return Promise.resolve();
+    if (this.isPrimary && this.networkEnabled) {
+      return this.remoteStore.enableNetwork();
```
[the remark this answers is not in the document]
mikelehen: That definitely helps, thanks.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -89,7 +89,11 @@ export class LocalDocumentsView {
     return PersistencePromise.waitFor(promises).next(() => results);
   }
 
-  /** Performs a query against the local view of all documents. */
+  /**
+   * Performs a query against the local view of all documents.
+   *
+   * Multi-Tab Note: This operation is safe to use from secondary clients.
```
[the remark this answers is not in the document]
mikelehen: SGTM but can you add a TODO(multitab) somewhere to make sure we actually do that audit then? :)

packages/firestore/src/local/local_store.ts
```diff
@@ -307,6 +307,28 @@ export class LocalStore {
     );
   }
 
+  /** Reads a mutation batch from persistence and returns its changes. */
+  lookupMutationBatch(batchId: BatchId): Promise<LocalWriteResult | null> {
```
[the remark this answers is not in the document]
mikelehen: You shouldn't let reality get in the way of my dreams!

packages/firestore/src/local/mutation_queue.ts
```diff
@@ -81,14 +81,22 @@ export interface MutationQueue extends GarbageSource {
...
     transaction: PersistenceTransaction,
     localWriteTime: Timestamp,
     mutations: Mutation[]
   ): PersistencePromise<MutationBatch>;
 
-  /** Loads the mutation batch with the given batchId. */
+  /**
+   * Loads the mutation batch with the given batchId.
+   *
+   * Multi-Tab Note: This operation is safe to use from secondary clients.
```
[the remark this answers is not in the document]
mikelehen: FWIW I wasn't imagining plumbing it all the way down to SimpleDbStore since I figured it'd be too context-sensitive (e.g. secondary tabs can add mutation queue objects, but only the primary can delete them).  So I was imagining they'd live in the MutationQueue / RemoteDocumentCache / etc. classes.  Classes that need the primary lease would just assert transaction.hasPrimaryLease() or something.

But I don't feel very strongly about doing this. It may be more noise than it's worth. So don't bother unless you are more convinced than me.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -134,14 +137,6 @@ export class RemoteStore {
   /** SyncEngine to notify of watch and write events. */
   syncEngine: RemoteSyncer;
 
-  /**
-   * Starts up the remote store, creating streams, restoring state from
-   * LocalStore, etc.
-   */
-  start(): Promise<void> {
```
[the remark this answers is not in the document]
mikelehen: Okay, cool. Thanks for explaining.  I'd be down to have a no-op start() method, I think.  I like symmetry, especially since we have it in our other components.

packages/firestore/src/core/sync_engine.ts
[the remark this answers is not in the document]
schmidt-sebastian: Added this as a TODO.

packages/firestore/src/local/local_documents_view.ts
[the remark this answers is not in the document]
schmidt-sebastian: Added in persistence.ts

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1112,59 +1119,87 @@ class MockDocument {
...
+  getStorageArea(storageListener: EventListener): Storage {
+    const clientIndex = this.activeClients.length;
+
+    const storageArea: Storage = {
+      get length() {
+        return this.length;
+      },
+      getItem: (key: string) => this.getItem(key),
+      key: (index: number) => this.key(index),
+      clear: () => this.clear,
```
mikelehen: Should this be `this.clear()`?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -137,6 +137,16 @@ export class RemoteStore {
   /** SyncEngine to notify of watch and write events. */
   syncEngine: RemoteSyncer;
 
+  /**
+   * Starts up the remote store, creating streams, restoring state from
+   * LocalStore, etc.
+   */
+  start(): Promise<void> {
+    console.log('starting remotestore');
```
mikelehen: 😱

Issue #503, opened 2018-02-09
messaging.getToken() throws in service worker scope


acostalima:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS High Sierra 10.13
  * Firebase SDK version: 4.9.0
  * Firebase Product: messaging

### [REQUIRED] Describe the problem

According to the error trace I get when I call `messaging.getToken()` in service worker scope, the root of the error lies in a call to `window.btoa()` in which `window` is, obviously, not defined. The culprit can be found in a helper function in [array-buffer-to-base64.ts](https://github.com/firebase/firebase-js-sdk/blob/7a611b645fbedcd6c732853ed19e3f6a1a2638d5/packages/messaging/src/helpers/array-buffer-to-base64.ts#L18):

```js
function toBase64(arrayBuffer) {
   const uint8Version = new Uint8Array(arrayBuffer);
   return window.btoa(String.fromCharCode.apply(null, uint8Version));
}
```

After some experimentation, I found out that if I define the following

```js
const btoaCopy = self.btoa; // self is ServiceWorkerGlobalScope
self.window = {
   btoa: (...args) => btoaCopy.call(self, ...args)
};
```
before calling `messaging.getToken()`, no error is thrown and the returned token string seems to be exactly the same as the one retrieved in the client.

Is this intended behavior, i.e., is `messaging.getToken()` not supposed to be called in service worker scope? If not, this should be an easy fix because `btoa` also defined in `ServiceWorkerGlobalScope`.

#### Steps to reproduce:

1. Ensure you are in service worker scope
2. Setup Firebase with `firebase.initializeApp()`
3. Retrieve messaging instance: `const messaging = firebase.messaging()`
4. Call `messaging.getToken()`

#### Relevant Code:

No code.

gauntface:
@acostalima how are you importing the code into your service worker?

acostalima:
ping @gauntface

gauntface:
@pinarx does your PR fix this issue: https://github.com/firebase/firebase-js-sdk/pull/579

pinarx:
@gauntface PR #579 should make messaging.getToken work from SW. So if that's the only issue in here, that should be resolved.

PR #623, opened 2018-04-03
Fixes getToken in SW

gauntface:
Fixes: https://github.com/firebase/firebase-js-sdk/issues/503


pinarx:
cc: @mmermerkaya

Woohoo, thank you! This was next on my list...

I'm curious why this wasn't caught on the SW tests though, which I guess ties into #624. In controller-get-token.test.ts, getToken tests are run both from SW and Window controllers. Is there too much mocking to hit this at the moment?

gauntface:
Yeah.

These tests are doing a bunch of stuff that I wouldn't do now (based of more working with testing and looking to others for inspiration).

PR #625, opened 2018-04-03
automaticDataCollectionEnabled Flag

jshcrowthe:
This PR changes `initializeApp` to support an options object as the second argument. It propagates the `name` and `automaticDataCollectionEnabled` flags to the `FirebaseAppImpl`.

### DIFF
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -57,6 +57,16 @@ export interface FirebaseNamespace {
    * Create (and initialize) a FirebaseApp.
    *
    * @param options Options to configure the services used in the App.
+   * @param config The optional config for your firebase app
+   */
+  initializeApp(
+    options: FirebaseOptions,
+    config?: FirebaseAppConfig
+  ): FirebaseApp;
+  /**
+   * Create (and initialize) a FirebaseApp.
+   *
+   * @param options Options to configure the services used in the App.
    * @param name The optional name of the app to initialize ('[DEFAULT]' if
    * omitted)
    */
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -24,6 +24,11 @@ export type FirebaseOptions = {
   [name: string]: any;
 };
 
+export interface FirebaseAppConfig {
+  name?: string;
+  automaticDataCollectionEnabled?: boolean;
+}
+
 export class FirebaseApp {
   /**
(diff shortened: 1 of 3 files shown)


packages/app-types/index.d.ts
```diff
@@ -36,13 +41,28 @@ export class FirebaseApp {
...
 export interface FirebaseNamespace {
+  /**
+   * Create (and initialize) a FirebaseApp.
+   *
+   * @param options Options to configure the services used in the App.
+   * @param config The optional config for your firebase app
+   */
+  initializeApp(
+    options: FirebaseOptions,
+    config?: FirebaseAppConfig
```
hiranya911: Can you define this as `string|FirebaseAppConfig`

packages/app/src/firebaseApp.ts
```diff
@@ -285,21 +299,36 @@ export function createFirebaseNamespace(): FirebaseNamespace {
...
-      if (typeof name !== 'string' || name === '') {
-        error('bad-app-name', { name: name + '' });
-      }
+  function initializeApp(
+    options: FirebaseOptions,
+    config?: FirebaseAppConfig
+  ): FirebaseApp;
+  function initializeApp(options: FirebaseOptions, name?: string): FirebaseApp;
+  function initializeApp(options: FirebaseOptions, rawConfig = {}) {
+    if (typeof rawConfig !== 'object' || rawConfig === null) {
```
hiranya911: Probably safer to check `typeof rawConfig === 'string'` here.

packages/app/src/firebaseApp.ts
```diff
@@ -285,21 +299,36 @@ export function createFirebaseNamespace(): FirebaseNamespace {
...
+      const name = rawConfig;
+      rawConfig = { name };
+    }
+
+    const config = rawConfig as FirebaseAppConfig;
+
+    if (config.name === undefined) {
```
hiranya911: `typeof config.name === 'undefined'`

packages/app-types/index.d.ts
jshcrowthe: We can, but Typescript supports overloading function definitions so I opted to do that instead.

packages/app/src/firebaseApp.ts
jshcrowthe: In this case, checking the `typeof config.name` isn't any different than checking if the value is `undefined`. If this were a top-level symbol, I'd agree.

PR #626, opened 2018-04-04
IID Model types

gauntface:
I was looking through the code and wanted to make a change but after noticing the lack of types I just figured it would be worth adding that first.

After tidying up the model, I noticed the tests were failing and it essentially opened a can of worms across the tests (A couple of passing tests that were throwing asynchronously or duplicate names.

### DIFF
--- a/packages/messaging/test/iid-model-delete-token.test.ts
+++ b/packages/messaging/test/iid-model-delete-token.test.ts
@@ -56,7 +56,7 @@ describe('Firebase Messaging > IIDModel.deleteToken()', function() {
       await globalIIDModel.deleteToken(fcmSenderId, fcmToken, fcmPushSet);
       throw new Error('Expected error to be thrown.');
     } catch (e) {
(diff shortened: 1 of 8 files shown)


gauntface:
cc @mmermerkaya

packages/messaging/test/iid-model-delete-token.test.ts
```diff
@@ -42,37 +42,38 @@ describe('Firebase Messaging > IIDModel.deleteToken()', function() {
...
     const errorMsg = 'invalid token';
-    let stubbedFetch = sinon.stub(window, 'fetch');
-    stubbedFetch.returns(fetchMock.jsonError(400, errorMsg));
+    const payload = JSON.stringify({
+      error: {
+        message: errorMsg
+      }
+    });
+
+    sandbox.stub(window, 'fetch').returns(fetchMock.jsonError(400, payload));
```
pinarx: fetchMock.jsonError already puts the error message into a JSON, you don't need the payload: https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/test/testing-utils/mock-fetch.ts

PR #632, opened 2018-04-04
Multi-Tab support for acknowledged batches

schmidt-sebastian:
Adding Multi-Tab support for acknowledged batches 

- The PR merges the SharedStareSyncer callbacks into a single function that is called applyBatchState.
- A user change now returns the list of mutations that are active after the change
- The PR splits the handling of outstanding writes and outstanding callbacks in the SpecTests. Outstanding writes are shared among clients (so that a secondary client can schedule a write that the primary client picks up). Outstanding callbacks are per client.


mikelehen:
Unassigning until https://github.com/firebase/firebase-js-sdk/pull/616 lands (or at least the bulk of the feedback has been resolved and this is rebased).

schmidt-sebastian:
Back to you, Mike.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -267,9 +267,13 @@ export class SpecBuilder {
     return this;
   }
 
-  drainQueue(): this {
+  drainQueue(options?: { expectUserCallback: boolean }): this {
```
mikelehen: Can you comment this, including the default behavior (especially since it's different than writeAcks and failWrite)?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -323,9 +323,28 @@ class EventAggregator implements Observer<ViewSnapshot> {
...
-interface OutstandingWrite {
-  mutations: Mutation[];
-  userCallback: Deferred<void>;
+/**
+ * FIFO queue that tracks all outstanding mutations for a single test run.
+ * As these mutations are shared among the set of active clients, any client can
+ * add or retrieve mutations.
+ */
+// PORTING NOTE: Multi-tab only.
+class OutstandingMutationTracker {
```
mikelehen: Agreed. Definitely the best MutationTracker I've seen!

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -323,9 +323,28 @@ class EventAggregator implements Observer<ViewSnapshot> {
...
+  private mutations: Mutation[][] = [];
```
mikelehen: "mutations" sounds like a list, but you have a list-of-lists.  

Methinks maybe:
1. OutstandingMutationTracker should be OutstandingBatchTracker or OutstandingWriteTracker.
2. "mutations" here should be "batches" or "writes" (and maybe add a comment).
3. push(mutation: Mutation[]) should be push(batch: Mutation[]) or push(write: Mutation[])

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -323,9 +323,28 @@ class EventAggregator implements Observer<ViewSnapshot> {
...
+  push(mutation: Mutation[]): void {
+    this.mutations.push(mutation);
+  }
+
+  peek(): Mutation[] {
+    assert(this.mutations.length > 0, 'No pending mutations');
+    return this.mutations[0];
+  }
+
+  poll(): Mutation[] {
```
mikelehen: Is poll() a standard name or a Java-ism?  I'd lean toward shift() to be more JS-idiomatic.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -360,7 +379,11 @@ abstract class TestRunner {
   private started = false;
   private serializer: JsonProtoSerializer;
 
-  constructor(protected readonly platform: TestPlatform, config: SpecConfig) {
+  constructor(
+    protected readonly platform: TestPlatform,
+    private outstandingMutations: OutstandingMutationTracker,
```
mikelehen: Rename this too if you rename the type.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -575,7 +598,8 @@ abstract class TestRunner {
 
   private doMutations(mutations: Mutation[]): Promise<void> {
     const userCallback = new Deferred<void>();
-    this.outstandingWrites.push({ mutations, userCallback });
+    this.outstandingMutations.push(mutations);
+    this.outstandingCallbacks.push(userCallback);
```
mikelehen: I originally wrote this:

> _So whenever we push to outstandingMutations we also push to outstandingCallbacks, but when we poll() from outstandingMutations we only _sometimes_ shift() from outstandingCallbacks (based on whether the step had expectUserCallback: true or not).  So it seems like these lists are going to end up out-of-sync (after a expectUserCallback: false step, subsequent steps will end up waiting on the wrong user callback).  But I'd expect to cause tests to fail...  so maybe I'm confused?_

But after staring at this for several minutes I figured out what's going on.  outstandingMutations is shared across clients but outstandingCallbacks is per-client, and outstandingCallbacks is sometimes consumed on its own without outstandingMutations (by doDrainQueue). And so everything should balance out.

This is a bit confusing / non-obvious. Maybe rename outstandingMutations to globalOutstandingMutations or crossClientOutstandingMutations or something? And can you comment outstandingMutations and outstandingCallbacks and their relationship to each other and the fact that outstandingMutations is shared across clients?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -821,8 +847,12 @@ abstract class TestRunner {
     await this.syncEngine.disableNetwork();
   }
 
-  private async doDrainQueue(): Promise<void> {
+  private async doDrainQueue(drainQueue: SpecDrainQueue): Promise<void> {
     await this.queue.drain();
+    if (drainQueue.expectUserCallback) {
```
mikelehen: Hrm...  It seems like we have an artificial testing limitation to only have a single mutation-result outstanding when we drain the queue?  So we can't test the primary ack'ing two secondary writes in a row?  Does this need to be a count or something?

On that note, I'm a little nervous with how decoupled the outstandingWrites and outstandingCallbacks are and how we're not really matching them up in any way or validating that callbacks that shouldn't be called aren't getting called, etc.  Seems like there's potential for accidental bugs / oversights.  Probably not something to solve right this moment, but if you have ideas, I think it might be worth a bit of investment at some point (e.g. assigning numbers to writes / callbacks and making the callback update a "lastUserCallback" number which we can then validate against the write or something).

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1559,6 +1598,11 @@ export type SpecWriteFailure = {
   expectUserCallback: boolean;
 };
 
+export type SpecDrainQueue = {
+  /** Whether the drain is expected to generate a user callback. */
```
mikelehen: As mentioned elsewhere I wonder if we can/should allow for multiple callbacks.

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -706,4 +706,53 @@ describeSpec('Writes:', [], () => {
...
+      .writeAcks(1000, { expectUserCallback: false })
+      .watchSends({ affects: [query] }, remoteDoc)
+      .watchSnapshots(1000)
+      .expectEvents(query, {
+        metadata: [remoteDoc]
+      })
+      .client(1)
+      .drainQueue({ expectUserCallback: true });
+    // TODO(heldacks): This event doesn't fire since we are holding the write
+    // acknowledgment.
```
mikelehen: I'd either change this to TODO(b/33446471) or add a comment to b/33446471 to grep the codebase for "heldacks" before resolving, else I don't think anybody will look for this. :)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -71,9 +71,26 @@ export interface SharedClientState {
...
+   */
   removeLocalPendingMutation(batchId: BatchId): void;
 
+  /**
+   * Verifies whether a Mutation Batch ID is associated with the current client.
+   */
+  hasLocalPendingMutation(batchId: BatchId): boolean;
+
+  /** Transitions a pending mutation to its final state. */
+  applyMutationState(
```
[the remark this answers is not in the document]
schmidt-sebastian: I named this `trackMutationResult`, which won my internal debate between `setMutationResult`, `persistMutationResult` and `recordMutationResult`.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -501,15 +517,31 @@ export class WebStorageSharedClientState implements SharedClientState {
 
   addLocalPendingMutation(batchId: BatchId): void {
     this.localClientState.addPendingMutation(batchId);
-    this.persistMutationState(batchId, 'pending');
     this.persistClientState();
+    this.persistMutationState(batchId, 'pending');
```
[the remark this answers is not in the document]
schmidt-sebastian: Hrm, I don't remember why this happened. Reverted it until it comes back to mind.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -501,15 +517,31 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+  }
+
+  applyMutationState(
+    batchId: BatchId,
+    state: 'acknowledged' | 'rejected',
+    error?: FirestoreError
+  ): void {
+    assert(
+      batchId >= this.getMinimumGlobalPendingMutation(),
+      `Cannot change mutation state for batch that is not pending: ${batchId}`
```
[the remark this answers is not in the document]
schmidt-sebastian: Oh yes, that is true. This shouldn't be needed in the first place. It did help me find some bugs in the tests though.

packages/firestore/test/unit/specs/spec_builder.ts
schmidt-sebastian: Reworked in https://github.com/firebase/firebase-js-sdk/pull/678 (and merged into this PR).

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: https://stackoverflow.com/questions/1590247/how-do-you-implement-a-stack-and-a-queue-in-javascript has 939 upvotes (that must mean it is the standard) and uses shift.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Renamed to `SharedWriteTracker` and made everything write. I feel the distinction that a write equals multiple mutations could be made more obvious by a typedef... but maybe I'll tackle that another day.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Done (`sharedWrites`).

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: As per https://github.com/firebase/firebase-js-sdk/pull/678, `outstandingCallbacks` is no longer a thing. `outstandingMutations` is now also `sharedWrites` and hopefully clears up some of the confusion.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Done and done :)

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: This is now gone and we do allow for multiple callbacks.

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: Changed to the bug number.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -793,13 +813,9 @@ abstract class TestRunner {
 
   private doWriteAck(writeAck: SpecWriteAck): Promise<void> {
     const updateTime = this.serializer.toVersion(version(writeAck.version));
-    const nextMutation = this.outstandingMutations.poll();
+    const nextMutation = this.sharedWrites.shift();
     return this.validateNextWriteRequest(nextMutation).then(() => {
       this.connection.ackWrite(updateTime, [{ updateTime }]);
```
mikelehen: FWIW- ackWrite() will end up calling SyncEngine.applySuccessfulWrite() which will end up resolving the Deferred<> representing the user callback, but since Promise continuations are called asynchronously, the actual user callback we created (that updates our acknowledged/rejectedDocs lists) won't run synchronously.  I'm guessing we do enough extra Promise/setTimeout gunk that the callback ends up deterministically running before we check the expectations, but since it's no longer explicit this could be fragile.  I think I'm okay with it... but it could be worth a comment somewhere or something perhaps.  Up to you.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -425,35 +423,38 @@ export class SpecBuilder {
    * expectUserCallback defaults to true if options are omitted.
    */
   writeAcks(
+    docs: string[],
```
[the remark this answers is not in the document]
schmidt-sebastian: Making this an array was indeed inspired by your previous comments. I debated adding an overload that just takes a string, but have now gone in favor of just accepting a string for now. It still makes sense to accept multiple callbacks though (after a drainQueue call).

I left the name as `doc` since that name used widely throughout the SpecTest framework.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: The Spec Tests call `this.queue.drain()` before invoking `validateStepExpectations. Since drain uses setTimeout(), it waits for all Promise chains to resolve beforehand.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -594,6 +595,29 @@ export class SpecBuilder {
...
 
+  expectUserCallbacks(docs: {
+    acknowledged?: string[];
+    rejected?: string[];
+  }): this {
+    this.assertStep('Expectations require previous step');
+    const currentStep = this.currentStep!;
+    currentStep.stateExpect = currentStep.stateExpect || {};
+    currentStep.stateExpect.userCallbacks = currentStep.stateExpect
+      .userCallbacks || { acknowledgedDocs: [], rejectedDocs: [] };
```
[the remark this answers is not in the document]
schmidt-sebastian: Filed https://b.corp.google.com/issues/78013824

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -597,11 +603,25 @@ abstract class TestRunner {
   }
 
   private doMutations(mutations: Mutation[]): Promise<void> {
+    const documentKeys = mutations.map(val => val.key.path.toString());
     const userCallback = new Deferred<void>();
```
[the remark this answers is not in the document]
schmidt-sebastian: Oh, nice! That makes this way cleaner.

PR #637, opened 2018-04-05
Add types

mmermerkaya:
Add types to @firebase/messaging

This makes src/ compatible with --noImplicitAny. test/ still needs work.

I'm aware that there are a lot of "any"s around. Think of this as a first pass :)


packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -66,8 +72,8 @@ export class ControllerInterface {
   async getToken(): Promise<string | null> {
     // Check with permissions
     const currentPermission = this.getNotificationPermission_();
-    if (currentPermission !== NotificationPermission.GRANTED) {
-      if (currentPermission === NotificationPermission.DENIED) {
+    if (currentPermission !== 'granted') {
+      if (currentPermission === 'denied') {
```
pinarx: why get rid of the enum?

mmermerkaya: TS already has a type for NotificationPermission: https://github.com/Microsoft/TypeScript/blob/5f9c34bcd2933075a382abbcc8606b0a61b582fe/lib/lib.dom.d.ts#L16034

It only allows these strings to be assigned to a variable with this type, so if I made a typo in 'granted' or 'denied' here, TS would not compile.

packages/messaging/src/interfaces/token-details.ts
```diff
@@ -14,8 +14,16 @@
...
-  DENIED = 'denied'
+export interface TokenDetails {
+  fcmToken: string;
+  swScope: string;
+  vapidKey: string | Uint8Array;
+  subscription: PushSubscription;
+  fcmSenderId: string;
+  fcmPushSet: string;
+
+  createTime?: number;
```
pinarx: Is ? still needed?

mmermerkaya: Yep, it's not defined everywhere TokenDetails is used.

I think we should eventually split the database version and code version (which doesn't define these optional members until db operations) of TokenDetails, and make the DB version extend the regular version. WDYT?

pinarx: synced offline, leaning towrads using Partial, or just setting the pushSubscription details when the TokenDetails are created, rather than in TokenDetails model.

PR #639, opened 2018-04-05
Fixing "Type could be null" errors

jshcrowthe:
Fixes some errors we are seeing in the messaging compilation.


mmermerkaya:
Please don't use "Object" as a type. Use "object" instead: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

(no need to do anything about it, as my PR (#637) will fix it)

PR #640, opened 2018-04-05
Misc Fixes

jshcrowthe:
There were a couple errors that were popping up causing the build to fail in a number of ways (e.g. gulp version was changing unexpectedly, messaging compilation was throwing errors, package-locks were causing issues, etc).

I have fixed all of these in this PR. PTAL.


packages/messaging/tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "../../config/tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "dist",
-    "strict": true
+    "outDir": "dist"
```
mmermerkaya: What errors are you getting because of this?

packages/messaging/tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "../../config/tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "dist",
-    "strict": true
```
hiranya911: Might make sense to create an issue to add this back later if possible. Admin Node SDK has had one for a while.

PR #641, opened 2018-04-05
Add TSLint and (mostly uncontroversial) rules

mmermerkaya:
Fixed lint errors. Added running TSLint to the test command in package.json.

PTAL at the rules and let me know if any of them don't make sense to you.


packages/messaging/package.json
```diff
@@ -7,10 +7,13 @@
   "module": "dist/esm/index.js",
   "scripts": {
     "dev": "gulp dev",
-    "test": "karma start --single-run && yarn type-check",
+    "test": "run-p test:karma type-check lint",
```
gauntface: I'm not seeing run-p being added anywhere as a devDependency.

gauntface: Also are logs going to be clean if one of these errors out?

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -14,33 +14,33 @@
...
 
 const SENDER_ID_OPTION_NAME = 'messagingSenderId';
 // Database cache should be invalidated once a week.
 export const TOKEN_EXPIRATION_MILLIS = 7 * 24 * 60 * 60 * 1000; // 7 days
 
 export abstract class ControllerInterface {
-  public app: FirebaseApp;
-  public INTERNAL: any;
+  app: FirebaseApp;
+  INTERNAL: any;
```
gauntface: These originally were public to ensure the Firebase SDK would work with the module (i.e. the name shouldn't be mangled).

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -215,7 +215,7 @@ export class SWController extends ControllerInterface {
       return;
     }
 
-    const notificationInformation = Object.assign({}, msgPayload.notification);
+    const notificationInformation = { ...msgPayload.notification };
```
gauntface: This feels less clear that you are creating a clone of the object.

packages/messaging/package.json
mmermerkaya: run-p is part of npm-run-all. There's a lot of run-p and run-s usage within this repo so I'm assuming it's going to be fine?

packages/messaging/src/controllers/controller-interface.ts
mmermerkaya: Public is the default, so they are still public. This is just removing the redundant "public" keywords.

packages/messaging/src/controllers/sw-controller.ts
mmermerkaya: I'll have to disagree, using the spread operator feels more natural to me in this case :)

I'd like to keep this anyway, because spread operator allows for better type checking and inference, so it's more than a style choice in this case.

https://palantir.github.io/tslint/rules/prefer-object-spread/

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -55,7 +55,7 @@ export abstract class ControllerInterface {
       throw this.errorFactory_.create(ERROR_CODES.BAD_SENDER_ID);
     }
 
-    this.messagingSenderId_ = app.options[SENDER_ID_OPTION_NAME]!;
+    this.messagingSenderId_ = app.options[SENDER_ID_OPTION_NAME];
```
pinarx: so this can be null? It shouldn't right?

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -121,14 +121,14 @@ export abstract class ControllerInterface {
...
-    const isTokenValid = this.isTokenStillValid(
+    const isTokenValid = isTokenStillValid(
       pushSubscription,
       publicVapidKey,
       tokenDetails
     );
     if (isTokenValid) {
       const now = Date.now();
-      if (now < tokenDetails.createTime! + TOKEN_EXPIRATION_MILLIS) {
+      if (now < tokenDetails.createTime + TOKEN_EXPIRATION_MILLIS) {
```
pinarx: again, why is createTime becoming nullable? Is it just for TS at this point?

pinarx: same comment for getKey in isTokenValid method.

packages/messaging/src/controllers/controller-interface.ts
mmermerkaya: I had to remove these because --strict (and --strictNullChecks with it) was disabled yesterday. I have another PR coming soon to reenable strict and revert this change.

PR #643, opened 2018-04-06
Re-enable --strict flag in messaging/

mmermerkaya:
Fixes the problems in app/ that were causing build issues.


packages/messaging/src/models/token-details-model.ts
```diff
@@ -138,7 +138,7 @@ export class TokenDetailsModel extends DBInterface {
           request.onsuccess = event => {
             const result: TokenDetails | null = (event.target as IDBRequest)
               .result;
-            resolve(result);
+            resolve(result!);
```
gauntface: The line above explicitly says this could be null.

packages/messaging/src/models/token-details-model.ts
```diff
@@ -170,7 +170,7 @@ export class TokenDetailsModel extends DBInterface {
           scopeRequest.onsuccess = event => {
             const result: TokenDetails | null = (event.target as IDBRequest)
               .result;
-            resolve(result);
+            resolve(result!);
```
gauntface: Same as previous comment - this explicitly is called as possibly being null.

packages/messaging/src/controllers/controller-interface.ts
```diff
@@ -128,7 +128,7 @@ export abstract class ControllerInterface {
     );
     if (isTokenValid) {
       const now = Date.now();
-      if (now < tokenDetails.createTime + TOKEN_EXPIRATION_MILLIS) {
+      if (now < tokenDetails.createTime! + TOKEN_EXPIRATION_MILLIS) {
```
gauntface: token details should be our own internal data structure so either the type is wrong or we should handle this being null.

If you want to just keep the current behavior can we add a TODO comment to handle if it's null.

packages/messaging/src/models/token-details-model.ts
mmermerkaya: Yeah. It is not right. That is why I want to enable --strict, so these issues become obvious :)

I tried not to change any functionality in this PR. I'm already working on fixing the types, and I'll take care of these next week.

packages/messaging/src/controllers/controller-interface.ts
mmermerkaya: TokenDetails without createTime is used throughout the codebase. CreateTime gets added right before it's saved to IndexedDB. We talked about it with @pinarx offline and decided to add createTime (and some other missing properties) as soon as the TokenDetails object is created. This will remove the need to make these types optional.

I didn't add a TODO because I already have a PR that fixes this, but I can't submit it before my other PRs are merged.

PR #645, opened 2018-04-06
 Refactor IndexedDb operations

mmermerkaya:
Simplifies a lot of IndexedDb operations and removes most of the duplicated code. Also refactors tests.

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -367,16 +367,19 @@ export class SWController extends ControllerInterface {
    * This will return the default VAPID key or the uint8array version of the
    * public VAPID key provided by the developer.
    */
-  getPublicVapidKey_(): Promise<Uint8Array> {
-    return this.getSWRegistration_()
-      .then(swReg => {
-        return this.getVapidDetailsModel().getVapidFromSWScope(swReg.scope);
-      })
-      .then(vapidKeyFromDatabase => {
-        if (vapidKeyFromDatabase === null) {
-          return DEFAULT_PUBLIC_VAPID_KEY;
-        }
-        return vapidKeyFromDatabase;
-      });
+  async getPublicVapidKey_(): Promise<Uint8Array> {
+    const swReg = await this.getSWRegistration_();
+    if (!swReg) {
+      return DEFAULT_PUBLIC_VAPID_KEY;
+    }
+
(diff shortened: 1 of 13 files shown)


packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -367,16 +367,19 @@ export class SWController extends ControllerInterface {
...
-      })
-      .then(vapidKeyFromDatabase => {
-        if (vapidKeyFromDatabase === null) {
-          return DEFAULT_PUBLIC_VAPID_KEY;
-        }
-        return vapidKeyFromDatabase;
-      });
+  async getPublicVapidKey_(): Promise<Uint8Array> {
+    const swReg = await this.getSWRegistration_();
+    if (!swReg) {
```
pinarx: this should never be the case right? Does it make more sense to throw an error instead?

packages/messaging/src/models/db-interface.ts
```diff
@@ -18,75 +18,129 @@ import { ErrorFactory } from '@firebase/util';
...
+          this.onDbUpgrade(request.result, event);
+      });
+    }
+
+    return this.dbPromise;
+  }
+}
+
+/** Promisifies an IDBRequest. Resolves with the IDBRequest's result. */
+function promisify<ReturnType>(request: IDBRequest): Promise<ReturnType> {
```
gauntface: Nit: I've generally seen people use `T` when using generics rather than `ReturnType`.

packages/messaging/src/models/token-details-model.ts
```diff
@@ -119,190 +107,114 @@ export class TokenDetailsModel extends DBInterface {
...
-        });
-      });
-    });
+    await this.delete(details.swScope);
+    return details;
   }
 }
+
+/** Promisifies an IDBRequest. Resolves with the IDBRequest's result. */
+function promisify<ReturnType>(request: IDBRequest): Promise<ReturnType> {
```
gauntface: It's probably worth pulling this function out into a shared utility file so it's not duplicated.

packages/messaging/src/models/db-interface.ts
mmermerkaya: Yeah that makes sense, done.

packages/messaging/src/models/token-details-model.ts
mmermerkaya: I don't think we should do this now. It's not going to be used anywhere else as this is the only file that interacts directly with IndexedDB.

PR #649, opened 2018-04-09
Fix TokenDetails in messaging/

mmermerkaya:
Makes sure that we use only one type of TokenDetails object, both in the code and in the database.

Updates DB and adds a function to fix the old values.

Fixes and adds tests.


packages/messaging/test/controller-get-token.test.ts
```diff
@@ -37,44 +38,47 @@ const ONE_DAY = 24 * 60 * 60 * 1000;
 describe('Firebase Messaging > *Controller.getToken()', () => {
   const sandbox = sinon.sandbox.create();
   const now = Date.now();
-  const expiredDate = new Date(now - ONE_DAY);
+  const expiredDate = now - ONE_DAY * 8; // 8 days ago
```
pinarx: i'm guessing the test wasn't supposed to be passing before this? Did you hit an issue when running the tests, or is the mocking making it work in either case?

mmermerkaya: Yeah, the date in the test was wrong. But the way the date is created (using a Date object instead of a timestamp) was also wrong, which made the tests pass. :man_shrugging:

mmermerkaya:
Ping :)

PR #651, opened 2018-04-09
Deprecate downloadURL property in Storage metadata types.

sphippen:
In the future, Storage backend changes will make it infeasible to automatically provide these properties. These properties will eventually be removed, but it will still be possible to generate download URLs with the Reference.getDownloadURL method.

The `@deprecated` property will show up in TSLint. I'll also update our [documentation](https://firebase.google.com/docs/) before the next release.


jshcrowthe:
This also needs to be updated in `packages/firebase/index.d.ts` and `packages/firebase/app/index.d.ts`

sphippen:
Thanks for pointing that out. Everything should be updated now.

PR #663, opened 2018-04-10
Remove redundant lib declarations

mmermerkaya:
es2015 includes es5 and es2015.promise

https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es2015.d.ts

### DIFF
--- a/config/tsconfig.base.json
+++ b/config/tsconfig.base.json
@@ -4,9 +4,7 @@
     "declaration": true,
     "importHelpers": true,
     "lib": [
-      "es5",
       "es2015",
-      "es2015.promise",
       "dom"
     ],
     "module": "commonjs",


jshcrowthe:
@mmermerkaya we are polyfilling as needed. You raise a very valid point that we could probably just use the lib option to fix this for us.

i.e. - Scope to only those things that are allowed and then if a user wants to use a new API they have to add it to the `@firebase/polyfill` package.

mmermerkaya:
That sounds good. I'll merge this as is for now, and maybe we can open an issue about that?

Also, Messaging doesn't really target ES5 (and need polyfills) as it only supports browsers with Service Worker support, all of which currently support ES2017. But I think we can just override the --lib flag in that package's own tsconfig.

jshcrowthe:
SGTM. FWIW I'm against having an API run ahead w/o getting the shims/polyfills into the `@firebase/polyfill` package. If you want to bump to ES2017 (which I'm totally fine w/ btw) lets just get the polyfills in place. 

If you wanted to dig into this deeper, ideally we would identify and generate the `@firebase/polyfill` package based on what we are using. If there is a good on demand TS polyfill packaing thing, I'd love to know more about it.

PR #667, opened 2018-04-10
Enabling TSLint for our test cases

schmidt-sebastian:
The way we ran TSLint didn't pick up our test cases in nested subfolders. This PR enables TSLint for all files and also addresses the warnings in the test files.

Note that we could probably spend a little more effort on the 'any' cleanup, but I'd like to punt this for now.

To work around a TSLint issue with Chai, this PR disables the no-unused-expression rule (see https://github.com/jshint/jshint/issues/3070)


packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -375,15 +375,16 @@ apiDescribe('Database transactions', persistence => {
     return integrationHelpers.withTestDb(persistence, db => {
       const doc1 = db.collection('counters').doc();
       const doc2 = db.collection('counters').doc();
-      let tries = 0;
+      // let tries = 0;
```
rsgowman: rm?

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -375,15 +375,16 @@ apiDescribe('Database transactions', persistence => {
...
       return (
         doc1
           .set({
             count: 15
           })
           .then(() => {
             return db.runTransaction(transaction => {
-              tries++;
+              // ++tries;
```
rsgowman: rm? And if you do so, I think you could optionally simplify line 385-390 slightly:

```
return db.runTransaction(transaction -> transaction.get(doc1).then(...).then(...))
```

packages/firestore/test/unit/model/path.test.ts
```diff
@@ -38,7 +38,7 @@ describe('Path', () => {
   addEqualityMatcher();
 
   it('can be constructed', () => {
-    const path = new ResourcePath(['rooms', 'Eros', 'messages']);
+    const _ = new ResourcePath(['rooms', 'Eros', 'messages']);
```
rsgowman: Could you `s/const _ = //` instead? (Or does the linter then complain that you've not used the returned value?)

schmidt-sebastian: For better of for worse, since we turned of no_unused_expressions, I can indeed just remove the assignment.

packages/firestore/test/integration/api/transactions.test.ts
schmidt-sebastian: This code is used in the commented out part of this tests that verifies the retry behavior. Removing it means that we cannot easily comment the block back in once b/34879758 gets addressed.

packages/firestore/test/integration/api/transactions.test.ts
schmidt-sebastian: See above.

packages/firestore/test/integration/api/transactions.test.ts
rsgowman: In that case, I'd recommend adding a TODO(b/34879758) or similar stating this. Otherwise, I'll quickly forget this conversation and remove that code next time I touch this file.

schmidt-sebastian: It's part of a TODO that is just below this block.

rsgowman: (after clicking expand a few times...)

Oh, I see. Given that I missed it once, there's a good chance without a TODO I'd miss it again. However, in order to actually change it, I'd have to load it up in my editor rather than the github diff viewer, so maybe I'd see it in that situation.

(No action required.)

PR #668, opened 2018-04-11
Porting Polyfills Fix (missing in master)

jshcrowthe:
Porting a fix that was made but didn't get added to the repo.


packages/polyfill/src/polyfills/fetch.ts
```diff
@@ -0,0 +1,22 @@
+/**
```
mmermerkaya: Is this file included in the build? I don't see it imported anywhere.

PR #673, opened 2018-04-11
 Add types for Service Worker Controller

mmermerkaya:
Also updates TypeScript to 2.8.1, which has much better WebWorker and Notification typings, which allowed me to remove some more "any" casts.


packages/messaging/src/controllers/sw-types.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Subset of Web Worker types from lib.webworker.d.ts
+ *
+ * Since it's not possible to have both "dom" and "webworker" libs in a single
+ * project, we have to manually declare the web worker types we need.
+ */
+
+// Not the whole interface, just the parts we're currently using.
+// If TS claims that something does not exist on this, feel free to add it.
+interface ServiceWorkerGlobalScope {
```
jshcrowthe: Does this not already exist as a TSLib?

mmermerkaya: It exists (as a type definition file), but it can't be used simultaneously with DOM.

https://github.com/Microsoft/TypeScript/issues/20595
https://github.com/Microsoft/TypeScript/issues/14687

So I took some parts of it that we need (and doesn't conflict with DOM) and put them here.

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -14,61 +14,67 @@
...
+} from '../models/worker-page-message';
 import { ControllerInterface } from './controller-interface';
 
+// Let TS know that this is a service worker
+declare const self: ServiceWorkerGlobalScope;
+
 const FCM_MSG = 'FCM_MSG';
 
-export type BgMessageHandler = (input: any) => Promise<any>;
+export type BgMessageHandler = (input: Payload) => void;
```
gauntface: You could be expecting a promise to be returned here so you can pass the promise to event.waitUntil(). Otherwise the service worker could be terminated before the developers work has finished.

mmermerkaya: That's a good point, thanks!

packages/messaging/src/controllers/sw-types.ts
mmermerkaya: Added link.

PR #674, opened 2018-04-11
Expand browsers support

yifanyang:

### Discussion

### Testing

### API Changes

    can discuss it together.


.travis.yml
```diff
@@ -31,11 +27,21 @@ addons:
...
-  on:
-    repo: firebase/firebase-js-sdk
-    branch: master
+jobs:
+  include:
+    - script: xvfb-run yarn test
+      after_success: yarn test:coverage
+      env: '#comment: yarn test'
+    - script: xvfb-run yarn test:saucelabs
+      env: '#comment: yarn test:saucelabs'
```
jshcrowthe: So I think we want to just completely disable this stage in the event that it is not a push build, I just realized that we run the same suite twice differently in two browsers and that seems unnecessary.

See: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ for an easy way of doing this.

config/karma.saucelabs.js
```diff
@@ -0,0 +1,153 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const glob = require('glob');
+const karma = require('karma');
+const path = require('path');
+const karmaBase = require('./karma.base');
+
+/** Tests in these packages are exluded due to flakiness or long run time. */
```
jshcrowthe: nit: exluded -> excluded

config/karma.saucelabs.js
```diff
@@ -0,0 +1,153 @@
...
+    sauceLabs: {
+      tunnelIdentifier: process.env.TRAVIS_JOB_NUMBER,
+      username: process.env.SAUCE_USERNAME,
+      accessKey: process.env.SAUCE_ACCESS_KEY,
+      startConnect: true
+    }
+  });
+
+  /* https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions */
+  if (process.env.TRAVIS_PULL_REQUEST != 'false') {
```
jshcrowthe: Per comment above: 

Let's go ahead and run these tests only during the push builds and preserve the resource if we can (see: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/)

We can probably remove this block.

config/karma.saucelabs.js
```diff
@@ -0,0 +1,153 @@
...
+};
+
+module.exports = function(config) {
+  const karmaConfig = Object.assign({}, karmaBase, {
+    basePath: '../',
+
+    files: getTestFiles(),
+
+    client: Object.assign({}, karmaBase.client, {
+      firestoreSettings: {
```
jshcrowthe: Can you leave a comment explaining why this is needed?

.travis.yml
yifanyang: Done.

I'm using env as a description for these two Travis jobs. It turns out that for if condition to work, there has to be a '=' in the env statement.

config/karma.saucelabs.js
yifanyang: Done.

config/karma.saucelabs.js
yifanyang: Done.

config/karma.saucelabs.js
yifanyang: Removed.

This was copied from karma config from packages/firestore. But we don't actually need it, as firestore tests are excluded for long running time right now. We still want to enable firestore tests eventually, and will figure out what exactly need to be done later.

yifanyang: Yes. Well, I guess we could added it back the day we enabled the firestore tests.

PR #675, opened 2018-04-11
Adjust test util withTestDoc() to accept initial data

rsgowman:
Similar to withTestCollection.

### DIFF
--- a/packages/firestore/test/integration/util/helpers.ts
+++ b/packages/firestore/test/integration/util/helpers.ts
@@ -200,20 +200,6 @@ export function withTestDbsSettings(
 
 export function withTestDoc(
   persistence: boolean,
-  fn: (doc: firestore.DocumentReference) => Promise<void>
-): Promise<void> {
-  return withTestDb(persistence, db => {
-    return fn(db.collection('test-collection').doc());
-  });
-}
-
-// TODO(rsgowman): Modify withTestDoc to take in (an optional) initialData and
-// fix existing usages of it. Then delete this function. This makes withTestDoc
-// more analogous to withTestCollection and eliminates the pattern of
-// `withTestDoc(..., docRef => { docRef.set(initialData) ...});` that otherwise is
-// quite common.
-export function withTestDocAndInitialData(
-  persistence: boolean,
   initialData: firestore.DocumentData | null,
   fn: (doc: firestore.DocumentReference) => Promise<void>
 ): Promise<void> {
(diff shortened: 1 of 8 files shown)


packages/firestore/test/integration/util/helpers.ts
```diff
@@ -199,20 +199,6 @@ export function withTestDbsSettings(
...
-    return fn(db.collection('test-collection').doc());
-  });
-}
-
-// TODO(rsgowman): Modify withTestDoc to take in (an optional) initialData and
-// fix existing usages of it. Then delete this function. This makes withTestDoc
-// more analogous to withTestCollection and eliminates the pattern of
-// `withTestDoc(..., docRef => { docRef.set(initialData) ...});` that otherwise is
-// quite common.
-export function withTestDocAndInitialData(
```
mikelehen: So TypeScript actually supports a mechanism for method overloads. It's a bit wacky (you basically list out all the overload signatures and then implement a function whose signature is a superset of all the overloads and then dynamically handle the various possible calls at runtime).

So that would look something like:
```
export type WithTestDocCallback = (doc: firestore.DocumentReference) => Promise<void>;
export function withTestDoc(
  persistence: boolean,
  fn: WithTestDocCallback
);
export function withTestDoc(
  persistence: boolean,
  initialData: firestore.DocumentData,
  fn: WithTestDocCallback
);
export function withTestDoc(
  persistence: boolean,
  initialDataOrFn: firestore.DocumentData | WithTestDocCallback,
  fn?: WithTestDocCallback
): Promise<void> {
  let initialData = null;
  if (fn) {
    initialData = initialDataOrFn;
  } else {
    fn = initialDataOrFn as WithTestDocCallback;
  }
  return withTestDb(persistence, db => {
    const docRef: firestore.DocumentReference = db
      .collection('test-collection')
      .doc();
    if (initialData) {
      return docRef.set(initialData).then(() => fn(docRef));
    } else {
      return fn(docRef);
    }
  });
}
```

And then all the places you're passing `null` can go back to how they were before.  

Sorry, this is what I had in my mind when I asked you to do the TODO and would have been a much easier migration. Though I like that you've gone ahead and cleaned up a bunch of tests to leverage initialData. That's a nice improvement. Thanks!

schmidt-sebastian:
Rich, when you change this to Master you will also notice that this now conflicts considerably with the changes that I merged in https://github.com/firebase/firebase-js-sdk/pull/667

I would suggest that you pick your changes over mine and fix the remaining lint errors manually.

packages/firestore/test/integration/util/helpers.ts
rsgowman: Ah, I see. Done.

It's just as well I initially did it this way though. Otherwise, I would've been tempted to not do the cleanups. :)

PR #677, opened 2018-04-12
Minor additional UserDataConverter cleanup.

mikelehen:
I made some minor tweaks while porting https://github.com/firebase/firebase-js-sdk/commit/b14678002bf6e8cc9ebd7561efbafe29315ceb8e to iOS. The big diff is because I rearranged methods but I did that in a separate commit to make reviewing easier.

### DIFF
--- a/packages/firestore/src/api/user_data_converter.ts
+++ b/packages/firestore/src/api/user_data_converter.ts
@@ -527,7 +527,7 @@ export class UserDataConverter {
         // No transform to add for a delete, so we do nothing.
       } else if (context.dataSource === UserDataSource.Update) {
         assert(
-          context.path == null || context.path.length > 0,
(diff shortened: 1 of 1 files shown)


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -574,7 +527,7 @@ export class UserDataConverter {
         // No transform to add for a delete, so we do nothing.
       } else if (context.dataSource === UserDataSource.Update) {
         assert(
-          context.path == null || context.path.length > 0,
+          context.path.length > 0,
```
mikelehen: Note that this is safe since at the top of the method we do:

```
    if (context.path === null) {
      throw context.createError(
        `${value.methodName} is not currently supported inside arrays`
      );
    }
```

Issue #658, opened 2018-04-10
Error in Internet Explorer 11 (Messaging)


metalgigio:
**Description of the Issue:**

Internet Explorer 11 fire a javascript error that obviously stop every following javascript

file: webpack:///messaging/dist/esm/src/helpers/base64-to-array-buffer.js
line: 17
Error (in italian): L'oggetto non supporta la proprietà o il metodo 'repeat'

**Environment Configuration:**

Operating System version: Windows 10
Firebase SDK version: 4.12.1
Firebase Product: Messaging

mmermerkaya:
Messaging is not supported by IE 11, because IE 11 does not support service workers.

For more information please see: https://firebase.google.com/docs/cloud-messaging/js/client

metalgigio:
Yes, i know isn't supported, but as i wrote that error broke all the js. So entire js doesnt work in ie11 since is, as best practice, minified all in one file.

mmermerkaya:
I'd either not include messaging in your ES5 build, or add the necessary polyfills to make your code work in IE 11.

I'll investigate this and check if we call base64ToArrayBuffer if the app is running in an unsupported browser.

metalgigio:
Well, ur right i should use polyfills but i prefer keep low js size since ie11 doesn not support messaging even with a polyfills.
For the record, or if someone needed too, I solved enclosing all messaging code like this:

```
if ('serviceWorker' in navigator) {
    var config = {
     ...
    };
    firebase.initializeApp(config);
    ...
}
```

PR #684, opened 2018-04-13
Lint rules, refactoring and better types

mmermerkaya:
Sorry for the large PR, but I wanted to get everything out before the weekend :)

Enables TSLint rules `no-redundant-jsdoc`, `variable-name` and `no-any` (no-any is still disabled in test/).

Enables `no-implicit-any` compiler option. With this change, we've enabled all of `--strict` in messaging.

Refactors ErrorFactory to have only one factory object for all of messaging.

Some more test and async/await refactoring.

Changes the `--lib` target to ES2017 for messaging. This is (totally) safe, as messaging only works in [browsers with service worker support](https://www.caniuse.com/#feat=serviceworkers). All of these browsers [support all of ES2017](http://kangax.github.io/compat-table/es2016plus/) (except shared memory and atomics, which is disabled because of meltdown/spectre).

~~To make this even safer, messaging now checks browser support **before** instantiating the service controller. This also prevents issues such as #658 by making sure we don't call any unsupported methods by accident.~~ This is removed from this PR and will be submitted separately.


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -556,10 +556,7 @@ export class UserDataConverter {
    *
    * @return The parsed value
    */
-  private parseScalarValue(
-    value: AnyJs,
-    context: ParseContext
-  ): FieldValue {
+  private parseScalarValue(value: AnyJs, context: ParseContext): FieldValue {
```
gauntface: @jshcrowthe this keeps on happening where another modules code is touched because of formatting - this really isn't good for PR's.

packages/messaging/index.ts
```diff
@@ -19,20 +19,29 @@ import {
...
     if (self && 'ServiceWorkerGlobalScope' in self) {
+      // Running in ServiceWorker context
       return new SWController(app);
-    }
+    } else {
+      // Assume we are in the window context.
+
+      // Check that the required API's are available
+      if (!isWindowControllerSupported()) {
+        throw errorFactory.create(ERROR_CODES.UNSUPPORTED_BROWSER);
```
gauntface: Won't this cause an app to completely fail just by loading the library?

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -68,13 +73,6 @@ export class WindowController extends ControllerInterface
    * permission isn't granted.
    */
   getToken(): Promise<string | null> {
-    // Check that the required API's are available
-    if (!this.isSupported_()) {
```
gauntface: The check should stay here - moving it to the module instantiation could break peoples applications.

packages/messaging/index.ts
mmermerkaya: No. It only throws when `firebase.messaging()` is called. You can still import '@firebase/messaging' or the complete firebase bundle without any errors.

I think that if a user calls `firebase.messaging()` to get the messaging service, they should receive an error instead of getting a service controller that will throw as soon as getToken is called. WDYT?

gauntface: It's going to break current developers and should be done with a major version bump.

mmermerkaya: So you're okay with it as long as we bump the version? We can release this with v5.

mmermerkaya:
I removed the isSupported changes from this PR.

PR #690, opened 2018-04-16
Fix data message not being handled on foreground when backgroundMessageHandler isn't set

### LINKED ISSUE #381: Data message not handled on foreground when backgroundMessageHandler isn't set

  * Operating System version: Windows 10 x64 
  * Firebase SDK version: 4.8.0
  * Firebase Product: messaging (0.1.5)

#### Steps to reproduce:

1. Don't call `messaging.setBackgroundMessageHandler(...)` in `firebase-messaging-sw`, i.e. don't handle background push.

2. Send a data message that doesn't have a "notification" field.
The message in my case:
image

3. The foreground receives the message, but isn't handled due to this line:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/src/controllers/sw-controller.ts#L76
  ```ts
if (msgPayload.notification || this.bgMessageHandler_) {
    // Send to page
    return this.sendMessageToWindowClients_(msgPayload);
}
  ```
  since neither of the conditions is true. (I'm not sure what's the purpose of the check, why wouldn't a window client want to handle data messages?)

#### Relevant Code:

Just basic examples from the guide, without using `setBackgroundMessageHandler`. (from https://firebase.google.com/docs/cloud-messaging/js/receive)

mmermerkaya:
According to https://firebase.google.com/docs/cloud-messaging/js/receive, a message should always be sent to the window client if it is in focus.

This change removes the checks that prevent the message from being sent if it is a data message and if bgMessageHandler isn't set.

Fixes #381.

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -98,12 +98,8 @@ export class SWController extends ControllerInterface {
 
     const hasVisibleClients = await this.hasVisibleClients_();
     if (hasVisibleClients) {
-      // Do not need to show a notification.
-      if (msgPayload.notification || this.bgMessageHandler) {
-        // Send to page
-        return this.sendMessageToWindowClients_(msgPayload);
-      }
-      return;
+      // App in foreground. Send to page.
+      return this.sendMessageToWindowClients_(msgPayload);
     }
 
     const notificationDetails = this.getNotificationData_(msgPayload);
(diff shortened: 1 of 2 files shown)


gauntface:
The reason for this specific behavior was that if the behavior of the SDK was completely not what someone wanted, that escape route would be to `data` payload and just set up your own push event handler (i.e. don't use the backgroundMessageHandler).

@pinarx I'm fine with this being simplified, but it could break some people if they are using this undocumented behaviour.

pinarx:
There are already some stuff we'd like to change on the message routing AFAIK. Let's come up with a plan for that, and fix this accordingly - that way we don't do multiple breaking changes.

gauntface:
@mmermerkaya the API is generally very confusing :(

pinarx:
+1 to have this on v5.

mmermerkaya:
Ping @gauntface.

PR #696, opened 2018-04-17
Options cleanup for includeMetadataChanges

schmidt-sebastian:
This PR unifies `includeQueryMetadataChanges` and `includeDocumentMetadataChanges` into `includeMetadataChanges` and modifies `docChanges` to be a method that can take options.


packages/firestore/src/core/event_manager.ts
```diff
@@ -266,18 +244,19 @@ export class QueryListener {
   }
 
   private shouldRaiseEvent(snap: ViewSnapshot): boolean {
-    // We don't need to handle includeDocumentMetadataChanges here because
```
schmidt-sebastian: Note: This is no longer true since we don't strip out the metadata-only changes beforehand (users could still ask for them even if the query-level metadata changes are turned off)

Furthermore, this code snippet uses a loop instead of find() for better IE support.

jshcrowthe: FWIW we currently shim `Array.prototype.find` for environments that don't support it (see https://github.com/firebase/firebase-js-sdk/blob/master/packages/polyfill/index.ts#L24-L25).

If it improves readability please don't hesitate to use it, however in this case, it's a freebie 😄

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
 # Unreleased
+- [breaking] `QuerySnapshot.docChanges()` is now a method. Please make sure
+  to audit your code for all existing usage of the old property `.docChanges`
+  and to change these to method invocations.
+- [changed] Replaced `QueryListenOptions` and `DocumentListenOptions` with
+  `SnapshotListenOptions`. `SnapshotListenOptions` only exposes the single
+  property `includeMetadataChanges` to turn on metadata changes.
```
mikelehen: Perhaps don't both mentioning these types since I doubt anybody is explicitly referencing them.  Instead reference the option names...  Something like:

Merged the `includeQueryMetadataChanges` and `includeDocumentMetadataChanges` options passed to `Query.onSnapshot()` into a single `includeMetadataChanges` option.

[I don't think we need to mention DocumentReference.onSnapshot() at all since the usage didn't change]

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
...
+- [changed] `QuerySnapshot.docChanges()` optionally takes
+  `SnapshotListenOptions`. By default, even when listening to a query with
```
mikelehen: Similarly, I'd say "'... optionally takes a `includeMetadataChanges` boolean option. By default, ..."

Also, can we either merge this with the top docChanges() change or at least put them next to each other?  Given your 3 entries right now, perhaps the optimal order would be 2,1,3.

packages/firestore/src/api/database.ts
```diff
@@ -922,8 +922,7 @@ export class DocumentReference implements firestore.DocumentReference {
     }
 
     const internalOptions = {
-      includeDocumentMetadataChanges: options.includeMetadataChanges,
-      includeQueryMetadataChanges: options.includeMetadataChanges
+      includeMetadataChanges: options.includeMetadataChanges
```
mikelehen: FWIW Gil / Rich didn't actually change the internal ListenOptions class... so TypeScript has diverged...  but simplifying it seems like an improvement and not a particularly confusing divergence, so it's probably fine for now.

packages/firestore/src/core/event_manager.ts
mikelehen: Hrm.  Thanks for flagging this...

This is another divergence from the Android / iOS implementations which still filter out the metadata-only changes, and I think filtering them out may make sense.  I don't think it's useful to call .docChanges({includeMetadataChanges: true}) unless you also specified {includeMetadatChanges: true} in onSnapshot()...  since you're not guaranteed to see every metadata-only change that way.  It would be kind of random.  We could even detect that and throw an exception ("If you pass includeMetadatChanges: true to docChanges() you must also pass it when calling onSnapshot()") if we wanted to be extra helpful.

WDYT?

packages/firestore/test/unit/core/event_manager.test.ts
```diff
@@ -320,91 +320,6 @@ describe('QueryListener', () => {
...
-        docs: snap2.docs,
-        oldDocs: snap1.docs,
-        docChanges: [change3],
-        fromCache: snap2.fromCache,
-        syncStateChanged: snap2.syncStateChanged,
-        hasPendingWrites: snap2.hasPendingWrites
-      };
-      expect(filteredEvents).to.deep.equal([snap1, expectedSnap2]);
-    }
-  );
```
mikelehen: Android / iOS kept these tests...

packages/firestore-types/index.d.ts
```diff
@@ -1098,6 +1074,17 @@ export class QuerySnapshot {
   /** True if there are no documents in the QuerySnapshot. */
   readonly empty: boolean;
 
+  /**
+   * An array of the documents that changed since the last snapshot. If this
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore-types/index.d.ts
```diff
@@ -1098,6 +1074,17 @@ export class QuerySnapshot {
...
+   * is the first snapshot, all documents will be in the list as added
+   * changes.
+   *
+   * @paran options `SnapshotListenOptions` that control whether metadata-only
+   * changes (i.e. only `DocumentSnapshot.metadata` changed) should trigger
+   * snapshot events.
+   */
+  docChanges(options?: SnapshotListenOptions): DocumentChange[];
+
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
 # Unreleased
+- [breaking] `QuerySnapshot.docChanges()` is now a method. Please make sure
```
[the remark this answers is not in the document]
schmidt-sebastian: `Breaking` is not really a thing, but it seemed worth pointing this out explicitly (granted, I don't know if the JS tooling would have liked this tag).

With the additional error message that I added, it seems fine to include this breaking change as part of the list of "changes".

packages/firestore/CHANGELOG.md
schmidt-sebastian: Sounds good! I like this much better.

packages/firestore/CHANGELOG.md
schmidt-sebastian: I merged this into two bullet points and moved the `docChanges` one further down. I wanted to emphasize the breaking change in the original changelog entry.

packages/firestore/src/api/database.ts
```diff
@@ -1817,13 +1808,35 @@ export class QuerySnapshot implements firestore.QuerySnapshot {
     return new Query(this._originalQuery, this._firestore);
   }
 
-  get docChanges(): firestore.DocumentChange[] {
-    if (!this._cachedChanges) {
+  docChanges(
+    options?: firestore.SnapshotListenOptions
+  ): firestore.DocumentChange[] {
```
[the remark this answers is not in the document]
schmidt-sebastian: I tried to make this work earlier but quickly ran into some issues:
- Symbol.iterator cannot actually be redefined.
- I didn't want to override .length since someone could actually be using it with the new docChanges() method. Granted, this is probably very unlikely.

I played around with this some more though and noticed that when someone writes `querySnap.docChanges`, it actually executes the method but without any `this` context. Hence, I have added an error check in the method that checks for `this === undefined`.

packages/firestore/src/api/database.ts
schmidt-sebastian: Unless we plan need to differentiate between these states, I would also argue that this is an improvement (a slightly biased argument I might say).

packages/firestore/src/core/event_manager.ts
schmidt-sebastian: I was kinda torn between this and throwing an error. I don't think we ever discussed this before and I noticed that this wasn't really addressed in the original PR. I think silently ignoring this option seems like an oversight, and I changed it to throw an error for now. The only benefit of supporting it the way I did originally was to try to do what the user asked for, but it does indeed create some rather strange semantics.

packages/firestore/test/unit/core/event_manager.test.ts
schmidt-sebastian: These tests require the internal distinction between document and query metadata changes. The only case that remains is tested in `raises metadata events only when specified`.

packages/firestore/src/api/database.ts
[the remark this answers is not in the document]
schmidt-sebastian: BTW, I just pushed a commit that added the changes you suggested. In my original attempt, I was calling the `defineProperty` code multiple times, which ended up breaking. The `this === undefined` seems to work in chai (`expect(snap.docChanges)` calls the method without context), but otherwise you are 100% correct (which is much more than 74%!).

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -200,11 +206,16 @@ apiDescribe('Queries', persistence => {
...
-        .then(querySnap => {
+        .then(async querySnap => {
           expect(toDataArray(querySnap)).to.deep.equal([
             { v: 'a' },
             { v: 'b' }
           ]);
+          if (querySnap.metadata.fromCache) {
+            // We might receive an additional event if the first query snapshot
+            // was served from cache.
+            await storeEventFull.awaitEvent();
```
schmidt-sebastian: I added this today as this test is pretty flaky when run as part of a larger test suite. It only ever fails with persistence enabled and seems to depend on whether listener 1 successfully received CURRENT before listener 2 is added.

packages/firestore/CHANGELOG.md
```diff
@@ -1,14 +1,11 @@
...
-  property `includeMetadataChanges` to turn on metadata changes.
-- [changed] `QuerySnapshot.docChanges()` optionally takes
-  `SnapshotListenOptions`. By default, even when listening to a query with
-  `{ includeMetadataChanges:true }`, metadata-only document changes
-  are suppressed in `docChanges()`. 
+- [changed] Merged the `includeQueryMetadataChanges` and 
+  `includeDocumentMetadataChanges` options passed to `Query.onSnapshot()` into 
+  a single `includeMetadataChanges` option.
+- [changed] `QuerySnapshot.docChanges()` is now a method that optionally takes
+  a `includeMetadataChanges` option. By default, even when listening to a query
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/CHANGELOG.md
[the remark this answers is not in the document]
jshcrowthe: I agree that it'd be nice to have the breaking tag. Maybe @samtstern has an idea bout how we can make that happen?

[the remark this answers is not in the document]
samtstern: You mean like a new pill style for "breaking"?  I think I could get that done, but CSS changes on devsite are kind of slow to deploy (they're part of the binary deploy and not something we can publish on-demand) so it depends on the timeline.

File a bug against me if you want me to look into it, detailing the syntax you want and when it should be ready.

PR #699, opened 2018-04-18
Remove unused Rollup plugins

mmermerkaya:
Removes `rollup-plugin-commonjs` and `rollup-plugin-node-resolve` from all packages except the main Firebase package.

These plugins are only needed to add dependencies from `node_modules` to bundles, but since only the firebase package creates a bundle that includes dependencies (no `external` in config), they are not needed in the other packages.

Changes the generation of `external` field in Rollup configs. This change prevents inclusion of submodule imports (like `import 'core-js/fn/array/find'`) in the bundle as non-external imports. This was not noticed before because they were just bundled with the package and increased the file size and build time but did not cause errors.


gauntface:
@mmermerkaya did you run this by @jshcrowthe before raising the PR?

mmermerkaya:
No. He's in the list of reviewers.

mmermerkaya:
I'm not sure what you mean by "Not having the `commonjs` and `node-resolve` plugins prevents this."

Currently these plugins are being used to bundle dependencies that should not be bundled (because these bundles are being served on NPM, and NPM already installs the dependencies for you separately).

If you'd like to make a browser (umd) bundle from each package as well, we can add the plugins only to that bundle's config. [Here](https://github.com/rollup/rollup-starter-lib/blob/master/rollup.config.js) is an example.

If you don't want me to remove the plugin devDependencies from `package.json`s, I can leave them in, but it doesn't make sense to me to keep dependencies that we don't use yet. We can always add them back in when we do need them.

mmermerkaya:
Ping

jshcrowthe:
Again, I'm **against** the dependencies change here. I'll try to clarify my reasoning.

If we remove these two plugins two things happen:

- Packages w/o them no longer understand named module resolutions
- Packages w/o them no longer understand CommonJS modules

Having the infrastructure in place for this is critical for components to adopt any external module (as it is not overly common to ship ES Modules by default). Having to add the infra to add external modules is a no-go.

The sub-module imports I am okay with though and we should definitely do that.

mmermerkaya:
I'm sorry but I still don't understand why we need these plugins for our module bundles. Removal of these plugins do not prevent the packages from having external modules as dependencies.

What is currently happening:
- Our CommonJS and ES Module bundles don't include "externals" and thus they don't need to know about Node module resolution or CommonJS modules. They get installed by users using NPM/Yarn and the users use module bundlers to bundle our modules and their dependencies for their browser builds. They might use Rollup, Webpack, etc.
- Our [browser bundle](https://github.com/firebase/firebase-js-sdk/blob/8922c3aeaf74493e1aefc4edd54e85606442479e/packages/firebase/rollup.config.js#L176) (which I didn't touch at all) takes our module bundle and adds the dependencies to create an `iife` build. This bundle includes the dependencies, which as you say might be using CommonJS. That is why we keep the plugins for this build.

This change will remove the plugins only from the CJS and ES bundles. These bundles still have the import statements for the external dependencies and the dependencies can be bundled by the users (who are the ones that need to add these plugins). And since they are not bundled, we don't need the two plugins I removed.

Is there anything wrong with this description?

dwoffinden:
Well that's... Insane :P

mmermerkaya:
Thanks for the review :smile: :tada: 

Looks like I still can't submit this, even though you are the code owner of the whole repo. Do you know why? :confused:

Feiyang1:
yea, each component's code owner has precedence over me. You would need to get them to approve as well. I can force merge it as an administrator, but that would be the last resort and I'd rather go through the proper process.

Feiyang1:
@schmidt-sebastian @mikelehen @bojeil-google @wti806 @ryanpbrewster friendly ping

bojeil-google:
This doesn't touch any auth component at all. It doesn't look like you need Auth approval.

Feiyang1:
I'm generally in favor of removing dependencies we don't use, even though we may need them in the future.

mmermerkaya:
Yeah, that was a long exchange, sorry :smile:

**TL;DR**: Those two plugins are only needed to bundle external dependencies and should not be used to bundle libraries (except browser builds, to use directly in `<script>` tags, but we do that only in the [main firebase package](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/rollup.config.js)). Rollup has an official example that shows this [here](https://github.com/rollup/rollup-starter-lib/blob/master/rollup.config.js).

alecmce:
I'm having a PTSD moment. I share Daniel's package sadness. Other than that, ;dr. Hope all's well, Mertcan.

PR #700, opened 2018-04-18
Check browser support before instantiating controllers

mmermerkaya:
Checks browser support before instantiating messaging service controllers and fixes some tests.

I think it's better to let the developer know that Messaging will not work as soon as possible, instead of instantiating a controller without an issue and then throwing an error as soon as they try to use it.

This also prevents issues such as #658 by making sure we don't call any unsupported methods by accident.

This can break users who were not checking browser support before calling `firebase.messaging()`, so it should go to the v5 branch.

### DIFF
--- a/integration/shared/validator.js
+++ b/integration/shared/validator.js
@@ -88,9 +88,16 @@ function validateNamespace(definition, candidate) {
       if (
         definitionChunk.__type === 'function' &&
         definitionChunk.__return &&
-        typeof candidateChunk === 'function' &&
-        candidateChunk()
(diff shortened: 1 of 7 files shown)


mmermerkaya:
I changed the base to `v5/master`, but that made @googlebot  a bit confused. All commits in this PR are mine.

mmermerkaya:
Permission change?

Sure, I can send it to API review.

mmermerkaya:
Went through API review, implemented requested changes.

PR #701, opened 2018-04-18
Adding mergeFields support to the Web SDK

schmidt-sebastian:
This PR adds field mask support to `.set(, {..})`. It's mostly a port of the missing features from https://critique.corp.google.com/#review/167190604 (internal)

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1,3 +1,5 @@
+import { FieldPath } from '../firestore-types/index';
+
 /**
  * Copyright 2017 Google Inc.
  *
@@ -1115,7 +1117,7 @@ declare namespace firebase.firestore {
    * An options object that configures the behavior of `set()` calls in
    * `DocumentReference`, `WriteBatch` and `Transaction`. These calls can be
    * configured to perform granular merges instead of overwriting the target
-   * documents in their entirety by providing a `SetOptions` with `merge: true`.
+   * documents in their entirety.
    */
   export interface SetOptions {
     /**
@@ -1124,6 +1126,16 @@ declare namespace firebase.firestore {
      * untouched.
      */
     readonly merge?: boolean;
+
+    /**
+     * Changes the behavior of set() calls to only replace the specified field
+     * paths. Any field path that is not specified is ignored and remains
+     * untouched.
+     *
+     * <p>It is an error to pass a SetOptions object to a set() call that is
+     * missing a value for any of the fields specified here.
+     */
+    readonly mergeFields?: (string | FieldPath)[];
   }
 
   /**
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1,5 +1,3 @@
-import { FieldPath } from '../firestore-types/index';
-
 /**
  * Copyright 2017 Google Inc.
  *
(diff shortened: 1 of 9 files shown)


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -297,19 +299,54 @@ export class UserDataConverter {
...
+
+    if (!fieldPaths) {
+      fieldMask = new FieldMask(context.fieldMask);
+    } else {
+      const validatedFieldPaths = [];
+
+      for (const stringOrFieldPath of fieldPaths) {
+        let fieldPath;
+
+        if (stringOrFieldPath instanceof ExternalFieldPath) {
```
schmidt-sebastian: Note that there are 3 different FieldPath in our API: The internal representation, the external interface (firestore.FieldPath) and the backing type for the external interface (ExternalFieldPath). By default, a firestore.FieldPath is not compatible with an ExternalFieldPath as it has properties missing.

schmidt-sebastian:
Note that Travis is failing because of a lint error in master (https://github.com/firebase/firebase-js-sdk/blob/master/packages/app-types/index.d.ts#L101)

packages/firebase/index.d.ts
```diff
@@ -1,3 +1,5 @@
+import { FieldPath } from '../firestore-types/index';
```
gsoltis: This should probably go below the copyright?

packages/firestore-types/index.d.ts
```diff
@@ -492,6 +492,16 @@ export interface SetOptions {
    * untouched.
    */
   readonly merge?: boolean;
+
+  /**
+   * Changes the behavior of set() calls to only replace the specified field
```
gsoltis: What if I pass `[]`?

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -297,19 +299,54 @@ export class UserDataConverter {
...
+          fail('Expected stringOrFieldPath to be a string or a FieldPath');
+        }
+
+        if (!updateData.contains(fieldPath)) {
+          throw new FirestoreError(
+            Code.INVALID_ARGUMENT,
+            `Field '${fieldPath}' is specified in your field mask but missing from your input data.`
+          );
+        }
+        validatedFieldPaths.push(fieldPath);
```
gsoltis: What about overlapping field paths?

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -207,6 +208,35 @@ apiDescribe('Database', persistence => {
     });
   });
 
+  it("can't overspecify field mask using set", () => {
```
gsoltis: Please include a test for `[]` and for overlapping field paths, i.e. `['foo', 'foo.bar']`.

packages/firebase/index.d.ts
schmidt-sebastian: This is actually not needed since FieldPath is defined in this very file.

packages/firestore-types/index.d.ts
schmidt-sebastian: This is allowed since `set({}, merge:true)` is also a valid use. We originally enforced that at least one field was updated, but got user reports asking for us to drop this validation.

packages/firestore/src/api/user_data_converter.ts
schmidt-sebastian: We discussed this during API review at one point and decided to "do whatever Android" does. Android doesn't validate and uses the most broad field path if multiple field paths match.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: I originally only ported the two tests that were missing from the original Android implementation. I have added more tests though, and while looking at the code noticed that the implementation didn't properly deal with field transforms. I am now also applying the mask to the transforms array. Let me know what you think and I will port it to Android.

PR #702, opened 2018-04-18
 Do nothing for action clicks

mmermerkaya:


packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -161,9 +161,13 @@ export class SWController extends ControllerInterface {
       return;
     }
 
+    if (event.action) {
+      // User clicked on an action button
```
pinarx: add a comment to indicate this will allow other listeners to pick up on this change. Can you also add an example somewhere to display how devs should be using this?

mmermerkaya: There's a comment about it just below this line that says "Prevent other listeners from receiving the event" before calling stopImmediatePropagation. I can put that call in an `else` block if you think that makes it clearer.

Where/how should I add the example? As a comment?

pinarx: Make it more explicit, saying this will allow devs to act on action button clicks by using a onNotificationClick listener that they define?

mmermerkaya: Done.

PR #706, opened 2018-04-19
Persisting mutation keys in memory

schmidt-sebastian:
This PR persists relevant mutation keys in memory. Relevant mutations are either mutations that are the mutations that are pending and being waited for by the specific clients. Primary clients wait for the mutations that they transmitted to the backend, and secondary clients wait for the mutations that originated from them.

This is arguable not very pretty, but I think local store is a better place for this than the mutation queue or shared client state. Let me know if you disagree :)

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -427,6 +427,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     const documents = await this.localStore.lookupMutationDocuments(batchId);
 
     if (documents === null) {
+      // A throttled tab may not have been able to cache the keys of affected
+      // documents before the mutation was removed from persistence. In this
+      // case, we cannot recompute our local views and instead ignore the
+      // update.
       log.debug(LOG_TAG, 'Cannot apply mutation batch with id: ' + batchId);
       return;
     }
@@ -442,7 +446,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       this.sharedClientState.removeLocalPendingMutation(batchId);
       this.processUserCallback(batchId, error ? error : null);
 
-      this.localStore.removeCachedMutationBatch(batchId);
+      this.localStore.removeCachedMutationBatchMetadata(batchId);
     } else {
       fail(`Unknown batchState: ${batchState}`);
     }
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -437,11 +437,12 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       // connection is disabled.
       await this.remoteStore.fillWritePipeline();
     } else if (batchState === 'acknowledged' || batchState === 'rejected') {
-      // NOTE: These methods are no-ops for batches that originated from other
-      // clients.
+      // NOTE: Both these methods are no-ops for batches that originated from
+      // other clients.
       this.sharedClientState.removeLocalPendingMutation(batchId);
-      this.localStore.removeMutationKeys(batchId);
       this.processUserCallback(batchId, error ? error : null);
+
+      this.localStore.removeCachedMutationBatch(batchId);
     } else {
       fail(`Unknown batchState: ${batchState}`);
     }
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -424,11 +424,12 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     error?: FirestoreError
   ): Promise<void> {
     this.assertSubscribed('applyBatchState()');
-    const mutationBatchResult = await this.localStore.lookupLocalWrite(batchId);
-    assert(
-      mutationBatchResult !== null,
-      'Unable to find mutation batch: ' + batchId
-    );
+    const documents = await this.localStore.lookupMutationDocuments(batchId);
+
+    if (documents === null) {
+      log.debug(LOG_TAG, 'Cannot apply mutation batch with id: ' + batchId);
+      return;
+    }
 
     if (batchState === 'pending') {
       // If we are the primary client, we need to send this write to the
@@ -436,15 +437,16 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       // connection is disabled.
       await this.remoteStore.fillWritePipeline();
     } else if (batchState === 'acknowledged' || batchState === 'rejected') {
-      // NOTE: Both these methods are no-ops for batches that originated from
-      // other clients.
+      // NOTE: These methods are no-ops for batches that originated from other
+      // clients.
       this.sharedClientState.removeLocalPendingMutation(batchId);
+      this.localStore.removeMutationKeys(batchId);
       this.processUserCallback(batchId, error ? error : null);
     } else {
       fail(`Unknown batchState: ${batchState}`);
     }
 
-    await this.emitNewSnapsAndNotifyLocalStore(mutationBatchResult.changes);
+    await this.emitNewSnapsAndNotifyLocalStore(documents);
   }
 
   applySuccessfulWrite(
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -427,10 +427,13 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     const documents = await this.localStore.lookupMutationDocuments(batchId);
 
     if (documents === null) {
-      // A throttled tab may not have been able to cache the keys of affected
-      // documents before the mutation was removed from persistence. In this
-      // case, we cannot recompute our local views and instead ignore the
-      // update.
+      // A throttled tab may not have seen the mutation before it was completed
+      // and removed from the mutation queue, in which case we won't have cached
+      // the affected documents. In this case we can safely ignore the update
+      // since that means we didn't apply the mutation locally at all (if we
+      // had, we would have cached the affected documents), and so we will just
+      // see any resulting document changes via normal remote document updates
+      // as applicable.
       log.debug(LOG_TAG, 'Cannot apply mutation batch with id: ' + batchId);
       return;
     }
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -41,7 +41,7 @@ import { MutationQueue } from './mutation_queue';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { SimpleDbStore, SimpleDbTransaction } from './simple_db';
-import { documentKeySet, DocumentKeySet } from '../model/collections';
+import { DocumentKeySet } from '../model/collections';
 
 /** A mutation queue for a specific user, backed by IndexedDB. */
 export class IndexedDbMutationQueue implements MutationQueue {
@@ -232,14 +232,9 @@ export class IndexedDbMutationQueue implements MutationQueue {
     const batchId = this.nextBatchId;
     this.nextBatchId++;
     const batch = new MutationBatch(batchId, localWriteTime, mutations);
-
     const dbBatch = this.serializer.toDbMutationBatch(this.userId, batch);
 
-    let keys = documentKeySet();
-    mutations.forEach(mutation => {
-      keys = keys.add(mutation.key);
-    });
-    this.documentKeysByBatchId[dbBatch.batchId] = keys;
+    this.documentKeysByBatchId[dbBatch.batchId] = batch.keys();
 
     return mutationsStore(transaction)
       .put(dbBatch)
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -66,8 +66,8 @@ export class IndexedDbMutationQueue implements MutationQueue {
    * locally, `removeCachedMutationKeys()` should be invoked either directly
    * or through `removeMutationBatches()`.
    *
-   * After the primary client acknowledged or rejected a mutation during
-   * multi-tab, this cache is used by secondary clients to invalidate the local
+   * With multi-tab, when the primary client acknowledges or rejects a mutation,
+   * this cache is used by secondary clients to invalidate the local
    * view of the documents that were previously affected by the mutation.
    */
   // PORTING NOTE: Multi-tab only.
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -41,6 +41,7 @@ import { MutationQueue } from './mutation_queue';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { SimpleDbStore, SimpleDbTransaction } from './simple_db';
+import { documentKeySet, DocumentKeySet } from '../model/collections';
 
 /** A mutation queue for a specific user, backed by IndexedDB. */
 export class IndexedDbMutationQueue implements MutationQueue {
@@ -57,6 +58,18 @@ export class IndexedDbMutationQueue implements MutationQueue {
    * A write-through cache copy of the metadata describing the current queue.
    */
   private metadata: DbMutationQueue;
+
+  /**
+   * Caches the document keys for pending mutation batches. If the mutation
+   * has been removed from IndexedDb, the cached value may continue to
+   * be used to retrieve the batch's document keys. To remove a cached value
+   * locally, `removeCachedMutationKeys()` should be invoked.
+   *
+   * During multi-tab, this cache is used by secondary clients to replay
+   * mutations after they were acknowledged or rejected by the primary client.
+   */
+  // PORTING NOTE: Multi-tab only.
+  private documentKeyCache = {} as { [batchId: number]: DocumentKeySet };
 
   private garbageCollector: GarbageCollector | null = null;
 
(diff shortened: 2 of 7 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -797,6 +809,11 @@ export class LocalStore {
     });
   }
 
+  // PORTING NOTE: Multi-tab only.
```
schmidt-sebastian: This was just moved up since it's public.

packages/firestore/src/local/local_store.ts
```diff
@@ -153,6 +153,15 @@ export class LocalStore {
   /** Used to generate targetIDs for queries tracked locally. */
   private targetIdGenerator = TargetIdGenerator.forLocalStore();
 
+  /**
+   * Stores the document keys for local pending mutation batches. While
+   * primary clients will store all keys for all pending mutations, secondary
+   * clients will only store the keys for mutations that originated in these
+   * clients.
```
mikelehen: It seems like all clients need to store all mutation keys, since secondary clients will still have applied the mutations from other clients locally (latency compensation) and need to revert to the remoteDocuments version when the write is no longer pending (i.e. stop latency compensating the value).

Else listeners would get stuck with documents with hasPendingWrites=true for other clients' writes, etc., right?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -424,27 +424,29 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
     if (batchState === 'pending') {
       // If we are the primary client, we need to send this write to the
       // backend. Secondary clients will ignore these writes since their remote
       // connection is disabled.
       await this.remoteStore.fillWritePipeline();
     } else if (batchState === 'acknowledged' || batchState === 'rejected') {
-      // NOTE: Both these methods are no-ops for batches that originated from
-      // other clients.
+      // NOTE: These methods are no-ops for batches that originated from other
+      // clients.
```
mikelehen: rework this comment if you end up tracking mutation keys for other clients...

packages/firestore/src/local/local_store.ts
```diff
@@ -918,7 +936,39 @@ export class LocalStore {
...
+  ): PersistencePromise<DocumentKeySet | null> {
+    if (this.localMutationBatchKeys[batchId]) {
+      return PersistencePromise.resolve(this.localMutationBatchKeys[batchId]);
+    } else {
+      return this.mutationQueue
+        .lookupMutationBatch(txn, batchId)
+        .next(batch => {
+          if (batch) {
+            const keys = batch.keys();
+            this.insertMutationKeys(batchId, keys);
```
mikelehen: FWIW, I think this code would be more readable if you got rid of the insertMutationKeys() method. That way it's clear that you're reading (on line 958) and writing to (here) the same data (`this.localMutationBatchKeys`)

packages/firestore/src/local/local_store.ts
```diff
@@ -918,7 +936,39 @@ export class LocalStore {
...
+    this.localMutationBatchKeys[batchId] = keys;
+  }
+
+  /**
+   * Looks up mutation keys for the given batch ID. If the mutation is not
+   * known locally (e.g. when another client added it), this method looks up the
+   * mutation from persistence.
+   */
+  // PORTING NOTE: Multi-tab only.
+  private lookupAndStoreMutationKeys(
```
mikelehen: lookupAndStore seems odd to me. :)  Perhaps lookupAndCache...?

packages/firestore/src/remote/serializer.ts
```diff
@@ -1090,8 +1090,8 @@ export class JsonProtoSerializer {
     } else if (filter.fieldFilter !== undefined) {
       return [this.fromRelationFilter(filter)];
     } else if (filter.compositeFilter !== undefined) {
-      return filter.compositeFilter.filters!
-        .map(f => this.fromFilter(f))
+      return filter.compositeFilter
+        .filters!.map(f => this.fromFilter(f))
```
mikelehen: I'm not sure why this code got touched?  A prettier fail somewhere?

tools/config.js
```diff
@@ -63,7 +63,9 @@ Promise.resolve(userToken || cachedToken)
         }
       ]);
 
-      const { projectId: { id } } = response;
+      const {
+        projectId: { id }
+      } = response;
```
mikelehen: ditto...  Can you make sure you're running the right prettier? (run yarn from the repo root)

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I think this will all work automatically... but please correct me if I am wrong. I don't think there are any changes that need to be relayed across tabs that are not either some client's own pending mutation or part of a watch target. If a document is part of a latency compensated view, then it should also raise a Watch event in the primary tab. The primary tab will then relay these changes with the Watch change log.

Am I missing something?

mikelehen: I think it's possible there's no watch update (the write didn't actually change anything) but the most obvious case is probably rejected mutations.  There definitely won't be a watch update, but all clients definitely need to stop latency-compensating the document and refresh from remoteDocuments cache.

schmidt-sebastian: Wouldn't that change the metadata on the document? I need to relay both the updates from the server as well as from local mutations for all Watch targets.

schmidt-sebastian: To clarify why I am a little bit concerned about this change: There is no guarantee that a client will see every pending mutation. For throttled clients, the mutation might already have been acknowledged and removed from the queue by the time it was picked up from Local Storage.

mikelehen: Maybe I don't understand this watch changelog thing...  I assumed it would only reflect what we receive from watch for active targets.  It would be unaware of mutations and the local view of documents (the contents of documents with local mutations applied).

In the case of a rejected mutation (that never shows up from watch), what do you expect to go on the changelog?  Keep in mind that in order for the primary tab to even know that targets could be affected by the mutation it would need to be evaluating those queries locally, which normally it wouldn't do unless the primary tab also had an active listener (with a View, etc.) for the query.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I moved this line below and added a space to indicate that the comment doesn't apply.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I moved the cache to the mutation queue for now. It definitely doesn't hurt to have it there, and the memory mutation queue's implementation is actually super straightforward.

I am still not 100% convinced that the original design doesn't work, but this is more of a matter of optimization and I much rather just look at that later. You are correct though that the primary tab won't have an active listener - and that might make it infeasible to compute the latency compensated views.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: No longer applicable, but I did rename it from `lookupAndCache` to `lookupAndStore`earlier since `cache` to me hinted to strongly that I might have implemented some cleanup magic.

packages/firestore/src/remote/serializer.ts
schmidt-sebastian: I think these problems show up for me when I ran yarn from master versus yarn from firestore-multi-tab. I re-ran it from firestore-multi-tab for this PR.

tools/config.js
schmidt-sebastian: Done

packages/firestore/src/local/local_store.ts
schmidt-sebastian: No longer applicable. I mostly did this so it was apparent that the "PORTING NOTES" applied.

packages/firestore/src/local/local_store.ts
```diff
@@ -332,26 +332,24 @@ export class LocalStore {
...
-                return { batchId: batch.batchId, changes: changedDocuments };
-              });
-          } else {
-            return PersistencePromise.resolve(null);
-          }
-        });
-    });
+  /** Returns the local view of the documents affected by a mutation batch. */
+  // PORTING NOTE: Multi-tab only.
+  lookupMutationDocuments(batchId: BatchId): Promise<MaybeDocumentMap | null> {
```
mikelehen: This method seems a bit odd to me. The name doesn't suggest to me that it looks up the local view of the documents (rather than just returning the list of documents in the mutation or something).  I sort of want to rename it to lookupMutatedDocuments() but that's not right because sometimes it is run against documents that have a mutation and sometimes it's run against documents that previously had a mutation.  So I can't come up with a great name...

It's also a bit odd that it may or may not succeed (it can return null if it fails to get the list of document keys).

Part of me is tempted to just inline the method where it's used since it's only used once and it's difficult to name / describe it well.  But I don't feel strongly that that's an improvement either.  So feel free to leave it as-is unless my mild griping has triggered some inspiration on your side. :)

packages/firestore/src/local/memory_mutation_queue.ts
```diff
@@ -173,6 +174,16 @@ export class MemoryMutationQueue implements MutationQueue {
     return PersistencePromise.resolve(this.findMutationBatch(batchId));
   }
 
+  lookupMutationKeys(
+    transaction: PersistenceTransaction,
+    batchId: BatchId
+  ): PersistencePromise<DocumentKeySet | null> {
+    const mutationBatch = this.findMutationBatch(batchId);
```
mikelehen: Should we assert that the mutationBatch is found since the memory mutation queue is only used in non-multi-tab and so it would be a bug to look up a mutation's keys after it's removed?

packages/firestore/src/local/mutation_queue.ts
```diff
@@ -102,6 +103,17 @@ export interface MutationQueue extends GarbageSource {
     batchId: BatchId
   ): PersistencePromise<MutationBatch | null>;
 
+  /**
+   * Returns the document keys for the mutation batch with the given batchId.
+   * This method returns `null` if the mutation has been removed.
```
mikelehen: This isn't exactly accurate right? Since the cache intentionally persists after the mutation has been removed...  Perhaps something like:

This method returns `null` after the primary client has called removeMutationBatches() or a secondary client has called `removeCachedMutationKeys()`

packages/firestore/src/local/mutation_queue.ts
```diff
@@ -194,6 +206,17 @@ export interface MutationQueue extends GarbageSource {
     batches: MutationBatch[]
   ): PersistencePromise<void>;
 
+  /**
+   * Clears the cached keys for a mutation batch. This method should be
+   * called by secondary clients after they process mutation updates.
+   *
+   * Note that this method does not have to be called from primary clients as
+   * `acknowledgeBatch()` and `removeMutationBatches()` clear these cache
```
mikelehen: I didn't see any code in acknowledgeBatch() that clears the cache...

packages/firestore/src/core/sync_engine.ts
```diff
@@ -424,11 +424,12 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
     this.assertSubscribed('applyBatchState()');
-    const mutationBatchResult = await this.localStore.lookupLocalWrite(batchId);
-    assert(
-      mutationBatchResult !== null,
-      'Unable to find mutation batch: ' + batchId
-    );
+    const documents = await this.localStore.lookupMutationDocuments(batchId);
+
+    if (documents === null) {
+      log.debug(LOG_TAG, 'Cannot apply mutation batch with id: ' + batchId);
```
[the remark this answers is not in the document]
schmidt-sebastian: I added a comment explaining when this might happen.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -440,11 +441,13 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       // other clients.
       this.sharedClientState.removeLocalPendingMutation(batchId);
       this.processUserCallback(batchId, error ? error : null);
+
+      this.localStore.removeCachedMutationBatch(batchId);
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -58,6 +59,18 @@ export class IndexedDbMutationQueue implements MutationQueue {
    */
   private metadata: DbMutationQueue;
 
+  /**
+   * Caches the document keys for pending mutation batches. If the mutation
+   * has been removed from IndexedDb, the cached value may continue to
+   * be used to retrieve the batch's document keys. To remove a cached value
+   * locally, `removeCachedMutationKeys()` should be invoked.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (reworded)

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -58,6 +59,18 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+   *
+   * During multi-tab, this cache is used by secondary clients to replay
+   * mutations after they were acknowledged or rejected by the primary client.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (reworded)

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -220,6 +233,12 @@ export class IndexedDbMutationQueue implements MutationQueue {
 
     const dbBatch = this.serializer.toDbMutationBatch(this.userId, batch);
 
+    let keys = documentKeySet();
+    mutations.forEach(mutation => {
+      keys = keys.add(mutation.key);
+    });
```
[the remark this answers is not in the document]
schmidt-sebastian: I am! I decided to showcase my engineering skills here.

(This is a Mutation[] and not a MutationBatch).

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I can't really come up with a name that alleviates all of these concerns. I think `lookupMutatedDocuments` might be slightly better, but I am also not 100% convinced. For now, i decided to keep it as is.

I would also rather not inline it as that would move the transaction logic into SyncEngine.

packages/firestore/src/local/memory_mutation_queue.ts
schmidt-sebastian: Yes, that makes sense. Albeit, this method should never be called in the first place.

I added a check for the tombstone-status instead.

packages/firestore/src/local/mutation_queue.ts
schmidt-sebastian: Reworded to:

```
Returns the document keys for the mutation batch with the given batchId.
For primary clients, this method returns `null` after
`removeMutationBatches()` has been called. Secondary clients return a
cached result until `removeCachedMutationKeys()` is invoked.
```

packages/firestore/src/local/mutation_queue.ts
schmidt-sebastian: It's done through `removeMutationBatches()`. I reworded this comment as well.

packages/firestore/src/local/indexeddb_mutation_queue.ts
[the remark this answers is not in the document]
mikelehen: What is `batch` though?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -427,6 +427,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     const documents = await this.localStore.lookupMutationDocuments(batchId);
 
     if (documents === null) {
+      // A throttled tab may not have been able to cache the keys of affected
+      // documents before the mutation was removed from persistence. In this
+      // case, we cannot recompute our local views and instead ignore the
+      // update.
```
mikelehen: Thanks for the updated comment. In trying to figure out what "ignore the update" means in practice, I've kinda' convinced me this is 100% safe / desirable behavior.  Can you confirm and if so, update the comment something like:

A throttled tab may not have seen the mutation before it was completed and removed from the mutation queue, in which case we won't have cached the affected documents.  In this case we can safely ignore the update since that means we didn't apply the mutation locally at all (if we had, we would have cached the affected documents), and so we will just see any resulting document changes via normal remote document updates as applicable.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -63,13 +63,15 @@ export class IndexedDbMutationQueue implements MutationQueue {
    * Caches the document keys for pending mutation batches. If the mutation
    * has been removed from IndexedDb, the cached value may continue to
    * be used to retrieve the batch's document keys. To remove a cached value
-   * locally, `removeCachedMutationKeys()` should be invoked.
+   * locally, `removeCachedMutationKeys()` should be invoked either directly
+   * or through `removeMutationBatches()`.
    *
-   * During multi-tab, this cache is used by secondary clients to replay
-   * mutations after they were acknowledged or rejected by the primary client.
+   * After the primary client acknowledged or rejected a mutation during
```
mikelehen: acknowledged => acknowledges, rejected => rejects

Perhaps instead of "during multi-tab" just preface the sentence with "With multi-tab, "

packages/firestore/src/local/indexeddb_mutation_queue.ts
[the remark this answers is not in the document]
schmidt-sebastian: It's a `DbMutationBatch`.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_mutation_queue.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_mutation_queue.ts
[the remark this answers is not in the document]
mikelehen: Are you sure?

PR #707, opened 2018-04-19
Adding multi-client support for rejected mutations

schmidt-sebastian:
This PR adds the missing elements to reject mutations for multi-client writes.


schmidt-sebastian:
This is a pretty tiny PR so sending this one over already.

PR #708, opened 2018-04-20
fix(@firebase/auth): export types from auth-types

Maistho:
Exports the types from `@firebase/auth-types` in `@firebase/auth`

This allows to import types from `@firebase/auth` when using typescript.

### DIFF
--- a/packages/auth/index.d.ts
+++ b/packages/auth/index.d.ts
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
-import "@firebase/auth-types";
+export * from "@firebase/auth-types";


jshcrowthe:
I see what you're doing here, but I don't think I agree with this change.

The main concern I see, is in reexporting these types, people will assume that they can use the types as the implementation (when in reality that is not the API exported by the package).

For example, in doing what you're suggesting I would have access to the `GoogleAuthProvider` class in my typings. I could write code like the following:

```typescript
import { GoogleAuthProvider } from '@firebase/auth';

const provider: GoogleAuthProvider = new firebase.auth.GoogleAuthProvider();
```

However many would be tempted to do:

```typescript
import { GoogleAuthProvider } from '@firebase/auth';

const provider = new GoogleAuthProvider();
```

The latter is not the exported API, would throw errors, and ultimately would result in more confusion than benefit. You are free to use the `@firebase/auth-types` package as you suggested and that will allow you to annotate your code as I illustrated in the first example. It still will not contain the actual implementation of a given interface/class and you would need to be explicit about that lest there is more confusion.

jshcrowthe:
Yeah, we are working on making the story better there. Appreciate your patience 😄.

The other modules only have a single function exported and that is purely so we can attach each module to `@firebase/app`. 

As to the question of "why." The Firebase API creates the `firebase` namespace and everything is generally attached to that for ease of use. We are doing some investigation here, but nothing to report yet. Stay tuned tho, and appreciate the feedback!

Issue #703, opened 2018-04-18
Expo and Firestore security rules


kusaeva:
* Operating System version: Android/iOS
  * Firebase SDK version: 4.12.1
  * Firebase Product: firestore

I’m using expo and the firebase web sdk (cause native firebase sdk is not working without detaching expo project).
When I decided to set up firestore security rules like this:
```
service cloud.firestore {
   match /databases/{database}/documents {
     match /{document=**} {
       allow read, write: if request.auth != null;
     }
   }
 }
```

suddenly I got permission-denied. It appears both on Android/iOS
<img width="260" alt="screen shot 2018-04-18 at 21 14 18" src="https://user-images.githubusercontent.com/5486933/38950314-b73072d0-434d-11e8-9b88-6ee665d930d3.png">

User is 100% authorized and I am not the only one who experiences this problem. I found it at StackOverflow, Firebase Google Group, Firebase Slack channel, Expo forums but it seems that no one creates the issue here.
The only solution was suggested at [stackoverflow](https://stackoverflow.com/questions/48054884/firebase-firestore-missing-or-insufficient-permissions-using-expo-react-native) - downgrading firebase sdk to 4.6.2. 
But I really don't like it, it has a problem with firestore on Android and need to be patched.

I understand that using firebase web sdk for Android/iOS is not the case but is there any chance that the issue will be fixed?

mikelehen:
So just to confirm, this happens with 4.12.1 but not 4.6.2?  If you can verify exactly what release the regression occurred in, that will help us track it down.

And can you capture logs by calling `firebase.firestore.setLogLevel('debug')` before calling any other firestore operations?

kusaeva:
I can confirm that security rules work with 4.6.2 and stop working with 4.7.0

```
17:49:18 [exp] Firestore (4.10.1) 2018-04-19T14:49:18.616Z [MemoryPersistence]: Starting transaction: Start LocalStore
17:49:18 [exp] Firestore (4.10.1) 2018-04-19T14:49:18.619Z [MemoryPersistence]: Starting transaction: Get last stream token
17:49:18 [exp] Firestore (4.10.1) 2018-04-19T14:49:18.621Z [MemoryPersistence]: Starting transaction: Get next mutation batch
17:49:18 [exp] Firestore (4.10.1) 2018-04-19T14:49:18.628Z [MemoryPersistence]: Starting transaction: Allocate query
[...]

PR #717, opened 2018-04-23
Disable httpHeadersOverwriteParam for ReactNative to address #703.

mikelehen:


packages/firestore/src/util/misc.ts
```diff
@@ -121,3 +121,12 @@ export function immediateSuccessor(s: string): string {
   // Return the input string, with an additional NUL byte appended.
   return s + '\0';
 }
+
+/**
+ * Detects whether we are running in a React Native app.
+ */
+export function isReactNative(): boolean {
```
wilhuff: Could we put this in a platform-detection module?

Maybe even combined with our browser version detection stuff?

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -215,6 +212,19 @@ export class WebChannelConnection implements Connection {
       sendRawJson: true,
       supportsCrossDomainXhr: true
     };
+
+    // Send our custom headers as a '$httpHeaders=' url param to avoid CORS
+    // preflight round-trip. This is formally defined here:
+    // https://github.com/google/closure-library/blob/b0e1815b13fb92a46d7c9b3c30de5d6a396a3245/closure/goog/net/rpc/httpcors.js#L40
+    //
+    // For some unclear reason (see
```
wilhuff: At a high level I understand what this comment is saying but I don't understand mechanically how this is actually working.

When we specify `httpHeadersOverwriteParam: '$httpHeaders'`, which headers are getting tucked into that variable? If we omit setting that, where do they get set instead?

PR #722, opened 2018-04-24
Port of https://github.com/firebase/firebase-ios-sdk/pull/1114

gsoltis:


packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -475,7 +507,8 @@ function genericLocalStoreTests(
       .afterAcknowledgingMutation({ documentVersion: 1 })
       .toReturnRemoved('foo/bar')
       .toNotContain('foo/bar')
-      .after(docUpdateRemoteEvent(doc('foo/bar', 1, { it: 'base' }), [1]))
+      .afterAllocatingQuery(Query.atPath(path('foo')))
+      .after(docUpdateRemoteEvent(doc('foo/bar', 1, { it: 'base' }), [2]))
```
schmidt-sebastian: Here and everywhere: Should we use this.lastTargetId instead of hardcoding targetId 2?

PR #724, opened 2018-04-25
Added TypeScript definitions for firebase.functions()

### LINKED ISSUE #705: Add TypeScript definitions for firebase.functions()
* Operating System version: All
  * Firebase SDK version: 4.12.1
  * Firebase Product: functions

In this version there are no TypeScript definitions for functions. Resulting in a ts error of:

`[ts] Property 'functions' does not exist on type 'typeof firebase'.`

mpoehler:
Fixes #705, "Add TypeScript definitions for firebase.functions()", also described here: https://stackoverflow.com/questions/49669360/using-firebase-functions-from-web-sdk-using-typescript

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -112,7 +112,9 @@ declare namespace firebase {
 
   function firestore(app?: firebase.app.App): firebase.firestore.Firestore;
 
-  function functions(app?: firebase.app.App): firebase.functions.FirebaseFunctions;
+  function functions(
+    app?: firebase.app.App
+  ): firebase.functions.FirebaseFunctions;
 }
 
 declare namespace firebase.app {
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1972,7 +1972,7 @@ declare namespace firebase.firestore {
    * - 'unauthenticated': The request does not have valid authentication
    *   credentials for the operation.
    */
-  export type ErrorStatus =
+  export type FirestoreErrorCode =
     | 'cancelled'
     | 'unknown'
     | 'invalid-argument'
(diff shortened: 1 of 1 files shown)


packages/firebase/index.d.ts
```diff
@@ -123,6 +127,42 @@ declare namespace firebase.app {
...
+    readonly data: any;
+  }
+  export interface HttpsCallable {
+    (data?: any): Promise<HttpsCallableResult>;
+  }
+  export class FirebaseFunctions {
+    private constructor();
+    httpsCallable(name: string): HttpsCallable;
+  }
+  export type FunctionsErrorCode =
```
bklimt: This should be named `ErrorStatus`.

mpoehler:
In addition to the changes mentioned, I renamed the FirestoreErrorCodes to ErrorStatus in the firestore namespace to be consistent.

jshcrowthe:
If we can roll back the `FirestoreErrorCodes` change I think this is good to go. Changing the name of an exported type (and not maintaining the previous name) is a breaking change, and I'd like to avoid that in this PR.

bklimt:
Sorry, my mistake. I agree with jshcrowthe@ that we renaming it isn't worth making a breaking change.

packages/firebase/index.d.ts
```diff
@@ -1934,7 +1972,7 @@ declare namespace firebase.firestore {
    * - 'unauthenticated': The request does not have valid authentication
    *   credentials for the operation.
    */
-  export type FirestoreErrorCode =
+  export type ErrorStatus =
```
jshcrowthe: Let's keep this change scoped to just the Functions SDK for now.

mpoehler:
@jshcrowthe I returned back to FirestoreErrorStatus.

jshcrowthe:
Awesome! Thanks for your help @mpoehler

PR #727, opened 2018-04-25
Port https://github.com/firebase/firebase-ios-sdk/pull/1122

gsoltis:


packages/firestore/src/core/view.ts
```diff
@@ -91,11 +91,15 @@ export class View {
   constructor(
     private query: Query,
     /** Documents included in the remote target */
-    private syncedDocuments: DocumentKeySet
+    private _syncedDocuments: DocumentKeySet
   ) {
     this.documentSet = new DocumentSet(query.docComparator.bind(query));
   }
 
+  get syncedDocuments(): DocumentKeySet {
```
schmidt-sebastian: Can you add the comment from the iOS implementation?

```
/**
 * The set of remote documents that the server has told us belongs to the target associated with
 * this view.
 */
```

packages/firestore/src/remote/remote_event.ts
```diff
@@ -70,6 +70,70 @@ export class RemoteEvent {
       resumeToken: emptyByteString()
     };
   }
+
+  /**
+   * Strips out mapping changes that aren't actually changes. That is, if the document already
+   * existed in the target, and is being added in the target, and this is not a reset, we can
+   * skip doing the work to associate the document with the target because it has already been done.
+   */
+  filterUpdatesFromTargetChange(
```
schmidt-sebastian: I am not a huge fan of this method.

1) A `filter` in JS usually returns a modified list that had the filter applied to it. The original list is not modified.

2) This method (and it's location) suggests that it modifies the encompassing RemoteEvent. It doesn't seem to do this, and could be implemented as a function outside of this class.

3) It's difficult to tell what is actually being updated here. The inline variable make the code much easier to read, but they make it hard to answer the question what `result` and `update` actually are.

packages/firestore/src/remote/remote_event.ts
```diff
@@ -70,6 +70,70 @@ export class RemoteEvent {
...
+    targetChange: TargetChange,
+    existingDocuments: DocumentKeySet
+  ): void {
+    if (targetChange.mapping instanceof UpdateMapping) {
+      const update = targetChange.mapping;
+      const added = update.addedDocuments;
+      let result = added;
+      added.forEach(key => {
```
schmidt-sebastian: Suggestion: We could also consider adding this functionality to SortedSet.

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+      targets,
+      noPendingResponses,
+      shouldSynthesize,
+      wrongState,
+      hasDocument,
+      docChange
+    );
+
+    const synthesized = DocumentKey.fromPathString('docs/2');
+    expect(event.documentUpdates.get(synthesized)).to.equal(null);
```
schmidt-sebastian: Nit: `to.be.null`

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+
+    const limboTargetChange = event.targetChanges[1];
+    event.synthesizeDeleteForLimboTargetChange(limboTargetChange, synthesized);
+    const expected = deletedDoc(
+      'docs/2',
+      event.snapshotVersion.toMicroseconds()
+    );
+    expectEqual(expected, event.documentUpdates.get(synthesized));
```
schmidt-sebastian: Flip the arguments here like in the rest of this class.

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+      event.targetChanges[2],
+      notSynthesized
+    );
+    expect(event.documentUpdates.get(notSynthesized)).to.not.exist;
+
+    event.synthesizeDeleteForLimboTargetChange(
+      event.targetChanges[3],
+      doc1.key
+    );
+    expect(event.documentUpdates.get(doc1.key) instanceof NoDocument).to.be
```
schmidt-sebastian: This is: `to.not.be.an.instanceof`

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+    event.synthesizeDeleteForLimboTargetChange(limboTargetChange, synthesized);
+    const expected = deletedDoc(
+      'docs/2',
+      event.snapshotVersion.toMicroseconds()
+    );
+    expectEqual(expected, event.documentUpdates.get(synthesized));
+
+    const notSynthesized = DocumentKey.fromPathString('docs/no1');
+    event.synthesizeDeleteForLimboTargetChange(
+      event.targetChanges[2],
```
schmidt-sebastian: Do you mind adding the little helper variables here and below like you did for `limboTargetChange`?

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+      1,
+      targets,
+      noPendingResponses,
+      newDocChange,
+      resetTargetChange,
+      existingDocChange
+    );
+
+    const updateChange = event.targetChanges[1];
+    expect(updateChange.mapping instanceof UpdateMapping).to.be.true;
```
schmidt-sebastian: This is `to.be.instanceof`.

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -460,4 +464,101 @@ describe('RemoteEvent', () => {
...
+    const update = updateChange.mapping as UpdateMapping;
+    expect(update.addedDocuments.has(existingDoc.key)).to.be.true;
+
+    const existingKeys = documentKeySet().add(existingDoc.key);
+    event.filterUpdatesFromTargetChange(updateChange, existingKeys);
+    expect(update.addedDocuments.has(existingDoc.key)).to.be.false;
+    expect(update.addedDocuments.has(newDoc.key)).to.be.true;
+
+    const resetChange = event.targetChanges[2];
+    expect(resetChange.mapping instanceof ResetMapping).to.be.true;
```
schmidt-sebastian: This is `to.be.instanceof`.

packages/firestore/src/remote/remote_event.ts
gsoltis: I'm not opposed to this, but from a quick look, it appears we want a few classic collection methods: `filter`, `foreach`, and maybe `collect`?

This seems like the sort of thing we would want to do in sync across all of the `SortedSet` and `SortedMap` implementations. Maybe file a bug for it? I don't think it should be part of this change.

packages/firestore/src/remote/remote_event.ts
schmidt-sebastian: I understand that this is the ported version of your original PR. I am still a little concerned though since this method changes internal and external state at the same time, while its naming indicates something else entirely.

I know that this is far from what iOS does, but I would prefer something like:


```
filterKnownDocuments(targetId:TargetID, knownDocuments:..) : RemoteEvent {
   ... this reads internal state, but doesn't mutatate it.
}
```
or:

```
removeKnownDocuments(targetId:TargetID, knownDocuments:..) : void {
   ... this reads internal state and also mutatates it.
}
```

In the interim, and to keep the port close, we could also make this a bit better by making it a static method.

packages/firestore/src/remote/remote_event.ts
schmidt-sebastian: That's fair, but I will ask you to do next time you implement `filter` :)

packages/firestore/src/remote/remote_event.ts
gsoltis: As discussed in person, kept the method non-static, but I switched the parameter to be `targetId` rather than `targetChange`, and added the requisite null check inside the method.

PR #732, opened 2018-04-26
Using an IndexedDB mock for Node testing

schmidt-sebastian:
This likely falls in the category of "hack" but it is going to make my life much more pleasant.

This PR allows running the Persistence-based tests on Node using https://github.com/axemclion/IndexedDBShim. There are some glitches that I worked around (like lingering files) but overall, I am very pleased to report that I can now debug tests using IntelliJ.

I also included default test configurations for IntelliJ users.

Note that while I am able to run the tests via IntelliJ, `test:node:persistence` still has some occasional failures that I punted on for now (about 3-10 tests fail for with persistence related errors, and 1 fails because XMLHTTPRequest can't be resolved).

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -13,6 +13,7 @@
     "test:browser": "karma start --single-run",
     "test:browser:debug": "karma start --browsers=Chrome --auto-watch",
     "test:node": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --require ts-node/register/type-check --require index.node.ts --retries 5 --timeout 5000 --exit",
+    "test:node:persistence": "USE_MOCK_PERSISTENCE=YES TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --require ts-node/register/type-check --require index.node.ts --require test/util/mock_persistence.ts --retries 5 --timeout 5000 --exit",
     "prepare": "npm run build"
   },
   "main": "dist/index.node.cjs.js",
@@ -39,6 +40,7 @@
     "@types/mocha": "5.0.0",
     "@types/sinon": "4.3.1",
     "chai": "4.1.2",
+    "indexeddbshim":"^3.6.1",
     "karma": "2.0.0",
     "karma-chrome-launcher": "2.2.0",
     "karma-cli": "1.0.1",
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -40,7 +40,7 @@
     "@types/mocha": "5.0.0",
     "@types/sinon": "4.3.1",
     "chai": "4.1.2",
-    "indexeddbshim":"^3.6.1",
+    "indexeddbshim":"3.6.1",
     "karma": "2.0.0",
     "karma-chrome-launcher": "2.2.0",
     "karma-cli": "1.0.1",
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -429,9 +429,11 @@ follow these steps, YOUR APP MAY BREAK.`);
   }
 
   INTERNAL = {
-    delete: async (): Promise<void> => {
+    delete: async (options?: {
+      purgePersistenceWithDataLoss?: boolean;
+    }): Promise<void> => {
       if (this._firestoreClient) {
-        return this._firestoreClient.shutdown();
+        return this._firestoreClient.shutdown(options);
       }
     }
   };
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -546,6 +546,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
 
 function convertStreamToken(token: ProtoByteString): string {
   if (token instanceof Uint8Array) {
+    // TODO(b/78771403): Convert tokens to strings during deserialization
     assert(
       process.env.USE_MOCK_PERSISTENCE === 'YES',
       'Persisting non-string stream tokens is only supported with mock persistence .'
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -190,7 +190,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
     );
 
     this.metadata.lastAcknowledgedBatchId = batchId;
-    this.metadata.lastStreamToken = validateStreamToken(streamToken);
+    this.metadata.lastStreamToken = convertStreamToken(streamToken);
 
     return mutationQueuesStore(transaction).put(this.metadata);
   }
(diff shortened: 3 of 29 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -429,9 +429,11 @@ follow these steps, YOUR APP MAY BREAK.`);
   }
 
   INTERNAL = {
-    delete: async (): Promise<void> => {
+    delete: async (options?: {
```
schmidt-sebastian: Note that I did not expose this option via the .d.ts definition for delete().

packages/firestore/src/local/local_serializer.ts
```diff
@@ -132,11 +132,18 @@ export class LocalSerializer {
...
+
+    if (queryData.resumeToken instanceof Uint8Array) {
+      assert(
+        process.env.USE_MOCK_PERSISTENCE === 'YES',
+        'Persisting non-string stream tokens is only supported with mock persistence .'
+      );
+      resumeToken = queryData.resumeToken.toString();
+    } else {
+      resumeToken = queryData.resumeToken;
+    }
```
mikelehen: Similar feedback as elsewhere.  If turning resume tokens into strings works, perhaps we should just do that as soon as we get them...  Add a TODO or log a bug?

packages/firestore/src/local/simple_db.ts
```diff
@@ -102,29 +102,30 @@ export class SimpleDb {
...
-
-    if (
-      ua.indexOf('MSIE ') > 0 ||
-      ua.indexOf('Trident/') > 0 ||
-      ua.indexOf('Edge/') > 0
-    ) {
-      return false;
-    } else {
-      return true;
+    if (window.navigator) {
```
mikelehen: Is this worth a comment or assertion to indicate that this check exists for MOCK_PERSISTENCE ?

packages/firestore/test/util/mock_persistence.ts
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+import * as registerIndexedDBShim from 'indexeddbshim';
+import * as fs from 'fs';
+
+// WARNING: The `indexeddbshim` installed via this module should only ever be
+// used during initial development. Do not use this code in your production apps
+// and always validate your changes via `yarn test:browser` (which uses a
+// browser-based IndexedDB implementation) before integrating with Firestore.
```
mikelehen: This warning reads a bit strange to me. I'm not sure if the intended audience is Firestore SDK contributors (mentioning "via 'yarn test:browser'"), or Firestore SDK consumers ("production apps"?)...

packages/firestore/package.json
```diff
@@ -39,6 +40,7 @@
     "@types/mocha": "5.0.0",
     "@types/sinon": "4.3.1",
     "chai": "4.1.2",
+    "indexeddbshim":"^3.6.1",
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed.

packages/firestore/src/api/database.ts
schmidt-sebastian: I played around with this more yesterday - and figured out why this isn't working (which sounds like both good and bad news, but is probably just bad news). If I pass ":memory:" or "file::memory:" to the IndexedDB Shim, it tries to create a file on disk with that name.. and subsequently times out all operations.  Since setting `basepath` also doesn't work for me ("The upgrade transaction was aborted."), I am inclined to keep this cleanup. Without it, a simple run creates hundreds of files in the main repo directory - which looks fairly scary for users that don't know what's going on.

I also think that we should do this cleanup for our tests in general - without it, we are leaving a ton of data behind in Chrome headless.

I can delete the lingering files in `withTestDb()` and not rely on `.shutdown()` if you do prefer this. We would then have to special case this testing mode there as well.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -190,7 +190,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
     );
 
     this.metadata.lastAcknowledgedBatchId = batchId;
-    this.metadata.lastStreamToken = validateStreamToken(streamToken);
+    this.metadata.lastStreamToken = convertStreamToken(streamToken);
```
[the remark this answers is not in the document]
schmidt-sebastian: I was kinda surprised that this works as is, but didn't dig to much deeper since my main goal was just to get the tests passing. I filed a bug to do this cleanup later (https://b.corp.google.com/issues/78771403).

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -544,12 +544,16 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
 }
 
-function validateStreamToken(token: ProtoByteString): string {
-  assert(
-    typeof token === 'string',
-    'Persisting non-string stream token not supported.'
-  );
-  return token as string;
+function convertStreamToken(token: ProtoByteString): string {
+  if (token instanceof Uint8Array) {
```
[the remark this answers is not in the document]
schmidt-sebastian: Exists everywhere: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array

packages/firestore/src/local/local_serializer.ts
schmidt-sebastian: Added TODO pointing to the bug.

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Added comment and a check that USE_MOCK_PERSISTENCE is "YES".

packages/firestore/test/util/mock_persistence.ts
schmidt-sebastian: It was meant for both :) But since our customers will likely never see this, I removed that portion.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -425,7 +425,7 @@ abstract class TestRunner {
 
   async shutdown(): Promise<void> {
     await this.remoteStore.shutdown();
-    await this.persistence.shutdown();
+    await this.persistence.shutdown(true);
```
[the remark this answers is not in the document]
schmidt-sebastian: Done (here and elsewhere).

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -21,7 +21,13 @@ import { DatabaseId, DatabaseInfo } from '../../../src/core/database_info';
 import { DEFAULT_PROJECT_ID } from '../util/helpers';
 import { getDefaultDatabaseInfo } from '../util/internal_helpers';
 
-const describeFn = typeof window !== 'undefined' ? describe : xdescribe;
+// We need to check both `window` and `window.navigator` to make sure we are
+// not running in Node with IndexedDBShim.
```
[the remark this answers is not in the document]
schmidt-sebastian: I think our pattern is wrong:

```
test/{,!(browser)/**/}*.test.ts
```

I think it should be something like:

```
test/{,!(**/browser)/**/}*.test.ts
```


... but that doesn't work for me, so I took your advice and ignored this comment.

PR #744, opened 2018-04-30
Port limbo document identification

gsoltis:
Port of https://github.com/firebase/firebase-ios-sdk/pull/1167

While aggregating updates into an `RemoteEvent`, track which documents are only changing due to limbo targets.

This is intended to be used in the future to make sure that non-eager GC schemes don't leak documents that are not attached to a persisted target.


packages/firestore/src/remote/watch_change.ts
```diff
@@ -188,30 +192,84 @@ export class WatchChangeAggregator {
    * so to know if a target is active, there must be no pending acks we're
    * waiting for and it must be in the current list of targets that the client
    * cares about.
```
schmidt-sebastian: This comment makes more sense `isActiveTarget`. We should probably combine it with the comment you just added (that describes that `isActiveTarget` calls this method).

Once that is done, I don't think `queryDataForActiveTarget` needs an actual comment. It's pretty self-explanatory.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -188,30 +192,84 @@ export class WatchChangeAggregator {
...
    */
   protected isActiveTarget(targetId: TargetId): boolean {
-    return (
-      !objUtils.contains(this.pendingTargetResponses, targetId) &&
-      objUtils.contains(this.listenTargets, targetId)
-    );
+    return this.queryDataForActiveTarget(targetId) !== null;
+  }
+
+  private updateLimboDocuments(
```
schmidt-sebastian: You wrote such a nice comment in the iOS implementation. It'd be awesome to see it here as well.

gsoltis: Oops, forgot to copy that. Fixed.

packages/firestore/src/remote/watch_change.ts
gsoltis: I think I disagree. This is the method that is actually determining whether or not a target is 'active'. This is confusing in the presence of `isActiveTarget`, hence the comment for that method that it defers to this one.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -188,30 +192,84 @@ export class WatchChangeAggregator {
...
   }
 
   private addDocumentChange(docChange: DocumentWatchChange): void {
     let relevant = false;
+    let isOnlyLimbo = true;
 
     for (const targetId of docChange.updatedTargetIds) {
-      if (this.isActiveTarget(targetId)) {
+      const queryData = this.queryDataForActiveTarget(targetId);
+      if (queryData) {
```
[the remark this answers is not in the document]
gsoltis: I'd prefer to keep this as-is, as it encapsulates all of the decision logic in one function, rather than splitting it.

packages/firestore/src/remote/watch_change.ts
schmidt-sebastian: The comment as is doesn't describe what this function does, but rather describes what `isActiveTarget` is used for. If you intend to keep this comment, can you emphasize that it returns QueryData? You can use the current comment to explain what an "active target" is. 

How about something along the lines of:

```
Returns the QueryData for a target that has been acknowledged by the backend and is still subscribed to by the user. This method is used to ensure that only events for these active targets are processed.
```

Please rewrite this as you please, as I am currently suffering from writer's block.

gsoltis: Rewritten.

PR #749, opened 2018-04-30
Port feedback on event filtering

gsoltis:
I got some good feedback on the java port internally, so I am porting that to typescript.

Filtering is moved onto the mapping subtypes, and some tests are split into distinct cases.

### DIFF
--- a/packages/firestore/src/remote/remote_event.ts
+++ b/packages/firestore/src/remote/remote_event.ts
@@ -101,7 +101,7 @@ export class RemoteEvent {
       // arrives, the document is not present in the snapshot and our
       // normal view handling would consider the document to remain in
       // limbo indefinitely because there are no updates to the document.
-      // To avoid this, we specially handle this just this case here:
+      // To avoid this, we specially handle this case here:
       // synthesizing a delete.
       //
       // TODO(dimond): Ideally we would have an explicit lookup query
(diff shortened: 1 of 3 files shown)


packages/firestore/src/remote/remote_event.ts
```diff
@@ -242,4 +219,19 @@ export class UpdateMapping {
       this.removedDocuments.isEqual(other.removedDocuments)
     );
   }
+
+  /**
+   * Strips out mapping changes that aren't actually changes. That is, if the document already
+   * existed in the target, and is being added in the target, and this is not a reset, we can
+   * skip doing the work to associate the document with the target because it has already been done.
+   */
+  filterUpdates(existingKeys: DocumentKeySet): void {
```
schmidt-sebastian: The name of this functions is kind of strange in this context (`UpdateMapping.fitlerUpdates`). I can't think of a much better name right now (and we would probably want to update the name of the class rather than the name of this function). Feel free to leave as is unless you can think of something better.

gsoltis: I'm going to leave it as-is to match other clients. If someone were to do some additional work on this method and prefer a different name, I'm sure they could rename it then and I'd be happy to approve it.

PR #761, opened 2018-05-02
Fix up changelog.

mikelehen:
A bunch of the "Unreleased" changes have actually been released, so I broke them out
into the appropriate versions. Also added a changelog entry for the 0.4.1 patch release
to fix Node.JS support.


wilhuff:
This seems like a Firestore-specific change. Does this need the Component-Auth label?

mikelehen:
Uhhhh...  No idea how Component-Auth ended up there. Thanks. :-)

PR #763, opened 2018-05-02
Add arrayUnion() and arrayRemove() Transforms (not publicly exposed yet)

mikelehen:
[This review is not super urgent if you're busy this week]

This is a port of The following iOS commits:
https://github.com/firebase/firebase-ios-sdk/commit/60d43d7b0980719beac21811da712f1cb4881308
https://github.com/firebase/firebase-ios-sdk/commit/360e58901c359d7d21da4fff8043894c843427b7
https://github.com/firebase/firebase-ios-sdk/commit/f2ec9e11fba82d4a76d989293d270a37ad1373a2
https://github.com/firebase/firebase-ios-sdk/commit/e36cc9610b11dfd2581ba5e3fda1917e0d5e697a
https://github.com/firebase/firebase-ios-sdk/commit/7dca635039efb00082624552c223ac7f47ea0cad

While I split this PR into separate commits, they don't match 1:1 with the iOS commits (some of the commits iterated on bits from previous ones) so it may be a little difficult to compare. So I'd probably just compare to iOS HEAD if in doubt.

### DIFF
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -16,18 +16,18 @@
 
 import { Timestamp } from '../api/timestamp';
 import { SnapshotVersion } from '../core/snapshot_version';
-import { assert, fail } from '../util/assert';
+import { assert } from '../util/assert';
 import * as misc from '../util/misc';
 
 import { Document, MaybeDocument, NoDocument } from './document';
 import { DocumentKey } from './document_key';
-import {
-  FieldValue,
-  ObjectValue,
-  ServerTimestampValue,
-  ArrayValue
-} from './field_value';
+import { FieldValue, ObjectValue } from './field_value';
 import { FieldPath } from './path';
+import {
+  TransformOperation,
+  ArrayRemoveTransformOperation,
+  ArrayUnionTransformOperation
+} from './transform_operation';
 
 /**
  * Provides a set of fields that can be used to partially patch a document.
@@ -62,45 +62,6 @@ export class FieldMask {
 
   isEqual(other: FieldMask): boolean {
     return misc.arrayEquals(this.fields, other.fields);
-  }
-}
-
-/** Represents a transform within a TransformMutation. */
-export interface TransformOperation {
-  isEqual(other: TransformOperation): boolean;
-}
-
-/** Transforms a value into a server-generated timestamp. */
-export class ServerTimestampTransform implements TransformOperation {
-  private constructor() {}
-  static instance = new ServerTimestampTransform();
-
-  isEqual(other: TransformOperation): boolean {
-    return other instanceof ServerTimestampTransform;
-  }
-}
-
-/** Transforms an array value via a union operation. */
-export class ArrayUnionTransformOperation implements TransformOperation {
-  constructor(readonly elements: FieldValue[]) {}
-
-  isEqual(other: TransformOperation): boolean {
-    return (
-      other instanceof ArrayUnionTransformOperation &&
-      misc.arrayEquals(other.elements, this.elements)
-    );
-  }
-}
-
-/** Transforms an array value via a remove operation. */
-export class ArrayRemoveTransformOperation implements TransformOperation {
-  constructor(readonly elements: FieldValue[]) {}
-
-  isEqual(other: TransformOperation): boolean {
-    return (
-      other instanceof ArrayRemoveTransformOperation &&
-      misc.arrayEquals(other.elements, this.elements)
-    );
   }
 }
 
@@ -619,26 +580,18 @@ export class TransformMutation extends Mutation {
         previousValue = baseDoc.field(fieldTransform.field) || null;
       }
 
-      let transformResult: FieldValue = null;
       // The server just sends null as the transform result for array union /
       // remove operations, so we have to calculate a result the same as we do
       // for local applications.
-      if (transform instanceof ArrayUnionTransformOperation) {
-        transformResult = this.arrayUnionTransformResult(
-          transform,
-          previousValue
-        );
-      } else if (transform instanceof ArrayRemoveTransformOperation) {
-        transformResult = this.arrayRemoveTransformResult(
-          transform,
-          previousValue
-        );
+      if (
+        transform instanceof ArrayUnionTransformOperation ||
+        transform instanceof ArrayRemoveTransformOperation
+      ) {
+        transformResults.push(transform.transform(previousValue));
       } else {
         // Just use the server-supplied result.
-        transformResult = serverTransformResults[i];
+        transformResults.push(serverTransformResults[i]);
       }
-
-      transformResults.push(transformResult);
     }
     return transformResults;
   }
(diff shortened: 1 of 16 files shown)


packages/firestore/src/model/mutation.ts
```diff
@@ -576,23 +661,83 @@ export class TransformMutation extends Mutation {
...
 
+  /**
+   * Transforms the provided `previousValue` via the provided `arrayUnion`. Used
+   * for both local application and after server acknowledgement.
+   */
+  private arrayUnionTransformResult(
+    arrayUnion: ArrayUnionTransformOperation,
+    previousValue: FieldValue | null
+  ): FieldValue {
+    const result = this.coercedFieldValuesArray(previousValue);
```
schmidt-sebastian: Would it make sense to have this live within the Transform class itself? If TransformOperation had an abstract method, the code in `localTransformResults` and `serverTransformResults ` would not need to know about each classes semantics.

packages/firestore/src/model/mutation.ts
```diff
@@ -576,23 +661,83 @@ export class TransformMutation extends Mutation {
...
+  /**
+   * Inspects the provided value, returning a mutable copy of the internal array
+   * if it's an ArrayValue and an empty mutable array if it's null or any other
+   * type of FieldValue.
+   */
+  private coercedFieldValuesArray(value: FieldValue | null): FieldValue[] {
+    if (value instanceof ArrayValue) {
+      return value.internalValue.slice();
+    } else {
+      // coerce to empty array.
```
schmidt-sebastian: This seems like it is rarely what the user wants, and as such, I feel that we should be logging it at the very least.

packages/firestore/test/integration/api/array_transforms.test.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  // Accumulator used to capture events during the test.
+  let accumulator: EventsAccumulator<firestore.DocumentSnapshot>;
+
+  // Listener registration for a listener maintained during the course of the
+  // test.
+  let listenerRegistration: () => void;
+
+  /** Waits for a latency compensated local snapshot. */
+  async function waitForLocalEvent(): Promise<firestore.DocumentSnapshot> {
```
schmidt-sebastian: Can we move this to helpers? This is all rock solid code, but having it twice doesn't make it all that much better :)

packages/firestore/test/integration/api/array_transforms.test.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Wraps a test, getting a docRef and event accumulator, and cleaning them
+   * up when done.
+   */
+  async function withTestSetup<T>(test: () => Promise<T>): Promise<void> {
+    await withTestDoc(persistence, async doc => {
+      docRef = doc;
+      accumulator = new EventsAccumulator<firestore.DocumentSnapshot>();
+      listenerRegistration = docRef.onSnapshot(
```
schmidt-sebastian: Nit: While you are moving this, I feel that in JS "unsubscribe" is a much better name for this variable.

packages/firestore/test/integration/api/array_transforms.test.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Unlike the withTestSetup() tests above, these tests intentionally avoid
+   * having any ongoing listeners so that we can test what gets stored in the
+   * offline cache based purely on the write acknowledgement (without receiving
+   * an updated document via watch). As such they also rely on persistence
+   * being enabled so documents remain in the cache after the write.
+   */
+  (persistence ? describe : describe.skip)('Server Application: ', () => {
+    it('set() with no cached base doc', async () => {
```
schmidt-sebastian: Maybe I am missing something here - but could we not simplify these tests by disabling the network and querying from snapshots purely from cache?

packages/firestore/src/model/mutation.ts
mikelehen: I like that idea. I reworked this (including moving the TransformOperation classes to their own file to keep mutation.ts sane and to match iOS). I definitely like it better (and it reduced ~30 lines of code) though it's a _little_ wonky since 1) ServerTimestampTransform.transform() requires localWriteTime (which we only have during local-application) and 2) .transform() is called on the array transforms during "remote" application (after a write ack) when we don't have localWriteTime available, and so it has to be an optional param.

packages/firestore/src/model/transform_operation.ts
```diff
@@ -0,0 +1,97 @@
...
+
+import { FieldValue, ServerTimestampValue, ArrayValue } from './field_value';
+import { Timestamp } from '../api/timestamp';
+import * as misc from '../util/misc';
+import { assert } from '../util/assert';
+
+/** Represents a transform within a TransformMutation. */
+export interface TransformOperation {
+  /** Transforms the provided `previousValue`. */
+  transform(previousValue: FieldValue, localWriteTime?: Timestamp): FieldValue;
```
schmidt-sebastian: Thanks for this refactor!

I played with this a little to see if we could get rid of `localWriteTime` as an argument. I am not entirely happy with it, but here is the result of the preliminary investigation: https://gist.github.com/schmidt-sebastian/6dd9cb25015d14bc96d742c3d27d6fc6

It's a lot of churn, which simplifies some parts of our Server Timestamp handling, but it also adds complexity to other parts (mainly to the serializer).

packages/firestore/test/integration/api/array_transforms.test.ts
```diff
@@ -0,0 +1,233 @@
...
+  async function writeInitialData(
+    initialData: firestore.DocumentData
+  ): Promise<void> {
+    await docRef.set(initialData);
+    await accumulator.awaitLocalEvent();
+    const snapshot = await accumulator.awaitRemoteEvent();
+    expect(snapshot.data()).to.deep.equal(initialData);
+  }
+
+  async function expectLocalAndRemoteEvent(
```
schmidt-sebastian: I'd suggest to move this to EventAccumulator as well, even though it is only used here.

packages/firestore/test/integration/api/array_transforms.test.ts
schmidt-sebastian: Yeah, that makes sense. I was thinking you could take your knowledge of the codebase into account and test `applyToRemoteDocument` by calling `applyToLocalView` (shrug).

packages/firestore/src/model/transform_operation.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  isEqual(other: TransformOperation): boolean;
+}
+
+/** Transforms a value into a server-generated timestamp. */
+export class ServerTimestampTransform implements TransformOperation {
+  private constructor() {}
+  static instance = new ServerTimestampTransform();
+
+  transform(previousValue: FieldValue, localWriteTime?: Timestamp): FieldValue {
```
schmidt-sebastian: If you end up following the changes linked to above, you could take the `serverTransformResults` as an argument (you might also have to split this up into `.localTransform` and `.serverTransform`).

With that: 
- You can just call `transform.transform()` in mutation.ts.
- You can assert that the server transform value in null in `ArrayUnion/RemoveTransformOperation` .

packages/firestore/src/model/transform_operation.ts
mikelehen: While I would like to like your change, the reason localWriteTime is stored in the MutationBatch rather than in the Mutation or ServerTimestampTransform is so that when you do a batch write, all of your server timestamp sentinels resolve to the same timestamp (i.e. when you committed the batch rather than each one resolving to when you added the mutation to the batch which could theoretically be over a longer time period).

If I didn't miss anything, your change ends up storing a timestamp in both the ServerTimestampTransform and in the MutationBatch, and uses the one in the ServerTimestampTransform for local-application...  but after serializing / deserializing, I think the version in the mutation will end up being replaced with the version in the MutationBatch (so after restarting your app, your local server timestamps may change which isn't good).

So I'm not sure this works out.  We could perhaps make it work by creating the ServerTimestampTransforms during parsing, and then going back and assigning a localWriteTime to each one at the time the MutationBatch is constructed.  That way they're guaranteed to be the same and to match the timestamp in the MutationBatch.

But I'm not sufficiently convinced this is a meaningful improvement. As you ran into with the serializer, our current model translates well to the backend protos / implementation, which has some advantages.  So I'm inclined to stick with the warts we have rather than exchange for new ones.

packages/firestore/src/model/transform_operation.ts
mikelehen: Though I didn't do the above changes, I did go ahead and change this as it's something I had already been considering. It added more lines of code, but is ultimately cleaner, I think.

NOTE: I did not add the assert for the transform value being null, since I don't want to prevent the server from changing that behavior in the future. We may decide that the server really should send a transformResult and we don't want that to break old clients.  In general, we should assert about our own state, but be careful about asserting on what the server sends us.

packages/firestore/test/integration/api/array_transforms.test.ts
mikelehen: Mostly I was nervous about the fact that the server sends us a null transform result and I wanted to make sure that we correctly ignore that and perform the local transform instead.  So I wanted to test that end-to-end.  It's possibly overkill or there are better ways to test it, but I'm inclined to just stick with these tests.

packages/firestore/src/model/transform_operation.ts
schmidt-sebastian: I am glad that I played around with this myself before blindly recommending it. Making the localWriteTime a constructor argument seemed much nicer before I changed the interactions between the mutation batches and the serializer. Thanks for looking at this as well, but I as it stands, I am 100% happy with where this PR landed.

packages/firestore/src/model/transform_operation.ts
schmidt-sebastian: Thanks, I do think this interaction is now much cleaner and much more obvious.

(... and I would hope that the backend would rev their protocol version before changing the semantics of a response).

packages/firestore/test/integration/api/array_transforms.test.ts
schmidt-sebastian: Looks like if we leave it as is, there's less code to review. We have a winner :)

PR #764, opened 2018-05-02
Array Contains support (not publicly exposed yet).

mikelehen:
[This review is not super urgent if you're busy this week]

Port of https://github.com/firebase/firebase-ios-sdk/commit/e4384c3e809556e75907df74cd116307f397472f

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -1282,7 +1282,15 @@ export class Query implements firestore.Query {
     validateDefined('Query.where', 3, value);
     let fieldValue;
     const fieldPath = fieldPathFromArgument('Query.where', field);
+    const relationOp = RelationOp.fromString(opStr);
     if (fieldPath.isKeyField()) {
+      if (relationOp === RelationOp.ARRAY_CONTAINS) {
+        throw new FirestoreError(
+          Code.INVALID_ARGUMENT,
+          "Invalid Query. You can't do array-contains queries on " +
+            'FieldPath.documentId() since document IDs are not arrays.'
+        );
+      }
       if (typeof value === 'string') {
         if (value.indexOf('/') !== -1) {
           // TODO(dimond): Allow slashes once ancestor queries are supported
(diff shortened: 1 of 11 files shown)


mikelehen:
Rebased against master since I removed usage of includesEqualElement() that I was relying on from the array transforms PR.

packages/firestore/src/api/database.ts
```diff
@@ -1282,7 +1282,15 @@ export class Query implements firestore.Query {
     validateDefined('Query.where', 3, value);
     let fieldValue;
     const fieldPath = fieldPathFromArgument('Query.where', field);
+    const relationOp = RelationOp.fromString(opStr);
     if (fieldPath.isKeyField()) {
+      if (relationOp === RelationOp.ARRAY_CONTAINS) {
+        throw new FirestoreError(
+          Code.INVALID_ARGUMENT,
+          "Invalid Query. You can't do array-contains queries on " +
```
schmidt-sebastian: Nit: s/do/perform/

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -426,6 +427,14 @@ apiDescribe('Database', persistence => {
       });
     });
 
+    it('inequality and array-contains on different fields works', () => {
+      return withTestCollection(persistence, {}, async coll => {
+        expect(() =>
+          coll.where('x', '>=', 32).where('y', arrayContainsOp, 'cat')
```
schmidt-sebastian: I think this would be more readable if you hardcoded 'array-contains' here. It's part of the public API and should not change anytime soon.

packages/firestore/src/api/database.ts
mikelehen: Thanks, done.

packages/firestore/test/integration/api/database.test.ts
mikelehen: Agreed! This exists because 'array-contains' is not exposed in the .d.ts. So arrayContainsOp contains 'array-contains' casted to `WhereFilterOp`.  Once we expose the API, it can go away. I've cleaned up the TODO to make that clearer.

PR #773, opened 2018-05-04
Enable Messaging tests on Sauce Labs

mmermerkaya:
Enables tests on all desktop browsers that currently support Push API and Notifications API (Chrome, Edge 17 and Firefox). Skips tests on Safari.

Thanks @yifanyang!


yifanyang:
LGTM!

PR #778, opened 2018-05-05
Add flag to setBackgroundMessageHandler for allowing all notifications to be handled

### LINKED ISSUE #777: FR: Allow FCMs setBackgroundMessageHandler to handle all notifications

  * Operating System version: macOS 10.13.3
  * Firebase SDK version: 4.13.1
  * Firebase Product: FCM (auth, database, storage, etc)

I propose with a flag, allow `setBackgroundMessageHandler` to be used to handle all notifications. This is an issue in some cases where the user wants to append some data to their notification object, or handle their notifications in certain ways.

Something like this would solve the issue and allow the user to handle everything themselves
```js
messaging.setBackgroundMessageHandler(payload => {
    // ...
}, true);
```

andrewfluck:
This fixes #777 

A small API change has been introduced to `setBackgroundMessageHandler`

```ts
setBackgroundMessageHandler(
    callback: (a: object) => Promise<void>,
    handlerMode: boolean
  ): void;
```

`handlerMode` when true, will disable the default handling, where notification is automatically sent. this fixes a lot of gotchas and will make the DX for some users better without finding a hacky workaround. I especially need this for work, we store extra data that isn't in the notification object, and we don't want to break our other services as a result of adding it.

### DIFF
--- a/packages/messaging-types/index.d.ts
+++ b/packages/messaging-types/index.d.ts
@@ -32,7 +32,7 @@ export class FirebaseMessaging {
     completed?: () => void
   ): Unsubscribe;
   requestPermission(): Promise<void>;
-  setBackgroundMessageHandler(callback: (a: object) => Promise<void>): void;
+  setBackgroundMessageHandler(callback: (a: object) => Promise<void>, handlerMode: boolean): void;
   useServiceWorker(registration: ServiceWorkerRegistration): void;
   usePublicVapidKey(b64PublicKey: string): void;
 }
--- a/packages/messaging-types/index.d.ts
+++ b/packages/messaging-types/index.d.ts
@@ -32,7 +32,10 @@ export class FirebaseMessaging {
     completed?: () => void
   ): Unsubscribe;
   requestPermission(): Promise<void>;
-  setBackgroundMessageHandler(callback: (a: object) => Promise<void>, handlerMode: boolean): void;
+  setBackgroundMessageHandler(
+    callback: (a: object) => Promise<void>,
+    handlerMode: boolean
+  ): void;
   useServiceWorker(registration: ServiceWorkerRegistration): void;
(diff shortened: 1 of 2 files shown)


andrewfluck:
I signed it

andrewfluck:
It is a very small fix and won't cause any broken code, as this mode defaults to false

andrewfluck:
I don't think there is a repo for the docs, but if there is, could you point me towards it? I can modify the docs accordingly if need be :)

andrewfluck:
Just as an example of this in use, our BE sends some extra information on data. For example, we send things like:
```
{data: { user.image_avatar: '...', user.user: '...', message.message: '', ... }, notification: { ... }}
```

https://www.youtube.com/watch?v=3iRb8hpMe6Y

gauntface:
I'm really sorry but we can't take this without going through a review process with the Firebase team.

Generally I agree that the experience here is pretty bad and we should fix it. We've discussed some approaches to simplify this but will be a breaking change so will take some time.

In future I would encourage opening an issue to discuss this before raising a PR to avoid any risk of wasting time :(

andrewfluck:
@gauntface Is there a recommended way to make a local copy of the library with this patch included? We need this as a feature and can't change anything without causing problems. If you could help me with that, that would be great! Either that, or is there a manual way to handle everything from scratch? Or is it tightly integrated?

andrewfluck:
Regarding the PR itself, it doesn't cause any breaking changes. Without the flag, by default it works fine. I thought I'd reiterate that just in case 😉

andrewfluck:
If you do want to discuss things, let's do discuss a better way to handle things in #777

mmermerkaya:
About this not being a breaking change, you're right. But since it's a change in the API, we still can't accept this without an internal review, as @gauntface mentioned.

PR #782, opened 2018-05-07
Storing a Query Changelog

schmidt-sebastian:
This might be more of an RFC before https://github.com/firebase/firebase-js-sdk/pull/784 gets merged, but here we go either way.

This PR:
- Combines the add/remove Query Cache callbacks in the persistence layer into one function (applyTargetState).  This function then deals with the query changelog as well.
- Adds functionality to the query cache to retrieve the query changes since a specific snapshot version.
- Changes the targetChange object store to only store a list of changed document keys (as discussed, it doesn't matter whether a document was added, modified or removed as the view computation just uses a list of modified keys).
- Adds a DbTimestampArray type since DbTimestamps can't be used in keys.

~~Also some general cleanup~~ 
~~Removes a bunch of error logs from running with Mock persistence by not logging the errors from accessing window.localStorage.get()~~


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -582,18 +588,19 @@ export class DbTargetChange {
...
+    public snapshotVersion: DbTimestampArray,
     /**
      * The keys of the changed documents in this snapshot.
      */
-    public changes: {
-      added?: EncodedResourcePath[];
-      modified?: EncodedResourcePath[];
-      removed?: EncodedResourcePath[];
-    }
+    public changes: EncodedResourcePath[]
```
mikelehen: So...  I think you merged added/modified/removed in response to my feedback. And I gave you that feedback because you don't need to know the type of the change in order to update local views...  but in thinking about this change and the later changes that will use this information, I'm left uncertain...

I guess it boils down to a question (which we may want to discuss offline):  Are you intending to keep `View._syncedDocuments` correctly updated in secondary tabs?  This is not necessary for local view evaluation, and so I think the reasons you might answer yes would be if: a) secondary tabs are responsible for detecting limbo documents and notifying the primary tab to resolve them, or b) we want secondary tabs to have this pre-populated in case they need to take over as primary (but I don't think this will work in general because secondary tabs won't have a View for every active listen across all tabs anyway)

If you are intending to keep it in sync, then I think you may want to preserve the change type here.  If you are not, then I'm wondering if a "Query Change Log" is actually what we want or if we should have a "Remote Document Cache Change Log" instead.  I think this might actually be simpler and may be my preference.  Secondary tabs would only care about keeping their local views correct, which is a consequence of the mutation queue and the remote document cache (well, and whether targets are current, which we need to solve still either way). This would also make sure that secondary clients see when we update remote documents after a write ack (https://github.com/firebase/firebase-js-sdk/blob/a56cbafd564a1e9cad5be366fa58643118745c92/packages/firestore/src/local/local_store.ts#L767) though maybe that isn't strictly necessary (in the cases that matter, we should [eventually at least] get an update from watch as well, which would show up in the query cache change log, I guess].

We can talk more but in general I'm nervous that secondary tabs will be surfacing query results based on remote document cache + mutation queue, but the notification mechanism for changes is via this query cache changelog which is tracking a slightly different thing...

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -327,6 +326,68 @@ export class IndexedDbQueryCache implements QueryCache {
       )
       .next(() => count > 0);
   }
+
+  getChangesSince(
+    transaction: PersistenceTransaction,
+    targetId: TargetId,
+    snapshotVersion: SnapshotVersion
+  ): PersistencePromise<DocumentKeySet> {
+    let documentUpdates = documentKeySet();
```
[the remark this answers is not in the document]
schmidt-sebastian: Changed the name to be `changedKeys`.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -86,6 +86,12 @@ export class DbTimestamp {
   constructor(public seconds: number, public nanoseconds: number) {}
 }
 
+/**
+ * An array of seconds and nanoseconds that allows storage of timestamps in an
```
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -86,6 +86,12 @@ export class DbTimestamp {
...
+ * an IndexedDb key.
```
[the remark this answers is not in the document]
schmidt-sebastian: That was supposed to be hidden by the linebreak!

packages/firestore/src/local/local_serializer.ts
```diff
@@ -90,6 +95,45 @@ export class LocalSerializer {
...
 
+  /** Converts a timestamp into an array of seconds and nanoseconds.  */
+  toTimestampArray(timestamp: Timestamp): DbTimestampArray {
+    return [timestamp.seconds, timestamp.nanoseconds];
+  }
+
+  /**
+   * Encodes a set of document keyss into a DbTargetChange.
+   */
+  toDbTargetChanges(
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore/src/local/local_serializer.ts
```diff
@@ -90,6 +95,45 @@ export class LocalSerializer {
...
+    );
+
+    changes.forEach(key => {
+      targetChange.changes.push(encode(key.path));
+    });
+
+    return targetChange;
+  }
+
+  /** Decodes a DbTargetChange into its set of updated document keys. */
```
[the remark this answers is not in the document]
schmidt-sebastian: I kept this but this is now `fromDbResourcePaths` (with an 's') and just takes a list of EncodedResourcePaths. And by list I mean array.

packages/firestore/src/local/query_cache.ts
```diff
@@ -120,40 +121,36 @@ export interface QueryCache extends GarbageSource {
...
-  removeMatchingKeysForTargetId(
+  getMatchingKeysForTargetId(
     transaction: PersistenceTransaction,
     targetId: TargetId
-  ): PersistencePromise<void>;
+  ): PersistencePromise<DocumentKeySet>;
 
   /**
-   * Returns the document keys that match the provided target ID.
+   * Returns the set of documents that were updated by the given target
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/query_cache.ts
```diff
@@ -120,40 +121,36 @@ export interface QueryCache extends GarbageSource {
...
+   * starting with the provided snapshot version (inclusive).
    *
-   * Multi-Tab Note: This operation is safe to use from secondary clients.
+   * Returned document keys are for documents that were added, modified or
+   * removed at any intermediate snapshots and may include documents that only
+   * existed temporarily.
    */
-  getMatchingKeysForTargetId(
+  getChangesSince(
```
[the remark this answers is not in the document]
schmidt-sebastian: `getChangedKeysForTargetId` works for me. I also made the snapshotVersion optional, which works well with the new name.

BTW, this used to be called `getAccumulatedChangesSince` but that was too hard to spell.

packages/firestore/src/local/query_cache.ts
```diff
@@ -120,40 +121,36 @@ export interface QueryCache extends GarbageSource {
...
+   * Applies the target change event to the cached query results of the given
+   * target.
    */
-  addMatchingKeys(
+  applyTargetChange(
     transaction: PersistenceTransaction,
-    keys: DocumentKeySet,
-    targetId: TargetId
+    tagetId: TargetId,
+    change: TargetChange
```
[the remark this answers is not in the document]
schmidt-sebastian: (This is no longer applicable).

Your arguments make a ton of sense (I would have changes this across platforms either way). I have moved this change to the remote document cache and am no longer touching these files.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: As we discussed last week, I have reworked this to use a document change log instead. My main argument against this was churn, and the main argument for this was that this will definitely solve the limbo document problem without any further work. Let me know what you think of the updated PR.

One nice outcome of this is that I no longer have to build up and buffer changes, since the RemoteDocumentBuffer already exists for this purpose.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -15,35 +15,71 @@
...
+    for (const maybeDocument of maybeDocuments) {
+      promises.push(
+        documentStore.put(
+          dbKey(maybeDocument.key),
+          this.serializer.toDbRemoteDocument(maybeDocument)
+        )
+      );
+
+      const existingChanges =
+        accumulatedChanges.get(maybeDocument.version) || documentKeySet();
```
mikelehen: I am not sure we can use the versions of the documents as the key in the change log...  We can / will receive "old" documents (e.g. you do a new query that includes documents that haven't changed in months, etc.).

We could probably use the global snapshot that the document changes came from or something along those lines, but it's probably easier / sufficient to just use an IndexedDb autoIncrement key generator (`db.createObjectStore("names", { autoIncrement : true });` or whatever).  FWIW that means we could also ditch DbTimestampArray.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -81,6 +122,38 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       })
       .next(() => results);
   }
+
+  getNextDocumentChanges(
```
mikelehen: I think getNewDocumentChanges() might convey better that this returns changes since the last call. Don't feel super strongly though.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -81,6 +122,38 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+      this.lastReturnedKey,
+      /*lowerOpen=*/ true
+    );
+
+    return documentChangesStore(transaction)
+      .iterate({ range }, (_, snapshotChange) => {
+        changedKeys = changedKeys.unionWith(
+          this.serializer.fromDbResourcePaths(snapshotChange.changes)
+        );
+        this.lastReturnedKey = snapshotChange.autoGeneratedKey;
```
mikelehen: For some reason I don't love 'autoGeneratedKey'.  I think it seems weird to put the generation mechanism into the field name...  I'd be more inclined to just name it 'id' or 'changeId' or something...  But at the same time, autoGeneratedKey is very clear and explicit and so if you like it, feel free to keep it.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -89,10 +162,33 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+  return getStore<DbDocumentChangesKey, DbDocumentChanges>(
+    txn,
+    DbDocumentChanges.store
+  );
+}
+
+/**
+ * Helper to get a typed SimpleDbStore from a transaction.
+ */
+function getStore<KeyType extends IDBValidKey, ValueType>(
```
mikelehen: Might be worth making this a static function on SimpleDbTransaction?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -560,43 +560,41 @@ function ensureTargetGlobalExists(
...
- * given target. This is used to facilitate storing query changelogs in the
- * targetChanges object store.
+ * An object store to store the keys of changed documents. This is used to
+ * facilitate storing document changelogs in the Remote Document Cache.
  *
  * PORTING NOTE: This is used for change propagation during multi-tab syncing
  * and not needed on iOS and Android.
  */
-export class DbTargetChange {
+export class DbDocumentChanges {
```
mikelehen: I'd name this to match DbRemoteDocuments since they're very tightly coupled... i.e. DbRemoteDocumentChanges

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -560,43 +560,41 @@ function ensureTargetGlobalExists(
...
+  static store = 'documentChanges';
 
-  /** Keys are automatically assigned via the targetId and snapshotVersion. */
-  static keyPath = ['targetId', 'snapshotVersion'];
+  /** Keys are auto-generated via the `autoGeneratedKey` property. */
+  static keyPath = 'autoGeneratedKey';
+
+  /**
+   * The auto-generated key of this entry.
+   */
```
mikelehen: put this comment on one line if you want...

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -560,43 +560,41 @@ function ensureTargetGlobalExists(
...
-    /**
-     * The targetId identifying a target.
-     */
-    public targetId: TargetId,
-    /**
-     * The snapshot version for this change.
-     */
-    public snapshotVersion: DbTimestamp,
     /**
      * The keys of the changed documents in this snapshot.
```
mikelehen: Drop "in this snapshot" or make it clearer what this is referring to?

packages/firestore/src/local/local_serializer.ts
```diff
@@ -90,6 +92,32 @@ export class LocalSerializer {
     return new MutationBatch(dbBatch.batchId, timestamp, mutations);
   }
 
+  /*
+   * Encodes a set of document keys into an array of EncodedResourcePaths.
+   */
+  toDbResourcePaths(changes: DocumentKeySet): EncodedResourcePath[] {
```
mikelehen: "changes" => "keys" ?

packages/firestore/src/local/local_serializer.ts
```diff
@@ -90,6 +92,32 @@ export class LocalSerializer {
...
+    changes.forEach(key => {
+      encodedKeys.push(encode(key.path));
+    });
+
+    return encodedKeys;
+  }
+
+  /** Decodes an array of EncodedResourcePaths into a set of document keys. */
+  fromDbResourcePaths(encodedPaths: EncodedResourcePath[]): DocumentKeySet {
+    let documentChanges = documentKeySet();
```
mikelehen: documentChanges => keys ?

packages/firestore/src/local/local_store.ts
```diff
@@ -183,7 +183,11 @@ export class LocalStore {
   start(): Promise<void> {
     // TODO(multitab): Ensure that we in fact don't need the primary lease.
     return this.persistence.runTransaction('Start LocalStore', false, txn => {
-      return this.startMutationQueue(txn).next(() => this.startQueryCache(txn));
+      return PersistencePromise.waitFor([
+        this.startMutationQueue(txn),
+        this.startQueryCache(txn),
+        this.startRemoteDocumentCache(txn)
```
mikelehen: Hrm...  I'm nervous about doing these in parallel.  Generally we do one thing at a time to keep our "threading" model simple and not have to worry about concurrent execution of things.  Maybe this is safe to parallelize but if you're not 100% sure or there isn't a meaningful perf reason to do it in parallel, I suggest putting it back. :-)

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -16,25 +16,42 @@
...
 import { DocumentKey } from '../model/document_key';
 
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { RemoteDocumentCache } from './remote_document_cache';
+import { SnapshotVersion } from '../core/snapshot_version';
 
 export class MemoryRemoteDocumentCache implements RemoteDocumentCache {
   private docs = maybeDocumentMap();
+  private nextDocumentChanges = documentKeySet();
```
mikelehen: Rename if you rename the method...

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -16,25 +16,42 @@
...
 
-  addEntry(
+  start(transaction: PersistenceTransaction): PersistencePromise<void> {
+    // Technically a no-op but we clear the set of changed documents to mimic
+    // the behavior of the IndexedDb counterpart.
+    //
+    // This is for testing only.
```
mikelehen: What is for testing?  The clearing?  Or the entire "nextDocumentChanges" collection?  Please be more explicit or remove.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -78,17 +78,17 @@ export class RemoteDocumentChangeBuffer {
    * the provided transaction.
    */
   apply(transaction: PersistenceTransaction): PersistencePromise<void> {
-    const changes = this.assertChanges();
+    const docs: MaybeDocument[] = [];
```
mikelehen: Would it make any sense to have RemoteDocumentCache.addEntries() accept a MaybeDocumentMap so we don't need to do this MaybeDocumentMap => array of documents conversion?  I don't care strongly.

packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -76,11 +83,31 @@ function genericRemoteDocumentCacheTests(): void {
       });
   }
 
+  function assertMatches<V>(
```
mikelehen: What is V?

packages/firestore/test/util/helpers.ts
```diff
@@ -129,6 +129,10 @@ export function deletedDoc(
   return new NoDocument(key(keyStr), version(ver));
 }
 
+export function removedDoc(keyStr: string): NoDocument {
```
mikelehen: So deletedDoc() returns a NoDocument with a specific version and removedDoc() returns a NoDocument with a forDeletedDoc() version.  This seems kinda' strange. :-)

How about just a noDoc() function that takes an optional version or something?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -15,35 +15,76 @@
...
 import { LocalSerializer } from './local_serializer';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { RemoteDocumentCache } from './remote_document_cache';
 import { SimpleDbStore, SimpleDbTransaction } from './simple_db';
+import { SnapshotVersion } from '../core/snapshot_version';
 
 export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
+  /** The last key read by `getNextDocumentChanges()`. */
+  private lastReturnedKey = 0;
```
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I am fine with either. Renamed to `getNewDocumentChanges`.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I originally called this an "id", then changes this to "changeKey", then to "autoGeneratedKey". I am fine with either. "id" is probably better for consumers of this data.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Sounds fair.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Done

packages/firestore/src/local/local_serializer.ts
schmidt-sebastian: Done

packages/firestore/src/local/local_serializer.ts
schmidt-sebastian: Done

packages/firestore/src/local/local_store.ts
schmidt-sebastian: Done. I just mostly didn't like the way this looked, but cosmetics shouldn't be the main factor here.

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: Done

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: Well, all of it I guess... But in this case I meant the clearing since start() should not have any data to clear. I removed that part of the comment.

packages/firestore/src/local/remote_document_change_buffer.ts
schmidt-sebastian: I'm leaning to leave it as is for now. `getChanges` returns an array of MaybeDocuments[] and matches this method.

packages/firestore/test/unit/local/remote_document_cache.test.ts
schmidt-sebastian: The value... which is now unused. This used to take a Map<DocumentKey, V> type.

packages/firestore/test/util/helpers.ts
schmidt-sebastian: I'm inclined to leave this as is as it matches the prodo semantics:

- A documentDelete message indicates that the document was deleted. That's why `deletedDoc` takes the version of the delete.

- A documentRemove just means that the document is no longer accessible, and we synthesize a delete locally.

mikelehen: Agree to disagree!  These helpers are just helpers for creating our model objects and shouldn't really have protocol semantics embedded into them...  

That said, I care about this much: --><--  And it's probably a pain to update existing usages. So whatevs.

PR #784, opened 2018-05-08
Refactor RemoteEvent

schmidt-sebastian:
This PR:

- Is way too large. If you guys know of L6 cleverness to split this up, I am more than happy to split this into multiple PRs.

- Changes the entire handling of RemoteEvents.
  - It turns the WatchChangeAggregator into a class that aggregates watch changes. That is, it stays around for the lifetime of the WatchStream. This moves some of the state that was previously passed around (such as pendingResponses) into the aggregator.
  - The aggregator combines all changes until `createRemoteEvent` is called, at which point it resets its updates.
  - The concept of mappings are gone. There is only a TargetChange, that includes the keys of the changed documents, regardless of whether we received an update or a reset.
  - Targets now contain lists of added, modified and removed documents. This is needed for multi-tab and GC.
  - RemoteEvents are now immutable and limbo resolution ( the part that synthesizes document deletes) now happens in the aggregator. Sync Engine still tracks limbo keys, but only for garbage collection.

- The CURRENT state is now a boolean instead of a "state machine".

- Changes the handling of existence filters to use the current state of the watch stream, and not the last snapshot.

- Fixes some spec tests that didn't send CURRENT after a reset and in return, marks RESET queries non-current.

Note that there is a lot of churn in remote_event.test.ts. I need to pass in the set of "existing keys", and they are different for every tests. I can separate out some of this refactor in a different PR.

### DIFF
--- a/packages/firestore/src/remote/watch_change.ts
+++ b/packages/firestore/src/remote/watch_change.ts
@@ -121,8 +121,12 @@ class TargetState {
   private _resumeToken: ProtoByteString = emptyByteString();
   private _current = false;
 
-  /** Whether this target state should be include in the next snapshot. */
-  private _hasPendingChanges = true; // Raise a snapshot for a new target state
+  /**
+   * Whether this target state should be included in the next snapshot. We
+   * initialize to true so that newly-added targets are included in the next
+   * RemoteEvent.
+   */
+  private _hasPendingChanges = true;
 
   /**
    * Whether this target has been marked 'current'.
@@ -402,11 +406,6 @@ export class WatchChangeAggregator {
     objUtils.forEachNumber(this.targetStates, (targetId, targetState) => {
       const queryData = this.queryDataForActiveTarget(targetId);
       if (queryData) {
-        if (targetState.hasPendingChanges) {
-          targetChanges[targetId] = targetState.toTargetChange();
-          targetState.clearPendingChanges();
-        }
-
         if (targetState.current && queryData.query.isDocumentQuery()) {
           // Document queries for document that don't exist can produce an empty
           // result set. To update our local cache, we synthesize a document
@@ -428,6 +427,11 @@ export class WatchChangeAggregator {
               new NoDocument(key, snapshotVersion)
             );
           }
+        }
+
+        if (targetState.hasPendingChanges) {
+          targetChanges[targetId] = targetState.toTargetChange();
+          targetState.clearPendingChanges();
         }
       }
     });
@@ -531,7 +535,7 @@ export class WatchChangeAggregator {
       this.ensureDocumentTargetMapping(key).delete(targetId)
     );
 
-    if (updatedDocument && !this.pendingDocumentUpdates.get(key)) {
+    if (updatedDocument) {
       this.pendingDocumentUpdates = this.pendingDocumentUpdates.insert(
         key,
         updatedDocument


packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -345,186 +446,178 @@ describe('RemoteEvent', () => {
...
-    const wrongStateChange = event.targetChanges[2];
-    event.synthesizeDeleteForLimboTargetChange(
-      wrongStateChange,
-      notSynthesized
-    );
-    expect(event.documentUpdates.get(notSynthesized)).to.not.exist;
-    expect(event.limboDocuments.has(notSynthesized)).to.be.false;
-  });
 
-  it("doesn't synthesize deletes with existing document", () => {
```
schmidt-sebastian: This test is not applicable anymore since we always synthesize deletes for single document lookups (provided we are marked current).

packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -544,59 +637,24 @@ describe('RemoteEvent', () => {
...
-    const update = updateChange.mapping as UpdateMapping;
-    expect(update.addedDocuments.has(existingDoc.key)).to.be.true;
-
-    const existingKeys = documentKeySet().add(existingDoc.key);
-    update.filterUpdates(existingKeys);
-    expect(update.addedDocuments.has(existingDoc.key)).to.be.false;
-    expect(update.addedDocuments.has(newDoc.key)).to.be.true;
-  });
 
-  it("doesn't filter resets", () => {
```
schmidt-sebastian: This tests tests a special case in the ResetMapping that no longer applies.

schmidt-sebastian:
The CI failures are likely due to https://b.corp.google.com/issues/69175563

schmidt-sebastian:
> The CI failures are likely due to b/69175563

This turned out to be an issue with one of the changes in the PR. I assumed that an empty change set for a document query meant that the document didn't exist. But that is not true when a target that has previously seen the document flips from not-current to current.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -248,7 +230,7 @@ export class RemoteStore {
    * from watch so we wait for the ack to process any messages from this target.
    */
   private sendWatchRequest(queryData: QueryData): void {
-    this.recordPendingTargetRequest(queryData.targetId);
+    this.watchChangeAggregator.recordPendingTargetRequest(queryData.targetId);
```
wilhuff: The watch change aggregator wouldn't have to ask if a target is active if it kept track of it itself.

For each `WatchTargetChange.Added` event: add to targetStates
For each `WatchTargetChange.Removed` event: remove from targetStates

One difference is that `WatchTargetChange.Reset` can't remove from targetStates but actually has to clear the entry.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -350,35 +222,318 @@ export class WatchChangeAggregator {
...
+
+  /**
+   * Converts the currently accumulated state into a remote event at the
+   * provided snapshot version. Resets the accumulated changes before returning.
+   */
+  createRemoteEvent(snapshotVersion: SnapshotVersion): RemoteEvent {
+    const targetChanges: { [targetId: number]: TargetChange } = {};
+
+    objUtils.forEachNumber(this.targetStates, (targetId, targetState) => {
+      if (this.isActiveTarget(targetId)) {
```
wilhuff: This makes each RemoteEvent proportional to all active targets rather than just the targets with changes. This will create a ton of bookkeeping writes in the local store that we should avoid if possible.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -350,35 +222,318 @@ export class WatchChangeAggregator {
...
+        targetChanges[targetId] = {
+          current: targetState.current,
+          resumeToken: targetState.resumeToken,
+          snapshotVersion,
+          addedDocuments,
+          modifiedDocuments,
+          removedDocuments
+        };
+
+        targetState.snapshotChanges = snapshotChangesMap();
```
wilhuff: Two other options for how to avoid `CurrentStatusUpdate`:

  * Keep a `lastCurrent` in `TargetState` and promote it here
  * Make an `existingCurrent` callback similar to the `existingKeys` mechanism you added for plumbing in existing prior key state.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -355,20 +339,20 @@ export class SyncEngine implements RemoteSyncer {
...
       );
-      const limboDocuments = documentKeySet().add(limboKey);
-      const event = new RemoteEvent(
-        SnapshotVersion.MIN,
-        {},
-        docMap,
-        limboDocuments
-      );
+      const resolvedLimboDocuments = documentKeySet().add(limboKey);
+      const event = {
```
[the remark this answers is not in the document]
schmidt-sebastian: I've undone and or changed how this is implemented. The types used in this PR are no longer duck-typed interfaces.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: This unfortunately doesn't quite work since a target is only active if the View is still interested in it _and_ if it has no pending ACKs. A target is usually long inactive before the Watch stream finishes sending us the REMOVE messages.

I think we can simplify this a lot if Views ignored updates for inactive targets, but I would rather not make this change right now.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -97,226 +98,97 @@ export class WatchTargetChange {
...
+  addTargetChange(targetChange: WatchTargetChange): void {
     targetChange.targetIds.forEach(targetId => {
-      const change = this.ensureTargetChange(targetId);
+      const targetState = this.ensureTargetState(targetId);
       switch (targetChange.state) {
         case WatchTargetChangeState.NoChange:
           if (this.isActiveTarget(targetId)) {
             // Creating the change above satisfies the semantics of no-change.
-            applyResumeToken(change, targetChange.resumeToken);
+            this.updateResumeToken(targetId, targetChange.resumeToken);
```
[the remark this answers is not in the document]
schmidt-sebastian: I moved this and the pending state counting to TargetState.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -97,226 +98,97 @@ export class WatchTargetChange {
...
-          if (!objUtils.contains(this.pendingTargetResponses, targetId)) {
+          if (targetState.pendingResponses === 0) {
             // We have a freshly added target, so we need to reset any state
-            // that we had previously This can happen e.g. when remove and add
+            // that we had previously. This can happen e.g. when remove and add
             // back a target for existence filter mismatches.
-            change.mapping = new UpdateMapping();
-            change.currentStatusUpdate = CurrentStatusUpdate.None;
-            delete this.existenceFilters[targetId];
+            targetState.current = false;
```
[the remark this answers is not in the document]
schmidt-sebastian: I think this now works just as before, but it is still internal to this class. The `shouldRaise` flag filters the no-op updates. I also added a new test for this ('only raises events for updated targets').

packages/firestore/src/remote/watch_change.ts
schmidt-sebastian: Addressed.

I am also garbage collection these states in TargetChange.REMOVE.

packages/firestore/src/remote/watch_change.ts
schmidt-sebastian: `shouldRaise` is essentially option one, but it also takes into account resume tokens and document updates.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -350,35 +324,253 @@ export class WatchChangeAggregator {
...
+  createRemoteEvent(snapshotVersion: SnapshotVersion): RemoteEvent {
+    const targetChanges: { [targetId: number]: TargetChange } = {};
+
+    objUtils.forEachNumber(this.targetStates, (targetId, targetState) => {
+      if (this.isActiveTarget(targetId)) {
+        const queryData = this.queryDataCallback(targetId);
+
+        if (targetState.shouldRaise) {
+          targetChanges[targetId] = targetState.toTargetChange();
+          targetState.clearChanges();
```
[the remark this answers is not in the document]
wilhuff: Ah okay. It clears `_shouldRaise`, but does not clear `_current`. LGTM

packages/firestore/test/unit/specs/existence_filter_spec.test.ts
```diff
@@ -82,6 +82,48 @@ describeSpec('Existence Filters:', [], () => {
     );
   });
 
+  specTest('Existence filter uses current state of query', [], () => {
```
wilhuff: This test case needs a better name--it's actually testing the positive case where existence filters successfully sync.

schmidt-sebastian: I split this into two test cases and gave them better names:

- "Existence filter match"
- "Existence filter match with pending update"

Note that we did not have any test cases with existence filter matches before.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -374,127 +437,28 @@ export class RemoteStore {
...
-  private handleWatchChangeBatch(
-    snapshotVersion: SnapshotVersion,
-    changes: WatchChange[]
-  ): Promise<void> {
-    const aggregator = new WatchChangeAggregator(
-      snapshotVersion,
-      this.listenTargets,
-      this.pendingTargetResponses
+  private raiseWatchSnapshot(snapshotVersion: SnapshotVersion): Promise<void> {
+    const remoteEvent = this.watchChangeAggregator.createRemoteEvent(
```
mikelehen: Might be worth an assert that snapshotVersion != SnapshotVersion.MIN (which would fail with the current existence filter code, I believe).

packages/firestore/src/core/view.ts
```diff
@@ -332,29 +327,19 @@ export class View {
...
-          );
-      }
+      targetChange.addedDocuments.forEach(
+        key => (this._syncedDocuments = this._syncedDocuments.add(key))
+      );
+      targetChange.modifiedDocuments.forEach(key =>
+        assert(
+          this._syncedDocuments.has(key),
+          `Modified document ${key} not found in view.`
+        )
```
mikelehen: FWIW it seems like we could put this same assert in the removeDocuments loop. I don't care strongly though.

packages/firestore/test/util/helpers.ts
```diff
@@ -233,6 +239,38 @@ export function bound(
...
 
+export function queryData(
+  targetId: TargetId,
+  queryPurpose: QueryPurpose,
+  path: string
+): QueryData {
+  return new QueryData(query(path)._query, targetId, queryPurpose);
+}
+
+export function docInsertRemoteEvent(
```
mikelehen: Why "Insert" ?  Do we use that term elsewhere?  Should this be docAddedRemoteEvent() or something?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -280,6 +280,62 @@ describeSpec('Listens:', [], () => {
...
+        .userListens(collQuery)
+        .watchAcksFull(collQuery, 1000, docA, docB)
+        .expectEvents(collQuery, { added: [docA, docB] })
+        .userUnlistens(collQuery)
+        .watchRemoves(collQuery)
+        // Verify that DocA and DocB exists
+        .userListens(collQuery, 'resume-token-1000')
+        .expectEvents(collQuery, { added: [docA, docB], fromCache: true })
+        .userUnlistens(collQuery)
+        // Now send a document query that produces no results
```
mikelehen: add "from the server" ?

packages/firestore/test/unit/specs/existence_filter_spec.test.ts
```diff
@@ -82,6 +108,34 @@ describeSpec('Existence Filters:', [], () => {
...
+    const doc1 = doc('collection/1', 1000, { v: 1 });
+    const doc2 = doc('collection/2', 2000, { v: 2 });
+    return (
+      spec()
+        .userListens(query)
+        .watchAcksFull(query, 1000, doc1)
+        .expectEvents(query, { added: [doc1] })
+        .watchSends({ affects: [query] }, doc2)
+        // Send an existence filter, but don't send a new global snapshot. We
+        // should not see an event.
```
mikelehen: I'm not sure what this is testing...  The comment implies that it's not raising an event because we haven't gotten a global snapshot, but as your code exists right now it's not raising an event because there's no mismatch...

If there was a mismatch then it'd raise an event without a global snapshot (which I assert is a bug).

packages/firestore/src/remote/watch_change.ts
```diff
@@ -389,33 +438,40 @@ export class WatchChangeAggregator {
 
     let resolvedLimboDocuments = documentKeySet();
 
-    this.documentTargetMapping.forEach((key, targets) => {
-      let isLimboTarget = true;
+    // We extract the set of limbo-only document updates as the GC logic
+    // special-cases documents that do not appear in the query cache.
+    //
+    // TODO(gsoltis): Expand on this comment once GC is available in the JS
+    // client.
```
mikelehen: Thank you for the comment, but I'm still struggling to make sense of this...

1) It seems like your PR changes the logic significantly.  It used to be that when we got a RemoteEvent for a target used for a limbo lookup, then we added the limbo document to this list.  But now the code seems to be going out of the way to make sure the document isn't _also_ in a non-limbo target? Is this change intentional?
2) Since this is only based off pendingDocumentTargetMapping which gets cleared between every remote event, I'm not sure it actually works in a meaningful way.  The document could belong to other non-limbo targets, but those targets didn't happen to get any mapping updates since the last snapshot, right?

If we can't resolve this without Greg, perhaps expand the TODO to flag these potential issues?

packages/firestore/src/remote/watch_change.ts
```diff
@@ -546,33 +611,33 @@ export class WatchChangeAggregator {
...
+   * Resets the state of a Watch target to its initial state (e.g. sets
    * 'current' to false, clears the resume token and removes its target mapping
    * from all documents).
    */
   private resetTarget(targetId: TargetId): void {
+    assert(
+      !this.targetStates[targetId].isPending,
+      'Should only reset active targets'
+    );
     delete this.targetStates[targetId];
```
mikelehen: I feel like we shouldn't completely delete the targetStates entry (or else we should immediately recreate it).  It may not be an issue because the removeDocumentFromTarget() calls below end up re-creating it, but I'm wondering if there's any case where that doesn't happen and as a result we don't include the target in our next RemoteEvent.  E.g. this could happen if we got an existence filter mismatch for an empty target (which would result in us failing to raise a fromCache=true event), except I don't think that's possible...

Since I can't actually come up with a broken case, I leave it up to you to decide if there's an issue warranting any change here.

packages/firestore/test/unit/specs/existence_filter_spec.test.ts
```diff
@@ -49,6 +49,29 @@ describeSpec('Existence Filters:', [], () => {
...
+        .userListens(query)
+        .watchAcksFull(query, 1000, doc1)
+        .expectEvents(query, { added: [doc1] })
+        .userUnlistens(query)
+        .watchRemoves(query)
+        .userListens(query, 'resume-token-1000')
+        .expectEvents(query, { added: [doc1], fromCache: true })
+        // The empty existence filter is ignored since Watch hasn't ACKed the
+        // target
+        .watchFilters([query])
```
mikelehen: I don't think watch would send an existence filter here since it already removed the target. Perhaps the `watchRemoves(query)` should be moved down below this?

packages/firestore/test/unit/specs/existence_filter_spec.test.ts
```diff
@@ -108,19 +131,29 @@ describeSpec('Existence Filters:', [], () => {
...
+        // the snapshot.
+        .watchFilters([query], doc1.key)
+        .watchSends({ affects: [query] }, doc3)
+        .watchSnapshots(2000)
+        // The query result includes doc3, but is marked as "inconsistent"
+        // due to the filter mismatch
+        .expectEvents(query, { added: [doc3], fromCache: true })
+        .expectActiveTargets({ query, resumeToken: '' })
+        .watchRemoves(query) // Acks removal of query
+        .watchAcksFull(query, 2000, doc1, doc2, doc3)
```
mikelehen: FWIW, it would probably be a more real-world use case if doc2 was gone now, since the existence filter apparently indicated it was no longer present.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -97,251 +98,238 @@ export class WatchTargetChange {
...
           // We need to decrement the number of pending acks needed from watch
           // for this targetId.
           this.recordTargetResponse(targetId);
+          if (!targetState.isPending) {
+            delete this.targetStates[targetId];
+          }
           assert(
             !targetChange.cause,
             'WatchChangeAggregator does not handle errored targets'
           );
```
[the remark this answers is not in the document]
mikelehen: only surfacing them back to the user...  What codepath is going to remove the WatchChangeAggregator.targetStates entry for a target if it's rejected for permission_denied?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -457,6 +387,36 @@ export class RemoteStore implements WatchMetadataProvider {
       }
     });
 
+    // Re-establish listens for the targets that have been invalidated by
+    // existence filter mismatches.
+    remoteEvent.targetResets.forEach(targetId => {
```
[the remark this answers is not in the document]
schmidt-sebastian: I was pondering about this earlier and didn't like any names I cam up with. I liked having this start with "target" to match "targetChanges". I renamed it to "targetMismatches" for now.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -338,111 +332,47 @@ export class RemoteStore implements WatchMetadataProvider {
...
+    assert(
+      !snapshotVersion.isEqual(SnapshotVersion.MIN),
+      "Can't raise event for unknown SnapshotVersion"
+    );
     const remoteEvent = this.watchChangeAggregator.createRemoteEvent(
       snapshotVersion
     );
 
-    // Update in-memory resume tokens. LocalStore will update the
+    // Update in-memory resume tokens and re-issue queries that have been
```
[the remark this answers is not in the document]
schmidt-sebastian: Sorry, I forgot to revert this. Fixed.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -350,35 +338,252 @@ export class WatchChangeAggregator {
...
+          ) {
+            this.documentUpdates = this.documentUpdates.insert(
+              key,
+              new NoDocument(key, snapshotVersion)
+            );
+
+            // While we don't add the document to a target, we potentially
+            // need to mark it as a resolved limbo key. This requires us
+            // to add the document to the list of document target mappings.
+            this.ensureDocumentTargetMapping(key);
```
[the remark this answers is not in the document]
schmidt-sebastian: I updated the code.

FWIW, the last version of this code didn't trigger limbo resolution before since we didn't write down that the target matched. There was nothing written down in the cache if we got no results back (which meant that the next time around we would not raise events from cache).

The test scenario that b/79422585 asked for is already handled by "Synthesizes deletes for missing document'.

packages/firestore/src/remote/watch_change.ts
```diff
@@ -535,14 +531,18 @@ export class WatchChangeAggregator {
       this.ensureDocumentTargetMapping(key).delete(targetId)
     );
 
-    if (updatedDocument) {
+    if (updatedDocument && !this.pendingDocumentUpdates.get(key)) {
```
mikelehen: Hrm. Not sure why you added the `!this.pendingDocumentUpdates.get(key)` check.  If it's correct, please add a comment that convinces me.

PR #786, opened 2018-05-08
Update Observer and Messaging types to match internal types

mmermerkaya:


packages/firebase/index.d.ts
```diff
@@ -23,10 +25,10 @@ declare namespace firebase {
     stack?: string;
   }
 
-  interface Observer<V, E> {
-    complete(): any;
-    error(error: E): any;
-    next(value: V | null): any;
+  interface Observer<T> {
```
jshcrowthe: this part of the change is breaking and we just shipped our `5.0.0` release.

You'll need to remove this portion.

mmermerkaya: It's not a breaking change, it's a bug fix. If they are annotating with `any`, then they are using wrong types and their application will not work as expected. In the [implementation](https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/subscribe.ts#L17), `ErrorFn` takes an `Error` instance as its argument, not `any`. If they expect it to be `any`, they're going to have a bad time.

Changed it to a generic with a default value of Error, so at least users can do `firebase.observer<any>` (and second generic will default to `Error`) as well as `firebase.Observer<any, any>`.

mmermerkaya: Sure you could, but if you tried to call [.message](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message) on a `number` or a `string`, you'd get a runtime error :)

PR #793, opened 2018-05-09
Fix messaging being optional and isSupported type definitions

mmermerkaya:
Fixes TS thinking `firebase.messaging()` method is optional after importing 'firebase/messaging' (removes the `?`). This should be done in other packages as well, unless there's a reason for the discrepancy between importing from 'firebase' and from 'firebase/packageName' that I don't know about.

Modifies `messaging.isSupported()` so it is a namespace export and not a property on the factory function, which is not actually exported. Fixes types accordingly.


packages/messaging/index.ts
```diff
@@ -67,13 +60,13 @@ registerMessaging(firebase as _FirebaseNamespace);
  */
 declare module '@firebase/app-types' {
   interface FirebaseNamespace {
-    messaging?: {
-      (app?: FirebaseApp): types.FirebaseMessaging;
-      Messaging: typeof types.FirebaseMessaging;
+    messaging: {
```
dwoffinden: Soo... @firebase/messaging *is* optional; you can use @firebase/database etc. but not messaging just fine.
firebase the metapackage bundles everything together, cool.
But if you *just* depend on @firebase/messaging, it would monkey-patch @firebase/app-types to add itself optionally? weird

packages/messaging/test/index.test.ts
```diff
@@ -18,9 +18,12 @@ import { expect } from 'chai';
 import { sandbox, SinonSandbox, SinonStub } from 'sinon';
 
 import { FirebaseApp } from '@firebase/app-types';
-import { _FirebaseNamespace } from '@firebase/app-types/private';
+import {
+  _FirebaseNamespace,
+  FirebaseServiceFactory
+} from '@firebase/app-types/private';
```
dwoffinden: It'd be really nice to have a test that works like [the docs](https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/README.md) suggests a user to, i.e. 

```
import '@firebase/messaging'
// Do stuff w/ `firebase` and `firebase.messaging`
```
rather than poking at the internals like this :/

mmermerkaya: Yeah, I agree. The tests need an overhaul (not just here). I'll get to that soon™.

PR #795, opened 2018-05-09
Saucelabs

yifanyang:
Enable multi-browser test for database and firestore.

Skipping below 4 flaky / failing test cases for now:

QueryListener
    ✗ raises error event
  AsyncQueue
    ✗ handles failures
Crawler Support
    ✗ "before each" hook for "set() is a no-op"
Transaction Tests
    ✗ server values: local timestamp should eventually (but not immediately) match the server with txns


config/karma.saucelabs.js
```diff
@@ -119,19 +114,50 @@ module.exports = function(config) {
...
+    retryLimit: 0,
 
     // concurrency: 10,
 
+    client: {
+      mocha: {
+        timeout: 20000,
+        retries: 3,
+        grep:
+          'Crawler Support|local timestamp|raises error event|handles failures',
```
mikelehen: As somebody viewing this code without context, I'll be curious about why these tests are disabled and wonder if/when they should be re-enabled.  Can you please log bugs and link them here?  E.g.:

// TODO(b/123, b/456, ...): The tests matched by this regex are currently failing in the saucelabs cross-browser tests. They're disabled until we have a chance to investigate / fix them.

I'd also try to make this regex as specific as possible so that we don't accidentally match unrelated tests (or if we could somehow exclude the tests directly at the test definitions, that would be better than having a golbal regex).  Is it possible to include the full test name (including the describe() part)?  e.g. "EventManager raises error event" instead of just "raises error event" ?

config/karma.saucelabs.js
```diff
@@ -119,19 +114,50 @@ module.exports = function(config) {
...
+        invert: true
+      }
+    },
+
+    browserConsoleLogOptions: { terminal: false },
```
mikelehen: Does this mean we won't get logs from these tests?  That'll make it harder to diagnose failures.  If nothing else, is a comment warranted?

config/karma.saucelabs.js
```diff
@@ -119,19 +114,50 @@ module.exports = function(config) {
...
+
+    specReporter: {
+      maxLogLines: 5,
+      suppressErrorSummary: true,
```
mikelehen: Out of curiosity, why do we want to suppress the error summary? That sounds useful offhand.

config/karma.saucelabs.js
```diff
@@ -119,19 +114,50 @@ module.exports = function(config) {
...
+    },
+
     sauceLabs: {
       tunnelIdentifier: process.env.TRAVIS_JOB_NUMBER,
       username: process.env.SAUCE_USERNAME,
       accessKey: process.env.SAUCE_ACCESS_KEY,
-      startConnect: true
+      startConnect: true,
+      connectOptions: {
+        noSslBumpDomains: 'all'
```
mikelehen: This looks interesting!  Worth a comment?  I'm also curious if 'all' is the best option or if this should just be scoped to *.firebaseio.com or something?

config/karma.saucelabs.js
```diff
@@ -119,19 +114,50 @@ module.exports = function(config) {
...
     port: 9876,
 
-    retryLimit: 3,
+    retryLimit: 0,
 
     // concurrency: 10,
 
+    client: {
+      mocha: {
+        timeout: 20000,
```
mikelehen: I assume these values are copied from the main karma.js or similar?  If we can't share code, perhaps add a comment referencing where these were copied from so we know how these values were chosen and can try to keep things in sync over time.

mikelehen:
Also, github is complaining "This branch is out-of-date with the base branch"  Can you please sync?

config/mocha.base.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright 2017 Google Inc.
```
mikelehen: 2018

config/karma.saucelabs.js
```diff
@@ -156,7 +159,8 @@ module.exports = function(config) {
       accessKey: process.env.SAUCE_ACCESS_KEY,
       startConnect: true,
       connectOptions: {
-        noSslBumpDomains: 'all'
+        // https://wiki.saucelabs.com/display/DOCS/Troubleshooting+Sauce+Connect
```
mikelehen: Perhaps:

Realtime Database uses WebSockets to connect to firebaseio.com so we have to set noSslBumpDomains. See https://wiki.saucelabs.com/display/DOCS/Troubleshooting+Sauce+Connect for more details.

yifanyang: Ooops, I didn't mean to publish this...

Theoretically, we only need to whitelist *.firebaseio.com instead of all, but I've noticed that requests to firestore.googleapis.com increased which maybe the cause of many failures in Safari and Edge.

I'll try a few more things and finalize my code change soon.

config/karma.saucelabs.js
yifanyang: Travis has a limit for total number of log lines at 10k. Since we are running 4 browsers now and have more to come (IE11 and mobile browsers), we basically generate a lot more logs, and the summary of test results can be only seen in the raw log. However, it's one more click away and it doesn't display some character correctly. We can still access the console log from saucelabs test result (e.g. https://saucelabs.com/beta/tests/f5ef37f915cb4ae8bc500b9b4ab3ba62).

So I sort of prefer to reduce the log verbosity here. But this is purely a cosmetic reason. Turned console log on now as we are not hitting the limit yet.

config/karma.saucelabs.js
yifanyang: We actually have another reporter below that provides a better summary of the run, as which test failed in which browser. So the summary from this reporter is not of much use. But there's no harm to turn it on.

config/karma.saucelabs.js
yifanyang: Commented below.

config/karma.saucelabs.js
yifanyang: Created a mocha base config file which all karma configs would read from. This new file will be the single source of truth.

config/mocha.base.js
yifanyang: Done. Good catch!

config/karma.saucelabs.js
yifanyang: Didn't figure out the root cause behind it. Only noticed that test time can easily exceed 30 mins with only SSL Bumping for '*.firebaseio.com' disabled, rather than ~22mins in case of 'all'.

yifanyang:
Have replied to each of the comments above. Thanks for your review!

jshcrowthe:
Seems like we had a bad merge here.

yifanyang:
I'm also thinking about setup something that can track failures and calculate flaky rate of different cases for us, but I'm not sure how to do it now.

jshcrowthe:
Sounds good @yifanyang, for now lets get this in as running the tests in all browsers (even if there are still some flakes) is better than not.

mikelehen:
So if we accidentally regress firestore so it fails on some browser (e.g. Edge), how will we notice?  Is there a notification somewhere, or would I have to actually click into the travis run to see it?  FWIW I never look at travis runs unless they fail my PR, and so it's unlikely we would notice. :-/

yifanyang:
We still have our original per package ChromeHeadless test running as blocking test. Any failure in that test will fail the entire build.

In my experiment runs, a lot more cases have failed at least once in some browser. If we make it blocking, that would be quite noisy and disruptive to our build process.

What I'm thinking now is to have sth that records failures and measures the flakiness of failed cases, so that when we want to investigate / fix them, we know which case to start with.

mikelehen:
Yeah, I get that and I don't want to break the build with flakiness...  but until we have /some/ sort of monitoring for the failures, they almost might as well not be running at all.  Building out infrastructure to measure flakiness sounds cool but not something we can do in the short-term.  If we could just get an email once a day for any failures in master, at least we could keep an eye on it and if a bunch of tests start failing on some browser then we would have an opportunity to notice.

For context, we regressed Firestore's Edge support not that long ago and only noticed because the auth team happened to manually test on Edge and tell us.  I'd like to not repeat that.

yifanyang:
I agree with this. I've opened pr #867 that sends out an email notification with test result summary after failures. Please help review it. Thanks!

PR #808, opened 2018-05-10
 Speed up validation tests

schmidt-sebastian:
My Remote Store PR is failing CI because of `Field paths must not have invalid segments`. The only reason I could come up with it that it times out because it issues a document write and a subsequent get for every iteration. This PR removes the document set.

### DIFF
--- a/scripts/release/cli.js
+++ b/scripts/release/cli.js
@@ -207,7 +207,11 @@ const { argv } = require('yargs');
     /**
      * Release new versions to NPM
      */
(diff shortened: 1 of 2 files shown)


scripts/release/cli.js
```diff
@@ -207,7 +207,11 @@ const { argv } = require('yargs');
     /**
      * Release new versions to NPM
      */
-    await publishToNpm(updates, releaseType, argv.canary ? 'canary': 'default');
+    await publishToNpm(
```
schmidt-sebastian: Note that this shows up even with a recent run of `yarn install`.

PR #812, opened 2018-05-11
Add listen spec tests

schmidt-sebastian:
This add some spec tests that tests behavior that I rely on in the RemoteEvent refactor. It also fixes existing test to send CURRENT after each RESET.

This PR also makes sending a `documentDelete` (rather than a `documentRemove`) message possible via the spec tests.

### DIFF
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -16,7 +16,7 @@
 
 import { Query } from '../../../src/core/query';
 import { Code } from '../../../src/util/error';
-import { doc, filter, path } from '../../util/helpers';
+import { deletedDoc, doc, filter, path } from '../../util/helpers';
 
 import { describeSpec, specTest } from './describe_spec';
 import { spec } from './spec_builder';
@@ -77,6 +77,19 @@ describeSpec('Listens:', [], () => {
       );
     }
   );
+
+  specTest('Does not raise event for initial document delete', [], () => {
+    const query = Query.atPath(path('collection'));
+    const missingDoc = deletedDoc('collection/a', 1000);
+    return spec()
+      .userListens(query)
+      .watchAcks(query)
+      .watchSends({ removed: [query] }, missingDoc)
+      .watchSnapshots(1000)
+      .watchCurrents(query, 'resume-token-2000')
+      .watchSnapshots(2000)
+      .expectEvents(query, { fromCache: false });
+  });
 
   specTest(
     'Will process removals without waiting for a consistent snapshot',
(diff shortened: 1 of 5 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -78,6 +78,19 @@ describeSpec('Listens:', [], () => {
     }
   );
 
+  specTest('Does not raise event for initial document delete', [], () => {
+    const query = Query.atPath(path('collection'));
+    const missingDoc = deletedDoc('collection/a', 1000);
+    return spec()
+      .userListens(query)
+      .watchAcks(query)
+      .watchSends({ removed: [query] }, missingDoc)
```
mikelehen: Hrm...  why is watch removing a document from the target that was never included?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -412,7 +416,7 @@ export class SpecBuilder {
 
   watchSends(
     targets: { affects?: Query[]; removed?: Query[] },
-    ...docs: Document[]
+    ...docs: MaybeDocument[]
```
mikelehen: Hrm...  I'm not sure this makes sense. Watch doesn't have a way to send "deleted" documents via the protos, does it?

schmidt-sebastian: It does when it knows when a document was deleted: https://github.com/firebase/firebase-js-sdk/blob/fa5d157751ae64524691150f6bb512b94d86b6f6/packages/firestore/src/remote/watch_change.ts#L63

packages/firestore/test/unit/specs/listen_spec.test.ts
schmidt-sebastian: This test basically follows the request trace I got from the validation test:

```
addTarget: 
   { documents: 
      { documents: 
         [ 'projects/node-review-with-fix/databases/(default)/documents/test/V9w4ini0kixUMxshKuhG' ] },
     targetId: 4 } }
Firestore (5.0.2) [Connection]: GRPC stream received: { response_type: 'targetChange',
  targetChange: 
   { targetChangeType: 'ADD',
     targetIds: [ 4 ],
     cause: null,
     resumeToken: <Buffer >,
     readTime: null },
  documentChange: null,
  documentDelete: null,
  documentRemove: null,
  filter: null }
Firestore (5.0.2) [Connection]: GRPC stream received: { response_type: 'documentDelete',
  targetChange: null,
  documentChange: null,
  documentDelete: 
   { document: 'projects/node-review-with-fix/databases/(default)/documents/test/V9w4ini0kixUMxshKuhG',
     removedTargetIds: [ 4 ],
     readTime: { seconds: '1526067903', nanos: 787493000 } },
  documentRemove: null,
  filter: null }
```

My resident Watch expert says "there is a good reason for this, but I don't remember".

mikelehen: So I didn't realize Watch does this...  Maybe add a comment to make it clear that watch really does this and so you're testing this intentionally.

// To indicate the document doesn't exist, watch sends a DocumentDelete message as if the document previously existed and now is being deleted/removed from the target.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -644,7 +645,13 @@ abstract class TestRunner {
         );
       });
     } else if (watchEntity.doc) {
-      const d = doc(watchEntity.doc[0], watchEntity.doc[1], watchEntity.doc[2]);
+      let d;
+      if (watchEntity.doc[2]) {
```
mikelehen: This code is unreadable. :-)  Can you add:

```
const [key, version, data] = watchEntity.doc;
```
and use the named variables?

schmidt-sebastian: I was going to use this part for my Go readability, but I guess not then.

packages/firestore/test/unit/specs/listen_spec.test.ts
schmidt-sebastian: Added.

PR #821, opened 2018-05-15
Enable IE11 testing on SauceLabs.

yifanyang:


packages/functions/src/serializer.ts
```diff
@@ -45,7 +45,7 @@ export class Serializer {
     if (data === true || data === false) {
       return data;
     }
-    if (toString.call(data) === '[object String]') {
+    if (Object.prototype.toString.call(data) === '[object String]') {
```
yifanyang: Interesting fact: `toString === Object.prototype.toString` resolves to true in all the other browsers except IE11.

config/babel.test.json
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
```
jshcrowthe: Instead of using the individual plugins for each syntax, I think it would make more sense to use the env preset and set it to our browser matrix. Take a look at: https://github.com/babel/babel/tree/master/packages/babel-preset-env

yifanyang: I assume the motivation of this pr is that we want to get IE11 tests running first, but eventually we don't want to depend on babel for our code to work? I just feel it shows the gap more clearly this way. But I don't know, what do you think?

jshcrowthe: I mean, for right now this is great. Using the preset-env plugin just gives us more confidence that we properly transpile **all** the syntaxes from our tests that would break some of our non-standard browsers.

jshcrowthe: Your call.

yifanyang: I agree. Done. Using @babel/preset-env that targets our browsers matrix now.

Store browsers in a standard .browserslistrc file now. Using babel 7 now, as only this new version knows to search for .browserslistrc file while doing karma preprocessing.

config/karma.saucelabs.js
```diff
@@ -86,11 +85,11 @@ function appiumLauncher(
...
-  Chrome_Windows: seleniumLauncher('chrome', 'Windows 10', 'latest'),
-  Firefox_Windows: seleniumLauncher('firefox', 'Windows 10', 'latest'),
-  Safari_macOS: seleniumLauncher('safari', 'macOS 10.13', 'latest'),
-  Edge_Windows: seleniumLauncher('MicrosoftEdge', 'Windows 10', 'latest'),
-  IE_Windows: seleniumLauncher('internet explorer', 'Windows 10', 'latest')
+  Chrome_Windows: seleniumLauncher('chrome', 'Windows 10', '66.0'),
+  Firefox_Windows: seleniumLauncher('firefox', 'Windows 10', '60.0'),
+  Safari_macOS: seleniumLauncher('safari', 'macOS 10.13', '11.0'),
+  Edge_Windows: seleniumLauncher('MicrosoftEdge', 'Windows 10', '17.17134'),
+  IE_Windows: seleniumLauncher('internet explorer', 'Windows 10', '11.103')
```
yifanyang: BrowsersList (used by Babel) and SauceLabs have different definition of "latest". We'll use specific version number here and in .browserslistrc file.

It'd be good if we can derive one file from the other, but BrowsersList and SauceLabs use different formats to represent a browser. I don't know if that's worth the effort.

jshcrowthe:
Looks great @yifanyang. Nicely done 👍

yifanyang:
Great! Thanks!

Issue #720, opened 2018-04-24
Firestore with offline persistence incorrectly throws 'multiple tabs open' error


gizmodus:
### Describe your environment

  * Operating System version: Android 6.0.1 - Android System WebView 66.0.3359.126
  * Firebase SDK version: 4.12.1-1
  * Firebase Product: firestore

### Describe the problem

I'm using Firestore with offline persistence enabled in an Ionic project that runs on Android as a standalone app (Android System WebView). Since I updated Android System WebView to the latest version, firestore is throwing the following error: 'Unhandled Promise rejection: There is another tab open with offline persistence enabled. Only one such tab is allowed at a time. The other tab must be closed or persistence must be disabled.'

As the the application runs in standalone mode there is of course no other tab open. Many users confirmed that they have the same issue on their phone.
The problem can be fixed if the Android System WebView updates are uninstalled.

#### Steps to reproduce:

I can not reproduce the problem as it only appears in the bundled app on Android. In Desktop Chrome the problem does not appear.

Log output with log level set to 'debug':

```
[2018-04-23T19:30:09.773Z]  @firebase/firestore: Firestore (4.10.1) [SimpleDb]: Opening database: firestore/[DEFAULT]/swim-coach-62d6b/main
vendor.js:142437 
[2018-04-23T19:30:09.929Z]  @firebase/firestore: Firestore (4.10.1) [SimpleDb]: GET owner owner {"ownerId":"X4UFPtF3GDvFzKd84JCa","leaseTimestampMs":1524511458400}
vendor.js:142437 
[2018-04-23T19:30:09.943Z]  @firebase/firestore: Firestore (4.10.1) [IndexedDbPersistence]: No valid owner. Acquiring owner lease. Current owner: {"ownerId":"X4UFPtF3GDvFzKd84JCa","leaseTimestampMs":1524511458400} New owner: {"ownerId":"BROWdCcPk9QpH4m5McjF","leaseTimestampMs":1524511809942}
vendor.js:142437 
[2018-04-23T19:30:09.953Z]  @firebase/firestore: Firestore (4.10.1) [SimpleDb]: PUT owner owner {"ownerId":"BROWdCcPk9QpH4m5McjF","leaseTimestampMs":1524511809942}
vendor.js:142437 
[2018-04-23T19:30:10.005Z]  @firebase/firestore: Firestore (4.10.1) [IndexedDbPersistence]: Starting transaction: Start LocalStore
vendor.js:142437 
[...]

PR #823, opened 2018-05-15
 Treating onabort as an error

schmidt-sebastian:
This PR removes our special handling of `onabort` for IndexedDb transactions and treats aborts as an error (as `onabort` is called for quota exceeded). We currently don't abort transactions in our SDK and hence removing this handling seems like the proper thing to do.

I tested this in a test app (not using the Firestore SDK) and saw this error get set during both the initial schema migration as well as during normal operation. The error is a DOMException with `{ code: 22, name: "QuotaExceededError" }`. To get there, I had to write 32 GB of random data to IndexedDb :)

The 'still propagates error if you throw after aborting an exception' test no longer makes sense since the abort already rejects the transaction.

Fixes: https://github.com/firebase/firebase-js-sdk/issues/720

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -235,7 +235,7 @@ export class FirestoreClient {
         error.code === Code.FAILED_PRECONDITION ||
         error.code === Code.UNIMPLEMENTED
       );
-    } else if (error instanceof DOMException) {
+    } else if (typeof DOMException !== 'undefined' && error instanceof DOMException) {
       return error.code === QUOTA_EXCEEDED;
     }
 
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -57,7 +57,7 @@ import { ViewSnapshot } from './view_snapshot';
 const LOG_TAG = 'FirestoreClient';
 
 /** The DOMException code for quota exceeded. */
-const QUOTA_EXCEEDED = 22;
+const DOM_EXCEPTION_QUOTA_EXCEEDED = 22;
 
 /**
  * FirestoreClient is a top-level class that constructs and owns all of the
@@ -229,6 +229,10 @@ export class FirestoreClient {
     }
   }
 
+  /**
+   * Decides whether the provided error allows us to gracefully disable
+   * persistence (as opposed to crashing the client).
+   */
   private canFallback(error: FirestoreError | DOMException): boolean {
     if (error instanceof FirestoreError) {
       return (
@@ -239,7 +243,7 @@ export class FirestoreClient {
       typeof DOMException !== 'undefined' &&
       error instanceof DOMException
     ) {
-      return error.code === QUOTA_EXCEEDED;
+      return error.code === DOM_EXCEPTION_QUOTA_EXCEEDED;
     }
 
     return true;
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -229,11 +229,11 @@ export class FirestoreClient {
     }
   }
 
-  private canFallback(error: FirestoreError|DOMException): boolean {
+  private canFallback(error: FirestoreError | DOMException): boolean {
     if (error instanceof FirestoreError) {
       return (
-          error.code === Code.FAILED_PRECONDITION ||
-          error.code === Code.UNIMPLEMENTED
+        error.code === Code.FAILED_PRECONDITION ||
+        error.code === Code.UNIMPLEMENTED
       );
     } else if (error instanceof DOMException) {
       return error.code === QUOTA_EXCEEDED;
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -56,6 +56,9 @@ import { ViewSnapshot } from './view_snapshot';
 
 const LOG_TAG = 'FirestoreClient';
 
+/** The DOMException code for quota exceeded. */
+const QUOTA_EXCEEDED = 22;
+
 /**
  * FirestoreClient is a top-level class that constructs and owns all of the
  * pieces of the client SDK architecture. It is responsible for creating the
@@ -199,7 +202,7 @@ export class FirestoreClient {
     if (usePersistence) {
       return this.startIndexedDbPersistence()
         .then(persistenceResult.resolve)
-        .catch((error: FirestoreError) => {
+        .catch(error => {
           // Regardless of whether or not the retry succeeds, from an user
           // perspective, offline persistence has failed.
           persistenceResult.reject(error);
@@ -226,11 +229,17 @@ export class FirestoreClient {
     }
   }
 
-  private canFallback(error: FirestoreError): boolean {
-    return (
-      error.code === Code.FAILED_PRECONDITION ||
-      error.code === Code.UNIMPLEMENTED
-    );
+  private canFallback(error: FirestoreError|DOMException): boolean {
+    if (error instanceof FirestoreError) {
+      return (
+          error.code === Code.FAILED_PRECONDITION ||
+          error.code === Code.UNIMPLEMENTED
+      );
+    } else if (error instanceof DOMException) {
+      return error.code === QUOTA_EXCEEDED;
+    }
+
+    return true;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/firestore/test/unit/local/simple_db.test.ts
```diff
@@ -128,11 +133,13 @@ describe('SimpleDb', () => {
...
-        txn.abort(); // JUST KIDDING!
-      });
-    });
+    expect(
+      runTransaction((store, txn) => {
+        return store.put(dummyUser).next(() => {
+          txn.abort();
+        });
+      })
+    ).to.eventually.be.rejectedWith('The IndexedDb transaction was aborted.');
```
mikelehen: FWIW, I don't think aborting a transaction should generally be an error case, so I'd maybe add a todo...

// TODO(mrschmidt): If in the future we end up adding code that intentionally aborts transactions we may need to turn this into a non-error case.

yarn.lock
```diff
@@ -233,10 +233,10 @@
     "@types/estree" "*"
 
 "@types/body-parser@*":
-  version "1.16.8"
-  resolved "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.16.8.tgz#687ec34140624a3bec2b1a8ea9268478ae8f3be3"
+  version "1.17.0"
```
mikelehen: uh, wow...  there was a lot of churn in this file.  All of that is from adding chai-as-promised, right?

schmidt-sebastian:
> By the way, from your comment I can't 100% tell if you verified that in the case of a quota exceeded error enablePersistence() will now return a rejected promise but leave the client functional (just with persistence disabled)... Can you confirm?

We need to special-case this error in `canRecover()`, which is something that I hadn't done. I updated the PR.

I also fixed an issue with our error bubbling. Before, we would call `abort()`, which then rejected the Promise before we had a chance to reject it with the error that we threw.

packages/firestore/test/unit/local/simple_db.test.ts
schmidt-sebastian: Done.

yarn.lock
schmidt-sebastian: This took all morning to resolve, but I am now happy to report that the diff is now as expected.

packages/firestore/src/local/simple_db.ts
```diff
@@ -263,19 +260,30 @@ export class SimpleDbTransaction {
...
   }
 
-  abort(): void {
+  abort(error?: Error): void {
+    if (error) {
+      this.completionPromise.reject(error);
+    }
+
     if (!this.aborted) {
       debug(LOG_TAG, 'Aborting transaction.');
```
mikelehen: perhaps log the error?

packages/firestore/src/local/simple_db.ts
```diff
@@ -147,15 +148,14 @@ export class SimpleDb {
       .catch(error => {
         // Abort the transaction if there was an
         // error.
-        transaction.abort();
-        return PersistencePromise.reject(error);
+        transaction.abort(error);
```
mikelehen: Can you rewrap the comment while you're here?

packages/firestore/src/local/simple_db.ts
```diff
@@ -246,11 +246,8 @@ export class SimpleDbTransaction {
 
   /**
    * A promise that resolves with the result of the IndexedDb transaction.
-   *
-   * Note: A transaction explicitly aborted via abort() is considered successful
-   * and this promise will resolve as successful.
    */
-  readonly completionPromise: Promise<void>;
+  readonly completionPromise: Deferred<void>;
```
mikelehen: Err...   exposing the raw Deferred<> as part of the public API of SimpleDbTransaction seems like leaking implementation details. External code shouldn't be able to call `transaction.completionPromise.resolve()` and external code shouldn't have to do `transaction.completionPromise.promise` to get the promise.  I would:

1. Rename completionPromise to completionDeferred and make it private.
2. Add a public completionPromise getter that returns completionDeferred.promise.

packages/firestore/src/local/simple_db.ts
```diff
@@ -263,19 +260,30 @@ export class SimpleDbTransaction {
...
+      this.completionPromise.resolve();
+    };
+    this.transaction.onabort = () => {
+      // TODO(mrschmidt): If in the future we end up adding code that
+      // intentionally aborts transactions we may need to turn this into a
+      // non-error case.
+      const error =
+        transaction.error ||
+        new Error('The IndexedDb transaction was aborted.');
+      this.completionPromise.reject(error);
```
mikelehen: So actually...  could we change this to:

```
if (transaction.error) {
  this.completionPromise.reject(error);
}
```
?

I think that would mean:
1. You can call transaction.abort() [no error] to gracefully abort the transaction but not trigger any error handling.
2. It would solve a wart where as-is, I think if you call transaction.abort(error), then we'll end up calling completionPromise.reject() in abort() and then again here which seems sloppy / confusing.

packages/firestore/test/unit/local/simple_db.test.ts
mikelehen: Per my other comment, I think we can make this work again without generating an error?

packages/firestore/test/unit/local/simple_db.test.ts
```diff
@@ -142,18 +149,13 @@ describe('SimpleDb', () => {
...
+          throw new Error('Generated error');
         });
-      });
-    } catch (error) {
-      expect(error.message).to.equal('error');
-      gotError = true;
-    }
-    expect(gotError).to.equal(true);
+      })
+    ).to.eventually.be.rejectedWith('Generated error');
```
mikelehen: FWIW, looking at the chai-as-promised examples, they don't tend to use expect() so this would become:

```
await runTransaction(...).should.eventually.be.rejectedWith('Generated error');
```

I don't care strongly though, so feel free to leave as-is.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -226,11 +229,20 @@ export class FirestoreClient {
     }
   }
 
-  private canFallback(error: FirestoreError): boolean {
-    return (
-      error.code === Code.FAILED_PRECONDITION ||
-      error.code === Code.UNIMPLEMENTED
-    );
+  private canFallback(error: FirestoreError | DOMException): boolean {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Done

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Done and done.

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Done

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Done

packages/firestore/test/unit/local/simple_db.test.ts
schmidt-sebastian: Yeah, with the conditional check for `transaction.error`, this once again becomes possible.

packages/firestore/test/unit/local/simple_db.test.ts
schmidt-sebastian: I changed this to use `expect` since I found it somewhat questionable to just extend the Promise implementation. Using `expect` hides this a little bit.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -56,6 +56,9 @@ import { ViewSnapshot } from './view_snapshot';
 
 const LOG_TAG = 'FirestoreClient';
 
+/** The DOMException code for quota exceeded. */
+const QUOTA_EXCEEDED = 22;
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore/src/local/simple_db.ts
```diff
@@ -263,21 +259,36 @@ export class SimpleDbTransaction {
...
+  abort(error?: Error): void {
+    if (error) {
+      this.completionDeferred.reject(error);
+    }
+
     if (!this.aborted) {
-      debug(LOG_TAG, 'Aborting transaction.');
+      debug(
+        LOG_TAG,
+        'Aborting transaction: %s',
```
mikelehen: Uhh...  Does `%s` work in JavaScript?

schmidt-sebastian: It does :)

Issue #824, opened 2018-05-15
Realtime Firestore query results in error for "docChanges" 


Thkasis:
### [REQUIRED] Describe your environment

  * Operating System version: Windows, Google Chrome
  * Firebase SDK version: 5.0.1
  * Firebase Product: Firestore

Simple query for realtime onSnapshot() results in error when trying to work with snapshot.docChanges. 
I don't see any doc.docChanges defined as it was in the previous version (4.9.1) that I was using and I get this error: 
snapshot.docChanges.forEach is not a function. Logging snapshot.docChanges is not an array of changes as it was before and instead shows the prototype method. 

#### Steps to reproduce:
Simply query data either using get() or realtime onSnapshot() similar to Firestore documents. 

For realtime query: 

> firebase.firestore(firebase.app("ProjectFirst")).collection('test').where("flg", "==", "i").onSnapshot(function(snapshot) {
        //there is data in database and I see them using get() method for instance
        snapshot.docChanges.forEach(function(change) {
             ...
        });
});

when I log snapshot.docChanges I see this instead of the array of results : 
 
ƒ (e){validateOptionNames("QuerySnapshot.docChanges",e,["includeMetadataChanges"]),e&&validateNamedOptionalType("QuerySnapshot.docChanges","boolean","includeMetadataChanges",e.includeMetadataChanges);v…

mikelehen:
@Thkasis Sorry for the trouble! We changed docChanges into a method in 5.0.0 so you must now do `snapshot.docChanges()` instead of `snapshot.docChanges`. You can find more details in our release notes here: https://firebase.google.com/support/release-notes/js#5.0.0

Thanks and sorry!

Thkasis:
@mikelehen, not a problem, thanks for the update. It would definitely be helpful to have this in the documentation instead if it's supposed to be the standard from now on as documentation still shows docChanges.forEach() 

Thanks for all the good work.
Cheers

mikelehen:
Thank you! I see that indeed https://firebase.google.com/docs/firestore/query-data/listen still has the old syntax. Whoops! We'll get that updated shortly. Thanks for mentioning it.

samtstern:
@mohshraim yep we will add this to the docs soon, sorry about the delay!

myleftshoe:
@google-cloud/firestore 0.16.1 has the opposite problem - snapshot.docChanges() errors with docChanges is not a function so I'm using snapshot.docChanges instead!

[...]

PR #825, opened 2018-05-15
Mikelehen/doc changes for each

mikelehen:
Add 'forEach' (and 'map' for good measure) to Array methods we intercept on QuerySnapshot.docChanges.

Would have prevented https://github.com/firebase/firebase-js-sdk/issues/824.

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -1914,10 +1914,11 @@ const docChangesPropertiesToOverride = [
 ];
 docChangesPropertiesToOverride.forEach(property => {
   /**
-   * This will technically overwrite the `Function.prototype.length` property of
-   * `docChanges`. On IE11, the property is improperly defined with
-   * `{ configurable: false }` which causes this line to throw. Wrap in a
-   * try-catch to ensure that we still have a functional SDK.
+   * We are (re-)defining properties on QuerySnapshot.prototype.docChanges which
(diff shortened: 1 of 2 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -1905,23 +1906,25 @@ function throwDocChangesMethodError(): never {
...
-}
+const docChangesPropertiesToOverride = [
+  'length',
+  'forEach',
+  'map',
+  ...(typeof Symbol !== 'undefined' ? [Symbol.iterator] : [])
+];
+docChangesPropertiesToOverride.forEach(property => {
+  /**
+   * This will technically overwrite the `Function.prototype.length` property of
```
schmidt-sebastian: Can you update the comment to indicate that "this" indicates "overwriting length".

mikelehen: Thanks, I rewrote the comment to hopefully be clearer:

```
  /**
   * We are (re-)defining properties on QuerySnapshot.prototype.docChanges which
   * is a Function. This could fail, in particular in the case of 'length' which
   * already exists on Function.prototype and on IE11 is improperly defined with
   * `{ configurable: false }`. So we wrap this in a try/catch to ensure that we
   * still have a functional SDK.
   */
```

I'll hold off on trying to submit this until you make the publishToNpm() style thing go away.

schmidt-sebastian: Thanks. I like it a lot!

PR #838, opened 2018-05-17
Generating target IDs from persistence

schmidt-sebastian:
This PR removes the in-memory state from the query cache. All tabs will generate target IDs based on the current state in IndexedDb.

I'm still using the TargetIdGenerator, but it now takes the value read from IndexedDb to generate the next Target ID.

### DIFF
--- a/packages/firestore/src/core/target_id_generator.ts
+++ b/packages/firestore/src/core/target_id_generator.ts
@@ -20,7 +20,7 @@ import { assert } from '../util/assert';
 const RESERVED_BITS = 1;
 
 enum GeneratorIds {
-  LocalStore = 0, // The target IDs for user queries are even (end in 0).
+  QueryCache = 0, // The target IDs for user-issued queries are even (end in 0).
   SyncEngine = 1 // The target IDs for limbo detection are odd (end in 1).
 }
 
@@ -31,6 +31,8 @@ enum GeneratorIds {
  * will ever produce the same ID. This is useful, because sometimes the backend
  * may group IDs from separate parts of the client into the same ID space.
  */
+// TODO(mrschmidt): Explore removing this class in favor of generating these IDs
+// directly in SyncEngine and LocalStore.
 export class TargetIdGenerator {
   private nextId: TargetId;
 
@@ -69,12 +71,12 @@ export class TargetIdGenerator {
     this.nextId = targetId;
   }
 
-  static forLocalStore(): TargetIdGenerator {
+  static forQueryCache(): TargetIdGenerator {
     // We seed the local store generator to return '2' as its first ID, as there
     // is no differentiation in the protocol layer between an unset number and
     // the number '0'. If we were to sent a target with target ID '0', the
     // backend would consider it unset and replace it with its own ID.
-    const targetIdGenerator = new TargetIdGenerator(GeneratorIds.LocalStore, 2);
+    const targetIdGenerator = new TargetIdGenerator(GeneratorIds.QueryCache, 2);
     return targetIdGenerator;
   }
 
--- a/packages/firestore/src/core/target_id_generator.ts
+++ b/packages/firestore/src/core/target_id_generator.ts
@@ -25,11 +25,18 @@ enum GeneratorIds {
 }
 
 /**
- * TargetIdGenerator generates monotonically increasing integer IDs. There are
- * separate generators for different scopes. While these generators will operate
- * independently of each other, they are scoped, such that no two generators
- * will ever produce the same ID. This is useful, because sometimes the backend
- * may group IDs from separate parts of the client into the same ID space.
+ * Generates monotonically increasing target IDs for sending targets to the
+ * watch stream.
+ *
+ * The client constructs two generators, one for the query cache (via
+ * forQueryCache()), and one for limbo documents (via forSyncEngine()). These
+ * two generators produce non-overlapping IDs (by using even and odd IDs
+ * respectively).
+ *
+ * By separating the target ID space, the query cache can generate target IDs
+ * that persist across client restarts, while sync engine can independently
+ * generate in-memory target IDs that are transient and can be reused after a
+ * restart.
  */
 // TODO(mrschmidt): Explore removing this class in favor of generating these IDs
 // directly in SyncEngine and LocalStore.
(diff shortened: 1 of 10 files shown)


packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -39,69 +39,66 @@ import { PersistencePromise } from './persistence_promise';
...
-   * The last received snapshot version. We store this seperately from the
-   * metadata to avoid the extra conversion to/from DbTimestamp.
-   */
-  private lastRemoteSnapshotVersion = SnapshotVersion.MIN;
-
-  /**
-   * A cached copy of the metadata for the query cache.
-   */
-  private metadata = null;
-
```
mikelehen: Can you add a comment something like:

PORTING NOTE: We don't cache global metadata for the query cache, since some of it (in particular highestTargetId) can be modified by secondary tabs. We could perhaps be more granular (and e.g. still cache lastRemoteSnapshotVersion in memory) but for simplicity we currently go to IndexedDb whenever we need to read metadata. We can revisit if it turns out to have a meaningful perf impact.

packages/firestore/src/core/target_id_generator.ts
```diff
@@ -31,44 +32,54 @@ enum GeneratorIds {
  * may group IDs from separate parts of the client into the same ID space.
  */
 export class TargetIdGenerator {
```
mikelehen: At this point this is a fancy class to add 2 to numbers.  next() adds 2 to the last value returned by after() or next(), and after() adds 2 to the value you pass in.  Perhaps rename it to AddTwoer for clarity?

Or maybe we should drop it and just define a few constants (either in types.ts next to the TargetId definition or perhaps we should create a target_id.ts file or something).

```
const FIRST_LIMBO_TARGET_ID = 1;
const FIRST_LOCAL_STORE_TARGET_ID = 2;
const TARGET_ID_INCREMENT = 2;
```
(with some appropriate comments)

I don't care super strongly but I do think the TargetIdGenerator abstraction isn't paying for itself.  And it's a bit suspicious that it has next() and after() methods but 1 consumer exclusively uses next() [limbo resolution] and the other exclusively uses after() [user queries]...  This also means that after() could literally just add 2 without updating this.nextID.

As-is I almost feel like the comments should call out that this is just a "add 2" helper. :-)

schmidt-sebastian: This is a fair comment, but after some extensive googling, I wasn't able to find an NPM package we could substitute for this generator.

... but in reality, I'd like to keep it as is. It more closely matches what we are doing on other platforms. I am not opposed to removing eventually for all clients (TODO added), but would prefer to defer this work until later.

packages/firestore/src/local/indexeddb_query_cache.ts
schmidt-sebastian: Added this comment. Thanks!

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -148,23 +157,29 @@ export class IndexedDbQueryCache implements QueryCache {
...
+    metadata: DbTargetGlobal
+  ): boolean {
+    if (queryData.targetId > metadata.highestTargetId) {
+      metadata.highestTargetId = queryData.targetId;
+      return true;
     }
 
     // TODO(GC): add sequence number check
-    return needsUpdate;
+    return false;
```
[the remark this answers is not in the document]
schmidt-sebastian: I am willing to take that risk.. 

Ask me again when Greg gets back though.

packages/firestore/src/local/local_store.ts
```diff
@@ -697,10 +706,12 @@ export class LocalStore {
             queryData = cached;
             return PersistencePromise.resolve();
           } else {
-            // TODO(multitab): targetIdGenerator is not multi-tab safe
-            const targetId = this.targetIdGenerator.next();
-            queryData = new QueryData(query, targetId, QueryPurpose.Listen);
-            return this.queryCache.addQueryData(txn, queryData);
+            return this.queryCache
+              .allocateTargetId(txn, this.targetIdGenerator)
```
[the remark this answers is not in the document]
schmidt-sebastian: That seems fair.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -364,7 +364,7 @@ export class RemoteStore {
     if (
       !snapshotVersion.isEqual(SnapshotVersion.MIN) &&
       snapshotVersion.compareTo(
-        this.localStore.getLastRemoteSnapshotVersion()
+        await this.localStore.getLastRemoteSnapshotVersion()
       ) >= 0
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

packages/firestore/test/unit/local/test_query_cache.ts
```diff
@@ -48,8 +51,10 @@ export class TestQueryCache {
     });
   }
 
-  count(): number {
-    return this.cache.count;
+  getQueryCount(): Promise<number> {
+    return this.persistence.runTransaction('getQueryCount', true, txn => {
```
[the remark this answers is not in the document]
schmidt-sebastian: I left `getQueryCount` as is for now (it is only used for testing), but changed `allocateTargetId` to not require a primary lease. This matches the usage in the rest of the code.

I added a TODO to address the existing callbacks.

packages/firestore/src/core/target_id_generator.ts
```diff
@@ -69,12 +71,12 @@ export class TargetIdGenerator {
     this.nextId = targetId;
   }
 
-  static forLocalStore(): TargetIdGenerator {
+  static forQueryCache(): TargetIdGenerator {
     // We seed the local store generator to return '2' as its first ID, as there
```
mikelehen: "local store" => "query cache" ?

packages/firestore/src/core/target_id_generator.ts
```diff
@@ -31,6 +31,8 @@ enum GeneratorIds {
  * will ever produce the same ID. This is useful, because sometimes the backend
  * may group IDs from separate parts of the client into the same ID space.
```
mikelehen: I think this comment could better reflect the state of the world with your simplifications...  Something like:

Generates target IDs for sending targets to the watch stream. The client constructs two generators, one for the query cache (via forQueryCache()), and one for limbo documents (via forSyncEngine()).  The two generators produce non-overlapping IDs (by using even and odd IDs respectively). 

Having two separate generators is useful because query cache target IDs persist across client restarts and must not be reused, while sync engine target IDs are transient and can be reused after a restart.

Issue #555, opened 2018-03-07
DataSnapshot forEach method should be able to return void


Francescu:
Code from the doc doesn't use the `return boolean`  from the `forEach` function when it doesn't want to interrupt the enumeration.

# Example
```javascript
var query = firebase.database().ref("users").orderByKey();
query.once("value")
  .then(function(snapshot) {
    snapshot.forEach(function(childSnapshot) {
      // key will be "ada" the first time and "alan" the second time
      var key = childSnapshot.key;
      // childData will be the actual contents of the child
      var childData = childSnapshot.val();
  });
});
```
# Problem
This code is not valid with the current Typescript definitions.
```
export interface DataSnapshot {
  forEach(action: (a: DataSnapshot) => boolean): boolean;
}
```
# Proposal
I'm new to Firebase so I'd be careful there. But it seems that adding to the closure return type `|void` would fix the issue. I can provide a PR, if you're OK with this solution.

```
export interface DataSnapshot {
  forEach(action: (a: DataSnapshot) => boolean|void): boolean;
}
```

schmidt-sebastian:
Thanks for sending this over @Francescu. We should indeed update the types. If you want to take a stab at this, I would be more than welcome to review your PR. 

Take a look here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database-types/index.d.ts

Crazometer:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/database-types/index.d.ts#L78 Can be changed to Promise<DataSnapshot>

jshcrowthe:
Closing this as it seems things have been taken care of. Thanks!

PR #840, opened 2018-05-18
Update rtdb types

### LINKED ISSUE #306: Updates to Database Type Definitions
* Expose `ThenableReference` type from the database type definitions.
* Currently `then()` and `catch()` methods are exposed from the `Reference` type, which doesn't seem to be correct. Those belong in `ThenableReference`.
* Add `EventType` type:

```
type EventType = 'value' | 'child_added' | 'child_changed' | 'child_moved' | 'child_removed';
```

Crazometer:
Updated some of the type definitions.

Fixes #306, #555.

### DIFF
--- a/packages/database-types/index.d.ts
+++ b/packages/database-types/index.d.ts
@@ -20,7 +20,7 @@ export interface DataSnapshot {
   child(path: string): DataSnapshot;
   exists(): boolean;
   exportVal(): any;
-  forEach(action: (a: DataSnapshot) => boolean): boolean;
+  forEach(action: (a: DataSnapshot) => boolean | void): boolean;
   getPriority(): string | number | null;
   hasChild(path: string): boolean;
   hasChildren(): boolean;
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
Thanks for sending this over. Due to how our repo is structured, these changes need to be replicated in one more place (https://github.com/firebase/firebase-js-sdk/blob/dd11cd17aada45a19732d6517a7d740e722546d2/packages/firebase/index.d.ts). I will do this myself once CI passes and this PR is merged.

Crazometer:
Thanks!

jamesdaniels:
FYI while being more explicit is useful, mind that changes such as this are breaking changes for typescript users. https://travis-ci.org/angular/angularfire2/jobs/386001074

jamesdaniels:
@schmidt-sebastian please watch out for what should be major changes to the types in future PRs. Angularfire is still in RC still so I'm not stressing out too much but this could break the build for anyone else using TS. 

IMO correct thing to do would be:

```ts
// TODO(semver): drop "string" in 6.0 release
type EventType = string | 'value' | 'child_added' | 'child_changed' | 'child_moved' | 'child_removed';
```

PR #842, opened 2018-05-18
DOC: Add a link to the release notes in README.md

mwaldstein:
Every time there is a new release, I need to go searching for this - not sure that everyone is aware of where these are placed online, seems to make sense to have a link included here.


mwaldstein:
I signed it!

PR #845, opened 2018-05-18
Propagating RTDB type changes

schmidt-sebastian:
This is a follow-up PR to #840

### DIFF
--- a/packages/database/src/core/snap/ChildrenNode.ts
+++ b/packages/database/src/core/snap/ChildrenNode.ts
@@ -327,7 +327,7 @@ export class ChildrenNode implements Node {
   forEachChild(
     index: Index,
     action: (key: string, node: Node) => boolean | void
-  ): any {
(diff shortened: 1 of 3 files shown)


packages/database/src/core/snap/ChildrenNode.ts
```diff
@@ -324,7 +324,10 @@ export class ChildrenNode implements Node {
   /**
    * @inheritDoc
    */
-  forEachChild(index: Index, action: (key: string, node: Node) => void): any {
+  forEachChild(
+    index: Index,
+    action: (key: string, node: Node) => boolean | void
+  ): any {
```
mikelehen: should the return type be boolean instead of any?

schmidt-sebastian: Yeah, good call. I changed this as well.

PR #846, opened 2018-05-18
 Falling back on ABORTED_ERROR

schmidt-sebastian:
This allows us to fall back to memory persistence when we get ABORTED_ERR 20.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/DOMException#Error_names

Fixes: https://github.com/firebase/firebase-js-sdk/issues/720

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ applications using Firebase services. This SDK is distributed via:
 To get started using Firebase, see
 [Add Firebase to your JavaScript Project](https://firebase.google.com/docs/web/setup).
 
+Current [Release Notes](https://firebase.google.com/support/release-notes/js)
+
 ## SDK Dev Workflow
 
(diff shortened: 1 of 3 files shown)


paulstelzer:
@schmidt-sebastian You made a small mistake ;) It's not "const DOM_EXCEPTION_ABORTED = 22;" - it's "const DOM_EXCEPTION_ABORTED = 20;"

schmidt-sebastian:
> @schmidt-sebastian You made a small mistake ;) It's not "const DOM_EXCEPTION_ABORTED = 22;" - it's "const DOM_EXCEPTION_ABORTED = 20;"

Obviously :/ Thanks for catching.

paulstelzer:
;) I need this fix for my app, so I take care everything is correctly :)

PR #851, opened 2018-05-21
Reworking PersistencePromise.waitFor to not use chaining

schmidt-sebastian:
This solves a potential performance bottleneck in `PersistencePromise.waitFor`. Instead of serially executing all promises, we should queue them all beforehand and enforce any ordering.


packages/firestore/test/unit/local/simple_db.test.ts
```diff
@@ -455,4 +455,33 @@ describe('SimpleDb', () => {
       });
     });
   });
+
+  // tslint:disable-next-line:ban A little perf test for convenient benchmarking
+  it.only('Perf', () => {
```
mikelehen: Should this be it.skip?

PR #854, opened 2018-05-21
 Adding ability to run prettier for Firestore

schmidt-sebastian:
Adding the ability to run `yarn prettier` in `packages/firestore`.


yarn.lock
```diff
@@ -8724,6 +8724,10 @@ prettier@1.12.0:
   version "1.12.0"
   resolved "https://registry.npmjs.org/prettier/-/prettier-1.12.0.tgz#d26fc5894b9230de97629b39cae225b503724ce8"
 
+prettier@^1.12.1:
```
wilhuff: Do we really want multiple versions of prettier in here?

packages/firestore/package.json
```diff
@@ -56,6 +57,7 @@
     "mocha": "5.0.5",
     "npm-run-all": "4.1.2",
     "nyc": "11.6.0",
+    "prettier": "^1.12.1",
```
mikelehen: This should be an explicit version (and as Gil points out, we should make sure it matches the root package.json version).

schmidt-sebastian:
Hmm, I am gonna blame yesterday's state of this two line PR on my lack of Git skills. I picked up a somewhat outdated version of master that still used the version ranges for dev-dependencies. It should now be fixed.

PR #856, opened 2018-05-22
Add missing tslib dependency

NeoLegends:
`tslib` was missing from the `package.json`, so the helpers ended up being inlined. This PR fixes this. :)


jshcrowthe:
Thanks a bunch! Great find 👍

PR #857, opened 2018-05-22
[esm/master]: Refactor Component Registration

jshcrowthe:
This refactors all of the component registration to happen in the `firebase` package instead of in the subpackage itself (notable exception here being auth).


jshcrowthe:
Just FYI: I'm expecting the tests here to fail initially. Will follow up with a fix for that, just wanted to get this out for now.

PR #862, opened 2018-05-23
Use isArrayBufferEqual to compare VAPID keys

### LINKED ISSUE #860: `messaging().getToken()` in service-worker context fails when not using own VAPID-keys
### Environment

  * Operating System version: Windows 10 LTSB
  * Firebase SDK version: 4.13.1
  * Firebase Product: messaging

### The problem

`messaging().getToken` fails in service-worker context with 401 from fcm.googleapis.com:
> Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential.

#### Steps to reproduce:

0) Perform initial setup of application with `messaging` for webpush. Just like in [tutorial](https://github.com/firebase/quickstart-js/tree/master/messaging), but without own VAPID-key;
1) Request token in service-worker context just like you would normally do;
2) I expect the request will fail with message described above.

#### Relevant Code:

I've narrowed my problem to [this code](https://github.com/firebase/firebase-js-sdk/blob/dd6937176d543352373181fce24ab67065ac1b96/packages/messaging/src/models/iid-model.ts#L113). Even though I'm not using own VAPID-key during firebase initialization, the key is being written to the IndexedDB anyway. And this is ok. Problems begin to appear when we are retrieving this key from IDB in service-worker-controller and comparing it with `DEFAULT_PUBLIC_VAPID_KEY`. And comparison performed in the code above will never succeed. We can't compare UInt8ByteArrays like that, unfortunately.

mmermerkaya:
Fixes #860.

### DIFF
--- a/packages/messaging/src/models/iid-model.ts
+++ b/packages/messaging/src/models/iid-model.ts
@@ -15,6 +15,7 @@
  */
 
 import { arrayBufferToBase64 } from '../helpers/array-buffer-to-base64';
+import { isArrayBufferEqual } from '../helpers/is-array-buffer-equal';
 import { ERROR_CODES, errorFactory } from './errors';
 import { DEFAULT_PUBLIC_VAPID_KEY, ENDPOINT } from './fcm-details';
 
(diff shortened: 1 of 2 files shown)


packages/messaging/src/models/iid-model.ts
```diff
@@ -42,7 +43,12 @@ export class IidModel {
       `encryption_key=${p256dh}&` +
       `encryption_auth=${auth}`;
 
-    if (publicVapidKey !== DEFAULT_PUBLIC_VAPID_KEY) {
+    if (
```
dwoffinden: This formatting looks like the formatter tried to smear your PR over as many lines as possible, turning it into some kind of weird symmetric ASCII-art :/

PR #864, opened 2018-05-23
Improves "Could not reach Firestore backend." log message.

mikelehen:
Improved message:

> Could not reach Cloud Firestore backend. Connection failed 2 times. Most recent error: FirebaseError: [code=unavailable]: 14 UNAVAILABLE: Connect Failed
> This is purely an informational message and typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.
> 

Or:

> Could not reach Cloud Firestore backend. Backend didn't respond within 10 seconds.
> This is purely an informational message and typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.

### DIFF
--- a/packages/firestore/src/remote/online_state_tracker.ts
+++ b/packages/firestore/src/remote/online_state_tracker.ts
@@ -18,6 +18,7 @@ import { OnlineState } from '../core/types';
 import * as log from '../util/log';
 import { assert } from '../util/assert';
 import { AsyncQueue, TimerId } from '../util/async_queue';
+import { FirestoreError } from '../util/error';
 import { CancelablePromise } from '../util/promise';
 
 const LOG_TAG = 'OnlineStateTracker';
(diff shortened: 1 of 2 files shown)


packages/firestore/src/remote/online_state_tracker.ts
```diff
@@ -166,10 +170,18 @@ export class OnlineStateTracker {
     }
   }
 
-  private logClientOfflineWarningIfNecessary(): void {
+  private logClientOfflineWarningIfNecessary(details: string): void {
+    const message =
+      `Could not reach Cloud Firestore backend. ${details}\n` +
+      `This is purely an informational message and typically indicates ` +
```
schmidt-sebastian: I would remove the first part of this sentence. Just saying "This typically indicates that your device ..." should make it obvious where the problem is.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -315,16 +315,17 @@ export class RemoteStore {
...
+  private async onWatchStreamClose(error?: FirestoreError): Promise<void> {
     assert(
       this.isNetworkEnabled(),
       'onWatchStreamClose() should only be called when the network is enabled'
     );
 
     this.cleanUpWatchStreamState();
-    this.onlineStateTracker.handleWatchStreamFailure();
+
+    if (error) {
```
schmidt-sebastian: I am tempted to make this part of the `this.shouldStartWatchStream()` if-check. That way we only change the state to `Unknown` if we actually care (and nicely works around the problem that the comment in line 330 seems to apply more to line 326 than 331).

packages/firestore/src/remote/online_state_tracker.ts
mikelehen: Good call, done.

packages/firestore/src/remote/remote_store.ts
mikelehen: done (and tweaked comments)

schmidt-sebastian:
Thanks. Changes look good.

PR #875, opened 2018-05-29
Tracking query state in local storage

schmidt-sebastian:
This PR adds local storage notifications for query updates.

There are two kinds of updates that we track:

- Query changes from `inactive` to `pending` or `rejected`.
- A `pending` query receives an update (by setting a new lastUpdateTime).

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -118,12 +118,12 @@ export interface SharedClientState {
   /**
    * Records that a query target has been updated.
    *
-   * Called by the primary client to notify secondary clients of updates to
-   * existing watch targets.
+   * Called by the primary client to notify secondary clients of document
+   * changes or state transitions for the provided query target.
    */
   trackQueryUpdate(
     targetId: TargetId,
-    state: 'active' | 'inactive' | 'rejected',
+    state: 'active' | 'current' | 'rejected',
     error?: FirestoreError
   ): void;
 
@@ -289,7 +289,7 @@ export class QueryTargetMetadata {
     let validData =
       typeof targetState === 'object' &&
       isSafeInteger(targetState.lastUpdateTime) &&
-      ['pending', 'active', 'inactive', 'rejected'].indexOf(
+      ['pending', 'active', 'current', 'rejected'].indexOf(
         targetState.state
       ) !== -1 &&
       (targetState.error === undefined ||
@@ -678,7 +678,7 @@ export class WebStorageSharedClientState implements SharedClientState {
 
   trackQueryUpdate(
     targetId: BatchId,
-    state: 'active' | 'inactive' | 'rejected',
+    state: 'active' | 'current' | 'rejected',
     error?: FirestoreError
   ): void {
     this.persistQueryTargetState(targetId, state, error);
@@ -954,7 +954,7 @@ export class MemorySharedClientState implements SharedClientState {
 
   trackQueryUpdate(
     targetId: BatchId,
-    state: 'active' | 'inactive' | 'rejected',
+    state: 'active' | 'current' | 'rejected',
     error?: FirestoreError
   ): void {
     // No op.
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -23,10 +23,7 @@ import { SortedSet } from '../util/sorted_set';
 import { isSafeInteger } from '../util/types';
 import * as objUtils from '../util/obj';
 import { User } from '../auth/user';
-import {
-  SharedClientStateSyncer,
-  QueryTargetState
-} from './shared_client_state_syncer';
+import { SharedClientStateSyncer } from './shared_client_state_syncer';
 import { AsyncQueue } from '../util/async_queue';
 import { Platform } from '../platform/platform';
 
@@ -123,7 +120,7 @@ export interface SharedClientState {
    */
   trackQueryUpdate(
     targetId: TargetId,
-    state: 'active' | 'current' | 'rejected',
+    state: 'incomplete' | 'current' | 'rejected',
     error?: FirestoreError
   ): void;
 
@@ -254,20 +251,20 @@ export class MutationMetadata {
  */
 interface QueryTargetStateSchema {
   lastUpdateTime: number;
-  state: QueryTargetState;
+  state: 'incomplete' | 'current' | 'rejected';
   error?: { code: string; message: string }; // Only set when state === 'rejected'
 }
 
 /**
  * Holds the state of a query target, including its target ID and whether the
- * target is 'pending', 'active', 'inactive' or 'rejected'.
+ * target is 'incomplete', 'current' or 'rejected'.
  */
 // Visible for testing
 export class QueryTargetMetadata {
   constructor(
     readonly targetId: TargetId,
     readonly lastUpdateTime: Date,
-    readonly state: QueryTargetState,
+    readonly state: 'incomplete' | 'current' | 'rejected',
     readonly error?: FirestoreError
   ) {
     assert(
@@ -289,9 +286,7 @@ export class QueryTargetMetadata {
     let validData =
       typeof targetState === 'object' &&
       isSafeInteger(targetState.lastUpdateTime) &&
-      ['pending', 'active', 'current', 'rejected'].indexOf(
-        targetState.state
-      ) !== -1 &&
+      ['incomplete', 'current', 'rejected'].indexOf(targetState.state) !== -1 &&
       (targetState.error === undefined ||
         typeof targetState.error === 'object');
 
@@ -400,7 +395,7 @@ class RemoteClientState implements ClientState {
       (clientState.maxMutationBatchId === null ||
         isSafeInteger(clientState.maxMutationBatchId));
 
-    let activeTargetIdsSet = new SortedSet<TargetId>(primitiveComparator);
+    let activeTargetIdsSet = targetIdSet();
 
     for (let i = 0; validData && i < clientState.activeTargetIds.length; ++i) {
       validData = isSafeInteger(clientState.activeTargetIds[i]);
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -231,6 +248,100 @@ export class MutationMetadata {
   }
 }
 
+/**
+ * The JSON representation of a query target's state as used during LocalStorage
+ * serialization. The TargetId is omitted as it is encoded as part of the key.
+ */
+interface QueryTargetStateSchema {
+  lastUpdateTime: number;
```
mikelehen: Remind me why we're storing these lastUpdateTime numbers?  Are we using them?

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -108,6 +128,29 @@ class TestSharedClientSyncer implements SharedClientStateSyncer {
...
+        this.pendingTargets.push(targetId);
+        return;
+      case 'active':
+        this.activeTargets.push(targetId);
+        return;
+      case 'inactive':
+        this.inactiveTargets.push(targetId);
+        return;
+      case 'rejected':
+        this.rejectedTargets[targetId] = error;
```
mikelehen: FWIW I'd kind of expect that if a target is in rejectedTargets then it's not in pendingTargets or activeTargets, etc.  But I don't think that's how this works...

You could rename this "targetRejections" or something, but I think you have the same problem with e.g. pendingTargets and activeTargets and I'm not coming up with a good naming solution there. :-/

Your tests don't actually run into this since they just do a single applyTargetState() call which isn't super realistic.

I wonder if it'd make sense to just store a TargetID=>state mapping, rather than have  a per-state list?  Or if you intend to test the sequence of state changes, store a TargetId=>list<state> mapping...  WDYT?

schmidt-sebastian: The tests merely verify that the states that are correctly written to Local Storage. They don't verify the state transitions. I don't rely on any order of transitions in the code and it will be hard to enforce them across multiple tabs.

That being said, I changed `TestSharedClientSyncer` to contain a mapping from target ID to state, but kept the assertions the same. The mapping gets translated into disjunct arrays in `get sharedClientState()`

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -60,49 +65,97 @@ function mutationKey(user: User, batchId: BatchId): string {
...
-      rejectedBatches: this.rejectedBatches
+    const state: TestSharedClientState = {
+      pendingBatches: [],
+      acknowledgedBatches: [],
+      rejectedBatches: {},
+      pendingTargets: [],
+      activeTargets: [],
+      currentTargets: [],
+      rejectedTargets: {}
     };
```
mikelehen: I don't care too strongly but it seems like this is optimizing for having a bunch of targets and then being able to assert that they end up in the right states like `expect(clientState.activeTargets).to.equal([1, 5, 7, 8])` but I think all of your tests only deal with a single target and you end up doing:

```
        expect(clientState.pendingTargets).to.be.empty;
        expect(clientState.activeTargets).to.be.empty;
        expect(clientState.currentTargets).to.have.members([1]);
        expect(clientState.rejectedTargets).to.be.empty;
```

And so it'd be simpler if you could just do:

```
expect(clientState.targetStates[1]).to.equal('pending')
```

Again I don't care that much, but this seems roundabout and more complicated than necessary given your current tests.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -231,6 +248,100 @@ export class MutationMetadata {
...
+  state: QueryTargetState;
+  error?: { code: string; message: string }; // Only set when state === 'rejected'
+}
+
+/**
+ * Holds the state of a query target, including its target ID and whether the
+ * target is 'pending', 'active', 'inactive' or 'rejected'.
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
schmidt-sebastian: I went ahead with this and updated the test assertions to directly assert the states on the map that I built up during the test run. These tests only exercise a single step at a time, so most of the assertions are fairly simple. The previous tests did a better job of hiding this :)

schmidt-sebastian:
I updated the PR with the changes we discussed offline. The 'pending' state is now read from the client state directly and no longer persisted in the target row. The target row only ever gets updated by the primary client.

With these changes, there are now a bunch of places that accept only the three states that can be set by the primary tab (incomplete, current, rejected), versus the four states that can be passed to SyncEngine (incomplete, current, rejected, pending). For now, I have manually defined the three state combination, and use `QueryTargetState` in SyncEngine. It might make sense to reverse this.

Also, what do you think of 'incomplete' versus 'active'?

packages/firestore/src/local/shared_client_state_syncer.ts
```diff
@@ -14,10 +14,26 @@
...
+ *
+ * When a secondary tab inserts a new target, it is marked 'pending'. The
+ * primary tab then transitions the target to 'incomplete' once Watch has
+ * delivered the first snapshot. Once a Watch target is current, it transitions
+ * to 'current'.
+ *
+ * Targets that get rejected from the backend are marked 'rejected'.
+ */
+export type QueryTargetState =
+  | 'pending'
```
mikelehen: It's not obvious to me why we need this state.  What will secondary tabs do in response to 'pending'?  If we got rid of it then we could use QueryTargetState everywhere, right?

packages/firestore/src/local/shared_client_state_syncer.ts
```diff
@@ -14,10 +14,26 @@
...
+  | 'incomplete'
```
mikelehen: What's the advantage of 'incomplete' over 'not-current' (which avoids inventing a new term we don't use in the rest of the client)?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -556,6 +666,16 @@ export class WebStorageSharedClientState implements SharedClientState {
   removeLocalQueryTarget(targetId: TargetId): void {
     this.localClientState.removeQueryTarget(targetId);
     this.persistClientState();
+    // TODO(multitab): Call `unlisten` on the primary tab and remove the query
+    // state from Local Storage.
+  }
+
+  trackQueryUpdate(
+    targetId: BatchId,
```
mikelehen: Unusual type choice!

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -732,6 +896,35 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+    const existingTargets = this.activeClients[clientState.clientId]
+      ? this.activeClients[clientState.clientId].activeTargetIds
+      : targetIdSet();
+    this.activeClients[clientState.clientId] = clientState;
+
+    let newTargets = Promise.resolve();
+    clientState.activeTargetIds.forEach(async targetId => {
+      if (!existingTargets.has(targetId)) {
+        newTargets = newTargets.then(() =>
+          this.syncEngine.applyTargetState(targetId, 'pending')
```
mikelehen: I am expecting that only the master tab cares about new targets, and it should also care about removed targets (so it knows to stop listening).  So I am thinking that this shouldn't use applyTargetState() and instead we should have new method(s) to notify the master of new/removed targets (though we need to somehow aggregate across all client states to know if it's new or not).

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -772,6 +965,14 @@ export class MemorySharedClientState implements SharedClientState {
     this.localState.addQueryTarget(targetId);
   }
 
+  trackQueryUpdate(
+    targetId: BatchId,
```
mikelehen: More confused types.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -800,3 +1001,11 @@ export class MemorySharedClientState implements SharedClientState {
...
   shutdown(): void {}
 }
+
+const EMPTY_NUMBERED_ID_SET = new SortedSet<number>(primitiveComparator);
+function targetIdSet(): SortedSet<TargetId> {
+  return EMPTY_NUMBERED_ID_SET;
+}
+function batchIdSet(): SortedSet<BatchId> {
+  return EMPTY_NUMBERED_ID_SET;
+}
```
mikelehen: Move these to collections.ts as EMPTY_BATCH_ID_SET and EMPTY_TARGET_ID_SET ?  (and I probably wouldn't bother sharing an underlying EMPTY_NUMBERED_ID_SET as I don't think we're that desperate to save on object count).

packages/firestore/src/local/shared_client_state_syncer.ts
schmidt-sebastian: This was a way to avoid another callback. I have added the callback with the name we discussed on Friday (applyActiveTargetsChange).

packages/firestore/src/local/shared_client_state_syncer.ts
schmidt-sebastian: It's not as clumsy as 'not-current', but I don't feel strongly. I changed it everywhere (hopefully!) to not-current.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Fixed

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: I added a new method to deal with this (appleActiveTargetsChange as suggested by our Kirkland team). I am now also `de-duping` these notifications in SharedClientState.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Fixed. I was in "think different" mode.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: That works for me. Done.

packages/firestore/src/model/collections.ts
```diff
@@ -52,3 +54,15 @@ const EMPTY_DOCUMENT_KEY_SET = new SortedSet(DocumentKey.comparator);
 export function documentKeySet(): DocumentKeySet {
   return EMPTY_DOCUMENT_KEY_SET;
 }
+
+export type TargetIdSet = SortedSet<TargetId>;
+const EMPTY_TARGET_ID_SET = new SortedSet<number>(primitiveComparator);
```
mikelehen: Is this intentionally "number" instead of "TargetId"? (same for BatchId below)

PR #877, opened 2018-05-29
Allow field deletes for SetOptions.mergeFields

schmidt-sebastian:
The validation for `set(data, mergeFields:...`) rejected field paths that pointed to field deletes/transforms since these are stripped from the update data.  This PR checks the field mask (for field deletes) and the list of field transforms (for server timestamps) to validate mergeFields.

Note that I could re-use the existing FieldMask.covers() function if I turned the array of field paths into a FieldMask.

### DIFF
--- a/packages/firestore/src/api/user_data_converter.ts
+++ b/packages/firestore/src/api/user_data_converter.ts
@@ -242,6 +242,16 @@ class ParseContext {
     );
   }
 
+  /** Returns 'true' if 'fieldPath' was traversed when creating this context. */
+  contains(fieldPath: FieldPath): boolean {
+    return (
+      this.fieldMask.find(field => fieldPath.isPrefixOf(field)) !== undefined ||
+      this.fieldTransforms.find(transform =>
+        fieldPath.isPrefixOf(transform.field)
+      ) !== undefined
+    );
+  }
+
   private validatePath(): void {
     // TODO(b/34871131): Remove null check once we have proper paths for fields
     // within arrays.
@@ -345,12 +355,13 @@ export class UserDataConverter {
           fail('Expected stringOrFieldPath to be a string or a FieldPath');
         }
 
-        if (!updateData.contains(fieldPath)) {
+        if (!context.contains(fieldPath)) {
           throw new FirestoreError(
             Code.INVALID_ARGUMENT,
             `Field '${fieldPath}' is specified in your field mask but missing from your input data.`
           );
         }
+
         validatedFieldPaths.push(fieldPath);
       }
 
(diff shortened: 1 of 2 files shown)


packages/firestore/test/integration/api/database.test.ts
```diff
@@ -158,14 +160,17 @@ apiDescribe('Database', persistence => {
       const initialData = {
         untouched: true,
         foo: 'bar',
+        inner: { untouched: true, foo: 'bar' },
```
mikelehen: inner looks identical to nested...  what am I missing?

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -242,6 +242,16 @@ class ParseContext {
     );
   }
 
+  /** Returns 'true' if 'fieldPath' was traversed when creating this context. */
+  contains(fieldPath: FieldPath): boolean {
```
mikelehen: I think contains() can be misinterpreted to think the specified path must be contained within one of the writes in the ParseContext, but really it's the opposite (the specified path must contain one of the writes).  Maybe affects(), modifies(), or containsWriteBelow() ?  That said, I guess this is consistent with the existing ObjectValue.contains() so you can leave it as-is if you prefer.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -149,6 +150,7 @@ apiDescribe('Database', persistence => {
           expect(docSnapshot.exists).to.be.ok;
           expect(docSnapshot.get('updated')).to.be.false;
           expect(docSnapshot.get('time')).to.be.an.instanceof(Timestamp);
+          expect(docSnapshot.get('nested.time')).to.be.an.instanceof(Timestamp);
```
mikelehen: changes to this test seem fine but unrelated to your change?  Are you making it match another platform or something?

packages/firestore/src/api/user_data_converter.ts
schmidt-sebastian: The function returns true if the original object that was being parsed contains a value at `fieldPath`. it is not necessarily a 1:1 match with writes, since a write at `a.b.c` contains `a.b`. The function returns false for an object `{a : b: {}}` and field path `a.b.c` (which I think is your second example). `contains` matches the semantics of `ObjectValue.contains()` and hence I'd prefer to keep the name.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: This is to match the new test `can set server timestamps using mergeFields`.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: This is meant to match `can delete field using mergeFields` which specifies two different field depths for the field mask. I removed it from here since it doesn't make sense for this test.

PR #878, opened 2018-05-29
Using LimboResolution for Limbo Resolution

schmidt-sebastian:
This PR also reduces the average line count of my recent PRs.


schmidt-sebastian:
Android is correct.

PR #885, opened 2018-05-31
Exporting firebase as namespace

screendriver:
You can't use firebase in a TypeScript project that loads firebase via a CDN. If you use a CDN then firebase should be available as a global variable because you don't import firebase in your codebase and just use the global variable.
Therefore (because it's a UMD library) I added a `export as namespace`


screendriver:
I signed it!

PR #892, opened 2018-06-01
Test for b/72533250.

mikelehen:
This is a test that demonstrates b/72533250 if you uncomment the line after the TODO.

I'm punting on the actual bugfix at the moment because I think the best fix is a semi-large refactor.


wilhuff:
Could you make this change description something that can be reasonably interpreted without access to buganizer?

PR #894, opened 2018-06-01
Update Chrome and Safari version in cross-browser test.

yifanyang:
The root cause of recent cannot start browser failure (https://travis-ci.org/firebase/firebase-js-sdk/jobs/386792434) is that Safari 11.0 used for our browser test is no longer supported by SauceLabs. This commit sets browsers to be always the lasted version available on SauceLabs.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,6 +1,5 @@
-chrome 66
-firefox 60
-safari 11
-edge 17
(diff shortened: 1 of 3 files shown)


yifanyang:
@mikelehen @jshcrowthe Thanks guys!

PR #903, opened 2018-06-04
Refine onSnapshot tests to ensure correct number of events emitted

rsgowman:
Behaviour was already correct. This PR is effectively just porting the
existing iOS tests to typescript.

http://b/32764756

### DIFF
--- a/packages/firestore/test/integration/util/events_accumulator.ts
+++ b/packages/firestore/test/integration/util/events_accumulator.ts
@@ -32,7 +32,9 @@ export class EventsAccumulator<
 
   storeEvent: (evt: T) => void = (evt: T) => {
     if (this.rejectAdditionalEvents) {
-      throw new Error("Additional event detected after assertNoAdditionalEvents called");
+      throw new Error(
+        'Additional event detected after assertNoAdditionalEvents called'
+      );
     }
     this.events.push(evt);
     this.checkFulfilled();
--- a/packages/firestore/test/integration/util/events_accumulator.ts
+++ b/packages/firestore/test/integration/util/events_accumulator.ts
@@ -34,6 +34,7 @@ export class EventsAccumulator<
     if (this.rejectAdditionalEvents) {
       throw new Error(
         'Additional event detected after assertNoAdditionalEvents called'
+        + JSON.stringify(evt)
       );
     }
     this.events.push(evt);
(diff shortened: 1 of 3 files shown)


packages/firestore/test/integration/util/events_accumulator.ts
```diff
@@ -28,8 +28,14 @@ export class EventsAccumulator<
   private events: T[] = [];
   private waitingFor: number;
   private deferred: Deferred<T[]> | null = null;
+  private rejectAdditionalEvents = false;
 
   storeEvent: (evt: T) => void = (evt: T) => {
+    if (this.rejectAdditionalEvents) {
+      throw new Error(
+        'Additional event detected after assertNoAdditionalEvents called'
```
mikelehen: To make this more actionable, I'd include the event... Something like:
```
    throw new Error(
      'Additional event detected after assertNoAdditionalEvents called:'
      JSON.stringify(evt)
    );
```

packages/firestore/test/integration/util/events_accumulator.ts
```diff
@@ -77,6 +83,7 @@ export class EventsAccumulator<
   }
 
   assertNoAdditionalEvents(): Promise<void> {
+    this.rejectAdditionalEvents = true;
     return new Promise((resolve: (val: void) => void, reject) => {
       setTimeout(() => {
         if (this.events.length > 0) {
```
mikelehen: I think this code is now unreachable, since storeEvent will throw rather than adding to `this.events`.

Perhaps we should remove it (but still do the setTimeout so that tests using assertNoAdditionalEvents() will wait long enough for local events to hit the storeEvent code)?  Or does the waiting for local events not actually work as-is?

rsgowman: TBH, I'm not sure if this function actually did anything. But I'm not certain of that either. I found that *in my situation*, this method would not object (prior to me adding 'rejectAdditionalEvents' at any rate.) But I'm not sure if other situations *would* trigger it. In the end, I took the coward's way out and left it there, figuring, whatever it was doing before (if anything) it'll continue doing (if anything). :)

It *seems* like it should be useful, for instance, if there were *already* events in the queue prior to calling this.

packages/firestore/test/integration/util/events_accumulator.ts
rsgowman: Done.

PR #915, opened 2018-06-11
Removing tsconfig.json from prod build

jshcrowthe:


packages/firebase/package.json
```diff
@@ -13,6 +13,14 @@
     "realtime",
     "storage"
   ],
+  "files": [
+    "**/dist/",
+    "**/package.json",
+    "/firebase*.js",
+    "/firebase*.map",
+    "/index.d.ts",
+    "/externs"
```
hiranya911: Do these get resolved as relative paths?

jshcrowthe: Relative to the root of the package I believe yes.

PR #920, opened 2018-06-13
[Multi-Tab]  Round-tripping queries (not small)

schmidt-sebastian:
This PR finishes up the query support for multi-tab. It adds support and/or tests for query updates, query rejections, held write acks, limit queries and limbo resolution (only performed by the primary tab).

To support held write acks, I also  had to expose SharedClientState to LocalStore. I hope that this is temporary, otherwise it may indeed make sense to move all of the SharedClientState to LocalStore.

To keep this PR from being small, I also changed the Remote Document Cache to return a MaybeDocumentMap in `getNextDocumentChanges` (as previously suggested). I need this map in the places where I consume getNextDocumentChanges.


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -473,18 +477,10 @@ export class LocalClientState implements ClientState {
   }
 
   addQueryTarget(targetId: TargetId): void {
-    assert(
-      !this.activeTargetIds.has(targetId),
-      `Target with ID '${targetId}' already active.`
-    );
```
mikelehen: Hrm. Why exactly did these asserts go away?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -663,9 +659,31 @@ export class WebStorageSharedClientState implements SharedClientState {
     }
   }
 
-  addLocalQueryTarget(targetId: TargetId): void {
+  addLocalQueryTarget(targetId: TargetId): QueryTargetState {
+    let queryState: QueryTargetState = 'not-current';
+
+    // Lookup an existing query state if the target ID was already registered
+    // by another tab
+    if (this.getAllActiveQueryTargets().has(targetId)) {
```
mikelehen: I notice there are several places (3-4?) where we call getAllActiveQueryTargets() when we only need to do something simple (like check for the existence of a single targetId).  This worries me a bit since it turns this into an unexpectedly O(n) operation. 

And this particular case could be problematic.  A single client doing _n_ listens will call addLocalQueryTarget() _n_ times and each one will be O(_n_), leading to O(_n_^2) perf overall...

If possible I suggest we find better data structures / plumbing to avoid these hidden O(n) taxes...

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: I added the first one back. I removed both of them to avoid the question why only this method has an assert.

 `removeQueryTarget` can't have an assert anymore since we call this method twice during  `SyncEngine.unlisten`. Once in the beginning to compute whether the target will remain active, and once when we call `SyncEngine.removeAndCleanUpQueryData`. We can potentially rework the `targetRemainsActive` logic (something along the lines of `SharedClientState.amITheOnlyClient(...)`) and add this assert back.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Added a `isActiveQueryTarget` that I use both here and in SyncEngine. The main difference is that I am no longer accumulating all targets into a new set. It is still not super-efficient, but I would hope that users will not run into this limitation. The number of active clients should be fairly manageable, and managing this directly in a dedicate data structure requires much more state keeping.

`getAllActiveQueryTargets` is now only used from tests.

packages/firestore/src/local/local_store.ts
```diff
@@ -737,6 +742,24 @@ export class LocalStore {
     });
   }
 
+  /**
+   * Locally unpin all the documents associated with the given query, but keep
+   * persisted query data.
+   */
+  // PORTING NOTE: Multi-tab only.
+  removeQuery(query: Query): Promise<void> {
```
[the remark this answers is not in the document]
mikelehen: I'm still struggling to understand and feel good about this code.  We used to have nice symmetry between allocateQuery() and releaseQuery() and that was easy to reason about.  But now primary tabs call allocateQuery()  / removeLocalQueryData() + releaseQuery().  While secondary tabs call allocateQuery() / removeLocalQueryData() [I think??].  That's less easy to reason about.  Somebody trying to write code against LocalStore based purely on its public API and method comments would have a really hard time.

If the only issue we're avoiding is the primary lease, then maybe we could do:

```
    const needPrimaryLease = this.garbageCollector.isEager;
    return this.persistence.runTransaction('Release query', needPrimaryLease, txn => {
      ...
```

And then we could still have a single releaseQuery() method?  Else, I think I'd lean towards either passing a flag to releaseQuery() or at least renaming removeLocalQueryData() to have some symmetry with allocateQuery() [maybe releaseQueryLocally() or something? Not sure.] and add method comments that touch on multi-tab and primary vs secondary counterparts to allocateQuery().

packages/firestore/src/local/local_store.ts
```diff
@@ -862,9 +882,16 @@ export class LocalStore {
...
-        this.applyWriteToRemoteDocuments(txn, batchResult, documentBuffer)
-      );
+      promiseChain = promiseChain
+        .next(() =>
+          this.applyWriteToRemoteDocuments(txn, batchResult, documentBuffer)
+        )
+        .next(() =>
+          this.sharedClientState.trackMutationResult(
+            batchResult.batch.batchId,
+            'acknowledged'
```
mikelehen: So we're notifying LocalStorage before the transaction has been committed, which seems like it could cause race condition issues...  This is perhaps a good reason for LocalStore not to have access to SharedClientState.  In general it won't be safe to do anything with SharedClientState in the middle of transactions (and most LocalStore methods are transactional).

packages/firestore/src/core/sync_engine.ts
```diff
@@ -671,22 +746,68 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+      this.queryViewsByTarget[targetId] ||
+      this.limboTargetIdGenerator.covers(targetId)
+    ) {
+      switch (state) {
+        case 'current':
+        case 'not-current': {
+          const changes = await this.localStore.getNewDocumentChanges();
+          return this.emitNewSnapsAndNotifyLocalStore(
+            changes,
+            state === 'current'
```
mikelehen: Er... this doesn't seem right since it's applying the per-target state to all document changes (which could apply to multiple targets).

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -110,8 +110,12 @@ export interface SharedClientState {
    */
   getMinimumGlobalPendingMutation(): BatchId | null;
 
-  /** Associates a new Query Target ID with the local Firestore clients. */
-  addLocalQueryTarget(targetId: TargetId): void;
+  /**
+   * Associates a new Query Target ID with the local Firestore clients. Returns
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -137,6 +137,18 @@ export class MemoryQueryCache implements QueryCache {
...
 
+  getQueryDataForTarget(
+    transaction: PersistenceTransaction,
+    targetId: TargetId
+  ): PersistencePromise<QueryData | null> {
+    this.queries.forEach((query, queryData) => {
+      if (queryData.targetId === targetId) {
+        return PersistencePromise.resolve(queryData);
+      }
+    });
```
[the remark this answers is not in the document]
schmidt-sebastian: I don't have strong feelings, but some of these multi-tab memory-only methods do called in unit tests. This method doesn't have a specific test at this point and thus doesn't ever get called. I went with your suggestion for now, but it may make sense to add a query cache test to cover this code directly.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -134,8 +138,18 @@ export interface SharedClientState {
    * The implementation for this may require O(n) runtime, where 'n' is the size
    * of the result set.
    */
+  // Visibile for testing
```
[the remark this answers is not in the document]
schmidt-sebastian: Without the test access, this method would not need to be part of the interface and could be private within `WebStorageSharedClientState`. I left the comment but fixed the spelling.

schmidt-sebastian:
I pushed one more commit that removes the `targetIdGenerator.covers()` method. If I am not mistaken, secondary tabs only need to know about resolved limbo documents if they have query target whose CURRENT flag flips from `false` to `true` as part of the limbo resolution. These tabs will then pick up the document change as their target change is processed.

packages/firestore/src/remote/remote_event.ts
```diff
@@ -49,6 +56,30 @@ export class RemoteEvent {
...
   ) {}
+
+  /**
+   * Creates a synthesized RemoteEvent that can be used to apply a CURRENT
+   * status change for a query that is executed by a different tab.
+   */
+  // PORTING NOTE: Multi-tab only
+  static createSynthesizedRemoteEventForSyncStateChange(
+    targetId: TargetId,
+    current: boolean
```
mikelehen: Can we avoid mixing current and syncState terminology and pick one? (probably 'current' since that's what you use as the parameter name and in the method description)

packages/firestore/src/remote/remote_event.ts
```diff
@@ -49,6 +56,30 @@ export class RemoteEvent {
      */
     readonly resolvedLimboDocuments: DocumentKeySet
   ) {}
+
+  /**
+   * Creates a synthesized RemoteEvent that can be used to apply a CURRENT
+   * status change for a query that is executed by a different tab.
```
mikelehen: I would maybe expand this comment and call out that this is still a bit hacky.  Maybe:

HACK: Views require RemoteEvents in order to determine whether the view is CURRENT, but secondary tabs don't receive remote events. So this method is used to create a synthesized RemoteEvent that can be used to apply a CURRENT status change to a View, for queries executed in a different tab.

packages/firestore/src/local/local_store.ts
```diff
@@ -709,52 +709,50 @@ export class LocalStore {
...
-            const documentBuffer = new RemoteDocumentChangeBuffer(
-              this.remoteDocuments
-            );
-            return this.releaseHeldBatchResults(txn, documentBuffer).next(
-              () => {
-                documentBuffer.apply(txn);
-              }
+  // PORTING NOTE: `keepPersistedQueryData` is multi-tab only.
+  releaseQuery(query: Query, keepPersistedQueryData: boolean): Promise<void> {
+    const requirePrimaryLease = keepPersistedQueryData === false;
```
mikelehen: FWIW I verified that if you change this to 
```
const requirePrimaryLease = this.garbageCollector.isEager;
```

and you remove the `!keepPersistedQueryData` check below, the tests all still pass.  So I still think we don't actually _need_ the `keepPersistedQueryData` argument.  But you can leave it if you think it's more correct or we'll need it in the future.

packages/firestore/src/local/local_store.ts
mikelehen: Ahh, good call!  I forgot that IndexedDb transactions are exclusive and we are inside of one.  Thanks for adding the comment though.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -282,9 +282,9 @@ describeSpec('Limbo Documents:', [], () => {
         .expectEvents(query, { fromCache: true })
         .client(0)
         .ackLimbo(1004, deletedDoc2)
-        .expectLimboDocs()
-        .client(1)
-        .expectEvents(query, { removed: [doc2] });
+        .expectLimboDocs();
+      // .client(1)
+      // .expectEvents(query, { removed: [doc2] });
```
mikelehen: intentional?

packages/firestore/src/remote/remote_event.ts
schmidt-sebastian: As discussed via Doodle, I updated this to use "forCurrentChange". I originally avoided this name because it is kind of ambiguous to non-Firestore folks.

packages/firestore/src/remote/remote_event.ts
schmidt-sebastian: Done (here and below)

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Reverted.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: In the multi-tab spec tests, this method only ever seems to be called this `this.garbageCollector.isEager` set to `false`. I don't necessarily want to rely on this, or use this state to toggle unrelated behavior. I prefer to be explicit.

PR #927, opened 2018-06-14
[Multi-Tab] Generating mutation IDs from persistence (medium)

schmidt-sebastian:
This PR removes the in-memory metadata state in the Mutation Queue and instead relies on IndexedDb as the source of truth. 

As part of this effort, I changed the schema to also store the last processed batch ID. This helps when mutations are rejected and prevents clients from re-using existing mutation IDs (which could include IDs assigned by other clients).

I suspect that we don't need `lastAcknowledgedBatchId` anymore, but kept it for now for now since it matches the other clients.


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -70,6 +70,7 @@ export function createOrUpgradeDb(
     p = p.next(() => {
       createClientMetadataStore(db);
       createRemoteDocumentChangesStore(db);
+      return saveProcessedMutationCount(txn);
```
mikelehen: `saveLastProcessedBatchIds` or `populateLastProcessedBatchId` or something?

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -103,38 +89,45 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
-            /*lastStreamToken=*/ ''
+            /*lastStreamToken=*/ '',
+            BATCHID_UNKNOWN
           );
         }
-        this.metadata = metadata;
+
+        // TODO(multitab): This should only run when no other client is
+        // active, as batch IDs may otherwise go back in time in clients that
+        // have already processed mutations.
```
mikelehen: I don't really understand this comment.  Since we're in an IndexedDb transaction, no other client can be accessing the queue metadata...  and since you've gotten rid of the in-memory cache, that should be sufficient, right?  So what is this TODO tracking exactly?

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -103,38 +89,45 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
 
         // On restart, nextBatchId may end up lower than
         // lastAcknowledgedBatchId since it's computed from the queue
```
mikelehen: Is this true any more? I thought your change was preventing reuse of batch IDs? And so we shouldn't ever reset lastAcknowledgedBatchId...

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -184,80 +177,101 @@ export class IndexedDbMutationQueue implements MutationQueue {
   getNextBatchId(
     transaction: PersistenceTransaction
   ): PersistencePromise<BatchId> {
-    return PersistencePromise.resolve(this.nextBatchId);
+    let maxPendingBatchId = BATCHID_UNKNOWN;
+    const range = IDBKeyRange.lowerBound([this.userId]);
```
mikelehen: Don't we need an upper bound to prevent iterating other users' batches?

FWIW- It's occurring to me that we could probably simplify things (and improve performance?) by switching to auto-generated IDs (this would also mean batchIDs would be unique across users which is not necessary but not harmful either).

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -510,6 +529,19 @@ export class IndexedDbMutationQueue implements MutationQueue {
         }
       }
     }
+    // PORTING NOTE: Multi-tab only.
+    // Keeping track of removed mutation batches allows us to ensure that batch
+    // IDs always advance during the lifetime of all clients.
+    promises.push(
+      this.getMutationQueueMetadata(transaction).next(metadata => {
+        if (maxProcessedBatchId > metadata.lastProcessedBatchId) {
```
mikelehen: FWIW, it seems like this should be always true, no?  If so we could consider making it an assert.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -147,13 +148,41 @@ export class DbMutationQueue {
...
 export type DbMutationBatchKey = [string, BatchId];
 
+/**
+ * Sets a mutation queue's `lastProcessedBatchId` to its
+ * `lastAcknowledgedBatchId` for all existing entries.
+ */
+function saveProcessedMutationCount(
+  txn: SimpleDbTransaction
+): PersistencePromise<void> {
+  const mutationsStore = txn.store<DbMutationBatchKey, DbMutationQueue>(
```
mikelehen: mutationsStore => mutationQueuesStore
DbMutationBatchKey => DbMutationQueueKey

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -147,13 +148,41 @@ export class DbMutationQueue {
...
+    DbMutationQueue.store
+  );
+
+  let promises = mutationsStore.loadAll().next(entries => {
+    for (const entry of entries) {
```
mikelehen: entries => queues
entry => queue

packages/firestore/src/local/indexeddb_mutation_queue.ts
mikelehen: +1 to not recycling IDs. I thought that was already a goal of this PR.

packages/firestore/src/local/indexeddb_mutation_queue.ts
mikelehen: Ahh...  Yes, I missed that.  I'm inclined to get rid of all this logic and just use auto-incrementing IDs that never recycle (either by leveraging IndexedDB's autoincrement feature or manually).

packages/firestore/src/local/indexeddb_mutation_queue.ts
mikelehen: But since you're iterating backwards, this will always just read the last mutation stored, regardless of what user it's for, right?  So you need an upper bound to get to the right user...  I think.

But this all gets simpler if we just have global (rather than per-user) auto-incrementing ids.

Oh... I guess autoincrement may not work with compound keys...  We could potentially change the key to just be the batchId and then add an index on [userId, batchId] though.  Not sure if that's more or less expensive than just doing our own auto-incrementing id (i.e. have a mutationQueueGlobals store with a nextBatchId value that we increment every time we create a batch).

packages/firestore/src/local/indexeddb_mutation_queue.ts
mikelehen: Oh, yeah.  I think I knew that once.  We should really get rid of held write acks. :)

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -181,8 +181,10 @@ export class IndexedDbMutationQueue implements MutationQueue {
     const range = IDBKeyRange.lowerBound([this.userId]);
     return mutationsStore(transaction)
       .iterate({ range, reverse: true }, (key, dbBatch, control) => {
-        const batch = this.serializer.fromDbMutationBatch(dbBatch);
-        maxPendingBatchId = batch.batchId;
+        if (dbBatch.userId === this.userId) {
+          const batch = this.serializer.fromDbMutationBatch(dbBatch);
+          maxPendingBatchId = batch.batchId;
+        }
```
mikelehen: I think once you add the upper bound you won't need this.

schmidt-sebastian:
I spent a bunch of time in the afternoon to see what this would look like with auto-generating IDs and had all of this working when I realized that I needed to rewrite the IDs in three different object stores to  make this work. This requires a full-fledged migration and is not all that trivial. I reverted most of the changes I made for this, but kept one of the ideas: Increment the counter of batches as batches are inserted. This nicely simplifies some of the logic, and doesn't require any reverse lookups to retrieve the next batch ID.

If you are interested, 20ccf635822ea49c1024eba879f8e7a2025f67f7 contains the approach that uses auto-incrementing keys, but it falls short when it comes to schema migration.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -139,7 +120,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
   static loadNextBatchIdFromDb(
     txn: PersistenceTransaction
   ): PersistencePromise<BatchId> {
-    let maxBatchId = BATCHID_UNKNOWN;
+    let maxBatchId = 0;
```
mikelehen: Hrm...  Why did this change?  You want the first batchId to be 1 instead of 0 for some reason?

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -243,14 +213,18 @@ export class IndexedDbMutationQueue implements MutationQueue {
     localWriteTime: Timestamp,
     mutations: Mutation[]
   ): PersistencePromise<MutationBatch> {
-    return this.getNextBatchId(transaction).next(batchId => {
+    return this.getMutationQueueMetadata(transaction).next(metadata => {
+      const batchId = metadata.nextBatchId;
+      ++metadata.nextBatchId;
```
mikelehen: Hrm...  metadata is per-user and we only populate metadata.nextBatchId with the globally-correct nextBatchId on startup...  after that we just increment it.  So if another tab managed to log in as a different user and create a batch, then we could end up with duplicate batchIds.  I think this is unlikely because auth tries to synchronize auth state across tabs, but I'm not sure we should rely on that.

That said, I don't think we _actually_ need batchIds to be unique across users.  The original intention was to keep them unique so that if the SDK was ever able to send mutations to the server for a non-logged-in user (e.g. the anonymous user, see b/33437802) then we could still send them in the correct order, across users.

So, uh I think we should probably either give up on globally-unique batch Ids or else we need to have a globalNextBatchId stored in IndexedDb somewhere (a new globals store with a single row or something) and use that instead of the per-user nextBatchId.  And in that case we could stop doing the loadNextBatchIdFromDb() thing in start() [and do it once on schema migration instead]...

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -151,32 +151,31 @@ export class DbMutationQueue {
...
     DbMutationQueue.store
   );
 
-  let promises = mutationsStore.loadAll().next(entries => {
-    for (const entry of entries) {
-      entry.lastProcessedBatchId = entry.lastAcknowledgedBatchId;
-      promises = promises.next(() => mutationsStore.put(entry));
+  let promises = mutationQueuesStore.loadAll().next(queues => {
+    for (const queue of queues) {
+      queue.nextBatchId = Math.max(queue.lastAcknowledgedBatchId + 1, 1);
```
mikelehen: This doesn't seem right.  A client can have lots of unacknowledged mutation batches stored in persistence, right? (e.g. it was operating offline last time the client was used)

But I think it also doesn't matter since IndexedDbMutationQueue.start() is going to replace this with a true globally-unique nextBatchId.  So maybe we don't need this schema migration at all (or we could just write -1 or whatever with a comment explaining why it doesn't matter).

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -89,45 +89,26 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
-              empty,
-              'Reset nextBatchID is only possible when the queue is empty'
-            );
-
-            metadata.lastAcknowledgedBatchId = BATCHID_UNKNOWN;
-            metadata.lastProcessedBatchId = BATCHID_UNKNOWN;
-            return mutationQueuesStore(transaction).put(metadata);
-          });
-        } else {
+        if (globalNextBatchId > metadata.nextBatchId) {
```
mikelehen: Hrm...  So I think this code still allows for re-use of batchIDs, since you might have:

1. user1 writes batchid 0 (user1 metadata.nextBatchId=1)
2. user1 acks batchid 0 (so nothing in the mutations store)
3. user1 logs out and new user2 logs in.
4. user2 metadata.nextBatchId will be initialized to 0 since there's no metadata for user2 and no mutations in the queue.
5. user2 writes batchId 0 (duplicate)

This is probably non-harmful since it's okay to "reset" the batchId if there are no mutations...  but I think your intention was to never reuse batch ids, right?  And I don't think this code does that.

To fix this you need to either iterate over all users' metadata and find the max() of their nextBatchIds... or introduce a global nextBatchId in IndexedDb (as I mention elsewhere).

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -93,9 +93,9 @@ describe('IndexedDbMutationQueue', () => {
       });
     }
 
-    it('returns zero when no mutations.', async () => {
+    it('returns one when no mutations.', async () => {
```
mikelehen: Ahh...  I guess your change was intentional, but I'm still left wondering why.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -204,13 +221,54 @@ function createMutationQueue(db: IDBDatabase): void {
...
+ */
+function upgradeMutationBatchSchemaAndMigrateData(
+  db: IDBDatabase,
+  txn: SimpleDbTransaction
+): PersistencePromise<void> {
+  const v1MutationsStore = txn.store<DbMutationBatchKey, DbMutationBatch>(
+    DbMutationBatch.store
+  );
+  return v1MutationsStore.loadAll().next(existingMutations => {
+    db.deleteObjectStore(DbMutationBatch.store);
```
mikelehen: Hrm...  I just noticed that this doesn't (apparently) happen within the context of the transaction...  which means an inopportune page close or similar could cause data loss.  :-/  We could avoid this by "copying" the mutations into a new store with the new schema rather than delete / rewrite in-place.  But I'm probably okay with the data loss risk. I'd add a comment to make it clear it's intentional though...

packages/firestore/test/unit/local/indexeddb_schema.test.ts
```diff
@@ -145,10 +147,65 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+
+    return withDb(2, db => {
+      const sdb = new SimpleDb(db);
+      return sdb.runTransaction('readwrite', [DbMutationBatch.store], txn => {
+        const store = txn.store(DbMutationBatch.store);
+        let p = PersistencePromise.resolve();
+        for (const testMutation of testMutations) {
+          p = p.next(() => store.put(testMutation));
+        }
+        return p;
```
mikelehen: Could you do something like this?

```
return PersistencePromise.waitFor(testMutations.map(mutation => store.put(mutation)));
```

Don't care too much though...

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -648,6 +648,17 @@ describeSpec('Writes:', [], () => {
...
+  specTest('New writes are sent after write failure', [], () => {
+    return spec()
+      .userSets('collection/a', { v: 1 })
+      .failWrite(
+        'collection/a',
+        new RpcError(Code.FAILED_PRECONDITION, 'failure')
+      )
+      .userSets('collection/b', { v: 1 })
+      .writeAcks('collection/b', 2000);
+  });
```
mikelehen: Err... just to confirm, is this related to this PR?

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -375,15 +310,12 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+        const mutationKey = this.rangeForSingleBatchId(batchId);
         // Look up the mutation batch in the store.
-        // PORTING NOTE: because iteration is callback driven in the web,
-        // we just look up the key instead of keeping an open iterator
-        // like iOS.
         return mutationsStore(transaction)
-          .get(mutationKey)
-          .next(dbBatch => {
-            if (dbBatch === null) {
+          .loadAll(DbMutationBatch.userMutationsIndex, mutationKey)
```
mikelehen: Can't we just use the batchId to read the batch directly without using the userMutationsIndex?  (same below)

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -103,142 +89,85 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
-          // is lexographically first. That also means that no other
-          // userIds can come before this one, so we can just exit early.
-          control.done();
-        } else {
-          const nextUser = immediatePredecessor(userId);
-          control.skip([nextUser]);
-        }
-      })
-      .next(() => maxBatchId + 1);
+    return PersistencePromise.resolve();
```
mikelehen: Perhaps add a porting note explaining why start() is empty.

packages/firestore/test/unit/local/indexeddb_schema.test.ts
schmidt-sebastian: Yup. Done.

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: Yeah, this test case fails if the client always just uses the last existing batch from IndexedDB, since both writes end up getting the same batch ID.

packages/firestore/src/local/indexeddb_schema.ts
mikelehen: Oh, nice!

PR #928, opened 2018-06-14
Multitab escaping

schmidt-sebastian:
We need to escape the persistence keys that we use when we parse local storage keys (the persistence keys contain brackets).


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -552,14 +552,21 @@ export class WebStorageSharedClientState implements SharedClientState {
       this.localClientId
     );
     this.activeClients[this.localClientId] = new LocalClientState();
+
+    // Escape the special characters mentioned here:
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Special_characters_meaning_in_regular_expressions
+    const escapedPersistenceKey = persistenceKey.replace(
+      /[\.\\\[\]\^\|\$\(\)\?\:\*\+\{\}\,\=\!]/g,
+      '\\$&'
+    );
```
mikelehen: I don't want to try to understand this regex and I don't know where you got it, so maybe update the doc link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

and use the code provided there?

```
string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
```

schmidt-sebastian: I replaced it with the example they provided. The RegExp above is basically the same (it also escapes '=' which seems to be unnecessary), but it's kind of ugly since it escaped all characters individually. The example nicely works around this.

PR #929, opened 2018-06-14
[Multi-Tab] Adding logging to SharedClientState (small)

schmidt-sebastian:
This PR adds logging for all CRUD operations in SharedClientState, similar to what we do in SimpleDb.

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -718,7 +718,7 @@ export class WebStorageSharedClientState implements SharedClientState {
   }
 
   private removeItem(key: string): void {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -702,10 +702,26 @@ export class WebStorageSharedClientState implements SharedClientState {
...
+    return value;
+  }
+
+  private setItem(key: string, value: string): void {
+    debug(LOG_TAG, 'SET', key, value);
+    this.storage.setItem(key, value);
+  }
+
+  private removeItem(key: string): void {
+    debug(LOG_TAG, 'REMOVE', this.localClientStorageKey);
```
mikelehen: Should this be logging `key`?

PR #930, opened 2018-06-15
[Multi-Tab] Handling mutations during failover

schmidt-sebastian:
This PR adds a test for mutations that are executed during a master failover.

The only client change is that we now remove the client metadata from IndexedDb at shutdown (since it would be included in the next round of master election until it expires).


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -199,20 +199,29 @@ export class IndexedDbPersistence implements Persistence {
   private updateClientMetadataAndTryBecomePrimary(): Promise<void> {
     return this.simpleDb.runTransaction('readwrite', ALL_STORES, txn => {
       const metadataStore = clientMetadataStore(txn);
-      metadataStore.put(
-        new DbClientMetadata(this.clientId, Date.now(), this.inForeground)
-      );
+      return metadataStore
+        .put(new DbClientMetadata(this.clientId, Date.now(), this.inForeground))
+        .next(() =>
+          this.canActAsPrimary(txn).next(canActAsPrimary => {
```
mikelehen: Can you move the paren on line 217 to after `this.canActAsPrimary(txn)` so this looks nicer / is less nested?

Also, just to confirm, nothing actually changed in this method, right? It's just syntax?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -296,20 +305,20 @@ export class IndexedDbPersistence implements Persistence {
...
     this.detachVisibilityHandler();
     this.detachWindowUnloadHook();
-    return this.releasePrimaryLeaseIfHeld().then(() => {
-      this.simpleDb.close();
-      if (deleteData) {
-        return SimpleDb.delete(this.dbName);
-      }
-    });
+    await this.releasePrimaryLeaseIfHeld();
+    await this.removeClientMetadata();
```
mikelehen: FWIW it's probably unlikely that this code will actually run to completion during tab close...  So we shouldn't rely on it.  And if we want to increase the probability of it running, we may want to use Promise.all() or something so things can at least go in parallel (uh, maybe? I guess they both will acquire IndexedDb locks... so really we'd need to refactor into the same transaction).

But if we really want failover to be fast then we need to only rely on localStorage like we do with the zombie thing...  Feel free to just add a TODO(multitab) if you want to do this but not right now.

Issue #850, opened 2018-05-21
onAuthStateChanged concurrency issue


Dipen-Dedania:
I'm facing an issue while checking if user is logged in or not. I have tried both the approaches to check if user is logged in or not.
1. firebase.auth().currentUser
2. firebaseApp.auth().onAuthStateChanged

My use case is something like this. I have hybrid application in which I'm displaying browser. I have login browser in which if user login then it will spawn multiple browsers. So in those browsers I need to check if user is logged-in (fail-check) and if he/she is logged in then I'll need a user info. But sometimes it returns null randomly in some browsers (may be concurrency issue while checking **onAuthStateChanged** listener) so I was trying to debug the issue. I put a breakpoints in both browser (chrome) and tried to refresh both browser at the same time. I came across a weird behavior that even it's a separate browser and they are trying to execute below code, one browser will freeze until the other instance doesn't resume the execution. Refer attached GIF for details. Not sure if it's related to the same file debugging. And it was also working fine before few days. Also tried updating a package to recent one but no luck. Looking forward to hear from you soon.

Environment
  * Operating System version: Windows 10, Chrome
  * Firebase JS SDK version: 4.10.1 / 5.0.3
  * Firebase Product: auth

```
let unsubscribe = firebaseApp.auth().onAuthStateChanged(function (user) {
	if (!user) {
		// User is not logged in
		unsubscribe();
	}
	else {
		// User is logged in - process to next
		unsubscribe();
	}
});
```
![fb_issue2](https://user-images.githubusercontent.com/5135662/40309293-97c511b0-5d26-11e8-8dbe-a11db89c0c63.gif)

Thanks!

bojeil-google:
I don't quite understand what you are trying to say: " I came across a weird behavior that even it's a separate browser and they are trying to execute below code, one browser will freeze until the other instance doesn't resume the execution. "
If it helps, `onAuthStateChanged` will not trigger at the exact time as the tab where the user signed in. There will be a fraction of second delay. This is expected.

Vindexus:
I have also been encountering an issue where the data returned by `onAuthStateChanged` is null. This commonly happens in my dev environment when I save a file and webpack rebuilds and refreshes the app.

On production this issue would sometimes happen on page refresh, or when a link from the app was opened in a new tab.

While trying to debug I did this
```
function getFirebaseAuthCookie () {
[...]

PR #937, opened 2018-06-16
fix(auth): Fixes issue when hybrid indexedDB incorrectly fallsback to localStorage

bojeil-google:
Fixes issue when hybrid indexedDB fallsback to localStorage because the write/read test fails because multiple windows or instances try to write/read/delete from the same indexedDB entry. Instead of a common key __sak, we use a random one minimizing the chances of collision.


bojeil-google:
Issue was reported here:
https://github.com/firebase/firebase-js-sdk/issues/850
https://stackoverflow.com/questions/50262301/firebase-logout-not-working-on-fb-hosting/50299173#50299173

PR #940, opened 2018-06-18
Force refresh token if RPC fails with "Unauthenticated" error

var-const:
"Unauthenticated" is presumed to mean that token is expired (which might happen if local clock is wrong) and retried, subject to the usual backoff logic.


packages/firestore/test/integration/remote/stream.test.ts
```diff
@@ -262,4 +281,51 @@ describe('Write Stream', () => {
...
+              new FirestoreError(Code.UNAVAILABLE, '')
+            );
+            return streamListener.awaitCallback('close');
+          })
+          .then(() => {
+            writeStream.start(streamListener);
+            return streamListener.awaitCallback('open');
+          })
+          .then(() => {
+            expect(credentials.observedStates).to.have.ordered.members([
```
var-const: In practice, the error message isn't particularly helpful (along the lines of `Array(4) does not equal Array(4)`. From a quick search, I only found references to old versions of Chai. Let me know if I'm using wrong comparators.

packages/firestore/test/integration/remote/stream.test.ts
```diff
@@ -262,4 +281,51 @@ describe('Write Stream', () => {
...
+            return streamListener.awaitCallback('close');
+          })
+          .then(() => {
+            writeStream.start(streamListener);
+            return streamListener.awaitCallback('open');
+          })
+          .then(() => {
+            // Simulate a different error -- token should not be invalidated this time.
+            writeStream.handleStreamClose(
+              new FirestoreError(Code.UNAVAILABLE, '')
```
var-const: This might be a bad idea -- feel free to push back. The compiler complains that `handleStreamClose` is private, but `yarn test` can run this test just fine. It's clear that the build process ignores the compile-time error, but I don't know if it's possible to rely on that.

mohshraim:
I was looking for solution to my case; i think this PR will solve it.
My Scenario;
-Saving some documents on offline mode; 
-Connection back after one hour or more.
-Make another operation directly after connection back; before Auth request refresh token.
-Get Error; Unauthenticated; form pending save operations. 
This PR will help??

var-const:
@mohshraim I'm not 100% sure about the exact issue you're experiencing, but yes, it looks like this PR might help. If you find out that the problem persists even with changes from this PR, please open an issue. Thanks!

mohshraim:
@var-const  
I will wait the next release that will include this PR, then will test my issue.
Thanks

PR #942, opened 2018-06-18
Add FcmOptions interface and link property

mmermerkaya:


mmermerkaya:
Done

packages/messaging/test/sw-controller.test.ts
```diff
@@ -638,162 +638,164 @@ describe('Firebase Messaging > *SwController', () => {
...
     });
 
-    it('should open click_action URL for FCM notification (third-party so no window client access)', async () => {
-      const clickAction = '/test-click-action';
-      const waitUntilSpy = sandbox.spy();
-      const event: any = {
-        notification: {
-          data: {
-            FCM_MSG: {
+    ['link', 'click_action'].forEach(key => {
```
dwoffinden: nit: prefer "for (... of ...) {"

go/tsstyle#iterating-arrays

packages/messaging/test/sw-controller.test.ts
```diff
@@ -638,162 +638,164 @@ describe('Firebase Messaging > *SwController', () => {
...
-              }
-            }
-          },
-          close: sandbox.spy()
-        },
-        waitUntil: waitUntilSpy,
-        stopImmediatePropagation: sandbox.spy()
-      };
+                data: {
+                  FCM_MSG: {
```
dwoffinden: nit: since these two paths differ only below FCM_MSG, pull it out? Alternatively, consider a helper method that just builds the event given the key?

const msg = key === 'link'
    ? {notification: {}, fcmOptions: {link}}
    : {notification: {click_action: link}};

event = {
  notification: {
    data: {FCM_MSG: msg},
    close: sandbox.spy()
  },
  waitUntil: waitUntilSpy,
  stopImmediatePropagation: sandbox.spy()
};

packages/messaging/test/sw-controller.test.ts
```diff
@@ -638,162 +638,164 @@ describe('Firebase Messaging > *SwController', () => {
...
+              stopImmediatePropagation: sandbox.spy()
+            };
+          }
+        });
 
-      const openWindowStub = sandbox.stub().callsFake(() => {
-        // Pretend it's a third party URL
-        return null;
-      });
+        it('should open URL (third-party so no window client access)', async () => {
```
dwoffinden: nit: Alec always suggested the "opens URL ..." form over "should open URL ...". It's shorter, and reads better when you consider that this is a property of the SUT that is either true or not? :stuck_out_tongue:

packages/messaging/test/sw-controller.test.ts
mmermerkaya: Sure, why not. I don't think you'll find any style guide conformance around here :smile: I'll rewrite all of these eventually...

packages/messaging/test/sw-controller.test.ts
mmermerkaya: I did consider the helper method but I'd like to keep logic in tests as simple as possible. Still, I guess I can pull FCM_MSG out...

Issue #900, opened 2018-06-04
Generic type 'Observer<V, E>' requires 2 type argument(s).


rtm:
I receive the following error:

> ./../../node_modules/@firebase/storage-types/index.d.ts(80,9): error TS2314: Generic type 'Observer<V, E>' requires 2 type argument(s).

The version of `@firebase/storage-types` is 0.2.3 (brought in by firebase@5.0.4). The line in question is

```
  on(
    event: TaskEvent,
    nextOrObserver?:
      | Observer<UploadTaskSnapshot>
      | null
      | ((a: UploadTaskSnapshot) => any),
    error?: ((a: Error) => any) | null,
    complete?: (Unsubscribe) | null
  ): Function;
```

The repo from which this is being brought in has

```
    "rxjs": "^6.0.0",
    "rxjs-compat": "^6.2.0"
```

hiepxanh:
![screen shot 2018-06-13 at 9 07 40 am](https://user-images.githubusercontent.com/16970990/41326455-41cb5182-6ee9-11e8-9f2f-35881a489727.png)

mmermerkaya:
Looks like `@firebase/storage` has [their own implementation](https://github.com/firebase/firebase-js-sdk/blob/73a586c92afe3f39a844b2be86086fddb6877bb7/packages/storage/src/implementation/observer.ts) but uses `@firebase/util` types in `@firebase/storage-types` (without adding that as a dependency in `package.json`): https://github.com/firebase/firebase-js-sdk/blob/785cc8b49f149068dbabb6f7482ac0763aeec85e/packages/storage-types/index.d.ts#L18

hiepxanh:
sir, is that a bug ? or what should we do?

hiepxanh:
@mmermerkaya is that a bug? what shoud I do sir?

mmermerkaya:
Try updating to the latest Firebase SDK version. See #787 for details.

The only issue I see here is that auth-types and storage-types packages don't have the `@firebase/util` dependency in their `package.json` files even though they are using it, which might have caused your version to stay behind. I opened another issue to track that: #952.

sreekanth-karini:
Update typescript,       `npm install typescript@2.3`       will fix the error.

PR #788, opened 2018-05-08
Fixed Generic type Observer

paulstelzer:
Generic type 'Observer' requires 1 type argument(s). Fix #787

Issue #787, opened 2018-05-08
[Error] Firebase 5.0.0: Generic type 'Observer<T>' requires 1 type argument(s).


paulstelzer:
Issue: Generic type 'Observer<T>' requires 1 type argument(s).

Observer types in Firebase 5.0.0 not fixed on all positions. In commit #770 the interface Observer is now "export interface Observer<T> {...}", but in "/packages/auth-types/index.d.ts" at line 271 and 276 it's still the old one. The same at "/packages/storage-types/index.d.ts" line 80.

### Environment

  * Operating System version: Windows 10
  * Firebase SDK version: 5.0.0

nayfin:
I'm seeing this too. Using `"angularfire2": "5.0.0-rc.8.", "firebase": "4.12.1", "@angular/core": "6.0.2"`. I started getting the following after upgraging to Angular 6. 

```
ERROR in node_modules/@firebase/auth-types/index.d.ts(255,21): error TS2314: Generic type 'Observer<T>' requires 1 type argument(s).
node_modules/@firebase/auth-types/index.d.ts(260,21): error TS2314: Generic type 'Observer<T>'
requires 1 type argument(s).
node_modules/@firebase/messaging-types/index.d.ts(25,38): error TS2314: Generic type 'Observer<T>' requires 1 type argument(s).
node_modules/@firebase/messaging-types/index.d.ts(30,38): error TS2314: Generic type 'Observer<T>' requires 1 type argument(s).
node_modules/@firebase/storage-types/index.d.ts(77,22): error TS2314: Generic type 'Observer<T>' requires 1 type argument(s).
```

Any help would be appreciated. I've been trying every possible configuration of firebase and angularfire that I can find.

paulstelzer:
@nayfin use the latest versions of angularfire2 and firebase. then everything should work again

sreekanth-karini:
Update typescript, `npm install typescript@2.3 --save` will fix the error.

PR #951, opened 2018-06-21
Hot fix: small quick update types for Server Side Rendering ( Angular Universal)

hiepxanh:
Hello, I'm fixing types error when compiling AOT with angular universal

### Discussion
  fixing this issue [Generic type 'Observer<V, E>' requires 2 type argument(s)](https://github.com/firebase/firebase-js-sdk/issues/900)

### Testing

tested

### API Changes

nope

### DIFF
--- a/packages/auth-types/index.d.ts
+++ b/packages/auth-types/index.d.ts
@@ -279,12 +279,12 @@ export class FirebaseAuth {
   languageCode: string | null;
   settings: AuthSettings;
   onAuthStateChanged(
-    nextOrObserver: Observer<any> | ((a: User | null) => any),
+    nextOrObserver: Observer<any,any> | ((a: User | null) => any),
     error?: (a: Error) => any,
     completed?: Unsubscribe
   ): Unsubscribe;
   onIdTokenChanged(
-    nextOrObserver: Observer<any> | ((a: User | null) => any),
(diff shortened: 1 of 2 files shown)


packages/auth-types/index.d.ts
```diff
@@ -279,12 +279,12 @@ export class FirebaseAuth {
   languageCode: string | null;
   settings: AuthSettings;
   onAuthStateChanged(
-    nextOrObserver: Observer<any> | ((a: User | null) => any),
+    nextOrObserver: Observer<any,Error> | ((a: User | null) => any),
```
bojeil-google: Add a space after the comma here and elsewhere.

hiepxanh:
@bojeil-google sir yes sir, sorry for my fault. I just updated it. Can you approve my pull request ? I'm very appreciate for that

hiepxanh:
screen shot 2018-06-21 at 4 37 02 pm
why is that: firebase-js-sdk project I cloned and run on the right is different from the code build on npm I install ? that causing the problem. 

I believe that the npm version available for now is out of date compare with github code. If you fix that, my PR is unnecessary and I will close it. @bojeil-google

mmermerkaya:
This should not be merged. As you can see from the Travis error, Observer requires one type argument.

The issue was #787 and it was fixed in #788. It will work if you use the latest version of the Firebase packages.

PR #955, opened 2018-06-21
[Multi-Tab] Don't disable network on shutdown

schmidt-sebastian:
This should bring make the multitab branch green again.

The additional asserts of https://github.com/firebase/firebase-js-sdk/pull/903 uncovered a problem in the Multi-Tab branch since the explicit call to `remoteStore.disableNetwork()` raised `isFromCache:true` events.

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -707,8 +707,8 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     return this.enableNetwork();
   }
 
-  shutdown(): Promise<void> {
-    return this.disableNetwork();
+  async shutdown(): Promise<void> {
+    this.networkAllowed = false;
   }
 
   getRemoteKeysForTarget(targetId: TargetId): DocumentKeySet {


wilhuff:
This doesn't seem right.

Users can delete the current firebase app in a tab and when they do so they expect Firestore to release resources including current connections.

In a multi-tab world, I would expect a shutdown to trigger release of mastership and end up with connections torn down in the current tab.

schmidt-sebastian:
> Users can delete the current firebase app in a tab and when they do so they expect Firestore to release resources including current connections.
> 
> In a multi-tab world, I would expect a shutdown to trigger release of mastership and end up with connections torn down in the current tab.

This already happens.

- FirestoreClient calls RemoteStore.shutdown() which changes the network state to Unknown.
- The shutdown handler in IndexedDb removes the primary lease on shutdown.

What this PR does is remove an additional call to `disableNetwork` that set the network state to Offline in between all the shutdown logic.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -707,8 +707,8 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     return this.enableNetwork();
   }
 
-  shutdown(): Promise<void> {
-    return this.disableNetwork();
+  async shutdown(): Promise<void> {
+    this.networkAllowed = false;
```
mikelehen: But don't we _want_ to close / clean up network connections when the client is shut down?

schmidt-sebastian: FirestoreClient does this on shutdown: https://github.com/firebase/firebase-js-sdk/blob/f6fd8f52be282cc8e5f2e7a21948262eaffb343c/packages/firestore/src/core/firestore_client.ts#L372

mikelehen: Agh, sorry. I missed this and your conversation with Gil about this already...

PR #959, opened 2018-06-22
Workaround for IndexedDb bug

schmidt-sebastian:
This works around an IndexedDb bug in Chrome and Firefox that omits index entires for auto-generated keys.

https://bugs.chromium.org/p/chromium/issues/detail?id=701972

### DIFF
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -161,34 +161,40 @@ export class IndexedDbMutationQueue implements MutationQueue {
     localWriteTime: Timestamp,
     mutations: Mutation[]
   ): PersistencePromise<MutationBatch> {
-    const batch = new MutationBatch(BATCHID_UNKNOWN, localWriteTime, mutations);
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -161,34 +161,40 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+    const documentStore = documentMutationsStore(transaction);
+    const mutationStore = mutationsStore(transaction);
 
-    return mutationsStore(transaction)
-      .add(dbBatch)
-      .next(batchId => {
-        batch.batchId = batchId;
-        this.documentKeysByBatchId[batchId] = batch.keys();
+    // The IndexedDb implementation in Chrome (and Firefox) does not handle
+    // compound indices with include auto-generated keys correctly. To ensure
```
mikelehen: with include => "with included?" or "which include"?

(this comment is great, by the way. Thanks for including it)

mikelehen: n/m you already fixed this. :)

PR #960, opened 2018-06-25
Remove startsWith because it's not supported in IE11

### LINKED ISSUE #958: Firebase 5.0 (database) is broken on IE11
* Operating System version: Any windows
  * Browser version: IE11
  * Firebase SDK version: 5.0.4
  * Firebase Product: Database
  * Webpack version: 4

#### Steps to reproduce:

Just try to use the realtime database on IE11, it'll complain that 

`Object doesn't support property or method 'startsWith'`

The offending line was added by db389db, you can see the exact line [here](https://github.com/firebase/firebase-js-sdk/commit/db389db97bf3698384a4d708f4be70beecfabb0c#diff-2f6dec17e1c2745a2b3e6ad8143df529) and it's still there in version 5.1

#### Relevant Code:

Just about anything using

```
firebase.database().ref(path).on('value', (data) => {});
```

#### Question

Is this intentional? I couldn't find [anything](https://firebase.google.com/docs/web/setup) about the minimum supported browser or wether we're required to use polyfills / transpilers to use this library.
Am I the only one that doesn't load polyfills on IE11?

piuccio:
Make IE great again. Fixes #958 

I tried adding the tslint [rule](https://github.com/Microsoft/tslint-microsoft-contrib) `"no-unsupported-browser-code": [true, ["IE 11"]],` but as it turns out the package `database` doesn't lint the code.
Linting is only enabled in `firestore` and `messaging`

I went a bit deeper trying to copy the config from `firestore` but as it turns out, there are loads of lint errors and I don't feel like touching more code than necessary.

For the moment simply remote the unsupported method in IE11, if anyone cares please add linting and configure it so that these errors won't show up again in the future

### DIFF
--- a/packages/database/src/core/util/libs/parser.ts
+++ b/packages/database/src/core/util/libs/parser.ts
@@ -43,7 +43,7 @@ function decodePath(pathString: string): string {
  */
 function decodeQuery(queryString: string): { [key: string]: string } {
   let results = {};
-  if (queryString.startsWith('?')) {
+  if (queryString.charAt(0) === '?') {
     queryString = queryString.substring(1);
   }
   for (const segment of queryString.split('&')) {


piuccio:
I signed it!

schmidt-sebastian:
We ran into this issue earlier and added support for `startsWith` via Polyfill: https://www.npmjs.com/package/core-js#ecmascript-6-collections. This Polyfill gets installed when you pull in firebase-app.js.

That being said, this fix is fine as is, and I am ok with merging it if ends up fixing some clients.

PR #968, opened 2018-06-28
[Multi-Tab] Recomputing views after failover

schmidt-sebastian:
This PR adds the functionality to reconcile views with the truth from persistence as tabs gain the primary lease. To do this, it updates state internal to the view, so we might want to wait with merging this until after the View refactor.

This also fixed an issue in the Spec test that cleared the target IDs on client shutdown. They should be cleared on global shutdown (or just re-recreated at test startup like we already do).

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -66,7 +66,6 @@ import {
 } from '../local/shared_client_state_syncer';
 import { ClientId, SharedClientState } from '../local/shared_client_state';
 import { SortedSet } from '../util/sorted_set';
-import * as objUtils from '../util/obj';
 
 const LOG_TAG = 'SyncEngine';
 
@@ -139,6 +138,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     [uidKey: string]: SortedMap<BatchId, Deferred<void>>;
   };
   private limboTargetIdGenerator = TargetIdGenerator.forSyncEngine();
+
+  // The primary state is set to `true` or `false` immediately after Firestore
+  // startup. In the interim, a client should only be considered primary if `
+  // isPrimary` is true.
   private isPrimary: undefined | boolean = undefined;
 
   constructor(
@@ -185,19 +188,23 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       const status = this.sharedClientState.addLocalQueryTarget(
         queryData.targetId
       );
-      return this.initializeListenAndNotifyViews(
+      return this.initializeViewAndComputeInitialSnapshot(
         queryData,
         status === 'current'
-      ).then(() => {
+      ).then(viewSnapshot => {
+        if (this.isPrimary) {
+          this.remoteStore.listen(queryData);
+        }
+        this.viewHandler!([viewSnapshot]);
         return queryData.targetId;
       });
     });
   }
 
-  private initializeListenAndNotifyViews(
+  private initializeViewAndComputeInitialSnapshot(
     queryData: QueryData,
     current: boolean
-  ): Promise<void> {
+  ): Promise<ViewSnapshot> {
     const query = queryData.query;
 
     return this.localStore.executeQuery(query).then(docs => {
@@ -232,25 +239,24 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
           );
           this.queryViewsByQuery.set(query, data);
           this.queryViewsByTarget[queryData.targetId] = data;
-          this.viewHandler!([viewChange.snapshot!]);
-          if (this.isPrimary) {
-            this.remoteStore.listen(queryData);
-          }
+          return viewChange.snapshot!;
         });
     });
   }
 
-  private synchronizeLocalView(queryData: QueryData): Promise<void> {
-    const query = queryData.query;
-
-    return this.localStore.executeQuery(query).then(docs => {
-      return this.localStore
-        .remoteDocumentKeys(queryData.targetId)
-        .then(async remoteKeys => {
-          const queryView = this.queryViewsByTarget[queryData.targetId];
-          queryView.view.synchronizeWithRemoteKeys(remoteKeys);
-        });
-    });
+  /**
+   * Reconcile the list of synced documents in the local views with those from
+   * persistence.
+   */
+  // PORTING NOTE: Multi-tab only.
+  private async synchronizeLocalView(queryData: QueryData): Promise<void> {
+    return this.localStore
+      .remoteDocumentKeys(queryData.targetId)
+      .then(async remoteKeys => {
+        const queryView = this.queryViewsByTarget[queryData.targetId];
+        assert(!!queryView, 'Expected queryView to be defined');
+        queryView.view.synchronizeWithRemoteKeys(remoteKeys);
+      });
   }
 
   /** Stops listening to the query. */
@@ -736,44 +742,50 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
 
   // PORTING NOTE: Multi-tab only
   async applyPrimaryState(isPrimary: boolean): Promise<void> {
-    const stateUpgraded = this.isPrimary === false && isPrimary;
+    if (isPrimary === true && this.isPrimary !== true) {
+      this.isPrimary = true;
 
-    this.isPrimary = isPrimary;
-    if (this.isPrimary) {
+      // Secondary tabs only maintain Views for their local listeners and the
+      // Views internal state may not be 100% populated (in particular
+      // secondary tabs don't track syncedDocuments, the set of documents the
+      // server considers to be in the target). So when a secondary becomes
+      // primary, we need to need to make sure that all views for all targets
+      // match the state on disk.
+      let p = Promise.resolve();
+      const activeTargets = this.sharedClientState.getAllActiveQueryTargets();
+      activeTargets.forEach(targetId => {
+        p = p.then(async () => {
+          let queryData;
+          const query = await this.localStore.getQueryForTarget(targetId);
+          if (this.queryViewsByTarget[targetId] === undefined) {
+            // For queries that never executed on this client, we need to
+            // allocate the query in LocalStore and initialize a new View.
+            queryData = await this.localStore.allocateQuery(query);
+            await this.initializeViewAndComputeInitialSnapshot(
+              queryData,
+              false
+            );
+          } else {
+            // For queries that have a local View, we need to update their state
+            // in LocalStore (as the resume token and the snapshot version
+            // might have changed) and reconcile their views with the persisted
+            // state (the list of syncedDocuments may have gotten out of sync).
+            await this.localStore.releaseQuery(query, true);
+            queryData = await this.localStore.allocateQuery(query);
+            await this.synchronizeLocalView(queryData);
+          }
+          this.remoteStore.listen(queryData);
+        });
+      });
+      await p;
+
       if (this.networkAllowed) {
         await this.remoteStore.enableNetwork();
       }
+    } else if (isPrimary === false && this.isPrimary !== false) {
+      this.isPrimary = false;
 
-      if (stateUpgraded) {
-        // Secondary tabs do not update their set of synced keys. When a
-        // secondary tab acquires a primary lease, we need to reconcile our
-        // state with the state in persistence. To do this, we release and
-        // re-allocate all active queries and update all views accordingly.
-        objUtils.forEachNumber(
-          this.queryViewsByTarget,
-          async (targetId, queryView) => {
-            await this.localStore.releaseQuery(
-              queryView.query,
-              /*keepPersistedQueryData=*/ true
-            );
-          }
-        );
-
-        let p = Promise.resolve();
-        const activeTargets = this.sharedClientState.getAllActiveQueryTargets();
-        activeTargets.forEach(targetId => {
-          p = p.then(async () => {
-            const queryData = await this.localStore.getQueryDataForTarget(
-              targetId
-            );
-            this.remoteStore.listen(queryData);
-            await this.localStore.allocateQuery(queryData.query);
-            await this.synchronizeLocalView(queryData);
-          });
-        });
-        await p;
-      }
-    } else {
+      // TODO: Unlisten from all active targets
       await this.remoteStore.disableNetwork();
     }
   }
@@ -833,10 +845,14 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
         !this.queryViewsByTarget[targetId],
         'Trying to add an already active target'
       );
-      const queryData = await this.localStore.getQueryDataForTarget(targetId);
-      assert(!!queryData, `Query data for active target ${targetId} not found`);
-      await this.localStore.allocateQuery(queryData.query);
-      await this.initializeListenAndNotifyViews(queryData, /*current=*/ false);
+      const query = await this.localStore.getQueryForTarget(targetId);
+      assert(!!query, `Query data for active target ${targetId} not found`);
+      const queryData = await this.localStore.allocateQuery(query);
+      await this.initializeViewAndComputeInitialSnapshot(
+        queryData,
+        /*current=*/ false
+      );
+      this.remoteStore.listen(queryData);
     }
 
     for (const targetId of removed) {
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -66,6 +66,7 @@ import {
 } from '../local/shared_client_state_syncer';
 import { ClientId, SharedClientState } from '../local/shared_client_state';
 import { SortedSet } from '../util/sorted_set';
+import * as objUtils from '../util/obj';
 
 const LOG_TAG = 'SyncEngine';
 
@@ -138,7 +139,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     [uidKey: string]: SortedMap<BatchId, Deferred<void>>;
   };
   private limboTargetIdGenerator = TargetIdGenerator.forSyncEngine();
-  private isPrimary = false;
+  private isPrimary: undefined | boolean = undefined;
 
   constructor(
     private localStore: LocalStore,
@@ -150,7 +151,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
 
   // Only used for testing.
   get isPrimaryClient(): boolean {
-    return this.isPrimary;
+    return this.isPrimary === true;
   }
 
   /** Subscribes view and error handler. Can be called only once. */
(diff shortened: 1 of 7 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -239,6 +240,19 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
   }
 
+  private synchronizeLocalView(queryData: QueryData): Promise<void> {
+    const query = queryData.query;
+
+    return this.localStore.executeQuery(query).then(docs => {
+      return this.localStore
+        .remoteDocumentKeys(queryData.targetId)
+        .then(async remoteKeys => {
+          const queryView = this.queryViewsByTarget[queryData.targetId];
```
mikelehen: As alluded to elsewhere, I think queryView could be undefined for targets from other tabs...  Depending on how you fix that, it may make sense to add an assert here (the following line will probably crash anyway, but an assert will be clearer).

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +735,46 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-  applyPrimaryState(isPrimary: boolean): Promise<void> {
+  async applyPrimaryState(isPrimary: boolean): Promise<void> {
+    const stateUpgraded = this.isPrimary === false && isPrimary;
+
     this.isPrimary = isPrimary;
-    if (this.isPrimary && this.networkAllowed) {
-      return this.remoteStore.enableNetwork();
+    if (this.isPrimary) {
+      if (this.networkAllowed) {
+        await this.remoteStore.enableNetwork();
```
mikelehen: FWIW it seems a little odd that we do this even when stateUpgraded == false.  It's tempting to only call enableNetwork() when we become primary and only call disableNetwork() when we lose primary.  So this method might end up:

```
const becamePrimary = this.isPrimary === false && isPrimary;
const lostPrimary = this.isPrimary === true && !isPrimary;
if (becamePrimary) {
  ...
} else if (lostPrimary) {
  ...
}
```

I also wonder if we can/should assert anything about the state of the RemoteStore before enabling the network.  I worry about the state of listenTargets, etc. Are we sure that all of the RemoteStore state is guaranteed to be in a known good state in secondary tabs?  It might be nice to add some asserts or comments about this.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -287,4 +288,49 @@ describeSpec('Limbo Documents:', [], () => {
...
     }
   );
+
+  specTest(
+    'Limbo documents are resolved after primary tab failover',
+    ['multi-client'],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const doc1 = doc('collection/a', 1000, { key: 'a' });
+      const doc2 = doc('collection/b', 1001, { key: 'b' });
```
mikelehen: nit: I like using docA / docB (to match the actual document names).

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -287,4 +288,49 @@ describeSpec('Limbo Documents:', [], () => {
...
+      const deletedDoc2 = deletedDoc('collection/b', 1005);
+
+      return (
+        client(0, false)
```
mikelehen: FWIW I couldn't remember or guess what this boolean argument was.  Please add `/*gcEnabled=*/` or make it so we can do .client(0).withGcEnabled(false)... or something.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -287,4 +288,49 @@ describeSpec('Limbo Documents:', [], () => {
...
+          .watchAcksFull(query, 1002, doc1, doc2)
+          .client(1)
+          .expectEvents(query, { added: [doc1, doc2] })
+          .client(0)
+          .watchRemovesDoc(doc2.key, query)
+          .watchSnapshots(1003)
+          .expectLimboDocs(doc2.key)
+          .shutdown()
+          .client(1)
+          .expectEvents(query, { fromCache: true })
```
mikelehen: Hrm...  Not 100% sure if this is expected / desirable.  I think ideally we'd pick a new primary and let it try to connect to the backend before triggering fromCache: true in all the secondary tabs.  Perhaps add a TODO(multi-tab)?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -702,4 +703,38 @@ describeSpec('Listens:', [], () => {
...
     }
   );
+
+  specTest(
+    'Query is re-listened to after primary tab failover',
+    ['multi-client'],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const docA = doc('collection/a', 1000, { key: 'a' });
+      const docB = doc('collection/b', 2000, { key: 'a' });
```
mikelehen: key: 'b'

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -287,4 +288,49 @@ describeSpec('Limbo Documents:', [], () => {
...
+    ['multi-client'],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const doc1 = doc('collection/a', 1000, { key: 'a' });
+      const doc2 = doc('collection/b', 1001, { key: 'b' });
+      const deletedDoc2 = deletedDoc('collection/b', 1005);
+
+      return (
+        client(0, false)
+          .becomeVisible()
```
mikelehen: Hrm.  I thought this was here to make sure the client becomes primary, but in your other test you omit it and have .expectPrimaryState(true) instead?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -702,4 +703,38 @@ describeSpec('Listens:', [], () => {
...
+        .watchAcksFull(query, 1000, docA)
+        .client(1)
+        .expectEvents(query, { added: [docA] })
+        .client(2)
+        .userListens(query)
+        .expectEvents(query, { added: [docA] })
+        .client(0)
+        .shutdown()
+        .client(1)
+        .runTimer(TimerId.ClientMetadataRefresh)
```
mikelehen: Per my other comments, I think you need a test where e.g. client(2) has a listen that client(1) does not...

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -984,10 +984,14 @@ abstract class TestRunner {
     this.validateLimboDocs();
     // Always validate that the expected active targets match the actual active
     // targets
-    this.validateActiveTargets();
+    await this.validateActiveTargets();
   }
 
   private validateLimboDocs(): void {
+    if (!this.started) {
```
mikelehen: uh... this looks a bit odd and should perhaps be commented (why do we call validateLimboDocs() when we're not started?)

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1010,12 +1014,20 @@ abstract class TestRunner {
...
       return;
     }
 
+    // In multi-tab mode, we cannot rely on the `waitForWatchOpen` call in
+    // `doUserListen` since primary tabs may execute queries from other tabs
+    // without any direct user interaction.
+    if (!obj.isEmpty(this.expectedActiveTargets)) {
+      await this.connection.waitForWatchOpen();
+      await this.queue.drain();
+    }
```
mikelehen: Hrm.  It seems a bit awkward to do this as part of every step (and do we end up draining the queue an extra time?)...  But I guess we don't have a mechanism to only do it on client changes or primary changes or something?

I don't care too much, but maybe it's worth a `// TODO(multitab): We'd ideally only do this after a primary change.` or something?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +735,46 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+      }
+
+      if (stateUpgraded) {
+        // Secondary tabs do not update their set of synced keys. When a
+        // secondary tab acquires a primary lease, we need to reconcile our
+        // state with the state in persistence. To do this, we release and
+        // re-allocate all active queries and update all views accordingly.
```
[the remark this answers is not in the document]
schmidt-sebastian: I updated the comment. I was going to add more detail here, but it got a little confusing. I added two more specific comments to the section below.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +735,46 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+        objUtils.forEachNumber(
+          this.queryViewsByTarget,
+          async (targetId, queryView) => {
+            await this.localStore.releaseQuery(
+              queryView.query,
+              /*keepPersistedQueryData=*/ true
+            );
```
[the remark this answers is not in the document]
schmidt-sebastian: I don't know that we 100% need to - but the snapshot version and the resume token stored in the LocalStore will go out of sync (they get updated with the next RemoteEvent). While releasing/allocating is a little bit more work than needed, it doesn't require any changes to LocalStore.

BTW, I noticed that I am retrieving the query data from LocalStore and then asking LocalStore to update it again. To make this less of a round-trip, I changed `getQueryDataForTarget` to only return the query.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +735,46 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+        let p = Promise.resolve();
+        const activeTargets = this.sharedClientState.getAllActiveQueryTargets();
+        activeTargets.forEach(targetId => {
+          p = p.then(async () => {
+            const queryData = await this.localStore.getQueryDataForTarget(
+              targetId
+            );
+            this.remoteStore.listen(queryData);
+            await this.localStore.allocateQuery(queryData.query);
+            await this.synchronizeLocalView(queryData);
```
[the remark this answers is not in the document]
schmidt-sebastian: I split this into two case: The first cases uses `initializeViewAndComputeInitialSnapshot` (formerly initializeListenAndNotifyViews) to populate the view, the second case uses `synchronizeLocalView`.

As part of this change, the renamed `initializeListenAndNotifyViews` no longer raises the view snapshots or registers with the remote store (which affects two of its callers).

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Fixed.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Added the "gcEnabled" comment.

I am not able to use `withGcEnabled` since `client(0)`  already drains the queue, and `withGcEnabled` rightfully validates that no steps have been added.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: It's not really needed anymore since we drain the queue to begin with. We just needed one step (which for the first client can be a no-op).

We still need `becomeVisible` in the tests that verify that if there are any foreground tabs only foreground tabs can acquire the primary lease.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: We raise `fromCache` since the query is in limbo.

packages/firestore/test/unit/specs/listen_spec.test.ts
schmidt-sebastian: Fixed

packages/firestore/test/unit/specs/listen_spec.test.ts
schmidt-sebastian: Agreed! Added a test for this below (`Query is established in new primary tab`).

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: I moved this up to `validateStateExpectations` and added a comment. It's a bit clearer now - but I am now also not validating the active targets (which we might still want to do).

Another option would be to reset the limbo docs to an empty array in SyncEngine and call `.expectLimboDocs()` after `.shutdown` in the spec tests.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: I looked at various places to put this... but there's no real good alternative. We could wait for the watch stream to open when we switch the clients, but it would seem really out of place (especially since we have to check that there are expected targets).

TODO added.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +735,46 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
   }
 
   // PORTING NOTE: Multi-tab only
-  applyPrimaryState(isPrimary: boolean): Promise<void> {
+  async applyPrimaryState(isPrimary: boolean): Promise<void> {
+    const stateUpgraded = this.isPrimary === false && isPrimary;
```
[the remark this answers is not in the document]
schmidt-sebastian: I removed this and now handle it as part of the if statement (similar to your suggestion but without the local helpers)

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I used `becamePrimary` and `lostPrimary` inline. We don't need to do any work when the state doesn't change.

As for brining the RemoteStore into a good state, we need to unlisten all targets when we lose the primary lease (or always call listen/unlisten even in secondary tabs). I haven't added a test for this yet (all existing tests just shut down the client) and will do this work when "throttling" a client becomes possible from the spec tests.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +741,52 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+          }
+          this.remoteStore.listen(queryData);
+        });
+      });
+      await p;
+
+      if (this.networkAllowed) {
+        await this.remoteStore.enableNetwork();
+      }
+    } else if (isPrimary === false && this.isPrimary !== false) {
```
mikelehen: I think we could also get rid of the subtlety with isPrimary===undefined by just calling `remoteStore.disableNetwork()` by default (in start() I guess) and wait for applyPrimaryState() to be called before we enable it.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -721,12 +741,52 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+            queryData = await this.localStore.allocateQuery(query);
+            await this.initializeViewAndComputeInitialSnapshot(
+              queryData,
+              false
+            );
+          } else {
+            // For queries that have a local View, we need to update their state
+            // in LocalStore (as the resume token and the snapshot version
+            // might have changed) and reconcile their views with the persisted
+            // state (the list of syncedDocuments may have gotten out of sync).
```
mikelehen: Great comment, thanks.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -231,14 +239,26 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
         });
     });
   }
 
+  /**
+   * Reconcile the list of synced documents in the local views with those from
+   * persistence.
+   */
+  // PORTING NOTE: Multi-tab only.
+  private async synchronizeLocalView(queryData: QueryData): Promise<void> {
```
mikelehen: This method seems to only need the targetId, not the full queryData.

packages/firestore/src/local/local_store.ts
```diff
@@ -971,10 +971,16 @@ export class LocalStore {
   }
 
   // PORTING NOTE: Multi-tab only.
-  getQueryDataForTarget(targetId: TargetId): Promise<QueryData | null> {
-    return this.persistence.runTransaction('Get query data', false, txn => {
-      return this.queryCache.getQueryDataForTarget(txn, targetId);
-    });
+  getQueryForTarget(targetId: TargetId): Promise<Query | null> {
+    if (this.targetIds[targetId]) {
```
mikelehen: I wonder if we should rename targetIds to allocatedQueriesByTargetId or similar.  As-is it's not entirely obvious if/when queries will be in targetIds.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -737,4 +737,32 @@ describeSpec('Listens:', [], () => {
...
+
+    // Client 1 and Client 3 listen to the same query. When client 1 shuts
+    // down, client 2 becomes primary and takes ownership of a query it
+    // did not previously listen to.
+    return client(0, /*withGcEnabled=*/ false)
+      .expectPrimaryState(true)
+      .userListens(query)
+      .watchAcksFull(query, 1000, docA)
+      .expectEvents(query, { added: [docA] })
+      .client(1)
```
mikelehen: just to sanity-check, we need this client(1) line even though there are no steps?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -138,7 +138,11 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     [uidKey: string]: SortedMap<BatchId, Deferred<void>>;
   };
   private limboTargetIdGenerator = TargetIdGenerator.forSyncEngine();
-  private isPrimary = false;
+
+  // The primary state is set to `true` or `false` immediately after Firestore
+  // startup. In the interim, a client should only be considered primary if `
+  // isPrimary` is true.
```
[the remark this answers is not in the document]
schmidt-sebastian: Line break fixed.

I opted to not change the current behavior for `isPrimary`. Calling `disableNetwork` in the constructor causes NPEs, calling in start() (instead of enable network) causes a bunch of test failures since `applyPrimaryState` might have been called beforehand, and passing the primary state into the constructor/start() is not an easy task since it needs to be read from IndexedDb. Sorry :/

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Good catch. Fixed.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: As discussed earlier, this created some hurdles so I opted to leave it as is.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I had this crazy though that we could just get rid of state... but then I didn't keep pursuing. I added a TODO to rename it to `queryDataByTarget`, which would  match the names in SyncEngine. I didn't want to do this as part of this PR as  I fear that this will create major merge conflicts at some point.

packages/firestore/test/unit/specs/listen_spec.test.ts
schmidt-sebastian: This starts up and initializes the second client. Comment added.

We probably don't need to do this here, but that would be a different test case: What happens when a client starts up, becomes primary, and there is already pending work. There's a TODO in SyncEngine to deal with this.

PR #969, opened 2018-06-28
Add spec test support for target-scoped resume tokens

schmidt-sebastian:
This PR adds some SpecTest functionality that I thought I needed for Multi-Tab but was then able to proceed without. This give us the option to either just ignore that this ever happened, or to see if we want to merge this change to Master nonetheless. 

BTW, this might also be a good time to treat `watchSnapshot` as a separate operation altogether and not piggy-back on the previous watch operation.


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1109,7 +1124,7 @@ export interface SpecStep {
    * Optional snapshot version that can be additionally specified on any other
    * watch event
    */
-  watchSnapshot?: SpecSnapshotVersion;
+  watchSnapshot?: SpecWatchSnapshot;
```
mikelehen: The comment ("snapshot version") is a bit out of date (maybe just change to "snapshot"?)

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -556,7 +559,7 @@ abstract class TestRunner {
 
   private doWatchAck(
     ackedTargets: SpecWatchAck,
-    watchSnapshot?: SpecSnapshotVersion
+    watchSnapshot?: SpecWatchSnapshot
```
mikelehen: FWIW it seems like we could probably remove "watchSnapshot?" from all of these and just have the test do the watchSnapshot step explicitly as a separate step.  I don't care strongly though.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -481,9 +481,20 @@ export class SpecBuilder {
...
-  watchSnapshots(version: TestSnapshotVersion): SpecBuilder {
+  watchSnapshots(
+    version: TestSnapshotVersion,
+    targets?: Query[],
+    resumeToken?: string
+  ): SpecBuilder {
     this.assertStep('Watch snapshot requires previous watch step');
-    this.currentStep!.watchSnapshot = version;
+
+    if (this.currentStep!.watchSnapshot !== undefined) {
```
mikelehen: Could we just always make the watchSnapshot a separate step (i.e. remove this check? and perhaps the assert above?)

schmidt-sebastian: Yeah, I considered that (and added it as a suggestion to the PR description. cough cough). Now consider it done.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Done

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -685,27 +688,39 @@ abstract class TestRunner {
...
-  ): Promise<void> {
-    let protoJSON = this.serializer.toTestWatchChange(watchChange);
-    this.connection.watchStream!.callOnMessage(protoJSON);
-
+  private doWatchSnapshot(watchSnapshot: SpecWatchSnapshot): Promise<void> {
     // The client will only respond to watchSnapshots if they are on a target
     // change with an empty set of target IDs. So we should be sure to send a
     // separate event. TODO(klimt): We should make the spec tests behave more
     // like the backend. Alternatively, we could hook in to the system at a
     // level higher than the stream.
```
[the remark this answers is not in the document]
schmidt-sebastian: TODO removed.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: Done. Updated the rest of this too.

PR #970, opened 2018-06-29
[Multi-Tab] Handling user changes

schmidt-sebastian:
Spec tests for user changes + small bugfix.

### DIFF
--- a/packages/firestore/test/unit/specs/write_spec.test.ts
+++ b/packages/firestore/test/unit/specs/write_spec.test.ts
@@ -1090,4 +1090,43 @@ describeSpec('Writes:', [], () => {
         })
     );
   });
+
+  specTest('Mutations are scoped by user', [, 'multi-client'], () => {
+    const query = Query.atPath(path('collection'));
+    const docALocal = doc(
+        'collection/a',
+        0,
+        { v: 1 },
+        { hasLocalMutations: true }
+    );
+    const docBLocal = doc(
+        'collection/b',
+        0,
+        { v: 1 },
+        { hasLocalMutations: true }
+    );
+
+    return (
+        client(0)
+            .changeUser('user1')
+            .userSets('collection/a', { v: 1 })
+            .client(1)
+            .changeUser('user2')
+            .userSets('collection/b', { v: 1 })
+                .client(0)
+                .userListens(query)
+            .expectEvents(query, {
+              added: [docALocal],
+              fromCache: true,
+              hasPendingWrites: true
+            })
+            .client(1)
+            .userListens(query)
+            .expectEvents(query, {
+              added: [docBLocal],
+              fromCache: true,
+              hasPendingWrites: true
+            })
+    );
+  });
 });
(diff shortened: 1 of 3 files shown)


packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -996,4 +996,98 @@ describeSpec('Writes:', [], () => {
...
+      { hasLocalMutations: true }
+    );
+    const docCLocal = doc(
+      'collection/c',
+      0,
+      { v: 1 },
+      { hasLocalMutations: true }
+    );
+
+    // Firebase Auth attempts to rapidly synchronize user changes across tab. We
```
mikelehen: tabs

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -996,4 +996,98 @@ describeSpec('Writes:', [], () => {
...
+    // emulate this behavior in this spec tests by calling `changeUser` manually
+    // for all clients.
+    return (
```
mikelehen: Is this paren necessary?  My brain explodes trying to find the match.

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -996,4 +996,98 @@ describeSpec('Writes:', [], () => {
...
+    const docCLocal = doc(
+      'collection/c',
+      0,
+      { v: 1 },
+      { hasLocalMutations: true }
+    );
+
+    // Firebase Auth attempts to rapidly synchronize user changes across tab. We
+    // emulate this behavior in this spec tests by calling `changeUser` manually
+    // for all clients.
```
mikelehen: FWIW, I think we should probably tolerate if auth doesn't successfully do this (or there's a delay)...  I'm tempted to ask you to write a test or two that intentionally has different tabs operating as different users and makes sure we still end up with sensible well-defined behavior.  But you could also try to convince me that it will "just work" or that we shouldn't really care (because we can trust auth to synchronize?)...  WDYT?

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: Done.

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: It took me 10 minutes to write this test and then about 2 hours to figure out why it wasn't passing. I was changing the user at the wrong time, and my mutations didn't get synced properly. If a user changes happens in tab A, but hasn't happened in tab B, then tab B will ignore the mutations for the new user and pick them up as we reload the mutation queue. Tab A will also ignore the mutations from the old user (this is all done in SharedUserState).

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: I believe `prettier` adds them. Let me remove it and see if it gets re-added :)

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: ... oh and I also just changed my mind on this and added a test.

Issue #975, opened 2018-07-02
FR: Support for manually setting the identity toolkit URL


matisiekpl:
### [REQUIRED] Describe your environment

  * Operating System version: Microsoft Windows 10
  * Browser version: Google Chrome 67
  * Firebase SDK version: 5.2.0
  * Firebase Product: auth

### [REQUIRED] Describe the problem

Hello.
I'm developing custom implementation of Google Identity Toolkit, which will help me to build private firebase implementation for testing apps via automated test runner. Now I can only specify Database URL and other parameters, but no Authentication URL, thus I cannot test my server with Firebase JS SDK. Can you implement a options param, which will be passed in Firebase initializeApp method to override default URL (https://www.googleapis.com/identitytoolkit/v3/relyingparty/)?

With best regards,
Mateusz Wozniak.

PR #977, opened 2018-07-03
Custom auth endpoints config

MatisiekPL:
Hey,
As I described in issue #975, I added not required option passing in initializeApp options, which provide custom auth servers endpoints. Option named "authConfig" is optional, and, if is passed, it must be an object. All tests in my private Travis CI account passed, coveralls reported +0.03 coverage increment ;) ,  all sdk features works fine in my local environment regardless of passing parameter or not. 

At the beginning I want to change constant variables in defines.js, but I in my opinion better solution is to pass authConfig via methods, someone may take advantage of this choice. It was hard, but finally it works :)

I hope, you will accept this pull request, but if not, please suggest me what is need to do, to accept next pull request with changes

Bye,
with Best Regards,
Mateusz Wozniak

### DIFF
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -21,7 +21,7 @@ export type FirebaseOptions = {
   projectId?: string;
   storageBucket?: string;
   messagingSenderId?: string;
-  authUrl?: string;
+  authUrl?: Object;
   [name: string]: any;
 };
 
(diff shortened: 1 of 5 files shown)


bojeil-google:
Thanks for the contribution but we have not agreed to support this feature. Feature requests must go through internal review before we support them.

MatisiekPL:
@bojeil-google Ok. Does my feature request have a chance to be in internal review? Can you say more about these reviews?

Mateusz Wozniak

PR #978, opened 2018-07-03
Add an InsecureGrpcConnection implementation

ryanpbrewster:


packages/firestore/src/platform_node/insecure_grpc_connection.ts
```diff
@@ -0,0 +1,283 @@
...
+      this.cachedStub = {
+        stub: new this.firestore.Firestore(this.databaseInfo.host, credentials),
+        token
+      };
+    }
+
+    return this.cachedStub.stub;
+  }
+
+  private getRpcCallable<Req, Resp>(
```
ryanpbrewster: There is a huge amount of duplicated code between `grpc_connection` and `insecure_grpc_connection`. This is basically the only method that's actually different. Instead of constructing a secure set of channel credentials with the authentication metadata baked in via call credentials, it uses insecure credentials and directly attaches the metadata to each RPC.

ryanpbrewster:
@wilhuff convinced me this was unnecessarily complex.

PR #979, opened 2018-07-03
Embed metadata directly into the RPC call

ryanpbrewster:
This is an alternative approach to https://github.com/firebase/firebase-js-sdk/pull/978 with less duplicated code.


packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -142,7 +103,27 @@ export class GrpcConnection implements Connection {
...
+    // backend.
+    metadata.set(
+      'google-cloud-resource-prefix',
+      `projects/${this.databaseInfo.databaseId.projectId}/` +
+        `databases/${this.databaseInfo.databaseId.database}`
+    );
+    const f = rpc.bind(stub);
+    return function() {
+      return f(arguments, metadata);
+    }
```
ryanpbrewster: The linter is complaining about this. Any ideas? I'm struggling with the fact that `rpc.bind(stub)` is a potentially 0-, 1-, or 2-ary function.

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -142,7 +103,27 @@ export class GrpcConnection implements Connection {
...
+      }
+    }
+    metadata.set('x-goog-api-client', X_GOOG_API_CLIENT_VALUE);
+    // This header is used to improve routing and project isolation by the
+    // backend.
+    metadata.set(
+      'google-cloud-resource-prefix',
+      `projects/${this.databaseInfo.databaseId.projectId}/` +
+        `databases/${this.databaseInfo.databaseId.database}`
+    );
```
mikelehen: FWIW, I think it's probably worth keeping this metadata creation in a helper.  `createMetadata(this.databaseInfo, token)` or whatever.

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -142,7 +103,27 @@ export class GrpcConnection implements Connection {
...
+    const f = rpc.bind(stub);
+    return function() {
+      return f(arguments, metadata);
+    };
```
mikelehen: I'd be a bit surprised if this works.  I think `arguments` will be an array-like object containing the arguments, which probably isn't what the RPC function expects?  I think you can do something like:

```
return function(...args) {
  f(...args, metadata);
}
```

and you could get rid of `f` and just do:

```
return function(...args) {
  rpc.bind(stub, ...args, metadata);
}
```

I _think_ that's right anyway...  I could be slightly off.

packages/firestore/src/platform_node/grpc_connection.ts
ryanpbrewster: Done

packages/firestore/src/platform_node/grpc_connection.ts
ryanpbrewster: Nice, that did the trick! ((It was definitely broken the way I had it.))

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -48,45 +48,32 @@ type UnaryRpc<Req, Resp> = (
...
+  metadata.set(
+    'google-cloud-resource-prefix',
+    `projects/${databaseInfo.databaseId.projectId}/` +
+      `databases/${databaseInfo.databaseId.database}`
   );
-
-  return grpc.credentials.combineChannelCredentials(
-    channelCredentials,
-    callCredentials
-  );
```
ryanpbrewster: Cool upside of this change: `ssl: false` now works with the Node SDK :smile:

ryanpbrewster:
Talked in-person about this and clarified a few things, plus identified a few key concerns:

1. What wire-level differences are there between metadata embedded in the call-credentials vs metadata directly attached to the RPC?
2. Do those wire-level differences manifest in different behavior on the backend (specifically, will the backend have to make more frequent authorization checks if we provide the auth token directly in the RPC metadata instead of embedding it in the call credentials)?

I believe there are no differences in the wire-level behavior. Looking at the gRPC source:

  - For a simple RPC, we start at: `client.ts: makeUnaryCall()` --- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-core/src/client.ts#L156
  - which sets up a stream to handle the RPC `channel.ts: createStream()` --- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-core/src/channel.ts#L307
  - which calls `channel.ts: startHttp2Stream()` --- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-core/src/channel.ts#L256
  - which generates metadata from the call credentials, then merges it with the per-RPC metadata:  `call-credentials-filter.ts: sendMetadata()` --- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-core/src/call-credentials-filter.ts#L28

so regardless of where you store the metadata, it ends up on the wire in the same place. The main differences, as far as I can tell, are:
  1. The call credentials cannot be used with an insecure channel (by design!)
  2. The call credentials expose a "generator" API, so you get token refreshes and such "for free". If we want to attach the metadata directly to each RPC we have to manually call the "generator" function. The `createMetadata()` method introduced in this PR does this.

mikelehen:
This may be worth discussing in person, but I find the gRPC design to be inconvenient and harmful to us.  We never want to send any data whatsoever (credentials or otherwise) to the production backend over an unencrypted channel.  Fortunately the production backend won't accept non-ssl traffic and so we are reasonably safe in that regard. Nobody can use ssl: false in production and have it work.  That said, for testing / development, it is entirely reasonable to use ssl: false against local emulators, etc. whether or not there's some sort of "credential" involved.  Making ssl: false actually work fully is quite useful to us (in addition to Ryan's goal, this will make the node client work against our hexa env).

I think at the point anybody (us or a customer going way off the beaten path) uses ssl: false, all security bets are off.  And this PR seems to be a straightforward way to make ssl: false work (albeit by circumventing gRPC's design goals, apparently).  If we wanted to re-add some minor guardrails we could try to make sure they've also configured to a non-production host or something.

mikelehen:
After talking to Gil, I think I just re-hashed what you guys already discussed and so my latest comments can likely be ignored. Thanks/sorry!

wilhuff:
Just to clarify: @ryanpbrewster and I discussed this yesterday and concluded that my earlier comments were unworkable. I'm on board with the way we're doing this now.

PR #980, opened 2018-07-05
[Multi-Tab] Making Multi-Tab optional

schmidt-sebastian:
This PR adds the yet-to-be approved API to make multi-tab optional.

It also fixes a race during the shutdown: We need to shut down the persistence layer before we shut down the remote store, as the persistence layer could end up calling the remote store even when it is already destroyed.

TODO: Rename `indexeddb_schema.test.ts` to something more applicable once Multi-Tab is merged.

### DIFF
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -57,6 +57,22 @@ export interface Settings {
   timestampsInSnapshots?: boolean;
 }
 
+/** Settings used to configure Firestore persistence. */
+export interface PersistenceSettings {
+  /**
+   * Whether to synchronize the in-memory state of multiple tabs. Setting this
+   * to 'true' in all open tabs enables shared access to local persistence,
+   * shared execution of queries and latency-compensated local document updates
+   * across all connected instances.
+   *
+   * To enable this mode, `synchronizeTabs:true` needs to be set globally in
+   * all active tabs.
+   *
+   * NOTE: This mode is experimental.
+   */
+  synchronizeTabs?: boolean;
+}
+
 export type LogLevel = 'debug' | 'error' | 'silent';
 
 export function setLogLevel(logLevel: LogLevel): void;
@@ -91,10 +107,11 @@ export class FirebaseFirestore {
    *   * unimplemented: The browser is incompatible with the offline
    *     persistence implementation.
    *
+   * @param settings Optional settings object to configure persistence.
    * @return A promise that represents successfully enabling persistent
    * storage.
    */
-  enablePersistence(): Promise<void>;
+  enablePersistence(settings?: PersistenceSettings): Promise<void>;
 
   /**
    * Gets a `CollectionReference` instance that refers to the collection at
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -57,7 +57,10 @@ export interface Settings {
   timestampsInSnapshots?: boolean;
 }
 
-/** Settings used to configure Firestore persistence. */
+/**
+ * Settings that can be passed to Firestore.enablePersistence() to configure
+ * Firestore persistence.
+ */
 export interface PersistenceSettings {
   /**
    * Whether to synchronize the in-memory state of multiple tabs. Setting this
@@ -66,9 +69,10 @@ export interface PersistenceSettings {
    * across all connected instances.
    *
    * To enable this mode, `synchronizeTabs:true` needs to be set globally in
-   * all active tabs.
+   * all active tabs. If omitted or set to 'false', `enablePersistence()` will
+   * will fail in all but the first tab.
    *
-   * NOTE: This mode is experimental.
+   * NOTE: This mode is experimental and not yet recommended for production use.
    */
   synchronizeTabs?: boolean;
 }
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -70,7 +70,7 @@ export interface PersistenceSettings {
    *
    * To enable this mode, `synchronizeTabs:true` needs to be set globally in
    * all active tabs. If omitted or set to 'false', `enablePersistence()` will
-   * will fail in all but the first tab.
+   * fail in all but the first tab.
    *
    * NOTE: This mode is experimental and not yet recommended for production use.
    */
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -98,9 +98,12 @@ import {
 // underscore to discourage their use.
 // tslint:disable:strip-private-property-underscore
 
+// settings() defaults:
 const DEFAULT_HOST = 'firestore.googleapis.com';
 const DEFAULT_SSL = true;
 const DEFAULT_TIMESTAMPS_IN_SNAPSHOTS = false;
+
+// enablePersistence() defaults:
 const DEFAULT_SYNCHRONIZE_TABS = false;
 
 /** Undocumented, private additional settings not exposed in our public API. */
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -101,6 +101,7 @@ import {
 const DEFAULT_HOST = 'firestore.googleapis.com';
 const DEFAULT_SSL = true;
 const DEFAULT_TIMESTAMPS_IN_SNAPSHOTS = false;
+const DEFAULT_SYNCHRONIZE_TABS = false;
 
 /** Undocumented, private additional settings not exposed in our public API. */
 interface PrivateSettings extends firestore.Settings {
@@ -199,6 +200,37 @@ class FirestoreConfig {
 }
 
 /**
+ * Encapsulates the settings that can be used to configure Firestore
+ * persistence.
+ */
+export class PersistenceSettings {
+  /** Whether to enable multi-tab synchronization. */
+  synchronizeTabs: boolean;
+
+  constructor(
+    readonly enabled: boolean,
+    settings?: firestore.PersistenceSettings
+  ) {
+    assert(
+      enabled || !settings,
+      'Can only provide PersistenceSettings with persistence enabled'
+    );
+    settings = settings || {};
+    this.synchronizeTabs = objUtils.defaulted(
+      settings.synchronizeTabs,
+      DEFAULT_SYNCHRONIZE_TABS
+    );
+  }
+
+  isEqual(other: PersistenceSettings): boolean {
+    return (
+      this.enabled === other.enabled &&
+      this.synchronizeTabs === other.synchronizeTabs
+    );
+  }
+}
+
+/**
  * The root reference to the database.
  */
 export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
@@ -291,7 +323,7 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return this._firestoreClient.disableNetwork();
   }
 
-  enablePersistence(): Promise<void> {
+  enablePersistence(settings?: firestore.PersistenceSettings): Promise<void> {
     if (this._firestoreClient) {
       throw new FirestoreError(
         Code.FAILED_PRECONDITION,
@@ -301,17 +333,21 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
       );
     }
 
-    return this.configureClient(/* persistence= */ true);
+    return this.configureClient(
+      new PersistenceSettings(/* enabled= */ true, settings)
+    );
   }
 
   ensureClientConfigured(): FirestoreClient {
     if (!this._firestoreClient) {
-      this.configureClient(/* persistence= */ false);
+      this.configureClient(new PersistenceSettings(/* enabled= */ false));
     }
     return this._firestoreClient as FirestoreClient;
   }
 
-  private configureClient(persistence: boolean): Promise<void> {
+  private configureClient(
+    persistenceSettings: PersistenceSettings
+  ): Promise<void> {
     assert(
       !!this._config.settings.host,
       'FirestoreSettings.host cannot be falsey'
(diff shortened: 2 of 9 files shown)


packages/firestore-types/index.d.ts
```diff
@@ -57,6 +57,22 @@ export interface Settings {
   timestampsInSnapshots?: boolean;
 }
 
+/** Settings used to configure Firestore persistence. */
```
mikelehen: To make life easier for people that are looking at docs / not using TypeScript, I'd say "Settings that can be passed to Firestore.enablePersistence() to configure ..." or something

packages/firestore-types/index.d.ts
```diff
@@ -57,6 +57,22 @@ export interface Settings {
...
+export interface PersistenceSettings {
+  /**
+   * Whether to synchronize the in-memory state of multiple tabs. Setting this
+   * to 'true' in all open tabs enables shared access to local persistence,
+   * shared execution of queries and latency-compensated local document updates
+   * across all connected instances.
+   *
+   * To enable this mode, `synchronizeTabs:true` needs to be set globally in
+   * all active tabs.
```
mikelehen: Seems like we should explain that /not/ setting this to true means enablePersistence() will fail in all but the first tab.

packages/firestore-types/index.d.ts
```diff
@@ -57,6 +57,22 @@ export interface Settings {
...
+   *
+   * NOTE: This mode is experimental.
```
mikelehen: FWIW- This isn't super actionable.  Consider adding "and not recommended for production use" or "and may cause bugs" or "and may corrupt your offline cache" or "Please log bugs on [github issues link] if you see any issues" or something (depending on how cautious we want people to be and what we want them to do).

packages/firestore/src/api/database.ts
```diff
@@ -101,6 +101,7 @@ import {
 const DEFAULT_HOST = 'firestore.googleapis.com';
 const DEFAULT_SSL = true;
 const DEFAULT_TIMESTAMPS_IN_SNAPSHOTS = false;
+const DEFAULT_SYNCHRONIZE_TABS = false;
```
mikelehen: Might be worth separating the .settings() default settings from the .enablePersistence() default settings...  Either via naming (DEFAULT_PERSISTENCE_SYNCHRONIZE_TABS) or just whitespace / comments:

```
// settings() defaults:
...

// enablePersistence() defaults:
...
```

packages/firestore/src/core/firestore_client.ts
```diff
@@ -135,13 +137,14 @@ export class FirestoreClient {
    * fallback succeeds we signal success to the async queue even though the
    * start() itself signals failure.
    *
-   * @param usePersistence Whether or not to attempt to enable persistence.
+   * @param persistenceSettings Settings object for the suggested persistence
+   *     mode.
```
mikelehen: "Settings object for the suggested persistence mode" sounds weird to me.  Maybe "Whether persistence is enabled and any associated settings" or use the same description as you used below ("Settings object to configure offline persistence")

packages/firestore/test/unit/local/indexeddb_schema.test.ts
```diff
@@ -29,8 +29,15 @@ import {
 } from '../../../src/local/indexeddb_schema';
 import { SimpleDb, SimpleDbTransaction } from '../../../src/local/simple_db';
 import { PersistencePromise } from '../../../src/local/persistence_promise';
+import { ClientId } from '../../../src/local/shared_client_state';
```
mikelehen: You have a TODO in your PR description about renaming this file, but that isn't greppable.  Can you add a TODO(multitab) in this file if you really intend to rename this?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -291,32 +302,37 @@ export class FirestoreClient {
     });
 
     return Promise.resolve().then(() => {
-      this.persistence = new IndexedDbPersistence(
+      const persistence = new IndexedDbPersistence(
```
mikelehen: nit: Not sure why you introduced `persistence` separate from `this.persistence`

packages/firestore/src/core/firestore_client.ts
```diff
@@ -291,32 +302,37 @@ export class FirestoreClient {
...
             'IndexedDB persistence is only available on platforms that support LocalStorage.'
           );
         }
-        debug(LOG_TAG, 'Starting Persistence in test-only non multi-tab mode');
-        this.sharedClientState = new MemorySharedClientState();
       }
-      return this.persistence.start();
+
+      this.sharedClientState = settings.synchronizeTabs
+        ? new WebStorageSharedClientState(
```
mikelehen: I'm confused about what's going on with USE_MOCK_PERSISTENCE.  It looks like this code used to fall back to MemorySharedClientState for mock persistence, but now it will always use WebStorageSharedClientState...  but the Code.UNIMPLEMENTED error condition above implies that WebStorageSharedClientState.isAvailable() will be false for node / USE_MOCK_PERSISTENCE.  So something seems out-of-sync.

I'm wondering if we don't need the memory fallback for mock persistence anymore since tests can explicitly opt out of synchronizeTabs, in which case we probably don't need the USE_MOCK_PERSISTENCE check above anymore?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,7 +220,11 @@ export class IndexedDbPersistence implements Persistence {
         .next(canActAsPrimary => {
           if (canActAsPrimary !== this.isPrimary) {
             this.isPrimary = canActAsPrimary;
-            this.queue.enqueue(() => this.primaryStateListener(this.isPrimary));
+            this.queue.enqueue(async () => {
+              if (this.started) {
```
mikelehen: It's non-obvious why this check is necessary. Maybe it's guarding against early shutdown? Can you add a comment?

// Ensure persistence hasn't been shut down before notifying the listener.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -263,7 +282,18 @@ export class IndexedDbPersistence implements Persistence {
           currentPrimary.ownerId !== this.getZombiedClientId();
 
         if (currentLeaseIsValid) {
-          return this.isLocalClient(currentPrimary);
+          if (this.isLocalClient(currentPrimary)) {
+            return true;
+          }
+
+          if (!currentPrimary.allowTabSynchronization) {
```
mikelehen: FWIW this seems like a non-obvious (and not 100% safe) place for this check.  It'd make more sense to me that we do this as part of runTransaction() so that every time we perform a transaction we make sure there's no external primary client that has allowTabSynchronization turned off.

As-is, a non-primary client could do up to 4 seconds of "corrupting" persistence before it notices that another tab is running without tab synchronization.

So it seems like we should either fix this or else consider it an intentional design choice and add some comments to explain.

packages/firestore/test/unit/local/indexeddb_schema.test.ts
```diff
@@ -210,3 +237,39 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+  }
+
+  beforeEach(() => SimpleDb.delete(INDEXEDDB_TEST_DATABASE));
+
+  after(() => SimpleDb.delete(INDEXEDDB_TEST_DATABASE));
+
+  it('rejects access when synchronization is disabled', () => {
+    return withPersistence('clientA', async db1 => {
+      await expect(db1.start(/*synchronizeTabs=*/ false)).to.eventually.be
+        .fulfilled;
```
mikelehen: Out of curiosity, is `expect(...).to.eventually.be.fulfilled` adding anything here?  Couldn't you just `await db1.start(...)` ?  If it fails, it'll throw an exception, which will fail the test.

I don't care strongly if this is intentional though...

packages/firestore/src/core/firestore_client.ts
schmidt-sebastian: this.persistence is of type Persistence, which doesn't expose `synchronizeTabs`. I made this more obvious by adding the type to `const persistence`.

packages/firestore-types/index.d.ts
schmidt-sebastian: Done

packages/firestore-types/index.d.ts
schmidt-sebastian: Added this to the comment.

packages/firestore-types/index.d.ts
schmidt-sebastian: I made it `not yet recommended for production use`.

packages/firestore/src/api/database.ts
schmidt-sebastian: That makes sense. I added the comment strings.

packages/firestore/src/core/firestore_client.ts
schmidt-sebastian: I was trying to indicate that this is more of Christmas-type situation and that not every wish is going to be fulfilled. But it does sound a little weird. Updated to use "Settings object to configure offline persistence".

packages/firestore/test/unit/local/indexeddb_schema.test.ts
schmidt-sebastian: Added

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: Comment added. This frequently happened during test failures, where the client is being shutdown, while another (unexpected) operation is still doing some leftover work.

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: We need to do this both here and in `runTransaction`. I added a comment that should make the reasoning for this a little more obvious. In essence, we need to fail the `enablePersistence()` promise, and running this check in `canActAsPrimary` does this at startup while also blocking the AsyncQueue if this check were to ever fail at a later point.

As you pointed out, we do need to run this check for every transaction . I added this to `runTransaction` though as I do need it to ensure correctness.

I could also call the new method `verifyAllowTabSynchronization` at startup, but that is an additional read for every client, and will in practice not make a difference.

packages/firestore/test/unit/local/indexeddb_schema.test.ts
schmidt-sebastian: Hm, I just used it because I wanted to use  the `rejectedWith` counterpart. It doesn't add any value though, beside the increase in complexity. Removed.

packages/firestore-types/index.d.ts
```diff
@@ -66,9 +69,10 @@ export interface PersistenceSettings {
    * across all connected instances.
    *
    * To enable this mode, `synchronizeTabs:true` needs to be set globally in
-   * all active tabs.
+   * all active tabs. If omitted or set to 'false', `enablePersistence()` will
+   * will fail in all but the first tab.
```
mikelehen: Too much will.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -287,6 +289,18 @@ export class IndexedDbPersistence implements Persistence {
           }
 
           if (!currentPrimary.allowTabSynchronization) {
+            // Fail the `canActAsPrimary` check if the current leaseholder has
+            // not opted into multi-tab synchronization. If this happens at
+            // client startup, we reject the Promise returned by
+            // `enablePersistence()` and the user can continue to use Firestore
+            // with in-memory persistence.
+            // If this fails during a lease refresh, we will instead block the
+            // AsyncQueue from execution further operations. Note that this is
```
mikelehen: execution => executing

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -287,6 +289,18 @@ export class IndexedDbPersistence implements Persistence {
...
+            // acceptable since mixing & matching different `synchronizeTabs`
+            // settings is not officially supported.
```
mikelehen: "not officially supported" makes it sound like there's some amount of unofficial support, but in reality it would be completely broken, right?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -421,7 +435,36 @@ export class IndexedDbPersistence implements Persistence {
...
+  ): PersistencePromise<void> {
+    const store = ownerStore(txn);
+    return store.get('owner').next(currentPrimary => {
+      const currentLeaseIsValid =
+        currentPrimary !== null &&
+        this.isWithinMaxAge(currentPrimary.leaseTimestampMs) &&
+        currentPrimary.ownerId !== this.getZombiedClientId();
+
+      if (currentLeaseIsValid && !this.isLocalClient(currentPrimary)) {
+        if (!currentPrimary.allowTabSynchronization) {
```
mikelehen: Intentionally using nested-if over `&&` ?

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: I removed `officially`. It's probably going to be very very broken.

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: Fixed

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: To me, this is easier to follow since it extracts the more important part into a separate statement.

packages/firestore-types/index.d.ts
schmidt-sebastian: Fixed.

PR #984, opened 2018-07-06
[Multi-Tab] Recover from unexpected primary lease loss

schmidt-sebastian:
This PR examines all the places where we need a Primary lease in IndexedDb and adds recovery logic. The recovery logic is pretty simple - it just flips the client state to secondary.

I tried to plumb through some checks on whether the client is actually no longer primary, but decided this is not worth it:
- Neither Sync Engine nor Remote Store have access to the persistence layer. I would have needed to plumb the primary state through to LocalStore.
- If the client should be primary, IndexedDb will reset the state in the next lease refresh.


packages/firestore/src/remote/remote_store.ts
```diff
@@ -558,12 +558,23 @@ export class RemoteStore implements TargetMetadataProvider {
...
-          this.writeStream.writeMutations(batch.mutations);
+      .then(
+        () => {
+          // Send the write pipeline now that the stream is established.
+          for (const batch of this.writePipeline) {
+            this.writeStream.writeMutations(batch.mutations);
+          }
+        },
+        err => {
+          if (err.code === Code.FAILED_PRECONDITION) {
```
schmidt-sebastian: FYI: Instead of relying on the error code, I tried to make this more specific and added a new error class at one point. I wanted to do an `instanceof CustomPrimaryLeaseLostError` check here. It turns out that instanceof checks don't work with errors though, and so I abandoned this approach.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -785,4 +785,62 @@ describeSpec('Listens:', [], () => {
...
+        .expectEvents(query, { added: [docA] })
+        .stealPrimaryLease()
+        .expectListen(query, 'resume-token-1000')
+        .watchAcksFull(query, 2000, docB)
+        .expectEvents(query, { added: [docB] })
+        .client(0)
+        .expectPrimaryState(false)
+        // TODO(multitab): Suppres the additional `fromCache`, which is raised
+        // locally because we disable the network when we lose the primary
+        // lease.
```
mikelehen: FWIW I'm not sure if this is wrong / harmful.  When a primary switch happens we collectively will be without a network connection for a period of time and it seems correct for views to trigger fromCache=true events...  but we _should_ return to fromCache=false once the primary switch completes and a new primary establishes the listens...

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -785,4 +785,62 @@ describeSpec('Listens:', [], () => {
...
+          .client(1)
+          .userListens(query)
+          .expectEvents(query, { fromCache: true })
+          .client(0)
+          .expectListen(query)
+          .client(1)
+          .stealPrimaryLease()
+          .client(0)
+          // Send a watch update to client 0, who is longer primary (but doesn't
+          // it yet). The watch update gets ignored.
```
mikelehen: missing "know"

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -457,12 +461,16 @@ export class SpecBuilder {
...
     version: TestSnapshotVersion,
-    options?: { expectUserCallback: boolean }
+    options?: { expectUserCallback?: boolean; keepInQueue?: boolean }
   ): this {
     this.nextStep();
-    this.currentStep = { writeAck: { version } };
+    options = options || {};
+
+    this.currentStep = {
+      writeAck: { version, keepInQueue: !!options.keepInQueue }
```
mikelehen: FWIW it may be worth adding a TODO(multitab) to update the spec schemas on the other clients with the changes you've made...  Just so we don't forget to do that before / soon-after merging multitab into master.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -883,10 +884,16 @@ abstract class TestRunner {
...
   }
 
-  private doApplyClientState(state: SpecClientState): Promise<void> {
+  private async doApplyClientState(state: SpecClientState): Promise<void> {
     if (state.visibility) {
       this.platform.raiseVisibilityEvent(state.visibility!);
     }
+
+    if (state.primary) {
+      await writeOwnerToIndexedDb(this.clientId);
```
mikelehen: nit: Would it be more realistic to just clear (or expire?) the current owner and then let ClientMetadataRefresh detect that it needs to steal the lease?  I don't care strongly though.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1004,8 +1011,10 @@ abstract class TestRunner {
     // without any direct user interaction.
     // TODO(multitab): Refactor so this is only executed after primary tab
     // change
-    if (!obj.isEmpty(this.expectedActiveTargets)) {
-      await this.connection.waitForWatchOpen();
+    if (this.expectedActiveTargets) {
+      if (!obj.isEmpty(this.expectedActiveTargets)) {
+        await this.connection.waitForWatchOpen();
+      }
       await this.queue.drain();
```
mikelehen: Er, why do we need to call/wait-for queue.drain() even when we don't call waitForWatchOpen()?  validateActiveTargets() is called by validateStateExpectations() which is called by run(step) which already calls queue.drain()...

packages/firestore/src/core/sync_engine.ts
```diff
@@ -490,6 +493,11 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       // connection is disabled.
       await this.remoteStore.fillWritePipeline();
     } else if (batchState === 'acknowledged' || batchState === 'rejected') {
+      // If we receive a notification of an `acknowledged` or `rejected` batch
+      // via Web Storage, we are either already secondary or another tab has
+      // taken the primary lease.
```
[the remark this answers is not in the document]
schmidt-sebastian: I do think this is racey, but it is racey both ways: If a client comes back from the dead, and thinks it is primary, it might take a while for it to catch up on Local Storage notifications before it realizes it is no longer primary. 

If I remove this line, I get additional query events about half the time and the tests become flakey. It loos like this is the only side effect, and if two primary tabs were to run this code in this function at the same time, the world would still be a happy place.

My argument for keeping it is another though: If a tab becomes primary even though it has not caught up on LocalStorage notifications, it seems much safer to go back into secondary mode and finish these notifications in the state it was supposed to be in until the client has caught up.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -801,6 +834,12 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     state: QueryTargetState,
     error?: FirestoreError
   ): Promise<void> {
+    if (this.isPrimary) {
+      // If we receive a target state notification via Web Storage, we are
+      // either already secondary or another tab has taken the primary lease.
+      await this.applyPrimaryState(false);
```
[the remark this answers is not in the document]
schmidt-sebastian: Removing these lines breaks some of the new tests. This is certainly a test issue, but in general multi-tab is only safe if there is only one primary (we don't want two simultaneous WriteStreams for example). I prefer having no primary temporarily than two primaries at once.

mikelehen:
Just to double check...  should this be assigned back to me? Or are you still working on review feedback?

schmidt-sebastian:
> Just to double check... should this be assigned back to me? Or are you still working on review feedback?

Back to you :)

schmidt-sebastian:
> Just to double check... should this be assigned back to me? Or are you still working on review feedback?

Back to you :)

packages/firestore/src/core/sync_engine.ts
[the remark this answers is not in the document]
mikelehen: I'm not sure I follow / agree-with your argument.  I don't think we can deterministically "finish these notifications in the state it was supposed to be in".

At the end of the day it's possible for two tabs to send the same write to the backend and broadcast LocalStorage notifications for it.  E.g. Tab A is primary, performs a write, gets the response, and gets paused / throttled right before sending the LocalStorage notification.  Tab B takes over as primary, re-sends the write, and processes the response (sending a LocalStorage notification). Tab A resumes and its LocalStorage notification (for the same write) now goes out.

I think this means that handling write acknowledgement notifications must be idempotent.

But I also don't understand where the extra query events you mentioned are coming from...  It may be worth talking through this or something.

packages/firestore/src/core/sync_engine.ts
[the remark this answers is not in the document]
mikelehen: I'd prefer that too, but I don't think we can guarantee it.  The only thing we can guarantee is that only one tab can make IndexedDb writes as primary at a time (via transactions that require the primary lease).  But we _could_ have two tabs with WriteStreams open for some period of time.

Ideally our design would allow for this.  If it's infeasible, then I guess we should acknowledge that there are race conditions but convince ourselves that it'll be sufficiently unlikely and/or non-catastrophic or something...

packages/firestore/src/core/sync_engine.ts
```diff
@@ -505,10 +504,16 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-      // via Web Storage, we are either already secondary or another tab has
-      // taken the primary lease.
-      await this.applyPrimaryState(false);
+      if (this.isPrimary) {
+        // If we receive a notification of an `acknowledged` or `rejected` batch
+        // via Web Storage, we are either already secondary or another tab has
+        // taken the primary lease.
+        log.debug(
+          LOG_TAG,
+          'Unexpectedly received mutation batch notification when already primary. Releasing primary lease.'
```
mikelehen: If we keep this logic, we should rephrase this comment since we're not _actually_ releasing the primary lease.  We're just pretending we don't have it. :-)

packages/firestore/src/core/sync_engine.ts
```diff
@@ -712,7 +724,27 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+   * primary lease has been taken by another client. This can happen when the
+   * client is temporarily CPU throttled and fails to renew its lease in time,
+   * in which we treat the current client as secondary. We can always revert
+   * back to primary status via the lease refresh in our persistence layer.
+   *
+   * @param err An error returned by an IndexedDb operation.
+   * @return A Promise that resolves after we recovered, or the original error.
+   */
+  private async tryRecoverClient(err: FirestoreError): Promise<void> {
+    if (err.code === Code.FAILED_PRECONDITION) {
```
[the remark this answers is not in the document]
mikelehen: We already re-use FAILED_PRECONDITION for the PRIMARY_LEASE_EXCLUSIVE_ERROR_MSG error, and I don't think it makes sense to fall back to secondary in that case.

I'd rather pessimistically assume that this error handling should _not_ apply to future unknown errors and only expand it if/when we have specific cases we want to let it handle.

If we do keep this error handling fairly inclusive, we should definitely include the error in the log message.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -824,41 +832,80 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+      this.remoteStore.unlisten(targetId);
+    });
+    this.limboKeysByTarget = [];
+    this.limboTargetsByKey = new SortedMap<DocumentKey, TargetId>(
+      DocumentKey.comparator
+    );
+  }
+
+  /**
+   * Reconcile the local query views with those the state from persistence.
```
mikelehen: those the?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -824,41 +832,80 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+        const query = await this.localStore.getQueryForTarget(targetId);
+        if (this.queryViewsByTarget[targetId] === undefined) {
+          assert(
+            this.isPrimary,
+            'A secondary tab should never have an active query without an active view.'
+          );
+          // For queries that never executed on this client, we need to
+          // allocate the query in LocalStore and initialize a new View.
+          queryData = await this.localStore.allocateQuery(query);
+          await this.initializeViewAndComputeSnapshot(queryData, false);
```
mikelehen: /*current=*/

packages/firestore/src/core/sync_engine.ts
```diff
@@ -824,41 +832,80 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+          // might have changed) and reconcile their views with the persisted
+          // state (the list of syncedDocuments may have gotten out of sync).
+          await this.localStore.releaseQuery(query, true);
+          queryData = await this.localStore.allocateQuery(query);
+          const viewChange = await this.synchronizeViewAndComputeSnapshot(
+            query,
+            queryData.targetId
+          );
+          if (viewChange.snapshot) {
+            this.syncEngineListener!.onWatchChange([viewChange.snapshot]);
```
mikelehen: Should we batch up all the snapshots and call onWatchChange() only once?

In practice I'm not sure how much it matters right now, but since all of the snapshots were the result of a single event, I think we should send them to onWatchChange() as a single batch.  In the future we might expose an API like:

```
firebase.firestore().onGlobalSnapshot(() => { /* I know that all listeners are in a consistent state. */ });
```

packages/firestore/src/core/sync_engine.ts
```diff
@@ -824,41 +832,80 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+    });
+    this.limboKeysByTarget = [];
+    this.limboTargetsByKey = new SortedMap<DocumentKey, TargetId>(
+      DocumentKey.comparator
+    );
+  }
+
+  /**
+   * Reconcile the local query views with those the state from persistence.
+   * Raises snapshots for any changes that affect the current client.
```
mikelehen: and "Returns the set of queries for which we have local views." or something?

packages/firestore/src/util/obj.ts
```diff
@@ -39,6 +39,15 @@ export function size<V>(obj: Dict<V>): number {
   return count;
 }
 
+/** Extracts the numeric indices from a dictionary. */
+export function indices<V>(obj: Dict<V>): number[] {
```
mikelehen: Would it make sense to use `[numberKey: number]: V` instead of `Dict<V>`?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -265,18 +271,33 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+      return this.localStore
+        .remoteDocumentKeys(targetId)
+        .then(async remoteKeys => {
+          const queryView = this.queryViewsByTarget[targetId];
+          assert(!!queryView, 'Cannot reconcile missing view');
+          const viewChange = queryView.view.synchronizeWithPersistedState(
+            docs,
+            remoteKeys,
+            /* resetLimboDocuments= */ this.isPrimary,
+            /* resetCurrent= */ this.isPrimary
```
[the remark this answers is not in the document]
schmidt-sebastian: I simplified this a little to unconditionally re-compute the list of limbo documents. In clients that are in-sync this is a no-op, and in clients that are out of sync, this will bring them back into sync. The only bit that I still need to provide is whether to set `current` to false. I added some documentation to clear this up a little.

Notice that I also added a TODO to mark the queries as non-current in SharedClientState. That's probably a three line change but I wanted to keep this PR from exploding even further.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -787,10 +787,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     err: FirestoreError
   ): Promise<void> {
     if (isPrimaryLeaseLostError(err)) {
-      log.debug(
-        LOG_TAG,
-        'Unexpectedly lost primary lease, attempting to recover'
-      );
+      log.debug(LOG_TAG, 'Unexpectedly lost primary lease');
```
mikelehen: You still need to update the method name and comment...  Please mention that when this error is thrown, there will automatically be a lease refresh scheduled so we should very quickly get an applyPrimaryState() call from the persistence layer.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -585,10 +585,7 @@ export class RemoteStore implements TargetMetadataProvider {
 
   private tryRecoverFromPrimaryLeaseLoss(err: FirestoreError): void {
```
mikelehen: Please rename this if you rename the other one.

schmidt-sebastian: Done

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Update the method and comment. Method is now `ignoreIfPrimaryLeaseLoss`.

PR #985, opened 2018-07-06
[Multi-Tab] Spec tests for clients that gain primary lease at startup

schmidt-sebastian:
These tests verify already implemented behavior at client startup and make sure that when we apply the primary lease, we pick up any remaining work.

### DIFF
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -725,7 +725,7 @@ describeSpec('Listens:', [], () => {
   );
 
   specTest(
-    'Query is re-listened to after primary tab failover',
+    'Listen is re-listened to after primary tab failover',
     ['multi-client'],
     () => {
       const query = Query.atPath(path('collection'));
@@ -758,7 +758,7 @@ describeSpec('Listens:', [], () => {
     }
   );
 
-  specTest('Query is established in new primary tab', ['multi-client'], () => {
+  specTest('Listen is established in new primary tab', ['multi-client'], () => {
     const query = Query.atPath(path('collection'));
     const docA = doc('collection/a', 1000, { key: 'a' });
     const docB = doc('collection/b', 2000, { key: 'b' });
@@ -787,7 +787,7 @@ describeSpec('Listens:', [], () => {
   });
 
   specTest(
-    'Query is established in newly started client',
+    'Listen is established in newly started primary',
     ['multi-client'],
     () => {
       const query = Query.atPath(path('collection'));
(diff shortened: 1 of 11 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -785,4 +785,36 @@ describeSpec('Listens:', [], () => {
...
+  specTest(
+    'Query is established in newly started client',
+    ['multi-client'],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const docA = doc('collection/a', 1000, { key: 'a' });
+      const docB = doc('collection/b', 2000, { key: 'b' });
+
+      // Client 0 executes a query on behalf of Client 1. When client 0 shuts
+      // down, client 1 starts up and becomes primary, taking ownership of the
```
mikelehen: I don't think this description is right.  You have a "client(2)" that starts up and becomes primary...

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -1127,4 +1127,21 @@ describeSpec('Writes:', [], () => {
...
+    return client(0)
+      .expectPrimaryState(true)
+      .client(1)
+      .expectPrimaryState(false)
+      .userSets('collection/a', { v: 1 })
+      .client(0)
+      .shutdown()
+      .client(2)
+      .expectPrimaryState(true)
+      .writeAcks('collection/a', 1000, { expectUserCallback: false })
```
mikelehen: Could we add a `.expectNumOutstandingWrites(1)` here?  It's probably not necessary (`writeAcks()` would hopefully fail if there's no outstanding write), but it'd make the test clearer, I think.

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -1127,4 +1127,21 @@ describeSpec('Writes:', [], () => {
         hasPendingWrites: true
       });
   });
+
+  specTest('Write is sent by newly started client', ['multi-client'], () => {
```
mikelehen: FWIW, I'm starting to wonder if it'd make sense to put all the multi-tab tests together (or maybe the primary transition ones at least) and have more contextual descriptions.  "Multi-tab: Primary-Transition: Write is sent by newly started client" or something...

I don't feel super strongly at the moment though.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -785,4 +785,36 @@ describeSpec('Listens:', [], () => {
       .client(2)
       .expectEvents(query, { added: [docB] });
   });
+
+  specTest(
+    'Query is established in newly started client',
```
mikelehen: "Query"=>"Listen" perhaps (here and elsewhere).

PR #988, opened 2018-07-06
Move fieldFilter (free function) to Filter.create()

rsgowman:
This is a refactoring to unify filter creation across platforms.


packages/firestore/src/core/query.ts
```diff
@@ -434,12 +463,12 @@ export class RelationOp {
   }
 }
 
-export class RelationFilter implements Filter {
+export class RelationFilter extends Filter {
   constructor(
```
schmidt-sebastian: You can omit constructors that match the super class constructor (here and everywhere)..

packages/firestore/src/core/query.ts
```diff
@@ -390,10 +390,39 @@ export class Query {
...
+  static create(
+    field: FieldPath,
+    op: RelationOp,
+    value: FieldValue
+  ): Filter {
+    if (value.isEqual(NullValue.INSTANCE)) {
+      if (op !== RelationOp.EQUAL) {
+        throw new FirestoreError(
+          Code.INVALID_ARGUMENT,
+          'Invalid query. You can only perform equals ' + 'comparisons on null.'
```
schmidt-sebastian: Do you mind removing the extraneous string concatenation?

rsgowman: Done.

packages/firestore/src/core/query.ts
rsgowman: Don't you need to keep them to ensure (in this case) field, op, and value are present? Or in other words, these ctors don't match the parent ctor.

schmidt-sebastian: Oh, yes, that's true. The constructors are much more powerful at second glance.

PR #990, opened 2018-07-09
[Multi-Tab]  Supporting non-initial initial queries

schmidt-sebastian:
This PR adds support for queries that SyncEngine knows about but EventManager doesn't.

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -212,6 +212,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     this.viewHandler!([viewSnapshot]);
     return targetId;
   }
+
   private initializeViewAndComputeInitialSnapshot(
     queryData: QueryData,
     current: boolean
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -177,30 +177,41 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
    * server. All the subsequent view snapshots or errors are sent to the
    * subscribed handlers. Returns the targetId of the query.
    */
-  listen(query: Query): Promise<TargetId> {
+  async listen(query: Query): Promise<TargetId> {
     this.assertSubscribed('listen()');
-    assert(
-      !this.queryViewsByQuery.has(query),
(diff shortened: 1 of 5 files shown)


packages/firestore/src/core/view_snapshot.ts
```diff
@@ -149,6 +149,30 @@ export class ViewSnapshot {
     readonly excludesMetadataChanges: boolean
   ) {}
 
+  /** Returns a view snapshot as if all documents in the snapshot were added. */
+  static fromSyncedDocuments(
```
mikelehen: I'm not really sure what "synced" means in this context.  Is this related to the existing `syncedDocuments` in the View?  Else maybe `fromDocuments()` or `fromInitialDocuments()` ?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -177,28 +177,40 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-        if (this.isPrimary) {
-          this.remoteStore.listen(queryData);
-        }
-        this.viewHandler!([viewSnapshot]);
-        return queryData.targetId;
-      });
-    });
+      );
+      if (this.isPrimary) {
+        this.remoteStore.listen(queryData);
```
mikelehen: Slightly tangential but since the primary ends up with listens that secondaries don't have I was trying to figure out how we reconcile that during primary transfers.  There's a "TODO: Unlisten from all active targets" comment in this file. Can you change that to TODO(multitab)?  I must have missed it in a previous review.

packages/firestore/src/core/view_snapshot.ts
schmidt-sebastian: I was debating between `fromSyncedDocuments` and `fromExistingDocuments`. `synced` is slightly wrong since this set of documents should also contain local edits, but I still liked it better than simply using `existing`. `initial` is much more clear though and doesn't suggest that these documents need to be in sync with the backend. I renamed it.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I'm gonna leave this as is for now to avoid merge conflicts with https://github.com/firebase/firebase-js-sdk/pull/994 and https://github.com/firebase/firebase-js-sdk/pull/984/files, both of which address this TODO.

PR #993, opened 2018-07-09
[Multi-Tab] Making Platform, Window and Document Mock available in all tests

schmidt-sebastian:
In order to raise visibility events in non-spec unit tests, this PR moves the Platform mocks out of the spec tests file and makes them public in `mock_platform.ts`. There is also a name change from `TestPlatform` to `MockPlatform`.

This is the base PR for what I assume will be #994.


schmidt-sebastian:
Pushed a commit that changes the names to FakeDocument, FakeWindow, SharedFakeWebStorage and TestPlatform.

PR #995, opened 2018-07-09
[Multi-Tab] Remove `withGcEnabled` for tests that don't call `unlisten`

schmidt-sebastian:
See amazing title.


schmidt-sebastian:
> Just to confirm, this doesn't overlap with any of the work Greg was doing on GC right? He was having a hard time with the tests assuming documents persisted across listens.

This is all just for new tests that use multi-tab. Greg hasn't seen these yet.

PR #1000, opened 2018-07-11
Multi-Tab Merge

schmidt-sebastian:
This PR merges the Multi-Tab changes into Master. The Multi-Tab feature itself is not turned on, and cannot be turned on by end users yet (the flag to do so is commented out).


schmidt-sebastian:
I am planning to squash this into Master (as merges seemed to be turned off these days). If you think we should merge this instead, then I will go ahead and clean up some of the earlier commit history in this PR that contains individual commits.

mikelehen:
Hrm...  This is going to be difficult to review in a meaningful way. :-)  Any chance you can help by trying to provide some sort of risk assessment?  Given the flag is disabled, what code paths are most affected and should be re-reviewed for sanity?  What perf impacts have you measured or expect there to be?  Or are you confident this will have very little impact?

schmidt-sebastian:
> Hrm... This is going to be difficult to review in a meaningful way. :-) Any chance you can help by trying to provide some sort of risk assessment? Given the flag is disabled, what code paths are most affected and should be re-reviewed for sanity? What perf impacts have you measured or expect there to be? Or are you confident this will have very little impact?

I looked through the changes in here to spot bad merges (without success so I consider this a success). 

The performance changes (and the code that can't turned off) all live in the persistence layer. 

- Mutation require an extra read on insert.
- Query targets require an extra  read on insert.
- The keys of changes documents are written to the `documentChangesStore` as we receive remote events.

I have tested the first two scenarios and will get back to you when I tested the third one.

Reading of mutations and raising of query snapshot should be unaffected and I was not able to spot a performance difference.

schmidt-sebastian:
I have added some not-so-great results to https://github.com/firebase/firebase-js-sdk/pull/1048

schmidt-sebastian:
Merged master :) Please note the changes in https://github.com/firebase/firebase-js-sdk/pull/1000/commits/76e0e1e54b459ce5943a0ecc6e459daea13f8a24#diff-cf8f425b1bd34112162e05d44fc3a464

PR #1001, opened 2018-07-11
Added missing types for optional database instance.

carlosacabrera:
Creating a reference to another DB instance (Database Sharding https://firebase.google.com/docs/database/usage/sharding) throws typing error: `[ts] Expected 0 arguments, but got 1.`

i.e.

```
app.database(`https://xxx.firebaseio.com/`);
```

I was able to find testing for these cases https://github.com/firebase/firebase-js-sdk/blob/81cd260aeb729c5003952988b19d5b32788510d4/packages/database/test/database.test.ts#L73 but the optional type is missing from https://github.com/firebase/firebase-js-sdk/blob/81cd260aeb729c5003952988b19d5b32788510d4/packages/firebase/index.d.ts#L128

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -125,7 +125,7 @@ declare namespace firebase {
 declare namespace firebase.app {
   interface App {
     auth(): firebase.auth.Auth;
-    database(): firebase.database.Database;
+    database(instance?: string): firebase.database.Database;
     delete(): Promise<any>;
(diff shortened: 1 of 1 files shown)


carlosacabrera:
I signed it!


Carlos A. Cabrera
Experience Technology Architect

S A P I E N T
R A Z O R F I S H _

300 W. 6th St. Suite 1600
Austin, Texas 78701, USA
M +1 305 501 0545

PR #1007, opened 2018-07-13
Enable firestore sdk to talk to emulator

tonymeng:
This change enables the testing sdk to talk to the firestore emulator. This is based off of https://docs.google.com/document/d/1dWAVm8-jK_JLsCOx4uZBj2MN_6auc7ood-0llWoojs8/edit?ts=5b47bec0#heading=h.sm63cr1eg2l2

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -148,8 +148,6 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
 
   private forceRefresh = false;
 
-  private tokenOverride: string = null;
-
   constructor(private readonly app: FirebaseApp) {
     // We listen for token changes but all we really care about is knowing when
     // the uid may have changed.
@@ -165,7 +163,6 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     };
 
     this.userCounter = 0;
-    this.tokenOverride = app.options.tokenOverride;
 
     // Will fire at least once where we set this.currentUser
     (this.app as _FirebaseApp).INTERNAL.addAuthTokenListener(
@@ -185,13 +182,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     const initialUserCounter = this.userCounter;
     const forceRefresh = this.forceRefresh;
     this.forceRefresh = false;
-    var token: Promise<FirebaseAuthTokenData>;
-    if (this.tokenOverride != null) {
-      token = Promise.resolve({ accessToken: this.tokenOverride });
-    } else {
-      token = (this.app as _FirebaseApp).INTERNAL.getToken(forceRefresh);
-    }
-    return token.then(tokenData => {
+    return (this.app as _FirebaseApp).INTERNAL.getToken(forceRefresh).then(tokenData => {
       // Cancel the request since the user changed while the request was
       // outstanding so the response is likely for a previous user (which
       // user, we can't be sure).
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -18,7 +18,10 @@ import { User } from '../auth/user';
 import { assert, fail } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { FirebaseApp } from '@firebase/app-types';
-import { _FirebaseApp, FirebaseAuthTokenData } from '@firebase/app-types/private';
+import {
+  _FirebaseApp,
+  FirebaseAuthTokenData
+} from '@firebase/app-types/private';
 
 // TODO(mikelehen): This should be split into multiple files and probably
 // moved to an auth/ folder to match other platforms.
@@ -188,29 +191,27 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     } else {
       token = (this.app as _FirebaseApp).INTERNAL.getToken(forceRefresh);
     }
-    return token.then(
-      tokenData => {
-        // Cancel the request since the user changed while the request was
-        // outstanding so the response is likely for a previous user (which
-        // user, we can't be sure).
-        if (this.userCounter !== initialUserCounter) {
-          throw new FirestoreError(
-            Code.ABORTED,
-            'getToken aborted due to uid change.'
+    return token.then(tokenData => {
+      // Cancel the request since the user changed while the request was
+      // outstanding so the response is likely for a previous user (which
+      // user, we can't be sure).
+      if (this.userCounter !== initialUserCounter) {
+        throw new FirestoreError(
+          Code.ABORTED,
+          'getToken aborted due to uid change.'
+        );
+      } else {
+        if (tokenData) {
+          assert(
+            typeof tokenData.accessToken === 'string',
+            'Invalid tokenData returned from getToken():' + tokenData
           );
+          return new OAuthToken(tokenData.accessToken, this.currentUser);
         } else {
-          if (tokenData) {
-            assert(
-              typeof tokenData.accessToken === 'string',
-              'Invalid tokenData returned from getToken():' + tokenData
-            );
-            return new OAuthToken(tokenData.accessToken, this.currentUser);
-          } else {
-            return null;
-          }
+          return null;
         }
       }
-    );
+    });
   }
 
   invalidateToken(): void {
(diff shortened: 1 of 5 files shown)


packages/testing/src/api/index.ts
```diff
@@ -14,9 +14,12 @@
  * limitations under the License.
  */
 
+import { firebase } from '@firebase/app';
 import * as admin from 'firebase-admin';
 import request from 'request-promise';
 import * as fs from 'fs';
+import { FirebaseApp } from '@firebase/app-types';
+import * as util from '@firebase/util';
```
tonymeng: is there a better way to import these?

packages/firestore/src/api/credentials.ts
```diff
@@ -179,7 +182,13 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     const initialUserCounter = this.userCounter;
     const forceRefresh = this.forceRefresh;
     this.forceRefresh = false;
-    return (this.app as _FirebaseApp).INTERNAL.getToken(forceRefresh).then(
+    var token: Promise<FirebaseAuthTokenData>;
```
ryanpbrewster: Do we have to modify the Firestore SDK in order to get this to work? I'd prefer not to introduce this `tokenOverride` value.

packages/testing/src/api/index.ts
```diff
@@ -64,6 +67,30 @@ export function initializeTestApp(options: any): admin.app.App {
...
+    util.base64.encodeString(JSON.stringify(header),/*webSafe=*/true),
+    util.base64.encodeString(JSON.stringify(options.auth),/*webSafe=*/true),
+    "fakesignature"
+  ].join(".");
+  return firebase.initializeApp({
+      projectId: options.projectId,
+      tokenOverride: fakeToken
+    },
+    'app-' + new Date().getTime() + "-" + Math.random()
+  );
```
ryanpbrewster: Can we initialize the app here, then immediately hijack the `INTERNAL.getToken` function?

packages/firestore/src/api/credentials.ts
tonymeng: AFAICT, we'd have to change `app-types` to export `FirebaseAppImpl` or to move `INTERNAL` into `FirebaseApp` to give visibility to the testing package. If that's the only way to do this, I'd rather modify the `firestore` package than `app` package

merlinnot:
I’m soooo happy to see this PR! You can’t even imagine!

packages/testing/src/api/index.ts
```diff
@@ -64,6 +67,34 @@ export function initializeTestApp(options: any): admin.app.App {
   );
 }
 
+export function initializeFirestoreTestApp(options: any): FirebaseApp {
+  if (!('projectId' in options)) {
+    throw new Error('projectId not specified');
+  }
+  if (typeof options.auth != 'object') {
+    throw new Error('auth must be an object');
+  }
```
ryanpbrewster: If you want, you could declare the type for `options` (instead of leaving it as `any`) and let the compiler construct these assertions for you.

packages/testing/src/api/index.ts
```diff
@@ -64,6 +67,34 @@ export function initializeTestApp(options: any): admin.app.App {
...
+    'fakesignature'
+  ].join('.');
+  let app = firebase.initializeApp(
+    {
+      projectId: options.projectId,
+      tokenOverride: fakeToken
+    },
+    'app-' + new Date().getTime() + '-' + Math.random()
+  );
+  (app as any).INTERNAL.getToken = () =>
```
ryanpbrewster: can you add a comment here explaining that this `INTERNAL.getToken` hijack is to inject the predefined authorization headers into the app (thus bypassing FirebaseAuth)?

ryanpbrewster:
Also please don't submit this until at least one of the other reviewers LGTMs it.

packages/testing/src/api/index.ts
tonymeng: done

packages/testing/src/api/index.ts
tonymeng: done

packages/testing/src/api/index.ts
```diff
@@ -49,10 +56,11 @@ export function initializeAdminApp(options: any): admin.app.App {
   );
 }
 
-export function initializeTestApp(options: any): admin.app.App {
-  if (!('databaseName' in options)) {
-    throw new Error('databaseName not specified');
-  }
+export type TestAppOptions = {
```
schmidt-sebastian: I would suggest to make this an interface or to just inline it and use two arguments in `initializeTestApp`.

packages/testing/src/api/index.ts
```diff
@@ -36,6 +39,10 @@ export function apps(): (admin.app.App | null)[] {
   return admin.apps;
 }
 
+export function firestoreApps(): (FirebaseApp | null)[] {
```
schmidt-sebastian: The distinction between `Admin App`, `Test App` and `Firestore Test App` is not very clear.  Would it be possible to add an argument to `initializeTestApp` that allows us to turn on Firestore support?

packages/testing/src/api/index.ts
```diff
@@ -49,10 +56,11 @@ export function initializeAdminApp(options: any): admin.app.App {
   );
 }
 
-export function initializeTestApp(options: any): admin.app.App {
-  if (!('databaseName' in options)) {
-    throw new Error('databaseName not specified');
-  }
+export type TestAppOptions = {
+  databaseName: string;
+  auth: Object;
```
schmidt-sebastian: Note that Object will only allow empty objects. You should use `object`.

https://blog.mariusschulz.com/2017/02/24/typescript-2-2-the-object-type

packages/testing/src/api/index.ts
```diff
@@ -64,6 +72,32 @@ export function initializeTestApp(options: any): admin.app.App {
...
+    kid: 'fakekid'
+  };
+  const fakeToken = [
+    base64.encodeString(JSON.stringify(header), /*webSafe=*/ false),
+    base64.encodeString(JSON.stringify(options.auth), /*webSafe=*/ false),
+    'fakesignature'
+  ].join('.');
+  const app = firebase.initializeApp(
+    { projectId: options.projectId },
+    'app-' + new Date().getTime() + '-' + Math.random()
```
schmidt-sebastian: You can make this a little more intuitive if you add a local variable that will help let users know that this is the `appName`.

packages/testing/test/database.test.ts
```diff
@@ -108,13 +126,27 @@ describe('Testing Module Tests', function() {
     ).to.throw(/Could not find file/);
   });
 
-  it('apps() returns all created apps', async function() {
+  it('apps() returns all apps created except by initializeFirestoreTestApp', async function() {
```
schmidt-sebastian: This behavior is strange and unexpected.

packages/testing/src/api/index.ts
```diff
@@ -32,36 +35,63 @@ class FakeCredentials {
...
+  projectId: string;
+  auth: object;
+}
+export function initializeTestApp(options: DatabaseAppOptions | FirestoreAppOptions): FirebaseApp {
+  let appOptions: FirebaseOptions;
+  if ('databaseName' in options) {
+    appOptions = {
+      databaseURL: DBURL + '?ns=' + options.databaseName
+    };
+  } else if ('projectId' in options) {
```
schmidt-sebastian: Is there a reason to not support both `projectId` and `databseURL`? This code might be a little cleaner if you supported both, even if the interface doesn't allow for it.

packages/testing/src/api/index.ts
schmidt-sebastian: Almost :) But it looks fine the way it is.

packages/testing/index.ts
```diff
@@ -26,5 +26,6 @@ export {
   assertSucceeds,
   initializeAdminApp,
   initializeTestApp,
+  initializeFirestoreTestApp,
```
schmidt-sebastian: You need to remove this to get the CI to pass.

tonymeng: done

packages/testing/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@firebase/testing",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
jshcrowthe: We don't need to do this, even though it is a breaking change major `0`s allow for breaking changes between other versions also.

tonymeng: done

PR #1013, opened 2018-07-17
[Multi-Tab] Zombie all client data

schmidt-sebastian:
This uses logic inspired by the current "zombied owner" logic to also zombie client metadata. This should speed up primary tab election, since any zombied tab will no longer be eligible.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -66,8 +66,6 @@ const CLIENT_METADATA_MAX_AGE_MS = 5000;
  * if they're already performing an IndexedDB operation.
  */
 const CLIENT_METADATA_REFRESH_INTERVAL_MS = 4000;
-/** LocalStorage location to indicate a zombied client id (see class comment). */
-const ZOMBIED_PRIMARY_LOCALSTORAGE_SUFFIX = 'zombiedClientId';
 /** User-facing error when the primary lease is required but not available. */
 const PRIMARY_LEASE_LOST_ERROR_MSG =
   'The current tab is not in the required state to perform this operation. ' +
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -352,9 +353,6 @@ export class IndexedDbPersistence implements Persistence {
     if (!this.started) {
       return Promise.resolve();
     }
-    // TODO(multitab): Similar to the zombied client ID, we should write an
-    // entry to Local Storage first to indicate that we are no longer alive.
-    // This will help us when the shutdown handler doesn't run to completion.
     this.started = false;
```
mikelehen: Should you have added a markClientZombied() call?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -631,15 +630,15 @@ export class IndexedDbPersistence implements Persistence {
...
-      );
+      const isZombied =
+        this.window.localStorage.getItem(
+          this.zombiedClientLocalStorageKey(clientId)
+        ) !== null;
       log.debug(
         LOG_TAG,
-        'Zombied clientId from LocalStorage:',
-        zombiedClientId
+        `Client ${isZombied ? 'is' : 'is not'} zombied in LocalStorage`
```
mikelehen: include clientId in this log message?

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: It's called in the window unload handler before all the rest of the stuff that runs on the AsyncQueue.

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_persistence.ts
mikelehen: Is that the only place shutdown() is called?

mikelehen: If we decide we don't want/need to mark zombie for other shutdown() cases, that's fine...  but then the removeClientZombiedEntry() should only be in the unload handler as well.

mikelehen: 👍

PR #1014, opened 2018-07-17
b/72533250: Fix issue with limbo resolutions triggering incorrect manufactured deletes.

mikelehen:
_NOTE: This is a minimal fix for the issue we discussed; I think a cleaner fix is to move View.syncedDocuments out of the view.  I'll poke at this but it'll probably take longer and be a bit more risky_

This fixes an issue occurring when a limbo target receives a documentUpdate,
then a global snapshot, and then a CURRENT. Because there was a global
snapshot before the CURRENT, WatchChangeAggregator has no pending document
updates and calls SyncEngine.targetContainsDocument() to see if we got any
document from the backend for the target. See:
https://github.com/firebase/firebase-js-sdk/blob/6905339235ad801291edc696dd75a08e80647f5b/packages/firestore/src/remote/watch_change.ts#L422

Prior to this change, targetContainsDocument() returned false because
it relies on our Views to track the contents of the target, and we don't
have Views for limbo targets. Thus WatchChangeAggregator incorrectly
manufactures a NoDocument document update which deletes data from our
cache.

The fix is to have SyncEngine track the fact that we did indeed get
a document for the limbo resolution and return true from
targetContainsDocument().

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -36,6 +36,7 @@ import { assert, fail } from '../util/assert';
 import { FirestoreError } from '../util/error';
 import * as log from '../util/log';
 import { AnyJs, primitiveComparator } from '../util/misc';
+import * as objUtils from '../util/obj';
 import { ObjectMap } from '../util/obj_map';
 import { Deferred } from '../util/promise';
 import { SortedMap } from '../util/sorted_map';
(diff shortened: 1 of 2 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -301,6 +317,21 @@ export class SyncEngine implements RemoteSyncer {
...
+      const limboResolution = this.limboResolutionsByTarget[targetId];
+      if (limboResolution) {
+        if (
+          targetChange.addedDocuments.size > 0 ||
+          targetChange.modifiedDocuments.size > 0
+        ) {
+          limboResolution.receivedDocument = true;
+        } else if (targetChange.removedDocuments.size > 0) {
+          limboResolution.receivedDocument = false;
+        }
```
wilhuff: Isn't there an `else` case here too? If we got a snapshot but there's no document at all it won't be added, modified, or removed. How are we supposed to handle that here?

If not, can you add a comment here/assertion about this?

PR #1017, opened 2018-07-18
Modify spec test to demonstrate deletedDoc issue.

mikelehen:
FYI- This demonstrates the modified spec test to verify we don't overwrite a (newer) deletedDoc with an (older) doc, which fails with your current branch.

### DIFF
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -283,6 +283,50 @@ describeSpec('Listens:', [], () => {
     );
   });
 
+  specTest('Individual (deleted) documents cannot revert', [], () => {
+    const allQuery = Query.atPath(path('collection'));
+    const visibleQuery = Query.atPath(path('collection')).addFilter(
+      filter('visible', '==', true)
+    );
+    const docAv1 = doc('collection/a', 1000, { visible: true, v: 'v1000' });
+    const docAv2 = doc('collection/a', 2000, { visible: false, v: 'v2000' });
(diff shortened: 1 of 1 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -283,6 +283,50 @@ describeSpec('Listens:', [], () => {
...
+        .watchSnapshots(3000)
+        .expectEvents(allQuery, { removed: [docAv1], fromCache: false })
+        .userUnlistens(allQuery)
+        .watchRemoves(allQuery)
+        // Supposing we sent a resume token for visibleQuery, watch could catch
+        // us up to docAV2 since that's the last relevant change to the query
+        // (the document falls out) and send us a snapshot that's ahead of
+        // docAv3 (which is already in our cache).
+        .userListens(visibleQuery, 'resume-token-1000')
+        .watchAcksFull(visibleQuery, 5000, docAv2)
```
schmidt-sebastian: I don't think Watch can ever send CURRENT to a query and not also send the last version of a document. We should receive v3 before we get CURRENT - even if the document doesn't match anymore.

mikelehen: That's the question we need Jonny or somebody from watch to answer.  If that's true, it would make life much easier...  but 1) I'm not sure how watch can guarantee that if it's using a per-target change log to catch us up, and 2) these spec tests (I just copied the one above) seem to specifically be assuming that watch _can_  send us a CURRENT after only giving us the latest version that affected the query.

PR #1019, opened 2018-07-18
Add a schema migration that drops the query cache

wilhuff:
This is a force fix for potential existence filter mismatches caused by
https://github.com/firebase/firebase-ios-sdk/issues/1548

The essential problem is that when resuming a query, the server is
allowed to forget deletes. If the number of incorrectly synthesized
deletes matches the number of server-forgotten deletes then the
existence filter can give a false positive, preventing the cache from
self healing.

Dropping the query cache clears any client-side resume token which
prevents a false positive existence filter mismatch.

Note that the remote document cache and mutation queues are unaffected
so any cached documents that do exist will still work while offline.

### DIFF
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -30,8 +30,11 @@ import { SnapshotVersion } from '../core/snapshot_version';
  * 1. Initial version including Mutation Queue, Query Cache, and Remote Document
  *    Cache
  * 2. Added targetCount to targetGlobal row.
+ * 3. Dropped and re-created Query Cache to deal with cache corruption related
+ *    to limbo resolution. Addresses
+ *    https://github.com/firebase/firebase-ios-sdk/issues/1548
  */
-export const SCHEMA_VERSION = 2;
+export const SCHEMA_VERSION = 3;
 
 /**
  * Performs database creation and schema upgrades.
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -67,6 +67,13 @@ export function createOrUpgradeDb(
       saveTargetCount(txn, targetGlobal)
     );
   }
+
+  // Brand new clients don't need to drop and recreate--only clients that
+  // potentially have corrupt data.
+  if (fromVersion !== 0 && fromVersion < 3 && toVersion >= 3) {
```
schmidt-sebastian: You should wait for `p` before you continue with the migration.

packages/firestore/test/unit/local/indexeddb_schema.test.ts
```diff
@@ -135,4 +138,63 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+          'readwrite',
+          [DbTarget.store, DbMutationBatch.store],
+          txn => {
+            const targets = txn.store<DbTargetKey, DbTarget>(DbTarget.store);
+            const mutations = txn.store<DbMutationBatchKey, DbMutationBatch>(
+              DbMutationBatch.store
+            );
+
+            return PersistencePromise.resolve()
+              .next(() => targets.get(targetId))
```
mikelehen: Per Sebastian's comment above, you could do `return targets.get(...).next() => ...)` here too, I think?

PR #1022, opened 2018-07-18
Add iat to fake access token payload

ryanpbrewster:


packages/testing/src/api/index.ts
```diff
@@ -43,39 +50,49 @@ export type AppOptions = {
...
+  projectId?: string;
+};
+/** Construct a FirebaseApp authenticated as an admin user. */
+export function initializeAdminApp(options: AdminAppOptions): FirebaseApp {
+  const app = initializeApp(options.databaseName, options.projectId);
+  // hijacking INTERNAL.getToken to bypass FirebaseAuth and allows specifying of auth headers
+  (app as any).INTERNAL.getToken = () =>
+    Promise.resolve({ accessToken: ADMIN_TOKEN });
+  return app;
+}
```
tonymeng: can you combine the logic for `initializeTestApp` and `initializeAdminApp` since they're essentially the same thing with a different accessToken

packages/testing/test/database.test.ts
```diff
@@ -47,56 +47,34 @@ describe('Testing Module Tests', function() {
...
-    let claims = base64.decodeString(
-      token.accessToken.split('.')[1],
-      /*webSafe=*/ false
+    const token = await (app as any).INTERNAL.getToken();
+    expect(token).to.have.keys('accessToken');
+    const claims = JSON.parse(
+      base64.decodeString(token.accessToken.split('.')[1], /*webSafe=*/ false)
     );
-    expect(claims).to.equal('{}');
+    expect(claims).to.have.property('uid', auth.uid);
```
tonymeng: expect iat to be set

tonymeng: let's make it closed

PR #1027, opened 2018-07-19
Unify local.QueryData with the other platforms

rsgowman:
This involves:
1) Making QueryData immutable
2) Renaming QueryData.update to QueryData.copy

### DIFF
--- a/packages/firestore/src/local/query_data.ts
+++ b/packages/firestore/src/local/query_data.ts
@@ -62,7 +62,7 @@ export class QueryData {
    * Creates a new query data instance with an updated snapshot version and
    * resume token.
    */
-  update(updated: {
+  copy(updated: {
     resumeToken: ProtoByteString;
     snapshotVersion: SnapshotVersion;
   }): QueryData {
--- a/packages/firestore/src/local/query_data.ts
+++ b/packages/firestore/src/local/query_data.ts
@@ -39,23 +39,23 @@ export enum QueryPurpose {
 export class QueryData {
   constructor(
     /** The query being listened to. */
-    public query: Query,
+    readonly query: Query,
     /**
      * The target ID to which the query corresponds; Assigned by the
      * LocalStore for user listens and by the SyncEngine for limbo watches.
      */
-    public targetId: TargetId,
+    readonly targetId: TargetId,
     /** The purpose of the query. */
-    public purpose: QueryPurpose,
+    readonly purpose: QueryPurpose,
     /** The latest snapshot version seen for this target. */
-    public snapshotVersion: SnapshotVersion = SnapshotVersion.MIN,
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/remote/remote_event.test.ts
```diff
@@ -668,8 +668,7 @@ describe('RemoteEvent', () => {
       WatchTargetChangeState.Current,
       [1, 2]
     );
-    const targets = listens(1, 2);
-    targets[2].purpose = QueryPurpose.LimboResolution;
+    const targets = { ...listens(1), ...limboListens(2) };
```
mikelehen: Nifty! I didn't know you could do this.

mikelehen:
Oops.  Didn't pay attention to the assignee. ;-)

packages/firestore/src/local/query_data.ts
```diff
@@ -39,30 +39,30 @@ export enum QueryPurpose {
...
-    public resumeToken: ProtoByteString = emptyByteString()
+    readonly resumeToken: ProtoByteString = emptyByteString()
   ) {}
 
   /**
    * Creates a new query data instance with an updated snapshot version and
    * resume token.
    */
-  update(updated: {
+  copy(updated: {
```
schmidt-sebastian: This is fine as is, but the API contract seems kind of strange. What do you think of 

```
copy(overwrite: {
  resumeToken?: ProtoByteString;
   snapshotVersion?: SnapshotVersion;
)
```

Then this could be an actual copy or a customized copy.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -424,7 +424,10 @@ export class RemoteStore implements TargetMetadataProvider {
 
       // Clear the resume token for the query, since we're in a known mismatch
       // state.
-      queryData.resumeToken = emptyByteString();
+      this.listenTargets[targetId] = queryData.copy({
+        snapshotVersion: queryData.snapshotVersion,
```
schmidt-sebastian: If you follow my suggestion above, this just becomes ` { resumeToken: empyByteString() }`.

packages/firestore/src/local/query_data.ts
```diff
@@ -39,39 +39,39 @@ export enum QueryPurpose {
...
+    resumeToken?: ProtoByteString;
+    snapshotVersion?: SnapshotVersion;
   }): QueryData {
     return new QueryData(
       this.query,
       this.targetId,
       this.purpose,
-      updated.snapshotVersion,
-      updated.resumeToken
+      overwrite.snapshotVersion || this.snapshotVersion,
```
schmidt-sebastian: Hm, sorry! I just noticed that this won't work. You will not be able to set `snapshotVersion` to 0 (SnapshotVersion.MIN) using this code. You need to explicitly check for `undefined`.

PR #1028, opened 2018-07-19
Use Alex's suggested phrasing instead

wilhuff:


packages/firestore/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Unreleased
-- [fixed] Fixed an issue where documents could be cached as deleted when
-  they should not have been, leading to queries returning fewer results than
-  they should (firebase/firebase-ios-sdk#1548).
+- [fixed] Fixed an issue where queries returned fewer results than they should,
+  caused by documents that were cached as deleted when they should not have
+  been (firebase/firebase-ios-sdk#1548). Some cache data is cleared and so
+  clients may use extra bandwidth the first time they launch with this version
+  of the SDK.
```
alexg33: Suggest: "Because some cache data is cleared, clients might use extra bandwidth the first time they launch..."

Makes it a little more direct, but not a big deal.

PR #1029, opened 2018-07-19
Ensure that we create an empty TargetGlobal row.

mikelehen:
Ensure the v3 migration unconditionally creates the TargetGlobal row.  Remove the no-longer-necessary v2 schema migration.


packages/firestore/test/unit/local/indexeddb_schema.test.ts
```diff
@@ -101,58 +91,35 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
-    );
-  });
-
   it('drops the query cache from 2 to 3', () => {
     const userId = 'user';
     const batchId = 1;
     const targetId = 2;
 
     const expectedMutation = new DbMutationBatch(userId, batchId, 1000, []);
+    const dummyTargetGlobal = new DbTargetGlobal(
```
wilhuff: Add parameter name= comments here too?

mikelehen: Sure, done.

PR #1039, opened 2018-07-24
 Comment out multi-tab flag

schmidt-sebastian:
This comments out the multi-tab flag in one `.d.ts` file and adds it to another (where I missed adding it).

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -769,6 +769,27 @@ declare namespace firebase.firestore {
     timestampsInSnapshots?: boolean;
   }
 
+// TODO(multitab): Uncomment when multi-tab is released publicly.
+// /**
+//  * Settings that can be passed to Firestore.enablePersistence() to configure
+//  * Firestore persistence.
+//  */
+// export interface PersistenceSettings {
+//   /**
+//    * Whether to synchronize the in-memory state of multiple tabs. Setting this
+//    * to 'true' in all open tabs enables shared access to local persistence,
+//    * shared execution of queries and latency-compensated local document updates
+//    * across all connected instances.
+//    *
+//    * To enable this mode, `experimentalTabSynchronization:true` needs to be set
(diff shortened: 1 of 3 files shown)


packages/firebase/index.d.ts
```diff
@@ -808,6 +829,29 @@ declare namespace firebase.firestore {
...
+    //  * Must be called before any other methods (other than settings()).
+    //  *
+    //  * If this fails, enablePersistence() will reject the promise it returns.
+    //  * Note that even after this failure, the firestore instance will remain
+    //  * usable, however offline persistence will be disabled.
+    //  *
+    //  * There are several reasons why this can fail, which can be identified by
+    //  * the `code` on the error.
+    //  *
+    //  *   * failed-precondition: The app is already open in another browser tab.
```
mikelehen: Technically this should probably have some mention of experimentalTabSynchronization...  but since it's experimental maybe we don't care to give it too much verbiage in this method description...

packages/firestore/src/api/database.ts
```diff
@@ -202,6 +202,10 @@ class FirestoreConfig {
   persistence: boolean;
 }
 
+// TODO(multi-tab): Replace with Firestore.PersistenceSettings
```
mikelehen: multitab (or else pinky swear that you'll grep for "TODO(multi" before we declare multi-tab done).

schmidt-sebastian: Fixed

packages/firebase/index.d.ts
```diff
@@ -808,6 +829,29 @@ declare namespace firebase.firestore {
      */
     enablePersistence(): Promise<void>;
 
+    // TODO(multitab): Uncomment when multi-tab is released publicly.
```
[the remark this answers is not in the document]
schmidt-sebastian: I'll take care of that when the day comes :)

PR #1041, opened 2018-07-24
Refactor PersistentStream.

mikelehen:
[FYI- I'll probably have @wilhuff take a look at this too since he had opinions about the stream listener stuff, but in the interest of saving him time, can you do the first pass?]

This breaks out a number of changes I made as prep for b/80402781 (Continue
retrying streams for 1 minute (idle delay)).

PersistentStream changes:
* Rather than providing a stream event listener to every call of start(),
  the stream listener is now provided once to the constructor and cannot
  be changed.
* Streams can now be restarted indefinitely, even after a call to stop().
  * PersistentStreamState.Stopped was removed and we just return to
    'Initial' after a stop() call.
  * Added `currentAuthAttempt` member to PersistentStream in order to avoid
    bleedthrough issues if stop() / start() are called while an auth attempt
    is in progress (i.e. so we correctly abandon the old auth attempt).
  * Calling stop() now triggers the onClose() event listener, which
    simplifies stream cleanup.
* PersistentStreamState.Auth renamed to 'Starting' to better reflect that
  it encompasses both authentication and opening the stream.

RemoteStore changes:
* Creates streams once and just stop() / start()s them as necessary,
  never recreating them completely.
  * Added networkEnabled flag to track whether the network is
    enabled or not, since we no longer null out the streams.
  * Refactored disableNetwork() / enableNetwork() to remove stream
    re-creation. They're now simple enough that I stopped re-using
    them from shutdown() and handleUserChange().

Misc:
* Comment improvements including a state diagram on PersistentStream.
* Fixed spec test shutdown to schedule via the AsyncQueue to fix
  sequencing order I ran into.

### DIFF
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -43,21 +43,38 @@ interface ListenRequest extends api.ListenRequest {
 export interface WriteRequest extends api.WriteRequest {
   database?: string;
 }
-
+/**
+ * PersistentStream can be in one of 5 states (each described in detail below)
+ * based on the following state transition diagram:
+ *
+ *          start() called             auth & connection succeeded
+ * INITIAL ----------------> STARTING -----------------------------> OPEN
+ *                             ^  |                                   |
+ *                             |  |                    error occurred |
+ *                             |  \-----------------------------v-----/
+ *                             |                                |
+ *                    backoff  |                                |
+ *                    elapsed  |              start() called    |
+ *                             \--- BACKOFF <---------------- ERROR
+ *
+ * [any state] --------------------------> INITIAL
+ *               stop() called or
+ *               idle timer expired
+ */
 enum PersistentStreamState {
   /**
-   * The streaming RPC is not running and there's no error condition.
+   * The streaming RPC is not yet running and there's no error condition.
    * Calling `start` will start the stream immediately without backoff.
    * While in this state isStarted will return false.
    */
   Initial,
 
   /**
-   * The stream is starting, and is waiting for an auth token to attach to
-   * the initial request. While in this state, isStarted will return
-   * true but isOpen will return false.
+   * The stream is starting, either waiting for an auth token or for the stream
+   * to successfully open. While in this state, isStarted will return true but
+   * isOpen will return false.
    */
-  Auth,
+  Starting,
 
   /**
    * The streaming RPC is up and running. Requests and responses can flow
@@ -68,22 +85,16 @@ enum PersistentStreamState {
   /**
    * The stream encountered an error. The next start attempt will back off.
    * While in this state isStarted() will return false.
-   *
    */
   Error,
 
   /**
    * An in-between state after an error where the stream is waiting before
-   * re-starting. After
-   * waiting is complete, the stream will try to open. While in this
-   * state isStarted() will return YES but isOpen will return false.
+   * re-starting. After waiting is complete, the stream will try to open.
+   * While in this state isStarted() will return true but isOpen will return
+   * false.
    */
-  Backoff,
-
-  /**
-   * The stream has been explicitly stopped; no further events will be emitted.
-   */
-  Stopped
+  Backoff
 }
 
 /**
@@ -125,6 +136,7 @@ const IDLE_TIMEOUT_MS = 60 * 1000;
  *   - Exponential backoff on failure
  *   - Authentication via CredentialsProvider
  *   - Dispatching all callbacks into the shared worker queue
+ *   - Closing idle streams after 60 seconds of inactivity
  *
  * Subclasses of PersistentStream implement serialization of models to and
  * from the JSON representation of the protocol buffers for a specific
@@ -157,16 +169,18 @@ export abstract class PersistentStream<
   private inactivityTimerPromise: CancelablePromise<void> | null = null;
   private stream: Stream<SendType, ReceiveType> | null = null;
 
-  protected backoff: ExponentialBackoff;
+  /** A sentinel object used to ignore auth callbacks for prior attempts. */
+  private currentAuthAttempt: object = null;
 
-  protected listener: ListenerType | null = null;
+  protected backoff: ExponentialBackoff;
 
   constructor(
     private queue: AsyncQueue,
     connectionTimerId: TimerId,
     private idleTimerId: TimerId,
     protected connection: Connection,
-    private credentialsProvider: CredentialsProvider
+    private credentialsProvider: CredentialsProvider,
+    protected listener: ListenerType
   ) {
     this.backoff = new ExponentialBackoff(
       queue,
@@ -187,14 +201,14 @@ export abstract class PersistentStream<
    */
   isStarted(): boolean {
     return (
-      this.state === PersistentStreamState.Backoff ||
-      this.state === PersistentStreamState.Auth ||
-      this.state === PersistentStreamState.Open
+      this.state === PersistentStreamState.Starting ||
+      this.state === PersistentStreamState.Open ||
+      this.state === PersistentStreamState.Backoff
     );
   }
 
   /**
-   * Returns true if the underlying RPC is open (the openHandler has been
+   * Returns true if the underlying RPC is open (the onOpen callback has been
    * called) and the stream is ready for outbound requests.
    */
   isOpen(): boolean {
@@ -206,16 +220,15 @@ export abstract class PersistentStream<
    * not immediately ready for use: onOpen will be invoked when the RPC is ready
    * for outbound requests, at which point isOpen will return true.
    *
-   *  When start returns, isStarted will return true.
+   * When start returns, isStarted will return true.
    */
-  start(listener: ListenerType): void {
+  start(): void {
     if (this.state === PersistentStreamState.Error) {
-      this.performBackoff(listener);
+      this.performBackoff();
       return;
     }
 
     assert(this.state === PersistentStreamState.Initial, 'Already started');
-    this.listener = listener;
     this.auth();
   }
 
@@ -227,7 +240,7 @@ export abstract class PersistentStream<
    */
   stop(): void {
     if (this.isStarted()) {
-      this.close(PersistentStreamState.Stopped);
+      this.close(PersistentStreamState.Initial);
     }
   }
 
@@ -299,8 +312,7 @@ export abstract class PersistentStream<
    * * sets internal stream state to 'finalState';
    * * adjusts the backoff timer based on the error
    *
-   * A new stream can be opened by calling `start` unless `finalState` is set to
-   * `PersistentStreamState.Stopped`.
+   * A new stream can be opened by calling `start`.
    *
    * @param finalState the intended state of the stream after closing.
    * @param error the error the connection was closed with.
@@ -309,6 +321,7 @@ export abstract class PersistentStream<
     finalState: PersistentStreamState,
     error?: FirestoreError
   ): Promise<void> {
+    assert(this.isStarted(), 'Only started streams should be closed.');
     assert(
       finalState === PersistentStreamState.Error || isNullOrUndefined(error),
       "Can't provide an error when not in an error state."
@@ -344,19 +357,15 @@ export abstract class PersistentStream<
       this.stream = null;
     }
 
+    // If there was a current auth attempt, make sure we ignore the response.
+    this.currentAuthAttempt = null;
+
     // This state must be assigned before calling onClose() to allow the callback to
     // inhibit backoff or otherwise manipulate the state in its non-started state.
     this.state = finalState;
-    const listener = this.listener!;
 
-    // Clear the listener to avoid bleeding of events from the underlying streams.
-    this.listener = null;
-
-    // If the caller explicitly requested a stream stop, don't notify them of a closing stream (it
-    // could trigger undesirable recovery logic, etc.).
-    if (finalState !== PersistentStreamState.Stopped) {
-      return listener.onClose(error);
-    }
+    // Notify the listener that the stream closed.
+    await this.listener.onClose(error);
   }
 
   /**
@@ -386,19 +395,26 @@ export abstract class PersistentStream<
       'Must be in initial state to auth'
     );
 
-    this.state = PersistentStreamState.Auth;
+    this.state = PersistentStreamState.Starting;
+
+    // Create sentinel object used to identify this auth attempt in case it
+    // is invalidated by a call to stop().
+    const authAttempt = (this.currentAuthAttempt = new Object());
 
     this.credentialsProvider.getToken().then(
       token => {
-        // Normally we'd have to schedule the callback on the AsyncQueue.
-        // However, the following calls are safe to be called outside the
-        // AsyncQueue since they don't chain asynchronous calls
-        this.startStream(token);
+        // Stream can be stopped while waiting for authentication.
+        if (this.currentAuthAttempt === authAttempt) {
+          // Normally we'd have to schedule the callback on the AsyncQueue.
+          // However, the following calls are safe to be called outside the
+          // AsyncQueue since they don't chain asynchronous calls
+          this.startStream(token);
+        }
       },
       (error: Error) => {
         this.queue.enqueue(async () => {
-          if (this.state !== PersistentStreamState.Stopped) {
-            // Stream can be stopped while waiting for authorization.
+          // Stream can be stopped while waiting for authentication.
+          if (this.currentAuthAttempt === authAttempt) {
             const rpcError = new FirestoreError(
               Code.UNKNOWN,
               'Fetching auth token failed: ' + error.message
@@ -411,15 +427,11 @@ export abstract class PersistentStream<
   }
 
   private startStream(token: Token | null): void {
-    if (this.state === PersistentStreamState.Stopped) {
-      // Stream can be stopped while waiting for authorization.
-      return;
-    }
-
     assert(
-      this.state === PersistentStreamState.Auth,
-      'Trying to start stream in a non-auth state'
+      this.state === PersistentStreamState.Starting,
+      'Trying to start stream in a non-starting state'
     );
+
     // Helper function to dispatch to AsyncQueue and make sure that any
     // close will seem instantaneous and events are prevented from being
     // raised after the close call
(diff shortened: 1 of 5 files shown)


packages/firestore/src/remote/remote_store.ts
```diff
@@ -136,82 +152,65 @@ export class RemoteStore implements TargetMetadataProvider {
...
-      this.writeStream = null;
-      this.watchStream = null;
     }
+
+    // Temporarily set networkEnabled to false so that our close handling
+    // doesn't try to restart the streams, since we are explicitly closing them.
+    const networkEnabled = this.networkEnabled;
+    this.networkEnabled = false;
+    this.writeStream.stop();
+    this.watchStream.stop();
```
schmidt-sebastian: Do we still need to restart the stream if we did not receive an error? It seems like that might be enough to decide whether we should restart or not.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -300,39 +295,40 @@ export class RemoteStore implements TargetMetadataProvider {
...
+      this.canUseNetwork() &&
       !this.watchStream.isStarted() &&
       !objUtils.isEmpty(this.listenTargets)
     );
   }
 
+  private canUseNetwork(): boolean {
+    // TODO(mikelehen): This could take into account isPrimary when we merge
+    // with multitab.
+    return this.networkEnabled;
```
schmidt-sebastian: Thanks! :)

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -425,9 +425,11 @@ abstract class TestRunner {
   }
 
   async shutdown(): Promise<void> {
-    await this.remoteStore.shutdown();
-    await this.persistence.shutdown(/* deleteData= */ true);
-    await this.destroyPersistence();
+    await this.queue.enqueue(async () => {
```
schmidt-sebastian: I hope this is going to solve all the shutdown troubles I see in the spec tests when I hit assertions. Fingers crossed!

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -43,21 +43,38 @@ interface ListenRequest extends api.ListenRequest {
...
+ *                             |  \-----------------------------v-----/
+ *                             |                                |
+ *                    backoff  |                                |
+ *                    elapsed  |              start() called    |
+ *                             \--- BACKOFF <---------------- ERROR
+ *
+ * [any state] --------------------------> INITIAL
+ *               stop() called or
+ *               idle timer expired
+ */
```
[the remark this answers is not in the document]
wilhuff: Needs more shading. Like this:
```
****:***:::::::::::::::::::::::::::::::::::::::::::::::::::::::::::.::::::::::::::.:.::.::::::::::::::.:::::::::::::::::::::::::
****:::::::::::::::::::::::::::::::::::::::::::::::::::::..:::::::..:....::.::......:.::::...:::::..:..:..:..::.:::::::..:.:::::
***::::::::::::::::::::::::::::::...:::::.:::..:.:..........::::::.::...:.:.:..:.....::.:.........................:::::::...::::
**:::::I***I:::::::::::::::::::::::::::...:.::...:::....:.:.:::::...:.....:............::......:::......:...:::...:::::::.::..::
*:::::::*I:*:::::::::::::::::::::::::::::**IV......I*:::::::::::::.::..::...:::.::::..:.........::::....::.....::.::::::.:::::::
**::::::IF:*:::::::::::::::::::::::I:............:::*:..*...:.:::...::..:.......:............:..:.....:.::.....:....:::..:::..::
***:*::::V.::::::::::::.........I...............:::...::I..I:::::.::..:................................:.:..:...:::....:::::.:::
*::::*:*:*:*:::::::::.......I...............II:I:II:...:.*...:.*..:..::..................................:.:::::::.:::::::::::::
*:::I:::.:.*::::.:::::...::........*:.:...I*.:I:*IIV:.::***I:I:::.::.::::................................:::::::::::::::::::::::
********::*::*::::::...I.........:**:*III:IIIFVFNF:*I..**::.:*.I*:::::...........................::::..:::::::::::::::.:::::::::
***VF*:****:*::::::::V.........**:*:.::NFFFIIFNNNNF..::*I:*.:.::::::.............:..............:::.::::::::::::::::::::::::::::
***IV***I:::*:::::V..........::.::*NF*FNNFFNNVFFFNNNV:::.I.*..:.....I:.**...................:....:::::::::::::::::::::::::::::::
*******I::****::::...........INFFN*FNNNVNNNNVFFNNVFIFV:II*::*:*..**I:*:.::......................::.:::::::::::::::::::::::::::::
****I***********.............INFFNFNNNVNNFNVNNNNVFNNNNN:..::.:*.::..I....*:........................:::::::::::::::::::::::::::::
*************V:...............:FNNVFNFVFNNINNFVFNNVNNFF*...I*.*II..::..:.I.I:............................:.:...:.....:::::::::::
II********VI:.:....................:I::NNFFFVFFNNFNIFVF:*.:*::::.::..*:II:..:::.:.:....................................:.:.:.:.:
FFFFFFVFV:::.:::....::I*......I...:II:VFNVNIFV:NFNFFF:......:.:*:I:..I:**I..:..................................................:
*IVV*::.:.:...::I:::I*:...:..::::V:..FNI.V*INFIFFFFF:.........:I..::*.*.::*:::..:..............................................:
:::::::::*:...:*::.:I::*I.:*.*I::II.:.:.:VFF*FFNNVFF...:...........*I****...:.....................................:..........*..
:::*I***I*V:*FFV*::.***:I.*:.****I:::*::.:.*I*:NNVFI:................:.:..:::::..::.........:......................:.:.....*....
:::::*V**I*:.NFNIFNV:::*IV*.::FNVNVFIN**:I*FVFNFNFV::...:..:.::.::.::..:::.::...:...............:.....:.................:*.....:
::I***::I*:*:IVFVNFVV.:I**:*...VF*:FVFIVVVVF:NFINNFI........::.....::...............................::........:......:........:*
:V****INF*:*V*V*FINNF:**I*::I:*.IIIFNFNFNVVIFNFFNNFV........:.........................::...:.............:.:.:::......I........:
::*:IV::FNVVNNIFFFF:*IV::*VFN*II..*INNFNFFFNVVFVNNNVI::::::.:::::..:::...:....:.:...::.:........:::..:::::::::::.I..I......I*:*:
I:*NVNV:*V*VNFFFFFVV*:FFFIFFFI*:*N**VVNFNNN**NFIINNF.V:::::::::::::.:.:.:::.::::.::.:::::::::::::::::::::::::::*:.*I.......:.VV*
I::::::I:I**VFFI*NIFF*FVNVV:FNFFVIFFFFFVN*......*VFFNFF:::*::::::::::::::::::::::::::::::::::*:::***********IIFFI*......II*::VVN
**::*II:::..*:IINNFFFFFFVFFNFFFNFVVFF*FF:...........N*NVI*I************.************************IIIIIIIIII*FVVFI.........I..IFFF
****I*:I*:*I**I:*IVINFFFFNVFFNFVFVFIIF...............IVNFIII:IIII**IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIVVVIIIIIFV*FV.......**:F:VFVFFF
*I****I:::I*V*:F:I:**IIVIVFN*IFNVVIF........I:.......***FFFVIVIIVVVIVIVVIIIVVVIII:*IIIIIIIVIVVVVVVVVVVVFFVVV....*.:.:VIFFFFFNNN*
***NNNVNF*NFIII*I**I:I::I*I*::*VF*.........***I:.......:..*NFIIVVVVVVVVVVVVVVVV.....IIIVVVVVVVVVVVVIVVII:*...*:.:..VFVFFFFF::::.
:::FNVFVNIVFFFI*II*I**IIFFF*F:..........:**I:*..........::.....VVVVVVVVIVVVVI......:..*VVVVVVVVVFF*::*I........VVFINFFF:.:VF....
:::FFNFNFFFFNIFI*:I*V*::V:.........:::.I:.*:I*.**..:....:*.V:.:.VVVVVVVV**:*IFFN.IFFVI.:VFF*:*::*V...**.I*.VI*IVV:.:.:VVV..:.::
:::::::*FFNFVI:..........:..........I:..:**.I...::*..:::....II*.....IFNNNNNNNN*VVVVVFVVFVVFFVV:.....**VVVVVV*..:....*VV:.:::::::
:::.:::::..:....::......:V::::..II.I*II::::.I:I..*.*:VV.::..IIVF.:VI:.*...*:.:*II***::.......*I.FIF***..:::::.:VFFVII.::::::IVFF
::::*****:......I::.....:*I::::..::::*:......:::..FVVF*F*:*F..IVFV:V**:FF*::.........:............:::..FFFFFFFFFFF..:::::FFVVFFV
::::::I***I*.I.*:::.I..*I*:.*.I.......*.I*:..*::..*:F*:*I:*:V...*FFIFFFV*FFV::.....:..::...*.:..::**:.VVVFVVVVVI....:IFNNNNNNFI*
:::::::::.**::::::.*NV.I:.I*:*:..:V.:::::..::I.**..FNFVNI**FFV:V...*FNNFFIINFF::::.:I*::.:::::VFVVVVI*:*VNNFF*..:::::::::*VVFFFN
::::::*V:I*VF*IFV*NVFFI:**::.VI::::*.::I.:*.:.I:*:..FNFFFFV*IIV:*FVVIV.INFFFIFFNFI::..FFFFI::VNNFFI:*IIVFFV..:::::::VNNNNFVFNV**
::::FVVFF*NNFFNNNNFFV::I::.I*I::.....:*::VF.FF*::*:..:VNFFNFF***III:II:*V*::IFNFFNNFFF*:.IFNFFFFFFFFFFVI......:.:::*I::..:VFFFFF
:::NNVFVFNNFNFNFVV::*V*F.FNVN:I:::*IV:::.IFFNVIVI.:::..NNFFFNVN*II**IIVIVIV***********IIIV:FVIF*VFIV.....:*VFFFFFFFFFNNNFFFFFFFF
::NFVNFVIN*VFN*NNIIFINVFNNINVNN:.::..........FNN:VVF.....:NFFFNNVF*I:::::*IIIIIV***********:**I*...::::**I:IFNNFNNNNNNF*::IIVFNN
::NVFFNVNFNVNNVNNNVNNIFNFVNFNN**:::IVNN.I.FN*IIINIVNNI:V:...VNNFNFNFNV**::::::::::::::*II*....:FNFFNNNFFFNNNNNF:.......:::::::..
.:NINNVFFFNIFNVNFNVNFVNNNVNVNNF:::IN:*V**NI:V::VFIFNNNFV***I:...:VFFFNNFFFNNNNNNNV:.......:::......::::....:::::::::YRREBRAP*NAI
```

packages/firestore/src/remote/remote_store.ts
```diff
@@ -300,39 +295,40 @@ export class RemoteStore implements TargetMetadataProvider {
...
     objUtils.forEachNumber(this.listenTargets, (targetId, queryData) => {
       this.sendWatchRequest(queryData);
     });
   }
 
   private async onWatchStreamClose(error?: FirestoreError): Promise<void> {
     assert(
-      this.isNetworkEnabled(),
-      'onWatchStreamClose() should only be called when the network is enabled'
+      error !== undefined || !this.shouldStartWatchStream(),
```
wilhuff: I'm not a fan of this assertion here because the condition is hard to reason about. It's better to make the simple assertion that `error !== undefined` inside the if block for `this.shouldStartWatchStream()`.

However, I'm not sure this is a good use of our time. In the caller we're mostly channeling gRPC's response. If it happens to close a stream without supplying an error do we really care? Is it really worth crashing our customers' apps over this?

I'm not arguing against assertions generally: the assertions in the remote store have saved our bacon repeatedly, but the good ones catch our bugs where we failed to anticipate a transition that was actually happening in the wild.

By contrast here it seems like this case of error being null is possible but harmless. If this assertion were to fire in the wild because gRPC was giving us a null error, would we do anything but remove this assertion?

Feel free to push back (especially if I'm not considering all the angles).

packages/firestore/src/remote/remote_store.ts
```diff
@@ -480,29 +476,30 @@ export class RemoteStore implements TargetMetadataProvider {
...
   /**
-   * Returns true if we can add to the write pipeline (i.e. it is not full and
-   * the network is enabled).
+   * Returns true if we can add to the write pipeline (i.e. it is not full).
    */
   private canAddToWritePipeline(): boolean {
-    return (
-      this.isNetworkEnabled() && this.writePipeline.length < MAX_PENDING_WRITES
-    );
+    return this.writePipeline.length < MAX_PENDING_WRITES;
```
wilhuff: `fillWritePipeline()` above is called by the syncEngine after a local write. Why would we want batches to be in the pipeline while the network is disabled?

mikelehen: TBH it just makes more sense to me.  I don't see what we gain by throwing away the write pipeline when the network is disabled and then re-filling it (with the exact same writes) once it's enabled again.  We may as well just always be filling the write pipeline and then it will just consistently contain the next 10 writes to send to the write stream.

This is similar to how `listenTargets` persists across the network being disabled / re-enabled.

The only time we actually want to clear the write pipeline is on a user change, and so I reworked the handleUserChange code to do that explicitly.

Does this explanation sway you?  We may just have different conceptual models of what the write pipeline represents...  I'm willing to be convinced to reconsider.

packages/firestore/test/unit/specs/spec_test_runner.ts
mikelehen: I hope so, but I'm not sure it will. Yesterday I fell into a quagmire of confusing errors that I think were due to us inconsistently queuing spec step stuff on the asyncQueue. :-/

packages/firestore/src/remote/remote_store.ts
mikelehen: So that is a potential option, though I would be nervous about relying on that...  And right now I actually have an assert that explicitly makes sure that shouldStartWatchStream() returns false if we did not receive an error (see my discussion with Gil about that), so if I remove the "this.networkEnabled = false;" line, we hit that assert.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -386,98 +395,86 @@ export abstract class PersistentStream<
...
+
+    this.stream = this.startRpc(token);
+    this.stream.onOpen(() => {
+      dispatchIfStateUnchanged(() => {
+        assert(
+          this.state === PersistentStreamState.Starting,
+          'Expected stream to be in state Starting, but was ' + this.state
+        );
+        this.updateState(PersistentStreamState.Open);
+        // Need to recreate our dispatcher since we changed the state.
```
mikelehen: This is the wart I mentioned...

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -435,49 +447,45 @@ export abstract class PersistentStream<
...
     );
     this.state = PersistentStreamState.Backoff;
 
     this.backoff.backoffAndRun(async () => {
-      if (this.state === PersistentStreamState.Stopped) {
-        // We should have canceled the backoff timer when the stream was
-        // closed, but just in case we make this a no-op.
-        return;
-      }
+      assert(
```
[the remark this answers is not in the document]
wilhuff: Interesting. I see what you mean re inherent safety.

On the other hand, I'd like to get us out of the business of having to make that kind of argument when reasoning about this code. I would prefer to just handle all the callbacks in a uniform manner so that we don't have to think about it.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -157,16 +169,18 @@ export abstract class PersistentStream<
   private inactivityTimerPromise: CancelablePromise<void> | null = null;
   private stream: Stream<SendType, ReceiveType> | null = null;
 
-  protected backoff: ExponentialBackoff;
+  /** A sentinel object used to ignore auth callbacks for prior attempts. */
+  private currentAuthAttempt: object = null;
```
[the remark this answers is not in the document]
wilhuff: I've addressed (1) elsewhere. I think what I was really asking for was a stream generation number, but read on.

Re (2) I think you're right: this is a mechanism for implementing centralized cancellation. We could instead force everything to be cancellation aware. We've partially done this on iOS by way of `FSTCallbackFilter`.

`dispatchIfStillActive` was always kind of a hack. It works because Javascript is so functional, but it's fundamentally a non-portable pattern. This came about back when we were seriously under the gun for time and I didn't push back on Jonny to stop and think about how to express this portably.

Functionally I think we have a few concerns to address in a centralized way:

  * we want to dispatch back onto the async queue
  * we need to be able to cancel the response if we don't need it anymore
  * cancelation needs to be checked as we're executing on the async queue

This is basically what `DelayedOperation` implements, but it does so in a non-reusable way. If we could share that between DelayedOperations and the auth callback we'd be golden on that front.

The stream callbacks could be treated the same way with something like `FSTCallbackFilter` here.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -68,22 +85,16 @@ enum PersistentStreamState {
...
    */
   Error,
 
   /**
    * An in-between state after an error where the stream is waiting before
-   * re-starting. After
-   * waiting is complete, the stream will try to open. While in this
-   * state isStarted() will return YES but isOpen will return false.
+   * re-starting. After waiting is complete, the stream will try to open.
+   * While in this state isStarted() will return true but isOpen will return
```
schmidt-sebastian: Nit: `While in this state, isStarted() will return true but isOpen() will return..`

I had to do a double take as I was reading this.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -386,98 +395,86 @@ export abstract class PersistentStream<
       'Must be in initial state to auth'
     );
 
-    this.state = PersistentStreamState.Auth;
+    this.updateState(PersistentStreamState.Starting);
+
+    const dispatchIfStateUnchanged = this.stateGuardedDispatcher();
+
+    // TODO(mikelehen): Just use dispatchIfStillActive, but see TODO below.
```
schmidt-sebastian: Method name is outdated in this comment.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -495,6 +492,44 @@ export abstract class PersistentStream<
...
+    this.stateSequenceNumber++;
+  }
+
+  /**
+   * Returns a "dispatcher" function that dispatches operations onto the
+   * AsyncQueue but only runs them if startSequenceNumber remains unchanged.
+   * This allows us to turn auth / stream callbacks into no-ops if the stream
+   * is closed / re-opened, etc.
+   */
+  private stateGuardedDispatcher(): (fn: () => Promise<void>) => void {
```
schmidt-sebastian: I think this would be cleaner if it:
- was called something along the lines of `dispatchIfStateUnchanged`
- you passed in the state sequence number

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: We need to make sure that the multi-tab code continues to throw away writes that were executed by other clients. This is simplified if we throw away the in-memory state and then re-fill the write pipeline.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: Ahh, yeah. That is better. I implemented this but called it closeCount (used with the now renamed `closeGuardedDispatcher`) because I think that's more precise (we increment it in close()) and understandable (what's a "stream generation"?)...  but if you don't like the naming, let me know.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -136,82 +152,66 @@ export class RemoteStore implements TargetMetadataProvider {
    * Starts up the remote store, creating streams, restoring state from
    * LocalStore, etc.
    */
-  start(): Promise<void> {
-    return this.enableNetwork();
-  }
+  async start(): Promise<void> {
+    // Load any saved stream token from persistent storage
+    this.writeStream.lastStreamToken = await this.localStore.getLastStreamToken();
```
schmidt-sebastian: We are doing this twice now. Once in`start()` and then right below in `enableNetwork()`.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -136,82 +152,66 @@ export class RemoteStore implements TargetMetadataProvider {
...
-          this.writePipeline.length +
-          ' pending writes'
-      );
-      // TODO(mikelehen): We only actually need to clear the write pipeline if
-      // this is being called as part of handleUserChange(). Consider reworking.
-      this.writePipeline = [];
+      if (this.writePipeline.length > 0) {
+        log.debug(
+          LOG_TAG,
+          'Stopping write stream with ' +
```
schmidt-sebastian: Nit: I'm guesstimating that this might fit on one line if you change it to use backticks:

```
`Stopping write stream with ${this.writePipeline.length} pending writes`
```

packages/firestore/src/remote/persistent_stream.ts
schmidt-sebastian: `closeGuardedDispatcher` still seems a little off to me. The name is very tied to its implementation, but at first glance it doesn't seem to convey what it is used for (it might also just be the choice of words).

I don't have a good suggestion for you, but even something like `getCloseGuardedDispatcher()` might clear up some of my confusion.

packages/firestore/src/remote/remote_store.ts
mikelehen: Oops, thanks! I meant to remove this one.

packages/firestore/src/remote/remote_store.ts
mikelehen: Nope! Nice try though.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: getCloseGuardedDispatcher() SGTM

PR #1043, opened 2018-07-24
Ensure persistence started

gsoltis:
This mostly fixes up the spec tests so that we can assert that `LocalStore` always gets an already-started `Persistence` instance.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -96,7 +96,10 @@ export class IndexedDbPersistence implements Persistence {
   static MAIN_DATABASE = 'main';
 
   private simpleDb: SimpleDb;
-  private started: boolean;
+  private hasStarted: boolean;
+  get started(): boolean {
+    return this.hasStarted;
+  }
   private dbName: string;
   private localStoragePrefix: string;
   private ownerId: string = this.generateOwnerId();
@@ -130,7 +133,6 @@ export class IndexedDbPersistence implements Persistence {
     }
 
     assert(!this.started, 'IndexedDbPersistence double-started!');
-    this.started = true;
 
     return SimpleDb.openOrCreate(this.dbName, SCHEMA_VERSION, createOrUpgradeDb)
       .then(db => {
@@ -140,12 +142,14 @@ export class IndexedDbPersistence implements Persistence {
       .then(() => {
         this.scheduleOwnerLeaseRefreshes();
         this.attachWindowUnloadHook();
+      }).then(() => {
+        this.hasStarted = true;
       });
   }
 
   shutdown(deleteData?: boolean): Promise<void> {
     assert(this.started, 'IndexedDbPersistence shutdown without start!');
-    this.started = false;
+    this.hasStarted = false;
     this.detachWindowUnloadHook();
     this.stopOwnerLeaseRefreshes();
     return this.releaseOwnerLease().then(() => {
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -96,10 +96,7 @@ export class IndexedDbPersistence implements Persistence {
   static MAIN_DATABASE = 'main';
 
   private simpleDb: SimpleDb;
-  private hasStarted: boolean;
-  get started(): boolean {
-    return this.hasStarted;
-  }
+  private _started = false;
   private dbName: string;
   private localStoragePrefix: string;
   private ownerId: string = this.generateOwnerId();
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -96,7 +96,10 @@ export class IndexedDbPersistence implements Persistence {
   static MAIN_DATABASE = 'main';
 
   private simpleDb: SimpleDb;
-  private started: boolean;
+  private hasStarted: boolean;
```
schmidt-sebastian: The general convention for exposing properties that are only readonly via the public API is to use an underscore. So this would be:

```
private _started: boolean;
get started(): boolean {
    return this._started;
}
```

Please also move the getter to where the other getters/methods are defined. You could also initialize _started with 'false'.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -140,12 +142,15 @@ export class IndexedDbPersistence implements Persistence {
       .then(() => {
         this.scheduleOwnerLeaseRefreshes();
         this.attachWindowUnloadHook();
+      })
+      .then(() => {
+        this.hasStarted = true;
```
schmidt-sebastian: This can be part of the previous `then` block.

But that being said, I don't this will work one I merge my multi-tab PR. In the new `scheduleOwnerLeaseRefreshes` I am using `started` to determine whether I should issue callbacks. We essentially need three states: not-started, starting, started. And since nothing else should be happening while we call `.start()`, it should be safe to keep flipping the started bit in the beginning?

packages/firestore/src/local/local_store.ts
```diff
@@ -163,6 +163,10 @@ export class LocalStore {
      */
     private garbageCollector: GarbageCollector
   ) {
+    assert(
+      persistence.started,
+      'Local Store passed an unstarted persistence implementation'
```
schmidt-sebastian: The LocalStore is the receiver here, which doesn't quite come across in this message (maybe it is just missing a `was` as in `was passed `)

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -45,18 +45,21 @@ export class MemoryPersistence implements Persistence {
   private remoteDocumentCache = new MemoryRemoteDocumentCache();
   private queryCache = new MemoryQueryCache();
 
-  private started = false;
+  private hasStarted = false;
```
schmidt-sebastian: Same comment about naming.

packages/firestore/src/local/persistence.ts
```diff
@@ -75,6 +75,11 @@ export interface Persistence {
    */
   start(): Promise<void>;
 
+  /**
```
schmidt-sebastian: Nit: I would move the property declarations to the top.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -356,26 +356,29 @@ abstract class TestRunner {
     this.serializer = new JsonProtoSerializer(this.databaseInfo.databaseId, {
       useProto3Json: true
     });
-    this.persistence = this.getPersistence(this.serializer);
+    this.persistence = new Promise(resolve => {
+      const p = this.getPersistence(this.serializer);
+      p.start().then(() => resolve(p));
+    });
```
schmidt-sebastian: Hmm, this is somewhat strange. Can we not start the persistence layer in the renamed `start()`? We can check there whether it is already started using your new nifty API.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -417,16 +421,9 @@ abstract class TestRunner {
...
-    this.connection.reset();
-    await this.persistence.start();
-    await this.localStore.start();
-    await this.remoteStore.start();
-  }
-
   async shutdown(): Promise<void> {
     await this.remoteStore.shutdown();
-    await this.persistence.shutdown(/* deleteData= */ true);
+    await (await this.persistence).shutdown(/* deleteData= */ true);
```
schmidt-sebastian: Please try to find a way to make this a non-Promise again.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: As discussed, leaving this at the end in its own `.then()` call.

packages/firestore/test/unit/specs/spec_test_runner.ts
gsoltis: We technically could, but `doRestart()` calls `start()`, and `doRestart()` intentionally does not want to start the persistence layer. As such, I think it makes more sense to keep the call to `persistence.start()` in the one place where we know we want to start it, rather than gate it with an `if (!persistence.started) await persistence.start()`, which then means you have to know why it would or wouldn't already be started.

packages/firestore/src/local/local_store.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_persistence.ts

packages/firestore/src/local/memory_persistence.ts
gsoltis: Done.

packages/firestore/src/local/persistence.ts
gsoltis: Done.

packages/firestore/test/unit/specs/spec_test_runner.ts
gsoltis: Restructured to have an `init` and a `start`.

PR #1047, opened 2018-07-25
Add a type parameter to Persistence

gsoltis:
Adds a type parameter to `Persistence` to specify the transaction type. Threads the parameter through where necessary. The end result is being able to remove some casts and adds the ability to add persistence-layer-specific fields to transactions in a type-safe manner.

No behavior changes are intended.


packages/firestore/src/local/memory_persistence.ts
```diff
@@ -33,15 +33,18 @@ const LOG_TAG = 'MemoryPersistence';
  * A memory-backed instance of Persistence. Data is stored only in RAM and
  * not persisted across sessions.
  */
-export class MemoryPersistence implements Persistence {
+export class MemoryPersistence
+  implements Persistence<MemoryPersistenceTransaction> {
```
schmidt-sebastian: I would rename `MemoryPersistenceTransaction` to `MemoryTransaction`.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -72,22 +75,27 @@ export class MemoryPersistence implements Persistence {
...
   }
 }
 
-/** Dummy class since memory persistence doesn't actually use transactions. */
-class MemoryPersistenceTransaction implements PersistenceTransaction {}
+/**
+ * Memory persistence is not actually transactional, but future implementations
+ * may have transaction-scoped state.
+ */
+export class MemoryPersistenceTransaction implements PersistenceTransaction {}
```
schmidt-sebastian: I would rename `MemoryPersistenceTransaction` to `MemoryTransaction`, which would match `IndexedDbTransaction`. Otherwise, maybe `MemoryPersistedTransaction`?

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -42,17 +47,8 @@ import * as persistenceHelpers from './persistence_test_helpers';
...
-
 describe('MemoryMutationQueue', () => {
-  beforeEach(() => {
-    return persistenceHelpers.testMemoryPersistence().then(p => {
-      persistence = p;
-    });
-  });
-
-  genericMutationQueueTests();
+  genericMutationQueueTests(persistenceHelpers.testMemoryPersistence);
```
schmidt-sebastian: We seem to be doing this in `local_store.test.ts` as well, but this might bite us later as this loses the `this` context. I would suggest you use an array function here.

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -61,13 +57,14 @@ describe('IndexedDbMutationQueue', () => {
...
-    });
+  let persistencePromise: Promise<Persistence<IndexedDbTransaction>>;
+  let persistence: Persistence<IndexedDbTransaction>;
+  beforeEach(async () => {
+    persistencePromise = persistenceHelpers.testIndexedDbPersistence();
+    persistence = await persistencePromise;
   });
 
-  genericMutationQueueTests();
+  genericMutationQueueTests(() => persistencePromise);
```
schmidt-sebastian: Is it possible to pass `persistence` here? Then you don't have to deal with `persistencePromise` anywhere. Alternatively, you could do:

`return persistencePromise.then(p => genericMutationQueueTests(p))`

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -35,17 +41,8 @@ import * as persistenceHelpers from './persistence_test_helpers';
...
-
 describe('MemoryQueryCache', () => {
-  beforeEach(() => {
-    return persistenceHelpers.testMemoryPersistence().then(p => {
-      persistence = p;
-    });
-  });
-
-  genericQueryCacheTests();
+  genericQueryCacheTests(persistenceHelpers.testMemoryPersistence);
```
schmidt-sebastian: Same comment as above regarding the arrow function.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -54,22 +51,22 @@ describe('IndexedDbQueryCache', () => {
...
-    });
+  let persistencePromise: Promise<Persistence<IndexedDbTransaction>>;
+  beforeEach(async () => {
+    persistencePromise = persistenceHelpers.testIndexedDbPersistence();
   });
 
-  afterEach(() => persistence.shutdown(/* deleteData= */ true));
-
-  genericQueryCacheTests();
+  genericQueryCacheTests(() => persistencePromise);
```
schmidt-sebastian: The previous implementation used a different `persistence` object for every test case. Can you verify that this is still the case?

packages/firestore/test/unit/local/mutation_queue.test.ts
gsoltis: I can change it, but `persistenceHelpers` is a module, not an instance of a class. This is a reference to a function, not a method. For this to bite us, we would need to change the type of `persistenceHelpers` to something where `this` would matter. Replacing a module with a class would I think be a significant enough change to warrant auditing the places where it was imported and catch things like this.

packages/firestore/src/local/memory_persistence.ts
gsoltis: Done. `MemoryTransaction`.

packages/firestore/test/unit/local/mutation_queue.test.ts
gsoltis: This unfortunately does not work, as it will break if another test uses `only`. The reason is that `describe` will always run, creating the persistence layer, but the `beforeEach` and `afterEach` methods will not, so it won't be properly cleaned up. The key is to only actually create the persistence layer in a `beforeEach` method.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Yes, the promise is recreated each time `beforeEach` runs, so each test run gets its own `persistence` instance.

packages/firestore/test/unit/local/mutation_queue.test.ts
gsoltis: Actually, I believe this will break even without another test using `only`. At the time that you would return this, `persistencePromise` is `undefined`. `beforeEach` needs to run before you can use `persistencePromise`.

mikelehen:
Uhhh.....   This seems okay, but honestly, I'm not a huge fan of it.  It adds a bit of compile-time safety but at the expense of code churn and making a bunch of our types generic (including LocalStore, LocalDocumentsView, etc. which don't really make sense as generic types and aren't generic on the other platforms).

A better option forward (for simplicity and platform consistency) might be to just drop the explicit transaction passing on web and make the current transaction implicit like it is on Android / iOS.  At the time we decided to make it explicit, iOS didn't have transactions and we hadn't tackled Android persistence yet.  So I think there's an argument to be made that web could be switched to implicit now.  But that's also a large refactoring with lots of code churn and I'm not sure there's anything specific to motivate doing that right now.

That said, I'm not going to object to merging this if the consensus is that it improves the state of affairs.  @wilhuff Any interest in weighing in? :)

mikelehen:
Fascinating!  What is a transaction-specific sequence number? :)

wilhuff:
In the iOS port, the assertion that you're currently in a transaction is [made centrally](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/Source/Local/FSTLevelDB.mm#L214) so I don't think there's any cognitive overhead to making the transaction implicit.

Given that we will only ever use one transaction at a time in the local store implicit vs explicit doesn't matter, so all things considered I'd rather change one port to conform to the other two than change two to conform to one. (To be specific, I prefer implicit.)

gsoltis:
We assign the transaction number to `QueryData` updates, as well as ack'd mutations. This lets us assess how recently used documents are. Here's the assert in `FSTLevelDB`: https://github.com/firebase/firebase-ios-sdk/blob/lru/Firestore/Source/Local/FSTLevelDB.mm#L118. After much back and forth, @wilhuff and I settled on reaching to the persistence layer and asserting that you're in a transaction as the least bad way of doing this on iOS. Explicit transactions are a more natural place to include this value, as they are valid and constant for the lifetime of a transaction.

My next branch after this one for the webclient adds `readonly currentSequenceNumber` as a field on `PersistenceTransaction`.

gsoltis:
I think implicit leads to a lot of hidden gotchas on what methods must be called in a transaction. Having an explicit transaction makes it much clearer. My opinion on this is based on having had to update the test suite to make sure that transactions existed whenever these methods that accessed the implicit transaction. It was painful, and in a lot of cases it is not clear why a transaction needs to exist other than if you run it, you will get a runtime error if you don't have one.

wilhuff:
I'm okay with explicit transactions too--I have no real preference. I'm coming at it from the point of conserving work.

gsoltis:
I'm not proposing that we move to explicit now, just that we don't remove the one place where it's already done.

gsoltis:
Summary of offline discussion:
 * Remove generic parameter on `Persistence`
 * Keep distinction between `SimpleDbTransaction` and `IndexedDbTransaction`.
   * `SimpleDbTransaction` will be a field on `IndexedDbTransaction` and will be internal to `IndexedDbPersistence` and `SimpleDb`.
 * Make `PersistenceTransaction` an `abstract class` rather than an `interface` to avoid the problem of overmatching.
 * `sequenceNumber`, when added in a subsequent PR, will be a property of `PersistenceTransaction`.

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -22,10 +22,9 @@ import {
 } from '../model/collections';
 import { Document, MaybeDocument } from '../model/document';
 import { DocumentKey } from '../model/document_key';
-
-import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { RemoteDocumentCache } from './remote_document_cache';
+import { PersistenceTransaction } from './persistence';
```
mikelehen: Hrm. Does prettier not order these? (I'm not sure why this moved here)

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -270,7 +273,7 @@ describe('LocalStore w/ IndexedDB Persistence', () => {
   genericLocalStoreTests(persistenceHelpers.testIndexedDbPersistence);
 });
 
-function genericLocalStoreTests(
+function genericLocalStoreTests<TransactionType extends PersistenceTransaction>(
```
mikelehen: obsolete?

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -61,31 +57,32 @@ describe('IndexedDbMutationQueue', () => {
...
   describe('loadNextBatchIdFromDb', () => {
     function loadNextBatchId(): Promise<BatchId> {
       return persistence.runTransaction('loadNextBatchIdFromDb', txn => {
-        return IndexedDbMutationQueue.loadNextBatchIdFromDb(txn).next(
-          batchId => {
-            return batchId;
-          }
-        );
+        return IndexedDbMutationQueue.loadNextBatchIdFromDb(
+          txn as IndexedDbTransaction
```
mikelehen: obsolete?

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -61,31 +57,32 @@ describe('IndexedDbMutationQueue', () => {
...
       });
     }
 
     function addDummyBatch(userId: string, batchId: BatchId): Promise<void> {
       return persistence.runTransaction('addDummyBatch', transaction => {
-        const txn = transaction as SimpleDbTransaction;
-        const store = txn.store<[string, number], DbMutationBatch>(
-          DbMutationBatch.store
-        );
+        const store = (transaction as IndexedDbTransaction).simpleDbTransaction.store<
```
mikelehen: use helper?

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -120,10 +117,15 @@ describe('IndexedDbMutationQueue', () => {
  * Defines the set of tests to run against both mutation queue
  * implementations.
  */
-function genericMutationQueueTests(): void {
+function genericMutationQueueTests<
+  TransactionType extends PersistenceTransaction
```
mikelehen: obsolete

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -98,11 +89,17 @@ function genericQueryCacheTests(): void {
...
 
+  let persistence: Persistence;
   beforeEach(async () => {
+    persistence = await persistencePromise();
     cache = new TestQueryCache(persistence, persistence.getQueryCache());
     await cache.start();
   });
 
+  afterEach(async () => {
+    persistence.shutdown(/* deletaData */ true);
```
mikelehen: `deleteData=`

packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -43,28 +37,25 @@ describe('IndexedDbRemoteDocumentCache', () => {
...
+  genericRemoteDocumentCacheTests(persistenceHelpers.testIndexedDbPersistence);
 });
 
 /**
  * Defines the set of tests to run against both remote document cache
  * implementations.
  */
-function genericRemoteDocumentCacheTests(): void {
+function genericRemoteDocumentCacheTests<
+  TransactionType extends PersistenceTransaction
```
mikelehen: obsolete

packages/firestore/test/unit/local/mutation_queue.test.ts
mikelehen: Is this refactoring related to the TransactionType generic?  If so, is it obsolete now?  It seems fine, but just wanting to make sure it's not just leftover churn.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -186,10 +192,17 @@ export class IndexedDbPersistence implements Persistence {
...
-      // Verify that we still have the owner lease as part of every transaction.
-      return this.ensureOwnerLease(txn).next(() => operation(txn));
-    });
+    return this.simpleDb.runTransaction(
+      'readwrite',
+      ALL_STORES,
+      simpleDbTxn => {
+        // Verify that we still have the owner lease as part of every transaction.
+        const txn = new IndexedDbTransaction(simpleDbTxn);
+        return this.ensureOwnerLease(txn.simpleDbTransaction).next(() =>
```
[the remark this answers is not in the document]
gsoltis: If it's all the same, I'd rather keep the longer name. I think the abbreviation is a little harder to read.

packages/firestore/src/local/memory_remote_document_cache.ts
gsoltis: I think I asked vscode to organize imports. And I guess prettier didn't object either way? I'll put it back and see what happens.

packages/firestore/src/local/indexeddb_persistence.ts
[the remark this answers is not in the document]
mikelehen: I was suggesting you use the existing variable that's in scope.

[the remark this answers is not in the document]
gsoltis: Ah. That makes more sense. Done.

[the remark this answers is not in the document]
mikelehen: As a general rule if I ever ask somebody to abbreviate a name in code review feedback you can assume my account has been hacked. 😁

packages/firestore/test/unit/local/local_store.test.ts
gsoltis: Fixed.

packages/firestore/test/unit/local/mutation_queue.test.ts
gsoltis: I think leftover churn. I'm removing it.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -297,6 +298,8 @@ export class IndexedDbQueryCache implements QueryCache {
     this.garbageCollector = gc;
   }
 
+  // TODO(gsoltis): we can let the compiler assert that txn !== null if we
+  // drop null from the type bounds on txn.
```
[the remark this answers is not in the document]
gsoltis: Well, kinda. It compiles, but only because we don't enforce strict function types. This is replicating the type signature from `GarbageSource`, which allows `null`. So actually removing it from this function gets us no safety and a more obscure runtime error than what you would currently get. 

The solution will eventually arrive when we get rid of `GarbageSource`.

gsoltis:
Ok, I think I have addressed everything, including adding a helper. I also left in some of the test refactor, since I think it's a little cleaner this way.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -98,11 +89,17 @@ function genericQueryCacheTests(): void {
...
 
+  let persistence: Persistence;
   beforeEach(async () => {
+    persistence = await persistencePromise();
     cache = new TestQueryCache(persistence, persistence.getQueryCache());
     await cache.start();
   });
 
+  afterEach(async () => {
+    persistence.shutdown(/* deletaData= */ true);
```
mikelehen: Actually, since it looks like you're going to need to merge master and let travis re-run anyway, I'll point out that you missed the second typo in this line.  delet_e_Data

gsoltis: This was apparently a load-bearing typo. CI has been failing since I fixed it...

PR #1051, opened 2018-07-26
[Multi-Tab] Don't add change log entry when there are no changes

schmidt-sebastian:
FriendlyEats writes a bunch of empty change events to the document change log. Let's not do that.

### DIFF
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -69,18 +69,18 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       let changedKeys = documentKeySet();
       for (const maybeDocument of maybeDocuments) {
         promises.push(
-            documentStore.put(
-                dbKey(maybeDocument.key),
-                this.serializer.toDbRemoteDocument(maybeDocument)
-            )
+          documentStore.put(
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -63,24 +63,27 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
     maybeDocuments: MaybeDocument[]
   ): PersistencePromise<void> {
     const promises: Array<PersistencePromise<void>> = [];
-    const documentStore = remoteDocumentsStore(transaction);
 
-    let changedKeys = documentKeySet();
-    for (const maybeDocument of maybeDocuments) {
+    if (maybeDocuments.length > 0) {
+      const documentStore = remoteDocumentsStore(transaction);
```
mikelehen: This seems fine...  but why are we calling addEntries() with an empty list anyway?  Should we be filtering this out at a higher-level (potentially preventing _even more_ unnecessary work?

schmidt-sebastian: I was pondering about doing this one layer up, but ended up picking the `remote_document_cache` for no particular reason (other than this would solve the same problem no matter who calls `addEntries`). Leaving as is.

PR #1055, opened 2018-07-27
Port optimizations to LocalDocumentsView from iOS

var-const:
* add a method to find batches affecting a set of keys (port of [1479](https://github.com/firebase/firebase-ios-sdk/pull/1479));
* use the newly-added method to avoid rereading batches when getting documents in LocalDocumentsView (port of [1505](https://github.com/firebase/firebase-ios-sdk/pull/1505));
* avoid rereading batches when searching for documents in a collection (port of [1533](https://github.com/firebase/firebase-ios-sdk/pull/1533)).

### DIFF
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -41,7 +41,8 @@ import { LocalSerializer } from './local_serializer';
 import { MutationQueue } from './mutation_queue';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
-import { SimpleDb, SimpleDbStore } from './simple_db';
+import { SimpleDbStore } from './simple_db';
+import { IndexedDbPersistence } from './indexeddb_persistence';
 
 /** A mutation queue for a specific user, backed by IndexedDB. */
 export class IndexedDbMutationQueue implements MutationQueue {
@@ -613,7 +614,7 @@ function convertStreamToken(token: ProtoByteString): string {
 function mutationsStore(
   txn: PersistenceTransaction
 ): SimpleDbStore<DbMutationBatchKey, DbMutationBatch> {
-  return SimpleDb.getStore<DbMutationBatchKey, DbMutationBatch>(
+  return IndexedDbPersistence.getStore<DbMutationBatchKey, DbMutationBatch>(
     txn,
     DbMutationBatch.store
   );
@@ -625,10 +626,10 @@ function mutationsStore(
 function documentMutationsStore(
   txn: PersistenceTransaction
 ): SimpleDbStore<DbDocumentMutationKey, DbDocumentMutation> {
-  return SimpleDb.getStore<DbDocumentMutationKey, DbDocumentMutation>(
-    txn,
-    DbDocumentMutation.store
-  );
+  return IndexedDbPersistence.getStore<
+    DbDocumentMutationKey,
+    DbDocumentMutation
+  >(txn, DbDocumentMutation.store);
 }
 
 /**
@@ -637,7 +638,7 @@ function documentMutationsStore(
 function mutationQueuesStore(
   txn: PersistenceTransaction
 ): SimpleDbStore<DbMutationQueueKey, DbMutationQueue> {
-  return SimpleDb.getStore<DbMutationQueueKey, DbMutationQueue>(
+  return IndexedDbPersistence.getStore<DbMutationQueueKey, DbMutationQueue>(
     txn,
     DbMutationQueue.store
   );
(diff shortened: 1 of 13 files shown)


packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -362,6 +363,41 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+        ) {
+          control.done();
+          return;
+        }
+
+        if (!documentKeys.has(new DocumentKey(path))) {
+          return;
+        }
+        uniqueBatchIDs = uniqueBatchIDs.add(batchID);
+      })
```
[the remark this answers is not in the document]
var-const: PTAL.

packages/firestore/src/local/memory_mutation_queue.ts
```diff
@@ -252,6 +253,26 @@ export class MemoryMutationQueue implements MutationQueue {
...
+    }
+    const start = new DocReference(documentKeys.first(), 0);
+    const end = new DocReference(documentKeys.last(), Number.POSITIVE_INFINITY);
+    let uniqueBatchIDs = new SortedSet<number>(primitiveComparator);
+
+    this.batchesByDocumentKey.forEachInRange([start, end], ref => {
+      if (documentKeys.has(ref.key)) {
+        uniqueBatchIDs = uniqueBatchIDs.add(ref.targetOrBatchId);
+      }
+    });
```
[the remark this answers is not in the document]
var-const: PTAL.

packages/firestore/src/local/memory_mutation_queue.ts
```diff
@@ -257,17 +257,18 @@ export class MemoryMutationQueue implements MutationQueue {
...
     let uniqueBatchIDs = new SortedSet<number>(primitiveComparator);
 
-    this.batchesByDocumentKey.forEachInRange([start, end], ref => {
-      if (documentKeys.has(ref.key)) {
+    documentKeys.forEach(documentKey => {
+      const start = new DocReference(documentKey, 0);
+      const end = new DocReference(documentKey, Number.POSITIVE_INFINITY);
+      this.batchesByDocumentKey.forEachInRange([start, end], ref => {
+        if (!documentKey.isEqual(ref.key)) {
+          return;
```
mikelehen: nit: Is it possible to hit this?  Seems like our range should constrain us to `documentKey`.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -56,11 +56,23 @@ export class LocalDocumentsView {
...
-      .next(remoteDoc => {
-        return this.computeLocalDocument(transaction, key, remoteDoc);
-      });
+    return this.mutationQueue
+      .getAllMutationBatchesAffectingDocumentKey(transaction, key)
+      .next(batches => this.getDocumentInBatches(transaction, key, batches));
+  }
+
+  /** Internal version of `getDocument` that allows reusing batches. */
+  private getDocumentInBatches(
```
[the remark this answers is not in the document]
mikelehen: No preference.   You're right that it would ideally be private... but TypeScript doesn't support "overloads" with mixed visibility. So I think `getDocumentInternal()` is good.

packages/firestore/src/local/memory_mutation_queue.ts
var-const: I'm not sure. This function is derived from `getAllMutationBatchesAffectingDocumentKey`, which [asserts](https://github.com/firebase/firebase-js-sdk/blob/87e0b06bdf5fdd78ff57fe2b7e47207b33542bdc/packages/firestore/src/local/memory_mutation_queue.ts#L241-L244) this equality, this led me to believe this case is not impossible. I can remove it altogether or replace with an assertion.

mikelehen: Asserts should generally mean the case is impossible.  They exist to document our expectations and/or catch false expectations as early as possible to make finding / fixing bugs later. So I think an assert here makes perfect sense.

Fine print: We're probably not 100% rigorous about asserts always being impossible.  E.g. there may be cases where we use asserts for cases that /are/ possible but represent rare error conditions that we don't feel compelled to provide proper error handling for.

mikelehen:
Nice!  Thanks for the performance data.

I can't 100% explain the "first time slow" issue without digging in, but it may be related to the "colRef.onSnapshot(() => {});" line...  That's going to spin up a full normal listener against the backend which involves reading documents and query metadata from persistence, sending the result to the backend, and handling the response data.

To cancel that out, you could make sure the listener has fully initialized / settled by doing something like:

```
  ...
  let deferred = new Deferred();
  colRef.onSnapshot(snap => if (!snap.metadata.fromCache) deferred.resolve());
  return deferred.promise;
}).then(() => {
  writeStartTime = Date.now();
  ...
```

packages/firestore/src/local/memory_mutation_queue.ts
var-const: Replicated the assertion.

packages/firestore/test/unit/local/mutation_queue.test.ts
```diff
@@ -433,7 +434,30 @@ function genericMutationQueueTests(): void {
     const matches = await mutationQueue.getAllMutationBatchesAffectingDocumentKey(
       key('foo/bar')
     );
-    expect(matches.length).to.deep.equal(expected.length);
```
var-const: Forgot to mention: I noticed that `expectEqualArrays` checks for equal length, so this line seems redundant.

PR #1056, opened 2018-07-27
[Multi-Tab] Don't write document change log when multi-tab is disabled

schmidt-sebastian:
This changes the remote document cache to no longer keep a document change log when multi-tab is off.


schmidt-sebastian:
Note that the performance differential between master and multitab for the spec perf tests is 1106 ms without this branch and 572 ms with it.

PR #1058, opened 2018-07-27
Firestore: in tests, silence the verbose warning about old timestamp behavior

var-const:
This cleans up the few leftover places where the warning was still being outputted in tests. Running locally, I get a clean log.

### DIFF
--- a/packages/firestore/test/util/api_helpers.ts
+++ b/packages/firestore/test/util/api_helpers.ts
@@ -46,15 +46,16 @@ export const FIRESTORE = new Firestore({
 });
 
 export function firestore(): Firestore {
+  FIRESTORE.settings({timestampsInSnapshots: true});
   return FIRESTORE;
 }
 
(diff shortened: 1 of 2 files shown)


packages/firestore/test/util/api_helpers.ts
```diff
@@ -46,15 +46,16 @@ export const FIRESTORE = new Firestore({
 });
 
 export function firestore(): Firestore {
+  FIRESTORE.settings({timestampsInSnapshots: true});
```
var-const: Not sure about this change. Can/should I modify `FIRESTORE` right after it is defined in the global scope?

mikelehen: I think it would be fine to move this out of the firestore() function, if that's what you're asking.  I don't have a strong preference.

var-const: Ok, I guess I'll merge as is, then.

PR #1063, opened 2018-07-30
Don't update query target metadata for updates

schmidt-sebastian:
This PR ports some of what I think I learned by reviewing Greg's latest Android PR to the Web. This save an additional read on every query update in Multi-Tab.

Greg - is this correct?

### DIFF
--- a/packages/firestore/src/local/indexeddb_query_cache.ts
+++ b/packages/firestore/src/local/indexeddb_query_cache.ts
@@ -87,12 +87,16 @@ export class IndexedDbQueryCache implements QueryCache {
     });
   }
 
-  setLastRemoteSnapshotVersion(
-    transaction: PersistenceTransaction,
-    snapshotVersion: SnapshotVersion
+  setTargetsMetadata(
+      transaction: PersistenceTransaction,
+      highestListenSequenceNumber: number,
+      lastRemoteSnapshotVersion?: SnapshotVersion
   ): PersistencePromise<void> {
     return this.retrieveMetadata(transaction).next(metadata => {
-      metadata.lastRemoteSnapshotVersion = snapshotVersion.toTimestamp();
+      metadata.highestListenSequenceNumber = highestListenSequenceNumber;
+      if (lastRemoteSnapshotVersion) {
+        metadata.lastRemoteSnapshotVersion = lastRemoteSnapshotVersion.toTimestamp();
+      }
       return this.saveMetadata(transaction, metadata);
     });
   }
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -114,15 +114,7 @@ export class IndexedDbQueryCache implements QueryCache {
...
-    return this.saveQueryData(transaction, queryData).next(() => {
-      return this.retrieveMetadata(transaction).next(metadata => {
-        if (this.updateMetadataFromQueryData(queryData, metadata)) {
-          return this.saveMetadata(transaction, metadata);
-        } else {
-          return PersistencePromise.resolve();
-        }
-      });
-    });
+    return this.saveQueryData(transaction, queryData);
```
gsoltis: You will still potentially need to update the metadata. Once `QueryCache` starts tracking the highest sequence number, it is likely that an update will trigger a metadata write.

schmidt-sebastian: I updated the PR to allow future Greg to set the sequence number as part of the `setLastRemoteSnapshotVersion` call. Let me know if that will work as is for GC.

I am also still accepting nominations for the API name voting challenge.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -87,12 +87,16 @@ export class IndexedDbQueryCache implements QueryCache {
     });
   }
 
-  setLastRemoteSnapshotVersion(
+  setTargetsMetadata(
     transaction: PersistenceTransaction,
-    snapshotVersion: SnapshotVersion
+    highestListenSequenceNumber: number,
+    lastRemoteSnapshotVersion?: SnapshotVersion
```
gsoltis: Why is this optional? Do we ever not supply it?

schmidt-sebastian: Local Store doesn't always call setLastRemoteSnapshotVersion. Do you still want to update the listen sequence number?

See https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/local_store.ts#L524

gsoltis: I see. I do still want to update the sequence number. I guess optional is ok, although it looks like in that case we could also pass in the highest snapshot version and have it be a no-op. I don't feel too strongly either way though.

schmidt-sebastian: We can refine the API when we start porting to Web, but I believe the highest snapshot version that you know of could be SnapshotVersion.MIN (unless you pass the version that you just got from getLastRemoteSnapshotVersion).

gsoltis: Yeah, I meant the one you just read. But yeah, we can update later if need be.

PR #1065, opened 2018-07-30
Don't look up TargetId in notifyLocalViewChanges

schmidt-sebastian:
This PR removes a QueryData read from `notifyLocalViewChanges`. This read happened for every target update and was merely used to look up the target ID.


packages/firestore/src/local/local_store.ts
```diff
@@ -553,29 +553,14 @@ export class LocalStore {
...
-                view.removedKeys,
-                targetId
-              );
-            })
-        );
-      }
-      return PersistencePromise.waitFor(promises);
-    });
+  notifyLocalViewChanges(viewChanges: LocalViewChanges[]): void {
+    for (const view of viewChanges) {
```
mikelehen: Since you've rewritten most of this loop, maybe go ahead and rename 'view' to 'viewChange' for sanity?

packages/firestore/src/local/local_view_changes.ts
```diff
@@ -26,11 +27,15 @@ import { documentKeySet, DocumentKeySet } from '../model/collections';
 export class LocalViewChanges {
   constructor(
     readonly query: Query,
```
mikelehen: I think query is unused now.  Remove it?

PR #1067, opened 2018-07-30
Adding Spec Test for Target Removal

schmidt-sebastian:
Spec test for:

  * Client adds a target
  * Client removes a target (without waiting for the response)
  * Watch fails the add target (perhaps with permission denied)
  * Watch now sees a remove target for a target not currently being listened to

### DIFF
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -141,17 +141,82 @@ describeSpec('Listens:', [], () => {
     }
   );
 
-  specTest("Doesn't raise events for non-active target", [], () => {
+  specTest('Will recover from out-of-band error', [], () => {
     const query = Query.atPath(path('collection'));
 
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -141,6 +141,19 @@ describeSpec('Listens:', [], () => {
     }
   );
 
+  specTest("Doesn't raise events for non-active target", [], () => {
```
wilhuff: The change that's coming is the sequence where userListens, userUnlistens, watchRemoves(with error),  watchRemoves(no error).

As specified this test is non-representative of the real system because Watch should respond to every request we make (or fail the stream).

schmidt-sebastian:
I have added a bunch of new test cases and fixed some of the error handling, by moving some of it into WatchChangeAggregator. I will pause work on this PR until b/112051702 is resolved.

PR #1069, opened 2018-07-31
Merging Persistent Stream refactor

schmidt-sebastian:
This "merges" the Persistent Stream refactor (#1041) into the multi-tab branch.


packages/firestore/src/remote/remote_store.ts
```diff
@@ -111,19 +108,25 @@ export class RemoteStore implements TargetMetadataProvider {
...
+
+  /**
+   * Indicates whether the network can accept requests (as determined by both
+   * the `isPrimary` flag and the user specified `networkEnabled` flag). If
+   * the connection is already established, new requests will be sent over the
+   * existing stream. If the stream has not yet been established, it will be
+   * established if there are outstanding requests.
    */
-  private networkAllowed = true;
+  private canUseNetwork = false;
```
mikelehen: Hrm.... My hope would be that we don't need to track 3 booleans and instead canUseNetwork would instead be a function of networkEnabled and isPrimary.  As-is I'm not sure what they're relation is.

packages/firestore/test/unit/specs/write_spec.test.ts
```diff
@@ -1245,4 +1245,30 @@ describeSpec('Writes:', [], () => {
       );
     }
   );
+
+  specTest(
+    'Mutation are not sent twice after primary failover',
```
mikelehen: Wow, did you really write a new test to test the merge of this?  Or where did this come from?

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: We need a boolean to indicate whether the network connection is established, so that we can shut it down when needed in `disableNetwork()`. At first, this was called `networkEnabledInternal`, but I then realizes that I need the same state in the `canAddToWritePipeline` and in the asserts that verify that a stream didn't close unexpectedly. 

If I want to use a separate `canUseNetwork()` method that uses both isPrimary and networkEnabled, then I need to flip networkEnabled to false during shutdown. Otherwise, I am hitting a bunch of asserts in the teardown.

packages/firestore/test/unit/specs/write_spec.test.ts
schmidt-sebastian: Yeah, I wanted to verify that we clear the write pipeline during failover.

schmidt-sebastian:
I merged in your changes. Let me know if the rest of this PR is good to merge.

PR #1071, opened 2018-07-31
Add a CHANGELOG entry for #1052

wilhuff:


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 # Unreleased
+- [changed] Improved how Firestore handles idle queries to reduce the cost of
+  re-listening within 30 minutes.
```
mikelehen: Do we want to go ahead and mention @var-const change as well?  Something like:

[changed] Improved offline performance with many outstanding writes.

wilhuff: Done.

hiepxanh:
I have two question.
1) what do your mean about "Improved offline performance with many outstanding writes." ? 
2) When user in offline mode, is that firebase using batched writes?. If so, that will cause a lot of action. Can I do some kind of action that only do 1 write query or this fixed is that?

wilhuff:
Regarding (1) the specific change @mikelehen is referencing and that I added to the changelog is #1055. The issue was that previously writes that were waiting to be sent were not handled efficiently: there were lots of lookups and batches could be re-read multiple times while replaying these changes to construct what the documents should look like. The #1055 change simplifies this code and loads the outstanding writes in a single pass.

Regarding (2) Firestore does not rearrange your changes. If you perform your writes as individual writes, they are applied as individual writes. If you perform your writes in batches they are applied in batches. What we changed was the performance of the way the client deals with your writes but does not change how your writes are sent to the server.

hiepxanh:
thank you for your answer. That very helpful to me

PR #1072, opened 2018-07-31
Ensure owner lease in owner lease refresh loop

gsoltis:
Add back call to `ensureOwnerLease`.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -353,10 +353,10 @@ export class IndexedDbPersistence implements Persistence {
         'readwrite',
         ALL_STORES,
         txn => {
-          // NOTE: We don't need to validate the current owner contents, since
-          // runTransaction does that automatically.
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -353,10 +353,10 @@ export class IndexedDbPersistence implements Persistence {
         'readwrite',
         ALL_STORES,
         txn => {
-          // NOTE: We don't need to validate the current owner contents, since
-          // runTransaction does that automatically.
-          const store = txn.store<DbOwnerKey, DbOwner>(DbOwner.store);
-          return store.put('owner', new DbOwner(this.ownerId, Date.now()));
+          this.ensureOwnerLease(txn).next(() => {
+            const store = txn.store<DbOwnerKey, DbOwner>(DbOwner.store);
```
schmidt-sebastian: This should be `return this.ensureOwnerLease(...`.

gsoltis: Oh yes, it's an easy change, I'll just let the CI run while I commute home...

Ok it's fixed now.

PR #1073, opened 2018-08-01
Merge

schmidt-sebastian:
This merges the following into multi-tab:

https://github.com/firebase/firebase-js-sdk/pull/1047
https://github.com/firebase/firebase-js-sdk/pull/1052
https://github.com/firebase/firebase-js-sdk/pull/1055


schmidt-sebastian:
@var-const @gsoltis  @wilhuff  Can you check that I didn't screw this up? This is a PR against the automatic merge into firestore-multi-tab.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -494,14 +485,10 @@ export class IndexedDbPersistence implements Persistence {
 
   runTransaction<T>(
     action: string,
-<<<<<<< HEAD
     requirePrimaryLease: boolean,
     transactionOperation: (
-      transaction: PersistenceTransaction
+      transaction: IndexedDbTransaction
```
gsoltis: This should be `PersistenceTransaction`. The same mistake exists in master, which I will fix.

schmidt-sebastian: I will backport your fix to master in the next merge.

PR #1074, opened 2018-08-01
Fixes servicerworker/client state syncing

bojeil-google:
Defines the MessageChannel based utility for communicating window <=> window and window <=> worker.

Adds receiver in indexedDB when operating from a service worker to listen to keyChanged events.
On detection, indexedDB is synced and a response is returned to sender whether the key change was processed. It may be that the key change was already detected. This would return false.

Adds a sender when an indexedDB write operation occurs and a service worker is available. The client will send a keyChanged message to the service worker to process the change and blocks on it. On response, the indexedDB write operation will resolve. The operation will resolve on success or failure. This is a best effort approach. If the service worker fails to process, the write operation should still succeed.

Updates obsoleted APIs in web worker.


bojeil-google:
I added them.

PR #1082, opened 2018-08-01
Enable strict function types

gsoltis:
After making some errors that would've been caught via more strict type checking, I explored what would be required to enable stricter checking.

Let me know what you think.

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -60,7 +60,7 @@ export enum TimerId {
  *
  * Supports cancellation (via cancel()) and early execution (via skipDelay()).
  */
-class DelayedOperation<T extends (AnyJs | void)> implements CancelablePromise<T> {
+class DelayedOperation<T extends AnyJs | void> implements CancelablePromise<T> {
   // handle for use with clearTimeout(), or null if the operation has been
   // executed or canceled already.
   private timerHandle: TimerHandle | null;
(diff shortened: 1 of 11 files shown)


packages/firestore/src/util/async_queue.ts
```diff
@@ -60,7 +60,7 @@ export enum TimerId {
  *
  * Supports cancellation (via cancel()) and early execution (via skipDelay()).
  */
-class DelayedOperation<T> implements CancelablePromise<T> {
+class DelayedOperation<T extends AnyJs | void> implements CancelablePromise<T> {
```
mikelehen: All of this `AnyJs | void` repetition makes me wonder if there's a useful abstraction to extract... but I'm not quite sure. Does an `OptionalAnyJs` make sense (and would it be useful anywhere besides here)?  Or perhaps an OperationReturnValue type alias in this file?

I don't feel strongly...  we just repeat it half a dozen times here, so I'm wondering...  It's probably fine as-is though.

gsoltis: I'll see what I can do. An alias local to one file it shouldn't be a problem, I don't think. `void` appears to be a bit of a weird concept at the intersection of Typescript and Javascript. See https://github.com/Microsoft/TypeScript/issues/20006#issuecomment-344469659 (but maybe avoid reading entire comment chain if you value your time and sanity). The gist of it is the `AnyJs` definition is "any value" (you can assign `undefined` to something), whereas, I think, `void` is intended to mean "no value" (`let x = void;` does not work). Per linked comment, you can technically end up with a value of `void` , but you probably shouldn't.

I *think* modifying our top type, `AnyJs`, to their recommended top type, `null | undefined | {}` might work, but I haven't tried it yet because that seems like it could have far-reaching implications. FWIW, Typescript 3.0 introduces `unknown` that I think is what we would actually want in this case: https://blogs.msdn.microsoft.com/typescript/2018/07/30/announcing-typescript-3-0/#the-unknown-type

mikelehen: Oooh, yes `unknown` will be nice to have. Good find.

To be clear, only "see what you can do" if you feel like it.  This is okay as-is too.

PR #1088, opened 2018-08-03
Multi-Tab Renames

schmidt-sebastian:
This PR:
- Renames the DbOwner store to DbPrimaryClient store (without renaming the IndexedDb location).
- Renames `targetIds` to `queryDataByTarget`
- Renames `indexeddb_schema.test.ts` to `indexeddb_persistence.test.ts`

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -29,8 +29,8 @@ import {
   createOrUpgradeDb,
   DbClientMetadataKey,
   DbClientMetadata,
-  DbOwner,
-  DbOwnerKey,
+  DbPrimaryClient,
+  DbPrimaryClientKey,
   SCHEMA_VERSION
 } from './indexeddb_schema';
 import { LocalSerializer } from './local_serializer';
@@ -306,7 +306,7 @@ export class IndexedDbPersistence implements Persistence {
   }
 
   /** Checks whether `client` is the local client. */
-  private isLocalClient(client: DbOwner | null): boolean {
+  private isLocalClient(client: DbPrimaryClient | null): boolean {
     return client ? client.ownerId === this.clientId : false;
   }
 
@@ -320,7 +320,7 @@ export class IndexedDbPersistence implements Persistence {
   private canActAsPrimary(
     txn: SimpleDbTransaction
   ): PersistencePromise<boolean> {
-    const store = ownerStore(txn);
+    const store = primaryClientStore(txn);
     return store
       .get('owner')
       .next(currentPrimary => {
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -649,7 +647,7 @@ export class IndexedDbPersistence implements Persistence {
       typeof this.document.addEventListener === 'function'
     ) {
       this.documentVisibilityHandler = () => {
-        this.queue.enqueueAndForget<DbOwner | void>(() => {
+        this.queue.enqueueAndForget(() => {
```
mikelehen: Did the generic type annotation intentionally get removed?

schmidt-sebastian: Yeah, it's always 'void' and can be inferred.

schmidt-sebastian:
> LGTM but if I grep for 'owner' there are still a bunch of references that I think we can clean up to unify on the new 'primary' terminology...

The key store is still called `owner`, the member variable in that store is still `owner` and the main comment on IndexedDbPersistence still mentions `owner` (there's another TODO to rewrite that comment).

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -295,14 +293,16 @@ describe('IndexedDb: canActAsPrimary', () => {
...
+  async function clearPrimaryLease(): Promise<void> {
     const simpleDb = await SimpleDb.openOrCreate(
       INDEXEDDB_TEST_DATABASE,
       SCHEMA_VERSION,
       createOrUpgradeDb
     );
-    await simpleDb.runTransaction('readwrite', [DbOwner.store], txn => {
-      const ownerStore = txn.store<DbOwnerKey, DbOwner>(DbOwner.store);
+    await simpleDb.runTransaction('readwrite', [DbPrimaryClient.store], txn => {
+      const ownerStore = txn.store<DbPrimaryClientKey, DbPrimaryClient>(
```
mikelehen: primaryStore?

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -295,14 +293,16 @@ describe('IndexedDb: canActAsPrimary', () => {
...
+        DbPrimaryClient.store
+      );
       return ownerStore.delete('owner');
```
mikelehen: How about creating `DbPrimaryClient.key` similar to how we have `DbTargetGlobal.key`?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1530,10 +1530,16 @@ async function writeOwnerToIndexedDb(clientId: ClientId): Promise<void> {
     createOrUpgradeDb
   );
   await db.runTransaction('readwrite', ['owner'], txn => {
-    const owner = txn.store<DbOwnerKey, DbOwner>(DbOwner.store);
+    const owner = txn.store<DbPrimaryClientKey, DbPrimaryClient>(
```
mikelehen: Any reason not to rename this variable and this method?

PR #1096, opened 2018-08-07
Fixed `get({source: 'cache'})` to return nonexistent documents from cache.

mikelehen:
Port of https://github.com/firebase/firebase-ios-sdk/pull/1642 (although there was no crash on web; it just unnecessarily returned an error instead of the cached nonexistent document)

### DIFF
--- a/packages/firestore/test/integration/api/get_options.test.ts
+++ b/packages/firestore/test/integration/api/get_options.test.ts
@@ -436,6 +436,7 @@ apiDescribe('GetOptions', persistence => {
   // TODO(b/112267729): We should raise a fromCache=true event with a
   // nonexistent snapshot, but because the default source goes through a normal
   // listener, we do not.
+  // tslint:disable-next-line:ban skipping test on purpose.
   it.skip('get deleted doc while offline with default get options', () => {
     return withTestDocAndInitialData(persistence, null, docRef => {
       return docRef
@@ -506,6 +507,7 @@ apiDescribe('GetOptions', persistence => {
   });
 
   // We need the deleted doc to stay in cache, so only run this with persistence.
+  // tslint:disable-next-line:ban skipping test on purpose.
   (persistence ? it : it.skip)(
     'get deleted doc while offline with source=cache',
     () => {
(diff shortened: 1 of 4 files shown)


packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -418,15 +418,37 @@ apiDescribe('GetOptions', persistence => {
...
+            expected => {}
+          )
+      );
+    });
+  });
+
+  // TODO(b/112267729): We should raise a fromCache=true event with a
+  // nonexistent snapshot, but because the default source goes through a normal
+  // listener, we do not.
+  // tslint:disable-next-line:ban skipping test on purpose.
```
schmidt-sebastian: You wrote the same thing twice here, once in tslint, and once in English. Feel free to leave as is :)

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -418,15 +418,37 @@ apiDescribe('GetOptions', persistence => {
...
         .then(() => docRef.get())
-        .then(
-          doc => {
-            expect.fail();
-          },
-          expected => {}
-        );
+        .then(doc => {
+          expect(doc.exists).to.be.false;
+          expect(doc.data).to.be.null;
```
schmidt-sebastian: This should be doc.data() and `undefined`.

packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -525,20 +566,15 @@ apiDescribe('GetOptions', persistence => {
...
-      );
+      return docRef.firestore
+        .disableNetwork()
+        .then(() => docRef.get({ source: 'server' }))
+        .then(
+          doc => {
+            expect.fail();
+          },
+          expected => {}
+        );
```
schmidt-sebastian: I know this is existing code, but it took me a while to parse this. The line breaks make it hard to understand that this is one function call with two arguments, and I believe they might not even be needed:

```
.then(() => expect.fail(), () => {});
```

This is extremely optional.

packages/firestore/test/integration/api/get_options.test.ts
mikelehen: hah, yeah.  Fixed.

packages/firestore/test/integration/api/get_options.test.ts
mikelehen: Wow, good catch. Now I know I can just send you my tests to evaluate rather than actually run them.

packages/firestore/test/integration/api/get_options.test.ts
mikelehen: Yeah...  this pattern exists throughout the file so I'm going to leave it as-is.

PR #1097, opened 2018-08-07
Removing/Addressing TODOs

schmidt-sebastian:
This a) removes two TODOs that are no longer applicable/valid and b) addresses one by changing the test files that wrap the cache to use the primaryLease mode that is used in the rest of the client. 
Note that I did the latter from memory, and if we want to make 100% certain that this is correct, I can go back and verify the callers manually.

### DIFF
--- a/packages/firestore/src/local/mutation_queue.ts
+++ b/packages/firestore/src/local/mutation_queue.ts
@@ -73,8 +73,6 @@ export interface MutationQueue extends GarbageSource {
 
   /**
    * Creates a new mutation batch and adds it to this mutation queue.
-   *
-   * TODO(multitab): Make this operation safe to use from secondary clients.
    */
   addMutationBatch(
     transaction: PersistenceTransaction,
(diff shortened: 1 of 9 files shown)


packages/firestore/src/local/mutation_queue.ts
```diff
@@ -73,8 +73,6 @@ export interface MutationQueue extends GarbageSource {
 
   /**
    * Creates a new mutation batch and adds it to this mutation queue.
-   *
-   * TODO(multitab): Make this operation safe to use from secondary clients.
```
mikelehen: Should this have "Multi-Tab Note: This operation is safe to use from secondary clients." ?  (please check all the comments for consistency)

schmidt-sebastian: Hm, I have added this comment in a few places, but forgot a bunch. Since most operations are now multi-tab safe, should we reverse the comment and put it on the few methods that are not safe to use from secondary tabs?

mikelehen: SGTM.

schmidt-sebastian: Done

PR #1102, opened 2018-08-08
Remove mutation data from client state

schmidt-sebastian:
This PR removes a bunch of code that was only ever used to save us from persisting a mutation state if no client is interested in it anymore.

### DIFF
--- a/packages/firestore/test/integration/api/get_options.test.ts
+++ b/packages/firestore/test/integration/api/get_options.test.ts
@@ -566,16 +566,18 @@ apiDescribe('GetOptions', persistence => {
 
   it('get non existing doc while offline with source=server', () => {
     return withTestDocAndInitialData(persistence, null, docRef => {
-      return docRef.firestore
-        .disableNetwork()
-        // Attempt to get doc.  This will fail since there's nothing in cache.
-        .then(() => docRef.get({ source: 'server' }))
-        .then(
-          doc => {
-            expect.fail();
-          },
-          expected => {}
-        );
+      return (
+        docRef.firestore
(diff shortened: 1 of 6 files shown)


packages/firestore/test/integration/api/get_options.test.ts
```diff
@@ -566,16 +566,18 @@ apiDescribe('GetOptions', persistence => {
...
-        // Attempt to get doc.  This will fail since there's nothing in cache.
-        .then(() => docRef.get({ source: 'server' }))
-        .then(
-          doc => {
-            expect.fail();
-          },
-          expected => {}
-        );
+      return (
+        docRef.firestore
```
schmidt-sebastian: Mike, did you remove the parentheses manually? I know you hate them :)

mikelehen: Bleh, that's so weird.  It seems like sometimes `git push` pushes the branch before running prettier, because my local branch has an "[AUTOMATED]: Prettier Code Styling" commit with this fix, but it apparently wasn't pushed. :-/

schmidt-sebastian:
Regarding the scenario described above: I don't think there are consequences to this race. Since every tab stores its mutation queues in memory, they can always "replay" the mutations (synchronize the affected documents). This replay read will go back to IndexedDb and should return the latest state of documents, even if they haven't changed. 
The only thing I can think of is the  possibility that a mutation stays "pending" if for some reason the IndexedDb transaction succeeds but the LocalStorage write gets dropped. I can't, however, convince myself that this would really ever happen.

PR #1105, opened 2018-08-08
Drop limbo collector

gsoltis:
Port of https://github.com/firebase/firebase-ios-sdk/pull/1413

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -326,7 +326,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
           .catch(err => this.ignoreIfPrimaryLeaseLoss(err));
       }
     } else {
-      await this.removeAndCleanupQuery(queryView);
+      this.removeAndCleanupQuery(queryView);
       await this.localStore.releaseQuery(
         query,
         /*keepPersistedQueryData=*/ true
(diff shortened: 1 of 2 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -680,12 +676,38 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     delete this.queryViewsByTarget[queryView.targetId];
 
     if (this.isPrimary) {
+      const limboKeys = this.limboDocumentRefs.referencesForId(
+        queryView.targetId
+      );
       this.limboDocumentRefs.removeReferencesForId(queryView.targetId);
-      await this.gcLimboDocuments();
+      limboKeys.forEach(async limboKey => {
```
schmidt-sebastian: I have to run to a meeting, but in case I forget this later: I don't think this works. You have to append this Promise to another data structure and then wait on all those promises. ForEach doesn't wait on the Promise resolution.

gsoltis: I suspect you're right. I've updated it with a chain of `PersistencePromise` with a final `await p.toPromise()` at the end.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -680,12 +677,40 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     delete this.queryViewsByTarget[queryView.targetId];
 
     if (this.isPrimary) {
+      const limboKeys = this.limboDocumentRefs.referencesForId(
+        queryView.targetId
+      );
       this.limboDocumentRefs.removeReferencesForId(queryView.targetId);
-      await this.gcLimboDocuments();
+      let p = PersistencePromise.resolve();
+      limboKeys.forEach(limboKey => {
```
schmidt-sebastian: Please update to use a normal Promise.

PR #1106, opened 2018-08-09
Garbage Collect Mutation Batch Keys

schmidt-sebastian:
As per go/garbage-collection. This works in Chrome, Firefox and Safari.

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -674,6 +674,15 @@ export class WebStorageSharedClientState implements SharedClientState {
     error?: FirestoreError
   ): void {
     this.persistMutationState(batchId, state, error);
+
+    // Once a final mutation result is observed by other clients, they no longer
+    // access the mutation's metadata entry. Since LocalStorage replays events
+    // in order, it is safe to delete the entry right after updating it.
+    // We enqueue the delete on the queue since LocalStorage writes acquire
+    // a lock internally, and we don't need to wait for the delete to succeed.
+    this.queue.enqueueAndForget(() =>
+      Promise.resolve(this.removeMutationState(batchId))
+    );
   }
 
   addLocalQueryTarget(targetId: TargetId): QueryTargetState {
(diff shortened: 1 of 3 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -674,6 +674,15 @@ export class WebStorageSharedClientState implements SharedClientState {
     error?: FirestoreError
   ): void {
     this.persistMutationState(batchId, state, error);
+
+    // Once a final mutation result is observed by other clients, they no longer
+    // access the mutation's metadata entry. Since LocalStorage replays events
+    // in order, it is safe to delete the entry right after updating it.
+    // We enqueue the delete on the queue since LocalStorage writes acquire
+    // a lock internally, and we don't need to wait for the delete to succeed.
```
mikelehen: So we're enqueuing the delete as a performance optimization?  Do we have reason to believe this is going to make a meaningful difference in some scenario?  Else I'd keep the code simple for now.

(FWIW, persistMutationState already is writing to LocalStorage, acquiring the lock, etc. so it's not like this will be any more expensive)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -722,8 +731,8 @@ export class WebStorageSharedClientState implements SharedClientState {
     removedBatchIds: BatchId[],
     addedBatchIds: BatchId[]
   ): void {
```
mikelehen: So it occurs to me that removedBatchIds is going to be the old mutation queue (for all clients) and addedBatchIds is going to be the new mutation queue (for all clients) and every client is going to run this handleUserChange() logic to add/remove the pending batches...  which probably isn't desirable and could potentially lead to race conditions...  Perhaps only the primary should do this or something?  Not 100% sure.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Optimize first, remove later. Ok, maybe not... I removed the only L6-worthy part of this PR.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Yeah, even if it is not racy, it is necessary to run this code more than once. I guarded it in a `isPrimary` check (in SyncEngine). Thanks for noticing.

schmidt-sebastian: Hm, this breaks the Spec Tests and hence I just added a TODO for now.

PR #1108, opened 2018-08-09
Garbage Collect the Remote Document Changelog

schmidt-sebastian:
This PR adds garbage collection of the Remote Document Change Log. Right now, this runs as part of the client metadata refresh, but I added a TODO for Greg to move it to the LRU callsite once it exists.

I only run the GC code every 30 minutes since it requires a read of all client metadata. 

I tried to unify some code which made `canActAsPrimary` slightly less efficient and this can be reverted.

This code is not well tested (and could potentially be better tested if I added a TimerId for it and scheduled it separately), but I did confirm that Chrome runs this code after 30 minutes (and at startup).

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -57,6 +57,15 @@ const LOG_TAG = 'IndexedDbPersistence';
  * are ignored.
  */
 const CLIENT_METADATA_MAX_AGE_MS = 5000;
+
+/**
+ * Oldest acceptable age in milliseconds for client metadata before it and its
+ * associated data (such as the remote document cache changelog) can be
+ * garbage collected. Clients that exceed this threshold will not be able to
+ * replay Watch events that occurred before this threshold.
+ */
+const CLIENT_METADATA_GARBAGE_COLLECTION_THRESHOLD_MS = 30 * 60 * 1000; // 30 Minutes
+
 /**
  * The interval at which clients will update their metadata, including
  * refreshing their primary lease if held or potentially trying to acquire it if
@@ -183,11 +192,17 @@ export class IndexedDbPersistence implements Persistence {
   /** The client metadata refresh task. */
   private clientMetadataRefresher: CancelablePromise<void>;
 
+  /** The last time we garbage collected the Remote Document Changelog. */
+  private lastGarbageCollectionTime = Number.NEGATIVE_INFINITY;
+
   /** Whether to allow shared multi-tab access to the persistence layer. */
   private allowTabSynchronization: boolean;
 
   /** A listener to notify on primary state changes. */
   private primaryStateListener: PrimaryStateListener = _ => Promise.resolve();
+
+  private queryCache: IndexedDbQueryCache;
+  private remoteDocumentCache: IndexedDbRemoteDocumentCache;
 
   constructor(
     private readonly persistenceKey: string,
@@ -226,6 +241,11 @@ export class IndexedDbPersistence implements Persistence {
     return SimpleDb.openOrCreate(this.dbName, SCHEMA_VERSION, createOrUpgradeDb)
       .then(db => {
         this.simpleDb = db;
+        this.queryCache = new IndexedDbQueryCache(this.serializer);
+        this.remoteDocumentCache = new IndexedDbRemoteDocumentCache(
+          this.serializer,
+          /*keepDocumentChangeLog=*/ this.allowTabSynchronization
+        );
       })
       .then(() => {
         this.attachVisibilityHandler();
@@ -274,7 +294,8 @@ export class IndexedDbPersistence implements Persistence {
             this.clientId,
             Date.now(),
             this.networkEnabled,
-            this.inForeground
+            this.inForeground,
+            this.remoteDocumentCache.lastProcessedDocumentChangesId
           )
         )
         .next(() => this.canActAsPrimary(txn))
@@ -295,7 +316,9 @@ export class IndexedDbPersistence implements Persistence {
           if (wasPrimary && !this.isPrimary) {
             return this.releasePrimaryLeaseIfHeld(txn);
           } else if (this.isPrimary) {
-            return this.acquireOrExtendPrimaryLease(txn);
+            return this.acquireOrExtendPrimaryLease(txn).next(() =>
+              this.maybeGarbageCollectRemoteDocumentChangelog(txn)
+            );
           }
         });
     });
@@ -306,6 +329,59 @@ export class IndexedDbPersistence implements Persistence {
   ): PersistencePromise<void> {
     const metadataStore = clientMetadataStore(txn);
     return metadataStore.delete(this.clientId);
+  }
+
+  /**
+   * If the garbage collection threshold has passed, prunes the
+   * RemoteDocumentChanges store based on the metadata state of all existing
+   * clients.
+   */
+  // TODO(gsoltis): Execute this as part of LRU GC
+  private maybeGarbageCollectRemoteDocumentChangelog(
+    txn: SimpleDbTransaction
+  ): PersistencePromise<void> {
+    assert(this.isPrimary, 'GC should only run in the primary client');
+
+    if (
+      !this.isWithinAge(
+        this.lastGarbageCollectionTime,
+        CLIENT_METADATA_GARBAGE_COLLECTION_THRESHOLD_MS
+      )
+    ) {
+      this.lastGarbageCollectionTime = Date.now();
+
+      return clientMetadataStore(txn)
+        .loadAll()
+        .next(existingClients => {
+          let activeClients = this.filterActiveClients(
+            existingClients,
+            CLIENT_METADATA_GARBAGE_COLLECTION_THRESHOLD_MS
+          );
+
+          // The primary client doesn't read from the document change log,
+          // and hence we exclude it when we determine the minimum
+          // `lastProcessedDocumentChangesId`.
+          activeClients = activeClients.filter(
+            client => client.clientId !== this.clientId
+          );
+
+          if (activeClients.length) {
+            const mostOutdatedClient = activeClients.reduce(
+              (previousValue, currentValue) =>
+                (currentValue.lastProcessedDocumentChangesId || 0) <
+                (previousValue.lastProcessedDocumentChangesId || 0)
+                  ? currentValue
+                  : previousValue
+            );
+            return this.remoteDocumentCache.removeDocumentChangesThroughChangeId(
+              new IndexedDbTransaction(txn),
+              mostOutdatedClient.lastProcessedDocumentChangesId
+            );
+          }
+        });
+    } else {
+      return PersistencePromise.resolve();
+    }
   }
 
   /**
@@ -345,7 +421,10 @@ export class IndexedDbPersistence implements Persistence {
       .next(currentPrimary => {
         const currentLeaseIsValid =
           currentPrimary !== null &&
-          this.isWithinMaxAge(currentPrimary.leaseTimestampMs) &&
+          this.isWithinAge(
+            currentPrimary.leaseTimestampMs,
+            CLIENT_METADATA_MAX_AGE_MS
+          ) &&
           !this.isClientZombied(currentPrimary.ownerId);
 
         // A client is eligible for the primary lease if:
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -308,6 +331,59 @@ export class IndexedDbPersistence implements Persistence {
...
+          );
+
+          // The primary client doesn't read from the document change log,
+          // and hence we exclude it when we determine the minimum
+          // `lastProcessedDocumentChangesId`.
+          activeClients = activeClients.filter(
+            client => client.clientId !== this.clientId
+          );
+
+          if (activeClients.length) {
```
mikelehen: `> 0` please?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -308,6 +331,59 @@ export class IndexedDbPersistence implements Persistence {
...
+            const mostOutdatedClient = activeClients.reduce(
+              (previousValue, currentValue) =>
+                (currentValue.lastProcessedDocumentChangesId || 0) <
+                (previousValue.lastProcessedDocumentChangesId || 0)
+                  ? currentValue
+                  : previousValue
+            );
+            return this.remoteDocumentCache.removeDocumentChangesThroughChangeId(
+              new IndexedDbTransaction(txn),
+              mostOutdatedClient.lastProcessedDocumentChangesId
```
mikelehen: Since we don't actually need `mostOutdatedClient` we can probably simplify all of this a bit, along the lines of:

```
const clientChangeIds = activeClients.map(client => client.lastProcessedDocumentChangesId || 0);
const oldestChangeId = Math.min(...clientChangeIds);
```

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -308,6 +331,59 @@ export class IndexedDbPersistence implements Persistence {
...
+            );
+          }
+        });
```
mikelehen: Are we supposed to GC the old metadata entries too?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -42,7 +42,7 @@ import { SimpleDbStore } from './simple_db';
 
 export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
   /** The last id read by `getNewDocumentChanges()`. */
-  private lastReturnedDocumentChangesId = 0;
+  private _lastProcessedDocumentChangesId = 0;
```
mikelehen: Can we be consistent with 'ChangeId' or 'ChangesId' ?  I think I prefer the former.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -184,6 +194,27 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+   * (inclusive).
+   */
+  removeDocumentChangesThroughChangeId(
+    transaction: PersistenceTransaction,
+    changeId: number
+  ): PersistencePromise<void> {
+    assert(
+      changeId <= this._lastProcessedDocumentChangesId,
+      'Cannot remove unprocessed changes'
+    );
```
mikelehen: I'm a little surprised this assert works since only the primary will remove changes, but the primary also doesn't actually need to process these changes.  Do we process them anyway or something?  Is that wasted work?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -184,6 +194,27 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+
+    if (changeId === 0) {
```
mikelehen: Is this an optimization?  Is it warranted (are we going to hit this frequently or in critical paths)?

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -51,12 +51,17 @@ const LOCAL_STORAGE_PREFIX = 'firestore_';
  * any previous contents if they existed.
  */
 export async function testIndexedDbPersistence(
-  synchronizeTabs?: boolean
+  options: {
+    dontPurgeData?: boolean;
```
mikelehen: What would you think of making this 'purgeData' and having it default to true if not provided?

packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -42,25 +57,55 @@ describe('IndexedDbRemoteDocumentCache', () => {
...
+    persistence = await persistenceHelpers.testIndexedDbPersistence({
+      synchronizeTabs: true,
+      dontPurgeData: true
+    });
+    const changedDocs = await persistence.runTransaction(
+      'getNewDocumentChanges',
+      false,
+      txn => {
+        const cache = persistence.getRemoteDocumentCache();
+        return cache.getNewDocumentChanges(txn);
```
mikelehen: So looking at this test it occurs to me to ask:  If we only GC these every 30 minutes, does that mean when you open a new tab it's going to have to replay 30 minutes of changes on start?

packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -216,3 +240,21 @@ function genericRemoteDocumentCacheTests(
...
+  actual.forEach((actualKey, actualDoc) => {
+    const found = expected.find(expectedDoc => {
+      if (actualKey.isEqual(expectedDoc.key)) {
+        expectEqual(actualDoc, expectedDoc);
+        return true;
+      }
+      return false;
+    });
+
+    expect(found).to.not.be.undefined;
```
mikelehen: is there no `.to.be.defined` ?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -57,6 +57,15 @@ const LOG_TAG = 'IndexedDbPersistence';
...
  */
 const CLIENT_METADATA_MAX_AGE_MS = 5000;
+
+/**
+ * Oldest acceptable age in milliseconds for client metadata before it and its
+ * associated data (such as the remote document cache changelog) can be
+ * garbage collected. Clients that exceed this threshold will not be able to
+ * replay Watch events that occurred before this threshold.
+ */
+const CLIENT_METADATA_GARBAGE_COLLECTION_THRESHOLD_MS = 30 * 60 * 1000; // 30 Minutes
```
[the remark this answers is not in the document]
schmidt-sebastian: I renamed it to `CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS`.  `CLIENT_STATE_GC_THRESHOLD_MS` looks kind of weird to me.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -183,12 +192,18 @@ export class IndexedDbPersistence implements Persistence {
   /** The client metadata refresh task. */
   private clientMetadataRefresher: CancelablePromise<void>;
 
+  /** The last time we garbage collected the Remote Document Changelog. */
+  private lastGarbageCollectionTime = Number.NEGATIVE_INFINITY;
```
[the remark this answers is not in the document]
schmidt-sebastian: I mostly just use the "ms" suffix for durations. This would be `lastGarbageCollectionTimeEpochMs` if I wanted to be 100% unambiguous.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -183,12 +192,18 @@ export class IndexedDbPersistence implements Persistence {
...
+
   /** Whether to allow shared multi-tab access to the persistence layer. */
   private allowTabSynchronization: boolean;
 
   /** A listener to notify on primary state changes. */
   private primaryStateListener: PrimaryStateListener = _ => Promise.resolve();
 
+  private queryCache: IndexedDbQueryCache;
+  private remoteDocumentCache: IndexedDbRemoteDocumentCache;
```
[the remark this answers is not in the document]
schmidt-sebastian: These are unfortunately assigned in `start()` and can't be marked readonly. I need to know the setting of the `sychronizeTabs` flag during construction.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -274,7 +294,8 @@ export class IndexedDbPersistence implements Persistence {
             this.clientId,
             Date.now(),
             this.networkEnabled,
-            this.inForeground
+            this.inForeground,
+            this.remoteDocumentCache.lastProcessedDocumentChangesId
```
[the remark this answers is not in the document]
schmidt-sebastian: FIxed here and everywhere.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -308,6 +331,59 @@ export class IndexedDbPersistence implements Persistence {
     return metadataStore.delete(this.clientId);
   }
 
+  /**
+   * If the garbage collection threshold has passed, prunes the
+   * RemoteDocumentChanges store based on the metadata state of all existing
+   * clients.
+   */
+  // TODO(gsoltis): Execute this as part of LRU GC
```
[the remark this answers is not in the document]
schmidt-sebastian: Hm. I don't have an answer to that. But I am ok with removing this TODO.

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: Mmmmkay.

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: ... but then I can't use `reduce`. It is much simpler, however..

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: Yeah, but I want to do this in a different PR.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I renamed all "changesId" to "changeId".

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I tested this over the course of an hour. I think this assert might become invalid as more changes accumulate. Note that the primary tab sets `_lastProcessedDocumentChangesId` to the existing highest value at startup.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: This gets hit during startup, but only for the very first time. It's probably ok to remove this.

packages/firestore/test/unit/local/persistence_test_helpers.ts
schmidt-sebastian: a) I don't like negative argument names
b) I don't like optional arguments that default to 'true'.

I slightly preferred compromising `a` over `b` in this case.

packages/firestore/test/unit/local/remote_document_cache.test.ts
schmidt-sebastian: But.... this is just code that I moved around.

Changed it to `expect(found).to.exist`

packages/firestore/test/unit/local/remote_document_cache.test.ts
schmidt-sebastian: This is probably not the cleanest test, but it doesn't call start() on the RemoteDocumentCache. start() forwards the internal pointer to the current location in the DB, and that means that older changes won't get replayed.

Short of fixing the test, I added a comment to explain this behavior.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -64,7 +64,7 @@ const CLIENT_METADATA_MAX_AGE_MS = 5000;
  * garbage collected. Clients that exceed this threshold will not be able to
  * replay Watch events that occurred before this threshold.
  */
-const CLIENT_METADATA_GARBAGE_COLLECTION_THRESHOLD_MS = 30 * 60 * 1000; // 30 Minutes
+const CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS = 30 * 60 * 1000; // 30 Minutes
```
mikelehen: [adding this comment here out of laziness]

Trying to reconcile the two timeouts here (CLIENT_METADATA_MAX_AGE_MS and CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS), I think:

1. CLIENT_METADATA_MAX_AGE_MS is basically how stale a client can be and still be eligible to be primary.
2. CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS is basically how stale a client can be and still "catch up" to current.

If that's right, we might want to do some renaming / comment tweaks...  But I also think the fact that `getActiveClients()` uses CLIENT_METADATA_MAX_AGE_MS is kinda' wrong.  If a primary change happens, any clients more stale than CLIENT_METADATA_MAX_AGE_MS will be ignored by the new primary and we won't start listeners for them.  So they will not properly "catch up" until they perform some extra action to re-persist their client state to local storage so that the primary notices them.

On that note, I'm wondering if we might want to ditch `getActiveClients()` and instead have the primary client write some sort of "report in please" notification to LocalStorage that triggers all secondary tabs to (re-)write their current state.  We could then use that mechanism on primary changes in order to get existing clients, rather than using `getActiveClients()`.  I think that might be cleaner and easier to reason about (we don't have an "active client" notion that crosses over between IndexedDbPersistence and SharedClientState)... WDYT?

schmidt-sebastian: I don't think I want to re-work the source of how we determine what clients are active. IndexedDb seems to be the more accurate source of truth here, and it has the only timestamp that actually gets updated (note that the row in LocalStorage has an updateTime, and I am wondering if I should remove it since it does not get refreshed).


As for everything else you said, I think you are correct: CLIENT_METADATA_MAX_AGE_MS is (right now) the age at which clients are considered active, and CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS is when we "give up on the client". I think you make a good case that we should not consider clients inactive that we don't yet "give up" on. It might make sense to use the 30 minute interval everywhere, and only use the 5s interval in the primary tab election. Since it is Friday, I am going to submit this PR, but I did file b/112478723.

PR #1109, opened 2018-08-09
GRPC Tweaks.

mikelehen:
* Use Client.waitForReady() to determine when the stream is actually open,
  rather than just considering it open right away.
* Don't lowercase RPC names because the gRPC stub supports both now.
* Remove event handlers / asserts that have been proven safe / unnecessary.

### DIFF
--- a/packages/firestore/src/platform_node/grpc_connection.ts
+++ b/packages/firestore/src/platform_node/grpc_connection.ts
@@ -125,11 +125,8 @@ export class GrpcConnection implements Connection {
     rpcName: string,
     token: Token | null
   ): UnaryRpc<Req, Resp> | ReadableRpc<Req, Resp> | DuplexRpc<Req, Resp> {
-    // RPC Methods have the first character lower-cased
-    // (e.g. Listen => listen(), BatchGetDocuments => batchGetDocuments()).
-    const rpcMethod = rpcName.charAt(0).toLowerCase() + rpcName.slice(1);
     const stub = this.ensureActiveStub(token);
-    const rpc = stub[rpcMethod];
+    const rpc = stub[rpcName];
     assert(rpc != null, 'Unknown RPC: ' + rpcName);
 
     const metadata = createMetadata(this.databaseInfo, token);
@@ -255,12 +252,6 @@ export class GrpcConnection implements Connection {
       close();
     });
 
-    grpcStream.on('finish', () => {
-      // TODO(mikelehen): I *believe* this assert is safe and we can just remove
-      // the 'finish' event if we don't see the assert getting hit for a while.
-      assert(closed, 'Received "finish" event without close() being called.');
-    });
-
     grpcStream.on('error', (grpcError: grpc.ServiceError) => {
       log.debug(
         LOG_TAG,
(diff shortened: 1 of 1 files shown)


packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -273,23 +264,19 @@ export class GrpcConnection implements Connection {
...
-    // registered
-    setTimeout(() => {
-      stream.callOnOpen();
-    }, 0);
+
+    // TODO(mikelehen): We will call waitForReady() on each stream attempt,
+    // but waitForReady() won't call our callback until the channel is
+    // successfully established (note that the callback is not called for most
+    // stream errors). So we'll accumulate more and more outstanding wait calls
+    // over time while the network is down, manifesting as a minor memory leak.
```
mikelehen: I can solve this TODO if you want but I think it's going to be a bit ugly so wanted to check with you first to see if we really care...  We basically need a singleton (per stub) ChannelWaiter object that you can subscribe to/from (and make sure that streams unsubscribe when they get an error).  

And to be fully leak-proof, it may need to use the raw connectivityState API rather than waitForReady(), but that isn't exposed in grpc's .d.ts file so we'd be in un-typed territory.

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -125,11 +125,8 @@ export class GrpcConnection implements Connection {
     rpcName: string,
     token: Token | null
   ): UnaryRpc<Req, Resp> | ReadableRpc<Req, Resp> | DuplexRpc<Req, Resp> {
-    // RPC Methods have the first character lower-cased
-    // (e.g. Listen => listen(), BatchGetDocuments => batchGetDocuments()).
-    const rpcMethod = rpcName.charAt(0).toLowerCase() + rpcName.slice(1);
     const stub = this.ensureActiveStub(token);
-    const rpc = stub[rpcMethod];
+    const rpc = stub[rpcName];
```
wilhuff: This is surprising that it works.

packages/firestore/src/platform_node/grpc_connection.ts
wilhuff: How is this any different than the new stream we're creating per invocation above?

Couldn't we make this another callback on the stream bridge?

packages/firestore/src/platform_node/grpc_connection.ts
mikelehen: Looking at the `stub` object in the debugger, it has 2 methods for every RPC, supporting either case now.  E.g. `listen` and `Listen`.

packages/firestore/src/platform_node/grpc_connection.ts
mikelehen: `grpcStream` is short-lived.  We create it on each retry and should throw it away once it fails.  Then we create a new one for the next retry.  No leak.

`client` is long-lived (it's the stub that we create per-token and lives at least for the life of the token, possibly longer).  Additionally our `client.waitForReady()` callbacks are long-lived (when the network is down) because they're not called until the client/channel successfully connects (potentially after many retry attempts).

So after 10 retries, we'll only have a single `grpcStream` of interest, but we'll have queued 10 `waitForReady()` callbacks.  Once we actually connect, all 10 will be called by grpc but all but the most recent one will be no-ops because those streams are closed / forgotten.

Note I haven't done any deep analysis of this code to convince myself we're not leaking lots of other things.  But I'm pretty certain we can easily build up an unbounded number of `waitForReady()` callbacks...  That said our max backoff is 60 seconds. With 2 streams, retrying once a minute, that's `2 * 24 * 60 = 2880` callbacks leaked per day of being offline and retrying.  I doubt that's a noticeable memory or CPU impact, so I'm semi-inclined not to care. :-)

mikelehen:
Bleh.  On further consideration, I think the waitForReady stuff should probably be abandoned (tl;dr: It could potentially cause us to wait for an unnecessary roundtrip).

I noticed that webchannel actually doesn't wait for open either: https://github.com/firebase/firebase-js-sdk/blob/d21071e913405533194504fceebad29cd5b05e61/packages/firestore/src/platform_browser/webchannel_connection.ts#L366

And it's quite intentional in order to make sure that we send our first actual stream request in the webchannel handshake...  And we should probably give gRPC the same opportunity.

I think this means we should actually probably remove our onOpen callback entirely since all of our platforms just artificially call it immediately apparently.  And consequently we should rethink the "Resilient Offline Gets" design a bit (I'll add comments to the doc tomorrow). :-/

mikelehen:
Removed the waitForReady change so this PR is just trivial cleanup now.

PR #1116, opened 2018-08-13
Stop starting the remote document cache

gsoltis:
Remove `start()` from the `RemoteDocumentCache` interface. The memory implementation was a no-op, and `IndexedDbPersistence` can guarantee an already-started `IndexedDbRemoteDocumentCache` instance after it has been started itself.

Additionally, move the specification of multi-tab support to the `IndexedDbPersistence` constructor.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -210,13 +210,13 @@ export class IndexedDbPersistence implements Persistence {
     platform: Platform,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    synchronizeTabs?: boolean
+    synchronizeTabs: boolean
   ) {
     this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
     this.serializer = new LocalSerializer(serializer);
     this.document = platform.document;
     this.window = platform.window;
-    this.allowTabSynchronization = !!synchronizeTabs;
+    this.allowTabSynchronization = synchronizeTabs;
     this.queryCache = new IndexedDbQueryCache(this.serializer);
     this.remoteDocumentCache = new IndexedDbRemoteDocumentCache(
       this.serializer,
@@ -245,6 +245,7 @@ export class IndexedDbPersistence implements Persistence {
       .then(db => {
         this.simpleDb = db;
       })
+      .then(() => this.startRemoteDocumentCache())
       .then(() => {
         this.attachVisibilityHandler();
         this.attachWindowUnloadHook();
@@ -252,7 +253,6 @@ export class IndexedDbPersistence implements Persistence {
           this.scheduleClientMetadataAndPrimaryLeaseRefreshes()
         );
       })
-      .then(() => this.startRemoteDocumentCache())
       .then(() => {
         this._started = true;
       });
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -209,22 +209,27 @@ export class IndexedDbPersistence implements Persistence {
     private readonly clientId: ClientId,
     platform: Platform,
     private readonly queue: AsyncQueue,
-    serializer: JsonProtoSerializer
+    serializer: JsonProtoSerializer,
+    synchronizeTabs?: boolean
   ) {
     this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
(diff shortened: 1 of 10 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -209,22 +209,27 @@ export class IndexedDbPersistence implements Persistence {
     private readonly clientId: ClientId,
     platform: Platform,
     private readonly queue: AsyncQueue,
-    serializer: JsonProtoSerializer
+    serializer: JsonProtoSerializer,
+    synchronizeTabs?: boolean
```
schmidt-sebastian: I would make the argument non-optional in internal code. I believe this was optional in start() to not break inheritance.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -254,11 +252,18 @@ export class IndexedDbPersistence implements Persistence {
           this.scheduleClientMetadataAndPrimaryLeaseRefreshes()
         );
       })
+      .then(() => this.startRemoteDocumentCache())
```
schmidt-sebastian: I think you want to start the remote document cache before attaching the visibility handler and invoking the metadata refresh. That allows the all code in this file to use the remote document cache (which is used from `updateClientMetadataAndTryBecomePrimary`).

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -100,13 +102,40 @@ async function withPersistence(
...
 
+async function withPersistence(
+  clientId: ClientId,
+  fn: (
+    persistence: IndexedDbPersistence,
+    platform: TestPlatform,
+    queue: AsyncQueue
+  ) => Promise<void>
+): Promise<void> {
+  return withConfiguredPersistence(clientId, new PersistenceSettings(true), fn);
```
schmidt-sebastian:
```
new PersistenceSettings(/* enabled= */ true)
```

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -77,8 +78,9 @@ function withDb(
     });
 }
 
-async function withPersistence(
+async function withConfiguredPersistence(
```
schmidt-sebastian: Nit: Maybe withPersistenceInternal/withCustomPersistence? I am not a big fan of these names either, but just throwing it out there since all of our persistence implementations are configured.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -100,13 +102,40 @@ async function withPersistence(
...
+  fn: (
+    persistence: IndexedDbPersistence,
+    platform: TestPlatform,
+    queue: AsyncQueue
+  ) => Promise<void>
+): Promise<void> {
+  return withConfiguredPersistence(clientId, new PersistenceSettings(true), fn);
+}
+
+async function withMultiTabPersistence(
```
schmidt-sebastian: BTW, I like these helpers a lot.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
gsoltis: I guess I'll do `withCustomPersistence`.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_persistence.ts

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: SGTM.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -100,13 +102,46 @@ async function withPersistence(
...
+  ) => Promise<void>
+): Promise<void> {
+  return withCustomPersistence(
+    clientId,
+    new PersistenceSettings(/* enabled */ true),
+    fn
+  );
+}
+
+async function withMultiTabPersistence(
```
schmidt-sebastian: One last suggestion if you like to add more iteration to this PR: The code refers to this feature as "multiclient" rather than "multitab", since it could technically be ported to the other platforms. Maybe. Probably not.

PR #1119, opened 2018-08-14
Delete stale client metadata

schmidt-sebastian:
This removes outdated client metadata entries from IndexedDb. It has been tested manually in Chrome.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -352,7 +352,7 @@ export class IndexedDbPersistence implements Persistence {
       return clientMetadataStore(txn)
         .loadAll()
         .next(existingClients => {
-          let activeClients = this.filterActiveClients(
+          const activeClients = this.filterActiveClients(
             existingClients,
             CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS
           );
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -352,12 +352,25 @@ export class IndexedDbPersistence implements Persistence {
...
-          // TODO(multitab): Remove outdated client metadata
           let activeClients = this.filterActiveClients(
             existingClients,
             CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS
           );
 
+          // Delete metadata for client that are no longer considered active.
+          let p = PersistencePromise.resolve();
+          existingClients
+            .filter(client => activeClients.indexOf(client) === -1)
```
mikelehen: [This is totally optional; feel free to ignore]

I'm sure it doesn't actually matter (number of clients should be small), but this is unnecessarily O(n^2) and in general we're doing a lot of extra enumeration.  It is tempting to:

1. Drop `filterActiveClients()` and instead create `isClientEligibleForPrimary()` and `isClientGarbageCollectable()` helper functions that perform the appropriate threshold check.
2. Replace all usages of `filterActiveClients(...)` with `.filter(isClientEligibleForPrimary)` or `.filter(isClientGarbageCollectable)`
3. Rework this specific piece of code to:

```
const inactiveClients = [], activeClients = [];
existingClients.forEach(client => {
  if (isClientGarbageCollectable(client)) {
    inactiveClients.push(client);
  } else {
    activeClients.push(client);
  }
});
```

Or else forego having `existingClients` and `activeClients` entirely and instead just inline the GC logic into that loop.

schmidt-sebastian: I'm gonna leave this as is for now since I am kinda short on time these days. Thanks for the suggestion though!

PR #1120, opened 2018-08-14
firebase-ios-sdk/1499: Restart streams on any token change.

mikelehen:
See https://github.com/firebase/firebase-ios-sdk/issues/1499

This reworks our UserListener into a CredentialChangeListener which
fires on any token change, even if the User did not change. This allows
us to restart our streams (but not switch mutation queues, etc.) on token
changes.

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -15,7 +15,7 @@
  */
 
 import { User } from '../auth/user';
-import { assert, fail } from '../util/assert';
+import { assert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { FirebaseApp } from '@firebase/app-types';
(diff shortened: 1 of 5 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -82,23 +84,24 @@ export interface CredentialsProvider {
...
 /** A CredentialsProvider that always yields an empty token. */
 export class EmptyCredentialsProvider implements CredentialsProvider {
   /**
-   * Stores the User listener registered with setUserChangeListener()
+   * Stores the listener registered with setChangeListener()
    * This isn't actually necessary since the UID never changes, but we use this
    * to verify the listen contract is adhered to in tests.
    */
-  private userListener: UserListener | null = null;
+  private listener: CredentialChangeListener | null = null;
```
wilhuff: We call this a `changeListener` below. Could we make these consistent?

PR #1123, opened 2018-08-15
Delete leftover zombied client IDs

schmidt-sebastian:
One more cleanup PR to delete zombied client IDs that may still linger in IndexedDb. This is "mostly" untested, since my client doesn't have an inactive clients with leftover zombie IDs.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -346,6 +346,9 @@ export class IndexedDbPersistence implements Persistence {
   ): PersistencePromise<void> {
     assert(this.isPrimary, 'GC should only run in the primary client');
 
+    let activeClients;
+    let inactiveClients;
+
     if (
       !this.isWithinAge(
         this.lastGarbageCollectionTime,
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -393,6 +398,15 @@ export class IndexedDbPersistence implements Persistence {
...
           }
+        })
+        .next(() => {
+          // Delete potential leftover entries that may continue to mark the
+          // inactive clients as zombied in LocalStorage.
+          inactiveClients.forEach(inactiveClient => {
+            this.window.localStorage.removeItem(
+              this.zombiedClientLocalStorageKey(inactiveClient.cliendId)
+            );
+          });
```
mikelehen: FWIW, this would be slightly more leakproof if you did this before deleting the IndexedDb entries.  (theoretically a crash after deleting the IndexedDb entries but before deleting the LocalStorage entry would leave the LocalStorage entries indefinitely).  I don't think I really care though.

schmidt-sebastian: Yes, that's true. But then the IndexedDb entries could come back to life, and I definitely don't want that.

schmidt-sebastian: Hm, I guess I need to do that after committing the transaction. I'll ponder about that some more.

PR #1127, opened 2018-08-16
Only delete zombied entries on commit success

schmidt-sebastian:
This is a follow up to https://github.com/firebase/firebase-js-sdk/pull/1123.

The idea is that we only want to delete the zombied entries if we actually deleted the client metadata from IndexedDb. To do this, I moved the transaction boundary inside the `maybeGarbageCollectMultiClientState` function. To keep things simple, I am no longer running this method at startup, but only after the first lease refresh.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -321,9 +321,7 @@ export class IndexedDbPersistence implements Persistence {
           if (wasPrimary && !this.isPrimary) {
             return this.releasePrimaryLeaseIfHeld(txn);
           } else if (this.isPrimary) {
-            return this.acquireOrExtendPrimaryLease(txn).next(() =>
-              this.maybeGarbageCollectMultiClientState(txn)
-            );
+            return this.acquireOrExtendPrimaryLease(txn);
           }
         });
     });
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -341,75 +339,77 @@ export class IndexedDbPersistence implements Persistence {
...
-        });
-    } else {
-      return PersistencePromise.resolve();
+      });
+
+      // Delete potential leftover entries that may continue to mark the
+      // inactive clients as zombied in LocalStorage.
+      // These deletes have to be executed only after the IndexedDb garbage
+      // collection commits successfully. Otherwise, stale clients' metadata
+      // may temporarily be included in another clients primary election.
```
mikelehen: FWIW I'm having trouble coming up with a case where this actually matters.  The zombie thing is mostly there to make sure that stale primaries get immediately replaced with a new primary (especially on e.g. a page refresh).  But this GC always runs by the primary client, so any stale primary has necessarily been "un-elected" already...

And suppose we deleted zombies from LocalStorage before deleting from IndexedDb.  In the case of a crash preventing us from doing the latter, we're screwed no matter what because if this tab crashed, it probably didn't write a zombie entry for itself and since it's primary, that's the only thing that really matters. We'll be stuck waiting for its lease to expire.

All this said, I don't have a problem with the current code.  Maybe there are cases where this is safer. But in general we're getting into edge case behavior that doesn't matter _too_ much since leases will expire after 5 seconds anyway so any problems should "fix themselves".

My main concern is that this "have to be executed" comment suggests more requirements on this code than I think probably exist in reality, which could potentially confuses somebody later.

I'd perhaps rephrase the comment:

> Ideally we'd delete the IndexedDb and LocalStorage zombie entries for the client atomically, but we can't. So we opt to delete the IndexedDb entries first to avoid potentially reviving a zombied client.

This should communicate the intent of the code without making somebody too afraid of changing it later. :-)

schmidt-sebastian: Hm. I thought about this some more. If the current primary garbage collects client metadata and somehow fails to delete the local storage entry, it probably also didn't mark itself as zombied either. So the new primary would have to wait 5 seconds for the current primary lease to expire. So, all things being equal, I am just going to go ahead with this PR because it's somewhat cleaner. I will update the comment before merging.

PR #1128, opened 2018-08-16
Garbage Collect Query State

schmidt-sebastian:
This PR removes the query target metadata row from LocalStorage on the last unlisten.

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -89,7 +89,7 @@ export interface SharedClientState {
    * Called by the primary client to notify secondary clients of mutation
    * results as they come back from the backend.
    */
-  trackMutationResult(
+  updateMutationState(
     batchId: BatchId,
     state: 'acknowledged' | 'rejected',
     error?: FirestoreError
@@ -676,7 +676,7 @@ export class WebStorageSharedClientState implements SharedClientState {
     this.persistMutationState(batchId, 'pending');
   }
 
-  trackMutationResult(
+  updateMutationState(
     batchId: BatchId,
     state: 'acknowledged' | 'rejected',
     error?: FirestoreError
@@ -1086,7 +1086,7 @@ export class MemorySharedClientState implements SharedClientState {
     // No op.
   }
 
-  trackMutationResult(
+  updateMutationState(
     batchId: BatchId,
     state: 'acknowledged' | 'rejected',
     error?: FirestoreError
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -102,9 +102,12 @@ export interface SharedClientState {
    */
   addLocalQueryTarget(targetId: TargetId): QueryTargetState;
 
-  /** Removes a Query Target ID for the local Firestore clients. */
+  /** Removes the Query Target ID association from the local client. */
   removeLocalQueryTarget(targetId: TargetId): void;
 
+  /** Removes the target's metadata entry. */
+  garbageCollectQueryState(targetId: TargetId): void;
```
mikelehen: garbageCollect carries more baggage than I think is warranted (this is active cleanup at the moment the query is no longer needed rather than any sort of finding/removing of garbage).  I understand `remove` could be confused with `removeLocalQueryTarget` though.  Maybe `clearQueryState()` or `deleteQueryState`?

It's also a little tempting to do some sort of rename to provide alignment with `trackQueryUpdate()`.  But `clearQueryUpdate()` seems weird to be singular and `clearQueryUpdates()` seems misleading (since there's only one row to clear).  So maybe the rename to do would be `trackQueryUpdate()` => `trackQueryState()` but that's a little misleading too since we call it even when the state didn't change, just to notify clients that something in the query changed.  So I guess I'm not fully bought in on any of these suggestions here. :-)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -102,9 +102,12 @@ export interface SharedClientState {
    */
   addLocalQueryTarget(targetId: TargetId): QueryTargetState;
 
-  /** Removes a Query Target ID for the local Firestore clients. */
+  /** Removes the Query Target ID association from the local client. */
```
mikelehen: Can you add "Called by the primary client to ..." ?

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: I renamed `garbageCollectQueryState()` to `clearQueryState()` and `trackQueryUpdate()` to `updateQueryState()`. That gives us some sort of alignment.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Done.

PR #1131, opened 2018-08-16
 Attempt to delete client metadata row

schmidt-sebastian:
We should try to delete the client metadata row from LocalStorage even if `.shutdown()` wasn't called. Because it won't be.

### DIFF
--- a/packages/firestore/src/local/shared_client_state.ts
+++ b/packages/firestore/src/local/shared_client_state.ts
@@ -651,9 +651,7 @@ export class WebStorageSharedClientState implements SharedClientState {
 
     // Register a window unload hook to remove the client metadata entry from
     // LocalStorage even if `shutdown()` was not called.
-    this.platform.window.addEventListener("unload", () =>
-      this.shutdown()
-    );
+    this.platform.window.addEventListener('unload', () => this.shutdown());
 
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/shared_client_state.ts
```diff
@@ -752,13 +757,11 @@ export class WebStorageSharedClientState implements SharedClientState {
...
   shutdown(): void {
-    assert(
-      this.started,
-      'WebStorageSharedClientState.shutdown() called when not started'
-    );
-    this.platform.window.removeEventListener('storage', this.storageListener);
-    this.removeItem(this.localClientStorageKey);
-    this.started = false;
+    if (this.started) {
+      this.platform.window.removeEventListener('storage', this.storageListener);
```
mikelehen: Should we remove the 'unload' hook too?  (FWIW that might mean we don't need the `this.started` check, though I don't know if that's guaranteed or not... so probably fine to leave it.)

schmidt-sebastian: I'm tempted not to. It provides very little gain and adds more state keeping (I would have to keep a reference to the original function pointer for it to work consistently in all browsers).

Issue #1138, opened 2018-08-18
Can't find variable: document


Bunkerbewohner:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10 Pro
  * Browser version: React Native (Expo)
  * Firebase SDK version: 5.4.0
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

I've been using Firebase and Firestore **Web SDK** successfully with React Native and Expo. However, after a recent update it broke, producing the error message "Can't find variable: document". I have identified this code as the problem:

```typescript
export class BrowserPlatform implements Platform {
  readonly base64Available: boolean;

  readonly emptyByteString = '';

  readonly document = document;

  readonly window = window;
 // ...
```

The declarations for "document" and "window" were added with this commit:
https://github.com/firebase/firebase-js-sdk/commit/2f17ba02a3a139678b849cd53b2199156e5bc6d1#diff-f48b4026a34711ba92518fa9e6d8eaf7R30

Obviously "document" doesn't exist on the React Native / Expo platform. To fix it locally, I simply added a check in the offending class: https://github.com/Bunkerbewohner/firebase-js-sdk/commit/06436255674a6911f270fd98f7325333ed59ea14

I'd appreciate if this or a similar solution makes it into master to fix the problem.

trakout:
Also experiencing this, it's breaking UMD builds for NodeJS targets. :-1:

koraytaylan:
I wonder when this fix will be released? and how to work around it in the meantime?

Bunkerbewohner:
@koraytaylan a work around for local development is to manually fix it in `./node_modules/@firebase/firestore/dist/index.cjs.js` in line 7298:

```js
var BrowserPlatform = /** @class */ (function () {
    function BrowserPlatform() {
        this.emptyByteString = '';
        this.document = document; // delete this line
        this.window = window; // delete this line
        this.base64Available = typeof atob !== 'undefined';
    }
```

schmidt-sebastian:
Note that you can already use the version of Firestore that addresses this problem by installing the staged version of the next release: `npm install firebase@next`

PR #1139, opened 2018-08-18
handle undefined document, window variables in BrowserPlatform

Bunkerbewohner:
Simple fix for the problem described in https://github.com/firebase/firebase-js-sdk/issues/1138

### DIFF
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -27,9 +27,9 @@ export class BrowserPlatform implements Platform {
 
   readonly emptyByteString = '';
 
-  readonly document = document || null;
+  readonly document = typeof document !== "undefined" ? document : null;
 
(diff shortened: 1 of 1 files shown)


Bunkerbewohner:
I signed the CLA!

schmidt-sebastian:
Thanks for sending this over. I would like to go one step further and verify that "window" and "document" are known types before accessing them. I have seen some platforms (such as Node) that error even when checks like "window || null" are performed.

See follow up PR #1140.

Bunkerbewohner:
Ooops seems like I accidently pushed the wrong code and didn't even notice it. My bad. You are of course right that just ORing with null is not sufficient. Thanks for the follow up PR!

I updated my PR just for the sake of correctness.

schmidt-sebastian:
Thanks for updating this. I have merged my follow-up PR, which uses the same logic you have added in this PR. I will close this PR as #1140 is merged. Thanks!

PR #1140, opened 2018-08-18
Support BrowserPlatforms with no window or document

### LINKED ISSUE #1138: Can't find variable: document

schmidt-sebastian:
This expand the checks of #1139 to support platforms where merely accessing "document" or "window" is an error condition (as is the case in Node for example).

Fixes #1138 which uses ReactNative.

### DIFF
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -27,12 +27,16 @@ export class BrowserPlatform implements Platform {
 
   readonly emptyByteString = '';
 
-  readonly document = document;
-
-  readonly window = window;
-
   constructor() {
     this.base64Available = typeof atob !== 'undefined';
+  }
+
+  get document(): Document | null {
+    return typeof document !== 'undefined' ? document : null;
+  }
+
+  get window(): Window | null {
+    return typeof window !== 'undefined' ? window : null;
   }
 
   loadConnection(databaseInfo: DatabaseInfo): Promise<Connection> {


jeffreyleeon:
@wilhuff  @schmidt-sebastian  Would like to know is this change deployed? Thank you!

wilhuff:
@jeffreyleeon This change will out with the next release, which should be soon. I can't promise an exact date, but we try to release weekly.

fredrivett:
+1 on needing this deployed, seems a blocker to using Firebase with any app created with `create-react-native-app`.

Understand it's in queue and will be soon, just registering a +1 that I'm blocked on this 👍

PR #1149, opened 2018-08-20
Schema migration that drops held write acks

schmidt-sebastian:
This PR adds the schema migration that drops existing held write acks one last time. This is part of https://github.com/firebase/firebase-js-sdk/pull/1135

The migration itself is not running yet since the SchemaVersion is still set to 4.

As part of this, I have to plumb through DatabaseId to the migration code. I also started using the same Database names in all tests.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -204,15 +204,18 @@ export class IndexedDbPersistence implements Persistence {
   private queryCache: IndexedDbQueryCache;
   private remoteDocumentCache: IndexedDbRemoteDocumentCache;
 
+  private readonly persistenceKey: string;
+
   constructor(
-    private readonly persistenceKey: string,
+    private readonly databaseInfo: DatabaseInfo,
     private readonly clientId: ClientId,
     platform: Platform,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
     synchronizeTabs: boolean
   ) {
-    this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
+    this.persistenceKey = IndexedDbPersistence.buildStoragePrefix(databaseInfo);
+    this.dbName = this.persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
     this.serializer = new LocalSerializer(serializer);
     this.document = platform.document;
     this.window = platform.window;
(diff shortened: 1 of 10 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -47,14 +56,15 @@ export const SCHEMA_VERSION = 4;
...
+  databaseId: DatabaseId,
   txn: SimpleDbTransaction,
   fromVersion: number,
   toVersion: number
 ): PersistencePromise<void> {
-  assert(
-    fromVersion < toVersion && fromVersion >= 0 && toVersion <= SCHEMA_VERSION,
-    'Unexpected schema upgrade from v${fromVersion} to v{toVersion}.'
-  );
+  // assert(
```
wilhuff: Should this still be commented out?

packages/firestore/src/local/simple_db.ts
```diff
@@ -35,10 +36,12 @@ const LOG_TAG = 'SimpleDb';
 export class SimpleDb {
   /** Opens the specified database, creating or upgrading it if necessary. */
   static openOrCreate(
+    databaseId: DatabaseId,
     name: string,
     version: number,
     runUpgrade: (
       db: IDBDatabase,
+      databaseId: DatabaseId,
```
wilhuff: SimpleDb is something we have previously discussed making into a separate package from Firestore so we should try to keep Firestore-specific concepts like "databaseId" out of the SimpleDb API.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -241,7 +244,11 @@ export class IndexedDbPersistence implements Persistence {
     assert(!this.started, 'IndexedDbPersistence double-started!');
     assert(this.window !== null, "Expected 'window' to be defined");
 
-    return SimpleDb.openOrCreate(this.dbName, SCHEMA_VERSION, createOrUpgradeDb)
+    return SimpleDb.openOrCreate(
+      this.dbName,
+      SCHEMA_VERSION,
+      new SchemaConverter(this.databaseInfo.databaseId)
```
wilhuff: You could just pass the `serializer` through here in which case we wouldn't need the `databaseInfo` as a member here anymore.

schmidt-sebastian: This is done. It allowed me to revert the changes in the IndexedDbPersistence constructor.

PR #1150, opened 2018-08-20
Adding forEach Helper

schmidt-sebastian:
This adds a small PersistencePromise.forEach() helper to make loop processing easier.

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -679,20 +679,16 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
         queryView.targetId
       );
       this.limboDocumentRefs.removeReferencesForId(queryView.targetId);
-      let p = PersistencePromise.resolve();
-      limboKeys.forEach(limboKey => {
-        p = p.next(() => {
-          return this.limboDocumentRefs
-            .containsKey(null, limboKey)
-            .next(isReferenced => {
-              if (!isReferenced) {
-                // We removed the last reference for this key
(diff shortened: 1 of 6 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -679,20 +679,16 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-            .next(isReferenced => {
-              if (!isReferenced) {
-                // We removed the last reference for this key
-                this.removeLimboTarget(limboKey);
-              }
-            });
-        });
-      });
-      await p.toPromise();
+      await PersistencePromise.forEach(limboKeys.toArray(), limboKey => {
```
wilhuff: Converting to an array like this isn't cheap. Is there any typescript concept for "iterable" things?

If there's no general concept then forget about it--this isn't worth spending too much time on.

schmidt-sebastian: forEach could be:

```
  static forEach<T>(
    elements: Iterable<T>,
    callback: (T) => PersistencePromise<void>
  ): PersistencePromise<void> {
    let p = PersistencePromise.resolve();
    for (const element of elements) {
      p = p.next(() => callback(element));
    }
    return p;
  }
```

This however only works in ES6 or if change the TS-Compiler setting for "downlevelIteration": to `true`. That increases our code size by 11kB so I opted to keep things as is. 

FWIW, I think the set of document keys should be fairly small here. We could also return a an Array from from `referencesForId` (since there are no duplicate references), but that changes the return type of `LocalStore.remoteDocumentKeys`.

PR #1153, opened 2018-08-21
Make MaybeDocument an abstract class

schmidt-sebastian:
... with the class comments from Android.

### DIFF
--- a/packages/firestore/src/model/document.ts
+++ b/packages/firestore/src/model/document.ts
@@ -31,23 +31,27 @@ export interface DocumentOptions {
  * marker that this document does not exist at a given version.
  */
 export abstract class MaybeDocument {
-  readonly key: DocumentKey;
-  readonly version: SnapshotVersion;
+  constructor(readonly key: DocumentKey, readonly version: SnapshotVersion) {}
+
+  static compareByKey(d1: MaybeDocument, d2: MaybeDocument): number {
+    return DocumentKey.comparator(d1.key, d2.key);
+  }
 }
 
 /**
  * Represents a document in Firestore with a key, version, data and whether the
  * data has local mutations applied to it.
  */
-export class Document implements MaybeDocument {
+export class Document extends MaybeDocument {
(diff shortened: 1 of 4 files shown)


packages/firestore/src/model/document.ts
```diff
@@ -26,7 +26,20 @@ export interface DocumentOptions {
...
+export abstract class MaybeDocument {
+  readonly key: DocumentKey;
+  readonly version: SnapshotVersion;
+}
+
+/**
+ * Represents a document in Firestore with a key, version, data and whether the
+ * data has local mutations applied to it.
+ */
+export class Document implements MaybeDocument {
```
wilhuff: Shouldn't this be `extends` rather than `implements`?

Here and below.

packages/firestore/src/model/document.ts
```diff
@@ -68,7 +81,7 @@ export class Document {
     );
   }
 
-  static compareByKey(d1: MaybeDocument, d2: MaybeDocument): number {
+  static compareByKey(d1: Document, d2: Document): number {
```
wilhuff: Why shouldn't this accept `MaybeDocument`?

wilhuff:
CI failed to build on this error:

TS2345 Argument of type \'MaybeDocument\' is not assignable to parameter of type \'Document\'.\n  Property \'hasLocalMutations\' is missing in type \'MaybeDocument\'

packages/firestore/src/model/document.ts
schmidt-sebastian: Yeah, I would have hoped that TS complains about things like this.

packages/firestore/src/model/document.ts
schmidt-sebastian: I moved this method up to the base class and changed it to accept MaybeDocument. I just thought it was odd that it accepts MaybeDocument, but is part of a subclass.

schmidt-sebastian:
Feedback addressed. Build passing :)

PR #1156, opened 2018-08-22
Adding CommitVersion to RemoteDocumentCache

schmidt-sebastian:
This PR adds the commit version to the RemoteDocumentCache.

As part of this:
- Document now has a remoteVersion and a commitVersion
- NoDocument doesn't have a commitVersion (unlike in the held write acks PR), since we never actually raise `hasPendingWrites` for deleted documents.
- I changed MutationResult to always have a version and populate it from either the updateTime or the commitTime of the mutation.

I also cleaned up the spec test document definition and made it a type, since I didn't think adding a fifth unnamed element was a such a great idea.


packages/firestore/src/model/document.ts
```diff
@@ -31,7 +31,10 @@ export interface DocumentOptions {
  * marker that this document does not exist at a given version.
  */
 export abstract class MaybeDocument {
-  constructor(readonly key: DocumentKey, readonly version: SnapshotVersion) {}
+  constructor(
+    readonly key: DocumentKey,
+    readonly remoteVersion: SnapshotVersion
```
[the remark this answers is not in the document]
schmidt-sebastian: For documents, we need to be able to tell the `commitTime` apart from `udpateTime`. Currently, one version is stored inside api.Document and serialized as part of the document, while the other is serialized in a separate field. 

We can just store a single version and add `dirty` boolean. We could then go back and name the one version that we have `version` and make it the `readTime` for NoDocument and the last modified time for Documents (which is either the commitTime or the updateTime).

What do you think?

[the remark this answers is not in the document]
schmidt-sebastian: FYI, I pushed a commit that replaces the two different versions with a "dirty" boolean that I named `hasCommittedMutations`.

schmidt-sebastian:
PR updated with the result of our discussion:
- We store a "dirty" bit in IndexedDb.
- We store "unknown documents" in IndexedDb. These are documents that are known to exist at a specific version, but have no data. This allows us to simply some of our View processing logic.
- We store deleted documents at the time of their delete.

packages/firestore/src/model/mutation.ts
```diff
@@ -484,17 +489,19 @@ export class TransformMutation extends Mutation {
     // patch, so we use the precondition to prevent incorrectly putting a
     // partial document into our cache.
     if (!this.precondition.isValidFor(maybeDoc)) {
-      return maybeDoc;
+      return new UnknownDocument(this.key, mutationResult.version);
```
wilhuff: Same comments as above (axe the TODO and describe why this document must actually exist).

packages/firestore/src/remote/datastore.ts
```diff
@@ -87,7 +87,10 @@ export class Datastore {
       'Commit',
       params
     ).then(response => {
-      return this.serializer.fromWriteResults(response.writeResults);
+      return this.serializer.fromWriteResults(
+        response.writeResults,
+        response.commitTime
```
wilhuff: Does this need a bang for strict null checks? (i.e. `response.commitTime!`)

I ask because the assignment to `const commitVersion` uses one below.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -446,7 +460,14 @@ export class DbRemoteDocument {
...
      */
-    public document: api.Document | null
+    public document: api.Document | null,
+    /**
+     * Documents that were written to the remote document store based on
+     * a write acknowledgment are marked with `hasCommittedMutations`. These
+     * documents are potentially inconsistent with the document's version on
+     * the backend.
+     */
+    public hasCommittedMutations: boolean | undefined
```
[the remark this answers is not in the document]
wilhuff: I can see that. Agreed.

packages/firestore/src/model/mutation.ts
```diff
@@ -188,6 +202,15 @@ export class Precondition {
  * DeleteMutation     NoDocument(v3)        NoDocument(v0)
  * DeleteMutation     null                  NoDocument(v0)
  *
+ * For acknowledged mutations, we use the updateTime of the WriteResponse as
+ * the resulting version for Set, Patch, and Transform mutations. As deletes
+ * have no explicit update time, we use the commitTime of the WriteResponse for
+ * Delete mutations.
+ *
+ * If due to precondition mismatches an acknowledged mutation can't be applied
```
wilhuff: This language is hard to understand. How about:

If a mutation is acknowledged but locally fails its precondition, we return an `UnknownDocument` and rely on Watch to send us the updated version.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -315,8 +312,12 @@ function genericLocalStoreTests(
       )
       .toContain(doc('foo/bar', 0, { foo: 'bar' }, { hasLocalMutations: true }))
       .afterAcknowledgingMutation({ documentVersion: 1 })
-      .toReturnChanged(doc('foo/bar', 0, { foo: 'bar' }))
-      .toContain(doc('foo/bar', 0, { foo: 'bar' }))
+      .toReturnChanged(
```
wilhuff: Nice!

packages/firestore/test/unit/model/mutation.test.ts
```diff
@@ -478,21 +481,34 @@ describe('Mutation', () => {
...
+      'collection/key',
+      7,
+      {},
+      { hasCommittedMutations: true }
+    );
 
-    assertVersionTransitions(set, docV3, mutationResult, docV3);
-    assertVersionTransitions(set, deletedV3, mutationResult, docV0);
-    assertVersionTransitions(set, null, mutationResult, docV0);
+    assertVersionTransitions(set, docV3, mutationResult, docV7Committed);
```
wilhuff: This test essentially implements the version transitions in that big comment on top of Mutation. Could you update the comment? Alternatively, if you didn't find the table helpful, maybe we we should just delete it?

packages/firestore/src/model/mutation.ts
schmidt-sebastian: Simplified to:

```
 * If a mutation is acknowledged by the backend but fails the precondition check
 * locally, we return an `UnknownDocument` and rely on Watch to send us the
 * updated version.
```

packages/firestore/test/unit/model/mutation.test.ts
schmidt-sebastian: The test invokes `applyToRemoteDocument`, which has changed a lot and hence the changes to the test. The table in the comment is for `applyToLocalView`, which should still be the same.

schmidt-sebastian:
> Specifically:
> 
> Don't we need a migration to convert acknowledged mutations still in the mutation queue to documents with hasCommittedChanges true?

I don't think so. They will get replayed as before and create Documents with `hasLocalMutatations`. With this PR, we now have two codepaths that do the same thing.

> Don't we need to change some of the code that handles NoDocument specially? For example: Precondition.exists needs to account for this new kind of document.

I will look at that.

In general, I wasn't going to merge this PR until I know that these changes work with my main held-write-acks-removal branch.

PR #1159, opened 2018-08-22
Enable strictNullChecks

schmidt-sebastian:
Hmmm.

(BTW: This is the dumb version of this PR. There are likely asserts we could add to make sure that our assumptions are valid).

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -252,7 +252,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
           );
           const viewChange = view.applyChanges(
             viewDocChanges,
-            /* updateLimboDocuments= */ this.isPrimary,
+            /* updateLimboDocuments= */ this.isPrimary === true,
             synthesizedTargetChange
           );
           assert(
@@ -781,7 +781,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
               remoteEvent && remoteEvent.targetChanges[queryView.targetId];
             const viewChange = queryView.view.applyChanges(
               viewDocChanges,
-              /* updateLimboDocuments= */ this.isPrimary,
+              /* updateLimboDocuments= */ this.isPrimary === true,
               targetChange
             );
             return this.updateTrackedLimbos(
@@ -809,7 +809,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     });
 
     await Promise.all(queriesProcessed);
-    this.syncEngineListener.onWatchChange(newSnaps);
+    this.syncEngineListener!.onWatchChange(newSnaps);
     this.localStore.notifyLocalViewChanges(docChangesInAllViews);
     // TODO(multitab): Multitab garbage collection
     if (this.isPrimary) {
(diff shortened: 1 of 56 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -1025,7 +1026,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       );
       const query = await this.localStore.getQueryForTarget(targetId);
       assert(!!query, `Query data for active target ${targetId} not found`);
```
wilhuff: I think we could add an `assertNotNull` type guard that that would make the `!` below unnecessary.

Something like:
```ts
assertNotNull<T>(value : T | null, explanation: string): arg is T {
  assert(!!value, explanation);
  return true;
}
```

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -773,6 +776,10 @@ export class WebStorageSharedClientState implements SharedClientState {
           return;
         }
 
+        if (event.key === null) {
```
wilhuff: Shouldn't this be `==`? (We test `event.newValue != null` just below.)

packages/firestore/test/integration/api/cursor.test.ts
```diff
@@ -26,7 +26,8 @@ import {
   withTestDbs
 } from '../util/helpers';
 
-const Timestamp = firebase.firestore.Timestamp;
+const Timestamp = firebase.firestore!.Timestamp;
```
wilhuff: Is something wrong with our typings? Would our end users have to do something like this to enable strict null checks?

If not why do we need to do this?

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: I'm surprised that the linter didn't complain about `event.newValue != null`. We should not rely on type coercion, and I would guess that implicit coercion is almost always unintended. So rather than using `==` here, I changed the line below to `!==`. The type for LocalStorage are defined as "string | null", so they should never be `undefined`.

packages/firestore/test/integration/api/cursor.test.ts
schmidt-sebastian: I asked Josh about this yesterday. The way our types are defined allows us to extend the `firebase` namespace after the fact without having to add all other product's APIs as well. Note that we don't ship generated typings, but only the hand-curated types in `firestore-types` and `firebase-types`.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Since these type guards only work in if-statements, this would change the calling site to:

```
  if (assertNotNull(query, `Query data for active target ${targetId} not found`)) {
        const queryData = await this.localStore.allocateQuery(query);
        await this.initializeViewAndComputeSnapshot(
            queryData,
            /*current=*/ false
        );
        this.remoteStore.listen(queryData);
      };
```

If we didn't have to put the `explanation`, then I would be all in favor. But with the explanation this looks a tad strange. If we do drop the explanation, we will never be able to trace these asserts back in minified code.

packages/firestore/src/local/shared_client_state.ts
wilhuff: I think the issue is that `== null` is the one case where double-equal is actually useful because it handles `undefined` in a sane way. I don't actually know which we prefer, but ISTR this point being discussed.

PR #1172, opened 2018-08-28
Add sequence numbers and sync them

gsoltis:
* Add sequence numbers to `PersistenceTransaction`
 * Provide a method of syncing sequence numbers across tabs
 * Refactor schema migrations to not directly use the mutation queue
 * Refactor a little bit of `IndexedDbPersistence` so we don't continually assert the presence of platform components, and instead only do it in the constructor.

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -321,13 +321,16 @@ export class FirestoreClient {
             user
           )
         : new MemorySharedClientState();
+      const multiClientParams = settings.experimentalTabSynchronization
+        ? { sequenceNumberSyncer: this.sharedClientState }
+        : undefined;
       this.persistence = new IndexedDbPersistence(
         storagePrefix,
         this.clientId,
         this.platform,
         this.asyncQueue,
         serializer,
-        { sequenceNumberSyncer: this.sharedClientState }
+        multiClientParams
       );
       return this.persistence.start();
     });
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -51,7 +51,7 @@ import { Deferred } from '../util/promise';
 import { DatabaseId, DatabaseInfo } from './database_info';
 import { Query } from './query';
 import { Transaction } from './transaction';
-import { OnlineStateSource, ListenSequenceNumber } from './types';
+import { OnlineStateSource } from './types';
 import { ViewSnapshot } from './view_snapshot';
 import {
   MemorySharedClientState,
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -301,7 +301,7 @@ export class FirestoreClient {
       useProto3Json: true
     });
 
-    return Promise.resolve().then(() => {
+    return Promise.resolve().then(async () => {
       if (
         settings.experimentalTabSynchronization &&
         !WebStorageSharedClientState.isAvailable(this.platform)
@@ -312,27 +312,32 @@ export class FirestoreClient {
         );
       }
 
-      this.sharedClientState = settings.experimentalTabSynchronization
-        ? new WebStorageSharedClientState(
-            this.asyncQueue,
-            this.platform,
-            storagePrefix,
-            this.clientId,
-            user
-          )
-        : new MemorySharedClientState();
-      const multiClientParams = settings.experimentalTabSynchronization
-        ? { sequenceNumberSyncer: this.sharedClientState }
-        : undefined;
-      this.persistence = new IndexedDbPersistence(
-        storagePrefix,
-        this.clientId,
-        this.platform,
-        this.asyncQueue,
-        serializer,
-        multiClientParams
-      );
-      return this.persistence.start();
+      if (settings.experimentalTabSynchronization) {
+        this.sharedClientState = new WebStorageSharedClientState(
+          this.asyncQueue,
+          this.platform,
+          storagePrefix,
+          this.clientId,
+          user
+        );
+        this.persistence = await IndexedDbPersistence.createMultiClientIndexedDbPersistence(
+          storagePrefix,
+          this.clientId,
+          this.platform,
+          this.asyncQueue,
+          serializer,
+          { sequenceNumberSyncer: this.sharedClientState }
+        );
+      } else {
+        this.sharedClientState = new MemorySharedClientState();
+        this.persistence = await IndexedDbPersistence.createIndexedDbPersistence(
+          storagePrefix,
+          this.clientId,
+          this.platform,
+          this.asyncQueue,
+          serializer
+        );
+      }
     });
   }
 
@@ -345,7 +350,7 @@ export class FirestoreClient {
     this.garbageCollector = new EagerGarbageCollector();
     this.persistence = new MemoryPersistence(this.clientId);
     this.sharedClientState = new MemorySharedClientState();
-    return this.persistence.start();
+    return Promise.resolve();
   }
 
   /**
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -51,7 +51,7 @@ import { Deferred } from '../util/promise';
 import { DatabaseId, DatabaseInfo } from './database_info';
 import { Query } from './query';
 import { Transaction } from './transaction';
-import { OnlineStateSource } from './types';
+import { OnlineStateSource, ListenSequenceNumber } from './types';
 import { ViewSnapshot } from './view_snapshot';
 import {
   MemorySharedClientState,
@@ -302,16 +302,6 @@ export class FirestoreClient {
     });
 
     return Promise.resolve().then(() => {
-      const persistence: IndexedDbPersistence = new IndexedDbPersistence(
-        storagePrefix,
-        this.clientId,
-        this.platform,
-        this.asyncQueue,
-        serializer,
-        settings.experimentalTabSynchronization
-      );
-      this.persistence = persistence;
-
       if (
         settings.experimentalTabSynchronization &&
         !WebStorageSharedClientState.isAvailable(this.platform)
@@ -331,6 +321,16 @@ export class FirestoreClient {
             user
           )
         : new MemorySharedClientState();
+      const persistence: IndexedDbPersistence = new IndexedDbPersistence(
+        storagePrefix,
+        this.clientId,
+        this.platform,
+        this.asyncQueue,
+        serializer,
+        { sequenceNumberSyncer: this.sharedClientState }
+      );
+      this.persistence = persistence;
+
       return persistence.start();
     });
   }
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -321,7 +321,7 @@ export class FirestoreClient {
             user
           )
         : new MemorySharedClientState();
-      const persistence: IndexedDbPersistence = new IndexedDbPersistence(
+      this.persistence = new IndexedDbPersistence(
         storagePrefix,
         this.clientId,
         this.platform,
@@ -329,9 +329,7 @@ export class FirestoreClient {
         serializer,
         { sequenceNumberSyncer: this.sharedClientState }
       );
-      this.persistence = persistence;
-
-      return persistence.start();
+      return this.persistence.start();
     });
   }
 
--- a/packages/firestore/src/core/listen_sequence.ts
+++ b/packages/firestore/src/core/listen_sequence.ts
@@ -16,8 +16,9 @@
 
 import { ListenSequenceNumber } from './types';
 
-const LOG_TAG = 'ListenSequence';
-
+/**
+ * `SequenceNumberSyncer` defines the methods required to keep multiple instances of a `ListenSequence` in sync.
+ */
 export interface SequenceNumberSyncer {
   writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
   setSequenceNumberListener(
@@ -25,6 +26,12 @@ export interface SequenceNumberSyncer {
   ): void;
 }
 
+/**
+ * `ListenSequence` is a monotonic sequence. It is initialized with a minimum value to
+ * exceed. All subsequent calls to next will return increasing values. If provided with a
+ * `SequenceNumberSyncer`, it will additionally bump its next value when told of a new value, as well as write out
+ * sequence numbers that it produces via `next()`.
+ */
 export class ListenSequence {
   static readonly INVALID: ListenSequenceNumber = -1;
 
--- a/packages/firestore/src/core/listen_sequence.ts
+++ b/packages/firestore/src/core/listen_sequence.ts
@@ -1,0 +1,62 @@
+/**
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ListenSequenceNumber } from './types';
+
+const LOG_TAG = 'ListenSequence';
+
+export interface SequenceNumberSyncer {
+  writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+  setSequenceNumberListener(
+    cb: (sequenceNumber: ListenSequenceNumber) => void
+  ): void;
+}
+
+export class ListenSequence {
+  static readonly INVALID: ListenSequenceNumber = -1;
+
+  private writeNewSequenceNumber?: (
+    newSequenceNumber: ListenSequenceNumber
+  ) => void;
+
+  constructor(
+    private previousValue: ListenSequenceNumber,
+    syncParams?: SequenceNumberSyncer
+  ) {
+    if (syncParams) {
+      syncParams.setSequenceNumberListener(sequenceNumber =>
+        this.setPreviousValue(sequenceNumber)
+      );
+      this.writeNewSequenceNumber = sequenceNumber =>
+        syncParams.writeSequenceNumber(sequenceNumber);
+    }
+  }
+
+  private setPreviousValue(
+    externalPreviousValue: ListenSequenceNumber
+  ): ListenSequenceNumber {
+    this.previousValue = Math.max(externalPreviousValue, this.previousValue);
+    return this.previousValue;
+  }
+
+  next(): ListenSequenceNumber {
+    const nextValue = ++this.previousValue;
+    if (this.writeNewSequenceNumber) {
+      this.writeNewSequenceNumber(nextValue);
+    }
+    return nextValue;
+  }
+}
--- a/packages/firestore/src/core/listen_sequence.ts
+++ b/packages/firestore/src/core/listen_sequence.ts
@@ -21,7 +21,9 @@ import { ListenSequenceNumber } from './types';
  */
 export interface SequenceNumberSyncer {
   updateSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
-  sequenceNumberHandler: ((sequenceNumber: ListenSequenceNumber) => void) | null;
+  sequenceNumberHandler:
+    | ((sequenceNumber: ListenSequenceNumber) => void)
+    | null;
 }
 
 /**
@@ -42,7 +44,8 @@ export class ListenSequence {
     sequenceNumberSyncer?: SequenceNumberSyncer
   ) {
     if (sequenceNumberSyncer) {
-      sequenceNumberSyncer.sequenceNumberHandler = sequenceNumber => this.setPreviousValue(sequenceNumber);
+      sequenceNumberSyncer.sequenceNumberHandler = sequenceNumber =>
+        this.setPreviousValue(sequenceNumber);
       this.writeNewSequenceNumber = sequenceNumber =>
         sequenceNumberSyncer.updateSequenceNumber(sequenceNumber);
     }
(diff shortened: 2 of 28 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -331,6 +321,16 @@ export class FirestoreClient {
             user
           )
         : new MemorySharedClientState();
+      const persistence: IndexedDbPersistence = new IndexedDbPersistence(
```
schmidt-sebastian: It looks like this can directly be assigned to `this.persistence` now. Before, we had a custom `start()` method and hence I added that indirection. This no longer needs to be typed as IndexedDbPersistence.

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+import { ListenSequenceNumber } from './types';
+
+const LOG_TAG = 'ListenSequence';
+
+export interface SequenceNumberSyncer {
+  writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+  setSequenceNumberListener(
```
schmidt-sebastian: We don't expose these setters in the other interfaces we have, leading me to think that we can likely drop it from this interface as well.

The name also suggests that there will be a receiving function.  I assume that's coming in a follow-up.

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,62 @@
...
+    private previousValue: ListenSequenceNumber,
+    syncParams?: SequenceNumberSyncer
+  ) {
+    if (syncParams) {
+      syncParams.setSequenceNumberListener(sequenceNumber =>
+        this.setPreviousValue(sequenceNumber)
+      );
+      this.writeNewSequenceNumber = sequenceNumber =>
+        syncParams.writeSequenceNumber(sequenceNumber);
+    }
```
schmidt-sebastian: Optional: If you plan to add more methods, I would suggest just storing the reference to the SequenceNumberSyncer.

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,62 @@
...
+export class ListenSequence {
+  static readonly INVALID: ListenSequenceNumber = -1;
+
+  private writeNewSequenceNumber?: (
+    newSequenceNumber: ListenSequenceNumber
+  ) => void;
+
+  constructor(
+    private previousValue: ListenSequenceNumber,
+    syncParams?: SequenceNumberSyncer
```
schmidt-sebastian: Nit: `params` suggests that this the syncing can be customized (such as the interval of syncs, etc).

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,62 @@
...
+  private setPreviousValue(
+    externalPreviousValue: ListenSequenceNumber
+  ): ListenSequenceNumber {
+    this.previousValue = Math.max(externalPreviousValue, this.previousValue);
+    return this.previousValue;
+  }
+
+  next(): ListenSequenceNumber {
+    const nextValue = ++this.previousValue;
+    if (this.writeNewSequenceNumber) {
```
schmidt-sebastian: What the use case for not setting it?

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -598,6 +578,58 @@ export class IndexedDbMutationQueue implements MutationQueue {
   }
 }
 
+/**
+ * Delete a mutation batch and the associated document mutations.
+ * @return done, A PersistencePromise that signals when the deletion has completed, and removedDocuments, the document
```
schmidt-sebastian: Nit: can you reflow this to either 80 or 100 characters? Pick one of your choice. I think we are pretty inconsistent already.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -598,6 +578,58 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+      mutation.key.path,
+      batch.batchId
+    );
+    promises.push(indexTxn.delete(indexKey));
+    removedDocuments.push(mutation.key);
+  }
+  return {
+    done: PersistencePromise.waitFor(promises),
+    removedDocuments
+  };
```
schmidt-sebastian: Isn't this just `return PersistencePromise.waitFor(promises).then(() => removedDocuments)`?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,25 +213,40 @@ export class IndexedDbPersistence implements Persistence {
...
   constructor(
     private readonly persistenceKey: string,
     private readonly clientId: ClientId,
     platform: Platform,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    synchronizeTabs: boolean
+    private readonly multiClientParams?: {
+      sequenceNumberSyncer: SequenceNumberSyncer;
+    }
```
schmidt-sebastian: Suggestion: Can we always pass this and have a no-op implementation?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,25 +213,40 @@ export class IndexedDbPersistence implements Persistence {
...
   ) {
     this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
     this.serializer = new LocalSerializer(serializer);
     this.document = platform.document;
     this.window = platform.window!;
-    this.allowTabSynchronization = synchronizeTabs;
+    if (multiClientParams !== undefined) {
```
schmidt-sebastian: Hm. Not sure I am big fan since this used to be very explicit and is now very implicit.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,25 +213,40 @@ export class IndexedDbPersistence implements Persistence {
...
+    } else {
+      this.allowTabSynchronization = false;
+    }
     this.queryCache = new IndexedDbQueryCache(this.serializer);
     this.remoteDocumentCache = new IndexedDbRemoteDocumentCache(
       this.serializer,
       /*keepDocumentChangeLog=*/ this.allowTabSynchronization
     );
+    this.webStorage = this.window.localStorage;
+    if (!this.webStorage) {
```
schmidt-sebastian: Since we mock WebStorage now as well, can you verify if we still need this?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -259,6 +282,25 @@ export class IndexedDbPersistence implements Persistence {
...
         );
       })
+      .then(() => {
+        return this.simpleDb.runTransaction(
+          'readonly',
+          [DbTargetGlobal.store],
+          txn => {
+            return getHighestListenSequenceNumber(txn).next(
+              highestListenSequenceNumber => {
+                const sequenceNumberSyncer = this.multiClientParams
```
schmidt-sebastian: This would just be `this. sequenceNumberSyncer` if you changed the constructor to awlays take a sequence number syncer and provided a NoOp syncer.

... on second thought: What's the side effect of always syncing the sequence numbers? Multi-Tab is optional because it might lead to corruption in IndexedDb if there is a bug/race. If the side effect of always using the synchronization code is that we are writing to LocalStorage, then we could consider always having this enabled.

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -385,6 +385,27 @@ function globalTargetStore(
...
+    txn,
+    DbTargetGlobal.store
+  );
+  return globalStore.get(DbTargetGlobal.key).next(metadata => {
+    assert(metadata !== null, 'Missing metadata row.');
+    return metadata!;
+  });
+}
+
+export function getHighestListenSequenceNumber(
```
schmidt-sebastian: These exported global functions don't port well and "pollute" the global namespace. Where possible, can you attach them to a class, similar to what I assume will be the case in Android/iOS?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -143,15 +137,11 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
                 `Cannot process batch ${dbBatch.batchId} from unexpected user`
               );
               const batch = this.serializer.fromDbMutationBatch(dbBatch);
-              return mutationQueue.removeMutationBatch(
-                indexedDbTransaction,
-                batch
-              );
+
+              const { done } = removeMutationBatch(txn, queue.userId, batch);
+              return done;
```
schmidt-sebastian: Thanks! 

(and this will be one line less if you used return the Promise)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -168,6 +172,11 @@ export interface SharedClientState {
 
   /** Changes the shared online state of all clients. */
   setOnlineState(onlineState: OnlineState): void;
+
+  writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+  setSequenceNumberListener(
+    cb: (sequenceNumber: ListenSequenceNumber) => void
+  ): void;
```
schmidt-sebastian: Please take a look at how the remaining "syncers" are set up in the top of this interface.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -633,6 +645,16 @@ export class WebStorageSharedClientState implements SharedClientState {
...
   }
 
+  setSequenceNumberListener(
+    cb: (sequenceNumber: ListenSequenceNumber) => void
+  ): void {
+    this.sequenceNumberListener = cb;
+  }
+
+  writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void {
+    this.setItem(SEQUENCE_NUMBER_KEY, JSON.stringify(sequenceNumber));
```
schmidt-sebastian: JSON.stringify(sequenceNumber) -> String(sequenceNumber) ?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -830,6 +852,13 @@ export class WebStorageSharedClientState implements SharedClientState {
               return this.handleOnlineStateEvent(onlineState);
             }
           }
+        } else if (event.key === SEQUENCE_NUMBER_KEY) {
+          if (this.sequenceNumberListener) {
+            const sequenceNumber = parseSequenceNumber(event.newValue);
```
schmidt-sebastian: Nit: Can you follow the existing naming conventions in this file (`this. fromLocalStorageOnlineState` for example).

packages/firestore/test/unit/core/listen_sequence.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { expect } from 'chai';
+import { fail } from 'assert';
+import { ListenSequence } from '../../../src/core/listen_sequence';
+import { ListenSequenceNumber } from '../../../src/core/types';
+
+type SequenceNumberCallback = (sequenceNumber: ListenSequenceNumber) => void;
+
+describe('ListenSequence', () => {
+  it('writes the new sequence number to local storage', () => {
```
schmidt-sebastian: Nit: These tests verify the interactions of the `syncer` interface, but don't use LocalStorage. Care to rename the descriptions? :)

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -598,6 +578,58 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
+/**
+ * Delete a mutation batch and the associated document mutations.
+ * @return done, A PersistencePromise that signals when the deletion has completed, and removedDocuments, the document
+ *         mutations that were removed.
+ */
+export function removeMutationBatch(
+  txn: SimpleDbTransaction,
+  userId: string,
+  batch: MutationBatch
+): { removedDocuments: DocumentKey[]; done: PersistencePromise<void> } {
```
zxu123: Should this match the order in return? `done` is before `removedDocuments` in return.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -874,29 +921,25 @@ export class IndexedDbPersistence implements Persistence {
...
+            isZombied ? 'is' : 'is not'
+          } zombied in LocalStorage`
+        );
+        return isZombied;
+      } catch (e) {
+        // Gracefully handle if LocalStorage isn't working.
+        log.error(LOG_TAG, 'Failed to get zombied client id.', e);
+        return false;
+      }
+    } else {
```
zxu123: tl;dr: return early for trivial cases.

Alternatively, you can check the negation of this if-condition and move the return early and move the try-statement out of the if. Does not matter for the code itself but will help human understanding the code.

packages/firestore/src/core/listen_sequence.ts
gsoltis: If we haven't enabled multi-tab, there's no reason to sync sequence numbers across tabs.

packages/firestore/src/core/listen_sequence.ts
gsoltis: changed to `sequenceNumberSyncer`.

packages/firestore/src/core/firestore_client.ts
gsoltis: Done.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: As discussed, let's not always write to local storage.

packages/firestore/src/local/indexeddb_mutation_queue.ts
gsoltis: Oh, yeah, I like that better. Done.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: Verified that the mock persistence works. I can't test browsers locally anymore, so i'm delegating that to travis

packages/firestore/src/local/shared_client_state.ts
gsoltis: We use `JSON.parse`, presumably we should use `JSON.stringify` to ensure it's compatible.

packages/firestore/src/local/shared_client_state.ts
gsoltis: Sure, but there doesn't seem to be a good reason to make this a method rather than just a function. It doesn't rely on any state from the instance.

gsoltis: Renamed to `fromLocalStorageSequenceNumber`

packages/firestore/test/unit/core/listen_sequence.test.ts
gsoltis: Done.

packages/firestore/src/core/listen_sequence.ts
gsoltis: Ah, ok, it looks like we expose writeable properties in the interface? I'll do that then.

packages/firestore/src/local/indexeddb_mutation_queue.ts
gsoltis: Rewrote the comment. But, we should configure prettier to do whatever is the appropriate thing here, it's a silly thing for humans to do.

packages/firestore/src/local/indexeddb_mutation_queue.ts
gsoltis: This has been refactored away, but in general I don't think it matters. Objects are unordered.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: This has also been refactored away.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: I'm not wild about that, especially if the `synchronizeTabs` flag is still present. It lets callers specify invalid configurations, mismatching the flag and the params.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: As discussed, this will eventually become mandatory, and the flag will go away. If we can come up with a way to not allow invalid configurations while also being more explicit, I'm happy to implement it. But, I would like to avoid being able to specify invalid configurations.

On a related note, I would like to eventually get rid of external callers calling the constructor followed immediately by `start()`. I think that would be a good time to switch to static methods that return a `Promise<IndexedDbPersistence>`, and it would be reasonable to name one of them such that it's obviously creating a multi-client instance.

packages/firestore/src/local/shared_client_state.ts
schmidt-sebastian: Two reasons to make this a method:

- It namespaces the method to the class that deals with Local Storage. You can then drop `LocalStorage` from the name, and it is immediately obvious that this method is only useful to one of the two classes in this file.
- It's consistent with the rest of the file. The new name is already a big step in that direction (thanks!)

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -56,6 +58,8 @@ const QUERY_TARGET_KEY_PREFIX = 'firestore_targets';
 
 // The LocalStorage key that stores the primary tab's online state.
 const ONLINE_STATE_KEY = 'firestore_online_state';
+// The LocalStorage key that stores the last sequence number allocated.
+const SEQUENCE_NUMBER_KEY = 'firestore_sequence_number';
```
schmidt-sebastian: If I am not mistaken, the `SEQUENCE_NUMBER_KEY` should include the persistence key prefix, so that it is only shared within a project? While you think about that, I'll figure out if I have to do the same thing for the ONLINE_STATE_KEY.

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * `SequenceNumberSyncer` defines the methods required to keep multiple instances of a `ListenSequence` in sync.
+ */
+export interface SequenceNumberSyncer {
+  updateSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+  sequenceNumberHandler:
+    | ((sequenceNumber: ListenSequenceNumber) => void)
+    | null;
+}
```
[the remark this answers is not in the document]
gsoltis: I'm confused. Do the existing `Syncer` classes not also write out information to local storage? That is what `updateSequenceNumber` does.

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor(
+    private previousValue: ListenSequenceNumber,
+    sequenceNumberSyncer?: SequenceNumberSyncer
+  ) {
+    if (sequenceNumberSyncer) {
+      sequenceNumberSyncer.sequenceNumberHandler = sequenceNumber =>
+        this.setPreviousValue(sequenceNumber);
+      this.writeNewSequenceNumber = sequenceNumber =>
+        sequenceNumberSyncer.updateSequenceNumber(sequenceNumber);
```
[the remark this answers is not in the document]
gsoltis: No, since we don't get events for local writes to local storage. Other tabs will get `setPreviousValue` called.

gsoltis:
Github won't let me reply to the function comment, so: the function is already namespaced. It is part of a module. Modules are the namespacing mechanism that javascript gives us, and I don't think we additionally need to stuff something into a class to get an extra level of indirection. This goes double for something that's not even exported. Even java would just have a static method for this sort of utility. If there are a bunch of other parsing methods that don't use any state from their instance, they should be moved out of the class. See, for example, `persistence_test_helpers` or the various `util` functions.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,41 +249,56 @@ export class IndexedDbPersistence implements Persistence {
...
     this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
     this.serializer = new LocalSerializer(serializer);
     this.document = platform.document;
-    this.window = platform.window!;
-    this.allowTabSynchronization = synchronizeTabs;
+    if (multiClientParams !== undefined) {
+      this.allowTabSynchronization = true;
+    } else {
+      this.allowTabSynchronization = false;
+    }
```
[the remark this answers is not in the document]

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -205,41 +249,56 @@ export class IndexedDbPersistence implements Persistence {
...
     this.queryCache = new IndexedDbQueryCache(this.serializer);
     this.remoteDocumentCache = new IndexedDbRemoteDocumentCache(
       this.serializer,
       /*keepDocumentChangeLog=*/ this.allowTabSynchronization
     );
+    if (!!platform.window && !!platform.window.localStorage) {
```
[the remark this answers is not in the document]

packages/firestore/src/local/shared_client_state.ts
gsoltis: Hmm, yeah, probably should be scoped to a project. I'll fix it.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -840,6 +861,15 @@ export class WebStorageSharedClientState implements SharedClientState {
               return this.handleOnlineStateEvent(onlineState);
             }
           }
+        } else if (event.key === this.sequenceNumberKey) {
+          if (this.sequenceNumberHandler) {
```
schmidt-sebastian: If I am understanding this correctly, this should always be set at this point. Can we change this to an assert?

packages/firestore/src/core/listen_sequence.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ListenSequenceNumber } from './types';
+
+/**
+ * `SequenceNumberSyncer` defines the methods required to keep multiple instances of a `ListenSequence` in sync.
+ */
+export interface SequenceNumberSyncer {
+  writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+  sequenceNumberHandler:
+    | ((sequenceNumber: ListenSequenceNumber) => void)
+    | null;
```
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/shared_client_state.ts
gsoltis: Done.

PR #1179, opened 2018-08-30
Remove start() from Persistence interface

gsoltis:
This PR:
 * removes `start()` from the `Persistence` interface. We always immediately called it after instantiating the persistence layer.
 * Makes the constructor for `IndexedDbPersistence` private.
 * Provides static factory methods on `IndexedDbPersistence`. They return a `Promise` of a fully started instance. The two methods also differentiate between multi-client and not. They can be consolidated when multi-client becomes required for disk persistence.
 * Switch instantiations to use static helpers.

The intention is that if you are able to obtain an instance of `Persistence`, then it is useable. Callers should not have access to unstarted instances.

Note that this PR merges into the `sequence_number` branch to resolve some issues that came up in that PR.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -299,7 +299,7 @@ export class IndexedDbPersistence implements Persistence {
    *
    * @return {Promise<void>} Whether persistence was enabled.
    */
-  start(): Promise<void> {
+  private start(): Promise<void> {
     assert(!this.started, 'IndexedDbPersistence double-started!');
     assert(this.window !== null, "Expected 'window' to be defined");
 
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -156,6 +156,9 @@ export class IndexedDbTransaction extends PersistenceTransaction {
  * TODO(multitab): Remove `experimentalTabSynchronization` section when
  * multi-tab is no longer optional.
  */
+type MultiClientParams = {
+  sequenceNumberSyncer: SequenceNumberSyncer;
+};
 export class IndexedDbPersistence implements Persistence {
   static getStore<Key extends IDBValidKey, Value>(
     txn: PersistenceTransaction,
@@ -200,9 +203,7 @@ export class IndexedDbPersistence implements Persistence {
     platform: Platform,
     queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    multiClientParams: {
-      sequenceNumberSyncer: SequenceNumberSyncer;
-    }
+    multiClientParams: MultiClientParams
   ): Promise<IndexedDbPersistence> {
     const persistence = new IndexedDbPersistence(
       persistenceKey,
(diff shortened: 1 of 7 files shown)


packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -106,20 +106,27 @@ async function withCustomPersistence(
...
-    TEST_PERSISTENCE_PREFIX,
-    clientId,
-    platform,
-    queue,
-    serializer,
-    multiClientParams
-  );
+  if (settings.experimentalTabSynchronization) {
+  } else {
+  }
```
schmidt-sebastian: Hm, I didn't see you add a test for this part of the PR :)

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -176,6 +176,46 @@ export class IndexedDbPersistence implements Persistence {
...
+  }
+
+  static async createMultiClientIndexedDbPersistence(
+    persistenceKey: string,
+    clientId: ClientId,
+    platform: Platform,
+    queue: AsyncQueue,
+    serializer: JsonProtoSerializer,
+    multiClientParams: {
+      sequenceNumberSyncer: SequenceNumberSyncer;
```
schmidt-sebastian: Optional: You could create a type for this since you are using it in three (?) places now.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -612,23 +616,19 @@ describe('IndexedDb: allowTabSynchronization', () => {
...
-  it('rejects access when synchronization is disabled', () => {
-    return withPersistence('clientA', async db1 => {
-      await db1.start();
-      await withPersistence('clientB', async db2 => {
-        await expect(db2.start()).to.eventually.be.rejectedWith(
-          'There is another tab open with offline persistence enabled.'
-        );
-      });
+  it('rejects access when synchronization is disabled', async () => {
+    await withPersistence('clientA', async db1 => {
```
schmidt-sebastian: Nit: This `await` is inconsistent with the rest of the surrounding tests.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1156,9 +1156,7 @@ class MemoryTestRunner extends TestRunner {
   protected async initPersistence(
     serializer: JsonProtoSerializer
   ): Promise<Persistence> {
-    const persistence = new MemoryPersistence(this.clientId);
-    await persistence.start();
-    return persistence;
+    return new MemoryPersistence(this.clientId);
```
schmidt-sebastian: This is kind of weird. Care to make the function not async and use a `Promise.resolve()` here?

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
gsoltis: The previous version had a bunch of `await`s...

packages/firestore/test/unit/specs/spec_test_runner.ts
gsoltis: Sure.

packages/firestore/src/local/indexeddb_persistence.ts

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
gsoltis: Obviously, it's trivially bug-free.

But since I don't want to write a test for it, I'll delete it.

PR #1183, opened 2018-08-31
Don't parse RegExps twice

schmidt-sebastian:
This changes SharedClientState to only parse the regular expression once per Local Storage key. Note that clientStateKeyRe, since it requires special handling for the `null` case.


gsoltis:
I think this is pushing more complexity into the value-parsing methods, when in at least few cases, we can get rid of them altogether. Here's an example of what I mean: https://github.com/firebase/firebase-js-sdk/compare/gsoltis/refactor_regex_usage?expand=1

PR #1184, opened 2018-08-31
Shallowing up the Typescript surface for firebase/app and exporting default

jamesdaniels:
`import firebase from 'firebase/app'` fails for Typescript developers, but is needed for correct behavior on both browser and Node as `import * as firebase from 'firebase/app'` ends up with an entirely different object in Node. 

The core of the matter is the types as they stand [suggest that `firebase/app` is a CJS/AMD](export--and-import--require), when we're actually an ESM+CJS with a default export. Therefore shallowing up, exporting each namespace individually, and exporting a default namespace is more correct.

As far as I'm aware this is not a breaking change.


jamesdaniels:
@jshcrowthe @davideast I'm not sure if this is 100% the correct / complete answer, but this is how I was able to get things working well locally. Happy to discuss further.

PR #1194, opened 2018-09-04
Add sequence number to QueryData

gsoltis:
Port of https://github.com/FirebasePrivate/firebase-android-sdk/pull/155

Adds sequence numbers to `QueryData` and starts tracking the highest value seen.


packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -86,6 +86,14 @@ export class IndexedDbQueryCache implements QueryCache {
     });
   }
 
+  getHighestSequenceNumber(
+    transaction: PersistenceTransaction
+  ): PersistencePromise<ListenSequenceNumber> {
+    return getHighestListenSequenceNumber(
+      (transaction as IndexedDbTransaction).simpleDbTransaction
```
schmidt-sebastian: I think that you can get rid of this cast if you change retrieveMetadata to use an `PersistenceTransaction` and `IndexedDbPersistence.getStore`.

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -81,7 +92,9 @@ export class MemoryQueryCache implements QueryCache {
     if (targetId > this.highestTargetId) {
       this.highestTargetId = targetId;
     }
-    // TODO(GC): track sequence number
+    if (queryData.sequenceNumber > this.highestSequenceNumber) {
```
schmidt-sebastian: You don't save the sequence number in the Android port here or in the IndexedDb version. I can't convince myself which one is correct, so I'll let you do that for me.

packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -93,11 +93,11 @@ export async function testIndexedDbPersistence(
   options: {
     dontPurgeData?: boolean;
     synchronizeTabs?: boolean;
-  } = {}
+  } = {},
+  prefix = `${TEST_PERSISTENCE_PREFIX}/`
```
schmidt-sebastian: Nit: What do you think of this signature?

```
 testIndexedDbPersistence(
  options: {	  options: {
    dontPurgeData?: boolean;	   
    synchronizeTabs?: boolean;	  
    prefix?: string (or ` = ${TEST_PERSISTENCE_PREFIX}/`)
  } = {}} = {},
```

Just my personal preference, so feel free to ignore.

packages/firestore/test/unit/local/query_cache.test.ts
```diff
@@ -51,6 +52,50 @@ describe('IndexedDbQueryCache', () => {
   });
 
   genericQueryCacheTests(() => persistencePromise);
+
+  it('persists metadata across restarts', async () => {
+    await (await persistencePromise).shutdown(/* deleteData= */ true);
+    const testPrefix = 'test-queryCache-restarts';
+    const db1 = await persistenceHelpers.testIndexedDbPersistence(
+      {},
+      testPrefix
```
schmidt-sebastian: Actually, upon further thought, do we need to specify a custom prefix here? Isn't `{ dontPurgeData: true }` enough?

packages/firestore/test/util/helpers.ts
```diff
@@ -246,7 +246,14 @@ export function queryData(
   queryPurpose: QueryPurpose,
   path: string
 ): QueryData {
-  return new QueryData(query(path)._query, targetId, queryPurpose);
+  // Arbitrary value.
+  const sequenceNumber = 0;
```
schmidt-sebastian: Does it make sense to use something other than `0` here for future debugging help?

packages/firestore/src/local/indexeddb_query_cache.ts
gsoltis: `IndexedDbPersistence` also calls `getHighestListenSequenceNumber` from a context in which a `PersistenceTransaction` does not make sense, but a `SimpleDbTRansaction` does. Namely, the `start()` method, where there is no sequence number and no visible-external-to-persistence-layer transaction.

packages/firestore/test/unit/local/persistence_test_helpers.ts
gsoltis: Reverted, see below.

packages/firestore/test/unit/local/query_cache.test.ts
gsoltis: Yeah, I tried that initially but must've messed something up because it nows appears that it works.

PR #1197, opened 2018-09-05
Offline get() improvements.

mikelehen:
[Roll-forward of c7e1ec9 (which was reviewed, submitted, and then reverted due to regression). The regression fix and added tests are in 3be0938]

1. Changed MAX_WATCH_STREAM_FAILURES to 1 per conversation with @wilhuff and
   @jdimond.
2. Fixed a "race" where when a get() triggered a watch stream error, we'd
   restart the stream, then get() would remove its listener, and so we had no
   listener left once the watch stream was "opened". Without a listen to send
   we'd be stuck in Offline state (even though we should be Unknown whenever
   we have no listens to send), resulting in the next get() returning data
   from cache without even attempting to reach the backend.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,4 +1,10 @@
-# 0.7.0 (Unreleased)
+# 0.7.1 (Unreleased)
+- [fixed] Fixed an issue where the first `get()` call made after being offline
+  could incorrectly return cached data without attempting to reach the backend.
+- [changed] Changed `get()` to only make 1 attempt to reach the backend before
+  returning cached data, potentially reducing delays while offline.
+
+# 0.7.0
 - [fixed] Fixed `get({source: 'cache'})` to be able to return nonexistent
   documents from cache.
 - [changed] Prepared the persistence layer to allow shared access from multiple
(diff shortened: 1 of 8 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-# Unreleased
+# 0.7.3 (Unreleased)
+- [fixed] Fixed an issue where the first `get()` call made after being offline
+  could incorrectly return cached data without attempting to reach the backend.
+- [changed] Changed `get()` to only make 1 attempt to reach the backend before
+  returning cached data, potentially reducing delays while offline.
```
wilhuff: Is it worth noting here that the two attempts mechanism was a work around for a specific backend bug that's no longer in effect?

packages/firestore/test/unit/specs/remote_store_spec.test.ts
```diff
@@ -77,6 +77,7 @@ describeSpec('Remote store:', [], () => {
         // Close before we get an ack, this should reset our pending
         // target counts.
         .watchStreamCloses(Code.UNAVAILABLE)
+        .expectEvents(query, { fromCache: true })
         // This should work now.
```
wilhuff: I don't understand what this comment means anymore now that we've raised the event above.

PR #1198, opened 2018-09-05
Update CHANGELOG.md

schmidt-sebastian:
Firestore change log for held write acks change.


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-# Unreleased
+# Unreleased (0.8.1)
+- [changed] The internal handling for locally updated documents that haven't yet
```
wilhuff: Most of these notes start with the verb (i.e. "Changed" in this case).

If possible it would be worthwhile to vaguely describe what the differences might be. Something like: `SnapshotMetadata.hasPendingWrites` may now be cleared earlier than it used to (or whatever the actual change is).

One key difference here is that after restart an acknowledged but unobserved document will have hasPendingWrites: false but fromCache: true. Should we make note that users wanting to make sure documents are non-dirty should check for either condition?

PR #1199, opened 2018-09-05
Expose multi-tab flag

schmidt-sebastian:
As per Josh: Don't merge until Friday to make sure it doesn't go out this week.

### DIFF
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -57,26 +57,25 @@ export interface Settings {
   timestampsInSnapshots?: boolean;
 }
 
-// TODO(multitab): Uncomment when multi-tab is released publicly.
-// /**
-//  * Settings that can be passed to Firestore.enablePersistence() to configure
-//  * Firestore persistence.
-//  */
-// export interface PersistenceSettings {
-//   /**
-//    * Whether to synchronize the in-memory state of multiple tabs. Setting this
-//    * to 'true' in all open tabs enables shared access to local persistence,
-//    * shared execution of queries and latency-compensated local document updates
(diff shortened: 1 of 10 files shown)


mikelehen:
What about the remaining TODOs?

```
~/src/firebase-js-sdk/packages/firestore (mrschmidt-multitabflag) $ grep -ir "todo(multi" src test
src/core/sync_engine.ts:      // TODO(multitab): Consider calling this only in the primary tab.
src/local/indexeddb_persistence.ts: * TODO(multitab): Remove `experimentalTabSynchronization` section when
src/local/indexeddb_persistence.ts:              // TODO(multitab): Remove this check when `synchronizeTabs` can no
src/local/indexeddb_persistence.ts:    // TODO(multitab): Consider removing `requirePrimaryLease` and exposing
src/local/indexeddb_persistence.ts:  // TODO(multitab): Remove this check when `synchronizeTabs` can no longer be
src/local/shared_client_state.ts:// TODO(multitab): Rename all usages of LocalStorage to WebStorage to better differentiate from LocalClient.
src/local/shared_client_state_syncer.ts:  // TODO(multitab): Consider different names for these methods that convey
src/local/indexeddb_remote_document_cache.ts:    // TODO(multitab): Another client may have garbage collected the remote
test/unit/specs/spec_test_runner.ts:    // TODO(multitab): Refactor so this is only executed after primary tab
```

I think we should either fix them or turn them into bugs or TODO(schmidt-sebastian) comments.

schmidt-sebastian:
> I think we should either fix them or turn them into bugs or TODO(schmidt-sebastian) comments.

Done. Filed bugs or assigned to myself.

PR #1201, opened 2018-09-05
Remove TODOs

schmidt-sebastian:
This PR removes two TODOs for things that I'd rather keep as is:

-  Consider removing `requirePrimaryLease` and exposing three different write modes (readonly, readwrite, readwrite_primary).

While that is a good idea per se, we don't expose the write mode in `IndexedDbPersistence.runTransaction`.

- SharedClientStateSyncer method renames.

I can't think of a better name that isn't super awkward. `applyBatchFromPersistence`/`synchronizeBatchWithPersistence` all imply things that are only 90% true, since these methods combine state from WebStorage and Persistence.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -717,8 +717,6 @@ export class IndexedDbPersistence implements Persistence {
       transaction: PersistenceTransaction
     ) => PersistencePromise<T>
   ): Promise<T> {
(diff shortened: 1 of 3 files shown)


mikelehen:
Per chat, please reconsider the requirePrimaryLease change.  I'm okay with removing the other TODO.

schmidt-sebastian:
> Per chat, please reconsider the requirePrimaryLease change. I'm okay with removing the other TODO.

Merged the other PR. This now removes the TODO.

Issue #1188, opened 2018-09-01
Type mismatch between Query#on and Query#off


mayhewluke:
### [REQUIRED] Describe your environment

  * Operating System version: N/A
  * Browser version: N/A
  * Firebase SDK version: 5.4.2
  * Firebase Product: database (auth, database, storage, etc)

#### Steps to reproduce:

`.off` is supposed to be able to take a function that has been used with `.on`. However, pass a function used with `.on` to `.off` and it throws a type error. If I had to hazard a guess, I would guess that this is because the type definitions of the callback functions that `.on` and `.off` accept are completely different.

#### Relevant Code:

```typescript
import firebase from "firebase";

const onFn = firebase
  .database()
  .ref("foo")
  .on("value", snapshot => snapshot);

firebase
  .database()
  .ref("foo")
  .off("value", onFn);
```

This produces an error, complaining about the `onFn` in the last line:

```
error TS2345: Argument of type '(a: DataSnapshot | null, b?: string | undefined) => any' is not assignable to parameter of type '(a: DataSnapshot, b?: string | null | undefined) => any'.
  Types of parameters 'b' and 'b' are incompatible.
    Type 'string | null | undefined' is not assignable to type 'string | undefined'.
      Type 'null' is not assignable to type 'string | undefined'.
```

Even if trying to specify types through explicit type declarations, or including the optional second parameter in the callback function, etc. it still produces varying errors to do with the types not matching:

[Definition of `.on`](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L510):

```typescript
    on(
      eventType: EventType,
      callback: (a: firebase.database.DataSnapshot | null, b?: string) => any,
      cancelCallbackOrContext?: Object | null,
      context?: Object | null
    ): (a: firebase.database.DataSnapshot | null, b?: string) => any;
```

[Definition of `.off`](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L505):

```typescript
    off(
      eventType?: EventType,
      callback?: (a: firebase.database.DataSnapshot, b?: string | null) => any,
      context?: Object | null
    ): any;
```

Specifically notice the lack of `| null` in the first argument in `.off`, and simultaneously the omission of `| null` in the second argument in `.on` (even though [the docs](https://firebase.google.com/docs/reference/js/firebase.database.Reference#on) say it may be passed `null`).

[...]

PR #1204, opened 2018-09-05
Fix the type annotations in the RTDB Query.on/off/once API.

schmidt-sebastian:
This fixes the nullability annotations in the RTDB Query.on/off API. The callback can be optional, but not the DataSnapshot. This is a breaking typing cleanup (as some types used to take "null"), but this updates the TS types to match the JS reference docs.

Fixes: https://github.com/firebase/firebase-js-sdk/issues/1188


packages/database-types/index.d.ts
```diff
@@ -68,20 +68,20 @@ export interface Query {
   limitToLast(limit: number): Query;
   off(
     eventType?: EventType,
-    callback?: (a: DataSnapshot, b?: string | null) => any,
-    context?: Object | null
-  ): any;
+    callback?: (a: DataSnapshot, b?: string) => any,
```
mikelehen: `b` (aka `prevChildName`) will be null in the case of child_added events where there's no predecessor.

packages/database-types/index.d.ts
```diff
@@ -68,20 +68,20 @@ export interface Query {
...
+    context?: Object
+  ): void;
   on(
     eventType: EventType,
-    callback: (a: DataSnapshot | null, b?: string) => any,
-    cancelCallbackOrContext?: Object | null,
-    context?: Object | null
-  ): (a: DataSnapshot | null, b?: string) => any;
+    callback: (a: DataSnapshot, b?: string) => any,
```
mikelehen: `b` (aka `prevChildName`) will be null in the case of child_added events where there's no predecessor.

packages/database/src/core/view/Event.ts
```diff
@@ -65,7 +65,7 @@ export class DataEvent implements Event {
       | ' child_removed',
     public eventRegistration: EventRegistration,
     public snapshot: DataSnapshot,
-    public prevName?: string | null
+    public prevName?: string
```
mikelehen: can be null

mikelehen:
FYI- Still LGTM after changelog addition.

schmidt-sebastian:
This will be merged into the Firebase I/O release as it contains minor breaking changes.

schmidt-sebastian:
@Feiyang1 Can you merge this with the version bump?

PR #1205, opened 2018-09-06
Removing MutationQueue.start()

schmidt-sebastian:
No longer needed on Web.

We could even go one step further and remove LocalStore.start().


schmidt-sebastian:
Investigating the test failure via https://b.corp.google.com/issues/114140836

PR #1206, opened 2018-09-06
Fixing the types for firebase/app

jamesdaniels:
```ts
import firebase from 'firebase/app'; // Does not work (it should however)
``` 

```ts
import * as firebase from 'firebase/app'; // Compiles but fails w/runtime errors
```

Closing #1184 in favor of this... While my heart was in the right place (in terms of `firebase/app`) I was modifying the types for `firebase` and introducing breaking changes.

Let's start fresh shall we, the root cause is that the typings for `firebase/app` are pointing to that of `firebase`. Which is not correct. 

`firebase/app` is an ESM+CJS with a default export, the types as defined in the root use `export = ...` which suggests AMD and does not play nice with ESM.

This fix simply reexports  `firebase` from `..` in a new `index.d.ts`.

### DIFF
--- a/packages/firebase/app/index.d.ts
+++ b/packages/firebase/app/index.d.ts
@@ -1,2 +1,18 @@
-import firebase from '@firebase/app';
+/**
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


jamesdaniels:
FYI @davideast

jamesdaniels:
@jshcrowthe this will break anyone pulling in types from `firebase/app`, are we ok with that or should we pull in from `firebase`?

jamesdaniels:
Alternatively we could do the reverse; move the types I did in #1184 to `app/index.d.ts` and then change `index.d.ts` to:

```ts
import * as firebase from 'app';
export = firebase;
export as namespace firebase;
```

Feiyang1:
@jamesdaniels Wouldn't the change break import * as firebase from 'firebase/app', unless we reexport everything from firebase namespace in firebase/app? Maybe we can migrate the index.d.ts file at the root of firebase to esm with both named and default exports?  Is there any downside to this approach? I'm not sure why we chose to use cjs format for the root index.d.ts.

jamesdaniels:
`import * as firebase from 'firebase/app'` is already broken.

PR #1210, opened 2018-09-06
b/112778345: Improve argument validation for get().

mikelehen:
In particular, we'll throw a decent error if you accidentally pass a callback
function to get().

### DIFF
--- a/packages/firestore/test/integration/api/validation.test.ts
+++ b/packages/firestore/test/integration/api/validation.test.ts
@@ -289,6 +289,26 @@ apiDescribe('Validation:', persistence => {
     );
   });
 
+  validationIt(persistence, 'get options are validated', db => {
+    const collection = db.collection('test');
+    const doc = collection.doc();
+    const fn = () => {};
+
+    expect(() => doc.get(fn as any)).to.throw(
+      'Function DocumentReference.get() requires its first argument to be of type object, but it was: a function'
+    );
+    expect(() => doc.get({ abc: 'cache' } as any)).to.throw(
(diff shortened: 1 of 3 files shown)


packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -289,6 +289,26 @@ apiDescribe('Validation:', persistence => {
...
+      `Unknown option 'abc' passed to function DocumentReference.get(). Available options: source`
+    );
+
+    expect(() => collection.get(fn as any)).to.throw(
+      'Function Query.get() requires its first argument to be of type object, but it was: a function'
+    );
+    expect(() => collection.get({ abc: 'cache' } as any)).to.throw(
+      `Unknown option 'abc' passed to function Query.get(). Available options: source`
+    );
+  });
```
rsgowman: Other possible checks:

```
const x = 42;  = "foo"; = [ "array", "of", "values"]; // or more generally, anything that isn't a map.
const x = { source: 'default', extra_param: 'oops' };  // but could become fragile if we extend options. Maybe skip this one.
```

Implement as many as you feel are appropriate (incl 0).

mikelehen: Ooops. The comment I wrote didn't get posted when I resolved the conversation, I guess.  I wrote something like:

Those are good suggestions, but would ultimately be testing the underlying validateArgType() and validateOptionNames() functions which are already pretty thoroughly exercised in aggregate by these tests.  Here I mostly just want to test that get() calls the right validators, rather than test all the cases that the validators cover.  So I'm leaving it as-is.

Issue #1213, opened 2018-09-06
index.d.ts file missing in @firebase/database package


hiranya911:
This used to be present until v0.3.1. Releases that followed 0.3.1 seem to be missing it.

```
======== Resolving module '@firebase/database' from '/home/travis/build/firebase/firebase-admin-node/src/firebase-namespace.ts'. ========
Module resolution kind is not specified, using 'NodeJs'.
Loading module '@firebase/database' from 'node_modules' folder, target file type 'TypeScript'.
Directory '/home/travis/build/firebase/firebase-admin-node/src/node_modules' does not exist, skipping all lookups in it.
Scoped package detected, looking in 'firebase__database'
'package.json' has 'typings' field 'dist/index.d.ts' that references '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts'.
Found 'package.json' at '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/package.json'. Package ID is '@firebase/database/dist/index.d.ts@0.3.4'.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database.ts' does not exist.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database.tsx' does not exist.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database.d.ts' does not exist.
'package.json' has 'typings' field 'dist/index.d.ts' that references '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts'.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts' does not exist.
Loading module as file / folder, candidate module location '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts', target file type 'TypeScript'.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts.ts' does not exist.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts.tsx' does not exist.
File '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts.d.ts' does not exist.
Directory '/home/travis/build/firebase/firebase-admin-node/node_modules/@firebase/database/dist/index.d.ts' does not exist, skipping all lookups in it.
[...]

PR #1223, opened 2018-09-12
change the build order, so declaration files are generated in the rig…

Feiyang1:
change the build order, so declaration files are generated correctly

### DIFF
--- a/packages/database/rollup.config.js
+++ b/packages/database/rollup.config.js
@@ -33,6 +33,15 @@ const external = Object.keys(
 
 export default [
   /**
+   * Node.js Build
+   */
+  {
+    input: 'index.node.ts',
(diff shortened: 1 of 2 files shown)


Feiyang1:
This is a fix for https://github.com/firebase/firebase-js-sdk/issues/1213
Some background on it - The issue is that typing for browser build is created in the wrong folder (dist/packages/database instead of dist), and weirdly typing for nodejs is created in the right place. I can only reproduce it in our repo, but not in the test repo I created with the same setup and dependencies. so I think it's an environmental issue specific to this repo. When I was debugging it, the typescript service emits the wrong path for the file,  but I couldn't determine what exactly is causing it. 
Changing the order of nodejs build and browser build somehow fixes it, so it's a workaround but we are not sure what the cause is.

PR #1230, opened 2018-09-17
Upgrade closure-builder dependency to 2.3.0.

mikelehen:
Java 10 removed the -d64 flag, which closure-builder was passing. Updating to the latest version resolves the issue.

### DIFF
--- a/packages/auth/package.json
+++ b/packages/auth/package.json
@@ -22,7 +22,6 @@
     "@firebase/auth-types": "0.3.4"
   },
   "devDependencies": {
-    "closure-builder": "2.2.39",
     "del": "3.0.0",
     "event-stream": "^3.3.4",
     "express": "4.16.3",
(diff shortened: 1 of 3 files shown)


wilhuff:
Version bump LGTM though I don't know what's going on with the tests :-(.

Feiyang1:
I see auth has dependency on closure-builder too. I suspect the build still fails for auth using java 10. I will give it a try. @bojeil-google , are you okay with upgrading the closure-builder version for auth? Do you have dependency on a specific version of closure-builder?

mikelehen:
@Feiyang1 Oh! Good catch.  I wonder if auth is not actually using it though (despite declaring a dependency).  Looks like they're using gulp-closure-compiler (https://github.com/firebase/firebase-js-sdk/blob/d257bf8cc8064c7fda74f341a1547157e56a0c28/packages/auth/gulpfile.js#L18) and google-closure-compiler (https://github.com/firebase/firebase-js-sdk/blob/d257bf8cc8064c7fda74f341a1547157e56a0c28/packages/auth/gulpfile.js#L60).

But I don't see any references to closure-builder.  I've gone ahead and removed it from auth/package.json.  Everything still builds.

mikelehen:
@bojeil-google Can you approve for auth?

PR #1232, opened 2018-09-18
Split UserDataConverter in two

wilhuff:
Splits ParseAccumulator from ParseContext and moves shared user data
classes like ParsedSetData and ParsedUpdateData to core.

This matches the reorganization in firebase/firebase-ios-sdk#1823 where
the changes have a more direct benefit. This change mostly just keeps
the ports in sync.


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -367,25 +155,19 @@ export class UserDataConverter {
...
 
   /** Parse update data from an update() call. */
   parseUpdateData(methodName: string, input: AnyJs): ParsedUpdateData {
-    const context = new ParseContext(
-      UserDataSource.Update,
-      methodName,
-      FieldPath.EMPTY_PATH
-    );
+    const accumulator = new ParseAccumulator(methodName, UserDataSource.Update);
+    const context = accumulator.rootContext();
```
mikelehen: Is it intentional that the rest of this method did not change (to use context.addToFieldMask() and accumulator.toUpdateData())?

PR #1233, opened 2018-09-18
Using SortedSet instead of a PriorityQueue

gsoltis:
As discussed, here's the implementation of `RollingSequenceNumberBuffer` with a `SortedSet`.

Note that merge base is the current `lru` PR: https://github.com/firebase/firebase-js-sdk/pull/1224


packages/firestore/src/local/lru_garbage_collector.ts
```diff
@@ -77,139 +77,52 @@ export type ActiveTargets = {
...
+      const highestValue = this.buffer.last()!;
+      if (bufferEntryComparator(entry, highestValue) < 0) {
+        this.buffer = this.buffer.delete(highestValue).add(entry);
       }
     }
   }
 
   get maxValue(): ListenSequenceNumber {
-    return this.queue.peek()!;
+    return this.buffer.last()![0];
```
mikelehen: Can you add a comment why it's guaranteed the buffer will not be empty?

PR #1237, opened 2018-09-19
Implement LRU Reference Delegate for Memory Persistence

gsoltis:
This PR adds only the lru reference delegate for memory. The merge base is into `lru-master`.

Android Port: FirebasePrivate/firebase-android-sdk#190
iOS Port: firebase/firebase-ios-sdk#1600

### DIFF
--- a/packages/firestore/src/local/memory_persistence.ts
+++ b/packages/firestore/src/local/memory_persistence.ts
@@ -15,7 +15,15 @@
  */
 
 import { User } from '../auth/user';
+import { DocumentKey } from '../model/document_key';
 import { debug } from '../util/log';
+import { ObjectMap } from '../util/obj_map';
+import { encode } from './encoded_resource_path';
+import {
+  ActiveTargets,
+  LruDelegate,
+  LruGarbageCollector
+} from './lru_garbage_collector';
 
 import { MemoryMutationQueue } from './memory_mutation_queue';
 import { MemoryQueryCache } from './memory_query_cache';
@@ -24,14 +32,15 @@ import { MutationQueue } from './mutation_queue';
 import {
   Persistence,
   PersistenceTransaction,
-  PrimaryStateListener
+  PrimaryStateListener, ReferenceDelegate
 } from './persistence';
 import { PersistencePromise } from './persistence_promise';
-import { QueryCache } from './query_cache';
-import { RemoteDocumentCache } from './remote_document_cache';
+import { QueryData } from './query_data';
+import { ReferenceSet } from './reference_set';
 import { ClientId } from './shared_client_state';
 import { ListenSequenceNumber } from '../core/types';
 import { ListenSequence } from '../core/listen_sequence';
+import * as obj from '../util/obj';
 
 const LOG_TAG = 'MemoryPersistence';
 
@@ -49,17 +58,31 @@ export class MemoryPersistence implements Persistence {
    */
   private mutationQueues: { [user: string]: MutationQueue } = {};
   private remoteDocumentCache = new MemoryRemoteDocumentCache();
-  private queryCache = new MemoryQueryCache();
+  private readonly queryCache: MemoryQueryCache;// = new MemoryQueryCache();
+  private readonly listenSequence = new ListenSequence(0);
 
   private _started = false;
 
-  constructor(private readonly clientId: ClientId) {
-    this._started = true;
+  private _referenceDelegate: ReferenceDelegate | null;
+  get referenceDelegate(): ReferenceDelegate {
+    return this._referenceDelegate!;
   }
 
-  async shutdown(deleteData?: boolean): Promise<void> {
+  static lruPersistence(clientId: ClientId): MemoryPersistence {
+    const persistence = new MemoryPersistence(clientId);
+    persistence._referenceDelegate = new MemoryLruDelegate(persistence);
+    return persistence;
+  }
+
+  constructor(private readonly clientId: ClientId) {
+    this._started = true;
+    this.queryCache = new MemoryQueryCache(this);
+  }
+
+  shutdown(deleteData?: boolean): Promise<void> {
     // No durable state to ensure is closed on shutdown.
     this._started = false;
+    return Promise.resolve();
   }
 
   get started(): boolean {
@@ -84,17 +107,17 @@ export class MemoryPersistence implements Persistence {
   getMutationQueue(user: User): MutationQueue {
     let queue = this.mutationQueues[user.toKey()];
     if (!queue) {
-      queue = new MemoryMutationQueue();
+      queue = new MemoryMutationQueue(this);
       this.mutationQueues[user.toKey()] = queue;
     }
     return queue;
   }
 
-  getQueryCache(): QueryCache {
+  getQueryCache(): MemoryQueryCache {
     return this.queryCache;
   }
 
-  getRemoteDocumentCache(): RemoteDocumentCache {
+  getRemoteDocumentCache(): MemoryRemoteDocumentCache {
     return this.remoteDocumentCache;
   }
 
@@ -107,8 +130,12 @@ export class MemoryPersistence implements Persistence {
   ): Promise<T> {
     debug(LOG_TAG, 'Starting transaction:', action);
     return transactionOperation(
-      new MemoryTransaction(ListenSequence.INVALID)
+      new MemoryTransaction(this.listenSequence.next())
     ).toPromise();
+  }
+
+  mutationQueuesContainKey(transaction: PersistenceTransaction, key: DocumentKey): PersistencePromise<boolean> {
+    return or(obj.values(this.mutationQueues).map(queue => () => queue.containsKey(transaction, key)));
   }
 }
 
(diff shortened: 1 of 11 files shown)


packages/firestore/src/local/memory_persistence.ts
```diff
@@ -49,17 +59,31 @@ export class MemoryPersistence implements Persistence {
    */
   private mutationQueues: { [user: string]: MutationQueue } = {};
   private remoteDocumentCache = new MemoryRemoteDocumentCache();
-  private queryCache = new MemoryQueryCache();
+  private readonly queryCache: MemoryQueryCache; // = new MemoryQueryCache();
+  private readonly listenSequence = new ListenSequence(0);
 
   private _started = false;
 
+  private _referenceDelegate: ReferenceDelegate | null;
```
schmidt-sebastian: We have been using just `ReferenceDelegate` as a the type if a member is essentially always set. Would it be possible to pass the delegate into the constructor from your factory function? You could then also make the variable itself public and remove the getter.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -49,17 +59,31 @@ export class MemoryPersistence implements Persistence {
...
+  get referenceDelegate(): ReferenceDelegate {
+    return this._referenceDelegate!;
+  }
+
+  static lruPersistence(clientId: ClientId): MemoryPersistence {
```
schmidt-sebastian: These methods in `indexeddb_persistence` are prefixed with `create`, which I like a lot. Can you do the same here?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -119,3 +154,154 @@ export class MemoryPersistence implements Persistence {
...
+/**
+ * Given an array of predicate functions that asynchronously evaluate to a
+ * boolean, implements a short-circuiting `or` between the results. Predicates
+ * will be evaluated until one of them returns `true`, then stop. The final
+ * result will be whether any of them returned `true`.
+ */
+function or(
+  predicates: Array<() => PersistencePromise<boolean>>
+): PersistencePromise<boolean> {
+  let p: PersistencePromise<boolean> = PersistencePromise.resolve(false);
```
schmidt-sebastian: I would move this to the PersistencePromise class as it is generally useful.

BTW, this is somewhat similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race, so you could maybe take inspiration from the name. It is not super similar tho :)

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -119,3 +154,154 @@ export class MemoryPersistence implements Persistence {
...
+  for (let i = 0; i < predicates.length; i++) {
+    const nextPredicate = predicates[i];
```
schmidt-sebastian: You don't seem to be using the index variable, so you could make this a for-each loop.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -119,3 +154,154 @@ export class MemoryPersistence implements Persistence {
...
+    txn: PersistenceTransaction,
+    f: (sequenceNumber: ListenSequenceNumber) => void
+  ): PersistencePromise<void> {
+    this.orphanedSequenceNumbers.forEach((_, sequenceNumber) =>
+      f(sequenceNumber)
+    );
+    return PersistencePromise.resolve();
+  }
+
+  setInMemoryPins(inMemoryPins: ReferenceSet): void {
```
schmidt-sebastian: I much prefer the name on Android (setAdditionalReferences). It conveys meaning even to those folks that don't know that Lru "pins" documents.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -119,3 +154,154 @@ export class MemoryPersistence implements Persistence {
...
+    this.additionalReferences = inMemoryPins;
+  }
+
+  removeTargets(
+    txn: PersistenceTransaction,
+    upperBound: ListenSequenceNumber,
+    activeTargetIds: ActiveTargets
```
schmidt-sebastian: The argument name should just be `activeTargets`.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -119,3 +154,154 @@ export class MemoryPersistence implements Persistence {
...
+  }
+
+  private isPinned(
+    txn: PersistenceTransaction,
+    key: DocumentKey,
+    upperBound: ListenSequenceNumber
+  ): PersistencePromise<boolean> {
+    return or([
+      () => this.persistence.mutationQueuesContainKey(txn, key),
+      () => this.additionalReferences!.containsKey(txn, key),
```
schmidt-sebastian: Is `additionalReferences` guaranteed to be set here? We could also initialize it with an empty set.

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -134,6 +150,28 @@ export class MemoryQueryCache implements QueryCache {
     return PersistencePromise.resolve();
   }
 
+  removeTargets(
+    transaction: PersistenceTransaction,
+    upperBound: ListenSequenceNumber,
+    activeTargetIds: ActiveTargets
```
schmidt-sebastian: Same comment as above, this should just be `activeTargets` to match the argument type.

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -163,7 +201,14 @@ export class MemoryQueryCache implements QueryCache {
...
     this.references.addReferences(keys, targetId);
-    return PersistencePromise.resolve();
+    const referenceDelegate = this.persistence.referenceDelegate;
+    const promises: Array<PersistencePromise<void>> = [];
+    if (referenceDelegate) {
+      keys.forEach(key => {
+        promises.push(referenceDelegate.addReference(txn, key));
+      });
+    }
+    return PersistencePromise.waitFor(promises);
```
schmidt-sebastian: If the set of keys here is small, you could also do:

```
return PersistencePromise.forEach(keys.toArray(), key -> referenceDelegate.addReference(txn, key));
```

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -172,7 +217,14 @@ export class MemoryQueryCache implements QueryCache {
     targetId: TargetId
   ): PersistencePromise<void> {
     this.references.removeReferences(keys, targetId);
-    return PersistencePromise.resolve();
+    const referenceDelegate = this.persistence.referenceDelegate;
+    const promises: Array<PersistencePromise<void>> = [];
+    if (referenceDelegate) {
+      keys.forEach(key => {
+        promises.push(referenceDelegate.removeReference(txn, key));
```
schmidt-sebastian: Same optional comment as above.

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -82,6 +82,19 @@ export class MemoryRemoteDocumentCache implements RemoteDocumentCache {
     return PersistencePromise.resolve(results);
   }
 
+  forEachDocumentKey(
+    transaction: PersistenceTransaction,
+    f: (key: DocumentKey) => PersistencePromise<void>
+  ): PersistencePromise<void> {
+    const iter = this.docs.getIterator();
```
schmidt-sebastian: Is there a reason you are not just using `this.docs.forEach()`?

packages/firestore/src/util/obj.ts
```diff
@@ -58,6 +58,12 @@ export function forEachNumber<V>(
   }
 }
 
+export function values<V>(obj: Dict<V>): V[] {
```
schmidt-sebastian: This strikes me as very similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values.

packages/firestore/src/local/memory_persistence.ts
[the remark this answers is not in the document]
gsoltis: Can I punt on this until the next PR? That one will introduce `MemoryEagerDelegate`, and then there will always be a reference delegate for memory persistence.

packages/firestore/src/local/memory_persistence.ts
gsoltis: The semantics are different. `race` runs all of the promises concurrently, whereas mine runs serially and short-circuits.

I will move it though.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -49,17 +59,31 @@ export class MemoryPersistence implements Persistence {
...
   private mutationQueues: { [user: string]: MutationQueue } = {};
   private remoteDocumentCache = new MemoryRemoteDocumentCache();
-  private queryCache = new MemoryQueryCache();
+  private readonly queryCache: MemoryQueryCache; // = new MemoryQueryCache();
+  private readonly listenSequence = new ListenSequence(0);
 
   private _started = false;
 
+  private _referenceDelegate: ReferenceDelegate | null;
+  get referenceDelegate(): ReferenceDelegate {
```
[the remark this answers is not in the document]
gsoltis: As above, this will eventually just be a property. It's just waiting on the final reference delegate implementation.

packages/firestore/src/local/memory_persistence.ts
gsoltis: Sure.

packages/firestore/src/local/memory_persistence.ts
gsoltis: It has already been changed on Android to `inMemoryPins` as part of the indexeddb portion of this port.

packages/firestore/src/local/memory_persistence.ts
gsoltis: This has already been changed once, I'd prefer to leave it as is at this point.

packages/firestore/src/local/memory_persistence.ts
gsoltis: It is required to be set. It originally could be set multiple times in spec tests, but that might actually be gone now. I'm going to leave this consistent with other ports but file a bug to investigate if this could be an argument to the constructor: b/116248679

packages/firestore/src/local/memory_persistence.ts
gsoltis: Done.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: Leaving as is. It's been renamed here and on Android already.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: I don't know if it's small? It's however many keys are added to a target.

packages/firestore/src/local/memory_query_cache.ts
gsoltis: Again, we don't know if the number of keys is small.

packages/firestore/src/local/memory_remote_document_cache.ts
gsoltis: Changed to `forEach`.

packages/firestore/src/util/obj.ts
gsoltis: Seems unavailable on some of our platforms :(

packages/firestore/src/local/memory_persistence.ts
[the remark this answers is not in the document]
schmidt-sebastian: Please add a TODO to clean this up then :)

packages/firestore/src/local/memory_persistence.ts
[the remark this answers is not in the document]
schmidt-sebastian: I would still like to push you to move this down to where we define other methods.

packages/firestore/src/local/memory_persistence.ts
schmidt-sebastian: Hm. Sad panda. Reminder for myself to never to go on parental leave.

packages/firestore/src/local/memory_persistence.ts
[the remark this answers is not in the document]
gsoltis: Done.

packages/firestore/src/local/memory_persistence.ts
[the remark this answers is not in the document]
gsoltis: Done.

Issue #1031, opened 2018-07-20
Uncaught SyntaxError: The requested module '../../webchannel-wrapper/dist/index.js' does not provide an export named 'ErrorCode'


lpellegr:
Using firebase package from NPM in version 5.3.0, importing firebase with ES6 module import:

```
import {firebase} from '../../node_modules/firebase/dist/index.esm.js';

firebase.initializeApp({...});

const auth = firebase.auth();
const database = firebase.database();
const firestore = firebase.firestore();
const functions = firebase.functions();
const storage = firebase.storage();
```
raises the following error:

```
Uncaught SyntaxError: The requested module '../../webchannel-wrapper/dist/index.js' does not provide an export named 'ErrorCode'
```

All related issues I have found are marked are closed or resolved. Is my usage wrong? 

I need to import `auth`, `database`, `firestore`, `functions` and `storage`

My usage is with a Polymer 3 app that aims to be eventually packaged.

lpellegr:
Context, error message and reproduction steps are given.

lpellegr:
Changing the import to `import firebase from 'firebase';` produces the exact same error as above.

lpellegr:
Here is a full example for reproducing:

https://github.com/lpellegr/pwa-starter-kit-firebase-import

The example is based on [`pwa-start-kit-no-redux` template](https://github.com/Polymer/pwa-starter-kit/tree/template-no-redux). I have simply added the following lines:

https://github.com/lpellegr/pwa-starter-kit-firebase-import/commit/b506824bb95a581525ba69a7857865a7bd01ad06#diff-842138042367a60b6d00ce81115f99ed

To run the example, execute `npm install`, `polymer serve`, then open `http://127.0.0.1:8081/` with Chrome. You will get the following error message on the console:

> Uncaught SyntaxError: The requested module '../../webchannel-wrapper/dist/index.js' does not provide an export named 'ErrorCode'

lpellegr:
@mikelehen @jshcrowthe I have added an example for reproduction. Any idea if the issue is due to a bad usage? or is it an issue that is going to be fixed?

mikelehen:
@lpellegr Sorry, unfortunately I have no idea and @jshcrowthe is unavailable for the next couple weeks.  It may be worth trying to isolate how your project is different from other projects that do work. There must be something about the build process, etc. that is causing an issue. Maybe that would shed some light.  Sorry I can't be of more help right now.

[...]

PR #1248, opened 2018-09-25
support es6 module in webchannel-wrapper

Feiyang1:
Added a build step to produce bundle in es6 module format.
@mikelehen the output file looks fine to me (it has all the expected exports in es6 format) and I tested it against the repo mentioned in https://github.com/firebase/firebase-js-sdk/issues/1031 and it works fine (just did firebase.firestore()). I don't know if the test is sufficient. If not, can you please let me know what else I should do? thanks!

### DIFF
--- a/packages/webchannel-wrapper/tools/build.js
+++ b/packages/webchannel-wrapper/tools/build.js
@@ -50,7 +50,8 @@ closureBuilder.build(
         compilation_level: 'ADVANCED'
       }
     }
-  }, async function (errors, warnings, files, results) {
+  },
+  async function(errors, warnings, files, results) {
     const filePath = resolve(__dirname, '../src/index.js');
     const inputOptions = {
       input: filePath,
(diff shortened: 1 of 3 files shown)


packages/webchannel-wrapper/tools/build.js
```diff
@@ -28,7 +33,44 @@ closureBuilder.build({
       output_wrapper:
         "(function() {%output%}).call(typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : typeof window !== 'undefined' ? window : {})",
       language_out: 'ECMASCRIPT5',
-      compilation_level: 'ADVANCED_OPTIMIZATIONS'
+      compilation_level: 'ADVANCED'
```
mikelehen: Interesting.  Were we not getting advanced optimizations before then?  Or does either value work?

Feiyang1: yep, either one works, but only ADVANCED is listed in their latest documentation. https://github.com/google/closure-compiler/wiki/Flags-and-Options

Feiyang1:
That's a great a point. I will address it in a separate pr.

PR #1252, opened 2018-09-26
remove incorrect typing

Feiyang1:
null is not possible in the apps array, hence I'm updating the typing.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -105,7 +105,7 @@ declare namespace firebase {
 
   function app(name?: string): firebase.app.App;
 
-  var apps: (firebase.app.App | null)[];
+  var apps: firebase.app.App[];
 
   function auth(app?: firebase.app.App): firebase.auth.Auth;
 


Feiyang1:
@mikelehen correct me if I'm wrong. My understanding is that {component}-types packages is the source of the truth for typings, but we don't want to expose all of them to developers, so we hand roll the firebase/index.d.ts that only expose what we want to expose. 
Admittedly creating and maintaining a .d.ts file is bad, so I hope I can tackle it as part of the documentation revamp (using typings).

mikelehen:
Unfortunately I'm not sure. :-)  I know for Firestore, we've been keeping them 100% in sync.  We don't have anything in our firestore-types that isn't public.  Looking at app-types/index.d.ts I'm fairly sure everything in there is meant to be public (and it has comments which suggest so as well). So I don't think there'd be any harm in copying them over.  That said if you don't want to mess with that right now, that's fine.

Tangentially, I don't really understand why we have the separate typings files in the first place.  If we can find a way to unify things, that would be great...  But I'm a bit nervous about breaking stuff.

Feiyang1:
I agree public interface typings should be kept in sync. Thanks for the feedback!

PR #1256, opened 2018-09-26
Implement Eager Reference Delegate for Memory Persistence, drop old GC

gsoltis:
In this PR:
 * Implement `MemoryEagerDelegate`, implements eager gc for memory persistence
 * Remove `GarbageCollector`, `GarbageSource` and implementations / extensions
 * Fix up tests, mostly in local store.

Final portion of LRU port to webclient
Android Port: FirebasePrivate/firebase-android-sdk#190
iOS Port: firebase/firebase-ios-sdk#1600


packages/firestore/src/local/persistence_promise.ts
```diff
@@ -238,4 +238,28 @@ export class PersistencePromise<T> {
     }
     return p;
   }
+
+  /**
+   * Given a collection that has a `forEach` method, call the given function on
+   * each element in the collection and wait for all of the resulting
+   * concurrent PersistencePromises to resolve.
+   */
+  static waitForEach<K>(
```
schmidt-sebastian: I just revived one of my older PRs that would allow you to merge this with `forEach` above. Let me know if you think this could work for you: https://github.com/firebase/firebase-js-sdk/pull/1269

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -470,12 +469,9 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
     ).next(removedDocuments => {
       this.removeCachedMutationKeys(batch.batchId);
-      if (this.garbageCollector !== null) {
-        for (const key of removedDocuments) {
-          // TODO(gsoltis): tell reference delegate that mutation was ack'd
-          this.garbageCollector.addPotentialGarbageKey(key);
-        }
-      }
+      return PersistencePromise.waitForEach(removedDocuments, key => {
+        return this.referenceDelegate.removeMutationReference(transaction, key);
```
schmidt-sebastian: This line seems to be missing from Android (at least in master). Is that still coming?

packages/firestore/src/local/indexeddb_query_cache.ts
```diff
@@ -295,17 +291,14 @@ export class IndexedDbQueryCache implements QueryCache {
...
       const path = EncodedResourcePath.encode(key.path);
-      promises.push(store.delete([targetId, path]));
-      if (this.garbageCollector !== null) {
-        this.garbageCollector.addPotentialGarbageKey(key);
-      }
-      promises.push(this.referenceDelegate.removeReference(txn, key));
+      return PersistencePromise.waitFor([
+        store.delete([targetId, path]),
+        this.referenceDelegate.removeReference(txn, key)
+      ]);
```
schmidt-sebastian: This could just be:
```
return store.delete(..).next(() => this.referenceDelegate.removeReference(txn, key))
```

Unless you have data that the `waitFor` parallelism meaningfully improves performance :)

packages/firestore/src/local/local_store.ts
```diff
@@ -439,7 +430,6 @@ export class LocalStore {
             // Do not ref/unref unassigned targetIds - it may lead to leaks.
             let queryData = this.queryDataByTarget[targetId];
             if (!queryData) return;
-
```
schmidt-sebastian: I would prefer if we keep this newline. It makes the early return stand out more.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -57,24 +57,30 @@ export class MemoryPersistence implements Persistence {
...
 
-  // TODO(gsoltis): remove option to be null once eager delegate is implemented.
-  private _referenceDelegate: ReferenceDelegate | null;
+  readonly referenceDelegate: MemoryLruDelegate | MemoryEagerDelegate;
 
   static createLruPersistence(clientId: ClientId): MemoryPersistence {
-    const persistence = new MemoryPersistence(clientId);
-    persistence._referenceDelegate = new MemoryLruDelegate(persistence);
-    return persistence;
+    return new MemoryPersistence(clientId, /* isEager= */ false);
```
schmidt-sebastian: What do you think about passing  in the `referenceDelegate` and removing the `isEager` flag?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -211,6 +318,7 @@ export class MemoryLruDelegate implements ReferenceDelegate, LruDelegate {
   ): PersistencePromise<number> {
     let count = 0;
     const cache = this.persistence.getRemoteDocumentCache();
+    // TODO(gsoltis): make this concurrent
```
schmidt-sebastian: Hm. This is an in-memory data structure. I would be surprised if this could be sped up if we allowed out-of-order execution.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -244,14 +235,28 @@ class LocalStoreTester {
     return this;
   }
 
+  toNotContainIfEager(
```
schmidt-sebastian: Nice! I would maybe flip this to `toContainIfLru`, since that feels like the stricter assert. But that's just my personal opinion :)

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -244,14 +235,28 @@ class LocalStoreTester {
...
+    gcIsEager: boolean,
+    doc: MaybeDocument
+  ): LocalStoreTester {
+    if (gcIsEager) {
```
schmidt-sebastian: Can we pass `gcIsEager` to LocalStoreTester? That way, the tests would not need to know what GC mode they are running under.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -643,22 +639,25 @@ function genericLocalStoreTests(
   });
 
   it('handles SetMutation -> Ack -> PatchMutation -> Reject', () => {
```
schmidt-sebastian: This test could be moved to the `LocalStore w/ Memory Persistence` "subclass".

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -688,37 +687,45 @@ function genericLocalStoreTests(
       .toContain(deletedDoc('foo/bar', 2))
       .afterAcknowledgingMutation({ documentVersion: 3 }) // patch mutation
       .toReturnChanged(unknownDoc('foo/bar', 3))
-      .toContain(unknownDoc('foo/bar', 3))
+      .toNotContainIfEager(gcIsEager, unknownDoc('foo/bar', 3))
       .finish();
   });
 
   it('collects garbage after ChangeBatch with no target ids', () => {
+    if (!gcIsEager) {
```
schmidt-sebastian: Same comment about moving this to `LocalStore w/ Memory Persistence`.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -427,15 +424,13 @@ abstract class TestRunner {
...
-    this.persistence = await this.initPersistence(this.serializer);
-    const garbageCollector = this.getGarbageCollector();
-
-    this.localStore = new LocalStore(
-      this.persistence,
-      this.user,
-      garbageCollector
+    this.persistence = await this.initPersistence(
+      this.serializer,
+      this.useGarbageCollection
```
schmidt-sebastian: Don't we always "use" garbage collection - we just switch between modes?

gsoltis:
Ugh. Merging master messed everything up. Will try to untangle.

gsoltis:
Whew, ok, I think it's straightened out now.

packages/firestore/src/local/indexeddb_query_cache.ts
schmidt-sebastian: There are only two promises here, and one of them is a delete for a single key.

packages/firestore/src/local/memory_persistence.ts
schmidt-sebastian: Sounds good. Thanks for the explanation.

packages/firestore/src/local/persistence_promise.ts
schmidt-sebastian: I feel like my comment on this got lost somehow. If you do think that the parallelism improves things, can we make the difference between `forEach`/`waitFor`/`waitForEach` more explicit? We could maybe rename `waitForEach` to `forEachParalellel` or something similar (sorry for the horrible name suggestion) or pass in a flag that allows parallelization. For what it's worth, there are currently other calls to `forEach` that could be parallelized.

gsoltis:
Sorry, `local_store.test.ts` comments got hidden. Will address those shortly.

packages/firestore/test/unit/local/local_store.test.ts
gsoltis: I started implementing this, but I think it gets a little too ugly with sharing the test infrastructure with the generic tests. 

If you feel strongly we can refactor, but I'd like to do it in a separate PR.

packages/firestore/src/local/persistence_promise.ts
```diff
@@ -259,4 +243,25 @@ export class PersistencePromise<T> {
     }
     return p;
   }
+
+  /**
+   * Given an iterable, call the given function on each element in the
+   * collection and wait for all of the resulting concurrent
+   * PersistencePromises to resolve.
+   */
```
schmidt-sebastian: I would make it more explicit that it runs in parallel:

```
/**
   * Given an iterable, call the given function on each element in
   * parallel and wait for all of the resulting concurrent
   * PersistencePromises to resolve.
   */
```

packages/firestore/test/unit/local/local_store.test.ts
schmidt-sebastian: I was hoping that you could just move the tests method to the respective class, similar to what we do here: https://github.com/firebase/firebase-js-sdk/blob/322357244dfc3f35ae97e7d522d40d345b216b51/packages/firestore/test/unit/local/remote_document_cache.test.ts#L168

This can of course be done in a separate PR.

packages/firestore/src/local/persistence_promise.ts
gsoltis: I think your update is actually subtly wrong. The callback is called on each element in sequence,  not in parallel. The resulting promises run in parallel.

packages/firestore/test/unit/local/local_store.test.ts
gsoltis: I filed b/117272361 to do it across all clients

PR #1262, opened 2018-09-28
 Use @grpc/proto-loader instead of the now-deprecated grpc.load

ryanpbrewster:
I get this warning every time I start up the SDK in Node:
```
(node:32614) DeprecationWarning: grpc.load: Use the @grpc/proto-loader module with grpc.loadPackageDefinition instead
```


ryanpbrewster:
Closing this, it's more complex than I had hoped and we'll either need to use protobufjs directly or re-think our approach.

PR #1269, opened 2018-10-01
Allow iterables in PersistencePromise.map/forEach/waitFor

schmidt-sebastian:
Revived this PR since https://github.com/firebase/firebase-js-sdk/pull/1256 could benefit from it.

This allows us to use SortedSet and SortedMap in the PersistencePromise utilities.

### DIFF
--- a/packages/firestore/src/local/persistence_promise.ts
+++ b/packages/firestore/src/local/persistence_promise.ts
@@ -170,49 +170,54 @@ export class PersistencePromise<T> {
 
   static waitFor(
     // tslint:disable-next-line:no-any Accept all Promise types in waitFor().
-    all: Array<PersistencePromise<any>>
+    all: Iterable<PersistencePromise<any>>
   ): PersistencePromise<void> {
-    const expectedCount = all.length;
-    if (expectedCount === 0) {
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/persistence_promise.ts
```diff
@@ -170,49 +170,54 @@ export class PersistencePromise<T> {
...
-    for (let i = 0; i < all.length; ++i) {
-      promises[i] = all[i].next(result => {
-        results[i] = result;
-      });
-    }
-    return PersistencePromise.waitFor(promises).next(() => {
-      return results;
-    });
+    return PersistencePromise.forEach(all, result => {
+      results.push(result);
```
schmidt-sebastian: I think this should append the elements in the correct order since forEach waits for each promise one by one...

schmidt-sebastian:
We install polyfills for Symbol.iterator:

import 'core-js/fn/array/find';
import 'core-js/fn/array/find-index';
import 'core-js/fn/object/assign';
import 'core-js/fn/string/starts-with';
import 'core-js/fn/string/repeat';
import 'core-js/fn/symbol';
import 'core-js/fn/symbol/iterator';

PR #1271, opened 2018-10-01
Fix map()

schmidt-sebastian:
Turns out that using foreach() for map() doesn't quite work. These two functions do very different things:

- Map takes the result of persistence promises and converts them to an array.
- forEach takes a list of values and "converts" them to a PersistencePromise by passing them through a callback.

`map` is so far only used in a single test that doesn't look at the return values, and hence nothing broke. I changed that test to use `waitFor`.

Alternatively, we could also remove `map`, but it might come in handy and is not all that trivial to implement (as I have just shown).

### DIFF
--- a/packages/firestore/src/local/persistence_promise.ts
+++ b/packages/firestore/src/local/persistence_promise.ts
@@ -224,7 +224,7 @@ export class PersistencePromise<T> {
 
   static forEach<T>(
     all: Iterable<T>,
-    callback: (elem:T) => PersistencePromise<void>
+    callback: (elem: T) => PersistencePromise<void>
   ): PersistencePromise<void> {
     const it = all[Symbol.iterator]();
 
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/persistence_promise.ts
```diff
@@ -200,10 +200,26 @@ export class PersistencePromise<T> {
 
   static map<R>(all: Iterable<PersistencePromise<R>>): PersistencePromise<R[]> {
     const results: R[] = [];
-    return PersistencePromise.forEach(all, result => {
-      results.push(result);
```
mikelehen: Why did this compile before?  Shouldn't `result` be a `PersistencePromise<R>` and therefore be incompatible with `results.push()` which should have expected an `R`?

schmidt-sebastian: The type signature for `forEach` just listed `(T) => callback`, but it needs to be `(elem:T) => callback`

mikelehen: Aha!  Might be time to add 'parameter' to our typedef tslint rules...  https://palantir.github.io/tslint/rules/typedef/

PR #1272, opened 2018-10-02
[Multi-Tab] Detect and recover from GCed Remote Document Changelog

schmidt-sebastian:
This is meant to address the case when one tab removed document changes from the document change log that a throttled tab has not yet seen. The detection relies on the fact that change IDs are incremented by 1 each time (as guaranteed by https://www.w3.org/TR/IndexedDB/#key-generator-construct)

### DIFF
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -175,6 +175,8 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
     return changesStore
       .iterate({ range }, (_, documentChange) => {
         if (firstIteration) {
+          firstIteration = false;
+
           // If our client was throttled for more than 30 minutes, another
           // client may have garbage collected the remote document changelog.
           if (this._lastProcessedDocumentChangeId + 1 !== documentChange.id) {
@@ -190,7 +192,6 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
               )
             );
           }
-          firstIteration = false;
         }
 
         changedKeys = changedKeys.unionWith(
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -27,6 +27,7 @@ import { DocumentKey } from '../model/document_key';
 
 import { SnapshotVersion } from '../core/snapshot_version';
 import { assert } from '../util/assert';
+import { Code, FirestoreError } from '../util/error';
 import { IndexedDbPersistence } from './indexeddb_persistence';
 import {
   DbRemoteDocument,
@@ -39,6 +40,10 @@ import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { RemoteDocumentCache } from './remote_document_cache';
 import { SimpleDb, SimpleDbStore, SimpleDbTransaction } from './simple_db';
+
+const REMOTE_DOCUMENT_CHANGELOG_GARBAGE_COLLECTED_ERR_MSG =
+  'The remote document changelog no longer contains all changes for all ' +
+  'local query views. It may be necessary to rebuild these views.';
 
 export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
   /** The last id read by `getNewDocumentChanges()`. */
@@ -69,21 +74,11 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
    */
   // PORTING NOTE: This is only used for multi-tab synchronization.
   start(transaction: SimpleDbTransaction): PersistencePromise<void> {
-    // If there are no existing changes, we set `lastProcessedDocumentChangeId`
-    // to 0 since IndexedDb's auto-generated keys start at 1.
-    this._lastProcessedDocumentChangeId = 0;
-
     const store = SimpleDb.getStore<
       DbRemoteDocumentChangesKey,
       DbRemoteDocumentChanges
     >(transaction, DbRemoteDocumentChanges.store);
-    return store.iterate(
-      { keysOnly: true, reverse: true },
-      (key, value, control) => {
-        this._lastProcessedDocumentChangeId = key;
-        control.done();
-      }
-    );
+    return this.synchronizeLastDocumentChangeId(store);
   }
 
   addEntries(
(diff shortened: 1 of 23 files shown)


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -172,18 +167,26 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+          // client may have garbage collected the remote document changelog.
+          if (this._lastProcessedDocumentChangeId + 1 !== documentChange.id) {
+            return PersistencePromise.reject(
+              new FirestoreError(
+                Code.DATA_LOSS,
+                REMOTE_DOCUMENT_CHANGELOG_GARBAGE_COLLECTED_ERR_MSG
+              )
+            );
+          }
+          firstIteration = false;
```
mikelehen: Per my comment elsewhere, perhaps rework this so that getNewDocumentChanges() always advances past all document changes, even if it ultimately returns an error...  e.g. by changing firstIteration to minChangeId and checking it in the next() block...

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -217,6 +220,40 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+      { keysOnly: true, reverse: true },
+      (key, value, control) => {
+        this._lastProcessedDocumentChangeId = key;
+        control.done();
+      }
+    );
+  }
+}
+
+export function isRemoteDocumentChangesGarbageCollectedError(
```
mikelehen: I like verbosity, but "is...Error" gets hard to parse when there are 5 words in between.  How about isDocumentChangeMissingError()?

I was also wondering if you could just:
```
export const DOCUMENT_CHANGE_MISSING_ERROR = new Error(...);
```

But I am not sure how JS works. Would that mess up the stack trace in the error or something?

It's also a little dubious to be using errors for control flow.  We could change getNewDocumentChanges() to return null if it encountered a missing change or something, but I think I'm not that thrilled about the idea.

packages/firestore/test/unit/local/test_remote_document_cache.ts
```diff
@@ -67,13 +68,42 @@ export class TestRemoteDocumentCache {
...
-      'getNextDocumentChanges',
+      'getNewDocumentChanges',
       'readonly',
       txn => {
         return this.cache.getNewDocumentChanges(txn);
       }
     );
   }
+
+  resetLastProcessDocumentChange(): Promise<void> {
```
mikelehen: I'm hoping this method goes away, but if not you're missing 'ed' here and below.

packages/firestore/test/unit/local/test_remote_document_cache.ts
```diff
@@ -67,13 +68,42 @@ export class TestRemoteDocumentCache {
...
+      throw new Error(
+        'Can only removeDocumentChangesThroughChangeId() in IndexedDb'
+      );
+    }
+    return this.persistence.runTransaction(
+      'removeDocumentChangesThroughChangeId',
+      'readwrite-primary',
+      txn => {
+        return (this
+          .cache as IndexedDbRemoteDocumentCache).removeDocumentChangesThroughChangeId(
```
mikelehen: Is this cast necessary?  If you moved your if check above into this closure it probably wouldn't be...

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: Consider it done.

packages/firestore/src/local/remote_document_cache.ts
```diff
@@ -94,4 +94,15 @@ export interface RemoteDocumentCache {
...
   ): PersistencePromise<MaybeDocumentMap>;
+
+  /**
+   * Skips all existing change log entries in IndexedDb and moves the change log
+   * cursor past the last existing change. This method should only be called if
+   * `getNewDocumentChanges()`can no longer replay all changes and all views
+   * have already been manually synchronized.
+   */
+  // PORTING NOTE: This is only used for multi-tab synchronization.
+  resetLastProcessedDocumentChange(
```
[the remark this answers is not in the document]
schmidt-sebastian: I removed the method and inlined the call that resets the change log pointer.

packages/firestore/test/unit/local/test_remote_document_cache.ts
schmidt-sebastian: And it's gone!

packages/firestore/test/unit/local/test_remote_document_cache.ts
schmidt-sebastian: No, I managed to get rid of it by moving the assert.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I changed the name to `isDocumentChangeMissingError`.

In general, I would like to shy away from statically allocating errors as it does mess up the stacktraces (even in JS).

I'm gonna use the current fire alarm in SFO as an excuse to not comment on the API contract. I can't really think of a much better way to convey this information.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -171,16 +171,22 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
           // client may have garbage collected the remote document changelog.
           if (this._lastProcessedDocumentChangeId + 1 !== documentChange.id) {
-            return PersistencePromise.reject(
-              new FirestoreError(
-                Code.DATA_LOSS,
-                REMOTE_DOCUMENT_CHANGELOG_GARBAGE_COLLECTED_ERR_MSG
+            // Reset the `lastProcessedDocumentChangeId` to allow further
+            // invocations to successfully return the changes after this
+            // rejection.
+            return this.synchronizeLastDocumentChangeId(changesStore).next(() =>
```
mikelehen: Just to verify, did you consider just letting this iteration continue so that line 199 can correctly update _lastProcessedDocumentChangeId, and then we could (I think?) get rid of the synchronizeLastDocumentChangeId?   I think the reduced code complexity would outweigh any potential performance saving of getting to skip the changedKeys accounting...

schmidt-sebastian:
I tested this PR manually this morning. I found one area of improvement (we should set firstIteration to false before we return, since the iterate function just keeps iterating).

Outside of this PR, I did find a slew of things that were not working/no longer working in Multi-Tab, so thanks for asking for an extra QA run :)

PR #1274, opened 2018-10-02
update documentation to show the correct import syntax

Feiyang1:
```javascript
import firebase from 'firebase'
``` 
doesn't work with the typings we currently have. It works with vanilla javascript though. But to keep it consistently across javascript and typescript, I'm updating it to 
```javascript
import * as firebase from 'firebase'
```

### DIFF
--- a/packages/firebase/README.md
+++ b/packages/firebase/README.md
@@ -138,7 +138,7 @@ services you use:
 
 ```
 // This import loads the firebase namespace along with all its type information.
-import firebase from 'firebase/app';
+import * as firebase from 'firebase/app';
 
 // These imports load individual services into the firebase namespace.
 import 'firebase/auth';
--- a/packages/firebase/README.md
+++ b/packages/firebase/README.md
@@ -173,6 +173,19 @@ var app = firebase.initializeApp({ ... });
 // ...
 ```
 
+If you are using native ES6 module with --experimental-modules flag, you should do:
+```
+// This import loads the firebase namespace.
+import firebase from 'firebase/app';
+
+// These imports load individual services into the firebase namespace.
+import 'firebase/auth';
+import 'firebase/database';
+```
+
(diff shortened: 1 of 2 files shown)


packages/firebase/README.md
```diff
@@ -83,7 +83,7 @@ var app = firebase.initializeApp({ ... });
 If you are using ES6 imports or TypeScript:
 
 ```
-import firebase from 'firebase';
```
mikelehen: I think there's another instance (`import firebase from 'firebase/app'`) that should be updated too?

Feiyang1: good catch! done.

mikelehen:
Oh, hrm.  Based on https://github.com/firebase/firebase-js-sdk/pull/1206#issuecomment-425694197 I guess this change isn't entirely a slam dunk (it'll break people targeting node)?  I'm guessing it's still an improvement (I suspect browser usage is much more common), but maybe @jamesdaniels will want to weigh in on what we should be documenting here.  Maybe we should describe both syntaxes and when to use each or something.

Feiyang1:
Good point. It's a problem when people tries to use native es6 module in node. Native es6 module support in node is still experimental and is hidden behind a flag, so it's probably a rare use case. Anyway, I only updated the section for browser use cases, and in node use case, we only advise to use require() syntax, so we are not misleading people hopefully. 
I don't think we need to, but I added a section to explain the native es6 module use case in node. @mikelehen, @jamesdaniels if you think it adds confusion, I will remove it.

mikelehen:
I'll let James weigh in as he's much more familiar with the nuances here. 😁

Feiyang1:
@jamesdaniels Are you okay with the change?

jamesdaniels:
LGTM. 

Sorry I've not had the time to dig in a bit more here. This seems to be the most consistent for the moment.

PR #1275, opened 2018-10-02
Backfill Firestore Changelog

schmidt-sebastian:


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
-# Unreleased (0.8.0)
+# Unreleased
+
+# 0.8.2
+- [fixed] Fixed an issue where native ES6 module loading was not working.
```
wilhuff: Is it worth mentioning webchannel in this at all?

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
...
+
+# 0.8.1
+- [fixed] Fixed an issue where typings are created in the wrong location.
```
mikelehen: Is there a way to rephrase this somehow to represent a user-visible difference in the SDK?  This seems to be describing the cause of the issue rather than the symptom.  Also "are" should be "were" to match the tense of "fixed"

schmidt-sebastian:
BTW, this just copies the existing Firestore entries from https://firebase.google.com/support/release-notes/js

packages/firestore/CHANGELOG.md
schmidt-sebastian: I just used the public release notes.

packages/firestore/CHANGELOG.md
schmidt-sebastian: Leaving as is since this matches the public release notes.

PR #1282, opened 2018-10-03
Always update LocalStorage state

schmidt-sebastian:
During today's QA run, I wasn't able to reliably get LocalStorage updates when the entry we wrote didn't change. I would like to add back the `updateTimeMs`.


schmidt-sebastian:
Comments added

PR #1287, opened 2018-10-04
Detect when another tab was primary before regaining lease

schmidt-sebastian:
We need to handle the case where Tab A lost the primary lease to Tab B but didn't notice it, since Tab B died before Tab A cared. This needs to be handled because Tab B could have changed persisted state.

As part of this `stealPrimaryLease` no longer writes the new owner to IndexedDb, since that prevents the client from noticing that it didn't have the lease.


schmidt-sebastian:
> I think this looks okay though we're adding more and more code to deal with primary lease logic so it's getting harder to reason about. It would be nice if we could find ways to simplify.

Agreed. I tried to do this a little bit with this PR, but ultimately failed :/ I'll keep it in the back of my mind.

PR #1295, opened 2018-10-08
[Multi-Tab] Separate GC threshold from Primary Election threshold

schmidt-sebastian:
This addresses https://www.google.com/url?q=https://github.com/firebase/firebase-js-sdk/pull/1108/files/34e6eecf092bf1f684c290c3accec3b9981c2f17..1ab51cbe48f4c5b3a9c1d36d51435f1487bbee7c%23diff-cf8f425b1bd34112162e05d44fc3a464&sa=D&usg=AFQjCNG-YmGkXVPvM2p5ywOjaZ09G6OEnw

- The 5 second threshold is only used during primary tab selection.
- The 30 minute threshold is used everywhere else. This means that the entire system tracks all states for clients that managed to rewrite their state within the last 30 minutes.  This may lead to additional query load for clients that went away without writing their zombied client ID.

I'm also debating if 30 minutes is maybe a bit too long, since the clients are supposed to update their state every 4 seconds.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -83,7 +83,7 @@ const MAX_CLIENT_AGE_MS = 30 * 60 * 1000; // 30 minutes
  * primary lease election. Clients that have not updated their client metadata
  * within 5 seconds are not eligible to receive a primary lease.
  */
-const PRIMARY_ELECTION_MAX_CLIENT_AGE_MS = 5000;
+const MAX_PRIMARY_ELIGIBLE_AGE_MS = 5000;
 
 /**
  * The interval at which clients will update their metadata, including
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -72,19 +72,18 @@ import { SimpleDb, SimpleDbStore, SimpleDbTransaction } from './simple_db';
...
- * Oldest acceptable age in milliseconds for client metadata before it and its
- * associated data (such as the remote document cache changelog) can be
- * garbage collected. Clients that exceed this threshold will not be able to
- * replay Watch events that occurred before this threshold.
+ * Oldest acceptable metadata age for clients that may participate in the
+ * primary lease election. Clients that have not updated their client metadata
+ * within 5 seconds are not eligible to receive a primary lease.
  */
-const CLIENT_STATE_GARBAGE_COLLECTION_THRESHOLD_MS = 30 * 60 * 1000; // 30 Minutes
+const PRIMARY_ELECTION_MAX_CLIENT_AGE_MS = 5000;
```
mikelehen: I'm still not wild about calling this an election...  Perhaps `CLIENT_ELIGIBLE_FOR_PRIMARY_MAX_AGE_MS` (or just `PRIMARY_ELIGIBLE_MAX_AGE_MS`)?

Issue #1292, opened 2018-10-06
The imports in the RxFire Auth docs should refer to the auth module


dungahk:
#### Steps to reproduce:
Access the RxFire Auth docs https://github.com/firebase/firebase-js-sdk/blob/master/packages/rxfire/docs/auth.md

Example:
Instead of 
```ts
import { authState } from 'rxfire/firestore';
```

It should be 
```ts
import { authState } from 'rxfire/auth';
```

All imports seem to be referring to `firestore` instead of `auth`.

Feiyang1:
Hi @dungahk, Thanks for pointing it out and PR is welcomed!

davideast:
@dungahk I would totally love an accept a PR here!

dungahk:
Done!
Btw, great job guys, RxFire is awesome!

dungahk:
Closed by #1297

PR #1297, opened 2018-10-09
Fix RxFire auth docs with wrong import path

dungahk:
The imports should be `rxfire/auth` instead of `rxfire/firestore`.
This could lead beginner developers to frustration since they wouldn't be able to use the examples from the docs successfully.

### Discussion

Issue raised #1292


dungahk:
I signed it!

PR #1299, opened 2018-10-09
Fix error in Memory LRU implementation

gsoltis:
Iterating orphaned documents was returning all documents, not just orphaned ones. They weren't being incorrectly GC'd, but they were incorrectly counting towards sequence numbers to GC. 

The test that caught this is coming in the port of https://github.com/firebase/firebase-ios-sdk/pull/1905

Additionally, I implemented `Symbol.iterator` for `ObjectMap` so that I could use `PersistencePromise.forEach()`.

### DIFF
--- a/packages/firestore/src/util/obj_map.ts
+++ b/packages/firestore/src/util/obj_map.ts
@@ -17,7 +17,7 @@
 import { Equatable } from './misc';
 import * as objUtil from './obj';
 
-export type Entry<K, V> = [K, V];
+type Entry<K, V> = [K, V];
 
 /**
  * A map implementation that uses objects as keys. Objects must implement the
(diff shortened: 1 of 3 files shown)


packages/firestore/src/util/obj_map.ts
```diff
@@ -17,7 +17,7 @@
 import { Equatable } from './misc';
 import * as objUtil from './obj';
 
-type Entry<K, V> = [K, V];
+export type Entry<K, V> = [K, V];
```
schmidt-sebastian: We already export a type Entry in sorted_map.ts. I would actually prefer if we use that type since it's easier to digest (it's an object on not an array)

packages/firestore/src/util/obj_map.ts
```diff
@@ -107,4 +107,64 @@ export class ObjectMap<KeyType extends Equatable<KeyType>, ValueType> {
   isEmpty(): boolean {
     return objUtil.isEmpty(this.inner);
   }
+
+  [Symbol.iterator](): Iterator<{ key: KeyType; value: ValueType }> {
+    // We don't care about the keys, all of the actual keys are in the
+    // arrays that are the values of the inner object.
+    const it = objUtil.values(this.inner)[Symbol.iterator]();
```
schmidt-sebastian: This creates a copy of the value array, which is a little surprising for an iterator. I don't know a good way around this without using generators (which I assume we can't), so I am going to ignore this.

I do think this could be simplified quite a bit though. Wouldn't this work just as well:

```
    const values : Array<{ key: KeyType; value: ValueType }> = [];
    this.forEach((key, value) => values.push({key, value}));
    return values[Symbol.iterator]();
```

You might not need the ObjectMapIterator.

packages/firestore/src/util/obj_map.ts
```diff
@@ -107,4 +107,64 @@ export class ObjectMap<KeyType extends Equatable<KeyType>, ValueType> {
...
+    return new ObjectMapIterator<KeyType, ValueType>(it);
+  }
+}
+
+/**
+ * Implements an Iterator over ObjectMap.
+ */
+export type IteratorEntry<KeyType, ValueType> = {
```
schmidt-sebastian: FWIW, This is the Entry type from SortedMap.

packages/firestore/src/util/obj_map.ts
gsoltis: Unfortunately, I don't know a way to avoid creating the array of values. Objects don't support iterators by default, and there's no ordering either to be able to start and stop iteration through the object.

As for using `forEach`, it kinda depends on whether or not we think there are collisions. `ObjectMapIterator` doesn't need to create all of the entries that will be returned at once. That's only a benefit though if there are collisions, since we do need the values array.

I'm not sure I have a strong opinion either way, since it depends on what's stored. Creating the whole array up front does have the advantage of not needing to export an extra `Entry` type, I suppose. 

Do you feel strongly either way?

packages/firestore/src/util/obj_map.ts
gsoltis: Yup, but I don't think a dependency between them is appropriate.

packages/firestore/src/util/obj_map.ts
schmidt-sebastian: If we do keep it, then please rename it. Having two Entry types that are very similar will make auto-importing a bit tricky.

packages/firestore/src/util/obj_map.ts
schmidt-sebastian: The two solutions (copy every value and copy every set of matching values) have such different complexities that I would much prefer to pick the simpler one.  
We can look at the usage in the SDK and see how common collisions are, but from memory it seems that they should be pretty rare (my IDE doesn't open at the moment...). The overhead from an occasional canonical ID collision can probably be ignored.

gsoltis: Ok, I went with the `forEach` method.

PR #1307, opened 2018-10-12
Removed warning messages when using testing SDK

scottcrossen:
Alternatively we don't need to set the default log level in the setLogLevel function (and just export two things)

### DIFF
--- a/packages/logger/index.ts
+++ b/packages/logger/index.ts
@@ -14,13 +14,12 @@
  * limitations under the License.
  */
 
-import { instances, LogLevel, setDefaultLogLevel } from './src/logger';
+import { instances, LogLevel } from './src/logger';
 
 export function setLogLevel(level: LogLevel) {
   instances.forEach(inst => {
     inst.logLevel = level;
   });
-  setDefaultLogLevel(level);
 }
 
 export { Logger, LogLevel, LogHandler } from './src/logger';
--- a/packages/logger/index.ts
+++ b/packages/logger/index.ts
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-import { instances, LogLevel, setDefaultLogLevel} from './src/logger';
+import { instances, LogLevel, setDefaultLogLevel } from './src/logger';
 
 export function setLogLevel(level: LogLevel) {
   instances.forEach(inst => {
     inst.logLevel = level;
   });
-  setDefaultLogLevel(level)
+  setDefaultLogLevel(level);
 }
 
 export { Logger, LogLevel, LogHandler } from './src/logger';
--- a/packages/logger/index.ts
+++ b/packages/logger/index.ts
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-import { instances, LogLevel } from './src/logger';
+import { instances, LogLevel, setDefaultLogLevel} from './src/logger';
 
 export function setLogLevel(level: LogLevel) {
   instances.forEach(inst => {
     inst.logLevel = level;
   });
+  setDefaultLogLevel(level)
 }
 
 export { Logger, LogLevel, LogHandler } from './src/logger';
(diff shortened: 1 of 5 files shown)


packages/testing/src/api/index.ts
```diff
@@ -18,11 +18,14 @@ import { firebase } from '@firebase/app';
 import request from 'request-promise';
 import { FirebaseApp, FirebaseOptions } from '@firebase/app-types';
 import { base64 } from '@firebase/util';
+import { setLogLevel, LogLevel } from '@firebase/logger';
 
 /** The default url for the local database emulator. */
 const DBURL = 'http://localhost:9000';
 /** Passing this in tells the emulator to treat you as an admin. */
 const ADMIN_TOKEN = 'owner';
+/** Warnings aren't needed when product is in testing. */
```
mikelehen: I'm curious what warnings exactly are you trying to suppress?  Normally I'd want a higher log level in testing, not lower...

mikelehen: Ahhhhh...  Yep, I see what you mean there. :-)

scottcrossen: Thanks Ryan for the prompt response!

packages/logger/index.ts
```diff
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-import { instances, LogLevel } from './src/logger';
+import { instances, LogLevel, setDefaultLogLevel } from './src/logger';
```
Feiyang1: Default log level should not be changed. It's meant to be a constant. To change the the log level, just change the value of the inst.logLevel.

packages/logger/src/logger.ts
```diff
@@ -42,7 +42,14 @@ export enum LogLevel {
 /**
  * The default log level
  */
-const defaultLogLevel: LogLevel = LogLevel.INFO;
+let defaultLogLevel: LogLevel = LogLevel.INFO;
```
Feiyang1: Default log level is a constant and should not be changed

scottcrossen: Resolving to consolidate similiar threads.

packages/logger/index.ts
scottcrossen: Yes, inst.logLevel works retroactively but fails to mute future instantiations of the logger. Under this design, changing the log-level globally from the "setLogLevel" function now requires either the implementer to design a monitoring service around the logger that changes newly-created instances back to the desired level or for each requiring class to be contracted to provide a logLevel from it's parent instantiation.

Seems like a hassle to me. Still, you're the maintainer of the code and if that's how you say it's gonna be then I'll make it happen. I'm open to suggestions on how to solve the global log level puzzle as well, but in the meantime I'll change this PR (within the next few minutes) with an alternative approach to spam the provided setLogLevel function every time the testing-sdk creates a new logger instance (which is only referenced by exported functions and not the database instance itself). It's not ideal for code quality but at least it will work.

scottcrossen:
Feiyang1 ptal

scottcrossen:
GitHub collapsed our previous comment thread when it should still be valid. Please review my comments there as well.

packages/logger/index.ts
mikelehen: FWIW- It seems like with the current design / implementation, `setLogLevel()` is really `setLogLevelOfAllExistingLoggerInstances()` and `setDefaultLogLevel()` is really `setLogLevelOfAllFutureLoggerInstances()`.

I'm not sure why we'd want to make this distinction.  I think calling `setLogLevel()` should change the log level for all existing and future instances.  So I think `defaultLogLevel` should really be `currentLogLevel` and `setLogLevel()` should go ahead and update it in addition to notifying all existing instances of the new log level.

scottcrossen: That's basically what I originally did! :1st_place_medal: I just didn't rename defaultLogLevel to currentLogLevel. I've made changes with an alterante method. I would prefer doing what you suggest. Lets see if @Feiyang1  has any thoughts!

PR #1311, opened 2018-10-17
update storage owner

Feiyang1:
Updating storage owner.

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -102,7 +102,7 @@ function initializeApp(
       timestampsInSnapshots: true
     });
   }
-/**
+  /**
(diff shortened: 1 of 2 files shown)


mikelehen:
@schmidt-sebastian I assume you're okay with this? 😀

packages/testing/src/api/index.ts
```diff
@@ -102,7 +102,7 @@ function initializeApp(
       timestampsInSnapshots: true
     });
   }
-/**
+  /**
```
mikelehen: I'm not sure why this changed?  It looks misaligned to me...

Feiyang1: It's the prettifier. Now it's aligned with the function it's commenting on. not sure how it was not caught before.

PR #1323, opened 2018-10-22
Introduce protos to communicate with the Firestore Emulator

ryanpbrewster:
This does two things:

In the short-term, this gives us the ability to set Firestore rules from the SDK, mirroring the database functionality

In the long-term, this proto API will give us the ability to expose interesting non-production functionality from the emulators.

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -15,10 +15,15 @@
  */
 
 import { firebase } from '@firebase/app';
-import request from 'request-promise';
+import * as request from 'request';
 import { FirebaseApp, FirebaseOptions } from '@firebase/app-types';
 import { base64 } from '@firebase/util';
 import { setLogLevel, LogLevel } from '@firebase/logger';
+import * as grpc from 'grpc';
+import * as fs from 'fs';
+
+const PROTOS = grpc.load({ root: __dirname + "/../protos", file: "google/firestore/emulator/v1/firestore_emulator.proto" });
+const EMULATOR = PROTOS["google"]["firestore"]["emulator"]["v1"];
 
 /** If this environment variable is set, use it for the database emulator's address. */
 const DATABASE_ADDRESS_ENV: string = 'FIREBASE_DATABASE_EMULATOR_ADDRESS';
@@ -128,24 +133,57 @@ export type LoadDatabaseRulesOptions = {
   databaseName: string;
   rules: string;
 };
-export function loadDatabaseRules(options: LoadDatabaseRulesOptions): void {
+export function loadDatabaseRules(options: LoadDatabaseRulesOptions): Promise<void> {
   if (!options.databaseName) {
-    throw new Error('databaseName not specified');
+    throw Error('databaseName not specified');
   }
 
   if (!options.rules) {
-    throw new Error('must provide rules to loadDatabaseRules');
+    throw Error('must provide rules to loadDatabaseRules');
   }
 
-  request({
-    uri: `http://${DATABASE_ADDRESS}/.settings/rules.json?ns=${
-      options.databaseName
-    }`,
-    method: 'PUT',
-    headers: { Authorization: 'Bearer owner' },
-    body: options.rules
-  }).catch(function(err) {
-    throw new Error('could not load rules: ' + err);
+  return new Promise((resolve, reject) => {
+    request.put({
+      uri: `http://${DATABASE_ADDRESS}/.settings/rules.json?ns=${
+        options.databaseName
+      }`,
+      headers: { Authorization: 'Bearer owner' },
+      body: options.rules
+    }, (err, resp, body) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
+export type LoadFirestoreRulesOptions = {
+  projectId: string;
+  rules: string;
+};
+export function loadFirestoreRules(options: LoadFirestoreRulesOptions): Promise<void> {
+  if (!options.projectId) {
+    throw new Error('projectId not specified');
+  }
+
+  if (!options.rules) {
+    throw new Error('must provide rules to loadFirestoreRules');
+  }
+
+  let client = new EMULATOR.FirestoreEmulator("localhost:8080", grpc.credentials.createInsecure());
+  return new Promise((resolve, reject) => {
+    client.setSecurityRules({
+      project: "projects/bar",
+      rules: { files: [{ content: options.rules }] }
+    }, (err, resp) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve(resp);
+      }
+    });
   });
 }
 
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -183,13 +183,13 @@ export function loadFirestoreRules(
   }
 
   let client = new EMULATOR.FirestoreEmulator(
-    'localhost:8080',
+    FIRESTORE_ADDRESS,
     grpc.credentials.createInsecure()
   );
   return new Promise((resolve, reject) => {
     client.setSecurityRules(
       {
-        project: 'projects/bar',
+        project: `projects/${options.projectId}`,
         rules: { files: [{ content: options.rules }] }
       },
       (err, resp) => {
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -22,8 +22,11 @@ import { setLogLevel, LogLevel } from '@firebase/logger';
 import * as grpc from 'grpc';
 import * as fs from 'fs';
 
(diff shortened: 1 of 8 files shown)


packages/testing/src/api/index.ts
```diff
@@ -15,10 +15,18 @@
...
-import request from 'request-promise';
+import * as request from 'request';
 import { FirebaseApp, FirebaseOptions } from '@firebase/app-types';
 import { base64 } from '@firebase/util';
 import { setLogLevel, LogLevel } from '@firebase/logger';
+import * as grpc from 'grpc';
+import * as fs from 'fs';
+
+const PROTOS = grpc.load({
+  root: __dirname + '/../protos',
```
tonymeng: what should this be? can you either get rid of the relative path or have a comment explaining what it is

packages/testing/src/api/index.ts
```diff
@@ -15,10 +15,18 @@
...
+  file: 'google/firestore/emulator/v1/firestore_emulator.proto'
+});
+const EMULATOR = PROTOS['google']['firestore']['emulator']['v1'];
```
tonymeng: are each of these all guaranteed to exist?

packages/testing/src/api/index.ts
```diff
@@ -128,24 +136,70 @@ export type LoadDatabaseRulesOptions = {
...
-    uri: `http://${DATABASE_ADDRESS}/.settings/rules.json?ns=${
-      options.databaseName
-    }`,
-    method: 'PUT',
-    headers: { Authorization: 'Bearer owner' },
-    body: options.rules
-  }).catch(function(err) {
-    throw new Error('could not load rules: ' + err);
+  let client = new EMULATOR.FirestoreEmulator(
+    'localhost:8080',
```
tonymeng: we allow the customer to change the port that the emulator runs on

packages/testing/src/api/index.ts
```diff
@@ -128,24 +136,70 @@ export type LoadDatabaseRulesOptions = {
...
+    grpc.credentials.createInsecure()
+  );
+  return new Promise((resolve, reject) => {
+    client.setSecurityRules(
+      {
+        project: 'projects/bar',
```
tonymeng: ??

packages/testing/src/protos/google/firestore/emulator/v1/firestore_emulator.proto
```diff
@@ -0,0 +1,42 @@
...
+  // Format: `projects/{project_id}
+  string project = 1;
+
+  // The ruleset to use for the project
+  google.firebase.rules.v1.Source rules = 2;
+}
+
+message ClearDataRequest {
+  // Format: `projects/{project_id}/databases/(default)`
+  string database = 1;
```
tonymeng: call this project

packages/testing/src/protos/google/protobuf/empty.proto
```diff
@@ -0,0 +1,52 @@
+// Protocol Buffers - Google's data interchange format
```
tonymeng: can we include where this came from (somewhere, doesn't have to be in this file)

packages/testing/src/protos/google/firebase/rules/v1/source.proto
```diff
@@ -0,0 +1,60 @@
...
+package google.firebase.rules.v1;
+
+option java_package = "com.google.firebase.rules.v1";
+option java_multiple_files = true;
+option java_outer_classname = "SourceProto";
+
+// `Source` is one or more `File` messages comprising a logical set of rules.
+message Source {
+  // `File` set constituting the `Source` bundle.
+  repeated File files = 1;
```
tonymeng: why would this be repeated?

packages/testing/src/protos/google/firebase/rules/v1/source.proto
```diff
@@ -0,0 +1,60 @@
...
+  // File name.
+  string name = 2;
+
+  // Fingerprint (e.g. github sha) associated with the `File`.
+  bytes fingerprint = 3;
+}
+
+// Position in the `Source` content including its line, column number, and an
+// index of the `File` in the `Source` message. Used for debug purposes.
+message SourcePosition {
```
tonymeng: this doesn't look to be used

packages/testing/src/api/index.ts
```diff
@@ -128,24 +136,70 @@ export type LoadDatabaseRulesOptions = {
   databaseName: string;
   rules: string;
 };
-export function loadDatabaseRules(options: LoadDatabaseRulesOptions): void {
+export function loadDatabaseRules(
+  options: LoadDatabaseRulesOptions
+): Promise<void> {
   if (!options.databaseName) {
-    throw new Error('databaseName not specified');
+    throw Error('databaseName not specified');
```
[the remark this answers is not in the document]
ryanpbrewster: Specifically, if I change these to `return Promise.reject('databaseName not specified')` then I have to change the tests from `expect(...).to.throw` to `expect(...).to.be.rejectedWith`, and it complains that it's not `thenable`, which I think means it's not returning a valid promise.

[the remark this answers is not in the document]
ryanpbrewster: The counterpoint would be that these exceptions are logic errors, and the failed promise is a network error, so these are more like assertions than errors.

packages/testing/src/api/index.ts
ryanpbrewster: :scream_cat: fixed

packages/testing/src/protos/google/firebase/rules/v1/source.proto
ryanpbrewster: That's just the API that they came up with. I think it's forward compatibility with multi-file rulesets (once we have a real module system).

packages/testing/src/protos/google/firebase/rules/v1/source.proto
ryanpbrewster: This is a pre-existing proto file. I don't want to modify it.

packages/testing/src/protos/google/firestore/emulator/v1/firestore_emulator.proto
ryanpbrewster: It should be database, projects do not contain data.

packages/testing/src/protos/google/protobuf/empty.proto
ryanpbrewster: I'll add a README

packages/testing/src/api/index.ts
ryanpbrewster: Good point, changed this to `FIRESTORE_ADDRESS`

packages/testing/src/api/index.ts
ryanpbrewster: Kinda. The `protos/` directory is structured such that these do exist. If they don't, it is a build error on our part.

packages/testing/src/api/index.ts
ryanpbrewster: I'm following the existing setup in `packages/firestore`, which does something similar. This is basically the cleanest way of "hardcoding" the path to `src/protos/`

packages/testing/src/protos/google/firestore/emulator/v1/firestore_emulator.proto
tonymeng: i see, how do you get what project it is then?

packages/testing/src/protos/google/firebase/rules/v1/source.proto
tonymeng: i see, where does this come from? do we need licenses or anything?

packages/testing/src/api/index.ts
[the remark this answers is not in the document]
scottcrossen: I'm cool with that answer. The **"thenable"** error might have something to do with it expecting a
```Promise.reject(new Error('databaseName not specified'))```
rather than
```Promise.reject('databaseName not specified')```
Not sure though. If they act more like assertions then that's probably good as is.

File: packages/testing/src/protos/google/firebase/rules/v1/source.proto (hunk 8 above)
Comment (ryanpbrewster) [reply packages/testing/src/protos/google/firebase/rules/v1/source.proto]: http://google3/google/firebase/rules/v1/source.proto

Added the licenses at the top of each proto file, good call.

File: packages/testing/src/protos/google/firestore/emulator/v1/firestore_emulator.proto (hunk 5 above)
Comment (ryanpbrewster) [reply packages/testing/src/protos/google/firestore/emulator/v1/firestore_emulator.proto]: Note that the database contains the project inside of it. So the database
```
projects/ryan-test/databases/(default)
```
maps to the project
```
projects/ryan-test
```

tonymeng: ah gotcha

Issue #1319, opened 2018-10-19
findIndex not supported by IE11


stucker71:
This is supposed to be compatible with IE11 but findIndex is used which IE11 does not support. Please fix!

I'm using latest version firebase 5.5.5 - firestore

PR #1328, opened 2018-10-23
Remove non-test only invocation of findIndex

schmidt-sebastian:
This is the only invocation of `findIndex` in a non test-only path. Alternatively, we could also rewrite this functionality.

Fixes https://github.com/firebase/firebase-js-sdk/issues/1319

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -260,6 +260,13 @@ export class AsyncQueue {
       `Attempted to schedule an operation with a negative delay of ${delayMs}`
     );
 
+    // While not necessarily harmful, we currently don't expect to have multiple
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
FYI: I changed my mind on this since we probably do want to support testing on IE11.

mmermerkaya:
FWIW, findIndex is in our polyfills already: https://github.com/firebase/firebase-js-sdk/blob/master/packages/polyfill/index.ts#L25

schmidt-sebastian:
> FWIW, findIndex is in our polyfills already: https://github.com/firebase/firebase-js-sdk/blob/master/packages/polyfill/index.ts#L25

It seems like people are somehow using Firestore without a dependency on Firebase App.

Issue #1308, opened 2018-10-15
Firestore 5.5.3 uses Symbol(Symbol.iterator) not supported by IE 11


harveylee:
* Operating System version: Windows 10
  * Browser version: Internet Explorer 11.345.17134.0
  * Firebase SDK version: 5.5.3
  * Firebase Product: firestore

#### Steps to reproduce:
1. Use Firebase JS SDK v5.5.3
2. Browse to page with Internet Explorer 11
3. Get error in console:
```
TypeError: Object doesn't support property or method 'Symbol(Symbol.iterator)_c.e9e7bht5pg3'
   at e.forEach (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:187742)
   at e.prototype.onTransactionCommitted (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:265406)
   at Anonymous function (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:264146)
   at Anonymous function (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:186886)
   at e.prototype.wrapUserFunction (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:186729)
   at e.prototype.wrapSuccess (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:186844)
   at e.prototype.next (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:186251)
   at e.prototype.runTransaction (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:264071)
   at e.prototype.start (https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js:1:242380)
   at Anonymo
```

Note that this issue didn't occur with v5.5.2

#### Relevant Code:
```javascript
<script type='text/javascript' src='https://www.gstatic.com/firebasejs/5.5.3/firebase-firestore.js'></script>
```

clintjansen:
Getting a similar error in React Native (Expo) after 5.3.1+ using the WebSDK

cmacdonnacha:
I'm getting the same issue. Reverting back to 5.5.1 and it works

schmidt-sebastian:
Thanks for reporting! We do have a Polyfill that is supposed to fix this. I will take a look tomorrow.

schmidt-sebastian:
[...]

PR #1330, opened 2018-10-23
Remove Symbol.iterator

schmidt-sebastian:
This allows use of Firestore in IE11 without Polyfills.

Fixes https://github.com/firebase/firebase-js-sdk/issues/1308

### DIFF
--- a/packages/firestore/src/local/persistence_promise.ts
+++ b/packages/firestore/src/local/persistence_promise.ts
@@ -170,56 +170,31 @@ export class PersistencePromise<T> {
 
   static waitFor(
     // tslint:disable-next-line:no-any Accept all Promise types in waitFor().
-    all: Iterable<PersistencePromise<any>>
+    all: { forEach: (cb: ((el: PersistencePromise<any>) => void)) => void }
   ): PersistencePromise<void> {
-    const it = all[Symbol.iterator]();
     return new PersistencePromise<void>((resolve, reject) => {
       let expectedCount = 0;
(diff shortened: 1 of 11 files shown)


packages/firestore/src/local/persistence_promise.ts
```diff
@@ -250,18 +225,15 @@ export class PersistencePromise<T> {
    * PersistencePromises to resolve.
    */
   static forEach<K>(
-    collection: Iterable<K>,
-    f: (item: K) => PersistencePromise<void>
+    // tslint:disable-next-line:no-any Accept any kind of `forEach`.
+    collection: { forEach: ((cb: ((k: K, ...args: any[]) => void)) => void) },
+    // tslint:disable-next-line:no-any Accept any kind of callback.
+    f: (k: K, ...args: any[]) => PersistencePromise<void>
```
mikelehen: Uh...  these type gymnastics are pretty crazy.  Seems like there has to be a better way?  

Maybe just define our own Iterable type (with a normal method instead of Symbol.iterator)?

FWIW I wouldn't be surprised if doing `...args` in an inner loop is a mildly bad practice for performance reasons.  Though I'd have to look at the generated code to be sure.

schmidt-sebastian: I changed the type a bit to remove the varargs. This gets us saner types and a bit more type safety (see the related changes in the last commit).

The generated code uses `apply()`.

PR #1331, opened 2018-10-23
Switch @firebase/testing dependency from '@firebase/app' to 'firebase'

ryanpbrewster:
This allows us to avoid a few hacky `(xyz as any)` conversions, at the (hopefully mild) cost of a larger dependency for testing.

### DIFF
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -11,18 +11,20 @@
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
-    "test": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --compilers ts:ts-node/register/type-check --retries 5 --timeout 5000 --exit",
+    "test": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --require ts-node/register/type-check --retries 5 --timeout 5000 --exit",
     "prepare": "npm run build"
   },
   "license": "Apache-2.0",
   "dependencies": {
     "@firebase/logger": "0.1.1",
     "@firebase/util": "0.2.2",
-    "request": "2.88.0",
-    "request-promise": "4.2.2"
+    "@types/request": "^2.47.1",
+    "grpc": "1.13.1",
+    "request": "2.88.0"
   },
   "devDependencies": {
     "chai": "4.1.2",
+    "chai-as-promised": "^7.1.1",
     "mocha": "5.0.5",
     "nyc": "11.6.0",
     "rollup": "0.57.1",
--- a/packages/testing/package.json
+++ b/packages/testing/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@firebase/testing",
   "version": "0.2.0",
-  "private": true,
   "description": "",
   "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
   "main": "dist/index.cjs.js",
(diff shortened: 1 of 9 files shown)


packages/testing/package.json
```diff
@@ -15,9 +15,10 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "firebase": "5.5.5",
     "@firebase/logger": "0.1.1",
     "@firebase/util": "0.2.2",
-    "@types/request": "^2.47.1",
+    "@types/request": "2.47.1",
```
ryanpbrewster: we are supposed to pin specific versions here

packages/testing/package.json
```diff
@@ -28,12 +29,11 @@
     "nyc": "11.6.0",
     "rollup": "0.57.1",
     "rollup-plugin-commonjs": "9.1.0",
+    "rollup-plugin-copy-assets": "1.0.0",
     "rollup-plugin-node-resolve": "3.3.0",
-    "rollup-plugin-typescript2": "0.12.0"
-  },
-  "peerDependencies": {
```
ryanpbrewster: for testing the size savings are not important and this makes it slightly simpler for our clients

packages/testing/rollup.config.js
```diff
@@ -33,6 +35,13 @@ const external = Object.keys(
...
   input: 'index.ts',
   output: [{ file: pkg.main, format: 'cjs' }],
-  plugins,
+  plugins: [
+    ...plugins,
+    // Needed as we also use the *.proto files
+    copy({ assets: ['./src/protos'] }),
+    replace({
+      'process.env.FIRESTORE_EMULATOR_PROTO_ROOT': JSON.stringify('src/protos')
+    })
```
ryanpbrewster: apparently we have to package the .proto files with the compiled javascript

packages/testing/src/api/index.ts
```diff
@@ -14,18 +14,23 @@
...
 import * as grpc from 'grpc';
+import { resolve } from 'path';
 import * as fs from 'fs';
 
-const PROTOS = grpc.load({
-  root: __dirname + '/../protos',
+const PROTO_ROOT = {
+  root: resolve(
+    __dirname,
+    process.env.FIRESTORE_EMULATOR_PROTO_ROOT || '../protos'
```
ryanpbrewster: the protos are in the `src` directory during tests but in a different places when we package and distribute them, so this is a workaround. This is how `@firebase/firestore` handles this issue.

packages/testing/src/api/index.ts
```diff
@@ -117,8 +121,15 @@ function initializeApp(
...
+  if (databaseName) {
+    // Toggle network connectivity to force a reauthentication attempt.
+    // This mitigates a minor race condition where the client can send the
+    // first database request before authenticating.
+    app.database().goOffline();
+    app.database().goOnline();
+  }
   if (projectId) {
-    (app as any).firestore().settings({
+    app.firestore().settings({
```
tonymeng: is most of this work for removing this single hack?

ryanpbrewster:
Yes, they no longer have to `require("@firebase/firestore")` just for the side-effects

packages/testing/src/api/index.ts
```diff
@@ -66,7 +70,7 @@ function createUnsecuredJwt(auth: object): string {
   ].join('.');
 }
 
-export function apps(): (FirebaseApp | null)[] {
```
tonymeng: is this no longer capable of returning null?

ryanpbrewster: Correct. I have no idea why this was ever returning an array of potentially null objects. That was pretty dangerous, since we were calling `.delete()` on the elements of this array, so those calls would throw exceptions if the element were `null`.

ryanpbrewster:
Yes, we have docs and quickstarts that we'll publish shortly. I'll update the `README.md` file here with links when they go live.

PR #1332, opened 2018-10-24
Export more functions from the firebase module

ryanpbrewster:
Courtesy of @tstirrat

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -22,6 +22,8 @@ import * as grpc from 'grpc';
 import { resolve } from 'path';
 import * as fs from 'fs';
 
+export { database, firestore } from 'firebase';
+
 const PROTO_ROOT = {
   root: resolve(
     __dirname,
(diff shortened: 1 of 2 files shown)


packages/testing/test/database.test.ts
```diff
@@ -124,4 +122,13 @@ describe('Testing Module Tests', function() {
     await firebase.initializeTestApp({ databaseName: 'bar', auth: {} });
     expect(firebase.apps().length).to.equal(numApps + 2);
   });
+
+  it('there is a way to get database timestamps', function() {
+    expect(firebase.database.ServerValue.TIMESTAMP).to.deep.equal({
+      '.sv': 'timestamp'
+    });
+  });
```
tonymeng: newline

packages/testing/src/api/index.ts
```diff
@@ -22,6 +22,8 @@ import * as grpc from 'grpc';
 import { resolve } from 'path';
 import * as fs from 'fs';
 
+export { database, firestore } from 'firebase';
```
tonymeng: why is this here?

ryanpbrewster: To expose the functions inside of these submodules without having to explicitly import them (for our clients)

PR #1335, opened 2018-10-24
update yarn.lock

Feiyang1:


mikelehen:
What exactly generated these diffs?  I thought we used explicit package versions in all of our package.json dependencies so that everything is frozen unless we intentionally upgrade.

ryanpbrewster:
Not sure about all of them, but `nyc@13.1` came because I upgraded the testing package dev dependencies to silence `npm audit` alerts every time I ran tests

ryanpbrewster:
Yep. Is there some way we can add a CI check that the `yarn.lock` file is up-to-date?

Feiyang1:
As @ryanpbrewster mentioned, a few dependencies are added/changed in testing package.
https://github.com/firebase/firebase-js-sdk/commit/6ebdb03d10a112190a138ed2f46d0ad017c39d3a
https://github.com/firebase/firebase-js-sdk/commit/ffd1526420a39a064ebdafcba02214141530a8f0
https://github.com/firebase/firebase-js-sdk/commit/0e85e42e84a592a6ce2028685983a22cf8db6fec

We should always check in the updated yarn.lock file when any dependency changes. just need to make sure we use the same version of yarn.

Feiyang1:
That's a good idea. I will look into adding the check.

Issue #1111, opened 2018-08-10
Typings not being told in package.json


degroote22:
[Here](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/auth/package.json) it doesn't have it.

But [here](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/app/package.json#L6) it does.

The Typescript compiler will warn me if I do something like this:
```typescript
const importFirebase = async () => {
  // dynamic import, will return promise
  const fb = await import("firebase/app");
  await import("firebase/auth"); // <-- Error: [ts] Could not find a declaration file for module 'firebase/auth'.
  // 'xxx/node_modules/firebase/auth/dist/index.cjs.js' implicitly has an 'any' type.
  return fb;
};
```

I'm using it with next.js and I don't know why but their's compilation step will fail if I don't add the typing lines to the related "package.json".

Is there a reason not to add the typings to other packages (auth, database, storage don't have it, for instance)?
If there isn't I'd be happy to submit a PR but I wouldn't mind if you did it by yourselves 👍

Feiyang1:
Hi @degroote22, firebase component packages don't export anything. What they do is patching the firebase namespace you import from firebase/app, so it is correct that they don't have any typings (because they don't have export).
You can try 
```javascript
(await import("firebase/auth")) as any
```
Hope it helps!

degroote22:
It did not make the compiler error go away.
My fix was creating a .d.ts file with this content
```typescript
declare module "firebase/storage" {

}
declare module "firebase/auth" {

}

```

ifiokjr:
The issue can be seen in the rxfire examples.

https://github.com/davideast/rxfire-samples/blob/f4b0b6b7959b7fb52b04fb7a8041edb17b92a872/src/preact/index.tsx#L20-L23

PR #1347, opened 2018-10-30
Create empty-import.d.ts and reference it in submodule imports

ifiokjr:
* Fixes https://github.com/firebase/firebase-js-sdk/issues/1111 

A full reproduction of the error is available below 

https://github.com/davideast/rxfire-samples/blob/f4b0b6b7959b7fb52b04fb7a8041edb17b92a872/src/preact/index.tsx#L20-L23

```typescript
// NOTE: The lazy load for Firestore does not work with TypeScipt
// w/out hacking the node_modules/firebase/firestore/package.json file
// to use "typings": "../index.d.ts".
const firestore$ = from(import('firebase/firestore'));
```

### DIFF
--- a/packages/firebase/auth/package.json
+++ b/packages/firebase/auth/package.json
@@ -1,5 +1,6 @@
 {
   "name": "firebase/auth",
   "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js"
+  "module": "dist/index.esm.js",
+  "typings": "../empty-import.d.ts"
 }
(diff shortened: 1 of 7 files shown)


Feiyang1:
Awesome, LGTM. Thank you very much. Can you please close this one and create a new PR on top of the HEAD? We had a hiccup in CI build, so the build fails for this PR.

ifiokjr:
@Feiyang1 thanks approving! 👍 
@davideast Any chance this could be merged soon. It'll save me from the ugly post-install bash script I'm currently using to bypass the issue.

Issue #1113, opened 2018-08-10
grpc deprecated error sdk 5.3.1 


ShailenNaidoo:
### [REQUIRED] Describe your environment

  * Operating System version: Ubuntu 18.04
  * Browser version: chrome 58
  * Firebase SDK version: 5.3.1
  * Firebase Product: firestore and functions

### [REQUIRED] Describe the problem
GRPC deprecated error

#### Steps to reproduce:
Install the sdk and use the firestore sdk

wilhuff:
Could you give some more details on where you're seeing this and how you're installing it? Are you using node.js? Where did you see the deprecated error?

Here's what I see when I install, which doesn't say anything about deprecation:
```
npm install firebase --save

> grpc@1.13.1 install /Users/mcg/tmp/node/node_modules/grpc
> node-pre-gyp install --fallback-to-build --library=static_library

node-pre-gyp WARN Using needle for node-pre-gyp https download 
[grpc] Success: "/Users/mcg/tmp/node/node_modules/grpc/src/node/extension_binary/node-v57-darwin-x64-unknown/grpc_node.node" is installed via remote

+ firebase@5.3.1
```

PR #1353, opened 2018-11-02
Upgrade GRPC to 1.16.0

squirly:
This removes a deprecated call used internally by GRPC, fixing the deprecation warning.

Fixes https://github.com/firebase/firebase-js-sdk/issues/1113


mikelehen:
Thanks!  Do you mind upgrading packages/testing/package.json as well, so the whole SDK is on the same version?

mikelehen:
Did you regenerate yarn.lock after upgrading packages/testing ? I expected to see 1.13.1 disappear.

squirly:
@mikelehen Updated!

Djiit:
Hi folks, thanks for that !

PR #1355, opened 2018-11-03
Use orphaned docs as part of GC calculation

gsoltis:
Port of https://github.com/firebase/firebase-android-sdk/pull/80

We had originally decided to only look at targets when assessing how much to run GC. However, based on running a write-heavy workload test, we need to include documents that ended up cached without being part of a target.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -1075,8 +1075,19 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
     this.garbageCollector = new LruGarbageCollector(this);
   }
 
-  getTargetCount(txn: PersistenceTransaction): PersistencePromise<number> {
-    return this.db.getQueryCache().getQueryCount(txn);
+  getSequenceNumberCount(txn: PersistenceTransaction): PersistencePromise<number> {
+    const docCountPromise = this.orphanedDocmentCount(txn);
+    const targetCountPromise = this.db.getQueryCache().getQueryCount(txn);
+    return targetCountPromise.next(targetCount => 
+      docCountPromise.next(docCount => targetCount + docCount)
+    );
+  }
+
+  private orphanedDocmentCount(txn: PersistenceTransaction): PersistencePromise<number> {
+    let orphanedCount = 0;
+    return this.forEachOrphanedDocumentSequenceNumber(txn, _ => {
+      orphanedCount++;
+    }).next(() => orphanedCount);
   }
 
   forEachTarget(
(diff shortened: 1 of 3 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1075,8 +1075,23 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
...
   }
 
-  getTargetCount(txn: PersistenceTransaction): PersistencePromise<number> {
-    return this.db.getQueryCache().getQueryCount(txn);
+  getSequenceNumberCount(
+    txn: PersistenceTransaction
+  ): PersistencePromise<number> {
+    const docCountPromise = this.orphanedDocmentCount(txn);
+    const targetCountPromise = this.db.getQueryCache().getQueryCount(txn);
+    return targetCountPromise.next(targetCount =>
```
schmidt-sebastian: We usually don't include `Promise` in the name. You can respond to this by just inlining the calls though :)

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -317,8 +317,25 @@ export class MemoryLruDelegate implements ReferenceDelegate, LruDelegate {
...
+    return targetCountPromise.next(targetCount =>
+      docCountPromise.next(docCount => targetCount + docCount)
+    );
+  }
+
+  private orphanedDocumentCount(
+    txn: PersistenceTransaction
+  ): PersistencePromise<number> {
+    let orphanedCount = 0;
+    return this.forEachOrphanedDocumentSequenceNumber(txn, _ => {
```
schmidt-sebastian: Isn't this use `this.orphanedSequenceNumbers.size()`?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -317,8 +317,25 @@ export class MemoryLruDelegate implements ReferenceDelegate, LruDelegate {
...
-  getTargetCount(txn: PersistenceTransaction): PersistencePromise<number> {
-    return this.persistence.getQueryCache().getTargetCount(txn);
+  getSequenceNumberCount(
+    txn: PersistenceTransaction
+  ): PersistencePromise<number> {
+    const docCountPromise = this.orphanedDocumentCount(txn);
+    const targetCountPromise = this.persistence
+      .getQueryCache()
+      .getTargetCount(txn);
+    return targetCountPromise.next(targetCount =>
```
schmidt-sebastian: Same comment regarding the naming as above.

packages/firestore/src/local/indexeddb_persistence.ts
gsoltis: I'm going to leave as-is. Inlining would put the operations in sequence, rather than in parallel. Also, the code is dealing with both the promise of the count, and the count itself, so I think it's worthwhile to distinguish them.

packages/firestore/src/local/memory_persistence.ts
gsoltis: No, `forEachOrphanedDocumentSequenceNumber` is doing some filtering to ensure that the values returned are actually orphaned.

packages/firestore/src/local/memory_persistence.ts
gsoltis: As above.

Issue #535, opened 2018-02-28
Travis CI Logs are difficult to read


jshcrowthe:
We are currently using the `spec` reporter for our CI environment. While this is consistent with our development env, it makes errors hard to find.

Consider switching reporters to only capture errors for CI envs.

yifanyang:
With this [code](https://github.com/firebase/firebase-js-sdk/compare/master...yifany-log-verbosity), the karma and mocha log for succeeded tests will not display in Travis CI environment. Local development still has the full log.

Here's an [example](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452111875) of how log look like now. Also please check out comparison between [previous log when some tests failed](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452094924) and [new log when some tests failed](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452096265).

The total length of the log is reduced, but I'm not so sure if it makes debugging any easier. Let me know what you think. @Feiyang1

PR #1372, opened 2018-11-09
Reduce log verbosity in Travis CI environment.

yifanyang:
Mocha and karma will not show logs for succeeded test cases in CI
environment. Local development still displays full logs.

The is to mitigate #535. The number of total log lines on Travis will drop by about 2/3 with this pull request.


config/mocha.browser.opts
```diff
@@ -0,0 +1,2 @@
+--timeout 20000
```
Feiyang1: should we add the google license as in the file you deleted?

config/ci.mocha.node.opts
```diff
@@ -0,0 +1,2 @@
+--reporter dot
+--exit
```
Feiyang1: can you please explain the 2 options? Is it possible to put comments in the file?

config/mocha.browser.opts
yifanyang: I think only js/ts files require a license header? Js configuration files under the 'config' directory do have a license header, while all the others don't. But I'm not sure. Your call.

P.S. Mocha version needs an upgrade to support comments in mocha config file, if we decide to add a header.

config/ci.mocha.node.opts
yifanyang: I think we can, although that requires us to upgrade mocha to the latest version (5.2.0) in all of our packages. We have 5.0.5 currently.

I can create another pull request to bump the mocha version in all of our SDK packages. This pull request will probably need approvals from all the SDK owners. Let me know if you want me to proceed with it.

PR #1373, opened 2018-11-09
Reduce log verbosity in Travis CI environment.

yifanyang:
For succeeded cases, the test in Travis CI environment will no longer display mocha framework logs (for node test) or karma framework logs (for browser test); logs generated by test code itself are still captured and displayed.

Local development still has full logs.

The is to mitigate #535. The number of total log lines on Travis will drop by about 2/3 with this pull request.


config/mocha.browser.opts
```diff
@@ -0,0 +1,2 @@
+--timeout 20000
```
Feiyang1: Do we need to include google license as in the file you deleted?

config/ci.mocha.node.opts
```diff
@@ -0,0 +1,2 @@
+--reporter dot
```
Feiyang1: what does the 2 options do? Is it possible to add comments in the configuration file?

config/mocha.browser.opts
yifanyang: Looks like only js/ts files require a license header? Js configuration files under the 'config' directory do have a license header, while all the others don't. But I'm not sure. Your call.

P.S. If we decide to add a header in this mocha config file, mocha version needs an upgrade to support the comments.

config/ci.mocha.node.opts
yifanyang: I think we can. That will require us to upgrade mocha to the latest version (5.2.0) in all of our packages. We have 5.0.5 currently.

I can create another pull request to bump the mocha version in all of our SDK packages. Does this kind of change require approvals from all the SDK owners? Or is it okay to upgrade as long as all test continue to pass?

Feiyang1: yep, it would require approvals from them. Upgrading should be fine.

yifanyang: Created #1375 to upgrade the mocha version. Feel free to add more reviewers who you think should be aware of this change to that pull request.

yifanyang: Added comments in the file to explain the usage of these flags.

yifanyang:
This pull request is to reduce the amount of logs generated during CI test. Another confusion in our log is that in case of an error, lerna will duplicate the same error messages multiple times. This is annoying especially when the log messages contains many lines (e.g. [here](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452223646) firestore logs which has thousands of lines got duplicated 4 times). This is tracked as a separate issue in lerna repo [here](https://github.com/lerna/lerna/issues/1790).

Feiyang1:
Does lerna duplicate it 4 times? I think lerna just duplicates it once per node and once for browser, so it amounts to 4 error messages?

packages/database/package.json
```diff
@@ -14,7 +14,7 @@
     "dev": "rollup -c -w",
     "test": "run-p test:browser test:node",
     "test:browser": "karma start --single-run",
-    "test:node": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --compilers ts:ts-node/register/type-check -r src/nodePatches.ts --retries 5 --timeout 5000 --exit",
+    "test:node": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --compilers ts:ts-node/register/type-check -r src/nodePatches.ts --retries 5 --timeout 5000 --opts ../../config/mocha.node.opts",
```
mikelehen: Hrm.  This sets retries / timeout and then pulls in options that are going to set different retries / timeout (and I don't know which one will win).  Same issue exists elsewhere.

FWIW- 3 retries seems excessive to me already. I don't know why we ended up with 5 retries in places...

config/ci.mocha.node.opts
```diff
@@ -0,0 +1,8 @@
+# With dot reporter, the full names of test case will not be displayed in log during execution.
+# The name and details of all failed cases will still be appended to log after execution.
+# See https://mochajs.org/#dot-matrix for more information.
+--reporter dot
+
+# Node doesn't exit after mocha test completion. Supplying this flag for force exit.
+# Not applicable in karma + mocha test (browser test).
+--exit
```
mikelehen: Am I confused or do we no longer set a timeout / retries in CI?

I'm not opposed to playing with changing these, but I'd want to do it in a separate PR since I think it's orthogonal to the logging output (well, mostly).

yifanyang:
> Does lerna duplicate it 4 times? I think lerna just duplicates it once per node and once for browser, so it amounts to 4 error messages?

> I think we also have mocha configured to retry tests 3 times:
> . So that could explain why you see 4 failures.

Not exactly. See this [example](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452183556) for better demonstration. The failed test in the example is a case in firebase/testing, for which we only have node test but no browser test:
https://github.com/firebase/firebase-js-sdk/blob/1baf7d6c18a8c314571a9f074ee7f034d5ad5250/packages/testing/package.json#L10-L15

The 4 duplicated messages are:
1. [messages](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452183556#L929) printed by mocha;
2. [messages](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452183556#L968) printed by lerna (prefixed with "lerna ERR!");
3. [messages](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452183556#L6490) printed by lerna (in the "Error" filed of the error object at the end of the log);
4. [messages](https://travis-ci.org/firebase/firebase-js-sdk/jobs/452183556#L6512) printed by lerna (in the "stdout" field of the error object at the end of the log).

packages/database/package.json
yifanyang: Flags on the command line will override the corresponding ones in mocha.opts, based on the mocha doc [here](https://mochajs.org/#mochaopts). And what I'm trying to propose is to put all default configurations in the config file and only override them with command line flags when necessary.

I have no idea of the 5 retries.

config/ci.mocha.node.opts
yifanyang: This is the mocha config for node testing.

I think it makes sense to separate out the change for mocha config file. I though it would be simple and straightforward at first, but it does feel confusing in this PR now.

packages/database/package.json
Feiyang1: Since we created a config file, we should move all the options into the file. It will be easier for people to read.

mikelehen: Agreed.  I don't think we actually want to override the defaults in these cases.  The defaults should work for both CI and local development AFAIK.

mikelehen: (to be clear, by "defaults" I mean what we put in the config file)

yifanyang:
This pr essentially tries to introduce below two changes. It's more clear to track them separately as these two changes are less related to each other. Closing this pr. 
- Remove duplicated log generated by lerna. #1468
- Consolidate mocha options into config file. #1469

PR #1374, opened 2018-11-09
Implement sequence number migration

gsoltis:
Port of https://github.com/firebase/firebase-android-sdk/pull/64

Ensures that every document in the cache has a corresponding sentinel row with a sequence number. Precursor to enabling LRU in production.

### DIFF
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -23,6 +23,7 @@ import { SnapshotVersion } from '../core/snapshot_version';
 import { BATCHID_UNKNOWN } from '../model/mutation_batch';
 import { encode, EncodedResourcePath } from './encoded_resource_path';
 import { removeMutationBatch } from './indexeddb_mutation_queue';
+import { getHighestListenSequenceNumber } from './indexeddb_query_cache';
 import { dbDocumentSize } from './indexeddb_remote_document_cache';
 import { LocalSerializer } from './local_serializer';
 import { PersistencePromise } from './persistence_promise';
@@ -40,7 +41,7 @@ import { SimpleDbSchemaConverter, SimpleDbTransaction } from './simple_db';
  * 4. Multi-Tab Support.
  * 5. Removal of held write acks.
  */
-export const SCHEMA_VERSION = 6;
+export const SCHEMA_VERSION = 7;
 
 /** Performs database creation and schema upgrades. */
 export class SchemaConverter implements SimpleDbSchemaConverter {
@@ -115,6 +116,12 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
       });
     }
 
+    if (fromVersion < 7 && toVersion >= 7) {
+      p = p.next(() => {
+        return this.ensureSequenceNumbers(txn);
+      });
+    }
+
     return p;
   }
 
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -40,7 +41,7 @@ import { SimpleDbSchemaConverter, SimpleDbTransaction } from './simple_db';
  * 4. Multi-Tab Support.
  * 5. Removal of held write acks.
  */
-export const SCHEMA_VERSION = 6;
+export const SCHEMA_VERSION = 7;
```
schmidt-sebastian: Do you mind backfilling the above comment with 6 and 7?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -115,6 +116,12 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
       });
     }
 
+    if (fromVersion < 7 && toVersion >= 7) {
+      p = p.next(() => {
+        return this.ensureSequenceNumbers(txn);
```
schmidt-sebastian: Nit: This can be one line like the statement in 109.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -172,6 +179,49 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+    const documentTargetStore = txn.store<
+      DbTargetDocumentKey,
+      DbTargetDocument
+    >(DbTargetDocument.store);
+    const documentsStore = txn.store<DbRemoteDocumentKey, DbRemoteDocument>(
+      DbRemoteDocument.store
+    );
+
+    return getHighestListenSequenceNumber(txn).next(currentSequenceNumber => {
+      const writeSentinelKey = (
```
schmidt-sebastian: I think this would be easier to read if a) this was declared one level higher (as a direct child of `ensureSequenceNumbers`) and b) if you used a normal JS function (`function writeSentinelKey() {}`).

Optional.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -172,6 +179,49 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+          new DbTargetDocument(0, encode(path), currentSequenceNumber)
+        );
+      };
+
+      const promises: Array<PersistencePromise<void>> = [];
+      return documentsStore
+        .iterate((key, doc) => {
+          const path = new ResourcePath(key);
+          const docSentinelKey = sentinelKey(path);
+          documentTargetStore.get(docSentinelKey).next(maybeSentinel => {
```
schmidt-sebastian: You need to push this promise (the one returned from `documentTargetStore.get`) rather then the result of `writeSentinelKey`. With the current code, you may miss some writes if the reads don't come back fast enough.

gsoltis: Good catch. I restructured it a little to remove the inner call to `push()` and instead push every read. Reads which require an additional write return a persistence promise in the `next()` call.

packages/firestore/src/local/indexeddb_schema.ts
gsoltis: Leaving as-is.

It can't go in a higher scope as it closes over the result of `getHighestListenSequenceNumber()`. And typescript disallows nested `function` declarations, so it needs to be a lambda. I'm guessing this is because `function` and lambdas have different bindings for `this`. Nested `function`s make it easy to make a mistake when calling methods off of `this`.

PR #1375, opened 2018-11-09
Upgrade mocha version to 5.2.0 in all packages.

yifanyang:
We are trying to consolidate flags passed to mocha test into [mocha.opts](https://mochajs.org/#mochaopts). In order to support comments in this file, mocha needs to be upgraded to its latest version (5.2.0 as of 11/09/2018).

See discussion in #1373 for more context.


yifanyang:
@mikelehen Thanks for the tips!

PR #94, opened 2017-07-08
Generate smaller bundles

jsayol:
### ~~Work in progress, do not merge yet.~~ Edit: Ready for review

I'm tweaking the build process to generate smaller bundles.

I still want to try a few more things but I'm already seeing some serious improvement. `firebase-database.js` is still not back to pre-TypeScript size but every other bundle is smaller:

| File | Before | After | % diff |
|----------|-------------|------|------|
| firebase-app.js | 18.87 KB | 16.19 KB | **-14.20%** |
| firebase-database.js | 240.49 KB | 181.47 KB | **-24.54%** |
| firebase-messaging.js | 24.93 KB | 19.94 KB | **-20.10%** |
| firebase-storage.js | 53.91 KB | 36.17 KB | **-32.91%** |
| firebase.js | 475.87 KB | 391.45 KB | **-17.74%** |

PR #653, opened 2018-04-09
Build Refactoring

jshcrowthe:
# Issues
There are several issues with the build process that I'd like to address:

- The current build process ignores Typescript compile errors, allowing invalid TS (that fails at release time) to get into the master of the SDK. This is WAI from a typescript viewpoint, so I'm addressing this in some refactoring I've been looking at for a while.
- The current build process is quite slow
- The current build process doesn't lend itself to highly optimized use cases

# Change Summary

The changes are broken down as follows:

## @firebase/* Packages (not including the types packages)

- Removed gulpfile.js and associated gulp dependencies
- Added rollup.config.js and associated rollup dependencies (including rollup-plugin-typescript2 which fails builds for TS errors)
- Updated the `main`, `module`, `browser` fields of each `package.json` to conform to rollup standards
- Updated the "scripts" for each package that were referencing `gulp` to now reference `rollup` (as pertinent)

## @firebase/*-types Packages

- Removed unused test files and associated test setup

# Results

## Build Time

**Before PR:** `130.11s`
**After PR:** `78.30s`

## Build Size

**Before PR**

| Asset | Size |
| ------------- | ------------- |
| firebase-app.js | 35.1 kB | 
| firebase-auth.js | 145 kB | 
| firebase-database.js | 178 kB | 
| firebase-firestore.js | 280 kB | 
| firebase-functions.js | 5.85 kB | 
| firebase-messaging.js | 26.5 kB | 
| firebase-storage.js | 33.7 kB | 

**After PR**

| Asset | Size |
| ------------- | ------------- |
| firebase-app.js | 13 kB | 
| firebase-auth.js | 141 kB | 
| firebase-database.js | 178 kB | 
| firebase-firestore.js | 302 kB | 
| firebase-functions.js | 6.8 kB | 
| firebase-messaging.js | 33 kB | 
| firebase-storage.js | 31 kB |

PR #238, opened 2017-10-18
Fixing named prop "_lat"

jshcrowthe:
There was a minification error causing some odd issues with auth state persistence.

Fixes #230

Issue #230, opened 2017-10-17
Database permission_denied after linking a provider (linkWithPopup)


khammami:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10 Chrome Version 63.0.3236.0 (Official Build) dev (64-bit)
  * Firebase SDK version: 4.5.1
  * Firebase Product: database, auth

### [REQUIRED] Describe your environment
database write permission denied after linking other providers with a popup. updates in the same page before works as intended but after linking an other provider I get a warning

it used to work fine before updating js sdk

```javascript
util.js:189 FIREBASE WARNING: update at /users/xxxxxxxx/data failed: permission_denied 
t.warn @ util.js:189
(anonymous) @ Repo.js:309
(anonymous) @ PersistentConnection.js:411
t.Tt @ PersistentConnection.js:444
e.Tt @ Connection.js:262
e.an @ Connection.js:256
(anonymous) @ Connection.js:157
t.Fn @ WebSocketConnection.js:197
t.handleIncomingFrame @ WebSocketConnection.js:247
mySock.onmessage @ WebSocketConnection.js:144
15:33:47.264 Repo.js:510 Uncaught (in promise) Error: PERMISSION_DENIED: Permission denied
    at Repo.js:510
    at Object.t.exceptionGuard (util.js:556)
    at e.callOnCompleteCallback (Repo.js:501)
    at Repo.js:314
    at PersistentConnection.js:411
    at t.Tt (PersistentConnection.js:444)
    at e.Tt (Connection.js:262)
    at e.an (Connection.js:256)
    at t.onMessage (Connection.js:157)
    at t.Fn (WebSocketConnection.js:197)
```

#### Steps to reproduce:

database rule ".write": "$uid === auth.uid"
main provider is Google to signin/ create an account in my app
try to write data to the current user ref after linking another provider with popup 

=> https://local-guides-directory.firebaseapp.com/account

stivaliserna:
I'm experiencing the same issue, after fresh install of my dependency tree which pulls latest version of firebase `4.5.2`.

stivaliserna:
@khammami can you try with firebase version `4.5.0`?

I forced `4.5.0` on my setup and the issue disappeared. Somehow it seems related to #223. I came across the idea to downgrade to `4.5.0` after ready this comment: https://github.com/firebase/firebase-js-sdk/issues/223#issuecomment-336610508

khammami:
[...]

PR #1385, opened 2018-11-16
Generate smaller bundles (take 2)

jsayol:
This reapplies the *uglify* optimizations from #94, which were lost after the build process was switched to rollup a while ago.

Here's the size diff compared with the current master:

| File | Before | After | % diff |
|----------|-------------|------|------|
| firebase-app.js | 34.03 KiB | 32.63 KiB | **-4.12%** |
| firebase-database.js | 178.22 KiB | 158.28 KiB | **-11.19%** |
| firebase-firestore.js | 354.32 KiB | 350.68 KiB | **-1.03%** |
| firebase-functions.js | 7.30 KiB | 7.26 KiB | **-0.51%** |
| firebase-messaging.js | 35.01 KiB | 34.11 KiB | **-2.55%** |
| firebase-storage.js | 35.57 KiB | 32.05 KiB | **-9.90%** |
| firebase.js | 779.88 KiB | 751.72 KiB | **-3.61%** |

This change doesn't affect `firebase-auth.js`, since applying these uglify options actually generated a slightly larger bundle for auth. It does apply to the global `firebase.js`, though, which includes auth.

All tests pass.

### DIFF
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -49,6 +49,20 @@ const plugins = [
 
 const external = Object.keys(pkg.dependencies || {});
 
+const uglifyOptions = {
+  mangle: {
+    properties: {
+      keep_quoted: true,
+      regex: /_$|^_/
+    }
+  },
+  compress: {
+    passes: 3,
+    unsafe: true,
+    warnings: false,
+  }
+};
+
 /**
  * Global UMD Build
  */
@@ -81,7 +95,7 @@ const appBuilds = [
       format: 'umd',
       name: GLOBAL_NAME
     },
-    plugins: [...plugins, uglify()]
+    plugins: [...plugins, uglify(uglifyOptions)]
   }
 ];
 
@@ -138,7 +152,7 @@ const componentBuilds = components
               );
             }`
         },
-        plugins: [...plugins, uglify()],
+        plugins: [...plugins, uglify(component !== 'auth' ? uglifyOptions : {})],
         external: ['@firebase/app']
       }
     ];
@@ -168,7 +182,7 @@ const completeBuilds = [
       format: 'umd',
       name: GLOBAL_NAME
     },
-    plugins: [...plugins, uglify()]
+    plugins: [...plugins, uglify(uglifyOptions)]
   },
   /**
    * App Node.js Builds
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -53,7 +53,7 @@ const uglifyOptions = {
   mangle: {
     properties: {
       keep_quoted: true,
-      regex: /_$|^_/
+      regex: /(^_[^_])|([^_]_$)/
     }
   },
   compress: {
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -57,9 +57,7 @@ const uglifyOptions = {
     }
   },
   compress: {
-    passes: 3,
-    unsafe: true,
-    warnings: false,
+    passes: 3
   }
 };
 
@@ -152,7 +150,10 @@ const componentBuilds = components
               );
             }`
         },
-        plugins: [...plugins, uglify(component !== 'auth' ? uglifyOptions : {})],
+        plugins: [
+          ...plugins,
+          uglify(component !== 'auth' ? uglifyOptions : {})
+        ],
         external: ['@firebase/app']
       }
     ];


jsayol:
It seems like something went wrong on the Travis build, I'll look into it later.

packages/firebase/rollup.config.js
```diff
@@ -49,6 +49,20 @@ const plugins = [
...
+const uglifyOptions = {
+  mangle: {
+    properties: {
+      keep_quoted: true,
+      regex: /_$|^_/
+    }
+  },
+  compress: {
+    passes: 3,
+    unsafe: true,
```
Feiyang1: Does it mean it could break things in some case?

jsayol: The `compress.unsafe` option applies these 4 transformations:
- `new Array(1, 2, 3)` or `Array(1, 2, 3)` → `[ 1, 2, 3 ]`
- `new Object()` → `{}`
- `String(exp)` or `exp.toString()` → `"" + exp`
- `new Object/RegExp/Function/Error/Array (...)` → the `new` is discarded

These should actually be safe in *most* cases. That being said, I just did a build without that option and it seems like it only provides a very marginal reduction in size, about an extra 0.2% on average.

I don't think it's worth it, so I'll remove it.

Feiyang1: It looks great, thank you!

mikelehen:
I think they are not. :-(  Unit tests, etc. of course won't work against the minified build.  A while ago, I split out the Firestore tests that could be run against a minified build.  firebase-js-sdk/integration/firestore used to do that, but it was removed from CI and then I think got broken by some build refactorings at some point, so it probably doesn't work anymore.

We should probably re-invest in some tests against the minified builds.  But until then, changes like this come with some risk we'll need to mitigate through manual testing or similar.

jsayol:
Ok, database works fine but there's definitely something wrong with the Firestore build. Getting lots of INTERNAL ASSERTION FAILED, I'm looking into it.

jsayol:
Fixed the issue, sorry about this.

I've run the samples from https://github.com/firebase/quickstart-js against my local build and everything is working fine now. I've tested Auth (anonymous and google sign-in), RTDB, Firestore, FCM, and Storage.

Feiyang1:
@jsayol Why does mangling double underscore broke the component? Now I'm a little concerned about the property mangling because what if some other patterns need to excluded, but we don't know yet. I wonder if it is why it was removed when refactoring. Unless we can run all the test cases on the minified bundle, I'm not very comfortable making this change.
@mikelehen what do you think? what is our convention of using _ vs __ and prefix _ vs postfix _.

jsayol:
> Unless we can run all the test cases on the minified bundle, I'm not very comfortable making this change.

Yeah, I agree. This isn't necessary anyway so it might be better to wait until there's at least some tests on the minified files.

> Why does mangling double underscore broke the component?

This one's a bit tricky, and I'm not really happy with the fix. For some reason uglify is ignoring the `keep_quoted: true` option (maybe it doesn't play well with other conditions like `regex`). The `goog.labs.net.webchannel` package from the google-closure-library uses `channelMsg['__data__']` to get the message data. Uglify was mangling `__data__` to something else, causing the message data to get lost, and that made Firestore fail.

That's definitely not an ideal situation, since similar errors might happen again in the future with any other quoted properties that begin or end with an underscore. I'll keep looking into it and see if I can figure out what's going on with `keep_quoted`.

mikelehen:
Here's where it was removed: https://github.com/firebase/firebase-js-sdk/pull/653/files#diff-f8545829bd5e24a8a540f499f5878fa7L57  I'm guessing it was just an accident.  That PR was massive.

It looks like the old regex was `/^_[^_][^lat]|[^_]_$/`.  So without `[^lat], this PR will likely break auth (see #238 and #230).  (sidenote: I think `[^lat]` probably should have been `(?!lat)` instead.)

So yeah, I agree this isn't a very safe change.  If we put it back to exactly what it was before, it may be fine, but there's a chance we have new weirdly-named properties we don't know about and won't find without exhaustive testing.

AFAIK we also don't use `_foo` or `foo_` conventions very much in our code, so this regex won't help reduce code size all that much (and incidentally, to measure meaningfully, we should look at the size after gzip).

That said, there is one nice benefit of the mangling we used to do.  It helps us hide non-public properties on our APIs.  So Firestore uses the `_foo` convention in all of our public API classes (e.g. https://github.com/firebase/firebase-js-sdk/blob/97cc3f89283d942a99c5c3e56e29eee331a0f5df/packages/firestore/src/api/database.ts#L240), but we don't bother in the rest of our code.  Without mangling these properties, people may notice them while debugging and then take a dependency on them in their app.  If they do, then we may some day break their app when making seemingly-innocuous changes to our implementation details. With name mangling, the names would end up unrecognizable and automatically change from release to release, so devs couldn't depend on them.

@jsayol Out of curiosity, did you have a motivation for this change beyond the size reduction? Also did you by chance look at the post-gzip size difference to see if that erodes the savings much?

Given the state of things (moderate risk of regression, low-ish benefit since we don't use the conventions much), this change may not be worth it.

jsayol:
> Out of curiosity, did you have a motivation for this change beyond the size reduction?

Not really. I initially did this last year because the database bundle had exploded in size after porting it to TypeScript and moving away from the closure compiler. A few days ago I was debugging something elsewhere on the SDK and noticed a `_foo`-style property that shouldn't have been there, so I realized the uglify changes had been lost and I figured I might as well reapply them.

> So without [^lat], this PR will likely break auth (see #238 and #230).

Yup, definitely will. And I think this case wouldn't be covered even if `keep_quoted` was being applied correctly, so it would still need an exception.

So yeah, I agree that this is not a safe approach overall. I'm going to close this PR unless anyone says otherwise. Thanks for the patience and for looking into all this :)

P.S.: Just a side note, have you given any thought to using closure compiler again but merely as a minifier replacing uglify? [This rollup plugin](https://github.com/ampproject/rollup-plugin-closure-compiler) looks interesting. I actually played with it earlier as a possible alternative here, but I have zero experience configuring closure compiler so I didn't get very far. Might be worth it, though, unless reducing the bundle size is not really a priority.

Feiyang1:
One of our top priorities is to modularize this SDK, so you can pull in the code you really use instead of the entire package, e.g. just signInWithEmailPassword instead of the entire auth SDK. It would require developers to use the SDK as modules and utilizing bundler's tree shaking capability. 

Clearly it won't help with the size of the minified bundles since they always include the entire component. We see the minified bundles as a convenient way to do experiments and quick prototypes, but discourage its usage in production in favor of ES6 modules. If you chose to use the minified bundle, you are already missing out the opportunity for size reduction, so I don't think we will focus too much on the minified bundle size.

jsayol:
> One of our top priorities is to modularize this SDK, so you can pull in the code you really use instead of the entire package

That makes perfect sense. From what you say I assume that's something the team is already working on internally? If not, I'd be happy to look into it.

I was thinking about this a couple days ago and it would require a slightly different API. The best approach would be if the public API methods were pure functions whenever possible, in particular doing away with any shared state between them.

So as an example, using database and auth could look like this:

```ts
import {
  auth,
  signInWithEmailAndPassword,
  onAuthStateChanged
} from 'firebase/auth';
import {
  database,
  reference,
  child,
  query,
  orderByValue,
  onValue
} from 'firebase/database';
import { config } from './config';

const app = initializeApp(config);
const appAuth = auth(app);
const appDatabase = database(appAuth);

const email = 'alice@example.com';
const password = 'bob123';
signInWithEmailAndPassword(appAuth, email, password);

onAuthStateChanged(appAuth).then(user => {
  if (user) {
    let dbRef = reference(appDatabase, 'userPosts');
    dbRef = child(dbRef, user.uid);

    let dbQuery = query(dbRef);
    dbQuery = orderByValue(dbQuery);

    onValue(dbQuery, snap => {
      console.log(snap.val());
    });
  }
});

```

This would allow a bundler to only include the methods that have been imported, along with what those methods use themselves.

The main issue I see, at least with the database package, is that all the internals are very tightly coupled so even if something like this is done, most of the package would still need to be pulled in.

Then there's also the fact that the whole auth package would need to be refactored as TypeScript, or as ES modules at the very least. But a tool like [gents](https://github.com/angular/clutz#gents---closure-to-typescript-converter) could probably help with that.

jsayol:
Just dumping some more thoughts here.

In certain cases it could make sense to offer a method to cascade a series of operations on an object, since repeatedly passing that object to those functions might not be very readable. I'm thinking of something akin to rxjs' `.pipe()`.

So for example, performing a query in Firestore could look like this:
```ts
const app = initializeApp(config);
const appFirestore = firestore(app);

const query = collection(appFirestore, 'myCollection').pipe(
  where('customer', '==', customerId),
  where('purchased', '<', someDate),
  orderBy('orderId'),
  limitTo(100)
);

getSnapshot(query).then( /* ... */);
```

P.S.: "pipe" is actually a bad name for it, since that could be confusing for those using firebase with rxjs. Just used it to illustrate my point.

Feiyang1:
These are all great points! Yes, we are actively looking at it internally. You are right to the point about making methods pure functions. At the same time we want to strike a balance between modularization (size reduction potential) and ease of use for developers. 

And yes, some components will go through major rewrite. Please stay tuned! :)

Issue #1400, opened 2018-11-26
Possible malicious code injected via npm dependency


jsayol:
### Describe your environment

  * Operating System version: Linux 4.19.2
  * Browser version: n/a
  * Firebase SDK version: 5.5.9
  * Firebase Product: core

### Describe the problem
I don't think you need to take any immediate action, but I figured I should give you a heads up just in case.

A few days ago a malicious package was added as a dependency to `event-stream`, which is a dependency to `ps-tree`, which in turn is a dependency to `npm-run-all`. This last one is used extensively as part of the build tools on this SDK.

To make sure this does not become a problem, it might be a good idea to update `npm-run-all` to the recently-published 4.1.5 version (it's currently pinned at 4.1.2 in all the `packages/**/package.json` files.)

Relevant discussions:
https://github.com/mysticatea/npm-run-all/issues/149
https://github.com/dominictarr/event-stream/issues/116

**Edit**: The current versions used here are at the moment are not affected, but upgrading is still a good idea to make sure they never are. And just to be clear, this would only affect the machines of those developing the SDK and not the end users of the library.

#### Steps to reproduce:
n/a

#### Relevant Code:
n/a

Feiyang1:
Thanks for the heads up! I will look into it.

Feiyang1:
Our repo is not affected by the exploit, thanks to the lock file. As a precaution, I'm updating the npm-run-all to 4.1.5. https://github.com/firebase/firebase-js-sdk/pull/1406

PR #1406, opened 2018-11-29
update npm-run-all to avoid potential exploit

Feiyang1:
A npm package we transitively depend on was hacked and was injected with malicious code. Our repo is not affected thanks to the fixed version and the lock file.  As a precaution, I'm upgrading the library which removed the dependency on the hacked library.
More detail: https://github.com/firebase/firebase-js-sdk/issues/1400

### DIFF
--- a/integration/typescript/package.json
+++ b/integration/typescript/package.json
@@ -24,7 +24,7 @@
     "karma-spec-reporter": "0.0.32",
     "karma-typescript": "3.0.12",
     "mocha": "5.2.0",
-    "npm-run-all": "4.1.2",
+    "npm-run-all": "4.1.5",
     "typescript": "2.8.1"
   }
 }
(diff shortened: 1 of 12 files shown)


packages/testing/src/api/index.ts
```diff
@@ -200,7 +200,13 @@ export function loadFirestoreRules(
...
   let client = new EMULATOR.FirestoreEmulator(
     FIRESTORE_ADDRESS,
-    grpc.credentials.createInsecure()
+    grpc.credentials.createInsecure(),
+    {
+      // Cap how much backoff gRPC will perform. This is testing code, so
+      // efficiency is less important than responsiveness.
+      'grpc.initial_reconnect_backoff_ms': 100,
+      'grpc.max_reconnect_backoff_ms': 100
+    }
```
mikelehen: Where is this change coming from?

Feiyang1: I think I know what's going on. I rebased the master yesterday after the release and the PRs are submitted before the rebase. Rebasing the PR should solve the problem.

Feiyang1: It is removed from the PR.

Feiyang1:
@davideast ping

Issue #32, opened 2017-06-02
FR: Ability to contribute to the Auth module


barakd:
It seems that currently, it is not possible to contribute to the Auth module.
Unless I missed something in my findings, as can be seen at [auth.build.js](https://github.com/firebase/firebase-js-sdk/blob/master/src/auth.build.js).
The whole of Auth module is already checked in in a post build file.
Thus creating PRs to improve this module is impossible at the moment.
The Auth module is one of the greatest features of Firebase and It a shame that the community can't contribute to improve this important piece of Firebase.
My request is simply to allow a way to contribute to the original pre-built file.
Would very appreciate your feedback on this matter, 
Thanks!

jshcrowthe:
Hey @barakd thanks for your interest in contributing to the SDK! We are working on open sourcing the Auth module, so stay tuned to future updates.

cc: @bojeil-google

leonardodino:
any updates on this @bojeil-google? the source code would alleviate some pressure on the support, where documentation is kinda scarce.

---

my use-case is implementing a  `react-native` `js only` solution to the following methods:
`auth.signInWithPhoneNumber/user.reauthenticateWithPhoneNumber/user.linkWithPhoneNumber`
using a WebView, to handle captcha, but it's not clear how to pass an anonymous logged in user from the `react-native` js side to the WebView.

(with the source code, i would navigate my way through it to find a solution instead of cluttering the support channels with more questions)

bojeil-google:
Hey @leonardodino, it's in the works.
By the way, do not use webviews. They are insecure. Google sign-in for example no longer works in an embedded webview. You need to use an SFSafariViewController for iOS or Chrome custom tab for android.

PR #1414, opened 2018-12-05
Enable database test with local emulator.

yifanyang:


scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+      'java',
+      ['-jar', EMULATOR_FILENAME, '--port', EMULATOR_PORT],
+      { cwd: EMULATOR_DIRECTORY }
+    );
+    promise.catch(reject);
+    const process = promise.childProcess;
+    console.log(`Bringing up emulator, pid: [${process.pid}] ...`);
+    process.stdout.on('data', data => {
+      console.log(`[emulator-logs]: ${data}`);
+      // Working on better ways to signal readiness.
```
ryanpbrewster: I'd suggest using a Node library (like `request-promise-native` or the like) to make REST requests in a loop and proceed once the request succeeds

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+      [
+        '-H',
+        'Authorization: Bearer owner',
+        '-X',
+        'PUT',
+        '-d',
+        '{"rules": {".read": true, ".write": true}}',
+        `http://localhost:${EMULATOR_PORT}/.settings/rules.json?ns=${EMULATOR_NAMESPACE}`
+      ],
+      { stdio: 'inherit' }
```
ryanpbrewster: same here, I think this will be slightly easier if you use a library to make these requests

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+  return Promise.all([
+    spawn('yarn', ['test:node'], options),
+    spawn('yarn', ['test:browser'], options)
+  ]);
+}
+
+(async () => {
+  let emulator;
+  try {
+    await downloadEmulator();
```
ryanpbrewster: I'm not familiar with your CI setup, but is it necessary to re-download the emulator for every run? Is there some way to configure the test environment in a more persistent way?

packages/database/test/helpers/util.ts
```diff
@@ -21,10 +21,23 @@ import '../../index';
 import { Reference } from '../../src/api/Reference';
 import { Query } from '../../src/api/Query';
 import { ConnectionTarget } from '../../src/api/test_access';
-import { RepoInfo } from '../../src/core/RepoInfo';
 
 export const TEST_PROJECT = require('../../../../config/project.json');
 
+const USE_EMULATOR = process.env.USE_RTDB_EMULATOR;
```
Feiyang1: You probably don't need it and can just check the port number.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+const { spawn } = require('child-process-promise');
+
+const EMULATOR_LINK =
+  'https://firebase.google.com/docs/database/downloads/database-emulator.jar';
+const EMULATOR_DIRECTORY = path.join(os.tmpdir(), 'firebase-emulators');
```
Feiyang1: You can use 'tmp' package for managing temporary directories. It generates a random name and creates the folder for you. 
https://www.npmjs.com/package/tmp

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+  return new Promise(resolve => setTimeout(resolve, 1000 * seconds));
+}
+
+async function downloadEmulator() {
+  return new Promise((resolve, reject) => {
+    console.log(`Creating directory [${EMULATOR_DIRECTORY}] ...`);
+    fs.mkdir(EMULATOR_DIRECTORY, err => {
+      if (err && err.code !== 'EEXIST') reject(err);
+      console.log(`Downloading emulator binary from [${EMULATOR_LINK}] ...`);
+      // const promise = spawn('firebase', ['setup:emulators:database']);
```
Feiyang1: remove comments

.travis.yml
```diff
@@ -34,19 +35,23 @@ jobs:
   allow_failures:
     - script: yarn test:saucelabs
     - script: yarn test:saucelabs --database-firestore-only
+    - script: node scripts/testing/database-emulator-test.js
```
mikelehen: FWIW it could be helpful to document these with comments to explain the current state and future plans.  e.g.:

```
# TODO(yifanyang): Once we verify the emulator tests are reliable, we should make these tests blocking rather than allow failures.
```

Ideally we'd comment the saucelabs ones too, but that's orthogonal to this PR, so feel free to leave as-is for now.

Also, you added names for the ones below.  Should these have names too?

package.json
```diff
@@ -45,7 +45,7 @@
     "clone": "2.1.1",
     "coveralls": "3.0.0",
     "dependency-graph": "0.7.0",
-    "firebase-tools": "5.1.1",
+    "firebase-tools": "6.1.1",
```
mikelehen: There should generally be an accompanying yarn.lock change any time you update a dependency in a package.json file.  Can you run "yarn" from the root and see if yarn.lock gets updated (and if so, add it to this PR)?

packages/database/test/helpers/util.ts
```diff
@@ -21,10 +21,23 @@ import '../../index';
...
+const EMULATOR_PORT = process.env.RTDB_EMULATOR_PORT;
+const EMULATOR_NAMESPACE = process.env.RTDB_EMULATOR_NAMESPACE;
+
+export const DATABASE_ADDRESS = USE_EMULATOR
+  ? `http://localhost:${EMULATOR_PORT}`
+  : TEST_PROJECT.databaseURL;
+
+export const DATABASE_URL = USE_EMULATOR
+  ? `http://localhost:${EMULATOR_PORT}?ns=${EMULATOR_NAMESPACE}`
+  : TEST_PROJECT.databaseURL;
```
mikelehen: I stared at these a bit to figure out what the difference between DATABASE_ADDRESS and DATABASE_URL is.  It would be slightly easier if we defined them in relation to each other (and added a comment).  E.g.:

```
// When running against the emulator, the hostname will be "localhost" rather than 
// something like  "namespace.firebaseio.com" and so we need to append the namespace 
// as a query param.
export const DATABASE_URL = USE_EMULATOR ? DATABASE_ADDRESS + '?ns=${EMULATOR_NAMESPACE}` : DATABASE_ADDRESS;
```

packages/database/test/info.test.ts
```diff
@@ -132,6 +132,8 @@ describe('.info Tests', function() {
   });
 
   it.skip('database.goOffline() / database.goOnline() connection management', function() {
+    // NOTE: getFreshRepo() no longer takes a hostname, so this test needs to be reworked.
+    // Need to figure out how to re-enable this test. Also see comment #27-#32.
```
mikelehen: What does `comment #27-#32` refer to?

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
+/**
```
mikelehen: Just to double-check, is there a reason this is written in JavaScript rather than TypeScript?  I don't care too much, but I think generally we'd prefer the type-safety and extended featureset of TypeScript for any new code we're writing...

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+ * limitations under the License.
+ */
+
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
+const { spawn } = require('child-process-promise');
+
+const EMULATOR_LINK =
+  'https://firebase.google.com/docs/database/downloads/database-emulator.jar';
```
mikelehen: cc/ @ryanpbrewster 

I'm surprised this link isn't versioned in any way (i.e. I'd expect database-emulator-0.1.jar or something).  As-is, if the emulator ever makes a breaking change, we'll be broken.  Has this been discussed at all?

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+        stdio: 'inherit'
+      });
+      resolve(promise);
+    });
+  });
+}
+
+async function launchEmulator(locked) {
+  const emulator = new Promise((resolve, reject) => {
+    // const promise = spawn('firebase', ['serve', '--only', 'database']);
```
mikelehen: more commented out code.

scripts/testing/database-emulator-test.js
mikelehen: And then hopefully you could get rid of the 5s timeout below.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+    console.log(`Bringing up emulator, pid: [${process.pid}] ...`);
+    process.stdout.on('data', data => {
+      console.log(`[emulator-logs]: ${data}`);
+      // Working on better ways to signal readiness.
+      if (data.includes(`Listening on port ${EMULATOR_PORT}`)) {
+        setTimeout(() => resolve(process), 5000);
+      }
+    });
+  });
+  await emulator;
```
mikelehen: I'd split this function into two.  One to start the emulator and one to initialize the rules.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+      const promise = spawn('curl', [EMULATOR_LINK, '-o', EMULATOR_FILENAME], {
+        cwd: EMULATOR_DIRECTORY,
+        stdio: 'inherit'
+      });
+      resolve(promise);
+    });
+  });
+}
+
+async function launchEmulator(locked) {
```
mikelehen: It's not obvious to me what `locked` indicates.  Please rename and/or add comments.

scripts/testing/database-emulator-test.js
mikelehen: +1.  Also more platform-independent (curl won't exist on windows, etc.).

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+    env: Object.assign({}, process.env, {
+      USE_RTDB_EMULATOR: true,
+      RTDB_EMULATOR_PORT: EMULATOR_PORT,
+      RTDB_EMULATOR_NAMESPACE: EMULATOR_NAMESPACE
+    }),
+    stdio: 'inherit'
+  };
+  return Promise.all([
+    spawn('yarn', ['test:node'], options),
+    spawn('yarn', ['test:browser'], options)
```
mikelehen: To be clear, this means we're running these in parallel, right?  Is that intentional?  I think it should probably be okay, but it may make debugging more confusing.

Also, does this rely on `yarn` being in the path?  That may be fine, but I don't know if it's 100% guaranteed.  If there was a way to invoke yarn programmatically, that might be cleaner and more reliable.  But if this is mostly for CI, which is a controlled environment, it's probably not a huge deal. We can always rework it later.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,121 @@
...
+  ]);
+}
+
+(async () => {
```
mikelehen: Hrm...  This looks a bit strange.

```
(async () => ...)();
```

That's going to define a lambda that returns a Promise, and then call the lambda, generating a Promise. But then it doesn't do anything with the returned Promise.  I guess it _does_ let you use await within the lambda code in order to sequence the actions.  But it's still a bit strange. I'd maybe extract the lambda into an actual named async function called `start()` or something. And then either handle the promise, or add a comment saying we don't need to.

scripts/testing/database-emulator-test.js
yifanyang: Done. Using 'request' node module to ping the emulator for readiness now.

scripts/testing/database-emulator-test.js
yifanyang: Done. Using 'request' node module for loading rules now.

scripts/testing/database-emulator-test.js
yifanyang: Travis does offer a cache mechanism. But in [Travis' words](https://docs.travis-ci.com/user/caching/#things-not-to-cache):

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source.

Caching doesn't seem to provide much benefit?

packages/database/test/helpers/util.ts
yifanyang: I hesitated as well. I guess code is more readable this way?

scripts/testing/database-emulator-test.js
yifanyang: Done. Using 'tmp' node module for downloading location now. Eventually we will leverage firebase cli to manage the download and startup of emulators once we have addressed discussion in this [doc](https://docs.google.com/document/d/1k_nmMPL9LRFzxk9yJceNYgHcxgiMeIaAc07YfKUgtqk/edit). But this should be good enough for now.

scripts/testing/database-emulator-test.js
yifanyang: Done.

.travis.yml
yifanyang: Done for adding comments.

Names for the below ones will actually display on Travis run page. But names here will have no use. And they can actually be different from the ones below, and everything will still work. Travis is only using the 'scirpt' content here to identify different jobs. So I feel like there's not much value for adding names here.

package.json
yifanyang: Oops. Reverted, since I'm not relying on firebase cli to download and start emulators. No need to update the version here.

Eventually we will leverage firebase cli to manage the download and startup of emulators once we have addressed discussion in this [doc](https://docs.google.com/document/d/1k_nmMPL9LRFzxk9yJceNYgHcxgiMeIaAc07YfKUgtqk/edit).

packages/database/test/helpers/util.ts
yifanyang: Done.

packages/database/test/info.test.ts
yifanyang: https://github.com/firebase/firebase-js-sdk/blob/d93e15307e400443c23467c6646e5cd104380834/packages/database/test/info.test.ts#L27-L36

Done. I guess there's not much value to put such reference in comments.

scripts/testing/database-emulator-test.js
yifanyang: Agreed.

Looks like all other file in the 'scripts' are written in JavaScript, so I figured I'd do the same. I can help convert them to TypeScript if I get a chance.

scripts/testing/database-emulator-test.js
yifanyang: It's using a link that specifies the version now.

It's been discussed in this [doc](https://docs.google.com/document/d/1k_nmMPL9LRFzxk9yJceNYgHcxgiMeIaAc07YfKUgtqk/edit) a bit. The emulator version is tied to firebase cli version, if we do download and startup through cli. After we have addresses thoughts in the doc, we will move to firebase cli eventually.

scripts/testing/database-emulator-test.js
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: It does means they will run in parallel. I've changed it to be:
`spawn('yarn', ['test'], options)` to be in accordance with how we start existing tests.

Note: 'yarn test' is calling 'run-p yarn test:node yarn test:browser' under the hood, which will run the tests in parallel as well.

scripts/testing/database-emulator-test.js
yifanyang: Done? The problem is that I need the reference to the emulator in the finally block. That's making the code not so streamlined.

scripts/testing/database-emulator-test.js
ryanpbrewster: Fair enough, and GCS can definitely handle the load.

packages/database/test/helpers/util.ts
```diff
@@ -243,3 +267,16 @@ export function testRepoInfo(url) {
   const [, ns] = match;
   return new ConnectionTarget(`${ns}.firebaseio.com`, true, ns, false);
 }
+
+export function repoInfoForConnectionTest() {
+  if (USE_EMULATOR) {
+    return new ConnectionTarget(
+      `localhost:${EMULATOR_PORT}`,
+      false, // emulator does not support https or wss
```
ryanpbrewster: can you document what this parameter is named (probably something like `/* useSsl = */ false,`)

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,135 @@
...
+      }
+    );
+
+    promise.catch(reject);
+    console.log(`Waiting for emulator to start up ...`);
+    setTimeout(function wait() {
+      console.log(`Ping emulator at [http://localhost:${EMULATOR_PORT}] ...`);
+      request(`http://localhost:${EMULATOR_PORT}`, (error, response) => {
+        if (error && error.code === 'ECONNREFUSED') {
+          setTimeout(wait, 1000);
```
ryanpbrewster: you may want to think about (or leave a TODO for) timing out here, for instance if the emulator has crashed on startup

packages/database/test/helpers/util.ts
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: Crash can actually get caught since it will cause the process to exit. But we do will run into problem if emulator hung during startup. Added code to handle this timeout.

packages/database/test/helpers/util.ts
Feiyang1: i'd like to have as few env variables as possible. I would do const USE_EMULATOR = !! EMULATOR_PORT;

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,140 @@
...
+  } finally {
+    if (emulator) {
+      console.log(`Shutting down emulator, pid: [${emulator.pid}] ...`);
+      emulator.kill();
+    }
+  }
+}
+
+start().catch(err => {
+  console.error(err);
```
Feiyang1: This catch block never runs unless `emulator.kill()` throws. Is it your intention?

yifanyang: Yes. I'm also wondering if I should omit the catch block in function 'start', and handle all the errors here. I'm not sure which one is better.

packages/database/test/helpers/util.ts
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: Agreed. I can help look into this one.

mikelehen: I think [ts-node](https://www.npmjs.com/package/ts-node) does that (and is quite popular).

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+const { spawn } = require('child-process-promise');
+
+const EMULATOR_LINK =
+  'https://storage.googleapis.com/firebase-preview-drop/emulator/firebase-database-emulator-v3.5.0.jar';
+const EMULATOR_FILENAME = 'database-emulator.jar';
+
+const EMULATOR_PORT = 8088;
+const EMULATOR_NAMESPACE = 'test-emulator';
+
+async function delay(seconds) {
```
mikelehen: Is this used somewhere?

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+      if (err) reject(err);
+      console.log(`Created temporary directory at [${dir}].`);
+      const filepath = path.resolve(dir, EMULATOR_FILENAME);
+      const writeStream = fs.createWriteStream(filepath);
+      console.log(`Downloading emulator from [${EMULATOR_LINK}] ...`);
+      request(EMULATOR_LINK)
+        .pipe(writeStream)
+        .on('finish', () => {
+          resolve(filepath);
+          console.log(`Saved emulator binary file to [${filepath}].`);
```
mikelehen: nit: I'd do the console.log() before resolving the promise.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+    const start = Date.now();
+    setTimeout(function wait() {
+      if (Date.now() - start > timeout * 1000) {
+        reject(`Emulator not ready after ${timeout}s. Exiting ...`);
+      } else {
+        console.log(`Ping emulator at [http://localhost:${EMULATOR_PORT}] ...`);
+        request(`http://localhost:${EMULATOR_PORT}`, (error, response) => {
+          if (error && error.code === 'ECONNREFUSED') {
+            setTimeout(wait, 1000);
+          } else if (response && response.statusCode === 400) {
```
mikelehen: 400 is BAD REQUEST...  Is there any way we can get a 200 instead?  Else, I'd at least add a comment why we are sending a bad request.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+            console.log('Emulator has started up successfully!');
+            resolve(promise.childProcess);
+          } else {
+            reject(error);
```
mikelehen: As written if we get a response but the statusCode is something other than 400, we'll reject the promise with a null error, which isn't very useful.  Might be worth reworking.

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+          }
+        });
+      }
+    }, 1000);
+  });
+}
+
+async function loadPublicRules() {
```
mikelehen: "load" doesn't quite seem like the right verb to me.  I'd prefer "upload" or "set" I think. (and please update console.log to match)

scripts/testing/database-emulator-test.js
```diff
@@ -0,0 +1,139 @@
...
+  return new Promise((resolve, reject) => {
+    request.put(
+      {
+        uri: `http://localhost:${EMULATOR_PORT}/.settings/rules.json?ns=${EMULATOR_NAMESPACE}`,
+        headers: { Authorization: 'Bearer owner' },
+        body: '{ "rules": { ".read": true, ".write": true } }'
+      },
+      (error, response, body) => {
+        if (error) reject(error);
+        console.log(body);
```
mikelehen: Is this intentional? Or was it just for debugging?

scripts/testing/database-emulator-test.js
mikelehen: I think I would remove the catch { } block above and move the `process.exitCode = 1;` to here.

scripts/testing/database-emulator-test.js
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: Added a comment here.

scripts/testing/database-emulator-test.js
yifanyang: Done.

scripts/testing/database-emulator-test.js
yifanyang: It's no longer used. Removed.

scripts/testing/database-emulator-test.js
yifanyang: Done. Code should never run to this line, but if it does, now we rejecting with both the error and response information.

scripts/testing/database-emulator-test.js
yifanyang: Changed to 'set'. 'load' is actually borrowed from emulator doc page for [database](https://firebase.google.com/docs/database/security/test-rules-emulator) and [firestore](https://firebase.google.com/docs/firestore/security/test-rules-emulator). (Search for 'load' in these pages.) Do we want to update the doc as well?

scripts/testing/database-emulator-test.js
yifanyang: Changed to `console.log(`Done setting public rule to emulator. Response: ${body}.`);`. This is to indicate success set of rules. `body` is {"status":"ok"} in normal cases, which should not be too disruptive.

PR #1422, opened 2018-12-07
Allow schema downgrades

gsoltis:
Adds a schema migration that switches to our own schema version number. In the future, we will be able to downgrade as far as this migration. 

Adds a test that we don't delete object stores.

Switches existing migrations to use guards as an example. In practice we can't safely downgrade far enough that they would be necessary.

### DIFF
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -318,7 +318,9 @@ function addManualSchemaVersion(
   return writeManualSchemaVersion(txn, FIRST_MANUAL_SCHEMA_VERSION);
 }
 
-function readManualSchemaVersion(txn: SimpleDbTransaction): PersistencePromise<number> {
+function readManualSchemaVersion(
+  txn: SimpleDbTransaction
+): PersistencePromise<number> {
   const store = txn.store<DbVersionGlobalKeyType, DbVersionGlobal>(
     DB_VERSION_GLOBAL_STORE
   );
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -21,6 +21,7 @@ import { assert } from '../util/assert';
 
 import { SnapshotVersion } from '../core/snapshot_version';
 import { BATCHID_UNKNOWN } from '../model/mutation_batch';
+import { Code, FirestoreError } from '../util/error';
 import { encode, EncodedResourcePath } from './encoded_resource_path';
 import { removeMutationBatch } from './indexeddb_mutation_queue';
 import { getHighestListenSequenceNumber } from './indexeddb_query_cache';
@@ -42,8 +43,14 @@ import { SimpleDbSchemaConverter, SimpleDbTransaction } from './simple_db';
  * 5. Removal of held write acks.
  * 6. Create document global for tracking document cache size.
  * 7. Ensure every cached document has a sentinel row with a sequence number.
+ * 8. Switch to independent schema version
  */
-export const SCHEMA_VERSION = 7;
+export const SCHEMA_VERSION = 8;
+/**
+ * This is the first version at which we started handling the version manually.
+ * If the manual version number is missing, start with this value.
+ */
+const FIRST_MANUAL_SCHEMA_VERSION = 8;
 
 /** Performs database creation and schema upgrades. */
 export class SchemaConverter implements SimpleDbSchemaConverter {
@@ -62,18 +69,24 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
     fromVersion: number,
     toVersion: number
   ): PersistencePromise<void> {
+    // We assert that we have sane values, but we now allow downgrades, so
+    // fromVersion may be less than toVersion.
     assert(
-      fromVersion < toVersion &&
-        fromVersion >= 0 &&
-        toVersion <= SCHEMA_VERSION,
-      `Unexpected schema upgrade from v${fromVersion} to v{toVersion}.`
+      fromVersion >= 0 && toVersion <= SCHEMA_VERSION,
+      `Unexpected schema upgrade from v${fromVersion} to v${toVersion}.`
     );
 
+    /*
+     * NOTE: Even though it is not possible to downgrade this far, existing
+     * migrations are following the patterns required for future migrations to
+     * allow downgrade as an example.
+     */
+
     if (fromVersion < 1 && toVersion >= 1) {
-      createPrimaryClientStore(db);
-      createMutationQueue(db);
-      createQueryCache(db);
-      createRemoteDocumentCache(db);
+      createV1PrimaryClientStore(db);
+      createV1MutationQueue(db);
+      createV1QueryCache(db);
+      createV1RemoteDocumentCache(db);
     }
 
     // Migration 2 to populate the targetGlobal object no longer needed since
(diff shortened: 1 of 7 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -122,6 +135,22 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+        'version. IndexedDB should never see its version go past 8.'
+    );
+    if (fromVersion < 8 && toVersion >= 8) {
+      p = p.next(() => this.addManualVersion(txn, db));
+    }
+
+    // IMPORTANT: new migrations should be implemented in
+    // `upgradeFromManualVersion`. After this point, we no longer use
+    // IndexedDB's versioning, as it does not allow downgrades.
+    if (toVersion >= FIRST_MANUAL_SCHEMA_VERSION) {
```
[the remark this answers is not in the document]
mikelehen: Oh...  So on this note (LAST_INDEXEDDB_SCHEMA_VERSION), are we missing the code that makes it so we *stop* updating the IndexedDB version number past 8 (else we still are stuck not being able to downgrade, right)?  Or am I confused?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -122,9 +141,43 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+    // IndexedDB's versioning, as it does not allow downgrades.
+
     return p;
   }
 
+  doManualMigrations(
+    db: IDBDatabase,
+    toVersion: number
+  ): PersistencePromise<void> {
+    return readManualSchemaVersion(db).next(fromVersion => {
```
mikelehen: 1. I'm having trouble reasoning about our transaction usage here.  readManualSchemaVersion creates a new transaction, does some stuff, and then in this next() callback (so we're still inside that transaction?), we create _another_ transaction.  Does that result in "nested" transactions?  Or does opening a new transaction end the old one?  If so, should we be doing something with the completionPromise on the old one?  This would be cleaner / safer if we just wrapped `db` in a `SimpleDB` instance early on, and then used db.runTransaction() rather than manually instantiating SimpleDbTransaction (since runTransaction returns a Promise that resolves after the transaction is committed).

2. Why are we creating two transactions anyway?  I think we should be reading the current schema version and doing the migrations atomically.  Else, technically, two tabs could end up doing the migrations on top of each other or something (maybe non-harmful since we made them idempotent, but still...).

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -122,9 +141,43 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+      // We don't yet know what migrations will run, and we don't know which ones might be skipped
+      // that added object stores that we will need to access. Rather than attempt to pick the exact
+      // right set of object stores based on the current version, add all of the existing object
+      // stores to the transaction.
```
mikelehen: FWIW I think you could drop this comment if you wanted.  It seems pretty reasonable to me to just lock all the stores when we're doing migrations. I don't think it requires explanation.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -222,6 +275,116 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+     *    reading and writing them.
+     *  * Guard schema additions. Check if stores exist before adding them.
+     *  * Data migrations should *probably* always run. Older versions of the SDK will not have
+     *    maintained invariants from later versions, so migrations that update values cannot
+     *    assume that existing values have been properly maintained. Calculate them again, if
+     *    applicable.
+     */
+
+    /*
+     * As above, the migrations should be in the form of
```
mikelehen: "As above" is pretty ambiguous at this point in the code. I'd drop it.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -222,6 +275,116 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+  version: number
+): PersistencePromise<void> {
+  const store = txn.store<DbVersionGlobalKeyType, DbVersionGlobal>(
+    DB_VERSION_GLOBAL_STORE
+  );
+  return store.put(DB_VERSION_GLOBAL_KEY, version);
+}
+
+function doTestMigration(txn: SimpleDbTransaction): PersistencePromise<void> {
+  // Dummy migration to test a version number not based on indexeddb's versioning scheme.
```
mikelehen: Suggested rewording to be a bit more explicit:

// Dummy migration that does nothing. It only exists for tests to exercise our "manual" version number mechanism and the ability to downgrade / re-upgrade successfully.

packages/firestore/src/local/simple_db.ts
```diff
@@ -31,6 +34,76 @@ export interface SimpleDbSchemaConverter {
...
     toVersion: number
   ): PersistencePromise<void>;
+
+  doManualMigrations(
+    db: IDBDatabase,
+    toVersion: number
+  ): PersistencePromise<void>;
+}
+
+export function openIndexedDb(
```
mikelehen: Perhaps comment that this is only exported for test usage?

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -602,6 +586,66 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+      await withDb(version, async db => {
+        const thisVersionStores = db.objectStoreNames;
+        for (const store of stores) {
+          expect(thisVersionStores.contains(store)).to.be.true;
+        }
+        // Reload the existing stores to check the next version.
+        stores = [];
+        for (let i = 0; i < thisVersionStores.length; i++) {
+          stores.push(thisVersionStores[i]);
+        }
```
mikelehen: `Array.from(thisVersionStores)` if you want.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -602,6 +586,66 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+      expect(version).to.equal(SCHEMA_VERSION);
+    });
+    // downgrade to first manual version
+    await withDb(FIRST_MANUAL_SCHEMA_VERSION, async db => {
+      expect(db.version).to.equal(FIRST_MANUAL_SCHEMA_VERSION);
+      const version = await readManualVersion(db);
+      expect(version).to.equal(FIRST_MANUAL_SCHEMA_VERSION);
+    });
+    // upgrade to latest version again
+    await withDb(SCHEMA_VERSION, async db => {
```
mikelehen: You've probably already done this, but maybe set a breakpoint here and on the dummy migration, and just make sure it gets called again.... since the test wouldn't notice if we updated the version number but didn't re-run the migrations (I think).

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -122,9 +140,35 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+  doManualMigrations(
+    db: IDBDatabase,
+    toVersion: number
+  ): PersistencePromise<void> {
+    const objectStores = Array.from(db.objectStoreNames);
+    const txn = new SimpleDbTransaction(
+      db.transaction(objectStores, 'readwrite')
+    );
+    return readManualSchemaVersion(txn).next(fromVersion =>
+      this.upgradeFromManualVersion(txn, fromVersion, toVersion)
```
mikelehen: I just realized:

1. upgradeFromManualVersion() will need a reference to `db` in order to create / delete object stores, etc.
2. But you can only call `db.createObjectStore()`, etc. from within the implicit "versionchange" transaction that's opened during the onupgradeneeded event...  but we're running migrations *after* the database has been opened and the "versionchange" transaction has completed. So any attempt to create object stores will fail.

So I think for this to work, we need to try to move the manual migrations into the onupgradeneeded handler...  Is that possible?  Or is it time to rethink web downgrade-ability?

gsoltis:
Ok, as discussed, let's abandon this approach in favor of a friendlier error message. While it may possibly be technically feasible to always get a version upgrade transaction, it would be a pretty horrendous implementation.

PR #1424, opened 2018-12-10
Finish type migration and get rid of AnyDuringMigration

gsoltis:
Fixes b/66916745

Of note: `sequence()` was barely being used, so rather than port the declared functionality, I removed what wasn't being used.

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -17,7 +17,6 @@
 import { assert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { debug } from '../util/log';
-import { AnyDuringMigration } from '../util/misc';
 import { AnyJs } from '../util/misc';
 import { Deferred } from '../util/promise';
 import { SCHEMA_VERSION } from './indexeddb_schema';
@@ -173,6 +172,7 @@ export class SimpleDb {
       .catch(error => {
         // Abort the transaction if there was an error.
         transaction.abort(error);
+        return PersistencePromise.reject<T>(error);
       })
       .toPromise();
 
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -173,15 +172,14 @@ export class SimpleDb {
       .catch(error => {
         // Abort the transaction if there was an error.
         transaction.abort(error);
+        return PersistencePromise.reject<T>(error);
```
mikelehen: I'm having trouble figuring out what this return value affects, if anything.  My guess is that it doesn't affect anything, because we'll have aborted the transaction and so transaction.completionPromise will be a failed promise below and we won't use transactionFnResult.  If that's correct, maybe add a comment to this effect just so it's clear how the error is expected to be propagated.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -161,7 +157,7 @@ class MockConnection implements Connection {
 
   waitForWriteRequest(): Promise<api.WriteRequest> {
     if (this.earlyWrites.length > 0) {
-      return Promise.resolve(this.earlyWrites.shift()) as AnyDuringMigration;
+      return Promise.resolve(this.earlyWrites.shift()!);
```
mikelehen: To avoid `!` you could probably rewrite this as:

```
const earlyWrite = this.earlyWrites.shift();
if (earlyWrite) {
  return Promise.resolve(earlyWrite);
}
```

Up to you though. I like to avoid `!`.

gsoltis: Oh yeah, good suggestion. I like that better.

packages/firestore/src/local/simple_db.ts
gsoltis: Done.

PR #1433, opened 2018-12-18
Port performance optimizations to speed up reading large collections from Android

var-const:
Straightforward port of https://github.com/firebase/firebase-android-sdk/pull/123.


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -178,6 +183,71 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+    let sizeMap = new SortedMap<DocumentKey, number>(DocumentKey.comparator);
+    if (documentKeys.isEmpty()) {
+      return PersistencePromise.resolve({ maybeDocuments: results, sizeMap });
+    }
+
+    const range = IDBKeyRange.bound(
+      documentKeys.first()!.path.toArray(),
+      documentKeys.last()!.path.toArray()
+    );
+    let key = documentKeys.first();
```
var-const: This is built upon the idea that `control.skip()` makes the algorithm more efficient. Will be happy with any feedback.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -178,6 +183,72 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+
+  /**
+   * Looks up several entries in the cache.
+   *
+   * @param documentKeys The set of keys entries to look up.
+   * @return A map of MaybeDocuments indexed by key (if a document cannot be
+   *     found, the key will be mapped to null) and a map of sizes indexed by
+   *     key (zero if the key cannot be found).
+   */
+  getSizedEntries(
```
var-const: Another approach would be for `getSizedEntries` to return a `SortedMap<DocumentKey, DocumentSizeEntry>`. I decided in favor of returning two maps because it makes it easier to avoid code duplication between `getEntries` and `getSizedEntries`.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -178,6 +183,72 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+    return remoteDocumentsStore(transaction)
+      .iterate({ range }, (potentialKeyRaw, dbRemoteDoc, control) => {
+        const potentialKey = DocumentKey.fromSegments(potentialKeyRaw);
+        while (DocumentKey.comparator(key!, potentialKey) != 1) {
+          if (key!.isEqual(potentialKey)) {
+            results = results.insert(
+              key!,
+              this.serializer.fromDbRemoteDocument(dbRemoteDoc)
+            );
+            sizeMap = sizeMap.insert(key!, dbDocumentSize(dbRemoteDoc));
```
var-const: Whether to update the `sizeMap` could be controlled by a flag to avoid doing unnecessary work when this function is invoked via `getEntries`, but it might be overkill.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -99,6 +107,32 @@ export abstract class RemoteDocumentChangeBuffer {
...
+   * @return The cached Document or NoDocument entry, or null if we have nothing
+   * cached.
+   */
+  getEntries(
+    transaction: PersistenceTransaction,
+    documentKeys: DocumentKeySet
+  ): PersistencePromise<NullableMaybeDocumentMap> {
+    const changes = this.assertChanges();
+
+    // Record the size of everything we load from the cache so we can compute a delta later.
```
var-const: Question: I could add code to look up the keys in the buffer. I decided to postpone it because it will necessitate some logic to merge the buffer lookup results with results from cache, so I'd like to check beforehand if you think this optimization is worthwhile.

var-const:
Michael, I'll port the tests and measure performance numbers tomorrow, but I think it should be in good enough shape for review. Thanks!

packages/firestore/src/model/collections.ts
```diff
@@ -37,6 +37,20 @@ export function maybeDocumentMap(): MaybeDocumentMap {
...
+export type NullableMaybeDocumentMap = SortedMap<
+  DocumentKey,
+  MaybeDocument | null
+>;
+
+export function nullableMaybeDocumentMap(): NullableMaybeDocumentMap {
+  return maybeDocumentMap();
+}
+
+export type DocumentSizeEntries = {
```
gsoltis: I am nervous about this type. I know currently that the keys are kept in sync, but there is nothing here that guarantees it.

gsoltis:
Ugh, I just saw I was marked as a reviewer but not assigned. Sorry about the drive-by review. I guess defer to whatever @mikelehen suggests.

packages/firestore/src/model/document.ts
```diff
@@ -59,7 +61,12 @@ export class Document extends MaybeDocument {
     key: DocumentKey,
     version: SnapshotVersion,
     readonly data: ObjectValue,
-    options: DocumentOptions
+    options: DocumentOptions,
+    /**
+     * Memoized serialized form of the document for optimization purposes (avoids repeated
+     * serialization). Might be undefined.
+     */
+    readonly proto?: api.Document
```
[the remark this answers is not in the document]
mikelehen: FWIW- If we are, I don't think it's the end of the world.  So I wouldn't go terribly far out-of-our-way to guarantee this unless it shows up as a problem.

packages/firestore/src/util/sorted_set.ts
```diff
@@ -97,6 +97,19 @@ export class SortedSet<T> {
     }
   }
 
+  /** Finds the least element greater than `elem`. */
+  firstAfter(elem: T): T | null {
```
mikelehen: So...  the consuming code is really just trying to iterate through the set.  But because we call firstAfter() every time we want to iterate to the next key, instead of it being an O(n) traversal it's an O(n log n) traversal...  which isn't really ideal.

So it seems like we should either try to expose an iterator interface of some kind on SortedSet or just take the lazy approach and call .toArray() (which creates an unnecessary copy but at least it's O(n)) and then iterate over that.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -74,6 +77,25 @@ export class LocalDocumentsView {
...
 
+  // Returns the view of the given `docs` as they would appear after applying
+  // all mutations in the given `batches`.
+  private applyLocalMutationsToDocuments(
+    transaction: PersistenceTransaction,
+    docs: NullableMaybeDocumentMap,
+    batches: MutationBatch[]
+  ): PersistencePromise<NullableMaybeDocumentMap> {
+    let results = nullableMaybeDocumentMap();
+    return new PersistencePromise<NullableMaybeDocumentMap>(resolve => {
```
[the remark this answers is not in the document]
mikelehen: @var-const In general, JavaScript is completely single-threaded (there's no way to block other than by spinning the CPU) and so wrapping a computation in a promise doesn't really help (in particular it won't enable any parallelism).

And PersistencePromise is extra weird because it's a specially-designed Promise-like construct that tries to be as synchronous as possible because IndexedDb has weird semantics where in the completion for one operation you *must* synchronously start the next operation or else your transaction will auto-close.  So even using PersistencePromise, this code is actually 100% synchronous.  So you can go ahead and just yank PersistencePromise out.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -84,28 +106,40 @@ export class LocalDocumentsView {
...
+
+  /**
+   * Similar to `getDocuments`, but creates the local view from the given
+   * `baseDocs` without retrieving documents from the local store.
+   */
+  getLocalViewOfDocuments(
+    transaction: PersistenceTransaction,
+    baseDocs: NullableMaybeDocumentMap
+  ): PersistencePromise<MaybeDocumentMap> {
+    let allKeys = documentKeySet();
```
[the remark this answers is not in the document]
mikelehen: Can you try using `AnyJs` instead of `any`?  

any is dangerous in that it basically opts out of typechecking.  So if you type the values as `any`, TypeScript will let you do anything with the values (call nonexistent methods, etc.) which is the opposite of what we want (we don't want to be able to do anything with the values).

`AnyJs` is a type we introduced essentially meant to be a supertype of any JS type (similar to `Object` in java). So TypeScript won't let you do anything with it without doing typeof / instanceof checks to narrow it down.

TypeScript 3.x actually introduced `unknown` which is similar to AnyJs and I'm hoping we can migrate to it soon.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
var-const: Done, please take a look.

packages/firestore/src/local/local_documents_view.ts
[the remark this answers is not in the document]
var-const: Sorry, missed this comment. Done now.

var-const:
Tried to do some performance measuring (I had a lot of trouble with Chrome apparently somehow caching the Firestore code being used, which only got resolved by restarting the browser. After that, the behavior seemed to be consistent, though I'm still slightly unsure).

Like other platforms, I tried reading a collection consisting of 1K somewhat large documents. Numbers fluctuate quite a bit:
* master -- 1.8s - 3.1s (raw numbers: 3104ms 2433ms 2863ms 2447ms 2545ms 1916ms 1894ms 1810ms 1865ms);
* branch -- 1.6s - 2s (raw numbers: 1559ms 1635ms 1915ms 2012ms 1734ms 1702ms 1733ms 1736ms 1843ms). The median seems to be around 1.7s

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -206,46 +218,72 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
   ): PersistencePromise<DocumentSizeEntries> {
     let results = nullableMaybeDocumentMap();
     let sizeMap = new SortedMap<DocumentKey, number>(DocumentKey.comparator);
```
mikelehen: Sorry, now that getEntries() is no longer a wrapper around getSizedEntries(), can we drop sizeMap and have this be new SortedMap<DocumentKey, DocumentSizeEntry|null> ?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -206,46 +218,72 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+          results = results.insert(key, null);
+          sizeMap = sizeMap.insert(key, 0);
+        }
+      }
+    ).next(() => {
+      return { maybeDocuments: results, sizeMap };
+    });
+  }
+
+  forEachDbEntry(
```
mikelehen: private please

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -108,8 +108,7 @@ export abstract class RemoteDocumentChangeBuffer {
   }
 
   /**
-   * Looks up several entries in the cache.
-   * checked, and if no buffered change applies, this will forward to
+   * Looks up several entries in the cache, orwarding to
```
mikelehen: forwarding :)

var-const: Thanks!

packages/firestore/src/local/indexeddb_remote_document_cache.ts
var-const: Done.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
var-const: First, I don't feel strongly about this. The reason I set it up that way is so that `getEntries` in `RemoteDocumentChangeBuffer` can return the `MaybeDocumentMap` directly. If this were a sorted map of `DocumentSizeEntry`s, then `getEntries` would have to create a new `MaybeDocumentMap` and fill it with just `MaybeDocument`s.

It's probably not a big deal, so if you think code clarity is more important here, I'll do the change.

mikelehen: Oh, sorry! I was confused.  I thought it was just so the old getEntries() implementation could return the MaybeDocumentMap directly.  But I see now that RemoteDocumentChangeBuffer.getEntries() ends up calling getSizedEntries() and using the sizes and also passing the MaybeDocumentMap straight through.  So it needs both, and the way it's structured right now makes sense.

So nevermind. Please keep it the way it is.

PR #1368, opened 2018-11-08
Add Numeric Add

schmidt-sebastian:
Port of firebase/firebase-android-sdk#105

Unlike Android, this PR doesn't treat INT64 overflows, as JavaScript's precision is limited to 53 bit. We likely can't properly expose the full backend semantics to our customers since we eventually surface the results as a number.

This limitation is currently - along with the rest of the API - going through API review.

PR #1443, opened 2018-12-26
Remove some Array to SortedSet<FieldPath> conversion

schmidt-sebastian:
There are a couple of places in UserDataConverter that construct an array of FieldPaths, which is then  immediately converted to a SortedSet. This PR directly constructs the SortedSet.

Note that I did not touch the remaining FieldPath array in ParseContext, as it is modified in place and requires a mutable data structure.

See commit https://github.com/firebase/firebase-js-sdk/pull/1368/commits/e62de99269434c4c55389dbb10b87fd645d02d4c#diff-af007bc1ca0cf1509f29df38bed7d308L609 for the rationale.

### DIFF
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -42,8 +42,12 @@ import { TransformOperation } from './transform_operation';
  *             containing foo
  */
 export class FieldMask {
-  constructor(readonly fields: SortedSet<FieldPath>) {
+  private constructor(readonly fields: SortedSet<FieldPath>) {
     // TODO(dimond): validation of FieldMask
+  }
+
+  static fromSet(fields: SortedSet<FieldPath>): FieldMask {
+    return new FieldMask(fields);
   }
 
   static fromArray(fields: FieldPath[]): FieldMask {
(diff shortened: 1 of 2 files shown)


packages/firestore/src/model/mutation.ts
```diff
@@ -42,7 +42,7 @@ import { TransformOperation } from './transform_operation';
  *             containing foo
  */
 export class FieldMask {
-  private constructor(readonly fields: SortedSet<FieldPath>) {
+  constructor(readonly fields: SortedSet<FieldPath>) {
```
rsgowman: For consistency with fromArray (as well as for consistency with the other platforms) you might consider leaving the ctor private and adding a 'fromSet' method.

In particular, the android sdk has (copy+pasted):

```
public final class FieldMask {
  public static FieldMask fromSet(Set<FieldPath> mask) {
    return new FieldMask(mask);
  }

  private final Set<FieldPath> mask;

  private FieldMask(Set<FieldPath> mask) {
    this.mask = mask;
  }
```

OTOH, the C++ just uses the ctor, so you'll end up being inconsistent with something, regardless of which you choose. :)

Optional.

schmidt-sebastian: Done.

Issue #1445, opened 2018-12-27
Testing: initializeAdminApp fails with `"projectId" not provided in firebase.initializeApp` if you also pass databaseName


sergeybekrin:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 10.14.2
  * Node version: v8.14.1
  * `firebase-admin`: 6.4.0
  * `@firebase/testing`: 0.5.1
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

Hey there 👋

I'm trying to run tests locally via `@firebase/testing` and with emulated firestore, but apprently it fails due to `projectId` is [not being passed to `firebase.initializeApp`](https://github.com/firebase/firebase-js-sdk/blob/master/packages/testing/src/api/index.ts#L113). I'm not quite sure if it's intentional or not.

#### Steps to reproduce:

1. Import `@firebase/testing`
2. Call `.initializeAdminApp` from it

#### Relevant Code:

```javascript
const testing = require('@firebase/testing');

const app = testing.initializeAdminApp({
  projectId: 'foo',
  databaseName: 'bar'
});
```

#### Error:
```
FirebaseError: "projectId" not provided in firebase.initializeApp.
    at new FirestoreError (./node_modules/@firebase/firestore/dist/index.node.cjs.js:348:28)
    at Function.Firestore.databaseIdFromApp (./node_modules/@firebase/firestore/dist/index.node.cjs.js:19361:19)
    at new Firestore (./node_modules/@firebase/firestore/dist/index.node.cjs.js:19270:43)
    at Object.firestore (./node_modules/@firebase/firestore/dist/index.node.cjs.js:20613:79)
    at FirebaseAppImpl._getService (./node_modules/@firebase/app/dist/index.node.cjs.js:138:66)
    at FirebaseAppImpl.(anonymous function) [as firestore] (./node_modules/@firebase/app/dist/index.node.cjs.js:327:31)
    at initializeApp (./node_modules/@firebase/testing/dist/index.cjs.js:116:9)
    at Object.initializeAdminApp (./node_modules/@firebase/testing/dist/index.cjs.js:84:10)
    at [eval]:1:30
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
```

mikelehen:
@ryanpbrewster It looks like the SDK is intended to be initialized with `projectId` OR `databaseName` but not both.  Is this intentional?  If so, perhaps we could throw an error if somebody tries...

[...]

PR #1449, opened 2019-01-02
Allow both databaseName and projectId

ryanpbrewster:
This should resolve https://github.com/firebase/firebase-js-sdk/issues/1445

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -109,15 +109,10 @@ function initializeApp(
 ): firebase.app.App {
   let appOptions = {};
   if (databaseName) {
-    appOptions = {
-      databaseURL: `http://${DATABASE_ADDRESS}?ns=${databaseName}`
-    };
-  } else if (projectId) {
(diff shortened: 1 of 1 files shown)


packages/testing/src/api/index.ts
```diff
@@ -109,15 +109,10 @@ function initializeApp(
...
   if (databaseName) {
-    appOptions = {
-      databaseURL: `http://${DATABASE_ADDRESS}?ns=${databaseName}`
-    };
-  } else if (projectId) {
-    appOptions = {
-      projectId: projectId
-    };
-  } else {
-    throw new Error('neither databaseName or projectId were specified');
```
tonymeng: we still need this check don't we?

ryanpbrewster: The underlying `FirebaseApp` already checks that `databaseURL` is defined when `.database()` is called, and that `projectId` is defined when `.firestore()` is called. I think these are just as intuitive as our existing error message.

PR #1451, opened 2019-01-02
Fixes #1450: ThenableReference should extend Promise, not PromiseLike.

### LINKED ISSUE #1450: Update @firebase/database-types
### Describe your environment

  * Operating System version: macOS 10.14.2
  * Browser version: N/A
  * Firebase SDK version: N/A
  * Firebase Product: database

### Describe the problem
`ThenableRefrence` doesn't have `catch`, while `Reference` does.

https://github.com/firebase/firebase-js-sdk/blob/c6357fbaf99e80257816dafb28e748de4a039edf/packages/database-types/index.d.ts#L101

https://github.com/firebase/firebase-js-sdk/blob/c6357fbaf99e80257816dafb28e748de4a039edf/packages/database/src/api/Reference.ts#L316

Reason for issue: https://github.com/firebase/firebase-admin-node/issues/433

mikelehen:
Since we implement .catch(), we should extend Promise.

### DIFF
--- a/packages/database-types/index.d.ts
+++ b/packages/database-types/index.d.ts
@@ -125,7 +125,7 @@ export interface ServerValue {
   };
 }
 
-export interface ThenableReference extends Reference, PromiseLike<Reference> {}
+export interface ThenableReference extends Reference, Promise<Reference> {}
 
 export function enableLogging(
   logger?: boolean | ((a: string) => any),
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -585,7 +585,7 @@ declare namespace firebase.database {
 
   interface ThenableReference
     extends firebase.database.Reference,
-      PromiseLike<any> {}
+      Promise<Reference> {}
 
   function enableLogging(
     logger?: boolean | ((a: string) => any),


packages/database-types/index.d.ts
```diff
@@ -125,7 +125,7 @@ export interface ServerValue {
   };
 }
 
-export interface ThenableReference extends Reference, PromiseLike<Reference> {}
+export interface ThenableReference extends Reference, Promise<Reference> {}
```
Feiyang1: ThenableReference doesn't really extend Promise, it just looks like it. The code below reassigns the catch to receive the reject callback instead of the real catch callback. I don't know if it's your intention, but the developer might be surprised the catch doesn't work as intended.
https://github.com/firebase/firebase-js-sdk/blob/7a41fe71fd3effd71b61548546e290015f58456b/packages/database/src/api/Reference.ts#L340-L341

I don't know why ThenableReference is implemented this way. Can we can change it to follow the Promise spec? Otherwise it's not correct to say it extends Promise and I think it will add confusion to the api.

mikelehen: Oh, hrm. I didn't realize that's how it's defined, but I _think_ that still matches the Promise spec.  Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch:

> The catch() method returns a Promise and deals with rejected cases only. It behaves the same as calling Promise.prototype.then(undefined, onRejected) (in fact, calling obj.catch(onRejected) internally calls obj.then(undefined, onRejected))

It seems like when Promises were still new-ish, not all implementations provided .catch(), so that may be why it was implemented that way...  Though I suspect that's not a concern anymore.

Feiyang1: Sorry, I take it back. You are right that it complies with the Promise Spec. I was thrown off by the `thennablePushRef.catch = promise.then.bind(promise, undefined); ` which is the same as `thennablePushRef.catch = promise.catch.bind(promise)`.

dimitarnestorov:
@mikelehen did you close by mistake?

PR #1453, opened 2019-01-04
Added firestoreClearData() testing sdk method

scottcrossen:
This PR exposes a new method for the @firebase/testing package to give users the ability to delete their entire database with one call. This is particularly useful in testing frameworks during the before/beforeEach lifecycle phases to isolate each test from one another. This only works with a properly updated Firestore Emulator (>v1.2.4)

### DIFF
--- a/packages/testing/index.ts
+++ b/packages/testing/index.ts
@@ -24,11 +24,11 @@ export {
   apps,
   assertFails,
   assertSucceeds,
+  clearFirestoreData,
   database,
   firestore,
   initializeAdminApp,
   initializeTestApp,
   loadDatabaseRules,
-  loadFirestoreRules,
-  clearFirestoreData
+  loadFirestoreRules
 } from './src/api';
--- a/packages/testing/index.ts
+++ b/packages/testing/index.ts
@@ -29,5 +29,6 @@ export {
   initializeAdminApp,
   initializeTestApp,
   loadDatabaseRules,
-  loadFirestoreRules
+  loadFirestoreRules,
+  clearFirestoreData
 } from './src/api';
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -220,6 +220,44 @@ export function loadFirestoreRules(
   });
 }
 
+export type clearFirestoreDataOptions = {
+  projectId: string;
+  databaseId?: string;
+};
+export function clearFirestoreData(
+  options: clearFirestoreDataOptions
+): Promise<void> {
+  if (!options.projectId) {
+    throw new Error('projectId not specified');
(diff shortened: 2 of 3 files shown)


packages/testing/index.ts
```diff
@@ -29,5 +29,6 @@ export {
   initializeAdminApp,
   initializeTestApp,
   loadDatabaseRules,
-  loadFirestoreRules
+  loadFirestoreRules,
+  clearFirestoreData
```
ryanpbrewster: sort alphabetically

packages/testing/src/api/index.ts
```diff
@@ -220,6 +220,44 @@ export function loadFirestoreRules(
   });
 }
 
+export type clearFirestoreDataOptions = {
+  projectId: string;
+  databaseId?: string;
```
ryanpbrewster: I would not implement a databaseId yet, these are not supported well

packages/testing/src/api/index.ts
```diff
@@ -220,6 +220,44 @@ export function loadFirestoreRules(
...
+export function clearFirestoreData(
+  options: clearFirestoreDataOptions
+): Promise<void> {
+  if (!options.projectId) {
+    throw new Error('projectId not specified');
+  }
+
+  const optionsWithDefaults = Object.assign({}, options, {databaseId: '(default)'})
+
+  let client = new EMULATOR.FirestoreEmulator(
```
ryanpbrewster: we should think about caching these stubs in the future

scottcrossen: I agree!

packages/testing/src/api/index.ts
```diff
@@ -220,6 +220,41 @@ export function loadFirestoreRules(
   });
 }
 
+export type clearFirestoreDataOptions = {
```
ryanpbrewster: types should be PascalCase

scottcrossen: [face-palm]

scottcrossen:
I'll merge this once the emulator gets updated to the next release.

tonymeng:
thoughts on mirroring this functionality for the rtdb?

ryanpbrewster:
This functionality already exists for the RTDB, because RTDB deletes are recursive by default

scottcrossen:
@tonymeng I'm not going to prioritize that because there's already a way as @ryanpbrewster mentioned. Additionally, we show them how in the [emulator quickstart](https://github.com/firebase/quickstart-nodejs/tree/master/database-emulator). If we get feedback from our developers that they'd like this testing-sdk method then I'd be happy to quickly write it out!

PR #1456, opened 2019-01-07
Treat ABORTED writes as retryable

wilhuff:
This corresponds to a server-side change that will be released for the v1 protocol only.


wilhuff:
I looked at the transaction code too. While our current handling seems specious there I think addressing that is out of scope for this change.

I've addressed the test failure by adding `isPermanentWriteError`. It's a big/subtle enough change that it likely should be re-reviewed. In particular, I'm not super happy that handshakes on the write streams can't be classified with `isPermanentWriteError` but ABORTED is not transient there so I couldn't see a better solution than just documenting that. LMK if you can think of a better arrangement.

PR #1458, opened 2019-01-08
use secret team as code owners to avoid review requests spam.

Feiyang1:
The PR addresses 2 issues:
1. the current CODEOWNERS setup is too rigid and it doesn't work in case like cross package changes and doesn't provide an escape hatch when owners are not available.
2. Automatic review requests are spammy for large teams.

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -109,10 +109,10 @@ function initializeApp(
 ): firebase.app.App {
   let appOptions = {};
   if (databaseName) {
-    appOptions["databaseURL"] = `http://${DATABASE_ADDRESS}?ns=${databaseName}`;
+    appOptions['databaseURL'] = `http://${DATABASE_ADDRESS}?ns=${databaseName}`;
   }
   if (projectId) {
-    appOptions["projectId"] = projectId;
+    appOptions['projectId'] = projectId;
   }
   const appName = 'app-' + new Date().getTime() + '-' + Math.random();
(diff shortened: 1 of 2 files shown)


packages/testing/src/api/index.ts
```diff
@@ -109,10 +109,10 @@ function initializeApp(
 ): firebase.app.App {
   let appOptions = {};
   if (databaseName) {
-    appOptions["databaseURL"] = `http://${DATABASE_ADDRESS}?ns=${databaseName}`;
+    appOptions['databaseURL'] = `http://${DATABASE_ADDRESS}?ns=${databaseName}`;
   }
   if (projectId) {
-    appOptions["projectId"] = projectId;
+    appOptions['projectId'] = projectId;
```
mikelehen: These changes seem unrelated...?  (but if that's what prettier wants, 🤷‍♂️ )

Feiyang1:
@bojeil-google, The global approver team has 2 purposes:
1. act as a fallback group when the owner(s) of a component is not available, e.g. on vacation. It's especially useful for small teams like rxFire, storage.
2. make cross package changes, large scale refactoring easier to get approved. E.g. library version upgrades, packaging changes, etc. 

The situation you mentioned could happen, but as a principle, we trust people do the right thing.

Feiyang1:
It won't be added automatically. Only visible owners will be added.
This is actually the reason why secret team is used - Firestore team doesn't want to receive any review request automatically.

PR #1464, opened 2019-01-09
Change timestampsInSnapshots default to true.

mikelehen:
Also update log messages instructing folks what to do if they're using the
soon-to-be-removed timestampsInSnapshots setting.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -774,22 +774,24 @@ declare namespace firebase.firestore {
     ssl?: boolean;
 
     /**
-     * Enables the use of `Timestamp`s for timestamp fields in
-     * `DocumentSnapshot`s.
+     * Specifies whether to use `Timestamp` objects for timestamp fields in
+     * `DocumentSnapshot`s. This is enabled by default and should not be
+     * disabled.
      *
-     * Currently, Firestore returns timestamp fields as `Date` but `Date` only
-     * supports millisecond precision, which leads to truncation and causes
-     * unexpected behavior when using a timestamp from a snapshot as a part
-     * of a subsequent query.
+     * Previously, Firestore returned timestamp fields as `Date` but `Date`
+     * only supports millisecond precision, which leads to truncation and
+     * causes unexpected behavior when using a timestamp from a snapshot as a
+     * part of a subsequent query.
      *
-     * Setting `timestampsInSnapshots` to true will cause Firestore to return
-     * `Timestamp` values instead of `Date` avoiding this kind of problem. To make
-     * this work you must also change any code that uses `Date` to use `Timestamp`
-     * instead.
+     * So now Firestore returns `Timestamp` values instead of `Date`, avoiding
+     * this kind of problem.
      *
-     * NOTE: in the future `timestampsInSnapshots: true` will become the
-     * default and this option will be removed so you should change your code to
-     * use Timestamp now and opt-in to this new behavior as soon as you can.
+     * To opt into the old behavior of returning `Date` objects, you can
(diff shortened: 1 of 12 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -192,6 +192,37 @@ class FirestoreSettings {
       'timestampsInSnapshots',
       settings.timestampsInSnapshots
     );
+
```
mikelehen: FYI- I moved these log messages from configureClient() to here because in configureClient(), the default has already been applied and so we can't differentiate if the setting is unset or set.

packages/firebase/index.d.ts
```diff
@@ -774,22 +774,24 @@ declare namespace firebase.firestore {
...
+     * causes unexpected behavior when using a timestamp from a snapshot as a
+     * part of a subsequent query.
+     *
+     * So now Firestore returns `Timestamp` values instead of `Date`, avoiding
+     * this kind of problem.
+     *
+     * To opt into the old behavior of returning `Date` objects, you can
+     * temporarily set `timestampsInSnapshots` to false.
+     *
+     * WARNING: This setting will be removed in a future release. You should
```
wilhuff: Deprecation is a common way to signal this kind of "don't-use-this-legacy-thing". Does typescript have a way to formally signal that?

Even with a formal signal it's probably worth explicitly spelling the deprecation out in the text. Something like:

Maybe "This setting is deprecated and will be removed ..."

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 # Unreleased
+- [changed] The `timestampsInSnapshots` setting is now enabled by default
+  so timestamp fields read from a `DocumentSnapshot` will be returned as
+  `Timestamp` objects instead of `Date`. Any cexpecting to receive a
```
wilhuff: s/Any cexpecting/Any code expecting/

packages/firestore/src/api/database.ts
```diff
@@ -192,6 +192,37 @@ class FirestoreSettings {
...
+    // Nobody should set timestampsInSnapshots anymore, but the error depends on
+    // whether they set it to true or false...
+    if (settings.timestampsInSnapshots === true) {
+      log.error(`
+  The timestampsInSnapshots setting now defaults to true and you no
```
wilhuff: Ideally we could mark the API itself deprecated and avoid complaining about this at run-time. People were not happy with us for clogging their logs when we added this the first time.

If there's no such mechanism in typescript then this is fine, but in iOS and Java where we have `NS_DEPRECATED` and `@Deprecated` I'd prefer use those and avoid logging in the case where a user has previously opted into the future behavior.

packages/firebase/index.d.ts
Feiyang1: Typescript doesn't provide a way to do it natively. You can add `@deprecated` annotation, and configure tslint to issue the warning. It's not ideal since we can't expect all our consumers to use tslint, or even typescript, so the best way might be just good documentation and a runtime warning.

mikelehen: Added `@deprecated` since it doesn't hurt anything, but I will keep the log message as well, which I think is the only real way to flag deprecations in JS...

packages/firestore/CHANGELOG.md
mikelehen: Er, woops. Thanks.

packages/firestore/src/api/database.ts
mikelehen: Sounds good.

PR #1465, opened 2019-01-09
Change downloaded emulator file mode to 'rwxr-xr-x'.

yifanyang:
The firestore emulator binary file cannot be started by 'java -jar' with
permission 'rw-r--r--' on Google corp workstation (Debian). The reason
why we didn't notice this before is that the emulator can run on macbook
(macOS Mojave) and Travis VM (Ubuntu Trusty) perfectly without this
change though.

### DIFF
--- a/scripts/emulator-testing/emulators/emulator.ts
+++ b/scripts/emulator-testing/emulators/emulator.ts
@@ -51,8 +51,12 @@ export abstract class Emulator {
           .pipe(writeStream)
           .on('finish', () => {
             console.log(`Saved emulator binary file to [${filepath}].`);
-            this.binaryPath = filepath;
-            resolve();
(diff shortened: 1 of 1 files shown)


scripts/emulator-testing/emulators/emulator.ts
```diff
@@ -51,8 +51,12 @@ export abstract class Emulator {
           .pipe(writeStream)
           .on('finish', () => {
             console.log(`Saved emulator binary file to [${filepath}].`);
-            this.binaryPath = filepath;
-            resolve();
+            fs.chmod(filepath, 0o755, err => {
```
Feiyang1: can you please add a comment here?

yifanyang: Done.

PR #1468, opened 2019-01-09
Do not print duplicate error messages to stdout.

yifanyang:
The [issue](https://github.com/lerna/lerna/issues/1790) with regard to lerna duplicating error messages was resolved recently. With this lerna option '--stream', lerna will output the same error message only once.

See the comparison below. In below demonstrations, I manually added a case that is meant to fail in firestore sdk test:
1. error message without this change: [code change](https://github.com/firebase/firebase-js-sdk/commit/f9dd422e316b), [travis logs](https://travis-ci.org/firebase/firebase-js-sdk/jobs/477506445)
2. error message with this change: [code change](https://github.com/firebase/firebase-js-sdk/compare/d36691d596b0...705c4bd2148f), [traivs logs](https://travis-ci.org/firebase/firebase-js-sdk/jobs/477512157)

Please check how the same error output are duplicated 4 times in the previous way:
- messages printed by mocha;
- messages printed by lerna (prefixed with "lerna ERR!");
- messages printed by lerna (in the "Error" filed of the error object at the end of the log);
- messages printed by lerna (in the "stdout" field of the error object at the end of the log).


mikelehen:
Sounds good to me, but I'll defer to Fei to officially approve.  Thanks!

yifanyang:
Sure. Thanks! I invited both of you since I think this should be helpful to firestore most, as firestore generates much more (usually thousands of) lines of log than other sdks. The duplication would cause a big confusion in such case.

mikelehen:
Yeah, I just want to defer to Fei since upgrading lerna from 2.x to 3.x could have other impacts on the repo and he's the person to sign off on that. :-)

Feiyang1:
Lerna 3 doesn't have any breaking change that affects us, so it should be fine.

yifanyang:
Tested both with and without the `--stream` option. Looks like the explicit `--stream` is necessary for lerna to not duplicate logs.

PR #1469, opened 2019-01-09
Consolidate mocha configurations into config file.

yifanyang:


config/mocha.browser.opts
```diff
@@ -0,0 +1,6 @@
+# See:
+#   - https://mochajs.org/#usage for more information on usage of mocha flags.
+#   - https://github.com/karma-runner/karma-mocha for more information on all mocha flags which the
+#     karma runner supports.
+--timeout 20000
+--retry 3
```
yifanyang: These are coming from mocha.base.js, which gets deleted in this pr.

yifanyang:
@Feiyang1 Can you please invite other folks who you think should also weigh in to review this pr?

mikelehen:
I'd be fine with trying a single config for node/browser and see if we run into issues.

Just to be clear, before this PR we didn't have a shared config for node, right (mocha.base.js was only being used for browser) and so we were just getting whatever the mocha defaults were (except for the packages that are explicitly passing timeouts on the command-line)?  On that note, I think we should remove all of the timeouts / retries being passed on the command-line.  I think those were only to override the mocha default, since we didn't have a shared mocha config for node.

yifanyang:
Yes. mocha.base.js was only meant for browser tests. For node tests that don't set timeouts or retries on the command line, they will get mocha default.

Even with the config file, the test itself can still override it by setting options on the cli explicitly.

Feiyang1:
I'm fine with sharing the same config for browser and nodejs. To Michael's point, let's move all the config options into the mocha config file.

yifanyang:
I have taken a closer look. There are some test options (e.g. `--require ts-node/register`, `--exit`) that node tests need but are not applicable to browser tests. If node and browser tests are to use the same mocha config file, then the node specific options will have to stay on command line.

So in our case, we probably will have to keep two mocha config files -- one for node and the other for browser. Also, this way we could use different test options for node and browser, which is also the way we have been doing since the beginning.

Feiyang1:
Okay. I'm okay with 2 config files since node tests and browser tests need different parameters. Thanks.

PR #1476, opened 2019-01-11
Port "Stop persisting last acknowledged batch ID" from iOS

var-const:
A straightforward port of https://github.com/firebase/firebase-ios-sdk/pull/2243.

### DIFF
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -305,6 +305,8 @@ export class DbMutationQueue {
      * by the server. All MutationBatches in this queue with batchIds less
      * than or equal to this value are considered to have been acknowledged by
      * the server.
+     *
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -305,6 +305,8 @@ export class DbMutationQueue {
      * by the server. All MutationBatches in this queue with batchIds less
      * than or equal to this value are considered to have been acknowledged by
      * the server.
+     *
+     * NOTE: this is deprecated and no longer used by the code.
```
var-const: I'm not sure if I can remove it from here, given that we can't remove `lastAcknowledgedBatchId` from the database schema?

mikelehen: Yeah, if for no other reason I think we need to keep it because it's used by the migration that deletes all of the acknowledged mutations.

Issue #952, opened 2018-06-21
Auth-types and storage-types should depend on util


mmermerkaya:
Both of these packages depend on `@firebase/util` but don't have this dependency listed in their `package.json` files.

https://github.com/firebase/firebase-js-sdk/blob/3c064a3a60fa4eb558012839cf9a77913996876a/packages/auth-types/index.d.ts#L18

https://github.com/firebase/firebase-js-sdk/blob/3c064a3a60fa4eb558012839cf9a77913996876a/packages/storage-types/index.d.ts#L18

This causes issues like #900.

hsubox76:
I've been looking into this and am not 100% sure but I think that since users npm install firebase and not the individual packages, and the firebase/app package (and others) include `@firebase/util` as a dependency, it will always be available for auth-types and storage-types.

I think what may have possibly happened in issue #900 is that the fix (changing 2 arguments to 1) was part of a major version bump for `@firebase/util` from 1.2.0 to 2.0.0?  And the user may have just run npm update, which doesn't update major versions.  It seems like a full reinstall as in issue #787 fixed problems like this for other people by forcing install of 2.0.0?

In principle it seems like it would make sense to add @firebase/util as a dependency but I've noticed none of the other "-types" packages have any actual "dependencies" in `package.json`, just peerDependencies (specifically `@firebase/app-types`, which they all import in the same way as `@firebase/util`) and devDependencies (on typescript), and I'm not sure if there's a reason for that.

For safety I could maybe add `@firebase/util` to the peerDependencies where `@firebase/app-types` already is.  This won't automatically install it for users but it will give a warning if it's behind and let them know to do the major update.

hsubox76:
Actually I can't reproduce how upgrading the `firebase` package would fail to upgrade `@firebase/util` so I guess the above guess is not correct.

Looks like every other "*-types" package that imports `@firebase/util` does not specify it as a dependency in `package.json`, but all the others (app, database, messaging, testing) do specify it as a dependency in the main package (e.g., "database-types" doesn't but "database" does).  Since the *-types can't be installed independently, they'll always have it if the main package does.

The main `auth` and `storage` packages don't depend on it, only their corresponding *-types do, so I guess it makes sense to put it in `auth-types` and `storage-types` package.json dependencies.

mmermerkaya:
Sounds good, thanks :slightly_smiling_face:

PR #1479, opened 2019-01-16
 Add @firebase/util dependencies to auth-types and storage-types

hsubox76:
Adding @firebase/util as an explicit dependency for `auth-types` and `storage-types` as unlike other packages, it is not guaranteed to come in through the corresponding main package, since they don't use it.

See #952 for details.


bojeil-google:
Approval for Auth side changes.

packages/testing/src/api/index.ts
```diff
@@ -131,7 +131,7 @@ function initializeApp(
   if (projectId) {
     app.firestore().settings({
       host: FIRESTORE_ADDRESS,
-      ssl: false,
+      ssl: false
```
ryanpbrewster: This change seems unrelated. I think that either we accidentally submitted some unformatted code, or you're on a slightly different version of the formatter (I've definitely had the latter happen to me on occasion). Would you mind manually limiting this PR to the dependency updates?

hsubox76: So I did a check and I am running the correct version of prettier, and its default behavior is to remove trailing commas, and we don't override that in the config, so it should be doing this.  It's also done the same to two firestore files modified 6 days ago that I think also didn't get a prettier pass on prepush.

I have a PR out to have prettier only run on files changed in the given PR (https://github.com/firebase/firebase-js-sdk/pull/1480) and then I guess the plan is that Travis will maybe run it on everything later to catch any stray files? But in the meantime it gets run on the whole repo on every push so this is where the stray files get caught.

ryanpbrewster: I'll defer to @Feiyang1 on this one.

PR #1483, opened 2019-01-18
update grpc to 1.18.0

Feiyang1:
https://github.com/grpc/grpc-node/releases/tag/grpc%401.18.0
With Electron 4 prebuilt binaries.


ryanpbrewster:
Once this goes in we can start removing the `'grpc.max_reconnect_backoff_ms': 100` options :)

PR #1502, opened 2019-01-29
Firestore Console Build

Feiyang1:
Reattempt at creating the build target for console.
The original PR: https://github.com/firebase/firebase-js-sdk/pull/374

### DIFF
--- a/packages/firestore/index.console.ts
+++ b/packages/firestore/index.console.ts
@@ -29,3 +29,4 @@ export { PublicBlob as Blob } from './src/api/blob';
 export { FirstPartyCredentialsSettings } from './src/api/credentials';
 export { PublicFieldValue as FieldValue } from './src/api/field_value';
 export { FieldPath } from './src/api/field_path';
+export { Timestamp } from './src/api/timestamp';
--- a/packages/firestore/index.console.ts
+++ b/packages/firestore/index.console.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018 Google Inc.
+ * Copyright 2019 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/firestore/index.console.ts
+++ b/packages/firestore/index.console.ts
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2018 Google Inc.
+ *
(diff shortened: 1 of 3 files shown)


packages/firestore/tools/console.build.js
```diff
@@ -0,0 +1,73 @@
...
+`;
+
+async function build() {
+  // create a bundle
+  const bundle = await rollup.rollup(inputOptions);
+
+  // generate code
+  const { code } = await bundle.generate(outputOptions);
+
+  const output = `${PREFIX}${JSON.stringify(String(code))}${POSTFIX}`;
```
tstirrat:
```js
JSON.stringify(String('a'));

// "\"a\""
```

I think that will output the code string wrapped.

```js
__firestore_exports__.firestore = "\"function(iife){}\"";
```

instead we want raw code:

```js
__firestore_exports__.firestore = function(iife){};
```

Feiyang1: They are wrapped in eval, so the extra quotes are needed to concat statements as strings.

packages/firestore/tools/console.build.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * Copyright 2018 Google Inc.
```
mikelehen: 2019

packages/firestore/tools/console.build.js
tstirrat: An update on this.. found a simpler wrap:

```js
goog.module('firestore');
var firestore = function(e){}();  // note, not stringified, just pure rollup code
exports = firestore;
```

packages/firestore/index.console.ts
```diff
@@ -0,0 +1,31 @@
...
+export {
+  Firestore,
+  FirestoreDatabase,
+  PublicCollectionReference as CollectionReference,
+  PublicDocumentReference as DocumentReference,
+  PublicDocumentSnapshot as DocumentSnapshot,
+  PublicQuerySnapshot as QuerySnapshot
+} from './src/api/database';
+export { GeoPoint } from './src/api/geo_point';
+export { PublicBlob as Blob } from './src/api/blob';
```
tstirrat: We'll also need Timestamp exported

Feiyang1: Done.

PR #1507, opened 2019-01-30
Add @license tags

hsubox76:
Adding @license tags to all existing license comments and configuring rollup to add license comments to generated code in the same format as Webpack was previously doing.  This is to conform to google3 third_party rules about licenses in code.

### DIFF
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -1,4 +1,5 @@
 /**
+ * @license
  * Copyright 2018 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -1,4 +1,5 @@
 /**
+ * @license
  * Copyright 2018 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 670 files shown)


packages/firebase/rollup.config.js
```diff
@@ -47,6 +50,12 @@ const plugins = [
   commonjs()
 ];
 
+const licenseOptions = {
+  banner: `@license Firebase v${pkg.version}
+    Build: rev-${gitRev.short()}
+    Terms: https://firebase.google.com/terms/`
```
mikelehen: So `@license` is supposed to designate the software license that applies to the code, right?  So what does `@license Firebase vX.Y.Z` mean?  The Terms URL doesn't really mention software licenses so I'm not sure what this header is communicating about our license.

Also, I believe in the distant past there was discussion about whether we needed to include license headers in our generated build products and the decision was that we did not, since these are closer to binaries than to sources (even though one could argue that any JavaScript code is "source").  So we omitted it since folks are pretty size-sensitive and the full header is 500+ bytes.  That said I don't feel too strongly about this.  So if we think the header is supposed to be here, fine.

hsubox76: The Third Party policy is pretty explicit about wanting it in the compiled code.  It's possible we could make a second Rollup config just for generating code to put in third_party.

I'm not sure about the unusual format for the compiled code headers, I copied it from the Webpack config that previously added the same format of header to our generated files before we switched to Rollup.  The Third Party policy doc seems to indicate we should be putting the entire license text in so I'm not sure if something changed or if the requirement has always been very loose.

mikelehen:
Oh, I think you also need to re-run `yarn` and check in the updated yarn.lock file, since you changed some package.json dependencies.

Feiyang1:
I just realized we should update license.js as well, so the annotation is added for new files.

hsubox76:
Now the plan is to add license headers to the generated text only for internal deployment, so most of the code for that workflow will be in a script living in the third_party dir.  The only part that remains in this repo is an alternate rollup config file, because it depends so much on the original.

PR #1508, opened 2019-01-31
Prevent skipped packages from being published

Feiyang1:

### DIFF
--- a/scripts/release/cli.js
+++ b/scripts/release/cli.js
@@ -144,7 +144,7 @@ const { argv } = require('yargs');
           delete versions[name];
 
           // remove skipped package from updates array
-          updates = updates.filter(pkgName => pkgName !== name );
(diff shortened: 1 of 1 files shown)


scripts/release/cli.js
```diff
@@ -136,13 +136,15 @@ const { argv } = require('yargs');
...
-
       /**
        * Remove the skipped packages
        */
       Object.keys(versions).forEach(name => {
         if (versions[name] === 'skip') {
           delete versions[name];
+
+          // remove skipped package from updates array
+          updates = updates.filter(pkgName => pkgName !== name);
```
hiranya911: `updates` gets overwritten for each skipped package. Is that what you want?

Feiyang1: yes, it's fine to overwrite it. It's not super efficient, but we don't need to be and it's much less/readable code.

PR #1511, opened 2019-02-01
Fix flaky test.

mikelehen:
The "can return previous value through consecutive updates" test was flaky
because it did two consecutive server timestamp writes to the same field and
expected 2 local events. But if you do both writes within 1ms then the
snapshots will be indistinguishable (the estimated value for the server
timestamps will be the same) and so only 1 event will be raised. I added an
additional change to the second write to guarantee it triggers an event.

While looking at the test, I also noticed it was slightly broken / incomplete.
There was an incorrect line attempting to disable the network and the test was
missing the remaining code to re-enable the network and verify the behavior.  I
fixed that as well.

### DIFF
--- a/packages/firestore/test/integration/api/server_timestamp.test.ts
+++ b/packages/firestore/test/integration/api/server_timestamp.test.ts
@@ -245,11 +245,12 @@ apiDescribe('Server Timestamps', persistence => {
   it('can return previous value through consecutive updates', () => {
     return withTestSetup(() => {
       return writeInitialData()
-        .then(() => docRef.firestore.disableNetwork)
+        .then(() => docRef.firestore.disableNetwork())
         .then(() => {
           // We set up two consecutive writes with server timestamps.
           docRef.update('a', FieldValue.serverTimestamp());
-          docRef.update('a', FieldValue.serverTimestamp());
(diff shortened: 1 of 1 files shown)


packages/firestore/test/integration/api/server_timestamp.test.ts
```diff
@@ -245,11 +245,12 @@ apiDescribe('Server Timestamps', persistence => {
...
     return withTestSetup(() => {
       return writeInitialData()
-        .then(() => docRef.firestore.disableNetwork)
+        .then(() => docRef.firestore.disableNetwork())
         .then(() => {
           // We set up two consecutive writes with server timestamps.
           docRef.update('a', FieldValue.serverTimestamp());
-          docRef.update('a', FieldValue.serverTimestamp());
+          // include b=1 to ensure there's a change resulting in a new snapshot.
+          docRef.update('a', FieldValue.serverTimestamp(), 'b', 1);
```
mikelehen: This is the actual fix. The rest is just making the test actually disable/re-enable the network and match the other ports.

mikelehen:
Er, actually I can fix this differently so that it both fixes the flakiness and is closer to the Android / iOS ports.  One sec.

Issue #1516, opened 2019-02-03
As ES6 module not working


pshihn:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS
  * Browser version: Chrome 71
  * Firebase SDK version: 5.8.2
  * Firebase Product:app

### [REQUIRED] Describe the problem

When importing firebase app as an es module, it fails because one of the dependencies (find.js)
uses `require`

#### Steps to reproduce:

Create an HTML file with following and see the console. 
```html
<body>
  <script type="module" src="./node_modules/firebase/app/dist/index.esm.js"></script>
</body>
```

It also happens if you are simply importing in any other module
```javascript
import * as firebase from 'firebase';
```

hsubox76:
I can't seem to reproduce this at the moment using either import method.  Can you paste in the console error message you are seeing?  Also, does the error go away if you roll back to a previous version of firebase?

pshihn:
Here's the error:
```
find.js:1 Uncaught ReferenceError: require is not defined
    at find.js:1
```

Tested with 5.7 and it *does not* happen

To reproduce this, I'd try in a clean workspace (node_modules) and installing v5.8.2

pshihn:
I quickly checked a few versions. 5.7.2 is the latest version that works. 
Since the code is in a dependency, I suspect the later version of a dependency does not provide valid esm

hsubox76:
I was able to reproduce using Webpack, and checking commits between 5.7.2 and 5.7.3 it seems like https://github.com/firebase/firebase-js-sdk/pull/699 might possibly be the cause because it deals with bundling and commonjs compatibility, specifically the removal of the plugins from @firebase/polyfill's rollup config: https://github.com/firebase/firebase-js-sdk/pull/699/files#diff-b63d69dff0c673fd93c711e0772b361a

The dependency in question is core-js (required by @firebase/polyfill) and the version of that package has not been changed since it was first installed, so it's not the dependency itself.

I'll try to see if reverting some of the code from that commit fixes this problem.

mmermerkaya:
My understanding is that the purpose of the `module` field (at least for now) isn't to point to a build that can be used natively in a browser. The `module` field is used by module bundlers, and [enables them to do their magic](https://github.com/rollup/rollup/wiki/ES6-modules).

[...]

PR #1521, opened 2019-02-06
Set up separate esm build for polyfill

hsubox76:
The `polyfill` package has a dependency on the `core-js` library, which is not an es module.  To create a build consumable as esm, we need to bundle the `core-js` code converted to esm format.  This seems to be the cause of issue https://github.com/firebase/firebase-js-sdk/issues/1516 and was caused during the removal of (mostly) unnecessary dependencies in PR https://github.com/firebase/firebase-js-sdk/pull/699.

### DIFF
--- a/packages/polyfill/package.json
+++ b/packages/polyfill/package.json
@@ -22,6 +22,8 @@
   },
   "devDependencies": {
     "rollup": "0.57.1",
+    "rollup-plugin-commonjs": "9.1.0",	
+    "rollup-plugin-node-resolve": "3.3.0",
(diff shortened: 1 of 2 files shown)


Feiyang1:
Now I think about it, it's probably not the best approach to solve the problem. It should be solved by the downstream dependencies.

In this case where the problem is in the polyfill libraries, it makes little sense to include them if a user is already loading modules natively. 

Instead we can suggest them to load scoped packages where polyfills are not included.

hsubox76:
Abandoning PR in favor of advising developers to bundle their own polyfills as needed.

PR #1529, opened 2019-02-11
Forbid queries endAt an uncommitted server timestamp.

rsgowman:
Without this, it still fails, but:
a) not until serializing the query, and
b) the error is an internal error, and
c) the error message is quite cryptic and has nothing to do with the problem.

Port of https://github.com/firebase/firebase-android-sdk/pull/138

### DIFF
--- a/packages/firestore/test/integration/api/validation.test.ts
+++ b/packages/firestore/test/integration/api/validation.test.ts
@@ -19,6 +19,7 @@ import * as firestore from '@firebase/firestore-types';
 import { expect } from 'chai';
 
 import { CACHE_SIZE_UNLIMITED } from '../../../src/api/database';
+import { EventsAccumulator } from '../util/events_accumulator';
 import firebase from '../util/firebase_export';
 import {
   ALT_PROJECT_ID,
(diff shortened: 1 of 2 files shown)


packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -795,6 +796,64 @@ apiDescribe('Validation:', persistence => {
...
+          /*docs=*/ {},
+          async (collection: firestore.CollectionReference) => {
+            await db.disableNetwork();
+
+            const offlineAccumulator = new EventsAccumulator<
+              firestore.QuerySnapshot
+            >();
+            const onlineAccumulator = new EventsAccumulator<
+              firestore.QuerySnapshot
+            >();
```
mikelehen: This is a slightly unconventional use of EventsAccumulator.  Normally we unconditionally accumulate events without any validation (usually by making it the direct receiver of the snapshot events) and then we call awaitEvent() to get the events out and then validate them.

I think what you're looking for is probably a Deferred. Take a look at https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/integration/api/database.test.ts#L700 for an example.

rsgowman: Yeah, that's better; done.

PR #1530, opened 2019-02-11
Fix bugs in internal rollup config

hsubox76:
Somehow did not catch some minor syntax mistakes in last commit.  This fix has been tested in the actual third_party environment by pulling this branch.

### DIFF
--- a/packages/firebase/rollup-internal.config.js
+++ b/packages/firebase/rollup-internal.config.js
@@ -16,10 +16,11 @@
  */
 
 /**
- * Config for internal deployment, adds required license header to generated code.
+ * Specialized config only for internal deployment to google3 repo, adds required license header to
+ * generated code.
  */
 
-// This will only be used in a script that first renames rollup.config.js to rollup-main.config.js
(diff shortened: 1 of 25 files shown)


packages/firebase/rollup-internal.config.js
```diff
@@ -19,18 +19,22 @@
  * Config for internal deployment, adds required license header to generated code.
  */
 
-import baseBuilds from './rollup.config.js';
+// This will only be used in a script that first renames rollup.config.js to rollup-main.config.js
+import baseBuilds from './rollup-main.config.js';
```
Feiyang1: Can you please add some comment for what this file is for? Something like 'It is only used in google3 to build bundles with license header '.

Does rollup not keep the license in the bundle ?

hsubox76: Added more comments.  Rollup (or more precisely, the uglify plugin) doesn't automatically preserve the license in generated files unless you add a setting in config, which seems to be the same in webpack.

Issue #1415, opened 2018-12-05
Improve handling of es-modules for browser/node


lindner:
### Describe your environment

  * Operating System version: MacOS Mojave
  * Browser version: N/A
  * Firebase SDK version: 5.6.0 (via npm)
  * Firebase Product: database

### Describe the problem

- Firebase should detect when a browser based bundle is running in a node environment and warn the user.
- Firebase should either provide esnext bundles or document module loading in differing environments.

#### Steps to reproduce:

Currently Firebase provides a 'main' bundle that is commonjs for Node.  A 'module' bundle that is browser based and a 'browser' bundle that is also browser based.  Here are the entries from package.json:

```
  "browser": "dist/index.cjs.js",
  "main": "dist/index.node.cjs.js",
  "module": "dist/index.esm.js",
```

When using a stock rollup config it will prefer the 'module' bundle.  If the circumstances are right (websocket shim?) it will work.  But often it will fail.  If you enable tracing using a call like this:

```
   firebase.database.enableLogging(true);
```

you will find error messages that look like this:

```
[2018-12-05T14:05:51.668Z]  @firebase/database: p:0: Making a connection attempt
[2018-12-05T14:05:51.690Z]  @firebase/database: getToken() completed. Creating connection.
[2018-12-05T14:05:51.690Z]  @firebase/database: c:0:2: Connection created
[2018-12-05T14:05:51.690Z]  @firebase/database: p:0: Failed to get token: Error: No transports available
[2018-12-05T14:05:51.690Z]  @firebase/database: p:0: data client disconnected
```

To fix this using Rollup you will need to disable `module` resolution for the node-resolve plugin.

```
  import resolve from 'rollup-plugin-node-resolve';
  ....
      resolve({jsnext: true, module: false, modulesOnly: true}),
  ....
```

Not sure of the best solution to this, but a couple come to mind:

- Use jsnext:main for an es6 modules node-oriented artifact.
- Switch module to use a Node version, and use jsnext:browser for the es6 browser artifact.
- Browser code could detect that it is running in the node environment and complain loudly.

Feiyang1:
Thanks for the issue! I think jsnext:main has been deprecated and is equivalent to module. Switching module to use a node version will cause problem for browser use cases (e.g. including unused polyfills), so there is no perfect solution here. 

I think we can do a better job on documenting the build process for the different build targets, and examples with the most popular bundlers. I will keep the issue open and report back once we have an update.

[...]

PR #1536, opened 2019-02-13
Add a console warning when browser bundle is loaded in node.

hsubox76:
Partially addresses issue https://github.com/firebase/firebase-js-sdk/issues/1415.

Open to suggestions on what the message should recommend to the user.

Tried a rollup node test project and browser webpack test project.  Warning is seen in node, not seen in browser.

### DIFF
--- a/packages/app/index.ts
+++ b/packages/app/index.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import { FirebaseNamespace } from '@firebase/app-types';
 import { createFirebaseNamespace } from './src/firebaseApp';
 
 // Node detection logic from: https://github.com/iliakan/detect-node/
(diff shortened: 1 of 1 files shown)


packages/app/index.ts
```diff
@@ -18,6 +18,23 @@
...
+let isNode = false;
+try {
+  isNode =
+    Object.prototype.toString.call(global.process) === '[object process]';
+} catch (e) {}
+
+isNode &&
+  console.warn(`
+Warning: This is a browser-targeted Firebase bundle but it appears it is being run in a Node
+environment.  This may be because you are using ES module resolution and the Node bundle is
```
Feiyang1: It's not clear to me what ES module resolution is. It's probably better to just state please use the "main" field in package.json.
We can also add the link to webpack doc - https://webpack.js.org/configuration/resolve/#resolvemainfields
and rollup doc - https://github.com/rollup/rollup-plugin-node-resolve

hsubox76: Modified the text.

arpowers:
The solution you guys are proposing is not going to work. Also you give no ways of debugging this issue since Webpack is compiling normally with the 'node' target set.

arpowers:
is says clearly here: 
https://webpack.js.org/configuration/resolve/#resolvemainfields

where you link that a node targeted environment will load 'module' first, YET, you've added this warning to the index.esm.js entry (you also list module: index.esm.js)

Feiyang1:
Thank you for the feedback. And sorry you had a bad experience.

To clarify how fields are used in firebase currently:
main: cjs build for Node
module: esm build for browser
browser: cjs build for browser
react-native: cjs build for react-native

Each firebase SDK (except storage) does something different for Node, so the browser version should really not be used for Node environment. and it could cause issues down the road if you are already using it, so I think the warning message is warranted.

We have both cjs and esm build for browser is because we'd like to support as many bundlers as possible, for example, browserify only works with cjs natively.

I'd like to have a `browser-module` field, so I can use `module` field to point to node version, but I don't think there is something like this at the moment, so we chose to use the module field for the browser build.

I'm open to suggestions, but given the constraint of a single module field, I don't know what more we can do than giving a warning message when a bundle is used in the unintended environment.

I'm not convinced that changing the global `resolve.mainFields` to read `main` field first is going to cause issue for other packages as all node packages should have a cjs build pointed by the `main` field.

If you are really worried about it, you could do something like: 
```javascript
  resolve: {
    alias: {
      firebase$: path.resolve(__dirname, 'node_modules/firebase/dist/index.node.cjs.js')
    }
  }
```

Anyway, thanks for voicing your concern. If you'd like to continue the discussion, please open an issue, so we can track it.

PR #1540, opened 2019-02-15
 Replace database tests that hit production with emulator test.

yifanyang:


scripts/emulator-testing/database-test-runner.ts
```diff
@@ -29,7 +29,7 @@ function runTest(port: number, namespace: string): ChildProcessPromise {
     }),
     stdio: 'inherit'
   };
-  return spawn('yarn', ['test'], options);
+  return spawn('npx', ['run-p', 'test:browser', 'test:node'], options);
```
Feiyang1: How about spawn('yarn', ['test:all'], options);

packages/database/package.json
```diff
@@ -12,7 +12,8 @@
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
-    "test": "run-p test:browser test:node",
+    "test": "yarn --cwd ../.. test:database:emulator",
```
Feiyang1: The indirection feels a little unnatural to me. I was confused for a moment where test:database:emulator is.
Can you move the test:database:emulator to the database package?

scripts/emulator-testing/database-test-runner.ts
yifanyang: Done.

packages/database/package.json
yifanyang: Yes, this makes more sense. Done.

scripts/emulator-testing/emulators/emulator.ts
```diff
@@ -80,11 +80,12 @@ export abstract class Emulator {
       this.emulator = promise.childProcess;
 
       console.log(`Waiting for emulator to start up ...`);
-      const timeout = 10; // seconds
+      const timeout = 60; // seconds
```
mikelehen: Is 10 seconds not enough?  That's too bad.  60 seconds seems excessive.  Could we merely bump to 20 or 30 initially and see if that's good enough?

yifanyang: I have actually tried 30s, unfortunately it's still not sufficient.

My observations on emulator startup time (from ~10 manual run for each case):

```
+----------------------------------------+--------------+------------------+-------------+
|                                        | MacBook Pro  | Corp Workstation |  Travis VM  |
+----------------------------------------------------------------------------------------+
| run database emulator test only:       |   2s         |     2s           |   2s        |
| `yarn test` under packages/database    |              |                  |             |
+----------------------------------------------------------------------------------------+
| run test from all packages:            |   16 ~ 17s   |     6 ~ 8s       |   49 ~ 53s  |
| `yarn test` under project root         |              |                  |             |
+----------------------------------------+--------------+------------------+-------------+
P.S. These numbers are pretty consistent though.
```

Our tests (from 20+ packages) runs in parallel, so I guess it makes sense that emulator takes longer to start up in this case. And it seems like we have really exhausted Travis VM's capability.

mikelehen: Oh, nice!  Thanks for gathering this data.  Can you add a comment something like:

```
// NOTE: On a normal workstation, the emulator should start much faster than 60s, but travis launches tests from 20+ packages in parallel and it can take 50+ seconds for the emulator to start.
```

FWIW if you've seen it take 53s, then I'm a little worried 60 won't be enough.  I usually set timeouts to at least double the expected time.  But your call.  We can always start at 60 and increase if it's an issue.

yifanyang: Done. Changed the timeout to `30s` in normal cases and `100s` when on Travis. Also added some comments for why we are setting it this way.

PR #1545, opened 2019-02-16
Made rulesets that don't compile throw errors in testing sdk

scottcrossen:
Case: User uses loadDatabaseRules method to upload broken rules.
Behavior before: Fail, but present success from SDK method.
Behavior now: Rejected promise from SDK method with compilation error message (i.e. line number)

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -147,6 +147,9 @@ export type LoadDatabaseRulesOptions = {
   databaseName: string;
   rules: string;
 };
+type LoadDatabaseRulesErrorBody = {
+    error?: string;
+}
 export function loadDatabaseRules(
   options: LoadDatabaseRulesOptions
 ): Promise<void> {
@@ -168,8 +171,16 @@ export function loadDatabaseRules(
         body: options.rules
       },
       (err, resp, body) => {
+        let parsedBody: LoadDatabaseRulesErrorBody
+        try {
+          parsedBody = JSON.parse(body)
+        } catch(e) {
+          parsedBody = {}
+        }
         if (err) {
           reject(err);
+        } else if (parsedBody.error) {
+          reject(parsedBody.error)
         } else {
           resolve();
         }
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -171,15 +171,11 @@ export function loadDatabaseRules(
         body: options.rules
       },
       (err, resp, body) => {
-        let parsedBody: LoadDatabaseRulesErrorBody;
-        try {
-          parsedBody = JSON.parse(body);
-        } catch (e) {
-          parsedBody = {};
-        }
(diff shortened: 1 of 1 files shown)


packages/testing/src/api/index.ts
```diff
@@ -168,8 +171,16 @@ export function loadDatabaseRules(
         body: options.rules
       },
       (err, resp, body) => {
+        let parsedBody: LoadDatabaseRulesErrorBody
+        try {
+          parsedBody = JSON.parse(body)
```
ryanpbrewster: I would consider the case where the body is invalid JSON to be an error (and an exceptional one, the emulator should always return valid JSON). Can we drop this `try` and blow up if that happens?

packages/testing/src/api/index.ts
```diff
@@ -168,8 +171,16 @@ export function loadDatabaseRules(
         body: options.rules
       },
       (err, resp, body) => {
+        let parsedBody: LoadDatabaseRulesErrorBody
```
ryanpbrewster: Can we check if `resp.statusCode === 200` first? If the rules don't compile we return a 400. I like that we parse the response and emit the error, though, let's definitely keep that.

packages/testing/src/api/index.ts
scottcrossen: Does the emulator return a body if the methods succeeds? We don't want the method to explode on a valid response.
If you don't know then I'll test further and implement accordingly on tuesday.

packages/testing/src/api/index.ts
scottcrossen: okay that makes a lot of sense! Will change on Tuesday.

packages/testing/src/api/index.ts
ryanpbrewster: It does, the body will look like `{"status":"ok"}`

scottcrossen: okay! will change on tuesday.

scottcrossen:
Tested the truthy case and error case.

packages/testing/src/api/index.ts
```diff
@@ -168,8 +171,12 @@ export function loadDatabaseRules(
         body: options.rules
       },
       (err, resp, body) => {
+        // This should always parse (so don't purposefully catch errors).
+        let parsedBody: LoadDatabaseRulesErrorBody = JSON.parse(body);
```
ryanpbrewster: This can be moved into the `if` block

packages/testing/src/api/index.ts
```diff
@@ -147,6 +147,9 @@ export type LoadDatabaseRulesOptions = {
   databaseName: string;
   rules: string;
 };
+type LoadDatabaseRulesErrorBody = {
+  error?: string;
+};
```
ryanpbrewster: not sure this type is doing much for us,

PR #1549, opened 2019-02-20
Update dependencies and prepare for enabling auto-update dependencies

Feiyang1:
A lot of our dependencies are out of date. It's very hard to keep up with the version updates manually, so we need a way to automate the process.

This PR updates the dependencies and fixes issues, so when we enable the bot to automate it, it will not be this massive and disruptive.

The most code changes are triggered by Typescript 3.x upgrade which provides new types and more accurate typings for native web APIs. As a result, it exposes some cases we don't consider before (e.g. we don't null check properties in some cases because the old typing says it's not nullable, but it is in the updated typing). I did my best effort to understand the code and believe we don't need to make any change. Please let me know if it's not the case, I will revise accordingly.


packages/database/src/core/util/Tree.ts
```diff
@@ -155,7 +155,7 @@ export class Tree<T> {
    * @return {boolean} true if the action callback returned true.
    */
   forEachAncestor(
-    action: (tree: Tree<T>) => void,
+    action: (tree: Tree<T>) => any,
```
Feiyang1: since return value of action is used in a if statement, it needs to return a type. Should it be boolean?

packages/database/test/helpers/util.ts
```diff
@@ -48,14 +48,14 @@ export const DATABASE_URL = USE_EMULATOR
...
 
-const qs = {};
-if ('location' in this) {
-  const search = (this.location.search.substr(1) || '').split('&');
-  for (let i = 0; i < search.length; ++i) {
-    const parts = search[i].split('=');
-    qs[parts[0]] = parts[1] || true; // support for foo=
-  }
-}
+// const qs = {};
```
Feiyang1: Can this block of code be removed? It populates qs with the querystring parameters, but qs is never used.

packages/firestore/src/local/simple_db.ts
```diff
@@ -111,7 +111,7 @@ export class SimpleDb {
         // We are provided a version upgrade transaction from the request, so
         // we wrap that in a SimpleDbTransaction to allow use of our friendlier
         // API for schema migration operations.
-        const txn = new SimpleDbTransaction(request.transaction);
+        const txn = new SimpleDbTransaction(request.transaction!);
```
Feiyang1: transaction is nullable - https://developer.mozilla.org/en-US/docs/Web/API/IDBRequest/transaction
It will always have a value since we use it within onupgradeneeded callback.

packages/firestore/src/util/input_validation.ts
```diff
@@ -330,7 +330,7 @@ export function valueDescription(input: AnyJs): string {
     if (input instanceof Array) {
       return 'an array';
     } else {
-      const customObjectName = tryGetCustomObjectType(input);
+      const customObjectName = tryGetCustomObjectType(input!);
```
Feiyang1: Typescript is not able to narrow it down to exclude null even if we did null check before.

packages/firestore/test/util/test_platform.ts
```diff
@@ -87,7 +87,11 @@ export class FakeWindow {
  * `Document` fake that implements the `visibilitychange` API used by Firestore.
  */
 export class FakeDocument {
-  private _visibilityState: VisibilityState = 'unloaded';
+  /**
+   * unload is no more a valid member of VisibilityState, is it a bug?
+   * https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState
+   */
+  private _visibilityState: VisibilityState = 'unloaded' as VisibilityState;
```
Feiyang1: What should we change the value to?

packages/firestore/.ncurc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "upgrade": true,
+    "reject": [
+        "long"
```
samtstern: It's a shame that these files don't allow for comments ...

mikelehen:
This looks like a large PR with multiple things going on.  Would it be possible to break it into separate smaller PRs?  E.g. a single PR that upgrades TypeScript, a single PR that does all the easy upgrades (don't require any additional changes), a single PR to prepare for auto-updates, etc. (you can decide what makes sense)

If it's possible to split out PRs with changes specific to a single package, that would be nice too, so we don't end up with 14 reviewers for a single PR (in which case it's really hard to know who is responsible for reviewing what).

bojeil-google:
Only reviewed Auth changes.

Feiyang1:
Closing the PR and will open a few smaller PRs as suggested by Michael.

PR #1553, opened 2019-02-21
Add autodocumentation scripts that run Typedoc

hsubox76:
Adds a script that runs Typedoc on `packages/firebase/index.d.ts` to generate our reference documentation.  Documents are written out to a gitignored `/html` dir in the same folder as the script.  The output html is optimally formatted for inserting into the internal devsite and doesn't have its own nav.

`toc.yaml` is used as the source of truth for the table of contents.  This affects the nav structure when added to the devsite, and the generated table of contents links on the JS SDK home page ("Overview").

`HOME.md` is the customized content at the top of the JS SDK home page, above the TOC links.

Also extensively updated `packages/firebase/index.d.ts` to try to match the current state of the API as best as possible.  It will probably need further passes by the tech writing team before we are ready to run this script and publish the results to our docs.


scripts/docgen/generate-docs.js
```diff
@@ -0,0 +1,133 @@
...
+    tocLower = tocRaw.replace(/path:.*/g, x => x.toLowerCase());
+    const { toc } = yaml.safeLoad(tocLower);
+    let tocPageLines = [homeRaw, '# API Reference'];
+    toc.forEach(group => {
+      tocPageLines.push(`\n## [${group.title}](${stripPath(group.path)})`);
+      group.section.forEach(item => {
+        tocPageLines.push(`- [${item.title}](${stripPath(item.path)})`);
+      });
+    });
+    return fs.writeFile(`${__dirname}/HOME_TEMP.md`, tocPageLines.join('\n'));
```
Feiyang1: nit: put HOME_TEMP.md in a variable since it's used multiple times.

scripts/docgen/generate-docs.js
```diff
@@ -0,0 +1,133 @@
...
+Promise.all([
+  fs.readFile(`${__dirname}/toc.yaml`, 'utf8'),
+  fs.readFile(`${__dirname}/HOME.md`, 'utf8')
+])
+  // Read TOC and homepage text and assemble a homepage markdown file.
+  .then(([tocRaw, homeRaw]) => {
+    // In case we need to paste in toc.yaml from old docs.
+    tocLower = tocRaw.replace(/path:.*/g, x => x.toLowerCase());
+    const { toc } = yaml.safeLoad(tocLower);
+    let tocPageLines = [homeRaw, '# API Reference'];
```
Feiyang1: Are these just for display or are clickable to go to the specific page.

I found this block a little hard to understand, though there is a comment. Can we put it (the whole then block to create the homepage) into a function with a self explanatory name? I think It will make the flow easier to follow.

scripts/docgen/toc.yaml
```diff
@@ -0,0 +1,145 @@
+toc:
+- title: "firebase"
+  path: /docs/reference/js/firebase
```
Feiyang1: How is the path determined?

scripts/docgen/generate-docs.js
```diff
@@ -0,0 +1,133 @@
...
+function moveFilesToRoot(subdir) {
+  return exec(`mv ${docPath}/${subdir}/* ${docPath}`)
+    .then(() => {
+      exec(`rmdir ${docPath}/${subdir}`);
+    })
+    .catch(e => console.error(e));
+}
+
+// Reformat links to match flat structure.
+function fixLinks(path) {
```
Feiyang1: I think we need more detailed comments to make the process easier to understand since there is a lot of customizations happening and they are very specific to our needs. 

For example, for each function and step I'd add to the comment why we do it as well.

scripts/docgen/toc.yaml
hsubox76: The directory is the one in google3, every file is at the top level, the file naming scheme is generated by Typedoc based on the hierarchy pulled from index.d.ts.

Feiyang1: Ah I see. I'm wondering if we need to put any google3 related process in google3.

@samtstern Is it an acceptable practice to expose google3 directory structure to the public?

samtstern: Yeah as long as it's not full of codenames or PII (like LDAPs) you don't have to worry about it.

scripts/docgen/generate-docs.js
hsubox76: Done

scripts/docgen/HOME.md
```diff
@@ -0,0 +1,12 @@
+# Firebase JavaScript SDK Reference
+The Firebase JavaScript SDK implements the client-side libraries used by
+applications using Firebase services. This SDK is distributed via:
+
+- CDN (`<script src="https://www.gstatic.com/firebasejs/5.5.3/firebase.js"></script>`)
```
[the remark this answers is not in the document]
hsubox76: Done with help from Eric.

scripts/docgen/generate-docs.js
hsubox76: Created a separate function.

scripts/docgen/generate-docs.js
hsubox76: Broke out a lot of functionality into separate functions, commented each function and its call.

PR #1554, opened 2019-02-22
Add accessToken field to firebase.auth.AuthCrendential typings

GCrispino:
Add accessToken field to firebase.auth.AuthCrendential typings defined at /packages/firebase/index.d.ts

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -294,6 +294,7 @@ declare namespace firebase.auth {
   interface AuthCredential {
     providerId: string;
     signInMethod: string;
+    accessToken: string;
   }
 
   interface AuthProvider {


GCrispino:
I signed it!

bojeil-google:
This is not correct. You need to cast the `AuthCredential` to an `OAuthCredential` which I assume you are getting after OAuth sign-in. The `accessToken` is defined on the latter: https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth-types/index.d.ts#L249

An `AuthCredential` can be a phone, email/password, etc and nothing to do with OAuth.

GCrispino:
Got it then! It just was not clear to me as following the firebase auth's getting started. Maybe adding a TypeScript example would be a nice idea?

GCrispino:
Got it then! It just was not clear to me as following the firebase auth's getting started. 
However, I can't find the `OAuthCredential` type when using the imported firebase's module. The only type that is visible to me is `OAuthProvider`, through the distributed index.d.ts file.

Feiyang1:
@bojeil-google Do we want to expose OAuthCredential to the public?

bojeil-google:
Hey @Feiyang1, `OAuthCredential` should already be public: https://firebase.google.com/docs/reference/js/firebase.auth.OAuthCredential

Feiyang1:
I made a PR https://github.com/firebase/firebase-js-sdk/pull/1559
Please take a look

tevanuwat:
Will my iphone be protected by this Firebase?

Feiyang1:
Closing it since https://github.com/firebase/firebase-js-sdk/pull/1559 addressed the issue.

PR #1556, opened 2019-02-22
Replace AnyJs and Unknown with unknown

Feiyang1:
Replace AnyJs and Unknown with generic unknown introduced in Typescript 3.0


Feiyang1:
@mikelehen I had to change unknown[] to Array<unknown> to get around tslint, because tslint has not updated to recognize unknown as a simple type.

tslint has made the change but not yet released it. Once the change is released, we can change it back to unknown[].

https://github.com/palantir/tslint/pull/4433

PR #1557, opened 2019-02-22
Upgrade closure-builder and optimize build size.

mikelehen:
* Upgraded closure-builder (necessary to be able to specify multiple defines).
* Disabled closure debug code (i.e. in particular logging) and several legacy
  fallbacks that we don't need.


schmidt-sebastian:
Hm, no more ActiveX?

PR #1561, opened 2019-02-25
build: update build tools

Feiyang1:
Update rollup and rollup plugins.


packages/testing/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "rollup -c",
     "dev": "rollup -c -w",
-    "test": "TS_NODE_CACHE=NO nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --opts ../../config/mocha.node.opts",
+    "test": "TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --opts ../../config/mocha.node.opts",
```
ryanpbrewster: should this go in `tsconfig.json` instead?

Feiyang1: rollup requires typescript output to be ES2015 or ESNext, so we have that in the tsconfig.

Instead of having a separate tsconfig just for testing, I'm overwriting the default config to use commonjs because Nodejs only works with it.

PR #1576, opened 2019-03-05
Enable sourcemap generation throughout repo.

mikelehen:
This enables (and improves) sourcemap generation throughout the repo.  My end goal was to be able to use source-map-explorer on any built artifact in order to better understand the bundle size.

NOTES:
1. I reworked the webchannel-wrapper build to actually generate the intermediate output file for our ESM build as it made it more straightforward to generate tho source map.
2. I had to add rollup-plugin-sourcemaps in order to make rollup notice / use existing sourcemaps (when we use compiled code as an input to a second build).
3. I enabled sourcemaps everywhere.  That may not be necessary, but AFAIK it doesn't hurt anything.

### DIFF
--- a/packages/webchannel-wrapper/tools/build.js
+++ b/packages/webchannel-wrapper/tools/build.js
@@ -22,6 +22,7 @@ const sourcemaps = require('rollup-plugin-sourcemaps');
 
 const glob = closureBuilder.globSupport();
 const { resolve } = require('path');
+const { tmpdir } = require('os');
 
 const closureDefines = [
   // Avoid unsafe eval() calls (https://github.com/firebase/firebase-js-sdk/issues/798)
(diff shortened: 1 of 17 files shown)


mikelehen:
FWIW, I didn't add any "analyze" build rules.  I'm happy to just `npx source-map-explorer foo.js` whenever I want to analyze something.  Though perhaps we should find a way to educate folks that this capability exists.

packages/webchannel-wrapper/tools/build.js
```diff
@@ -59,31 +60,29 @@ closureBuilder.build(
...
         language_out: 'ECMASCRIPT5',
         compilation_level: 'ADVANCED',
         define: closureDefines
       }
     }
   },
-  async function(errors, warnings, files, results) {
-    const filePath = resolve(__dirname, '../src/index.js');
+  async function() {
+    const filePath = resolve(__dirname, '../dist/index.closure-es.js');
```
Feiyang1: We probably don't want to publish intermediate files to npm.
We can either change the files array in package.json to include only the files we want to publish( it's the entire dist folder right now), or we can put the intermediate files in a tmp folder and exclude the tmp folder in .gitignore

mikelehen: I used `os.tmpdir()` to put them in a temp dir outside the git repo (so we don't need to add anything to .gitignore)...  PTAL

Feiyang1: Nice!

PR #1582, opened 2019-03-06
Update firestore externs file to avoid closure compiler errors

hsubox76:
Upon importing `firebase-firestore-externs.js` for internal use for the first time, some errors came up when a project attempted to build it using closure compiler, mostly regarding unrecognized jsdoc annotations. Fixed errors brought up by the compiler.

### DIFF
--- a/packages/firebase/externs/firebase-firestore-externs.js
+++ b/packages/firebase/externs/firebase-firestore-externs.js
@@ -474,7 +474,7 @@ firebase.firestore.SetOptions.prototype.merge;
  *
  * @typedef {Array<string>|Array<firebase.firestore.FieldPath>}
  */
-firebase.firestore.SetOptions.prototype.mergeFields
+firebase.firestore.SetOptions.prototype.mergeFields;
 
 /**
  * A `DocumentReference` refers to a document location in a Firestore database
@@ -662,7 +662,7 @@ firebase.firestore.GetOptions = function() {};
  *
  * @typedef {string|undefined}
  */
-firebase.firestore.GetOptions.prototype.source
+firebase.firestore.GetOptions.prototype.source;
 
 /**
  * Metadata about a snapshot, describing the state of the snapshot.
@@ -679,7 +679,7 @@ firebase.firestore.SnapshotMetadata = function() {};
  *
  * @typedef {boolean}
  */
-firebase.firestore.SnapshotMetadata.prototype.fromCache
+firebase.firestore.SnapshotMetadata.prototype.fromCache;
 
 /**
  * True if the snapshot includes local writes (`set()` or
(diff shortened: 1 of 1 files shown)


packages/firebase/externs/firebase-firestore-externs.js
```diff
@@ -712,42 +714,45 @@ firebase.firestore.SnapshotMetadata.prototype.isEqual = function(other) {};
  * access will return 'undefined'. You can use the `exists` property to
  * explicitly verify a document's existence.
  *
- * @interface
+ * @constructor
```
mikelehen: Why did this change?  We don't allow creating DocumentSnapshot objects directly...  and I see we (successfully?) use `@constructor` elsewhere in this file...

mikelehen: woops, I meant we successfully use `@interface` elsewhere in this file.

hsubox76: The closure compiler error was "constructors can only extend constructors" when `QueryDocumentSnapshot` tried to extend this.

hsubox76:
Figuring out a way to generate this from index.d.ts might be a good idea for the future, especially if we are going to be expected to keep the third_party version updated regularly. The new Typedoc autodocumentation tools might help somehow.

Feiyang1:
They are used by 1p apps as dependencies for their compilation.
We have another set of externs in another location in google3 that are used for reference doc generation. Unfortunately it is also used as dependency in some 1p projects. So we can't remove either of them without breaking something.

Two things I think we can do:
1. 1p projects should really only take dependency on firebase externs in third_party/firebase, so we can try to change the 1p projects to depend on the right externs and we can remove the externs in firebase/jscore/api.

2. There is a tool called Tsickle that compiles d.ts file to externs. I have tried it before in the attempt to create externs files and use our existing jsdoc scripts to create reference docs. Not sure if the output externs can be directly used for other projects, but It's worth exploring.

mikelehen:
Cool, thanks!

FWIW- If there are only a few 1p projects using these and they're relatively slow-moving (don't rely on our newest features), it's tempting to push the burden of maintaining externs onto them...  just as they would have to do for any other non-closure dependencies that they're using.

Basically I'd assert that closure usage should be going down over time, and we should focus on supporting TypeScript.  We can't break people, but we might be able to move towards deprecating first-class closure support.

Of course if we can support them for free via tsickle generation or something, then this becomes a non-issue.

packages/firebase/externs/firebase-firestore-externs.js
mikelehen: I see.  Perhaps the better fix is to change QueryDocumentSnapshot to use`@interface` then.  I don't care super strongly about these externs being perfectly correct though. :)

PR #1586, opened 2019-03-07
Fix doc generation script to preserve old filename capitalization

hsubox76:
From the docs side, it's a lot easier on them if we preserve the same capitalization of files.  Typedoc generates lowercase files and it doesn't seem easy to fix on Typedoc's side so the script will now get the correct casing for a filename from the `toc.yaml` entry that matches the generated filename in a case-insensitive comparison, and rename the file to that.

### DIFF
--- a/scripts/docgen/generate-docs.js
+++ b/scripts/docgen/generate-docs.js
@@ -76,7 +76,7 @@ function fixLinks(file) {
       .replace(/\.\.\//g, '')
       .replace(/(modules|interfaces|classes)\//g, '');
     let caseFixedLinks = flattenedLinks;
-    for (let lower in lowerToUpperLookup) {
+    for (const lower in lowerToUpperLookup) {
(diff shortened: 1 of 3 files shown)


scripts/docgen/generate-docs.js
```diff
@@ -72,14 +72,19 @@ function moveFilesToRoot(subdir) {
  */
 function fixLinks(file) {
   return fs.readFile(file, 'utf8').then(data => {
-    const fixedLinks = data
+    const flattenedLinks = data
       .replace(/\.\.\//g, '')
       .replace(/(modules|interfaces|classes)\//g, '');
-    return fs.writeFile(file, fixedLinks);
+    let caseFixedLinks = flattenedLinks;
+    for (let lower in lowerToUpperLookup) {
```
Feiyang1: const lower

hsubox76: Done.

PR #1552, opened 2019-02-21
build: update typescript version

Feiyang1:
Part of https://github.com/firebase/firebase-js-sdk/pull/1549
Update dependencies in preparation to enabling auto-updates.

PR #1589, opened 2019-03-08
update remaining outdated deps before enabling auto update

Feiyang1:
Related PRs: 
https://github.com/firebase/firebase-js-sdk/pull/1564
https://github.com/firebase/firebase-js-sdk/pull/1561
https://github.com/firebase/firebase-js-sdk/pull/1552

@mikelehen the only code change is to replace `Array<unknown>` with unknown[]. I changed any[] to `Array<unknown>` in https://github.com/firebase/firebase-js-sdk/pull/1556


updatedeps.js
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const glob = require('glob');
+const { exec, spawn } = require('child_process');
+
+const { workspaces: rawWorkspaces } = require(`./package.json`);
```
mikelehen: nit: I think this would read better as:
```
const rawWorkspaces = require(`./package.json`).workspaces;
```

updatedeps.js
```diff
@@ -0,0 +1,56 @@
...
+    workspaces.map(
+      workspace =>
+        new Promise(resolve => {
+          glob(workspace, (err, paths) => {
+            if (err) throw err;
+            resolve(paths);
+          });
+        })
+    )
+  ).then(paths => paths.reduce((arr, val) => arr.concat(val), []));
```
mikelehen: Is this just flattening the array?  Could we use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat instead?

updatedeps.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+mapWorkspaceToPackages(workspaces).then(paths => {
+  for (const path of paths) {
+    const handle = spawn('ncu', ['-u', '--packageFile', 'package.json'], {
```
mikelehen: I guess this comes from npm-check-updates?  But I don't see it added to our package.json...  Should it be?

Feiyang1: Sorry, This file shouldn't be included. We will use oss bot to run a similar script.

PR #1591, opened 2019-03-09
Rename prepare to build, so build is not triggered after yarn install

Feiyang1:
`yarn install` takes a long time at the repo level. Part of it is because we automatically trigger a build for all packages after installation. This change uncouples build script from yarn install at repo level.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   ],
   "scripts": {
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
-    "prepare": "lerna run --scope @firebase/* --scope firebase --scope rxfire prepare",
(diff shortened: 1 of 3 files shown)


mikelehen:
I take it this didn't work out? :-(

schmidt-sebastian:
Note that `prepare` also gets run when someone installs the SDK straight from GitHub (via `npm install https://github.com/firebase/firebase-js-sdk`). We probably want the compile/transpile to take place for these installs. Something to keep in mind for the next iteration of this PR :)

mikelehen:
Looks like the next iteration is over in #1592...  

So does that mean our options are either to have "prepare" (and therefore "install" as well) build the world, or else check in built artifacts to the repo?  :-/

PR #1592, opened 2019-03-09
 rename prepare to build, so build is not triggered after yarn install

Feiyang1:
`yarn install` takes a long time at the repo level. Part of it is because we automatically trigger a build for all packages after installation. This change uncouples build script from yarn install at repo level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
 
 before_script:
   - cp config/ci.config.json config/project.json
+  - yarn build
 
 # Misc Addons/Configs
 dist: trusty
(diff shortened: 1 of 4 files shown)


Feiyang1:
@mikelehen @schmidt-sebastian It appears neither npm nor yarn supports installing directly from github from a monorepo.
https://github.com/npm/npm/issues/2974
https://github.com/yarnpkg/yarn/issues/4725

This change only removed prepare script at the repo level, and I don't think it makes sense to install the entire monorepo.
Once the subdir support is added to the tools, our packages are going to work automatically since each package still has the prepare script.

PR #1593, opened 2019-03-10
Stop using deprecated grpc.load, switch to @grpc/proto-loader

ryanpbrewster:
`grpc.load` is deprecated, guidance is to switch to `@grpc/proto-loader` ([link](https://github.com/grpc/grpc-node/blob/38637b51038e3c0a7dae8339c7d5fbeab888adb4/packages/grpc-native-core/index.js#L100))


packages/firestore/src/platform_node/load_protos.ts
```diff
@@ -24,19 +25,13 @@ import { resolve } from 'path';
  * @returns The GrpcObject representing our protos.
  */
 export function loadProtos(): grpc.GrpcObject {
-  const options = {
-    // Beware that converting fields to camel case does not convert the tag
-    // fields in oneof groups (!!!). This will likely be fixed when we upgrade
-    // to protobufjs 6.x
-    convertFieldsToCamelCase: true
```
ryanpbrewster: This PR apparently upgrades us to protobufjs 6.x, and the default is now to convert to camelcase:
```
 * @param options.keepCase Preserve field names. The default is to change them
 *     to camel case.
```

mikelehen: Hrm.  FWIW- The comment seems to be warning us about the fact that tags are still snake case, e.g.: https://github.com/firebase/firebase-js-sdk/blob/9bb3603c2d840986009b4cbb810db5b8a6a0e70a/packages/firestore/src/remote/serializer.ts#L452

But it suggests this would be fixed by upgrading to protobufjs 6.x...  but I guess that didn't come true?

ryanpbrewster:
This approach is incompatible with our existing tests (specifically for the `Value` proto). I have some ideas for how to resolve this (the simplest of which is to open a PR against the @grpc/proto-loader code). I will write up my findings.

For now, though, this seems like a dead end.

Issue #1524, opened 2019-02-08
Persistency issue with unlisten in 5.8.2 or I'm missing something


LanderBeeuwsaert:
### [REQUIRED] Describe your environment

  * Operating System version: windows
  * Browser version: chrome 72
  * Firebase SDK version: 5.8.2
  * Firebase Product: atuh, database

### [REQUIRED] Describe the problem
So I've been on and off enabling persistency. Having for syncing between a lot of clients always had some issues so I disable it unless experimenting.
After upgrade today to 5.8.2 I thought I give it another try.
I now consistently get a:
FIRESTORE (5.8.1) INTERNAL ASSERTION FAILED: Trying to unlisten on query not found.
Because of the VERY HELPFUL Query you now add to the error it was easy to pinpoint the offending subscription.
However repairing it has been not that easy.
I have to unsubscribe/subscribe to the query with different parameters, the first subscribe works but the second one fails already.
I've tried unsubscribing in multiple ways, I normally use the excellenct untilDestroyed(this) library. But I switched to manual unsubscribing (each time I have to subscribe to the query with new parameters), but also that doesn't work.
Disabling persistency solves the issue.

some code:
` 
ngOnChanges(changes: { [propertyName: string]: SimpleChange }) {
    // console.log("changes: ", changes);

    if (changes['selectedOrganisation']) {

      this.loading = true;

      if (this.dataSource) this.dataSource.disconnect();
      if (this.subscription && !this.subscription.closed) this.subscription.unsubscribe();

      if (this.selectedOrganisation) {
        this.subscription = this.OrganisationService.getChildOrganisationsForOrganisationObs(this.selectedOrganisation).subscribe((data: Organisation[]) => { //.pipe(untilDestroyed(this))
          this.dataSource = new MatTableDataSource(data);

          this.dataSource.sort = this.sort;
          this.dataSource.filterPredicate = this.helpers.filterAllStringsPredicate;
          this.loading = false;
        });
      }
    }
  }
`

`
  ngOnDestroy() {
    if (this.subscription && !this.subscription.closed) this.subscription.unsubscribe();
    // if (this.dataSource) this.dataSource.disconnect();
  }`

mikelehen:
Thanks for the report!  By definition, "INTERNAL ASSERTION FAILED" always indicates a bug in the SDK (although occasionally the bug is that we should have a better user-facing error for that circumstance).  So I'd like to get to the bottom of this.

[...]

PR #1596, opened 2019-03-12
Ensure consistent JSON Serialization of DocumentKeys

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-js-sdk/issues/1524

The DocumentKey serialization to canonical Query IDs is different once a Query is round-tripped through IndexedDb. After the round-trip, we include the project and database ID in the JSON serialized canonical ID. Since we use this ID as a Map key in SyncEngine, we need to ensure that these IDs are consistent.

This is technically a minor breaking change since it means that existing Queries that filter by DocumentReferences will no longer be read back from persistence. Instead, we will treat them as new queries and GC the old query.

### DIFF
--- a/packages/firestore/src/model/path.ts
+++ b/packages/firestore/src/model/path.ts
@@ -23,7 +23,7 @@ export const DOCUMENT_KEY_NAME = '__name__';
 /**
  * Path represents an ordered sequence of string segments.
  */
-export abstract class Path {
+abstract class Path {
   private segments: string[];
   private offset: number;
   private len: number;
@@ -31,6 +31,14 @@ export abstract class Path {
   constructor(segments: string[], offset?: number, length?: number) {
     this.init(segments, offset, length);
   }
+
+  /**
+   * Returns a String representation.
+   *
+   * Implementing classes are required to provide deterministic implementations as
+   * the String representation is used to obtain canonical Query IDs.
+   */
+  abstract toString(): string;
 
   /**
    * An initialization method that can be called from outside the constructor.
@@ -71,10 +79,6 @@ export abstract class Path {
 
   get length(): number {
     return this.len;
-  }
-
-  toJSON(): string[] {
-    return this.segments.slice(this.offset, this.limit());
   }
 
   isEqual(other: Path): boolean {
--- a/packages/firestore/src/model/path.ts
+++ b/packages/firestore/src/model/path.ts
@@ -73,6 +73,10 @@ export abstract class Path {
     return this.len;
   }
 
+  toJSON(): string[] {
+    return this.segments.slice(this.offset, this.limit());
+  }
+
   isEqual(other: Path): boolean {
     return Path.comparator(this, other) === 0;
   }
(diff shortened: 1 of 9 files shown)


packages/firestore/src/model/path.ts
```diff
@@ -73,6 +73,10 @@ export abstract class Path {
     return this.len;
   }
 
+  toJSON(): string[] {
+    return this.segments.slice(this.offset, this.limit());
+  }
```
mikelehen: Oh, wow.  I had to set a breakpoint on this code and run your test to understand why this fixes it...  It looks like it's because:
1. RelationFilter.canonicalId() [calls FieldValue.toString()](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/query.ts#L580).
2. ObjectValue.toString() [uses JSON.stringify()](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/model/field_value.ts#L617)

So while your fix works, I'm concerned it's not holistic.  Using .toString() as part of our canonicalId() implementation is perhaps questionable, and using JSON.stringify() is almost certainly problematic, since 1) JSON.stringify() doesn't provide any guarantees as to order, etc. and 2) we have made no effort to implement sane toJSON() methods on any of our types.  The one you fixed here may not be the only issue!

So I think at best, this fix is a workaround.  If we're in a hurry to get this fixed, we could check it in but open a bug to fix this properly.

A better fix would be to port what we do on Android in ObjectValue.toString() which [calls .toString() on the underlying SortedMap](https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/model/value/ObjectValue.java#L85).  This at least guarantees a consistent ordering and makes sure we are relying on the FieldValue.toString() method (as opposed to the largely undefined JSON representation).  Unfortunately it looks like the JS SortedMap doesn't have .toString(), but you could either add it or else just do the traversal in ObjectValue.toString() manually.

Ultimately I think it's dangerous to be using toString() this freely in our construction of canonicalString().  It imposes a non-obvious requirement on our .toString() implementations.  So a complete fix would be to add canonicalString() to FieldValue and use that instead of .toString().  But that'd need to be back-ported to the other platforms, so perhaps a project for another day.

mikelehen: Oh, and note that ArrayValue will need the same fix as it relies on JSON.stringify() too.

mikelehen: By the way, another approach for solving this issue would be to change RelationFilter to generate a simpler but more collision-prone canonicalID. e.g. if the FieldValue is an ObjectValue just use "[object]" as the canonicalId representation.  This will generate collisions, but it avoids any chance for mismatches.  

This probably isn't my top choice though.

schmidt-sebastian: Hm, looks like someone really doesn't like `JSON.stringify()`:) 

What you suggested feels right to me, but some of it does feels a little out of scope for this bug fix. I did drop `JSON.stringify` and added a `toString()` on SortedMap, which is relatively safe and straightforward fix. As you pointed out, that should already a big step in the right direction. I would hope that we can leave the rest as is for now, and rely on the fact that our `toString()` implementations are deterministic. That `ResourcePath.toString()` already works as is a good indication.

PR #1597, opened 2019-03-12
Fix FCM in Chrome Extensions

mmermerkaya:
Prevents sending messages to Chrome Extension background pages, which are always considered visible in Chrome. That causes our SDK try to send the message to the background page and to not show the notification.

I'll also investigate if background page being considered a visible client is a bug in Chrome itself, and try to get it fixed if I can. But for now this seems like a good workaround.

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -333,7 +333,11 @@ export class SwController extends BaseController {
     const clientList = await getClientList();
 
     return clientList.some(
-      (client: WindowClient) => client.visibilityState === 'visible'
+      (client: WindowClient) =>
+        client.visibilityState === 'visible' &&
+        // Ignore chrome-extension clients as that matches the background pages
+        // of extensions, which are always considered visible.
+        !client.url.startsWith('chrome-extension://')
     );
   }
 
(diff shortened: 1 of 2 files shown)


dwoffinden:
Is there a use-case for extensions handling the message themselves in the background page?

Is CI just broken? :P

mmermerkaya:
> extensions handling the message themselves in the background page

The event handler doesn't get triggered when you try to do that. I'm not yet sure why.

I fixed the tests.

evanfrawley:
Hi @mmermerkaya & @dwoffinden!

I have a use case that this change broke. My use case is: Using a background script in a Chrome extension that listens for messages from FCM and when the user is using Chrome in the foreground, I relay the data in a customized way through the web page itself (think similarly to Honey's modifications to an Amazon product detail page). 

Screen Shot 2019-03-29 at 1 42 03 PM

Would it be possible to opt-in allowing Chrome extension background scripts to receive FCM messages? It would be awesome to stay up to date with the newest versions of the JS SDK and not be stuck on the version before this fix forever!

I'd be happy to open a PR for this as well, but it may take some time since I'm unfamiliar with the tests & CI/CD system that is being used.

mmermerkaya:
Hi @evanfrawley, thanks for the report. Could you open an issue with this information? And could you also include a notification payload example please?

PR #1598, opened 2019-03-12
Remove source links from generated reference docs

hsubox76:
Typedoc adds a link to the source code for every symbol documented, and since we are currently using `index.d.ts` as our source, this is not very useful, and the link wrongly points to the local generated files dir anyway.  If we reach our goal of generating docs from source code we can restore this and fix the bad link prefix too.

This file basically overrides the [default Typedoc theme file of the same name](https://github.com/TypeStrong/typedoc-default-themes/blob/f992ea329904db58671a553c30171eb8fdfd3690/src/default/partials/member.sources.hbs) and removes the sources section.

### DIFF
--- a/scripts/docgen/theme/partials/member.sources.hbs
+++ b/scripts/docgen/theme/partials/member.sources.hbs
@@ -1,0 +1,11 @@
+<aside class="tsd-sources">
+    {{#if implementationOf}}
+        <p>Implementation of {{#with implementationOf}}{{> typeAndParent}}{{/with}}</p>
+    {{/if}}
+    {{#if inheritedFrom}}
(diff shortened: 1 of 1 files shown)


scripts/docgen/theme/partials/member.sources.hbs
```diff
@@ -0,0 +1,11 @@
+<aside class="tsd-sources">
```
Feiyang1: Does it work because it overwrites the default behavior and doesn't include link to source?

hsubox76: Yeah, it is identical to the Typedoc template file it's overriding except the source link section is removed.

PR #1602, opened 2019-03-13
Externalize Storage Domain Constant

fredzqm:
Extract a constant `domain` for Storage, so it is easier to switch to a different backend.

### DIFF
--- a/packages/auth/test/utils_test.js
+++ b/packages/auth/test/utils_test.js
@@ -438,11 +438,11 @@ function testMatchDomain_ipAddressDomain() {
 
 function testMatchDomain_caseInsensitiveMatch() {
   assertTrue(fireauth.util.matchDomain('localhost', 'localhost', 'http'));
-  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN_NAME.COM', 'http'));
+  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN.COM', 'http'));
   assertTrue(fireauth.util.matchDomain(
       'doMAin.com', 'abC.domain.COM', 'http'));
   assertTrue(fireauth.util.matchDomain('localhost', 'localhost', 'https'));
-  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN_NAME.COM', 'https'));
+  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN.COM', 'https'));
   assertTrue(fireauth.util.matchDomain(
       'doMAin.com', 'abC.domain.COM', 'https'));
   assertFalse(fireauth.util.matchDomain(
--- a/packages/auth/test/utils_test.js
+++ b/packages/auth/test/utils_test.js
@@ -438,11 +438,11 @@ function testMatchDomain_ipAddressDomain() {
 
 function testMatchDomain_caseInsensitiveMatch() {
   assertTrue(fireauth.util.matchDomain('localhost', 'localhost', 'http'));
-  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN.COM', 'http'));
+  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN_NAME.COM', 'http'));
   assertTrue(fireauth.util.matchDomain(
       'doMAin.com', 'abC.domain.COM', 'http'));
   assertTrue(fireauth.util.matchDomain('localhost', 'localhost', 'https'));
-  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN.COM', 'https'));
+  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN_NAME.COM', 'https'));
   assertTrue(fireauth.util.matchDomain(
       'doMAin.com', 'abC.domain.COM', 'https'));
   assertFalse(fireauth.util.matchDomain(
--- a/packages/storage/src/implementation/authwrapper.ts
+++ b/packages/storage/src/implementation/authwrapper.ts
@@ -82,7 +82,7 @@ export class AuthWrapper {
   private static extractBucket_(config: {
     [prop: string]: any;
   }): string | null {
-    let bucketString = config[constants.configOption] || null;
+    let bucketString = config[constants.CONFIG_STORAGE_BUCKET_KEY] || null;
     if (bucketString == null) {
       return null;
     }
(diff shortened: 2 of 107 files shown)


packages/storage/src/implementation/constants.ts
```diff
@@ -18,15 +18,17 @@
  * @fileoverview Constants used in the Firebase Storage library.
  */
 
+export const domain: string = "firebasestorage.googleapis.com";
```
avolkovi: minor, but this should probably be called something like `api_base_url`, domain is a little ambiguous

Feiyang1: Just want to point out that constant variable names should look like CONSTANT_VALUES_LIKE_THIS per our style guide, but they are all in camelCases in this file.

Would you be able to update them ?

fredzqm: Done

fredzqm: I picked `DOMAIN_NAME`, `API_BASE_URL` is taken below

packages/auth/test/utils_test.js
```diff
@@ -438,11 +438,11 @@ function testMatchDomain_ipAddressDomain() {
 
 function testMatchDomain_caseInsensitiveMatch() {
   assertTrue(fireauth.util.matchDomain('localhost', 'localhost', 'http'));
-  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN.COM', 'http'));
+  assertTrue(fireauth.util.matchDomain('domain.com', 'DOMAIN_NAME.COM', 'http'));
```
Feiyang1: I think you accidentally replaced the text here.

fredzqm: Thanks! Good catch. I reverted it.
I am quite surprised that this did not break the build.

packages/storage/src/implementation/constants.ts
```diff
@@ -18,49 +18,50 @@
  * @fileoverview Constants used in the Firebase Storage library.
  */
 
+/**
+ * Domain name for firebase storage.
+ */
+export const DOMAIN_NAME: string = 'firebasestorage.googleapis.com';
```
schmidt-sebastian: FWIW, `DEFAULT_HOST` might be a better name here.

packages/storage/src/implementation/constants.ts
```diff
@@ -18,49 +18,50 @@
...
+
 /**
  * Domain and scheme for API calls.
  */
-export const domainBase: string = 'https://firebasestorage.googleapis.com';
+export const API_DOMAIN: string = `https://${DOMAIN_NAME}`;
```
schmidt-sebastian: Endpoint or something similar?

packages/storage/src/implementation/constants.ts
```diff
@@ -18,49 +18,50 @@
...
-}
+export const API_UPLOAD_BASE_URL: string = '/v0';
 
 export const configOption: string = 'storageBucket';
 
 /**
  * 1 minute
  */
-export const shortMaxOperationRetryTime: number = 1 * 60 * 1000;
+export const SHORT_MAX_OPERATION_RETRY_TIME: number = 1 * 60 * 1000;
```
schmidt-sebastian: Since these are now global constants, we should use much more descriptive names. Can this name be changed to reflect better what it is used for?

packages/storage/src/implementation/constants.ts
```diff
@@ -18,49 +18,50 @@
...
  * 2 minutes
  */
-export const defaultMaxOperationRetryTime: number = 2 * 60 * 1000;
+export const DEFAULT_MAX_OPERATION_RETRY_TIME: number = 2 * 60 * 1000;
 
 /**
  * 10 minutes
  */
-export const defaultMaxUploadRetryTime: number = 10 * 60 * 100;
+export const DEFAULT_MAX_UPLOAD_RETRY_TIME: number = 10 * 60 * 100;
```
schmidt-sebastian: `DEFAULT_MAX_UPLOAD_RETRY_TIME` is a good example for a descriptive name.

packages/storage/src/implementation/constants.ts
schmidt-sebastian: On second thought, these are only used sparingly, so we could avoid the naming discussion by simply removing them.

packages/storage/src/implementation/location.ts
```diff
@@ -80,13 +80,9 @@ export class Location {
       loc.path_ = decodeURIComponent(loc.path);
     }
     let version = 'v[A-Za-z0-9_]+';
+    let domainRegex = DEFAULT_HOST.replace(/[.]/g, '\\.');
```
schmidt-sebastian: Optional: Rename to hostRegex

packages/storage/src/implementation/authwrapper.ts
```diff
@@ -74,15 +74,15 @@ export class AuthWrapper {
     this.requestMaker_ = requestMaker;
     this.pool_ = pool;
     this.service_ = service;
-    this.maxOperationRetryTime_ = constants.defaultMaxOperationRetryTime;
-    this.maxUploadRetryTime_ = constants.defaultMaxUploadRetryTime;
+    this.maxOperationRetryTime_ = constants.DEFAULT_MAX_OPERATION_RETRY_TIME;
```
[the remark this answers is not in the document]
schmidt-sebastian: I don't have a strong opinion, but all else being equal, `timeout` is the more commonly used term.

packages/storage/src/implementation/url.ts
```diff
@@ -18,19 +18,11 @@
...
-}
-
-export function makeDownloadUrl(urlPart: string): string {
-  return constants.downloadBase + constants.apiBaseUrl + urlPart;
-}
-
-export function makeUploadUrl(urlPart: string): string {
-  return constants.domainBase + constants.apiUploadBaseUrl + urlPart;
+export function makeUrl(urlPart: string): string {
+  return `https://${DEFAULT_HOST}/v0/${urlPart}`;
```
schmidt-sebastian: BTW, the last `/` is breaking CI.

packages/storage/src/implementation/authwrapper.ts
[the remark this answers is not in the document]
fredzqm: I think this name is already part of the public API, so we cannot change it.

packages/storage/src/implementation/url.ts
fredzqm: thx

Issue #1447, opened 2019-01-02
Why does firestore / AngularFire SDK packs multiple promises to one request?


jimmykane:
### Version info

**Angular:**  7.1.4

**Firebase:**5.7.1 

**AngularFire:**5.1.1

**Other (e.g. Ionic/Cordova, Node, browser, operating system):**

### How to reproduce these conditions

**Failing test unit, Plunkr, or JSFiddle demonstrating the problem**

**Steps to set up and reproduce**

Enforce Slow network

**Sample data and security rules**

<-- include/attach/link to some json sample data (or provide credentials to a sanitized, test Firebase project) -->

### Debug output

** Errors in the JavaScript console **

** Output from `firebase.database().enableLogging(true);` **

** Screenshots **

### Expected behavior

Many requests for multiple promises 

### Actual behavior

One request for multiple promises and that is problematic 

Here is how I write a document and it's subcollections:

```
public async setEvent(event: EventInterface): Promise<void[]> {
    return new Promise<void[]>(async (resolve, reject) => {
      const writePromises: Promise<void>[] = [];
      event.setID(event.getID() || this.afs.createId());
      event.getActivities()
        .forEach((activity) => {
          activity.setID(activity.getID() || this.afs.createId());
          writePromises.push(this.afs.collection('events').doc(event.getID()).collection('activities').doc(activity.getID()).set(activity.toJSON()));
          activity.getAllStreams().forEach((stream) => {
            this.logger.info(`Steam ${stream.type} has size of GZIP ${getSize(this.getBlobFromStreamData(stream.data))}`);
            writePromises.push(this.afs
              .collection('events')
              .doc(event.getID())
              .collection('activities')
              .doc(activity.getID())
              .collection('streams')
              .doc(stream.type) // @todo check this how it behaves
              .set({
                type: stream.type,
                data: this.getBlobFromStreamData(stream.data),
              }))
          });
        });
      try {
        await Promise.all(writePromises);
        await this.afs.collection('events').doc(event.getID()).set(event.toJSON());
        resolve()
      } catch (e) {
        Raven.captureException(e);
        // Try to delete the parent entity and all subdata
        await this.deleteEvent(event.getID());
        reject('Something went wrong')
      }
    })
  }
``` 

However when I look at the network tab:

![image](https://user-images.githubusercontent.com/1268723/50444433-c7eb8480-0908-11e9-8536-1f64b570024e.png)

[...]

PR #1603, opened 2019-03-13
b/123095740: Increase webchannel timeout to accommodate slow internet connections (#1447).

mikelehen:


packages/webchannel-wrapper/externs/overrides.js
```diff
@@ -62,3 +62,6 @@ goog.net.WebChannel.Options.backgroundChannelTest;
 
 /** @type {boolean|undefined} */
 goog.net.WebChannel.Options.fastHandshake;
+
+/** @type {(!Object<string, boolean|number>|undefined)|undefined} */
```
schmidt-sebastian: I'm not sure I fully follow this type definition. Can we not remove one of the `undefined`s and remove the parentheses?

See https://cs.corp.google.com/piper///depot/google3/javascript/closure/labs/net/webchannel.js?q=webchannel+internalChannelParams&dr=CSs&l=211

mikelehen: Er, good catch!  I had a brain hiccup when I copy/pasted that I misread it to think the existing "|undefined" was inside the Object definition and so I added another one... :-/  Fixed.

PR #1604, opened 2019-03-13
Modify doc generation script to also generate node reference docs

hsubox76:
This creates 2 versions of the HOME.md and toc.yaml content source files, one for js and one for node, and adds an `--api` command line option to choose which API to generate docs for.  The output files will be written to `html/js` and `html/node` respectively.

Since Typedoc generates all files derived from reading `index.d.ts` and node docs only contain a subset of those files, the script also removes extra files generated.


package.json
```diff
@@ -30,7 +30,7 @@
     "test:setup": "node tools/config.js",
     "pretest:saucelabs": "lerna run --parallel pretest",
     "test:saucelabs": "karma start config/karma.saucelabs.js --single-run",
-    "docgen": "node scripts/docgen/generate-docs.js"
+    "docgen": "node scripts/docgen/generate-docs.js --api js; node scripts/docgen/generate-docs.js --api node"
```
Feiyang1: I think it'd be helpful to be able to run js or node separately. So I'd suggest to add 2 more scripts:
```javascript
{
"docgen:js": ...
"docgen:node": ...
"docgen": "npm run docgen:js; npm run docgen:node"
}
```

scripts/docgen/generate-docs.js
```diff
@@ -21,13 +21,16 @@ const fs = require('mz/fs');
 const path = require('path');
 const yaml = require('js-yaml');
 
+// API to generate docs for, 'node' or 'js' - default to 'js'
+const apiType = argv.api || 'js';
```
Feiyang1: I know it's a very simple API. It would be nice if we can show the option usage with `-help` command.

package.json
hsubox76: Done.

scripts/docgen/generate-docs.js
hsubox76: Done.

package.json
```diff
@@ -30,7 +30,9 @@
     "test:setup": "node tools/config.js",
     "pretest:saucelabs": "lerna run --parallel pretest",
     "test:saucelabs": "karma start config/karma.saucelabs.js --single-run",
-    "docgen": "node scripts/docgen/generate-docs.js --api js; node scripts/docgen/generate-docs.js --api node"
+    "docgen": "node scripts/docgen/generate-docs.js --api js; node scripts/docgen/generate-docs.js --api node",
```
Feiyang1: no need to repeat the command -
`
"docgen": "npm run docgen:js; npm run docgen:node"
`

scripts/docgen/generate-docs.js
```diff
@@ -16,18 +16,29 @@
...
 const fs = require('mz/fs');
 const path = require('path');
 const yaml = require('js-yaml');
 
-// API to generate docs for, 'node' or 'js' - default to 'js'
-const apiType = argv.api || 'js';
 const repoPath = path.resolve(`${__dirname}/../..`);
+
+// Command-line options.
+const { api: apiType, source: sourceFile } = yargs
```
Feiyang1: nice!

package.json
hsubox76: Oh right, done.

PR #1605, opened 2019-03-14
b/125385470: Fix transactions w/ auth issue on NodeJS.

mikelehen:
In https://github.com/firebase/firebase-js-sdk/commit/81cd260aeb729c5003952988b19d5b32788510d4 we moved to sending our auth metadata directly in each RPC call, but for RPCs that take a callback (i.e. non-streaming RPCs like we use to commit a transaction), the auth metadata must be passed *before* the callback, but we were unconditionally passing it as the last argument: https://github.com/firebase/firebase-js-sdk/blob/a18b2905a161f302bf0fad7a32dd82507932093e/packages/firestore/src/platform_node/grpc_connection.ts#L145

I've fixed this by removing the getRpcCallable() helper which I felt wasn't being that helpful anymore. I also cleaned up ensureActiveStub() which accepted a token but didn't do anything with it.

### DIFF
--- a/packages/firestore/src/platform_node/grpc_connection.ts
+++ b/packages/firestore/src/platform_node/grpc_connection.ts
@@ -96,16 +96,16 @@ export class GrpcConnection implements Connection {
         ? grpc.credentials.createSsl()
         : grpc.credentials.createInsecure();
       this.cachedStub = new this.firestore.Firestore(
-          this.databaseInfo.host,
-          credentials,
-          {
-            // We do our own connection backoff (that for example is aware of whether or
-            // not a write stream error is permanent or not) so we don't want gRPC to do
-            // backoff on top of that. 100ms is the minimum value that gRPC allows.
-            'grpc.initial_reconnect_backoff_ms': 100,
-            'grpc.max_reconnect_backoff_ms': 100
-          }
-        );
+        this.databaseInfo.host,
+        credentials,
+        {
+          // We do our own connection backoff (that for example is aware of whether or
+          // not a write stream error is permanent or not) so we don't want gRPC to do
+          // backoff on top of that. 100ms is the minimum value that gRPC allows.
+          'grpc.initial_reconnect_backoff_ms': 100,
+          'grpc.max_reconnect_backoff_ms': 100
+        }
+      );
     }
     return this.cachedStub;
   }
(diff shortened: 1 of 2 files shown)


packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -95,26 +83,19 @@ export class GrpcConnection implements Connection {
...
-  private sameToken(tokenA: Token | null, tokenB: Token | null): boolean {
-    const valueA = tokenA && tokenA.authHeaders['Authorization'];
-    const valueB = tokenB && tokenB.authHeaders['Authorization'];
-    return valueA === valueB;
-  }
-
-  private ensureActiveStub(token: Token | null): GeneratedGrpcStub {
-    if (!this.cachedStub || !this.sameToken(this.cachedStub.token, token)) {
+  private ensureActiveStub(): GeneratedGrpcStub {
+    if (!this.cachedStub) {
```
ryanpbrewster: What consequences will removing this `||` statement have? Does this mean that we'll no longer reconnect with fresh credentials when a new token comes in?

mikelehen: I'm hoping it has no consequences! 😄 

We used to associate the auth token with the stub (which might correspond to a gRPC channel?) which would allow all our RPCs to pick up the token for free.  And so in that world we had to recreate the stub whenever the token changed.  Hence this logic.

But after your rework, the reason for re-creating the stub went away because we no longer associate the token with the stub and so then we were just recreating the stub.

As for "no longer reconnect", gRPC manages the actual connection to the backend under the covers and I think even recreating the stub probably didn't trigger an actual TCP-level reconnect, though I don't have an easy way to verify that.

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -82,11 +75,6 @@ function createMetadata(
 // tslint:disable-next-line:no-any
 type GeneratedGrpcStub = any;
```
schmidt-sebastian: I am wondering if it makes sense to define the type for GeneratedGrpcStub. It should end up looking something like
https://github.com/googleapis/nodejs-firestore/blob/master/dev/test/util/helpers.ts#L65 and might prevent issues like this in the feature.

mikelehen: Oh, nice...  That is interesting, but is probably hard to do in a useful way right now since we rely on having a relatively-untyped `Connection` interface that's shared by WebChannel and gRPC (e.g. https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/connection.ts#L46)...  But I'll keep that in the back of my head in case we ever feel like reworking that.

PR #1610, opened 2019-03-17
Storage List API

fredzqm:
API proposal: internal golink firebase-storage-list-api


packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,90 @@
...
+  const validType = p && type.isObject(p);
+  if (!validType) {
+    throw 'Expected ListOptions object.';
+  }
+  const maxResultsKey = 'maxResults';
+  const pageTokenKey = 'pageToken';
+  for (const key in p) {
+    if (key === maxResultsKey) {
+      if (!type.isInteger(p[maxResultsKey]) || p[maxResultsKey] <= 0) {
+        throw 'Expected maxResults to be positive.';
```
schmidt-sebastian: Nit: If you change this to 'to be a positive number', your error message will cover both cases.

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+    } else if (key === pageTokenKey) {
+      if (!type.isString(p[pageTokenKey])) {
+        throw 'Expected pageToken to be string.';
+      }
+    } else {
+      throw 'Unknown option ' + key;
```
schmidt-sebastian: Nit: I would add a column after 'option'.

packages/storage/src/list.ts
```diff
@@ -0,0 +1,33 @@
...
+type ListOptions = {
+  maxResults?: number | null | undefined;
+  pageToken?: string | null | undefined;
+};
+
+type ListResult = {
+  prefixes: Reference[];
+  items: Reference[];
+  nextPageToken?: string | null | undefined;
+};
```
schmidt-sebastian: You can remove `undefined` from all these definitions.

packages/storage/src/list.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
schmidt-sebastian: 2019

packages/storage/src/list.ts
```diff
@@ -0,0 +1,33 @@
...
+  pageToken?: string | null | undefined;
+};
+
+type ListResult = {
+  prefixes: Reference[];
+  items: Reference[];
+  nextPageToken?: string | null | undefined;
+};
+
+export { ListOptions, ListResult };
```
schmidt-sebastian: Nit: You can export the types directly (`export type ...`)

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,28 @@ export class Reference {
     });
   }
 
+  /**
+   * @param options.maxResults limits the total number of prefixes and items to return..
+   * @param options.pageToken nextPageToken from a previous list(). Resume from that position.
```
schmidt-sebastian: Maybe change this to:

'The `nextPageToken` from a previous list() response. If provided, listing is resumed from that position.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,28 @@ export class Reference {
     });
   }
 
+  /**
+   * @param options.maxResults limits the total number of prefixes and items to return..
```
schmidt-sebastian: Maybe prefix with `If set, limits...`

packages/storage/test/requests.test.ts
```diff
@@ -172,9 +171,9 @@ describe('Firebase Storage > Requests', () => {
       [locationNormal, locationNormalUrl],
       [locationEscapes, locationEscapesUrl]
     ];
-    for (let i = 0; i < maps.length; i++) {
-      const location = maps[i][0] as Location;
-      const url = maps[i][1] as string;
+    for (const d of maps) {
```
schmidt-sebastian: While you are here, you could make this:

```
    for (const [location, url ] of maps) {
```

packages/storage/test/requests.test.ts
```diff
@@ -196,6 +195,82 @@ describe('Firebase Storage > Requests', () => {
...
+        headers: {},
+        urlParams: {
+          prefix: '',
+          delimiter: '/'
+        }
+      },
+      requestInfo
+    );
+  });
+  it('list request info', () => {
```
schmidt-sebastian: Nit: add empty line above

packages/storage/test/requests.test.ts
```diff
@@ -196,6 +195,82 @@ describe('Firebase Storage > Requests', () => {
...
+            delimiter: '/',
+            pageToken: pageToken,
+            maxResults: maxResults
+          }
+        },
+        requestInfo
+      );
+    }
+  });
+  it('list handler', () => {
```
schmidt-sebastian: Nit: add empty line above

packages/storage/test/requests.test.ts
fredzqm: Done

packages/storage/test/requests.test.ts
fredzqm: Done

packages/storage/test/requests.test.ts
fredzqm: done

packages/storage-types/index.d.ts
```diff
@@ -51,6 +51,18 @@ export interface Reference {
   storage: Storage;
   toString(): string;
   updateMetadata(metadata: SettableMetadata): Promise<FullMetadata>;
+  list(options?: ListOptions): Promise<ListResult>;
```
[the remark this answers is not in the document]
fredzqm: Done

packages/storage-types/index.d.ts
```diff
@@ -51,6 +51,18 @@ export interface Reference {
...
+}
+
+export interface ListResult {
+  prefixes: Reference[];
+  items: Reference[];
+  nextPageToken: string | null;
+}
+
+export interface ListOptions {
+  maxResults?: number | null | undefined;
```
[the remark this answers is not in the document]
fredzqm: Good point. Fixed

packages/firebase/index.d.ts
```diff
@@ -5472,6 +5472,52 @@ declare namespace firebase.storage {
...
+    /**
+     * The maximum number of results returned.
+     */
+    maxResults?: number | null;
+    /**
+     * A pageToken returned from a previous list().
+     * If set, resumes list from the previous location. prefixes and items already listed
+     * will not show up.
+     */
+    pageToken?: string | null;
```
fredzqm: @alexg33 Could you take a look at the documentation here?
Internal API link firebase-storage-list-api
The backend roughly follows https://cloud.google.com/storage/docs/json_api/v1/objects/list

packages/firebase/index.d.ts
```diff
@@ -5472,6 +5472,52 @@ declare namespace firebase.storage {
      *     including if the object did not exist.
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
+    /**
+     * List items and folders (prefixes) within this directory.
```
schmidt-sebastian: Please wrap this to the prevalent line length in the file. I would also add an empty line between these paragraphs.

packages/firebase/index.d.ts
```diff
@@ -5472,6 +5472,52 @@ declare namespace firebase.storage {
...
+     * "/" is treated as a path delimiter. Firebase storage does not support invalid object path that ends with "/" or contains two consecutive "//".
+     * All invalid objects in GCS will be filtered.
+     *
+     * @param {!options.maxResults} If set, limits the total number of prefixes and items to return..
```
schmidt-sebastian: I know that the rest of the file uses these exclamation marks, but they are only used in JSDoc to describe the type of the argument. Here (and everywhere else in the Storage section), they seem to be used to define the argument name. That doesn't seem correct. 

You should also drop the curly braces for the argument names.

packages/firebase/index.d.ts
```diff
@@ -5472,6 +5472,52 @@ declare namespace firebase.storage {
...
+     * @param {!options.pageToken} The nextPageToken from a previous list() response. If provided, listing is resumed from that position.
+     * @return {!firebase.Promise<firebase.storage.ListResult>} A promise that resolves with the sub-directories and items.
```
schmidt-sebastian: @Feiyang1 Should we drop the return types? And if not, does it make sense to use a standard `Promise` here, since that is what we are returning?

packages/storage/src/implementation/list.ts
```diff
@@ -24,26 +24,41 @@ import * as json from './json';
...
+  bucket: string;
+}
+
+interface ListResultResource {
+  prefixes: string[];
+  items: ListMetadataReource[];
+  nextPageToken?: string;
+}
+
+const maxResultsKey = 'maxResults';
```
schmidt-sebastian: I believe you should use all uppercase here: https://g3doc.corp.google.com/javascript/typescript/g3doc/styleguide.md?cl=head#identifiers

packages/storage/src/implementation/list.ts
```diff
@@ -24,26 +24,41 @@ import * as json from './json';
 import * as type from './type';
 import { ListResult } from '../list';
 
-export function fromResource(
+interface ListMetadataReource {
```
schmidt-sebastian: Missing "s"

schmidt-sebastian: Can you add a short comment for these interfaces?

packages/storage/src/implementation/list.ts
```diff
@@ -24,26 +24,41 @@ import * as json from './json';
...
+
+const maxResultsKey = 'maxResults';
+const pageTokenKey = 'pageToken';
+const prefixesKey = 'prefixes';
+const itemsKey = 'items';
+
+function fromResource(
   authWrapper: AuthWrapper,
-  resource: { [name: string]: any }
+  resource: ListResultResource
```
schmidt-sebastian: Is "resource" really a response here? Please push back if "resource" is the expected terminology here, but I think we could still improve our terminology here a bit. 

```
function convertBackendResponse(authWrapper: AuthWrapper, response: ListResponse) : ListResult {...}
```

packages/storage/src/reference.ts
```diff
@@ -204,10 +204,12 @@ export class Reference {
   }
 
   /**
-   * @param options.maxResults limits the total number of prefixes and items to return..
-   * @param options.pageToken nextPageToken from a previous list(). Resume from that position.
+   * List items and folders (prefixes) within this directory.
+   *
+   * @param {!options.maxResults} If set, limits the total number of prefixes and items to return..
+   * @param {!options.pageToken} The nextPageToken from a previous list() response. If provided, listing is resumed from that position.
```
schmidt-sebastian: Please also re-wrap these lines to match the file's line length.

packages/firebase/index.d.ts
alexg33: I actually don't have the bandwidth for this at the moment. Let me get back to you with a writer that might cover.

packages/firebase/index.d.ts
```diff
@@ -5467,11 +5467,62 @@ declare namespace firebase.storage {
...
+   * The options list() accepts
+   */
+  interface ListOptions {
+    /**
+     * The maximum number of results returned.
+     */
+    maxResults?: number | null;
+    /**
+     * A pageToken returned from a previous list().
+     * If set, resumes list from the previous location. prefixes and items already listed
```
schmidt-sebastian: Add backticks here too.

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Represents the JSON response of List API.
+ */
+interface ListResultResponse {
+  prefixes: string[];
+  items: ListMetadataResponse[];
+  nextPageToken?: string;
+}
+
+const MAXRESULTS_KEY = 'maxResults';
```
schmidt-sebastian: Should this be "MAX_RESULTS_KEY" and "PAGE_TOKEN_KEY"?

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,110 @@
...
+const PAGETOKEN_KEY = 'pageToken';
+const PREFIXES_KEY = 'prefixes';
+const ITEMS_KEY = 'items';
+
+function fromBackendResponse(
```
schmidt-sebastian: This is a much much clearer name (especially to someone that doesn't know the rest of the codebase (*)). Thanks!

(*) Like me.

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,110 @@
...
+      const reference = authWrapper.makeStorageReference(
+        new Location(authWrapper.bucket(), item['name'])
+      );
+      listResult.items.push(reference);
+    }
+  }
+  return listResult;
+}
+
+export function fromResourceString(
```
schmidt-sebastian: Do you think this would be worth renaming to `fromResponseString`?

packages/storage/src/list.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Reference } from './reference';
+
+/**
+ * @fileoverview Documentation for ListOptions and ListResult format.
+ */
+export type ListOptions = {
```
schmidt-sebastian: Can you check if the the tooling that we use for documentation generation will attach the comment above to `ListOptions`? I am not sure how `@fileoverview` works and if you need an empty line in between.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,34 @@ export class Reference {
     });
   }
 
+  /**
+   * List items and folders (prefixes) within this directory.
+   *
+   * @param {!options.maxResults} If set, limits the total number of prefixes
```
schmidt-sebastian: Drop just the curly braces here to.

If you look at this in GitHub, you can actually see that the `If` is highlighted differently. The format is:

```
@param {type} argName Bla bla bla
```

Since this is TypeScript, you can drop `{type}`.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,34 @@ export class Reference {
...
+   *      and items to return..
+   * @param {!options.pageToken} The nextPageToken from a previous list()
+   *      response. If provided, listing is resumed from that position.
+   * @return A promise that resolves with ListResult. `prefixes` contains
```
schmidt-sebastian: Yay backticks! :) Thanks!

packages/storage/src/list.ts
fredzqm: I realized that I need to add those types to `toc.yaml` in order for `docgen` to work

fredzqm:
@schmidt-sebastian Could you give a final approval?
Do you know why there is a Cloud Build blocking the merge? Do I need to create a different PR to get around it?

schmidt-sebastian:
I'll go through this and approve this by EOW.

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Represents the JSON response of List API.
+ */
+interface ListResultResponse {
+  prefixes: string[];
+  items: ListMetadataResponse[];
+  nextPageToken?: string;
+}
+
+const MAXR_ESULTS_KEY = 'maxResults';
```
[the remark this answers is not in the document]
fredzqm: Thanks for catching it.
I will do another round of proof read as well.

packages/firebase/index.d.ts
```diff
@@ -5473,6 +5473,57 @@ declare namespace firebase.storage {
      *     including if the object did not exist.
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
+    /**
+     * List items and sub-directories within this directory.
+     *
+     * "/" is treated as a path delimiter. Firebase storage does not support
+     * invalid object path that ends with "/" or contains two consecutive "//".
```
schmidt-sebastian: s/path/paths/

packages/firebase/index.d.ts
```diff
@@ -5473,6 +5473,57 @@ declare namespace firebase.storage {
...
+   * The options list() accepts
+   */
+  interface ListOptions {
+    /**
+     * The maximum number of results returned.
+     */
+    maxResults?: number | null;
+    /**
+     * A pageToken returned from a previous call to list().
+     * If set, resumes list from the previous location. prefixes and items already listed
```
schmidt-sebastian: I would remove the second sentence. If not, s/prefixes/Prefixes/

packages/firebase/index.d.ts
```diff
@@ -5473,6 +5473,57 @@ declare namespace firebase.storage {
...
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
+    /**
+     * List items and sub-directories within this directory.
+     *
+     * "/" is treated as a path delimiter. Firebase storage does not support
+     * invalid object path that ends with "/" or contains two consecutive "//".
+     * All invalid objects in GCS will be filtered.
+     *
+     * @param options.maxResults If set, limits the total number of `prefixes`
```
schmidt-sebastian: Please validate that this number doesn't exceed 1000. It would also be good to document this.

fredzqm: Done

packages/firebase/index.d.ts
fredzqm: Done

packages/firebase/index.d.ts
fredzqm: Removed

packages/storage/src/implementation/list.ts
```diff
@@ -0,0 +1,110 @@
...
+}
+
+export function listOptionsValidator(p: any) {
+  const validType = p && type.isObject(p);
+  if (!validType) {
+    throw 'Expected ListOptions object.';
+  }
+  for (const key in p) {
+    if (key === MAX_RESULTS_KEY) {
+      if (!type.isInteger(p[MAX_RESULTS_KEY]) || p[MAX_RESULTS_KEY] <= 0) {
```
schmidt-sebastian: Where do we validate that this number is less than or equal to 1000? Is that still missing?

fredzqm: Sorry, just added that and a test.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,38 @@ export class Reference {
...
+   * @param options.maxResults If set, limits the total number of `prefixes`
+   *      and `items` to return.
+   * @param options.pageToken The `nextPageToken` from a previous call to
+   *      list(). If provided, listing is resumed from the previous position.
+   * @return A promise that resolves with the items and sub-directories.
+   *      `prefixes` contains references to sub-directories and `items`
+   *      contains references to objects in this folder. `nextPageToken`
+   *      can be passed as `options.pageToken` to get the rest of results.
+   */
+  list(options?: ListOptions | null): Promise<ListResult> {
```
fredzqm: I don't want to make changes twice everytime
TODO(@fredzqm): Once the PR is approved, copy the documentation from `index.d.ts` here.

packages/firebase/index.d.ts
```diff
@@ -5474,35 +5474,50 @@ declare namespace firebase.storage {
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
     /**
-     * List items and sub-directories within this directory.
+     * List items(files) and prefixes(folders) with this prefix of ref.name+'/'.
```
schmidt-sebastian: Nit: add spaces before the parenthesis. I also think we should go back to "within this directory".

My suggestion:

```
List items (files) and prefixes (folders) with this directory.
```

packages/firebase/index.d.ts
```diff
@@ -5474,35 +5474,50 @@ declare namespace firebase.storage {
...
 
   /**
-   * Result returned by list()
+   * Result returned by list().
    */
   interface ListResult {
     /**
-     * References to sub-directories. You can call list() on them to get its contents.
+     * References to sub-folders. You can call list() on them to get its contents.
+     * GCS's
```
schmidt-sebastian: remove "GCS's"

packages/firebase/index.d.ts
```diff
@@ -5512,11 +5527,11 @@ declare namespace firebase.storage {
   }
 
   /**
-   * The options list() accepts
+   * The options list() accepts.
    */
   interface ListOptions {
     /**
-     * The maximum number of results returned.
+     * The maximum number of results (items+prefixs) returned.
```
schmidt-sebastian: s/(items+prefixes)/(items and prefixes)

fredzqm: Done

packages/firebase/index.d.ts
fredzqm: Done

packages/firebase/index.d.ts
fredzqm: Done

packages/storage/src/reference.ts
fredzqm: Synced them once. Will do it again after approval.

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
      *     including if the object did not exist.
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
+    /**
+     * List all items (files) and prefixes (folders) under this storage reference.
+     *
+     * It is a helper method for calling list() recursively until there is no
```
schmidt-sebastian: s/It is/This is/
s/recursively/repeatedly/
s/there is/there are/

:)

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     * more results. The default pagination size is 200.
```
schmidt-sebastian: Why do we not use the maximum page size? It seems like it would be faster (less roundtrips) and less likely to be inconsistent.

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     *
+     * Note: the results may not be a consistent snapshot if objects are changed
```
schmidt-sebastian: s/the/The

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     * between paginating requests.
```
schmidt-sebastian: "while this operation is running" ?

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     *
+     * Warning: If there are
```
schmidt-sebastian: Remove?

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     *
+     * @return A promise that resolves with all the items and prefixes under
```
schmidt-sebastian: s/promise/Promise

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     *      the current storage reference. `prefixes` contains references to
+     *      sub-directories and `items` contains references to objects in this
+     *      folder. `nextPageToken` is never returned.
+     */
+    listAll(): Promise<ListResult>;
```
schmidt-sebastian: Add empty line above.

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+    /**
+     * List items (files) and prefixes (folders) under this storage reference.
+     *
+     * GCS is a key-blob store. Firebase storage impose the semantic of '/'
+     * delimited folder structure.
+     * Objects whose paths, aside from the prefix, contain '/' will have
```
schmidt-sebastian: Can you find a way simplify this?

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,88 @@ declare namespace firebase.storage {
...
+     * their path, truncated after the '/', returned in prefixes.
+     * Duplicate prefixes are omitted.
+     * Objects whose paths, aside from the prefix, do not contain '/' will
+     * be returned in items.
+     *
+     * Firebase storage does not support invalid object paths that end with
```
schmidt-sebastian: s/Firebase storage/Cloud Storage for Firebase (?) or "Firebase Storage"

I would also remove "invalid" since they are not invalid in GCS.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,89 @@ export class Reference {
...
+  /**
+   * List all items (files) and prefixes (folders) under this storage reference.
+   *
+   * It is a helper method for calling list() recursively until there is no
+   * more results. The default pagination size is 200.
+   *
+   * Note: the results may not be a consistent snapshot if objects are changed
+   * between paginating requests.
+   *
+   * Warning: If there are
```
schmidt-sebastian: Please update this comment as well.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,89 @@ export class Reference {
...
+    };
+    return this.listAllHelper(accumulator, null).then(() => accumulator);
+  }
+
+  private async listAllHelper(
+    accumulator: ListResult,
+    pageToken?: string
+  ): Promise<void> {
+    let opt: ListOptions = {
+      maxResults: 1,
```
schmidt-sebastian: Please make this 1000.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,89 @@ export class Reference {
...
+      pageToken
+    };
+    const nextPage = await this.list(opt);
+    Array.prototype.push.apply(accumulator.prefixes, nextPage.prefixes);
```
schmidt-sebastian: The new and modern syntax for this is `accumulator.prefixes(...nextPage.prefixes)`, which runs in O(n).

You could also use `concat` and make this potentially faster.

packages/storage/src/reference.ts
```diff
@@ -201,6 +203,89 @@ export class Reference {
...
+    Array.prototype.push.apply(accumulator.items, nextPage.items);
+    if (nextPage.nextPageToken) {
```
schmidt-sebastian: This should technically check against "undefined", even though this will never make a difference :)

packages/storage/test/reference.test.ts
```diff
@@ -493,6 +576,19 @@ describe('Firebase Storage > Reference', () => {
         'storage/invalid-root-operation'
       );
     });
+    it("listAll doesn't throws", () => {
```
schmidt-sebastian: s/throws/throw/

packages/storage/test/reference.test.ts
```diff
@@ -493,6 +576,19 @@ describe('Firebase Storage > Reference', () => {
...
+      assert.doesNotThrow(() => {
+        root.listAll();
+      });
+    });
+    it("list doesn't throws", () => {
```
schmidt-sebastian: s/throws/throw/

packages/storage/src/reference.ts
fredzqm: Removed to let is use the default.

packages/storage/src/reference.ts
fredzqm: Done.
I thought `concat` does not modify the existing list, so it is O(n2) for many pages.

packages/storage/src/reference.ts
fredzqm: Done

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,85 @@ declare namespace firebase.storage {
      *     including if the object did not exist.
      */
     updateMetadata(metadata: firebase.storage.SettableMetadata): Promise<any>;
+    /**
+     * List all items (files) and prefixes (folders) under this storage reference.
+     *
+     * This is a helper method for calling list() repeatedly until there are
+     * no more results. The default pagination size is 1000.
+     *
+     * Note: The results may not be a consistent snapshot if objects are
```
schmidt-sebastian: s/a consistent snapshot/consistent/

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,85 @@ declare namespace firebase.storage {
...
+     */
+    listAll(): Promise<ListResult>;
+    /**
+     * List items (files) and prefixes (folders) under this storage reference.
+     *
+     * GCS is a key-blob store. Firebase Storage imposes the semantic of '/'
+     * delimited folder structure.
+     * Refer to GCS's List API if you want to learn more.
+     *
+     * To adhere to Firebase Rules's Semantics, Firebase storage does not
```
schmidt-sebastian: s/Firebase storage/Firebase Storage/

packages/firebase/index.d.ts
```diff
@@ -5434,6 +5434,85 @@ declare namespace firebase.storage {
...
+     * support objects whose paths end with "/" or contain two consecutive
+     * "/"s. Firebase Storage List API will filter these unsupported objects.
+     * list() may fail if there are too many unsupported objects in the bucket.
+     *
+     * @param options.maxResults If set, limits the total number of `prefixes`
+     *      and `items` to return. The default and maximum maxResults is 1000.
+     *      Use the nextPageToken to retrieve more objects.
+     * @param options.pageToken The `nextPageToken` from a previous call to
+     *      list(). If provided, listing is resumed from the previous position.
+     * @return A promise that resolves with the items and prefixes.
```
schmidt-sebastian: s/promise/Promise

fredzqm:
@schmidt-sebastian Addressed all the doc suggestions, and copy js doc over once again.

It looks like this PR is still blocked on an owner's approval.
@Feiyang1 Could you take a look?

packages/firebase/index.d.ts
```diff
@@ -5484,6 +5484,87 @@ declare namespace firebase.storage {
...
+     *
+     * To adhere to Firebase Rules's Semantics, Firebase Storage does not
+     * support objects whose paths end with "/" or contain two consecutive
+     * "/"s. Firebase Storage List API will filter these unsupported objects.
+     * list() may fail if there are too many unsupported objects in the bucket.
+     *
+     * @param options.maxResults If set, limits the total number of `prefixes`
+     *      and `items` to return. The default and maximum maxResults is 1000.
+     *      Use the nextPageToken to retrieve more objects.
+     * @param options.pageToken The `nextPageToken` from a previous call to
```
Feiyang1: This style of describing the parameter only works if you define the type directly in this method, like the following:
```javascript
    list(options?: {maxResults: number | null, pageToken: string | null}
```
If an interface is used, the comments should coexist with the interface definition.

Since you have defined an interface, can you move the comments to the interface?

scripts/docgen/content-sources/node/toc.yaml
```diff
@@ -124,3 +124,25 @@ toc:
     path: /docs/reference/node/firebase.firestore.Transaction
   - title: "WriteBatch"
     path: /docs/reference/node/firebase.firestore.WriteBatch
+
+- title: "firebase.storage"
```
Feiyang1: I don't think storage is supported in Nodejs.

packages/firebase/index.d.ts
fredzqm: Done. Updated the doc.

PR #1614, opened 2019-03-19
b/32935141: Clean up remaining TODOs in first-party-auth implementation.

mikelehen:
Low-pri cleanup...

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -26,7 +26,7 @@ import { Code, FirestoreError } from '../util/error';
 
 export interface FirstPartyCredentialsSettings {
   type: 'gapi';
-  client: Gapi;
+  client: unknown;
   sessionIndex: string;
 }
 
@@ -280,16 +280,7 @@ export class FirstPartyToken implements Token {
  * to applications hosted by Google.
  */
 export class FirstPartyCredentialsProvider implements CredentialsProvider {
-  constructor(private gapi: Gapi, private sessionIndex: string) {
-    assert(
-      !!(
-        this.gapi &&
-        this.gapi.auth &&
-        this.gapi.auth.getAuthHeaderValueForFirstParty
-      ),
-      'unexpected gapi interface'
-    );
-  }
+  constructor(private gapi: Gapi, private sessionIndex: string) {}
 
   getToken(): Promise<Token | null> {
     return Promise.resolve(new FirstPartyToken(this.gapi, this.sessionIndex));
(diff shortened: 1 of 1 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -240,35 +240,37 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
   user = User.FIRST_PARTY;
 
-  constructor(private gapi: Gapi, private sessionIndex: string) {
-    assert(
-      this.gapi &&
-        this.gapi['auth'] &&
-        this.gapi['auth']['getAuthHeaderValueForFirstParty'],
-      'unexpected gapi interface'
-    );
-  }
```
mikelehen: This assert was duplicated so I removed this one, but kept the other one.

packages/firestore/src/api/credentials.ts
```diff
@@ -240,35 +240,37 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
+  constructor(private gapi: Gapi, private sessionIndex: string) {}
 
   get authHeaders(): { [header: string]: string } {
-    return {
-      Authorization: this.gapi['auth']['getAuthHeaderValueForFirstParty']([]),
```
mikelehen: getAuthHeaderValueForFirstParty() can technically return null, which would violate this method's return value.  So I tweaked this to omit 'Authorization' in that case.

packages/firestore/src/api/credentials.ts
```diff
@@ -280,9 +282,11 @@ export class FirstPartyToken implements Token {
 export class FirstPartyCredentialsProvider implements CredentialsProvider {
   constructor(private gapi: Gapi, private sessionIndex: string) {
     assert(
-      this.gapi &&
-        this.gapi['auth'] &&
-        this.gapi['auth']['getAuthHeaderValueForFirstParty'],
+      !!(
```
schmidt-sebastian: Would it make sense to move this assert to `makeCredentialsProvider`? If we do that, we could then change the type of [FirestoreSettings.credentials](https://github.com/firebase/firebase-js-sdk/blob/54bdfc4cebb038ffe19b16bb263b1007132ca356/packages/firestore/src/api/database.ts#L156) to `unknown` and convert it to `Gapi` in  `makeCredentialsProvider()`.

Feel free to punt since this may very well not be worth your time :)

mikelehen: Oh!  I like that idea quite a bit.  Done.

packages/firestore/src/api/credentials.ts
```diff
@@ -26,7 +26,7 @@ import { Code, FirestoreError } from '../util/error';
 
 export interface FirstPartyCredentialsSettings {
   type: 'gapi';
-  client: Gapi;
+  client: unknown;
```
goldblatt: Following up on this change-- might intro issues if there was any usage of client.x since Gapi was previously an any

mikelehen:
Chatted with Susan offline and I think we're in agreement that this should be okay, but if it causes any issues on the next console import of the SDK, I'm happy to take a look.

PR #1620, opened 2019-03-20
s/firebase.Promise/Promise

fredzqm:
Refactor PR

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -108,7 +108,7 @@ declare namespace firebase {
      *     resolve. This does not apply if the user is anonymous.</dd>
      * </dl>
      *
-     * @return {!firebase.Promise<void>}
+     * @return {!Promise<void>}
      */
     delete(): Promise<void>;
     emailVerified: boolean;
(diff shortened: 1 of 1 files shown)


schmidt-sebastian:
LGTM but I don't have rubberstamp power over this part of the repo.

packages/firebase/index.d.ts
```diff
@@ -108,7 +108,7 @@ declare namespace firebase {
      *     resolve. This does not apply if the user is anonymous.</dd>
      * </dl>
      *
-     * @return {!firebase.Promise<void>}
+     * @return {!Promise<void>}
```
Feiyang1: Please remove all @return that doesn't have any comment since it can be inferred from the type definition itself.

packages/firebase/index.d.ts
```diff
@@ -1119,7 +1119,7 @@ declare namespace firebase.app {
      *   });
      * ```
      *
-     * @return {!firebase.Promise<void>} An empty promise fulfilled when the app has
+     * @return {!Promise<void>} An empty promise fulfilled when the app has
```
Feiyang1: Please remove the type, so it will be just `*  @return An empty promise fulfilled when the app has....`

packages/firebase/index.d.ts
fredzqm: Done

packages/firebase/index.d.ts
fredzqm: Done. vim macros.

PR #1628, opened 2019-03-22
Fixing a few typos.

egilmorez:


packages/firebase/index.d.ts
```diff
@@ -3018,6 +3018,7 @@ declare namespace firebase.auth {
    */
   class GithubAuthProvider_Instance implements firebase.auth.AuthProvider {
     /**
+     * @return {!firebase.auth.AuthProvider} The provider instance itself.
```
hsubox76: This line should be removed.

egilmorez: Got it, thanks!

PR #1629, opened 2019-03-22
Expose toJSON and fromJSON on AuthCredential

wti806:
- Exposed toJSON and fromJSON on AuthCredential.
- Changed AuthCredential from interface to abstract class as fromJSON is a static method.
- Changed the OAuthCredential constructor to take a OAuthProviderOptions object, but not expose the API to public. The constructor change is backward compatible as it still can take idToken and accessToken string.

### DIFF
--- a/packages/auth-types/index.d.ts
+++ b/packages/auth-types/index.d.ts
@@ -248,7 +248,8 @@ export interface UserMetadata {
 
 export type Persistence = string;
 
-export abstract class OAuthCredential extends AuthCredential {
+export class OAuthCredential extends AuthCredential {
+  private constructor();
   idToken?: string;
   accessToken?: string;
   secret?: string;
--- a/packages/auth-types/index.d.ts
+++ b/packages/auth-types/index.d.ts
@@ -104,7 +104,7 @@ export interface ApplicationVerifier {
   verify(): Promise<string>;
 }
 
-export class AuthCredential {
+export abstract class AuthCredential {
   providerId: string;
   signInMethod: string;
   toJSON(): Object;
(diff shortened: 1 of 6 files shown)


packages/auth-types/index.d.ts
```diff
@@ -246,7 +248,7 @@ export interface UserMetadata {
 
 export type Persistence = string;
 
-export interface OAuthCredential extends AuthCredential {
+export abstract class OAuthCredential extends AuthCredential {
```
bojeil-google: `OAuthCredential` is not really abstract.

wti806: Changed to concrete class and made the constructor private.

packages/auth/src/authcredential.js
```diff
@@ -635,23 +635,31 @@ fireauth.OAuthProvider.prototype.getScopes = function() {
...
  * @param {?string=} opt_accessToken The optional OAuth access token.
- * @param {?string=} opt_rawNonce The optional raw nonce.
  * @return {!fireauth.AuthCredential} The Auth credential object.
  */
-fireauth.OAuthProvider.prototype.credential = function(opt_idToken,
-    opt_accessToken, opt_rawNonce) {
-  if (!opt_idToken && !opt_accessToken) {
+fireauth.OAuthProvider.prototype.credential =
+    function(optionsOrIdToken, opt_accessToken) {
+  var oauthResponse;
```
Feiyang1: Nit: Can you please run prettifier through the code? The formatting is off.

wti806: I believe the prettifier will be automatically triggered before pushing to canonical repo. As you can see, there are 2 commits from the prettier in this PR: 
[AUTOMATED]: Prettier Code Styling
 Do I need to manually run the prettier?

PR #1632, opened 2019-03-25
Add firestore.FieldValue.increment to index.d.ts

egilmorez:
Based on details from an externs.js update that happened during our move to TypeScript source, I'm adding firestore.FieldValue.increment here.

This update is (or can be) staged for Google internal reviewers.

Thanks!

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7209,21 +7209,21 @@ declare namespace firebase.firestore {
     /**
      * Returns a special value that can be used with `set()` or `update()` that tells
      * the server to increment the field's current value by the given value.
-     *
+     * 
      * If either the operand or the current field value uses floating point precision,
      * all arithmetic follows IEEE 754 semantics. If both values are integers,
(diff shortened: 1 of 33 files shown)


packages/firebase/index.d.ts
```diff
@@ -7206,6 +7206,25 @@ declare namespace firebase.firestore {
...
+     * once processed by the Firestore backend, all integer operations are capped
+     * between -2^63 and 2^63-1.
+     * 
+     * If the current field value is not of type `number`, or if the field does not
+     * yet exist, the transformation sets the field to the given value.
+     *
+     * @param n The value to increment by.
+     * @return The FieldValue sentinel for use in a call to `set()` or `update()`.
+     */
+    static increment(...elements: any[]): FieldValue;
```
hsubox76: I think the param should be `n: number` not `...elements: any[]`? https://github.com/firebase/firebase-js-sdk/blob/c923efab982900f50141554d284b3d762fda97d1/packages/firestore-types/index.d.ts#L1315

egilmorez: Oh -- yes.  Pushed a fix for that.  Thanks!

PR #1634, opened 2019-03-25
b/66181677: WebChannel: Disable log redaction for potential performance improvement.

mikelehen:
See https://github.com/google/closure-library/blob/7d6ec4f903cefed13f0b53c6815f7ec62213b4f1/closure/goog/labs/net/webchannel.js#L168 for context.


mikelehen:
Actually, digging a bit deeper, I've convinced myself the redaction code isn't even included in our webchannel build, since we strip logging.  So this is unnecessary.

PR #1645, opened 2019-03-28
Improve AppError typings

mmermerkaya:
Randomly noticed this while looking into FirebaseError. This way TS will complain if the `errors` object is missing any `AppError` string, or contains a string that is not in `AppError`.

I'd still recommend using a `const enum` instead of a union type as `AppError` though.


Feiyang1:
Thanks very much! I updated it to use enum in the app-v6 branch. Also PR is welcomed ! <3

mmermerkaya:
Oh awesome, haha I didn't even realize we had that branch. I guess there's no need to switch to enums here since this will go away soon enough.

PR #1646, opened 2019-03-28
Use a const enum for ErrorCode

mmermerkaya:
Following my own advice from #1645 :slightly_smiling_face:

### DIFF
--- a/packages/messaging/src/models/errors.ts
+++ b/packages/messaging/src/models/errors.ts
@@ -17,7 +17,7 @@
 
 import { ErrorFactory } from '@firebase/util';
 
-export const enum ErrorCodes {
+export const enum ErrorCode {
   AVAILABLE_IN_WINDOW = 'only-available-in-window',
   AVAILABLE_IN_SW = 'only-available-in-sw',
   SHOULD_BE_INHERITED = 'should-be-overriden',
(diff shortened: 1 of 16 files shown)


packages/messaging/src/models/errors.ts
```diff
@@ -17,139 +17,142 @@
...
-  BAD_VAPID_KEY: 'bad-vapid-key',
-  BAD_SUBSCRIPTION: 'bad-subscription',
-  BAD_TOKEN: 'bad-token',
-  BAD_PUSH_SET: 'bad-push-set',
-  FAILED_DELETE_VAPID_KEY: 'failed-delete-vapid-key',
-  INVALID_PUBLIC_VAPID_KEY: 'invalid-public-vapid-key',
-  USE_PUBLIC_KEY_BEFORE_GET_TOKEN: 'use-public-key-before-get-token',
-  PUBLIC_KEY_DECRYPTION_FAILED: 'public-vapid-key-decryption-failed'
-};
+export const enum ErrorCodes {
```
dwoffinden: Don't you usually give enums singular names, i.e. ErrorCode? As a type it describes what it means to be an ErrorCode value, this is no longer just an untyped bag of error codes.

mmermerkaya: Yeah that's true, thanks. Done.

Feiyang1:
nice!

Issue #1642, opened 2019-03-27
AsyncQueue is already failed: The transaction was aborted, so the request cannot be fulfilled.


larssn:
### Describe your environment

  * Operating System version:  Android, so the version varies
  * Browser version: Usually latest Chrome (73.0.3683.90)
  * Firebase SDK version: 5.9.3
  * Firebase Product: Firestore

### Describe the problem
We're seeing the following error more and more lately (via our error reporting service):
```
Error: Uncaught (in promise): Error: FIRESTORE (5.9.1) INTERNAL ASSERTION FAILED: AsyncQueue is already failed: The transaction was aborted, so the request cannot be fulfilled.
Error: FIRESTORE (5.9.1) INTERNAL ASSERTION FAILED: AsyncQueue is already failed: The transaction was aborted, so the request cannot be fulfilled.
    at fail (file:///android_asset/www/main.js:86695:11)
    at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.verifyNotFailed (file:///android_asset/www/main.js:94221:13)
    at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.enqueue (file:///android_asset/www/main.js:94173:14)
    at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.enqueueAndForget (file:///android_asset/www/main.js:94165:14)
    at FirestoreClient.push../node_modules/@firebase/firestore/dist/index.cjs.js.FirestoreClient.listen (file:///android_asset/www/main.js:105490:25)
    at Query.push../node_modules/@firebase/firestore/dist/in...
  at fail(./node_modules/@firebase/firestore/dist/index.cjs.js:155:1)
  at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.verifyNotFailed(./node_modules/@firebase/firestore/dist/index.cjs.js:7681:1)
  at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.enqueue(./node_modules/@firebase/firestore/dist/index.cjs.js:7633:1)
  at AsyncQueue.push../node_modules/@firebase/firestore/dist/index.cjs.js.AsyncQueue.enqueueAndForget(./node_modules/@firebase/firestore/dist/index.cjs.js:7625:1)
[...]

PR #1649, opened 2019-03-29
[DO NOT SUBMIT] Add more logging to abort.

mikelehen:
Not for submission. Just suggesting this to possibly help debug https://github.com/firebase/firebase-js-sdk/issues/1642

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -18,6 +18,7 @@
 import { assert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { debug } from '../util/log';
+import * as log from '../util/log';
 import { Deferred } from '../util/promise';
 import { SCHEMA_VERSION } from './indexeddb_schema';
 import { PersistencePromise } from './persistence_promise';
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -312,7 +314,8 @@ export class SimpleDbTransaction {
     this.transaction.oncomplete = () => {
       this.completionDeferred.resolve();
     };
-    this.transaction.onabort = () => {
+    this.transaction.onabort = (event) => {
```
larssn: Is this correct? `event` is not getting used.

mikelehen: Oh, that could be reverted...  I was just hacking this together real quick and am not planning to submit it as-is. 😄

larssn: Alright, just wanted to be sure where you'd want the error from. 😅 

I'm pushing a new release with your suggestions here; so we should have some kind of data fairly soon.

mikelehen: Ah, fantastic!  I think while writing this I was wondering the same thing (if there was error information in the `event`) but discovered that there was not.

PR #1650, opened 2019-04-01
Remove extra *.d.ts files from npm packaging

hsubox76:
Subdirectories in dist/ folders contain many *.d.ts files that don't need to be put into the npm package.  Set more specific "files" rules in each package.json to avoid including these.  Removed ".npmignore" (blacklist) files as they don't override package.json's "files" field (whitelist) and just add confusion.  We're only supposed to use one method, whitelist or blacklist.


mikelehen:
Do you know how I can "test" this and see what actually will end up in npm?

FWIW- (I haven't tested, but...) I'm a bit surprised `dist/*.*` does what we want, since `ls packages/firestore/dist/*.*` still includes all the .d.ts files. But maybe npm uses a different glob implementation?

Feiyang1:
Actually... we may need to keep everything in the /dist folder, except for /dist/test. 

The typing file in each scoped package is directly created from source by Typescript, so the root index.d.ts has references to d.ts files in the /src folder. For example, https://unpkg.com/@firebase/database@0.3.16/dist/index.d.ts
and
https://unpkg.com/@firebase/firestore@1.1.2/dist/index.d.ts

We could manually create an index.d.ts file for each scoped package and point to it, but I don't think we should do it (more maintenance).

mmermerkaya:
There are two tools that I know of which can help with generating a `.d.ts` bundle. They are [API Extractor](https://api-extractor.com/) (I haven't used this one) and [dts-bundle-generator](https://github.com/timocov/dts-bundle-generator/) (works great for a simple package, but I haven't tried it with anything big).

mikelehen:
FWIW- Looking at what we have in https://unpkg.com/@firebase/database@0.3.16/dist/index.d.ts, it seems pretty problematic.  We're exporting raw types (e.g. `Query`) which means we're exposing way more API surface area than is actually public (e.g. `Query.repo`), and so people could easily be using non-public parts of our API (without realizing it) and basically any change we make (to internal implementation stuff) could potentially break people. :-/

Firestore is in much better shape.  It looks like the only internal d.ts file it actually pulls in is https://unpkg.com/@firebase/firestore@1.1.2/dist/src/platform_browser/browser_init.d.ts which is empty.

I definitely think for v6 we need to rework all of this and be pretty careful about what we're exposing.

I'm okay dropping this work item for now, since it seems to be bumping into other tech debt we need to deal with first.

Feiyang1:
Yup, we will rework the type definitions for v6. @hsubox76 can you please close the PR and Issue?

PR #1657, opened 2019-04-03
Delete Hash base class in util

mmermerkaya:
Removes the `Hash` base class that is only used in `Sha1` in `@firebase/util`. It only has a single property and only one class extends from it, so I think it's fair to say that it's not very useful.

The actual reason of me doing this is that Rollup doesn't realize that extended classes that are transpiled to ES5 don't have side effects. Without this change, any package using Rollup that imports anything from `@firebase/util` (like `FirebaseError` or `ErrorFactory`) will also get `Sha1` and `Hash` classes included in their bundle. I suspect that Webpack does something similar but I haven't checked.

After this change, using Rollup, an iife build of a package that only exports `ErrorFactory` and `FirebaseError` from `@firebase/util` went from **1478 bytes** to **544 bytes** (minified + gzipped).


mmermerkaya:
There's a comment in `hash.ts` that says "See Sha1 and Md5 for sample implementations." but I think Md5 disappeared at some point. :man_shrugging:

Feiyang1:
I'm curious as to which part of the class is considered as sideeffect, the extending part? Does that mean all extended classes are not treeshakable by rollup?

I'm going to try webpack. but anyway webpack reads 'sideEffects' which can be set to false for util (we probably should).

Feiyang1:
Verified webpack handles it correctly.
Rollup has an open issue for this - https://github.com/rollup/rollup/issues/1890

Otherwise I'm fine with the PR.

mmermerkaya:
Thanks for checking Webpack, good to know that it works fine. The Rollup issues https://github.com/rollup/rollup/issues/1890 and https://github.com/rollup/rollup/issues/1763 have been open for a while. Also even without these issues, this class seems unnecessary anyway. So I'll go ahead and merge this.

PR #1662, opened 2019-04-05
Add "experimentalForceLongPolling" option.

mikelehen:


packages/firebase/index.d.ts
```diff
@@ -5853,6 +5853,19 @@ declare namespace firebase.firestore {
...
+    /**
+     * Forces the SDK’s underlying network transport (WebChannel) to use
+     * long-polling. Each response from the backend will be closed immediately
+     * after the backend sends data (by default responses are kept open in
+     * case the backend has more data to send). This avoids incompatibility
+     * issues with certain proxies, antivirus software, etc. that incorrectly
+     * buffer traffic indefinitely. Use of this option will cause some
+     * performance degradation though.
+     *
+     * This setting may be removed in a future release.
```
wilhuff: Should we invite users to discuss with us when they're using this?

The backend will have logs of when this is actually used, but if the rest of the world starts using this by default we won't know why otherwise.

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+- [feature] Added an `experimentalForceLongPolling` setting that may be helpful
```
wilhuff: Could we make the call to action more concrete here?

Maybe "... setting that can be used to work around proxies that prevent the Firestore client from connecting to its backend"?

mikelehen:
@hsubox76 Can you approve for the .d.ts updates?

FWIW- This option is actually only applicable to web, and so if there's a way to hide it from the node docs, let me know.  Else, it's not a big deal (you can actually set the option from node, it'll just be a no-op).

packages/firebase/index.d.ts
```diff
@@ -5853,6 +5853,22 @@ declare namespace firebase.firestore {
...
+     * case the backend has more data to send). This avoids incompatibility
+     * issues with certain proxies, antivirus software, etc. that incorrectly
+     * buffer traffic indefinitely. Use of this option will cause some
+     * performance degradation though.
+     *
+     * This setting may be removed in a future release. If you find yourself
+     * using it to work around a specific network reliability issue, please
+     * tell us about it in
+     * https://github.com/firebase/firebase-js-sdk/issues/1674.
+     */
```
hsubox76: Can you put a `@webonly` tag at the end of this comment? PR https://github.com/firebase/firebase-js-sdk/pull/1679 will add ability for the doc generation process to filter out symbols with this tag when generating the Node version of the reference docs.

PR #1669, opened 2019-04-08
Remove use of deprecated grpc.load() method

rsgowman:
This also requires a workaround to keep our tests passing.


packages/firestore/src/platform_node/load_protos.ts
```diff
@@ -38,5 +61,8 @@ export function loadProtos(): grpc.GrpcObject {
     root,
     file: 'google/firestore/v1/firestore.proto'
   };
-  return grpc.load(firestoreProtoFile, /*format=*/ 'proto', options);
+
+  let builder = ProtobufJS.newBuilder(options);
```
mikelehen: I think you can (and probably should) add 'protobufjs' to our package.json (as a devDependency, since it's only needed for tests) with a 6.x version, and then you can use the same options as in the loadProtos() method, and we'll be sure we're loading the protos in the actual SDK and the tests in the same manner.

rsgowman: Done.

rsgowman:
Ready for re-review. This is a meld of this plus #1699.

Note that the upgrade to protobufjs 6.x required a minor change to packages/firestore/src/remote/serializer.ts... it seems as though protobufjs now always returns this as a unit8array? But my knowledge of our typescript port in general is a little thin, so definitely take a close look to make sure that makes sense during your review.

I've been a little paranoid on the types (i.e. `const value: fieldValue.FieldValue = fieldValue.FieldValue.fromWhatever(x)`). I found that useful while developing (since the type system caught some of my silly errors early) but we can pare that down somewhat if you'd prefer.

mikelehen:
Sorry for the slow review on this, but there are a couple things I'm trying to figure out and I think I need to spend a bit of time debugging through this.  I'm not sure why serializer.ts needed to be changed.  And I think our hasTag() code is not being used like it was before and can perhaps be removed.  I'll poke at this more Monday and get back to you.

packages/firestore/src/remote/serializer.ts
```diff
@@ -1364,35 +1354,3 @@ export class JsonProtoSerializer {
     return FieldMask.fromArray(fields);
   }
 }
-
```
wilhuff: http://www.nooooooooooooooo.com/

PR #1677, opened 2019-04-09
Validate operator enum in Query.where() calls

thebrianchen:
Reference: b/64156656

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -59,7 +59,6 @@ import { AsyncQueue } from '../util/async_queue';
 import { Code, FirestoreError } from '../util/error';
 import {
   invalidClassError,
-  validateArgEnum,
   validateArgType,
   validateAtLeastNumberOfArgs,
   validateBetweenNumberOfArgs,
@@ -71,6 +70,7 @@ import {
   validateOptionalArgType,
   validateOptionalArrayElements,
   validateOptionNames,
+  validateStringEnum,
   valueDescription
 } from '../util/input_validation';
 import * as log from '../util/log';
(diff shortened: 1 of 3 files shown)


packages/firestore/src/util/input_validation.ts
```diff
@@ -292,6 +292,31 @@ export function validateNamedOptionalPropertyEquals<T>(
...
+/**
+ * Validates that the provided argument is part of the
+ *
+ * @param functionName Function making the validation call.
+ * @param enums Array containing all possible values for the enum.
+ * @param enumName Custom type name.
+ * @param position Position of the argument in `functionName`.
+ * @param argument Arugment to validate.
+ */
+export function validateArgEnum<T>(
```
mikelehen: I'd actually drop "Arg" (most of our validation functions don't have it).  This also isn't technically an enum (it's a string union or whatever), though I'm not wild about validateStringUnion().  Maybe `validateStringEnum()`?

packages/firestore/src/api/database.ts
```diff
@@ -1354,6 +1355,15 @@ export class Query implements firestore.Query {
     validateExactNumberOfArgs('Query.where', arguments, 3);
     validateArgType('Query.where', 'non-empty string', 2, opStr);
```
mikelehen: As mentioned below, I think we could drop this and rely on the new validation function to handle this case.

packages/firestore/src/util/input_validation.ts
```diff
@@ -152,6 +151,7 @@ export function validateArgType(
   position: number,
   argument: unknown
 ): void {
+  // console.log('validating: ', argument, 'as', type);
```
mikelehen: Please remove. 😄

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -800,6 +800,14 @@ apiDescribe('Validation:', persistence => {
       );
     });
 
+    validationIt(persistence, 'enum', db => {
+      const collection = db.collection('test') as any;
```
mikelehen: nit: maybe do `'foo' as any` below instead of `as any` here to minimize the scope of the untyped code we're dealing with (and to make the code better self-document exactly what we're doing that is invalid).

packages/firestore/src/util/input_validation.ts
```diff
@@ -292,6 +291,29 @@ export function validateNamedOptionalPropertyEquals<T>(
...
+export function validateStringEnum<T>(
+  functionName: string,
+  enums: string[],
+  position: number,
+  argument: unknown
+): void {
+  if (!enums.some(element => element === argument)) {
+    throw new FirestoreError(
+      Code.INVALID_ARGUMENT,
+      `Invalid value ${argument} provided to function ${functionName}() for its` +
```
mikelehen: I'd wrap ${argument} in quotes to clearly delineate it from the rest of the sentence and make the error clearer e.g. if the user accidentally passes an empty string they should see `Invalid value "" provided ...` instead of `Invalid value  provided ...`.

Or to make sure the error is relatively readable for non-strings as well, you could use the `valueDescription()` helper method that we use elsewhere (which wraps strings in quotes as well as massages other types into readable descriptions).

PR #1681, opened 2019-04-09
Update documentation to use yarn build 

thebrianchen:
* Adding `yarn build` command to main firebase-sdk-js repo
* Updating firestore documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 # Firebase Javascript SDK
 
 <!-- BADGES -->
+
 [![Build Status](https://travis-ci.org/firebase/firebase-js-sdk.svg?branch=master)](https://travis-ci.org/firebase/firebase-js-sdk)
 [![Build Status](https://saucelabs.com/buildstatus/firebase-oss)](https://saucelabs.com/u/firebase-oss)
+
 <!-- END BADGES -->
 
 The Firebase JavaScript SDK implements the client-side libraries used by
@@ -36,7 +38,7 @@ or [`N`](https://github.com/tj/n) to install and manage multiple node versions_
 
 In addition to Node.js we use `yarn` to facilitate multi package development.
 
-To install `yarn` follow the instructions listed on their website: 
+To install `yarn` follow the instructions listed on their website:
 https://yarnpkg.com/en/docs/install
 
 #### Java
@@ -85,16 +87,17 @@ them below.
 
 #### Authentication Support
 
-Visit the authentication config in your project and enable the `Anonymous` 
+Visit the authentication config in your project and enable the `Anonymous`
 sign-in provider to complete your project config.
 
 #### Automated Setup
 
-The remainder of the test setup can be done by running the following command at
+The remainder of the test setup can be done by running the following commands at
 the root of the package:
 
 ```bash
 yarn test:setup
+yarn build
 ```
 
 ### Running the tests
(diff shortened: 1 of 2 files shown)


packages/firestore/CONTRIBUTING.md
```diff
@@ -2,10 +2,30 @@
...
+environment.
+
+## Setting up an account for testing
+
+You will need a production project to test the Firestore SDK. You can create
+a new project by visiting the [Firebase Console](https://console.firebase.google.com/).
+Make sure that the project has Cloud Firestore enabled in the database section of the
+console.
+
+While setting up tests, you can choose the project manually, or specify the project directly with the following command:
```
wilhuff: Line length < 80?

README.md
```diff
@@ -85,16 +85,17 @@ them below.
...
 
 #### Automated Setup
 
-The remainder of the test setup can be done by running the following command at
+The remainder of the test setup can be done by running the following commands at
 the root of the package:
 
 ```bash
 yarn test:setup
+yarn build
```
wilhuff: Actually, doesn't this belong with Running the tests?

`yarn test:setup` is a one-time thing, but `yarn build` is something you do every time you've made changes, no?

Feiyang1: Correct. `yarn build` doesn't belong here. 
Can you please add a section below Install Dependencies, something like:
### Build The SDK
Once you installed dependencies, you can build the entire SDK by running the following at the root of the SDK:

```bash
$yarn build
```

packages/firestore/CONTRIBUTING.md
```diff
@@ -2,10 +2,31 @@
...
+
+## Setting up an account for testing
+
+You will need a production project to test the Firestore SDK. You can create
+a new project by visiting the 
+[Firebase Console](https://console.firebase.google.com/). Make sure that the 
+project has Cloud Firestore enabled in the database section of the console.
+
+While setting up tests, you can choose the project manually, or specify the 
+project directly with the following command:
```
Feiyang1: We should call out that it needs to be done at the root of the SDK, not within the firestore folder.

Feiyang1: Or better yet, can you just point to the Automated Setup section in the main README?

wilhuff: I think a better of of thinking about this is that we should link out to the main README for one-time setup, but keep instructions here for how to work with Firestore, including over time.

For example, when doing a `git pull` to come up to date with changes others have made, do we need to `yarn build` at the root?

thebrianchen:
Added more information to the "Automated Setup" section of the main README, and linked to there in the Firestore CONTRIBUTING file.

README.md
```diff
@@ -85,23 +85,29 @@ them below.
...
 
-Visit the authentication config in your project and enable the `Anonymous` 
+Visit the authentication config in your project and enable the `Anonymous`
 sign-in provider to complete your project config.
 
 #### Automated Setup
 
-The remainder of the test setup can be done by running the following command at
-the root of the package:
+The remainder of the test setup requiers choosing a test project. You can
```
wilhuff: s/requiers/requires/

README.md
```diff
@@ -112,16 +118,23 @@ an individual package directory.
 
 ## Building the SDK
 
+Once you have installed all the dependencies, you can build the entire SDK by
+running the following command the root of the SDK:
```
Feiyang1: at the root of the SDK.

I actually meant to create a new section 'Build the SDK'  immediately after 'Install Dependencies', because it's the natural flow. It's also good to have it before the Testing section, because testing requires building the projects first.

We probably don't need a new section, also it avoids having 2 'Build the SDK' sections. Can you please move this paragraph to the end of 'Install Dependencies' ?

PR #1685, opened 2019-04-11
Adding other enum types to be validated

thebrianchen:


packages/firestore/src/api/database.ts
```diff
@@ -1448,12 +1450,16 @@ export class Query implements firestore.Query {
     directionStr?: firestore.OrderByDirection
   ): firestore.Query {
     validateBetweenNumberOfArgs('Query.orderBy', arguments, 1, 2);
-    validateOptionalArgType(
-      'Query.orderBy',
-      'non-empty string',
-      2,
-      directionStr
-    );
+    if (!!directionStr) {
```
mikelehen: We usually just check for `undefined` explicitly, i.e. `if (directionSrc !== undefined) {`.

As-is, if you did `orderBy('foo', false)` it'll skip your validation.

The other option would be to have a `validateOptionalStringEnum()` version of the method that deals with optional arguments.  But given this would be the only usage it may not be worth it.

thebrianchen:
I was so focused on trying to replace the enums that I didn't notice it was pretty much implemented already. Going to close this PR.

Issue #1683, opened 2019-04-10
firebase.app().delete() throw error 


mohshraim:
### [REQUIRED] Describe your environment

  * Operating System version: Mac & Windows
  * Browser version: chrome and safari
  * Firebase SDK version: 5.9.3
  * Firebase Product:  firebase APP

### [REQUIRED] Describe the problem

`firebase.app().delete() ` was working before but now throw error as follow

![Screen Shot 2019-04-10 at 4 23 36 PM](https://user-images.githubusercontent.com/16401732/55882596-77452200-5bad-11e9-9365-7d8060d4a7f0.png)

Thanks 
Moh'd Shraim

hsubox76:
Hi, quick question: when you say "was working before", did you have a previous version of Firebase where this code worked, or was it working before you made a change to your code?

The reason I'm asking is that I do see this error, but it occurs in every version of Firebase as far back as I've gone.  The service that seems to be throwing the error is Functions, so is it possible that you started seeing this error after adding Functions to your app?

It's definitely a bug, I'm just trying to make sure if we need to revert something we broke or add something that was never there.  Thanks!

mohshraim:
Hello @hsubox76 

What i can confirm it was working on 5.0.4 and some after that... 
Which i started using it since this comment https://github.com/firebase/firebase-js-sdk/issues/947#issuecomment-398782452

Thanks

hsubox76:
I think I found the problem and the fix should go out in the next release, but if I'm correct it shouldn't have ever worked in any previous versions either, as long as the app was bringing in the `firebase.functions` component.  I was finding the same error message when I downgraded to 5.0.4. 
 In any case, try out the next release (Thursday)? and see if it fixes things for you.

mohshraim:
Thanks 
Will feed you back

mohshraim:
solved 
Thanks

PR #1687, opened 2019-04-11
Add internal delete method to functions

hsubox76:
The functions service lacks an `INTERNAL.delete()` method which I think is the cause of https://github.com/firebase/firebase-js-sdk/issues/1683. This adds a delete method which resolves a top-level promise on the service instance which should cancel any ongoing request promises.

### DIFF
--- a/packages/functions/src/api/service.ts
+++ b/packages/functions/src/api/service.ts
@@ -57,6 +57,8 @@ export class Service implements FirebaseFunctions {
   private readonly contextProvider: ContextProvider;
   private readonly serializer = new Serializer();
   private emulatorOrigin: string | null = null;
+  private cancelAllRequests: Promise<void>;
+  private deleteService: Function;
 
   /**
    * Creates a new Functions service for the given app and (optional) region.
@@ -68,11 +70,24 @@ export class Service implements FirebaseFunctions {
     private region_: string = 'us-central1'
   ) {
     this.contextProvider = new ContextProvider(app_);
+    // Cancels all ongoing requests when resolved.
+    this.cancelAllRequests = new Promise(resolve => {
+      this.deleteService = () => {
+        resolve();
+      }
+    });
   }
 
   get app(): FirebaseApp {
     return this.app_;
   }
+
+  INTERNAL = {
+    delete: (): Promise<void> => {
+      this.deleteService();
+      return Promise.resolve();
+    }
+  };
 
   /**
    * Returns the URL for a callable with the given name.
@@ -184,8 +199,14 @@ export class Service implements FirebaseFunctions {
 
     const response = await Promise.race([
       this.postJSON(url, body, headers),
-      failAfter(timeout)
+      failAfter(timeout),
+      this.cancelAllRequests
     ]);
+
+    // If service was deleted, interrupted response returns undefined.
+    if (!response) {
+      return;
+    }
 
     // Check for an error status, regardless of http status.
     const error = _errorForResponse(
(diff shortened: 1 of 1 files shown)


packages/functions/src/api/service.ts
```diff
@@ -68,12 +70,25 @@ export class Service implements FirebaseFunctions {
...
+        resolve();
+      };
+    });
   }
 
   get app(): FirebaseApp {
     return this.app_;
   }
 
+  INTERNAL = {
```
Feiyang1: Make function Service implement FirebaseService, so INTERNAL has proper type.
`import { FirebaseService } from '@firebase/app-types/private';`

hsubox76: Done.

packages/functions/src/api/service.ts
```diff
@@ -68,12 +71,25 @@ export class Service implements FirebaseFunctions {
...
+    delete: (): Promise<void> => {
+      this.deleteService();
+      return Promise.resolve();
```
bklimt: The fact that `delete` is supposed to return a `Promise` kind of implies to me that when the returned `Promise` is resolved, then the service should be finished being deleted. But that won't be the case here. `this.deleteService()` will have signaled the various function calls to cancel, but they won't be cancelled yet. Is that important?

It seems like it would be pretty annoying to have to implement some kind of countdown latch or something just to make sure all the functions have completed. So, what guarantees is `delete` supposed to make?

packages/functions/src/api/service.ts
```diff
@@ -184,9 +200,15 @@ export class Service implements FirebaseFunctions {
...
     const response = await Promise.race([
       this.postJSON(url, body, headers),
-      failAfter(timeout)
+      failAfter(timeout),
+      this.cancelAllRequests
     ]);
 
+    // If service was deleted, interrupted response returns undefined.
+    if (!response) {
+      return;
```
bklimt: Returning `undefined` seems unnecessarily harsh to me. Code relying on this function returning a `Promise<HttpsCallableResult>` would crash. (It's not clear to me whether it would return `undefined` or a `Promise` that is resolved with `undefined`, actually.) But either way, if we were okay with crashing, we wouldn't be bothering to implement `delete` to begin with. So, I think it would be better to throw a proper error, as documented. Something like this:

```
      throw new HttpsErrorImpl(
        'cancelled',
        'Firebase Functions instance was deleted.'
      );
```

hsubox76: Sounds good, done.

packages/functions/src/api/service.ts
hsubox76: Looking at Firestore and Auth, it seems like `delete` only guarantees that the deletion process has been started. Auth delete doesn't wait for anything before returning `Promise.resolve()`, Firestore shuts down some services and waits for those to complete but doesn't cancel pending promises.  I don't know if this is ideal, but it's consistent for now.

Made a small change to return the result of `deleteService()` directly because it reads clearer, but it doesn't solve anything as far as timing.

packages/functions/src/api/service.ts
```diff
@@ -184,9 +199,18 @@ export class Service implements FirebaseFunctions {
 
     const response = await Promise.race([
       this.postJSON(url, body, headers),
-      failAfter(timeout)
+      failAfter(timeout),
+      this.cancelAllRequests
     ]);
 
+    // If service was deleted, interrupted response returns undefined.
```
bklimt: This comment needs updated.

hsubox76: Done.

PR #1688, opened 2019-04-11
Fix test:node:persistence command and make travis run it.

mikelehen:
Since our code is TypeScript we use `--require ts-node/register` for mocha to
be able to run our code. We have this require in our global
`../../config/mocha.node.opts` file, but it apparently isn't applied before the
`--require test/util/node_persistence.ts` in our test:node:persistence
command-line. So I've had to add an explicit `--require ts-node/register` to
our command-line.

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -11,7 +11,7 @@
     "lint:fix": "tslint --fix -p tsconfig.json -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
     "prettier": "prettier --write 'src/**/*.js' 'test/**/*.js' 'src/**/*.ts' 'test/**/*.ts'",
     "test": "run-s lint test:all",
-    "test:all": "run-p test:browser test:node:persistence",
+    "test:all": "run-p test:browser test:node",
     "test:browser": "karma start --single-run",
     "test:browser:debug": "karma start --browsers=Chrome --auto-watch",
     "test:node": "TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file index.node.ts --opts ../../config/mocha.node.opts",
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -11,11 +11,11 @@
     "lint:fix": "tslint --fix -p tsconfig.json -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
     "prettier": "prettier --write 'src/**/*.js' 'test/**/*.js' 'src/**/*.ts' 'test/**/*.ts'",
     "test": "run-s lint test:all",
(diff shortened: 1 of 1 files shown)


packages/firestore/package.json
```diff
@@ -11,11 +11,11 @@
     "lint:fix": "tslint --fix -p tsconfig.json -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
     "prettier": "prettier --write 'src/**/*.js' 'test/**/*.js' 'src/**/*.ts' 'test/**/*.ts'",
     "test": "run-s lint test:all",
-    "test:all": "run-p test:browser test:node",
+    "test:all": "run-p test:browser test:node:persistence",
```
schmidt-sebastian: I'm not 100% sold that we should add this to our CI, after all it is really hacky. Plus, it looks like it just broke CI by hitting a memory limit. Your mileage may vary, and I let you decide what you do with this comment :)

mikelehen: Haha.  The memory issue has come up a couple times recently actually. I'm not convinced I caused it...

It's been long enough that I don't remember what was so hacky about this...  But I run these tests all the time as I'm developing and they seem to be 100% reliable.  I don't feel strongly about adding them to CI other than that it'll make sure they don't get broken again. 🤷‍♂️

schmidt-sebastian: The setup that we do is a bit unorthodox (re-defining globals to provide custom IndexedDb and WebStorage implementation that just happen to support the subset of features that we are using). But I have personally not see any flakes due to this. If the CI failure is unrelated to this change, then let's ship it. We can always revisit and increase someone's PR count (maybe make it mine next time?).

mikelehen: You win! It's no longer running out of memory, but it is still not reliable:
```
@firebase/firestore:   2 failing
@firebase/firestore:   1) IndexedDbLruDelegate
@firebase/firestore:        runs when the cache is large enough:
@firebase/firestore:      Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/firebase/firebase-js-sdk/packages/firestore/test/unit/local/lru_garbage_collector.test.ts)
@firebase/firestore:   
@firebase/firestore:   2) IndexedDbLruDelegate
@firebase/firestore:        caps sequence numbers to collect:
@firebase/firestore:      Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/travis/build/firebase/firebase-js-sdk/packages/firestore/test/unit/local/lru_garbage_collector.test.ts)
@firebase/firestore:   
@firebase/firestore: error Command failed with exit code 2.
@firebase/firestore: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
@firebase/firestore: ERROR: "test:node:persistence" exited with 1.
```

So I have reverted this line and we won't run the node persistence tests in CI.  That's possibly for the best since I was starting to feel bad about regressing my CI performance improvement only an hour later...

PR #1691, opened 2019-04-12
Fix firestore.Timestamp case.

mikelehen:


hsubox76:
You probably want to get https://github.com/firebase/firebase-js-sdk/blob/master/scripts/docgen/content-sources/node/toc.yaml too (node docs toc).

mikelehen:
Oops, thanks!  Updated.

PR #1695, opened 2019-04-13
Throw error in subsequent function calls after the client is shutdown

thebrianchen:
Context: b/38324305

Please take a look! Couldn't figure out how to cancel existing listeners and pending promises just yet, but will try to figure it out.

For some reason, the test passed the first time, but no longer passes with persistence = false. Gonna close this PR until I figure out what's wrong.

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -164,7 +164,7 @@ export class FirestoreClient {
    *     unconditionally resolved.
    */
   start(persistenceSettings: InternalPersistenceSettings): Promise<void> {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     // We defer our initialization until we get the current user from
     // setChangeListener(). We block the async queue until we got the initial
     // user and the initialization is completed. This will prevent any scheduled
@@ -210,7 +210,7 @@ export class FirestoreClient {
 
   /** Enables the network connection and requeues all pending operations. */
   enableNetwork(): Promise<void> {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     return this.asyncQueue.enqueue(() => {
       return this.syncEngine.enableNetwork();
     });
@@ -304,14 +304,13 @@ export class FirestoreClient {
    * Checks that the client has not been shutdown. Ensures that other methods on
    * this class cannot be called after the client is shutdown.
    */
-  private verifyClientRunning(): void {
-    if (this.isShutdown === false) {
-      return;
+  private verifyNotShutdown(): void {
+    if (this.isShutdown) {
+      throw new FirestoreError(
+        Code.FAILED_PRECONDITION,
+        'The client has already been shutdown.'
+      );
     }
-    throw new FirestoreError(
-      Code.FAILED_PRECONDITION,
-      'The client has already been shutdown.'
-    );
   }
 
   /**
@@ -483,7 +482,7 @@ export class FirestoreClient {
 
   /** Disables the network connection. Pending operations will not complete. */
   disableNetwork(): Promise<void> {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     return this.asyncQueue.enqueue(() => {
       return this.syncEngine.disableNetwork();
     });
@@ -519,7 +518,7 @@ export class FirestoreClient {
     observer: Observer<ViewSnapshot>,
     options: ListenOptions
   ): QueryListener {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     const listener = new QueryListener(query, observer, options);
     this.asyncQueue.enqueueAndForget(() => {
       return this.eventMgr.listen(listener);
@@ -528,14 +527,14 @@ export class FirestoreClient {
   }
 
   unlisten(listener: QueryListener): void {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     this.asyncQueue.enqueueAndForget(() => {
       return this.eventMgr.unlisten(listener);
     });
   }
 
   getDocumentFromLocalCache(docKey: DocumentKey): Promise<Document | null> {
-    this.verifyClientRunning();
+    this.verifyNotShutdown();
     return this.asyncQueue
       .enqueue(() => {
         return this.localStore.readDocument(docKey);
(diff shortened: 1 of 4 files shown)


mikelehen:
Did you close this on purpose?  I'm confused. 😄

mikelehen:
FYI- You can assign this to me whenever it's ready for review.  I will ignore it until then. 😄 

(In general, we use "assignee" as the singular indicator of who needs to take action on a given PR, so you can always open a PR but assign it to yourself (or nobody) until it's ready for review).

thebrianchen:
Sum of changes since original:
* Call `ensureClientConfigured()` in `delete()` to ensure error is thrown in subsequent API usage after client is shutdown
* Added `invokeShutdown` param to `withTestDbsSettings()` to allow for calling `shutdown()` in tests.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -297,6 +300,20 @@ export class FirestoreClient {
     return true;
   }
 
+  /**
+   * Checks that the client has not been shutdown. Ensures that other methods on
+   * this class cannot be called after the client is shutdown.
+   */
+  private checkClientRunning(): void {
```
mikelehen: naming nit: I think `verifyClientRunning()` might make it a bit clearer that the method throws and be consistent with similar patterns in our code (e.g. https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/database.ts#L850)

And to use consistent terminology (reducing ambiguity), consider `verifyNotShutdown()`

packages/firestore/src/core/firestore_client.ts
```diff
@@ -108,6 +108,7 @@ export class FirestoreClient {
   private lruScheduler?: LruScheduler;
 
   private readonly clientId = AutoId.newId();
+  private isClientShutdown = false;
```
mikelehen: nit: I think this could just be `isShutdown`

packages/firestore/src/core/firestore_client.ts
```diff
@@ -297,6 +300,20 @@ export class FirestoreClient {
...
+   * Checks that the client has not been shutdown. Ensures that other methods on
+   * this class cannot be called after the client is shutdown.
+   */
+  private checkClientRunning(): void {
+    if (this.isClientShutdown === false) {
+      return;
+    }
+    throw new FirestoreError(
+      Code.FAILED_PRECONDITION,
+      'The client has already been shutdown. Please restart the client.'
```
mikelehen: There's not really a way to "restart" the client...  You'd need to get a new instance of `Firestore` by calling `firebase.firestore()` again.  Perhaps just drop "Please restart the client."

packages/firestore/src/core/firestore_client.ts
```diff
@@ -474,6 +492,7 @@ export class FirestoreClient {
   shutdown(options?: {
     purgePersistenceWithDataLoss?: boolean;
   }): Promise<void> {
+    this.checkClientRunning();
```
mikelehen: FWIW- It's tempting to make shutdown() idempotent (i.e. do nothing if it's already been called).  I think that might also mean you wouldn't need to add the `withTestDbNoShutdown()` helper...  WDYT?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -560,12 +585,14 @@ export class FirestoreClient {
   }
 
   databaseId(): DatabaseId {
+    this.checkClientRunning();
```
mikelehen: FWIW- since this doesn't *do* anything (and doesn't correspond to a public API method), it would probably be fine to drop this call.  I don't feel strongly though.

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -144,6 +144,26 @@ export function withTestDb(
...
+export function withTestDbNoShutdown(
+  persistence: boolean,
+  fn: (db: firestore.FirebaseFirestore) => Promise<void>
+): Promise<void> {
+  return withTestDbsSettings(
+    persistence,
+    DEFAULT_PROJECT_ID,
+    DEFAULT_SETTINGS,
+    1,
+    /* invokeShutdown= */ false,
```
mikelehen: ❤️ for commenting boolean parameters!

thebrianchen:
Resolved comments.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -297,6 +300,20 @@ export class FirestoreClient {
     return true;
   }
 
+  /**
+   * Checks that the client has not been shutdown. Ensures that other methods on
+   * this class cannot be called after the client is shutdown.
+   */
+  private verifyClientRunning(): void {
+    if (this.isShutdown === false) {
+      return;
```
mikelehen: Sorry, I'm not sure why I didn't notice this before, but is there a reason this isn't just:
```
if (this.isShutdown) {
  throw new FirestoreError(
      Code.FAILED_PRECONDITION,
      'The client has already been shutdown.'
    );
}
```

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -31,6 +31,7 @@ import {
   apiDescribe,
   withTestCollection,
   withTestDb,
+  withTestDbNoShutdown,
```
mikelehen: leftover?

packages/firestore/src/core/firestore_client.ts
mikelehen: Just to double-check, did you see my `verifyNotShutdown()` suggestion? It looks like you didn't implement it and I'm not sure if that was an oversight or intentional. 😄

packages/firestore/src/core/firestore_client.ts
thebrianchen: my bad, that's a careless mistake from just blindly implementing changes and not considering how the rest of the code looks.

packages/firestore/test/integration/api/database.test.ts
thebrianchen: removed in subsequent commit, another careless error.

thebrianchen:
Okay! This time it should be actually ready for a re-requested review. Got too excited last time and resolved comments without pausing to consider other changes those initial changes would require.

PR #1699, opened 2019-04-15
Switch serializer tests to test round trips + remove deprecated grpc.load()

rsgowman:
* Round trip testing move this port closer to the others.
* Also changes the way protos in the test are constructed

Separately (though related) this PR also removes the deprecated grpc.load() method. The old tests relied on grpc.load() exposing the proto constructors (which the new protoLoader does not do.)

### DIFF
--- a/packages/firestore/test/unit/remote/node/serializer.test.ts
+++ b/packages/firestore/test/unit/remote/node/serializer.test.ts
@@ -16,7 +16,6 @@
  */
 
 import { expect } from 'chai';
-import * as Long from 'long';
 
 import { Blob } from '../../../../src/api/blob';
 import { PublicFieldValue as FieldValue } from '../../../../src/api/field_value';
@@ -41,7 +40,6 @@ import {
   SetMutation
 } from '../../../../src/model/mutation';
 import { DOCUMENT_KEY_NAME, FieldPath } from '../../../../src/model/path';
-import { loadProtos } from '../../../../src/platform_node/load_protos';
 import * as api from '../../../../src/protos/firestore_proto_api';
 import { JsonProtoSerializer } from '../../../../src/remote/serializer';
 import {
@@ -77,8 +75,6 @@ describe('Serializer', () => {
   const partition = new DatabaseId('p', 'd');
   const s = new JsonProtoSerializer(partition, { useProto3Json: false });
   const emptyResumeToken = new Uint8Array(0);
-  const protos = loadProtos();
-  const ds = protos['google']['firestore']['v1'];
 
   /**
    * Wraps the given query in QueryData. This is useful because the APIs we're
@@ -104,7 +100,7 @@ describe('Serializer', () => {
       proto: api.Value,
       tag: string,
       serializer: JsonProtoSerializer = s
-    ) {
+    ): void {
       const actual: api.Value = serializer.toValue(value);
 
       // Unlike the other ports, protobufjs doesn't seem to expose a
--- a/packages/firestore/test/unit/remote/node/serializer.test.ts
+++ b/packages/firestore/test/unit/remote/node/serializer.test.ts
@@ -96,29 +96,33 @@ describe('Serializer', () => {
     );
   }
 
-  describe('toValue', () => {
-    /**
-     * Verifies that the given object can be parsed as a datastore Value proto.
-     */
-    function expectValue(obj: unknown, tag: string): Chai.Assertion {
-      const proto = new ds.Value(obj);
-      expect(proto.value_type).to.equal(tag);
-      return expect(proto[tag]);
+  describe('convertsValue', () => {
+    function expectRoundTrip(value: fieldValue.FieldValue, proto: api.Value, tag: string, serializer: JsonProtoSerializer = s) {
+      const actual: api.Value = serializer.toValue(value);
+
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -200,16 +204,13 @@ describe('Serializer', () => {
...
-        const expected = { timestampValue: expectedJson[i] };
-        expect(obj).to.deep.equal(expected);
-        expectValue(obj, 'timestampValue').to.deep.equal(
-          new protos['google']['protobuf'].Timestamp(expectedJson[i]),
-          'for date ' + example
+        const value: fieldValue.FieldValue = new fieldValue.TimestampValue(
+          Timestamp.fromDate(examples[i])
         );
+        const proto: api.Value = {
+          timestampValue: (expectedJson[i] as unknown) as string
```
rsgowman: I'm not sure why the proto's timestampValue is typed as a string? But I also have no idea how protobufjs works.

I also needed the same trick down below in the bytes section, though at least in that case, it sort of makes sense... since at least some of the time this is actually a base64 encoded string.

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -337,220 +346,10 @@ describe('Serializer', () => {
...
-      // Compare the raw object representation rather than the Message because
-      // occasionally Jasmine will hang trying to generate the string form in
-      // the event of an inequality.
-      expect(new ds.Value(obj).toRaw(true, true)).to.deep.equal(
-        new ds.Value(expectedJson).toRaw(true, true)
-      );
-      // clang-format on
-    });
-  });
+      const proto: api.Value = expectedJson;
```
rsgowman: Regarding this comment which I deleted: "Compare the raw object ... in the event of an inequality".

After my change (I didn't test it before my change), I found this to be incorrect. If I introduced an inequality and ran the tests a few times, the tests always completed with the error string properly generated. As a result, I eliminated the special casing here.

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -337,208 +365,10 @@ describe('Serializer', () => {
...
-      const proto = { bytesValue: base64 };
-      const blob = Blob.fromUint8Array(new Uint8Array(bytes));
-      const serializer = new JsonProtoSerializer(partition, {
-        useProto3Json: true
-      });
-      expect(serializer.fromValue(proto)).to.deep.equal(
-        new fieldValue.BlobValue(blob)
-      );
+      expectRoundTrip(objValue, proto, 'mapValue');
+      // clang-format on
```
mikelehen: what is this?

rsgowman:
> I haven't done a fully detailed review yet but at a high-level I think I'm mostly on board with this (thanks a ton for tackling all of this). I left a few comments and I do have a bit of reservation about no longer constructing actual proto objects. As-is if you make a mistake in serializer.ts with the JSON we generate such that the JSON is wrong, the serializer tests won't notice as long as you make the mistake in both the serialization and deserialization methods. In _theory_ the JSON representation of the protos is strongly typed based on https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/protos/firestore_proto_api.d.ts#L1, but there are some nuances / hacks because grpc and webchannel speak slightly different dialects of JSON protos so it's still conceivable we could screw something up... though presumably integration tests would catch it.
> 
> If we wanted to round-trip through the _actual_ proto object that gRPC uses (similar to the way the tests did before), I think we could just load the protos directly in our tests via protobufjs instead of going through `@grpc/proto-loader` and then we'd get access to the constructors as before. Basically just copy the gprc/proto-loader loadSync() code: https://github.com/grpc/grpc-node/blob/073dc563858a2d74bd114e5cc50d02203123078f/packages/proto-loader/src/index.ts#L314
> 
> I'd be a fan of that change, but if you don't feel like taking it on, I can probably get behind this PR as-is or else I might take a stab at it myself...

Take a look at this: https://github.com/firebase/firebase-js-sdk/pull/1669. It's my first attempt; I don't like it as much, but I think it does what you're suggesting. I think it may be possible to meld these two together, so that we still get access to the underlying ctors, but also can eliminate all the fromFoo tests. If that's what you had in mind, then I'll tackle it.

mikelehen:
Oh!  Yes, #1669 is very much what I had in mind, but yeah, I think combining that with all the cleanup you did here would be great.  Basically make your "roundtrip" methods also construct the real proto instead of just the JSON representation (maybe even encode to bytes and back to JSON?).  That would make these tests really nice.

wilhuff:
Could we meld these two approaches together by essentially running the round trip twice? Once directly and once through protos? This would have the same benefit of reducing the volume of tests, while giving us the same net coverage.

Note that this is at least conceptually similar to what we're doing in the C++ serialization tests, where we're validating that our serialization generates bytes that libprotobuf can parse. There we modified the round trip validation to check both paths. Though the particulars are different here, could the same approach work?

wilhuff:
LOL. I wish I refreshed before posting.

rsgowman:
Closing in favour of #1669 (which contains the relevant bits from this pr)

Issue #443, opened 2018-01-16
Firestore persistence isAvailable() check should exclude iOS Safari, Android 4.4, etc.


nathanfan46:
### [REQUIRED] Describe your environment

  * Operating System version: iOS 9.3.5
  * Firebase SDK version: 4.8.0
  * Firebase Product: firestore

### [REQUIRED] Describe the problem
When enable persistence, iPad mini keep showing following error messages:
![unadjustednonraw_thumb_2](https://user-images.githubusercontent.com/914545/35009775-a57ed53a-face-11e7-917c-d54ee8be659b.jpg)

#### Steps to reproduce:
Just enable persistence

`AngularFirestoreModule.enablePersistence()`

#### Relevant Code:

```javascript
// TODO(you): code here to reproduce the problem
```

nikhilag:
I have been seeing the same issue for a while now. Is there any fix planned?

jshcrowthe:
@nathanfan46 Can you get us some code, or a sample repo to illustrate the problem?

mikelehen:
We rely on IndexedDB which is unfortunately not available in Safari 9.  The right "fix" here is to expand our isAvailable() check here [https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/simple_db.ts#L85] to detect Safari 9 so we gracefully fail to enable persistence instead of crashing...  We'll take a look at making this change in an upcoming release.

nathanfan46:
@jshcrowthe I created a simple ionic project with AngularFirestoreModule.enablePersistence()
you can reproduce the issue by running command `ionic cordova run ios -lcs`
https://github.com/nathanfan46/rttest1

@mikelehen Thank you for the response, it would be better to not having persistence than crash.

paulstelzer:
I can confirm the same issue on Android 4.4 (Android 5 and above all working fine)! If I enable persistence I get "Unhandled Promise rejection" and firestore is not working. I check for IndexedDb and he returns that this is available. Is there maybe another reason why enable persistence causes the issue?

sengoontoh:
following

nikhilag:
@mikelehen : Is this issue fixed? I have not been able to enable offline persistence in my web app due to this issue. Please help.

paulstelzer:
I don't know if this always fits, but you can check for `navigator.serviceWorker`. If this is available, you can activate persistence, otherwise not. This seems to work good for me

```
[...]

PR #1703, opened 2019-04-17
Exclude iOS Safari < 10.0 and Android < 4.4 when checking isAvailable() to enable persistence

thebrianchen:
from: b/80403585, #443 

Reference for IndexedDB support: https://caniuse.com/#feat=indexeddb

Old error message [iOS Safari (iOS 9.1)] (error):
```
[2019-04-17T00:04:02.109Z]  @firebase/firestore: – "Firestore (5.9.4): INTERNAL UNHANDLED ERROR: " – "transaction@[native code]↵open@http://localhost:8080/firebase-firestore.js:1:16616…"
```

New error message (warning):
```
Error enabling offline storage. Falling back to storage disabled: FirebaseError: [code=unimplemented]: This platform is either missing IndexedDB or is known to have an incomplete implementation. Offline persistence has been disabled.
```

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -162,16 +162,16 @@ export class SimpleDb {
     // ua = 'Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML,
     // like Gecko) Chrome/39.0.2171.71 Safari/537.36 Edge/12.0';
 
-    const iOS = !!ua.match(/iPad/i) || !!ua.match(/iPhone/i);
-    const webkit = !!ua.match(/WebKit/i);
-    const iOSSafari = iOS && webkit && !ua.match(/CriOS/i);
-    console.log('is iOSSafari?: ', iOSSafari);
+    // iOS Safari: Disable for users running iOS version < 10.
+    const iOSSafari =
+      (!!ua.match(/iPad/i) || !!ua.match(/iPhone/i)) && !!ua.match(/WebKit/i);
+    const isUnsupportedIOS = iOSSafari && SimpleDb.getIOSVersion(ua) < 10;
 
     if (
       ua.indexOf('MSIE ') > 0 ||
       ua.indexOf('Trident/') > 0 ||
       ua.indexOf('Edge/') > 0 ||
-      iOSSafari
+      isUnsupportedIOS
     ) {
       return false;
     } else {
@@ -185,6 +185,15 @@ export class SimpleDb {
     store: string
   ): SimpleDbStore<KeyType, ValueType> {
     return txn.store<KeyType, ValueType>(store);
+  }
+
+  /** Parse User Agent to determine iOS version. Returns 0 if not found. */
+  static getIOSVersion(ua: string): number {
+    const iOSVersionRegex = ua.match(/i(?:phone|pad|pod) os ([\d_]+)/i);
+    const version = iOSVersionRegex
+      ? iOSVersionRegex[1].split('_')[0]
+      : '0';
+    return parseInt(version);
   }
 
   constructor(private db: IDBDatabase) {}
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -163,7 +163,7 @@ export class SimpleDb {
     // like Gecko) Chrome/39.0.2171.71 Safari/537.36 Edge/12.0';
 
     // iOS Safari: Disable for users running iOS version < 10.
-    const iOSVersion = getIOSVersion(ua);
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -181,6 +191,25 @@ export class SimpleDb {
     return txn.store<KeyType, ValueType>(store);
   }
 
+  /** Parse User Agent to determine iOS version. Returns -1 if not found. */
```
mikelehen: Did you write these regex yourself or did you borrow them from a trustworthy source? 😄 If the former, I'll review them more closely...  If the latter, perhaps add a comment declaring where they came from.

packages/firestore/src/local/simple_db.ts
```diff
@@ -181,6 +191,25 @@ export class SimpleDb {
...
+  static getIOSVersion(ua: string): number {
```
mikelehen: Can you add a "// visible for testing" or similar comment to these methods to make it clear why they're public (and discourage dependencies from other code).

packages/firestore/src/local/simple_db.ts
thebrianchen: I combined them from several sources. 

iOS: [source](https://stackoverflow.com/questions/48009956/php-get-exact-ios-version-from-user-agent?rq=1): Original one was found in php, but javacsript does not support `\k` in regex queries, so I ran multiple manually to get the regex code right. Tested on iOS simulator with iphone + ipad.

Android: Looked at several android user agents, applied the same regex as iOS but with `.` instead of `_`.

packages/firestore/src/local/simple_db.ts
```diff
@@ -162,10 +162,20 @@ export class SimpleDb {
     // ua = 'Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML,
     // like Gecko) Chrome/39.0.2171.71 Safari/537.36 Edge/12.0';
 
+    // iOS Safari: Disable for users running iOS version < 10.
+    const iOSVersion = SimpleDb.getIOSVersion(ua);
+    const isUnsupportedIOS = 0 < iOSVersion && iOSVersion < 10;
+
+    // Android browser: Disable for userse running version < 4.4.
+    const androidVersion = SimpleDb.getAndroidVersion(ua);
+    const isUnsupportedAndroid = 0 < androidVersion && androidVersion < 4.4;
```
Feiyang1: Can't this be caught by feature detection?  https://caniuse.com/#feat=indexeddb shows there is no support for indexed db for android version < 4.4

Feiyang1:
Generally I think feature detection is preferable, since features can be polyfilled. But in this case, iOS has partial indexeddb support, so I'm okay with UA check (even in this case, the partial/buggy implementation can be polyfilled).

thebrianchen:
@Feiyang1 added general check for IndexedDB in `window`. Thanks for the tip!

packages/firestore/src/local/simple_db.ts
```diff
@@ -162,10 +162,20 @@ export class SimpleDb {
     // ua = 'Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML,
     // like Gecko) Chrome/39.0.2171.71 Safari/537.36 Edge/12.0';
 
+    // iOS Safari: Disable for users running iOS version < 10.
+    const iOSVersion = SimpleDb.getIOSVersion(ua);
+    const isUnsupportedIOS = 0 < iOSVersion && iOSVersion < 10;
+
+    // Checks whether IndexedDB is supported on the browser
+    const noIndexedDB =
```
Feiyang1: It is redundant since we already check it at the top of the function.

thebrianchen: Thanks for catching that

mikelehen:
I think the reason to check for Android 4.4 explicitly is that it's not (always?) fully supported: https://github.com/firebase/firebase-js-sdk/issues/443#issuecomment-374041939

thebrianchen:
@mikelehen  That's an interesting point. Looking the detailed description at [caniuse](https://caniuse.com/#feat=indexeddb), it seems that indexedDB is only supported for 4.4.3-4.4.4, but it doesn't include 4.4.1 and 4.4.2. Not sure if those two versions have partial support or no support. 

If there is no support for 4.4.1 and 4.4.2, the existing `typeof window === 'undefined' || window.indexedDB == null` check should have triggered and turned persistence off. The fact that it didn't do that leads me to infer that (1) the 4.4.1 and 4.4.2 have partial support or (2) the caniuse site is incorrect. 

Given that, should we just opt to return false for Android < 4.5?

mikelehen:
That is my inclination, yeah.  I'd rather err towards occasionally disabling persistence unnecessarily than occasionally crashing.

PR #1712, opened 2019-04-19
Add clearPersistence(), separate functionality out from shutdown()

thebrianchen:


packages/firestore/src/api/database.ts
```diff
@@ -406,6 +410,20 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+      throw new FirestoreError(
+        Code.FAILED_PRECONDITION,
+        'Persistence cannot be cleared while the client is running'
+      );
+    }
+    const persistenceKey = IndexedDbPersistence.buildStoragePrefix(
+      this.getDatabaseInfo()
+    );
+    const dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
+    return IndexedDbPersistence.clearPersistence(dbName);
```
schmidt-sebastian: I would suggest that you generate the `dbName` in IndexedDbPersistence and only pass the persistence key to `clearPersistence()`.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -732,6 +732,10 @@ export class IndexedDbPersistence implements Persistence {
     );
   }
 
+  static async clearPersistence(dbName: string): Promise<void> {
+    await SimpleDb.delete(dbName);
```
schmidt-sebastian: In this case, I would just `return SimpleDb.delete(dbName)`. This function will also no longer need to be marked `async`.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -928,6 +928,58 @@ apiDescribe('Database', persistence => {
...
+        await app2.delete();
+        await firestore2.clearPersistence();
+        const app3 = firebase.initializeApp(options, name);
+        const firestore3 = firebase.firestore!(app3);
+        await firestore3.enablePersistence();
+        const docRef3 = firestore3.doc(path);
+        await expect(
+          docRef3.get({ source: 'cache' })
+        ).to.eventually.be.rejectedWith(
+          'Failed to get document from cache. (However, this document may ' +
```
schmidt-sebastian: Does this pass if you only pass in the first part of this error message?

packages/firestore/src/api/database.ts
```diff
@@ -458,6 +480,8 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
       this._config.credentials,
       this._queue
     );
+
+    this.isClientRunning = true;
```
schmidt-sebastian: I'd move this line up, so that if we ever add any async code above, the flag will still be set to true even if the user manages to sneak in a call to `clearPersistence()` while we are waiting for an async operation to complete.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -697,7 +697,7 @@ export class IndexedDbPersistence implements Persistence {
     // we successfully deleted its metadata from IndexedDb.
     this.removeClientZombiedEntry();
     if (deleteData) {
-      await SimpleDb.delete(this.dbName);
+      await IndexedDbPersistence.clearPersistence(this.dbName);
```
schmidt-sebastian: I believe that we should now be able to delete this code pass and the deleteData argument.

packages/firestore-types/index.d.ts
```diff
@@ -216,6 +216,13 @@ export class FirebaseFirestore {
    */
   app: any;
 
+  /**
+   * Clears the persistence cache. This can only be called when the client is not running.
+   *
+   * @return A promise that is resolved once the persistence has been cleared.
+   */
+  clearPersistence(): Promise<void>;
```
schmidt-sebastian: FYI: This change also needs to be made in packages/firebase/index.d.ts :(

packages/firestore/test/integration/api/database.test.ts
thebrianchen: It's a direct string match passed in from FirestoreClient. 

I'm not sure how to partial string match when using `rejectedWith()`

schmidt-sebastian: According to https://stackoverflow.com/questions/45227354/how-can-i-use-rejectedwith-to-perform-an-exact-match-of-the-error-message `chai` performs a substring match already.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -928,6 +928,58 @@ apiDescribe('Database', persistence => {
...
+
+        // Verify data in cache persists after restart.
+        await app.delete();
+        const app2 = firebase.initializeApp(options, name);
+        const firestore2 = firebase.firestore!(app2);
+        await firestore2.enablePersistence();
+        const docRef2 = firestore2.doc(path);
+        const docSnap2 = await docRef2.get({ source: 'cache' });
+        expect(docSnap2.exists).to.be.true;
+        expect(docSnap2.data()).to.deep.equal({ foo: 'bar' });
```
schmidt-sebastian: Consider removing this from the current test and creating a new test that verifies that persisted data is available after a restart.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -480,6 +480,14 @@ export class FirestoreClient {
     return this.syncEngine.handleCredentialChange(user);
   }
 
+  clearPersistence(): Promise<void> {
+    const persistenceKey = IndexedDbPersistence.buildStoragePrefix(
+      this.databaseInfo
+    );
+    const dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
+    return IndexedDbPersistence.clearPersistence(dbName);
```
schmidt-sebastian: Consider wrapping the browser error in something nicer.

packages/firestore-types/index.d.ts
wilhuff: Unfortunately, you should hold both changes back for a separate PR.

These changes essentially make the API public and we need to wait until the API has been approved before we can do that. We can commit the implementation though, and that's worth doing so this isn't sitting around in a branch and rotting while we work through approval.

packages/firestore/src/api/database.ts
```diff
@@ -307,13 +308,16 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
   // are already set to synchronize on the async queue.
   private _firestoreClient: FirestoreClient | undefined;
 
+  private isClientRunning: boolean;
```
wilhuff: Boolean variables tend to named without the leading "is" (see e.g. fromCache, ssl, etc).

packages/firestore-types/index.d.ts
```diff
@@ -216,6 +216,13 @@ export class FirebaseFirestore {
    */
   app: any;
 
+  /**
+   * Clears the persistence cache. This can only be called when the client is not running.
```
wilhuff: A few thoughts on this comment:
  * In the `enablePersistence` comment above we say we "enable persistent storage". Let's describe this comparably.
  * We should also define "not running" concretely: see how `enablePersistence` defines the before state, but also after deleting the app.
  * We also need to adjust the comment for enablePersistence(), since `clearPersistence` is allowed.
  * You should indicate that we'll throw an exception if the client is still running
  * You should indicate under which conditions the promise can be rejected.

(And again, all in a separate PR.)

packages/firestore/src/api/database.ts
```diff
@@ -415,6 +432,16 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return this._firestoreClient as FirestoreClient;
   }
 
+  private getDatabaseInfo(): DatabaseInfo {
```
wilhuff: Since this isn't a getter exactly, consider naming this with a `make` or `create` prefix. Try to be consistent with other examples nearby (e.g. `makeCredentialsProvider`).

packages/firestore/src/core/firestore_client.ts
```diff
@@ -480,6 +480,20 @@ export class FirestoreClient {
     return this.syncEngine.handleCredentialChange(user);
   }
 
+  clearPersistence(): Promise<void> {
```
wilhuff: This seems as if its unused?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -732,6 +729,11 @@ export class IndexedDbPersistence implements Persistence {
     );
   }
 
+  static clearPersistence(persistenceKey: string): Promise<void> {
+    const dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
+    return SimpleDb.delete(dbName);
```
wilhuff: Prefer making this an async function and doing an `await` here to returning the promise directly.

I know this doesn't make much of a difference but makes it conceptually cleaner and easier to refactor later.

Basically we want to be in the habit of making things async and avoiding the Promises as much we can.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -928,6 +928,61 @@ apiDescribe('Database', persistence => {
     });
   });
 
+  (persistence ? it : it.skip)(
+    'maintains persistence after restarting app',
+    async () => {
+      await withTestDoc(persistence, async docRef => {
+        await docRef.set({ foo: 'bar' });
+        const path = docRef.path;
```
wilhuff: There's a lot of noise in this test and the one below that's related to re-recreating the Firestore instance. So much so that it obscures the "interesting" parts of the test.

Is it possible to extract the logic of capturing the details of recreating a Firestore instance? Or maybe make this a shared function near `withTestDoc` that can rely on its implementation details?

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -213,11 +215,11 @@ export function withTestDbsSettings(
       return wipeDb(dbs[0]).then(() =>
         dbs.reduce(
           (chain, db) =>
-            chain.then(
-              db.INTERNAL.delete.bind(this, {
-                purgePersistenceWithDataLoss: true
-              })
-            ),
+            chain
```
wilhuff: It's worth rewriting gunk like this as an async function rather than piling more crazy thens in here.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -301,7 +301,12 @@ function genericLocalStoreTests(
     localStore = new LocalStore(persistence, User.UNAUTHENTICATED);
   });
 
-  afterEach(() => persistence.shutdown(/* deleteData= */ true));
+  afterEach(async () => {
+    await persistence.shutdown();
+    await IndexedDbPersistence.clearPersistence(
+      persistenceHelpers.TEST_PERSISTENCE_PREFIX
```
wilhuff: Since we're repeating this phrase so often, maybe extract this into `persistenceHelpers.clearPersistence()`?

packages/firestore/test/integration/api/database.test.ts
thebrianchen: Decided on not doing so after an in-person conversation, mainly b/c we would only be reducing this by 3 lines.

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -105,6 +106,14 @@ function apiDescribeInternal(
   }
 }
 
+// TODO(b/131094514): Remove after clearPersistence() is updated in index.d.ts.
+export async function _clearPersistence(
```
mikelehen: You can rename this to clearPersistence() if you want.  We basically only ever use `_` as a way to hide methods on our public APIs (like FirebaseFirestore).

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -105,6 +106,14 @@ function apiDescribeInternal(
...
+  firestore: firestore.FirebaseFirestore
+): Promise<void> {
+  // tslint:disable-next-line:no-any
+  await (firestore as any).clearPersistence();
```
mikelehen: I think this should be `_clearPersistence()`?

packages/firestore/src/api/database.ts
```diff
@@ -492,13 +515,13 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
   INTERNAL = {
-    delete: async (options?: {
-      purgePersistenceWithDataLoss?: boolean;
-    }): Promise<void> => {
+    delete: async (): Promise<void> => {
       // The client must be initalized to ensure that all subsequent API usage
       // throws an exception.
       this.ensureClientConfigured();
-      return this._firestoreClient!.shutdown(options);
+      return this._firestoreClient!.shutdown().then(() => {
```
wilhuff: `await` instead of `then()`?

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -208,32 +217,19 @@ export function withTestDbsSettings(
...
-      }
-    );
-  });
+  const dbs = await Promise.all(promises);
+
+  try {
+    await fn(dbs);
+  } finally {
+    await wipeDb(dbs[0]);
+    for (let i = 0; i < dbs.length; i++) {
```
wilhuff: Since you don't otherwise need the index, consider using a `for..of` loop instead:

```ts
for (const db of dbs) {
  await db.INTERNAL.delete();
}
```

wilhuff:
Good to go!

PR #1715, opened 2019-04-22
Adding documentation for WriteBatch size limit

thebrianchen:
Reference: b/70222712

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -6047,8 +6047,8 @@ declare namespace firebase.firestore {
 
     /**
      * Creates a write batch, used for performing multiple writes as a single
-     * atomic operation. The maximum number of writes that can be passed to a commit operation or
-     * performed in a transaction is 500.
+     * atomic operation. The maximum number of writes that can be passed to a
+     * commit operation or performed in a transaction is 500.
      *
      * @return
      *   A `WriteBatch` that can be used to atomically execute multiple writes.
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -6047,7 +6047,8 @@ declare namespace firebase.firestore {
 
     /**
      * Creates a write batch, used for performing multiple writes as a single
-     * atomic operation.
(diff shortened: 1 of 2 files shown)


packages/firebase/index.d.ts
```diff
@@ -6047,7 +6047,8 @@ declare namespace firebase.firestore {
 
     /**
      * Creates a write batch, used for performing multiple writes as a single
-     * atomic operation.
+     * atomic operation. The maximum number of writes that can be passed to a
+     * commit operation or performed in a transaction is 500.
```
mikelehen: 1. This isn't really phrased in terms of our API (you build up a batch one write at a time rather than passing lots of writes to a "commit operation").

2. There's actually a subtlety here that I think is probably worth capturing...

Because of the way the protocol works (see https://github.com/googleapis/googleapis/blob/45199a22a4e47c8d53213d1ac4a5ad7f22382c56/google/firestore/v1beta1/write.proto#L49), any field transform counts as its own write.  This means that this will actually count as 3 writes:
```
bactch.update(doc, {
  a: 'normal',
  b: 'normal',
  c: firebase.firestore.FieldValue.serverTimestamp(),
  d: firebase.firestore.FieldValue.increment(1)
})
```

So maybe say something like:

> Creates a write batch, used for performing multiple writes as a single atomic operation. The maximum number of writes allowed is 500, but note that each usage `FieldValue.serverTimestamp()`, `FieldValue.arrayUnion()`, `FieldValue.arrayRemove()`, or `FieldValue.increment()` counts as an additional write.

thebrianchen:
Oh wow, that is a subtlety that is hard to glean from just skimming the documentation. Thanks for pointing it out and explaining it to me!

packages/firestore-types/index.d.ts
```diff
@@ -206,8 +212,14 @@ export class FirebaseFirestore {
   ): Promise<T>;
 
   /**
-   * Creates a write batch, used for performing multiple writes as a single
-   * atomic operation.
+   *  Creates a write batch, used for performing multiple writes as a single
```
mikelehen: Oooh, thanks for bringing this into sync.

PR #1717, opened 2019-04-22
add clearPersistence() to index.d.ts

thebrianchen:
Using this PR as a draft to get the correct description for clearPersistence().


packages/firebase/index.d.ts
```diff
@@ -6060,6 +6061,20 @@ declare namespace firebase.firestore {
      */
     app: firebase.app.App;
 
+    /**
+     * Clears the persistent storage.
+     *
+     * Must be called after the app is shutdown or before other methods (other
```
wilhuff: It's OK to mention the idea of "not started", which could give users some conceptual purchase on what we're trying to do here. What I was pointing out earlier was that describing this restriction only as not started without defining what that meant would be hard for someone who doesn't live and breathe this stuff to understand.

packages/firebase/index.d.ts
```diff
@@ -6060,6 +6061,20 @@ declare namespace firebase.firestore {
...
+     * than settings()) on startup. If the client is still running, an exception
+     * `code` will be thrown.
+     *
+     *   * failed-precondition: The client is still running in a browser tab.
```
wilhuff: Use bulleted lists only when you have more than one item in the list. I suggest:

> ..., an exception with a code of `failed-precondition` will be thrown.

packages/firebase/index.d.ts
```diff
@@ -6060,6 +6061,20 @@ declare namespace firebase.firestore {
...
+     *
+     * @return A promise that is resolved once the persistent storage has been
+     * cleared.
```
wilhuff: If IndexedDB fails to clear persistence, does that result in a rejected promise?

packages/firebase/index.d.ts
```diff
@@ -6060,6 +6061,20 @@ declare namespace firebase.firestore {
      */
     app: firebase.app.App;
 
+    /**
+     * Clears the persistent storage.
+     *
+     * Must be called while the client is not started (after the app is shutdown
+     * or when the app is first initialized). On startup, this method must called
+     * before other methods (other than settings()). If the client is still
+     * running, an exception with the a code of `failed-precondition` will be
```
wilhuff: s/the a/a/

packages/firebase/index.d.ts
```diff
@@ -6064,14 +6064,14 @@ declare namespace firebase.firestore {
     /**
      * Clears the persistent storage.
      *
-     * Must be called after the app is shutdown or before other methods (other
-     * than settings()) on startup. If the client is still running, an exception
-     * `code` will be thrown.
-     *
-     *   * failed-precondition: The client is still running in a browser tab.
+     * Must be called while the client is not started (after the app is shutdown
+     * or when the app is first initialized). On startup, this method must called
```
Feiyang1: this method must "be" called.

packages/firestore/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 # Unreleased
 - [changed] Firestore now recovers more quickly after network connectivity
```
wilhuff: This item was released as part of 1.3.3. Could you create that entry and move it there?

packages/firestore/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
...
   changes (airplane mode, Wi-Fi availability, etc.).
+- [feature] Added `clearPersistence()`, clears the persistent storage
```
wilhuff: Don't use commas in place of words like "which" and "that". Also "cached tests" should be "cached documents".

I think we can drop the "For applications ..." bit here.

This should refer to any issues we're addressing (in this case, 449).

PR #1720, opened 2019-04-24
Add types to ErrorFactory message parameters

mmermerkaya:
This makes it possible to have static type checking for error parameters.

Example:

```typescript
const enum ErrorCode {
  WITH_PARAMS = 'with-params',
  WITHOUT_PARAMS = 'without-params'
}

const ERROR_MAP: ErrorMap<ErrorCode> = {
  [ErrorCode.WITH_PARAMS]:
    'This message has {$param} and maybe also {$optParam}.',
  [ErrorCode.WITHOUT_PARAMS]: "This message doesn't have any parameters."
};

interface ErrorParams {
  [ErrorCode.WITH_PARAMS]: { param: string; optParam?: string };
}

const ERROR_FACTORY = new ErrorFactory<ErrorCode, ErrorParams>(
  'service',
  'serviceName',
  ERROR_MAP
);

ERROR_FACTORY.create(ErrorCode.WITH_PARAMS); // error
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, undefined); // error with strict
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, null); // error with strict
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {}); // error
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, { param: 'param' }); // good
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  optParam: 'optParam'
}); // good
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  otherParam: 'otherParam'
}); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS); // good
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, undefined); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, null); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, {}); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, { param: 'param' }); // error
ERROR_FACTORY.create('randomCode'); // error

// This works the same for all ErrorCode values.
const ERROR_FACTORY_WITHOUT_PARAMS = new ErrorFactory<ErrorCode>(
  'service',
  'serviceName',
  ERROR_MAP
);

// Same as WITHOUT_PARAMS before, only works with a single argument.
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS); // good
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, undefined); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, null); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, {}); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, {
  param: 'param'
}); // error
ERROR_FACTORY_WITHOUT_PARAMS.create('randomCode'); // error
```

### DIFF
--- a/packages/util/src/errors.ts
+++ b/packages/util/src/errors.ts
@@ -103,18 +103,25 @@ export class FirebaseError extends Error {
   }
 }
 
-export class ErrorFactory<ErrorCode extends string> {
+export class ErrorFactory<
+  ErrorCode extends string,
+  ErrorParams extends { readonly [K in ErrorCode]?: ErrorData } = {}
+> {
   constructor(
     private readonly service: string,
     private readonly serviceName: string,
     private readonly errors: ErrorMap<ErrorCode>
   ) {}
 
-  create(code: ErrorCode, data: ErrorData = {}): FirebaseError {
+  create<K extends ErrorCode>(
+    code: K,
+    ...data: K extends keyof ErrorParams ? [ErrorParams[K]] : []
+  ): FirebaseError {
+    const customData = data[0] || {};
     const fullCode = `${this.service}/${code}`;
     const template = this.errors[code];
 
-    const message = template ? replaceTemplate(template, data) : 'Error';
+    const message = template ? replaceTemplate(template, customData) : 'Error';
     // Service Name: Error message (service/code).
     const fullMessage = `${this.serviceName}: ${message} (${fullCode}).`;
 
@@ -123,14 +130,14 @@ export class ErrorFactory<ErrorCode extends string> {
     // Keys with an underscore at the end of their name are not included in
     // error.data for some reason.
     // TODO: Replace with Object.entries when lib is updated to es2017.
-    for (const key of Object.keys(data)) {
+    for (const key of Object.keys(customData)) {
       if (key.slice(-1) !== '_') {
         if (key in error) {
           console.warn(
             `Overwriting FirebaseError base field "${key}" can cause unexpected behavior.`
           );
         }
-        error[key] = data[key];
+        error[key] = customData[key];
       }
     }
 
--- a/packages/app/src/errors.ts
+++ b/packages/app/src/errors.ts
@@ -40,8 +40,10 @@ const ERRORS: ErrorMap<AppError> = {
     'Firebase App instance.'
 };
 
-const appErrors = new ErrorFactory<AppError>('app', 'Firebase', ERRORS);
+type ErrorParams = { [key in AppError]: { name: string } };
 
-export function error(code: AppError, args?: { [name: string]: any }) {
-  throw appErrors.create(code, args);
-}
+export const ERROR_FACTORY = new ErrorFactory<AppError, ErrorParams>(
+  'app',
+  'Firebase',
+  ERRORS
+);
--- a/packages/app/src/firebaseApp.ts
+++ b/packages/app/src/firebaseApp.ts
@@ -27,7 +27,7 @@ import {
   FirebaseAppInternals
 } from '@firebase/app-types/private';
 import { deepCopy, deepExtend } from '@firebase/util';
-import { error, AppError } from './errors';
+import { AppError, ERROR_FACTORY } from './errors';
 import { DEFAULT_ENTRY_NAME } from './constants';
 
 interface ServicesCache {
@@ -202,7 +202,7 @@ export class FirebaseAppImpl implements FirebaseApp {
    */
   private checkDestroyed_(): void {
     if (this.isDeleted_) {
-      error(AppError.APP_DELETED, { name: this.name_ });
+      throw ERROR_FACTORY.create(AppError.APP_DELETED, { name: this.name_ });
     }
   }
 }
--- a/packages/app/src/firebaseNamespaceCore.ts
+++ b/packages/app/src/firebaseNamespaceCore.ts
@@ -31,7 +31,7 @@ import {
 } from '@firebase/app-types/private';
 import { deepExtend, patchProperty } from '@firebase/util';
 import { FirebaseAppImpl } from './firebaseApp';
-import { error, AppError } from './errors';
+import { ERROR_FACTORY, AppError } from './errors';
 import { FirebaseAppLiteImpl } from './lite/firebaseAppLite';
 import { DEFAULT_ENTRY_NAME } from './constants';
 
@@ -104,7 +104,7 @@ export function createFirebaseNamespaceCore(
   function app(name?: string): FirebaseApp {
     name = name || DEFAULT_ENTRY_NAME;
     if (!contains(apps, name)) {
-      error(AppError.NO_APP, { name: name });
+      throw ERROR_FACTORY.create(AppError.NO_APP, { name: name });
     }
     return apps[name];
   }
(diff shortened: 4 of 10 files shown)


Feiyang1:
You can use the rest parameter with tuple type to conditionally require the second parameter. But it still doesn't work in this case, because `ErrorParams[K]` evaluates to `unknown` instead of `undefined`. I tried `ErrorParams[K] extends unknown ? [] : [ErrorParams[K]]`, but it doesn't work since everything extends unknown. : (

```javascript
export class ErrorFactory<
  ErrorCode extends string,
  ErrorParams extends Partial<{ readonly [K in ErrorCode]: ErrorData }> = {}
> {
  create<K extends ErrorCode>(
    code: K,
    ...data: ErrorParams[K] extends undefined ? [] : [ErrorParams[K]]
  ): FirebaseError {
  }
}
```

so to make `ErrorParams[k] extends undefined` work, I changed ErrorParams interface to define fields without parameters as undefined.
```javascript
interface AppErrorParams {
    [AppErrorCode.SOME_ERROR]: undefined,

    [AppErrorCode.WITH_PARAMS]: { param: string; optParam?: string };
    // See https://github.com/Microsoft/TypeScript/issues/8032
    [AppErrorCode.WITHOUT_PARAMS]: undefined;
}
```
Then it works as expected in all cases.
The result for your examples

```javascript

ERROR_FACTORY.create(ErrorCode.SOME_ERROR); // good
ERROR_FACTORY.create(ErrorCode.SOME_ERROR, {}); // error
ERROR_FACTORY.create(ErrorCode.SOME_ERROR, { param: 'param' }); // error

ERROR_FACTORY.create(ErrorCode.WITH_PARAMS); // error
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, { param: 'param' }); // good
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  optParam: 'optParam'
}); // good

ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS); // good
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, {}); // error


ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {}); // error

ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  otherParam: 'otherParam'
}); // error

// This one doesn't have any params.
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, { param: 'whatever' });
```

The solution is not ideal as we need to define fields without parameters as undefined, but it's 100% type safe. Let me know if you find better solutions...

mmermerkaya:
Thanks for taking a look!

Yeah, my first version of this was using a rest parameter as well, but it looks pretty bad in VSCode IntelliSense: `create(code: ErrorCode.FILE_NOT_FOUND, __1_0: { file: string; }): FirebaseError`. Second parameter appears as `__0_1` if you do `...[data]` and `data_0` if you do `...data`, so I tried to find another solution. You're right though, there's no better solution AFAICT.

For the `unknown` stuff, you can do this: `ErrorParams[K] extends ErrorData ? [ErrorParams[K]] : []` instead of `ErrorParams[K] extends undefined ? [] : [ErrorParams[K]]` and it works, without explicitly defining the errors without parameters. I think it's because TS doesn't ["track negated constraints in the false branch of a conditional type"](https://github.com/Microsoft/TypeScript/issues/26240#issuecomment-411236132). So when you do `ErrorParams[K] extends undefined ? [] : [ErrorParams[K]]` and `ErrorParams[K]` isn't `undefined`, TS doesn't infer that the type in the `false` side (`[ErrorParams[K]`) can't be `undefined`, and just gives it the type `unknown`. They should be able to fix this issue after finishing [negated types](https://github.com/Microsoft/TypeScript/pull/29317) I think.

In the end, I changed it to `...data: K extends keyof ErrorParams ? [ErrorParams[K]] : []`. After all that's what we really want to know, if the key is in the interface.

Changing to the rest parameter also makes this stricter, meaning if there's no entry in the interface for that `ErrorCode`, you can't define data for it. I tried to not make this a breaking change, but I suppose it's okay now that we're about to release a new major version. I'm happier with the result.

One last question is whether we should allow omitting `ErrorParams` or require an explicit `{}`? I'm okay with allowing it to be omitted but maybe it'll be more confusing this way, WDYT?

Here's how it works with the current version:

```ts
const enum ErrorCode {
  WITH_PARAMS = 'with-params',
  WITHOUT_PARAMS = 'without-params'
}

const ERROR_MAP: ErrorMap<ErrorCode> = {
  [ErrorCode.WITH_PARAMS]:
    'This message has {$param} and maybe also {$optParam}.',
  [ErrorCode.WITHOUT_PARAMS]: "This message doesn't have any parameters."
};

interface ErrorParams {
  [ErrorCode.WITH_PARAMS]: { param: string; optParam?: string };
}

const ERROR_FACTORY = new ErrorFactory<ErrorCode, ErrorParams>(
  'service',
  'serviceName',
  ERROR_MAP
);

ERROR_FACTORY.create(ErrorCode.WITH_PARAMS); // error
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, undefined); // error with strict
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, null); // error with strict
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {}); // error
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, { param: 'param' }); // good
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  optParam: 'optParam'
}); // good
ERROR_FACTORY.create(ErrorCode.WITH_PARAMS, {
  param: 'param',
  otherParam: 'otherParam'
}); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS); // good
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, undefined); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, null); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, {}); // error
ERROR_FACTORY.create(ErrorCode.WITHOUT_PARAMS, { param: 'param' }); // error
ERROR_FACTORY.create('randomCode'); // error

// This works the same for all ErrorCode values.
const ERROR_FACTORY_WITHOUT_PARAMS = new ErrorFactory<ErrorCode>(
  'service',
  'serviceName',
  ERROR_MAP
);

// Same as WITHOUT_PARAMS before, only works with a single argument.
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS); // good
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, undefined); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, null); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, {}); // error
ERROR_FACTORY_WITHOUT_PARAMS.create(ErrorCode.WITHOUT_PARAMS, {
  param: 'param'
}); // error
ERROR_FACTORY_WITHOUT_PARAMS.create('randomCode'); // error
```

packages/util/src/errors.ts
```diff
@@ -103,18 +103,25 @@ export class FirebaseError extends Error {
...
   constructor(
     private readonly service: string,
     private readonly serviceName: string,
     private readonly errors: ErrorMap<ErrorCode>
   ) {}
 
-  create(code: ErrorCode, data: ErrorData = {}): FirebaseError {
+  create<K extends ErrorCode>(
+    code: K,
+    ...data: K extends keyof ErrorParams ? [ErrorParams[K]] : []
```
Feiyang1: really nice. 👍

mmermerkaya: Thanks :slightly_smiling_face:

Do you think we can merge this today?

PR #1728, opened 2019-04-26
Rename multi-tab setting to synchronizeTabs

schmidt-sebastian:
Note: To be merged into next major release.


packages/firestore/src/api/database.ts
```diff
@@ -398,14 +398,30 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+            '`synchronizeTabs`. In a future release, the setting  will be removed ' +
+            'and it is recommended that you remove it from your ' +
+            'firestore.enablePersistence() call now.'
+        );
+      }
+      synchronizeTabs = objUtils.defaulted(
+        settings.synchronizeTabs !== undefined
+          ? settings.synchronizeTabs
+          : settings.experimentalTabSynchronization,
+        DEFAULT_SYNCHRONIZE_TABS
```
mikelehen: Way to go naming the default correctly ahead of time!

packages/firestore/src/api/database.ts
```diff
@@ -398,14 +398,30 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
           'any other methods on a Firestore object.'
       );
     }
+
+    let synchronizeTabs = false;
+
+    if (settings) {
+      if (settings.experimentalTabSynchronization !== undefined) {
+        log.error(
+          'The `experimentalTabSynchronization` setting has been renamed to ' +
```
mikelehen: This is probably super nitty but I don't think we usually use backticks within log messages.  I'd remove or replace with `'`

packages/firestore/src/api/database.ts
```diff
@@ -398,14 +398,30 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+            '`synchronizeTabs`. In a future release, the setting  will be removed ' +
```
mikelehen: extra space in `setting  will`

packages/firestore/src/api/database.ts
```diff
@@ -398,14 +398,30 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+            'and it is recommended that you remove it from your ' +
```
mikelehen: I might be overly paranoid but I worry "the setting will be removed" and "recommend that you remove it" could potentially be ambiguous and lead people to remove the setting entirely.  Perhaps:

> In a future release, 'experimentalTabSynchronization' will be removed, so you should update your firestore.enablePersistence() call to use 'synchronizeTabs' now.

packages/firestore/src/api/database.ts
schmidt-sebastian: I didn't like the message without any quoting, but I am fine just using single quotes.

packages/firestore/src/api/database.ts
schmidt-sebastian: Ugh, I thought I fixed this before sending this out. It's fixed now.

packages/firestore/src/api/database.ts
schmidt-sebastian: Sure, that works and reduces code duplication :)

packages/firestore/src/api/database.ts
schmidt-sebastian: Why do you assume the API proposal wasn't based on this name? :)

schmidt-sebastian:
Back to you Mike for merging! Thanls

mikelehen:
cc/ @Feiyang1 Another change for the next release...

PR #1730, opened 2019-04-29
Fix stack traces for custom equality mismatches.

mikelehen:
While messing with your serializer changes I kept having issues where the test failure wouldn't include a file / line number and so I couldn't tell for sure which expectation had failed.

Through debugging / reverse-engineering, I came up with this change which seems to work, though this doesn't seem to be a very well-documented area of chai.

### DIFF
--- a/packages/firestore/test/util/equality_matcher.ts
+++ b/packages/firestore/test/util/equality_matcher.ts
@@ -68,19 +68,20 @@ export function addEqualityMatcher(): void {
         originalFunction = originalFunction || _super;
         return function(...args): void {
           if (isActive) {
-            const [right, msg] = args;
-            const left = utils.flag(this, 'object');
+            const [expected, msg] = args;
+            utils.flag(this, 'message', msg);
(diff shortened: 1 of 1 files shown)


packages/firestore/test/util/equality_matcher.ts
```diff
@@ -69,10 +69,13 @@ export function addEqualityMatcher(): void {
         return function(...args): void {
           if (isActive) {
             const [right, msg] = args;
-            utils.flag(this, 'message', msg);
             const left = utils.flag(this, 'object');
 
-            new chai.Assertion(left).assert(
+            // These are used by chai to log the relevant stack trace.
```
rsgowman: Would using this make more sense: https://www.chaijs.com/api/plugins/#method_transferflags ?

Optional.

mikelehen: Good idea, done.  While testing this I also noticed that the actual/expected was backwards in the error message due to confusing chai docs, so I fixed that as well.

PR #1732, opened 2019-04-29
Clean up some serializer stuff.

mikelehen:
Sorry, I decided to review https://github.com/firebase/firebase-js-sdk/pull/1669 in the form of a PR. 😄 Thanks for doing the hard work to get protobufjs upgraded and the tests working again.  This is basically a cleanup pass to convince myself that everything was working right and simplify the tests. In particular:

1. The switch to protobufjs 6.x seems to have removed the need for hasTag() (and actually it had broken our usage of it since the tags ended up changing from snake_case to camelCase so they never matched). So I've removed hasTag() and set `oneofs: false` in protoLoaderOptions so the virtual "fooType" fields are no longer added at all.
2. 'seconds' in timestamps can and should always be serialized as a string, since it's an int64 value but we weren't doing that. I fixed it, which removed some of the special casing in your roundtrip testing.
3. I reworked the expect...RoundTrip methods into a single unified one. I think this is simpler (and reduced some repetition) but if you have concerns, let me know.
4. I re-added the disabled useProto3Json assert that you had removed from serializer.ts and instead tweaked the offending test so it doesn't try to round-trip through protobufjs.

### DIFF
--- a/packages/firestore/test/unit/remote/node/serializer.test.ts
+++ b/packages/firestore/test/unit/remote/node/serializer.test.ts
@@ -80,6 +80,9 @@ import {
 describe('Serializer', () => {
   const partition = new DatabaseId('p', 'd');
   const s = new JsonProtoSerializer(partition, { useProto3Json: false });
+  const proto3JsonSerializer = new JsonProtoSerializer(partition, {
+    useProto3Json: true
+  });
   const emptyResumeToken = new Uint8Array(0);
   const protos = loadRawProtos();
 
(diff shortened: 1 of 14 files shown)


packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -421,60 +324,42 @@ describe('Serializer', () => {
...
-
-      expectFieldValueJsonProtoRoundTrip(value, proto, serializer);
-      expectJsonProtoProtobufjsMessageRoundTrip(
-        proto,
-        protobufjsField,
-        'bytesValue'
-      );
-      expectFullRoundTrip(value, serializer);
+      const obj = serializer.toValue(new fieldValue.BlobValue(example));
+      expect(obj).to.deep.equal(expected);
```
rsgowman: Undoing the call to verifyFieldValueRoundTrip essentially means we never test the round trip of a fieldValue.BlobValue with proto3. Is this ok? (I'm not sure how/where we use proto3json, so I wasn't sure.)

mikelehen: Oh, that's a fair point.  I thought I was reverting this to the state in master but I forgot there's a "converts BlobValue from string" test in master that was removed in your PR...  That said, I think your PR wasn't testing the full round-trip either, since it included round-tripping through protobufJS which yields a UInt8Array instead of base64 string (and this is why you had to remove that assert in serializer.ts).

So I tweaked verifyFieldValueRoundTrip() to accept a `useProto3Json` option that uses the proto3 JSON serializer and skips the round-trip through protobufJS.

mikelehen:
Manually merged (with history) to rsgowman/grpc_load.

PR #1744, opened 2019-05-02
Use Sinon fake timers in tests that deal with time

mmermerkaya:
Fixes flaky tests. Removes some redundant code.

### DIFF
--- a/packages/installations/src/functions/delete-installation.test.ts
+++ b/packages/installations/src/functions/delete-installation.test.ts
@@ -51,13 +51,13 @@ describe('deleteInstallation', () => {
     deleteInstallationSpy = stub(
       deleteInstallationModule,
       'deleteInstallation'
-    ).callsFake(
-      () => sleep(50) // Request would take some time
-    );
+    ).callsFake(async () => {
+      await sleep(100); // Request would take some time
+    });
   });
 
   it('resolves without calling server API if there is no installation', async () => {
-    await expect(deleteInstallation(app)).to.eventually.be.fulfilled;
+    await expect(deleteInstallation(app)).to.be.fulfilled;
     expect(deleteInstallationSpy).not.to.have.been.called;
   });
 
@@ -68,7 +68,7 @@ describe('deleteInstallation', () => {
     };
     await set(appConfig, entry);
 
-    await expect(deleteInstallation(app)).to.eventually.be.fulfilled;
+    await expect(deleteInstallation(app)).to.be.fulfilled;
     expect(deleteInstallationSpy).not.to.have.been.called;
     await expect(get(appConfig)).to.eventually.be.undefined;
   });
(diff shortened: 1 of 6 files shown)


packages/installations/src/functions/delete-installation.test.ts
```diff
@@ -51,13 +51,13 @@ describe('deleteInstallation', () => {
     deleteInstallationSpy = stub(
       deleteInstallationModule,
       'deleteInstallation'
-    ).callsFake(
-      () => sleep(50) // Request would take some time
-    );
+    ).callsFake(async () => {
+      await sleep(100); // Request would take some time
```
dwoffinden: most pointless async+await in history :stuck_out_tongue:

packages/installations/src/functions/get-token.test.ts
```diff
@@ -96,10 +96,10 @@ const setupInstallationEntryMap: Map<
         }
       };
 
-      set(appConfig, entry);
+      await set(appConfig, entry);
 
-      // Finish pending request in 10 ms
-      sleep(50).then(() => {
+      // Finish pending request after 500 ms
+      sleep(500).then(async () => {
```
dwoffinden: why not await this sleep instead of then()ning it?

packages/installations/src/functions/get-token.test.ts
```diff
@@ -217,16 +217,11 @@ describe('getToken', () => {
...
-    await clear();
-  });
-
   describe('basic functionality', () => {
     for (const [title, setup] of setupInstallationEntryMap.entries()) {
       describe(`when ${title} in the DB`, () => {
-        beforeEach(() => {
-          setup(appConfig);
+        beforeEach(async () => {
+          await setup(appConfig);
```
dwoffinden: why not just `beforeEach(() => setup(appConfig))`?

packages/installations/src/functions/get-token.test.ts
dwoffinden: ok, I see :P

you don't need the async/await if you're deliberately not doing anything with the Promise that set() returns, though :stuck_out_tongue:

mmermerkaya:
I'd rather await every promise (except [this](https://palantir.github.io/tslint/rules/no-return-await/)) to make it more explicit that it is a promise, but it's fine either way.

dwoffinden:
> @dwoffinden I'd rather await every promise (except [this](https://palantir.github.io/tslint/rules/no-return-await/)) to make it more explicit that it is a promise, but it's fine either way.

Eh, as you like. It'd be shorter to just add the return type to the arrow function than to wrap it in async + await + curly brackets, though :stuck_out_tongue: 

const a = async () => {
  await sleep();
};
const b = (): Promise<void> => sleep();

PR #1756, opened 2019-05-07
Separate Saucelabs cross-browser tests by package

hsubox76:
Setting up our Saucelabs test to run tests for each package separately and sequentially, making it easier to identify which package/browser combination has failed.

Each run of all tests has one build number and counts as a build in Saucelabs' dashboard.  Each test has a separate name based on package name and browser environment.  Tests are run on one package at a time, on 5 browsers in parallel.

Will try to address some of the test failures in future PRs.


packages/installations/karma.conf.js
```diff
@@ -17,11 +17,15 @@
 
 const karmaBase = require('../../config/karma.base');
 
+const files = ['src/**/*.test.ts'];
+
 module.exports = function(config) {
   config.set({
     ...karmaBase,
-    files: ['src/**/*.test.ts'],
+    files: files,
```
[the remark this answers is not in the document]
mmermerkaya: Thanks :slightly_smiling_face:

config/karma.saucelabs.js
```diff
@@ -16,42 +16,96 @@
...
+  // Chrome_Android: appiumLauncher('Chrome', 'Android Emulator', 'Android', '6.0')
+};
+
+/**
+ * Any special options per package.
+ */
+const packageConfigs = {
+  messaging: {
+    // Messaging only supports these two browsers.
+    browsers: ['Chrome_Windows', 'Firefox_Windows']
```
mmermerkaya: Can we add Edge as well (as long as it's Edge 17+)? I don't know if the tests will pass, it's been a while since I checked Edge, but they should (in theory) pass.

https://caniuse.com/#feat=push-api

hsubox76: They did pass actually! But Edge wasn't listed as supported in [ENVIRONMENTS.md](ENVIRONMENTS.md) so I left it out.  Should I add it to the ENVIRONMENTS.md table too?

mmermerkaya: We added it to the new environments doc, but I'm not sure if that's public yet. I'm guessing that we'll replace that file with a link to the doc, so no need to change it for now IMO.

hsubox76: Added Edge to the messaging browser tests list.

config/karma.saucelabs.js
```diff
@@ -16,42 +16,96 @@
  */
 
 const argv = require('yargs').argv;
-const glob = require('glob');
-const karma = require('karma');
 const path = require('path');
 const karmaBase = require('./karma.base');
 
+// karma.conf.js test configuration file to run.
+const testFile = argv['testfile'];
```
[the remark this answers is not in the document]
hsubox76: Done.

scripts/run_saucelabs.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+const { spawn } = require('child-process-promise');
+const yargs = require('yargs');
+const glob = require('glob');
+
+// Check for 'files' flag to run on specified karma.conf.js files instead of
+// on all files.
+const { files } = yargs
+  .option('files', {
```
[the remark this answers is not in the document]
hsubox76: Done.

hsubox76:
- Set it to only run npm `pretest` scripts right before the package it is for.
- Make sure it exits with 1 if any package fails (so CI will show an overall failure)

Issue #1754, opened 2019-05-06
Angular with SSR firebase bug


vforv:
There is a bug with firebase@5.11.1

When install last version of firebase getting error on my server:

```
/dist/server.js:75602
if (self && 'firebase' in self) {
^
ReferenceError: self is not defined
```

With version 5.7.3 works fine.

I am using Angular 7 with server side rendering and @angular/fire": "^5.1.3" pkg as well.

Feiyang1:
Thanks for reporting the issue! We didn't consider SSR use case when adding the condition. We will fix it in the next release. 

Can you please downgrade firebase to <= 5.10.0 as a workaround?

amit-ahire:
/var/www/html/project-name/dist/server.js:102897
if (self && 'firebase' in self) {
^
ReferenceError: self is not defined

I'm also Facing the same issue after I updated the packages
Now I have angular v7.2.14, firebase v5.11.1 & @angular/fire v5.1.3
Earlier it was working perfectly

mezzo9:
Why is this closed?
Will you please let us know if there have there been any solutions?
I tried the latest version, and all the other ones back to 5.0.0 and none of them work. the build:ssr works just fine, but serve:ssr throws the error mentioned above.

Feiyang1:
Github automatically closed it when I mentioned it is fixed in https://github.com/firebase/firebase-js-sdk/pull/1757. Reopening it till it's released.

This check is only added in 5.11.0, so previous versions should work, or you are facing a different issue.

mezzo9:
Thanks for your clarification.
You are correct in that this change is introduced in this version. When using other versions, I get different errors.  which didn't make sense to open issues for those. if we can fix this one, then we are all happy :)

Is there a way I can use the fix in #1757 before release if it is going to take a while to release?
Thank you very much for your help.

Feiyang1:
You can use `npm install firebase@canary`. We should have a release soon though.

mezzo9:
Thank you very much.
After uninstalling the latest, and then installing the canary version, That error went away.
however, it was replaced by a new one:
Please note it builds fine, but when SSR app, this happens.

var major = parseInt(firebase_app__WEBPACK_IMPORTED_MODULE_6__["SDK_VERSION"].split('.')[0]);
                                                                              ^

TypeError: Cannot read property 'split' of undefined

mezzo9:
Found a temporary solution in case others are interested too.
[...]

PR #1757, opened 2019-05-07
Revamp environment check for firebase app 

Feiyang1:
1. Proper browser env detection. Fixed https://github.com/firebase/firebase-js-sdk/issues/1754
2. Delay Nodejs detection and warning util initializeApp, so people have a chance to suppress it by setting logLevel in @firebase/logger


Feiyang1:
Warning messages are shown by default, so they don't need to do anything unless they want to disable  it. Is it your concern?
The FirebaseLite message can't be suppressed because it will execute before people have a chance to set log level. It is desirable because having firebase imported twice is always the wrong thing to do.

PR #1762, opened 2019-05-07
Add OnVersionChange listener to Firestore

thebrianchen:
Calling clearPersistence() in any tab will cause other open tabs to be shutdown.


packages/firestore/src/core/firestore_client.ts
```diff
@@ -470,6 +470,12 @@ export class FirestoreClient {
             }
           }
         });
+
+        // If the persistence database has been deleted in another client,
+        // shutdown the current client.
```
wilhuff: The phrase "persistence database" is imprecise. Also, a comment here could help give larger context, so maybe rephrasing in terms of the user action might make this clearer. Something like:

> When a user calls `clearPersistence()` in one client all other clients need to shut down to allow the delete to succeed.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -387,6 +398,14 @@ export class IndexedDbPersistence implements Persistence {
     return primaryStateListener(this.isPrimary);
   }
 
+  setDatabaseDeletedListener(
+    databaseDeletedListener: DatabaseDeletedListener
+  ): void {
+    this.databaseDeletedListener = async () => {
```
wilhuff: Same feedback as below. Is there any value in retaining the `DatabaseDeletedListener` here? The actual events are emitted by IndexedDB so it seems like this is materially different from `primaryStateListener` above.

packages/firestore/src/local/simple_db.ts
```diff
@@ -212,7 +212,23 @@ export class SimpleDb {
...
+  constructor(private db: IDBDatabase) {
+    this.db.onversionchange = (event: IDBVersionChangeEvent) => {
+      return this.versionChangeListener(event);
+    };
+  }
+
+  setVersionChangeListener(
+    versionChangeListener: (event: IDBVersionChangeEvent) => Promise<void>
+  ): void {
+    this.versionChangeListener = async event => {
```
wilhuff: Do we need to manage our own `versionChangeListener` or could we just have this setter assign to `this.db.onversionchange` directly?

Or does this added indirection help out somehow?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -377,6 +385,14 @@ export class SpecBuilder {
     return this;
   }
 
+  assertIsShutdown(): this {
```
wilhuff: Other "assertions" in this package use the "expect" prefix. Is this sufficiently different that it warrants inconsistency? If so could you document that here?

wilhuff:
Try running the JSON export to verify that continues to work. Run it with `packages/firestore/test/unit/generate_spec_json.sh json-dir`, where `json-dir` is any directory, though most usefully check writing into the `Firestore/Example/Tests/SpecTests/json` directory in the iOS SDK and `firebase-firestore/src/test/resources/json` in the Android SDK.

mikelehen:
Nice work on the spec test! 👍

PR #1763, opened 2019-05-07
Add Installations to Firebase package

mmermerkaya:


Feiyang1:
We haven't created an entry point for installations in firebase package. A couple of things we need to do together with this change:
1. create an entry point for installations in firebase package
2. update the toc.yaml, so reference docs are added to the dev site. https://github.com/firebase/firebase-js-sdk/commit/932d7bea1568e6923bdd617d3830404f5578af84#diff-abd7b818fc44968da080027f237cb27eR149

Feiyang1:
Thanks. Just a question before it's merged - Are you targeting a specific time to release it, or anytime from now?

packages/firebase/index.d.ts
```diff
@@ -1220,6 +1234,39 @@ declare namespace firebase.app {
...
+     * @return Firebase Installation ID
+     */
+    getId(): Promise<string>;
+
+    /**
+     * Returns an Authentication Token for the current Firebase Installation.
+     *
+     * @return Firebase Installation Authentication Token
+     */
+    getToken(): Promise<string>;
```
andirayo: Why is this not called #getAuthToken()?

The term "token" is heavily overloaded in the space of Firebase and particularly with the background of Instance-ID, where a "token" referred to an FCM-Token, which is an FCM-address that can be used to send notifications to a Firebase Installation.
This token will not be able to do that, but is a dedicated authentication-token.
Thus, I would strongly recommend to add this distinction to the method name for clarification.

packages/firebase/index.d.ts
```diff
@@ -1220,6 +1234,39 @@ declare namespace firebase.app {
...
+   * {@link firebase.installations `firebase.installations()`}.
+   */
+  export interface Installations {
+    /**
+     * Creates a Firebase Installation if there isn't one for the app and
+     * returns the Installation ID.
+     *
+     * @return Firebase Installation ID
+     */
+    getId(): Promise<string>;
```
andirayo: Why is this not called #getFid()?

In the context of Firebase Installations, ID might be clear, but still there are so many IDs (Android-IDs, Gaia-IDs, Project-IDs, Instance-IDs, and so on and so forth) that I would like to have a distinctive name.

packages/firebase/index.d.ts
```diff
@@ -1220,6 +1234,39 @@ declare namespace firebase.app {
...
+     * Returns an Authentication Token for the current Firebase Installation.
+     *
+     * @return Firebase Installation Authentication Token
+     */
+    getToken(): Promise<string>;
+
+    /**
+     * Deletes the Firebase Installation and all associated data.
+     */
+    delete(): Promise<void>;
```
andirayo: Why is this not called #deleteInstallation()?

Like getId above this might be clear in the context of Firebase Installations, but still why not be explicit about what is happening here?

packages/firebase/installations/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
andirayo: why 2017?

packages/firebase/package.json
```diff
@@ -42,6 +42,7 @@
     "@firebase/database": "0.4.0",
     "@firebase/firestore": "1.3.0",
     "@firebase/functions": "0.4.6",
+    "@firebase/installations": "0.1.0",
```
andirayo: out of curiosity:
when do we plan to go to "1.0.0"?

packages/installations-types/index.d.ts
```diff
@@ -18,7 +18,23 @@
...
+   * @return Firebase Installation ID
+   */
   getId(): Promise<string>;
+
+  /**
+   * Returns an Authentication Token for the current Firebase Installation.
+   *
+   * @return Firebase Installation Authentication Token
+   */
   getToken(): Promise<string>;
```
andirayo: naming: as before...
the name #getToken() is asking for confusion.

packages/firebase/installations/index.ts
mmermerkaya: Copy paste mistake. Fixed.

packages/firebase/package.json
mmermerkaya: Explained [here](http://doc/1jWfkOoxk2mtjD0H62SUxFfAAm4c8GXSdLvIkdwDC6b4#heading=h.kstmuj6ckw41).

mmermerkaya:
@andirayo I resolved all questions about naming, as we're discussing that elsewhere. LMK if you have any comments about the documentation.

@Feiyang1 I don't mind, there's no rush. But let's not merge this until the naming discussion is done.

mmermerkaya:
Since the naming discussions are done, I think we can submit this. Any objections @Feiyang1 @andirayo?

Feiyang1:
SGTM

Issue #914, opened 2018-06-11
ENVIRONMENTS.md is missing Firestore and Functions


nicolasgarnier:
https://github.com/firebase/firebase-js-sdk/blob/master/ENVIRONMENTS.md lists the supported browsers for most Firebase features but it's currently missing Firestore and the Cloud Functions's Callable Functions feature. Would be cool to add them in.

Especially would be nice to know if some features are working or not depending on the browser. For instance Firestore's offline support.

nicolasgarnier:
I hate you Bot. My issue is great :D

jshcrowthe:
I'd love a PR 😉

Feiyang1:
fixed by https://github.com/firebase/firebase-js-sdk/pull/1772

Issue #880, opened 2018-05-30
5.0.4 Does not compile in node environment / typescript definitions..


swftvsn:
### [REQUIRED] Describe your environment

  * Operating System version: High Sierra
  * Firebase SDK version: 5.0.4
  * Firebase Product: core

#### Steps to reproduce:
Can't compile Node project. Compiles fine with 4.13.1.

Update: just tried this again, and 5.0.0 works, but 5.0.1, 5.0.2, 5.0.3 and 5.0.4 all fail with that message.
#### Relevant Code:

```bash
node_modules/firebase/index.d.ts(578,36): error TS2304: Cannot find name 'ServiceWorkerRegistration'.
```

colincapurso:
I have the same problem. Had to go back to firebase@5.0.0 to be able to deploy my functions.

curvecode:
I also have same problem when upgrade to firebase@5.0.4 when compile for my probject.

henriquecf:
I am getting an error: error TS2339: Property 'auth' does not exist on type 'FirebaseNamespace'

forest:
+1

kglowacki:
+1

casperstorm:
I had the same problem:

```
error TS2304: Cannot find name 'ServiceWorkerRegistration'.
578 useServiceWorker(registration: ServiceWorkerRegistration): void;
```

A fix, as mentioned above, is to downgrade `firebase` to `5.0.0`.

WCarrollSTO:
After updating to 5.1.0, the error still exists, but the line number is different.

`node_modules/firebase/index.d.ts:600:36 - error TS2304: Cannot find name 'ServiceWorkerRegistration'.`

mmermerkaya:
A workaround is to add `"dom"` to the `"lib"` compiler option in your `tsconfig.json`.

@jshcrowthe This is not an issue with Messaging. Our packages [assume that DOM types exist](https://github.com/firebase/firebase-js-sdk/blob/63e141488d9025eb5433b97132e6266a87f35310/config/tsconfig.base.json#L8). The problem is that importing `firebase/app` [imports typings of all packages](https://github.com/firebase/firebase-js-sdk/blob/63e141488d9025eb5433b97132e6266a87f35310/packages/firebase/app/package.json#L6).

Some of our packages (like Messaging) only work in the browser. These packages (or their types) should not be imported in a Node environment.

WCarrollSTO:
[...]

PR #1772, opened 2019-05-09
point ENVIRONMENTS.md to devsite

Feiyang1:


mmermerkaya:
Can we put this in README instead, and delete this file?

hsubox76:
Seems like a good idea to put in README.md unless there are links to this page from somewhere or there is a convention for people to check for a file with this name.

mmermerkaya:
Not sure about links to the file, but I'm pretty sure that there's no such convention. And I know that at least one user didn't notice the file :slightly_smiling_face: https://github.com/firebase/firebase-js-sdk/issues/880#issuecomment-401910371

Feiyang1:
Moved the section to README

mmermerkaya:
BTW, I think this PR fixes https://github.com/firebase/firebase-js-sdk/issues/914.

Feiyang1:
Yay! One more closed issue.

PR #1773, opened 2019-05-09
Fix browser detection

mmermerkaya:
Web Workers are browser too.

### DIFF
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -68,7 +68,7 @@ export function isNode(): boolean {
  * Detect Browser Environment
  */
 export function isBrowser(): boolean {
-  return typeof window !== 'undefined';
+  return typeof self === 'object' && self.self === self;
 }
 
 /**
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -19,7 +19,7 @@ import { CONSTANTS } from './constants';
 
 /**
  * Returns navigator.userAgent string or '' if it's not defined.
- * @return {string} user agent string
+ * @return user agent string
  */
 export function getUA(): string {
   if (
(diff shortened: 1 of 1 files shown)


mmermerkaya:
BTW, should we merge these `isX` functions into a single `detectEnvironment` function?

packages/util/src/environment.ts
```diff
@@ -68,13 +67,13 @@ export function isNode(): boolean {
  * Detect Browser Environment
  */
 export function isBrowser(): boolean {
-  return typeof window !== 'undefined';
+  return typeof self === 'object' && self.self === self;
```
Feiyang1: Just curious why checking self.self? Does some none browser env have self, but not self.self, or just to handle the case where someone defined a variable named self in none browse env? 

The intention of this function was just to check if it's a browser window env.
My intuition is that webworker probably needs a dedicated env check logic, but we don't need it anywhere. Do you see any implication of treating webworker as a browser env?

Feiyang1:
I was thinking something along that line. We could just export the result of the function. Then you can do something like
```javascript
import { Environment, ENVIRONMENT } from `@firebase/util`
if (Environment === ENVIRONMENT.Browser) ....
```

We can also extend the function to return an object if there is other useful information. I think it's a good idea.

packages/util/src/environment.ts
mmermerkaya: To make sure that self is actually the global self, and not some random global that a developer defined.

I thought that in [the place you use this function](https://github.com/firebase/firebase-js-sdk/blob/f6afa45fe24f7adf5ea9a672100049c2ca8571db/packages/app/index.ts#L26), you'd want the result to be true for web workers as well. It's a useful check in both window and web worker contexts.

PR #1776, opened 2019-05-09
Standardize shutdown checking logic

thebrianchen:


packages/firestore/src/core/firestore_client.ts
```diff
@@ -495,10 +495,10 @@ export class FirestoreClient {
   }
 
   shutdown(): Promise<void> {
-    if (this.isShutdown === true) {
-      return Promise.resolve();
-    }
     return this.asyncQueue.enqueue(async () => {
+      if (this.isShutdown === true) {
+        return Promise.resolve();
```
wilhuff: nit: This is already inside an async function returning `Promise<void>`, so can't you just `return;`?

Also on Android you made it such that the `if` tests the equivalent `!this.isShutdown`. This is different.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -495,23 +495,22 @@ export class FirestoreClient {
...
 
   shutdown(): Promise<void> {
-    if (this.isShutdown === true) {
-      return Promise.resolve();
-    }
     return this.asyncQueue.enqueue(async () => {
-      // PORTING NOTE: LocalStore does not need an explicit shutdown on web.
-      if (this.lruScheduler) {
-        this.lruScheduler.stop();
+      if (this.isShutdown === false) {
```
wilhuff: Is it imperative that this value be strictly checked against `false`? If not, I'd consider the simpler `if (!this.isShutdown)`.

Looking through our source the convention seems to heavily favor the simpler form:
```
grep -r 'if (!' packages/firestore/src | wc -l   # => 158
grep -r '=== false' packages/firestore/src | wc -l  # => 2
```

If strictness really is required, please comment on why because this seems like something to fix.

PR #1791, opened 2019-05-16
Deprecate messaging.requestPermission

mmermerkaya:
I'll also update our documentation and guides to use `Notification.requestPermission()` instead of `messaging.requestPermission()`.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -5247,6 +5247,9 @@ declare namespace firebase.messaging {
       completed?: firebase.CompleteFn
     ): firebase.Unsubscribe;
     /**
+     * @deprecated Use Notification.requestPermission() instead.
+     * https://developer.mozilla.org/en-US/docs/Web/API/Notification/requestPermission
+     *
      * Notification permissions are required to send a user push messages.
      * Calling this method displays the permission dialog to the user and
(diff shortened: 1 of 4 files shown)


packages/firebase/index.d.ts
```diff
@@ -5247,6 +5247,9 @@ declare namespace firebase.messaging {
       completed?: firebase.CompleteFn
     ): firebase.Unsubscribe;
     /**
+     * @deprecated Use Notification.requestPermission() instead.
```
hsubox76: Can you move the `@deprecated` below the main body text of the comment? When Typedoc sees a tag in a comment, it thinks everything below it is part of the tag until it sees another tag.  Some of the deprecated tags in Auth have this problem too.  I'm fixing those in another PR now.

I'm also making a hack to the Typedoc templates to make sure the `@deprecated` block always displays at the top of the entry even though in the source code it's below the main comment text. I wish I could change the parser to not need all tags at the bottom but I think that requires forking Typedoc.

hsubox76:
Adding @egilmorez from docs to keep him aware of docs changes.

packages/firebase/index.d.ts
mmermerkaya: Yeah that makes sense, done.

PR #1792, opened 2019-05-16
Empty import typings for firebase/performance

jamesdaniels:
`firebase/performance` is missing a typing reference.


Feiyang1:
resolved by https://github.com/firebase/firebase-js-sdk/pull/1763

PR #1793, opened 2019-05-17
Remove exception throwing from clearPersistence()

thebrianchen:


packages/firestore/src/api/database.ts
```diff
@@ -428,18 +425,24 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
       this.makeDatabaseInfo()
     );
     const deferred = new Deferred<void>();
     this._queue.enqueueAndForget(async () => {
       try {
+        if (
+          this._firestoreClient !== undefined &&
+          !this._firestoreClient.isShutdown()
+        ) {
+          deferred.reject(
```
wilhuff: Since you're already in a `try`/`catch` block you could accomplish the same without an explicit reject/return. You could just `throw new FirestoreError(...)` and allow the `catch` to perform the rejection for you.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -590,6 +590,10 @@ export class FirestoreClient {
     return this.databaseInfo.databaseId;
   }
 
+  isShutdown(): boolean {
```
wilhuff: You could make this look like a readonly property to callers by marking this a getter.

See e.g. how `_started` is handled in `MemoryPersistence`

packages/firestore/src/core/firestore_client.ts
```diff
@@ -108,7 +108,7 @@ export class FirestoreClient {
   private lruScheduler?: LruScheduler;
 
   private readonly clientId = AutoId.newId();
-  private isShutdown = false;
+  private _isShutdown = false;
```
wilhuff: Optional: You might consider dropping the "is" prefix for the name. Other boolean properties in the source tree are typically named without it (e.g. started, committed, current).

PR #1796, opened 2019-05-17
Generate `syntax:esmodules` package for @firebase/performance

kristoferbaxter:
This PR generates an output using the soon to be specified `syntax` field for `package.json`.

The only value added is `esmodules` indicating the distributed output will be run using the module script loader. As a result, the output is roughly the same size as 'es2017' output but can be safely loaded using the module/nomodule pattern.

### DIFF
--- a/packages/performance/package.json
+++ b/packages/performance/package.json
@@ -7,6 +7,9 @@
   "browser": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "esm2017": "dist/index.esm2017.js",
+  "syntax": {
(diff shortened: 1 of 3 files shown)


Feiyang1:
@kristoferbaxter Thanks so much for the contribution! I couldn't find anything for the new `syntax` field online. Do you have some pointers where I can read more about it? Thanks!

kristoferbaxter:
We couldn’t come to consensus on this approach for code shipped to npm.

Dropping the PR for now.

PR #1802, opened 2019-05-20
Add performance spec tests for queries against large collections

schmidt-sebastian:
This PR:
- adds a new spec tests for queries that only return a small subset of documents from a large collection
- increases the test timeout for benchmark spec tests to 10 seconds
- fixes two performance tests failures that likely broke when we removed held write acks

### DIFF
--- a/packages/firestore/test/unit/specs/describe_spec.ts
+++ b/packages/firestore/test/unit/specs/describe_spec.ts
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+import { ExclusiveTestFunction, PendingTestFunction } from 'mocha';
+
 import { IndexedDbPersistence } from '../../../src/local/indexeddb_persistence';
 import { assert } from '../../../src/util/assert';
 import { addEqualityMatcher } from '../../util/equality_matcher';
@@ -47,7 +49,8 @@ const KNOWN_TAGS = [
 ];
 
 // TODO(mrschmidt): Make this configurable with mocha options.
-const RUN_BENCHMARK_TESTS = false;
+const RUN_BENCHMARK_TESTS = true;
+const BENCHMARK_TEST_TIMEOUT_MS = 10 * 1000;
 
 // The format of one describeSpec written to a JSON file.
 interface SpecOutputFormat {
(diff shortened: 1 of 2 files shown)


packages/firestore/test/unit/specs/perf_spec.test.ts
```diff
@@ -66,7 +66,7 @@ describeSpec(
               fromCache: true,
               hasPendingWrites: true
             })
-            .writeAcks(`collection/${i}`, ++currentVersion)
+            .writeAcks(`collection/${i}`, docRemote.version.toMicroseconds())
             .watchAcksFull(query, ++currentVersion, docRemote)
```
schmidt-sebastian: Note that in the existing code, this watchAck will get ignored since docRemote.version here is greater than the writeAck version.

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -47,7 +49,8 @@ const KNOWN_TAGS = [
 ];
 
 // TODO(mrschmidt): Make this configurable with mocha options.
-const RUN_BENCHMARK_TESTS = false;
+const RUN_BENCHMARK_TESTS = true;
```
mikelehen: Intentionally enabling these?  How long do they take?

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -79,7 +82,10 @@ export function setSpecJSONHandler(writer: (json: string) => void): void {
 }
 
 /** Gets the test runner based on the specified tags. */
-function getTestRunner(tags, persistenceEnabled): Function {
+function getTestRunner(
+  tags,
+  persistenceEnabled
+): ExclusiveTestFunction | PendingTestFunction {
```
mikelehen: Neat.

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -153,9 +168,10 @@ export function specTest(
     for (const usePersistence of persistenceModes) {
       const spec = builder();
       const runner = getTestRunner(tags, usePersistence);
+      const timeout = getTestTimeout(tags);
       const mode = usePersistence ? '(Persistence)' : '(Memory)';
       const fullName = `${mode} ${name}`;
-      runner(fullName, async () => {
+      const testFunction = runner(fullName, async () => {
```
mikelehen: Thoughts on renaming `testFunction` to `queuedTest` or something?  I misunderstood what was going on at first.

packages/firestore/test/unit/specs/describe_spec.ts
schmidt-sebastian: Hm. I woke up too early today.

packages/firestore/test/unit/specs/describe_spec.ts
schmidt-sebastian: Renamed

PR #1804, opened 2019-05-20
Switch from 'grpc' to '@grpc/grpc-js' dependency.

### LINKED ISSUE #1783: FR: Replace grpc with grpc-js
Hi, I noticed that last week the https://github.com/googleapis/nodejs-firestore package got rid of the grpc node native addon and replaced it with `@grpc/grpc-js`. I was wondering if this package was planning on upgrading to `grpc-js` to remove the complexity of dealing with compiled node addons for different platforms like node and electron. I couldn't find any discussion on this in the issue tracker. Thanks!

mikelehen:
Fixes #1783, reduces package size, and should ease compatibility issues with Electron, etc.

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -31,8 +31,8 @@
     "@firebase/firestore-types": "1.3.0",
     "@firebase/logger": "0.1.14",
     "@firebase/webchannel-wrapper": "0.2.20",
+    "@grpc/grpc-js": "0.4.0",
     "@grpc/proto-loader": "^0.5.0",
-    "grpc": "1.20.3",
     "tslib": "1.9.3"
   },
   "peerDependencies": {
--- a/packages/firestore/src/platform_node/grpc_connection.ts
+++ b/packages/firestore/src/platform_node/grpc_connection.ts
@@ -15,12 +15,12 @@
  * limitations under the License.
  */
 
-import * as grpc from 'grpc';
+import * as grpc from '@grpc/grpc-js';
 
 import firebase from '@firebase/app';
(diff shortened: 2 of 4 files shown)


jeffreypriebe:
@mikelehen This seems great, but the latest release of @firebase/firestore now doesn't work on node 6 with this, given [@grpc/grpc-js requirement for node v8](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/package.json#L9).

Is there a reasonable remedy?
Is this node v6 support a non-goal for firestore at this point (we're using this with firestore and have had trouble with Google cloud functions in node 8 [we moved to node 8 and had to roll back] [per this issue](https://github.com/firebase/firebase-functions/issues/429)).

mikelehen:
@jeffreypriebe To be clear, are you using the [firebase](https://www.npmjs.com/package/firebase) npm module (in which case this is the right github repo)?  Or are you using [https://www.npmjs.com/package/firebase-admin](https://www.npmjs.com/package/firebase-admin) or [@google-cloud/firestore](https://www.npmjs.com/package/@google-cloud/firestore)?

mikelehen:
@jeffreypriebe (I'm asking because usually you would use firebase-admin if you're using Cloud Functions...

mikelehen:
@jeffreypriebe We've ended up reverting this change (will be in the next release) since it is a breaking change to remove support for v6.  We will hold off until our next breaking change release.

That said, node v6 has reached end-of-life (https://github.com/nodejs/Release#end-of-life-releases) and you should definitely strive to upgrade when you can.

PR #1805, opened 2019-05-20
Apply strictNullChecks to all packages except database

hsubox76:
Applying Typescript option strictNullChecks to all packages and fixing existing code where it violates this rule.

Database has an override on this for now because it has so many strict null check violations to fix that it should be a separate PR.


packages/storage/src/task.ts
```diff
@@ -455,7 +455,7 @@ export class UploadTask {
    */
   on(
     type: TaskEvent,
-    nextOrObserver = undefined,
+    nextOrObserver = null,
```
schmidt-sebastian: I don't know if this is the right fix. The API allows for the observer function to be undefined (https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage-types/index.d.ts#L80). I think we need to update https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage/src/implementation/observer.ts#L41 to accept `undefined`.

Feiyang1: How about changing the interface at (https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage-types/index.d.ts#L80) to
```javascript
on(
...
    nextOrObserver:
      | Observer<UploadTaskSnapshot>
      | ((a: UploadTaskSnapshot) => any) = null,
...
  ): Function
```

It seems unnecessary to accept undefined if null means the same thing. Please correct me if undefined and null mean different things.

schmidt-sebastian: If I understand your suggestion correctly, that's a breaking change that affects the public API.

packages/storage/test/blob.test.ts
```diff
@@ -42,16 +42,18 @@ describe('Firebase Storage > Blob', () => {
   it('Slicing works', () => {
     const blob = new FbsBlob(new Uint8Array([1, 2, 3, 4, 5, 6, 7]));
     const sliced = blob.slice(1, 5);
+    assert.isNotNull(sliced);
```
mmermerkaya: Since this is a test, I think it's safe to just define `sliced` like this:

```ts
const sliced = blob.slice(1, 5)!;
```

Then you can remove the `!`s below.

packages/storage/test/blob.test.ts
```diff
@@ -42,16 +42,18 @@ describe('Firebase Storage > Blob', () => {
...
     testShared.assertUint8ArrayEquals(
-      sliced.uploadData() as Uint8Array,
+      sliced!.uploadData() as Uint8Array,
       new Uint8Array([2, 3, 4, 5])
     );
   });
   it('Blobs are merged with strings correctly', () => {
     const blob = new FbsBlob(new Uint8Array([1, 2, 3, 4]));
     const merged = FbsBlob.getBlob('what', blob, '\ud83d\ude0a ');
+    assert.isNotNull(merged);
```
mmermerkaya: Same here.

packages/storage/test/blob.test.ts
```diff
@@ -80,6 +82,7 @@ describe('Firebase Storage > Blob', () => {
 
     const concatenated = FbsBlob.getBlob(blob1, blob2);
 
-    assert.equal(20, concatenated.size());
+    assert.isNotNull(concatenated);
```
mmermerkaya: And here :smile:

packages/storage/test/requests.test.ts
```diff
@@ -298,7 +298,7 @@ describe('Firebase Storage > Requests', () => {
       [locationNormal, locationNormalNoObjUrl],
       [locationEscapes, locationEscapesNoObjUrl]
     ];
-    const promises = [];
+    const promises: Promise<void>[] = [];
```
mmermerkaya: Here I'd use `Array<Promise<void>>`. http://go/tsstyle#arrayt-type

packages/storage/test/testshared.ts
```diff
@@ -104,8 +107,9 @@ export function assertThrows(f: () => void, code: Code): FirebaseStorageError {
     }
   }).to.throw();
   expect(captured).to.be.an.instanceof(FirebaseStorageError);
+  // @ts-ignore Compiler thinks catch block is unreachable.
```
mmermerkaya: I think it's because TS doesn't know that the callback is invoked immediately.

Hopefully they'll fix it at some point.

https://github.com/Microsoft/TypeScript/issues/11498

packages/util/src/crypt.ts
```diff
@@ -49,12 +49,12 @@ const stringToByteArray = function(str) {
...
  * Turns an array of numbers into the string given by the concatenation of the
  * characters to which the numbers correspond.
- * @param {Array<number>} bytes Array of numbers representing characters.
+ * @param {number[]} bytes Array of numbers representing characters.
  * @return {string} Stringification of the array.
  */
 const byteArrayToString = function(bytes) {
   // TODO(user): Use native implementations if/when available
-  var out = [],
+  var out: Array<string> = [],
```
mmermerkaya: `string[]`

packages/util/src/obj.ts
```diff
@@ -118,7 +118,7 @@ export const getAnyKey = function(obj) {
 };
 
 export const getValues = function(obj) {
-  var res = [];
+  var res: any[] = [];
```
mmermerkaya: Can we replace this whole function with `Object.values()`?

packages/storage/src/reference.ts
```diff
@@ -175,7 +175,7 @@ export class Reference {
     let data = fbsString.dataFromString(format, string);
     let metadata = object.clone<Metadata>(opt_metadata);
     if (!type.isDef(metadata['contentType']) && type.isDef(data.contentType)) {
-      metadata['contentType'] = data.contentType;
+      metadata['contentType'] = data.contentType || undefined;
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/util/test/errors.test.ts
```diff
@@ -93,8 +93,9 @@ describe('FirebaseError', () => {
 
   it('has stack', () => {
     const e = ERROR_FACTORY.create('generic-error');
+    assert.isDefined(e.stack);
     // Multi-line match trick - .* does not match \n
-    assert.match(e.stack, /FirebaseError[\s\S]/);
+    assert.match(e.stack as string, /FirebaseError[\s\S]/);
```
mmermerkaya: `e.stack!` should work I think.

packages/app/src/firebaseApp.ts
```diff
@@ -115,7 +115,9 @@ export class FirebaseAppImpl implements FirebaseApp {
 
         return Promise.all(
           services.map(service => {
-            return service.INTERNAL.delete();
+            return service.INTERNAL
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/app/test/firebaseApp.test.ts
```diff
@@ -161,8 +161,8 @@ function executeFirebaseTests() {
           const [app, , instanceIdentifier] = args;
           return new TestService(app, instanceIdentifier);
         },
-        null,
-        null,
+        undefined,
```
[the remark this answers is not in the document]
hsubox76: Will tackle this in another PR.

packages/storage/src/implementation/metadata.ts
```diff
@@ -70,7 +70,7 @@ export function getMappings(): Mappings {
   if (mappings_) {
     return mappings_;
   }
-  let mappings = [];
+  let mappings: Array<Mapping> = [];
```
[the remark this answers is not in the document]
Feiyang1: It says use [] only with simple types and Array<> will complex types. Isn't Mapping a complex type?

packages/logger/test/logger.test.ts
```diff
@@ -16,15 +16,15 @@
...
+import { spy as Spy, SinonSpy } from 'sinon';
 import { Logger, LogLevel } from '../src/logger';
 import { setLogLevel } from '../index';
 import { debug } from 'util';
 
 describe('@firebase/logger', () => {
   const message = 'Hello there!';
   let client: Logger;
-  const spies = {
+  const spies: { [key: string]: SinonSpy | null } = {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/rxfire/database/utils.ts
```diff
@@ -26,14 +26,15 @@ export function isNil(obj: any): boolean {
  * that is populated with all the Realtime Database child events.
  * @param events
  */
-export function validateEventsArray(events?: ListenEvent[]) {
+export function validateEventsArray(events?: ListenEvent[]): ListenEvent[] {
+  let validatedEvents: ListenEvent[] = [];
   if (isNil(events) || events!.length === 0) {
```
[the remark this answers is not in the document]
hsubox76: Changed the return type of isNil since it's used elsewhere and this method doesn't touch implementation (playing it safe).  Will use `== null` for future code.

packages/storage/src/implementation/metadata.ts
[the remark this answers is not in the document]
mmermerkaya: A simple type is defined as "containing just alphanumeric characters and dot". It doesn't have to be a primitive type.

BTW, there's a [TSLint rule](https://palantir.github.io/tslint/rules/array-type/) for this (`array-simple` option), which [Google TSLint uses](https://github.com/google/gts/blob/d8461f302156dd66098d65d7da1a0a48b401e2b2/tslint-rules.json#L3).

packages/rxfire/test/database.test.ts
```diff
@@ -41,11 +41,11 @@ let batch = (items: any[]) => {
 };
 
 describe('RxFire Database', () => {
-  let app: app.App = null;
-  let database: database.Database = null;
+  let app: app.App | null = null;
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/rxfire/test/firestore.test.ts
```diff
@@ -67,8 +67,8 @@ const seedTest = firestore => {
 };
 
 describe('RxFire Firestore', () => {
-  let app: app.App = null;
-  let firestore: firestore.Firestore = null;
+  let app: app.App | null = null;
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/storage/src/implementation/observer.ts
```diff
@@ -15,16 +15,17 @@
...
-type ErrorFn = (error: Error) => void;
+type ErrorFn = (error: Error | FirebaseStorageError) => void;
 type CompleteFn = () => void;
 type Unsubscribe = () => void;
 
 type Subscribe<T> = (
-  next: NextFn<T> | { [name: string]: string | null },
-  error?: ErrorFn,
-  complete?: CompleteFn
+  next: NextFn<T> | { [name: string]: string | null } | null | undefined,
```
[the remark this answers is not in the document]
hsubox76: Fixed all null | undefined to `var?: ... | null` at least.

packages/storage/src/implementation/requests.ts
```diff
@@ -271,7 +271,7 @@ export function multipartUpload(
   if (body === null) {
     throw errorsExports.cannotSliceBlob();
   }
-  let urlParams = { name: metadata['fullPath'] };
+  let urlParams: UrlParams = { name: metadata['fullPath'] || '' };
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/storage/test/blob.test.ts
hsubox76: Done.

packages/storage/test/blob.test.ts
hsubox76: Done.

packages/storage/src/task.ts
Feiyang1: I take it back - we don't need to change the interface.

My point is that the change here (nextOrObserver = null) is sufficient. User can still pass `undefined` to nextOrObserver, and in this case, we will default it to `null` internally, so https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage/src/implementation/observer.ts#L41  doesn't need to accept `undefined`.

packages/storage/test/requests.test.ts
hsubox76: Done.

packages/storage/src/implementation/metadata.ts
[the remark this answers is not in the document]
hsubox76: Oh, I see, it's about visual complexity of the variable name, not the actual object it represents.  So `Mapping` is a simple name. Done.

packages/storage/test/blob.test.ts
hsubox76: Done.

packages/storage/src/implementation/observer.ts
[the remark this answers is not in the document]
mmermerkaya: Oh, it's okay either way. I was mostly just complaining about this API accepting `T | null | undefined` everywhere, which we should fix at some point. Thanks though. :slightly_smiling_face:

packages/app/test/firebaseApp.test.ts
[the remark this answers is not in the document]
mmermerkaya: Sounds good, thanks. :slightly_smiling_face:

packages/util/src/obj.ts
hsubox76: Probably. The only difference is that `Object.values()` doesn't get properties in the prototype chain? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values But in the one case is invoked, that doesn't seem to matter. I don't want to mess with functionality too much in a TS PR though, maybe next time.

mmermerkaya: Getting properties from the prototype chain sounds like a great footgun TBH. :smile:

I followed the calls to this function, and the only reason it's not caused an error yet is because everywhere it's used, there's a check like `if (change.type === some_type)` before any other field is accessed, and luckily, this just results in a check like `undefined === some_type` for the values and functions from the prototype chain.

I understand that you don't want to change this in this PR, but this is very likely to cause an error, and we should definitely change it.

I can send a PR if you'd like.

packages/app/src/firebaseApp.ts
```diff
@@ -114,11 +114,11 @@ export class FirebaseAppImpl implements FirebaseApp {
...
         return Promise.all(
-          services.map(service => {
-            return service.INTERNAL
-              ? service.INTERNAL.delete()
-              : Promise.resolve();
-          })
+          services
+            .filter(service => 'INTERNAL' in service)
+            .map(service => {
+              return service.INTERNAL!.delete();
```
mmermerkaya: `.map(service => service.INTERNAL!.delete());` should work, no need for the extra `return`.

packages/rxfire/database/utils.ts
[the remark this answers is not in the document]
mmermerkaya: You can use generics instead of `any`.

```ts
function isNil<T>(obj: T | null | undefined): obj is null | undefined {
```

I'd still just change it to `== null` though. It's exactly the same as this function. :man_shrugging:

packages/app/src/lite/firebaseAppLite.ts
```diff
@@ -97,7 +97,9 @@ export class FirebaseAppLiteImpl implements FirebaseApp {
 
         return Promise.all(
           services.map(service => {
-            return service.INTERNAL.delete();
+            return service.INTERNAL
```
mmermerkaya: Can you change this one as well?

packages/storage/src/implementation/requests.ts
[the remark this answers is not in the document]
mmermerkaya: Also the others in this file please.

packages/rxfire/database/utils.ts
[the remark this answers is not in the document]
hsubox76: Removed isNil and also changed the one other usage in packages/rxfire/database/list/index.ts

packages/storage/src/implementation/requests.ts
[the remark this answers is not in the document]
hsubox76: Done.

packages/util/src/obj.ts
hsubox76: Sure, send a PR, sounds good.

packages/app/src/lite/firebaseAppLite.ts
hsubox76: Yep, done.

packages/app/src/firebaseApp.ts
hsubox76: Got it.

packages/rxfire/database/utils.ts
```diff
@@ -17,17 +17,13 @@
...
-}
-
 /**
  * Check the length of the provided array. If it is empty return an array
  * that is populated with all the Realtime Database child events.
  * @param events
  */
 export function validateEventsArray(events?: ListenEvent[]): ListenEvent[] {
-  if (isNil(events) || events!.length === 0) {
+  if (events == null || events!.length === 0) {
```
mmermerkaya: You can remove the `!` now.

packages/storage/test/testshared.ts
mmermerkaya: Can you update the comment with the actual reason?

packages/util/src/sha1.ts
```diff
@@ -43,32 +43,32 @@ export class Sha1 {
   /**
    * Holds the previous values of accumulated variables a-e in the compress_
    * function.
-   * @type {!Array<number>}
+   * @type {!number[]}
```
mmermerkaya: Since you're changing these, can you actually remove the `type` comments? They are useless with TS, and can even be misleading.

http://go/tsstyle#comments-documentation

packages/rxfire/database/utils.ts
hsubox76: Done.

packages/util/src/sha1.ts
hsubox76: Done.

packages/storage/test/testshared.ts
hsubox76: Done.

packages/rxfire/database/list/index.ts
```diff
@@ -42,11 +42,11 @@ export function list(
   query: database.Query,
   events?: ListenEvent[]
 ): Observable<QueryChange[]> {
-  events = validateEventsArray(events);
+  const eventsToList = validateEventsArray(events);
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/rxfire/database/list/index.ts
```diff
@@ -42,11 +42,11 @@ export function list(
...
   return fromOnce(query).pipe(
     switchMap(change => {
       const childEvent$ = [of(change)];
-      events.forEach(event => childEvent$.push(fromRef(query, event)));
+      eventsToList.forEach(event => childEvent$.push(fromRef(query, event)));
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/storage/test/blob.test.ts
hsubox76: oops, done.

packages/storage/test/service.test.ts
```diff
@@ -45,19 +46,19 @@ describe('Firebase Storage > Service', () => {
       const error = testShared.assertThrows(() => {
         service.ref('gs://bucket/object');
       }, 'storage/invalid-argument');
-      assert.match(error.message, /refFromURL/);
+      error && assert.match(error.message, /refFromURL/);
```
schmidt-sebastian: Can you see if this is really necessary? assertThrows should always return an error and if it doesn't, the test would pass when it shouldn't.

packages/storage/src/task.ts
schmidt-sebastian: I looked through the code and the saving grace seems to be https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage/src/implementation/observer.ts#L50

With that being said, I still don't think it is a good idea to diverge between out `.d.ts` file and the source. At least for the RTDB, this has caused a few issues in the past. Ideally, we would at some point auto-generate our types and that can only work if the types match the interfaces that we use in our sources.

With that being said, the current state of the PR resolves this issue and uses the type we use in index.d.ts.

packages/storage/test/service.test.ts
hsubox76: Changing assertThrows back to assuming it always returns an error (instead of allowing it to return null), and should fail a test first if not? Needed some more `@ts-ignore` tags to make it work.

packages/functions/src/api/service.ts
```diff
@@ -110,7 +110,7 @@ export class Service implements FirebaseFunctions, FirebaseService {
    * @param origin The origin of the local emulator, such as
    * "http://localhost:5005".
    */
-  useFunctionsEmulator(origin: string) {
+  useFunctionsEmulator(origin: string | null) {
```
Feiyang1: Don't think this function is intended to reset the emulator. Plus it's an API change. I would just leave it to only accept string.

packages/rxfire/test/database.test.ts
```diff
@@ -147,7 +147,7 @@ describe('RxFire Database', () => {
           count = count + 1;
           const { event, snapshot } = change;
           expect(event).to.equal(ListenEvent.added);
-          expect(snapshot.val()).to.eql(data[snapshot.key]);
+          expect(snapshot.val()).to.eql(data[snapshot.key || '']);
```
Feiyang1: could it happen? Should it be `snapshot.key!` ?

packages/storage/src/implementation/list.ts
```diff
@@ -58,11 +59,15 @@ function fromBackendResponse(
     items: [],
     nextPageToken: resource['nextPageToken']
   };
+  const bucket = authWrapper.bucket();
+  if (bucket === null) {
```
Feiyang1: This is why I don't like mixing `==null` in the code. I have to look up `authWrapper.bucket()` signature to understand if it is actually meant to be `==` and made a mistake.

packages/storage/src/reference.ts
```diff
@@ -226,7 +226,7 @@ export class Reference {
       prefixes: [],
       items: []
     };
-    return this.listAllHelper(accumulator, null).then(() => accumulator);
+    return this.listAllHelper(accumulator, undefined).then(() => accumulator);
```
Feiyang1: You can just omit the second param

packages/storage/test/service.test.ts
```diff
@@ -111,7 +112,7 @@ describe('Firebase Storage > Service', () => {
       const error = testShared.assertThrows(() => {
         new Service(testShared.fakeApp, xhrIoPool, 'gs://bucket/object/');
       }, 'storage/invalid-default-bucket');
-      assert.match(error.message, /Invalid default bucket/);
+      error && assert.match(error.message, /Invalid default bucket/);
```
Feiyang1: `assert.match()` should always execute. `testShared.assertThrows()` returns a `FirebaseStorageError`, I don't think we need to null check error.

packages/storage/test/service.test.ts
```diff
@@ -135,7 +136,7 @@ describe('Firebase Storage > Service', () => {
       const error = testShared.assertThrows(() => {
         service.refFromURL('path/to/child');
       }, 'storage/invalid-argument');
-      assert.match(error.message, /invalid/i);
+      error && assert.match(error.message, /invalid/i);
```
Feiyang1: The same as above.

packages/util/src/crypt.ts
```diff
@@ -49,12 +49,12 @@ const stringToByteArray = function(str) {
 /**
  * Turns an array of numbers into the string given by the concatenation of the
  * characters to which the numbers correspond.
- * @param {Array<number>} bytes Array of numbers representing characters.
+ * @param {number[]} bytes Array of numbers representing characters.
```
Feiyang1: all the param/return types should be moved from the comment to the definition directly.

packages/functions/src/api/service.ts
hsubox76: Ok, got rid of the test line that uses null to reset this.  Doesn't seem necessary.

packages/rxfire/test/database.test.ts
hsubox76: Theoretically it could be null but if it ever happens I think it's so far back in the stack that it's beyond the scope of this test to check for it. Changed to `snapshot.key!`

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/test/service.test.ts
hsubox76: Done.

packages/storage/test/service.test.ts
hsubox76: Done.

packages/util/src/crypt.ts
hsubox76: Done, fixed similar in rest of file.

PR #1806, opened 2019-05-20
Use @grpc/proto-loader in @firebase/testing

ryanpbrewster:


packages/testing/src/api/index.ts
```diff
@@ -202,13 +204,7 @@ export function loadFirestoreRules(
...
   let client = new EMULATOR.FirestoreEmulator(
     FIRESTORE_ADDRESS,
-    grpc.credentials.createInsecure(),
-    {
-      // Cap how much backoff gRPC will perform. This is testing code, so
-      // efficiency is less important than responsiveness.
-      'grpc.initial_reconnect_backoff_ms': 100,
-      'grpc.max_reconnect_backoff_ms': 100
-    }
+    grpc.credentials.createInsecure()
```
yuchenshi: So we don't need these any more?

ryanpbrewster: correct

PR #1809, opened 2019-05-22
Allow firestore to recover quicker when a network change occurs

thebrianchen:
reference: b/35928432
[Android PR](https://github.com/firebase/firebase-android-sdk/pull/217/files)

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -6,6 +6,8 @@
   if you continue to use `experimentalTabSynchronization`.
 - [feature] You can now query across all collections in your database with a
   given collection ID using the `FirebaseFirestore.collectionGroup()` method.
+- [changed] Firestore now recovers more quickly from long periods without
+  network access (#1809).
 
 # 1.1.4
 - [feature] Added an `experimentalForceLongPolling` setting that that can be
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,13 +1,15 @@
 
 # Unreleased
+- [changed] Firestore now recovers more quickly after network connectivity
+  changes (airplane mode, Wi-Fi availability, etc.).
+
+# 1.3.0
 - [changed] Deprecated the `experimentalTabSynchronization` setting in favor of 
   `synchronizeTabs`. If you use multi-tab synchronization, it is recommended
   that you update your call to `enablePersistence()`. Firestore logs an error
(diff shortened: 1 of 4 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   if you continue to use `experimentalTabSynchronization`.
 - [feature] You can now query across all collections in your database with a
   given collection ID using the `FirebaseFirestore.collectionGroup()` method.
+- [changed] Firestore now recovers more quickly from long periods without
+  network access (#1809).
```
mikelehen: I see you modeled this off Android, but:
1) This isn't really specific to "long periods without network access"
2) We usually don't include the PR number in changelog entries.  If there is a customer-reported issue that the PR is fixing (and that issue contains useful context), then sometimes we *do* reference it (e.g. "Fixes a potential crash when reading from persistence (#<insert issue ID with all the gory details)").  But all else being equal, ideally the changelog entry will be sufficient to describe the change on its own.

So I propose something like: Firestore now recovers more quickly after network connectivity changes (airplane mode, wifi / cellular availability, etc.).

packages/firestore/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   if you continue to use `experimentalTabSynchronization`.
 - [feature] You can now query across all collections in your database with a
```
schmidt-sebastian: To double the impact of your PR, can you move these two items to a new section? They were releases as part of 1.3.0 if I am not mistaken.

Also (@mikelehen) - Should we just use the version numbers from https://firebase.google.com/support/release-notes/js? In which case this would 6.0. We don't seem to have a changelog for the Firestore NPM package.

mikelehen: Personally I would be fine with that and it would be a lot easier. But I also don't want to volunteer to go back and fix all the version numbers. 😛

thebrianchen:
After talking with @wilhuff  and @schmidt-sebastian, we concluded that testing this functionality is not feasible. `window` events indicate when the network is available/unavailable, and there is no way to mock out `window` in the integration tests. I tried writing this as a spec test, but that does not work because WatchStream is mocked out there. ConnectivityMonitor kicks off Watch Stream's error handling to reset the exponential backoff, and it's not possible to wait for exponential backoff in a spec test.

I have verified it manually and made sure that the event emitted on reconnect is caused by the ConnectivityMonitor and *not* by exponential backoff. See b/35928432 for the logs.

PR #1811, opened 2019-05-23
Refactor and remove object utility functions

mmermerkaya:
Used some newer JavaScript methods instead of custom implementations. Fixed types.

Also did the same things in packages that used the refactored object utility functions.


packages/util/src/obj.ts
```diff
@@ -15,135 +15,137 @@
...
-export const forEach = function(obj, fn) {
-  for (var key in obj) {
-    if (Object.prototype.hasOwnProperty.call(obj, key)) {
-      fn(key, obj[key]);
-    }
+export function forEach<T extends object>(
+  obj: T,
+  fn: (key: Keys<T>, value: Values<T>) => void
+) {
+  for (var key of Object.keys(obj)) {
```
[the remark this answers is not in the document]
mmermerkaya: :smile: I'm so used to TSLint fixing these.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,137 @@
  * limitations under the License.
  */
 
-// See http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/
+type Keys<T> = Extract<keyof T, string>;
+type Values<T> = T[keyof T];
 
-export const contains = function(obj, key) {
+export function contains<T extends object>(obj: T, key: keyof T): boolean {
```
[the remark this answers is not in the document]
mmermerkaya: Thanks :slightly_smiling_face:

packages/util/src/obj.ts
```diff
@@ -15,135 +15,137 @@
...
-export const getValues = function(obj) {
-  var res = [];
-  var i = 0;
-  for (var key in obj) {
-    res[i++] = obj[key];
-  }
-  return res;
-};
+export function getValues<T extends object>(obj: T): Array<Values<T>> {
+  return Object.values(obj);
```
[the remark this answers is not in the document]
hsubox76: It's only used in one place as far as I can find.  Is it worth having a util function or should we just call Object.values directly there?

packages/database/src/core/PersistentConnection.ts
```diff
@@ -948,8 +948,8 @@ export class PersistentConnection extends ServerActions {
 
     // Puts depend on having received the corresponding data update from the server before they complete, so we must
     // make sure to send listens before puts.
-    forEach(this.listens_, (pathString: string, queries: Object) => {
-      forEach(queries, (key: string, listenSpec: ListenSpec) => {
+    forEach(this.listens_, (_, queries) => {
+      forEach(queries, (_, listenSpec) => {
```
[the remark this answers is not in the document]
mmermerkaya: It means this variable is exempt from unused variable checks (this check is not turned on in our repo right now): https://github.com/Microsoft/TypeScript/issues/9458

It can just start with an underscore though, the rest of the name doesn't matter, so I'll add the names back.

packages/util/src/obj.ts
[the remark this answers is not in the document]
Feiyang1: If values in the prototype chain don't matter, we should use Object.values directly.

packages/database/src/core/PersistentConnection.ts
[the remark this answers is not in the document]
Feiyang1: 👍

packages/util/src/obj.ts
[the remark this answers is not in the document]
mmermerkaya: Where it's called its type implies that it doesn't include values from the prototype chain. I also checked where the value returned was used, and it's always used as the values from the object, without the prototype stuff, so yeah it's pretty safe IMO.

I just replaced the object that this was called with in that file with an actual `Map`. :smile:

mmermerkaya:
@Feiyang1 No problem! Also LMK if you need any help with the TSLint stuff.

Added those two to polyfills.

I think that the polyfills package is going to be difficult to keep in sync unless we change `lib` in the base `TSConfig` to `es5` and then add the libs one by one, which would cause an error if we use a function that's not included there. It's not possible to add every single method independently, but there's `ES2017.object` at least. So we'd add that to `lib` in `TSConfig` and then `import 'core-js/features/object';` to the polyfill package.

It's fine like this for now though. Just an idea.

packages/database/src/core/snap/IndexMap.ts
```diff
@@ -75,7 +75,7 @@ export class IndexMap {
       // regular child map
       return null;
     } else {
-      return sortedMap;
+      return sortedMap as SortedMap<NamedNode, Node>;
```
schmidt-sebastian: You made SafeGet type-safe below. I think you should need to update the index.d.ts of the util package as this `any` would go away.

packages/database/src/core/snap/Node.ts
```diff
@@ -119,7 +119,7 @@ export interface Node {
    * @param {boolean=} exportFormat True for export format (also wire protocol format).
    * @return {*} Value of this node as JSON.
    */
-  val(exportFormat?: boolean): Object;
+  val(exportFormat?: boolean): any;
```
schmidt-sebastian: Can we use lowercase `object` here?

packages/database/src/core/view/ChildChangeAccumulator.ts
```diff
@@ -89,14 +88,11 @@ export class ChildChangeAccumulator {
...
+      this.changeMap.set(childKey, change);
     }
   }
 
-  /**
-   * @return {!Array.<!Change>}
-   */
   getChanges(): Change[] {
-    return getValues(this.changeMap_);
+    return Array.from(this.changeMap.values());
```
schmidt-sebastian: This is not supported in IE11. Do you think the changes in this file warrant another Polyfill requirement?

packages/database/src/core/view/ChildChangeAccumulator.ts
```diff
@@ -41,44 +34,50 @@ export class ChildChangeAccumulator {
       childKey !== '.priority',
       'Only non-priority child changes can be tracked.'
     );
-    const oldChange = safeGet(this.changeMap_, childKey) as Change;
+    const oldChange = this.changeMap.get(childKey);
```
schmidt-sebastian: @mikelehen Do you know if there is a reason for `safeGet()`?. Are we guarding against queries for `length` or something similar?

packages/polyfill/index.ts
```diff
@@ -21,10 +21,12 @@ import 'whatwg-fetch';
 // Promise
 import 'promise-polyfill/lib/polyfill';
 
-// ES6
+// ES Stable
 import 'core-js/features/array/find';
 import 'core-js/features/array/find-index';
 import 'core-js/features/object/assign';
+import 'core-js/features/object/entries';
+import 'core-js/features/object/values';
```
schmidt-sebastian: If we keep using the Map in ChildChangeAccumulator, do we need to add `map/values` here?

packages/storage/src/implementation/request.ts
```diff
@@ -291,7 +291,7 @@ export function makeRequest<T>(
 ): Request<T> {
   let queryPart = UrlUtils.makeQueryString(requestInfo.urlParams);
   let url = requestInfo.url + queryPart;
-  let headers = object.clone<Headers>(requestInfo.headers);
+  let headers = clone(requestInfo.headers);
```
schmidt-sebastian: This could be `{...requestInfo.headers}` or `Object.assign({}, requestInfo.headers)`.

packages/storage/src/implementation/requestmap.ts
```diff
@@ -51,7 +50,7 @@ export class RequestMap {
    * Cancels all registered requests.
    */
   clear() {
-    object.forEach(this.map_, (key: string, val: Request<any>) => {
+    forEach(this.map_, (key: string, val: Request<any>) => {
```
schmidt-sebastian: I would prefer if we use `Object.keys`.

packages/storage/src/implementation/requests.ts
```diff
@@ -218,7 +218,7 @@ export function metadataForUpload_(
   blob: FbsBlob,
   opt_metadata?: Metadata | null
 ): Metadata {
-  let metadata = object.clone<Metadata>(opt_metadata);
+  let metadata = clone(opt_metadata);
```
schmidt-sebastian: This could be `Object.assign({}, opt_metadata)`.

packages/storage/src/implementation/url.ts
```diff
@@ -28,7 +28,7 @@ export function makeUrl(urlPart: string): string {
 export function makeQueryString(params: { [key: string]: string }): string {
   let encode = encodeURIComponent;
   let queryPart = '?';
-  object.forEach(params, function(key, val) {
+  forEach(params, function(key, val) {
```
schmidt-sebastian: Please use `Object.keys`.

packages/storage/src/implementation/xhrio_network.ts
```diff
@@ -65,7 +65,7 @@ export class NetworkXhrIo implements XhrIo {
     this.xhr_.open(method, url, true);
     if (type.isDef(opt_headers)) {
       const headers = opt_headers as Headers;
-      object.forEach(headers, (key, val) => {
+      forEach(headers, (key, val) => {
```
schmidt-sebastian: Please use `Object.keys`.

packages/storage/src/reference.ts
```diff
@@ -173,7 +173,7 @@ export class Reference {
     );
     this.throwIfRoot_('putString');
     let data = fbsString.dataFromString(format, string);
-    let metadata = object.clone<Metadata>(opt_metadata);
+    let metadata = clone(opt_metadata);
```
schmidt-sebastian: Consider `Object.assign({}, opt_metadata)`

packages/storage/test/testshared.ts
```diff
@@ -16,9 +16,9 @@
  */
 import { expect } from 'chai';
 import { FirebaseApp } from '@firebase/app-types';
+import { contains, forEach } from '@firebase/util';
```
schmidt-sebastian: I would also consider just using the built-in primitives.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,133 @@
  * limitations under the License.
  */
 
-// See http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/
+type Keys<T> = Extract<keyof T, string>;
```
schmidt-sebastian: I am huge fan of the changes in this file... but I feel that some of the methods here don't add very much value.  Have we considered getting rid of at least of some of them?

packages/util/src/obj.ts
```diff
@@ -15,135 +15,133 @@
...
- * @param {!Object} objFrom
- * @return {!Object} objTo
  */
-export const extend = function(objTo, objFrom) {
-  forEach(objFrom, function(key, value) {
-    objTo[key] = value;
-  });
-  return objTo;
-};
+export function extend<T extends object, U extends object>(
```
schmidt-sebastian: If this is not used often, this is a good candidate for removal.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,133 @@
...
 /**
  * Returns a clone of the specified object.
- * @param {!Object} obj
- * @return {!Object} cloned obj.
+ * @return cloned obj.
  */
-export const clone = function(obj) {
-  return extend({}, obj);
-};
+export function clone<T extends object>(obj: T): T {
```
schmidt-sebastian: If this is not used often, this is a good candidate for removal.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,133 @@
...
+}
 
-export const isEmpty = function(obj) {
-  for (var key in obj) {
-    return false;
-  }
-  return true;
-};
+export function isEmpty(obj: object): obj is {} {
+  return Object.keys(obj).length === 0;
```
schmidt-sebastian: This is slower than before.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,133 @@
...
-  }
-};
+}
 
-export const getValues = function(obj) {
-  var res = [];
-  var i = 0;
-  for (var key in obj) {
-    res[i++] = obj[key];
+export function getAnyKey<T extends object>(obj: T): keyof T | undefined {
```
schmidt-sebastian: This is slower than before.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -948,8 +948,8 @@ export class PersistentConnection extends ServerActions {
 
     // Puts depend on having received the corresponding data update from the server before they complete, so we must
     // make sure to send listens before puts.
-    forEach(this.listens_, (pathString: string, queries: Object) => {
-      forEach(queries, (key: string, listenSpec: ListenSpec) => {
+    forEach(this.listens_, (_pathString, queries) => {
```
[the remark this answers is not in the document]
mmermerkaya: That's what I did initially, but @Feiyang1 asked me to add the names. I don't mind either way, as long as the variable starts with a `_`. You two can decide. :smile:

packages/database/src/core/snap/IndexMap.ts
mmermerkaya: This isn't about that though. The "fallback type" here is object, which is a subtype of `SortedMap`, so TS can't narrow the type to `SortedMap`.

I reversed the condition and used `instanceof` instead.

packages/database/src/core/snap/Node.ts
mmermerkaya: It [can be a primitive value](https://github.com/firebase/firebase-js-sdk/blob/ae32b2b58a3969266ab701a6b2a4895b790c8ea8/packages/database/src/core/snap/LeafNode.ts#L189). The pre-TS types in the comments confirm this as well. Sometimes the best type to use is `any`. :man_shrugging:

packages/database/src/core/view/ChildChangeAccumulator.ts
mmermerkaya: Probably. Man, that file's going to be really hard to keep in sync if we add every feature one by one. :confused:

packages/polyfill/index.ts
mmermerkaya: Added. :man_shrugging:

packages/storage/src/implementation/request.ts
mmermerkaya: I tried not to change anything that was working fine after my refactoring. If I remove these functions that can be replaced by native functions (and I want to), I'll just change all the packages in a single PR and get approval from everyone. So I'll leave these as is for now.

packages/util/src/obj.ts
mmermerkaya: I tried not to change anything that was working fine after my refactoring. If I remove these functions that can be replaced by native functions (and I want to), I'll just change all the packages in a single PR and get approval from everyone. So I'll leave these as is for now.

That's why I'm also leaving the util functions in your other comments instead of replacing them now. I'll send a PR right after this one.

packages/database/src/core/view/ChildChangeAccumulator.ts
mmermerkaya: Added to polyfill package.

packages/util/src/obj.ts
mmermerkaya: Done.

packages/util/src/obj.ts
mmermerkaya: Done.

packages/database/src/core/view/ChildChangeAccumulator.ts
mmermerkaya: I think it's just guarding against accidentally getting a value from the prototype chain.

packages/storage/src/implementation/request.ts
mmermerkaya: Done in next PR.

packages/storage/src/implementation/requestmap.ts
mmermerkaya: Done in next PR.

packages/storage/src/implementation/url.ts
mmermerkaya: Done in next PR.

packages/util/src/obj.ts
mmermerkaya: Done in next PR.

packages/util/src/obj.ts
mmermerkaya: Done in next PR.

packages/storage/test/testshared.ts
mmermerkaya: Done in next PR.

packages/storage/src/reference.ts
mmermerkaya: Done in next PR.

packages/storage/src/implementation/xhrio_network.ts
mmermerkaya: Done in next PR.

packages/storage/src/implementation/requests.ts
mmermerkaya: Done in next PR.

packages/database/src/core/PersistentConnection.ts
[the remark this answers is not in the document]
Feiyang1: I just find it more clear with a meaningful name. Maybe we need to use the first parameter one day, the name can provide a context that makes it easier to understand.

mmermerkaya:
I'm removing some of these functions in the next PR: https://github.com/firebase/firebase-js-sdk/pull/1816

packages/database/src/core/snap/Node.ts
schmidt-sebastian: At least internally, for cases like this we have migrated to `unknown`, which enforces type safety in the calling function.

packages/polyfill/index.ts
```diff
@@ -21,11 +21,16 @@ import 'whatwg-fetch';
...
+import 'core-js/features/array/from';
 import 'core-js/features/object/assign';
+import 'core-js/features/object/entries';
+import 'core-js/features/object/values';
 import 'core-js/features/string/starts-with';
 import 'core-js/features/string/repeat';
 import 'core-js/features/symbol';
 import 'core-js/features/symbol/iterator';
+import 'core-js/features/map';
+import 'core-js/features/set';
```
schmidt-sebastian: @Feiyang1 Since we no longer ship these Polyfills by default, are changes like this considered breaking changes?

@mmermerkaya What is the size impact of these Polyfills?

packages/util/src/obj.ts
```diff
@@ -15,135 +15,138 @@
...
-  for (var key in obj) {
-    res[key] = f.call(opt_obj, obj[key], key, obj);
+export function map<T extends object, V, U extends { [key in keyof T]: V }>(
+  obj: T,
+  fn: (value: Values<T>, key: Keys<T>, obj: T) => V,
+  opt_obj?: unknown
+): U {
+  const res: Partial<U> = {};
+  for (const [key, value] of Object.entries(obj)) {
+    res[key] = fn.call(opt_obj, value, key, obj);
```
schmidt-sebastian: This function now also loops twice.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,138 @@
...
-export const findKey = function(obj, fn, opt_this?) {
-  for (var key in obj) {
-    if (fn.call(opt_this, obj[key], key, obj)) {
-      return key;
+export function findKey<T extends object>(
+  obj: T,
+  fn: <K extends Keys<T>>(value: T[K], key: K, obj: T) => boolean,
+  opt_this?: unknown
+): keyof T | undefined {
+  for (const [key, value] of Object.entries(obj)) {
```
schmidt-sebastian: This function now also loops twice.

packages/util/src/obj.ts
```diff
@@ -15,135 +15,138 @@
...
+export function every<T extends object>(
+  obj: T,
+  fn: (key: Keys<T>, value: Values<T>) => boolean
 ): boolean {
-  for (let key in obj) {
-    if (Object.prototype.hasOwnProperty.call(obj, key)) {
-      if (!fn(key, obj[key])) {
-        return false;
-      }
+  for (const [key, value] of Object.entries(obj)) {
```
schmidt-sebastian: This function now also loops twice.

packages/polyfill/index.ts
Feiyang1: They shouldn't be considered as breaking changes.

Feiyang1:
@mmermerkaya What is the state of the PR? any blocker?

packages/util/src/obj.ts
mmermerkaya: I merged the other PR into this one, so most of these functions are now gone.

packages/database/src/core/SyncPoint.ts
```diff
@@ -237,71 +222,57 @@ export class SyncPoint {
...
-   *
-   * @param {!Path} path The path to the desired complete snapshot
-   * @return {?Node} A complete cache, if it exists
+   * @param path The path to the desired complete snapshot
+   * @return A complete cache, if it exists
    */
   getCompleteServerCache(path: Path): Node | null {
     let serverCache: Node | null = null;
-    forEach(this.views_, (key: string, view: View) => {
+    for (const view of this.views.values()) {
```
schmidt-sebastian: This creates a copy.

packages/database/src/core/SyncPoint.ts
```diff
@@ -237,71 +222,57 @@ export class SyncPoint {
...
-   */
   getCompleteView(): View | null {
-    const completeView = findValue(this.views_, (view: View) =>
-      view
-        .getQuery()
-        .getQueryParams()
-        .loadsAllData()
-    );
-    return completeView || null;
+    for (const view of this.views.values()) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/SyncTree.ts
```diff
@@ -171,9 +158,9 @@ export class SyncTree {
         // overwrite
         affectedTree = affectedTree.set(Path.Empty, true);
       } else {
-        forEach(write.children, function(pathString: string, node: Node) {
+        for (const [pathString, node] of Object.entries(write.children)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/snap/nodeFromJSON.ts
```diff
@@ -69,17 +69,17 @@ export function nodeFromJSON(
     const children: NamedNode[] = [];
     let childrenHavePriority = false;
     const hinzeJsonObj: { [k: string]: any } = json as object;
-    forEach(hinzeJsonObj, (key: string, child: any) => {
-      if (typeof key !== 'string' || key.substring(0, 1) !== '.') {
+    for (const [key, child] of Object.entries(hinzeJsonObj)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/util/ImmutableTree.ts
```diff
@@ -50,9 +49,9 @@ export class ImmutableTree<T> {
    */
   static fromObject<T>(obj: { [k: string]: T }): ImmutableTree<T> {
     let tree: ImmutableTree<T> = ImmutableTree.Empty;
-    forEach(obj, (childPath: string, childSnap: T) => {
+    for (const [childPath, childSnap] of Object.entries(obj)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/util/Tree.ts
```diff
@@ -118,9 +118,9 @@ export class Tree<T> {
    * @param {function(!Tree.<T>)} action Action to be called for each child.
    */
   forEachChild(action: (tree: Tree<T>) => void) {
-    forEach(this.node_.children, (child: string, childTree: TreeNode<T>) => {
+    for (const [child, childTree] of Object.entries(this.node_.children)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/util/validation.ts
```diff
@@ -173,7 +173,7 @@ export const validateFirebaseData = function(
   if (data && typeof data === 'object') {
     let hasDotValue = false,
       hasActualChild = false;
-    forEach(data, function(key: string, value: any) {
+    for (const [key, value] of Object.entries(data)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/util/validation.ts
```diff
@@ -286,7 +286,7 @@ export const validateFirebaseMergeDataArg = function(
   }
 
   const mergePaths: Path[] = [];
-  forEach(data, function(key: string, value: any) {
+  for (const [key, value] of Object.entries(data)) {
```
schmidt-sebastian: This creates a a copy.

packages/database/src/core/SyncPoint.ts
mmermerkaya: `this.views` is a `Map`. This returns an iterator, not a copy.

packages/database/src/core/SyncPoint.ts
mmermerkaya: `this.views` is a `Map`. This returns an iterator, not a copy.

mmermerkaya:
Reverted (almost) all changes that added `Object.*` calls in Database. Replaced the `forEach` function from `@firebase/util` with DB's own util function `each`.

FWIW, the need to iterate over plain JS objects looks like a code smell to me. I'd recommend using ES6 `Map`s over `object`s if the keys aren't known in compile time and/or can change.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -76,9 +76,8 @@ export class PersistentConnection extends ServerActions {
   id = PersistentConnection.nextPersistentConnectionId_++;
   private log_ = logWrapper('p:' + this.id + ':');
 
-  /** @private {Object} */
   private interruptReasons_: { [reason: string]: boolean } = {};
-  private listens_: { [path: string]: { [queryId: string]: ListenSpec } } = {};
+  private readonly listens: Map<string, Map<string, ListenSpec>> = new Map();
```
schmidt-sebastian: Nit: Can you add a comment to explain what the different strings mean? Maybe something easy:

```
 private readonly listens: Map</* path */ string, Map</* queryId */string, ListenSpec>> = new Map();
```

mmermerkaya: Done. Added as JSDoc as well, so that it shows up in IntelliSense.

packages/database/src/core/Repo.ts
```diff
@@ -651,10 +573,13 @@ export class Repo {
...
-    forEach(stats, (stat: string, value: any) => {
+    each(stats, (stat: string, value: any) => {
+      let paddedStat = stat;
       // pad stat names to be the same length (plus 2 extra spaces).
-      for (let i = stat.length; i < longestName + 2; i++) stat += ' ';
-      console.log(stat + value);
+      for (let i = stat.length; i < longestName + 2; i++) {
+        paddedStat += ' ';
+      }
+      console.log(paddedStat + value);
```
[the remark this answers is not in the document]
mmermerkaya: I don't think so, but this PR is already big enough so I don't think I should change anything else.

PR #1816, opened 2019-05-24
Remove some object utility functions

mmermerkaya:
Replaces some functions from `@firebase/util` with native functions.

### DIFF
--- a/packages/database/src/core/PersistentConnection.ts
+++ b/packages/database/src/core/PersistentConnection.ts
@@ -76,7 +76,6 @@ export class PersistentConnection extends ServerActions {
   id = PersistentConnection.nextPersistentConnectionId_++;
   private log_ = logWrapper('p:' + this.id + ':');
 
-  /** @private {Object} */
   private interruptReasons_: { [reason: string]: boolean } = {};
   private listens_: { [path: string]: { [queryId: string]: ListenSpec } } = {};
   private outstandingPuts_: OutstandingPut[] = [];
@@ -88,26 +87,19 @@ export class PersistentConnection extends ServerActions {
   private securityDebugCallback_: ((a: Object) => void) | null = null;
   lastSessionId: string | null = null;
 
-  /** @private {number|null} */
   private establishConnectionTimer_: number | null = null;
 
-  /** @private {boolean} */
   private visible_: boolean = false;
 
   // Before we get connected, we keep a queue of pending messages to send.
   private requestCBHash_: { [k: number]: (a: any) => void } = {};
   private requestNumber_ = 0;
 
-  /** @private {?{
-   *   sendRequest(Object),
-   *   close()
-   * }} */
   private realtime_: {
     sendRequest(a: Object): void;
     close(): void;
   } | null = null;
 
-  /** @private {string|null} */
   private authToken_: string | null = null;
   private forceTokenRefresh_ = false;
   private invalidAuthTokenCount_ = 0;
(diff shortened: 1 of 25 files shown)


packages/database/src/core/PersistentConnection.ts
```diff
@@ -948,11 +948,11 @@ export class PersistentConnection extends ServerActions {
 
     // Puts depend on having received the corresponding data update from the server before they complete, so we must
     // make sure to send listens before puts.
-    forEach(this.listens_, (_pathString, queries) => {
-      forEach(queries, (_key, listenSpec) => {
+    for (const queries of Object.values(this.listens_)) {
```
schmidt-sebastian: Object.values() already loops through the object once (see https://github.com/es-shims/Object.values/blob/master/implementation.js).

While not the most beautiful JS to write in 2019, this would be more efficient as:

```
for (const key in this.listens_) {
  if (this.listens_. hasOwnProperty(key)) {
     ...
  }
}
```

This applies in other places too. Sorry for not pointing this out earlier (or rather: Sorry for leading you down this path).

packages/database/src/core/SyncPoint.ts
```diff
@@ -237,66 +222,48 @@ export class SyncPoint {
...
   hasCompleteView(): boolean {
     return this.getCompleteView() != null;
   }
 
-  /**
-   * @return {?View}
-   */
   getCompleteView(): View | null {
-    const completeView = findValue(this.views_, (view: View) =>
+    const completeView = Array.from(this.views.values()).find((view: View) =>
```
schmidt-sebastian: This creates a copy of the array now, whereas before it didn't.

packages/database/src/core/WriteTree.ts
```diff
@@ -504,29 +493,18 @@ export class WriteTree {
...
       return writeRecord.path.contains(path);
     } else {
-      // findKey can return undefined, so use !! to coerce to boolean
-      return !!findKey(writeRecord.children, function(
-        childSnap: Node,
-        childName: string
-      ) {
-        return writeRecord.path.child(childName).contains(path);
-      });
+      return Object.keys(writeRecord.children).some((childName: string) =>
```
schmidt-sebastian: This creates a copy now. 

It's also not supported by IE11 and needs another Polyfill: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/database/src/core/util/util.ts
```diff
@@ -410,9 +406,11 @@ export const each = function(
      * a single impl that does a key, value callback. So we invert
      * to not have to touch the `each` code points
      */
-    forEach(obj, (key: any, val: any) => fn(val, key));
+    for (const [key, value] of Object.entries(obj)) {
```
schmidt-sebastian: Please make sure that you do not create extra copies/loops in these utility functions.

packages/database/src/core/PersistentConnection.ts
mmermerkaya: Used a `Map` here instead of a plain `object`.

Unless these objects have 1000s of values, the performance difference is negligible and I'd rather have more readable code, but it's your package, so up to you.

packages/database/src/core/SyncPoint.ts
mmermerkaya: Done.

packages/database/src/core/WriteTree.ts
mmermerkaya: Done. I'm sure that at least 99% of people who use this just load all polyfills and not pick them one by one though.

packages/database/src/core/util/util.ts
mmermerkaya: Done.

mmermerkaya:
I'm merging this to the other PR so that we can submit both of them together.

PR #1818, opened 2019-05-24
Use HTTP method provided by NetworkRequest object instead of hardcode…

zijianjoy:
Currently we hardcoded "GET" as HttpMethod for network request in Performance. Switch to use the real value now. Since browser API doesn't support HTTPmethod retrieval, the value will be set as "UNKNOWN" for now.


Feiyang1:
@zijianjoy What is the status of this PR? are you waiting for someone to review? If so, can you please assign it to the appropriate person?

Issue #1817, opened 2019-05-24
Firebase 6.0.4 and web worker


universse:
Hi. I'm using firebase authentication, firestore and storage in web worker. After upgrading from version 6.02 to 6.04 i'm starting to get ```ReferenceError: window is not defined```. It seems to come from firestore.

Feiyang1:
Thanks for reporting it! We have made a fix https://github.com/firebase/firebase-js-sdk/pull/1819

schmidt-sebastian:
@universse Are you able to pin your Firestore release to 6.0.2?

Feiyang1:
@universse Can you also post a detailed log, so we can verify we actually fixed what is causing the issue for you?

universse:
Hi. I reverted back to 6.02 when I reported this issue. After upgrading to 6.04 again, there's no more issue. Thank you very much!

schmidt-sebastian:
I'm glad to hear that you are no longer blocked! Please note that the fix for your issue is, however, not yet released and will likely be part of 6.0.5. If you do see this issue again, please upgrade to 6.0.5 when it gets released.

universse:
The issue cropped up again. Somehow it did not show up earlier. Sorry for any confusion caused.
![image](https://user-images.githubusercontent.com/15166715/58365475-972f6b80-7ef7-11e9-9e10-ac054b7e21b3.png)

PR #1819, opened 2019-05-24
fix the condition that checks if window is available

Feiyang1:
fixes https://github.com/firebase/firebase-js-sdk/issues/1817

### DIFF
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -47,7 +47,7 @@ export class BrowserPlatform implements Platform {
   }
 
   newConnectivityMonitor(): ConnectivityMonitor {
-    if (this.window !== undefined) {
+    if (this.window) {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -47,7 +47,7 @@ export class BrowserPlatform implements Platform {
   }
 
   newConnectivityMonitor(): ConnectivityMonitor {
-    if (this.window !== undefined) {
+    if (this.window) {
```
schmidt-sebastian: This could also just be `typeof window !== 'undefined'`

Feiyang1: this.window does exactly that, so I figured we should just use it.

Feiyang1: @schmidt-sebastian mind approving the PR? Thanks!

PR #1822, opened 2019-05-24
Update license checking script.

hsubox76:
We have a script that automatically adds the entire license block but it doesn't add `@license` tags (required for internal Google use) if the block exists and is just missing the tag.  Updated the script to add these tags as well, and ran it on existing files.

### DIFF
--- a/packages/auth/demo/public/sample-config.js
+++ b/packages/auth/demo/public/sample-config.js
@@ -1,4 +1,5 @@
 /*
+ * @license
  * Copyright 2017 Google Inc. All Rights Reserved.
  *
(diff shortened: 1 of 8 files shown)


packages/auth/demo/public/sample-config.js
```diff
@@ -1,4 +1,5 @@
 /*
+* @license
```
Feiyang1: Nit: The line is misaligned.

hsubox76: Fixed script to match indent of line below when inserting license tags.

PR #1828, opened 2019-05-28
Retry API requests once if the server returns a 5xx response

mmermerkaya:


packages/installations/src/api/common.ts
```diff
@@ -75,10 +75,30 @@ export function getHeadersWithAuth(
...
+ * Calls the passed in fetch wrapper and returns the response.
+ * If the returned response has a status of 5xx, re-runs the function once and
+ * returns the response.
+ */
+export async function retryIfServerError(
+  fn: () => Promise<Response>
+): Promise<Response> {
+  const result = await fn();
+
+  if (String(result.status)[0] === '5') {
```
dwoffinden: That's pretty gross. Why not just check >=500&&<600?

packages/installations/src/api/create-installation.test.ts
```diff
@@ -104,24 +105,41 @@ describe('createInstallation', () => {
...
       );
-    });
 
-    it('throws a FirebaseError with the error information from the server', async () => {
       await expect(
         createInstallation(appConfig, inProgressInstallationEntry)
       ).to.be.rejectedWith(FirebaseError);
     });
+
+    it('retries once if the server returns a 5xx error', async () => {
```
dwoffinden: as well as a test that it does retry once, we want tests that it doesn't retry infinitely?

packages/installations/src/api/common.ts
mmermerkaya: Ehh, it's fine either way, but sure.

packages/installations/src/api/create-installation.test.ts
mmermerkaya: Added tests for `retryIfServerError`. No need to duplicate them in all API functions.

PR #1830, opened 2019-05-29
Remove return type generics

mmermerkaya:
It wasn't a good idea.


Feiyang1:
Why is it not a good idea? I kinda like it.

PR #1831, opened 2019-05-30
eslint rules for firebase projects

Feiyang1:
Create the base eslint rules for firebase projects. We will apply it to all repos eventually.
As the first step, we need to agree on the set of rules, so feedback is welcomed.

I have run this set of rules against all projects, and there are about 4000 issues. Given the large number of issues, we will enable the eslint rules for project one by one in a few followup PRs. 

The set of rules derives from the google tslint rules. I made a [document](https://docs.google.com/document/d/1oIJrTmqUFDVxH15RKaUneJAux-LgitLDojuLChyrBsI/edit#) to show what rules from google tslint rules are included/excluded and what additional rules are included in this PR.

### DIFF
--- a/config/.eslintrc.json
+++ b/config/.eslintrc.json
@@ -1,0 +1,165 @@
+{
+    "env": {
+        "browser": true,
+        "es6": true,
+        "node": true
+    },
+    "parser": "@typescript-eslint/parser",
+    "plugins": [
+        "@typescript-eslint",
+        "@typescript-eslint/tslint"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 2018,
+        "sourceType": "module",
+        "project": "packages/firestore/tsconfig.json"
+    },
+    "overrides": [
+        {
+            "files": [
+                "*test.ts",
+                "**/test/**/*.ts"
+            ],
+            "rules": {
+                "no-unused-expressions": "off"
+            }
+        }
+    ],
+    "rules": {
+        "arrow-body-style": [
+            "error",
+            "as-needed",
+            {
+                "requireReturnForObjectLiteral": true
+            }
+        ],
+        "curly": [
+            "error",
+            "multi-line"
+        ],
+        "guard-for-in": "error",
+        "no-extra-label": "error",
+        "no-unused-labels": "error",
+        "new-parens": "error",
+        "no-new-wrappers": "error",
+        "no-debugger": "error",
+        "no-duplicate-case": "error",
+        "no-throw-literal": "error",
+        "no-return-await": "error",
+        "no-unsafe-finally": "error",
+        "no-unused-expressions": [
+            "error",
+            {
+                "allowShortCircuit": true
+            }
+        ],
+        "no-var": "error",
+        "object-shorthand": "error",
+        "prefer-arrow-callback": [
+            "error",
+            {
+                "allowNamedFunctions": true
+            }
+        ],
+        "prefer-const": [
+            "error",
+            {
+                "destructuring": "all"
+            }
+        ],
+        "radix": "error",
+        "default-case": "error",
+        "eqeqeq": [
+            "error",
+            "always",
+            {
+                "null": "ignore"
+            }
+        ],
+        "no-caller": "error",
+        "no-cond-assign": ["error", "always"],
+        "use-isnan": "error",
+        "camelcase": "error",
+        "id-blacklist": [
+            "error", 
+            "any", 
+            "number", 
+            "string", 
+            "boolean"
+        ],
+        "@typescript-eslint/array-type": [
+            "error",
+            "array-simple"
+        ],
+        "@typescript-eslint/ban-types": [
+            "error",
+            {
+                "types": {
+                    "Object": "Use object instead.",
+                    "String": "Use 'string' instead.",
+                    "Number": "Use 'number' instead.",
+                    "Boolean": "Use 'boolean' instead."
+                }
+            }
+        ],
+        "@typescript-eslint/class-name-casing": "error",
+        "@typescript-eslint/interface-name-prefix": [
+            "error",
+            "never"
+        ],
+        "@typescript-eslint/prefer-interface": "error",
+        "@typescript-eslint/explicit-member-accessibility": [
+            "error",
+            {
+                "accessibility": "no-public",
+                "overrides": {
+                    "parameterProperties": "explicit"
+                }
+            }
+        ],
+        "@typescript-eslint/no-angle-bracket-type-assertion": "error",
+        "@typescript-eslint/no-explicit-any": "error",
+        "@typescript-eslint/no-namespace": [
+            "error",
+            {
+                "allowDeclarations": true
+            }
+        ],
+        "@typescript-eslint/no-triple-slash-reference": "error",
+        "@typescript-eslint/no-require-imports": "error",
+        "@typescript-eslint/no-useless-constructor": "error",
+        "@typescript-eslint/semi": "error",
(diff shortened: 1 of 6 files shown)


mmermerkaya:
Is `typescript-eslint` feature complete? Shouldn't we wait until they deprecate TSLint, which they will do after ESLint is feature complete?

You can see what's implemented and what's not [here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md).

Also I recommend using [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) and [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) instead of defining style rules. Or if we do TSLint instead, there's TSLint versions of these two as well. Example [here](https://github.com/firebase/firebase-js-sdk/blob/6045a08995ed4549bce033eed90d31eb892c29a2/packages/installations/tslint.json).

Feiyang1:
`typescript-eslint` is not feature complete according to the roadmap you linked, but it has implemented the majority of the features and a lot of the tslint rules have an equivalent rule in eslint already.

For the tslint rules that are not available in `eslint`/`@typescript-eslint` yet, I'm using `@typescript-eslint/tslint` to fill the gap.  We will remove this plugin eventually as the features are implemented in `eslint`/`@typescript-eslint`.

I don't think we need to include any code formatting related rules in eslint, since it will be dealt with by prettier. I also don't want to fail compilation because of the formatting issue. Though we could make prettier violations warnings instead of errors, I feel it will just clutter the output without too much benefits.

config/.eslintrc.json
```diff
@@ -0,0 +1,150 @@
...
+                "no-unused-expressions": "off"
+            }
+        }
+    ],
+    "rules": {
+        "arrow-body-style": [
+            "error",
+            "as-needed",
+            {
+                "requireReturnForObjectLiteral": true
```
mmermerkaya: But why? :disappointed: I always tell people that they can do `() => ({ ... })` in readability reviews.

mmermerkaya:
> I don't think we need to include any code formatting related rules in eslint, since it will be dealt with by prettier.

That's exactly what these two do. One disables formatting related rules of ESLint and the other checks Prettier for formatting. So formatting will be dealt with by Prettier, but within ESLint. I think that would make things easier, since we wouldn't have to run two different scripts for prepush and CI tests.

> I also don't want to fail compilation because of the formatting issue.

It wouldn't fail compilation, it would fail tests. That's where my packages (messaging and installations) run linters. And I thought CI tests checked formatting already? Or is that only the prepush script?

> clutter the output without too much benefits

That is true, but you can run fix / format on save in most editors (including VSCode, which is what I use).

Anyway, I don't feel strongly about any of these, just my two cents.

config/.eslintrc.json
```diff
@@ -0,0 +1,150 @@
...
+            }
+        ],
+        "curly": [
+            "error",
+            "multi-line"
```
mmermerkaya: Even though it's a bit stricter than Google TS style (doesn't allow braceless single line if), I vote for `"all"`.

config/.eslintrc.json
Feiyang1: I was confused by the syntax the first time I saw it, but I don't feel strongly about it.

config/.eslintrc.json
Feiyang1: I'd be fine with "all". I'm leaving it unresolved for a couple of days to see if anyone objects.

config/.eslintrc.json
mikelehen: After going through Firestore errors, we are not a fan of "arrow-body-style" at all.  We have 566 violations. 
 Some of them would be fine to change (though unclear benefit), but some would be worse.  For example, it would cause all of our tests to change from, e.g.:
```
  it('can set documents', () => {
    return integrationHelpers.withTestDoc(persistence, doc => {
      return doc.firestore
        .batch()
        .set(doc, { foo: 'bar' })
        .commit()
        .then(() => doc.get())
        .then(snapshot => {
          expect(snapshot.exists).to.equal(true);
          expect(snapshot.data()).to.deep.equal({ foo: 'bar' });
        });
    });
  });
```
to:
```
  it('can set documents', () =>
    integrationHelpers.withTestDoc(persistence, doc =>
      doc.firestore
        .batch()
        .set(doc, { foo: 'bar' })
        .commit()
        .then(() => doc.get())
        .then(snapshot => {
          expect(snapshot.exists).to.equal(true);
          expect(snapshot.data()).to.deep.equal({ foo: 'bar' });
        })
    ));
```

We are not in favor of this, since:
1) We think braces can be a useful tool when coding for denoting long blocks of code (even if it's just chaining method calls).
2) [quoting @wilhuff] Any multiline block that currently contains a single return statement can end up with multiple statements easily. Refactoring the extra statements in is error prone. This is why we require braces around if statements

So I'd advocate for removing this rule at the Firebase level.  And if we keep it, we (Firestore) will likely decide to disable it.

config/.eslintrc.json
mikelehen: Firestore is a fan of "all" too, thought this will probably create some violations since currently we allow no braces iff the statement is on the same line as the if, e.g.:
```
if (result.skip) continue;
```

Looks like eslint doesn't have a similar configuration, so we'll probably lose that... but I think that's okay.

packages/firestore/package.json
```diff
@@ -8,6 +8,7 @@
     "build:console": "node tools/console.build.js",
     "dev": "rollup -c -w",
     "lint": "tslint -p tsconfig.json -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
+    "lint:eslint": "eslint -c .eslintrc.json '**/*.ts'",
```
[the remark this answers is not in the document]
Feiyang1: Added the `--ignore-path ../../.gitignore`

config/.eslintrc.json
```diff
@@ -0,0 +1,165 @@
...
+        "@typescript-eslint/explicit-function-return-type": [
+            "error",
+            {
+                "allowExpressions": true
+            }
+        ],
+        "@typescript-eslint/no-unused-vars": [
+            "error",
+            {
+                "varsIgnorePattern": "^_*$"
```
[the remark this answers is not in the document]
Feiyang1: Added `"argsIgnorePattern": "^_"`.
I didn't realize `argsIgnorePattern` and `varsIgnorePattern` are 2 different options util now. :(

config/.eslintrc.json
```diff
@@ -0,0 +1,165 @@
...
+        "@typescript-eslint/no-require-imports": "error",
+        "@typescript-eslint/no-useless-constructor": "error",
+        "@typescript-eslint/semi": "error",
+        "@typescript-eslint/explicit-function-return-type": [
+            "error",
+            {
+                "allowExpressions": true
+            }
+        ],
+        "@typescript-eslint/no-unused-vars": [
```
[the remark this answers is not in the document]
Feiyang1: Yes! That's the intention, but I made a mistake in the regex. Update the regex to "^_".

config/.eslintrc.json
Feiyang1: Changed to `all`

config/.eslintrc.json
Feiyang1: Replaced this rule with tslint `"arrow-return-shorthand": true` which is currently enabled in firestore.
It enforces arrow function shorthand, but will ignore functions that span multiple lines (this option is not available in the eslint rule `arrow-body-style`).

PR #1838, opened 2019-06-03
Update protos to include IN and ARRAY_CONTAINS_ANY

thebrianchen:


packages/firestore/src/protos/firestore_proto_api.d.ts
```diff
@@ -44,6 +45,7 @@ export interface IFieldFilterOpEnum {
   GREATER_THAN_OR_EQUAL: FieldFilterOp;
   EQUAL: FieldFilterOp;
   ARRAY_CONTAINS: FieldFilterOp;
+  IN: FieldFilterOp;
```
mikelehen: I _think_ we probably want ARRAY_CONTAINS_ANY as well.  This is essentially the version of "in" that works on array fields.  So if your documents have a field 'tags' containing an array of tags you could do something like:  `movies.where('tags', 'array-contains-any', ['action', 'comedy'])`

reyawn: @mikelehen I did some searching and I'm not seeing any references to "array-contains-any", where are docs on this?

PR #1842, opened 2019-06-04
Replace faye-websocket with ws

ryanpbrewster:
This seems to be more a more modern websocket library. In my local benchmarks its performance is comparable with faye-websocket.

I encountered an inexplicable race condition when testing against the RTDB emulator and eventually root-caused it to an issue with faye-websocket occasionally dropping the first frame it receives. ws does not have this issue.

I do not think this will have any significant impact on clients, but it may give us more breathing room in terms of future maintenance.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1947,6 +1947,14 @@
   resolved "https://registry.npmjs.org/@types/tough-cookie/-/tough-cookie-2.3.3.tgz#7f226d67d654ec9070e755f46daebf014628e9d9"
   integrity sha512-MDQLxNFRLasqS4UlkWMSACMKeSm1x4Q3TxzUC7KQUsh6RK1ZrQ0VEyE3yzXcBu+K8ejVj4wuX32eUG02yNp+YQ==
 
+"@types/ws@6.0.1":
+  version "6.0.1"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -5820,7 +5828,7 @@ fast-url-parser@^1.1.3:
   dependencies:
     punycode "^1.3.2"
 
-faye-websocket@0.11.1, faye-websocket@>=0.6.0:
+faye-websocket@>=0.6.0:
```
ryanpbrewster: this is used by the dev dependency on `firebase-tools`, which in turn pulls in an old version of `firebase` (this repo)

ryanpbrewster:
ws seems to support the same constructor arguments (which feels extremely lucky): https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketaddress-protocols-options

mikelehen:
I don't see "headers" or "proxy/origin" in that list of options (or the referenced [http.request() options](https://nodejs.org/api/http.html#http_http_request_options_callback)).

I _do_ see "origin" listed, but it's a top-level option instead of proxy/origin and looks like it has a different meaning.

ryanpbrewster:
Ah, darn, I misread this and thought they were all just headers. This may not be a drop-in replacement, which would make replacing it challenging.

I'm closing this for now, I'll investigate why faye-websocket is breaking and see if we can fix it.

PR #1843, opened 2019-06-04
Add IN and ARRAY_CONTAINS_ANY (not publicly exposed)

thebrianchen:


packages/firestore/src/core/query.ts
```diff
@@ -568,7 +595,10 @@ export class RelationFilter extends Filter {
 
   isInequality(): boolean {
     return (
-      this.op !== RelationOp.EQUAL && this.op !== RelationOp.ARRAY_CONTAINS
+      this.op !== RelationOp.EQUAL &&
+      this.op !== RelationOp.ARRAY_CONTAINS &&
+      this.op !== RelationOp.IN &&
+      this.op !== RelationOp.ARRAY_CONTAINS_ANY
```
mikelehen: When there was only one (or two) exceptions, it probably made sense to write it this way, but at this point we should probably flip the logic and write
```
return (this.op === RelationOp.LESS_THAN || this.op === RelationOp.LESS_THAN_OR_EQUAL || ...);
```
or perhaps:
```
return [RelationOp.LESS_THAN, RelationOp.LESS_THAN_OR_EQUAL, ...].indexOf(this.op) >= 0;
```

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -30,6 +30,9 @@ import { EventsAccumulator } from '../util/events_accumulator';
 import firebase from '../util/firebase_export';
 import {
   apiDescribe,
+  arrayContainsAnyOp,
+  clearPersistence,
```
mikelehen: clearPersistence added on purpose?

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -555,6 +557,58 @@ apiDescribe('Queries', persistence => {
...
+  });
+
+  it('can use array-contains-any filters', async () => {
+    const testDocs = {
+      a: { array: [42] },
+      b: { array: ['a', 42, 'c'] },
+      c: { array: [41.999, '42', { a: [42] }] },
+      d: { array: [42], array2: ['bingo'] },
+      e: { array: [43] },
+      f: { array: [{ a: 42 }] }
```
mikelehen: Maybe add `{ array: 42 }` to make sure it doesn't get picked up?

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -555,6 +557,58 @@ apiDescribe('Queries', persistence => {
     });
   });
 
+  it('can use IN filters', async () => {
+    const testDocs = {
+      a: { zip: 98101 },
+      b: { zip: 91102 },
+      c: { zip: 98103 },
+      d: { zip: [98101] },
+      e: { zip: ['98101', { zip: 98101 }] },
```
mikelehen: nice negative cases! :)

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -555,6 +557,58 @@ apiDescribe('Queries', persistence => {
     });
   });
 
+  it('can use IN filters', async () => {
```
mikelehen: Per the conversation at standup, we'll need to figure out a way to make these only run against the emulator for now (e.g. check that we're running in node and the emulator environment variable is set).

packages/firestore/test/integration/api/database.test.ts
thebrianchen: From a screwed up merge master :(

packages/firestore/test/integration/api/query.test.ts
thebrianchen: Inspired by your tests for array-contains!

packages/firestore/src/api/database.ts
```diff
@@ -1936,13 +1968,40 @@ export class Query implements firestore.Query {
...
+          this._query.getRelationOpFilter(RelationOp.ARRAY_CONTAINS_ANY) ||
+          this._query.getRelationOpFilter(RelationOp.IN);
+        const hasInAndArrayContains =
+          (existingFilter === RelationOp.ARRAY_CONTAINS &&
+            filter.op === RelationOp.IN) ||
+          (existingFilter === RelationOp.IN &&
+            filter.op === RelationOp.ARRAY_CONTAINS);
+        // The only permitted combination of these 3 filters is IN with ARRAY_CONTAINS.
+        if (hasInAndArrayContains) {
+          return;
```
mikelehen: We try to avoid early-return except in special cases since it breaks the linear flow of the code.  In this case somebody might add additional validation logic at the end of the method and not realize that it'll get skipped if we hit this case here.  Fortunately there's a super easy solution in this case. Just remove the return.

packages/firestore/src/api/database.ts
```diff
@@ -1936,13 +1968,40 @@ export class Query implements firestore.Query {
...
+          this._query.getRelationOpFilter(RelationOp.ARRAY_CONTAINS) ||
+          this._query.getRelationOpFilter(RelationOp.ARRAY_CONTAINS_ANY) ||
+          this._query.getRelationOpFilter(RelationOp.IN);
+        const hasInAndArrayContains =
+          (existingFilter === RelationOp.ARRAY_CONTAINS &&
+            filter.op === RelationOp.IN) ||
+          (existingFilter === RelationOp.IN &&
+            filter.op === RelationOp.ARRAY_CONTAINS);
+        // The only permitted combination of these 3 filters is IN with ARRAY_CONTAINS.
+        if (hasInAndArrayContains) {
```
mikelehen: I think the following filter will incorrectly be allowed by this code: `.where('foo', 'array_contains', 5).where('bar', 'in', [1, 2, 3]).where('baz', 'in', [4, 5, 6])`

The problem is that existingFilter will be the _first_ of the matching filter types, but there could be more than one (since you can combine in / array-contains)

packages/firestore/src/core/query.ts
```diff
@@ -349,14 +349,16 @@ export class Query {
     return null;
   }
 
-  hasArrayContainsFilter(): boolean {
-    return (
+  getRelationOpFilter(relationOp: RelationOp): RelationOp | null {
```
mikelehen: This is a pretty unusual function contract, returning the relationOp back to you if it's found, else returning null.  Somebody looking at this in isolation would wonder why it doesn't just return a boolean.  I guess this unconventional contract is letting you be clever and save a bit of repetition in the calling code.  If we decided that was warranted, I'd recommend we put a comment here to provide that context...

But really, I think we can probably go back to returning a boolean here and still find a way to make the calling code clean / readable. :-)

packages/firestore/src/core/query.ts
```diff
@@ -540,6 +550,28 @@ export class RelationFilter extends Filter {
...
       );
+    } else if (this.op === RelationOp.IN) {
+      if (this.value instanceof ArrayValue) {
+        return (
+          this.value.internalValue.find(element => element.isEqual(other)) !==
+          undefined
+        );
+      } else {
+        fail('IN filter has invalid value: ' + this.value.toString());
+        return false;
```
mikelehen: You can actually `return fail(...)` here rather than have a misleading `return false;` that'll never get hit.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1038,12 +1038,68 @@ apiDescribe('Validation:', persistence => {
...
+            .where('foo', inOp, [2, 3])
+            .where('foo', 'array-contains', 1)
+        ).not.to.throw();
       }
     );
 
     validationIt(
       persistence,
-      'with non-array values fail for IN and array-contains-any queries.',
+      'cannot have an IN or array-contains-any filter with on-array values.',
```
mikelehen: non-array

packages/firestore/src/api/database.ts
```diff
@@ -1417,18 +1417,31 @@ export class Query implements firestore.Query {
...
     validateExactNumberOfArgs('Query.where', arguments, 3);
     validateDefined('Query.where', 3, value);
-    // Enumerated from the WhereFilterOp type in index.d.ts.
-    const whereFilterOpEnums = ['<', '<=', '==', '>=', '>', 'array-contains'];
-    validateStringEnum('Query.where', whereFilterOpEnums, 2, opStr);
+
+    // TODO(in-queries): Add 'in' and 'array-contains-any' to validation.
+    if (
+      opStr.toString() !== 'in' &&
+      opStr.toString() !== 'array-contains-any'
```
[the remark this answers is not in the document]
thebrianchen: `opStr` is of type `WhereFilterOp`, which does not include the `in` and `array-contains-any` in its type yet b/c this isn't publicly exposed. Is adding a null check here enough?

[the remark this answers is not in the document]
mikelehen: Oh! I see.  `null` was just an example. You'd have the same problem with `undefined` or `{foo: 1}` or whatever...  although it's pretty unlikely that any of these will come up between now and when we can expose these for real, I think using .toString() as a workaround for an (intentional) TypeScript typing issue is misleading.  Perhaps just change these checks to something like `(opStr as unknown) !== 'in'`?

packages/firestore/src/api/database.ts
```diff
@@ -1936,13 +1976,31 @@ export class Query implements firestore.Query {
...
+        }
+        if (conflictingOp === null && isArrayOp) {
+          conflictingOp = this._query.findRelationOpFilter(arrayOps);
+        }
+        if (conflictingOp != null) {
+          // We special case when it's a duplicate op to give a slightly clearer error message.
+          if (conflictingOp === filter.op) {
+            throw new FirestoreError(
+              Code.INVALID_ARGUMENT,
+              'Invalid query. Queries only support a single ' +
```
mikelehen: Consider "You cannot use more than one 'x' filter."

packages/firestore/src/api/database.ts
```diff
@@ -1417,18 +1417,30 @@ export class Query implements firestore.Query {
...
+
     let fieldValue;
     const fieldPath = fieldPathFromArgument('Query.where', field);
     const relationOp = RelationOp.fromString(opStr);
     if (fieldPath.isKeyField()) {
-      if (relationOp === RelationOp.ARRAY_CONTAINS) {
+      if (
+        relationOp === RelationOp.ARRAY_CONTAINS ||
+        relationOp === RelationOp.ARRAY_CONTAINS_ANY ||
+        relationOp === RelationOp.IN
```
rockwotj: @thebrianchen can we update this based on the Android PR?

PR #1844, opened 2019-06-04
Rename value to docValue in RelationFilter

thebrianchen:
Inside `matchesValue()` of `RelationFilter`, the document's value is passed in as `value`. This leads to some confusion because `value` refers to the document's value, whereas `this.value` refers to the `RelationFilter`'s value. Changing `value` to `docValue` should make the code easier to understand.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -533,18 +533,18 @@ export class RelationFilter extends Filter {
     }
   }
 
-  private matchesValue(docValue: FieldValue): boolean {
+  private matchesValue(other: FieldValue): boolean {
     if (this.op === RelationOp.ARRAY_CONTAINS) {
       return (
-        docValue instanceof ArrayValue &&
-        docValue.internalValue.find(element => element.isEqual(this.value)) !==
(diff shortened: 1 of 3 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -533,18 +533,18 @@ export class RelationFilter extends Filter {
     }
   }
 
-  private matchesValue(value: FieldValue): boolean {
+  private matchesValue(docValue: FieldValue): boolean {
```
mikelehen: How about "other"?  This actually matches what we have in [Android](https://github.com/firebase/firebase-android-sdk/blob/59af48cc23b5a23c29720d142f7aa2cd7036916a/firebase-firestore/src/main/java/com/google/firebase/firestore/core/RelationFilter.java#L74) and [iOS](https://github.com/firebase/firebase-ios-sdk/blob/4908f68a8feb0dc91bf6b01d93328508f8be0e4e/Firestore/core/src/firebase/firestore/core/relation_filter.h#L49) and will save you the effort of porting the rename. :-)

PR #1845, opened 2019-06-04
Enable linting and fix lint issues for app, util and logger

Feiyang1:


packages/util/src/crypt.ts
```diff
@@ -154,34 +154,34 @@ export const base64 = {
    *
    * @param input An array of bytes (numbers with
    *     value in [0, 255]) to encode.
-   * @param opt_webSafe Boolean indicating we should use the
+   * @param optwebSafe Boolean indicating we should use the
```
hsubox76: Did you mean to get rid of the `opt_` entirely? https://engdoc.corp.google.com/eng/doc/devguide/js/styleguide.md?cl=head#features-functions-default-parameters

If not, it should still be camelCase I think.

packages/util/src/sha1.ts
```diff
@@ -103,19 +103,19 @@ export class Sha1 {
...
    * Internal compress helper function.
    * @param {!number[]|!Uint8Array|string} buf Block to compress.
-   * @param {number=} opt_offset Offset of the block in the buffer.
+   * @param {number=} optOffset Offset of the block in the buffer.
    * @private
    */
-  compress_(buf, opt_offset?) {
-    if (!opt_offset) {
-      opt_offset = 0;
+  compress_(buf, optOffset?): void {
```
hsubox76: Should just be `offset` if we are following Google style guide (see previous comment)

packages/util/src/sha1.ts
```diff
@@ -190,53 +190,53 @@ export class Sha1 {
     this.chain_[4] = (this.chain_[4] + e) & 0xffffffff;
   }
 
-  update(bytes, opt_length?) {
+  update(bytes, optLength?): void {
```
hsubox76: Same as above, `length`?

packages/util/src/crypt.ts
Feiyang1: Good catch. Updated to `webSafe`.

Feiyang1: I think they mean `option` instead of `optional` which the google styleguide references to, but it is redundant.

PR #1846, opened 2019-06-05
Refactor Query.where()

thebrianchen:
WIP

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -1420,72 +1420,14 @@ export class Query implements firestore.Query {
     // Enumerated from the WhereFilterOp type in index.d.ts.
     const whereFilterOpEnums = ['<', '<=', '==', '>=', '>', 'array-contains'];
     validateStringEnum('Query.where', whereFilterOpEnums, 2, opStr);
-    let fieldValue;
+
     const fieldPath = fieldPathFromArgument('Query.where', field);
     const relationOp = RelationOp.fromString(opStr);
-    if (fieldPath.isKeyField()) {
-      if (relationOp === RelationOp.ARRAY_CONTAINS) {
-        throw new FirestoreError(
-          Code.INVALID_ARGUMENT,
-          "Invalid Query. You can't perform array-contains queries on " +
-            'FieldPath.documentId() since document IDs are not arrays.'
-        );
-      }
-      if (typeof value === 'string') {
-        if (value === '') {
-          throw new FirestoreError(
-            Code.INVALID_ARGUMENT,
-            'Function Query.where() requires its third parameter to be a ' +
-              'valid document ID if the first parameter is ' +
-              'FieldPath.documentId(), but it was an empty string.'
-          );
-        }
-        if (
-          !this._query.isCollectionGroupQuery() &&
-          value.indexOf('/') !== -1
-        ) {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -1915,7 +1857,95 @@ export class Query implements firestore.Query {
     );
   }
 
-  private validateNewFilter(filter: Filter): void {
+  /**
+   * Verifies that the arguments passed in can create a valid filter.
+   */
+  private validateFilter(
```
mikelehen: Perhaps validateFilterArguments() would be a tad more explicit / clear?

packages/firestore/src/api/database.ts
```diff
@@ -1915,7 +1857,95 @@ export class Query implements firestore.Query {
...
+          value
+        );
+      }
+      return fieldValue;
+    }
+
+  /**
+   * Verifies that a filter is valid when combined with existing filters on the query.
+   */
+  private validateQueryWithFilter(filter: Filter): void {
```
mikelehen: Maybe verifyCanAddFilter() ?

packages/firestore/src/api/database.ts
```diff
@@ -1915,7 +1857,95 @@ export class Query implements firestore.Query {
...
+        );
+      }
+    }
+  }
+
+  /**
+   * Given an unknown value, finds the field value.
+   */
+  private findFieldValue( fieldPath: FieldPath,
+    value: unknown): FieldValue {
```
mikelehen: FWIW- This looks like it's not been formatted with prettier...  which is strange, since it's supposed to run automatically when you push... 🤔

packages/firestore/src/api/database.ts
```diff
@@ -1915,7 +1857,95 @@ export class Query implements firestore.Query {
...
+        if (typeof value === 'string') {
+          const path = this._query.path.child(ResourcePath.fromString(value));
+          fieldValue = new RefValue(
+            this.firestore._databaseId,
+            new DocumentKey(path)
+          );
+        } else if (value instanceof DocumentReference) {
+          const ref = value as DocumentReference;
+          fieldValue = new RefValue(this.firestore._databaseId, ref._key);
+        }
```
mikelehen: I'm not sure I like splitting findFiledValue() apart from the validation logic since looking at this function in isolation, I might wonder "What happens if value isn't a string or DocumentReference?" or "What if path is an invalid path?" I can't answer that without knowing that 1) There's a validateFilter() method that detects / prevents those cases and 2) that every piece of code that calls findFieldValue() first calls validateFilter().

So although the code *looks* simpler, I think it's actually more complex because in order to read this function and know if it's implemented correctly, you now need additional context from elsewhere in the code...  and if you wanted to expand this logic somehow, you'd very likely need to modify both findFieldValue() and the validation logic in validateFilter().

Another way to look at it is that this code now has hidden assumptions about the fieldPath and value parameters that aren't obvious from the code or comments.  We usually try to aggressively document and validate these sorts of assumptions by having assertions.  E.g. you'll notice that we have lots of verbose validation / error handling in our public Query class in database.ts...  and then in query.ts (our internal Query implementation) we use assertions in places that we're *relying* on that validation to have already taken place so that the contract is clear and enforced.

PR #1850, opened 2019-06-06
enable eslint for Storage

Feiyang1:
Notes
1. 'no-throw-literal' is disabled. Storage always throws string literal errors. We should throw `Error` object instead, otherwise stack trace is not accessible. 
Changing to throw `Error` is a breaking change, so we will need to align this change with a major release.
2. There are a few floating promises (no error handling). @schmidt-sebastian, Can you review them to see if they are harmless or if we should add error handling ?


packages/storage/src/implementation/metadata.ts
```diff
@@ -37,27 +39,27 @@ export function noXform_(metadata: Metadata, value: any): any {
...
 export { Mappings };
 
 let mappings_: Mappings | null = null;
-
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function xformPath(fullPath: any): string {
-  let valid = type.isString(fullPath);
+  const valid = type.isString(fullPath);
   if (!valid || fullPath.length < 2) {
     return fullPath;
```
Feiyang1: Based on the function, it could return non-string value. Not sure what the right fix is. @schmidt-sebastian, What do you think?

packages/storage/src/implementation/metadata.ts
```diff
@@ -161,57 +167,60 @@ export function downloadUrlFromResourceString(
...
-    let val = p[key];
-    if (key === 'customMetadata') {
-      if (!type.isObject(val)) {
-        throw 'Expected object for \'customMetadata\' mapping.';
-      }
-    } else {
-      if (type.isNonNullObject(val)) {
-        throw "Mapping for '" + key + "' cannot be an object.";
+  for (const key in p) {
+    if (contains(p, key)) {
```
Feiyang1: added the hasownproperty guard. Is it correct? @schmidt-sebastian

schmidt-sebastian:
Yay! Thanks for doing this. Instead of going through every one of your changes one by one, I decided to take this one small notch further: https://github.com/firebase/firebase-js-sdk/pull/1860 

Let me know what you think.

packages/storage/src/implementation/metadata.ts
schmidt-sebastian: I think so. I inline `contains` in my follow-up PR.

packages/storage/src/implementation/metadata.ts
schmidt-sebastian: I am not 100% certain, but I tried to fix the types in this file. I got it to work without `any` in the follow up. Wasn't pleasant :)

packages/storage/package.json
```diff
@@ -10,9 +10,11 @@
     "dist"
   ],
   "scripts": {
+    "lint": "eslint -c .eslintrc.json '**/*.ts' --ignore-path '../../.gitignore'",
+    "lint:fix": "eslint --fix -c .eslintrc.json '**/*.ts'",
     "build": "rollup -c",
     "dev": "rollup -c -w",
-    "test": "run-p test:browser",
+    "test": "run-p lint test:browser",
```
schmidt-sebastian: Note: With this change, it's no longer possible to add `it.only` and run the tests, since the linter will reject the change before the tests run.

packages/storage/src/implementation/args.ts
```diff
@@ -60,49 +64,52 @@ export function validate(name: string, specs: ArgSpec[], passed: IArguments) {
...
 export class ArgSpec {
-  validator: (p1: any) => void;
+  validator: (p1: unknown) => void;
   optional: boolean;
 
-  constructor(validator: (p1: any) => void, opt_optional?: boolean) {
-    let self = this;
-    this.validator = function(p: any) {
+  constructor(validator: (p1: unknown) => void, optional?: boolean) {
+    const self = this;
```
schmidt-sebastian: I removed a bunch of the "self" assignments in the follow-up.

packages/storage/src/implementation/async.ts
```diff
@@ -26,8 +26,9 @@ import * as promiseimpl from './promise_external';
  * into browser code.
  */
 export function async(f: Function): Function {
-  return function(...argsToForward: any[]) {
-    promiseimpl.resolve(true).then(function() {
+  return function(...argsToForward: unknown[]) {
+    // tslint:disable-next-line:no-floating-promises
+    promiseimpl.resolve(true).then(() => {
```
schmidt-sebastian: I think this floating Promises are intended and we don't need to handle them.

schmidt-sebastian:
As commented in the Firestore PR, can you play around with `after-used` (https://eslint.org/docs/rules/no-unused-vars#args) so we don't have to underscores as prefixes  as often?

packages/storage/package.json
Feiyang1: We can add a new command `"test:no-lint": "run-p test:all"` to facilitate local development.

packages/storage/src/implementation/metadata.ts
Feiyang1: Awesome! I was so frustrated by this file that I initially disabled eslint for it entirely

packages/storage/package.json
schmidt-sebastian: Can we just flip them? It allows to run the tests, but still prevents us from checking this in.

Feiyang1: Do you mean make `yarn test` not lint, but make `yarn test:lint` lint?
It's supposed to prevent us from checking code in if any lint rule fails, so CI should always run the version that does linting.
Our CI run `yarn test`, so I'm making lint part of it. Am I understanding your concern correctly?

Feiyang1:
@schmidt-sebastian Do you want to disable `no-used-vars` for function parameters for Storage?

packages/storage/package.json
schmidt-sebastian: This is what I mean:

```
"test": "run-p test:browser lint"
```

That way we run the tests before we fail the lint step. I can ignore the failed lint step in my local development.

schmidt-sebastian:
> @schmidt-sebastian Do you want to disable `no-used-vars` for function parameters for Storage? I disabled it for Firestore.

Cough cough. Yes please :)

packages/storage/package.json
Feiyang1: Ah, yes. I thought about it, but decided to lint first because I want it to fail fast - tests take much longer to finish.
I'm prioritizing CI over local development because travis is very slow and is time consuming to restart. Imagine running tests for 10 minutes only to find that lint failed.

That being said - I think it's okay to flip the order of lint and test since you will run lint locally anyway (just later), so lint should always pass in CI. Though I worry sometimes people forget to run test at all before committing changes, it should be rare.

so.. I updated the test command to `"test": "run-p test:browser lint"`

Feiyang1:
1. Disabled `no-unused-vars` check for function parameters
2. Removed duplicate imports

schmidt-sebastian:
@Feiyang1 Can you resolve the conflicts? I'll review it right after . I am OOO Tuesday though.

Feiyang1:
Rebased on master. @schmidt-sebastian

PR #1851, opened 2019-06-06
Add support for running Firestore integration tests against the emulator

thebrianchen:
Hardcoding a port for the firestore emulator causes collisions with the RTDB emulator in database tests. Instead, find an open port to connect to for testing.

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -48,6 +48,7 @@
     "chai": "4.2.0",
     "chai-as-promised": "7.1.1",
     "indexeddbshim": "4.1.0",
+    "find-free-port": "2.0.0",
     "karma": "4.1.0",
     "karma-chrome-launcher": "2.2.0",
     "karma-cli": "2.0.0",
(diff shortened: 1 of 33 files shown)


packages/firestore/package.json
```diff
@@ -11,7 +11,7 @@
     "lint:fix": "tslint --fix -p tsconfig.json -c tslint.json 'src/**/*.ts' 'test/**/*.ts'",
     "lint:eslint": "eslint -c .eslintrc.json '**/*.ts' --ignore-path '../../.gitignore'",    
     "prettier": "prettier --write 'src/**/*.js' 'test/**/*.js' 'src/**/*.ts' 'test/**/*.ts'",
-    "test": "run-s lint test:all",
+    "test": "yarn test:emulator",
```
Feiyang1: I think we still want to run the browser tests, so we should replace `test:node` with `test:emulator` in the "test:all" instead.

thebrianchen: How does changing `test:node` to `test:emulator` in `test:all` make browser tests run? My understanding is that `test:emulator` runs `test:node` against the emulator rather than prod.

thebrianchen: How does changing test:node to test:emulator in test:all make browser tests run? My understanding is that test:emulator runs test:node against the emulator rather than prod.

Feiyang1: Travis runs `yarn test` which in turn runs `yarn test:all`, so both node and browser tests will run.

PR #1852, opened 2019-06-06
 enable eslint in Performance

Feiyang1:
There is a few floating promises (no error handler). @alikn, Can you check if they are harmless, or if we need to add error handlers for them?


packages/performance/src/resources/trace.test.ts
```diff
@@ -26,7 +26,7 @@ import '../../test/setup';
 describe('Firebase Performance > trace', () => {
   setupApi(window);
   let trace: Trace;
-  const createTrace = () => {
+  const createTrace = (): Trace => {
```
alikn: Is this intended?

Issue #1670, opened 2019-04-08
IOS 12.2 - Safari 12.1 - INTERNAL UNHANDLED ERROR:  ... error ... Indexed Database server


Frou-HOS:
### My environment

  * Operating System version: IOS 12.2
  * Browser version: Mobile Safari 12.1
  * Firebase SDK version: 5.9.3
  * Firebase Product: Firestore
  * JS Framework: Vue.js

### The problem
This problem comming with the IOS 12.2 update.
Every time i send my App to the background, launch an other app and come back to my app, my app freeze due to this error :
- @firebase/firestore: Firestore (5.5.0): INTERNAL UNHANDLED ERROR:  An internal error was encountered in the Indexed Database server

When i disable persitence it seems to be working but it's really slow.
An other error throw after for each action :
- AsyncQueue is already failed: An internal error was encountered in the Indexed Database server

Sentry issue :
https://sentry.io/share/issue/e4a452668dd54497b20c20d720a0aa3e/

#### Steps to reproduce:
Have a snapshot on collection on  your app and enable persistence on
IOS 12.2
Mobile Safari 12.1
Kill all your app
Launch your app -> make some action calling snapshot
Back IOS Home 
Launch an other app (ex: twitter)
Back IOS Home
Lanch your app -> error throw

JeffAtGoog:
Hi Frou-HOS - I was wondering if you might be able to provide some more information to help us reproduce the issue. I have attempted but have been unsuccessful in reproducing the issue you have reported above.

My setup:
- iPhone XS 64 GB
- iOS 12.2 / Safari 12.1
- Firebase SDK 5.9.0
- Simple application that creates a snapshot to a document (also tried to an entire collection)

I followed your steps:

1. Kill all running applications
2. Launch My Application and trigger the snapshot
3. Put application into background
4. Launch Twitter (+ many other applications)
5. Back to iOS home screen
6. Open My Application again

Based on this I was wondering if you may be able to share with us more details on how you are able to reproduce the issue.

If possible please provide the device you are using, how you are establishing the snapshots, how much storage space is free on the device, and if there is anything happening when your application sleeps/resumes.

It would be greatly appreciated if you could create a MVCE (https://stackoverflow.com/help/mcve) for this.

Thanks,
Jeff

jbgomez21:
I have exactly the same problem.
Until now only iOS 12.2.
The problem is with enablePersistence, if it's enable fail approx 50% times

In my case:

```
[...]

PR #1854, opened 2019-06-07
Provide warning and potential workaround for iOS Safari bug described by #1670.

mikelehen:
* On iOS 12.2 we proactively warn about the "internal error was encountered in
  Indexed Database Server" Bug.
* When the bug is actually encountered on iOS >= 12.2 an < 13 we log a custom
  error message.

Both log message and error link to https://stackoverflow.com/q/56496296/110915 which
provides potential error handling / recovery example.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -2,16 +2,17 @@
 # Unreleased
 - [feature] Added `clearPersistence()`, which clears the persistent storage
   including pending writes and cached documents. This is intended to help
-  write reliable tests (#449). 
-- [changed] Added logging / custom error message for iOS safari bug described
-  in #1670.
+  write reliable tests (#449).
+- [changed] Added logging and a custom error message to help users hitting
+  https://bugs.webkit.org/show_bug.cgi?id=197050 (a bug in iOS 12.2 causing
+  the SDK to potentially crash when persistence is enabled).
 
 # 1.3.3
 - [changed] Firestore now recovers more quickly after network connectivity
   changes (airplane mode, Wi-Fi availability, etc.).
 
 # 1.3.0
-- [changed] Deprecated the `experimentalTabSynchronization` setting in favor of 
+- [changed] Deprecated the `experimentalTabSynchronization` setting in favor of
   `synchronizeTabs`. If you use multi-tab synchronization, it is recommended
   that you update your call to `enablePersistence()`. Firestore logs an error
   if you continue to use `experimentalTabSynchronization`.
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -3,6 +3,8 @@
 - [feature] Added `clearPersistence()`, which clears the persistent storage
   including pending writes and cached documents. This is intended to help
   write reliable tests (#449). 
+- [changed] Added logging / custom error message for iOS safari bug described
+  in #1670.
 
 # 1.3.3
 - [changed] Firestore now recovers more quickly after network connectivity
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -15,13 +15,13 @@
  * limitations under the License.
  */
 
+import { getUA } from '@firebase/util';
 import { assert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { debug, error } from '../util/log';
 import { Deferred } from '../util/promise';
 import { SCHEMA_VERSION } from './indexeddb_schema';
 import { PersistencePromise } from './persistence_promise';
-import { getUA } from '@firebase/util';
 
 const LOG_TAG = 'SimpleDb';
 
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -17,10 +17,11 @@
 
 import { assert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
-import { debug } from '../util/log';
+import { debug, error } from '../util/log';
 import { Deferred } from '../util/promise';
 import { SCHEMA_VERSION } from './indexeddb_schema';
 import { PersistencePromise } from './persistence_promise';
+import { getUA } from '@firebase/util';
 
 const LOG_TAG = 'SimpleDb';
 
@@ -149,8 +150,7 @@ export class SimpleDb {
     }
 
     // Check the UA string to find out the browser.
-    // TODO(mikelehen): Move this logic into packages/util/environment.ts
-    const ua = window.navigator.userAgent;
+    const ua = getUA();
 
     // IE 10
     // ua = 'Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.2; Trident/6.0)';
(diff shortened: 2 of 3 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - [feature] Added `clearPersistence()`, which clears the persistent storage
   including pending writes and cached documents. This is intended to help
   write reliable tests (#449). 
+- [changed] Added logging / custom error message for iOS safari bug described
+  in #1670.
```
schmidt-sebastian: Nit: I would prefer if this was a little more self-explanatory.

packages/firestore/src/local/simple_db.ts
```diff
@@ -15,9 +15,10 @@
  * limitations under the License.
  */
 
+import { getUA } from '@firebase/util';
```
schmidt-sebastian: I take it we don't have API reviews for `@firebase/util`? :)

packages/firestore/src/local/simple_db.ts
```diff
@@ -212,7 +217,21 @@ export class SimpleDb {
...
-  constructor(private db: IDBDatabase) {}
+  constructor(private db: IDBDatabase) {
+    const iOSVersion = SimpleDb.getIOSVersion(getUA());
+    // NOTE: According to https://bugs.webkit.org/show_bug.cgi?id=197050, the
+    // bug we're checking for should exist in iOS >= 12.2 and < 13, but for
+    // whatever reason it's much harder to hit after 12.2 so we only proactively
+    // log on 12.2.
+    if (iOSVersion === 12.2) {
+      error(
+        'Firestore persistence on iOS 12.2 suffers from a bug in iOS 12.2 ' +
```
schmidt-sebastian: Do you think mentioning the version number twice is enough? :) I would remove the first `on iOS 12.2`.

packages/firestore/src/local/simple_db.ts
```diff
@@ -212,7 +217,21 @@ export class SimpleDb {
...
+          'Safari that may cause your app to stop working. See ' +
+          'https://stackoverflow.com/q/56496296/110915 for details ' +
+          'and potential workaround.'
```
schmidt-sebastian: s/and potential/and a potential/ seems more grammatically correct.

packages/firestore/src/local/simple_db.ts
```diff
@@ -195,7 +195,12 @@ export class SimpleDb {
...
   static getIOSVersion(ua: string): number {
     const iOSVersionRegex = ua.match(/i(?:phone|pad|pod) os ([\d_]+)/i);
-    const version = iOSVersionRegex ? iOSVersionRegex[1].split('_')[0] : '-1';
+    const version = iOSVersionRegex
+      ? iOSVersionRegex[1]
+          .split('_')
+          .slice(0, 2)
+          .join('.')
+      : '-1';
     return Number(version);
```
schmidt-sebastian: My gut feeling is that version numbers are better represented as strings.

packages/firestore/src/local/simple_db.ts
```diff
@@ -578,7 +599,9 @@ export class SimpleDbStore<
     const cursorRequest = this.cursor({});
     return new PersistencePromise((resolve, reject) => {
       cursorRequest.onerror = (event: Event) => {
-        reject((event.target as IDBRequest).error!);
+        const error = (event.target as IDBRequest).error!;
```
schmidt-sebastian: You could move this cast into `checkForAndReportiOSError`, which could then do an `instanceof` check..

packages/firestore/src/local/simple_db.ts
```diff
@@ -692,7 +715,35 @@ function wrapRequest<R>(request: IDBRequest): PersistencePromise<R> {
...
 }
+
+// Guard so we only report the error once.
+let reportedIOSError = false;
+function checkForAndReportiOSError(error: DOMException): void {
+  if (reportedIOSError) {
+    return;
+  }
+  const iOSVersion = SimpleDb.getIOSVersion(getUA());
+  if (iOSVersion >= 12.2 && iOSVersion < 13) {
```
schmidt-sebastian: 1) Now I understand why we use a version number and not a version number string.
2) Do we need to guard this check on the iOS version? We might just want to log the console message whenever we encounter the matching error string.

packages/firestore/src/local/simple_db.ts
```diff
@@ -692,7 +715,35 @@ function wrapRequest<R>(request: IDBRequest): PersistencePromise<R> {
...
+    const IOS_ERROR =
+      'An internal error was encountered in the Indexed Database server';
+    if (error.message.indexOf(IOS_ERROR) >= 0) {
+      reportedIOSError = true;
+      // Throw a global exception outside of this promise chain, for the user to
+      // potentially catch.
+      setTimeout(() => {
```
schmidt-sebastian: This could be `setImmediate`.

packages/firestore/src/local/simple_db.ts
```diff
@@ -692,7 +715,35 @@ function wrapRequest<R>(request: IDBRequest): PersistencePromise<R> {
...
+        throw new FirestoreError(
+          'internal',
+          `IOS_INDEXEDDB_BUG1: IndexedDb has thrown '${IOS_ERROR}'. This is likely ` +
```
schmidt-sebastian: Is `IOS_INDEXEDDB_BUG1` intentional?

packages/firestore/src/local/simple_db.ts
mikelehen: hah, woops.

packages/firestore/src/local/simple_db.ts
mikelehen: done

packages/firestore/src/local/simple_db.ts
mikelehen: we need the cast result for `reject(error)` too though...  and IDBRequest is an interface not a class, so no instanceof checks. :-/

packages/firestore/src/local/simple_db.ts
mikelehen: Given the iOS error message is so generic, I am a little worried about making this check too open-ended and possibly having it hide future bugs in iOS Safari's IndexedDB implementation (I also named this "IOS_INDEXEDDB_BUG1" to allow for future bugs in the future 😛)...  So I prefer keeping the version scoped for now.

packages/firestore/src/local/simple_db.ts
mikelehen: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

packages/firestore/src/local/simple_db.ts
mikelehen: Yup.  Future proofing for the next iOS bug that bites us. 😬

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: We should use Go-style variable names everywhere :)

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Sounds good. Will bug you about it again when we need to block 12.3.1.

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Sounds good. Glad you didn't use IOS_INDEXEDDB_BUG_001 :)

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Hm. I learned something today.

(Never develop JS for browsers).

PR #1858, opened 2019-06-07
Taking a stab at simplifying validation for array / disjunctive ops.

mikelehen:
> If you can think of a better way to structure the array-contains, in, array-contains-any logic, please advise :)

I accepted this challenge and this is what I came up with (after fiddling around for a while).  If you're on board I can merge it into your PR...  else we can drop it or you can take it and tweak it or whatever...

I also re-arranged the test cases a tad.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -349,12 +349,17 @@ export class Query {
     return null;
   }
 
-  hasRelationOpFilter(relationOp: RelationOp): boolean {
-    return (
-      this.filters.find(
-        filter => filter instanceof RelationFilter && filter.op === relationOp
-      ) !== undefined
-    );
+  // Checks if any of the provided RelationOps are included in the query and
+  // returns the first one that is, or null if none are.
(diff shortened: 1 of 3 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -349,12 +349,17 @@ export class Query {
...
 
-  hasRelationOpFilter(relationOp: RelationOp): boolean {
-    return (
-      this.filters.find(
-        filter => filter instanceof RelationFilter && filter.op === relationOp
-      ) !== undefined
-    );
+  // Checks if any of the provided RelationOps are included in the query and
+  // returns the first one that is, or null if none are.
+  findRelationOpFilter(relationOps: RelationOp[]): RelationOp | null {
```
mikelehen: By the way, I realize this is similar to what you did with getRelationOp() that I asked you to undo...  I _think_ this version is reasonable, since it operates on an array and returns the first match (rather than taking a single op and returning it back to you)...  But I apologize for not suggesting this to you, and sorry if I seem hypocritical / inconsistent. :)

PR #1862, opened 2019-06-10
Enable noImplicitAny Typescript compiler flag

hsubox76:
Enabled noImplicitAny Typescript compiler flag on all packages with an exception for database, which would require extensive updates out of scope for this PR.


packages/storage/index.ts
```diff
@@ -17,7 +17,10 @@
 
 import firebase from '@firebase/app';
 import { FirebaseApp } from '@firebase/app-types';
-import { FirebaseServiceFactory } from '@firebase/app-types/private';
+import {
+  FirebaseServiceFactory,
+  _FirebaseNamespace
+} from '@firebase/app-types/private';
```
schmidt-sebastian: Could this just be `FirebaseNamespace`? It's not immediately obvious to me why it needs an underscore, but I am willing to learn :)

packages/storage/src/implementation/requests.ts
```diff
@@ -163,7 +163,7 @@ export function list(
   pageToken?: string | null,
   maxResults?: number | null
 ): RequestInfo<ListResult> {
-  let urlParams = {};
+  let urlParams: { [key: string]: any } = {};
```
schmidt-sebastian: We have another PR pending that removes `any` from Storage. Can you use `unknown` here?

packages/storage/index.ts
schmidt-sebastian: Hm, I guess we have:

```
import { FirebaseNamespace } from '@firebase/app-types';
import { _FirebaseNamespace } from '@firebase/app-types/private';	
```

hsubox76: Sure, the public `FirebaseNamespace` type doesn't have an `INTERNAL` property which causes a TS error on line 54.

packages/storage/src/implementation/requests.ts
hsubox76: I actually just saw that there was a `UrlParams` type so I used that.  The only change is I have to explicitly cast `maxResults` to string if that's okay.

hsubox76: Oops I guess it wasn't okay.  I changed UrlParams to accept `string | number` values, that seems like a better idea.

packages/util/src/obj.ts
```diff
@@ -18,12 +18,20 @@
 /* eslint-disable */
 // will enable after https://github.com/firebase/firebase-js-sdk/pull/1811 is merged
 
+interface UtilObject<V> {
+  [key: string]: V;
+  [index: number]: V;
```
Feiyang1: No need to define a number index. All number indexing are string indexing under the hood.

packages/app/index.node.ts
```diff
@@ -18,8 +18,11 @@
 import { FirebaseNamespace } from '@firebase/app-types';
 import { _FirebaseNamespace } from '@firebase/app-types/private';
 import { createFirebaseNamespace } from './src/firebaseNamespace';
-import Storage from 'dom-storage';
-import { XMLHttpRequest } from 'xmlhttprequest';
+// Node specific packages.
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const Storage = require('dom-storage');
```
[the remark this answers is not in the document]
hsubox76: `dom-storage` and `xmlhttprequest` don't have types so the type of `Storage` and `XMLHttpRequest` is implicitly `any`.  Maybe it would be simpler to just `@ts-ignore` and explain it is a node package.

packages/app/src/firebaseNamespaceCore.ts
```diff
@@ -80,6 +80,7 @@ export function createFirebaseNamespaceCore(
   //
   //   import * as firebase from 'firebase';
   //   which becomes: var firebase = require('firebase');
+  // @ts-ignore
```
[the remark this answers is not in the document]

packages/functions/src/serializer.ts
```diff
@@ -18,8 +18,8 @@
 const LONG_TYPE = 'type.googleapis.com/google.protobuf.Int64Value';
 const UNSIGNED_LONG_TYPE = 'type.googleapis.com/google.protobuf.UInt64Value';
 
-function mapValues(o: object, f: (any) => any): object {
-  let result = {};
+function mapValues(o: { [key: string]: any }, f: (arg0: any) => any): object {
+  let result: { [key: string]: any } = {};
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/storage/src/implementation/requests.ts
```diff
@@ -427,13 +427,16 @@ export function getResumableUploadStatus(
 
   function handler(xhr: XhrIo, text: string): ResumableUploadStatus {
     let status = checkResumeHeader_(xhr, ['active', 'final']);
-    let sizeString;
+    let sizeString: string | null = null;
     try {
       sizeString = xhr.getResponseHeader('X-Goog-Upload-Size-Received');
+      if (sizeString === null) {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/testing/src/api/index.ts
```diff
@@ -212,7 +212,8 @@ export function loadFirestoreRules(
         project: `projects/${options.projectId}`,
         rules: { files: [{ content: options.rules }] }
       },
-      (err, resp) => {
+      // @ts-ignore Defined in protobuf.
```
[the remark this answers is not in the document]
hsubox76: `resolve()` won't accept `unknown`

packages/util/src/jwt.ts
```diff
@@ -18,26 +18,34 @@
 import { base64Decode } from './crypt';
 import { jsonEval } from './json';
 
+interface Claims {
+  [key: string]: object | number | string | boolean;
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/util/src/jwt.ts
```diff
@@ -100,14 +110,11 @@ export const isValidTimestamp = function(token): boolean {
...
- * @param {?string} token
- * @return {?number}
  */
-export const issuedAtTime = function(token): number | null {
-  const claims = decode(token).claims;
+export const issuedAtTime = function(token: string): number | null {
+  const claims: Claims = decode(token).claims;
   if (typeof claims === 'object' && claims.hasOwnProperty('iat')) {
-    return claims['iat'];
+    return claims['iat'] as number | null;
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/util/src/obj.ts
hsubox76: Weird, I was previously getting a lot of errors in places that used this functions with number-indexed objects but I guess those errors got fixed on the consuming end.  Done.

PR #1864, opened 2019-06-11
enable eslint for functions

Feiyang1:


packages/functions/src/api/error.ts
```diff
@@ -146,7 +148,7 @@ export function _errorForResponse(
       }
       // TODO(klimt): Add better default descriptions for error enums.
       // The default description needs to be updated for the new code.
-      description = status;
+      description = status as string;
```
Feiyang1: @bklimt Can `status` be anything other than string (`description` is defined as a string)? I'm confused as we check if `status` is a string at line 142.

bklimt: As far as the API is concerned, it's supposed to be a string, but this is the code that should be enforcing that, not assuming it. It looks like this line should be moved up into the if-statement above.

packages/functions/src/api/service.ts
```diff
@@ -34,7 +32,17 @@ import { Serializer } from '../serializer';
  */
 interface HttpResponse {
   status: number;
-  json: any;
+  json: HttpResponseJson | null;
+}
+
+export interface HttpResponseJson {
```
bklimt: 1. Please add a comment to this interface.
2. I don't know why this file has `Http` vs `JSON` capitalization everywhere, but this interface isn't consistent with it. I hesitate to suggest renaming it to `HttpResponseJSON`, though. That's too horrible. :-(
3. In the comment, I would explicitly say that this interface is to make functions that would otherwise take `{}` able to access the possible elements in the body more easily. This interface doesn't actually make any stronger type guarantee than `{}` would.

I don't really love this interface, but I can't think of any nicer alternative.

packages/functions/src/api/service.ts
```diff
@@ -119,7 +127,7 @@ export class Service implements FirebaseFunctions, FirebaseService {
    * @param name The name of the trigger.
    */
   httpsCallable(name: string, options?: HttpsCallableOptions): HttpsCallable {
-    let callable = <HttpsCallable>(data?: any) => {
+    const callable = (data?: {}): Promise<HttpsCallableResult> => {
```
bklimt: `data?:` should be `{} | null`, not `{}`. I'm surprised this works, since there's a test that calls it with `null`.

packages/functions/src/context.ts
```diff
@@ -50,9 +49,11 @@ export class ContextProvider {
     try {
       // HACK: Until we have a separate instanceId package, this is a quick way
       // to load in the messaging instance for this app.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
bklimt: Do we still not have a separate instance id package we could depend on?

packages/functions/src/serializer.ts
```diff
@@ -31,8 +31,8 @@ function mapValues(o: object, f: (any) => any): object {
 export class Serializer {
   // Takes data and encodes it in a JSON-friendly way, such that types such as
   // Date are preserved.
-  encode(data: any): any {
-    if (data === null || data === undefined) {
+  encode(data: unknown): unknown {
+    if (data == null) {
```
bklimt: `==` 😱

This seems less readable to me, but if the linter demands it, fine.

packages/functions/src/serializer.ts
```diff
@@ -53,27 +53,27 @@ export class Serializer {
...
     }
     // If we got this far, the data is not encodable.
     throw new Error('Data cannot be encoded in JSON: ' + data);
   }
 
   // Takes data that's been encoded in a JSON-friendly form and returns a form
   // with richer datatypes, such as Dates, etc.
-  decode(json: any): any {
-    if (json === null) {
+  decode(json: unknown): unknown {
```
bklimt: This should be `decode(json: {} | null): {} | null {`. The callers already seem fine with that. Passing in `undefined` would've thrown anyway.

packages/functions/src/serializer.ts
```diff
@@ -53,27 +53,27 @@ export class Serializer {
...
+    if (json == null) {
```
bklimt: This is a change in behavior. But if the type `json` is changed to `{} | null`, it'll be fine.

packages/functions/src/serializer.ts
```diff
@@ -53,27 +53,27 @@ export class Serializer {
...
       return json;
     }
-    if (json['@type']) {
-      switch (json['@type']) {
+    if ((json as {})['@type']) {
```
bklimt: You shouldn't need all these casts if the type is changed as suggested above.

packages/functions/src/api/service.ts
Feiyang1: Gotcha. It works because `HttpsCallable` is defined in `functions-types` which still has `any` type. :( 
I will update `functions-types` as well.

packages/functions/src/api/service.ts
Feiyang1: Updated the interface name to `HttpResponseBody`. WDYT?

packages/functions/src/context.ts
Feiyang1: Yes, we have `installations` package which is the replacement for IID. I'm not sure if we can start using it in functions. We should talk about it. @mmermerkaya

packages/functions/src/serializer.ts
Feiyang1: it's allowed but not demanded. I'm not a fan of it, but we as a team decided to allow it for this specific use case ( Google styleguide also allows it).
I just updated it for the sake of consistency.

packages/functions/src/serializer.ts
Feiyang1: Good point. updated.

packages/functions/src/api/service.ts
bklimt: Very clever. I love it.

packages/functions/src/context.ts
bklimt: Let's punt that to a future PR anyway.

packages/functions/src/api/service.ts
```diff
@@ -32,10 +32,14 @@ import { Serializer } from '../serializer';
  */
 interface HttpResponse {
   status: number;
-  json: HttpResponseJson | null;
+  json: HttpResponseBody | null;
 }
-
-export interface HttpResponseJson {
+/**
+ * Describe the shape of the HttpResponse body.
```
bklimt: s/Describe/Describes/

Feiyang1: done

packages/functions/test/browser/callable.test.ts
```diff
@@ -18,36 +18,16 @@ import { expect } from 'chai';
 import * as sinon from 'sinon';
 import { FirebaseApp } from '@firebase/app-types';
 import { _FirebaseApp } from '@firebase/app-types/private';
-import { HttpsError, FunctionsErrorCode } from '@firebase/functions-types';
 import firebase from '@firebase/app';
-import '@firebase/messaging';
 import { Service } from '../../src/api/service';
+
+/* eslint-disable import/no-duplicates */
+import '@firebase/messaging';
```
hsubox76: Just curious why the duplicate imports are needed.

Feiyang1: Technically, it's not needed, but I worry that hiding the bare import with the named import makes the code less clear and more error prone. E.g. people removes the named import without knowing that we rely on the side effect of the import.

PR #1868, opened 2019-06-11
Check that self.firebase is not undefined

### LINKED ISSUE #1867: CJS and ESM builds of @firebase/app throw when used in a service worker bundle
## Description of the Issue

The dist files `@firebase/app/dist/index.cjs.js` and `@firebase/app/dist/index.esm.js` both contain code that causes an FCM service worker to fail to load.

[Firebase Lite detection](https://github.com/firebase/firebase-js-sdk/blob/76539be9b3ab19f5be70275f2334bee9b022e3c4/packages/app/index.ts#L26) uses `'firebase' in self` to check if Firebase is in the service worker global scope. However, in `index.cjs.js` and `index.esm.js`, the compiler uses the `var` keyword to declare `firebase`, which is hoisted and initialized as `undefined`, so `'firebase' in self` evaluates to `true`.

This causes `var sdkVersion = self.firebase.SDK_VERSION` to throw because `self.firebase` is `undefined`, and the service worker fails to register.

This bug is not present in the code distributed via the gstatic CDN because that version is compiled into a module loader which does not declare `firebase` using `var`.

I have included code from the relevant files on unpkg.com below, and a link to the problem code on GitHub.

## Motivation for or Use Case

Firebase Cloud Messaging won't work if using code from NPM.

## Environment Configuration

This bug is present in `firebase@6.1.1` and `firebase@5.11.1`, but not present in `firebase@5.9.4`.

### [index.cjs.js](https://unpkg.com/@firebase/app@0.4.4/dist/index.cjs.js)

```javascript
// Firebase Lite detection
// BUG: firebase is already initialized as undefined
if (util.isBrowser() && 'firebase' in self) {
  logger.warn(
    '\n    Warning: Firebase is already defined in the global scope. Please make sure\n    Firebase library is only loaded once.\n  ',
  )
  var sdkVersion = self.firebase.SDK_VERSION
  if (sdkVersion && sdkVersion.indexOf('LITE') >= 0) {
    logger.warn(
      '\n    Warning: You are trying to load Firebase while using Firebase Performance standalone script.\n    You should load Firebase Performance with this instance of Firebase to avoid loading duplicate code.\n    ',
    )
  }
}
var firebaseNamespace = createFirebaseNamespace()
var initializeApp = firebaseNamespace.initializeApp
firebaseNamespace.initializeApp = function() {
  // Environment check before initializing app
  // Do the check in initializeApp, so people have a chance to disable it by setting logLevel
  // in @firebase/logger
  if (util.isNode()) {
    logger.warn(
      '\n      Warning: This is a browser-targeted Firebase bundle but it appears it is being\n      run in a Node environment.  If running in a Node environment, make sure you\n      are using the bundle specified by the "main" field in package.json.\n      \n      If you are using Webpack, you can specify "main" as the first item in\n      "resolve.mainFields":\n      https://webpack.js.org/configuration/resolve/#resolvemainfields\n      \n      If using Rollup, use the rollup-plugin-node-resolve plugin and set "module"\n      to false and "main" to true:\n      https://github.com/rollup/rollup-plugin-node-resolve\n      ',
    )
  }
  return initializeApp.apply(undefined, arguments)
}
// BUG: var firebase is hoisted and initialized as undefined
var firebase = firebaseNamespace
```

### [index.esm.js](https://unpkg.com/@firebase/app@0.4.4/dist/index.esm.js)

```javascript
// Firebase Lite detection
if (isBrowser() && 'firebase' in self) {
  logger.warn(
    '\n    Warning: Firebase is already defined in the global scope. Please make sure\n    Firebase library is only loaded once.\n  ',
  )
  var sdkVersion = self.firebase.SDK_VERSION
  if (sdkVersion && sdkVersion.indexOf('LITE') >= 0) {
    logger.warn(
      '\n    Warning: You are trying to load Firebase while using Firebase Performance standalone script.\n    You should load Firebase Performance with this instance of Firebase to avoid loading duplicate code.\n    ',
    )
  }
}
var firebaseNamespace = createFirebaseNamespace()
var initializeApp = firebaseNamespace.initializeApp
firebaseNamespace.initializeApp = function() {
  // Environment check before initializing app
  // Do the check in initializeApp, so people have a chance to disable it by setting logLevel
  // in @firebase/logger
  if (isNode()) {
    logger.warn(
      '\n      Warning: This is a browser-targeted Firebase bundle but it appears it is being\n      run in a Node environment.  If running in a Node environment, make sure you\n      are using the bundle specified by the "main" field in package.json.\n      \n      If you are using Webpack, you can specify "main" as the first item in\n      "resolve.mainFields":\n      https://webpack.js.org/configuration/resolve/#resolvemainfields\n      \n      If using Rollup, use the rollup-plugin-node-resolve plugin and set "module"\n      to false and "main" to true:\n      https://github.com/rollup/rollup-plugin-node-resolve\n      ',
    )
  }
  return initializeApp.apply(undefined, arguments)
}
var firebase = firebaseNamespace
```

https://github.com/firebase/firebase-js-sdk/blob/76539be9b3ab19f5be70275f2334bee9b022e3c4/packages/app/index.ts#L25-L68

jacksteamdev:
Fixes #1867.

### DIFF
--- a/packages/app/index.ts
+++ b/packages/app/index.ts
@@ -23,12 +23,8 @@ import { Logger } from '@firebase/logger';
 const logger = new Logger('@firebase/app');
 
 // Firebase Lite detection
-if (
-  isBrowser() &&
-  'firebase' in self &&
-  // Check that firebase is not undefined
-  ((self as any).firebase as FirebaseNamespace) !== undefined
-) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+if (isBrowser() && (self as any).firebase !== undefined) {
   logger.warn(`
     Warning: Firebase is already defined in the global scope. Please make sure
     Firebase library is only loaded once.
--- a/packages/app/index.ts
+++ b/packages/app/index.ts
@@ -23,7 +23,12 @@ import { Logger } from '@firebase/logger';
 const logger = new Logger('@firebase/app');
 
 // Firebase Lite detection
-if (isBrowser() && 'firebase' in self) {
+if (
+  isBrowser() &&
+  'firebase' in self &&
+  // Check that firebase is not undefined
+  ((self as any).firebase as FirebaseNamespace) !== undefined
+) {
   logger.warn(`
     Warning: Firebase is already defined in the global scope. Please make sure
     Firebase library is only loaded once.


jacksteamdev:
I signed the CLA!

packages/app/index.ts
```diff
@@ -23,7 +23,12 @@ import { Logger } from '@firebase/logger';
 const logger = new Logger('@firebase/app');
 
 // Firebase Lite detection
-if (isBrowser() && 'firebase' in self) {
+if (
+  isBrowser() &&
+  'firebase' in self &&
+  // Check that firebase is not undefined
+  ((self as any).firebase as FirebaseNamespace) !== undefined
```
Feiyang1: `'firebase' in self` becomes redundant, so we can remove it.

Can you add `// elsint-disable-next-line @typescript-eslint/no-explicit-any` on the top of the line to pass lint ?
Also `as FirebaseNamespace` is not necessary, so it can be dropped.

PR #1872, opened 2019-06-11
Migrate messaging to eslint

Feiyang1:
Let me know if you want to enable/disable additional rules for messaging.

### DIFF
--- a/packages/app/.eslintrc.json
+++ b/packages/app/.eslintrc.json
@@ -2,8 +2,5 @@
     "extends": "../../config/.eslintrc.json",
     "parserOptions": {
         "project": "tsconfig.json"
-    },
-    "rules": {
-        "import/no-default-export": "off"
     }
 }
--- a/packages/messaging/src/controllers/sw-types.ts
+++ b/packages/messaging/src/controllers/sw-types.ts
@@ -23,7 +23,7 @@
  * project, we have to manually declare the web worker types we need.
  */
 
-// tslint:disable no-any These types are from TS.
+ /* eslint-disable @typescript-eslint/no-explicit-any, These types are from TS */
 
 // Not the whole interface, just the parts we're currently using.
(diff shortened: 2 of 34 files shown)


packages/app/.eslintrc.json
```diff
@@ -2,5 +2,8 @@
     "extends": "../../config/.eslintrc.json",
     "parserOptions": {
         "project": "tsconfig.json"
+    },
+    "rules": {
+        "import/no-default-export": "off"
```
mmermerkaya: Unless it's used everywhere (I didn't check), I'd recommend adding a `// eslint:disable-next-line` (this is TSLint format, but you know what I mean) comment to the default export lines instead of disabling the rule for the whole package.

packages/messaging/src/controllers/sw-types.ts
```diff
@@ -63,7 +63,7 @@ interface ClientQueryOptions {
 }
 
 interface WindowClient extends Client {
-  readonly ancestorOrigins: ReadonlyArray<string>;
+  readonly ancestorOrigins: readonly string[];
```
mmermerkaya: Nice. Is this checked by ESLint?

Feiyang1: I didn't touch it, so it must be auto fixed.

packages/app/.eslintrc.json
Feiyang1: That's true. Will update before merge.

packages/messaging/src/controllers/sw-types.ts
mmermerkaya: Cool. It's a TS 3.4 feature, good to see ESLint being quick to update stuff like this. It has a lot of extra rules that we didn't have in TSLint as well. Initially I had my doubts, but I'm sold on ESLint. :smile: :+1:

PR #1877, opened 2019-06-12
enable eslint for installations

Feiyang1:


config/.eslintrc.json
```diff
@@ -17,8 +17,9 @@
     "overrides": [
         {
             "files": [
-                "*test.ts",
-                "**/test/**/*.ts"
+                "**/*.test.ts",
+                "**/test/**/*.ts",
+                "**/testing/**/*.ts"
```
mmermerkaya: I see code under `testing/` as regular code that is used for testing. I don't think we should disable these two rules under `testing/`.

packages/installations/src/helpers/idb-manager.ts
```diff
@@ -33,6 +33,8 @@ const dbPromise: Promise<DB> = openDb(
     switch (upgradeDB.oldVersion) {
       case 0:
         upgradeDB.createObjectStore(OBJECT_STORE_NAME);
+        break;
```
mmermerkaya: Lack of `break` statements here is intentional. Please see the comment above the `switch` statement.

Feiyang1: oops. will fix.

config/.eslintrc.json
Feiyang1: Okay. will remove it.

PR #1879, opened 2019-06-13
Update React Native AsyncStorage import

hsubox76:
As pointed out in https://github.com/firebase/firebase-js-sdk/issues/1847, React Native is deprecating its included `AsyncStorage` as of v0.59 and asking users to import this dependency from `@react-native-community/async-storage` instead.

There doesn't seem to be a reliable way to detect React Native version in runtime so this change tries to `require` the new package, and if not found, falls back to `require`ing the deprecated included package, which should result in the user seeing the console warning from React Native.

We should also add something in our documentation letting our React Native users know they will have to install this package.

Tested with current version of React Native: if `@react-native-community/async-storage` is installed in the test project, there is no error or warning, if it is not installed, React Native's deprecation warning appears.


leochoo:
This error completely broken down my dev environment. I deleted yarn, node_modules, reset metro, clear cache, AND re-created a completely new project. But still, It is showing me the error `Unable to resolve module react-native-community/async-storage` 
How is this even possible? Where is this thing locating the package I deleted??

HarisSpahija:
Same issue here, havent touched the project in a couple of weeks, no changes but still have an issue with async-storage. I am not even using the library, so it must be used as a dependency by a package.

hsubox76:
I've added a reply to both of your comments in the Github issue (https://github.com/firebase/firebase-js-sdk/issues/1898). Can you leave further comments in the Github issue for this?  Thanks!

SasankhM:
Facing the same issue which has totally halted all work. Please fix ASAP!

sertony:
It seems like @hsubox76 just forgot to add `.default` at the end here when importing via `require`:
https://github.com/firebase/firebase-js-sdk/pull/1879/files#diff-024dba18a8ed0c50f8d5b08c7f14d9ffR32

PR #1886, opened 2019-06-15
Overhaul Documents

wilhuff:
Remove different internal types for `MaybeDocument`, `Document`, `NoDocument`,
and `UnknownDocument`. These are all now just `Document`.

Also make Documents non-nullable nearly everywhere.

The change introduces a `DocumentType` enum and maps the old classes like
so:

| Before          | After                         |
| --------------- | ----------------------------- |
| `null`            | `DocumentType.UNKNOWN`          |
| `UnknownDocument` | `DocumentType.CONTENTS_UNKNOWN` |
| `NoDocument`      | `DocumentType.MISSING`          |
| `Document`        | `DocumentType.EXISTING`         |

Part of why this change works is that most of the system only cares
whether or not a document exists. We have several refinements of
non-existing documents for the purposes of caching but everywhere else
we can ignore it and just check `document.exists`. The `instanceof` checks
were usually over-constraining things.

The motivation for this change is in two parts:

  * It's always bugged me that we had so many `instanceof` checks around
  * In C++ we don't have `instanceof` anyway, so we have to do something
    else.

Also, this definitely addresses b/32275378.

### DIFF
--- a/packages/firestore/src/model/document.ts
+++ b/packages/firestore/src/model/document.ts
@@ -24,55 +24,162 @@ import { FieldPath } from './path';
 
 import * as api from '../protos/firestore_proto_api';
 
-export interface DocumentOptions {
-  hasLocalMutations?: boolean;
+export interface MissingDocumentOptions {
   hasCommittedMutations?: boolean;
 }
 
-/**
- * The result of a lookup for a given path may be an existing document or a
- * marker that this document does not exist at a given version.
- */
-export abstract class MaybeDocument {
-  constructor(readonly key: DocumentKey, readonly version: SnapshotVersion) {}
-
-  static compareByKey(d1: MaybeDocument, d2: MaybeDocument): number {
-    return DocumentKey.comparator(d1.key, d2.key);
-  }
+export interface DocumentOptions {
+  hasLocalMutations?: boolean;
+  hasCommittedMutations?: boolean;
 
   /**
-   * Whether this document had a local mutation applied that has not yet been
-   * acknowledged by Watch.
+   * Memoized serialized form of the document for optimization purposes (avoids
+   * repeated serialization). Might be undefined.
    */
-  abstract get hasPendingWrites(): boolean;
+  memoizedProto?: api.Document;
+}
 
-  abstract isEqual(other: MaybeDocument | null | undefined): boolean;
+/**
+ * At a high level, Documents either exist or don't exist. However, when they
+ * don't exist there are different reasons why they don't exist and we treat
+ * these all differently in the cache.
+ */
+export enum DocumentType {
+  /**
+   * A document that is unknown to the client. It doesn't exist locally but
+   * may or may not exist on the server.
+   */
+  UNKNOWN,
 
-  abstract toString(): string;
+  /**
+   * A special case of UNKNOWN document whose contents are unknown but that is
+   * known to exist on the server. This kind of document comes about when the
+   * server accepts an update (i.e. a mutation with a precondition that the
+   * document exists) but the client didn't have any base document to which
+   * the mutation applied.
+   *
+   * Note that even though the server considers the document to exist, the
+   * client still presents a local view as if it didn't exist. This type exists
+   * to allow the client to record the version at which the server said the
+   * document did exist, so that we can avoid flickering back to old versions
+   * while listening for updates.
+   */
+  CONTENTS_UNKNOWN,
+
+  /**
+   * A document that's definitely known not to exist, either because the server
+   * said it was missing or because a local mutation deleted it.
+   */
+  MISSING,
+
+  /**
+   * A document that's known to exist and we know what's inside it.
+   */
+  EXISTS
+}
+
+enum DocumentState {
+  /** No mutations applied. Document was sent to us by Watch. */
+  SYNCED,
+
+  /**
+   * Local mutations applied via the mutation queue. Document is potentially
+   * inconsistent.
+   */
+  LOCAL_MUTATIONS,
+
+  /**
+   * Mutations applied based on a write acknowledgment. Document is potentially
+   * inconsistent.
+   */
+  COMMITTED_MUTATIONS
 }
 
 /**
  * Represents a document in Firestore with a key, version, data and whether the
  * data has local mutations applied to it.
+ *
+ *
+ * A class representing a deleted document.
+ * Version is set to 0 if we don't point to any specific time, otherwise it
+ * denotes time we know it didn't exist at.
+ *
+ * A class representing an existing document whose data is unknown (e.g. a
+ * document that was updated without a known base document).
  */
-export class Document extends MaybeDocument {
-  readonly hasLocalMutations: boolean;
-  readonly hasCommittedMutations: boolean;
+export class Document {
+  static unknown(key: DocumentKey): Document {
+    return new Document(
+      DocumentType.UNKNOWN,
+      key,
+      SnapshotVersion.MIN,
+      ObjectValue.EMPTY,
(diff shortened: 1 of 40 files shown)


wilhuff:
Note that this PR is requesting feedback on the idea in here as well. I think this makes things considerably easier to reason about internally and it will be definitely make the C++ implementation easier (no subclasses make things so much easier), but I wanted to vet this outside the noise of actually porting iOS Objective-C code to C++.

packages/firestore/src/model/document.ts
```diff
@@ -100,85 +207,68 @@ export class Document extends MaybeDocument {
...
-  ) {
-    super(key, version);
-    this.hasCommittedMutations = !!(options && options.hasCommittedMutations);
+  get exists(): boolean {
+    return this.type === DocumentType.EXISTS;
   }
 
-  toString(): string {
-    return `NoDocument(${this.key}, ${this.version})`;
+  get missing(): boolean {
```
mikelehen: isMissing?

packages/firestore/src/model/document.ts
```diff
@@ -24,55 +24,162 @@ import { FieldPath } from './path';
...
 /**
- * The result of a lookup for a given path may be an existing document or a
- * marker that this document does not exist at a given version.
+ * At a high level, Documents either exist or don't exist. However, when they
+ * don't exist there are different reasons why they don't exist and we treat
+ * these all differently in the cache.
  */
-export abstract class MaybeDocument {
-  constructor(readonly key: DocumentKey, readonly version: SnapshotVersion) {}
+export enum DocumentType {
```
mikelehen: I would put MISSNG / EXISTS at the top of this enum, since they are the simple cases...

packages/firestore/src/model/document.ts
```diff
@@ -100,85 +207,68 @@ export class Document extends MaybeDocument {
...
 
-  constructor(
-    key: DocumentKey,
-    version: SnapshotVersion,
-    options?: DocumentOptions
-  ) {
-    super(key, version);
-    this.hasCommittedMutations = !!(options && options.hasCommittedMutations);
+  get exists(): boolean {
+    return this.type === DocumentType.EXISTS;
```
mikelehen: Naively I would think that CONTENTS_UNKNOWN should return true.  If it shouldn't, then perhaps `.exists` is misnamed and really means something else?

packages/firestore/src/model/document.ts
```diff
@@ -100,85 +207,68 @@ export class Document extends MaybeDocument {
...
 
-  toString(): string {
-    return `NoDocument(${this.key}, ${this.version})`;
+  get missing(): boolean {
+    return this.type === DocumentType.MISSING;
   }
 
-  get hasPendingWrites(): boolean {
-    return this.hasCommittedMutations;
+  get unknown(): boolean {
```
mikelehen: isUnknown?  Though I'm not sure that's descriptive enough...  I'm not immediately thinking of a better name though.

packages/firestore/src/model/document.ts
```diff
@@ -100,85 +207,68 @@ export class Document extends MaybeDocument {
...
+    return this.type === DocumentType.UNKNOWN;
   }
 
-  isEqual(other: MaybeDocument | null | undefined): boolean {
+  /**
+   * Returns true if the document is of some definite type: either existing
+   * or definitely missing.
+   */
+  get definite(): boolean {
```
mikelehen: I have a really hard time guessing what `if (fooDoc.definite) { ... }` means.

In general, I'm nervous about all of these properties (`exists`, `missing`, `unknown`, `definite`) because I'm worried they could be confusing / misleading.  We're mapping 4 DocumentTypes (actually 3 since CONTENTS_UNKNOWN doesn't show up) onto 4 helper properties and the mapping isn't immediately obvious to me from the names.  You might expect that `!exists` implies `missing` but that would be wrong. 

There may be better terminology that makes everything clearer, but failing that I would be inclined to just drop them all and move the explicit `DocumentType` checks into the calling code...  if that's too onerous though, we can work harder on the names / comments.

packages/firestore/src/model/document.ts
```diff
@@ -24,55 +24,162 @@ import { FieldPath } from './path';
...
-export class Document extends MaybeDocument {
-  readonly hasLocalMutations: boolean;
-  readonly hasCommittedMutations: boolean;
+export class Document {
+  static unknown(key: DocumentKey): Document {
+    return new Document(
+      DocumentType.UNKNOWN,
+      key,
+      SnapshotVersion.MIN,
+      ObjectValue.EMPTY,
```
mikelehen: Per my comments elsewhere I think I would prefer if we made .data throw for missing / unknown documents (and optionally added `.dataOrEmptyObject` or something so you can explicitly opt into `ObjectValue.EMPTY`

packages/firestore/src/model/document.ts
```diff
@@ -100,85 +207,68 @@ export class Document extends MaybeDocument {
...
-}
 
-/**
- * A class representing an existing document whose data is unknown (e.g. a
- * document that was updated without a known base document).
- */
-export class UnknownDocument extends MaybeDocument {
-  constructor(key: DocumentKey, version: SnapshotVersion) {
-    super(key, version);
+  get hasLocalMutations(): boolean {
```
mikelehen: If possible I think I'd like hasLocalMutations to throw for unknown / missing documents and hasCommittedMutations to throw for unknown documents...  We can loosen if/when we need to, but ensuring code isn't checking for impossible things would be nice.

wilhuff:
As I'm reviewing your feedback in occurs to me that an assumption underlying this change was that in practice, most of our code doesn't care about whether or not a document actually exists so long as non-existing document and existing documents behavior uniformly. The uniformity would be a feature of the change, not a bug.

Building from there, whether or not a document actually exists or not is just additional metadata that we happen to attach to the document.

For example, this change makes most mutations simpler because no matter the actual state of the document you can apply the mutation and any tests around it.

Of course, there are exceptions:

  * Patch mutations have different behavior for documents that exist vs not.
  * Queries only want to return known existing documents
  * Document gets want known existing or known not existing documents
  * etc

However, naming concerns aside, most of your feedback is essentially pushing back toward non-uniformity or overspecificity. This [comment](https://github.com/firebase/firebase-js-sdk/pull/1886#discussion_r294601169) encapsulates this disconnect.

My contention is that this code has two behaviors based on whether or not the document exists or not. Which form of non-existence doesn't matter here so there's no reason to apply a constraint.

If we can't agree that this basic premise is worthwhile then this change isn't really worth fixing. It's gone way too far toward that goal to bring back to true.

mikelehen:
Hrm.  I'm not 100% sure if I'm interpreting your comment correctly.  When you say "so long as non-existing document and existing documents behavior uniformly" I'm not sure if you mean "non-existing documents behave uniformly and existing documents behave uniformly" or "(non-existing documents and existing documents) behave uniformly" and in either case I'm not sure if you are considering unknown documents to be included in "non-existing documents"...

FWIW- In the example [you linked](https://github.com/firebase/firebase-js-sdk/pull/1886/files/622d9d521e9bf607efed4ac9869fe3bce040248f#diff-3cb595fdd1282a582c6e6e7310e1d571L50): 
1. I think treating unknown the same as missing would be incorrect.  We should only use a baseVersion of 0 if the document really is a missing document.
2. The calling code already protects us from .unknown documents, but because that's no longer enforced by the type system, it's no longer obvious / self-documenting.
3. If part of our goal is to get rid of instanceof checks, we could add .version and .exists (or .isDocument or whatever) to MaybeDocument.

Anyway, let's chat in person after sprint planning.

wilhuff:
As discussed, this approach isn't going to work.

PR #1892, opened 2019-06-18
Fix a few linter issues

mmermerkaya:


mmermerkaya:
The messaging changes are mostly the "convert all to async functions" button, not me :smile:.

PR #1894, opened 2019-06-19
Refactor Filters

wilhuff:
This is a bunch of related changes:

  * Extract specific filter types for each special kind of operation
  * Rename things to match
  * Simplify

Each commit describes what it does individually.

cc @thebrianchen

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -110,7 +110,7 @@ export class Query {
   addFilter(filter: Filter): Query {
     assert(
       this.getInequalityFilterField() == null ||
-        !(filter instanceof RelationFilter) ||
+        !(filter instanceof FieldFilter) ||
         !filter.isInequality() ||
         filter.field.isEqual(this.getInequalityFilterField()!),
       'Query must only have one inequality field.'
@@ -342,7 +342,7 @@ export class Query {
 
   getInequalityFilterField(): FieldPath | null {
     for (const filter of this.filters) {
-      if (filter instanceof RelationFilter && filter.isInequality()) {
+      if (filter instanceof FieldFilter && filter.isInequality()) {
         return filter.field;
       }
     }
@@ -353,7 +353,7 @@ export class Query {
   // returns the first one that is, or null if none are.
   findRelationOpFilter(relationOps: RelationOp[]): RelationOp | null {
     for (const filter of this.filters) {
-      if (filter instanceof RelationFilter) {
+      if (filter instanceof FieldFilter) {
         if (relationOps.indexOf(filter.op) >= 0) {
           return filter.op;
         }
@@ -465,7 +465,7 @@ export abstract class Filter {
       }
       return new NanFilter(field);
     } else {
-      return new RelationFilter(field, op, value);
+      return new FieldFilter(field, op, value);
     }
   }
 }
@@ -514,7 +514,7 @@ export class RelationOp {
   }
 }
 
-export class RelationFilter extends Filter {
+export class FieldFilter extends Filter {
   constructor(
     public field: FieldPath,
     public op: RelationOp,
@@ -619,7 +619,7 @@ export class RelationFilter extends Filter {
   }
 
   isEqual(other: Filter): boolean {
-    if (other instanceof RelationFilter) {
+    if (other instanceof FieldFilter) {
       return (
         this.op.isEqual(other.op) &&
         this.field.isEqual(other.field) &&
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -476,6 +476,8 @@ export abstract class Filter {
         );
       }
       return new NanFilter(field);
+    } else if (op === Operator.ARRAY_CONTAINS) {
+      return new ArrayContainsFilter(field, value);
     } else {
       return new FieldFilter(field, op, value);
     }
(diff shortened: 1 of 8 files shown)


packages/firestore/src/remote/serializer.ts
```diff
@@ -1229,28 +1230,28 @@ export class JsonProtoSerializer {
...
-        return RelationOp.ARRAY_CONTAINS;
+        return Operator.ARRAY_CONTAINS;
       case 'IN':
-        return RelationOp.IN;
+        return Operator.IN;
       case 'ARRAY_CONTAINS_ANY':
-        return RelationOp.ARRAY_CONTAINS_ANY;
+        return Operator.ARRAY_CONTAINS_ANY;
       case 'OPERATOR_UNSPECIFIED':
         return fail('Unspecified relation');
```
mikelehen: relation => operator ? (same 2 lines down)

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -845,7 +845,7 @@ export class SpecBuilder {
     }
     if (query.filters) {
       spec.filters = query.filters.map((filter: Filter) => {
-        if (filter instanceof RelationFilter) {
```
mikelehen: Can you update the comment on line 1504 (`This currently only supports relation filters (<, <=, ==, >=, >)`)?  Maybe the comment can be completely removed (I think we accidentally support nan / null filters now...)

packages/firestore/src/core/query.ts
```diff
@@ -110,7 +110,7 @@ export class Query {
   addFilter(filter: Filter): Query {
     assert(
       this.getInequalityFilterField() == null ||
-        !(filter instanceof RelationFilter) ||
+        !(filter instanceof FieldFilter) ||
         !filter.isInequality() ||
```
mikelehen: In your interest of avoiding instanceof and over-constraining, we could promote isInequality() to `Filter` if we wanted. 🤷‍♂

packages/firestore/src/core/query.ts
```diff
@@ -447,57 +447,79 @@ export abstract class Filter {
...
+    } else if (op === Operator.ARRAY_CONTAINS) {
+      return new ArrayContainsFilter(field, value);
+    } else if (op === Operator.IN) {
+      assert(
+        value instanceof ArrayValue,
+        'IN filter has invalid value: ' + value.toString()
+      );
+      return new InFilter(field, value as ArrayValue);
+    } else if (op === Operator.ARRAY_CONTAINS_ANY) {
+      return new ArrayContainsAnyFilter(field, value as ArrayValue);
```
mikelehen: We should presumably have the same `value instanceof ArrayValue` assert here (as above).

packages/firestore/src/core/query.ts
```diff
@@ -635,17 +610,14 @@ export class RelationFilter extends Filter {
   }
 }
 
+export class UnaryFilter extends FieldFilter {}
```
mikelehen: I don't care strongly, but this empty class looks a little silly.  I guess it only exists for a single instanceof check in the serializer.  I say we drop it or else add some kind of comment justifying its existence ("for parity with the proto representation" or something)

packages/firestore/src/core/query.ts
```diff
@@ -635,17 +610,14 @@ export class RelationFilter extends Filter {
...
  */
-export class NullFilter extends Filter {
-  constructor(public field: FieldPath) {
-    super();
-  }
-
-  matches(doc: Document): boolean {
-    const val = doc.field(this.field);
-    return val !== undefined && val.value() === null;
+export class NullFilter extends UnaryFilter {
```
mikelehen: Concretely, why are we keeping NullFilter / NaNFilter rather than just relying on FieldFilter?  Removing them would be a nice simplification (we could probably get this PR to be net negative lines of code) and I'd rather forget a few resume tokens (in practice, NaN queries are probably nonexistent and null queries are probably rare-ish) than carry along these classes forever...

If we keep them, we should have some sort of comment explaining that they exist due to legacy canonicalIds  (unless there's another reason I'm missing)?

packages/firestore/src/core/query.ts
```diff
@@ -686,13 +644,86 @@ export class NanFilter extends Filter {
...
+
+/**
+ * A Filter that implements the array-contains operator. When a user specifies
+ * query.where('some.field', 'array-contains', 42), this filter will match
+ * documents that:
+ *
+ *   * contain a field at some.field;
+ *   * that field is an array; and
+ *   * that array contains the value 42.
+ */
```
mikelehen: (you are 110% welcome to ignore, but...)  TBH these comments feel a little overkill and out-of-place here (they would be better suited for inclusion in our public API docs or something).  I'm also not wild about `*` bullet lists inside `/*` comments.  I also want to get this PR to be net negative lines of code (and these descriptions are longer than the actual implementation). 😛

packages/firestore/src/model/field_value.ts
```diff
@@ -641,6 +641,18 @@ export class ArrayValue extends FieldValue {
     return this.internalValue.map(v => v.value(options));
   }
 
+  /**
+   * Returns true if the given value is contained in this array.
+   */
+  contains(value: FieldValue): boolean {
```
mikelehen: Oh, nice idea!

mikelehen:
BOOM!
image

packages/firestore/src/core/query.ts
mikelehen: Yeah.  I don't entirely disagree but when the cast happens so far from the validation, it's hard to know it's safe... so I don't really have a better answer... Might end up cleaner if we moved some of the create() logic into the public API or something.

Issue #1774, opened 2019-05-09
Crash if cookies are disabled


neelance:
The line
https://github.com/firebase/firebase-js-sdk/blob/6064f2744bc01b00fa771ffd08e5309c89110f84/packages/performance/src/services/api_service.ts#L60
crashes with `SecurityError: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.` if cookies are disabled (Chrome 74.0.3729.131).

Feiyang1:
Thanks for reporting it! Very interesting that disabling cookie disables localStorage access.
I can see a couple of options:
1.  wrap the code with try catch and falls back to memory 
2. use indexeddb instead

I will take it back to the team and see what we will do.

mmermerkaya:
At least in Chrome, if cookies are disabled IndexedDB doesn't work either. Disabling cookies blocks all client storage features AFAIK.

PR #1895, opened 2019-06-19
Do not initialize performance if required apis are not available

alikn:
Context: https://github.com/firebase/firebase-js-sdk/issues/1774

### DIFF
--- a/packages/performance/src/controllers/perf.test.ts
+++ b/packages/performance/src/controllers/perf.test.ts
@@ -22,7 +22,7 @@ import { Trace } from '../resources/trace';
 import { Api, setupApi } from '../services/api_service';
 import { FirebaseApp } from '@firebase/app-types';
 import * as initializationService from '../services/initialization_service';
-
+import { consoleLogger } from '../utils/console_logger';
 import '../../test/setup';
 
 describe('Firebase Performance Test', () => {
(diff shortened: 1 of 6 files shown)


packages/performance/src/services/perf_logger.ts
```diff
@@ -107,6 +107,10 @@ export function logTrace(trace: Trace): void {
   if (!settingsService.dataCollectionEnabled && !trace.isAuto) {
     return;
   }
+  // Do not log if required apis are not available.
+  if (!Api.getInstance().requiredApisAvailable()) {
```
zijianjoy: Should we also add these "if condition" to logNetworkRequest() in this file?

packages/performance/src/controllers/perf.test.ts
```diff
@@ -38,6 +42,16 @@ describe('Firebase Performance Test', () => {
     options: fakeFirebaseConfig
   } as unknown) as FirebaseApp;
 
+  describe('#constructor', () => {
+    it('does not initialize performance if the required apis are not available', () => {
+      stub(Api.prototype, 'requiredApisAvailable').returns(false);
+      stub(initializationService, 'getInitializationPromise');
+      new PerformanceController(fakeFirebaseApp);
+
+      expect(initializationService.getInitializationPromise).not.be.called;
```
zijianjoy: Add a check on the existence of console log?

For example:
expect( console.log.calledWith('Firebase Performance cannot start if browser does not support fetch and Promise or cookie is disabled.') ).to.be.true;

packages/performance/src/services/perf_logger.ts
alikn: No need, the trace api can be used to create custom trace even if the sdk is not initialized, but there is no public api for networks.

packages/performance/src/controllers/perf.test.ts
alikn: Added a check for the consoleLogger to be called. Not sure about adding the whole message though.

packages/performance/src/services/perf_logger.ts
zijianjoy: Got it, suggest to use console logger to log that required APIs are not available for the scenario above.

PR #1897, opened 2019-06-20
Enable Typescript strict flag for app, logger, and util packages

hsubox76:
Turned on Typescript `strict` flag for app, logger, and util packages.  This encompasses a number of more specific flags.

### DIFF
--- a/packages/app/src/firebaseNamespaceCore.ts
+++ b/packages/app/src/firebaseNamespaceCore.ts
@@ -225,10 +225,7 @@ export function createFirebaseNamespaceCore(
     // @ts-ignore
     firebaseAppImpl.prototype[name] = function(...args: any) {
       const serviceFxn = this._getService.bind(this, name);
-      return serviceFxn.apply(
-        this,
-        allowMultipleInstances ? args : []
-      );
+      return serviceFxn.apply(this, allowMultipleInstances ? args : []);
     };
 
     return serviceNamespace;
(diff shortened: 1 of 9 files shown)


packages/app/src/firebaseNamespaceCore.ts
```diff
@@ -222,11 +222,15 @@ export function createFirebaseNamespaceCore(
     namespace[name] = serviceNamespace;
 
     // Patch the FirebaseAppImpl prototype
+    // TODO: The eslint disable can be removed and the 'ignoreRestArgs' option added to
```
Feiyang1: nit: move it to directly above the eslint suppression

packages/util/src/obj.ts
```diff
@@ -108,7 +108,7 @@ export const map = function<V>(
 ) {
   var res: UtilObject<V> = {};
   for (var key in obj) {
-    res[key] = fn.call(context, obj[key], key, obj);
+    res[key] = fn.call(context, obj[key], key, obj) as V;
```
Feiyang1: Change fn's return signature to V ?

packages/util/src/subscribe.ts
```diff
@@ -78,7 +78,7 @@ class ObserverProxy<T> implements Observer<T> {
   // Micro-task scheduling by calling task.then().
   private task = Promise.resolve();
   private finalized = false;
-  private finalError: Error;
+  private finalError?: Error;
```
Feiyang1: What does "?" mean here? It could be undefined?

packages/app/src/firebaseNamespaceCore.ts

packages/util/src/subscribe.ts
hsubox76: Yeah, either that or it should be `Error | null` and initialized to null, because it's not initialized and is only assigned if there is an error.  There is a falsy check later when it is used.  What do you think?

packages/util/src/obj.ts
hsubox76: Done.

Issue #560, opened 2018-03-13
FR: Enable firestore web offline persistence by default when running as standalone


zevdg:
### [REQUIRED] Describe your environment

  * Operating System version: linux 4.15
  * Firebase SDK version: 4.11.0
  * Firebase Product: firestore

### [REQUIRED] Describe the problem
Until #190 is addressed, it doesn't make sense to enable this feature across the board.  However, this feature is more or less required to make a true PWA.  While we wait for multi-tab support, there may be a decent stopgap to support PWAs better without breaking normal websites.  

When an app is running in standalone mode, it may be (relatively) safe to assume that it will be the only version of that app running in that way.  For example, when an android user selects "Add to Homescreen" in chrome, the OS seems to enforce a single instance of the standalone version of the app.  

In my own PWA, i have the following code

```javascript
function isRunningStandalone() {
  return (
    // Chrome
    window.matchMedia("(display-mode: standalone)").matches ||
    window.navigator.standalone // iOS
  );
}

if (isRunningStandalone()) {
    try {
      await firebase.firestore().enablePersistence();
    } catch (err) {
      console.error("Error enabling firestore offline persistence", err);
    }
  }
```
This gives my PWA offline functionality safely on android.

That said, I'm unsure of how other browsers and/or OSs treat PWAs.  For example, the "Add to shelf" equivalent for Desktop Chrome does appear to allow the user to open multiple instances of an app in standby mode.  ChromeOS or iOS may also allow multiple instances.

#### Steps to reproduce:

N/A
#### Relevant Code:

N/A

zevdg:
closely related to #262

Issue #519, opened 2018-02-17
FR: Easier access to google calendar


mesqueeb:
Currently firebase auth is not enough for easy access to a user's google calendar even though he's logged in via the google auth provider.

It would be great if Firebase could natively add easier support and documentation to integrate the google calendar scopes through the firebase authentication.

bojeil-google:
Hey @mesqueeb, Firebase Auth only focuses on authentication and not authorization.
We only return the OAuth access token based on the provided OAuth scopes you requested which can include Google Calendar access. If you wish to use Google Calendar API, you should use their API directly. If you need continuous access to Google Calendar, you are better off using the Google Sign-in library for web: https://developers.google.com/identity/sign-in/web/sign-in

You can still use that to sign in with Firebase Auth using the returned Google ID token or Google Access token.

Can you elaborate more on your request? Do you need a mechanism to manage your OAuth access tokens via Firebase Auth? or are you looking into built-in support for the Calendar API via Firebase Auth? The latter is out of scope as this is an authentication library.

mesqueeb:
Dear @bojeil-google Thank you for your elaborate answers always. 

I have a firebase app where users have tasks. I want a way to synchronize the tasks to show to their google calendar and view the calendar events from their existing calendar. 

I'm now researching for the past several ways the best way to take this approach.

bojeil-google:
Hey @mesqueeb, it is unlikely we will have any dedicated APIs for Calendar, Gmail, Drive, Youtube, etc. The list is extremely long: https://developers.google.com/identity/protocols/googlescopes
In addition, Firebase Auth is not only meant to facilitate sign-in with Google. There are many other OAuth providers that we support and plan to support in the future.
It is clearly out of scope for what Firebase Auth is designed for. 

The most reasonable option would be the capability to manage OAuth access tokens which is currently not the focus of Firebase Authentication and not supported, but is in the realm of possibility. This would facilitate access to such APIs (not just Google but other OAuth providers). Currently, Firebase Auth only returns the initial OAuth access token on sign-in and discards the returned OAuth refresh token. You are also welcome to file an official feature request for that if you haven't already done so.

mesqueeb:
@bojeil-google Currently every time a user who's logged into firebase can see their Firestore data, but then need to log in a second time through the _Javascript GAPI_ to get the access to their calendar via GAPI.

[...]

PR #1912, opened 2019-06-24
Backport updates from Android (multiple changes)

thebrianchen:
Backporting changes from Android IN queries: 
- [#519](https://github.com/firebase/firebase-android-sdk/pull/519#discussion_r294032999): filter for `null` and `NaN` and reorganize tests to match Android SDK
- [#550](https://github.com/firebase/firebase-android-sdk/pull/550): update doc comments for `clearPersistence()` to include Rachel's edits
- [#560](https://github.com/firebase/firebase-android-sdk/pull/560): Allow IN queries with field path and array of documentIds

This change introduces `KeyFieldInFilter`, which stacks on top of @wilhuff's refactor (#1894). This will be ported to Android when he refactors the Android client.

TODO: After this change goes in, I will update the error messages inside `parseDocumentIdValue()` to match the Android SDK, since those error messages are more generally applicable to both IN queries and KeyFieldFilter queries.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -518,7 +518,7 @@ export class FieldFilter extends Filter {
           }),
           'Comparing on key with IN, but an array value was not a RefValue'
         );
-        return new KeyFieldInFilter(field, op, value as ArrayValue);
+        return new KeyFieldInFilter(field, value as ArrayValue);
       } else {
         assert(
           value instanceof RefValue,
@@ -645,18 +645,14 @@ export class KeyFieldFilter extends FieldFilter {
 
 /** Filter that matches on key fields within an array. */
 export class KeyFieldInFilter extends FieldFilter {
-  constructor(field: FieldPath, op: Operator, value: ArrayValue) {
-    super(field, op, value);
+  constructor(field: FieldPath, public value: ArrayValue) {
+    super(field, Operator.IN, value);
   }
 
   matches(doc: Document): boolean {
-    const arrayValue = this.value as ArrayValue;
+    const arrayValue = this.value;
     return arrayValue.internalValue.some(refValue => {
-      const comparison = DocumentKey.comparator(
-        doc.key,
-        (refValue as RefValue).key
-      );
-      return this.matchesComparison(comparison);
+      return doc.key.isEqual((refValue as RefValue).key);
     });
   }
 }
@@ -675,12 +671,12 @@ export class ArrayContainsFilter extends FieldFilter {
 
 /** A Filter that implements the IN operator. */
 export class InFilter extends FieldFilter {
-  constructor(field: FieldPath, value: ArrayValue) {
+  constructor(field: FieldPath, public value: ArrayValue) {
     super(field, Operator.IN, value);
   }
 
   matches(doc: Document): boolean {
-    const arrayValue = this.value as ArrayValue;
+    const arrayValue = this.value;
     const other = doc.field(this.field);
     return other !== undefined && arrayValue.contains(other);
   }
@@ -688,12 +684,12 @@ export class InFilter extends FieldFilter {
 
 /** A Filter that implements the array-contains-any operator. */
 export class ArrayContainsAnyFilter extends FieldFilter {
-  constructor(field: FieldPath, value: ArrayValue) {
+  constructor(field: FieldPath, public value: ArrayValue) {
     super(field, Operator.ARRAY_CONTAINS_ANY, value);
   }
 
   matches(doc: Document): boolean {
-    const arrayValue = this.value as ArrayValue;
+    const arrayValue = this.value;
     const other = doc.field(this.field);
     return (
       other instanceof ArrayValue &&
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -449,17 +449,29 @@ export abstract class Filter {
    */
   static create(field: FieldPath, op: Operator, value: FieldValue): Filter {
     if (field.isKeyField()) {
-      assert(
-        value instanceof RefValue,
-        'Comparing on key, but filter value not a RefValue'
-      );
-      assert(
-        op !== Operator.ARRAY_CONTAINS &&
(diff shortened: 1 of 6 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -631,6 +643,24 @@ export class KeyFieldFilter extends FieldFilter {
   }
 }
 
+/** Filter that matches on key fields within an array. */
+export class KeyFieldInFilter extends FieldFilter {
```
mikelehen: On the surface it's bugging me a little that we have `KeyFieldFilter`, `InFilter` and `KeyFieldInFilter` as peers... though thinking about this, I think it probably makes sense.  cc/ @wilhuff to sanity-check since he just put thought into our filter hierarchy...

packages/firestore/src/core/query.ts
```diff
@@ -631,6 +643,24 @@ export class KeyFieldFilter extends FieldFilter {
...
+  constructor(field: FieldPath, op: Operator, value: ArrayValue) {
+    super(field, op, value);
+  }
+
+  matches(doc: Document): boolean {
+    const arrayValue = this.value as ArrayValue;
```
mikelehen: We can avoid this `as` cast by changing `value: ArrayValue` in the constructor definition to `public value: ArrayValue` (i.e. we override the public `value` property inherited from `FieldFilter` to be a more specific type (`ArrayValue` instead of `FieldValue`)).

But I see we are already doing this cast in `KeyFieldFilter`, `InFilter`, and `ArrayContainsFilter`.  Can you fix them too?

packages/firestore/src/core/query.ts
```diff
@@ -631,6 +643,24 @@ export class KeyFieldFilter extends FieldFilter {
...
+    return arrayValue.internalValue.some(refValue => {
+      const comparison = DocumentKey.comparator(
+        doc.key,
+        (refValue as RefValue).key
+      );
+      return this.matchesComparison(comparison);
```
mikelehen: I am having trouble seeing how this works...  matchesComparison() is going to use `this.op` to decide what comparison to match for (`<`, `==`, etc.).  But `this.op` is going to be `in` which `matchesComparison()` doesn't support.  So I'm overlooking something or this code isn't being hit by our tests...?

In any case, I _think_ this block could be replaced with something like:
```
return doc.key.isEqual((refValue as RefValue).key);
```

Since all we only need to do an equality check.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1003,125 +1003,99 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
-          "Invalid query. You cannot use 'array-contains-any' filters with " +
-            "'in' filters."
-        );
+      // This is redundant with the above tests, but makes sure our validation
+      // doesn't get confused.
+      expect(() =>
+        db
+          .collection('test')
+          .where('foo', inOp, [2, 3])
+          .where('foo', 'array-contains', 1)
```
mikelehen: Hrm...  shouldn't this fail to compile since you're not using `arrayContainsAnyOp`?

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1163,7 +1137,7 @@ apiDescribe('Validation:', (persistence: boolean) => {
 
     validationIt(
       persistence,
-      'cannot have an IN or array-contains-any filter with non-array values.',
+      'follow array requirements for disjunctive filters',
```
mikelehen: maybe "follow" => "enforce" ?

packages/firestore/test/integration/api/validation.test.ts
thebrianchen: `array-contains` has been released already, so this test is checking that `in` with `array-contains-any` still fails even if its separated by an `array-contains`.

packages/firestore/src/core/query.ts
wilhuff: Abstractly, `KeyFieldInFilter` seems like it is-a `KeyFieldFilter` and `KeyFieldInFilter` seems like it is-a `InFilter`, but it's also neither of them because it operates on different arguments than either.

That is to say, none of these are practically substitutable for each other and subclasses that violate LSP seem like a bad idea.

packages/firestore/src/core/query.ts
thebrianchen: Thanks for catching another breaking bug. I had validation tests to catch the errors, but not an integration test to check that the end to end query worked. I've added another integration test to check that IN queries actually work with document IDs.

packages/firestore/src/core/query.ts
mikelehen: 👍 Thanks for the sanity check.

packages/firestore/src/core/query.ts
```diff
@@ -675,25 +671,25 @@ export class ArrayContainsFilter extends FieldFilter {
...
 /** A Filter that implements the array-contains-any operator. */
 export class ArrayContainsAnyFilter extends FieldFilter {
-  constructor(field: FieldPath, value: ArrayValue) {
+  constructor(field: FieldPath, public value: ArrayValue) {
     super(field, Operator.ARRAY_CONTAINS_ANY, value);
   }
 
   matches(doc: Document): boolean {
-    const arrayValue = this.value as ArrayValue;
+    const arrayValue = this.value;
```
mikelehen: nit: I would just drop the intermediate variables (`arrayValue`) at this point. It is saving 0 characters!

packages/firestore/test/integration/api/validation.test.ts
mikelehen: Oh, right! Brain hiccup. :)

thebrianchen:
@mikelehen I just realized that the current implementation of `matches` in the Android SDK doesn't support the integration test for IN queries on doc IDs yet... 

This means that Gil's refactor of filter PR needs to go in first, or else I'll be doing the logic twice. I'll submit this PR for now, and back-back-backport(?) the integration test once the Android refactor of Filter is done to include `KeyFieldInFilter`.

mikelehen:
👍

PR #1913, opened 2019-06-24
Fix array-contains queries

### LINKED ISSUE #1904: Firestore 1.4.2 in Firebase 6.2.2: array-contains query returns empty array with offline persistence

  * Operating System version: Win10
  * Browser version: Chrome 75.0.3770.100
  * Firebase SDK version: 6.2.2
  * Firebase Product: firestore

#### Steps to reproduce:

Make a query with array-contains, no error from console but empty array returned.

#### Relevant Code:

```javascript

db
        .collection('Foo')
        .where('barArray', 'array-contains', 'something')
        


```

var-const:
#1894 introduced a regression that made persisted queries deserialize incorrectly. Fix the regression and add tests.

The root cause is that the refactoring introduced a type hierarchy of classes derived from `FieldFilter`, but serializer always deserializes persisted filters to the base class `FieldFilter`, never to the derived types. This would have affected array-contains queries, in queries, and key field queries.

Fixes #1904.


packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -25,6 +25,9 @@ import { GeoPoint } from '../../../../src/api/geo_point';
 import { Timestamp } from '../../../../src/api/timestamp';
```
var-const: I couldn't figure out how to test `KeyFieldFilter` due to problems with creating a query with this filter.

packages/firestore/src/remote/serializer.ts
```diff
@@ -1282,7 +1282,7 @@ export class JsonProtoSerializer {
   }
 
   fromFieldFilter(filter: api.Filter): Filter {
-    return new FieldFilter(
+    return Filter.create(
```
mikelehen: There are two other instances of `new FieldFilter()` in this file that I think should optimally be `FieldFilter.create()` as well (though in practice they're not harmful).

To make sure we don't mess this up going forward, I think it might make sense to move `Filter.create()` to be `FieldFilter.create()` so we can make the `FieldFilter` constructor `protected`.  This is probably more correct anyway, since the only kind of filter that Filter.create() could conceivably create with its current arguments is `FieldFilter`.

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -693,7 +696,9 @@ describe('Serializer', () => {
           value: { stringValue: 'food' }
         }
       });
-      expect(s.fromFieldFilter(actual)).to.deep.equal(input);
+      const roundtripped = s.fromFieldFilter(actual);
+      expect(roundtripped).to.deep.equal(input);
+      expect(roundtripped).to.be.instanceof(FieldFilter);
```
mikelehen: FWIW we could instead do:
```
expect(Object.getPrototypeOf(roundtripped)).to.equal(Object.getPrototypeOf(input));
```
to avoid the redundancy of specifying the type of `input`...  but in practice it doesn't matter and it's easier to read as-is... so probably not a good improvement (unless we wanted to refactor these tests to share code or something).

packages/firestore/test/unit/remote/node/serializer.test.ts
mikelehen: Not sure if this is what you mean, but it might look something like:

```
      const input = filter(DOCUMENT_KEY_NAME, '==', ref('project/database', 'coll/doc'));
```

var-const: Thanks! That worked.

packages/firestore/src/remote/serializer.ts
var-const: I agree, and I like this suggestion. Done, PTAL.

packages/firestore/test/unit/remote/node/serializer.test.ts
var-const: I don't know how `getPrototypeOf` works -- if it checks for the _exact_ class, then it could be an improvement, because given, say, `ArrayContainsFilter`, it will nevertheless satisfy a check for `instanceof FieldFilter` (due to inheritance).

commits.txt
```diff
@@ -0,0 +1,26 @@
...
+c55d1e9e269db59a74ea745e8744e8c33d57c007 Check that self.firebase is not undefined (#1868)
+fb45cd65e27ec858ca7170d5d8a9f47f49a70dcd Upgrade faye-websocket from 0.11.1 to 0.11.3 (#1880)
+5511cbd6c2659868bae524209a833b6894ddc8c9 add code fonts (#1871)
+b91f690d83aba450697e5c73ea5523fcf1d58b83 Fix storage Reference.put param type (#1876)
+9678bf9a5692b0a33822cfdef27ddb27834cebb2 Add IN and ARRAY_CONTAINS_ANY (not publicly exposed) (#1843)
+dd098c6a87f23ddf54a7f9b21b87f7bb3fd56bdd Migrate messaging to eslint (#1872)
+fd5a31a6bba01b79f9fd269c96926c8aa6248701 Generate a new token after getId calls if there is no valid token (#1869)
+99e6cc8267d4eea4b2893a45d097c6cb20eddd15 Add support for running Firestore integration tests against the emulator (#1851)
+3f2c7bf8a08978bea065ac1706af016e8646f234 Check that window is complete in BrowserConnectivityMonitor (#1870)
+73f10767a8473468aa6fb6300abe93f998b59ecf Minor style changes for a RTDB comment block (#1857)
```
mikelehen: 🤔 

Think you accidentally added some extra files...

var-const: I think it's the prettier script striking again. Removed now.

packages/firestore/test/unit/remote/node/serializer.test.ts
mikelehen: Ah, yes.  I believe the prototype check would be more strict than that...  that said it's not super normal to deal with an object's prototype directly or do equal checks on it.  I am not 100% confident it wouldn't bite us in the future some day (due to browser differences or something).  So I don't feel strongly that it's a better approach or anything (i.e. feel free to leave as-is).

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -745,7 +757,31 @@ describe('Serializer', () => {
           value: { doubleValue: 1e100 }
         }
       });
-      expect(s.fromFieldFilter(actual)).to.deep.equal(input);
+      const roundtripped = s.fromFieldFilter(actual);
+      expect(roundtripped).to.deep.equal(input);
+      expect(roundtripped).to.be.instanceof(FieldFilter);
+    });
+
+    it('converts key field', () => {
```
mikelehen: Thanks for adding!

PR #1915, opened 2019-06-25
Set typescript strict option to true for functions and testing

hsubox76:
Turn on "strict: true" option in Typescript compiler for functions, testing, and main firebase package.

### DIFF
--- a/packages/functions/src/api/service.ts
+++ b/packages/functions/src/api/service.ts
@@ -71,7 +71,7 @@ export class Service implements FirebaseFunctions, FirebaseService {
   private readonly serializer = new Serializer();
   private emulatorOrigin: string | null = null;
   private cancelAllRequests: Promise<void>;
-  private deleteService: Function;
+  private deleteService: Function = () => {};
 
   /**
    * Creates a new Functions service for the given app and (optional) region.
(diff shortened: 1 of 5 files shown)


packages/functions/src/api/service.ts
```diff
@@ -71,7 +71,7 @@ export class Service implements FirebaseFunctions, FirebaseService {
   private readonly serializer = new Serializer();
   private emulatorOrigin: string | null = null;
   private cancelAllRequests: Promise<void>;
-  private deleteService: Function;
+  private deleteService: Function = () => {};
```
Feiyang1: `private deleteService!: Function` - the `!` means it will be initialized before any usage.

packages/testing/test/database.test.ts
```diff
@@ -87,13 +87,16 @@ describe('Testing Module Tests', function() {
   });
 
   it('loadDatabaseRules() throws if no databaseName or rules', async function() {
+    //@ts-ignore This test is checking an invalid param.
     await expect(firebase.loadDatabaseRules.bind(null, {})).to.throw(
```
Feiyang1: `(firebase as any).loadDatabaseRules` and use eslint suppression instead.

packages/functions/src/api/service.ts
hsubox76: Got it, done.

packages/testing/test/database.test.ts
hsubox76: Done for 3 lines.

Issue #1696, opened 2019-04-13
TypeError: app.auth is not a function


ramen-mukherjee:
#### Introduction
After the recent updates (I update regularly), I am getting following error while running the application. Project is created with Angular CLI. As of now, no build errors.

#### System
  * Operating System version: Windows 10 (v1803 - OS Build 17134.753) & 8 x64
  * Browser version: Chrome v74.0.3729.131 x64
  * Firebase SDK version: v6.0.2
  * Firebase Product: auth
  * Node: v12.2.0
  * NPM: v6.9.0
  * Angular: v7.2.15
  * Angular CLI: v7.3.9
  * @angular/fire: v5.1.3

#### Steps to reproduce:
1. Update your existing App.
2. Either use JIT or AOT build of Angular.
3. Observe the below error in browser console at runtime.

#### Screenshot
![app variable](https://user-images.githubusercontent.com/20390998/57704628-ea93f500-767f-11e9-84d8-51e566172530.png)

#### Code
`import { User } from './../models/user.model';
import { Observable, Observer } from 'rxjs';
import { Injectable } from '@angular/core';
import { AngularFireAuth } from '@angular/fire/auth';
import { auth } from 'firebase/app';

@Injectable({
  providedIn: 'root'
})
export class AuthService {
  user: User = {
    name: '',
    email: '',
    photo: '',
    token: {
      value: '',
      uid: '',
      createdOn: '',
      expiredOn: '',
      isAuthenticated: false,
      isAdmin: false
    }
  };

  constructor(public afAuth: AngularFireAuth) { }

  observable$ = Observable.create((observer) => {
    try {
      this.getUserDetails();
      observer.next(this.user);
    } catch {
      this.resetUserData();
      observer.error();
    } finally {
      observer.complete();
    }
  });

  getAuthUser(): Observable<User> {
    return this.observable$;
  }

  login(): User | void {
    this.afAuth.auth.setPersistence(auth.Auth.Persistence.SESSION);
    this.afAuth.auth.signInWithPopup(new auth.GoogleAuthProvider())
    .then((user) => {
      if (user) {
        this.getUserDetails();
      }
    })
    .catch((error) => {
      this.resetUserData();
      console.error(error);
    });
    return this.user;
  }

  logout(): void {
    this.afAuth.auth.signOut()
    .then(() => {
      this.resetUserData();
    })
    .catch(
      response => console.error('Error Occured While Logging Out' + response)
    );
  }

[...]

PR #1916, opened 2019-06-25
Address version conflict when used together with firebase-admin

Feiyang1:
This PR will allow firebase-admin to remove `@firebase/app` from its dependency list. `@firebase/app` being a dependency in `firebase-admin` causes version conflict when `firebase-admin` and `firebase` are used together. More detail: https://github.com/firebase/firebase-js-sdk/issues/1696#issuecomment-501546596

The PR includes the following changes
1. Introduce a new file `version.ts` in `@firebase/database`, so the source code doesn't import `@firebase/app` directly for SDK_VERSION.
2. set `SDK_VERSION` in `version.ts` in the entry files (index.ts and index.node.ts), so we can set `SDK_VERSION` differently depending on the target.
3. change the static `import firebase from '@firebase/app'` to `require('@firebase/app').default` in `index.node.ts` for handling `firebase-admin` use case. It allows `@firebase/database` to work in `firebase-admin` without `@firebase/app`

The corresponding PR in `firebase-admin`: https://github.com/firebase/firebase-admin-node/pull/586


packages/database/index.node.ts
```diff
@@ -88,13 +85,20 @@ export function registerDatabase(instance: FirebaseNamespace) {
...
 
-registerDatabase(firebase);
+try {
+  // If @firebase/app is not present, skip registering database.
+  // It could happen when this package is used in firebase-admin which doesn't depend on @firebase/app.
+  // Previously firebase-admin depends on @firebase/app, which causes version conflict on
+  // @firebase/app when used together with the js sdk. More detail:
+  // https://github.com/firebase/firebase-js-sdk/issues/1696#issuecomment-501546596
+  const firebase = require('@firebase/app').default;
+  registerDatabase(firebase);
```
hsubox76: Could this swallow a legit error in `registerDatabase()`?  If the require works fine but there is some error thrown in `registerDatabase()` itself?

packages/database/src/core/version.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** The semver (www.semver.org) version of the SDK. */
+export let SDK_VERSION = '';
```
schmidt-sebastian: I believe this export in combination with `setSDKVersion`  doesn't work, as it exports the current state of the variable:

deps.js:

```
let foo = 'foo'
function setToBar() {
	foo = 'bar';
}
module.exports = {foo, setToBar};
```

```
let deps = require('./deps');
console.log(deps.foo);
deps.setToBar();
console.log(deps.foo);
```

Prints:

```
foo
foo
```

Feiyang1: It actually works because esm only exports references. The code compiled to cjs format would look like:
```javascript
exports.foo = 'foo';
function setToBar() {
    exports.foo = 'bar';
}
exports.setToBar = setToBar;
```

packages/database/index.node.ts
Feiyang1: Yeah, updated to only ignore MODULE_NOT_FOUND error

Feiyang1:
@schmidt-sebastian gentle ping :)

PR #1920, opened 2019-06-26
Fix SauceLabs tests

mmermerkaya:


Feiyang1:
1. Installations tests are working correctly in saucelabs.
2. performance is failing in Firefox and Edge because localStorage can't be directly stubbed in them. https://github.com/sinonjs/sinon/issues/662
The workaround is to have a wrapper around the native APIs and stub the wrapper. @alikn - I can do it if you feel okay.

performance also doesn't work in IE because of a syntax error I don't know what is causing it yet.

mmermerkaya:
Thanks for figuring all this out! Sounds like there's nothing for me to do. :smile:

packages/installations/src/helpers/buffer-to-base64-url-safe.test.ts
```diff
@@ -19,8 +19,13 @@ import { expect } from 'chai';
 import '../testing/setup';
 import { bufferToBase64UrlSafe } from './buffer-to-base64-url-safe';
 
-const TYPED_ARRAY_REPRESENTATION = new TextEncoder().encode('hello world');
-const BASE_64_REPRESENTATION = btoa('hello world');
+const str = 'hello world';
+const TYPED_ARRAY_REPRESENTATION = new Uint8Array(str.length);
```
Feiyang1: Changed so it works in IE11 without TextEncoder polyfill, though there still are many tests failing in IE11 (due to indexedDB?). Is installations designed to work in IE11 ?

Feiyang1: @mmermerkaya ping

mmermerkaya: Installations should work in IE11. I have not received any bug reports about it not working in IE11.

Feiyang1:
The PR fixes the saucelabs testing infrastructure and unit tests for some SDKs.
Performance and Installations are still failing for IE11. We will fix them in the separate PRs.

PR #1921, opened 2019-06-26
Multi tenant beta

wti806:
beta release for multi-tenant support


packages/auth-types/index.d.ts
```diff
@@ -325,6 +353,7 @@ export class FirebaseAuth {
   signInWithPopup(provider: AuthProvider): Promise<UserCredential>;
   signInWithRedirect(provider: AuthProvider): Promise<void>;
   signOut(): Promise<void>;
+  tenantId: string | null;
```
bojeil-google: Should we mark readonly properties? It is not clear when tenantId can be set and when it can only be read.

packages/firebase/index.d.ts
```diff
@@ -1496,6 +1516,46 @@ declare namespace firebase.functions {
...
+     * Parse the email action link string and returns an ActionCodeURL object
+     * if the link is valid, otherwise returns null.
+     *
+     * @param link The email action link string.
+     * @return The ActionCodeURL object, or null if the link is invalid.
+     */
+    static parseLink(link: string): firebase.auth.ActionCodeURL | null;
+    /**
+     * The tenant ID of the email action link. Null if the email action
+     * is from agent project.
```
bojeil-google: We don't use "agent" term externally.
Instead change to:
is from the parent project.

packages/firebase/index.d.ts
```diff
@@ -2610,6 +2670,29 @@ declare namespace firebase.auth {
      * Signs out the current user.
      */
     signOut(): Promise<void>;
+    /**
+     * The current Auth instance's tenant ID. This is a readable/writable
+     * property. When set the enant ID on a Auth instance, all future
+     * sign-in/sign-up operations will pass this tenant ID and sign in
+     * or sign up users to the specified project as identified by the tenant.
+     * When set to null, users are signed in to the agent project. By default,
```
bojeil-google: replace "agent project" with "parent project"

packages/firebase/index.d.ts
```diff
@@ -707,6 +707,26 @@ declare namespace firebase {
     sendEmailVerification(
       actionCodeSettings?: firebase.auth.ActionCodeSettings | null
     ): Promise<void>;
+    /**
+     * The current user's tenant ID. This is a read-only property, which indicates
+     * the tenant ID used to sign in the current user. This is null if the user is
+     * signed in from the agent project.
```
bojeil-google: Replace "agent project" with "parent project"

packages/auth-types/index.d.ts
```diff
@@ -353,7 +353,7 @@ export class FirebaseAuth {
   signInWithPopup(provider: AuthProvider): Promise<UserCredential>;
   signInWithRedirect(provider: AuthProvider): Promise<void>;
   signOut(): Promise<void>;
-  tenantId: string | null;
+  readonly tenantId: string | null;
```
bojeil-google: this is not readonly. The user one is.

wti806: done

packages/firebase/index.d.ts
wti806: done

packages/firebase/index.d.ts
wti806: done

packages/firebase/index.d.ts
wti806: done

packages/auth-types/index.d.ts
wti806: done

packages/firebase/index.d.ts
```diff
@@ -3577,6 +3732,31 @@ declare namespace firebase.auth.Auth {
...
 }
 
+declare namespace firebase.auth.ActionCodeInfo {
+  type Operation = string;
+  /**
+   * An enumeration of the possible email action types.
+   */
+  var Operation: {
+    /**
+     * The email link sign in email action.
```
kevinthecheung: The email link sign-in action.

packages/firebase/index.d.ts
```diff
@@ -2610,6 +2670,29 @@ declare namespace firebase.auth {
      * Signs out the current user.
      */
     signOut(): Promise<void>;
+    /**
+     * The current Auth instance's tenant ID. This is a readable/writable
+     * property. When set the enant ID on a Auth instance, all future
```
kevinthecheung: When you set the tenant ID of an Auth instance, all future sign-in/sign-up operations will pass this tenant ID and sign in or sign up users to the specified tenant project.

PR #1922, opened 2019-06-26
cherry pick es2015+ libs

Feiyang1:
It helps us track what es2015+ features are used in firebase.


Feiyang1:
Removed `lib` option from installations

PR #1925, opened 2019-06-28
Use the FID from the server if client FID generation fails

mmermerkaya:
Starting from SDK v0.2.0, the server will start returning a valid FID instead of returning an error if the client tries to register an invalid FID, and the SDK will save that FID to the DB, replacing the invalid one. In case of a failure with client side FID generation, `getId` will wait for a valid FID from the server before resolving instead of throwing or returning an invalid FID.

### DIFF
--- a/packages/installations/src/helpers/generate-fid.ts
+++ b/packages/installations/src/helpers/generate-fid.ts
@@ -28,6 +28,8 @@ export function generateFid(): string {
     // A valid FID has exactly 22 base64 characters, which is 132 bits, or 16.5
     // bytes. our implementation generates a 17 byte array instead.
     const fidByteArray = new Uint8Array(17);
+    const crypto =
+      self.crypto || ((self as unknown) as { msCrypto: Crypto }).msCrypto;
     crypto.getRandomValues(fidByteArray);
 
     // Replace the first 4 random bits with the constant FID header of 0b0111.
--- a/packages/installations/src/helpers/generate-fid.ts
+++ b/packages/installations/src/helpers/generate-fid.ts
@@ -17,17 +17,29 @@
 
 import { bufferToBase64UrlSafe } from './buffer-to-base64-url-safe';
 
-/** Generates a new FID using random values from Web Crypto API. */
+export const VALID_FID = /^[cdef][\w-]{21}$/;
+
+/**
+ * Generates a new FID using random values from Web Crypto API.
+ * Returns an empty string if FID generation fails for any reason.
+ */
 export function generateFid(): string {
-  // A valid FID has exactly 22 base64 characters, which is 132 bits, or 16.5
-  // bytes. our implementation generates a 17 byte array instead.
-  const fidByteArray = new Uint8Array(17);
-  crypto.getRandomValues(fidByteArray);
+  try {
+    // A valid FID has exactly 22 base64 characters, which is 132 bits, or 16.5
+    // bytes. our implementation generates a 17 byte array instead.
+    const fidByteArray = new Uint8Array(17);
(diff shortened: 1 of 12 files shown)


packages/installations/src/helpers/generate-fid.ts
```diff
@@ -17,17 +17,31 @@
...
-  // A valid FID has exactly 22 base64 characters, which is 132 bits, or 16.5
-  // bytes. our implementation generates a 17 byte array instead.
-  const fidByteArray = new Uint8Array(17);
-  crypto.getRandomValues(fidByteArray);
+  try {
+    // A valid FID has exactly 22 base64 characters, which is 132 bits, or 16.5
+    // bytes. our implementation generates a 17 byte array instead.
+    const fidByteArray = new Uint8Array(17);
+    const crypto =
+      self.crypto || ((self as unknown) as { msCrypto: Crypto }).msCrypto;
```
andirayo: out of curiosity:
is this an attempt to fix the clients that currently create "AA=="?

packages/installations/src/helpers/generate-fid.ts
```diff
@@ -17,17 +17,31 @@
...
+
+    // Replace the first 4 random bits with the constant FID header of 0b0111.
+    fidByteArray[0] = 0b01110000 + (fidByteArray[0] % 0b00010000);
 
-  // Replace the first 4 random bits with the constant FID header of 0b0111.
-  fidByteArray[0] = 0b01110000 + (fidByteArray[0] % 0b00010000);
+    const fid = encode(fidByteArray);
 
-  return encode(fidByteArray);
+    return VALID_FID.test(fid) ? fid : '';
```
andirayo: how about using a constant here, e.g. INVALID_FID
with INVALID_FID = '' or INVALID_FID = null

packages/installations/src/helpers/get-installation-entry.ts
```diff
@@ -40,26 +41,33 @@ export interface InstallationEntryWithRegistrationPromise {
...
+      const entryWithPromise = triggerRegistrationIfNecessary(
+        appConfig,
+        installationEntry
+      );
+      registrationPromise = entryWithPromise.registrationPromise;
+      return entryWithPromise.installationEntry;
+    }
+  );
+
+  if (installationEntry.fid === '') {
```
andirayo: INVALID_FID instead of ''?

packages/installations/src/helpers/generate-fid.ts
```diff
@@ -17,17 +17,31 @@
 
 import { bufferToBase64UrlSafe } from './buffer-to-base64-url-safe';
 
-/** Generates a new FID using random values from Web Crypto API. */
+export const VALID_FID = /^[cdef][\w-]{21}$/;
```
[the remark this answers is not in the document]
mmermerkaya: JS doesn't support `\A` and `\z`.

packages/installations/src/helpers/generate-fid.ts
mmermerkaya: Done.

packages/installations/src/helpers/get-installation-entry.ts
mmermerkaya: Done.

packages/installations/src/helpers/generate-fid.ts
mmermerkaya: No, it's an unrelated fix for IE 11.

https://developer.mozilla.org/en-US/docs/Web/API/Window/crypto#Syntax

PR #1927, opened 2019-06-28
 Make Document.field() nullable

schmidt-sebastian:
Small cleanup PR to make us do less of `doc.field(path) || null`.

Note that the Public API still returns `undefined` for missing fields (to differentiate from null), but the private API doesn't need to since it returns a wrapped FieldValue.

### DIFF
--- a/packages/firestore/src/model/document.ts
+++ b/packages/firestore/src/model/document.ts
@@ -76,7 +76,7 @@ export class Document extends MaybeDocument {
   }
 
   field(path: FieldPath): FieldValue | null {
-    return this.data.field(path) || null;
+    return this.data.field(path);
   }
 
   fieldValue(path: FieldPath): unknown {
--- a/packages/firestore/src/model/document.ts
+++ b/packages/firestore/src/model/document.ts
@@ -75,8 +75,8 @@ export class Document extends MaybeDocument {
     this.hasCommittedMutations = !!options.hasCommittedMutations;
   }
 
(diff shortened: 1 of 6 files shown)


packages/firestore/src/model/document.ts
```diff
@@ -75,8 +75,8 @@ export class Document extends MaybeDocument {
     this.hasCommittedMutations = !!options.hasCommittedMutations;
   }
 
-  field(path: FieldPath): FieldValue | undefined {
-    return this.data.field(path);
+  field(path: FieldPath): FieldValue | null {
+    return this.data.field(path) || null;
```
mikelehen: It seems like this PR replaces 2 `|| null`s with 1 `|| null` which is technically better, I guess...  except now we have two different `field()` methods (one on Document and one on `ObjectValue`) that treat nonexistence differently (one as null, one as undefined) which seems worse.

If we're going to make this change, I think we should probably at least plumb it down into `ObjectValue.field()` as well...

schmidt-sebastian: I added two more "|| null" in https://github.com/firebase/firebase-js-sdk/pull/1928/ and made an internal value of type "FieldValue | undefined" instead of "undefined" (https://github.com/firebase/firebase-js-sdk/pull/1928/files#diff-af007bc1ca0cf1509f29df38bed7d308R693) which I think it pretty ugly.

I also changed ObjectValue.

PR #1928, opened 2019-06-28
Store the coerced base value in the BaseMutation

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/564

The PR changes what we store in a BaseMutation. With the changes here, every transform operation computes what it needs to store in the BaseMutation. For Numeric increment, this is either an existing numeric value or the IntegerValue.of(0), replacing all invalid base values at time of creation.

### DIFF
--- a/packages/firestore/src/model/field_value.ts
+++ b/packages/firestore/src/model/field_value.ts
@@ -22,10 +22,11 @@ import { Timestamp } from '../api/timestamp';
 import { DatabaseId } from '../core/database_info';
 import { assert, fail } from '../util/assert';
 import { primitiveComparator } from '../util/misc';
-import { SortedMap } from '../util/sorted_map';
-
 import { DocumentKey } from './document_key';
+import { FieldMask } from './mutation';
 import { FieldPath } from './path';
+import { SortedMap } from '../util/sorted_map';
+import { SortedSet } from '../util/sorted_set';
 
 /**
  * Supported data value types:
@@ -603,6 +604,31 @@ export class ObjectValue extends FieldValue {
       }
     });
     return field;
+  }
+
+  /** Recursively extracts the FieldPaths that are set in this ObjectValue. */
+  fieldMask(): FieldMask {
+    let fields = new SortedSet<FieldPath>(FieldPath.comparator);
+    this.internalValue.forEach((key, value) => {
+      const currentPath = new FieldPath([key]);
+      if (value instanceof ObjectValue) {
+        const nestedMask = value.fieldMask();
+        const nestedFields = nestedMask.fields;
+        if (nestedFields.isEmpty()) {
+          // Preserve the empty map by adding it to the FieldMask.
+          fields = fields.add(currentPath);
+        } else {
+          // For nested and non-empty ObjectValues, add the FieldPath of the
+          // leaf nodes.
+          nestedFields.forEach(nestedPath => {
+            fields = fields.add(currentPath.child(nestedPath));
+          });
+        }
+      } else {
+        fields = fields.add(currentPath);
+      }
+    });
+    return FieldMask.fromSet(fields);
   }
 
   toString(): string {
--- a/packages/firestore/src/model/field_value.ts
+++ b/packages/firestore/src/model/field_value.ts
@@ -606,7 +606,10 @@ export class ObjectValue extends FieldValue {
     return field;
   }
 
-  /** Recursively extracts the FieldPaths that are set in this ObjectValue. */
+  /**
+   * Returns a FieldMask built from all FieldPaths starting from this ObjectValue,
+   * including paths from nested objects.
+   */
   fieldMask(): FieldMask {
     let fields = new SortedSet<FieldPath>(FieldPath.comparator);
     this.internalValue.forEach((key, value) => {
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -277,9 +277,9 @@ export abstract class Mutation {
   ): MaybeDocument | null;
 
   /**
-   * If applicable, returns the base value to persist with this mutation. If a
-   * base value is provided, the mutation is always applied to this base value,
-   * even if document has already beenupdated.
+   * If this mutation is not idempotent, returns the base value to persist with
+   * this mutation. If a base value is returned, the mutation is always applied
+   * to this base value, even if document has already been updated.
    *
    * The base value is a sparse object that consists of only the document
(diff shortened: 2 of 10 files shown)


packages/firestore/src/model/field_value.ts
```diff
@@ -605,6 +606,31 @@ export class ObjectValue extends FieldValue {
     return field;
   }
 
+  /** Recursively extracts the FieldPaths that are set in this ObjectValue. */
```
Hui-Wu: I feel like the comments here can be elaborated, something like:

Returns a `FieldMask` built from all `FieldPath`s starting from this `ObjectValue`, including paths from nested objects.

packages/firestore/src/model/mutation.ts
```diff
@@ -302,17 +276,25 @@ export abstract class Mutation {
...
 
-  abstract isEqual(other: Mutation): boolean;
-
   /**
-   * If applicable, returns the field mask for this mutation. Fields that are
-   * not included in this field mask are not modified when this mutation is
-   * applied. Mutations that replace the entire document return 'null'.
+   * If applicable, returns the base value to persist with this mutation. If a
+   * base value is provided, the mutation is always applied to this base value,
+   * even if document has already beenupdated.
```
Hui-Wu: beenupdated -> been updated

packages/firestore/src/model/mutation.ts
```diff
@@ -302,17 +276,25 @@ export abstract class Mutation {
...
+   * behavior for non-idempotent transforms and allow us to return the same
+   * latency-compensated value even if the backend has already applied the
+   * mutation. The base value is null for idempotent mutations, as they can be
+   * re-played even if the backend has already applied them.
+   *
+   * @return a base value to store along with the mutation, or null for
+   * idempotent mutations.
    */
-  abstract get fieldMask(): FieldMask | null;
+  abstract extractBaseValue(maybeDoc: MaybeDocument | null): ObjectValue | null;
```
Hui-Wu: As a reader who don't have enough knowledge, `BaseValue` is a little vague. I could not understand it until after reading the comments. I feel like maybe we should try to come up with a better name, my attempt is `computeBaseValueToApplyMutation`... a bit wordy.

packages/firestore/src/model/mutation.ts
```diff
@@ -302,17 +276,25 @@ export abstract class Mutation {
     localWriteTime: Timestamp
   ): MaybeDocument | null;
 
-  abstract isEqual(other: Mutation): boolean;
-
   /**
-   * If applicable, returns the field mask for this mutation. Fields that are
-   * not included in this field mask are not modified when this mutation is
-   * applied. Mutations that replace the entire document return 'null'.
+   * If applicable, returns the base value to persist with this mutation. If a
```
Hui-Wu: When is this not `applicable`?

packages/firestore/src/model/mutation.ts
```diff
@@ -600,22 +579,29 @@ export class TransformMutation extends Mutation {
...
+  extractBaseValue(maybeDoc: MaybeDocument | null): ObjectValue | null {
+    let baseObject: ObjectValue | null = null;
     for (const fieldTransform of this.fieldTransforms) {
-      if (!fieldTransform.isIdempotent) {
-        return false;
+      const existingValue =
+        maybeDoc instanceof Document
+          ? maybeDoc.field(fieldTransform.field)
+          : undefined;
+      const coercedValue = fieldTransform.transform.computeBaseValue(
```
Hui-Wu: Not a comment but a question: where is the implementation of `computeBaseValue` defined. I cannot find it..

packages/firestore/src/model/field_value.ts
schmidt-sebastian: Done (went with your version).

packages/firestore/src/model/mutation.ts
schmidt-sebastian: Replaced with "If this mutation is not idempotent"

packages/firestore/src/model/mutation.ts
schmidt-sebastian: Fixed

packages/firestore/src/model/mutation.ts
schmidt-sebastian: I can't come up with a better term that is still concise. For better or for worse, the "base" concept is already used in our serialization (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/indexeddb_schema.ts#L429) and may just be one of the core concepts that someone has to ramp up on before they can understand how our mutations are applied. I agree that it would be nicer if the name alone conveyed what this value is used for, but I can't come up with anything short enough to win me over.

packages/firestore/src/model/mutation.ts
schmidt-sebastian: It's added in this PR: https://github.com/firebase/firebase-js-sdk/pull/1928/files#diff-eb38c9da3263f013aac932a74339324dR228

Hui-Wu: Ah, I overlooked..thx.

schmidt-sebastian:
Adding @wilhuff for Owners approval.

packages/firestore/test/integration/api/server_timestamp.test.ts
```diff
@@ -209,6 +209,14 @@ apiDescribe('Server Timestamps', (persistence: boolean) => {
   });
 
   it('can return previous value', () => {
+    // The following update includes an update of the nested map "deep", which
+    // updates it to contain a single ServerTimestamp. As such, the update is
+    // split into two mutations: One that sets "deep" to an empty map and
+    // overwrites the previous ServerTimestamp value and a second ransform that
```
wilhuff: s/ransform/transform/

packages/firestore/test/integration/api/server_timestamp.test.ts
```diff
@@ -209,6 +209,14 @@ apiDescribe('Server Timestamps', (persistence: boolean) => {
   });
 
   it('can return previous value', () => {
+    // The following update includes an update of the nested map "deep", which
```
wilhuff: "The following update includes an update" sounds wrong. Maybe you meant "The following test includes an update"?

Also "As such" doesn't work here. I think what you're trying to say is that the implementation of the update call is to split it into two mutations ...

schmidt-sebastian: Removed "as such" and replaced with "This update".

PR #1935, opened 2019-07-01
Enable Typescript strict flag for Storage

hsubox76:
Enable `"strict": true` Typescript flag for Storage.

### DIFF
--- a/packages/storage/src/implementation/backoff.ts
+++ b/packages/storage/src/implementation/backoff.ts
@@ -66,17 +66,20 @@ export function start(
     }, millis);
   }
 
-  function handler(success: boolean): void {
+  // TODO: This disable can be removed and the 'ignoreRestArgs' option added to
+  // the no-explicit-any rule when ESlint releases it.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function handler(success: boolean, ...args: any[]): void {
     if (triggeredCallback) {
       return;
     }
     if (success) {
-      triggerCallback.apply(null, [...arguments]);
+      triggerCallback.call(null, success, ...args);
       return;
     }
     const mustStop = canceled() || hitTimeout;
     if (mustStop) {
-      triggerCallback.apply(null, [...arguments]);
+      triggerCallback.call(null, success, ...args);
       return;
     }
     if (waitSeconds < 64) {
--- a/packages/storage/src/implementation/backoff.ts
+++ b/packages/storage/src/implementation/backoff.ts
@@ -49,10 +49,13 @@ export function start(
   }
   let triggeredCallback = false;
 
-  function triggerCallback(): void {
+  // TODO: This disable can be removed and the 'ignoreRestArgs' option added to
+  // the no-explicit-any rule when ESlint releases it.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function triggerCallback(...args: any[]): void {
     if (!triggeredCallback) {
       triggeredCallback = true;
-      callback.apply(null, arguments);
+      callback.apply(null, args);
     }
   }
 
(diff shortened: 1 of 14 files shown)


packages/storage/src/implementation/backoff.ts
```diff
@@ -49,10 +49,13 @@ export function start(
   }
   let triggeredCallback = false;
 
-  function triggerCallback(): void {
+  // TODO: This disable can be removed and the 'ignoreRestArgs' option added to
+  // the no-explicit-any rule when ESlint releases it.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function triggerCallback(...args: any[]): void {
```
schmidt-sebastian: Should we follow the same pattern in handler() and change the signature as well? Something like:

```
function handler(success:boolean, ...args:any[])
```

packages/storage/src/implementation/blob.ts
```diff
@@ -30,7 +30,7 @@ import * as type from './type';
  *     modified after this blob's construction.
  */
 export class FbsBlob {
-  private data_: Blob | Uint8Array;
+  private data_!: Blob | Uint8Array;
```
schmidt-sebastian: Can you explain to me what the bang here means?

packages/storage/src/implementation/metadata.ts
```diff
@@ -69,15 +69,15 @@ export function getMappings(): Mappings {
   if (mappings_) {
     return mappings_;
   }
-  const mappings: Array<Mapping<unknown>> = [];
-  mappings.push(new Mapping('bucket'));
-  mappings.push(new Mapping('generation'));
-  mappings.push(new Mapping('metageneration'));
-  mappings.push(new Mapping('name', 'fullPath', true));
+  const mappings: Array<Mapping<string> | Mapping<number>> = [];
```
schmidt-sebastian: Is this not just `const mappings: Mappings`?

packages/storage/src/service.ts
```diff
@@ -63,8 +63,8 @@ export class Service {
    * bucket.
    */
   ref(path?: string): Reference {
-    function validator(path: string): void {
-      if (/^[A-Za-z]+:\/\//.test(path)) {
+    function validator(path: unknown): void {
+      if (/^[A-Za-z]+:\/\//.test(path as string)) {
```
schmidt-sebastian: If you want to go the extra mile, then I would suggest we also validate explicitly that `path` is a string. Optional,

packages/storage/src/implementation/blob.ts
hsubox76: It is a way of us promising the compiler that this variable will be initialized before anything calls it.  The compiler thinks data_ isn't definitively set in the constructor because there's an `if` and 2 `else if`s where it is set, but no `else`.  Since the constructor requires the `data` param to be of a type that would meet all possible `if` conditions, it should always get set in the constructor.

I can fix it by adding an else at the end that assigns `data_` some default value, which seems risky unless you have a suggestion, or I can just look at the types and promise the compiler `data_` is going to get set.  Thoughts?

packages/storage/src/implementation/metadata.ts
hsubox76: Oh right, done.

packages/storage/src/implementation/backoff.ts
hsubox76: Done.

packages/storage/src/service.ts
hsubox76: Will give it a shot.

packages/storage/src/implementation/blob.ts
schmidt-sebastian: Interesting. I have never seen this before. We have a bunch of code in Firestore that we could simplify this way. So keep it as is please :)

packages/storage/src/service.ts
schmidt-sebastian: Can you do the same thing for `refFromURL` below? Thanks!

hsubox76: Done.

PR #1940, opened 2019-07-03
 stub storage differently so tests work in Firefox and IE

Feiyang1:


packages/performance/src/services/remote_config_service.test.ts
```diff
@@ -69,7 +70,12 @@ describe('Performance Monitoring > remote_config_service', () => {
 
   beforeEach(() => {
     fetchStub = stub(self, 'fetch');
-    storageGetItemStub = stub(self.localStorage, 'getItem');
+    // we need to stub the entire localStorage, because storage can't be stubbed in Firefox and IE.
+    // stubbing on self(window) seems to only work the first time (at least in Firefox), the subsequent
+    // tests will have the same stub. stub.reset() in afterEach doesn't help either. As a result, we stub on ApiInstance
```
alikn: Line length. Also, missing period at the end.

packages/performance/src/services/remote_config_service.test.ts
```diff
@@ -88,15 +94,18 @@ describe('Performance Monitoring > remote_config_service', () => {
...
-        storageGetItemFakeFactory(
-          EXPIRY_LOCAL_STORAGE_VALUE,
-          STRINGIFIED_CONFIG
+      const getItemStub = stub();
+      storageStub.value({
+        getItem: getItemStub.callsFake(
+          storageGetItemFakeFactory(
+            EXPIRY_LOCAL_STORAGE_VALUE,
+            STRINGIFIED_CONFIG
+          )
```
alikn: can we put the stubbing part in a function which returns the stub? Something like the pattern discussed here: go/sifers.

Feiyang1: Done. PTAL.

packages/performance/src/services/remote_config_service.test.ts
Feiyang1: done.

PR #1941, opened 2019-07-03
Turn on Typescript strict option for Firestore

hsubox76:
Set `strict: true` in Typescript compiler options.  This required setting a lot of initial values for variables  or promising the compiler (with `!`) that these values would be set before being used, so please check I have guessed the correct values or that the guarantees are correct.

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -134,7 +134,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
   private tokenListener: ((token: string | null) => void) | null = null;
 
   /** Tracks the current User. */
-  private currentUser: User;
+  private currentUser?: User;
 
   /**
    * Counter used to detect if the token changed while a getToken request was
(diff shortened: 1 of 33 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -134,7 +134,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
   private tokenListener: ((token: string | null) => void) | null = null;
 
   /** Tracks the current User. */
-  private currentUser: User;
+  private currentUser?: User;
```
mikelehen: I think this should be initialized to `User.UNAUTHENTICATED` instead.

packages/firestore/src/api/credentials.ts
```diff
@@ -192,7 +192,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
               typeof tokenData.accessToken === 'string',
               'Invalid tokenData returned from getToken():' + tokenData
             );
-            return new OAuthToken(tokenData.accessToken, this.currentUser);
+            return new OAuthToken(tokenData.accessToken, this.currentUser!);
```
mikelehen: Remove `!` after above suggestion.

hsubox76:
This isn't urgent so I can definitely put this on the back burner until someone has time to properly review it.

mikelehen:
I've addressed the strict violations and enabled strict mode via https://github.com/firebase/firebase-js-sdk/pull/2078.  Thanks!

PR #1957, opened 2019-07-10
Fix missing query results when it is a match after a local patch mutation.

Hui-Wu:
Fixing https://github.com/firebase/firebase-android-sdk/issues/155

### DIFF
--- a/packages/firestore/src/local/local_documents_view.ts
+++ b/packages/firestore/src/local/local_documents_view.ts
@@ -19,13 +19,13 @@ import { Query } from '../core/query';
 import { SnapshotVersion } from '../core/snapshot_version';
 import {
   DocumentKeySet,
+  documentKeySet,
   DocumentMap,
   documentMap,
   MaybeDocumentMap,
   maybeDocumentMap,
   NullableMaybeDocumentMap,
   nullableMaybeDocumentMap,
-  documentKeySet
 } from '../model/collections';
 import { Document, MaybeDocument, NoDocument } from '../model/document';
 import { DocumentKey } from '../model/document_key';
@@ -35,10 +35,10 @@ import { ResourcePath } from '../model/path';
 import { assert } from '../util/assert';
 import { IndexManager } from './index_manager';
 import { MutationQueue } from './mutation_queue';
+import { PatchMutation } from '../model/mutation';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { RemoteDocumentCache } from './remote_document_cache';
-import { PatchMutation } from '../model/mutation';
 
 /**
  * A readonly view of the local state of all documents we're tracking (i.e. we
@@ -51,7 +51,7 @@ export class LocalDocumentsView {
     private remoteDocumentCache: RemoteDocumentCache,
     private mutationQueue: MutationQueue,
     private indexManager: IndexManager
-  ) {}
+  ) { }
 
   /**
    * Get the local view of the document identified by `key`.
@@ -228,57 +228,31 @@ export class LocalDocumentsView {
         // that are not in `result` yet, and back fill them via `remoteDocumentCache.getEntries`,
         // otherwise those `PatchMutations` will be ignored because no base document can be found,
         // and lead to missing result for the query.
-        let missingBaseDocEntriesForPatching = documentKeySet();
-        for (const batch of matchingMutationBatches) {
-          for (const mutation of batch.mutations) {
-            // Only process documents belonging to the collection.
-            if (!query.path.isImmediateParentOf(mutation.key.path)) {
-              continue;
+        return this.getDocumentWithPendingPatches(transaction, mutationBatches, results)
+          .next(docsWithPendingPatches => {
+            docsWithPendingPatches.forEach((key, doc) => {
+              if (doc !== null && doc instanceof Document) {
+                results = results.insert(key, doc);
+              }
+            });
+            for (const batch of mutationBatches) {
+              for (const mutation of batch.mutations) {
+                const key = mutation.key;
+                const baseDoc = results.get(key);
+                const mutatedDoc = mutation.applyToLocalView(
+                  baseDoc,
+                  baseDoc,
+                  batch.localWriteTime
+                );
+                if (mutatedDoc instanceof Document) {
+                  results = results.insert(key, mutatedDoc);
+                } else {
+                  results = results.remove(key);
+                }
+              }
             }
-            if (
-              mutation instanceof PatchMutation &&
-              results.get(mutation.key) === null
-            ) {
-              missingBaseDocEntriesForPatching = missingBaseDocEntriesForPatching.add(
-                mutation.key
-              );
-            }
-          }
-        }
-        return this.remoteDocumentCache.getEntries(
-          transaction,
-          missingBaseDocEntriesForPatching
-        );
-      })
-      .next(docsWithPendingPatches => {
-        docsWithPendingPatches.forEach((key, doc) => {
-          if (doc !== null && doc instanceof Document) {
-            results = results.insert(key, doc);
-          }
-        });
-        for (const batch of mutationBatches) {
-          for (const mutation of batch.mutations) {
-            const key = mutation.key;
-            // Only process documents belonging to the collection.
-            if (!query.path.isImmediateParentOf(key.path)) {
-              continue;
-            }
-
-            const baseDoc = results.get(key);
-            const mutatedDoc = mutation.applyToLocalView(
-              baseDoc,
-              baseDoc,
-              batch.localWriteTime
-            );
-            if (mutatedDoc instanceof Document) {
-              results = results.insert(key, mutatedDoc);
-            } else {
-              results = results.remove(key);
-            }
-          }
-        }
-      })
-      .next(() => {
+          });
+      }).next(() => {
         // Finally, filter out any documents that don't actually match
         // the query.
         results.forEach((key, doc) => {
@@ -289,5 +263,29 @@ export class LocalDocumentsView {
 
         return results;
       });
+  }
+
+  private getDocumentWithPendingPatches(
+    transaction: PersistenceTransaction,
+    matchingMutationBatches: MutationBatch[],
+    results: DocumentMap
+  ): PersistencePromise<NullableMaybeDocumentMap> {
+    let missingBaseDocEntriesForPatching = documentKeySet();
+    for (const batch of matchingMutationBatches) {
+      for (const mutation of batch.mutations) {
+        if (
+          mutation instanceof PatchMutation &&
+          results.get(mutation.key) === null
+        ) {
+          missingBaseDocEntriesForPatching = missingBaseDocEntriesForPatching.add(
+            mutation.key
+          );
+        }
+      }
+    }
+    return this.remoteDocumentCache.getEntries(
+      transaction,
+      missingBaseDocEntriesForPatching
+    );
   }
 }
--- a/packages/firestore/src/local/local_documents_view.ts
+++ b/packages/firestore/src/local/local_documents_view.ts
@@ -25,7 +25,7 @@ import {
   maybeDocumentMap,
   NullableMaybeDocumentMap,
   nullableMaybeDocumentMap,
-  documentKeySet,
+  documentKeySet
 } from '../model/collections';
 import { Document, MaybeDocument, NoDocument } from '../model/document';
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,7 +221,43 @@ export class LocalDocumentsView {
...
       })
       .next(matchingMutationBatches => {
+        mutationBatches = matchingMutationBatches;
+        // It is possible that a PatchMutation can make a document a match to the query,
+        // but the version in `remoteDocumentCache` is not a match yet (waiting for server
+        // to ack). To handle this, we find all document keys affected by some`PatchMutation`s
+        // that are not in `result` yet, and back fill them via `remoteDocumentCache.getEntries`,
+        // otherwise those `PatchMutations` will be ignored because no base document can be found,
+        // and lead to missing result for the query.
+        let missingBaseDocEntriesForPatching = documentKeySet();
```
var-const: Nit: the added code makes the function considerably longer. Perhaps refactor it out into a helper function? Something like:
```javascript
.next(matchingMutationBatches => {
  mutationBatches = matchingMutationBatches;
  // Comment without changes here (or move it into the function as well)
  return this.getDocsWithLocalPatches(transaction, query, matchingMutationBatches);
})
.next(baseDocsForPatching) {
```

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,7 +221,43 @@ export class LocalDocumentsView {
...
+            }
+          }
+        }
+        return this.remoteDocumentCache.getEntries(
+          transaction,
+          missingBaseDocEntriesForPatching
+        );
+      })
+      .next(baseDocsForPatching => {
+        baseDocsForPatching.forEach((key, doc) => {
```
var-const: Optional: if you decide to follow the suggestion to create a helper function, perhaps you can move this filtering step there as well.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,7 +221,43 @@ export class LocalDocumentsView {
...
+              );
+            }
+          }
+        }
+        return this.remoteDocumentCache.getEntries(
+          transaction,
+          missingBaseDocEntriesForPatching
+        );
+      })
+      .next(baseDocsForPatching => {
```
var-const: Nit: this variable name is perhaps slightly misleading; without context, it sounds like it contains _all_ base documents, but in fact it's only a subset of those documents (those that have pending writes making them potentially match). Consider renaming (`docsWithPendingWrites`? `docsWithLocalPatches`?).

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,7 +221,43 @@ export class LocalDocumentsView {
...
+            // Only process documents belonging to the collection.
+            if (!query.path.isImmediateParentOf(key.path)) {
+              continue;
+            }
+            if (
+              mutation instanceof PatchMutation &&
+              results.get(mutation.key) === null
+            ) {
+              missingBaseDocEntriesForPatching = missingBaseDocEntriesForPatching.add(
+                mutation.key
```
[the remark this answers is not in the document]
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: changed to docsWithPendingPatches.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: If possible, I'd like to keep it as-is. 

It is a part of the main logic of `getDocumentsMatchingCollectionQuery`, read would need to understand this block to understand the whole function. I think in this case, it justifies to have a longer method.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: As i wrote in other reply, i'd like to keep the block as-is.

packages/firestore/src/local/local_documents_view.ts
var-const: I disagree, but I won't insist.

To me, it seems like the general flow of this function would be clearer if it only contained high-level steps. Currently, it delegates a lot of important logic to `remoteDocumentsCache`, and I would say the fact that some parts are hidden behind an abstraction is actually helpful to readability, because the reader can focus on a single concern at a time. However, once again, I don't feel too strongly about it; feel free to keep as is.

var-const:
@schmidt-sebastian Sebastian, since you previously volunteered, could you please check this implementation doesn't leave out any important steps? It looks fine to me, but I'm concerned I might be missing some of the context here.

packages/firestore/test/unit/specs/query_spec.test.ts
```diff
@@ -98,4 +98,41 @@ describeSpec('Queries:', [], () => {
...
+        { match: true },
+        { hasLocalMutations: true }
+      );
+
+      return (
+        spec()
+          .userListens(fullQuery)
+          .watchAcksFull(fullQuery, 1000, docA)
+          .expectEvents(fullQuery, { added: [docA] })
+
```
schmidt-sebastian: I believe that if you do `userUnlisten(fullQuery)` here then you can remove the event right below and further narrow down your test case.

schmidt-sebastian: I am myself no longer fully sold on my idea. But let me explain what I had in mind:

```
    spec()
          .withGcEnabled(false)
          .userListens(fullQuery)
          .watchAcksFull(fullQuery, 1000, docA)
          .expectEvents(fullQuery, { added: [docA] })
          .userUnlistens(fullQuery)
          .userPatches('collection/a', { match: true })
          .userListens(filteredQuery)
          .expectEvents(filteredQuery, {
            added: [docAv2Local],
            fromCache: true,
            hasPendingWrites: true
          })
```

I believe this would work and it would remove the event in line 124 that is not directly related to your change. The reason I am no longer fully sold on this is that you would have to run the test with GC disabled to make this work. Otherwise, the SDK would delete `docA` from the RemoteDocumentCache as the user unlistens.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,29 +221,38 @@ export class LocalDocumentsView {
...
-              baseDoc,
-              batch.localWriteTime
-            );
-            if (mutatedDoc instanceof Document) {
-              results = results.insert(key, mutatedDoc);
-            } else {
-              results = results.remove(key);
+        mutationBatches = matchingMutationBatches;
+        // It is possible that a PatchMutation can make a document a match to the query,
+        // but the version in `remoteDocumentCache` is not a match yet (waiting for server
```
schmidt-sebastian: Nit: s/`remoteDocumentCache`/the RemoteDocumentCache/ (since it is a widely used concept in Firestore)

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,29 +221,38 @@ export class LocalDocumentsView {
...
+        // to ack). To handle this, we find all document keys affected by some`PatchMutation`s
```
schmidt-sebastian: s/some`PatchMutation`s/the PatchMutations/

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,29 +221,38 @@ export class LocalDocumentsView {
...
+        // that are not in `result` yet, and back fill them via `remoteDocumentCache.getEntries`,
+        // otherwise those `PatchMutations` will be ignored because no base document can be found,
+        // and lead to missing result for the query.
+        return this.getDocumentWithPendingPatches(transaction, mutationBatches, results)
```
schmidt-sebastian: `getDocumentWithPendingPatches` only conveys half of the story. Maybe `getMissingBaseDocuments()?

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -252,4 +264,28 @@ export class LocalDocumentsView {
         return results;
       });
   }
+
+  private getDocumentWithPendingPatches(
+    transaction: PersistenceTransaction,
+    matchingMutationBatches: MutationBatch[],
+    results: DocumentMap
+  ): PersistencePromise<NullableMaybeDocumentMap> {
```
schmidt-sebastian: Would it make sense for this to return a DocumentMap? It would simplify the code above, but you would need to post-filter in here. Alternatively, you could also assemble the final result map here and make this a `populateDocumentMapWithMissingBaseDocuments()` or something similar. Optional.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -252,4 +264,28 @@ export class LocalDocumentsView {
         return results;
       });
   }
+
+  private getDocumentWithPendingPatches(
+    transaction: PersistenceTransaction,
+    matchingMutationBatches: MutationBatch[],
+    results: DocumentMap
```
schmidt-sebastian: `results` is not very clear. Maybe something like `existingDocuments`?

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,29 +221,38 @@ export class LocalDocumentsView {
...
+        // that are not in `result` yet, and back fill them via `remoteDocumentCache.getEntries`,
+        // otherwise those `PatchMutations` will be ignored because no base document can be found,
+        // and lead to missing result for the query.
+        return this.getDocumentWithPendingPatches(transaction, mutationBatches, results)
+          .next(docsWithPendingPatches => {
+            docsWithPendingPatches.forEach((key, doc) => {
+              if (doc !== null && doc instanceof Document) {
+                results = results.insert(key, doc);
+              }
+            });
```
schmidt-sebastian: Nit: Add empty line to convey that the loop below does something very different (we move from one stage to another).

packages/firestore/test/unit/specs/query_spec.test.ts
Hui-Wu: Yep, can confirm what you described. Without GC disabled this is not working.

I am keep it as-is as that is slightly more likely what happens in real world.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -218,7 +221,42 @@ export class LocalDocumentsView {
         );
       })
       .next(matchingMutationBatches => {
+        mutationBatches = matchingMutationBatches;
+        // It is possible that a PatchMutation can make a document a match to the query,
+        // but the version in `remoteDocumentCache` is not a match yet (waiting for server
+        // to ack). To handle this, we find all document keys affected by some`PatchMutation`s
+        // that are not in `result` yet, and back fill them via `remoteDocumentCache.getEntries`,
+        // otherwise those `PatchMutations` will be ignored because no base document can be found,
+        // and lead to missing result for the query.
```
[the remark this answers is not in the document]
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: Done.

packages/firestore/src/local/local_documents_view.ts
Hui-Wu: My understanding is `DocumentMap` being an immutable structure makes it awkward to update from the helper function. I have not tried it..but don't expect it to work without extra tricks.

schmidt-sebastian: You can get return the combined set of `existingDocuments` and the result of this function. The goal would be to simplify the `getDocumentsMatchingCollectionQuery` function.

Optional.

Hui-Wu: Decide not to do it because it is not a significant improvement..

PR #1960, opened 2019-07-10
Adding Limit/Limbo spec tests for go/index-free

schmidt-sebastian:
Adding 2 tests for edge cases that require us to parse more documents that contained in the TargetQueryMapping.

### DIFF
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
         .expectEvents(query, { removed: [docC] });
     }
   );
+
+  specTest('Limbo documents stay consistent between views', [], () => {
+    // This tests verifies that a document is consistent between views, even
+    // if the document is only in Limbo in one of the.
+    const originalQuery = Query.atPath(path('collection')).addFilter(
+      filter('matches', '==', true)
+    );
+    const filteredQuery = Query.atPath(path('collection'))
+      .addFilter(filter('matches', '==', true))
+      .addFilter(filter('key', '>=', 'a'));
+
+    const docA = doc('collection/a', 1000, { key: 'a', matches: true });
+    const docACommitted = doc(
+      'collection/a',
+      1000,
+      { key: 'a', matches: true },
+      { hasCommittedMutations: true }
+    );
+    const docBCommitted = doc(
+      'collection/b',
+      1001,
+      { key: 'b', matches: true },
+      { hasCommittedMutations: true }
+    );
+
+    return (
+      client(0)
+        .userSets('collection/a', { key: 'a', matches: true })
+        .userSets('collection/b', { key: 'b', matches: true })
+        .writeAcks('collection/a', 1000)
+        .writeAcks('collection/b', 1001)
+        .userListens(originalQuery)
+        .expectEvents(originalQuery, {
+          added: [docACommitted, docBCommitted],
+          fromCache: true
+        })
+        .watchAcksFull(originalQuery, 2000, docACommitted)
+        .expectLimboDocs(docBCommitted.key)
+        .userListens(filteredQuery)
+        .expectEvents(filteredQuery, {
+          added: [docA, docBCommitted],
+          fromCache: true
+        })
+        .userUnlistens(originalQuery)
+        .expectLimboDocs()
+        // Re-run the query. Note that we still return the unresolved limbo
+        // document `docBCommitted` even though the backend told us earlier that
+        // it does not match.
+        .userListens(originalQuery, 'resume-token-2000')
+        .expectEvents(originalQuery, {
+          added: [docA, docBCommitted],
+          fromCache: true
+        })
+    );
+  });
 });
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -437,54 +437,54 @@ describeSpec('Limbo Documents:', [], () => {
 
   specTest('Limbo documents stay consistent between views', [], () => {
     // This tests verifies that a document is consistent between views, even
-    // if the document is only in Limbo in one of the.
-    const originalQuery = Query.atPath(path('collection')).addFilter(
-      filter('matches', '==', true)
-    );
+    // if the document is only in Limbo in one of them.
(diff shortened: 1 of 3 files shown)


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1021,11 +1021,11 @@ abstract class TestRunner {
       );
     });
     for (const expectedLimboDoc of this.expectedLimboDocs) {
-      expect(
-        actualLimboDocs.get(expectedLimboDoc),
+      expect(actualLimboDocs.get(expectedLimboDoc)).to.not.equal(
```
schmidt-sebastian: FYI: Using `equal` makes the custom error message show up.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
...
+        .userSets('collection/a', { key: 'a', matches: true })
+        .userSets('collection/b', { key: 'b', matches: true })
+        .writeAcks('collection/a', 1000)
+        .writeAcks('collection/b', 1001)
+        .userListens(originalQuery)
+        .expectEvents(originalQuery, {
+          added: [docACommitted, docBCommitted],
+          fromCache: true
+        })
+        .watchAcksFull(originalQuery, 2000, docACommitted)
```
wilhuff: I don't understand why watch would only ack `docACommitted` here. Does this presuppose that docB has been modified out of band here? If so please comment.

Also a point of clarity: isn't the thing that watch is acknowledging docA?

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
...
+    // This tests verifies that a document is consistent between views, even
+    // if the document is only in Limbo in one of the.
+    const originalQuery = Query.atPath(path('collection')).addFilter(
+      filter('matches', '==', true)
+    );
+    const filteredQuery = Query.atPath(path('collection'))
+      .addFilter(filter('matches', '==', true))
+      .addFilter(filter('key', '>=', 'a'));
+
+    const docA = doc('collection/a', 1000, { key: 'a', matches: true });
```
wilhuff: FWIW These names read almost backwards to me. Without looking too hard at the contents of the docs, it seems to me as if docACommitted would be the state after we've committed. As it stands it's docACommitted while we're waiting for the commit and then docA after.

docADirty might be a better name for docA with hasCommittedMutations.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
...
+          fromCache: true
+        })
+        .watchAcksFull(originalQuery, 2000, docACommitted)
+        .expectLimboDocs(docBCommitted.key)
+        .userListens(filteredQuery)
+        .expectEvents(filteredQuery, {
+          added: [docA, docBCommitted],
+          fromCache: true
+        })
+        .userUnlistens(originalQuery)
```
wilhuff: Shouldn't watch ack the filteredQuery somewhere in here?

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
...
+        .expectLimboDocs()
+        // Re-run the query. Note that we still return the unresolved limbo
+        // document `docBCommitted` even though the backend told us earlier that
+        // it does not match.
```
wilhuff: And this is because we never saw the event that would resolve the state of the limbo resolution, right?

packages/firestore/test/unit/specs/limit_spec.test.ts
```diff
@@ -155,6 +155,35 @@ describeSpec('Limits:', [], () => {
     }
   );
 
+  specTest('Limits are re-filled from cache', [], () => {
+    // This test verifies that our Query handling backfills a limit query from
+    // cache even if the backend has not told us that an existing
+    // RemoteDocument is within the limit.
+    const fullQuery = Query.atPath(path('collection')).addFilter(
```
wilhuff: Consider naming these consistently fullQuery/limitQuery or queryAll/queryWithLimit

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -434,4 +434,59 @@ describeSpec('Limbo Documents:', [], () => {
         .expectEvents(query, { removed: [docC] });
     }
   );
+
+  specTest('Limbo documents stay consistent between views', [], () => {
+    // This tests verifies that a document is consistent between views, even
+    // if the document is only in Limbo in one of the.
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Those names work for me, and they are a little clearer (at least for everyone that remembers Jonny making up songs about the "dirty bit"). I changed the names to docA, docADirty and docBDirty.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: This makes more sense if docACommitted is just docA here (since Watch doesn't consider it dirty). I also added a comment about the out of band modification of docB, which cases it to go into Limbo.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: It can - but it doesn't matter for the purpose of the test. For all we know, the user's network could be slow at this point.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Yes. Added to comment.

packages/firestore/test/unit/specs/limit_spec.test.ts
schmidt-sebastian: Done

PR #1980, opened 2019-07-11
Fix preconditions in transactions

thebrianchen:
Context: [b/135709736](http://b/135709736)
Currently,  if you read a non-existent document, then modify the document twice in the same transaction, each Write will include a precondition that the document did not exist. The server will reject this sequence since each precondition is applied on top of preceding writes in the transaction.

Instead, for the first write to a given key, generate a precondition that takes into account the read time and existence of the document, as we do today. For subsequent writes to a given key, generate a general precondition that only preserves the semantics of the operation: `set()` and `delete()` use  `Precondition(none)` and `update()` uses `Precondition(exists)`.

- Update precondition logic to buganizer spec.
- Reorder public and private methods in `transaction.ts`.
- Add table tests for testing transaction combinations.

### DIFF
--- a/packages/firestore/src/core/transaction.ts
+++ b/packages/firestore/src/core/transaction.ts
@@ -31,9 +31,7 @@ import { SnapshotVersion } from './snapshot_version';
  * perform and the base versions for any documents read.
  */
 export class Transaction {
-  /**
-   * The version of each document that was read during this transaction.
-   */
+  // The version of each document that was read during this transaction.
   private readVersions = documentVersionMap();
   private mutations: Mutation[] = [];
   private committed = false;
@@ -46,6 +44,9 @@ export class Transaction {
 
   /**
    * Set of documents that have been written in the transaction.
+   *
+   * When there's more than one write to the same key in a transaction, any
+   * writes after hte first are handled differently.
    */
   private writtenDocs: Set<DocumentKey> = new Set();
 
@@ -88,9 +89,6 @@ export class Transaction {
 
   delete(key: DocumentKey): void {
     this.write([new DeleteMutation(key, this.precondition(key))]);
-    // Since the delete will be applied before all following writes, we need to
-    // ensure that the precondition for the next write will be exists: false.
-    // nomore
     this.writtenDocs.add(key);
   }
 
@@ -142,11 +140,6 @@ export class Transaction {
     }
   }
 
-  private write(mutations: Mutation[]): void {
-    this.ensureCommitNotCalled();
-    this.mutations = this.mutations.concat(mutations);
-  }
-
   /**
    * Returns the version of this document when it was read in this transaction,
    * as a precondition, or no precondition if it was not read.
@@ -171,14 +164,15 @@ export class Transaction {
       if (version.isEqual(SnapshotVersion.forDeletedDoc())) {
         // The document doesn't exist, so fail the transaction.
 
-        // This has to be validated locally because you can't send a precondition
-        // that a document does not exist without changing the semantics of the
-        // backend write to be an insert. This is the reverse of what we want,
-        // since we want to assert that the document doesn't exist but then send
-        // the update and have it fail. Since we can't express that to the backend,
-        // we have to validate locally.
+        // This has to be validated locally because you can't send a
+        // precondition that a document does not exist without changing the
+        // semantics of the backend write to be an insert. This is the reverse
+        // of what we want, since we want to assert that the document doesn't
+        // exist but then send the update and have it fail. Since we can't
+        // express that to the backend, we have to validate locally.
 
-        // Note: this can change once we can send separate verify writes in the transaction.
+        // Note: this can change once we can send separate verify writes in the
+        // transaction.
         throw new FirestoreError(
           Code.INVALID_ARGUMENT,
           "Can't update a document that doesn't exist."
@@ -186,12 +180,16 @@ export class Transaction {
       }
       // Document exists, base precondition on document update time.
       return Precondition.updateTime(version);
-      // else !firstWrite GENERAL
     } else {
       // Document was not read, so we just use the preconditions for a blind
       // update.
       return Precondition.exists(true);
     }
+  }
+
+  private write(mutations: Mutation[]): void {
+    this.ensureCommitNotCalled();
+    this.mutations = this.mutations.concat(mutations);
   }
 
   private ensureCommitNotCalled(): void {
--- a/packages/firestore/src/core/transaction.ts
+++ b/packages/firestore/src/core/transaction.ts
@@ -31,7 +31,9 @@ import { SnapshotVersion } from './snapshot_version';
  * perform and the base versions for any documents read.
  */
 export class Transaction {
-  // The version of each document that was read during this transaction.
+  /**
+   * The version of each document that was read during this transaction.
+   */
   private readVersions = documentVersionMap();
   private mutations: Mutation[] = [];
   private committed = false;
@@ -42,7 +44,77 @@ export class Transaction {
    */
   private lastWriteError: FirestoreError;
 
+  /**
+   * Set of documents that have been written in the transaction.
+   */
+  private writtenDocs: Set<DocumentKey> = new Set();
+
   constructor(private datastore: Datastore) {}
+
+  lookup(keys: DocumentKey[]): Promise<MaybeDocument[]> {
+    this.ensureCommitNotCalled();
+
+    if (this.mutations.length > 0) {
+      throw new FirestoreError(
+        Code.INVALID_ARGUMENT,
+        'Firestore transactions require all reads to be executed before all writes.'
+      );
+    }
+    return this.datastore.lookup(keys).then(docs => {
+      docs.forEach(doc => {
+        if (doc instanceof NoDocument || doc instanceof Document) {
+          this.recordVersion(doc);
+        } else {
+          fail('Document in a transaction was a ' + doc.constructor.name);
+        }
+      });
+      return docs;
+    });
+  }
+
+  set(key: DocumentKey, data: ParsedSetData): void {
+    this.write(data.toMutations(key, this.precondition(key)));
+    this.writtenDocs.add(key);
+  }
+
+  update(key: DocumentKey, data: ParsedUpdateData): void {
+    try {
+      this.write(data.toMutations(key, this.preconditionForUpdate(key)));
+    } catch (e) {
+      this.lastWriteError = e;
+    }
+    this.writtenDocs.add(key);
+  }
+
+  delete(key: DocumentKey): void {
+    this.write([new DeleteMutation(key, this.precondition(key))]);
+    // Since the delete will be applied before all following writes, we need to
+    // ensure that the precondition for the next write will be exists: false.
+    // nomore
+    this.writtenDocs.add(key);
+  }
+
+  commit(): Promise<void> {
+    this.ensureCommitNotCalled();
+
+    if (this.lastWriteError) {
+      throw this.lastWriteError;
+    }
+    let unwritten = this.readVersions;
+    // For each mutation, note that the doc was written.
+    this.mutations.forEach(mutation => {
+      unwritten = unwritten.remove(mutation.key);
+    });
+    if (!unwritten.isEmpty()) {
+      throw new FirestoreError(
+        Code.INVALID_ARGUMENT,
+        'Every document read in a transaction must also be written.'
+      );
+    }
+    return this.datastore.commit(this.mutations).then(() => {
+      this.committed = true;
+    });
+  }
 
   private recordVersion(doc: MaybeDocument): void {
     let docVersion: SnapshotVersion;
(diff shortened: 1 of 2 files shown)


packages/firestore/src/core/transaction.ts
```diff
@@ -42,33 +44,12 @@ export class Transaction {
...
-    if (doc instanceof Document) {
-      docVersion = doc.version;
-    } else if (doc instanceof NoDocument) {
-      // For deleted docs, we must use baseVersion 0 when we overwrite them.
-      docVersion = SnapshotVersion.forDeletedDoc();
-    } else {
-      throw fail('Document in a transaction was a ' + doc.constructor.name);
-    }
+  /**
+   * Set of documents that have been written in the transaction.
```
wilhuff: Optional: consider helping your readers understand why this is useful to track. For example, adding a sentence like "When there's more than one write to the same key in a transaction, any writes after the first are handled differently." could help.

Then again, could be overkill.

packages/firestore/src/core/transaction.ts
```diff
@@ -149,16 +83,15 @@ export class Transaction {
...
 
   delete(key: DocumentKey): void {
     this.write([new DeleteMutation(key, this.precondition(key))]);
     // Since the delete will be applied before all following writes, we need to
     // ensure that the precondition for the next write will be exists: false.
-    this.readVersions = this.readVersions.insert(
-      key,
-      SnapshotVersion.forDeletedDoc()
-    );
+    // nomore
```
wilhuff: Uhhhh.

Maybe you meant to delete this whole comment?

packages/firestore/src/core/transaction.ts
```diff
@@ -183,6 +116,84 @@ export class Transaction {
...
+          Code.ABORTED,
+          'Document version changed between two reads.'
+        );
+      }
+    } else {
+      this.readVersions = this.readVersions.insert(doc.key, docVersion);
+    }
+  }
+
+  private write(mutations: Mutation[]): void {
```
wilhuff: Optional: In your new ordering you might put `write` at the bottom to make this flow more (generally a transaction reads a value (which records a version), we generate the precondition for any write and then actually write.

packages/firestore/src/core/transaction.ts
```diff
@@ -183,6 +116,84 @@ export class Transaction {
...
+        // The document doesn't exist, so fail the transaction.
+
+        // This has to be validated locally because you can't send a precondition
+        // that a document does not exist without changing the semantics of the
+        // backend write to be an insert. This is the reverse of what we want,
+        // since we want to assert that the document doesn't exist but then send
+        // the update and have it fail. Since we can't express that to the backend,
+        // we have to validate locally.
+
+        // Note: this can change once we can send separate verify writes in the transaction.
```
wilhuff: Line length limit is 80.

packages/firestore/src/core/transaction.ts
```diff
@@ -183,6 +116,84 @@ export class Transaction {
...
+        throw new FirestoreError(
+          Code.INVALID_ARGUMENT,
+          "Can't update a document that doesn't exist."
+        );
+      }
+      // Document exists, base precondition on document update time.
+      return Precondition.updateTime(version);
+      // else !firstWrite GENERAL
```
wilhuff: ? Please clean up

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,7 +23,184 @@ import * as integrationHelpers from '../util/helpers';
 
 const apiDescribe = integrationHelpers.apiDescribe;
 
-apiDescribe('Database transactions', (persistence: boolean) => {
+apiDescribe.only('Database transactions', (persistence: boolean) => {
```
wilhuff: Remove `.only`

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -63,131 +240,261 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
-            expect(docSnap.exists).to.equal(false);
-            transaction.update(docRef, { foo: 'bar' });
-          });
-        })
-        .then(() => expect.fail('transaction should fail'))
-        .catch((err: firestore.FirestoreError) => {
-          expect(err).to.exist;
-          expect(err.code).to.equal('invalid-argument');
-        });
+  it('run transactions after getting existing document', async () => {
```
wilhuff: These descriptions should read like a sentence. This should read like "It runs transactions ...", not "It run transactions ..."

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -63,131 +240,261 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
-              transaction.update(docRef, { foo: 'new-bar' });
-            });
-          })
-          .catch((err: firestore.FirestoreError) => {
-            expect(err).to.exist;
-            // This is the error surfaced by the backend.
-            expect(err.code).to.equal('invalid-argument');
-          });
-      });
+  it('run transactions after getting non-existent document', async () => {
```
wilhuff: This test description is the same as above but docExists = false here.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -63,131 +240,261 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
-          expect(err).to.exist;
-          expect(err.code).to.equal('invalid-argument');
-        });
+  it('run transactions after getting existing document', async () => {
+    return integrationHelpers.withTestDb(persistence, async db => {
+      const tt = new TransactionTester(db);
+      await tt.validateSequenceSuccess(
+        undefined,
+        /** docExists = */ true,
+        tt.getExisting,
```
wilhuff: If you made these free functions you could avoid prefixing with `tt.` for each of these.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,7 +23,184 @@ import * as integrationHelpers from '../util/helpers';
...
+      return transaction.get(docRef).then(docSnap => {
+        expect(docSnap.exists).to.equal(false);
+      });
+    }
+
+    /**
+     * Validates that executing the provided steps in a transaction results in
+     * the doc value equal to the expected value.
+     *
+     * @param expected The expected value of the doc.
```
wilhuff: Add ... or undefined for no document (if delete is the last step).

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,7 +23,184 @@ import * as integrationHelpers from '../util/helpers';
...
+     * @param docExists Whether the document the transaction is being run on
+     * should exist beforehand.
+     * @param stages The series of transaction steps to execute inside the
+     * transaction.
+     */
+    async validateSequenceSuccess(
+      expected: object | undefined,
+      docExists: boolean,
+      ...stages: TransactionStep[]
```
wilhuff: You're calling these `TransactionStep` but the variable here is `stages`. Consider naming these consistently. Otherwise it prompts the question of how is a stage different from a step?

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,7 +23,184 @@ import * as integrationHelpers from '../util/helpers';
...
+     * the doc value equal to the expected value.
+     *
+     * @param expected The expected value of the doc.
+     * @param docExists Whether the document the transaction is being run on
+     * should exist beforehand.
+     * @param stages The series of transaction steps to execute inside the
+     * transaction.
+     */
+    async validateSequenceSuccess(
+      expected: object | undefined,
```
wilhuff: The ordering of these arguments ends up reading weirdly.

First is the expected outcome, then the precondition for the test, then the actual steps. You can't reorder the varargs, but you could make the varargs `TransactionStep | object | undefined` and then pop the last one off as the actual expected outcome.

thebrianchen: I originally implemented it with popping off the last arg, but I realized that if this case where you get an existing document, the document has to be set outside of the `runTransaction` block, which means you can't run it as a `TransactionStage`. I think popping off the last value and the 2nd to last value to do different things is worse for readability compared to having a different order that have comments in the input. Thoughts?

packages/firestore/test/integration/api/transactions.test.ts
thebrianchen: One gets an "existing" document; the other gets a "non-existent" document.

packages/firestore/test/integration/api/transactions.test.ts
wilhuff: We have to balance readability in the caller vs readability in the implementation. Commenting without reordering is a sort of band-aid. It's fine, I guess, but I think we can do better.

This would read better in the caller of we could make it read like: establish precondition, run stages, verify result.

Some alternatives:

  * Take an explicit array of steps as a middle argument rather than varargs
  * Define overloads that enumerate the number of stages
  * Split into multiple chained methods

The overloads thing would essentially define fixed number of stages overloads like this:

```ts
async validateSequenceSuccess(docExists: boolean, stage1: TransactionStep, expected: object | undefined)
async validateSequenceSuccess(docExists: boolean, stage1: TransactionStep, stage2: TransactionStep, expected: object | undefined)
async validateSequenceSuccess(docExists: boolean, stage1: TransactionStep, stage2: TransactionStep, stage3: TransactionStep, expected: object | undefined)
```

This preserves this structure but changes the argument order. You'll need to unpack the arguments here in the usual way with typescript overloads. The approach works in the other languages, which all support proper overloads.

Splitting into chained methods will make the caller look like this:

```ts
tt.withDoc().run(getExisting, delete, delete).expectNoDoc();
tt.withDoc().run(getExisting, delete, update2).expectError('invalid-argument');
```

Note that `withDoc` and `run` don't actually do anything except assign into variables in the tester so they're not async and can return `this`. It's the expect methods that actually do everything.

I think the last form would read best but acknowledge that it's the most effort.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -240,261 +247,184 @@ apiDescribe.only('Database transactions', (persistence: boolean) => {
...
+        .expectDoc({ foo: 'bar2' });
+      await tt
+        .withNonexistentDoc()
+        .run(get, set1, set2)
+        .expectDoc({ foo: 'bar2' });
     });
   });
 
-  it('run transactions on existing document ', async () => {
+  it('runs transactions on existing document ', async () => {
```
wilhuff: nit: remove space after "document"

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,175 +23,182 @@ import * as integrationHelpers from '../util/helpers';
...
+  type TransactionStage = (
     transaction: firestore.Transaction,
     docRef: firestore.DocumentReference
   ) => void;
 
   /**
-   * Used for testing that all possible combinations of executing transactions
-   * result in the desired document value or error.
-   *
    * The transaction steps are postfixed by numbers to indicate the calling
```
wilhuff: s/steps/stages/. Also maybe "These transaction stages" or "The transaction stages that follow" or something like that.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,175 +23,182 @@ import * as integrationHelpers from '../util/helpers';
...
-      await this.db
-        .runTransaction(async transaction => {
-          for (let stage of stages) {
-            await stage(transaction, docRef);
-          }
-          return docRef;
+    async expectDoc(expected: object): Promise<void> {
+      await this.prepareDoc();
+      await this.runTransaction()
+        .then(async () => {
```
wilhuff: Would this get rid of the final chaining then/catch in here?

```ts
try {
  await this.runTransaction();
  const snapshot = await this.docRef.get();
  expect(snapshot.exists).to.equal(true);
  expect(snapshot.data()).to.deep.equal(expected);
} catch (err) {
  expect.fail(...);
}
```

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,175 +23,182 @@ import * as integrationHelpers from '../util/helpers';
...
               ') to succeed, but got ' +
               err
           );
         });
+      this.cleanupTester();
+    }
+
+    async expectError(expected: string): Promise<void> {
+      await this.prepareDoc();
+      await this.runTransaction().then(
```
wilhuff: Similar to the above:

```ts
bool succeeded = false;
try {
  await this.prepareDoc();
  await this.runTransaction();
  succeeded = true;
} catch (err) {
  expect((err as firestore.FirestoreError).code).to.equal(expected);
}
if (succeeded) {
  expect.fail(...);
}
```

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -23,175 +23,182 @@ import * as integrationHelpers from '../util/helpers';
...
-        await docRef.set({ foo: 'bar0' });
-        const docSnap = await docRef.get();
+    private async prepareDoc(): Promise<void> {
+      this.docRef = this.db.collection('tester-docref').doc();
+      if (this.fromExistingDoc) {
+        await this.docRef.set({ foo: 'bar0' });
+        const docSnap = await this.docRef.get();
         expect(docSnap.exists).to.equal(true);
+      } else {
+        const docSnap = await this.docRef.get();
```
wilhuff: This effectively guaranteed true by creating a random document ID. You can drop this check and speed things up.

PR #1983, opened 2019-07-12
Refactor Transaction and SyncEngine to use async/await

thebrianchen:


packages/firestore/src/core/sync_engine.ts
```diff
@@ -381,92 +363,81 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-          }
-          // TODO(klimt): Put in a retry delay?
-          return this.runTransaction(updateFunction, retries - 1);
-        });
-    });
+      userPromise.catch(error => {
+        return Promise.reject<T>(error);
+      });
+      result = await userPromise;
+    } catch (error) {
```
wilhuff: IIUC if you throw in an `async` function that's automatically converted to a rejected promise. You don't have to make this conversion yourself.

That is, I think you can remove this `try/catch`.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -381,92 +363,81 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-        })
-        .catch(error => {
-          if (retries === 0) {
-            return Promise.reject<T>(error);
-          }
-          // TODO(klimt): Put in a retry delay?
-          return this.runTransaction(updateFunction, retries - 1);
-        });
-    });
+      userPromise.catch(error => {
```
wilhuff: This dance seems unnecessary. `userPromise` is already defined as returning `Promise<T>`, so if it fails the result will already be of that type. Meanwhile, the `return Promise.reject<T>(error);` is local only to the function you passed to catch, so you've effectively passed an identity function to catch.

Maybe what's going on is that previously, this was calling `wrapUpdateFunctionError(error)` so it did serve some purpose. Now that you've removed that translation step this can entirely go away.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -973,31 +940,28 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-          return this.localStore.getNewDocumentChanges().then(
-            async changes => {
-              const synthesizedRemoteEvent = RemoteEvent.createSynthesizedRemoteEventForCurrentChange(
-                targetId,
-                state === 'current'
-              );
-              await this.emitNewSnapsAndNotifyLocalStore(
-                changes,
-                synthesizedRemoteEvent
+          try {
```
wilhuff: You only need the extra braces on a `case` statement if there are declarations you want to keep local to the case. Any other construct that introduces a scope serves the same purpose, so the try/catch you've added now makes these braces redundant.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -764,55 +740,51 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+          docChangesInAllViews.push(docChanges);
+        }
+      };
+      queriesProcessed.push(processor);
     });
 
-    await Promise.all(queriesProcessed);
+    for (const queryProcess of queriesProcessed) {
+      await queryProcess();
+    }
```
mikelehen: I think this change makes it so we process the queries serially (we wait for each promise before processing the next query) instead of in parallel (we start processing all the queries and then wait for them all to complete), which will reduce performance.

I think reverting this to how it was would probably be better.

packages/firestore/src/core/sync_engine.ts
mikelehen: I think technically they probably weren't needed before either...  so maybe prettier is just adding them automatically?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -973,31 +940,28 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+              targetId,
+              state === 'current'
+            );
+            await this.emitNewSnapsAndNotifyLocalStore(
+              changes,
+              synthesizedRemoteEvent
+            );
+            break;
+          } catch (error) {
+            if (isDocumentChangeMissingError(error)) {
```
mikelehen: Hrm. I think the prior code was a bit more precisely correct in that it was clear that the DocumentChangeMissingError would come from the getNewDocumentChanges() call, whereas now it could (theoretically) come from createSynthesizedRemoteEventForCurrentChange() or emitNewSnapsAndNotifyLocalStore().

It's probably fine, but I'm noticing that the promise chaining code is in some ways better...  though the new code is perhaps more readable since the many line `.then(..., ...)` construct is probably easy to misread.

Anyway, this is probably good. I'm just thinking out loud.

thebrianchen: Yeah, I see what you mean. Would commenting the source of errors in the `catch()` block be helpful in this case?

packages/firestore/src/core/sync_engine.ts
thebrianchen: yes, prettier is adding them.

packages/firestore/src/core/sync_engine.ts
mikelehen: Probably not necessary, but wouldn't hurt (I leave it up to you. 😄 )

PR #1989, opened 2019-07-15
change getMissingBaseDocuments

Hui-Wu:
`getMissingBaseDocuments` -> `addMissingBaseDocuments`.

Might not make a big difference here, but will make ios/android code more readable.

### DIFF
--- a/packages/firestore/src/local/local_documents_view.ts
+++ b/packages/firestore/src/local/local_documents_view.ts
@@ -51,7 +51,7 @@ export class LocalDocumentsView {
     private remoteDocumentCache: RemoteDocumentCache,
     private mutationQueue: MutationQueue,
     private indexManager: IndexManager
-  ) { }
+  ) {}
 
   /**
    * Get the local view of the document identified by `key`.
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/local_documents_view.ts
```diff
@@ -288,9 +284,16 @@ export class LocalDocumentsView {
         }
       }
     }
-    return this.remoteDocumentCache.getEntries(
-      transaction,
-      missingBaseDocEntriesForPatching
-    );
+    let mergedDocuments = existingDocuments;
```
var-const: Nit: please add a newline above this line for readability. It's easier to scan code when it's split into blocks.

var-const: Question: the only use of `existingDocuments` is to initialize `mergeDocuments`. Is the idea to avoid modifying a function parameter?

Hui-Wu: Done.

Hui-Wu: No, it's mainly there to be more readable: explicit about the merging happening.

I don't feel very strong about this though, i can just use `existingDocuments` if you think it's better.

var-const: I don't feel strongly either way. If you've already started with the iOS port, just use whatever is more similar to the iOS code (in general, we try to keep the code as similar as practical across platforms).

PR #1990, opened 2019-07-15
Add ts-ignore line to allow test:node:persistence to compile

thebrianchen:
Will add node:persistence tests after they are running against emulator

### DIFF
--- a/packages/firestore/test/util/node_persistence.ts
+++ b/packages/firestore/test/util/node_persistence.ts
@@ -16,6 +16,7 @@
  */
 
 import * as fs from 'fs';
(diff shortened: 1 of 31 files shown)


packages/firestore/test/util/node_persistence.ts
```diff
@@ -16,6 +16,7 @@
  */
 
 import * as fs from 'fs';
+// @ts-ignore
```
wilhuff: Consider adding to this comment indicating what's going on. Something like, "there are no types for indexdbshim" or something like that.

PR #1994, opened 2019-07-16
 Fix Spec tests 'Limbo documents stay consistent'

schmidt-sebastian:
The  'Limbo documents stay consistent' used the multi-client runner instead of the spec-runner. This PR fixes the test and adds an assert to make sure that this doesn't happen again.

### DIFF
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -439,25 +439,27 @@ describeSpec('Limbo Documents:', [], () => {
     // This tests verifies that a document is consistent between views, even
     // if the document is only in Limbo in one of them.
     const originalQuery = Query.atPath(path('collection'));
-    const filteredQuery = Query.atPath(path('collection'))
-      .addFilter(filter('matches', '==', true));
+    const filteredQuery = Query.atPath(path('collection')).addFilter(
+      filter('matches', '==', true)
+    );
 
     const docA = doc('collection/a', 1000, { matches: true });
     const docADirty = doc(
       'collection/a',
       1000,
-      {  matches: true },
+      { matches: true },
       { hasCommittedMutations: true }
     );
     const docBDirty = doc(
       'collection/b',
       1001,
-      {  matches: true },
+      { matches: true },
       { hasCommittedMutations: true }
     );
 
     return (
-      client(0)
+      spec()
+        .withGCEnabled(false)
         .userSets('collection/a', { matches: true })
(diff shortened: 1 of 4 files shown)


schmidt-sebastian:
Assigning this to @mikelehen since he is back and its more of his area of expertise :) You are welcome!

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1279,6 +1281,12 @@ export async function runSpec(
   let count = 0;
   try {
     await sequence(steps, async step => {
+      assert(
+        step.clientIndex === undefined || tags.indexOf(MULTI_CLIENT_TAG) !== -1,
+        "Cannot use 'client()' to initialize a test that is not tagged with " +
+          "'multi-client'. Did you mean to use 'spec()'?"
+      );
```
mikelehen: I wonder if it'd make sense to assert the inverse as well (if this is a multi-client test, then clientIndex must be defined)?  Basically:
```
const isMultiClientTest = tags.indexOf(MULTI_CLIENT_TAG) !== -1;
const hasClientIndex = step.clientIndex !== undefined;
assert(isMultiClientTest === hasClientIndex, "Tests can use client() if-and-only-if they are tagged with 'multi-client'.");
```

I don't care strongly if that seems pointless though.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -439,25 +439,27 @@ describeSpec('Limbo Documents:', [], () => {
...
+    const filteredQuery = Query.atPath(path('collection')).addFilter(
+      filter('matches', '==', true)
+    );
 
     const docA = doc('collection/a', 1000, { matches: true });
     const docADirty = doc(
       'collection/a',
       1000,
-      {  matches: true },
+      { matches: true },
```
mikelehen: Hrm. Strange that prettier allowed an extra space before...

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: It's not pointless, but not necessarily worth opening an IDE for and fixing since it would not break anything. I'll leave it as an imaginary TODO :)

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: It didn't. But the change was never pushed.

PR #1995, opened 2019-07-16
Make running against the emulator the default for Firestore integration tests

thebrianchen:
Currently `yarn test:emulator` downloads and runs the emulator, which is necessary when CI runs tests. However, there is currently no way to run tests against a locally running emulator. 

This adds the option to run tests against a locally running emulator on port 8080, which matches the Android SDK's functionality.


packages/firestore/CONTRIBUTING.md
```diff
@@ -16,10 +18,30 @@ See
...
+    ```
+    firebase setup:emulators:firestore
+    ```
+  * Run the emulator
+    ```
+    firebase serve --only firestore
+    ```
 
-## Running Firestore Tests
+### Running Integration Tests
```
mikelehen: We cover both unit and integration tests below so not sure "Integration" is appropriate.

packages/firestore/CONTRIBUTING.md
```diff
@@ -43,7 +65,17 @@ yarn test:browser --integration
...
 # Run a subset of tests whose names match a filter.
 yarn test:browser --grep 'SortedSet keeps elements in the right order'
 yarn test:node --grep 'SortedSet keeps elements in the right order'
+
+# Run tests against against a Firestore emulator (emulator will be downloaded).
+yarn test:emulator
+
+# Run tests against a Firestore emulator running on localhost:8080.
+yarn test:node:emulator
+yarn test:node:persistence:emulator
```
mikelehen: It seems weird that test:emulator and test:node:emulator behave differently wrt the emulator.  I think "node" and "emulator" are separate concerns and adding ":node" shouldn't affect the behavior of ":emulator".

So I think this argues for having ":emulator" and ":running-emulator" or something...  (or ":downloaded-emulator" and ":emulator")

mikelehen:
FWIW another alternative to ":running-emulator" is renaming "test:emulator" to "test:travis" or something, if travis is the only place we want to download the emulator manually.  But again, I'm not clear on the intention.

thebrianchen:
After an offline conversation with @mikelehen, we decided to make running tests against the emulator the default, which is the case for the Android and iOS SDKs. `yarn test:emulator` is renamed to `yarn test:travis` to avoid confusion between downloaded vs. running emulators.

packages/firestore/CONTRIBUTING.md
```diff
@@ -16,10 +18,28 @@ See
 [Automated Setup](https://github.com/firebase/firebase-js-sdk#automated-setup) 
 for more details.
 
+### Setting up the Firestore emulator
+
+The unit and integration tests require that the Firestore emulator is running
```
mikelehen: nit: I _think_ all the unit tests shouldn't connect to the backend at all.

PR #2003, opened 2019-07-17
[Performance] do not access window before initialization

alikn:


packages/performance/index.ts
```diff
@@ -36,6 +36,10 @@ export function registerPerformance(instance: FirebaseNamespace): void {
     if (app.name !== DEFAULT_ENTRY_NAME) {
       throw ERROR_FACTORY.create(ErrorCode.FB_NOT_DEFAULT);
     }
+    if (!window) {
```
Feiyang1: It will still throw exception in Nodejs. `typeof window !== undefined` should work.

alikn: Good point! typeof returns a string. Updated the condition.

imjared:
this seems to still noisily throw errors in a `window`-less environment. for example, we're using in a next.js app and it just throws errors and prevents the app from initializing. is there another way to handle initialization that wouldn't `throw` but just silently fail if in a server environemnt?

```
{
  ReferenceError: window is not defined
    at Object.<anonymous> (../index.ts:51:0)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> ([...]/node_modules/firebase/performance/dist/index.cjs.js:3:1)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Module.require (internal/modules/cjs/loader.js:636:17)
    at require (internal/modules/cjs/helpers.js:20:18) sourceMapsApplied: true }

alikn:
Performance sdk can be imported but should not be initialized in window-less environments. I suggest to make the initialization of Performance (firebase.performance()) conditional.

PR #2005, opened 2019-07-18
support FIREBASE_DATABASE_EMULATOR_HOST env var

IanWyszynski:
This change add an environment variable for pointing the database emulator SDK at a local realtime database emulator instance. 

There is also a change that implements this functionality in the firebase admin sdk: https://github.com/firebase/firebase-admin-node/pull/589/files

### DIFF
--- a/packages/database/index.ts
+++ b/packages/database/index.ts
@@ -32,7 +32,6 @@ import { setSDKVersion } from './src/core/version';
 
 const ServerValue = Database.ServerValue;
 
-
 export function registerDatabase(instance: FirebaseNamespace) {
   // set SDK_VERSION
   setSDKVersion(instance.SDK_VERSION);
--- a/packages/database/index.ts
+++ b/packages/database/index.ts
@@ -32,7 +32,6 @@ import { setSDKVersion } from './src/core/version';
 
 const ServerValue = Database.ServerValue;
 
-export { FIREBASE_DATABASE_EMULATOR_HOST_VAR } from './src/core/RepoManager';
 
 export function registerDatabase(instance: FirebaseNamespace) {
   // set SDK_VERSION
--- a/packages/database/index.ts
+++ b/packages/database/index.ts
@@ -32,6 +32,8 @@ import { setSDKVersion } from './src/core/version';
 
 const ServerValue = Database.ServerValue;
 
+export { FIREBASE_DATABASE_EMULATOR_HOST_VAR } from './src/core/RepoManager';
+
 export function registerDatabase(instance: FirebaseNamespace) {
   // set SDK_VERSION
   setSDKVersion(instance.SDK_VERSION);
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -27,9 +27,7 @@ import { RepoInfo } from './RepoInfo';
 
 /** @const {string} */
 const DATABASE_URL_OPTION = 'databaseURL';
-
-export const FIREBASE_DATABASE_EMULATOR_HOST_VAR =
-  'FIREBASE_DATABASE_EMULATOR_HOST';
+const FIREBASE_DATABASE_EMULATOR_HOST_VAR = 'FIREBASE_DATABASE_EMULATOR_HOST';
 
 let _staticInstance: RepoManager;
 
@@ -83,11 +81,11 @@ export class RepoManager {
    * @return {!Database}
    */
   databaseFromApp(app: FirebaseApp, url?: string): Database {
-    const dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
-    const dbUrl: string =
-      url || app.options[DATABASE_URL_OPTION] || dbEmulatorHost
-        ? `http://${dbEmulatorHost}`
-        : undefined;
+    let dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
+    if (dbEmulatorHost) {
+      dbEmulatorHost = `http://${dbEmulatorHost}`;
+    }
+    const dbUrl: string = url || app.options[DATABASE_URL_OPTION] || dbEmulatorHost
     if (dbUrl === undefined) {
       fatal(
         "Can't determine Firebase Database URL.  Be sure to include " +
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -85,7 +85,8 @@ export class RepoManager {
     if (dbEmulatorHost) {
       dbEmulatorHost = `http://${dbEmulatorHost}`;
     }
-    const dbUrl: string = url || app.options[DATABASE_URL_OPTION] || dbEmulatorHost
+    const dbUrl: string =
+      url || app.options[DATABASE_URL_OPTION] || dbEmulatorHost;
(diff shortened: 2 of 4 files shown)


packages/database/src/core/RepoManager.ts
```diff
@@ -27,6 +27,9 @@ import { RepoInfo } from './RepoInfo';
 
 /** @const {string} */
 const DATABASE_URL_OPTION = 'databaseURL';
+const PROJECT_ID_OPTION = 'projectId';
```
ryanpbrewster: `projectId` is not quite what we want here, usually the namespace name is encoded directly into the `databaseURL` field. This only matters for projects with multiple databases. For instance, one of my projects, `rpb-firestore-3`, has a database instance `rpb-firestore-3-a`. I would connect to it using the SDK like
```
firebase.initializeApp({
  project: "rpb-firestore-3",
  databaseURL: "https://rpb-firestore-3-a.firebaseio.com"
})
```
or, even simpler,
```
firebase.initializeApp({ databaseURL: "https://rpb-firestore-3-a.firebaseio.com" })
```

There is a method, `parseRepoInfo`, that extracts the namespace name and stuffs it into a `RepoInfo` class. Can we use that?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -19,6 +19,8 @@ import { Path } from '../Path';
 import { RepoInfo } from '../../RepoInfo';
 import { warnIfPageIsSecure, warn, fatal } from '../util';
 
+const FIREBASE_DATABASE_EMULATOR_HOST_VAR = 'FIREBASE_DATABASE_EMULATOR_HOST';
```
ryanpbrewster: Is there some way we can define this constant once and re-use it?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -134,6 +136,11 @@ export const parseURL = function(
     scheme = 'https',
     port = 443;
 
+  // Don't use SSL if we are parsing an emulator URL.
+  if (process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR]) {
+    scheme = 'http';
```
ryanpbrewster: This seems unnecessary, aren't we directly extracting the address in the other file modified in this PR?

packages/database/src/core/RepoManager.ts
IanWyszynski: Yep, will remove this.

packages/database/src/core/util/libs/parser.ts
IanWyszynski: To keep consistent with `FIRESTORE_EMULATOR_HOST`, the `FIREBASE_DATABASE_EMULATOR_HOST` must not specify a protocol. Before some changes to the node.js admin sdk which append the protocol identifier before passing a url to this module, this code was needed. It is no longer needed.

packages/database/src/core/RepoManager.ts
```diff
@@ -80,7 +83,10 @@ export class RepoManager {
    * @return {!Database}
    */
   databaseFromApp(app: FirebaseApp, url?: string): Database {
-    const dbUrl: string = url || app.options[DATABASE_URL_OPTION];
+    const dbUrl: string =
+      url ||
+      app.options[DATABASE_URL_OPTION] ||
+      'http://' + process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
```
ryanpbrewster: If the environment variable is not set, we're going to end up with `dbUrl = "http://undefined"`, which seems pretty surprising.

IanWyszynski: Thanks for catching that.

packages/database/index.ts
```diff
@@ -32,6 +32,8 @@ import { setSDKVersion } from './src/core/version';
 
 const ServerValue = Database.ServerValue;
 
+export { FIREBASE_DATABASE_EMULATOR_HOST_VAR } from './src/core/RepoManager';
```
mikelehen: Why are we exporting this here?  Are you hoping to make this part of the public API?  I'm not sure that's desirable...  If we do want to make it part of the public API then it needs to be added to our .d.ts files and we would need to go through an internal API review process...

packages/database/src/core/util/libs/parser.ts
```diff
@@ -18,6 +18,7 @@
 import { Path } from '../Path';
 import { RepoInfo } from '../../RepoInfo';
 import { warnIfPageIsSecure, warn, fatal } from '../util';
+import { FIREBASE_DATABASE_EMULATOR_HOST_VAR } from '../../RepoManager';
```
mikelehen: Is this unused?

packages/database/src/core/RepoManager.ts
```diff
@@ -80,7 +83,11 @@ export class RepoManager {
    * @return {!Database}
    */
   databaseFromApp(app: FirebaseApp, url?: string): Database {
-    const dbUrl: string = url || app.options[DATABASE_URL_OPTION];
+    const dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
+    const dbUrl: string =
+      url || app.options[DATABASE_URL_OPTION] || dbEmulatorUrl
+        ? `http://${dbEmulatorHost}`
+        : undefined;
```
mikelehen: 1) Isn't app.options[DATABASE_URL_OPTION] almost always going to be set?  And so I'm not sure what we gain by falling back to the environment variable?  It seems like we'd want the environment variable to *override* the options. But maybe I'm misunderstanding the use case?
2) Offhand I don't know the operator precedence between `||` and the ternary operator (`... ? ... : ...`) and so this is hard to read.  I would advocate adding parentheses or splitting this up.

packages/database/index.ts
IanWyszynski: Perhaps it shouldn't be exposed. The intended use case is to be able to set this variable from the firebase nodejs admin sdk and I figured having a single source of truth might've been useful, but we can also just define the variable once in each repo and keep them in sync.

mikelehen: Oh!  I see.  That actually is probably fine.  How about just adding a comment then?

packages/database/src/core/RepoManager.ts
IanWyszynski: 1. One use case is when a developer calls `initializeApp` with an empty options object, but has the environment variable set. In this case we want to default to pointing at the emulator.
2. I'll split it up.

mikelehen: Why would the developer call initializeApp() with an empty options object though?  That won't ever work (except if you're using this environment variable), right?  And if you are modifying you code to work with the emulator, you might as well just call initializeApp() and pass the emulator URL for your databaseURL, right?

IanWyszynski: Yep, that's true. I think maybe the variable should override app.options, but not the passed url parameter. The reason is that if the firebase admin sdk already sees the variable, it will pass in the appropriate url as an argument to this package (in which case we want to prefer the argument). However, if the developer is using the package directly, they may want to use this variable to override what is in `databaseURL`.

packages/database/index.ts
IanWyszynski: I don't think it's worth putting this change up through API review. I will define the variables separately and add a cross-referencing comment to both repos.

PR #2006, opened 2019-07-18
Fix a regression that leads to larger bundle size

Feiyang1:
1. `downlevelIteration` adds support for iteration protocol for es5 target, but it adds extra codes for simple array iteration. 
Since it is only required by database, I am only setting it in database's tsconfig.json instead of in the global tsconfig.json to prevent the bloat.

2. `base64` was defined as an object literal before. When we enabled typescript `strict` flag, it was changed to a class. Class that are compiled down to es5 can't be treeshaked in rollup, so base64 is included in all packages that depend on `@firebase/util` though they don't use base64. This changes base64 back to object literal with type safety.

### DIFF
--- a/packages/util/src/crypt.ts
+++ b/packages/util/src/crypt.ts
@@ -86,10 +86,10 @@ const byteArrayToString = function(bytes: number[]): string {
 };
 
 interface Base64 {
-  byteToCharMap_: { [key: number]: string } | null ;
-  charToByteMap_: { [key: string]: number } | null ;
-  byteToCharMapWebSafe_: { [key: number]: string } | null ;
-  charToByteMapWebSafe_: { [key: string]: number } | null ;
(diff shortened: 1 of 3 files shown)


packages/util/src/crypt.ts
```diff
@@ -85,56 +85,87 @@ const byteArrayToString = function(bytes: number[]): string {
...
+}
+
+// We define it as an object literal instead of a class because a class compiled down to es5 can't
+// be treeshaked. https://github.com/rollup/rollup/issues/1691
 // Static lookup maps, lazily populated by init_()
-class Base64 {
+export const base64: Base64 = {
   /**
    * Maps bytes to characters.
+   * @type {Object}
```
hsubox76: Probably can re-remove the jsdoc type annotations.

PR #2016, opened 2019-07-23
infer namespace from subdomain in database emulator environment variable

IanWyszynski:
https://github.com/firebase/firebase-js-sdk/pull/2005 added support for point the RTDB javascript sdk at the database emulator. This change improves it by inferring the namespace name from the parsed `databaseURL`  app option field.

### DIFF
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -104,7 +104,7 @@ export class RepoManager {
 
     let dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
     if (dbEmulatorHost) {
-      dbUrl = `http://${dbEmulatorHost}?ns=${parseURL(dbUrl).subdomain}`;
+      dbUrl = `http://${dbEmulatorHost}?ns=${repoInfo.namespace}`;
     }
 
     validateUrl('Invalid Firebase Database URL', 1, parsedUrl);
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -19,7 +19,7 @@ import { FirebaseApp } from '@firebase/app-types';
 import { safeGet } from '@firebase/util';
 import { Repo } from './Repo';
 import { fatal } from './util/util';
-import { parseRepoInfo } from './util/libs/parser';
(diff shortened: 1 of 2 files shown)


packages/database/src/core/RepoManager.ts
```diff
@@ -107,6 +102,11 @@ export class RepoManager {
     const parsedUrl = parseRepoInfo(dbUrl);
     const repoInfo = parsedUrl.repoInfo;
 
+    let dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
+    if (dbEmulatorHost) {
+      dbUrl = `http://${dbEmulatorHost}?ns=${parseURL(dbUrl).subdomain}`;
```
yuchenshi: What if `dbUrl` is already an Emulator URL, like `http://localhost:9000/?ns=foo`? For example, I have a Node.js script that is already configured to write to the emulator but I still want to run it through `emulator:exec` (which will set the env var).

IanWyszynski: In this case the result will be a no-op: https://github.com/firebase/firebase-js-sdk/blob/6b53e0058483c9002d2fe56119f86fc9fb96b56c/packages/database/src/core/util/libs/parser.ts#L185

yuchenshi: Good stuff!

mikelehen: Can we replace `parseURL(dbUrl).subdomain` with `repoInfo.namespace` ?  Seems like we're parsing twice in two different ways...

packages/database/src/core/RepoManager.ts
```diff
@@ -90,12 +90,7 @@ export class RepoManager {
    * @return {!Database}
    */
   databaseFromApp(app: FirebaseApp, url?: string): Database {
-    let dbEmulatorHost = process.env[FIREBASE_DATABASE_EMULATOR_HOST_VAR];
-    if (dbEmulatorHost) {
-      dbEmulatorHost = `http://${dbEmulatorHost}`;
-    }
-    const dbUrl: string =
-      url || dbEmulatorHost || app.options[DATABASE_URL_OPTION];
+    let dbUrl: string = url || app.options[DATABASE_URL_OPTION];
```
mikelehen: This type should probably be `string|undefined` ...

PR #2017, opened 2019-07-24
Surface transaction errors in browser

thebrianchen:
Transactions made via http surface all 400 errors as INVALID_ARGUMENT, when they could also be FAILED_PRECONDITION or OUT_OF_RANGE ([code](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/rpc_error.ts#L247)).

When possible, parse the error message to determine if the server returned a more specific error code.

### DIFF
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -123,12 +123,24 @@ export class WebChannelConnection implements Connection {
                 xhr.getResponseText()
               );
               if (status > 0) {
-                reject(
-                  new FirestoreError(
-                    mapCodeFromHttpStatus(status),
-                    'Server responded with status ' + xhr.getStatusText()
-                  )
-                );
+                const response = xhr.getResponseJson().error;
+                if (!!response) {
+                  const serverError = response.status
+                    .toLowerCase()
+                    .replace('_', '-');
+                  const firestoreError: Code =
+                    Object.values(Code).indexOf(serverError) >= 0
+                      ? (serverError as Code)
+                      : Code.UNKNOWN;
+                  reject(new FirestoreError(firestoreError, response.message));
+                } else {
+                  reject(
+                    new FirestoreError(
+                      mapCodeFromHttpStatus(status),
+                      'Server responded with status ' + xhr.getStatus()
+                    )
+                  );
+                }
               } else {
                 // If we received an HTTP_ERROR but there's no status code,
                 // it's most probably a connection issue
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -32,7 +32,7 @@ import { SDK_VERSION } from '../core/version';
 import { Connection, Stream } from '../remote/connection';
 import {
   mapCodeFromRpcStatus,
-  mapCodeFromHttpReponse
+  mapCodeFromHttpResponseErrorStatus
 } from '../remote/rpc_error';
 import { StreamBridge } from '../remote/stream_bridge';
 import { assert, fail } from '../util/assert';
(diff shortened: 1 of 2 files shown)


packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -123,12 +123,24 @@ export class WebChannelConnection implements Connection {
...
-                );
+                const response = xhr.getResponseJson().error;
+                if (!!response) {
+                  const serverError = response.status
+                    .toLowerCase()
+                    .replace('_', '-');
+                  const firestoreError: Code =
+                    Object.values(Code).indexOf(serverError) >= 0
+                      ? (serverError as Code)
+                      : Code.UNKNOWN;
```
mikelehen: Can you throw this conversion code in a helper function (either in this file or in rpc_error.ts)?  mapCodeFromHttpResponse() or something?  I think giving it a name (and maybe a short comment) will make it more readable and maintainable.

`const serverError = response.status.toLowerCase().replace('_', '-');` looks a bit too arbitrary / magical out-of-context.

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -123,12 +123,24 @@ export class WebChannelConnection implements Connection {
...
+                  reject(new FirestoreError(firestoreError, response.message));
+                } else {
+                  reject(
+                    new FirestoreError(
+                      mapCodeFromHttpStatus(status),
```
mikelehen: I kinda' think we should drop mapCodeFromHttpStatus() since there isn't a one-to-one mapping and just use Code.UNKNOWN here...  WDYT?

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -123,12 +123,24 @@ export class WebChannelConnection implements Connection {
                 xhr.getResponseText()
               );
               if (status > 0) {
-                reject(
-                  new FirestoreError(
-                    mapCodeFromHttpStatus(status),
-                    'Server responded with status ' + xhr.getStatusText()
-                  )
-                );
+                const response = xhr.getResponseJson().error;
```
mikelehen: `response` is a bit misleading since this is only the "error" part of the response.

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -123,12 +123,24 @@ export class WebChannelConnection implements Connection {
...
+                if (!!response) {
+                  const serverError = response.status
+                    .toLowerCase()
+                    .replace('_', '-');
+                  const firestoreError: Code =
```
mikelehen: Since this is a `Code` not a `FirestoreError` I'd rename to avoid confusion (`firestoreErrorCode` or whatever would be fine).

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -123,12 +123,25 @@ export class WebChannelConnection implements Connection {
...
               if (status > 0) {
-                reject(
-                  new FirestoreError(
-                    mapCodeFromHttpStatus(status),
-                    'Server responded with status ' + xhr.getStatusText()
-                  )
-                );
+                const responseError = xhr.getResponseJson().error;
+                if (!!responseError) {
+                  const firestoreErrorCode = mapCodeFromHttpReponse(
```
mikelehen: Reponse => Response

(Optional) I also wonder if we could/should factor out a bit more code.  Maybe call it `firestoreErrorFromRestError()` and pass in the full `responseError` value and return a full `FirestoreError` object?

As-is, this should technically be named something like mapCodeFromHttpResponseErrorStatusCode() or something...

thebrianchen: Doing so would require that we define the type for the responseError or pass it in as `any` with a ts-ignore line. I also added some checks to make sure that the fields are defined.

PR #2018, opened 2019-07-24
Only retry transactions on non-permanent errors

thebrianchen:
Based on Android [port](https://github.com/firebase/firebase-android-sdk/pull/628).

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -71,6 +71,7 @@ import {
   ViewDocumentChanges
 } from './view';
 import { ViewSnapshot } from './view_snapshot';
+import { isPermanentError } from '../remote/rpc_error';
 
 const LOG_TAG = 'SyncEngine';
 
@@ -380,15 +381,22 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
         Error('Transaction callback must return a Promise')
       );
     }
-    const result = await userPromise;
     try {
-      await transaction.commit();
-      return result;
-    } catch (error) {
-      if (retries === 0) {
+      const result = await userPromise;
+      try {
+        await transaction.commit();
+        return result;
+      } catch (error) {
+        if (retries > 0 && this.isRetryableTransactionError(error)) {
+          return await this.runTransaction(updateFunction, retries - 1);
+        }
         return Promise.reject<T>(error);
       }
-      return this.runTransaction(updateFunction, retries - 1);
+    } catch (error) {
+      if (retries > 0 && this.isRetryableTransactionError(error)) {
+        return this.runTransaction(updateFunction, retries - 1);
+      }
+      return Promise.reject<T>(error);
     }
   }
 
@@ -913,6 +921,20 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     }
     this.syncEngineListener!.onWatchChange(newViewSnapshots);
     return activeQueries;
+  }
+
+  private isRetryableTransactionError(error: Error): boolean {
+    if (error.name === 'FirebaseError') {
+      // In transactions, the backend will fail outdated reads with FAILED_PRECONDITION and
+      // non-matching document versions with ABORTED. These errors should be retried.
+      const code = (error as FirestoreError).code;
+      return (
+        code === 'aborted' ||
+        code === 'failed-precondition' ||
+        !isPermanentError(code)
+      );
+    }
+    return false;
   }
 
   // PORTING NOTE: Multi-tab only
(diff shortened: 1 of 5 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -380,15 +381,22 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+          return await this.runTransaction(updateFunction, retries - 1);
+        }
         return Promise.reject<T>(error);
       }
-      return this.runTransaction(updateFunction, retries - 1);
+    } catch (error) {
+      if (retries > 0 && this.isRetryableTransactionError(error)) {
+        return this.runTransaction(updateFunction, retries - 1);
+      }
+      return Promise.reject<T>(error);
```
mikelehen: Why do we need the nested try/try/catch/catch with identical catch bodies?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -915,6 +923,20 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     return activeQueries;
   }
 
+  private isRetryableTransactionError(error: Error): boolean {
+    if (error.name === 'FirebaseError') {
```
mikelehen: Is there a reason not to do an `error instanceof FirestoreError` here? And then you shouldn't need the `as` cast below...

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -556,6 +555,9 @@ apiDescribe('Database transactions', (persistence: boolean) => {
         })
         .then(() => {
           // Now all transaction should be completed, so check the result.
+          // There should be a maximum of 3 retries: once for the 2nd update,
+          // and twice for the 3rd update.
+          expect(started).to.be.lessThan(7);
```
mikelehen: Why is this `7` instead of `6`?  Also, on Android:
1) You renamed started to counter which was a good change.
2) You added a "There should be 3 initial transaction runs." comment above which was a good change.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -800,6 +805,28 @@ apiDescribe('Database transactions', (persistence: boolean) => {
     });
   });
 
+  it('does not retry on permanent errors', () => {
+    return integrationHelpers.withTestDb(persistence, db => {
+      let count = 0;
+      return db
```
mikelehen: On android you had a "Make a transaction that should fail with a permanent error" comment...

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -809,18 +836,18 @@ apiDescribe('Database transactions', (persistence: boolean) => {
   it('are cancelled on rejected promise', () => {
     return integrationHelpers.withTestDb(persistence, db => {
       const doc = db.collection('towns').doc();
-      let count = 0;
+      let counter = 0;
```
mikelehen: I'm not opposed to renaming `count` to `counter`, but you just added a new test above that uses `count`...  so we're not being particularly consistent here... :)

packages/firestore/test/integration/api/transactions.test.ts
thebrianchen: On android it's <= 6, but I don't think `expect` has a `lessThanOrEqualTo` equivalent.

mikelehen: Oh!  Right, that makes sense. :)

packages/firestore/src/core/sync_engine.ts
thebrianchen: Yeah, I initially used `error instanceof FirestoreError`, but the error that is surfaced for the transaction tests `increment transactionally` and `update transactionally` fail that check:
```
 ✗ increment transactionally
	FirebaseError: the stored version (1564089858307124) does not match the required base version (1564089858018004)
	    at new FirestoreError (test/integration/bootstrap.js:16790:11599)
	    at X.<anonymous> (test/integration/bootstrap.js:16331:45296)
	    at nb (test/integration/bootstrap.js:14538:195)
	    at X.../webchannel-wrapper/dist/index.esm.js.g.dispatchEvent (test/integration/bootstrap.js:14536:304)
	    at vd (test/integration/bootstrap.js:14563:390)
	    at xd (test/integration/bootstrap.js:14566:62)
	    at X.../webchannel-wrapper/dist/index.esm.js.g.$a (test/integration/bootstrap.js:14564:363)
	    at X.../webchannel-wrapper/dist/index.esm.js.g.Fa (test/integration/bootstrap.js:14564:339)
```

I'm not sure why this is the case, since I am throwing a new Firestore error (from the most recent webchannel PR).

mikelehen: Huh, interesting...  I'm going to guess it's related to this: https://stackoverflow.com/a/30403029

(the way transpiled inheritance works and the way the native Error object is implemented are incompatible or something)

PR #2029, opened 2019-07-27
use fake "owner" credential when rtdb emulator env var is set

IanWyszynski:
Today, the RTDB emulator accepts any production access token. We would like to support fully-offline emulator interaction that doesn't require an external auth provider. Some questions to follow up on:

- Should we store a bit describing whether the SDK is talking to an emulator in the `RepoInfo` class? 
- Should we only infer that we're talking to an emulator if the environment variable is set? Can we infer it in the additional situation in which an `ns` query parameter is needed (`RepoInfo#needsQueryParam`)?


mikelehen:
Hrm...  I guess I have several interrelated questions...

1. What is the overall user experience we're trying to build here?  I'm surprised that we want to always communicate with the emulator as "owner".  Don't we want to be able to test security rules, etc. which will require authenticating with different credentials?
2. There are a couple of existing "hooks" to replace auth already.  FirebaseAuth hooks into FirebaseApp like (this)[https://github.com/firebase/firebase-js-sdk/blob/8271edfb49a17082c379bae160a0ebcbfbe215cb/packages/auth/src/exports_auth.js#L618] to provide auth functionality.  And RTDB wraps FirebaseApp's token fetching capabilities in its own abstraction: https://github.com/firebase/firebase-js-sdk/blob/8271edfb49a17082c379bae160a0ebcbfbe215cb/packages/database/src/core/AuthTokenProvider.ts#L23. Would it make sense to modify / replace this AuthTokenProvider to incorporate whatever emulator functionality we're aiming for?
3. The Cloud Firestore SDK already has emulator support.  Are we doing something similar there?  Or is RTDB different for some reason?

IanWyszynski:
> Hrm... I guess I have several interrelated questions...
> 
> 1. What is the overall user experience we're trying to build here?  I'm surprised that we want to always communicate with the emulator as "owner".  Don't we want to be able to test security rules, etc. which will require authenticating with different credentials?
> 2. There are a couple of existing "hooks" to replace auth already.  FirebaseAuth hooks into FirebaseApp like (this)[https://github.com/firebase/firebase-js-sdk/blob/8271edfb49a17082c379bae160a0ebcbfbe215cb/packages/auth/src/exports_auth.js#L618] to provide auth functionality.  And RTDB wraps FirebaseApp's token fetching capabilities in its own abstraction: https://github.com/firebase/firebase-js-sdk/blob/8271edfb49a17082c379bae160a0ebcbfbe215cb/packages/database/src/core/AuthTokenProvider.ts#L23
>    . Would it make sense to modify / replace this AuthTokenProvider to incorporate whatever emulator functionality we're aiming for?
> 3. The Cloud Firestore SDK already has emulator support.  Are we doing something similar there?  Or is RTDB different for some reason?

The use case we want to support here is to enable the emulators in environments where we can't get a production access token (e.g., CI). The issue I'm trying to work around in this PR is that we're using a single global application credential for all emulators. 

I just talked offline with @ryanpbrewster and @yuchenshi and I think one path forward is to enforce the following policy in the admin sdk: if either FIRESTORE_EMULATOR_HOST or FIREBASE_DATABASE_EMULATOR_HOST is set, we default to using a single fake credential globally. We are still discussing this, but if we decide to move forward, I will restore [this](https://github.com/firebase/firebase-admin-node/pull/589) PR in the admin sdk and get rid of this one.

Apologies for the back-and-forth on this.

mikelehen:
Per chat discussion, if we move forward with this approach, I think the special handling for RTDB should probably live in a new EmulatorAuthTokenProvider class akin to https://github.com/firebase/firebase-js-sdk/blob/4e84beff2ef5246397cf812d806624ab111c144c/packages/database/src/core/AuthTokenProvider.ts#L25 that just returns "owner" or whatever and swap it in when the env var is set.

packages/database/src/core/EmulatorAuthTokenProvider.ts
```diff
@@ -0,0 +1,44 @@
...
+
+class EmulatorAuthToken implements FirebaseAuthTokenData {
+  constructor(public accessToken: string) {}
+}
+
+export class EmulatorAuthTokenProvider implements TokenProvider {
+  constructor(private app_: FirebaseApp) {}
+
+  getToken(forceRefresh: boolean): Promise<FirebaseAuthTokenData> {
+    return new Promise(() => new EmulatorAuthToken('owner'));
```
yuchenshi:
```suggestion
    return Promise.resolve(EmulatorAuthToken('owner'));
```

yuchenshi: I don't think the original one is correct -- it will probably create a promise that is never resolved or rejected.

mikelehen:
When this is ready for another review, can you please assign it to me?

packages/database/src/core/Constants.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
mikelehen: 2019

packages/database/src/core/Constants.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const FIREBASE_DATABASE_EMULATOR_HOST_VAR =
+  'FIREBASE_DATABASE_EMULATOR_HOST';
```
mikelehen: Having a dedicated file for 1 constant seems overly forward-looking. 😛 

Maybe just add it to core/util/util.ts ?

packages/database/src/core/AuthTokenProvider.ts
```diff
@@ -19,10 +19,24 @@ import { FirebaseApp } from '@firebase/app-types';
...
+  removeTokenChangeListener(listener: (token: string | null) => void);
+
+  notifyForInvalidToken();
+}
+
 /**
  * Abstraction around FirebaseApp's token fetching capabilities.
  */
-export class AuthTokenProvider {
+export class AuthTokenProvider implements TokenProvider {
```
mikelehen: I'd be inclined to rename
AuthTokenProvider => FirebaseAuthTokenProvider
TokenProvider => AuthTokenProvider

packages/database/src/core/ReadonlyRestClient.ts
```diff
@@ -78,7 +81,7 @@ export class ReadonlyRestClient extends ServerActions {
       c: boolean,
       d: number | null
     ) => void,
-    private authTokenProvider_: TokenProvider
+    private authTokenProvider_: AuthTokenProvider
```
mikelehen: You used FirebaseAuthTokenProvider above.  I think they should both be AuthTokenProvider?

packages/database/src/core/AuthTokenProvider.ts
```diff
@@ -19,10 +19,27 @@ import { FirebaseApp } from '@firebase/app-types';
...
 
+/**
+ * An interface for token fetchers.
+ */
+export interface AuthTokenProvider {
+  /**
+   * @param {boolean} forceRefresh
+   * @return {!Promise<FirebaseAuthTokenData>}
+   */
+  getToken(forceRefresh: boolean): Promise<FirebaseAuthTokenData>;
```
schmidt-sebastian: The JS Doc comment adds little value. We can probably remove it, otherwise I'd suggest we document what `forceRefresh` is for.

IanWyszynski: I'd prefer to remove it and add this documentation in a later PR.

PR #2033, opened 2019-07-30
Add spec tests for limit queries with limbo documents

schmidt-sebastian:
This PR adds two spec tests that summarize an offline conversation about potential edge cases for index-free queries (good news is that they pass in the branch). I also made some improvements to our logging which hopefully can cut down future spec test debugging time.

### DIFF
--- a/packages/firestore/src/util/log.ts
+++ b/packages/firestore/src/util/log.ts
@@ -59,10 +59,10 @@ export function setLogLevel(newLevel: LogLevel): void {
 }
 
 export function debug(tag: string, msg: string, ...obj: unknown[]): void {
-  //if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
-  const args = obj.map(argToString);
-  console.debug(`Firestore (${SDK_VERSION}) [${tag}]: ${msg}`, ...args);
-  //}
+  if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
+    const args = obj.map(argToString);
+    logClient.debug(`Firestore (${SDK_VERSION}) [${tag}]: ${msg}`, ...args);
+  }
 }
 
 export function error(msg: string, ...obj: unknown[]): void {
--- a/packages/firestore/src/util/log.ts
+++ b/packages/firestore/src/util/log.ts
@@ -59,10 +59,10 @@ export function setLogLevel(newLevel: LogLevel): void {
 }
 
 export function debug(tag: string, msg: string, ...obj: unknown[]): void {
-  if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
-    const args = obj.map(argToString);
-    logClient.debug(`Firestore (${SDK_VERSION}) [${tag}]: ${msg}`, ...args);
-  }
+  //if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
+  const args = obj.map(argToString);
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/specs/limit_spec.test.ts
```diff
@@ -184,6 +184,109 @@ describeSpec('Limits:', [], () => {
...
+          .restart()
+          // We listen to the limit query again. Note that we include
+          // `firstDocument` in the result since we did not resolve its limbo
+          // state.
+          .userListens(limitQuery, 'resume-token-1001')
+          .expectEvents(limitQuery, { added: [firstDocument], fromCache: true })
+          .watchAcks(limitQuery)
+          // Watch lets us know that it can't reconstruct the query state from
+          // our resume token. Instead, we treat it as a the full query.
+          .watchResets(limitQuery)
```
wilhuff: As discussed in person, this isn't what would customarily happen. Normally watch will send a result with an existence filter and the client will decide it doesn't match and reset.

packages/firestore/test/unit/specs/limit_spec.test.ts
```diff
@@ -184,6 +184,109 @@ describeSpec('Limits:', [], () => {
...
+          .userListens(limitQuery)
+          .watchAcksFull(limitQuery, 1001, firstDocument)
+          .expectEvents(limitQuery, { added: [firstDocument] })
+          .userUnlistens(limitQuery)
+          .watchRemoves(limitQuery)
+          .userListens(fullQuery)
+          .expectEvents(fullQuery, { added: [firstDocument], fromCache: true })
+          .watchAcksFull(fullQuery, 1002, firstDocument, secondDocument)
+          .expectEvents(fullQuery, { added: [secondDocument] })
+          // Watch sends us a remove for `firstDocument`, but doesn't tell us
```
wilhuff: It would be helpful for this comment to describe the scenario we're simulating at a higher level. Something like:

> Assuming some other client deletes `firstDocument`, watch sends us a remove ...

schmidt-sebastian:
I addressed the comments and updated the spec tests. Unfortunately (or fortunately) I am now seeing a breakage with my Index-free PRs.

packages/firestore/src/util/log.ts
```diff
@@ -59,10 +59,10 @@ export function setLogLevel(newLevel: LogLevel): void {
 }
 
 export function debug(tag: string, msg: string, ...obj: unknown[]): void {
-  if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
-    const args = obj.map(argToString);
-    logClient.debug(`Firestore (${SDK_VERSION}) [${tag}]: ${msg}`, ...args);
-  }
+  //if (logClient.logLevel <= FirebaseLogLevel.DEBUG) {
```
wilhuff: Commented out code.

packages/firestore/src/util/log.ts
```diff
@@ -59,10 +59,10 @@ export function setLogLevel(newLevel: LogLevel): void {
...
+  const args = obj.map(argToString);
+  console.debug(`Firestore (${SDK_VERSION}) [${tag}]: ${msg}`, ...args);
```
wilhuff: Intended?

schmidt-sebastian: No. Sorry that is just my preferred way of debugging :/

PR #2034, opened 2019-07-30
Fill License author(google)

SuhwanCha:
fill license author


Feiyang1:
It's the template example, not to be filled by the actual value.

Issue #2035, opened 2019-07-30
Encoding Issue with long polling


glauberramos:
* Operating System version: macos Mojave 10.14.5
  * Browser version: Chrome 75.0.3770.142
  * Firebase SDK version: 6.3.3
  * Firebase Product: database

#### Describe the problem:
Given long polling is enabled, saving strings with an accent (and other non-standard chars) to the realtime database, is throwing an encoding issue.

#### Steps to reproduce:
- Import firebase-app and firebase-database from gstatic. 
- Create a project and realtime database.
- Initialize firebase
- Enabled long polling
- Try to write a string like "éáã"

I am not sure if it is a problem with uglifyJs, but one person found out uglifyJs was causing this same issue. But in my example, I am using the minified version provided by Firebase, I am not using uglifyJs.

https://github.com/firebase/firebase-js-sdk/issues/1356
https://github.com/mishoo/UglifyJS2/issues/3278

#### Relevant Code:
You can see the issue happening here:
https://codepen.io/glauberramos/pen/KOmzRR

```javascript
firebase.initializeApp(firebaseConfig);
firebase.database.INTERNAL.forceLongPolling();
firebase.database().ref('test/').set('éñå')
```

puf:
I can reproduce the issue, here: https://jsbin.com/nasuwem/edit?js,console

```
firebase.database.INTERNAL.forceLongPolling();

ref.set("Česká republika")
   .then(() => console.log("Written") )
   .catch((e) => console.error(e))

ref.on("value", (snapshot) => { console.log(snapshot.val()); })
```
Running this logs:
```
"Česká republika"
"��esk�� republika"
"Written"
```
So the first line is the local event, the second line the value from the server, where it is indeed also broken (see [json](https://stackoverflow.firebaseio.com/57178991.json)).

Removing the `firebase.database.INTERNAL.forceLongPolling()` fixes the problem, ruling out issues in the application code.

schmidt-sebastian:
@glauberramos Thanks for filing this issue. Unfortunately, we are internally using UglifyJs to build our sources, which I believe introduces this problem. Since we have gotten a number of user reports lately, I will get this prioritized again and see if we can find a workaround.

glauberramos:
Thanks, @schmidt-sebastian, it will be very nice if you could fix this issue because it is currently affecting users of my product from many countries. From my understanding, it is a critical bug because people cannot write text normally.

[...]

PR #2039, opened 2019-07-31
fix an obnoxious uglify issue

Feiyang1:
@ryanpbrewster reported an issue where `uglify-js` uglify [this function](https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/crypt.ts#L18) incorrectly. As a result, the uglified code does wrong thing on multi-byte utf8 characters.
This function is being used in RTDB, so some user data in RTDB might have been corrupted.

As an illustration 
```javascript
arr[idx++] = 0;
arr[idx++] = 1;
```
is turned into:
```
arr[idx++] = (arr[idx++] = 1, 0)
```

 The issue doesn't exist in the latest `uglify-js` anymore, so we upgrade `uglify-js` to 3.6.0

Fixes: #2035

### DIFF
--- a/packages/database/CHANGELOG.md
+++ b/packages/database/CHANGELOG.md
@@ -1,3 +1,4 @@
 # Unreleased
+- [fixed] Fixed an uglify issue in minified scripts (`firebase.js` and `firebase-database.js`) where multi-byte UTF-8 characters were encoded incorrectly. 
 - [changed] Improved consistency between the type annotations for `Query.on`/`Reference.on`, 
   `Query.off`/`Reference.off` and `Query.once`/`Reference.once` (#1188, #1204).
--- a/packages/database/CHANGELOG.md
+++ b/packages/database/CHANGELOG.md
@@ -1,4 +1,4 @@
 # Unreleased
-- [fixed] Fixed an uglify issue in minified scripts (`firebase.js` and `firebase-database.js`) where multi-byte UTF-8 characters were encoded incorrectly. 
+- [fixed] Fixed an issue where multi-byte UTF-8 characters would not be written correctly when using `firebase.js` or `firebase-database.js` (https://github.com/firebase/firebase-js-sdk/issues/2035).
 - [changed] Improved consistency between the type annotations for `Query.on`/`Reference.on`, 
   `Query.off`/`Reference.off` and `Query.once`/`Reference.once` (#1188, #1204).
--- a/packages/database/CHANGELOG.md
+++ b/packages/database/CHANGELOG.md
@@ -1,4 +1,4 @@
 # Unreleased
-- [fixed] Fixed an issue where multi-byte UTF-8 characters would not be written correctly when using `firebase.js` or `firebase-database.js` (https://github.com/firebase/firebase-js-sdk/issues/2035).
+- [fixed] Fixed an issue where multi-byte UTF-8 characters would not be written correctly when using `firebase.js` or `firebase-database.js` (#2035).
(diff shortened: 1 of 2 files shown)


schmidt-sebastian:
> I can't think of any way to prevent regressions here. Can we maybe add this to a package.json somewhere?

Isn't this a regression in UglifyJS? They could break us again with a 3.6.1 release if they don't have tests for this.

schmidt-sebastian:
@Feiyang1 Can you add a Changelog entry for the RTDB component?

Feiyang1:
Okay. I added `uglify-js@3.6.0` as a direct dependency in Firebase and excluded it from auto dep update, so we should always get 3.6.0 going forward.

mikelehen:
Is the plan to pin indefinitely?  That seems non-ideal.  Presumably uglify will make improvements in the future that improve bundle size, etc. and we won't get them. 😢 

Sebastian thinks this is the fix on their end: https://github.com/mishoo/UglifyJS2/pull/3430/files  and they added a regression test.

It would not be unreasonable for us to *also* add a regression test, assuming it's not too convoluted to trigger the bug...  So we could do that too.

Not psyched about pinning forever. :)

Feiyang1:
@schmidt-sebastian @ryanpbrewster `uglify-js` did add a test for this use case - https://github.com/mishoo/UglifyJS2/commit/008c23613769a34e48129a5671f6ea22a989d171
but I still think it's a good idea to fix `uglify-js` version as there is no way for us to verify our minified bundle.

mikelehen:
FWIW- Firestore has (had?) the ability to run (most of) our integration tests against minified builds here: https://github.com/firebase/firebase-js-sdk/tree/master/integration/firestore  But unfortunately that _somehow_ got disabled and so they've rotted now.  

Ideally we'd invest in testing our minified builds at some point.  There are lots of ways to break the minified build through our own code changes in addition to UglifyJS changing out from under us.  And I suspect that would be more common (it's happened several times, whereas this is the first/only time UglifyJS has broken us).

So I still think pinning is an over-reaction.  UglifyJS has been very stable in the past and they (seem to) do a good job of adding regression tests.

ryanpbrewster:
My main goal for this PR is to ensure that we don't fall back behind v3.4.10 (which fixed this particular issue). IMO it is safer for us to do this directly in `package.json` than in `yarn.lock`.

Agreed that we should run integration tests against the minified builds.

No particular opinion on pinning. It does seem like the number of bugs in uglify-js is going down rather than up, and I appreciate that each fix has a regression test accompanying it. I think there's not a ton of danger in ratcheting _forward_.

Feiyang1:
Decision: Do not pin `uglify-js` version as it is fairly stable, so we are not too concerned with it and we want to get minification improvements with new versions.

packages/database/CHANGELOG.md
```diff
@@ -1,3 +1,4 @@
 # Unreleased
+- [fixed] Fixed an issue where multi-byte UTF-8 characters would not be written correctly when using `firebase.js` or `firebase-database.js` (https://github.com/firebase/firebase-js-sdk/issues/2035).
```
schmidt-sebastian: s/https://github.com/firebase/firebase-js-sdk/issues/2035/#2035/

puf:
Is this issue just the root cause for https://github.com/firebase/firebase-js-sdk/issues/2035?

schmidt-sebastian:
> Is this issue just the root cause for #2035?

Yes.

packages/database/CHANGELOG.md
Feiyang1: Done.

PR #2045, opened 2019-08-01
port public shutdown to web sdk.

Hui-Wu:


packages/firestore/src/util/async_queue.ts
```diff
@@ -208,12 +218,65 @@ export class AsyncQueue {
...
+
+  /**
+   * Regardless if the queue has initialized shutdown, adds a new operation to the
+   * queue.
+   */
+  private enqueueEvenAfterShutdown<T extends unknown>(
+    op: () => Promise<T>
+  ): Promise<T> {
+    this.verifyNotFailed();
+    // tslint:disable-next-line:no-floating-promises
```
mikelehen: this shouldn't be necessary.

packages/firestore/src/util/async_queue.ts
```diff
@@ -208,12 +218,65 @@ export class AsyncQueue {
...
+   * Once this method is called, the only possible way to request running an operation
+   * is through `enqueueAndForgetEvenAfterShutdown`.
+   */
+  enqueueAndInitilizeShutdown<T extends unknown>(
+    op: () => Promise<T>
+  ): Promise<T> {
+    this.verifyNotFailed();
+    if (this._isShuttingDown) {
+      // Return a Promise resolves right away if it is already shutdown.
+      return new Promise<T>(resolve => resolve(undefined));
```
mikelehen: You would normally use `Promise.resolve()` for this...  but I see that gives a type error because `undefined` isn't convertible to `T`, which is a valid complaint. So I think if this is the behavior we want, the right thing to do here is `return Promise.resolve(undefined as T)` but we are subverting the type system (T likely isn't supposed to include `undefined`).

But as it happens, the only code we pass to `enqueueAndInitilizeShutdown()` returns `Promise<void>` so we should probably just make this method non-generic, which solves our problems.  Finally, you could:
1. Make this method `async` so you can use `await` and the compiler will automatically make all code paths return `Promise` (i.e. it'll generate a `Promise.resolve(undefined)` for the no-op case.
2. Use your `enqueueEvenAfterShutdown()` so you don't need to do the enqueueInternal() / isShuttingDown=true reverse ordering trick.

So you end up with:
```
  async enqueueAndInitilizeShutdown(op: () => Promise<void>): Promise<void> {
    this.verifyNotFailed();
    if (!this._isShuttingDown) {
      this._isShuttingDown = true;
      await this.enqueueEvenAfterShutdown(op);
    }
  }
```

This will cause an error in your async_queue.test.ts test which you can fix by changing:
```
    queue.enqueueAndInitilizeShutdown(() => doStep(2));
```
to:
```
    queue.enqueueAndInitilizeShutdown(async () => { doStep(2); });
```

or you could just change doStop() to return `Promise<void>` (as I mention again below).

packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -209,6 +209,28 @@ describe('AsyncQueue', () => {
...
     expect(completedSteps).to.deep.equal([1, 2]);
   });
+
+  it('Schedules operaions with respect to shut down', async () => {
+    const queue = new AsyncQueue();
+    const completedSteps: number[] = [];
+    const doStep = (n: number): Promise<number> =>
+      defer(() => {
+        const result = completedSteps.push(n);
+        return result;
```
mikelehen: Since you're not using the result, you could drop this return and change this to return `Promise<void>`.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -507,22 +506,19 @@ export class FirestoreClient {
...
-        await this.remoteStore.shutdown();
-        await this.sharedClientState.shutdown();
-        await this.persistence.shutdown();
-
-        // `removeChangeListener` must be called after shutting down the
-        // RemoteStore as it will prevent the RemoteStore from retrieving
-        // auth tokens.
-        this.credentials.removeChangeListener();
-        this._clientShutdown = true;
+    return this.asyncQueue.enqueueAndInitilizeShutdown(async () => {
```
mikelehen: `Initilize` => `Initialize`

FWIW, If you are using VS Code (if not, disregard), there's a "Code Spell Checker" extension from "Street Side Software" which is handy.

wilhuff: Actually, `Initialize` => `Initiate` to match the other ports.

mikelehen:
Fair enough.  My naive alternative is basically "Don't shut down the queue" since it's not actually shutting down and we still actively need it (for clearPersistence())".  But I don't know what breaks if you don't shut it down, and it's probably not worth re-exploring this, so I'll defer my criticism until a point in time where I feel motivated to explore it myself. :)

wilhuff:
Well, the thing that we want is in a few parts:

  * clearPersistence should be able to run after shutdown
  * shutdown should be able to run after shutdown (and have no effect)
  * anything else should fail
  * callbacks submitted to other APIs should not fail.

So far we've chosen to push the is shutdown check down into the async queue, but another way to go would be to push it up into FirestoreClient.

packages/firestore/src/api/database.ts
```diff
@@ -447,6 +447,39 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return deferred.promise;
   }
 
+  /**
+     * Shuts down this Firestore instance.
```
mikelehen: leading spaces seems off...  If you run `yarn lint` from packages/firestore, it should complain about this (and it's complaining in the CI build too).

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1073,7 +1073,7 @@ apiDescribe('Database', (persistence: boolean) => {
   it('can start a new instance after shut down', async () => {
     return withTestDoc(persistence, async docRef => {
       const firestore = docRef.firestore;
-      await firestore.INTERNAL.shutdown();
+      await (firestore as any)._shutdown();
```
mikelehen: `any` is disallowed by the linter (again, running `yarn lint` from packages/firestore will complain).  So you'll need a lint suppression. I'd recommend trying not to repeat it over and over.  So you could create a helper like Brian did for clearPersistence(): https://github.com/firebase/firebase-js-sdk/commit/7a15e7e54bbd80d04311865c46588b68eff93d3b#diff-1422c3d56f022a40f00b40d264351ffdL109

packages/firestore/src/util/async_queue.ts
Hui-Wu: Actually `yarn lint` complains about this.

packages/firestore/test/integration/api/database.test.ts
Hui-Wu: Done.

PR #2050, opened 2019-08-02
Add spec test for limit query initialization

schmidt-sebastian:
Adding a spec test that makes sure that we "refill" Views from cache even for the initial query snapshot.

### DIFF
--- a/packages/firestore/test/unit/specs/limit_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limit_spec.test.ts
@@ -185,6 +185,33 @@ describeSpec('Limits:', [], () => {
   });
 
   specTest(
+    'Initial snapshots for limit queries are re-filled from cache',
+    [],
+    () => {
+      // Verify that views for limit queries are re-filled even if the initial
+      // snapshot does not contain the requested number of results.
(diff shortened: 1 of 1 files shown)


packages/firestore/test/unit/specs/limit_spec.test.ts
```diff
@@ -184,6 +184,33 @@ describeSpec('Limits:', [], () => {
...
+      const limitQuery = Query.atPath(path('collection'))
+        .addFilter(filter('matches', '==', true))
+        .withLimit(2);
+      const doc1 = doc('collection/a', 1001, { matches: true });
+      const doc2 = doc('collection/b', 1002, { matches: true });
+      const doc3 = doc('collection/c', 1003, { matches: true });
+      return spec()
+        .withGCEnabled(false)
+        .userListens(fullQuery)
+        .watchAcksFull(fullQuery, 1003, doc1, doc2, doc3)
```
wilhuff: Shouldn't the result only include two documents, given the limit?

schmidt-sebastian: This is the "full query"

wilhuff: Doh.

PR #2051, opened 2019-08-03
Make initial view test actually use index-free queries

schmidt-sebastian:
Update the spec test I wrote on Friday to actually use index-free queries. Note that I still have all tests passing even with this change.


schmidt-sebastian:
Updated this PR with a test that actually now breaks on Android (since we need to refill the query from cache when it is resumed)

PR #2057, opened 2019-08-06
make renovate update transitive deps in the lock file

Feiyang1:


ryanpbrewster:
I'll defer to @mikelehen here, I assume it's fine to update transitive dependencies but lockfiles are only useful if they're stable so I'm not super confident here.

Feiyang1:
Updating the transitive deps is no different than updating the direct dependencies IMO, so I don't see more risk doing it. 
The lockfile is stable as long as the deps are up to date.

PR #2060, opened 2019-08-07
treat the `ns=` query param as the namespace name if it is present

IanWyszynski:


packages/database/src/core/util/libs/parser.ts
```diff
@@ -122,12 +124,14 @@ export const parseURL = function(
   secure: boolean;
   scheme: string;
   pathString: string;
+  namespace: string;
```
mikelehen: Do we still need `subdomain`?  (seems like `namespace` should probably replace it)

packages/database/src/core/util/libs/parser.ts
```diff
@@ -101,7 +101,9 @@ export const parseRepoInfo = function(
       parsedUrl.host,
       parsedUrl.secure,
       namespace,
-      webSocketOnly
+      webSocketOnly,
+      '',
+      parsedUrl.namespace != parsedUrl.subdomain
```
mikelehen: When using literals or boolean expressions as function arguments, it's often helpful to comment them for the reader, something like:

```
    webSocketOnly,
    /*persistenceKey=*/ '',
    /*withQueryParam=*/ parsedUrl.namespace != parsedUrl.subdomain
```

packages/database/src/core/RepoInfo.ts
```diff
@@ -42,15 +42,20 @@ export class RepoInfo {
     public secure: boolean,
     public namespace: string,
     public webSocketOnly: boolean,
-    public persistenceKey: string = ''
+    public persistenceKey: string = '',
+    private withQueryParam: boolean = false
```
mikelehen: `withQueryParam` is hard to interpret out of context.  Consider `includeNamespaceInQueryParams` or something?

Alternatively I wonder if we really need this...  We could probably change `needsQueryParam()` to include `this.host.indexOf(this.namespace + '.') !== 0` in its checks.

IanWyszynski: I don't think this is what we want. Somewhat unfortunately, the database emulator treats the first subdomain as the namespace name only if the host has 3 path components, and otherwise defaults to the value of the `ns` query parameter, if it exists. 

If, for whatever reason, the host url only contained 2 components, and we didn't pass the ns query parameter, we'd get a "missing namespace" response.

The database emulator does prefer the `ns` query parameter in all cases, if it exists. If the semantics we want are to prefer the `ns` query parameter, I think we should not add more logic that depends on the format of the host url and instead do the thing that is simplest to reason about.

packages/database/src/core/util/libs/parser.ts
IanWyszynski: I'm not sure. `subdomain` means something different than `namespace` and we are relying on that distinction to determine whether we need to pass query parameters in the `RepoInfo` class.

packages/database/src/core/RepoInfo.ts
mikelehen: Bleh.  I don't really know why we have all this code that treats 3 path components specially.  It seems silly.  But rather than deal with potentially breaking stuff, this is fine as-is.

Part of me is tempted to just always include ?ns=blah in the URL as this seems non-harmful (other than a _tiny_ bit of wasted bandwidth) and would mean we don't need all this funky logic.

But it's probably safer to keep this as-is.  Hopefully we don't have to touch this code again.

IanWyszynski:
> Awesome!
> 
> Since this is a breaking change, we should add a change log entry that calls this out. Do you mind doing this?
> 
> I also want your verbal (or rather written) nod that you tested this manually. Thanks.

I tested this manually.

packages/database/src/core/RepoInfo.ts
IanWyszynski: I think we should explore this in later PRs.

PR #2063, opened 2019-08-08
Retrying transactions with backoff

thebrianchen:
I was trying to come up with ways to skip the timeouts in integration tests, and though I'm not a fan overwriting the AsyncQueue instead of initializing it, I couldn't find another way, since we instantiate Firestore through the public API in tests ([link](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/integration/util/helpers.ts)). The downside to doing this that the `readonly` property of `_queue` in the internal `Firestore` has to be removed, but there aren't many places where `_queue` is accessed or set.

Another option to consider is mocking out `setTimeout` within the scope of the `runTransaction` call, but that seems more hacky.


packages/firestore/src/core/firestore_client.ts
```diff
@@ -70,6 +71,17 @@ const DOM_EXCEPTION_INVALID_STATE = 11;
...
+/**
+ * Initial backoff time in milliseconds after an error.
+ * Set to 1s according to https://cloud.google.com/apis/design/errors.
+ */
+const BACKOFF_INITIAL_DELAY_MS = 1000;
+
+/** Maximum backoff time in milliseconds */
+const BACKOFF_MAX_DELAY_MS = 60 * 1000;
+
+const BACKOFF_FACTOR = 1.5;
```
mikelehen: These constants look really familiar!

Maybe we should move them to backoff.ts and just make them be default values for the ExponentialBackoff() constructor?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -367,30 +368,41 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-        return this.runTransaction(updateFunction, retries - 1);
+    let transactionError;
+    for (let i = 0; i < retries; i++) {
+      const backoffDeferred = new Deferred<void>();
+      backoff.backoffAndRun(async () => {
+        backoffDeferred.resolve();
+      });
+
+      try {
+        await backoffDeferred.promise;
```
mikelehen: This is a neat use of async/await to do the retries in a for-loop, but since we don't have async/await on the other platforms, it's probably not going to be very portable. :-/  I'd suggest implementing transaction retries on Android/iOS and then revisiting this approach to try to keep the platforms in sync.

If the non-async/await approach ends up being sufficiently heinous, then perhaps it's worth still doing it on JavaScript, but I'm hoping it's not too bad.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -367,30 +368,41 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-      return result;
-    } catch (error) {
-      if (retries > 0 && this.isRetryableTransactionError(error)) {
-        return this.runTransaction(updateFunction, retries - 1);
+    let transactionError;
+    for (let i = 0; i < retries; i++) {
+      const backoffDeferred = new Deferred<void>();
+      backoff.backoffAndRun(async () => {
+        backoffDeferred.resolve();
+      });
```
mikelehen: This is interesting, but feels like a weird dance to have to do.  I'd be tempted to make backoff.backoffAndRun() return a Promise<> so you could instead do something like:

```
await backoff.backoffAndRun(async () => { });
```
But that's a bit of an odd dance too.  So I'd be tempted to add a new backoff method so you can do:
```
await backoff.awaitBackoff();
```
But now this really highlights the pitfall of this pattern.  By structuring your code like this, it will *not* run on the AsyncQueue which is usually a problem (since it may run intertwined with other SDK code leading to race conditions / inconsistency).  For transactions, this is arguably not a problem since we are already not running them on the AsyncQueue. But TBH I'm unclear about to what degree that's intentional.

Anyway, I guess if we stick with this pattern (see below), I would advocate for backoffAndRun() returning a promise so you don't have to do this Deferred dance.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -20,9 +20,31 @@ import { expect } from 'chai';
...
 import firebase from '../util/firebase_export';
 import * as integrationHelpers from '../util/helpers';
+import { setAsyncQueue } from '../util/internal_helpers';
+import { AsyncQueue, TimerId } from '../../../src/util/async_queue';
+import { CancelablePromise } from '../../../src/util/promise';
 
-const apiDescribe = integrationHelpers.apiDescribe;
+/**
+ * Version of the AsyncQueue that overrides `enqueueAfterDelay` to fast-forward
+ * the backoffs used in retrying transactions.
```
mikelehen: So the whole point of having `TimerId` and a `runDelayedOperationsEarly()` method was so that tests could explicitly fast-forward timers as necessary, _without_ needing to override the AsyncQueue, etc.  So what's the deal here?  Why can't the tests explicitly call `runDelayedOperationsEarly()` as necessary?  Is there some other way we could modify the existing `AsyncQueue` to make transaction retries easily testable?  

I'd like to avoid needing to override / replace the AsyncQueue if at all possible.

mikelehen: In case it's useful, an example of a test using runDelayedOperationsEarly(): https://github.com/firebase/firebase-js-sdk/blob/4be253c0443f59dce79c0a4114b5f4a124ce6c06/packages/firestore/test/integration/api_internal/idle_timeout.test.ts#L30

thebrianchen: The reason I overrode the AsyncQueue was because creating and awaiting the backoff happen inside `runTransaction`. Because of that, once `runTransaction` is called, it will not complete until has completed. How can we separate awaiting the backoff from the rest of the logic?

mikelehen: As a somewhat lame solution, if you just did:

setInterval(() => asyncQueue(db).runDelayedOperationsEarly(TimerId.RetryTransaction), 20);

at the start of your test, it seems like that would probably work?  The lame part is we're just calling it incessantly instead of only when a retry backoff is actually enqueued.

I'm not sure what the best solution would be.  Without mocking the network it seems like we can't know when the  retries will be enqueued so we can't know when to call runDelayedOperationsEarly().  So for this kind of test, maybe we need another AsyncQueue test helper method like `queue.overrideDelaysForTimerId(TimerId.RetryTransaction, 0);` to essentially disable the delays.

WDYT?

thebrianchen: The idea of overriding the delays sounds much more elegant -- wish I could have thought of it sooner! I gave it sho, WDYT?

I'm currently in the process of trying to port this over to android in order to see how the async/await stuff will turn out, but let me know what you think of this approach please

packages/firestore/src/core/sync_engine.ts
thebrianchen: I attempted a draft on Android. I'm not certain that it works yet since my integration tests won't run, but can you take a look to see if it's in the right spiritual direction? ([link](https://github.com/firebase/firebase-android-sdk/compare/bc/tx-backoff?expand=1))

thebrianchen:
@mikelehen I've updated the web version to also use the stubbed exponential backoff in tests to match the Android version.

packages/firestore/src/util/async_queue.ts
```diff
@@ -275,6 +278,15 @@ export class AsyncQueue {
...
     this.delayedOperations.push(delayedOp);
 
+    // Fast-forward delays for timerIds that have been overriden.
+    for (const timerIdToSkip of this.timerIdsToSkip) {
+      if (this.containsDelayedOperation(timerIdToSkip)) {
+        this.runDelayedOperationsEarly(timerIdToSkip)
+          .then(() => {})
+          .catch(() => 'obligatory catch');
+      }
+    }
```
mikelehen: Rather than use runDelayedOperationsEarly(), I think you could just call `delayedOp.skipDelay()` which should be portable to Android?

thebrianchen: I forgot to remove this code snippet, but I've replaced whole `timerIdsToSkip` with `removeTransactionBackoffs` to override the exponential backoff directly. Just pushed again

mikelehen:
Per my Android feedback, I'd prefer go back to  the skipDelaysForTimerId() approach.  I'm also in favor of trying to bring JS more closely inline with Android (and probably iOS).  But we can defer that discussion until the iOS port is further along.

mikelehen:
Random note relating to unifying the platforms and whether to change web to run the transaction logic on the AsyncQueue or not:  Right now, transactions will likely continue to retry even after you call shutdown() on the client.  If we hooked the transaction code into the AsyncQueue, that would automatically be fixed.

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Deferred } from './../util/promise';
+import { TimerId, AsyncQueue } from './../util/async_queue';
+import { ExponentialBackoff } from './../remote/backoff';
```
mikelehen: nit: unnecessary leading `./` on these 3 imports.

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+    private readonly remoteStore: RemoteStore,
+    private readonly updateFunction: (transaction: Transaction) => Promise<T>,
+    private readonly deferred: Deferred<T>
+  ) {
+    this.backoff = new ExponentialBackoff(
+      this.asyncQueue,
+      TimerId.RetryTransaction
+    );
+  }
+  /** Runs the transaction and returns a Promise containing the result. */
```
mikelehen: nit: newline please

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+import { TimerId, AsyncQueue } from './../util/async_queue';
+import { ExponentialBackoff } from './../remote/backoff';
+import { Transaction } from './transaction';
+import { RemoteStore } from '../remote/remote_store';
+import { isNullOrUndefined } from '../util/types';
+import { isPermanentError } from '../remote/rpc_error';
+import { FirestoreError } from '../util/error';
+
+const RETRY_COUNT = 5;
+/**
```
mikelehen: nit: newline please

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+                });
+            });
+          })
+          .catch(error => {
+            this.handleTransactionError(error);
+          });
+      } catch (error) {
+        // Do not retry errors thrown by user provided updateFunction.
+        this.deferred.reject(error);
+      }
```
mikelehen: It's hard to tell what is really inside the try/catch or not because most of this code is run asynchronously and so although it looks like it's inside the try/catch, it's really not (which I think is correct).  To make this clearer you could introduce a helper function:

```
  private tryRunUpdateFunction(transaction: Transaction): Promise<T> | null {
    try {
      const userPromise = this.updateFunction(transaction);
      if (
        isNullOrUndefined(userPromise) ||
        !userPromise.catch ||
        !userPromise.then
      ) {
        this.deferred.reject(
          Error('Transaction callback must return a Promise')
        );
        return null;
      }
      return userPromise;
    } catch (error) {
      // Do not retry errors thrown by user provided updateFunction.
      this.deferred.reject(error);
      return null;
    }
  }
```

And then in here do something like:

```
      const userPromise = this.tryRunUpdateFunction(transaction);
      if (userPromise) {
        ...
      }
```

Up to you.

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+          !userPromise.then
+        ) {
+          this.deferred.reject(
+            Error('Transaction callback must return a Promise')
+          );
+          return;
+        }
+        userPromise
+          .then(result => {
+            return this.asyncQueue.enqueue(() => {
```
mikelehen: I think this should probably be `enqueueAndForget()` since we're not returning the result to anybody (and we don't want errors from here to be handled by the `catch(error => ...)` clause on line 81 below...

packages/firestore/src/core/transaction_runner.ts
```diff
@@ -0,0 +1,113 @@
...
+                .then(() => {
+                  this.deferred.resolve(result);
+                })
+                .catch(error => {
+                  this.handleTransactionError(error);
+                });
+            });
+          })
+          .catch(error => {
+            this.handleTransactionError(error);
```
mikelehen: I think to make sure we're always running our code on the async queue, we should wrap this in an asyncQueue.enqueueAndForget().

packages/firestore/src/core/transaction_runner.ts
thebrianchen: wow, thanks for taking the time to type this out. This is so much more beautiful and readable!

PR #2064, opened 2019-08-08
database sdk not correctly inferring emulator namespace name

IanWyszynski:
In https://github.com/firebase/firebase-js-sdk/pull/2016, we added logic for inferring the namespace name to pass to the RTDB emulator from the result of `parseRepoInfo`, but we didn't actually create a repo class to communicate with it.

### DIFF
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -99,8 +99,8 @@ export class RepoManager {
       );
     }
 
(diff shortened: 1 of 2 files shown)


IanWyszynski:
> Looks reasonable... I guess we didn't test this end-to-end before? Have we done so now?

Yes, I'm not sure how this slipped through. I think I may have been testing this changes concurrently with changes in the Admin SDK when we thought those were going to be necessary. I've tested this manually using only the js sdk.

PR #2066, opened 2019-08-08
Enable strictNullChecks for scripts/emulator-testing/

mikelehen:
I took the emulator-related part out of https://github.com/firebase/firebase-js-sdk/pull/1941/files and reworked it a bit (trying not to use `!` assertions).  I'll work on all the Firestore stuff next (sorry for being so slow).

NOTE: I left out the `stdio: 'inherit'` lines you changed to `stdio: ['inherit', process.stdout, process.stderr]` as I couldn't figure out what that was for.  Let me know if I should pull that in for some reason.


mikelehen:
Interesting...  I merged this as-is, but if you remember how to reproduce the issue with child process output getting lost, I can take a peek and reinstate your change if necessary.  Else, we can leave this as-is and revisit if it comes up again.

PR #2067, opened 2019-08-08
Remove unnecessary tsconfig.json flags.

mikelehen:
tsconfig.base.json enables "strict" now which implicitly enables
"strictFunctionTypes", "strictNullChecks", and "noImplicitAny" (see https://www.typescriptlang.org/docs/handbook/compiler-options.html) so we don't need
to do so explicitly.

We were disabling "alwaysStrict" but that was unnecessary since we have no violations.

### DIFF
--- a/packages/firestore/tsconfig.json
+++ b/packages/firestore/tsconfig.json
@@ -2,6 +2,7 @@
   "extends": "../../config/tsconfig.base.json",
   "compilerOptions": {
     "outDir": "dist",
+    "noImplicitThis": false,
     "strictBindCallApply": false,
     "strictPropertyInitialization": false
   },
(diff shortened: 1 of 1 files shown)


mikelehen:
Oops.  `yarn build` succeeds` but that doesn't validate our test files. So this isn't quite ready.

schmidt-sebastian:
LGTM. I wonder if it makes sense to add a different tsconfig setting for our tests, so that we can enable `noImplicitThis` in the main client code.

schmidt-sebastian:
Please ignore comment (just saw https://github.com/firebase/firebase-js-sdk/pull/2068/files)

Issue #1898, opened 2019-06-21
[6.2.1] Unable to resolve async-storage


KazuyaHara:
Since upgrading to v6.2.1 from 6.2.0, app unable to resolve async-storage correctly. My environment is

  * Firebase SDK version: 6.2.1
  * Expo SDK version: 33.0.6

Maybe it occurred by [this commit](https://github.com/firebase/firebase-js-sdk/commit/5856b9d60ea6c98b5f534a1893572d8437c04a43) and I know it was made to correspond to [this change](https://github.com/firebase/firebase-js-sdk/issues/1847#issue-452686444).
> Starting from RN0.59 AsyncStorage is deprecated from react-native and moved into its own package react-native-async-storage.

It seems like expo-related error that app doesn't return 'MODULE_NOT_FOUND' exception as described in following code but can we have some workaround by firebase side?
https://github.com/firebase/firebase-js-sdk/blob/master/packages/app/index.rn.ts#L34-L36

paarasbhandari:
I'm having the same problem. 
![Simulator Screen Shot - iPhone X - 2019-06-21 at 00 20 38](https://user-images.githubusercontent.com/31348197/59897380-676e6900-93ba-11e9-8a0d-f3872c9d1c5e.png)

aiham:
Looks like it's related to https://github.com/firebase/firebase-js-sdk/commit/5856b9d60ea6c98b5f534a1893572d8437c04a43

aiham:
Would be great to have a workaround for Expo, thanks.

hsubox76:
This is caused by this recent change: https://github.com/firebase/firebase-js-sdk/pull/1879

I think that the problem is that it's only checking for the very specific error code 'MODULE_NOT_FOUND' and perhaps a different error code is being thrown in Expo.  Currently working on a fix that will try to require the React Native core bundled AsyncStorage after any error on the first require call.

In the meantime, the quickest fix would be to install `@react-native-community/async-storage` into your React Native project as it will be required in the future anyway (React Native is deprecating the version bundled with React Native core).

We're also working on getting documentation to reflect this.

hsubox76:
Sorry, it turns out that workaround won't work.  I tried it out in Expo and wasn't able to get it working which turns out to be because Expo currently doesn't allow adding additional native modules without ejecting: https://github.com/react-native-community/async-storage/issues/72

[...]

PR #2073, opened 2019-08-09
Support React Native ^0.59.0

sertony:
Fixing the `import` path for **AsyncStorage** in order to support React Native ^0.59.0


hsubox76:
We attempted this before but had to roll it back when it caused issues: https://github.com/firebase/firebase-js-sdk/issues/1898

The problem is that the React Native packager does not allow conditional requires, so the try/catch block does not gracefully fall back to the built-in package when the community package is not installed.

PR #2074, opened 2019-08-09
Enable strictBindCallApply.

mikelehen:


packages/firestore/src/local/persistence_promise.ts
```diff
@@ -224,10 +224,18 @@ export class PersistencePromise<T> {
...
       | ((r: R, s: S) => PersistencePromise<void>)
       | ((r: R) => PersistencePromise<void>)
+  ): PersistencePromise<void>;
+  static forEach<R>(
+    collection: { forEach: (cb: (r: R) => void) => void },
+    f: (r: R) => PersistencePromise<void>
+  ): PersistencePromise<void>;
+  static forEach<R, S>(
+    collection: { forEach: (cb: (r: R, s?: S) => void) => void },
+    f: (r: R, s?: S) => PersistencePromise<void>
```
mikelehen: I split this into two overloads. One for collections with keys and values, one for collections with just keys.  Without this, you get:

> Argument of type 'S | undefined' is not assignable to parameter of type 'S'.  Type 'undefined' is not assignable to type 'S'.

from the `f.call(this, r, s)` line below (for `s` being undefined).

packages/firestore/src/util/async_queue.ts
```diff
@@ -326,14 +326,14 @@ export class AsyncQueue {
...
+    const delayedOp = DelayedOperation.createAndSchedule<T>(
       this,
       timerId,
       delayMs,
       op,
-      op => this.removeDelayedOperation(op)
+      op => this.removeDelayedOperation(op as DelayedOperation<unknown>)
     );
-    this.delayedOperations.push(delayedOp);
+    this.delayedOperations.push(delayedOp as DelayedOperation<unknown>);
```
mikelehen: Without this you get:

> Type 'DelayedOperation<unknown>' is not assignable to type 'CancelablePromise<T>'.

for the `return delayedOp` below.

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -42,7 +42,7 @@ describeFn('WebChannel', () => {
     const makeUrl = conn.makeUrl.bind(conn);
 
     it('includes project ID and database ID', () => {
-      const url = makeUrl('Commit', {});
+      const url = makeUrl('Commit');
```
mikelehen: This was just a bogus extra parameter.

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -111,6 +107,23 @@ function apiDescribeInternal(
...
+  only: ApiSuiteFunction;
+}
+
+export const apiDescribe = apiDescribeInternal.bind(
+  null,
+  describe
+) as ApiDescribe;
+apiDescribe.skip = apiDescribeInternal.bind(null, describe.skip);
+apiDescribe.only = apiDescribeInternal.bind(null, describe.only);
+
```
mikelehen: Without these changes, we were getting two errors:

> Argument of type 'PendingSuiteFunction' is not assignable to parameter of type 'IContextDefinition'. Type 'PendingSuiteFunction' is missing the following properties from type 'IContextDefinition': only, skip

For passing `describe.skip` to `apiDescribeInternal()` and:

> Property 'skip' does not exist on type '(message: string, testSuite: (persistence: boolean) => void) => void'.ts(2339)

for assigning to `apiDescribe.skip`.

packages/firestore/test/util/equality_matcher.ts
```diff
@@ -106,7 +105,7 @@ export function addEqualityMatcher(): void {
               /*showDiff=*/ true
             );
           } else if (originalFunction) {
-            originalFunction.apply(this, args);
+            originalFunction.call(this, expected, msg);
```
mikelehen: Without these changes we were getting:

> Argument of type 'unknown[]' is not assignable to parameter of type '[unknown, unknown]'.
  Type 'unknown[]' is missing the following properties from type '[unknown, unknown]': 0, 1

packages/firestore/src/util/async_queue.ts
```diff
@@ -326,14 +326,14 @@ export class AsyncQueue {
...
     );
 
-    const delayedOp = DelayedOperation.createAndSchedule<unknown>(
+    const delayedOp = DelayedOperation.createAndSchedule<T>(
       this,
       timerId,
       delayMs,
       op,
-      op => this.removeDelayedOperation(op)
+      op => this.removeDelayedOperation(op as DelayedOperation<unknown>)
```
schmidt-sebastian: Hmmm. This looks wrong to me. Not the code you wrote - but don't we need to call `removeDelayedOperation` and `delayedOperations.push` with delayedOp?

mikelehen: I agree it looks wrong, but it's actually right because unlike the line above, the removal callback is called with the delayedOp. It's pretty confusing though since we call it "op" which is already an in-scope variable.  I renamed it to `removedOp` to avoid ambiguity.  I tried `removedDelayedOp` but that is a bit long and prettier ended up splitting `DelayedOperation<unknown>` into 3 lines which looked horrible.

schmidt-sebastian: Oh. Yes, that is confusing. I even checked out the branch locally and it didn't clear up this confusion :/ Thanks for renaming.

PR #2078, opened 2019-08-12
Enable strictPropertyInitialization for Firestore

mikelehen:
This "fixes" all class members that were left uninitialized by the class constructor, allowing us to enable "strictPropertyInitialization" in our tsconfig.json.

I have tried to do this as a series of commits that can be reviewed individually, with the less trivial commits at the beginning.  If you'd like me to separate some of these out into separate PRs, let me know.

cc/ @hsubox76 This replaces https://github.com/firebase/firebase-js-sdk/pull/1941.  I borrowed your fixes where applicable (thanks!), but there were a number of places where I wanted to do a deeper fix to avoid suppressions or unnecessarily making things nullable, etc.

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -280,15 +280,6 @@ class FirestoreSettings {
   }
 }
 
-class FirestoreConfig {
-  databaseId: DatabaseId;
-  persistenceKey: string;
-  credentials: CredentialsProvider;
-  firebaseApp: FirebaseApp;
-  settings: FirestoreSettings;
-  persistence: boolean;
-}
-
 /**
  * The root reference to the database.
  */
@@ -296,8 +287,11 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
   // The objects that are a part of this API are exposed to third-parties as
   // compiled javascript so we want to flag our private members with a leading
   // underscore to discourage their use.
-  private readonly _config: FirestoreConfig;
   readonly _databaseId: DatabaseId;
+  private readonly _persistenceKey: string;
+  private _credentials: CredentialsProvider;
+  private readonly _firebaseApp: FirebaseApp | null = null;
+  private _settings: FirestoreSettings;
 
   // The firestore client instance. This will be available as soon as
   // configureClient is called, but any calls against it will block until
@@ -314,15 +308,14 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
   _dataConverter: UserDataConverter;
 
   constructor(databaseIdOrApp: FirestoreDatabase | FirebaseApp) {
-    const config = new FirestoreConfig();
     if (typeof (databaseIdOrApp as FirebaseApp).options === 'object') {
       // This is very likely a Firebase app object
       // TODO(b/34177605): Can we somehow use instanceof?
       const app = databaseIdOrApp as FirebaseApp;
-      config.firebaseApp = app;
-      config.databaseId = Firestore.databaseIdFromApp(app);
-      config.persistenceKey = config.firebaseApp.name;
-      config.credentials = new FirebaseCredentialsProvider(app);
+      this._firebaseApp = app;
+      this._databaseId = Firestore.databaseIdFromApp(app);
+      this._persistenceKey = app.name;
+      this._credentials = new FirebaseCredentialsProvider(app);
     } else {
       const external = databaseIdOrApp as FirestoreDatabase;
       if (!external.projectId) {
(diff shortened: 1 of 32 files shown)


mikelehen:
Note: I had to diffbase this on https://github.com/firebase/firebase-js-sdk/pull/2066 because otherwise our tests fail to build/run due to strict violations in the emulator startup code.

packages/firestore/src/model/path.ts
```diff
@@ -63,31 +46,34 @@ abstract class Path {
...
+   */
+  abstract toString(): string;
 
   get length(): number {
     return this.len;
   }
 
-  isEqual(other: Path): boolean {
-    return Path.comparator(this, other) === 0;
+  isEqual(other: this): boolean {
```
schmidt-sebastian: Should this now also be `B`?

packages/firestore/src/model/path.ts
```diff
@@ -63,31 +46,34 @@ abstract class Path {
...
+    return BasePath.comparator(this, other) === 0;
   }
 
-  child(nameOrPath: string | this): this {
+  child(nameOrPath: string | this): B {
     const segments = this.segments.slice(this.offset, this.limit());
-    if (nameOrPath instanceof Path) {
+    if (nameOrPath instanceof BasePath) {
```
schmidt-sebastian: Maybe make this `nameOrPath instanceof this.constructor`. Or leave as is, if you don't like regressions :)

packages/firestore/src/api/database.ts
```diff
@@ -489,30 +478,27 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
   }
 
   private configureClient(
     persistenceSettings: InternalPersistenceSettings
   ): Promise<void> {
-    assert(
-      !!this._config.settings.host,
-      'FirestoreSettings.host cannot be falsey'
-    );
+    assert(!!this._settings.host, 'FirestoreSettings.host cannot be falsey');
```
schmidt-sebastian: s/cannot be falsey/is not set/ if you want :)

packages/firestore/src/api/database.ts
```diff
@@ -495,9 +496,20 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+      PlatformSupport.getPlatform(),
+      databaseInfo,
+      this._credentials,
+      this._queue
+    );
+
+    return this._firestoreClient.start(persistenceSettings);
+  }
+
+  private static makeDataConverter(databaseId: DatabaseId): UserDataConverter {
```
schmidt-sebastian: I think I would like "createDataConverter" or "newDataConverter" slightly better. 

But... could we just make this a non-static function? As long as we return the UserDataConverter, we can still assign it in the constructor.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -207,8 +207,15 @@ export class MemoryTransaction implements PersistenceTransaction {
 }
 
 export class MemoryEagerDelegate implements ReferenceDelegate {
-  private inMemoryPins: ReferenceSet | null;
-  private orphanedDocuments: Set<DocumentKey>;
+  private inMemoryPins: ReferenceSet | null = null;
+  private _orphanedDocuments: Set<DocumentKey> | null = null;
+  private get orphanedDocuments(): Set<DocumentKey> {
```
schmidt-sebastian: Nit: Move this below the constructor.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -677,7 +677,7 @@ export class PersistentWriteStream extends PersistentStream<
    * PersistentWriteStream manages propagating this value from responses to the
    * next request.
    */
-  lastStreamToken: ProtoByteString;
+  lastStreamToken: ProtoByteString = '';
```
schmidt-sebastian: Can we use `emptyByteString()` here? https://github.com/firebase/firebase-js-sdk/blob/df1b58816a5f502853f4b6fde59033abca92f3c9/packages/firestore/src/platform/platform.ts#L85

packages/firestore/src/util/sorted_map.ts
```diff
@@ -526,11 +526,21 @@ export class LLRBNode<K, V> {
 
 // Represents an empty node (a leaf node in the Red-Black Tree).
 export class LLRBEmptyNode<K, V> {
-  key: K;
-  value: V;
-  color: boolean;
-  left: LLRBNode<K, V>;
-  right: LLRBNode<K, V>;
+  get key(): K {
```
schmidt-sebastian: You could change the return type of all of these to "never"

packages/firestore/src/model/path.ts
mikelehen: Sure.  That makes it so you can't use .isEqual() to check for equality between a `ResourcePath` and a `FieldPath` but that's probably good. 👍

packages/firestore/src/model/path.ts
mikelehen: Neat trick, but that breaks TypeScript's type narrowing (so `nameOrPath.foreach()` doesn't work since TypeScript still thinks it may be a string.

But I was able to change nameOrPath to be `string | B` instead of `string | this` which gives us better compile-time type-safety.

packages/firestore/src/api/database.ts
mikelehen: Heh.  SGTM.

packages/firestore/src/local/memory_persistence.ts
mikelehen: Done.

packages/firestore/src/remote/persistent_stream.ts
mikelehen: Yes, good call!

packages/firestore/src/api/database.ts
mikelehen: SGTM.

packages/firestore/src/util/sorted_map.ts
mikelehen: Looks like you're right. Done!

PR #2079, opened 2019-08-12
Don't persist manufactured documents

schmidt-sebastian:
This ports the two line change from Android (https://github.com/firebase/firebase-android-sdk/pull/695), which removes manufactured NoDocuments from cache rather than persisting them.

The main part of this is plumbing through deletes through the RemoteDocumentChangeBuffer. I changed the backing data structure in the change buffer to an ObjectMap since it differentiates between null and undefined. That way, I can add `null` entries and use them for removal.

You might notice that there is a fair bit of overlap between the IndexedDb and Memory implementations of the newly named `modifyEntries`. I tried to unify them but it is actually tricker than I thought (the storage format and the size computation is different between the two implementations).


packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -67,8 +68,14 @@ export abstract class RemoteDocumentChangeBuffer {
 
   /** Buffers a `RemoteDocumentCache.addEntry()` call. */
   addEntry(maybeDocument: MaybeDocument): void {
-    const changes = this.assertChanges();
-    this.changes = changes.insert(maybeDocument.key, maybeDocument);
+    this.assertChanges();
+    this.changes!.set(maybeDocument.key, maybeDocument);
+  }
+
+  /** Buffers a `RemoteDocumentCache.removeEntry()` call. */
```
mikelehen: Is the contract here that you must call getEntry() prior to calling removeEntry() [in order for `documentSizes` to be populated so we can update the size delta]?  Can you document that here (and on addEntry() which presumably has the same contract)?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -487,26 +490,34 @@ class IndexedDbRemoteDocumentChangeBuffer extends RemoteDocumentChangeBuffer {
...
+        delta += size - previousSize!;
+        toApply.push({ key, doc, size });
+      } else {
+        delta -= previousSize!;
+        toApply.push({ key });
+      }
     });
 
-    return this.documentCache.addEntries(transaction, toApply, delta);
+    return this.documentCache.modifyEntries(transaction, toApply, delta);
```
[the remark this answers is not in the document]
schmidt-sebastian: I'll play with this more but the main reason was that we have to apply the size difference twice. I'll see if I can get a clean version of this, especially since the delta for the deletes will likely be zero.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -67,8 +68,14 @@ export abstract class RemoteDocumentChangeBuffer {
 
   /** Buffers a `RemoteDocumentCache.addEntry()` call. */
   addEntry(maybeDocument: MaybeDocument): void {
-    const changes = this.assertChanges();
-    this.changes = changes.insert(maybeDocument.key, maybeDocument);
+    this.assertChanges();
+    this.changes!.set(maybeDocument.key, maybeDocument);
```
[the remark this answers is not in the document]
schmidt-sebastian: I can go back to this. The main reason I removed this was that we now directly modify the instance variable and I thought it was cleaner to make this explicit.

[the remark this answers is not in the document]
mikelehen: FWIW- It might feel slightly cleaner to have a nullable `_changes` property and then make `changes` a non-nullable getter that does the assert, similar to https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/memory_persistence.ts#L215

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1235,21 +1236,27 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
     txn: PersistenceTransaction,
     upperBound: ListenSequenceNumber
   ): PersistencePromise<number> {
```
mikelehen: Is there a reason we can't create / use a RemoteDocumentChangeBuffer here so that it does the size-tracking instead of us?  That would clean up this code and allow us to potentially completely encapsulate size-tracking to an implementation detail of RemoteDocumentCache/RemoteDocumentChangeBuffer (by making RemoteDocumentChangeBuffer.removeEntry / addEntry private).

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -95,66 +96,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
   }
 
   /**
-   * Adds the supplied entries to the cache. Adds the given size delta to the cached size.
+   * Adds the supplied entries to the cache.
+   *
+   * Callees must ensure that the total document size in the cache is updated
+   * after all entries are written (via `updateMetadata()`).
```
mikelehen: Callees => Callers

Also, not sure if it's worth doing in this PR, but I think if we made IndexedDbRemoteDocumentChangeBuffer into a ["nested class"](https://stackoverflow.com/a/32494175) of IndexedDbRemoteDocumentCache, then we could make addEntry() [and potentially removeEntry(); see my comment on removeOrphanedDocuments] into private methods only called by applyChanges().  Then the size accounting is very well encapsulated in RemoteDocumentCache/RemoteDocumentChangeBuffer with callers being oblivious, and that seems kinda' great.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -95,66 +96,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+    });
   }
 
   /**
-   * Removes a document from the cache. Note that this method does *not* do any
-   * size accounting. It is the responsibility of the caller to count the bytes removed
-   * and issue a final updateSize() call after removing documents.
+   * Removes a document from the cache.
+   *
+   * Callees must ensure that the total document size in the cache is updated
```
mikelehen: Callees => Callers  (if we keep this public anyway)

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -95,66 +96,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
-      } else {
-        return PersistencePromise.resolve(0);
-      }
+    return store.delete(key);
+  }
+
+  /**
+   * Udpates the document change log and adds the given delta to the cached current size.
+   * Callers to `addEntry()` and `removeEntry()` *must* call this afterwards to update the
+   * cache's metadata.
```
mikelehen: Again, we could potentially make this private and not worry about callers messing up.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1261,9 +1268,11 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
     return iteration
       .next(() => PersistencePromise.waitFor(promises))
       .next(() =>
-        this.db.getRemoteDocumentCache().updateSize(txn, -bytesRemoved)
+        this.db
+          .getRemoteDocumentCache()
+          .updateMetadata(txn, changedKeys, -bytesRemoved)
```
mikelehen: So it seems like before we weren't properly propagating changedKeys to the documentChangesStore, but now we are?  I assume that was essentially a bug?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -277,9 +277,8 @@ export class MemoryEagerDelegate implements ReferenceDelegate {
             // Since this is the eager delegate and memory persistence,
             // we don't care about the size of documents. We don't track
             // the size of the cache for eager GC.
-            return cache.removeEntry(txn, key).next(() => {});
+            cache.removeEntry(key);
```
mikelehen: Consider using RemoteDocumentChangeBuffer here too so that we can make removeEntry private...

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -235,34 +194,29 @@ export class MemoryRemoteDocumentChangeBuffer extends RemoteDocumentChangeBuffer
...
-        `Attempting to change document ${key.toString()} without having read it first`
-      );
-      const size = this.sizer(maybeDocument);
-      delta += size - previousSize!;
-      docs.push({ maybeDocument, size });
+    const promises: Array<PersistencePromise<void>> = [];
+    this.changes.forEach((key, doc) => {
+      if (doc) {
+        const size = this.sizer(doc);
+        promises.push(this.documentCache.addEntry(transaction, doc, size));
```
mikelehen: Hrm...  Maybe not a big deal, but it seems like you could move `this.sizer(doc)` into the implementation of `addEntry()` and then IndexedDbRemoteDocumentCache.addEntry() and MemoryRemoteDocumentCache.addEntry() would have similar signature which seems appealing.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -89,22 +99,12 @@ export abstract class RemoteDocumentChangeBuffer {
     transaction: PersistenceTransaction,
     documentKey: DocumentKey
   ): PersistencePromise<MaybeDocument | null> {
-    const changes = this.assertChanges();
-
-    const bufferedEntry = changes.get(documentKey);
-    if (bufferedEntry) {
+    this.assertNotApplied();
+    const bufferedEntry = this.changes!.get(documentKey);
```
mikelehen: Unnecessary `!`?

packages/firestore/src/local/memory_persistence.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I reworded this to point users to the new API.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: It might not have been worth it, but it is done. It solves some of the data encapsulation issues in this class hierarchy.

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: Done. It keeps the size tracking even more encapsulated, which is something I am in favor of.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -46,7 +44,10 @@ import { PersistencePromise } from './persistence_promise';
  * porting this class as part of that implementation work.
  */
 export abstract class RemoteDocumentChangeBuffer {
-  private changes: MaybeDocumentMap | null = maybeDocumentMap();
+  protected changes: ObjectMap<
+    DocumentKey,
+    MaybeDocument | null
```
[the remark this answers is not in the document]
schmidt-sebastian: Done.

PR #2081, opened 2019-08-13
Port waitForPendingWrites from android

wu-hui:
* Port waitForPendingWrites

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -482,7 +482,7 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
    * Waits until all currently pending writes for the active user have been acknowledged by the
    * backend.
    *
-   * The returned Promise completes immediately if there are no outstanding writes. Otherwise, the
+   * The returned Promise resolves immediately if there are no outstanding writes. Otherwise, the
    * Task waits for all previously issued writes (including those written in a previous app
    * session), but it does not wait for writes that were added after the method is called. If you
    * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
@@ -493,6 +493,7 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
    * acknowledged by the backend.
    */
   _waitForPendingWrites(): Promise<void> {
+    this.ensureClientConfigured();
     return this._firestoreClient!.waitForPendingWrites();
   }
 
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -483,13 +483,13 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
    * backend.
    *
    * The returned Promise resolves immediately if there are no outstanding writes. Otherwise, the
-   * Task waits for all previously issued writes (including those written in a previous app
+   * Promise waits for all previously issued writes (including those written in a previous app
    * session), but it does not wait for writes that were added after the method is called. If you
    * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
    *
    * Any outstanding `waitForPendingWrites()` Promises are rejected during user changes.
    *
-   * @return A `Promise` which resolves when all currently pending writes have been
+   * @return A Promise which resolves when all currently pending writes have been
    * acknowledged by the backend.
    */
   _waitForPendingWrites(): Promise<void> {
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -478,6 +478,24 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return this._firestoreClient!.clientShutdown;
   }
 
+  /**
+   * Waits until all currently pending writes for the active user have been acknowledged by the
+   * backend.
+   *
+   * The returned Promise completes immediately if there are no outstanding writes. Otherwise, the
+   * Task waits for all previously issued writes (including those written in a previous app
+   * session), but it does not wait for writes that were added after the method is called. If you
+   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
+   *
+   * Any outstanding `waitForPendingWrites()` Promises are rejected during user changes.
+   *
+   * @return A `Promise` which resolves when all currently pending writes have been
+   * acknowledged by the backend.
+   */
+  _waitForPendingWrites(): Promise<void> {
+    return this._firestoreClient!.waitForPendingWrites();
+  }
+
   ensureClientConfigured(): FirestoreClient {
     if (!this._firestoreClient) {
       // Kick off starting the client but don't actually wait for it.
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -522,6 +522,20 @@ export class FirestoreClient {
     });
   }
 
+  /**
+   * Returns a `Promise` resolves when all the pending writes at the time when this method is called
+   * received server acknowledgement. An acknowledgement can be either acceptance or rejections.
+   */
+  waitForPendingWrites(): Promise<void> {
+    this.verifyNotShutdown();
+
+    const deferred = new Deferred<void>();
+    this.asyncQueue.enqueueAndForget(() => {
+      return this.syncEngine.registerPendingWritesCallback(deferred);
+    });
+    return deferred.promise;
+  }
+
   listen(
     query: Query,
     observer: Observer<ViewSnapshot>,
@@ -536,7 +550,6 @@ export class FirestoreClient {
   }
 
   unlisten(listener: QueryListener): void {
-    this.verifyNotShutdown();
     this.asyncQueue.enqueueAndForget(() => {
       return this.eventMgr.unlisten(listener);
     });
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -523,8 +523,9 @@ export class FirestoreClient {
   }
 
   /**
-   * Returns a `Promise` resolves when all the pending writes at the time when this method is called
-   * received server acknowledgement. An acknowledgement can be either acceptance or rejections.
+   * Returns a Promise that resolves when when all writes that were pending at the time when this
+   * method is called received server acknowledgement. An acknowledgement can be either acceptance
+   * or rejections.
    */
   waitForPendingWrites(): Promise<void> {
     this.verifyNotShutdown();
@@ -550,6 +551,7 @@ export class FirestoreClient {
   }
 
   unlisten(listener: QueryListener): void {
+    this.verifyNotShutdown();
     this.asyncQueue.enqueueAndForget(() => {
       return this.eventMgr.unlisten(listener);
     });
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -523,9 +523,9 @@ export class FirestoreClient {
   }
 
   /**
-   * Returns a Promise that resolves when when all writes that were pending at the time when this
-   * method is called received server acknowledgement. An acknowledgement can be either acceptance
-   * or rejections.
+   * Returns a Promise that resolves when all writes that were pending at the time this
+   * method was called received server acknowledgement. An acknowledgement can be either acceptance
+   * or rejection.
    */
   waitForPendingWrites(): Promise<void> {
     this.verifyNotShutdown();
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -28,12 +28,12 @@ import {
 import { MaybeDocument, NoDocument } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { Mutation } from '../model/mutation';
-import { MutationBatchResult } from '../model/mutation_batch';
+import { MutationBatchResult, BATCHID_UNKNOWN } from '../model/mutation_batch';
 import { RemoteEvent, TargetChange } from '../remote/remote_event';
 import { RemoteStore } from '../remote/remote_store';
 import { RemoteSyncer } from '../remote/remote_syncer';
 import { assert, fail } from '../util/assert';
-import { FirestoreError } from '../util/error';
+import { Code, FirestoreError } from '../util/error';
 import * as log from '../util/log';
 import { primitiveComparator } from '../util/misc';
 import { ObjectMap } from '../util/obj_map';
@@ -160,6 +160,8 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
   private mutationUserCallbacks = {} as {
     [uidKey: string]: SortedMap<BatchId, Deferred<void>>;
   };
+  /** Stores user callbacks waiting for all pending writes to be acknowledged. */
+  private pendingWritesCallbacks = new Map<BatchId, Array<Deferred<void>>>();
   private limboTargetIdGenerator = TargetIdGenerator.forSyncEngine();
 
   // The primary state is set to `true` or `false` immediately after Firestore
@@ -450,6 +452,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       (this.isPrimary && source === OnlineStateSource.RemoteStore) ||
       (!this.isPrimary && source === OnlineStateSource.SharedClientState)
     ) {
+      this.assertSubscribed('applyOnlineStateChange()');
       const newViewSnapshots = [] as ViewSnapshot[];
       this.queryViewsByQuery.forEach((query, queryView) => {
         const viewChange = queryView.view.applyOnlineStateChange(onlineState);
@@ -570,6 +573,8 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     // before listen events.
     this.processUserCallback(batchId, /*error=*/ null);
 
+    this.triggerPendingWritesCallbacks(batchId);
+
     try {
       const changes = await this.localStore.acknowledgeBatch(
         mutationBatchResult
@@ -593,6 +598,8 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     // listen events.
     this.processUserCallback(batchId, error);
 
+    this.triggerPendingWritesCallbacks(batchId);
+
     try {
       const changes = await this.localStore.rejectBatch(batchId);
       this.sharedClientState.updateMutationState(batchId, 'rejected', error);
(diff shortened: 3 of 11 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -478,6 +478,24 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return this._firestoreClient!.clientShutdown;
   }
 
+  /**
+   * Waits until all currently pending writes for the active user have been acknowledged by the
+   * backend.
+   *
+   * The returned Promise completes immediately if there are no outstanding writes. Otherwise, the
```
schmidt-sebastian: s/completes/resolves/

packages/firestore/src/api/database.ts
```diff
@@ -478,6 +478,24 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+   * session), but it does not wait for writes that were added after the method is called. If you
+   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
+   *
+   * Any outstanding `waitForPendingWrites()` Promises are rejected during user changes.
+   *
+   * @return A `Promise` which resolves when all currently pending writes have been
+   * acknowledged by the backend.
+   */
+  _waitForPendingWrites(): Promise<void> {
+    return this._firestoreClient!.waitForPendingWrites();
```
schmidt-sebastian: Please call `ensureClientConfigured()` to make sure this works even when it is the first operation that is scheduled.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -522,6 +522,20 @@ export class FirestoreClient {
     });
   }
 
+  /**
+   * Returns a `Promise` resolves when all the pending writes at the time when this method is called
```
schmidt-sebastian: I would remove the backticks for Promise. It's just a normal term in JS, not a class or method name (also it should be "Promise that resolves");

Also: s/when all the pending writes at the time/when all writes that were pending at the time/

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     }
   }
 
+  /**
+   * Takes a snapshot of current mutation queue, and register a user callback
```
schmidt-sebastian: s/register/registers

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+   * which will be called when all those mutations in the snapshot are either
```
schmidt-sebastian: What do you mean by "snapshot" here? Snapshot usually means something different in our code base.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+  async registerPendingWritesCallback(callback: Deferred<void>): Promise<void> {
+    if (!this.remoteStore.canUseNetwork()) {
+      log.debug(
+        LOG_TAG,
+        'The network is disabled. The task returned by ' +
+          "'awaitPendingWrites()' will not complete until the network is enabled."
+      );
+    }
+
+    const largestBatchId = await this.localStore.getHighestUnacknowledgedBatchId();
```
schmidt-sebastian: Nit: Make this `highestBatchId`.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     }
   }
 
+  /**
+   * Takes a snapshot of current mutation queue, and register a user callback
+   * which will be called when all those mutations in the snapshot are either
+   * accepted or rejected by the server.
+   */
+  async registerPendingWritesCallback(callback: Deferred<void>): Promise<void> {
```
schmidt-sebastian: This method doesn't seem to have to be async. You can just return void.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+
+    const largestBatchId = await this.localStore.getHighestUnacknowledgedBatchId();
+
+    if (largestBatchId === BATCHID_UNKNOWN) {
+      // Trigger the callback right away if there is no pending writes at the moment.
+      callback.resolve();
+      return Promise.resolve();
+    }
+
+    const callbacks = this.pendingWritesCallbacks.get(largestBatchId) || [];
```
schmidt-sebastian: It would be beneficial to verify what the transpiled JS looks like, but this could potentially be:

```
const callbacks = [... this.pendingWritesCallbacks.get(largestBatchId), callback]
```

If we don't transpile the spread operator away, there is some chance we are breaking users, so this might be not worth exploring.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+    callbacks.push(callback);
+    this.pendingWritesCallbacks.set(largestBatchId, callbacks);
+    return Promise.resolve();
+  }
+
+  /**
+   * Triggers callbacks waiting for this batch id to get acknowledged by server,
```
schmidt-sebastian: s/callbacks/the callbacks that are/

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+   * if there are any.
+   */
+  private triggerPendingWritesCallbacks(batchId: BatchId): void {
+    (this.pendingWritesCallbacks.get(batchId) || []).forEach(callback => {
```
schmidt-sebastian: If you use a `for ... of` loop then you don't need to create an empty array here.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,62 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+  }
+
+  /** Reject all outstanding callbacks waiting for pending writes to complete. */
+  private rejectOutstandingPendingWritesCallbacks(): void {
+    this.pendingWritesCallbacks.forEach(callbacks => {
+      callbacks.forEach(callback => {
+        callback.reject(
+          new FirestoreError(
+            Code.CANCELLED,
+            "'waitForPendingWrites' task is cancelled due to User change."
```
schmidt-sebastian: This might be applicable to Android to: Since this is a very generalized function it might make sense to pass in the error (at least the message if you want to keep the stacktrace).

Also nit: s/User/a user/.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -265,6 +265,26 @@ export class IndexedDbMutationQueue implements MutationQueue {
       .next(() => foundBatch);
   }
 
+  getHighestUnacknowledgedBatchId(
+    transaction: PersistenceTransaction
+  ): PersistencePromise<BatchId> {
+    const range = IDBKeyRange.bound(
+      [this.userId, Number.NEGATIVE_INFINITY],
```
schmidt-sebastian: While this is fine, the method below uses `BATCHID_UNKNOWN` which also works.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -265,6 +265,26 @@ export class IndexedDbMutationQueue implements MutationQueue {
       .next(() => foundBatch);
   }
 
+  getHighestUnacknowledgedBatchId(
+    transaction: PersistenceTransaction
+  ): PersistencePromise<BatchId> {
+    const range = IDBKeyRange.bound(
```
schmidt-sebastian: Is it possible to just use `upperBound`?

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1098,6 +1101,21 @@ apiDescribe('Database', (persistence: boolean) => {
     });
   });
 
+  it('can unlisten queries after shutdown', async () => {
```
schmidt-sebastian: Please move these changes to a different PR. We don't want to revert two PRs worth of work if we happen to find a problem with one of them.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1138,50 @@ apiDescribe('Database', (persistence: boolean) => {
       }).to.throw();
     });
   });
+
+  it('can wait for pending writes as expected', async () => {
```
schmidt-sebastian: "as expected" should be redundant for our tests :)

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1138,50 @@ apiDescribe('Database', (persistence: boolean) => {
...
+      // Prevent pending writes receiving acknowledgement.
+      await firestore.disableNetwork();
+
+      const awaitPendingWrites1 = waitForPendingWrites(firestore);
+      const pendingWrites = docRef.set({ foo: 'bar' });
+      const awaitPendingWrites2 = waitForPendingWrites(firestore);
+
+      // `awaitsPendingWrites1` resolves immediately because there are no pending writes at
+      // the time it is created.
+      await expect(awaitPendingWrites1).to.be.eventually.fulfilled;
```
schmidt-sebastian: Hm. I don't know how you can check in JS if a Promise is already resolved. If you want to keep this line (I think you should) I would adjust the comment to explain why the Promise does resolve even though the network is offline.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1138,50 @@ apiDescribe('Database', (persistence: boolean) => {
...
+
+      // pending writes can receive acknowledgements now.
+      await firestore.enableNetwork();
+      await expect(pendingWrites).to.be.eventually.fulfilled;
```
schmidt-sebastian: `await expect(pendingWrites).to.be.eventually.fulfilled` is the same as `await pendingWrites`.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1138,50 @@ apiDescribe('Database', (persistence: boolean) => {
...
+  it('waiting for pending writes should fail when user changes', async () => {
+    await withMockCredentialProviderTestDb(persistence, async (db, mockCredentialsProvider) => {
+      // Prevent pending writes receiving acknowledgement.
+      await db.disableNetwork();
+      db.doc('abc/123').set({ foo: 'bar' });
+      const awaitPendingWrite = waitForPendingWrites(db);
+
+      mockCredentialsProvider.changeUserTo(new User('user_1'));
+
+      await expect(awaitPendingWrite).to.be.eventually.rejected;
```
schmidt-sebastian: Please do test the error message as well. It will also make the test much clearer.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1138,50 @@ apiDescribe('Database', (persistence: boolean) => {
...
+
+  it('waiting for pending writes resolves immediately when offline and no pending writes',
+    async () => {
+      await withTestDoc(persistence, async docRef => {
+        const firestore = docRef.firestore;
+        // Prevent pending writes receiving acknowledgement.
+        await firestore.disableNetwork();
+
+        const awaitPendingWrites = waitForPendingWrites(firestore);
+        await expect(awaitPendingWrites).to.be.eventually.fulfilled;
```
schmidt-sebastian: This is the same test case as above. In general, it is usually better to have smaller and more focused test cases. Maybe we should remove the "offline but no writes" case from above and only test it here?

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -124,6 +126,20 @@ export const apiDescribe = apiDescribeInternal.bind(
 apiDescribe.skip = apiDescribeInternal.bind(null, describe.skip);
 apiDescribe.only = apiDescribeInternal.bind(null, describe.only);
 
+export class MockCredentialsProvider extends EmptyCredentialsProvider {
+
+  private listener: CredentialChangeListener | null = null;
+
+  changeUserTo(user: User): void {
```
schmidt-sebastian: Do you think it makes sense to call this `handleUserChange` to match SyncEngine?

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -163,6 +179,20 @@ export function withTestDb(
   });
 }
 
+export function withMockCredentialProviderTestDb(
+  persistence: boolean,
+  fn: (db: firestore.FirebaseFirestore, mockCredential: MockCredentialsProvider) => Promise<void>
+): Promise<void> {
+  const mockCredentialsProvider = new MockCredentialsProvider();
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const settings: any = { ...DEFAULT_SETTINGS };
```
schmidt-sebastian: You can potentially remove the `any`:

```
const settings = {... DEFAULT_SETTINGS, credentials: { client: mockCredentialsProvider, type: 'provider' }}
```

packages/firestore/test/integration/api/database.test.ts
wu-hui: Done.

packages/firestore/test/integration/api/database.test.ts
wu-hui: Done.

packages/firestore/test/integration/api/database.test.ts
wu-hui: Done.

packages/firestore/test/integration/util/helpers.ts
wu-hui: `handleUserChange` implies it's handling something, which is not true. Changed to `triggerUserChange` instead.

packages/firestore/test/integration/util/helpers.ts
wu-hui: Done.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Please follow the pattern that we use here: https://github.com/firebase/firebase-js-sdk/pull/2063/files#diff-3bab365c04d7fc7ae24002035afdd01fR614

It makes the interactions in the client simpler if we limit async methods to their smallest possible scope. Right now, registerPendingWritesCallback only has one consumer, but once it has two or more this would really make a difference (and even with just one consumer, you can already remove two Promise.resolve() invocations that the transpiler adds for you.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: You can invoke `yarn build`  in "packages/firestore". It will spit out a couple different files into "packages/firestore/dist"

In packages/firestore/dist/index.esm.js, your current code looks as such:

```
callbacks = this.pendingWritesCallbacks.get(highestBatchId) || [];
callbacks.push(callback);
this.pendingWritesCallbacks.set(highestBatchId, callbacks);
```

It looks like we have to specify different compiler settings to make my suggestion work (--downlevelIterators)/ Since it is just an optional suggestion, let's keep it as is.

packages/firestore/src/api/database.ts
```diff
@@ -478,6 +478,25 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return this._firestoreClient!.clientShutdown;
   }
 
+  /**
+   * Waits until all currently pending writes for the active user have been acknowledged by the
+   * backend.
+   *
+   * The returned Promise resolves immediately if there are no outstanding writes. Otherwise, the
+   * Task waits for all previously issued writes (including those written in a previous app
```
schmidt-sebastian: s/Task/Promise/

packages/firestore/src/api/database.ts
```diff
@@ -478,6 +478,25 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+   * session), but it does not wait for writes that were added after the method is called. If you
+   * wish to wait for additional writes, you have to call `waitForPendingWrites()` again.
+   *
+   * Any outstanding `waitForPendingWrites()` Promises are rejected during user changes.
+   *
+   * @return A `Promise` which resolves when all currently pending writes have been
```
schmidt-sebastian: Nit: remove backticks.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -522,6 +522,21 @@ export class FirestoreClient {
     });
   }
 
+  /**
+   * Returns a Promise that resolves when when all writes that were pending at the time when this
```
schmidt-sebastian: s/when when/when/
s/at the time when/at the time/

packages/firestore/src/core/firestore_client.ts
```diff
@@ -522,6 +522,21 @@ export class FirestoreClient {
...
+   * method is called received server acknowledgement. An acknowledgement can be either acceptance
```
schmidt-sebastian: s/is/was/

packages/firestore/src/core/firestore_client.ts
```diff
@@ -522,6 +522,21 @@ export class FirestoreClient {
...
+   * or rejections.
```
schmidt-sebastian: s/rejections/rejection/

packages/firestore/src/core/sync_engine.ts
```diff
@@ -450,6 +452,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       (this.isPrimary && source === OnlineStateSource.RemoteStore) ||
       (!this.isPrimary && source === OnlineStateSource.SharedClientState)
     ) {
+      this.assertSubscribed('applyOnlineStateChange()');
```
schmidt-sebastian: Is this intended?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -602,6 +609,59 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     }
   }
 
+  /**
+   * Registers a user callback with all the pending mutations at the moment of calling.
```
schmidt-sebastian: Can you update this to:

Registers a user callback that resolves when all pending mutations at the moment of calling are acknowledged .

You can then remove the second sentence.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I learned something today :)

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1123,49 @@ apiDescribe('Database', (persistence: boolean) => {
...
+  });
+
+  it('waiting for pending writes resolves immediately when offline and no pending writes',
+    async () => {
+      await withTestDoc(persistence, async docRef => {
+        const firestore = docRef.firestore;
+        // Prevent pending writes receiving acknowledgement.
+        await firestore.disableNetwork();
+
+      // `awaitsPendingWrites` is created when there is no pending writes, it will resolve
```
schmidt-sebastian: This looks like it needs reformatting.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1120,4 +1123,49 @@ apiDescribe('Database', (persistence: boolean) => {
...
+      // immediately even if we are offline.
+        const awaitPendingWrites = waitForPendingWrites(firestore);
```
schmidt-sebastian: Suggest to remove the redundant variable here.

schmidt-sebastian:
> I do want you to make `registerPendingWritesCallback` non-async though :)

Looks like this is not possible as `getHighestUnacknowledgedBatchId` returns a Promise. Thanks for pointing this out.

packages/firestore/src/api/database.ts

packages/firestore/src/api/database.ts

packages/firestore/src/core/firestore_client.ts

packages/firestore/src/core/firestore_client.ts

packages/firestore/src/core/firestore_client.ts

packages/firestore/src/core/sync_engine.ts
wu-hui: Yes. This was missing but should be present. Otherwise `this.syncEngineListener!.*` might hit errors.

packages/firestore/src/core/sync_engine.ts
wu-hui: Done.

packages/firestore/src/core/sync_engine.ts
wu-hui: `this.localStore.getHighestUnacknowledgedBatchId()` returns a promise, which demands it to be async.

packages/firestore/src/core/sync_engine.ts
wu-hui: :)

packages/firestore/test/integration/api/database.test.ts
wu-hui: Done.

packages/firestore/test/integration/api/database.test.ts
wu-hui: Done.

PR #2082, opened 2019-08-13
remove @firebase/app and @firebase/app-types from peerDependencies

Feiyang1:
This is to solve the unmet `peerDependency` warning when `@firebase/database` is used in Admin SDK. https://github.com/firebase/firebase-admin-node/issues/614

`@firebase/app` was intentionally removed from Admin SDK's dependency to solve a version conflict issue when it's used together with JS SDK. https://github.com/firebase/firebase-js-sdk/pull/1916

Though this PR is not the most correct way to solve the issue, it is quick and easy with little impact on user experience. The only user experience change is that user won't get unmet peer dependency warnings if `@firebase/database` is installed without installing `@firebase/app`.
It should be a rare case as we ask user to install `firebase` package in all our guides, which automatically includes `@firebase/app`.

The alternative solution is create a separate npm package for admin sdk consumption. It feels like an overkill and adds a lot of maintenance burden on us for little benefit.


packages/database-types/package.json
```diff
@@ -11,7 +11,6 @@
     "index.d.ts"
   ],
   "peerDependencies": {
```
hiranya911: Should this become a regular dependency instead?

packages/database/package.json
```diff
@@ -22,8 +22,6 @@
   },
   "license": "Apache-2.0",
   "peerDependencies": {
```
hiranya911: Should these become dev dependencies?

packages/database-types/package.json
Feiyang1: Do you mean making `@firebase/app` a regular dependency instead of removing it? No, that would install `@firebase/app` in Admin SDK causing version conflict potentially.

Or do you just mean renaming it to `dependencies` ?

hiranya911: Not if we database-types a dev dependency of database.

Feiyang1: hmm, good catch. I think we probably should make app-types a dependency of database-types, or tsc will complain app-types is not found.

Feiyang1: hmm, good catch. I think we probably should do it, or tsc may complain app-types is not found.

Feiyang1: Or we can keep it as a peerDependency here and make app-types a dependency in admin sdk. WDTY?

packages/database/package.json
Feiyang1: `@firebase/app` can't be made a dependency because then it will be installed with admin sdk, which can cause potential version conflict if `firebase` is installed as well.

packages/database-types/package.json
Feiyang1: Made `app-types` a regular dependency, so we don't entirely lose the dependency.

packages/database/package.json
Feiyang1: Just realized you said dev dependencies - I don't feel strongly about it, but I think it's nice the dependencies are preserved this way (though not 100% correctly).

Feiyang1:
@schmidt-sebastian WDYT

PR #2083, opened 2019-08-14
Allow user to unlisten to queries after shutdown.

wu-hui:


packages/firestore/src/core/firestore_client.ts
```diff
@@ -536,7 +536,6 @@ export class FirestoreClient {
   }
 
   unlisten(listener: QueryListener): void {
-    this.verifyNotShutdown();
     this.asyncQueue.enqueueAndForget(() => {
```
schmidt-sebastian: This potentially calls into the SyncEngine. I wonder if we should guard this on the shutdown state and not call `unlisten` if we are shut down?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -536,7 +536,10 @@ export class FirestoreClient {
   }
 
   unlisten(listener: QueryListener): void {
-    this.verifyNotShutdown();
+    // Make unlisten a no-op if client is shut down.
```
schmidt-sebastian: Your code is so clear that I think we can remove this comment :)

packages/firestore/src/core/firestore_client.ts
```diff
@@ -536,7 +536,10 @@ export class FirestoreClient {
...
+    if(this.clientShutdown) {
```
schmidt-sebastian: There is something going on with your formatter. This should be `if (this.clientShutdown)`

I would also slightly prefer if you wrote the inverse and removed the early return. Optional.

packages/firestore/src/core/firestore_client.ts
wu-hui: I actually want to keep it. The reason is `FirestoreClient` methods generally run `verifyNotShutdown` before doing anything. This one deviate from that. 

I updated the comment to be more clear.

packages/firestore/src/core/firestore_client.ts
wu-hui: Yeah, it's strange, the automated prettier commit did not make into the push somehow..

PR #2085, opened 2019-08-15
Do not throw error on duplicate service registration.

hsubox76:
A change needed for when multiple components each internally depend on another component and may each try to register it.

### DIFF
--- a/packages/app/src/firebaseNamespaceCore.ts
+++ b/packages/app/src/firebaseNamespaceCore.ts
@@ -182,7 +182,10 @@ export function createFirebaseNamespaceCore(
   ): FirebaseServiceNamespace<FirebaseService> {
     // Cannot re-register a service that already exists
     if (factories[name]) {
-      throw ERROR_FACTORY.create(AppError.DUPLICATE_SERVICE, { appName: name });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return (namespace as any)[name] as FirebaseServiceNamespace<
+        FirebaseService
+      >;
     }
 
     // Capture the service factory for later service instantiation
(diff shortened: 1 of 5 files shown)


packages/app/src/firebaseNamespaceCore.ts
```diff
@@ -182,7 +182,10 @@ export function createFirebaseNamespaceCore(
   ): FirebaseServiceNamespace<FirebaseService> {
     // Cannot re-register a service that already exists
```
Feiyang1: The comment is outdated and needs update.

Feiyang1: I think we should log a message at DEBUG log level for catching the possible version mixup. We already use logger in `@firebase/app`, we can make it sharable and use it here.

packages/app/test/firebaseApp.test.ts
```diff
@@ -87,6 +87,32 @@ function executeFirebaseTests(): void {
...
+
+      (firebase as _FirebaseNamespace).INTERNAL.registerService(
+        'test',
+        (app: FirebaseApp) => {
+          registrations += 1;
+          return new TestService(app);
+        }
+      );
+
+      (firebase as any).test();
```
Feiyang1: We should just test no exception is thrown the second time and that `firebase.test === the previous firebase.test`

packages/app/src/firebaseNamespaceCore.ts

packages/app/test/firebaseApp.test.ts
hsubox76: Added equality test.  To check for no exception, do we have to do anything?  Or just leave it alone and throwing will fail the test?

hsubox76: Updated test to check for exception and service namespace equality.

lookfirst:
THANK YOU!

PR #2099, opened 2019-08-20
Don't persist documents that we already have

schmidt-sebastian:
This PR changes the handling of remote documents to skip any document updates for document versions that we already have cached. To keep the code readable, I removed the concept of `authoritativeUpdates` (as discussed in the API discussion phase of Sprint planning).

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -459,43 +459,46 @@ export class LocalStore {
           remoteEvent.targetChanges,
           (targetId: TargetId, change: TargetChange) => {
             const oldQueryData = this.queryDataByTarget[targetId];
-            if (oldQueryData) {
-              // Only update the remote keys if the query is still active. This
-              // ensures that we can persist the updated query data along with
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -580,22 +558,21 @@ export class LocalStore {
         // can synthesize remote events when we get permission denied errors while
         // trying to resolve the state of a locally cached document that is in
         // limbo.
-        const remoteVersion = remoteEvent.snapshotVersion;
-        if (!remoteVersion.isEqual(SnapshotVersion.MIN)) {
+        if (!snapshotVersion.isEqual(SnapshotVersion.MIN)) {
```
var-const: Question: why doesn't this change apply to the Android port?

schmidt-sebastian: This is just a rename to which allows me to use an object shorthand notation (instead of {snapshotVersion:remoteVersion} I can now use {snapshotVersion}.

PR #2100, opened 2019-08-20
Adding onSnapshotsInSync (not public)

thebrianchen:
First pass at plumbing through `onSnapshotsInSync`. 
- I wasn't sure how to add comments to the changed `onViewSnapshots` and `applyOnlineStateChanges` mainly because I'm not sure how they do. I'd be happy to add comments to the individual methods in EventManager since it's pretty hard to follow along with just the code.
- Please fill me in on any other test cases I should also include.
- Inside EventManager, I'm only calling the listener once per call to `applyOnlineStateChanges` and `onViewSnapshots`. Otherwise, the inSync callback would fire for each query listener.


packages/firestore/test/integration/api/database.test.ts
```diff
@@ -542,6 +543,364 @@ apiDescribe('Database', (persistence: boolean) => {
     });
   });
 
+  apiDescribe('onSnapshotsInSync listener', () => {
```
mikelehen: So I think Gil is write that these could be written as spec tests.  But this will require some plumbing work... basically the ability for spec tests to add / remove snapshots-in-sync listeners, and the ability for tests to express when the event is expected to be raised.  As a rough outline (probably not 100% right and all the details are TBD) I expect that you'll need to:

1. Introduce a new SpecBuilder method that we can call from spec tests similar to the existing userListen one: https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/test/unit/specs/spec_builder.ts#L217
2. Introduce a new `SpecStep` type: https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/src/core/event_manager.ts#L214
3. Make the `SpecTestRunner` code handle the new SpecStep: https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/test/unit/specs/spec_builder.ts#L217
4. Need to add a way for tests to express expectations regarding when the event fires, similar to expectEvents(): https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/test/unit/specs/spec_builder.ts#L722
5. Plumb the expected events into the `SpecStep` somehow.  This may mean adding a new field to `SpecStep` that's similar to `expect` (which really should be called `expectedEvents`) or else somehow extending the existing `SpecExpectation` so that it can cover both snapshot events and snapshots-in-sync events.  https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/test/unit/specs/spec_test_runner.ts#L1406
6. Update our step runner code to validate the new expectations as appropriate. https://github.com/firebase/firebase-js-sdk/blob/beb72676a75dabf5e0f965f5ee06208b60c9ea45/packages/firestore/test/unit/specs/spec_test_runner.ts#L520

The benefit of all this is that you'll be able to write the tests once and then run them on all platforms (once you make the appropriate SpecTestRunner changes on the other platforms). They'll also be faster since they'll run as synchronous unit tests instead of as slow integration tests that are network-dependent.  And they'll hopefully end up more concise / more readable (less boilerplate), etc.  It's also a great opportunity for you to ramp up more deeply on spec tests! :-)

I would probably keep 1 or 2 integration tests that just smoke test the public API, but we can hopefully use spec tests to test all the edge cases.

packages/firestore/src/core/event_manager.ts
```diff
@@ -105,16 +111,23 @@ export class EventManager implements SyncEngineListener {
   }
 
   onWatchChange(viewSnaps: ViewSnapshot[]): void {
+    let shouldRunInSyncListeners = false;
```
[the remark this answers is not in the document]
thebrianchen: What do you feel about `didRaiseEvent`? To me, `raisedEvent` seems like the actual event that was raised, rather than a boolean value.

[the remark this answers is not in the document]
mikelehen: I'm okay with that...  But should `raisedInitialEvent` be renamed to `haveRaisedInitialEvent` or something to follow suit? (Note that I don't think `didRaiseInitialEvent` would make sense)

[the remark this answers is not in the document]
thebrianchen: Ah, I didn't notice the `raisedInitialEvent` precedence. In that case, `raisedEvent` makes much more sense.

packages/firestore/src/core/event_manager.ts
```diff
@@ -132,11 +145,37 @@ export class EventManager implements SyncEngineListener {
...
+    if (shouldRunInSyncListeners) {
+      this.runSnapshotsInSyncListeners();
+    }
+  }
+
+  hasQueries(): boolean {
+    return !this.queries.isEmpty();
+  }
+
+  addSnapshotsInSyncListener(asyncObserver: AsyncObserver<void>): void {
```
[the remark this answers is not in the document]
thebrianchen: Ooooh that makes a lot more sense. Thanks for the detailed explanation with examples!

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1236,4 +1236,164 @@ describeSpec('Listens:', [], () => {
...
+        .expectEvents(query2, {
+          metadata: [docAv2]
+        })
+        .expectSnapshotsInSyncEvent();
+    }
+  );
+
+  specTest(
+    'onSnapshotsInSync fires for multiple listeners',
+    ['exclusive'],
```
mikelehen: This mean the rest of the tests aren't being run. :-)

thebrianchen: ahhhhhh sorry for all the careless errors I've been leaving around in this PR 😢

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -541,59 +541,29 @@ apiDescribe('Database', (persistence: boolean) => {
...
+        events.push('snapshots-in-sync');
+        done.resolve();
       });
+      await done.promise;
+      done = new Deferred<void>();
+      events = [];
+
+      await doc.set({ foo: 3 });
+      await done.promise;
+      expect(events).to.deep.equal(['doc', 'snapshots-in-sync']);
```
mikelehen: While this test is probably deterministic in JavaScript since it is single-threaded, it will be racey on Android / iOS since there is a race between onSnapshot() getting an initial event from the server and it getting the event from your local set() call.  If the server event comes first, you could end up with ['doc', 'snapshots-in-sync', 'doc', 'snapshots-in-sync']

There are a variety of ways to make this non-racey.  One way would be to explicitly wait for the initial snapshot event.  It's not super elegant, but something like:

```
  it.only('onSnapshotsInSync fires after listeners are in sync', () => {
    const testDocs = {
      a: { foo: 1 }
    };
    return withTestCollection(persistence, testDocs, async coll => {
      let events: string[] = [];
      const doc = coll.doc('a');
      const gotInitialSnapshot = new Deferred<void>();
      doc.onSnapshot(snap => {
        events.push('doc');
        gotInitialSnapshot.resolve();
      });
      await (gotInitialSnapshot.promise);
      events = [];

      const done = new Deferred<void>();
      onSnapshotsInSync(doc.firestore, () => {
        events.push('snapshots-in-sync');
        if (events.length === 3) {
          // We should have an initial snapshots-in-sync event, then snapshot event for set(), then
          // another event to indicate we're in sync again.
          expect(events).to.deep.equal(['snapshots-in-sync', 'doc', 'snapshots-in-sync']);
          done.resolve();
        }
      });

      await doc.set({ foo: 3 });
      await done.promise;
    });
  });
```

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -862,8 +862,9 @@ export class SpecBuilder {
   expectSnapshotsInSyncEvent(count = 1): this {
     this.assertStep('Expectations require previous step');
     const currentStep = this.currentStep!;
-    currentStep.stateExpect = currentStep.stateExpect || {};
-    currentStep.stateExpect.numSnapshotsInSyncEvents = count;
+    if (!currentStep.expectedSnapshotsInSyncEvents) {
```
mikelehen: Why is this check needed?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1047,6 +1041,11 @@ abstract class TestRunner {
     }
   }
 
+  private validateSnapshotsInSyncEvents(count: number): void {
```
mikelehen: I think this should be `number | undefined`?  And then you wouldn't need the `!` on line 525?
For clarity maybe rename `count` to `expectedCount`

PR #2102, opened 2019-08-21
Move isShutdown to state expectations

schmidt-sebastian:
I would like to move the `isShutdown` to the Spec test state expectations. That should help me in my futile attempts to auto-generate spec tests.

### DIFF
--- a/packages/firestore/test/unit/local/persistence_test_helpers.ts
+++ b/packages/firestore/test/unit/local/persistence_test_helpers.ts
@@ -144,8 +144,8 @@ export async function testMemoryLruPersistence(
 }
 
 /** Clears the persistence in tests */
-export async function clearTestPersistence(): Promise<void> {
-  await IndexedDbPersistence.clearPersistence(TEST_PERSISTENCE_PREFIX);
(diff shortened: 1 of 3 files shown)


packages/firestore/test/unit/local/persistence_test_helpers.ts
```diff
@@ -144,8 +144,8 @@ export async function testMemoryLruPersistence(
 }
 
 /** Clears the persistence in tests */
-export async function clearTestPersistence(): Promise<void> {
-  await IndexedDbPersistence.clearPersistence(TEST_PERSISTENCE_PREFIX);
+export function clearTestPersistence(): Promise<void> {
+  return IndexedDbPersistence.clearPersistence(TEST_PERSISTENCE_PREFIX);
```
mikelehen: I feel like this is a quiz on how async/await works, but I'm stumped...  Does your change do anything?

schmidt-sebastian: No... but I find it weird to use async in a method that only uses a single Promise in its return type. I can turn it into a quiz though!

PR #2109, opened 2019-08-23
make waitForPendingWrites public

wu-hui:


wu-hui:
> Can you add a Changelog entry? Thanks!

I did add a CHANGELOG entry, now also fixed missing versions of previous entries.

PR #2111, opened 2019-08-23
Fixing formatting of CDN link

egilmorez:
Per internal b/138222317.

Will this format work?

Thanks!


hsubox76:
I think we ended up settling on linking users to the documentation page for how to include the CDNs, instead of directly to the CDN download.

PR #2125, opened 2019-08-29
Index-Free: Track readTime in the RemoteDocument store

schmidt-sebastian:
This a loosely based port of https://github.com/firebase/firebase-android-sdk/pull/615 and https://github.com/firebase/firebase-android-sdk/pull/686

It adds a read time index to the Remote Document Store that can be used for Index-Free queries.   I also wasn't able to find good documentation on what happens if I add an index to an existing ObjectStore, but I wrote a test that verifies that this works. 

A follow up PR uses a similar index for the Multi-Tab changelog.

PR #2128, opened 2019-08-30
Use Read-Time Index for Multi-Tab

schmidt-sebastian:
This uses the read time we persist for Index-Free queries in the change log for multi-tab. The big benefit is that we no longer need to GC the change log (since LRU GC does it for us) and we no longer need to recover from GCed change logs. The downside is that we now need to store NoDocuments for deleted rows, but these are internal to the IndexedDbRemoteDocumentStore (I added a new option to the RemoteDocumentChangeBuffer for that).

This is a follow-up to https://github.com/firebase/firebase-js-sdk/pull/2125


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1144,7 +1144,7 @@ describeSpec('Listens:', [], () => {
       .client(1)
       .stealPrimaryLease()
       .expectListen(query, 'resume-token-2000')
-      .watchAcksFull(query, 2000, docC)
+      .watchAcksFull(query, 3000, docC)
```
schmidt-sebastian: This changed since we can use the read time to get a change diff, rather than a separate counter.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -331,95 +295,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
-                new NoDocument(key, SnapshotVersion.forDeletedDoc());
-              changedDocs = changedDocs.insert(key, doc);
-            })
+    const documentsStore = remoteDocumentsStore(transaction);
+    const range = IDBKeyRange.lowerBound(lastReadTime, true);
+    return documentsStore
+      .iterate(
+        { index: DbRemoteDocument.readTimeIndex, range },
+        (_, dbRemoteDoc) => {
+          const doc = this.serializer.fromDbRemoteDocument(dbRemoteDoc);
```
mikelehen: Maybe add a comment that we explicitly don't use maybeDecodeDocument() because we want to keep sentinel deletes?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -486,15 +447,22 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
           promises.push(this.documentCache.addEntry(transaction, key, doc));
         } else {
           sizeDelta -= previousSize!;
-          promises.push(this.documentCache.removeEntry(transaction, key));
+          if (this.trackDeletes) {
+            // A sentinel delete is symbolized by a NoDocument with a version of 0.
```
mikelehen: Maybe expand on this comment?

> In order to track deletes, we store "sentinel delete" entries in the Remote Document Cache. These entries are represented by a NoDocument with a version of 0 and are ignored by maybeDecodeDocument() but preserved in getNewDocumentChanges().

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -331,95 +295,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+        }
         control.done();
       }
     );
   }
 
-  newChangeBuffer(): RemoteDocumentChangeBuffer {
-    return new IndexedDbRemoteDocumentCache.RemoteDocumentChangeBuffer(this);
+  newChangeBuffer(options: {
+    createSentinelDocumentsToTrackDeletes: boolean;
```
mikelehen: 1. You say "sentinelDocuments" here but refer to them as "sentinel deletes" elsewhere.  Please use consistent terminology.
2. We have this long name here but just `trackDeletes` in the RemoteDocumentChangeBuffer constructor.  I am not sure which I prefer, but I can't think of a great reason for them to be different.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -331,95 +295,62 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+    this.lastProcessedReadTime = SnapshotVersion.forDeletedDoc();
+    return documentsStore.iterate(
+      { index: DbRemoteDocument.readTimeIndex, reverse: true },
       (key, value, control) => {
-        this._lastProcessedDocumentChangeId = key;
+        if (value.readTime) {
+          this.lastProcessedReadTime = SnapshotVersion.fromTimestamp(
+            new Timestamp(value.readTime![0], value.readTime![1])
+          );
+        }
```
mikelehen: What happens immediately after the schema upgrade when we've just added the index?  Will the index have no entries and so this iterate() is a no-op?  Or will it have empty entries for every document in the cache, so this iterate() will scan every over every document?  (possibly still okay, but it might be worth a comment flagging that this could be expensive)

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -137,18 +117,11 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
   ): PersistencePromise<void> {
     return this.getMetadata(transaction).next(metadata => {
       metadata.byteSize += sizeDelta;
-      return this.setMetadata(transaction, metadata).next(() => {
-        if (this.keepDocumentChangeLog) {
-          return documentChangesStore(transaction).put({
-            changes: this.serializer.toDbResourcePaths(changedKeys)
-          });
-        }
-      });
```
mikelehen: The method comment still says "Updates the document change log and ..."  Perhaps do a grep over the codebase for "changelog" and "change log" to make sure there aren't other stale instances?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -444,6 +375,29 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+    dbRemoteDoc: DbRemoteDocument | null
+  ): MaybeDocument | null {
+    if (dbRemoteDoc) {
+      const doc = this.serializer.fromDbRemoteDocument(dbRemoteDoc);
+      if (
+        doc instanceof NoDocument &&
+        doc.version.isEqual(SnapshotVersion.forDeletedDoc())
+      ) {
+        // The document is a sentinel delete and should only be used in the
+        // change log.
```
mikelehen: Since we don't have a proper "change log" anymore, perhaps change to "used in getNewDocumentChanges()" ?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -1017,19 +1005,17 @@ export class DbClientMetadata {
   static keyPath = 'clientId';
 
   constructor(
+    // Note: Previous schema versions included a field
+    // "lastProcessedDocumentChangeId". Don't use anymore.
```
mikelehen: nit: Might be better to keep this note at the bottom of the list, since that's where we would be adding new fields...

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -1060,11 +1046,7 @@ export const V1_STORES = [
 export const V3_STORES = V1_STORES;
 
 // Visible for testing
-export const V4_STORES = [
-  ...V3_STORES,
-  DbClientMetadata.store,
-  DbRemoteDocumentChanges.store
-];
+export const V4_STORES = [...V3_STORES, DbClientMetadata.store];
```
mikelehen: Hrm...  I'm having uneasy feelings about revising history like this.  I wonder if we should keep some hint about the prior existence of this store to keep our history of schema changes in-tact in this file.  I'm not sure though.  It also seems strange to keep dead code.  Maybe ask @wilhuff what he thinks.

packages/firestore/src/local/local_store.ts
```diff
@@ -345,7 +345,9 @@ export class LocalStore {
       'readwrite-primary',
       txn => {
         const affected = batchResult.batch.keys();
-        const documentBuffer = this.remoteDocuments.newChangeBuffer();
+        const documentBuffer = this.remoteDocuments.newChangeBuffer({
+          createSentinelDocumentsToTrackDeletes: true
```
mikelehen: Part of me wonders if this should be the default behavior and GC should explicitly opt-out (e.g. flip the logic and rename to `suppressSentinelDeleteDocuments` or something) since reading this code, I would have trouble guessing what this is for and why it is enabled in this code path.  By contrast, the GC code could have a comment explaining why we don't want to track deletes in that case.

Alternatively, since this option probably corresponds 100% to whether this is GC code or not, we could consider specializing the name to match (e.g. `forGarbageCollection: true` or something)

WDYT?

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -177,7 +178,9 @@ export class MemoryRemoteDocumentCache implements RemoteDocumentCache {
     return PersistencePromise.resolve(changedDocs);
   }
 
-  newChangeBuffer(): RemoteDocumentChangeBuffer {
+  newChangeBuffer(options: {
+    createSentinelDocumentsToTrackDeletes: boolean;
```
mikelehen: Perhaps there should be a comment indicating we explicitly ignore this.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -791,6 +791,24 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+            return remoteDocumentStore
+              .iterate(
+                { index: DbRemoteDocument.readTimeIndex, range },
+                (_, dbRemoteDoc) => {
+                  const doc = TEST_SERIALIZER.fromDbRemoteDocument(dbRemoteDoc);
+                  docsRead.push(doc.key.path.toString());
+                }
+              )
+              .next(() => {
+                // The old documents are not included since readTime was not populated
```
mikelehen: Would be nice if this test added some documents with index entries but that were older than version=1 to verify that are not included either...

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: Added. This makes me thing we should never have added the change log to the MemoryRemoteDocumentCache.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
schmidt-sebastian: I extracted this into a separate test ("RemoteDocumentCache can filter documents by read time") that does.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: This loop only ever reads the first row. If there is no processed read time it doesn't read any entries at all (verified via a quick unit test hack). Manually setting a read time range that ends with [0,0] also excludes entries with an undefined read time.

schmidt-sebastian:
I used "Sebastian's guide to screw with your commit history for this PR". The review changes are part of https://github.com/firebase/firebase-js-sdk/pull/2128/commits/8142123abdddc511ca9580e0a7e51ad7de4cb612

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1251,7 +1225,10 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
               // Our size accounting requires us to read all documents before
               // removing them.
               return changeBuffer.getEntry(txn, docKey).next(() => {
-                changeBuffer.removeEntry(docKey);
+                changeBuffer.removeEntry(
+                  docKey,
+                  SnapshotVersion.forDeletedDoc()
+                );
```
mikelehen: Per chat conversation, I will not stand for removeEntry() requiring a version that is MEANINGLESS!!! j/k...  But I am realizing that there is exactly one place where we want to track removed entries and that is for failed limbo resolution, so I'm wondering if we should:

1. Remove the `trackRemovals` option.
2. Make it so removeEntry() doesn't _require_ a version.
3. Let you optionally pass an entry in.  This could be `removeEntry(key, version)` or if we want to make it a little more explicit (and borrow a bit from the existing shape of the API), maybe `removeEntry(key, {trackRemovalVersion: version})` or something.

Alternatively, since this such a specific case, we could consider _not_ storing the sentinel delete and inventing a localStorage notification for it instead.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -131,24 +110,18 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
   }
 
   /**
-   * Updates the document change log and adds the given delta to the cached current size.
+   * Updates the current size.
```
mikelehen: maybe "current cache size" ?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -331,95 +295,65 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+        }
         control.done();
       }
     );
   }
 
-  newChangeBuffer(): RemoteDocumentChangeBuffer {
-    return new IndexedDbRemoteDocumentCache.RemoteDocumentChangeBuffer(this);
+  newChangeBuffer(options?: {
+    trackRemovals: boolean;
```
mikelehen: Per feedback elsewhere, I recommend dropping this option and "moving" it to the removeEntry() call.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -617,6 +630,12 @@ export class DbRemoteDocument {
 
   static collectionReadTimeIndexPath = ['parentPath', 'readTime'];
 
+  // TODO: We are currently storing full document keys almost three times
+  // (once as part of the primary key, once - partly - as `parentPath` and once
+  // inside the encoded documents). During our next migration, we should
```
mikelehen: We store it 3 more times if you include the indexes (collectionReadTimeIndex stores it twice, readTimeIndex stores it once).  Hopefully people don't use deeply-nested subcollections and small documents!

It occurs to me that we could maintain our own collectionReadTimeIndex that instead of being a true index was just an object store containing `{collectionPath, readTime, documentId}` objects.  This would save us storing the full documentKey.  It's probably not worth doing for web since it'd be less efficient than a real index, but with LevelDB (where we have to implement our own indexes anyway), it probably makes sense.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: Done

PR #2137, opened 2019-09-03
Avoid auth triggers if we haven't yet received the initial user.

rsgowman:
https://github.com/firebase/firebase-js-sdk/pull/2078/ altered the
initial state of currentUser from undefined to not undefined, which
caused the if statement to unconditionally evaluate to true. This
results in a race condition that could cause some initial writes to the
database to be dropped.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2135

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -135,6 +135,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
 
   /** Tracks the current User. */
   private currentUser: User = User.UNAUTHENTICATED;
+  private receivedInitialUser: boolean = false;
 
   /**
    * Counter used to detect if the token changed while a getToken request was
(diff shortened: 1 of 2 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -151,6 +152,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     this.tokenListener = () => {
       this.tokenCounter++;
       this.currentUser = this.getUser();
+      this.receivedInitialUser = true;
```
mikelehen: Could we just add a `this.currentUser = this.getUser();` call below so that it gets initialized in the constructor?  This would match Android (https://github.com/firebase/firebase-android-sdk/blob/29a1573ac1c6b7765aa6ed95ceafc95ec3b670f0/firebase-firestore/src/main/java/com/google/firebase/firestore/auth/FirebaseAuthCredentialsProvider.java#L73)

rsgowman: Good call; done.

PR #2139, opened 2019-09-04
Implement Promise finally

Feiyang1:
`finally` was added to the latest Promise spec. It [broke our CI](https://travis-ci.org/firebase/firebase-js-sdk/builds/580486955#L690) because `@types/node` is updated and starts using the latest Promise types.

I'm changing the interface to explicitly exclude `finally` to fix the issue.

### DIFF
--- a/packages/firestore/src/util/promise.ts
+++ b/packages/firestore/src/util/promise.ts
@@ -23,7 +23,8 @@ export interface Rejecter {
   (reason?: Error): void;
 }
 
-export interface CancelablePromise<T> extends Pick<Promise<T>, Exclude<keyof Promise<T>, 'finally'>> {
+export interface CancelablePromise<T>
+  extends Pick<Promise<T>, Exclude<keyof Promise<T>, 'finally'>> {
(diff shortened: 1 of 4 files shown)


packages/firestore/src/util/promise.ts
```diff
@@ -23,7 +23,8 @@ export interface Rejecter {
   (reason?: Error): void;
 }
 
-export interface CancelablePromise<T> extends Promise<T> {
+export interface CancelablePromise<T>
+  extends Pick<Promise<T>, Exclude<keyof Promise<T>, 'finally'>> {
```
merlinnot: This can be simplified to:

```suggestion
  extends Omit<Promise<T>, 'finally'> {
```

It would probably improve readability.

Feiyang1:
As per offline discussion, we will implement finally without check.
`finally = this.deferred.promise.finally.bind(this.deferred.promise);`

Feiyang1:
Closing the PR because it will break Node8 users.

PR #2142, opened 2019-09-04
[database] Add `request` to `optionalDependencies`

styfle:
The `database` package uses the `request` dependency but doesn't include it in package.json.

See this line that calls `require('request')`:
https://github.com/firebase/firebase-js-sdk/blob/6b53e0058483c9002d2fe56119f86fc9fb96b56c/packages/database/src/nodePatches.ts#L162

This adds `request` to the `optionalDependencies` (perhaps it should be under `dependencies`?)

I did not create an issue because it's easier to discuss with the code in this PR 😄

### DIFF
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -22,6 +22,9 @@
   },
   "license": "Apache-2.0",
   "peerDependencies": {},
+  "optionalDependencies": {
+    "request": "*"
+  },
   "dependencies": {
     "@firebase/database-types": "0.4.3",
     "@firebase/logger": "0.1.23",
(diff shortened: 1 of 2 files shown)


packages/database/package.json
```diff
@@ -22,6 +22,9 @@
   },
   "license": "Apache-2.0",
   "peerDependencies": {},
+  "optionalDependencies": {
+    "request": "*"
```
styfle: Would you prefer "2.x" instead of asterisk?

mikelehen:
Thanks for the PR!  Your change looks reasonable, but I'm discussing with a colleague and I think we may opt to just delete this code. It's not really used at this point.  Hang tight.

styfle:
@mikelehen That sounds good too, thanks!

Feel free to close this PR if you need to.

mikelehen:
Moving forward with the delete: https://github.com/firebase/firebase-js-sdk/pull/2144

So I'll close this but we do appreciate you brining this to our attention. 😀

PR #2144, opened 2019-09-04
Remove nodePatches.ts hackery.

mikelehen:
This had some very stale code to:
1. Make it possible to use long-polling instead of WebSockets from the Node.JS SDK. This existed purely for testing purposes, wasn't reliable, and hasn't been used in ages.
2. Work around a bad Node.JS bug that was fixed 6 years ago (https://github.com/joyent/node/issues/6506).


mikelehen:
Changelog added.

PR #2159, opened 2019-09-06
Index-Free: Filter document queries by read time

schmidt-sebastian:
This PR adds the ability to getAllDocumentsMatchingQuery() to skip documents older than a given update time. This allows us to only scan for documents that have been modified since the Query's Target Mapping has been persisted.

Port of https://github.com/firebase/firebase-android-sdk/pull/617

### DIFF
--- a/packages/firestore/test/unit/local/remote_document_cache.test.ts
+++ b/packages/firestore/test/unit/local/remote_document_cache.test.ts
@@ -17,6 +17,7 @@
 
 import { expect } from 'chai';
 import { Query } from '../../../src/core/query';
+import { SnapshotVersion } from '../../../src/core/snapshot_version';
 import { IndexedDbPersistence } from '../../../src/local/indexeddb_persistence';
 import { MaybeDocument } from '../../../src/model/document';
 import {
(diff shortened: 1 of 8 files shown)


packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -312,14 +313,57 @@ function genericRemoteDocumentCacheTests(
...
+  });
+
+  it('query matching uses read time rather than update time', async () => {
+    await cache.addEntries(
+      [doc('b/old', 1, DOC_DATA)],
+      /* readTime= */ version(2)
+    );
+    await cache.addEntries(
+      [doc('b/new', 2, DOC_DATA)],
+      /* readTime= */ version(1)
```
mikelehen: Reading versions that haven't been written yet!  Impressive.

(I assume this is intentional, though I'm also not sure it's a valuable test.  I can't immediately think of a bug that this test would catch that the previous one wouldn't).

schmidt-sebastian: Yeah, this might have been more valuable on Android since I switched the implementation from updateTime to readTime. I'll keep it for consistency.

PR #2160, opened 2019-09-06
Add SimpleQueryEngine

schmidt-sebastian:
Adds SimpleQueryEngine and wires it up.

Android: https://github.com/firebase/firebase-android-sdk/blob/715f3f03a96681e8dfd816b2645276b2cd58b8d7/firebase-firestore/src/main/java/com/google/firebase/firestore/local/SimpleQueryEngine.java
https://github.com/firebase/firebase-android-sdk/blob/715f3f03a96681e8dfd816b2645276b2cd58b8d7/firebase-firestore/src/main/java/com/google/firebase/firestore/local/QueryEngine.java

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -48,6 +48,7 @@ import { Persistence, PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { QueryCache } from './query_cache';
 import { QueryData, QueryPurpose } from './query_data';
+import { QueryEngine } from './query_engine';
 import { ReferenceSet } from './reference_set';
 import { RemoteDocumentCache } from './remote_document_cache';
 import { RemoteDocumentChangeBuffer } from './remote_document_change_buffer';
@@ -158,6 +159,7 @@ export class LocalStore {
   constructor(
     /** Manages our in-memory or durable persistence. */
     private persistence: Persistence,
+    private queryEngine: QueryEngine,
     initialUser: User
   ) {
     assert(
@@ -175,6 +177,7 @@ export class LocalStore {
       this.mutationQueue,
       this.persistence.getIndexManager()
     );
+    this.queryEngine.setLocalDocumentsView(this.localDocuments);
   }
 
   /**
@@ -819,7 +822,7 @@ export class LocalStore {
    */
   executeQuery(query: Query): Promise<DocumentMap> {
     return this.persistence.runTransaction('Execute query', 'readonly', txn => {
-      return this.localDocuments.getDocumentsMatchingQuery(
+      return this.queryEngine.getDocumentsMatchingQuery(
         txn,
         query,
         SnapshotVersion.MIN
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -175,6 +177,7 @@ export class LocalStore {
       this.mutationQueue,
       this.persistence.getIndexManager()
     );
+    this.queryEngine.setLocalDocumentsView(this.localDocuments);
```
mikelehen: How come we construct the QueryEngine in FirestoreClient and pass it into LocalStore, rather than just construct it here in LocalStore?  If we could do that, then we wouldn't need this awkward `setLocalDocumentsView()` method that must be called before use...

packages/firestore/src/local/simple_query_engine.ts
```diff
@@ -0,0 +1,57 @@
...
+    query: Query,
+    sinceReadTime: SnapshotVersion
+  ): PersistencePromise<DocumentMap> {
+    assert(
+      this.localDocumentsView !== undefined,
+      'setLocalDocumentsView() not called'
+    );
+
+    // TODO: Once LocalDocumentsView provides a getCollectionDocuments()
+    // method, we should call that here and then filter the results.
```
mikelehen: Hrm.  What's the context for this TODO (other than the fact that it exists in Android too)?  Why do we expect LocalDocumentsView is going to provide a `getCollectionDocuments()` method in the future?

packages/firestore/src/local/local_store.ts
schmidt-sebastian: It allows to parameterize the tests for the LocalStore. Fun times.

packages/firestore/src/local/simple_query_engine.ts
schmidt-sebastian: Hm. I don't know. I ported it cause it was there an Android. Unfortunately, all traces of who added it are gone by now. I could invest some research and dig this up on Codesearch, but I suspect you would not like the outcome :)

schmidt-sebastian:
Let me know if you want me to remove the TODO, otherwise I will merge once CI goes green.

packages/firestore/src/local/local_store.ts
mikelehen: It's tempting to suggest that we just have a `useIndexFreeQueryEngine` boolean argument or something that tests can use...  But I don't care that much.

packages/firestore/src/local/simple_query_engine.ts
mikelehen: Ohhh...  GRRRR... I forgot about that.  I suspect it used to be TODO(indexing) which would imply it doesn't apply to web.  So I would prefer if you don't port it.  But that means you have to run CI again, so maybe not worth removing (it's friday, YOLO, etc.)

schmidt-sebastian: I'll drop it in the next PR.

PR #2163, opened 2019-09-09
Clean up IndexedDb initialization

schmidt-sebastian:
My next PR is going to add one more argument to IndexedDbPersistence, so I thought it was a good idea to clean up the initialization a tiny bit.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -257,9 +257,6 @@ export class IndexedDbPersistence implements Persistence {
   private readonly webStorage: Storage;
   readonly referenceDelegate: IndexedDbLruDelegate;
 
-  // Note that `multiClientParams` must be present to enable multi-client support while multi-tab
-  // is still experimental. When multi-client is switched to always on, `multiClientParams` will
-  // no longer be optional.
   private constructor(
     private readonly allowTabSynchronization: boolean,
     private readonly persistenceKey: string,
@@ -268,7 +265,7 @@ export class IndexedDbPersistence implements Persistence {
     lruParams: LruParams,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    private readonly sequenceNumberSyncer?: SequenceNumberSyncer
+    private readonly sequenceNumberSyncer: SequenceNumberSyncer
   ) {
     this.referenceDelegate = new IndexedDbLruDelegate(this, lruParams);
     this.dbName = persistenceKey + IndexedDbPersistence.MAIN_DATABASE;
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -191,43 +188,32 @@ export class IndexedDbPersistence implements Persistence {
...
-      lruParams,
-      multiClientParams
+      options.allowTabSynchronization,
+      options.persistenceKey,
+      options.clientId,
+      options.platform,
+      options.lruParams,
+      options.queue,
+      options.serializer,
+      options.sequenceNumberSyncer
```
mikelehen: It seems slightly odd and arbitrary for createIndexedDbPersistence() to take an options object but then we flatten that into separate arguments when calling the IndexedDbPersistence constructor.

I don't care very strongly, but it seems like it'd be cleaner to either convert both to options objects or just use plain arguments for both.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -278,25 +261,19 @@ export class IndexedDbPersistence implements Persistence {
...
+    private readonly allowTabSynchronization: boolean,
     private readonly persistenceKey: string,
     private readonly clientId: ClientId,
     platform: Platform,
+    lruParams: LruParams,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    lruParams: LruParams,
-    private readonly multiClientParams?: MultiClientParams
+    private readonly sequenceNumberSyncer?: SequenceNumberSyncer
```
mikelehen: It looks like with this change, you are always passing a sequenceNumberSyncer, so this may not need to be optional?

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -278,25 +261,19 @@ export class IndexedDbPersistence implements Persistence {
   // is still experimental. When multi-client is switched to always on, `multiClientParams` will
   // no longer be optional.
```
mikelehen: stale comment

schmidt-sebastian: Removed

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: For IndexedDb it doesn't have to be optional anymore.

packages/firestore/src/local/indexeddb_persistence.ts
mikelehen: Oh, right!  That's a pretty good reason.

PR #2164, opened 2019-09-09
Expose operation counts from Persistence layer

schmidt-sebastian:
This is a port of https://github.com/firebase/firebase-android-sdk/pull/595

The ultimate goal of this PR is to allow us write unit tests that count how many items are processed during query execution. For this to work, this PR adds a StatsCollector. The RemoteDocumentCache and the MutationQueue have been updated to provide stats for every read, delete and insert operation. I tried to ensure that they counts between the different backing stores would match and added a unit test that ensures that this is true for a small subset of the operations. 

I also changed the way that MemoryPersistence gets initialized to align it a bit closer with IndexedDbPersistence.

There are likely a bunch of issues with the stats accounting that we could either:
- Ignore.
- Uncover by adding more unit test.
- Address by removing the stats accounting for areas of this PR that are not tested.

We could also drop this PR altogether and only do the counting on one platform.

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -384,7 +384,9 @@ export class FirestoreClient {
    * @returns A promise that will successfully resolve.
    */
   private startMemoryPersistence(): Promise<LruGarbageCollector | null> {
-    this.persistence = MemoryPersistence.createEagerPersistence(this.clientId);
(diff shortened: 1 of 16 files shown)


schmidt-sebastian:
Dropping in favor of more isolated approach in https://github.com/firebase/firebase-js-sdk/pull/2165

PR #2165, opened 2019-09-10
Expose operation counts from Persistence layer (take 2)

schmidt-sebastian:
"Sebastian uses his time wisely."

This is another take on https://github.com/firebase/firebase-js-sdk/pull/2164.

It solves the following issues:
- It doesn't add any stats that are not used/not tested.
- It doesn't add code to the main client.
- It removes the need to plumb a counter through the Persistence layer.
- it reduces the reliance on the exact implementation, but instead focuses more on the result of each operation.
- it gives us a reason to port this back to Android.

### DIFF
--- a/packages/firestore/src/local/local_documents_view.ts
+++ b/packages/firestore/src/local/local_documents_view.ts
@@ -48,9 +48,9 @@ import { RemoteDocumentCache } from './remote_document_cache';
  */
 export class LocalDocumentsView {
   constructor(
-    private remoteDocumentCache: RemoteDocumentCache,
-    private mutationQueue: MutationQueue,
-    private indexManager: IndexManager
+    readonly remoteDocumentCache: RemoteDocumentCache,
+    readonly mutationQueue: MutationQueue,
+    readonly indexManager: IndexManager
   ) {}
 
   /**
--- a/packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
+++ b/packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
@@ -27,8 +27,8 @@ import { MutationQueue } from '../../../src/local/mutation_queue';
 
 /**
  * A test-only query engine that forwards all calls to the query engine
- * provided at construction time, but provides access to the number of
- * documents and mutations read.
+ * provided at construction time, while exposing to the number of documents and
+ * mutations read.
  */
 export class ForwardingCountingQueryEngine implements QueryEngine {
   /**
@@ -45,7 +45,7 @@ export class ForwardingCountingQueryEngine implements QueryEngine {
 
   constructor(private readonly queryEngine: QueryEngine) {}
 
-  resetCounts() : void {
+  resetCounts(): void {
     this.mutationsRead = 0;
     this.documentsRead = 0;
   }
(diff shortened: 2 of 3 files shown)


schmidt-sebastian:
Note that I did experiment with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy, but was never able to get it to work via TS Node. It would also require us to use a different tsconfig for the unit tests (to add a dependency on es2015.proxy).

schmidt-sebastian:
"Coverage decreased (-0.05%) to 89.621%"

I don't believe you.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -48,9 +48,9 @@ import { RemoteDocumentCache } from './remote_document_cache';
  */
 export class LocalDocumentsView {
   constructor(
-    private remoteDocumentCache: RemoteDocumentCache,
-    private mutationQueue: MutationQueue,
-    private indexManager: IndexManager
+    readonly remoteDocumentCache: RemoteDocumentCache,
+    readonly mutationQueue: MutationQueue,
+    readonly indexManager: IndexManager
```
mikelehen: I'm guessing this is why coverage decreased.  Making these readonly probably costs us a few bytes. 😄

packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
```diff
@@ -0,0 +1,153 @@
...
+import { PersistencePromise } from '../../../src/local/persistence_promise';
+import { DocumentMap } from '../../../src/model/collections';
+import { RemoteDocumentCache } from '../../../src/local/remote_document_cache';
+import { MutationQueue } from '../../../src/local/mutation_queue';
+
+/**
+ * A test-only query engine that forwards all API calls and exposes the number
+ * of documents and mutations read.
+ */
+export class ForwardingCountingQueryEngine implements QueryEngine {
```
mikelehen: `ForwardingCountingQueryEngine` doesn't read super nicely.  I think I would do `OpCountingQueryEngine` or `OpCountingQueryEngineWrapper` or something.

packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
```diff
@@ -0,0 +1,153 @@
...
+      this.wrapRemoteDocumentCache(localDocuments.remoteDocumentCache),
+      this.wrapMutationQueue(localDocuments.mutationQueue),
+      localDocuments.indexManager
+    );
+
+    return this.queryEngine.setLocalDocumentsView(view);
+  }
+
+  private wrapRemoteDocumentCache(
+    target: RemoteDocumentCache
```
mikelehen: "target" is already used for something else in the client. Find a new term. 😄   E.g.: `remoteDocumentCache`, `internal`, `wrapped`

packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
schmidt-sebastian: I dropped the "Forwarding" part. Not sold on your suggestions :)

packages/firestore/test/unit/local/forwarding_counting_query_engine.ts
schmidt-sebastian: I used "source" (which is also used, but much more lightly...)

schmidt-sebastian:
Merging this since the pervious CI run passed and changes are just renames/minor.

PR #2166, opened 2019-09-10
Port matchesAllDocuments()

schmidt-sebastian:
Ports:
- https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/core/Query.java#L123
- https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/test/java/com/google/firebase/firestore/core/QueryTest.java#L534

This will be used to opt out Index-Free query processing for queries that match all documents.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -199,6 +199,22 @@ export class Query {
     );
   }
 
+  /**
+   * Returns true if this query does not specify any query constraints that
+   * could remove results.
+   */
+  matchesAllDocuments(): boolean {
+    return (
+      this.filters.length === 0 &&
+      this.limit === null &&
+      this.startAt == null &&
+      this.endAt == null &&
+      (this.explicitOrderBy.length === 0 ||
+        (this.explicitOrderBy.length === 1 &&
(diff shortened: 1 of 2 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -199,6 +199,22 @@ export class Query {
...
+   */
+  matchesAllDocuments(): boolean {
+    return (
+      this.filters.length === 0 &&
+      this.limit === null &&
+      this.startAt == null &&
+      this.endAt == null &&
+      (this.explicitOrderBy.length === 0 ||
+        (this.explicitOrderBy.length === 1 &&
+          this.getFirstOrderByField()!.isKeyField()))
```
mikelehen: You could do `this.explicitOrderBy[0].field.isKeyField()` which would be more congruous with the previous check and avoid the need for `!`.

schmidt-sebastian: Done. Thanks for the suggestion.

schmidt-sebastian:
Merging this since the pervious CI run passed and changes are just renames.

PR #2168, opened 2019-09-11
Fixing Document.toString()

schmidt-sebastian:
This broke when data became data() in https://github.com/firebase/firebase-js-sdk/pull/2115


schmidt-sebastian:
> Should this have a test?

Hmm. Maybe. I'll add one if anyone else chimes in :)

PR #2169, opened 2019-09-11
Add IndexFreeQueryEngine

schmidt-sebastian:
This PR ports the current state of https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/local/IndexFreeQueryEngine.java and https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/test/java/com/google/firebase/firestore/local/IndexFreeQueryEngineTest.java, as well as the updated plumbing for `executeQuery()` (which allows us to pass in the remoteKeys if we already have them, since we don't want to fetch them twice).

This is the last PR for Index-Free queries that adds logic to the client. There will be one more PR that adds new tests to the LocalStore.

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -615,6 +615,23 @@ export class FirestoreClient {
     return this.databaseInfo.databaseId;
   }
 
+  addSnapshotsInSyncListener(observer: Observer<void>): void {
+    this.verifyNotTerminated();
+    this.asyncQueue.enqueueAndForget(() => {
+      this.eventMgr.addSnapshotsInSyncListener(observer);
+      return Promise.resolve();
+    });
+  }
+
+  removeSnapshotsInSyncListener(observer: Observer<void>): void {
+    // Checks for shutdown but does not raise error, allowing remove after
+    // shutdown to be a no-op.
+    if (this.clientTerminated) {
+      return;
+    }
+    this.eventMgr.removeSnapshotsInSyncListener(observer);
+  }
+
   get clientTerminated(): boolean {
     // Technically, the asyncQueue is still running, but only accepting operations
     // related to termination or supposed to be run after termination. It is effectively
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -22,11 +22,6 @@ import { LocalStore } from '../local/local_store';
 import { MemoryPersistence } from '../local/memory_persistence';
 import { Persistence } from '../local/persistence';
 import { SimpleQueryEngine } from '../local/simple_query_engine';
-import {
-  DocumentKeySet,
-  documentKeySet,
-  DocumentMap
-} from '../model/collections';
 import { Document, MaybeDocument, NoDocument } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { Mutation } from '../model/mutation';
@@ -45,7 +40,7 @@ import {
   QueryListener
 } from './event_manager';
 import { SyncEngine } from './sync_engine';
-import { View, ViewDocumentChanges } from './view';
+import { View } from './view';
 
 import {
   LruGarbageCollector,
@@ -584,22 +579,18 @@ export class FirestoreClient {
 
   getDocumentsFromLocalCache(query: Query): Promise<ViewSnapshot> {
     this.verifyNotTerminated();
-    return this.asyncQueue
-      .enqueue(() => {
-        return this.localStore.executeQuery(query);
-      })
-      .then((docs: DocumentMap) => {
-        const remoteKeys: DocumentKeySet = documentKeySet();
-
-        const view = new View(query, remoteKeys);
-        const viewDocChanges: ViewDocumentChanges = view.computeDocChanges(
-          docs
-        );
-        return view.applyChanges(
-          viewDocChanges,
-          /* updateLimboDocuments= */ false
-        ).snapshot!;
-      });
+    return this.asyncQueue.enqueue(async () => {
+      const queryResult = await this.localStore.executeQuery(
+        query,
+        /* usePreviousResults= */ true
+      );
+      const view = new View(query, queryResult.remoteKeys);
+      const viewDocChanges = view.computeDocChanges(queryResult.documents);
+      return view.applyChanges(
+        viewDocChanges,
+        /* updateLimboDocuments= */ false
+      ).snapshot!;
+    });
   }
 
   write(mutations: Mutation[]): Promise<void> {
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -246,11 +246,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     const remoteKeys = await this.localStore.remoteDocumentKeys(
       queryData.targetId
     );
-    const docs = await this.localStore.executeQuery(
-      query,
+    const docs = await this.localStore.executeQuery(query, {
       queryData,
       remoteKeys
-    );
+    });
 
     const view = new View(query, remoteKeys);
     const viewDocChanges = view.computeDocChanges(docs);
@@ -290,11 +289,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
     const remoteKeys = await this.localStore.remoteDocumentKeys(
       queryData.targetId
     );
-    const docs = await this.localStore.executeQuery(
-      queryData.query,
+    const docs = await this.localStore.executeQuery(queryData.query, {
       queryData,
       remoteKeys
-    );
+    });
     const viewSnapshot = view.synchronizeWithPersistedState(docs, remoteKeys);
     if (this.isPrimary) {
       this.updateTrackedLimbos(queryData.targetId, viewSnapshot.limboChanges);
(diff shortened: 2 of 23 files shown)


packages/firestore/test/unit/local/index_free_query_engine.test.ts
```diff
@@ -0,0 +1,362 @@
...
+  1,
+  { matches: false, order: 1 },
+  { hasLocalMutations: true }
+);
+const UPDATED_DOC_A = doc('coll/a', 11, { matches: true, order: 1 });
+const MATCHING_DOC_B = doc('coll/b', 1, { matches: true, order: 2 });
+const UPDATED_MATCHING_DOC_B = doc('coll/b', 11, { matches: true, order: 2 });
+
+/**
+ * A LocalDocumentsView that inspects the arguments to
```
mikelehen: Maybe add "wrapper" to this description.

packages/firestore/test/unit/local/index_free_query_engine.test.ts
```diff
@@ -0,0 +1,362 @@
...
+
+    await addDocument(MATCHING_DOC_A, MATCHING_DOC_B);
+    await persistQueryMapping(MATCHING_DOC_A.key, MATCHING_DOC_B.key);
+
+    const docs = await expectIndexFreeQuery(() => runQuery(query, queryData));
+
+    verifyResult(docs, [MATCHING_DOC_A, MATCHING_DOC_B]);
+  });
+
+  it('filters non-matching initial results', async () => {
```
mikelehen: I'm not sure what "initial results" refers to...  should this be "filters non-matching changes since initial results" by chance?

packages/firestore/test/unit/local/index_free_query_engine.test.ts
```diff
@@ -0,0 +1,362 @@
...
+  });
+
+  it('does not use initial results for limit query when last document has been updated out of band', async () => {
+    const query = Query.atPath(path('coll'))
+      .addFilter(filter('matches', '==', true))
+      .addOrderBy(orderBy('order', 'desc'))
+      .withLimit(1);
+
+    // Add a query mapping for a document that matches, but that sorts below
+    // another document based due to an update that the SDK received after the
```
mikelehen: "based due"?

packages/firestore/test/unit/local/index_free_query_engine.test.ts
```diff
@@ -0,0 +1,362 @@
...
+function testQueryData(query: Query, hasLimboFreeSnapshot: boolean): QueryData {
+  return new QueryData(
+    query,
+    TEST_TARGET_ID,
+    QueryPurpose.Listen,
+    ListenSequence.INVALID,
+    version(10),
+    hasLimboFreeSnapshot ? version(10) : SnapshotVersion.MIN
+  );
+}
```
mikelehen: Why are these helpers at the bottom instead of the top?

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -73,7 +73,7 @@ import {
 } from '../../util/helpers';
 
 import { FieldValue, IntegerValue } from '../../../src/model/field_value';
-import { CountingQueryEngine } from './forwarding_counting_query_engine';
+import { CountingQueryEngine } from './counting_query_engine';
```
mikelehen: 😲

packages/firestore/src/local/local_store.ts
```diff
@@ -783,51 +800,84 @@ export class LocalStore {
...
+  executeQuery(query: Query): Promise<DocumentMap>;
+  /**
+   * Runs the specified query against the local store and returns the results,
+   * potentially taking advantage of the provided query data and the set of
+   * remote document keys.
+   */
+  executeQuery(
+    query: Query,
+    queryData: QueryData | null,
+    remoteKeys: DocumentKeySet
```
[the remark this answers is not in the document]
mikelehen: Sorry, but I'm not really a fan:
1. I think the default behavior should be a full collection scan and you should have to explicitly opt into the code paths that take advantage of optimizations that might not always be safe.
2. "needsRefill" is specifically used for a different purpose in the SDK and so we shouldn't overload it...  and I don't think it's a good term for this situation anyway (what are we "refilling"?).
3. I still don't like passing in all of `QueryData`.  It's kind of a grab bag data structure that was born out of convenience (it's hard to describe to somebody what it represents).  Propagating it to more layers of our API does not seem beneficial (there's probably some OO principle about giving methods the minimum data they need to accomplish their task or something...?)

I'm not 100% sure what it means for this to be implementation agnostic (or else why it's important).  I think if you wrapped this up in a `CachedResultsInfo` or whatever it would become very clear why e.g. `SimpleQueryEngine` just ignores it... whereas right now it's not clear what a QueryEngine implementer should / shouldn't do with remoteKeys and queryData.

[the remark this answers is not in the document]
mikelehen: FWIW (you're still free to push back / suggest alternatives) - Fleshing out my proposal a bit more, I think I'd maybe change CachedResultsInfo.lookup(targetId) to CachedResultsInfo.unknown(targetId) and add CachedResultsInfo.None, which would be the default behavior if no CachedResultsInfo was passed.

packages/firestore/test/unit/local/index_free_query_engine.test.ts
```diff
@@ -349,14 +412,3 @@ function verifyResult(
...
-function testQueryData(query: Query, hasLimboFreeSnapshot: boolean): QueryData {
-  return new QueryData(
-    query,
-    TEST_TARGET_ID,
-    QueryPurpose.Listen,
-    ListenSequence.INVALID,
-    version(10),
-    hasLimboFreeSnapshot ? version(10) : SnapshotVersion.MIN
-  );
-}
```
mikelehen: I don't care super strongly, but as a general rule I hate test boilerplate and so I preferred having the specialized helper.  Was there a reason to remove it?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -584,22 +579,18 @@ export class FirestoreClient {
...
-        const viewDocChanges: ViewDocumentChanges = view.computeDocChanges(
-          docs
-        );
-        return view.applyChanges(
-          viewDocChanges,
-          /* updateLimboDocuments= */ false
-        ).snapshot!;
-      });
+    return this.asyncQueue.enqueue(async () => {
+      const queryResult = await this.localStore.executeQuery(
```
schmidt-sebastian: Updated this to all run on the async queue.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -584,22 +579,18 @@ export class FirestoreClient {
...
+        query,
+        /* usePreviousResults= */ true
+      );
+      const view = new View(query, queryResult.remoteKeys);
```
schmidt-sebastian: `queryResult.remoteKeys` is used now instead of `null`, which seems more correct, but doesn't change the state of the `fromCache` flag since it is set to true either way.

packages/firestore/src/local/local_store.ts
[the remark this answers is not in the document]
schmidt-sebastian: I updated this based on our offline discussion:

- `executeQuery()` now returns the remoteKeys, which means that there are now only two cases this method needs to address (Index-Free: yes or no). We don't need to pass remote keys since the method returns them. As discussed, this adds slight overhead to limit Query refills.

- This change also solves a potential multi-tab race since the set of remote keys and the set of remote documents are now fetched in a single transaction.

- I also changed the QueryEngine to only take the lastLimbFreeSnapshotVersion instead of QueryData.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -786,9 +783,14 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
             // The query has a limit and some docs were removed, so we need
             // to re-run the query against the local store to make sure we
             // didn't lose any good docs that had been past the limit.
-            return this.localStore.executeQuery(queryView.query).then(docs => {
-              return queryView.view.computeDocChanges(docs, viewDocChanges);
-            });
+            return this.localStore
+              .executeQuery(queryView.query, /* usePreviousResults= */ false)
+              .then(({ documents }) => {
```
schmidt-sebastian: Note: I verified that the transpiler unpacks this (in `cjs` and `esm` mode).

packages/firestore/src/local/local_store.ts
```diff
@@ -782,13 +788,22 @@ export class LocalStore {
     transaction: PersistenceTransaction,
     query: Query
   ): PersistencePromise<QueryData | null> {
-    return this.queryCache.getQueryData(transaction, query).next(queryData => {
-      if (queryData == null) {
-        return null;
+    let queryData: QueryData | null = null;
+
+    // Look up the query data in our local map first, as this avoid a potential
```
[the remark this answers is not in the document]
schmidt-sebastian: Fixed.

packages/firestore/src/local/local_store.ts
```diff
@@ -782,13 +788,22 @@ export class LocalStore {
...
+      this.queryDataByTarget,
+      (targetId, cachedQueryData) => {
+        if (cachedQueryData.query.isEqual(query)) {
+          queryData = cachedQueryData;
+        }
       }
-      const cachedQueryData = this.queryDataByTarget[queryData!.targetId];
-      return cachedQueryData || queryData;
-    });
+    );
```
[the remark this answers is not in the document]
schmidt-sebastian: I'll merge this as is and will send you a follow-up PR soonish. My gut feeling is that `n` has to be fairly large to be slower than an IndexedDb lookup. This is mostly based on how much faster our Memory persistence layer is in compared to our IndexedDb-backed implementation.

PR #2176, opened 2019-09-13
Change the order of checks in applyRemoteEvent()

schmidt-sebastian:
If there is a chance that a limbo resolution fails for a document that no longer exists, we could get into a state where we write a document into the RemoteDocumentCache with a snapshot version of zero. By re-ordering the statements, we make sure this does not happen

Port of https://github.com/firebase/firebase-android-sdk/pull/802

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -516,21 +516,32 @@ export class LocalStore {
           documentBuffer.getEntries(txn, updatedKeys).next(existingDocs => {
             remoteEvent.documentUpdates.forEach((key, doc) => {
               const existingDoc = existingDocs.get(key);
+
+              // Note: The order of the steps below is important, since we want
+              // to ensure that rejected limbo resolutions (which fabricate
+              // NoDocuments with SnapshotVersion.MIN) never add documents to
+              // cache.
               if (
+                doc instanceof NoDocument &&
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -516,23 +516,34 @@ export class LocalStore {
...
                 doc.version.compareTo(existingDoc.version) > 0 ||
                 (doc.version.compareTo(existingDoc.version) === 0 &&
                   existingDoc.hasPendingWrites)
               ) {
+                // TODO(index-free): Comment in this assert when we enable
+                // Index-Free queries
+                // assert(
+                //  !SnapshotVersion.MIN.isEqual(remoteEvent.snapshotVersion),
+                //  'Cannot add a document when the remote version is zero'
+                // );
```
mikelehen: What do you think of adding this now but as a non-fatal error log statement that suggests people file a github issue (or maybe we should open a tracking issue now and just ask people to add a comment, just in case a massive number of people hit the log statement)?  That would give us some extra confidence that we won't hit issues when we add the assert for real.

schmidt-sebastian: That's a great idea. I changed this to a logline right now, and hope that people will flock to GitHub if they see this. We probably want to do the same from Android if we don't get reproduction steps before the next code freeze.

PR #2179, opened 2019-09-17
Add Query to Target ID Map

schmidt-sebastian:
This PR replaces the O(n) loop in getQueryData() with a key lookup in an ObjectMap. This follows the precedent in SyncEngine (and maybe eventually leads us down the path of merging the different query tracking mechanisms).

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -39,6 +39,7 @@ import { RemoteEvent, TargetChange } from '../remote/remote_event';
 import { assert } from '../util/assert';
 import * as log from '../util/log';
 import * as objUtils from '../util/obj';
+import { ObjectMap } from '../util/obj_map';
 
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -816,32 +815,31 @@ export class LocalStore {
...
-          return PersistencePromise.forEach(removed, (key: DocumentKey) =>
-            this.persistence.referenceDelegate.removeReference(txn, key)
-          ).next(() =>
-            this.persistence.referenceDelegate.removeTarget(txn, queryData!)
-          );
-        } else {
-          return PersistencePromise.resolve();
-        }
-      });
+      const targetId = this.targetIdByQuery.get(query);
```
schmidt-sebastian: We now have a map to lookup targetIds for active queries, so we no longer need to go to the QueryCache.

mikelehen: Good call!

PR #2180, opened 2019-09-17
Index-Free Queries (feature, code still disabled)

schmidt-sebastian:
This PR adds (but doesn't enable) Index-Free queries for Web.


schmidt-sebastian:
Note: Do not merge until after Fireconf.

PR #2182, opened 2019-09-19
Sort classes properly in doc.

gabrc:
The ordering currently doesn't make sense:
https://firebase.google.com/docs/reference/js/firebase.auth
https://firebase.google.com/docs/reference/js/firebase.storage

This fixes b/141303832.


hsubox76:
Thanks!
1) I'm adding Eric Gilmore from docs, if he approves we can merge.
2) Don't forget to also change the [Node TOC file](https://github.com/firebase/firebase-js-sdk/blob/master/scripts/docgen/content-sources/node/toc.yaml) as well (it is identical minus auth methods that are not supported in Node)

gabrc:
Thank you. I've updated the other file.

Sorry for the force push. I'm new to Git, and was told I shouldn't have done that.

PR #2187, opened 2019-09-23
implement limitToLast

wu-hui:
Implements `limitToLast` based on https://github.com/firebase/firebase-js-sdk/commit/af1502abc0d63c8b393b0671e22f1030994993cf

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7260,15 +7260,6 @@ declare namespace firebase.firestore {
     limit(limit: number): Query;
 
     /**
-     * Creates and returns a new Query where only the first matching documents
-     * are returned as results.
-     *
-     * @param limit The maximum number of items to return.
-     * @return The created Query.
-     */
-    limitToFirst(limit: number): Query;
-
-    /**
      * Creates and returns a new Query where only the last matching documents
      * are returned as results.
      *
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7251,13 +7251,31 @@ declare namespace firebase.firestore {
     ): Query;
 
     /**
-     * Creates and returns a new Query where the results are limited to the
-     * specified number of documents.
+     * Creates and returns a new Query where only the first matching documents
+     * are returned as results.
      *
      * @param limit The maximum number of items to return.
      * @return The created Query.
      */
     limit(limit: number): Query;
+
+    /**
+     * Creates and returns a new Query where only the first matching documents
+     * are returned as results.
+     *
+     * @param limit The maximum number of items to return.
+     * @return The created Query.
+     */
+    limitToFirst(limit: number): Query;
+
+    /**
+     * Creates and returns a new Query where only the last matching documents
+     * are returned as results.
+     *
+     * @param limit The maximum number of items to return.
+     * @return The created Query.
+     */
+    limitToLast(limit: number): Query;
 
     /**
      * Creates and returns a new Query that starts at the provided document
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -168,7 +168,7 @@ export class Query {
   }
 
   withLimitToLast(limit: number | null): Query {
-    const q = new Query(
+    return new Query(
       this.path,
       this.collectionGroup,
       this.explicitOrderBy.slice(),
@@ -178,7 +178,6 @@ export class Query {
       this.startAt,
       this.endAt
     );
-    return q;
   }
 
   withStartAt(bound: Bound): Query {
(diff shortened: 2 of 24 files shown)


packages/firebase/index.d.ts
```diff
@@ -7251,14 +7251,32 @@ declare namespace firebase.firestore {
...
     limit(limit: number): Query;
 
+    /**
+     * Creates and returns a new Query where only the first matching documents
+     * are returned as results.
+     *
+     * @param limit The maximum number of items to return.
+     * @return The created Query.
+     */
+    limitToFirst(limit: number): Query;
```
mikelehen: Just FYI- Per the API proposal review, we should keep limit() as-is (don't deprecate it) and not add limitToFirst().  So the only public API change is adding Query.limitToLast().

FWIW I think I would be inclined to still _internally_ differentiate them as limitToFirst / limitToLast though.

packages/firestore/src/core/query.ts
```diff
@@ -47,6 +52,7 @@ export class Query {
     readonly explicitOrderBy: OrderBy[] = [],
     readonly filters: Filter[] = [],
     readonly limit: number | null = null,
+    readonly limitType: LimitType = LimitType.First,
```
mikelehen: Talking to @wilhuff I think we want to start trying to differentiate `Query` and `Target`, where `Query` represents any query the SDK supports (including limit-to-last queries and in theory could include local "and", "or", etc. queries in the future) and `Target` represents only queries that the backend supports (and corresponds directly to the `Target` proto in the wire protocol).

To fully realize that, we could create a new `Target` type and replace many (most?) usages of `Query` with that new type, but in the short-term I think we could just have a `Query.toTargetQuery()` method and add a `isTargetQuery` flag.  When you call `.toTargetQuery()` it'll translate the query as necessary (in particular the limit-to-last translation) and set the `isTargetQuery` flag to true.

We can then add assertions in the rest of our code to verify `isTargetQuery` is true in places where we expect to be dealing with backend queries, in particular we should assert in:
1. The QueryData constructor  (and we should eventually rename QueryData to TargetData)
2. The IndexedDb and Memory implementations of `QueryCache.getQueryData()`.  This will ensure that we are only storing (and generating canonicalIds for) the Target version of the query.  FWIW- If we wanted we could rename QueryCache to TargetCache in the future.
3. `Serializer.ToQueryTarget()`.  This ensures we only send the Target version of the query to the backend or write it to persistence.

And once you add those asserts you'll need to presumably fix the calling code to properly do the conversions so that the asserts pass...  and there may be other places we can/should assert.

I haven't fully explored this so if this ends up messy / complicated / whatever, feel free to holler.

packages/firestore/src/local/memory_query_cache.ts
```diff
@@ -182,7 +182,7 @@ export class MemoryQueryCache implements QueryCache {
     transaction: PersistenceTransaction,
     query: Query
   ): PersistencePromise<QueryData | null> {
-    const queryData = this.queries.get(query) || null;
+    const queryData = this.queries.getWithoutValueEqualityCheck(query) || null;
```
mikelehen: I think we want to undo this but add an assert that query.isTargetQuery is true...

packages/firestore/src/remote/serializer.ts
```diff
@@ -1008,6 +1009,14 @@ export class JsonProtoSerializer {
   }
 
   toQueryTarget(query: Query): api.QueryTarget {
+    // limitToLast is implemented client-side and must be "normalized" before
+    // sending to the backend.
+    query = query.convertLimitToFirstIfNecessary();
```
mikelehen: I think we should remove this (and fix calling code instead) and the assert below should become query.isTargetQuery == true

packages/firestore/src/remote/serializer.ts
```diff
@@ -1101,12 +1110,14 @@ export class JsonProtoSerializer {
       endAt = this.fromCursor(query.endAt);
     }
 
+    // TODO(limitToLast): Persist limitType somehow? :-/
```
mikelehen: We're explicitly not persisting it so this todo can go away.

packages/firestore/src/core/query.ts
wu-hui: One place `Query` should be used is what `View` references to. Right now it is constructed with `QueryData.query` and it will become a target query once we introduced such notion. But it needs to be still a `Query`, such that we keep the wanted order, is that right?

mikelehen: That sounds right.  We will need to somehow initialize it with the original query instead of queryData.query, I guess...  but if that ends up too hard / awkward, holler.

PR #2188, opened 2019-09-24
Remove database field from REST API request body.

yuchenshi:
(Discussed internally.)

### DIFF
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -38,6 +38,7 @@ import { StreamBridge } from '../remote/stream_bridge';
 import { assert, fail } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import * as log from '../util/log';
+import { Indexable } from '../util/misc';
 import { Rejecter, Resolver } from '../util/promise';
 import { StringMap } from '../util/types';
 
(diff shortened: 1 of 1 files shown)


packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -172,7 +172,7 @@ export class WebChannelConnection implements Connection {
         }
       });
 
-      const jsonObj = {...request} as any;
+      const jsonObj = { ...request } as any;
```
mikelehen: We try not to use `any` (and the linter won't let you) since it essentially disables all type-checking.  Try `object` instead?

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -172,7 +172,9 @@ export class WebChannelConnection implements Connection {
         }
       });
 
-      const requestString = JSON.stringify(request);
+      const jsonObj = { ...request } as any;
+      delete jsonObj.database; // The database field is already encoded in URL.
```
mikelehen: Your comment makes it sound like we are only doing this as an optimization, but as I understand it, this is _required_ to make the SDK work with the emulator and so I would make that clearer in the comment so somebody modifying this code in the future knows the consequences of changing this.

packages/firestore/src/platform_browser/webchannel_connection.ts
mikelehen: Poking at this a little more, this may be what you need:

```
import { Indexable } from '../util/misc';
...
const jsonObj = { ...request } as unknown as Indexable;
```

yuchenshi: Done.

packages/firestore/src/platform_browser/webchannel_connection.ts
yuchenshi: Good point! I've added more comments.

PR #2192, opened 2019-09-25
Add Firebase Analytics package and integrate with FCM

Feiyang1:
Add Firebase Analytics and integrate with FCM

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,4 +1,5 @@
 {
   singleQuote: true,
-  printWidth: 80
+  printWidth: 80,
+  quoteProps: preserve
 }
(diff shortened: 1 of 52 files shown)


.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   singleQuote: true,
-  printWidth: 80
+  printWidth: 80,
+  quoteProps: preserve
```
mikelehen: I'm surprised to see a prettier change included in this merge.  I assume it's intentional?  Do we need to re-run prettier on the repo now?  (maybe not, since "preserve" apparently respects the existing code?)

Feiyang1: @hsubox76 Do you remember why we need to introduce `quoteProps: preserve`?

PR #2195, opened 2019-09-26
correct package name for remote-config

Feiyang1:
The package name is wrong in README.md. This pr corrects it.


Feiyang1:
@erikeldridge ping :)

Issue #983, opened 2018-07-05
FR: Firestore offline persistence in web workers


lamplightdev:
* Operating System version: n/a
  * Browser version: n/a
  * Firebase SDK version: n/a
  * Firebase Product: firestore

Are there any plans to make offline persistence available in web workers? Inspired by @davideast 's article [I dropped 95% of my Firebase bundle size using this one weird trick](https://davidea.st/articles/firebase-bundle-size) to use firebase in a worker, I've noticed offline persistence isn't available in that context. Browsing the source, I suspect it's the reliance on (and absence in workers of) localStorage that's the issue - so I was wondering if there's any alternative to using synchronous storage in this case? Thanks.

joshdifabio:
Any news on this guys? The lack of offline persistence in service workers & web workers is a bit of a killer for PWA.

MarvinMiles:
Same here

mikelehen:
Thanks for your use case.  One option would of course be to pull the data from Firestore on the main event loop and then just do the processing on the webworker, but if it's a lot of data, perhaps there are performance wins for moving the Firestore interaction off the main event loop as well. So it's a reasonable request.

Service workers are a bit more complicated. They don't support XHR, which Firestore (and possibly other parts of Firebase) depends on.  And service worker lifetime is generally short-lived. Our supposition so far is that it doesn't make that much sense to use Firestore from a service worker. But if folks have use cases, please do share.

joshdifabio:
> Thanks for your use case. One option would of course be to pull the data from Firestore on the main event loop and then just do the processing on the webworker

Thanks. I did consider this, but it wasn't practical for a few reasons. For example:
1) This approach would significantly increase the amount of boilerplate code I'd need in order to get data in and out of the worker thread
2) When processing Firestore snapshots I often need access to Firestore's snapshot objects in order to process the data more efficiently, but class instances cannot be passed between the main thread and workers.

joshdifabio:
Something I forgot to add previously: every now and then I see the main thread blocked for 100ms+ by Firestore XHR-related scripts shortly after cancelling snapshot listeners. (Note that this timing is taken on a powerful MacBook Pro and I'd expect the main thread to be blocked for much longer on mobile.) This is another reason I offload most of my Firestore interaction to a worker thread.

mikelehen:
[...]

PR #2197, opened 2019-09-26
Add `experimentalForceOwningTab` for Web Worker support

thebrianchen:
Original issue: #983 
Provides an experimental feature to support offline persistence in web workers.

I've updated the sample app with the latest changes as well ([link](https://github.com/thebrianchen/firestore-web-worker)).

There also seem to be some changes from running prettier that probably weren't pushed in other commits.


packages/firestore/src/core/firestore_client.ts
```diff
@@ -343,7 +344,8 @@ export class FirestoreClient {
       ) {
         throw new FirestoreError(
           Code.UNIMPLEMENTED,
-          'IndexedDB persistence is only available on platforms that support LocalStorage.'
+          'IndexedDB persistence with synchronizedTabs is only available on ' +
```
mikelehen: synchronizedTabs => synchronizeTabs

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -281,13 +284,14 @@ export class IndexedDbPersistence implements Persistence {
...
-      this.webStorage = this.window.localStorage;
+      this.webStorage = platform.window.localStorage;
     } else {
-      throw new FirestoreError(
-        Code.UNIMPLEMENTED,
-        'IndexedDB persistence is only available on platforms that support LocalStorage.'
+      this.webStorage = null;
+      log.error(
+        LOG_TAG,
+        'LocalStorage is unavailable. Multi-tab functionality may not be supported.'
```
mikelehen: They have not enabled multi-tab (since we would have thrown a different error already if they did).  So I think this error should say, 'LocalStorage is unavailable. As a result, persistence may not work reliably. In particular enablePersistence() could fail immediately after refreshing the page.'

or something like that (maybe we should tell them they can enable `force`?  Not sure...  Probably best to leave it ambiguous and then if we get feedback, find out how people are hitting this and what advice we should be giving them.)

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -281,13 +284,14 @@ export class IndexedDbPersistence implements Persistence {
...
     if (platform.window && platform.window.localStorage) {
-      this.window = platform.window;
-      this.webStorage = this.window.localStorage;
+      this.webStorage = platform.window.localStorage;
     } else {
-      throw new FirestoreError(
-        Code.UNIMPLEMENTED,
-        'IndexedDB persistence is only available on platforms that support LocalStorage.'
+      this.webStorage = null;
+      log.error(
```
mikelehen: I think maybe we should only log the error if force is false, because if force==true I don't think it matters that localStorage is not available.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -526,9 +529,11 @@ export class IndexedDbPersistence implements Persistence {
       // the client atomically, but we can't. So we opt to delete the IndexedDb
       // entries first to avoid potentially reviving a zombied client.
       inactiveClients.forEach(inactiveClient => {
-        this.window.localStorage.removeItem(
-          this.zombiedClientLocalStorageKey(inactiveClient.clientId)
-        );
+        if (this.webStorage) {
```
mikelehen: nit: I would do this check once on the outside of the for-loop rather than repeat it on every iteration.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -590,6 +595,9 @@ export class IndexedDbPersistence implements Persistence {
           }
 
           if (!this.isLocalClient(currentPrimary)) {
+            if (this.force) {
+              return true;
+            }
```
mikelehen: I'm not sure this is the right place to do this.  We only want to respect `force` on initial client startup.  Else, if you launch two tabs with `force`: true, I think they could both end up broken because they both steal the lease from the other one or something (it's kind of race-y so I'm not 100% sure if it would happen in practice).

So I think maybe we want to add a `force` _parameter_ to updateClientMetadataAndTryBecomePrimary() and then if it's true we can just skip the canActAsPrimary() check.  And we will pass true only from IndexedDbPersistence.start() if settings.force == true.

packages/firestore/src/local/simple_db.ts
```diff
@@ -66,7 +66,14 @@ export class SimpleDb {
       // suggests IE9 and older WebKit browsers handle upgrade
       // differently. They expect setVersion, as described here:
       // https://developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest/setVersion
-      const request = window.indexedDB.open(name, version);
+
+      let request: IDBOpenDBRequest;
+      // Support Web Workers, which don't have `window` global variable.
+      if (indexedDB !== undefined) {
```
mikelehen: AFAIK the only way to safely check for the existence of a global is `typeof indexedDB !== 'undefined'`

https://imgur.com/a/28vgzqX

So your code here only works because we've already (safely) checked for the existence of the global in isAvailable().

packages/firestore/src/local/simple_db.ts
```diff
@@ -66,7 +66,14 @@ export class SimpleDb {
...
+        request = indexedDB.open(name, version);
+      } else {
+        request = window.indexedDB.open(name, version);
```
mikelehen: Since `window` is _usually_ the global object and we've already checked for the `indexedDB` global directly, is there a case where we will ever hit this?

Maybe we have a fake `window` object in node tests or something?  Do you know?

packages/firestore/src/local/simple_db.ts
```diff
@@ -132,6 +139,10 @@ export class SimpleDb {
 
   /** Returns true if IndexedDB is available in the current environment. */
   static isAvailable(): boolean {
+    // If IndexedDb is available directly (ex: in case of web workers).
+    if (typeof indexedDB !== 'undefined') {
+      return true;
+    }
     if (typeof window === 'undefined' || window.indexedDB == null) {
```
mikelehen: Per above, since `window` is _usually_ the global object and we've already checked for the indexedDB global directly, is there a case where we will ever hit this?

packages/firebase/index.d.ts
```diff
@@ -7768,10 +7768,23 @@ declare namespace firebase.firestore {
...
+     * with `synchronizeTabs:true` and is primarily intended for use with Web
+     * Workers. Setting this to 'true' will enable persistence, but cause other
+     * tabs using persistence to fail.
+     *
+     * This setting may be removed in a future release. If you find yourself
+     * using it for a specific use case or run into any issues, please tell us
+     * about it in
+     * https://github.com/firebase/firebase-js-sdk/issues/983.
+     */
+    experimentalForceOwningTab?: boolean;
```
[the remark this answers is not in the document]
schmidt-sebastian: The use case might be to pin an owner tab. Not sure why that might be useful, but it would avoid primary to secondary tab transition which AFAIK haven't been a problem.

As for the name:

To me, `WebWorker` is the umbrella term, but I might be mistaken (and propose `experimentalBackgroundWorkerPersistence` instead).

My main problem with the API name is that `experimentalForceOwningTab` only says what it does but doesn't say why it does it or what feature it unblocks. The fact that it enables WebWorker/SharedWorker/ServiceWorker persistence is not obvious from the name, and it isn't even document in the API description. The only place where it shows up is in the title of the issue that is linked. A user that wants to use Firestore in a WebWorker has no obvious way to discover that this is the setting they need to enable.

If you are happy with the name, feel free to ignore. This has been discussed for quite some time.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -981,6 +1001,9 @@ export class IndexedDbPersistence implements Persistence {
   }
 
   private detachWindowUnloadHook(): void {
+    if (!this.window) {
```
schmidt-sebastian, line 1004: This could be an assert in the `if` block below.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -963,6 +980,9 @@ export class IndexedDbPersistence implements Persistence {
    * handler.
    */
   private attachWindowUnloadHook(): void {
+    if (!this.window) {
+      return;
+    }
     if (typeof this.window.addEventListener === 'function') {
```
schmidt-sebastian, line 986: Not sure what effect this has on code size, but you could remove some of extra lines in this PR but using optional chaining:

```
if (typeof this.window?.addEventListener === 'function') {}
this.webStorage?.setValue(...)
```

packages/firestore/src/local/simple_db.ts
```diff
@@ -78,7 +78,7 @@ export class SimpleDb {
       // suggests IE9 and older WebKit browsers handle upgrade
       // differently. They expect setVersion, as described here:
       // https://developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest/setVersion
-      const request = window.indexedDB.open(name, version);
+      const request = indexedDB.open(name, version);
```
schmidt-sebastian, line 81: I am surprised this works with the IndexedDB shim. We only register on `window.indexedDB`:

https://github.com/firebase/firebase-js-sdk/blob/631a0ec5982b36382a96c15be9916bf29f98a01f/packages/firestore/test/util/test_platform.ts#L57

The CI is green though.

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: I think we just need to register the IndexedDb shim directly on `globalAny` in `node_persistence.ts`.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1108,6 +1136,12 @@ describe('IndexedDb: canActAsPrimary', () => {
       expect(await getCurrentLeaseOwner()).to.not.be.null;
     });
   });
+
+  it('obtains lease if forceOwningTab is set', () => {
+    return withPersistence('clientA', async (persistence, platform, queue) => {
+      await withForcedPersistence('clientB', db2 => Promise.resolve());
```
schmidt-sebastian, line 1142: Can we verify somehow that `clientA` is not useable anymore? We could invoke `runTransaction` or something.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1005,11 +1024,12 @@ export class IndexedDbPersistence implements Persistence {
 
   private detachWindowUnloadHook(): void {
     if (this.windowUnloadHandler) {
+      assert(this.window != null, "'window' should be defined");
```
schmidt-sebastian, line 1027: We only use `debugAssert` and `hardAssert` and not the asserts from Node. This should be a `debugAssert` since this would be an SDK bug (versus an external bug from a hardAssert). I do think you can combine it with the assert below though:

```
debugAssert(typeof this.window?.removeEventListener === 'function',	    ...   );
```

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1127,6 +1156,18 @@ describe('IndexedDb: canActAsPrimary', () => {
       expect(await getCurrentLeaseOwner()).to.not.be.null;
     });
   });
+
+  it('obtains lease if forceOwningTab is set', () => {
+    return withPersistence('clientA', async db => {
+      await withForcedPersistence('clientB', async () => {
+        return expect(
+          db.runTransaction('tx', 'readwrite-primary', () =>
```
schmidt-sebastian, line 1164: s/db/some descriptive name that makes it clear this is from the prior instance/

packages/firestore/src/local/simple_db.ts
schmidt-sebastian: Doesn't that mean that "if (typeof window === 'undefined' || window.indexedDB == null)" is now no longer needed? Is your distinction from your first response here still relevant? if so, we should probably talk this through offline.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -239,7 +242,8 @@ export class IndexedDbPersistence implements Persistence {
     lruParams: LruParams,
     private readonly queue: AsyncQueue,
     serializer: JsonProtoSerializer,
-    private readonly sequenceNumberSyncer: SequenceNumberSyncer
+    private readonly sequenceNumberSyncer: SequenceNumberSyncer,
+    private readonly forceOwningTab: boolean
```
schmidt-sebastian, line 246: Do you mind adding a short inline comment for this and `allowTabSynchronization`? The rest of the arguments seem obvious.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -261,14 +265,19 @@ export class IndexedDbPersistence implements Persistence {
...
-      this.window = platform.window;
-      this.webStorage = this.window.localStorage;
+      this.webStorage = platform.window.localStorage;
     } else {
-      throw new FirestoreError(
-        Code.UNIMPLEMENTED,
-        'IndexedDB persistence is only available on platforms that support LocalStorage.'
-      );
+      this.webStorage = null;
+      if (forceOwningTab === false) {
```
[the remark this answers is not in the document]
thebrianchen: We now throw an error if `synchronizeTabs` is used with `experimentalForceOwningTab`, and we throw an error if LocalStorage is unavailable in WebStorageSharedClientState. Which case are we not covering?

packages/firestore/src/local/indexeddb_persistence.ts
thebrianchen: done.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -585,6 +585,9 @@ export class IndexedDbPersistence implements Persistence {
   private canActAsPrimary(
     txn: PersistenceTransaction
   ): PersistencePromise<boolean> {
+    if (this.forceOwningTab) {
+      return PersistencePromise.resolve<boolean>(true);
```
schmidt-sebastian, line 589: You can probably drop the generic since it can be inferred.

thebrianchen: Not sure why TS can't infer it, but the error I get is:
`Type 'PersistencePromise<true>' is not assignable to type 'PersistencePromise<boolean>'.`

packages/firebase/index.d.ts
```diff
@@ -7776,10 +7776,23 @@ declare namespace firebase.firestore {
...
-     * @deprecated This setting is deprecated. To enabled synchronization between
+     * @deprecated This setting is deprecated. To enable synchronization between
      * multiple tabs, please use `synchronizeTabs: true` instead.
      */
     experimentalTabSynchronization?: boolean;
+
+    /**
+     * Whether to force enable persistence for the client. This cannot be used
+     * with `synchronizeTabs:true` and is primarily intended for use with Web
+     * Workers. Setting this to 'true' will enable persistence, but cause other
```
egilmorez, line 7787: Was the intention to use backticks around `true` ?  That would work.

thebrianchen: done.

config/functions/project.json
```diff
@@ -0,0 +1,9 @@
+ {
```
Feiyang1, line 1: Do you mean to commit this file?

thebrianchen: thanks for catching

PR #2198, opened 2019-09-26
Make onSnapshotsInSync() public

thebrianchen:


packages/firebase/index.d.ts
```diff
@@ -7209,6 +7209,43 @@ declare namespace firebase.firestore {
      */
     waitForPendingWrites(): Promise<void>;
 
+    /**
```
mikelehen: extra whitespace

packages/firebase/index.d.ts
```diff
@@ -7209,6 +7209,43 @@ declare namespace firebase.firestore {
...
+   * with each other, but does not relate to whether those snapshots are in sync
+   * with the server. Use SnapshotMetadata in the individual listeners to
+   * determine if a snapshot is from the cache or the server.
+   *
+   * @param onSync A callback to be called every time all snapshot listeners are
+   * in sync with each other.
+   * @return An unsubscribe function that can be called to cancel the snapshot
+   * listener.
+   */
+  onSnapshotsInSync(onSync: () => void): () => void;
```
mikelehen: So for onSnapshot() we seem to have 4 overloads but a single doc comment describing the method.  Can we do that here too?

packages/firestore/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
   small subset of the documents in a collection.
 - [fixed] Fixed a race condition between authenticating and initializing
   Firestore that could result in initial writes to the database being dropped.
+- [feature] Added a `Firestore.onSnapshotsInSync()` method that notifies users
+  when all their snapshot listeners are in sync with each other.
```
mikelehen: nit: We use 2nd-person for some of the other changelog entries and I think that's probably a bit cleaner.  In particular, "users" is always a little problematic since our _users_ are the developers that are writing apps which have their own _users_.  So it can sometimes be ambiguous (not a big deal in this case probably).

> Added a `Firestore.onSnapshotsInSync()` method that notifies you when all of your snapshot listeners are in sync with each other.

packages/firebase/index.d.ts
thebrianchen: Where is that? I saw that each `onSnapshot()` [link](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L7758) had it own doc comment so that's what I did as well.

mikelehen: OHHH...  Interesting.  I was looking at firestore-types/index.d.ts instead of firebase/index.d.ts.  see here: https://github.com/firebase/firebase-js-sdk/blob/66655d0ba42c2ebf153bd950d73612f0a1533c7f/packages/firestore-types/index.d.ts#L1174

I thought our types were basically in-sync in both files, but I guess not.  Hrm.  I'm guessing @hsubox76 had to separate them when transitioning our public docs to be generated from the .d.ts files...  in which case we should probably update firestore-types/index.d.ts to match.  If the files are out-of-sync now, they'll just get worse over time.

PR #2202, opened 2019-09-27
Fixed default timeout for uploads

samuelgozi:
It seems like there was a missing zero for the milliseconds, which resulted in the max time being one minute instead of ten.


schmidt-sebastian:
Thanks for this contribution! This will be included in the next release.

Issue #2204, opened 2019-09-27
Error thrown when empty map is present in firestore document


david0178418:
### [REQUIRED] Describe your environment

  * Operating System version: Kubuntu 18.10
  * Browser version: Chrome 77.0.3865.90 (Official Build) (64-bit)
  * Firebase SDK version: 6.6.0
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

Error produced when empty map is present.  `Document.getProtoField` incorrectly assumes a "fields" prop is present in mapValue.  I don't know if this assumption is correct or if the mapValue should always have a "fields" property even if it is empty.

Method first appeared in [firebase v6.6.0](https://github.com/firebase/firebase-js-sdk/compare/firebase@6.5.0...firebase@6.6.0#diff-65780527e92e17915504f19ee06e6ffbR167) as part of https://github.com/firebase/firebase-js-sdk/pull/2115

#### Steps to reproduce:
1) Create a document containing an empty map
2) Load the document in the firebase sdk

Expected Behavior: Loads data without error
Actual: Crashes

Adding a dummy prop works around the issue.

#### Relevant Code:

![image](https://user-images.githubusercontent.com/128739/65805073-d68df300-e149-11e9-850b-ce3680752e56.png)

Full error:

```Uncaught TypeError: Cannot read property 'FOO' of undefined
    at Document.push../node_modules/@firebase/firestore/dist/index.cjs.js.Document.getProtoField (index.cjs.js:5792)
    at Document.push../node_modules/@firebase/firestore/dist/index.cjs.js.Document.field (index.cjs.js:5732)
    at FieldFilter.push../node_modules/@firebase/firestore/dist/index.cjs.js.FieldFilter.matches (index.cjs.js:14503)
    at Query.push../node_modules/@firebase/firestore/dist/index.cjs.js.Query.matchesFilters (index.cjs.js:14378)
    at Query.push../node_modules/@firebase/firestore/dist/index.cjs.js.Query.matches (index.cjs.js:14303)
    at index.cjs.js:6297
    at IDBRequest.cursorRequest.onsuccess [as __zone_symbol__ON_PROPERTYsuccess] (index.cjs.js:4121)
    at IDBRequest.wrapFn (zone-evergreen.js:1202)
    at ZoneDelegate.invokeTask (zone-evergreen.js:400)
    at Zone.runTask (zone-evergreen.js:168)
[...]

PR #2206, opened 2019-09-28
Treat empty maps correctly in Document.get()

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-js-sdk/issues/2204

This fixes the types for MapValue.fields  so that tslint catches undefined violations and removes the violation in `getProtoField` that caused https://github.com/firebase/firebase-js-sdk/issues/2204

Note that is seems pretty involved to write a test for this since this code is only hit in Query processing and not in DocumentSnapshot.get() (see https://github.com/firebase/firebase-js-sdk/blob/4016987a72fa87eea5171dd718d20f9e5dde8123/packages/firestore/src/api/database.ts#L1346). If requested, I can add this test next week.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,4 +1,9 @@
-# Unreleased (1.5.0)
+# Unreleased
+- [fixed] Fixed a regression that caused queries with nested field filters to
+  crash the client if the field was not present in the local copy of the
+  document. 
+
+# 1.5.0
 - [feature] Added a `Firestore.waitForPendingWrites()` method that
   allows users to wait until all pending writes are acknowledged by the
   Firestore backend.
--- a/packages/firestore/src/model/document.ts
+++ b/packages/firestore/src/model/document.ts
@@ -120,7 +120,7 @@ export class Document extends MaybeDocument {
   data(): ObjectValue {
     if (!this.objectValue) {
       let result = ObjectValue.EMPTY;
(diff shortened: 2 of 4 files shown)


schmidt-sebastian:
I fixed the browser tests and was now able to write a test that fails on master and succeeds with this PR. I did not add any inline comments as the types should speak for themselves.

As for your questions - I don't think `this.proto.fields` should ever be missing (but the types allow for it now, since `this.proto` is a `fields?: ApiClientObjectMap<Value>`. The issue the customer ran into is a nested empty map that did not contain the `fields` property when read back from IndexedDb (despite https://github.com/firebase/firebase-js-sdk/blob/c7e501259eff8db6a576df36da9c2396a6d28fa1/packages/firestore/src/remote/serializer.ts#L563).

schmidt-sebastian:
I fixed the browser tests and was now able to write a test that fails on master and succeeds with this PR. I did not add any inline comments as the types should speak for themselves.

As for your questions - I don't think `this.proto.fields` should ever be missing (but the types allow for it now, since `this.proto` is a `fields?: ApiClientObjectMap<Value>`. The issue the customer ran into is a nested empty map that did not contain the `fields` property when read back from IndexedDb (despite https://github.com/firebase/firebase-js-sdk/blob/c7e501259eff8db6a576df36da9c2396a6d28fa1/packages/firestore/src/remote/serializer.ts#L563).

PR #2212, opened 2019-09-30
Allow leading zeros in tryParseInt

schmidt-sebastian:
Fixes https://buganizer.corp.google.com/issues/141298281

A customer noticed that we don't correctly order their documents. Some of their documents are using keys of the form "00x", where x is a 9 digit number. Our number check only allows for ten digits, and leading zeros can break this limit. This PR relaxes the check and matches the backend semantics.

### DIFF
--- a/packages/database/test/path.test.ts
+++ b/packages/database/test/path.test.ts
@@ -20,8 +20,8 @@ import { Path } from '../src/core/util/Path';
 
 describe('Path Tests', function() {
   const expectGreater = function(left, right) {
-    expect(Path.comparePaths(new Path(left), new Path(right))).to.equal(1);
-    expect(Path.comparePaths(new Path(right), new Path(left))).to.equal(-1);
+    expect(Path.comparePaths(new Path(left), new Path(right))).to.be.greaterThan(0)
+    expect(Path.comparePaths(new Path(right), new Path(left))).to.be.lessThan(0);
   };
 
   const expectEqual = function(left, right) {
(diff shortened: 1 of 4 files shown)


packages/database/test/path.test.ts
```diff
@@ -117,5 +117,6 @@ describe('Path Tests', function() {
     expectGreater('/ab', '/a');
     expectGreater('/a/b', '/a');
     expectGreater('/a/b', '/a//');
+    expectGreater('/a/0971500000', '/a/00403311635');
```
jsdt:
```suggestion
    expectGreater('/a/0971500000', '/a/00403311635');
    expectGreater('/a/0971500000', '/a/971500000');
```

Does this handle that case correctly (when both parse to the same Int, we fall back to comparing the length of the strings)?

schmidt-sebastian: I think it should, but let's find out :)

packages/database/test/path.test.ts
```diff
@@ -20,8 +20,8 @@ import { Path } from '../src/core/util/Path';
 
 describe('Path Tests', function() {
   const expectGreater = function(left, right) {
-    expect(Path.comparePaths(new Path(left), new Path(right))).to.equal(1);
-    expect(Path.comparePaths(new Path(right), new Path(left))).to.equal(-1);
+    expect(Path.comparePaths(new Path(left), new Path(right))).to.be.greaterThan(0)
```
mikelehen: missing semicolon?

schmidt-sebastian: Well, on GitHub it is missing. But I swear that I have this locally (via the Prettier commit that once again got stuck somewhere...)

schmidt-sebastian:
CI is passing now after a re-run.

PR #2214, opened 2019-09-30
Split `Query` into `Query` and `Target`.

wu-hui:
Splitting `Query` and `Target` to serve different purpose: `Query` for queries with local features, `Target` for queries sent to backend.

It's probably best to review this in the commit order, the most significant commits come first, the bulk renaming commits come after those.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -27,158 +27,103 @@ import {
 import { FieldPath, ResourcePath } from '../model/path';
 import { assert, fail } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
-import { isNullOrUndefined } from '../util/types';
+import { Target } from './target';
 
+/**
+ * Represents a local query: queries with features implemented by the SDK on top
+ * of server implemented features.
+ */
 export class Query {
   static atPath(path: ResourcePath): Query {
-    return new Query(path);
+    return new Query(new Target(path));
   }
-
-  private memoizedCanonicalId: string | null = null;
-  private memoizedOrderBy: OrderBy[] | null = null;
 
   /**
    * Initializes a Query with a path and optional additional query constraints.
    * Path must currently be empty if this is a collection group query.
    */
-  constructor(
-    readonly path: ResourcePath,
-    readonly collectionGroup: string | null = null,
-    readonly explicitOrderBy: OrderBy[] = [],
-    readonly filters: Filter[] = [],
-    readonly limit: number | null = null,
-    readonly startAt: Bound | null = null,
-    readonly endAt: Bound | null = null
-  ) {
-    if (this.startAt) {
-      this.assertValidBound(this.startAt);
-    }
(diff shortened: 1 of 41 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -27,158 +27,103 @@ import {
...
+import { Target } from './target';
 
+/**
+ * Represents a local query: queries with features implemented by the SDK on top
+ * of server implemented features.
+ */
 export class Query {
   static atPath(path: ResourcePath): Query {
-    return new Query(path);
+    return new Query(new Target(path));
```
mikelehen: Per chat conversation, rather than have `Query` _wrap_ `Target` I think they should perhaps be 100% separate, with `Target` being a trivial class, perhaps something like:

```
export class Target {
  constructor(
    readonly path: ResourcePath,
    readonly collectionGroup: string | null = null,
    readonly explicitOrderBy: OrderBy[] = [],
    readonly filters: Filter[] = [],
    readonly limit: number | null = null,
    readonly startAt: Bound | null = null,
    readonly endAt: Bound | null = null
  ) { };
}
```

We might end up needing a few methods as well, but in particular, we should _not_ need to "build up" a Target by calling `.addFilter()`, `.withStartAt()`, etc.  Instead `Query.toTarget()` should just construct the `Target` on demand (or for perf it could create it once and then cache it if we expect toTarget() to be called repeatedly).

I don't feel super strongly about this, but I _think_ it'll be easier to reason about the types if we separate them completely (so `Query` isn't a wrapper) and make `Target` just a very simple type that's (basically) only used for serializing to/from the backend and persistence.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -240,19 +240,19 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
    */
   private async initializeViewAndComputeSnapshot(
-    queryData: QueryData,
+    targetData: TargetData,
     current: boolean
   ): Promise<ViewSnapshot> {
-    const query = queryData.query;
-    const docs = await this.localStore.executeQuery(query);
+    const target = targetData.target;
+    const docs = await this.localStore.executeTargetQuery(target);
```
mikelehen: Per chat conversation, I think we should never "execute" a `Target` because `Target` represents a server query and so we can send it to the backend and get the server results, or we can look up cached server results from the targetDocuments table in persistence, but we should not execute it directly.  Instead our local query execution should happen on `Query` objects.

That may mean we need to refactor this method so it takes a `Query` and a `targetId` instead of `TargetData`.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1053,7 +1061,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       // removed if it has been rejected by the backend.
       if (queryView) {
         await this.localStore
```
edd0408: R

PR #2216, opened 2019-09-30
Fix Firestore Browser tests

schmidt-sebastian:
This is the set of changes I needed to run `yarn test:browser` in packages/firestore.

Notes:
- `forEachType` is unused so I optimized for code size.
- I don't know how to properly depend on the newest set of Firestore types, which are needed to expose `onSnapshotsInSync` (cc @Feiyang1)

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -30,7 +30,7 @@
     "dist"
   ],
   "dependencies": {
-    "@firebase/firestore-types": "1.5.0",
+    "@firebase/firestore-types": "file:../firestore-types/",
     "@firebase/logger": "0.1.25",
     "@firebase/webchannel-wrapper": "0.2.26",
     "@grpc/proto-loader": "^0.5.0",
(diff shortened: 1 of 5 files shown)


packages/firestore/package.json
```diff
@@ -30,7 +30,7 @@
     "dist"
   ],
   "dependencies": {
-    "@firebase/firestore-types": "1.5.0",
+    "@firebase/firestore-types": "file:../firestore-types/",
```
Feiyang1: No need to manually change package version. Lerna will make the new types available to all packages in the workspace in our dev environment.

Package versions are automatically updated by the release script during a release.

mikelehen: I agree in principle, but in practice, `yarn test:browser` is failing from packages/firestore. 😄 It seems like with our current setup we have a period of brokenness between when we check in a change to firestore-types and when we do the next release.  Any suggestions?

PR #2223, opened 2019-10-01
rxfire: Support collection groups & a minor change detection fix

jamesdaniels:
* rxfire was comparing documents based off ids, this was fine before collection group queries. Here I've switched to `ref.isEqual` like I have done in AngularFire.
* When a document is modified, we were only replacing the single doc in the array; this would not always trip change detection in frameworks as `===` would still equal true on the top level. Splicing it creates a new array, which will fail `===` and trip change detection.
* Further, this was an issue with the other operations as we were using splice. I'm now slicing to clone the array before modifying.
* NIT: Visual Studio Code's settings, which are committed, were not setup to track the workspace's version of Typescript. This tripped me up a little at first.

### DIFF
--- a/packages/rxfire/firestore/collection/index.ts
+++ b/packages/rxfire/firestore/collection/index.ts
@@ -88,7 +88,7 @@ function processIndividualChange(
           combined.splice(change.oldIndex, 1);
           combined.splice(change.newIndex, 0, change);
         } else {
-          combined[change.newIndex] = change;
+          combined.splice(change.newIndex, 1, change);
         }
       }
       break;
--- a/packages/rxfire/firestore/collection/index.ts
+++ b/packages/rxfire/firestore/collection/index.ts
@@ -18,7 +18,7 @@
 import { firestore } from 'firebase/app';
 import { fromCollectionRef } from '../fromRef';
 import { Observable, MonoTypeOperatorFunction } from 'rxjs';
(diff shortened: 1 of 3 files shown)


packages/rxfire/firestore/collection/index.ts
```diff
@@ -80,22 +80,22 @@ function processIndividualChange(
...
+        combined[change.oldIndex].doc.ref.isEqual(change.doc.ref)
       ) {
         // When an item changes position we first remove it
         // and then add it's new position
         if (change.oldIndex !== change.newIndex) {
           combined.splice(change.oldIndex, 1);
           combined.splice(change.newIndex, 0, change);
         } else {
-          combined[change.newIndex] = change;
+          combined.splice(change.newIndex, 1, change);
```
Feiyang1: Splicing doesn't create a new array, but modify the existing array. Do you mean to use slice() ?

jamesdaniels: Derp, yes, good catch. Yes I meant to use slice... further it should be used in the other spots too. Will make that change, thanks @Feiyang1

jamesdaniels: I'll write some tests for this while I'm at it.

jamesdaniels:
FYI @davideast @jhuleatt

PR #2228, opened 2019-10-01
Send raw values when using WebChannel

thebrianchen:
As requested by @yuchenshi

### DIFF
--- a/packages/firestore/src/remote/serializer.ts
+++ b/packages/firestore/src/remote/serializer.ts
@@ -167,12 +167,13 @@ export class JsonProtoSerializer {
    * our generated proto interfaces say Int32Value must be. But GRPC actually
    * expects a { value: <number> } struct.
    */
-  private toInt32Value(val: number | null): number | undefined {
-    if (!typeUtils.isNullOrUndefined(val)) {
+  private toInt32Value(val: number | null): number | null {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/remote/serializer.ts
```diff
@@ -167,12 +167,13 @@ export class JsonProtoSerializer {
    * our generated proto interfaces say Int32Value must be. But GRPC actually
    * expects a { value: <number> } struct.
    */
-  private toInt32Value(val: number | null): number | undefined {
-    if (!typeUtils.isNullOrUndefined(val)) {
+  private toInt32Value(val: number | null): number | null {
```
mikelehen: You changed the return type (to `null`) but need to update the comment to match.

thebrianchen: oops, good catch!

yuchenshi:
I think we can totally test this against the old Firestore emulator v1.7.1: https://github.com/firebase/firebase-tools/commit/b1ee5c29001dc498a56556638e52cdfc37d910d7#diff-e1b9303f298bab6ec7fe10ee0fad137bR43

yuchenshi:
Talked offline. I'll do a manual test and get back to you

PR #2229, opened 2019-10-02
Fixing the RemoteConfig main entry

jamesdaniels:
Hey @Feiyang1, not sure the intent but looks like RC's main entry was wrong; referencing a non-existent file.


Feiyang1:
Oops! Thanks for flagging it!

PR #2239, opened 2019-10-04
Add the forceRefresh parameter to Installations' getToken method

mmermerkaya:


packages/installations/src/functions/get-token.ts
```diff
@@ -23,14 +23,17 @@ import { AppConfig } from '../interfaces/app-config';
...
+  forceRefresh?: boolean
+): Promise<string> {
   const appConfig = extractAppConfig(app);
 
   await completeInstallationRegistration(appConfig);
 
   // At this point we either have a Registered Installation in the DB, or we've
   // already thrown an error.
-  return refreshAuthToken(appConfig);
+  return refreshAuthToken(appConfig, forceRefresh);
```
dwoffinden: If you don't pass forceRefresh to getToken, this'll be undefined, right? Tbh I'd rather be explicit, make it `= false` here and pass the boolean down? I didn't even realise that that worked with an explicit undefined parameter :/

packages/installations/src/functions/get-token.test.ts
```diff
@@ -226,7 +227,7 @@ describe('getToken', () => {
 
         it('resolves with an auth token', async () => {
           const token = await getToken(app);
-          expect(token).to.equal(AUTH_TOKEN);
+          expect(token).to.be.oneOf([AUTH_TOKEN, NEW_AUTH_TOKEN]);
```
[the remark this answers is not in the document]
mmermerkaya: It is deterministic, it's just that this test is looping through different situations and making sure that the function always returns a valid auth token, whether it's the one in the DB or it's a new one from the server.

packages/installations/src/functions/get-token.ts
mmermerkaya: Passing `undefined` explicitly or not passing anything work exactly the same in JS, so I decided to keep the default value definition in the actual function that uses the value, but if it does look that scary, I can change it. :smile:

Feiyang1:
@mmermerkaya  Seems we forgot to update `firebase/index.d.ts`. Can you please update it ?

PR #2240, opened 2019-10-04
Add support for Typescript Custom Mapping 

### LINKED ISSUE #1888: FR: Generic Document Types
### Describe your environment

  * Operating System version: N/A
  * Browser version: N/A
  * Firebase SDK version: 6.2.0
  * Firebase Product: Firestore

#### Steps to reproduce:

Currently:

```ts
let post = await firebase.firestore().collection('posts').doc('123').get();
post.data().title; // Ideally, should know this is a string
post.data().someNonExistentProperty; // Ideally, should error in Typescript
```

It would be extremely helpful if the Firestore APIs accepted generic type arguments to describe the shape of items in a given collection:

```ts
interface Post {
  title: string
}
let post = await firebase.firestore().collection<Post>('posts').doc('123').get();
post.data().title; // string
post.data().someNonExistentProperty; // TS error
```

I know I can cast the result of `.data()`:

```
post.data() as Post
```

But that's less than ideal, because:

1. I have to remember to cast each and every time, vs. being able to supply generic type args to the `.collection()` method could ensure types for all document references from that collection, and
2. Casting the `.data()` call to a type misses out on type improvements to `.set()` and other methods.

thebrianchen:
API proposal: [internal link](http://go/firestore-ts-generics)

Fixes #1888 via a `withConverter()` methods on `CollectionReference`, `DocumentReference`, and `Query` that allow developers to use Firestore classes with generics. 

Example usage:
```
class Post {
  constructor(readonly title: string, readonly author: string) {}
  byline(): string {
    return this.title + ', by ' + this.author;
  }
}

const PostConverter = {
  toFirestore(post: Post): DocumentData {
    return {title: post.title, author: post.author};
  },
  fromFirestore(snapshot: QueryDocumentSnapshot, options: SnapshotOptions): Post {
    const data = snapshot.data(options)!;
    return new Post(data.title, data.author);
  }
};

const postSnap = await firebase.firestore().collection('posts').withConverter(PostConverter).doc().get();
const post = postSnap.data();
if (post !== undefined) {
  post.title; // string
  post.byline(); // Should be defined 
  post.someNonExistentProperty; // TS error
}
```


packages/firestore-types/index.d.ts
```diff
@@ -181,7 +192,11 @@ export class FirebaseFirestore {
    * @param documentPath A slash-separated path to a document.
    * @return The `DocumentReference` instance.
    */
-  doc(documentPath: string): DocumentReference;
+  doc(documentPath: string): DocumentReference<DocumentData>;
+  doc<T>(
```
schmidt-sebastian: I don't think this should be typed. You can have a typed collection that has a typed DocumentReferences, but if you call `doc()`  from an untyped Collection or FirebaseFirestore, my gut feeling is that it is better not to allow types.

packages/firestore-types/index.d.ts
```diff
@@ -542,7 +561,7 @@ export class Transaction {
    * within the document.
    * @return This `Transaction` instance. Used for chaining method calls.
    */
-  update(documentRef: DocumentReference, data: UpdateData): Transaction;
+  update(documentRef: DocumentReference<any>, data: UpdateData): Transaction;
```
schmidt-sebastian: I think this could be "unknown"

packages/firestore-types/index.d.ts
```diff
@@ -560,7 +579,7 @@ export class Transaction {
    * to the backend (Note that it won't resolve while you're offline).
    */
   update(
-    documentRef: DocumentReference,
+    documentRef: DocumentReference<any>,
```
schmidt-sebastian: Same.

packages/firestore-types/index.d.ts
```diff
@@ -572,7 +591,7 @@ export class Transaction {
    * @param documentRef A reference to the document to be deleted.
    * @return This `Transaction` instance. Used for chaining method calls.
    */
-  delete(documentRef: DocumentReference): Transaction;
+  delete(documentRef: DocumentReference<any>): Transaction;
```
schmidt-sebastian: Also unknown here.

packages/firestore-types/index.d.ts
```diff
@@ -616,7 +635,7 @@ export class WriteBatch {
    * within the document.
    * @return This `WriteBatch` instance. Used for chaining method calls.
    */
-  update(documentRef: DocumentReference, data: UpdateData): WriteBatch;
+  update(documentRef: DocumentReference<any>, data: UpdateData): WriteBatch;
```
schmidt-sebastian: s/any/unknown

packages/firestore-types/index.d.ts
```diff
@@ -633,7 +652,7 @@ export class WriteBatch {
    * to the backend (Note that it won't resolve while you're offline).
    */
   update(
-    documentRef: DocumentReference,
+    documentRef: DocumentReference<any>,
```
schmidt-sebastian: s/any/unknown

packages/firestore-types/index.d.ts
```diff
@@ -645,7 +664,7 @@ export class WriteBatch {
    * @param documentRef A reference to the document to be deleted.
    * @return This `WriteBatch` instance. Used for chaining method calls.
    */
-  delete(documentRef: DocumentReference): WriteBatch;
+  delete(documentRef: DocumentReference<any>): WriteBatch;
```
schmidt-sebastian: s/any/unknown

packages/firestore-types/index.d.ts
```diff
@@ -744,30 +763,39 @@ export class DocumentReference {
...
   /**
    * Gets a `CollectionReference` instance that refers to the collection at
    * the specified path.
    *
    * @param collectionPath A slash-separated path to a collection.
    * @return The `CollectionReference` instance.
    */
-  collection(collectionPath: string): CollectionReference;
+  collection(collectionPath: string): CollectionReference<DocumentData>;
+  collection<C>(
```
schmidt-sebastian: Maybe use R here - it is more common to use R,S,T in generics.

packages/firestore-types/index.d.ts
```diff
@@ -744,30 +763,39 @@ export class DocumentReference {
...
+  ): CollectionReference<C>;
 
   /**
    * Returns true if this `DocumentReference` is equal to the provided one.
    *
    * @param other The `DocumentReference` to compare against.
    * @return true if this `DocumentReference` is equal to the provided one.
    */
-  isEqual(other: DocumentReference): boolean;
+  isEqual(other: DocumentReference<T>): boolean;
```
schmidt-sebastian: How do you think we should handle DocumentReference<A>.isEqual(DocumentReference<B>) when both point to the same document? Does Typescipt allow us to do an equal check on the type of the Generic?

packages/firestore-types/index.d.ts
```diff
@@ -971,7 +999,8 @@ export class DocumentSnapshot {
    * @return An Object containing all fields in the document or 'undefined' if
    * the document doesn't exist.
    */
-  data(options?: SnapshotOptions): DocumentData | undefined;
+  // data(options?: SnapshotOptions): DocumentData | undefined;
```
schmidt-sebastian: Please remove.

packages/firestore-types/index.d.ts
```diff
@@ -996,7 +1025,7 @@ export class DocumentSnapshot {
    * @param other The `DocumentSnapshot` to compare against.
    * @return true if this `DocumentSnapshot` is equal to the provided one.
    */
-  isEqual(other: DocumentSnapshot): boolean;
+  isEqual(other: DocumentSnapshot<T>): boolean;
```
schmidt-sebastian: Same question regarding isEqual.

mikelehen:
FYI- I took the luxury of editing your example in the PR description to make the converter its own `PostConverter` object since I think that's a bit cleaner and probably closer to expected usage (we expect people to want to re-use the same converter in multiple codepaths).

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -889,4 +889,46 @@ apiDescribe('Database transactions', (persistence: boolean) => {
         });
     });
   });
+
+  // PORTING NOTE: These tests are for generics support and apply only to web.
+  apiDescribe('Generics support', (persistence: boolean) => {
```
mikelehen: same naming feedback.

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -889,4 +889,46 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
+    class Post {
+      constructor(readonly title: string, readonly author: string) {}
+      byline(): string {
+        return this.title + ', by ' + this.author;
+      }
+    }
+
+    it('can set and get a document with transactions', () => {
```
mikelehen: "with transactions" is redundant.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -858,7 +858,7 @@ apiDescribe('Database', (persistence: boolean) => {
         () => {
           throw new Error('Promise resolved even though error was expected.');
         },
-        err => {
+        (err: firestore.FirestoreError) => {
```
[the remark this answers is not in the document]
thebrianchen: When I first made the changes, it didn't compile unless I added the type, but it seems to work now...

packages/firestore/src/api/database.ts
```diff
@@ -592,14 +592,26 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     validateExactNumberOfArgs('Firestore.collection', arguments, 1);
     validateArgType('Firestore.collection', 'non-empty string', 1, pathString);
     this.ensureClientConfigured();
-    return new CollectionReference(ResourcePath.fromString(pathString), this);
+    return new CollectionReference(ResourcePath.fromString(pathString), this, {
+      toFirestore: value => value,
+      fromFirestore: (snapshot, options) =>
+        snapshot.data(options)! as firestore.DocumentData
```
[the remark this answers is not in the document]
thebrianchen: I'm leaning towards replacing `DocumentSnapshot` with `QueryDocumentSnapshot` in the converter, because it seems more straightforward and understandable. It's also easier to guarantee that we return `undefined` when `.data()` is in fact undefined, since the TS compiler will thrown an error if you try to access `.data()` without first making sure it's defined.

It seems to me that forcing the developer to convert non-existent DocumentSnapshots to custom model objects is more complicated for them as opposed to returning `undefined`.

packages/firestore/src/api/database.ts
```diff
@@ -1297,29 +1354,43 @@ class SnapshotMetadata implements firestore.SnapshotMetadata {
...
+      );
+      const snapshot = new DocumentSnapshot(
+        this._firestore,
+        this._key,
+        this._document,
+        this._fromCache,
+        this._hasPendingWrites,
+        {
+          toFirestore: value => value,
+          fromFirestore: _snapshot => documentData as T
```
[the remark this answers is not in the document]
thebrianchen: Wow, this is very clean! I did not consider the overhead of constructing a new DocumentSnapshot.

packages/firestore/src/api/database.ts
```diff
@@ -592,14 +592,26 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
     this.ensureClientConfigured();
-    return DocumentReference.forPath(ResourcePath.fromString(pathString), this);
+    return DocumentReference.forPath(
+      ResourcePath.fromString(pathString),
+      this,
+      {
+        toFirestore: (value: firestore.DocumentData) => value,
+        fromFirestore: (snapshot, options) =>
+          snapshot.data(options)! as firestore.DocumentData
+      }
```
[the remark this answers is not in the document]
thebrianchen: I ended up making the converter optional, in order to lazily construct the DocumentSnapshot only as necessary when calling `DocumentSnapshot.data()`. However, `Transaction.get<T>()` and `.set<T>()` require a typed converter that lives on the DocumentReference, so I added a `defaultConverter(): FirestoreDataConverter<T>` that acts like the default.

packages/firestore/src/api/database.ts
```diff
@@ -1297,29 +1354,43 @@ class SnapshotMetadata implements firestore.SnapshotMetadata {
...
-          this._document.data(),
-          FieldValueOptions.fromSnapshotOptions(
-            options,
-            this._firestore._areTimestampsInSnapshotsEnabled()
-          )
-        );
+    if (!this._document) {
+      return undefined;
+    } else {
+      const documentData = this.convertObject(
```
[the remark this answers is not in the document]
thebrianchen: Thoughts on `convertObjectValue()` or `convertObjectToDocumentData()`?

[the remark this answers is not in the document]
thebrianchen: Good point! Renamed to `convertObjectValue()`.

packages/firebase/index.d.ts
```diff
@@ -7087,6 +7087,13 @@ declare namespace firebase.firestore {
...
   export function setLogLevel(logLevel: LogLevel): void;
 
+  export interface FirestoreDataConverter<T> {
+    // Converts a model object of type T into plain Firestore DocumentData.
+    toFirestore(modelObject: T): DocumentData;
+    // Converts a Firestore DocumentSnapshot into a model object of type T.
+    fromFirestore(snapshot: DocumentSnapshot, options: SnapshotOptions): T;
+  }
+
   /**
```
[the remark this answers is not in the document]
thebrianchen: Wrote a basic checker for the sake of this, but I'll definitely explore an automated script that checks this. It's a little tricky right now because some names are different and some one-liner functions are spread out across multiple lines in firestore-types, but I will bother Gil and learn how to script.

packages/firestore-types/index.d.ts
thebrianchen: Since we can't check for the type of Generic, Sebastian proposed adding an additional check to make sure the converters are the same, in addition to DocumentKey and underlying Firestore instance.

packages/firestore-types/index.d.ts
thebrianchen: https://github.com/firebase/firebase-js-sdk/pull/2240#discussion_r350901467

packages/firebase/index.d.ts
```diff
@@ -7087,6 +7087,13 @@ declare namespace firebase.firestore {
    */
   export function setLogLevel(logLevel: LogLevel): void;
 
+  export interface FirestoreDataConverter<T> {
```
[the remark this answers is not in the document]
thebrianchen: Spoke with Sam -- I'll add Mark Arndt as a reviewer once this PR is more solidified.

packages/firestore/src/api/database.ts
```diff
@@ -1004,20 +1050,21 @@ export class DocumentReference implements firestore.DocumentReference {
...
             'DocumentReference.set',
-            value,
+            convertedValue,
             options.mergeFields
           )
         : this.firestore._dataConverter.parseSetData(
             'DocumentReference.set',
-            value
+            convertedValue
           );
```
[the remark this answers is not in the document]
thebrianchen: Added a custom message of `toFirestore() from DocumentReference.set` iff the user's converter was used.

packages/firestore/test/util/test_platform.ts
```diff
@@ -45,7 +45,8 @@ export class FakeWindow {
     });
     this.fakeIndexedDb =
       fakeIndexedDb ||
-      (typeof window !== 'undefined' && window.indexedDB) || null;
+      (typeof window !== 'undefined' && window.indexedDB) ||
```
thebrianchen: This file got submitted without linter changes, so this is just the result of running prettier.

thebrianchen:
When diffing the files, here's the sed command I've been running to delete comments, empty lines, and leading whitespaces:
```
sed -e '/*/d' -e '/^[[:space:]]*$/d' -e 's/^ *//g' firebase.ts > firebase.ts.tmp
```

packages/firestore/src/api/database.ts
```diff
@@ -1451,11 +1480,12 @@ export class DocumentSnapshot<T = firestore.DocumentData>
...
       (this._document === null
         ? other._document === null
-        : this._document.isEqual(other._document))
+        : this._document.isEqual(other._document)) &&
+      isEqualConverter(this._converter, other._converter)
     );
   }
 
-  private convertObject(
+  private convertObjectValue(
```
mikelehen: It would be nice to avoid overloading "convert"...  and try to choose different terminology for converting a FieldValue to plain JS vs converting plain JS to the user's custom model type.

Maybe:
convertValue => toJSValue()
convertObjectValue => toJSObject()
convertArray => toJSArray()

?

packages/firestore/test/util/api_helpers.ts
```diff
@@ -42,6 +42,7 @@ import {
   key,
   path as pathFrom
 } from './helpers';
+import { Provider, ComponentContainer } from '@firebase/component';
```
mikelehen: Can you verify this is expected?  Or is it a bad merge or something?

packages/firestore/src/api/database.ts
thebrianchen: Wow, this is much nicer!

packages/firestore/test/util/api_helpers.ts
thebrianchen: The only thing that was imported was the `defaultDocumentDataConverter`, which prompted the linter to split these into separate lines.

packages/firestore/src/api/database.ts
```diff
@@ -2652,6 +2682,16 @@ function resultChangeType(type: ChangeType): firestore.DocumentChangeType {
...
 }
 
+function isEqualConverter(
+  first: firestore.FirestoreDataConverter<unknown> | undefined,
+  second: firestore.FirestoreDataConverter<unknown> | undefined
+): boolean {
+  if (first === undefined) {
+    return second === undefined;
+  } else {
+    return second !== undefined && first.toString() === second.toString();
```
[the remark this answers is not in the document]
thebrianchen: I originally tried comparing with `first === second`, but it caused the equality tests to fail because I think it compares the memory at which the function is defined rather than the internals of the function. I read about other ways to make equality work by putting it on the Object prototype, but the `toString()` way seemed the most straightforward.

packages/firestore/src/api/database.ts
```diff
@@ -2461,7 +2491,7 @@ export class CollectionReference<T = firestore.DocumentData> extends Query<T>
     return DocumentReference.forPath<T>(
       this._query.path.child(path),
       this.firestore,
-      this.converter
+      this._converter
```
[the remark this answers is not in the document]
thebrianchen: Resolved in group chat where we decided to keep it as is.

packages/firestore/src/api/database.ts
```diff
@@ -1039,17 +1047,22 @@ export class DocumentReference<T = firestore.DocumentData>
...
     }
     const path = ResourcePath.fromString(pathString);
-    return new CollectionReference(this._key.path.child(path), this.firestore, {
-      toFirestore: value => value,
-      fromFirestore: snapshot => snapshot.data() as T
-    });
+    return new CollectionReference(
+      this._key.path.child(path),
+      this.firestore,
+      this._converter
```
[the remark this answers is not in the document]
thebrianchen: Good catch, thanks!

packages/firestore/src/api/database.ts
```diff
@@ -1014,16 +1011,27 @@ export class DocumentReference<T = firestore.DocumentData>
...
 
   get path(): string {
     return this._key.path.canonicalString();
   }
 
+  /**
+   * Returns a default FirestoreDataConverter that is used when a converter
+   * is not provided.
+   */
+  get defaultConverter(): firestore.FirestoreDataConverter<T> {
```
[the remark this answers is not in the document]
thebrianchen: Thanks for walking me through it, from level 1 thinking to mikelehen thinking!

packages/firestore/src/api/database.ts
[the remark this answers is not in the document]
mikelehen: Hrm!  So the "compares the memory at which the function is defined rather than the internals of the function" is what I was expecting.  So if you have two converters that happen to be identical but are defined separately in your code, then they wouldn't compare equal.  I think this would be OK.

That said, comparing .toString() is perhaps better.  I was initially skeptical because I thought it was essentially undefined behavior (i.e. browser-dependent), but I guess it was standardized (maybe in ES2015? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString#Specifications) and so I guess it's okay to rely on it.

I'm still a tiny bit hesitant, but I don't have a good reason.  My biggest argument would perhaps be that it might be kinda' slow (I'm not sure how the browser generates the string). But it's probably fine.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1229,4 +1230,111 @@ apiDescribe('Database', (persistence: boolean) => {
...
+
+    it('drops the converter when calling CollectionReference<T>.parent()', () => {
+      return withTestDb(persistence, async db => {
+        const postsCollection = db
+          .collection('users/user1/posts')
+          .withConverter(postConverter);
+
+        const usersCollection = postsCollection.parent;
+        expect((usersCollection as DocumentReference)._converter).to.be
+          .undefined;
```
mikelehen: Note that accessing `_` properties from tests is kinda' bad practice and it may break if @schmidt-sebastian gets his property minification stuff to work. 😛 

So you might consider finding another way to write this...  Perhaps use `isEqual()`?

packages/firestore/test/util/api_helpers.ts
```diff
@@ -55,11 +60,19 @@ export function firestore(): Firestore {
 }
 
 export function collectionReference(path: string): CollectionReference {
-  return new CollectionReference(pathFrom(path), firestore());
+  return new CollectionReference(
+    pathFrom(path),
+    firestore(),
+    defaultDocumentDataConverter()
```
mikelehen: We made converter optional, so perhaps you can drop all of these (and get rid of `defaultDocumentDataConverter` entirely)?

packages/firebase/index.d.ts
```diff
@@ -7080,13 +7080,69 @@ declare namespace firebase.firestore {
...
+   * Firestore data.
+   *
+   * Using the converter allows you to specify generic type arguments when
+   * storing and retrieving objects from Firestore.
+   *
+   * @example
+   *  ```typescript
+   * class Post {
+   *   constructor(readonly title: string, readonly author: string) {}
+   *     byline(): string {
```
schmidt-sebastian: It is late in the day, but it took me way too long to figure that these are two methods. Can you add a new line and use a common term for the method, such as `toString()`?

packages/firebase/index.d.ts
```diff
@@ -7080,13 +7080,69 @@ declare namespace firebase.firestore {
...
+   *     return this.title + ', by ' + this.author;
+   *   }
+   * }
+   *
+   * const PostConverter = {
```
schmidt-sebastian: Nit: Specify the type of the converter here (`const converter: FirestoreDataConverter()`).

`PostConverter` should also be lowercase since it is an instance of a converter.

packages/firestore/src/api/database.ts
```diff
@@ -2214,13 +2345,14 @@ export class QuerySnapshot implements firestore.QuerySnapshot {
     );
```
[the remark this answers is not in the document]
thebrianchen: Good catch. Thank you

packages/firestore/src/api/database.ts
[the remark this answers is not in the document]
thebrianchen: API discussion decision was to use `first === second` direct comparison.

thebrianchen:
Also went over the docs with Mark, and updated docs after his review.

packages/firebase/index.d.ts
```diff
@@ -7551,7 +7614,7 @@ declare namespace firebase.firestore {
      * to the backend (Note that it won't resolve while you're offline).
      */
     update(
-      documentRef: DocumentReference,
+      documentRef: DocumentReference<unknown>,
```
schmidt-sebastian: Using `unknown` here breaks consumers of our SDK that are still on TypeScript 2.x. Can you confirm with the JS team that that is ok?

packages/firebase/index.d.ts
```diff
@@ -7906,10 +7972,24 @@ declare namespace firebase.firestore {
...
+      onNext: (snapshot: DocumentSnapshot<T>) => void,
       onError?: (error: Error) => void,
       onCompletion?: () => void
     ): () => void;
+
+    /**
+     * Applies a custom data converter to this DocumentReference, allowing you
+     * to use your own custom model objects with Firestore. When you call
+     * set(), get(), etc. on the returned DocumentReference instance, it will
+     * use the provided converter to convert data to/from type U (your custom
```
schmidt-sebastian: Maybe: "... returned DocumentReference instance, the provided converter will convert between Firestore data and your custom type U".

packages/firestore/src/api/database.ts
```diff
@@ -2496,6 +2646,30 @@ function resultChangeType(type: ChangeType): firestore.DocumentChangeType {
...
+ */
+function applyFirestoreDataConverter<T>(
+  converter: firestore.FirestoreDataConverter<T> | undefined,
+  value: T,
+  functionName: string
+): [firestore.DocumentData, string] {
+  let convertedValue;
+  if (converter) {
+    convertedValue = converter.toFirestore(value);
+    functionName = 'toFirestore() in Transaction.set';
```
schmidt-sebastian: This is surprising to me (I wouldn't expect the error message to contain a Transaction string if I use a WriteBatch). Can this be done at the callsite?

packages/firebase/index.d.ts
```diff
@@ -7080,13 +7080,73 @@ declare namespace firebase.firestore {
...
+   * ```typescript
+   * class Post {
+   *   constructor(readonly title: string, readonly author: string) {}
+   *
+   *   toString(): string {
+   *     return this.title + ', by ' + this.author;
+   *   }
+   * }
+   *
+   * const postConverter = {
```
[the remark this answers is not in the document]
mikelehen: If you make it a class, then they would have to do `withConverter(new PostConverter())` instead of `withConverter(PostConverter)` so I'm not sure it's an improvement...

packages/firebase/index.d.ts
thebrianchen: Confirmed w/ Feiyang that they currently only support 2.8 and up.

schmidt-sebastian: I think that means we have to drop "unknown".

thebrianchen: Oops didn't mean to resolve. Is the right thing to do to replace with `<any> ` and add a comment to disable linter?

mikelehen: FWIW @schmidt-sebastian recently "discovered" that

```
function foo(x: any): Bar { ... }
```
is essentially equivalent to:
```
function foo<T>(x: T): Bar { ... }
```
TypeScript will let you pass any type as `x`...  and so maybe the right thing to do here is just make update generic as well:
```
update<T>(documentRef: DocumentReference<T>, ...)
```

schmidt-sebastian: Looks like `any` is already used in the file without disabling tslint. I can't think of a better way to express this.

thebrianchen: It works in the `index.d.ts` files, but when I used it in `database.ts`, it would trigger the "unexpected any" error. Should I just disable tslint, or switch to the other notation? I think using `<any>` is more readable and easier to understand for the end user, but I'm not sure if there are any consequences to disabling tslint.

packages/firebase/index.d.ts
```diff
@@ -7080,13 +7080,73 @@ declare namespace firebase.firestore {
...
+   *     return new Post(data.title, data.author);
+   *   }
+   * };
+   *
+   * const postSnap = await firebase.firestore()
+   *   .collection('posts')
+   *   .withConverter(postConverter)
+   *   .doc().get();
+   * const post = postSnap.data();
+   * if (post !== undefined) {
```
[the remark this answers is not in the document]
thebrianchen: I think specifying the custom types in the code snippet provides an exact example of what custom types are and what they enable the user to do that they otherwise could not.

packages/firestore/src/api/database.ts
thebrianchen: Oops, typo. Meant to add the functionName that is passed in, so in this case, `WriteBatch.set()` would be passed in if the user used a WriteBatch.

packages/firebase/index.d.ts
mikelehen: Using `any` and disabling the tslint check (on a single-line basis) in database.ts should be fine.

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-# Unreleased (1.8.0)
+# Unreleased
+- [feature] Added support for strongly typed collections, documents, and
```
schmidt-sebastian: I don't think average Joe Schmidt (a JavaScript user) will be able to decipher the true impact of this change. You should point out that users can now store and retrieve custom classes in Firestore documents.

packages/firestore/src/api/database.ts
```diff
@@ -772,17 +772,20 @@ export class Transaction implements firestore.Transaction {
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    documentRef: firestore.DocumentReference<any>,
     field: string | ExternalFieldPath,
     value: unknown,
     ...moreFieldsAndValues: unknown[]
   ): Transaction;
   update(
-    documentRef: firestore.DocumentReference<unknown>,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    documentRef: firestore.DocumentReference<any>,
```
schmidt-sebastian: Since this "catch all" implementation is not part of the Public API, you could still use "unknown" here (if the TypeScript compiler allows it.

Optional, but it does provide a bit more type safety.

PR #2249, opened 2019-10-09
Serialize Timestamps as string (per proto3 JSON format).

yuchenshi:
This fixes using timestamps with the Firestore Emulator. This is similar to #2228, but for timestamps.

### Discussion

(Internal tracking ID: 37282237)

### Testing

Tested manually against the Firestore Emulator and production Firestore.

### API Changes

None.

### DIFF
--- a/packages/firestore/test/unit/remote/node/serializer.test.ts
+++ b/packages/firestore/test/unit/remote/node/serializer.test.ts
@@ -321,6 +321,12 @@ describe('Serializer', () => {
 
       expect(
         proto3JsonSerializer.toValue(
+          new fieldValue.TimestampValue(new Timestamp(1488872578, 916000))
+        )
+      ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.000916000Z' });
+
+      expect(
+        proto3JsonSerializer.toValue(
           new fieldValue.TimestampValue(new Timestamp(1488872578, 0))
         )
       ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.000000000Z' });
--- a/packages/firestore/test/unit/remote/node/serializer.test.ts
+++ b/packages/firestore/test/unit/remote/node/serializer.test.ts
@@ -303,31 +303,27 @@ describe('Serializer', () => {
     it('converts TimestampValue to string (useProto3Json=true)', () => {
       expect(
         proto3JsonSerializer.toValue(
-          new fieldValue.TimestampValue(new Timestamp(1488872578, 916123456)))
-      ).to.deep.equal(
-        { timestampValue: '2017-03-07T07:42:58.916123456Z' }
-      );
+          new fieldValue.TimestampValue(new Timestamp(1488872578, 916123456))
+        )
+      ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.916123456Z' });
 
       expect(
         proto3JsonSerializer.toValue(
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
Looks like this broke our tests. Can you take a look?

yuchenshi:
It's actually broken since we lost nanosecond precision. I'll revise the PR.

yuchenshi:
Fixed

packages/firestore/src/api/timestamp.ts
```diff
@@ -83,6 +83,14 @@ export class Timestamp {
     );
   }
 
+  _toISOString(): string {
```
schmidt-sebastian: Hrm. This is no longer an ISO String, so I think we should drop that naming. Furthermore, I am not a huge fan of adding this to our public API class (even when it is meant to be private). I would suggest either inlining it into the callsite or adding another dedicated method to RemoteSerializer.

packages/firestore/src/api/timestamp.ts
```diff
@@ -83,6 +83,14 @@ export class Timestamp {
...
+    const strMillis = this.toDate().toISOString(); // JS Date only has millis.
+    // So let's replace the fraction w/ up to 9 digits (nano resolution).
+    const fracs = (this.nanoseconds / 1e9).toFixed(9).substr(1); // .123456789
+    const strNanos = strMillis.replace(/\.\d*/, '').replace('Z', fracs + 'Z');
+    return strNanos;
```
schmidt-sebastian: Looks like my comment got dropped. I think we should try to make it more obvious what is going on, so that a casual read can understand this.

My idea would be:

- Create a Date just from the seconds
- Drop the last 4 characters from its toISOString
- Zero pad the nanos by prefixing with "000000000" and using substring
- Concat both strings and add "Z"

packages/firestore/src/api/timestamp.ts
yuchenshi: I'll inline this but just to clarify, the result is still an ISO string, just with more digits. The standard does not specify limits on the number of digits, and the fact that JS limits it to 3 digits is more like an issue with Date resolution.

packages/firestore/src/api/timestamp.ts
yuchenshi: Done

packages/firestore/test/unit/remote/node/serializer.test.ts
```diff
@@ -300,6 +300,32 @@ describe('Serializer', () => {
...
+        )
+      ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.916123456Z' });
+
+      expect(
+        proto3JsonSerializer.toValue(
+          new fieldValue.TimestampValue(new Timestamp(1488872578, 916123000))
+        )
+      ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.916123000Z' });
+
+      expect(
```
schmidt-sebastian: Can you also add a test for:

```
 expect(
        proto3JsonSerializer.toValue(
          new fieldValue.TimestampValue(new Timestamp(1488872578, 916000))
        )
      ).to.deep.equal({ timestampValue: '2017-03-07T07:42:58.000916000Z' });
```

yuchenshi: Done.

PR #2251, opened 2019-10-09
Marking SimpleDb calls as idempotent

schmidt-sebastian:
This PR marks all straightforward calls of SimpleDb.runTransaction as idempotent. Most of the usage is in test.

This was verified via a manual code audit and via a test hack (not checked in) that fails the first run of idempotent transactions.

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -321,6 +321,7 @@ export class IndexedDbPersistence implements Persistence {
       .then(() => {
         return this.simpleDb.runTransaction(
           'readonly',
+          /* idempotent= */ false,
           [DbTargetGlobal.store],
           txn => {
             return getHighestListenSequenceNumber(txn).next(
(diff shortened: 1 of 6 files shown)


packages/firestore/test/unit/local/encoded_resource_path.test.ts
```diff
@@ -184,6 +184,7 @@ async function assertOrdered(paths: ResourcePath[]): Promise<void> {
 
   const selected: string[] = [];
   await runTransaction(simpleStore => {
+    selected.splice(0); // The function might get re-run.
```
wilhuff: This would be cleaner if the selected array were just returned from the transaction.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -329,7 +329,7 @@ export class IndexedDbPersistence implements Persistence {
       })
       .then(() => {
         return this.simpleDb.runTransaction(
-          'readonly',
+          'readonly-idempotent',
```
wilhuff: This appears to modify `this.listenSequence` in the transaction.

Why not move that out of the transaction and into the first `then` block below?

schmidt-sebastian: There are a bunch of calls in the SDK that follow this pattern and re-compute a value inside a transaction, but the re-computation has no side effects. In this case, refactoring the change out of this transaction is pretty simple, so I applied the change. It might not always be so simple in other cases though.

packages/firestore/test/unit/local/encoded_resource_path.test.ts
schmidt-sebastian: That's much cleaner. Updated.

PR #2252, opened 2019-10-09
Mark mostly readonly calls as idempotent

schmidt-sebastian:
This PR marks the simple callsites of IndexedDbPersistence.runTransaction() as idempotent.

Prerequisite was to make 'readonly-primary' and friends available via the public Persistence interface.

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -273,7 +273,16 @@ export class SimpleDb {
         objectStores
       );
       try {
+        // TODO(schmidt-sebastian): Remove this code/comment or find a way to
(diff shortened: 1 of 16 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -273,7 +273,16 @@ export class SimpleDb {
         objectStores
       );
       try {
+        // TODO(schmidt-sebastian): Remove this code/comment or find a way to
```
schmidt-sebastian: I left this here for now to make my development process easier. This actually caught an issue during development, when I marked `getNewDocumentChanges` as idempotent. It is not yet since it modifies internal state that is used during execution.

PR #2254, opened 2019-10-10
[1/3] Query/Target split -- create Target and make remote/ and local/ work with Target.

wu-hui:
Changes:

1. Created `Target` and `Query.toTarget`. 
2. Make local/ and remote/ work with `Target`, except for `LocalStore.executeQuery`.

Future PRs:
[2/3] Make sync engine work with both `Query` and `Target`.
[3/3] Rename `QueryData` to `TargetData` and `QueryCache` to `TargetCache`.


packages/firestore/src/core/sync_engine.ts
```diff
@@ -949,11 +950,12 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
         // allocate the query in LocalStore and initialize a new View.
-        const query = await this.localStore.getQueryForTarget(targetId);
-        assert(!!query, `Query data for target ${targetId} not found`);
-        queryData = await this.localStore.allocateQuery(query!);
+        const target = await this.localStore.getTarget(targetId);
+        assert(!!target, `Target for id ${targetId} not found`);
+        queryData = await this.localStore.allocateTarget(target!);
         await this.initializeViewAndComputeSnapshot(
-          queryData,
+          target!.toQuery(),
```
schmidt-sebastian: I am not sure converting a query to a target here will always work. If a secondary tab issues a `limitToLast` query, and the primary tab executes the query without knowing this, then the computed View is likely different. Unfortunately, we use the View to determine which documents to return for Limit queries (https://github.com/firebase/firebase-js-sdk/blob/5beb23cd47312ffc415d3ce2ae309cc3a3fde39f/packages/firestore/src/core/view.ts#L226). I think executing the "wrong query" might break Limbo resolution.

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+import { ResourcePath } from '../model/path';
+import { isNullOrUndefined } from '../util/types';
+import { Bound, Filter, OrderBy, Query } from './query';
+import { DocumentKey } from '../model/document_key';
+
+/**
+ * Represents a backend query: the query protos SDK sent to backend and persisted
+ * in local storage are from this class.
```
schmidt-sebastian: Same comment regarding "local storage". Can you be more specific here?

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+    if (this.endAt) {
+      str += ', endAt: ' + this.endAt.canonicalId();
+    }
+
+    return str + ')';
+  }
+
+  isEqual(other: Target): boolean {
+    if (this.limit !== other.limit) {
```
schmidt-sebastian: It would be nice if Target and Query could share some of this code.

packages/firestore/src/local/local_store.ts
```diff
@@ -163,9 +164,9 @@ export class LocalStore {
   /** Maps a targetID to data about its query. */
   private queryDataByTarget = {} as { [targetId: number]: QueryData };
 
-  /** Maps a query to its targetID. */
-  private targetIdByQuery = new ObjectMap<Query, TargetId>(q =>
-    q.canonicalId()
+  /** Maps a target to its targetID. */
+  private targetIdByTarget = new ObjectMap<Target, TargetId>(t =>
```
schmidt-sebastian: Have you considered making the TargetId a part of Target?

packages/firestore/src/local/local_store.ts
```diff
@@ -754,19 +755,24 @@ export class LocalStore {
     });
   }
 
+  //TODO(wuandy): delete this temp method, it's only for change isolation.
```
schmidt-sebastian: Please add space between "//" and "TODO" and consider using your GitHub handle.

packages/firestore/src/local/local_store.ts
```diff
@@ -806,31 +812,41 @@ export class LocalStore {
...
+  releaseQuery(query: Query, keepPersistedQueryData: boolean): Promise<void> {
+    return this.releaseTarget(query.toTarget(), keepPersistedQueryData);
+  }
+
   /**
    * Unpin all the documents associated with the given query. If
    * `keepPersistedQueryData` is set to false and Eager GC enabled, the method
    * directly removes the associated query data from the query cache.
+   *
+   * Calling `releaseTarget` multiple times with the same target ID is an error.
```
schmidt-sebastian: s/target ID/target ?

packages/firestore/src/local/query_data.ts
```diff
@@ -37,10 +37,11 @@ export enum QueryPurpose {
 /**
  * An immutable set of metadata that the local store tracks for each query.
  */
+//TODO(wuandy): rename this to TargetData.
```
schmidt-sebastian: Same comment regarding spacing and usage of GitHub handle.

packages/firestore/src/core/query.ts
```diff
@@ -28,7 +28,12 @@ import { FieldPath, ResourcePath } from '../model/path';
 import { assert, fail } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { isNullOrUndefined } from '../util/types';
+import { Target } from './target';
 
+/**
+ * Represents a local query, it can be run against local storage, as well as be converted
```
schmidt-sebastian: Nit: s/it can/which can/

"Local storage" is also a very ambiguous term on Web. I believe you are referring to the Firestore concept (LocalStore) here rather than Web Storage.

schmidt-sebastian:
@wu-hui Did you mean to assign this back to me already? :)

wu-hui:
> @wu-hui Did you mean to assign this back to me already? :)

My bad, I thought I forgot to assign to you..turned out you already replied..

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,82 @@
...
+  canonicalId,
+  toString,
+  isEqual,
+  isDocumentQuery
+} from './query_target_common';
+
+/**
+ * Represents a backend query: the query protos SDK sent to backend and persisted
+ * in the LocalStore are from this class.
+ */
```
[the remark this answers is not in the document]
wu-hui: Gave it a try..PTAL.

packages/firestore/src/core/query.ts
```diff
@@ -220,106 +231,17 @@ export class Query {
...
-      if (this.startAt) {
-        canonicalId += '|lb:';
-        canonicalId += this.startAt.canonicalId();
-      }
-      if (this.endAt) {
-        canonicalId += '|ub:';
-        canonicalId += this.endAt.canonicalId();
-      }
-      this.memoizedCanonicalId = canonicalId;
+      this.memoizedCanonicalId = canonicalId(this);
```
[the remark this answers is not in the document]
wu-hui: Done. `Query` now has-a `Target`.

packages/firestore/src/core/query.ts
```diff
@@ -37,6 +43,9 @@ export class Query {
   private memoizedCanonicalId: string | null = null;
   private memoizedOrderBy: OrderBy[] | null = null;
 
+  // The corresponding `Target` this `Query` instance.
+  private target: Target | null = null;
```
schmidt-sebastian: Is there ever a case where `toTarget()` is not called? If there isn't, we can probably assign `this.target` in the constructor.

packages/firestore/src/core/query.ts
```diff
@@ -220,106 +229,17 @@ export class Query {
...
-      if (this.startAt) {
-        canonicalId += '|lb:';
-        canonicalId += this.startAt.canonicalId();
-      }
-      if (this.endAt) {
-        canonicalId += '|ub:';
-        canonicalId += this.endAt.canonicalId();
-      }
-      this.memoizedCanonicalId = canonicalId;
+      this.memoizedCanonicalId = this.toTarget().canonicalId();
```
schmidt-sebastian: At least for now, you don't need to memoize the canonical ID since this is done by Target.

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
schmidt-sebastian: My attempt would be:

"A Target represents the WatchTarget representation of a Query, which is used
by the LocalStore and the RemoteStore to keep track of and to execute backend
queries. While a Query can represent multiple Targets, each Targets maps to
a single WatchTarget in RemoteStore and a single TargetData entry in persistence."

Feel free to combine yours and mine :)

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,181 @@
...
+export class Target {
+  private memoizedCanonicalId: string | null = null;
+
+  /**
+   * Initializes a Target with a path and optional additional query constraints.
+   * Path must currently be empty if this is a collection group query.
+   *
+   * NOTE: you should always construct `Target` from `Query.toTarget` instead of
+   * using this constructor, because `Query` provides sensible default `orderBy`
```
schmidt-sebastian: What is "sensible"? In the rest of the code base we differentiate between "implicit" and "explicit". Can we use the same wording here?

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,174 @@
...
+  isDocumentQuery(): boolean {
+    return (
+      DocumentKey.isDocumentKey(this.path) &&
+      this.collectionGroup === null &&
+      this.filters.length === 0
+    );
+  }
+
+  toQuery(): Query {
+    return new Query(
```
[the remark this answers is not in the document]
schmidt-sebastian: Sounds good.

packages/firestore/src/core/target.ts
```diff
@@ -0,0 +1,181 @@
...
+    return (
+      DocumentKey.isDocumentKey(this.path) &&
+      this.collectionGroup === null &&
+      this.filters.length === 0
+    );
+  }
+
+  /**
+   * Creates a `Query` object from this `Target`. Note the result might be
+   * different from the original `Query` from which we get this instance.
```
schmidt-sebastian: s/get/got/ (or obtained?)

packages/firestore/src/core/query.ts
```diff
@@ -220,106 +229,17 @@ export class Query {
...
-    }
-    if (this.startAt) {
-      str += ', startAt: ' + this.startAt.canonicalId();
-    }
-    if (this.endAt) {
-      str += ', endAt: ' + this.endAt.canonicalId();
-    }
-
-    return str + ')';
+    return `Query(${this.toTarget().toContentString()})`;
```
schmidt-sebastian: I don't have a strong preference here, but I would optimize for less code/a smaller public API and just write this as:

```
return `Query(target=${this.toTarget().toString()})`;
```

packages/firestore/src/core/query.ts
wu-hui: Yes. `LocalStore.executeQuery` is one example.

More importantly, `Query` is built in a way where there might be many temporary instances, every `with*` will create one for instance.

packages/firestore/src/core/query.ts
wu-hui: Done.

packages/firestore/src/core/query.ts
wu-hui: Done.

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
wu-hui: Took yours.

packages/firestore/src/core/target.ts
wu-hui: Done.

packages/firestore/src/core/target.ts
wu-hui: Done.

packages/firestore/src/core/target.ts
```diff
@@ -34,7 +35,7 @@ export class Target {
    * Path must currently be empty if this is a collection group query.
    *
    * NOTE: you should always construct `Target` from `Query.toTarget` instead of
-   * using this constructor, because `Query` provides sensible default `orderBy`
+   * using this constructor, because `Query` provides implicit `orderBy`
```
schmidt-sebastian: s/provides implicit/provides an implicit/

wu-hui: Done.

PR #2255, opened 2019-10-10
Make getNewDocumentChanges() idempotent

schmidt-sebastian:
Ugh. I believe this is one of the more difficult "idempotent" migration since getNewDocumentChanges() changes state in-memory state in the Persistence layer, whereas most other transactions just change in-memory state in LocalStore.

To make `getNewDocumentChanges()` idempotent, I move the change watermark into LocalStore, which then lets me set it only after the transaction is committed. Instead of making this work for MemoryRemoteDocumentStore, I opted to just assert that `getNewDocumentChanges()` is never called (it is only used for multi-tab).

### DIFF
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -55,9 +55,6 @@ import {
 import { ObjectMap } from '../util/obj_map';
 
 export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
-  /** The read time of the last entry consumed by `getNewDocumentChanges()`. */
-  private lastProcessedReadTime = SnapshotVersion.MIN;
-
   /**
    * @param {LocalSerializer} serializer The document serializer.
    * @param {IndexManager} indexManager The query indexes that need to be maintained.
@@ -66,18 +63,6 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
     readonly serializer: LocalSerializer,
     private readonly indexManager: IndexManager
   ) {}
-
-  /**
-   * Starts up the remote document cache.
-   *
-   * Reads the ID of the last  document change from the documentChanges store.
-   * Existing changes will not be returned as part of
-   * `getNewDocumentChanges()`.
-   */
-  // PORTING NOTE: This is only used for multi-tab synchronization.
-  start(transaction: SimpleDbTransaction): PersistencePromise<void> {
-    return this.synchronizeLastProcessedReadTime(transaction);
-  }
 
   /**
    * Adds the supplied entries to the cache.
@@ -313,14 +298,21 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       .next(() => results);
   }
 
-  getNewDocumentChanges(
-    transaction: PersistenceTransaction
-  ): PersistencePromise<MaybeDocumentMap> {
+  /**
+   * Returns the set of documents that have been updated since the specified read
+   * time.
+   */
+  // PORTING NOTE: This is only used for multi-tab synchronization.
+  getDocumentChanges(
+    transaction: PersistenceTransaction,
+    sinceReadTime: SnapshotVersion
+  ): PersistencePromise<{
+    changedDocs: MaybeDocumentMap;
+    readTime: SnapshotVersion;
+  }> {
     let changedDocs = maybeDocumentMap();
 
-    const lastReadTime = this.serializer.toDbTimestampKey(
-      this.lastProcessedReadTime
-    );
+    let lastReadTime = this.serializer.toDbTimestampKey(sinceReadTime);
 
     const documentsStore = remoteDocumentsStore(transaction);
     const range = IDBKeyRange.lowerBound(lastReadTime, true);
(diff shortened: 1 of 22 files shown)


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -313,59 +298,74 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
-    this.lastProcessedReadTime = SnapshotVersion.forDeletedDoc();
-    return documentsStore.iterate(
-      { index: DbRemoteDocument.readTimeIndex, reverse: true },
-      (key, value, control) => {
-        if (value.readTime) {
-          this.lastProcessedReadTime = this.serializer.fromDbTimestampKey(
-            value.readTime
-          );
+  // PORTING NOTE: This is only used for multi-tab synchronization.
+  getLastDocumentChange(
```
schmidt-sebastian: This is used in the `synchronizeLastProcessedReadTime` version of LocalStore. The `changedDoc` is never read, but I opted to return it to make the API similar to getDocumentChanges above.

packages/firestore/src/local/local_store.ts
```diff
@@ -192,6 +200,11 @@ export class LocalStore {
     this.queryEngine.setLocalDocumentsView(this.localDocuments);
   }
 
+  /** Starts the LocalStore. */
+  start(): Promise<void> {
+    return this.synchronizeLastDocumentChangeReadTime();
```
schmidt-sebastian: This is equivalent to the previous functionality in RemoteDocumentStore.start().

packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -101,10 +131,62 @@ describe('IndexedDbRemoteDocumentCache', () => {
       dontPurgeData: true
     });
     cache = new TestRemoteDocumentCache(persistence);
-    const changedDocs = await cache.getNewDocumentChanges();
+    const { readTime } = await getLastDocumentChange();
+    const { changedDocs } = await getDocumentChanges(readTime);
+    assertMatches([], changedDocs);
+  });
+
+  it('can get changes', async () => {
```
schmidt-sebastian: These tests moved so they are only exercised against IndexedDb.

packages/firestore/src/local/local_store.ts
```diff
@@ -168,6 +169,13 @@ export class LocalStore {
     q.canonicalId()
   );
 
+  /**
+   * The read time of the last entry processed by `getNewDocumentChanges()`.
```
wilhuff: `getDocumentChanges`?

Oh! I see: this refers to `getNewDocumentChanges` in `LocalStore` and not the one in the `RemoteDocumentCache`. This is slightly confusing and I have a slight preference for keeping the names the same, though up to you.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -313,59 +293,74 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
 
-  getNewDocumentChanges(
-    transaction: PersistenceTransaction
-  ): PersistencePromise<MaybeDocumentMap> {
+  /**
+   * Returns the set of documents that have been updated since the specified read
+   * time.
+   */
+  // PORTING NOTE: This is only used for multi-tab synchronization.
+  getDocumentChanges(
```
wilhuff: FWIW, the `getNewDocumentChanges` still makes sense while explicitly passing the `readTime`. That name helps convey that it's a diff of changes.

I don't oppose renaming it to `getDocumentChanges` if that's your preference, but I don't think it needed to change.

packages/firestore/src/local/remote_document_cache.ts
```diff
@@ -82,17 +81,6 @@ export interface RemoteDocumentCache {
...
   ): PersistencePromise<DocumentMap>;
 
-  /**
-   * Returns the set of documents that have been updated since the last call.
-   * If this is the first call, returns the set of changes since client
-   * initialization. Further invocations will return document changes since
-   * the point of rejection.
-   */
-  // PORTING NOTE: This is only used for multi-tab synchronization.
-  getNewDocumentChanges(
```
wilhuff: An alternative to making LocalStore aware that it's using IndexedDB persistence would be to leave this method in the interface and allow persistence implementations to throw some kind of unsupported operation error if they're ever called.

This would have the effect of making the LocalStore code cleaner. Both would fail if MemoryPersistence was ever used with multi-tab. As is we fail the assertion; the proposed change would throw unsupported operation.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I see your confusion:) I tried to be to smart about it and reverted back to using getNewDocumentChanges() everywhere, which still makes sense even when a read time is passed (at least I see it that way now).

packages/firestore/src/local/remote_document_cache.ts
schmidt-sebastian: I added the method back to the interface and am throwing an exception now. I debated between the two approaches, but it probably makes sense to reduce this clutter in LocalStore.

I did leave `getLastDocumentChange()` IndexdedDb only. This is a new method and there are no asserts for this in LocalStore (just an if-statement that makes the intent clear).

PR #2257, opened 2019-10-10
Make handleUserChange idempotent

wilhuff:


wilhuff:
I ran with the forced rerun and tests related to user changes did fail. The issue is that on the rerun the “old” list is computed from the new mutation queue so the diff becomes empty.

PR #2259, opened 2019-10-10
 Improve test hack

schmidt-sebastian:
Improve the idempotent test hack to prevent writes to IndexedDb by failing the transaciton.


schmidt-sebastian:
Commented this back in for now.

PR #2260, opened 2019-10-11
Expose nonce

wti806:
1. Updated the OAuthProvider.credential API to accept Object for the first argument. 
2. Exposed nonce property on OAuthCredential object.


packages/auth-types/index.d.ts
```diff
@@ -215,7 +215,10 @@ export class OAuthProvider implements AuthProvider {
   constructor(providerId: string);
   providerId: string;
   addScope(scope: string): AuthProvider;
-  credential(idToken?: string, accessToken?: string): OAuthCredential;
+  credential(
+    optionsOrIdToken: Object | string | null,
```
bojeil-google: Let's define the `Object` interface here.

packages/firebase/index.d.ts
```diff
@@ -3615,23 +3620,27 @@ declare namespace firebase.auth {
...
      * ```
      *
-     * @param idToken The OAuth ID token if OIDC compliant.
+     * @param optionsOrIdToken Either the options object containing
+     *     the ID token, access token and raw nonce or the ID token string.
      * @param accessToken The OAuth access token.
      */
     credential(
-      idToken?: string,
+      optionsOrIdToken: Object | string | null,
```
bojeil-google: Same here on defining the interface.

packages/auth-types/index.d.ts
```diff
@@ -284,6 +287,13 @@ export class OAuthCredential extends AuthCredential {
   idToken?: string;
   accessToken?: string;
   secret?: string;
+  nonce?: string;
```
bojeil-google: This is not exposed. Let's remove it from the type definitions.

packages/firebase/externs/firebase-auth-externs.js
```diff
@@ -122,6 +154,14 @@ firebase.auth.OAuthCredential.prototype.accessToken;
  */
 firebase.auth.OAuthCredential.prototype.secret;
 
+/**
+ * The raw nonce associated with the ID token if the ID token is created with
+ * a nonce.
+ *
+ * @type {?string|undefined}
+ */
+firebase.auth.OAuthCredential.prototype.nonce;
```
bojeil-google: Same here.

PR #2263, opened 2019-10-11
Make applyRemoteEvent idempotent

schmidt-sebastian:
This one was the most "fun" so far.

Major changes:
- Make queryDataByTarget an immutable map, so I can discard changes if the transaction inside of applyRemoteEvent fails.
- Move the checks whether a CollectionParent index has been written outside of the RDC and the MC and instead move it to the callsites (this can lead to slightly more calls since the data structure is no longer used throughout the entire lifetime of the client).

### DIFF
--- a/packages/firestore/src/local/indexeddb_index_manager.ts
+++ b/packages/firestore/src/local/indexeddb_index_manager.ts
@@ -22,6 +22,7 @@ import { decode, encode } from './encoded_resource_path';
 import { IndexManager } from './index_manager';
 import { IndexedDbPersistence } from './indexeddb_persistence';
 import { DbCollectionParent, DbCollectionParentKey } from './indexeddb_schema';
+import { MemoryCollectionParentIndex } from './memory_index_manager';
 import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { SimpleDbStore } from './simple_db';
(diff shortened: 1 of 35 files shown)


packages/firestore/src/local/lru_garbage_collector.ts
```diff
@@ -83,12 +84,10 @@ export interface LruDelegate {
...
 
 /**
- * Describes an object whose keys are active target ids. We do not care about the type of the
+ * Describes a map whose keys are active target ids. We do not care about the type of the
  * values.
  */
-export interface ActiveTargets {
-  [id: number]: unknown;
-}
+export type ActiveTargets = SortedMap<TargetId, unknown>;
```
schmidt-sebastian: Changed this type so I could continue to parse the target map from LocalStore.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -31,30 +30,17 @@ import { SimpleDbStore } from './simple_db';
  * A persisted implementation of IndexManager.
  */
 export class IndexedDbIndexManager implements IndexManager {
-  /**
```
wilhuff: I can't find where this cache is kept now. Unconditionally writing it would be a pretty major regression.

Or is the equivalent supposed to be the `collectionParents` addition in `indexeddb_mutation_queue.ts`? This cache ensures that we only ever write a collection parent index entry once for the entire duration of the app. The `indexeddb_mutation_queue.ts` changes ensure uniqueness within a mutation batch, but that's a far lesser benefit.

On the whole I think unwinding all of this in the name of idempotency is a bad deal. If we added `transaction.onCommitted` this wouldn't be impacted at all. This is the second case we were looking for--I think it makes sense to pursue that instead of these changes.

schmidt-sebastian:
I am not sure why this doesn't compile according to Travis (I will check after 1 pm), but I modified this PR to pull in the latest changes. Unfortunately, I still had to keep some of the new logic that determines whether we have already seen a collection parent before we write it, since the accounting now only happens after a transaction commit. We might be able to clean this up more if we add a `onAborted` callback, but that adds even more complexity.

PR #2264, opened 2019-10-11
idempotent `allocateQuery` and `notifyLocalViewChanges`

wu-hui:


packages/firestore/src/local/local_store.ts
```diff
@@ -812,24 +809,25 @@ export class LocalStore {
                   QueryPurpose.Listen,
                   txn.currentSequenceNumber
                 );
-                return this.queryCache.addQueryData(txn, queryData);
+                return this.queryCache.addQueryData(txn, queryData).next(() => {
```
schmidt-sebastian: I think you can just write this as:

```
 return this.queryCache.addQueryData(txn, queryData).next(() => queryData)
```

packages/firestore/src/local/local_store.ts
```diff
@@ -702,41 +702,40 @@ export class LocalStore {
...
+        );
+        this.queryDataByTarget = this.queryDataByTarget.insert(
+          targetId,
+          updatedQueryData
+        );
+      }
+    }
     return this.persistence.runTransaction(
       'notifyLocalViewChanges',
       'readwrite',
```
schmidt-sebastian: Here and below: Can you change this to `readwrite-idempotent`? Note that if you merge in the latest changes from mrschmidt/idempotent, the test runner will automatically run the transaction twice (so you can verify that this works).

wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Done.

PR #2265, opened 2019-10-11
Add onCommitted listeners for transactions

schmidt-sebastian:
This should simply some of the idempotent development by allowing any PersistenceTransaction consumer to register onCommitted listeners.

### DIFF
--- a/packages/firestore/src/local/memory_persistence.ts
+++ b/packages/firestore/src/local/memory_persistence.ts
@@ -180,12 +180,15 @@ export class MemoryPersistence implements Persistence {
     this.referenceDelegate.onTransactionStarted();
     return transactionOperation(txn)
       .next(result => {
-        txn.raiseOnCommittedEvent();
         return this.referenceDelegate
           .onTransactionCommitted(txn)
           .next(() => result);
       })
-      .toPromise();
+      .toPromise()
+      .then(result => {
+        txn.raiseOnCommittedEvent();
+        return result;
+      });
   }
 
   mutationQueuesContainKey(
--- a/packages/firestore/src/local/memory_persistence.ts
+++ b/packages/firestore/src/local/memory_persistence.ts
@@ -180,6 +180,7 @@ export class MemoryPersistence implements Persistence {
     this.referenceDelegate.onTransactionStarted();
     return transactionOperation(txn)
       .next(result => {
+        txn.raiseOnCommittedEvent();
         return this.referenceDelegate
           .onTransactionCommitted(txn)
           .next(() => result);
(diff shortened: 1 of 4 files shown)


packages/firestore/src/local/persistence.ts
```diff
@@ -29,14 +29,25 @@ import { RemoteDocumentCache } from './remote_document_cache';
...
  * When you call Persistence.runTransaction(), it will create a transaction and
  * pass it to your callback. You then pass it to any method that operates
  * on persistence.
  */
 export abstract class PersistenceTransaction {
+  private readonly onCommittedListeners: Array<() => {}> = [];
+
   abstract readonly currentSequenceNumber: ListenSequenceNumber;
+
+  addOnCommittedListener(listener: () => {}): void {
```
wilhuff: Shouldn't we require that the committed listeners return `void`? That way you can't accidentally have a return statement in your callback that you intend for the outer method.

Or would that interfere with using short arrow functions in the callback?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -180,6 +180,7 @@ export class MemoryPersistence implements Persistence {
     this.referenceDelegate.onTransactionStarted();
     return transactionOperation(txn)
       .next(result => {
+        txn.raiseOnCommittedEvent();
```
wilhuff: This ordering seems off. Couldn't the reference delegate conceivably modify the transaction?

IndexedDbPersistence has the better implementation. Chaining this off the promise makes it clearer that it's altogether outside the transaction.

packages/firestore/src/local/persistence.ts
schmidt-sebastian: I just realized this when I started to modify my "applyRemoteEvent" PR. It is much better if we return void.

packages/firestore/src/local/memory_persistence.ts
schmidt-sebastian: Changed this to invoke the committed handler as the last step.

schmidt-sebastian:
Tests added. Whee.

packages/firestore/test/unit/local/persistence_transaction.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        const targetsStore = IndexedDbPersistence.getStore<
+          DbTargetKey,
+          { targetId: TargetId }
+        >(txn, DbTarget.store);
+
+        txn.addOnCommittedListener(() => {
+          ++commitCount;
+        });
+
+        ++runCount;
```
wilhuff: Optional: you might also consider expecting that commitCount is zero during the transaction run.

schmidt-sebastian: Done

PR #2268, opened 2019-10-11
Make "Locally write mutations" idempotent

schmidt-sebastian:


wu-hui:
The name and description is wrong though, please correct.

Issue #2232, opened 2019-10-03
Firestore: Attempt to get a record without an in-progress transaction


Saturnyn:
Hello,

_sorry if this is not a proper bug report, but I am out of ideas and thought I'd try and ask for help and advice here._

Our team is maintaining an ionic app that uses firestore with the firebase js sdk.
The persistence is activated on firestore: this is a requirement for the app, it has to be able to save important data while being offline for a long period of time.

For a while now, we've been getting different flavours of the following error:
*FIRESTORE (6.0.2) INTERNAL ASSERTION FAILED: AsyncQueue is already failed*.
- An internal error was encountered in the Indexed Database server
- The transaction was aborted, so the request cannot be fulfilled
- Transaction timed out due to inactivity

We haven't been able to reproduce those errors consistently on our own devices, and this thread https://github.com/firebase/firebase-js-sdk/issues/1670 made us hope that the new iOS version would solve the problem, so we haven't been dealing with it directly and hoped that it would disappear after the iOS update.

However, now we have customers using iOS 13 and iPadOS and they're still getting "AsyncQueue is already failed" errors, expect with new messages:
- Attempt to get a record from database without an in-progress transaction
- Attempt to open a cursor in database without an in-progress transaction

We have run some tests on the device and were able to reproduce those new errors more reliably by just waiting for the iPad to lock itself and then waking it up.
It looks like updating to firebase sdk 6.6 makes this new kind of error disappear (on our testing device at least) but we have no idea if the previous errors are still there or not.

Digging into the bug reports more, we've seen that some of the "AsyncQueue is already failed" errors had actually also been coming from Android devices, so it looks like the problem is more complicated than we thought.

Can anyone tell me if firestore will work reliably on all platforms if we update to 6.6, or if there is anything else we can do to fix the problem ?

We are considering dropping firestore completely, but we want to make sure there is no solution before starting working on an alternative.

### [REQUIRED] Describe your environment

* ionic info:
```
Ionic:
   ionic (Ionic CLI)  : 4.1.0 (/usr/local/lib/node_modules/ionic)
   Ionic Framework    : ionic-angular 3.9.5
   @ionic/app-scripts : 3.2.3
Cordova:
   cordova (Cordova CLI) : 6.5.0
   Cordova Platforms     : android 6.4.0, ios 4.5.5
   Cordova Plugins       : cordova-plugin-ionic-webview 1.2.1, (and 19 other plugins)
System:
[...]

PR #2270, opened 2019-10-15
Idempotency: Address TODOs, add Changelog

schmidt-sebastian:
Address TODOs, add Changelog.

The only outstanding TODO is the clean up of the different modes we pass to runPersistence, but we should discuss offline if we want to clean up the unit tests first.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Unreleased
+- [changed] Improved iOS 13 support by addressing a change in Mobile Safari
+  that broke IndexedDB persistence if the Firestore SDK was used in a 
+  background tab.
+
+# 1.6.0
 - [fixed] Fixed a regression that caused queries with nested field filters to
   crash the client if the field was not present in the local copy of the
   document. 
-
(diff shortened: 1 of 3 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
 # Unreleased
+- [changed] Improved iOS 13 support by addressing a change in Mobile Safari
```
wilhuff: It's probably more important to tie this to observed user behavior and link to the details if someone is really curious.

> Fixed a crash on iOS 13 that could happen if Firestore was used in a tab in the background (#2232).

schmidt-sebastian: Updated to say:

"Fixed a crash on iOS 13 that occurred when persistence was enabled in a background tab."

This still points out that it is a persistence-only issue. As for linking to the original issue, we unfortunately don't do this in our public release notes (see https://firebase.google.com/support/release-notes/js), but I added it here in case someone uses our internal release notes.

wilhuff: FWIW, we do link to issues in the iOS release notes (https://firebase.google.com/support/release-notes/ios) so there's definitely precedent for it. I think it would be helpful.

PR #2271, opened 2019-10-15
Support iOS 13 with offline persistence

schmidt-sebastian:
This PR addresses a behavior change in iOS 13 that causes transaction to be aborted (https://bugs.webkit.org/show_bug.cgi?id=202705). As this is now common in Mobile Safari, we changed the transaction logic in Firestore to allow for retries. If a transaction is not committed successfully, it can now be retried.

This PR makes all non-unit tests transactions idempotent.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2232


packages/firestore/CHANGELOG.md
```diff
@@ -1,9 +1,15 @@
 # Unreleased
+- [changed] Fixed a crash on iOS 13 that occurred when persistence was enabled
+  in a background tab (#2232).
+- [fixed] Fixed an issue in the interaction with the Firestore Emulator that
+  caused requests with timestamps to fail.
```
schmidt-sebastian: I rewrote this slightly when I addressed the merge conflict (to add a missing "the").

cc @yuchenshi

PR #2276, opened 2019-10-17
Include add_payment_info and page_view events

cdoe:
Typescript currently throws a fit when trying to log either the `add_payment_info` or `page_view` events with `firebase.analytics.logEvent()`.  The error goes a little something like... 

`Argument of type '"add_payment_info"' is not assignable to parameter of type 'never'.`

These two are the only default events without parameters and also don't have function overloads defined—meaning the following line assigns them to `never`.
https://github.com/firebase/firebase-js-sdk/blob/07cd70d4e694080c0114076fe2be821c951ed75b/packages/firebase/index.d.ts#L4336

I think adding these two definitions patches things right up.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -3961,8 +3961,8 @@ declare namespace firebase.analytics {
      * ```
      */
     app: firebase.app.App;
-              
-     /**
+
+    /**
      * Sends analytics event with given `eventParams`. This method
      * automatically associates this logged event with this Firebase web
      * app instance on this device.
@@ -3972,10 +3972,10 @@ declare namespace firebase.analytics {
      */
     logEvent(
       eventName: 'add_payment_info',
-      eventParams?: {},
+      eventParams?: { [key: string]: any },
       options?: firebase.analytics.AnalyticsCallOptions
     ): void;
-    
+
     /**
      * Sends analytics event with given `eventParams`. This method
      * automatically associates this logged event with this Firebase web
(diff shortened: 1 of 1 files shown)


packages/firebase/index.d.ts
```diff
@@ -4076,6 +4090,20 @@ declare namespace firebase.analytics {
...
+     * Sends analytics event with given `eventParams`. This method
+     * automatically associates this logged event with this Firebase web
+     * app instance on this device.
+     * List of official event parameters can be found in
+     * {@link https://developers.google.com/gtagjs/reference/event
+     * the gtag.js reference documentation}.
+     */
+    logEvent(
+      eventName: 'page_view',
+      eventParams?: {},
```
hsubox76: For `page_view` I think it should be:

```
eventParams: {
        page_title?: string;
        page_location?: string;
        page_path?: string;
        [key: string]: any;
      }
```
Source: https://developers.google.com/analytics/devguides/collection/gtagjs/pages

packages/firebase/index.d.ts
```diff
@@ -3961,7 +3961,21 @@ declare namespace firebase.analytics {
...
+     * Sends analytics event with given `eventParams`. This method
+     * automatically associates this logged event with this Firebase web
+     * app instance on this device.
+     * List of official event parameters can be found in
+     * {@link https://developers.google.com/gtagjs/reference/event
+     * the gtag.js reference documentation}.
+     */
+    logEvent(
+      eventName: 'add_payment_info',
+      eventParams?: {},
```
hsubox76: In the case of `add_payment_info` it seems there's no specific recommended parameters, so can you just replace the `eventParams?: {}` with `eventParams?: { [key: string]: any }` so it will at least have an index signature?

cdoe:
Thanks for pointing that out, @hsubox76.  New commit corrects those two eventParams.  And sorry for some of the line-break weirdness.  I think some extra spaces were included in my first commit that are now removed with this new one.

PR #2277, opened 2019-10-18
Don't use process when not available

schmidt-sebastian:
This breaks a compile test I am doing with google-closure-compiler


packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -675,7 +675,8 @@ function convertStreamToken(token: ProtoByteString): string {
   if (token instanceof Uint8Array) {
     // TODO(b/78771403): Convert tokens to strings during deserialization
     assert(
-      process.env.USE_MOCK_PERSISTENCE === 'YES',
+      typeof process !== 'undefined' &&
```
wilhuff: Could be a function in SimpleDb? Something like `isMockPersistence` or `isUsingMockPersistence`?

schmidt-sebastian: That's a good idea.

PR #2281, opened 2019-10-18
[2/3] Query/Target split: make SyncEngine able to handle the mapping between Target and Query.

wu-hui:


wu-hui:
Tests for this PR will be in the next one, together with limitToLast implementation.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -925,28 +946,13 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
-          queryView
-        );
-        if (viewChange.snapshot) {
-          newViewSnapshots.push(viewChange.snapshot);
-        }
-      } else {
+      let queryData: QueryData | undefined = undefined;
+      const queries = this.queriesByTarget[targetId];
+
+      if (!queries || queries.isEmpty()) {
```
schmidt-sebastian: In general, a smaller diff is better and make this easier to review. Do you mind flipping the if-statement around again?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -958,9 +964,42 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+        // For queries that have a local View, we need to update their state
+        // in LocalStore (as the resume token and the snapshot version
+        // might have changed) and reconcile their views with the persisted
+        // state (the list of syncedDocuments may have gotten out of sync).
+        await this.localStore.releaseTarget(
+          targetId,
+          /*keepPersistedQueryData=*/ true
+        );
+
+        assert(
```
schmidt-sebastian: If I am not mistaken, since assert should never trigger since it matches the if-check in line 952? Is it meant to guard against the `releaseQuery` call above?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -958,9 +964,42 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+          !!this.queriesByTarget[targetId] &&
+            !this.queriesByTarget[targetId].isEmpty(),
+          `There is no queries mapped to target id ${targetId}`
+        );
+
+        const queries = this.queriesByTarget[targetId].toArray();
+        queryData = await this.localStore.allocateTarget(queries[0].toTarget());
```
schmidt-sebastian: I think this code (and the logic behing it) would be easier to follow if `releaseQuery` and `allocateQuery` would continue to happen right after each other.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -958,9 +964,42 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+            queryView!
+          );
+          if (viewChange.snapshot) {
+            newViewSnapshots.push(viewChange.snapshot);
+          }
+        }
       }
-      activeQueries.push(queryData);
+
+      assert(!!queryData, `Target data for target ${targetId} not found`);
```
schmidt-sebastian: I slightly prefer doing the wrong thing here and changing the queryData type back to just `QueryData` (rather than `QueryData|undefined`. queryData is always assigned to above and you could get rid of the assert, which is also slightly misleading since the LocalStore is meant to allocate QueryData, rather than retrieve it.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1040,18 +1077,20 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
       // removed if it has been rejected by the backend.
-      if (queryView) {
-        await this.localStore
-          .releaseQuery(queryView.query, /*keepPersistedQueryData=*/ false)
-          .then(() => {
-            this.remoteStore.unlisten(targetId);
-            this.removeAndCleanupQuery(queryView);
-          })
-          .catch(ignoreIfPrimaryLeaseLoss);
+      if (!this.queriesByTarget[targetId]) {
```
schmidt-sebastian: Above, you check for targetId existence and emptiness. It would be nice if we could guarantee that we only need to check for targetId existence.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -199,22 +200,26 @@ class LocalStoreTester {
   }
 
   afterAllocatingQuery(query: Query): LocalStoreTester {
+    return this.afterAllocatingTarget(query.toTarget());
```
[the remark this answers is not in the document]
wu-hui: I actually intend to keep it as a convenient method, otherwise we would have a lot of `Query.toTarget` calls in the tests.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -495,13 +500,11 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       );
       return this.applyRemoteEvent(event);
     } else {
-      const queryView = this.queryViewsByTarget[targetId];
-      assert(!!queryView, 'Unknown targetId: ' + targetId);
+      assert(!!this.queriesByTarget[targetId], 'Unknown targetId: ' + targetId);
```
schmidt-sebastian: In general, we only want to add assert that are needed for the client to function properly. Since `this.queriesByTarget[targetId]` is not accessed here, you can drop the assert.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I don't feel strongly. If what you have works and it's consistent, then we can probably keep it as is. If you wanted to make this more consistent, I would do something like this:

```
function getQueriesForTarget(targetId: number) {
  return this.queriesByTarget || [];
}

const queries = this.getQueriesForTarget(targetId);

if (queries.length > 0) {...}

```

But this just adds code, without too much benefit.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -270,7 +270,10 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
 
     const data = new QueryView(query, targetId, view);
     this.queryViewsByQuery.set(query, data);
-    this.queryViewsByTarget[targetId] = data;
+    if (!this.queriesByTarget[targetId]) {
+      this.queriesByTarget[targetId] = new Array<Query>();
```
schmidt-sebastian: Optional Nit: I prefer `this.queriesByTarget[targetId] = []; `since it is more concise and used in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

packages/firestore/src/core/sync_engine.ts
wu-hui: Done.

packages/firestore/src/core/sync_engine.ts
wu-hui: OK, I'll keep it as-is for now.

packages/firestore/src/core/sync_engine.ts
wu-hui: Done.

PR #2286, opened 2019-10-21
Add assertExists helper

schmidt-sebastian:
This should reduce code size and pain slightly by removing the need to write asserts for type-safety.

### DIFF
--- a/packages/firestore/src/util/assert.ts
+++ b/packages/firestore/src/util/assert.ts
@@ -48,3 +48,11 @@ export function assert(assertion: boolean, message: string): void {
     fail(message);
   }
 }
+
+/**
+ * Return the given value if it is neither null nor undefined, otherwise throws.
+ */
+export function assertExists<T>(value: T | undefined | null): T {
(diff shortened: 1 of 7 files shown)


packages/firestore/src/util/assert.ts
```diff
@@ -48,3 +48,11 @@ export function assert(assertion: boolean, message: string): void {
     fail(message);
   }
 }
+
+/**
+ * Return the given value if it is neither null nor undefined, otherwise throws.
```
wu-hui: `Returns`?

packages/firestore/src/util/assert.ts
```diff
@@ -48,3 +48,11 @@ export function assert(assertion: boolean, message: string): void {
...
+ */
+export function assertExists<T>(value: T | undefined | null): T {
```
wu-hui: This should take an optional message parameter? The default message does not provide a lot of value.

schmidt-sebastian:
Dropping this for now as we come up with an overall strategy for our Web asserts.

PR #2288, opened 2019-10-21
Retry all non-Firestore exceptions

schmidt-sebastian:
Our users are still seeing IndexedDB errors on iOS 13 (using IONIC/Cordova). One of the possible reasons is that some environments might not throw DOMExceptions, which would lead us to not retry. This PR flips our logic around and only skips retries if the underlying exception is a non-Firestore exception.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2232

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -298,6 +298,9 @@ export class SimpleDb {
         // TODO(schmidt-sebastian): We could probably be smarter about this and
         // not retry exceptions that are likely unrecoverable (such as quota
         // exceeded errors).
+        
+        // Note: We cannot use an instanceof check for FirestoreException, since the
+        // exception is wrapped in a generic error by our async/await handling.
         const retryable =
           idempotent &&
           error.name !== 'FirebaseError' &&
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -294,23 +294,23 @@ export class SimpleDb {
...
         return transactionFnResult;
-      } catch (e) {
+      } catch (error) {
         // TODO(schmidt-sebastian): We could probably be smarter about this and
         // not retry exceptions that are likely unrecoverable (such as quota
         // exceeded errors).
         const retryable =
           idempotent &&
-          isDomException(e) &&
+          error.name !== 'FirebaseError' &&
```
schmidt-sebastian: Note that the instanceof check doesn't work here since the Async/Await wraps our FirestoreError in a generic Error.

wilhuff: This might be worthwhile to add as an actual comment to the code rather than just the PR.

schmidt-sebastian: Done

PR #2290, opened 2019-10-21
rxfire: docs update, add collectionData() method usage

jmmendivil:
I use this method already, but noticed there is no docs about it.


jmmendivil:
@googlebot I signed it!

PR #2295, opened 2019-10-23
fix renovate config

Feiyang1:
Renovate seems to have a problem with regex(https://github.com/renovatebot/renovate/issues/4655). Our regex are causing config errors as a result, so I'm changing  to use a different field to workaround the issue.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,13 +7,11 @@
   },
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


Rebel13Lvt:
Okay keep us posted thank you

Naomi
Phoenix Legacy Photography ©️
2019

________________________________

PR #2297, opened 2019-10-25
Try API requests again if they failed in a different function call

mmermerkaya:
There isn't any useful error information if a request that was started in a different function call fails. Instead of throwing a useless error, try it again in the current function call.

Also makes the "missing app config values" error a little more helpful.

### DIFF
--- a/packages/installations/src/functions/get-id.ts
+++ b/packages/installations/src/functions/get-id.ts
@@ -19,7 +19,6 @@ import { FirebaseApp } from '@firebase/app-types';
 import { extractAppConfig } from '../helpers/extract-app-config';
 import { getInstallationEntry } from '../helpers/get-installation-entry';
 import { refreshAuthToken } from '../helpers/refresh-auth-token';
-import { RequestStatus } from '../interfaces/installation-entry';
 
 export async function getId(app: FirebaseApp): Promise<string> {
   const appConfig = extractAppConfig(app);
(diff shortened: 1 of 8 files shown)


packages/installations/src/functions/get-id.ts
```diff
@@ -28,14 +27,11 @@ export async function getId(app: FirebaseApp): Promise<string> {
...
-  }
-
-  if (installationEntry.registrationStatus === RequestStatus.COMPLETED) {
+    registrationPromise.catch(console.error);
+  } else {
     // If the installation is already registered, update the authentication
-    // token if needed. Suppress errors as they are not relevant to getId.
-    refreshAuthToken(appConfig).catch(() => {});
+    // token if needed.
+    refreshAuthToken(appConfig).catch(console.error);
```
dwoffinden: Huh, this never used to work without a .bind(console)

packages/installations/src/helpers/extract-app-config.ts
```diff
@@ -15,21 +15,44 @@
...
+    'projectId',
+    'apiKey',
+    'appId'
+  ];
 
-  if (!appName || !projectId || !apiKey || !appId) {
-    throw ERROR_FACTORY.create(ErrorCode.MISSING_APP_CONFIG_VALUES);
+  for (const keyName of configKeys) {
+    if (!app.options[keyName]) {
+      // Cast required becase FirebaseOptions includes "[name: string]: any;".
```
dwoffinden: :/

packages/installations/src/functions/get-id.ts
mmermerkaya: The future is now.

packages/installations/src/helpers/extract-app-config.ts
mmermerkaya: Be the change you want to see in the world: #2298.

dwoffinden: 👏🎉

PR #2300, opened 2019-10-26
[3/3] Spec tests for limitToLast+multi-tab

wu-hui:
I am finally able to make spec tests run against 1:N Target:Query mappings...Some cleanup is still needed, like `Watch*` methods in `spec_builder.ts` should work with `Target` not `Query`. I intend to do this with all other cleanups, in next PR.

Let me know if you have more spec tests ideas.

### DIFF
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { Query } from '../../../src/core/query';
-import { deletedDoc, doc, filter, path } from '../../util/helpers';
+import { deletedDoc, doc, filter, path, orderBy } from '../../util/helpers';
 
 import { TimerId } from '../../../src/util/async_queue';
 import { Code } from '../../../src/util/error';
@@ -554,4 +554,87 @@ describeSpec('Limbo Documents:', [], () => {
         })
     );
   });
+
+  specTest(
+    'LimitToLast query from secondary results in no expected limbo doc',
+    ['multi-client'],
+    () => {
+      const limitToLast = Query.atPath(path('collection'))
+        .addOrderBy(orderBy('val', 'desc'))
+        .withLimitToLast(3);
+      const docA = doc('collection/a', 1000, { val: 11 });
+      const docB = doc('collection/b', 1000, { val: 12 });
+      const docC = doc('collection/c', 1000, { val: 13 });
+
+      const docA2 = doc('collection/a', 2000, { val: 5 });
+      const docB2 = doc('collection/b', 2000, { val: 6 });
+      const docD2 = doc('collection/d', 2000, { val: 7 });
+
+      return (
+        client(0)
+          .becomeVisible()
+          .client(1)
+          .userListens(limitToLast)
+          .client(0)
+          .expectListen(limitToLast)
+          .watchAcksFull(limitToLast, 1000, docA, docB, docC)
+          .client(1)
+          .expectEvents(limitToLast, { added: [docC, docB, docA] })
+          .client(0)
+          .watchResets()
+          .watchSends({ affects: [limitToLast] }, docA2, docB2, docD2)
+          .watchCurrents(limitToLast, 'resume-token-2000')
+          .watchSnapshots(2000)
+          .client(1)
+          .expectEvents(limitToLast, {
+            added: [docD2],
+            removed: [docC],
+            modified: [docB2, docA2]
+          })
+          .client(0)
+          // docC dropped out of limit in both local and backend result, hence
+          // it's not a limbo doc.
+          .expectLimboDocs()
+      );
+    }
+  );
+
+  specTest(
+    'LimitToLast query from secondary results in expected limbo doc',
+    ['multi-client'],
+    () => {
+      const limitToLast = Query.atPath(path('collection'))
+        .addOrderBy(orderBy('val', 'desc'))
+        .withLimitToLast(3);
+      const docA = doc('collection/a', 1000, { val: 11 });
+      const docB = doc('collection/b', 1000, { val: 12 });
+      const docC = doc('collection/c', 1000, { val: 13 });
+
+      const docA2 = doc('collection/a', 2000, { val: 11 });
+      const docB2 = doc('collection/b', 2000, { val: 12 });
+      const docD2 = doc('collection/d', 2000, { val: 100 });
+
+      return (
+        client(0)
+          .becomeVisible()
+          .client(1)
+          .userListens(limitToLast)
+          .client(0)
+          .expectListen(limitToLast)
+          .watchAcksFull(limitToLast, 1000, docA, docB, docC)
+          .client(1)
+          .expectEvents(limitToLast, { added: [docC, docB, docA] })
+          .client(0)
+          .watchResets()
+          .watchSends({ affects: [limitToLast] }, docA2, docB2, docD2)
+          .watchCurrents(limitToLast, 'resume-token-2000')
+          .watchSnapshots(2000)
+          // docC dropped out of limit in from backend result, but still
+          // in local results, so it is in limbo now.
+          .expectLimboDocs(docC.key)
+          .client(1)
+          .expectEvents(limitToLast, { fromCache: true })
+      );
+    }
+  );
 });
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -196,7 +196,7 @@ describeSpec('Limbo Documents:', [], () => {
     const fullQuery = Query.atPath(path('collection'));
     const limitQuery = Query.atPath(path('collection'))
       .addFilter(filter('include', '==', true))
-      .withLimit(1);
+      .withLimitToFirst(1);
     const docA = doc('collection/a', 1000, { key: 'a', include: true });
     const docB = doc('collection/b', 1000, { key: 'b', include: true });
     const docBQuery = Query.atPath(docB.key.path);
@@ -266,7 +266,7 @@ describeSpec('Limbo Documents:', [], () => {
       const fullQuery = Query.atPath(path('collection'));
       const limitQuery = Query.atPath(path('collection'))
         .addFilter(filter('include', '==', true))
-        .withLimit(1);
+        .withLimitToFirst(1);
       const docA = doc('collection/a', 1000, { key: 'a', include: true });
       const docB = doc('collection/b', 1000, { key: 'b', include: true });
       const docBQuery = Query.atPath(docB.key.path);
--- a/packages/firestore/test/unit/specs/limbo_spec.test.ts
+++ b/packages/firestore/test/unit/specs/limbo_spec.test.ts
@@ -568,7 +568,8 @@ describeSpec('Limbo Documents:', [], () => {
 
       const docA2 = doc('collection/a', 2000, { val: 5 });
       const docB2 = doc('collection/b', 2000, { val: 6 });
-      const docD2 = doc('collection/d', 2000, { val: 7 });
+
+      const docD = doc('collection/d', 2000, { val: 7 });
 
       return (
         client(0)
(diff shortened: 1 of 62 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -886,6 +886,127 @@ describeSpec('Listens:', [], () => {
...
+      const docC = doc('collection/c', 2000, { val: 0 });
+
+      return client(0)
+        .becomeVisible()
+        .client(1)
+        .userListens(limit)
+        .userListens(limitToLast)
+        .client(0)
+        .expectListen(limit)
+        .expectListen(limitToLast)
```
schmidt-sebastian: Should this not just a be a single listen?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -886,6 +886,127 @@ describeSpec('Listens:', [], () => {
...
+        .client(1)
+        .expectEvents(limit, { added: [docA, docB] })
+        .expectEvents(limitToLast, { added: [docB, docA] })
+        .userUnlistens(limit)
+        .client(0)
+        .expectUnlisten(limit)
+        .watchSends({ affects: [limitToLast] }, docC)
+        .watchSnapshots(2000)
+        .client(1)
+        .expectEvents(limitToLast, { added: [docC], removed: [docB] });
```
schmidt-sebastian: What happens if the user unlistens from the mirror query?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -886,6 +886,127 @@ describeSpec('Listens:', [], () => {
...
+        .expectEvents(limit, { added: [docA, docB] })
+        .client(1)
+        .expectEvents(limitToLast, { added: [docB, docA] })
+        .userUnlistens(limitToLast)
+        .client(0)
+        .expectUnlisten(limitToLast)
+        .watchSends({ affects: [limit] }, docC)
+        .watchSnapshots(2000)
+        .expectEvents(limit, { added: [docC], removed: [docB] })
+        .client(1)
```
schmidt-sebastian: Should we split this up into a separate spec test ("User listens, unlistens and re-listens to mirror query")

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -116,30 +118,30 @@ class CachedTargetIdGenerator {
    * target ID has ever been assigned.
    */
   next(query: Query): TargetId {
```
schmidt-sebastian: Did you play around with this to see what it would take to make `next`, `purged` and `cachedId` take a Target as an argument? It looks like that might be cleaner.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -886,6 +886,127 @@ describeSpec('Listens:', [], () => {
...
+        .expectListen(limitToLast)
+        .userUnlistens(limit)
+        .expectUnlisten(limit)
+        .watchSends({ affects: [limitToLast] }, docD)
+        .watchSnapshots(3000)
+        .client(1)
+        .expectEvents(limitToLast, { added: [docD], removed: [docC] });
+    }
+  );
+
```
schmidt-sebastian: I think we now also have the ability to spec test the scenario I pointed out in https://docs.google.com/document/d/1u-N-tTBTaMe6k2XTbByYd8CBukgsKf7ijZ7pGhm_NpM. Do you want to give it a stab?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -266,15 +258,15 @@ export class SpecBuilder {
 
   userUnlistens(query: Query): this {
     this.nextStep();
-    if (!objUtils.contains(this.queryMapping, query.canonicalId())) {
+    if (!objUtils.contains(this.queryMapping, query.toTarget().canonicalId())) {
```
schmidt-sebastian: Nit: This could be simplified if you asserted that targetId is not undefined after line 264. I am mostly just pointing this out to see if we can remove calls to `toTarget()`.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -108,6 +109,7 @@ export class ClientMemoryState {
  * active in multiple tabs.
  */
 class CachedTargetIdGenerator {
+  // TODO(wuandy): rename this to targetMapping.
   private queryMapping: QueryMap = {};
```
schmidt-sebastian: Should we make this an ObjectMap<Target, number>? We should then be able to remove most calls to `canonicalId()`.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -458,11 +451,12 @@ export class SpecBuilder {
       if (!objUtils.contains(this.limboMapping, path)) {
         this.limboMapping[path] = this.limboIdGenerator.next();
       }
-      this.activeTargets[this.limboMapping[path]] = {
-        query: SpecBuilder.queryToSpec(Query.atPath(key.path)),
-        // Limbo doc queries are currently always without resume token
-        resumeToken: ''
-      };
+      // Limbo doc queries are currently always without resume token
+      this.addQueryToActiveTargetsIfNotYet(
```
schmidt-sebastian: Limbo queries are single document queries that request the latest state of a document. Since we know that the document is out of sync, it probably would not make sense to use resume tokens.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -936,14 +933,55 @@ export class SpecBuilder {
     }
   }
 
+  private addQueryToActiveTargetsIfNotYet(
```
schmidt-sebastian: To be honest, I am not the biggest fan of this method name. Do you have alternatives?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -936,14 +933,55 @@ export class SpecBuilder {
...
+  ): void {
+    if (this.activeTargets[targetId]) {
+      const specQueriesClone = [...this.activeTargets[targetId].queries];
+      // query is not added to active targets yet.
+      if (
+        !specQueriesClone.some(specQuery =>
+          parseQuery(specQuery).isEqual(query)
+        )
+      ) {
+        specQueriesClone.push(SpecBuilder.queryToSpec(query));
```
schmidt-sebastian: This would flow a bit better if line 942, 943, and 951 onwards were inside of the if statement. You could even inline specQueriesClone.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -936,14 +933,55 @@ export class SpecBuilder {
...
+    } else {
+      this.activeTargets[targetId] = {
+        queries: [SpecBuilder.queryToSpec(query)],
+        resumeToken: resumeToken || ''
+      };
+    }
+  }
+
+  private removeQueryFromActiveTargets(query: Query, targetId: number): void {
+    const specQueries = this.activeTargets[targetId].queries.filter(
```
schmidt-sebastian: As above, this should be inside the if statement.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1566,6 +1567,8 @@ export interface SpecWatchFilter
   '1': string | undefined;
 }
 
+export type SpecLimitType = 'LimitToFirst' | 'LimitToLast';
```
schmidt-sebastian: Could this use the Enum type that you defined in your previous PR (which already has a String mapping)?

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Good idea, done.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Done.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: No longer calling canonicalId.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: I thought this is what the code does here? I did not change this..

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Done.

I still keep the temp variable to save some typing.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: I intend to keep this unchanged, except a better name. This cannot be in if statement, given the way it is written..

packages/firestore/test/unit/specs/spec_test_runner.ts
wu-hui: Hmm, I am merely following other cases (filter, orderBy) here that they all have separate types instead of using prod types. I thought this was by design?

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -554,4 +554,87 @@ describeSpec('Limbo Documents:', [], () => {
...
+      const limitToLast = Query.atPath(path('collection'))
+        .addOrderBy(orderBy('val', 'desc'))
+        .withLimitToLast(3);
+      const docA = doc('collection/a', 1000, { val: 11 });
+      const docB = doc('collection/b', 1000, { val: 12 });
+      const docC = doc('collection/c', 1000, { val: 13 });
+
+      const docA2 = doc('collection/a', 2000, { val: 5 });
+      const docB2 = doc('collection/b', 2000, { val: 6 });
+      const docD2 = doc('collection/d', 2000, { val: 7 });
```
schmidt-sebastian: Nit: This would be easier to read if this was:

```
const docA = doc('collection/a', 1000, { val: 11 });
const docB = doc('collection/b', 1000, { val: 12 });
const docC = doc('collection/c', 1000, { val: 13 });

const docA2 = doc('collection/a', 2000, { val: 5 });
const docB2 = doc('collection/b', 2000, { val: 6 });

const docD = doc('collection/d', 2000, { val: 7 }); // No "2" and space
```

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -975,35 +979,46 @@ describeSpec('Listens:', [], () => {
...
+          .client(1)
+          .userListens(limitToLast)
+          .client(0)
+          .expectListen(limit)
+          .expectListen(limitToLast)
+          .watchAcksFull(limit, 1000, docA, docB)
+          .expectEvents(limit, { added: [docA, docB] })
+          .client(1)
+          .expectEvents(limitToLast, { added: [docB, docA] })
+          // Secondary tab unlistens it's query
```
schmidt-sebastian: s/it's/from its/

(here and below)

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -936,14 +932,58 @@ export class SpecBuilder {
...
+      const activeQueries = this.activeTargets[targetId].queries;
+      if (
+        !activeQueries.some(specQuery => parseQuery(specQuery).isEqual(query))
+      ) {
+        this.activeTargets[targetId] = {
+          queries: [SpecBuilder.queryToSpec(query), ...activeQueries],
+          resumeToken: resumeToken || ''
+        };
+      } else {
+        this.activeTargets[targetId] = {
```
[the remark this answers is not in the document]
wu-hui: See comment below.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -936,14 +932,58 @@ export class SpecBuilder {
...
+          queries: [...activeQueries],
```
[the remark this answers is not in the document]
wu-hui: Your suggestion leads to test failures, looks like the resumeToken get be updated by future steps this way.

I prefer the current way where it is copied everywhere..this is the trickiest thing about spec tests.

schmidt-sebastian:
Unit tests are generally much more reliable and you can fine-tune all input. I think you will find that having them will make your current and future development much simpler. Plus, you have all the tests already and just need to rip out the multi-tab part.

PR #2301, opened 2019-10-27
Use string concatenation instead of array join for making RPC URLs

httpsOmkar:
From array to string template converted for better readability

### DIFF
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -415,6 +415,16 @@ export class WebChannelConnection implements Connection {
   makeUrl(rpcName: string): string {
     const urlRpcName = RPC_NAME_REST_MAPPING[rpcName];
     assert(urlRpcName !== undefined, 'Unknown REST mapping for: ' + rpcName);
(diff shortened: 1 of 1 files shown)


packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -415,16 +415,6 @@ export class WebChannelConnection implements Connection {
...
-    url.push(this.databaseId.projectId);
-
-    url.push('/databases/');
-    url.push(this.databaseId.database);
-    url.push('/documents');
-
-    url.push(':');
-    url.push(urlRpcName);
-    return url.join('');
+    return `${ this.baseUrl }/${ RPC_URL_VERSION }/projects/${ this.databaseId.projectId }/databases/${ this.databaseId.database }/documents:${ urlRpcName }`;
```
wilhuff: Is there any way to break this up? 159 characters is almost double the allowed line length.

httpsOmkar: yeah sure

httpsOmkar: prettier is reverting my changes

wilhuff: FWIW, I think the prior arrangement was essentially a workaround for this disconnect.

What about splitting with a `+`?

httpsOmkar: fixed

PR #2307, opened 2019-10-30
Don't fail client if lease refresh fails

schmidt-sebastian:
This PR:
- Makes the lease refresh idempotent by moving the `primaryStateListener` callback outside of the transaction function. Note that it call still get called with `primaryStateListener(false)`, as calling that repeatedly should not have any side effects.
- Changes transactions to obtain the IndexedDB lease even if they don't already have it.

Addresses https://github.com/firebase/firebase-js-sdk/issues/2232

### DIFF
--- a/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
+++ b/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
@@ -933,6 +933,28 @@ describe('IndexedDb: canActAsPrimary', () => {
     simpleDb.close();
   }
 
+  async function getCurrentLeaseOwner(): Promise<ClientId | null> {
+    const simpleDb = await SimpleDb.openOrCreate(
+      INDEXEDDB_TEST_DATABASE_NAME,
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1035,6 +1057,23 @@ describe('IndexedDb: canActAsPrimary', () => {
...
+  it('regains lease if available', () => {
+    return withPersistence('clientA', async persistence => {
+      expect(await getCurrentLeaseOwner()).to.not.be.null;
+
+      await clearPrimaryLease();
+      expect(await getCurrentLeaseOwner()).to.be.null;
+
+      await persistence.runTransaction(
+        'regain lease',
+        'readwrite-primary',
```
wilhuff: readwrite-primary-idempotent?

schmidt-sebastian: Done.

Issue #2191, opened 2019-09-24
Realtime Database rules are not being loaded if launching tests with `emulators:exec`


evindor:
### [REQUIRED] Describe your environment

  * Operating System version: Mac OS 10.14.4, Alpine 3.9
  * Browser version: Irrelevant
  * Firebase SDK version: @firebase/testing@0.12.0 & up, firebase@6.4.0 & up
  * Firebase Product: database emulator, @firebase/testing

### [REQUIRED] Describe the problem
`firebase.loadDatabaseRules` does not load rules if tests are being run using `firebase emulators:exec`.
#### Steps to reproduce:
- Clone [database emulator example](https://github.com/firebase/quickstart-nodejs/tree/master/database-emulator/typescript-quickstart)
- Bump `@firebase/testing` to any version from `0.12.0` and up to `0.13.2`
- `yarn install`
- `firebase emulators:exec --only database "yarn test"`

3/4 tests will fail, reporting `Error: Expected request to fail, but it succeeded.` indicating that rules were not set by testing code and default emulator rules apply (allow all reads & writes).

Latest version of `@firebase/testing` which works fine is `0.11.10`

mikelehen:
cc/ @ryanpbrewster

ryanpbrewster:
This has something to do with the `FIREBASE_DATABASE_EMULATOR_HOST=localhost:9000` environment variable being set. I think it's automatically configuring the SDK to act as an admin.

yuchenshi:
This has been fixed `@firebase/testing`. Please `npm i -D @firebase/testing@latest` and that should fix it. I also bumped the version in the quickstarts.

PR #2308, opened 2019-10-30
Revert "use fake "owner" credential when rtdb emulator env var is set…

IanWyszynski:
… (#2029)"

This reverts commit 4ebedd071ddd4f4f7ab427a87fa6a725622225c3.

### Discussion

https://github.com/firebase/firebase-js-sdk/issues/2191 is caused by the fact that the `FIREBASE_DATABASE_EMULATOR_HOST` environment variable is set by the `emulators:exec` command automatically. This causes the firebase-js-sdk to send the database emulator an administrative access token and therefore breaks the quickstart tests.

Reverting this behavior will fix the testing SDK by sending the emulator a production access token. Though this is not ideal, we have long term plans to introduce an API that will allow the various SDKs to specify a custom access token at app initialization time.

### Testing

I ran the `@firebase/testing` package tests in this repo, and worked through the reproduction steps described in the issue.


packages/database/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@firebase/database",
-  "version": "0.5.9",
+  "version": "0.5.10",
```
schmidt-sebastian: We should not set our versions manually. The JS SDK team will do so for us.

IanWyszynski:
> LGTM. Pretty clean rollback. I'd like another person to look at this though.
> 
> Also, what do people feel about also bumping `@firebase/testing` so we can tell the affected developers to upgrade to `@firebase/testing@>=x.x.x`?

per @schmidt-sebastian comment, I'd defer to members of the firebase-js-sdk on this. I think it's a good idea.

PR #2313, opened 2019-10-31
Add spec test for single client

wu-hui:
Also fixes a bug in spec test builder.

### DIFF
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -1016,16 +1016,19 @@ describeSpec('Listens:', [], () => {
           // are unlistened by now.
           .userUnlistens(limitToLast)
           .client(0)
-          // TODO(b/143693491) `expectListen` would also pass, which is wrong.
-          // the subsequent `expectActiveTargets` makes sure the target is
-          // removed, but we still need to fix `spec_test_runner`.
+          // TODO(b/143693491) If we use `expectListen` here, the test would
+          // also pass, which is wrong. The reason is `TestRunner` only check
+          // the expected Queries against the actual target, in the case of
+          // mirror queries, both query will be able to find an actual target.
+          // We need to change `TestRunner` to track actual client queries
+          // being listened to in addition to the targets to fix this.
           .expectUnlisten(limitToLast)
           .expectActiveTargets()
       );
     }
   );
 
-  specTest('Mirror Queries in primary client', [], () => {
+  specTest('Can listen/unlisten to mirror queries.', [], () => {
     const limit = Query.atPath(path('collection'))
       .addOrderBy(orderBy('val', 'asc'))
       .withLimitToFirst(2);
@@ -1045,15 +1048,16 @@ describeSpec('Listens:', [], () => {
         .watchAcksFull(limit, 1000, docA, docB)
         .expectEvents(limit, { added: [docA, docB] })
         .expectEvents(limitToLast, { added: [docB, docA] })
-        // Un-listen to limitToLast
         .userUnlistens(limitToLast)
         .expectUnlisten(limitToLast)
         .watchSends({ affects: [limit] }, docC)
+        .watchCurrents(limit, 'resume-token-2000')
         .watchSnapshots(2000)
         .expectEvents(limit, { added: [docC], removed: [docB] })
-        // Re-listens to limitToLast
         .userListens(limitToLast)
         .expectListen(limitToLast)
+        // Note the result is not from cache because the target is kept
+        // alive since `limit` is still being listened to.
         .expectEvents(limitToLast, { added: [docC, docA] })
         // Backend fails the query.
         .watchRemoves(
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -1018,10 +1018,10 @@ describeSpec('Listens:', [], () => {
           .client(0)
           // TODO(b/143693491) If we use `expectListen` here, the test would
           // also pass, which is wrong. The reason is `TestRunner` only check
-          // the expected Queries against the actual target, in the case of
-          // mirror queries, both query will be able to find an actual target.
-          // We need to change `TestRunner` to track actual client queries
-          // being listened to in addition to the targets to fix this.
+          // the expected Queries against the actual target. In the case of
+          // mirror queries, both queries will be able to find an actual target.
+          // We need to change `TestRunner` to track the actual client queries
+          // in addition to the targets to fix this.
           .expectUnlisten(limitToLast)
           .expectActiveTargets()
       );
--- a/packages/firestore/test/unit/specs/listen_spec.test.ts
+++ b/packages/firestore/test/unit/specs/listen_spec.test.ts
@@ -1016,11 +1016,55 @@ describeSpec('Listens:', [], () => {
           // are unlistened by now.
           .userUnlistens(limitToLast)
           .client(0)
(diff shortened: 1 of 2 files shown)


packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,56 @@ describeSpec('Listens:', [], () => {
...
+          // TODO(b/143693491) `expectListen` would also pass, which is wrong.
+          // the subsequent `expectActiveTargets` makes sure the target is
+          // removed, but we still need to fix `spec_test_runner`.
+          .expectUnlisten(limitToLast)
           .expectActiveTargets()
       );
     }
   );
 
+  specTest('Mirror Queries in primary client', [], () => {
```
schmidt-sebastian: Suggest to drop "primary client" form here as this name won't be applicable to iOS and Android

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,56 @@ describeSpec('Listens:', [], () => {
           // are unlistened by now.
           .userUnlistens(limitToLast)
           .client(0)
-          .expectListen(limitToLast)
+          // TODO(b/143693491) `expectListen` would also pass, which is wrong.
+          // the subsequent `expectActiveTargets` makes sure the target is
+          // removed, but we still need to fix `spec_test_runner`.
+          .expectUnlisten(limitToLast)
```
schmidt-sebastian: Can you rewrite this comment? I am not quite following :/

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,56 @@ describeSpec('Listens:', [], () => {
...
+      spec()
+        .userListens(limit)
+        .expectListen(limit)
+        .userListens(limitToLast)
+        .expectListen(limitToLast)
+        .watchAcksFull(limit, 1000, docA, docB)
+        .expectEvents(limit, { added: [docA, docB] })
+        .expectEvents(limitToLast, { added: [docB, docA] })
+        // Un-listen to limitToLast
+        .userUnlistens(limitToLast)
```
schmidt-sebastian: FWIW, these comments describe the code as is and not the intention behind the calls. I suggest you drop them.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,56 @@ describeSpec('Listens:', [], () => {
...
+        .expectUnlisten(limitToLast)
+        .watchSends({ affects: [limit] }, docC)
+        .watchSnapshots(2000)
+        .expectEvents(limit, { added: [docC], removed: [docB] })
+        // Re-listens to limitToLast
+        .userListens(limitToLast)
+        .expectListen(limitToLast)
+        .expectEvents(limitToLast, { added: [docC, docA] })
```
schmidt-sebastian: Oh, this is interesting. Is this event not marked from cache since the target is already active? This would be something to note in a comment :)

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,56 @@ describeSpec('Listens:', [], () => {
...
+        .expectEvents(limitToLast, { added: [docB, docA] })
+        // Un-listen to limitToLast
+        .userUnlistens(limitToLast)
+        .expectUnlisten(limitToLast)
+        .watchSends({ affects: [limit] }, docC)
+        .watchSnapshots(2000)
+        .expectEvents(limit, { added: [docC], removed: [docB] })
+        // Re-listens to limitToLast
+        .userListens(limitToLast)
+        .expectListen(limitToLast)
```
schmidt-sebastian: This is kind of strange since it should either not exist or have a resume token. Can you add a comment to b/143693491 that we should clean this up (ideally by removing the expectListen).

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Should be slightly improved?

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: We can remove `expectListen` here, because it is actually checking actual target, as explained above. 

I want to keep it still, such that when `expectListen` gets fixed, this can be one of the case to cover.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

Yeah, mirror queries map to the same target, and as long as one of them is alive, the data is kept update to date.

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,60 @@ describeSpec('Listens:', [], () => {
           // are unlistened by now.
           .userUnlistens(limitToLast)
           .client(0)
-          .expectListen(limitToLast)
+          // TODO(b/143693491) If we use `expectListen` here, the test would
+          // also pass, which is wrong. The reason is `TestRunner` only check
+          // the expected Queries against the actual target, in the case of
```
schmidt-sebastian: s/, in/. In

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,60 @@ describeSpec('Listens:', [], () => {
...
+          // mirror queries, both query will be able to find an actual target.
```
schmidt-sebastian: s/query/queries/

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,60 @@ describeSpec('Listens:', [], () => {
...
+          // We need to change `TestRunner` to track actual client queries
```
schmidt-sebastian: s/actual/the actual/

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1016,12 +1016,60 @@ describeSpec('Listens:', [], () => {
...
+          // being listened to in addition to the targets to fix this.
```
schmidt-sebastian: I'd drop "being listened to"

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

packages/firestore/test/unit/specs/listen_spec.test.ts
wu-hui: Done.

PR #2316, opened 2019-11-01
Component framework implementation

Feiyang1:
Implements go/firebase-js-platform


packages/component/src/component_container.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export class ComponentContainer {
+  private providers = new Map<string, Provider>();
+
+  constructor(private name: string) {}
+
+  /**
+   *
+   * @param component Component being added
+   * @param overwrite When a component with the same name has already been registered,
```
mmermerkaya: What about two methods, like `addComponent` and `overwriteComponent`?

packages/component/src/component_container.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.providers.delete(component.name);
+        // create a new provider
+        provider = this.getProvider(component.name);
+      }
+    }
+
+    provider.setComponent(component);
+  }
+
+  getProvider<T>(name: string): Provider<T> {
```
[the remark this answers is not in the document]
Feiyang1: Good point. We actually don't need the generics anymore because individual components define overloads.
so we  just have `getProvider(name: string): Provider<unknown>` here, WDYT?

packages/component/src/component_container.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    // create a Provider for a service that hasn't registered with Firebase
+    const provider = new Provider(name, this);
+    this.providers.set(name, provider);
+
+    return provider as Provider<T>;
+  }
+
+  getProviders(): Provider[] {
+    return Array.from(this.providers, entry => entry[1]);
```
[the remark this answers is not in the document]
Feiyang1: sweet!

packages/component/src/provider.ts
```diff
@@ -0,0 +1,161 @@
...
+    // if multipleInstances is not supported, use the default name
+    const normalizedIdentifier = this.normalizeInstanceIdentifier(identifier);
+
+    const instance = this.getOrInitializeService(normalizedIdentifier);
+
+    if (!instance) {
+      if (options && options.optional) {
+        return null;
+      }
+      throw Error(`Service ${this.name} is not available`);
```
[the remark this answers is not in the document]
Feiyang1: `getImmediate()` is intended for 2 workflows: 
1. Getting a required dependency. Throw if the dependency is not available.
2. Getting an optional dependency synchronously. Do things differently depending on the null check.

A missing required dependency should always throw an exception IMO, so having a central place to do that achieves it and simplifies the caller's code. So I'd like to keep `optional` to support the 2 use cases.

packages/component/src/provider.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  clearInstance(identifier: string = DEFAULT_ENTRY_NAME): void {
+    this.instancesDeferred.delete(identifier);
+    this.instances.delete(identifier);
+  }
+
+  // app.delete() will call this method on every provider to delete the services
+  // TODO: should we mark the provider as deleted?
+  delete(): Promise<unknown[]> {
```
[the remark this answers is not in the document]
Feiyang1: That's true. Changed to `Promise<void>`

packages/component/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Factory to create an instance of type T, given a ComponentContainer.
+ * ComponentContainer is the IOC container that provides {@link Provider}
+ * for dependencies.
+ *
+ * NOTE: The container only provides {@link Provider} rather than the actual instances of dependencies.
+ * It is useful for lazily loaded dependencies and optional dependencies.
+ */
+export type InstanceFactory<T = unknown> = (
```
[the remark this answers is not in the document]
Feiyang1: IIUC, return only generic only applies to the concrete function declaration.
Here It's just a generic type to help create concrete types which are in turn used to type functions.

packages/component/src/component_container.ts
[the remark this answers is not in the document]
Feiyang1: I went ahead and removed the generic from the function declaration.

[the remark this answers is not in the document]
mmermerkaya: Nice, thanks.

This is perfectly fine already, but I think that if the components defined the `NameServicePairs`, it would be more type safe (you wouldn't be able to call this with any string) and it would have the same complexity in the extending components, since they're declaration merging anyway. They'd just extend the interface instead of overloading this method.

Also described here (kind of): http://go/typescript/return_only_generics#mapped-types

packages/component/src/types.ts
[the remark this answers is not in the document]
mmermerkaya: Yeah that's true, I just realized how it's actually used. :thinking: :sweat_smile: 

You could make this `InstanceFactory<T extends Service>` if you do `NameServicePairs`, but if not, this is fine too.

packages/component/src/component_container.ts
mmermerkaya: I usually don't like flag arguments, since it's harder to tell what it's doing when you're reading code that calls this function (or you have to add comments like `addComponent(component, /* overWrite */ true);`).

I guess it's okay if it's mostly for testing, but TBH I'd rather not use the same app instance across tests :smile:. Up to you.

packages/component/src/provider.ts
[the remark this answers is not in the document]
mmermerkaya: My thinking was that removing an argument would simplify the caller's code, but what you said makes sense as well. SGTM, let's leave it like this.

packages/component/src/component_container.ts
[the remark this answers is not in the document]
Feiyang1: Done. PTAL.

packages/component/src/types.ts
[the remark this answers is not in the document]
Feiyang1: I think we can keep this type independent of  the `NameServicePairs`. To make everything 100% type safe, we'd need to tie `Component` with `NameServicePairs` as well. I feel it might be too restricting because people should be able to use `Component` without doing extra things.

packages/component/src/component_container.ts
Feiyang1: Yeah, flag arguments are not ideal. Added a new method `addOrOverwriteComponent()`.

`database` tests share `app` a lot, but I don't feel like changing it at the moment. Also there is a request from `@firebase/testing` to be able to overwrite `auth-internal` component inside an initialized app that has the default `auth-internal`.

mmermerkaya: Fair enough, that sounds like a lot of work.

packages/component/src/types.ts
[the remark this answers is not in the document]
mmermerkaya: Isn't a `Component` supposed to be implementing a single service? It feels like a perfect match to me, you could define `Component` as `export class Component<T extends Name> {`, and still all a service would need to do is to extend the `NameServicePairs` interface, which they're already doing AFAICT. But TBH I might not be aware of all the use cases of `Component`.

packages/component/src/component_container.ts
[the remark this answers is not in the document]
mmermerkaya: Looks good, thanks. :slightly_smiling_face:

packages/analytics/index.ts
```diff
@@ -59,7 +60,7 @@ export function registerAnalytics(instance: _FirebaseNamespace): void {
             .getImmediate();
           return {
             logEvent: analytics.logEvent
-          };
+          } as FirebaseAnalyticsInternal;
```
mmermerkaya: I think you can assign the factory method to a const before calling `registerComponent`, and set the type to `InstanceFactory` there instead of casting.

packages/component/src/component_container.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { Provider } from './provider';
+import { Component } from './component';
+import { Name } from './types';
+
+/**
+ * ComponentContainer that provides Providers for service name T, e.g. `auth`, `auth-internal`
+ */
+export class ComponentContainer {
+  private readonly providers = new Map<string, Provider>();
```
mmermerkaya: When the types of values are different and known, instead of a `Map` I prefer using an object, which allows you to type the values independently. Something like:

```typescript
type ProviderMap = { [T in Name]?: Provider<T> };

const providers: ProviderMap = {};
```

And then you should be able to remove the casts in `getProvider`.

packages/component/src/types.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface Dictionary {
+  [key: string]: unknown;
+}
+
+/**
+ * This interface will be extended by Firebase SDKs to provide service name and service type mapping.
+ * It is used as a generic constraint to ensure type safety.
+ */
+export interface NameServiceMapping {
+  'app': string;
```
mmermerkaya: I don't get this one :sweat_smile: Why is app a string?

packages/app/src/firebaseApp.ts
```diff
@@ -152,28 +122,15 @@ export class FirebaseAppImpl implements FirebaseApp {
...
-      );
-      this.services_[name][instanceIdentifier] = service;
-    }
-
-    return this.services_[name][instanceIdentifier];
+    // getImmediate will always succeed because _getService is only called for registered components.
+    return (
+      (this.container
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        .getProvider(name as any)
```
mmermerkaya: If you type `name` as `Name`, does it work? I was thinking that the compiler wouldn't complain and that it would understand that it is a `Name` even though it doesn't know all the members in the type `Name`.

Feiyang1: Yes, but I need to cast it in a different place when calling the function :(

packages/component/src/types.ts
Feiyang1: oops, testing code.

packages/component/src/provider.ts
```diff
@@ -0,0 +1,179 @@
...
+import { ComponentContainer } from './component_container';
+import { DEFAULT_ENTRY_NAME } from './constants';
+import { InstantiationMode, Name, NameServiceMapping } from './types';
+import { Component } from './component';
+
+/**
+ * Provider for instance for service name T, e.g. 'auth', 'auth-internal'
+ * U is an alias for the type of the instance
+ */
+export class Provider<T extends Name = Name, U = NameServiceMapping[T]> {
```
mmermerkaya: I think the existence of `U` here might confuse the compiler, making it think that U can actually be set to something else than `NameServiceMapping[T]`. Since we'd never want something like `Provider<'app', FirebaseInstallations>` I think we should remove `U` and replace its uses with `NameServiceMapping[T]`, even though it makes the code below more verbose.

Also I'd rather remove the default value for T, and type this as `Provider<Name>` explicitly where needed, which I  assume isn't that many places.

packages/component/src/types.ts
```diff
@@ -0,0 +1,64 @@
...
+ * PUBLIC: A public component provides a set of public APIs to customers. A service namespace will be patched
+ * onto `firebase` namespace. Assume the component name is `test`, customers will be able
+ * to get the service by calling `firebase.test()` or `app.test()` where `app` is a `FirebaseApp` instance.
+ *
+ * PRIVATE: A private component provides a set of private APIs that are used internally by other
+ * Firebase SDKs. No serivce namespace is created in `firebase` namespace and customers have no way to get them.
+ */
+export const enum ComponentType {
+  PUBLIC = 'PUBLIC',
+  PRIVATE = 'PRIVARE'
```
hsubox76: Typo: 'PRIVARE'

packages/analytics/index.ts
Feiyang1: I removed it and there is no error. It's probably something I added before the type safety changes.

packages/component/src/component_container.ts
Feiyang1: Hmm, I got an error without casting - 
`'"test"' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint '"test" | "badtest"'.`

`test` and `badtest` are the service names I defined and used for testing.

packages/component/src/types.ts
Feiyang1: removed.

packages/app/src/firebaseApp.ts
Feiyang1: Casted it to `Name` and removed the eslint disable line, but it still needs casting because `FirebaseApp` doesn't implement `FirebaseService` interface.  🤷‍♂

packages/component/src/provider.ts
Feiyang1: Yeah, removed the generic. I wish there is a way to alias types within a class, something like:
```
class Provider<T extends Name = Name> {
   define type U = NameServiceMapping[T];
   // then use U as an alias in methods and properties
}
``` 🤷‍♂

packages/component/src/types.ts
Feiyang1: Good catch, fixed!

packages/component/src/component_container.ts
mmermerkaya: Just tried it myself, I think the problem is that without casting TypeScript thinks the values of the object are `Provider<'test'> | Provider<'badTest'>` etc. instead of `Provider<'test' | 'badTest'>`, but I might be wrong.

Sorry about that, yeah I think we should keep the `Map` :sweat_smile:

packages/app/src/firebaseApp.ts
mmermerkaya: > `FirebaseApp` doesn't implement `FirebaseService` interface

Haha, nothing we can do about that here.

PR #2323, opened 2019-11-01
Migrate messaging to component framework

Feiyang1:
### Context

go/firebase-js-platform
firebase/firebase-js-sdk#2316 implements the proposal.

### Changes

This PR changes `@firebase/messaging` to use the component platform implemented in firebase/firebase-js-sdk#2316

### Dependencies

firebase/firebase-js-sdk#2316
https://github.com/firebase/firebase-js-sdk/pull/2322


mmermerkaya:
Looks like there are a lot of dependencies being passed around. Can't we retrieve the components where they are needed, like we are doing now?

Feiyang1:
implemented the proposal where firebase services are bundled in a single object.

packages/messaging/src/controllers/base-controller.ts
```diff
@@ -43,11 +44,14 @@ export const TOKEN_EXPIRATION_MILLIS = 7 * 24 * 60 * 60 * 1000; // 7 days
 
 export abstract class BaseController implements FirebaseMessaging {
   INTERNAL: FirebaseServiceInternals;
+  readonly app!: FirebaseApp;
```
mmermerkaya: Looks like you don't need the `!` since you are defining it in the constructor.

packages/messaging/index.ts
```diff
@@ -16,41 +16,55 @@
...
   const messagingName = 'messaging';
 
-  const factoryMethod: FirebaseServiceFactory = app => {
+  const factoryMethod = (container: ComponentContainer): FirebaseService => {
+    /* Dependencies */
+    const app = container.getProvider('app').getImmediate();
+    const installations = container.getProvider('installations').getImmediate();
+    const analyticsProvider = container.getProvider('analytics-internal');
+
+    const firebaseServices = { app, installations, analyticsProvider };
```
mmermerkaya: Could you give the `FirebaseInternalServices` type to this object?

packages/messaging/src/interfaces/external-services.ts
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '@firebase/app-types';
+import { FirebaseInstallations } from '@firebase/installations-types';
+import { FirebaseAnalyticsInternal } from '@firebase/analytics-interop-types';
+import { Provider } from '@firebase/component';
+
+export interface FirebaseInternalServices {
```
mmermerkaya: Filename says external, this says internal :smile:

packages/messaging/src/controllers/base-controller.ts
Feiyang1: Done.

packages/messaging/src/interfaces/external-services.ts
Feiyang1: renamed the file to `internal-services.ts`

PR #2317, opened 2019-11-01
Migrate auth to component platform

Feiyang1:
### Context

go/firebase-js-platform
firebase/firebase-js-sdk#2316 implements the proposal.

### Changes

This PR changes `@firebase/auth` to use the component platform implemented in firebase/firebase-js-sdk#2316

PR #2326, opened 2019-11-01
Migrate storage to component framework

Feiyang1:
### Context

go/firebase-js-platform

### Changes

This PR changes `@firebase/storage` to use the component platform implemented in https://github.com/firebase/firebase-js-sdk/pull/2316

### Dependencies

### DIFF
--- a/packages/storage/src/implementation/authwrapper.ts
+++ b/packages/storage/src/implementation/authwrapper.ts
@@ -31,6 +31,8 @@ import {
   _FirebaseApp,
   FirebaseAuthTokenData
 } from '@firebase/app-types/private';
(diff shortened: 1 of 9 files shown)


packages/storage/src/implementation/authwrapper.ts
```diff
@@ -40,6 +42,7 @@ import {
  */
 export class AuthWrapper {
   private app_: FirebaseApp | null;
+  private authProvider: Provider<FirebaseAuthInternal>;
```
schmidt-sebastian: Nit: Match style and add underscore.

Feiyang1: Done.

PR #2327, opened 2019-11-01
Migrate Database to component framework

Feiyang1:
### Context

go/firebase-js-platform

### Changes

This PR changes `@firebase/database` to use the component platform implemented in https://github.com/firebase/firebase-js-sdk/pull/2316

### Dependencies

### DIFF
--- a/packages/database/index.node.ts
+++ b/packages/database/index.node.ts
@@ -16,7 +16,7 @@
  */
 
 import { FirebaseNamespace, FirebaseApp } from '@firebase/app-types';
-import { _FirebaseNamespace } from '@firebase/app-types/private';
+import { _FirebaseNamespace, _FirebaseApp } from '@firebase/app-types/private';
 import { Database } from './src/api/Database';
 import { DataSnapshot } from './src/api/DataSnapshot';
 import { Query } from './src/api/Query';
@@ -30,6 +30,13 @@ import { setSDKVersion } from './src/core/version';
 import { CONSTANTS, isNodeSdk } from '@firebase/util';
 import { setWebSocketImpl } from './src/realtime/WebSocketConnection';
 import { Client } from 'faye-websocket';
+import {
+  Component,
+  ComponentType,
+  Provider,
+  ComponentContainer
+} from '@firebase/component';
+import { FirebaseAuthInternal } from '@firebase/auth-interop-types';
 
 setWebSocketImpl(Client);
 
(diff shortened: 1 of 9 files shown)


packages/database/test/browser/crawler_support.test.ts
```diff
@@ -38,13 +38,13 @@ describe('Crawler Support', function() {
     restRef = getFreshRepoFromReference(normalRef);
     forceRestClient(false);
 
-    tokenProvider = testAuthTokenProvider(restRef.database.app);
+    // tokenProvider = testAuthTokenProvider(restRef.database.app);
```
Feiyang1: It seems we are patching auth in app after app.database() has been called.
Under the component framework, authProvider is given as a parameter during database construction, so to mock auth, we should register the mock auth before app.database() is called. It also means we need a new app instance for each test, otherwise database and its authProvider is cached.

My question is how is the mocked auth used in tests. My intention is to ensure the tests with component framework cover all the use cases that are covered today.

I'm also confused because tests pass with these lines commented out.

schmidt-sebastian: Hm. I don't understand the interactions here well enough to provide an adequate answer (I can do some research if required). If the tests pass against Prod and against the Emulator without mocking the auth token then I see little reason to do so.

packages/database/src/core/Repo.ts
```diff
@@ -79,7 +81,8 @@ export class Repo {
   constructor(
     public repoInfo_: RepoInfo,
     forceRestClient: boolean,
-    public app: FirebaseApp
+    public app: FirebaseApp,
+    authProvider: Provider<FirebaseAuthInternal>
   ) {
     const authTokenProvider = new AuthTokenProvider(app);
```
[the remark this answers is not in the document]
Feiyang1: Fixed.  Rebasing is dangerous. 😅

packages/database/test/browser/crawler_support.test.ts
Feiyang1: I double checked all tests (browser, node, emulator) pass without mocking the auth token, so I removed these 2 lines and `testAuthTokenProvider`.

packages/database/index.node.ts
```diff
@@ -51,8 +58,26 @@ export function initStandalone(app: FirebaseApp, url: string, version: string) {
   CONSTANTS.NODE_ADMIN = true;
   setSDKVersion(version);
 
+  /**
+   * Create a 'auth-internal' component using firebase-admin-node's implementation that implements FirebaseAuthInternal.
```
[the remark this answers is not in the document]
Feiyang1: Done.

PR #2339, opened 2019-11-06
Directly check for existence of gtag script tag

hsubox76:
We previously checked for pre-existence of a gtag script by looking for `window.dataLayer`. This is not an accurate check in some cases, such as if the user is using Google Tag Manager on the page, which creates a `window.dataLayer` but does not insert a script tag to download `gtag.js`.

This change directly checks the DOM for existence of such a script tag, and we will insert it if not found.

### DIFF
--- a/packages/analytics/src/helpers.test.ts
+++ b/packages/analytics/src/helpers.test.ts
@@ -51,8 +51,8 @@ describe('FirebaseAnalytics methods', () => {
   });
 
   it('getOrCreateDataLayer is able to correctly identify an existing data layer', () => {
-    delete window['dataLayer'];
-    expect(getOrCreateDataLayer('dataLayer')).to.deep.equal([]);
+    const existingDataLayer = window['dataLayer'] = [];
+    expect(getOrCreateDataLayer('dataLayer')).to.equal(existingDataLayer);
   });
 
   it('insertScriptIfNeeded inserts script tag', () => {
--- a/packages/analytics/src/helpers.test.ts
+++ b/packages/analytics/src/helpers.test.ts
@@ -249,27 +249,6 @@ describe('FirebaseAnalytics methods', () => {
       existingGtagStub.reset();
     });
 
-    // it('wrapped window.gtag function waits for initialization promises before sending events', async () => {
-    //   const deferred = new Deferred<void>();
-    //   wrapOrCreateGtag(
-    //     { [mockAnalyticsId]: deferred.promise },
-    //     'dataLayer',
-    //     'gtag'
-    //   );
-    //   (window['gtag'] as Gtag)(GtagCommand.EVENT, 'purchase', {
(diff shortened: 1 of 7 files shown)


packages/analytics/src/helpers.ts
```diff
@@ -80,7 +77,7 @@ export function insertScriptTag(dataLayerName: string): void {
 export function getOrCreateDataLayer(dataLayerName: string): DataLayer {
   // Check for existing dataLayer and create if needed.
   let dataLayer: DataLayer = [];
-  if (hasDataLayer(dataLayerName)) {
+  if (Array.isArray(window[dataLayerName])) {
     dataLayer = window[dataLayerName] as DataLayer;
   } else {
     dataLayer = window[dataLayerName] = [];
```
Feiyang1: Since dataLayer is already initialized with `[]`, you can avoid one memory allocation by `window[dataLayerName] = dataLayer`

packages/analytics/src/helpers.ts
```diff
@@ -206,3 +203,16 @@ export function wrapOrCreateGtag(
     wrappedGtag: window[gtagFunctionName] as Gtag
   };
 }
+
+/**
+ * Returns first script tag in DOM matching our gtag url pattern.
+ */
+export function findGtagScriptOnPage(): HTMLScriptElement | null {
+  const scriptTags = window.document.getElementsByTagName('script');
+  for (const tag of Object.values(scriptTags)) {
```
Feiyang1: You an iterate directly on `scriptTags` - `for (const tag of scriptTags)`

packages/analytics/src/helpers.test.ts
```diff
@@ -45,13 +45,14 @@ describe('FirebaseAnalytics methods', () => {
...
-    window['dataLayer'] = [];
-    expect(hasDataLayer('dataLayer')).to.be.true;
-    window['dataLayer'] = 'hello';
-    expect(hasDataLayer('dataLayer')).to.be.false;
+  it('getOrCreateDataLayer is able to create a new data layer if none exists', () => {
     delete window['dataLayer'];
+    expect(getOrCreateDataLayer('dataLayer')).to.deep.equal([]);
+  });
+
+  it('getOrCreateDataLayer is able to correctly identify an existing data layer', () => {
```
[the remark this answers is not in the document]
hsubox76: Oops, yes, changed.

packages/analytics/src/helpers.ts
hsubox76: Done.

packages/analytics/src/helpers.ts
hsubox76: Confirmed we cannot do this without `downlevelIteration` and adding `dom.iterable` TS lib, which we agreed adds unnecessary bloat to the analytics package.

PR #2341, opened 2019-11-07
Fixed failed auth tests with component framework

Feiyang1:
Some auth tests failed because it uses `firebase.INTERNAL` which has been removed.


Feiyang1:
Merged into https://github.com/firebase/firebase-js-sdk/pull/2342

PR #2346, opened 2019-11-08
Allow nested arrays for IN queries

thebrianchen:
Also making another IN filters integration test run against prod.

### DIFF
--- a/packages/firestore/src/api/user_data_converter.ts
+++ b/packages/firestore/src/api/user_data_converter.ts
@@ -134,7 +134,11 @@ enum UserDataSource {
    * Indicates the source is a where clause, cursor bound, arrayUnion()
    * element, etc. Of note, isWrite(source) will return false.
    */
-  Argument
+  Argument,
+  /**
+   * Indicates that the source is an Argument that allows for nested arrays.
+   */
+  ArrayArgument
 }
 
 function isWrite(dataSource: UserDataSource): boolean {
@@ -144,6 +148,7 @@ function isWrite(dataSource: UserDataSource): boolean {
     case UserDataSource.Update:
       return true;
     case UserDataSource.Argument:
+    case UserDataSource.ArrayArgument:
       return false;
     default:
       throw fail(`Unexpected case for UserDataSource: ${dataSource}`);
@@ -478,10 +483,16 @@ export class UserDataConverter {
   /**
    * Parse a "query value" (e.g. value in a where filter or a value in a cursor
    * bound).
+   *
+   * @param allowArrays Whether the dataContext should allow nested arrays.
    */
-  parseQueryValue(methodName: string, input: unknown): FieldValue {
+  parseQueryValue(
+    methodName: string,
+    input: unknown,
+    allowArrays = false
+  ): FieldValue {
     const context = new ParseContext(
-      UserDataSource.Argument,
+      allowArrays ? UserDataSource.ArrayArgument : UserDataSource.Argument,
       methodName,
       FieldPath.EMPTY_PATH
     );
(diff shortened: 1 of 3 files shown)


packages/firestore/src/api/user_data_converter.ts
```diff
@@ -536,7 +536,9 @@ export class UserDataConverter {
       if (input instanceof Array) {
         // TODO(b/34871131): Include the path containing the array in the error
         // message.
-        if (context.arrayElement) {
+        // Nested arrays are allowed when making IN queries, so we make the
+        // exception here.
+        if (context.arrayElement && context.methodName !== 'Query.where') {
```
mikelehen: This works in JS but is a little fragile (relies on string equality of magic vales) and won't be portable to other platforms since they don't have access to the methodName. [The only reason we have the methodName in JS is because we try to include it in the error message since (historically at least) in some cases you may not have a meaningful callstack to show you what method the error came from.]

I think it'd probably be better to use `context.dataSource` which is a more formalized way to know where the data came from and how we should parse it.  We could probably get away with `context.dataSource !== UserDataSource.Argument` here.  This means we will disable the nested-array validation in more cases, but I think it's probably okay.  If we want to keep it more targeted, we could introduce `UserDataSource.ArrayArgument`, indicating that we're parsing an array and so "nested" arrays is okay.

The other option I had mentioned previously would be to just parse the elements of the array individually for 'in' filters, e.g. [here](https://github.com/firebase/firebase-js-sdk/blob/efc0a78803990eb01888b7b2121c15377c4007b1/packages/firestore/src/api/database.ts#L1488) we would do:
```
        fieldValue = new ArrayValue((value as unknown[]).map(val => this.firestore._dataConverter.parseQueryValue('Query.where', value)));
```

I don't have strong feelings, so whichever of these sounds good to you works for me.

thebrianchen:
I realized that we can't use the `context.dataSource !== UserDataSource.Argument` check since calling `doc.set({ x: FieldValue.arrayUnion(1, ['nested']) })` results in a timeout (didn't dig too deep). 

I tried naming, but I'm not sure if `allowArrays` captures the essence of what we're going for here.

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -478,10 +483,16 @@ export class UserDataConverter {
   /**
    * Parse a "query value" (e.g. value in a where filter or a value in a cursor
    * bound).
+   *
+   * @param allowArrays Whether the dataContext should allow nested arrays.
```
mikelehen: dataContext is an internal implementation detail of this method and so it shouldn't be referenced in the method comments (i.e. this is a ["leaky abstraction"](https://en.wikipedia.org/wiki/Leaky_abstraction)).  I'd change it to something like "Whether the query value is an array that may directly contain additional arrays (e.g. the operand of an 'in' query)."

packages/firestore/src/api/user_data_converter.ts
```diff
@@ -536,7 +547,12 @@ export class UserDataConverter {
       if (input instanceof Array) {
         // TODO(b/34871131): Include the path containing the array in the error
         // message.
-        if (context.arrayElement) {
+        // Nested arrays are allowed when making IN queries, so we make the
+        // exception here.
```
mikelehen: I'd tweak this to:

> In the case of IN queries, the parsed data is an array (representing the set of values to be included for the IN query) that may directly contain additional arrays (each representing an individual field value) so we disable this validation.

packages/firestore/src/api/user_data_converter.ts
thebrianchen: oh woah, I did not know there was an official name for that. Thanks for sharing!

packages/firestore/src/api/user_data_converter.ts
thebrianchen: This is so much clearer, thanks!

PR #2348, opened 2019-11-08
Add Database.Servervalue.increment(x)

inlined:
Per offline discussions, it looks like this is pretty simple to add; we just change resolveDeferredValueSnapshot to also accept a cached value.

Added tests for ServeValues in general.

Forgive my lake of familiarity: do we edit packages/database-types directly? While I understand we don't actually want the feature to be exposed directly yet, it seems weird that extending ServerValues in database.ts didn't change database-types/index.ts

### DIFF
--- a/packages/database/src/core/util/ServerValues.ts
+++ b/packages/database/src/core/util/ServerValues.ts
@@ -23,6 +23,7 @@ import { nodeFromJSON } from '../snap/nodeFromJSON';
 import { PRIORITY_INDEX } from '../snap/indexes/PriorityIndex';
 import { Node } from '../snap/Node';
 import { ChildrenNode } from '../snap/ChildrenNode';
+import { SyncTree } from '../SyncTree';
 
 /**
  * Generate placeholders for deferred values.
@@ -48,14 +49,62 @@ export const generateWithValues = function(
  */
 export const resolveDeferredValue = function(
   value: { [k: string]: any } | string | number | boolean,
+  existing: Node,
   serverValues: { [k: string]: any }
 ): string | number | boolean {
   if (!value || typeof value !== 'object') {
     return value as string | number | boolean;
-  } else {
-    assert('.sv' in value, 'Unexpected leaf node or priority contents');
-    return serverValues[value['.sv']];
   }
+  assert('.sv' in value, 'Unexpected leaf node or priority contents');
+
+  if (typeof value['.sv'] === 'string') {
+    return resolveScalarDeferredValue(value['.sv'], existing, serverValues);
+  } else if (typeof value['.sv'] === 'object') {
+    return resolveComplexDeferredValue(value['.sv'], existing, serverValues);
+  } else {
+    assert(false, 'Unexpected server value: ' + JSON.stringify(value, null, 2));
+  }
+};
+
+const resolveScalarDeferredValue = function(
+  op: string,
+  existing: Node,
+  serverValues: { [k: string]: any }
+): string | number | boolean {
+  switch (op) {
+    case 'timestamp':
+      return serverValues['timestamp'];
+    default:
+      assert(false, 'Unexpected server value: ' + op);
+  }
+};
+
+const resolveComplexDeferredValue = function(
+  op: Object,
+  existing: Node,
+  unused: { [k: string]: any }
+): string | number | boolean {
+  if (!op.hasOwnProperty('increment')) {
+    assert(false, 'Unexpected server value: ' + JSON.stringify(op, null, 2));
+  }
+  const delta = op['increment'];
+  if (typeof delta !== 'number') {
+    assert(false, 'Unexpected increment value: ' + delta);
+  }
+
+  // Incrementing a non-number sets the value to the incremented amount
+  if (!existing.isLeafNode()) {
+    return delta;
+  }
+
+  const leaf = existing as LeafNode;
+  const existingVal = leaf.getValue();
+  if (typeof existingVal !== 'number') {
+    return delta;
+  }
+
+  // No need to do over/underflow arithmetic here because JS only handles floats under the covers
+  return existingVal + delta;
 };
 
 /**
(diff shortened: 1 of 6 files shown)


packages/database/src/core/util/ServerValues.ts
```diff
@@ -48,13 +49,62 @@ export const generateWithValues = function(
...
     return value as string | number | boolean;
+  }
+  assert('.sv' in value, 'Unexpected leaf node or priority contents');
+
+  if (typeof value['.sv'] === 'string') {
+    return resolveScalarDeferredValue(value['.sv'], existing, serverValues);
+  } else if (typeof value['.sv'] === 'object') {
+    return resolveComplexDeferredValue(value['.sv'], existing, serverValues);
+  } else {
+    assert(false, 'Unexpected server value: ' + JSON.stringify(value, null, 2));
```
schmidt-sebastian: Can we just throw an Error here? We don't yet use the asserts from ./util in database.

packages/database/test/servervalues.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    // server version of the timestamp.
+    expect(values.length).to.equal(2);
+    values.forEach(serverTime => {
+      const delta = Math.abs(serverTime - start);
+      expect(delta).to.be.lessThan(1000);
+    });
+  });
+
+  it('handles increments locally', async () => {
+    // Must go offline because the latest emulator may not support this serer op
```
schmidt-sebastian: s/sever/server and move comment to line 53

packages/database/test/servervalues.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        values.push(snap.val() as number);
+      });
+
+      // Because we're offline, we shouldn't await the node operations. This would block the test.
+      node.set(addOne);
+      node.set(5);
+      node.set(addOne);
+
+      node.off('value');
+      expect(values).to.deep.equal([1, 5, 6]);
```
schmidt-sebastian: Using this test setup it seems relatively straightforward to also test type conversions (both from null and other scalar values). Do you mind adding that?

inlined: Done

packages/database/test/servervalues.test.ts
inlined: Done

packages/database/src/core/util/ServerValues.ts
```diff
@@ -67,13 +117,18 @@ export const resolveDeferredValue = function(
...
   serverValues: Object
 ): SparseSnapshotTree {
   const resolvedTree = new SparseSnapshotTree();
   tree.forEachTree(new Path(''), function(path, node) {
     resolvedTree.remember(
       path,
-      resolveDeferredValueSnapshot(node, serverValues)
+      resolveDeferredValueSnapshot(
+        node,
+        existing.calcCompleteEventCache(path),
```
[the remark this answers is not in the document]
inlined: My understanding is that calcCompleteEventCache will return a null node (ChildrenNode.EMPTY_NODE) in this case. I've added a unit (smoke) test with no listeners and an assert against null/undefined.

packages/database/src/core/util/ServerValues.ts
schmidt-sebastian: Oh, sorry, I looked at your PR in an IDE with no diffbase. You can leave it as is.

packages/database/src/core/util/ServerValues.ts
```diff
@@ -67,13 +116,17 @@ export const resolveDeferredValue = function(
  */
 export const resolveDeferredValueTree = function(
   tree: SparseSnapshotTree,
+  syncTree: SyncTree,
   serverValues: Object
 ): SparseSnapshotTree {
   const resolvedTree = new SparseSnapshotTree();
   tree.forEachTree(new Path(''), function(path, node) {
+    const existing = syncTree.calcCompleteEventCache(path);
+    assert(existing != null && typeof existing != 'undefined',
```
schmidt-sebastian: In Firestore we just do `assert(!!existing)`.

If you want to keep it more explicit, please change to `existing !== null`. `existing != null` already implicitly checks for undefined, but should be rejected by the linter (which we may not have set up correctly).

packages/database/src/core/util/ServerValues.ts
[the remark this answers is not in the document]
schmidt-sebastian: Do you mind changing the return type of `calcCompleteEventCache()` to ensure that this remains true?

packages/database/src/core/util/ServerValues.ts
inlined: ok

packages/database/src/core/util/ServerValues.ts
inlined: Used !==

schmidt-sebastian:
> Even though you approved the PR is blocked. Do I need to add a second approver?

@Feiyang1 Can you approve the .vscode change?

inlined:
> > Even though you approved the PR is blocked. Do I need to add a second approver?
> 
> @Feiyang1 Can you approve the .vscode change?

Moved to #2365; it would unblock me and seems like a cleaner git history anyway.

packages/database/src/core/util/ServerValues.ts
```diff
@@ -67,13 +116,19 @@ export const resolveDeferredValue = function(
  */
 export const resolveDeferredValueTree = function(
   tree: SparseSnapshotTree,
+  syncTree: SyncTree,
   serverValues: Object
 ): SparseSnapshotTree {
   const resolvedTree = new SparseSnapshotTree();
   tree.forEachTree(new Path(''), function(path, node) {
+    const existing = syncTree.calcCompleteEventCache(path);
```
schmidt-sebastian: @inlined - Would it be possible to move this call into `resolveComplexDeferredValue`?

Context https://github.com/firebase/firebase-js-sdk/issues/2487

PR #2351, opened 2019-11-12
remove prerender and unloaded from visibilityState

Feiyang1:
`prerender` is deprecated and removed from the spec according to https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

Typescript has removed it from `VisibilityState` type and https://github.com/firebase/firebase-js-sdk/pull/2344 fails because of it.

This PR removes `prerender` from our own code.

### DIFF
--- a/packages/performance/src/utils/attributes_utils.ts
+++ b/packages/performance/src/utils/attributes_utils.ts
@@ -28,8 +28,7 @@ const enum ServiceWorkerStatus {
 export enum VisibilityState {
   UNKNOWN = 0,
   VISIBLE = 1,
-  HIDDEN = 2,
-  UNLOADED = 4
+  HIDDEN = 2
 }
 
 const enum EffectiveConnectionType {
--- a/packages/performance/src/utils/attributes_utils.ts
+++ b/packages/performance/src/utils/attributes_utils.ts
@@ -29,7 +29,6 @@ export enum VisibilityState {
   UNKNOWN = 0,
   VISIBLE = 1,
   HIDDEN = 2,
-  PRERENDER = 3,
   UNLOADED = 4
 }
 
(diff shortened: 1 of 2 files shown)


packages/performance/src/utils/attributes_utils.ts
```diff
@@ -29,7 +29,6 @@ export enum VisibilityState {
   UNKNOWN = 0,
   VISIBLE = 1,
   HIDDEN = 2,
-  PRERENDER = 3,
   UNLOADED = 4
```
Feiyang1: Is `UNLOADED` needed ? It's not part of the spec and we don't use it anywhere.

packages/performance/src/utils/attributes_utils.ts
```diff
@@ -29,7 +29,6 @@ export enum VisibilityState {
   UNKNOWN = 0,
   VISIBLE = 1,
   HIDDEN = 2,
-  PRERENDER = 3,
```
alikn: Can you please add a comment here so it is not weird that the enum jumps from 2 to 4?

alikn: Nvm, we can remove unloaded as well.

packages/performance/src/utils/attributes_utils.ts
alikn: We can remove this one as well.

Feiyang1: Done.

packages/performance/src/utils/attributes_utils.ts
Feiyang1: Removed `UNLOADED`

PR #2357, opened 2019-11-14
Enable Index-Free Queries

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/730

@Feiyang1 / @hsubox76: Can we bump the minor version for the release that this will be part of?

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,4 +1,11 @@
-# Unreleased
+# Unreleased (1.8.0)
+- [changed] Improved the performance of repeatedly executed queries when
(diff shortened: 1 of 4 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-# Unreleased
+# Unreleased (1.8.0)
```
wilhuff: Don't you mean to put this down on a separate line?

schmidt-sebastian: This is my way of suggesting the next version number. It's still unreleased.

PR #2367, opened 2019-11-17
Compile time asserts

schmidt-sebastian:


packages/firestore/src/util/assert.ts
```diff
@@ -43,7 +43,7 @@ export function fail(failure: string): never {
  * Fails if the given assertion condition is false, throwing an Error with the
  * given message if it did.
  */
-export function assert(assertion: boolean, message: string): void {
+export function assert(assertion: boolean, message: string): asserts assertion {
```
Feiyang1: Love it!

packages/firestore/src/remote/serializer.ts
```diff
@@ -595,11 +595,11 @@ export class JsonProtoSerializer {
       !!doc.found,
       'Tried to deserialize a found document from a missing document.'
     );
-    assertPresent(doc.found!.name, 'doc.found.name');
-    assertPresent(doc.found!.updateTime, 'doc.found.updateTime');
-    const key = this.fromName(doc.found!.name!);
-    const version = this.fromVersion(doc.found!.updateTime!);
-    return new Document(key, version, {}, undefined, doc.found!, v =>
+    assertPresent(doc.found.name, 'doc.found.name');
+    assertPresent(doc.found.updateTime, 'doc.found.updateTime');
```
mikelehen: Feel free to punt, but could we magic up assertPresent() to get rid of the remaining `!` usage below?

schmidt-sebastian: Updated. Do you want to take another look?

PR #2378, opened 2019-11-25
Component Framework

Feiyang1:
Adopt the component framework to manage SDK registration and  inter SDK dependencies.
Design doc: go/firebase-js-platform


packages/messaging/src/controllers/window-controller.ts
```diff
@@ -308,16 +308,23 @@ export class WindowController extends BaseController {
...
           const eventType = getEventType(firebaseMessagingType);
-          (this.app as _FirebaseApp).INTERNAL.analytics.logEvent(
-            eventType,
-            /* eslint-disable camelcase */
-            {
-              message_name: data[FN_CAMPAIGN_NAME],
-              message_id: data[FN_CAMPAIGN_ID],
-              message_time: data[FN_CAMPAIGN_TIME],
-              message_device_time: Math.floor(Date.now() / 1000)
+          this.services.analyticsProvider.get().then(
```
mmermerkaya: Can you make the event listener `async` and use `await` here?

packages/messaging/test/testing-utils/make-fake-firebase-services.ts
```diff
@@ -34,12 +48,23 @@ export function makeFakeApp(options: FirebaseOptions = {}): FirebaseApp {
...
-        delete: () => Promise.resolve()
-      };
-    }
+    installations: null as any
+  };
+}
+
+export function makeFakeInstallations(): FirebaseInstallations {
+  return {
+    getId: () => Promise.resolve('FID'),
```
mmermerkaya: Very much a nit, but you can do `getId: async () => 'FID',`. Also the other ones.

packages/component/README.md
```diff
@@ -0,0 +1,20 @@
+# @firebase/component
+
+_NOTE: This is specifically tailored for Firebase JS SDK usage, if you are not a
+member of the Firebase team, please avoid using this package_
+
+## Installation
```
hsubox76: Should we get rid of the rest of the instructions from here on down if nobody should be installing it anyway?

packages/messaging/src/controllers/window-controller.ts
Feiyang1: Done.

packages/messaging/test/testing-utils/make-fake-firebase-services.ts
Feiyang1: Nice! Changed here and other places where we return a promise.

packages/component/README.md
Feiyang1: Good point, but I think it's harmless to keep it, and it's consistent with other packages, e.g. `@firebase/util`.

PR #2382, opened 2019-11-26
LimitToLast: back port some changes from android

wu-hui:
* s/memorized/memoized
* Disallow double-unlisten (remote store) and double-releasing (local store)
* Consolidate duplicated integration tests
* Mark synthesizeTargetChange as non-multitab

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -890,7 +890,10 @@ export class LocalStore {
     keepPersistedTargetData: boolean
   ): Promise<void> {
     const targetData = this.targetDataByTarget.get(targetId);
-    assert(targetData !== null, `Tried to release nonexistent target: ${targetId}`);
+    assert(
+      targetData !== null,
+      `Tried to release nonexistent target: ${targetId}`
+    );
 
     const mode = keepPersistedTargetData
       ? 'readwrite-idempotent'
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -890,9 +890,7 @@ export class LocalStore {
     keepPersistedTargetData: boolean
   ): Promise<void> {
     const targetData = this.targetDataByTarget.get(targetId);
-    if (!targetData) {
-      return Promise.resolve();
-    }
+    assert(targetData !== null, `Tried to release nonexistent target: ${targetId}`);
 
     const mode = keepPersistedTargetData
       ? 'readwrite-idempotent'
@@ -917,7 +915,7 @@ export class LocalStore {
           return PersistencePromise.forEach(removed, (key: DocumentKey) =>
             this.persistence.referenceDelegate.removeReference(txn, key)
           ).next(() => {
-            this.persistence.referenceDelegate.removeTarget(txn, targetData);
+            this.persistence.referenceDelegate.removeTarget(txn, targetData!);
           });
         } else {
           return PersistencePromise.resolve();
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -917,15 +915,15 @@ export class LocalStore {
           return PersistencePromise.forEach(removed, (key: DocumentKey) =>
             this.persistence.referenceDelegate.removeReference(txn, key)
           ).next(() => {
-            this.persistence.referenceDelegate.removeTarget(txn, targetData);
+            this.persistence.referenceDelegate.removeTarget(txn, targetData!);
```
schmidt-sebastian: Can you check if this is still needed? With TypeScript 3.7, `assert` should now be type-aware.

packages/firestore/src/local/local_store.ts
```diff
@@ -917,15 +915,15 @@ export class LocalStore {
...
           });
         } else {
           return PersistencePromise.resolve();
         }
       })
       .then(() => {
         this.targetDataByTarget = this.targetDataByTarget.remove(targetId);
-        this.targetIdByTarget.delete(targetData.target);
+        this.targetIdByTarget.delete(targetData!.target);
```
schmidt-sebastian: Same as above.

packages/firestore/src/remote/remote_event.ts
```diff
@@ -127,11 +127,12 @@ export class TargetChange {
 
   /**
    * HACK: Views require TargetChanges in order to determine whether the view is
-   * CURRENT, but secondary tabs don't receive remote events. So this method is
-   * used to create a synthesized TargetChanges that can be used to apply a
-   * CURRENT status change to a View, for queries executed in a different tab.
+   * CURRENT, but secondary tabs or new queries mapped to an active target don't
+   * receive remote events (yet). So this method is used to create a synthesized
+   * TargetChanges that can be used to apply a CURRENT status change to a View,
+   * for queries executed in a different tab, or for the new queries to raise
```
schmidt-sebastian: How about:

This method is used to create a synthesized TargetChanges that can be used to apply a CURRENT status change to a View (for queries executed in a different tab) or for new queries (to raise snapshots with correct CURRENT status).

packages/firestore/src/remote/remote_event.ts
```diff
@@ -127,11 +127,12 @@ export class TargetChange {
 
   /**
    * HACK: Views require TargetChanges in order to determine whether the view is
-   * CURRENT, but secondary tabs don't receive remote events. So this method is
-   * used to create a synthesized TargetChanges that can be used to apply a
-   * CURRENT status change to a View, for queries executed in a different tab.
+   * CURRENT, but secondary tabs or new queries mapped to an active target don't
```
schmidt-sebastian: Nit: Would be easier to parse if you made it "to an already active target".

packages/firestore/src/local/local_store.ts
wu-hui: Leads to compilation error.

packages/firestore/src/local/local_store.ts
wu-hui: Same.

packages/firestore/src/remote/remote_event.ts
wu-hui: Done.

packages/firestore/src/remote/remote_event.ts
wu-hui: Took your comment below.

Issue #2383, opened 2019-11-27
collectionData in RxFire with React Native Firebase does not provide any values 


dtengeri:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Mojave 10.14.5
  * Browser version: -
  * Firebase SDK version: RxFire 3.9.2
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

We tried to use RxFire with [React Native Firebase](https://github.com/invertase/react-native-firebase) but the `collectionData()` does not provide any values.

#### Steps to reproduce:

Created a sample application which is available here:
https://github.com/dtengeri/rxfiresample

#### Relevant Code:

The following code does not provide any values:
```
collectionData(citiesRef, "id")
  .pipe(tap(() => console.log("This is just an observable!")))
  .subscribe(cities => {
    this.setState({ cities });
  });
```

rommelpe:
Thanks for reporting @dtengeri. I was able to get the same behavior in android, and I'll file this internally to see what's causing the issue. I'll let you know once I have an update.

joshdlcruz:
Any update on this issue?

jamesdaniels:
https://github.com/FirebaseExtended/rxfire/pull/9

PR #2384, opened 2019-11-27
Modify onSnapshot arguments

dtengeri:
* Issue: https://github.com/firebase/firebase-js-sdk/issues/2383
  * There are two issues which are resolved by the PR:
    1. React Native Firebase could not handle empty options in `onSnapshot()`
    2. The `next`, `error` and `complete` callbacks lost the context of `this` and cause error when RNFirebase calls `next()`

### Testing

  * RxFire tests passed


dtengeri:
@googlebot I signed it!

PR #2385, opened 2019-11-27
Add typings for WebChannel

schmidt-sebastian:
With inspiration from

https://github.com/fivetran/typescript-closure-tools/blob/master/index/closure-library/closure/goog/net/errorcode.d.ts
and  https://github.com/fivetran/typescript-closure-tools/blob/master/index/closure-library/closure/goog/net/xhrio.d.ts

### DIFF
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -313,9 +313,9 @@ export class WebChannelConnection implements Connection {
     ): void => {
       // TODO(dimond): closure typing seems broken because WebChannel does
       // not implement goog.events.Listenable
-      channel.listen(type, (param?: T) => {
+      channel.listen(type, (param: unknown) => {
         try {
-          fn(param);
+          fn(param as T);
         } catch (e) {
           setTimeout(() => {
             throw e;
@@ -369,9 +369,9 @@ export class WebChannelConnection implements Connection {
           // compatible with the bug we need to check either condition. The latter
           // can be removed once the fix has been rolled out.
           // Use any because msgData.error is not typed.
-          const msgDataAsAny: WebChannelError | object = msgData;
+          const msgDataOrError: WebChannelError | object = msgData;
           const error =
-            msgDataAsAny.error || (msgDataAsAny as WebChannelError[])[0]?.error;
+            msgDataOrError.error || (msgDataOrError as WebChannelError[])[0]?.error;
           if (error) {
             log.debug(LOG_TAG, 'WebChannel received error:', error);
             // error.status will be a string like 'OK' or 'NOT_FOUND'.
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -371,7 +371,8 @@ export class WebChannelConnection implements Connection {
           // Use any because msgData.error is not typed.
           const msgDataOrError: WebChannelError | object = msgData;
           const error =
-            msgDataOrError.error || (msgDataOrError as WebChannelError[])[0]?.error;
+            msgDataOrError.error ||
+            (msgDataOrError as WebChannelError[])[0]?.error;
           if (error) {
             log.debug(LOG_TAG, 'WebChannel received error:', error);
             // error.status will be a string like 'OK' or 'NOT_FOUND'.
--- a/packages/firestore/src/platform_browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform_browser/webchannel_connection.ts
@@ -20,6 +20,7 @@ import {
   ErrorCode,
   EventType,
   WebChannel,
+  WebChannelError,
   WebChannelOptions,
   XhrIo
 } from '@firebase/webchannel-wrapper';
@@ -123,7 +124,7 @@ export class WebChannelConnection implements Connection {
                 xhr.getResponseText()
               );
               if (status > 0) {
-                const responseError = xhr.getResponseJson().error;
+                const responseError = (xhr.getResponseJson() as WebChannelError).error;
                 if (
                   !!responseError &&
                   !!responseError.status &&
@@ -189,7 +190,6 @@ export class WebChannelConnection implements Connection {
 
       this.modifyHeadersForRequest(headers, token);
 
-      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       xhr.send(url, 'POST', requestString, headers, XHR_TIMEOUT_SECS);
     });
   }
@@ -271,9 +271,7 @@ export class WebChannelConnection implements Connection {
 
     const url = urlParts.join('');
     log.debug(LOG_TAG, 'Creating WebChannel: ' + url + ' ' + request);
-    // Use any because listen isn't defined on it.
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const channel = webchannelTransport.createWebChannel(url, request) as any;
+    const channel = webchannelTransport.createWebChannel(url, request);
 
     // WebChannel supports sending the first message with the handshake - saving
     // a network round trip. However, it will have to call send in the same
@@ -309,7 +307,7 @@ export class WebChannelConnection implements Connection {
     // Note that eventually this function could go away if we are confident
     // enough the code is exception free.
     const unguardedEventListen = <T>(
-      type: WebChannel.EventType,
+      type: string,
       fn: (param?: T) => void
     ): void => {
       // TODO(dimond): closure typing seems broken because WebChannel does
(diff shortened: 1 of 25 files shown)


packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -31,7 +82,9 @@ export interface WebChannelOptions {
   messageHeaders?: StringMap;
   initMessageHeaders?: StringMap;
```
mikelehen: Are these usages of StringMap safe for some reason?  I'm not clear why you had to change `messageUrlParams` and `internalChannelParams` but not these.  

That said, it would be nice if you didn't have to change any of these and our minification logic was able to handle `StringMap`.  Is that not possible?

packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -44,13 +97,23 @@ export interface WebChannelOptions {
...
   xmlHttpFactory?: unknown;
   requestRefreshThresholds?: { [key: string]: number };
 }
 
+export interface WebChannel {
+  open(): void;
+  close(): void;
+  halfClose(): void;
+  listen<T>(type: EventType, cb: (param: T) => void): void;
+  send<T>(msg: T): void;
```
mikelehen: nit: Is there a benefit to making these generic instead of typing them as `unknown` or similar?

schmidt-sebastian: `Send` can be unknown (it should probably not be - but then we should both the generic type on WebChannel). I made it unknown for now.
The <T> in `listen` calls back to us so it cannot be unknown.

packages/webchannel-wrapper/src/index.js
```diff
@@ -68,8 +68,6 @@ goog.net.XhrIo.prototype['getLastError'] =
 goog.net.XhrIo.prototype['getLastErrorCode'] =
   goog.net.XhrIo.prototype.getLastErrorCode;
 goog.net.XhrIo.prototype['getStatus'] = goog.net.XhrIo.prototype.getStatus;
-goog.net.XhrIo.prototype['getStatusText'] =
-  goog.net.XhrIo.prototype.getStatusText;
```
schmidt-sebastian: This is not used by our code. Removing it is technically a breaking change though for consumers of this package. We might be ok though: https://github.com/search?l=JavaScript&q=%22%40firebase%2Fwebchannel-wrapper%22&type=Code

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -371,10 +369,9 @@ export class WebChannelConnection implements Connection {
           // compatible with the bug we need to check either condition. The latter
           // can be removed once the fix has been rolled out.
           // Use any because msgData.error is not typed.
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          const msgDataAsAny: any = msgData;
+          const msgDataAsAny: WebChannelError | object = msgData;
```
mikelehen: The name `msgDataAsAny` did not survive your refactor very well.  `msgDataOrError` ?

packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -15,23 +15,65 @@
...
+}
+
+export var ErrorCode: {
+  NO_ERROR: number;
+  HTTP_ERROR: number;
+  TIMEOUT: number;
+};
+
+export interface Headers {
+  [name: string]: string | number;
```
mikelehen: Out of curiosity, why doesn't this cause problems like StringMap did?

packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -15,23 +15,65 @@
...
+  getResponseJson(): WebChannelError | object;
+
+  listenOnce<T>(type: string, cb: (param: T) => void): void;
+}
 
 export interface WebChannelOptions {
-  messageHeaders?: StringMap;
-  initMessageHeaders?: StringMap;
+  messageHeaders?: {};
+  initMessageHeaders?: {};
```
mikelehen: Is your intention to require that we set these to an empty object?  Because I don't think that's what this actually does.  This will allow us to pass any object (e.g. `{'foo': 'bar'}` since TypeScript does structural typing (aka duck typing).  

Not sure it's possible to specify a type that indicates that only an empty object is valid.

That said, in general this is feeling somewhat fragile which is a bit worrisome since webchannel is already a bit fragile. It wouldn't be too surprising if we tweak some webchannel setting for performance/reliability reasons and then it gets minified wrong because we don't properly update this d.ts file and therefore it's ignored, and we don't notice because it's not something we can readily test.

If that's the world we're heading towards, I'd like to discuss if there are ways to avoid it...  and if not, then we will perhaps eventually (once your minification stuff lands) want some big warnings in this file or something.

packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -44,13 +86,22 @@ export interface WebChannelOptions {
...
+    forwardChannelRequestTimeoutMs?: number;
+  };
   xmlHttpFactory?: unknown;
   requestRefreshThresholds?: { [key: string]: number };
 }
 
+export interface WebChannel {
+  open(): void;
+  close(): void;
+  listen<T>(type: string, cb: (param: T) => void): void;
```
mikelehen: I still don't really understand this usage of generics.  Is there even any practical difference between:
```
function foo(x: any)
// and
function foo<T>(x: T)
```
It seems like it's a way of telling TypeScript to let you pass anything.  The generic way looks more type-safe, but I don't think it is?

It seems like the correct type here is `unknown` and we should explicitly cast to something else if we need to.

packages/webchannel-wrapper/src/index.d.ts
```diff
@@ -15,23 +15,65 @@
...
+
+  getLastError(): string;
+
+  getStatus(): number;
+
+  getResponseText(): string;
+
+  getResponseJson(): WebChannelError | object;
+
+  listenOnce<T>(type: string, cb: (param: T) => void): void;
```
mikelehen: same feedback about generics here.

packages/firestore/src/platform_browser/webchannel_connection.ts
```diff
@@ -310,7 +308,7 @@ export class WebChannelConnection implements Connection {
     // Note that eventually this function could go away if we are confident
     // enough the code is exception free.
     const unguardedEventListen = <T>(
-      type: WebChannel.EventType,
+      type: string,
```
mikelehen: Was it not possible to make WebChannel.EventType work?  It seems like we're losing a bit of help from the type system here by switching to just `string`.  Not the end of the world though.

packages/firestore/src/platform_browser/webchannel_connection.ts
schmidt-sebastian: Sounds good. Renamed.

packages/firestore/src/platform_browser/webchannel_connection.ts
schmidt-sebastian: WebChannel.EventType is now a variable declaration and no longer a string. I had tried both `WebChannel.EventType` and `typepof WebChannel.EventType`.

packages/webchannel-wrapper/src/index.d.ts
schmidt-sebastian: I opted not to enable renaming for Object literal access. Since all headers are invalid JavaScript identifiers, they are only ever set using the literal access notation (such as `header['X-Google-Foo'] = 42`).

packages/webchannel-wrapper/src/index.d.ts
schmidt-sebastian: Removed. I am "casting" in the callsite now.

packages/webchannel-wrapper/src/index.d.ts
schmidt-sebastian: To be perfectly honest, that was my intention. I did realize this doesn't work but I thought that it might still convey some of the original intent. 

But you can ignore all this because I came up with a way that actually works. I can disallow all properties with:

`messageHeaders: { [k : string] : never }`

Furthermore, all members that have at least a single property specified do not allow any extra properties as is.

packages/webchannel-wrapper/src/index.d.ts
schmidt-sebastian: No - there is no difference, but one looks safer! I changed this to `unknown` and modified the consumer.

packages/firestore/src/platform_browser/webchannel_connection.ts
mikelehen: Bleh...  I came up with several ways to get this to compile (let us use `EventType` as a type as well as hang property values off it).  E.g.:

```
export interface EventType {}
export var EventType: {
  readonly COMPLETE: EventType
}
```
or:
```
export class EventType {
  static readonly COMPLETE: EventType;
}
```

but they didn't provide type-safety, since TypeScript's structural typing would let you pass anything (e.g. `{x: 1}`) for a parameter typed as `EventType`. 😕 

So the only thing I found that actually works for type-safety is using `export enum EventType { COMPLETE }` like you originally had.  So if you feel like putting that back, I would be okay with that.  Else if you're tired of messing with it, this is okay too.

schmidt-sebastian: Hm. My "I couldn't get it to work" response above is based on the current version that used the variable declaration w/ nested property syntax. I am inclined to leave that as is for now, as it matches the underlying library closer than the Enum values.

PR #2386, opened 2019-11-27
Don't use "long" in Proto types

schmidt-sebastian:
"long" is not defined and throws a diagnostic error:

Error: TypeScript Diagnostic Error : ../../firebase/firebase-js-sdk/packages/firestore/src/protos/firestore_proto_api.d.ts(366,42): error TS2304: Cannot find name 'long'.

### DIFF
--- a/packages/firestore/src/protos/firestore_proto_api.d.ts
+++ b/packages/firestore/src/protos/firestore_proto_api.d.ts
@@ -363,7 +363,7 @@ export declare namespace firestoreV1ApiClientInterfaces {
     booleanValue?: boolean;
     integerValue?: string;
     doubleValue?: number;
-    timestampValue?: string | { seconds: long; nanos: long };
+    timestampValue?: string | { seconds: string; nanos: number };
     stringValue?: string;
     bytesValue?: string | Uint8Array;
     referenceValue?: string;


packages/firestore/src/protos/firestore_proto_api.d.ts
```diff
@@ -363,7 +363,7 @@ export declare namespace firestoreV1ApiClientInterfaces {
     booleanValue?: boolean;
     integerValue?: string;
     doubleValue?: number;
-    timestampValue?: string | { seconds: long; nanos: long };
+    timestampValue?: string | { seconds: string; nanos: number };
```
mikelehen: Hrm.  I wonder if this should technically be `string|number` ?  I don't think it's required to send `seconds` as a string, but we do to avoid number overflows.

I don't care strongly though.  If it compiles, then it's good enough for now.

schmidt-sebastian: I think you are right, but that breaks our internal type:

```
// This is a supplement to the generated proto interfaces, which fail to account
// for the fact that a timestamp may be encoded as either a string OR this.
interface TimestampProto {
  seconds?: string;
  nanos?: number;
}
```

That's why I just use strings.

PR #2396, opened 2019-12-02
Update auth.md

CreativeBuilds:
The return value of the idToken() had a breaking style change for how markdown works as the "|" symbol cut off the rest of the line as Markdown thinks that | was closing the table.


jamesdaniels:
@CreativeBuilds thanks for the catch, I actually went with escaping the pipe char instead.

PR #2401, opened 2019-12-04
Add Closure Externs for Firebase Performance

richieforeman:
Adding some minimal closure externs for the Firebase Performance SDK

### DIFF
--- a/packages/firebase/externs/firebase-performance-externs.js
+++ b/packages/firebase/externs/firebase-performance-externs.js
@@ -1,0 +1,47 @@
+/**
+ * @license Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


Feiyang1:
Would it be possible to define the type of `Trace` as well?

packages/firebase/externs/firebase-performance-externs.js
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Firebase Performance SDK externs
+ */
+
+/**
+ * @param {!firebase.app.App=} app The app to create a fireperf service for.  If
```
alikn: Performance sdk only accepts the default app. Throws an exception otherwise.

PR #2402, opened 2019-12-05
Tighten the typing in the Firebase Performance package

richieforeman:
- Remove usage of 'any' for accessing navigator.connection -- prefer a defined interface.
- Make fields in the `ApiService` readonly where possible
- Remove the assertion operator on `ApiService.localStorage` and truly declare it as optional; Update downstream usage to reflect this.


richieforeman:
> Thanks Richie! Welcome to the other side!

Yeehaw! :cowboy_hat_face:

PR #2403, opened 2019-12-05
Small assorted fixes for better property name mangling support

schmidt-sebastian:
I am sure there are more to come... but most of the explicit return type assignments that I had earlier might not be needed anymore (https://github.com/FirebasePrivate/sdk-minifier/pull/2)

### DIFF
--- a/packages/firestore/src/local/indexeddb_index_manager.ts
+++ b/packages/firestore/src/local/indexeddb_index_manager.ts
@@ -62,7 +62,7 @@ export class IndexedDbIndexManager implements IndexManager {
         this.collectionParentsCache.add(collectionPath);
       });
 
-      const collectionParent : DbCollectionParent = {
+      const collectionParent: DbCollectionParent = {
         collectionId,
         parent: encode(parentPath)
       };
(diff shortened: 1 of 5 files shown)


packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -62,10 +62,11 @@ export class IndexedDbIndexManager implements IndexManager {
         this.collectionParentsCache.add(collectionPath);
       });
 
-      return collectionParentsStore(transaction).put({
+      const collectionParent: DbCollectionParent = {
         collectionId,
         parent: encode(parentPath)
-      });
+      };
+      return collectionParentsStore(transaction).put(collectionParent);
```
mikelehen: FWIW this seems like another case we could detect.  Since collectionParent gets passed to a function that accepts `DbCollectionParent`...

schmidt-sebastian: That's definitely true. I haven't seen any other cases like this yet, but I can spend some time on this if there are more (or I run out of other quirks).

PR #2405, opened 2019-12-05
Add version reporting for all packages

hsubox76:
Add platform version reporting to:
- auth
- database
- firestore
- installations
- messaging
- performance
- remote-config
- storage

Add suffixes for bundle variants where appropriate ("-node", "-rn", etc.)


Feiyang1:
Looks good generally. For SDKs that have different builds for browser/node/rn, we should give non-browser ones unique names (with a postfix, e.g. *-node) when registering their versions, so we can tell which build is used. WDYT?

PR #2421, opened 2019-12-09
Request notification permission in getToken if permission is "default"

mmermerkaya:
Instead of returning null and expecting the user to request the notification permission, do it automatically.

Also some (not enough) refactoring of tests.

### DIFF
--- a/packages/messaging/src/controllers/base-controller.ts
+++ b/packages/messaging/src/controllers/base-controller.ts
@@ -70,14 +70,16 @@ export abstract class BaseController
     this.tokenDetailsModel = new TokenDetailsModel(services);
   }
 
-  async getToken(): Promise<string | null> {
-    // Check with permissions
-    const currentPermission = this.getNotificationPermission_();
-    if (currentPermission === 'denied') {
+  async getToken(): Promise<string> {
+    // Check notification permission.
+    let permission = this.getNotificationPermission();
+    if (permission === 'default') {
+      // The user hasn't allowed or denied notifications yet. Ask them.
+      permission = await this.requestNotificationPermission();
+    }
+
+    if (permission !== 'granted') {
       throw errorFactory.create(ErrorCode.NOTIFICATIONS_BLOCKED);
-    } else if (currentPermission !== 'granted') {
-      // We must wait for permission to be granted
-      return null;
     }
 
     const swReg = await this.getSWRegistration_();
@@ -327,8 +329,23 @@ export abstract class BaseController
   /**
    * Returns the current Notification Permission state.
    */
-  getNotificationPermission_(): NotificationPermission {
+  private getNotificationPermission(): NotificationPermission {
     return Notification.permission;
+  }
+
+  /**
+   * Requests notification permission from the user.
+   */
+  private async requestNotificationPermission(): Promise<
+    NotificationPermission
+  > {
+    if (!Notification.requestPermission) {
+      // Notification.requestPermission() is not available in web workers.
+      // Return the current permission.
+      return Notification.permission;
+    }
+
+    return Notification.requestPermission();
   }
 
   getTokenDetailsModel(): TokenDetailsModel {
--- a/packages/messaging/src/controllers/base-controller.ts
+++ b/packages/messaging/src/controllers/base-controller.ts
@@ -340,7 +340,7 @@ export abstract class BaseController
     NotificationPermission
   > {
     if (!Notification.requestPermission) {
-      // Notification.requestPermission() is not available in web workers.
+      // Notification.requestPermission() is not available in service workers.
       // Return the current permission.
       return Notification.permission;
     }
--- a/packages/messaging/test/controller-interface.test.ts
+++ b/packages/messaging/test/controller-interface.test.ts
@@ -216,11 +216,11 @@ describe('Firebase Messaging > *BaseController', () => {
     });
   });
 
-  describe('getNotificationPermission_', () => {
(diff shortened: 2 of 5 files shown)


packages/messaging/src/controllers/base-controller.ts
```diff
@@ -327,10 +329,25 @@ export abstract class BaseController
...
   }
 
+  /**
+   * Requests notification permission from the user.
+   */
+  private async requestNotificationPermission(): Promise<
+    NotificationPermission
+  > {
+    if (!Notification.requestPermission) {
+      // Notification.requestPermission() is not available in web workers.
```
dwoffinden: > https://developer.mozilla.org/en-US/docs/Web/API/Notification/requestPermission
> Note: This feature is available in Web Workers.

ಠ_ಠ

packages/messaging/test/controller-interface.test.ts
```diff
@@ -216,11 +216,11 @@ describe('Firebase Messaging > *BaseController', () => {
...
   });
 
-  describe('getNotificationPermission_', () => {
+  describe('getNotificationPermission', () => {
     it('should return current permission', () => {
-      stub(Notification as any, 'permission').value('test');
+      stub(Notification, 'permission').value('test');
       const controller = new MockBaseController(firebaseInternalServices);
-      const result = controller.getNotificationPermission_();
+      const result = (controller as any).getNotificationPermission();
```
dwoffinden: /sigh

packages/messaging/src/controllers/base-controller.ts
```diff
@@ -70,14 +70,16 @@ export abstract class BaseController
     this.tokenDetailsModel = new TokenDetailsModel(services);
   }
 
-  async getToken(): Promise<string | null> {
```
dwoffinden: Do we have release notes where we can tell callers they can remove their null checks?

packages/messaging/src/controllers/base-controller.ts
mmermerkaya: Haha, fixed.

packages/messaging/test/controller-interface.test.ts
mmermerkaya: Yeah I know...

packages/messaging/src/controllers/base-controller.ts
mmermerkaya: Yup.

packages/messaging/src/controllers/base-controller.ts
```diff
@@ -70,14 +70,16 @@ export abstract class BaseController
...
-    if (currentPermission === 'denied') {
+  async getToken(): Promise<string> {
+    // Check notification permission.
+    let permission = this.getNotificationPermission();
+    if (permission === 'default') {
+      // The user hasn't allowed or denied notifications yet. Ask them.
+      permission = await this.requestNotificationPermission();
+    }
+
+    if (permission !== 'granted') {
```
Feiyang1: Wouldn't it be a breaking change for users who have special handling for `null`? Perhaps they want to show a different message than when user clicks block, although I don't know how often people do it.

Perhaps a safer option is keep returning `null` for `default`?

mmermerkaya: I thought about that too, but I'm pretty sure that like 95% of our users just copy paste what we do in quickstart or our guides, which is requesting permission and calling `getToken` if that succeeds, which would prevent this change from ever affecting them. And if people were calling `getToken` without making sure they have permission, their program wouldn't have worked anyway.

Besides, previously `default` meant that the developer probably forgot to ask for permission, but since `getToken` will do that automatically, `default` here means "we displayed the permission dialog to the user and they dismissed it without clicking allow or deny", which is much closer to the throw behavior before (denying explicitly).

So if a developer were handling `null` by requesting permission, if we still return `null` when the user dismisses the popup, that would make the developer request permission for a second time, which doesn't seem like a good idea to me.

Feiyang1: That makes sense , but wouldn't this change make us request permission for a second time if the developer followed the quickstart where they call `Notification.requestPermission()` before calling `getToken()`, and the user dismisses the dialog?

I don't know how I feel about releasing it without a major version bump. FCM is one of the most popular SDKs, so it could affect a lot of people.  What is the motivation for this change?

Feiyang1: Discussed offline, this change would not affect apps that followed the quickstart, because they won't call `getToken()` if user dismissed the dialog in the first place.

VagnerGon: Actually, this change affects functions.httpsCallable. Since it aquires FCM token automatically, when calling some function by httpsCallable, the browser asks user for notification permission.
So, there are cases when the user does not use FCM but uses Clould Functions and get a dialog asking to allow notifications, unecessarlly

ChromeQ: I've just ran into this issue pointed out by @VagnerGon, I had an httpsCallable which doesn't require a user to be logged in, but on executing the function it auto pops up the notification permission on page load which is really not best practice. 
This is a breaking change and I hope it can be revisited to not call getToken for httpsCallable, perhaps pass an option to httpsCallable so it returns null if permission is default

mmermerkaya: Yeah I didn't realize Functions was doing that, sorry about that. This was fixed in #2489.

Issue #2364, opened 2019-11-15
firebase.messaging.getToken() edgecase when push notification is set back to default (ask)


thomasmburke:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Mojave Version 10.14.6
  * Browser version: Google Chrome Version 78.0.3904.97
  * Firebase SDK version: 7.0.0
  * Firebase Product: Firebase Cloud Messaging

### [REQUIRED] Describe the problem
The issue I have identified was introduced in version 7.0.0 of the firebase-js-sdk. I do not run into this issue in versions previous to 7.0.0. The issue occurs when a use grants notification permissions, gets a FCM token, then updates the notification permissions in the browser back to Ask (default), reloads the page, grants notification permissions again in Chrome, and get an error when trying to getToken at this point. Note: subsequent calls to getToken don't face this issue.

#### Steps to reproduce:

The problem occurs after taking the following steps:
- Call Notification.requestPermission() and user grants permissions
- Call firebase.messaging.getToken()
- Set notification permission in browser back to Ask (default)
![image](https://user-images.githubusercontent.com/40719837/68976050-c8536f00-07a9-11ea-926d-d81af6eea498.png)

- Call Notification.requestPermission() and user grants permissions
- call firebase.messaging.getToken()
- get DELETE  https://fcmregistrations.googleapis.com/v1/projects/{project_id}/regis.../{FCM_token} 500 error and this specific error from FCM: "FirebaseError: Messaging: A problem occured while unsubscribing the user from FCM: FirebaseError: Messaging: A problem occured while unsubscribing the user from FCM: Internal error encountered. (messaging/token-unsubscribe-failed). (messaging/token-unsubscribe-failed)."
- call messaging.getToken() again
- get new token, no error this time.

Others have encountered this issue as well on Google Groups:
https://groups.google.com/a/google.com/forum/#!msg/firebase-developer-cuf/dGU1SnJ1cz0/GMa_WvWGBwAJ

#### Relevant Code:

The firebase javascript messaging quickstart encounters this issue:
https://github.com/firebase/quickstart-js/tree/master/messaging

Specifically, I think the update needs to be made here:
firebase-js-sdk/packages/messaging/src/controllers/base-controller.ts
Lines: 136-137

bjgrosse:
I also ran into this error just as the OP described.

mindiii006:
I am also getting this issue.

rommelpe:
[...]

PR #2428, opened 2019-12-10
Suppress server errors in deleteToken request

mmermerkaya:
Band-aid for #2364 until we find and fix the problem.

### DIFF
--- a/packages/messaging/src/controllers/base-controller.ts
+++ b/packages/messaging/src/controllers/base-controller.ts
@@ -238,7 +238,12 @@ export abstract class BaseController
    */
   private async deleteTokenFromDB(token: string): Promise<void> {
     const tokenDetails = await this.tokenDetailsModel.deleteToken(token);
-    await this.subscriptionManager.deleteToken(this.services, tokenDetails);
+    try {
+      await this.subscriptionManager.deleteToken(this.services, tokenDetails);
+    } catch (e) {
+      // A failed server-side delete does not need to break the app.
+      console.error(e);
+    }
   }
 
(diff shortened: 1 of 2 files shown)


packages/messaging/src/controllers/base-controller.ts
```diff
@@ -238,7 +238,12 @@ export abstract class BaseController
    */
   private async deleteTokenFromDB(token: string): Promise<void> {
     const tokenDetails = await this.tokenDetailsModel.deleteToken(token);
-    await this.subscriptionManager.deleteToken(this.services, tokenDetails);
+    try {
+      await this.subscriptionManager.deleteToken(this.services, tokenDetails);
+    } catch (e) {
```
dwoffinden: Is it possible to check the response code for this particular error, rather than catching all errors?

packages/messaging/src/controllers/base-controller.ts
```diff
@@ -238,7 +238,12 @@ export abstract class BaseController
...
+      // A failed server-side delete does not need to break the app.
```
dwoffinden: 💩

packages/messaging/src/controllers/base-controller.ts
mmermerkaya: The response is just "500 Internal Error" so it's not really distinguishable from other errors :man_shrugging:

packages/messaging/src/controllers/base-controller.ts
mmermerkaya: :roll_eyes:

PR #2434, opened 2019-12-11
Client logging API: methods for users to access the SDK's log messages

hsubox76:
Adds
- `firebase.setLogLevel()` - set threshhold log level to show
- `firebase.onLog()` - add a user specified callback that has access to log level, message, and name of package (e.g. app, firestore, auth) that logged it


packages/logger/src/logger.ts
```diff
@@ -137,23 +147,82 @@ export class Logger {
...
+    }
+    if (logCallback === null) {
+      instance.userLogHandler = null;
+    } else {
+      instance.userLogHandler = (
+        instance: Logger,
+        level: LogLevel,
+        ...args: unknown[]
+      ) => {
+        const message = args.map(arg => (arg as object).toString()).join(' ');
```
Feiyang1: Looked at how we use logger internally and thought about it again and , I think we should just assign `args[0]` to `message` and pass the rest args as is.

`toString()` produces `object [Object]` for objects, which is not very helpful. WDYT?

packages/logger/src/logger.ts
```diff
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+import { LogCallback, LogLevelString, LogOptions } from '@firebase/app-types';
```
Feiyang1: It feels like these 3 types belong in the logger package and should be imported to `@firebase/app`

packages/logger/src/logger.ts
```diff
@@ -137,23 +147,82 @@ export class Logger {
...
+    inst.logLevel = newLevel;
+  });
+}
+
+export function setUserLogHandler(
+  logCallback: LogCallback | null,
+  options: LogOptions
+): void {
+  if (typeof logCallback !== 'function') {
+    console.warn('First argument to `onLog` must be a function.');
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/logger/src/logger.ts
hsubox76: Done.

packages/logger/src/logger.ts
hsubox76: We would have to enforce the convention of putting the entire message string in the first arg which I guess is possible, there are only a few existing cases where it wouldn't work, such as https://github.com/firebase/firebase-js-sdk/blob/542240d093ad54f3c8e5739dcb8449e8b9cab131/packages/firestore/src/local/indexeddb_persistence.ts#L1018.  In that case the message should be the first 2 args.  It might be easier to change the few existing lines in Firestore that don't fit, and enforce the convention going forward, than to handle all cases.

If we decided to do a robust handling of all args, I was thinking we could check arg type and do `toString()` on primitives, and `JSON.stringify()` on arrays and objects.

hsubox76:
I am making a PR for the perf logger conflict issue: https://github.com/firebase/firebase-js-sdk/pull/2460.

packages/firebase/index.d.ts
```diff
@@ -78,6 +78,15 @@ declare namespace firebase {
     complete: CompleteFn;
   }
 
+  /**
+   * The JS SDK supports 5 log levels and also allows a user the ability to
+   * silence the logs altogether.
+   *
+   * The order is as follows:
+   * debug < verbose < info < warn < error
```
egilmorez: Should  "< silent" be rightmost here?

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
     variant?: string
   ): void;
 
+  /**
+   * Sets log level for all Firebase packages.
+   *
+   * All of the log types above the current log level will be captured (i.e. if
```
egilmorez: Suggest "are captured"

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
...
+   * you set the log level to `info`, errors will still be logged, but `debug` and
```
egilmorez: Suggest "are logged"

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
...
+   * `verbose` logs will not)
```
egilmorez: Suggest "are not"

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
...
+   */
+  function setLogLevel(logLevel: LogLevel): void;
+
+  /**
+   * Sets log handler for all Firebase packages.
+   * @param logCallback An optional custom log handler that will execute user code whenever
```
egilmorez: Suggest "executes"

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
...
+      args: unknown[];
+      /**
+       * A string indicating the source of the log call, usually a package name
+       * such as `@firebase/firestore`.
+       */
+      type: string;
+    }) => void,
+    options?: {
+      /**
+       * Threshhold log level. Only logs at or above this level will trigger the `logCallback`
```
egilmorez: Suggest omitting "will"

packages/app/test/clientLogger.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        new Component(
+          'test-shell',
+          async () => {
+            const logger = new Logger('@firebase/logger-test');
+            (firebase as _FirebaseNamespace).onLog((logData) => {
+              result = logData;
+            });
+            logger.info('hi');
+            expect(result.level).to.equal('info');
+            expect(result.message).to.equal('hi');
```
Feiyang1: Can you please add tests for `args` and `type` ?

packages/firebase/index.d.ts
```diff
@@ -95,6 +104,49 @@ declare namespace firebase {
...
+      /**
+       * Any text from logged arguments joined into one string.
+       */
+      message: string;
+      /**
+       * The raw arguments passed to the log call.
+       */
+      args: unknown[];
+      /**
+       * A string indicating the source of the log call, usually a package name
```
Feiyang1: Is it not always a package name? What are the exceptions ?

packages/logger/src/logger.ts
```diff
@@ -137,23 +166,102 @@ export class Logger {
...
+  });
+}
+
+export function setUserLogHandler(
+  logCallback: LogCallback | null,
+  options?: LogOptions
+): void {
+  if (logCallback !== null && typeof logCallback !== 'function') {
+    throw new TypeError(
+      'First argument to `onLog` must be null or a function.'
```
Feiyang1: This function should not know about `onLog` that's defined on firebase namespace. Instead, we should catch this error in `@firebase/app` and wrap it in a `FirebaseError` with the correct message.

packages/logger/src/logger.ts
```diff
@@ -137,23 +166,102 @@ export class Logger {
...
+              try {
+                return JSON.stringify(arg);
+              } catch (ignored) {
+                return null;
+              }
+            }
+          })
+          .filter(arg => arg)
+          .join(' ');
+        if (level >= threshhold) {
```
Feiyang1: `threshhold` should be the loglevel that's passed in `options`. If `options` doesn't have it, it should be the current `instance.logLevel`. 

on line [224](https://github.com/firebase/firebase-js-sdk/pull/2434/files#diff-6d4e0e31c88a6d05d9b0a1fc0c63338fR224), `threshhold` should be assigned with `null` instead of the `instance.logLevel` at that time.

packages/logger/src/logger.ts
```diff
@@ -56,42 +86,40 @@ export type LogHandler = (
...
+  [LogLevel.ERROR]: 'error'
+};
+
 /**
  * The default log handler will forward DEBUG, VERBOSE, INFO, WARN, and ERROR
  * messages on to their corresponding console counterparts (if the log method
  * is supported by the current log level)
  */
-const defaultLogHandler: LogHandler = (instance, logType, ...args): void => {
+const defaultLogHandler: LogHandler = (instance, logType, message): void => {
```
Feiyang1: Aren't we going to miss the additional args in log messages?

packages/logger/test/custom-logger.test.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    it('parses multiple arguments correctly', () => {
+      client1.info('info message!', ['hello'], 1, { a: 3 });
+      expect(result.message).to.equal('info message! ["hello"] 1 {"a":3}');
+      expect(result.args).to.deep.equal(['info message!', ['hello'], 1, { a: 3 }]);
+      expect(result.level).to.equal('info');
+      expect(result.type).to.equal('@firebase/test-logger');
+    });
+
+    it('calls custom callback when log call is above set log level', () => {
```
Feiyang1: Can you please add a comment that log level is defaulted to`INFO`  here and in other tests where we assume the default log level?

packages/firebase/index.d.ts
hsubox76: Done.

packages/firebase/index.d.ts
hsubox76: Done.

packages/firebase/index.d.ts
hsubox76: Done.

packages/firebase/index.d.ts
hsubox76: Done.

packages/firebase/index.d.ts
hsubox76: There aren't any exceptions I can think of but I wanted to be safe in case we log a message from some util or pass through an error from a dep or something?

packages/firebase/index.d.ts
hsubox76: Done.

packages/firebase/index.d.ts
hsubox76: Done.

packages/logger/src/logger.ts
hsubox76: I think this must have been overlooked, fixed.

packages/logger/src/logger.ts
hsubox76: Unless I misunderstand, I think that's the current behavior.  It will be set to `instance.logLevel` but if `options.level` exists, that will override it.

packages/logger/test/custom-logger.test.ts
hsubox76: Done. It is also in this `describe` block title.

packages/logger/src/logger.ts
Feiyang1: You don't want to fix the loglevel to the value of `instance.logLevel` at the point of callback registration as it could be changed later. You want to use the up to date value of `instance.logLevel` when the callback is invoked.

packages/logger/src/logger.ts
hsubox76: Moved.

packages/logger/src/logger.ts
hsubox76: Oh, got it, fixed.

packages/firebase/index.d.ts
Feiyang1: I feel we should just say it will be a package name since it's the case at the moment.

hsubox76: Done.

PR #2436, opened 2019-12-12
Add verify support

thebrianchen:
Implements b/135487220.

First pass, not sure what the best failure messages are. I also thought about adding additional tests, but I think the two current ones capture its functionality.


packages/firestore/src/core/transaction.ts
```diff
@@ -102,12 +107,11 @@ export class Transaction {
...
-    if (!unwritten.isEmpty()) {
-      throw new FirestoreError(
-        Code.INVALID_ARGUMENT,
-        'Every document read in a transaction must also be written.'
-      );
-    }
+    // For each document that was read but not written to, we want to perform
+    // a `verify` operation.
+    unwritten.forEach((key, _version) => {
+      this.mutations.unshift(new VerifyMutation(key, this.precondition(key)));
```
wilhuff: Verify operations only need to precede writes for the same key--there's no requirement for ordering among keys. Each unwritten key has no other mutations for that key in the `mutations` array (by definition). This means that there's no constraint on where in the array the verify operations for the unwritten keys live. As a result I think you can just use `push` here instead of `unshift`: `unshift` is (usually) an O(n) operation where `push` is typically amortized O(1).

PR #2444, opened 2019-12-13
Extract WebStorage schema

schmidt-sebastian:
This PR moves all the schema parts of SharedClientState to a new file. We can then use the generated .d.ts file to make sure the schema doesn't get minified.

AFAIK, this is the only file that uses a schema in WebStorage. The Zombie key is assembled as a string, and the value is also only a string.

### DIFF
--- a/packages/firestore/src/local/shared_client_state_schema.ts
+++ b/packages/firestore/src/local/shared_client_state_schema.ts
@@ -1,0 +1,141 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { BatchId, MutationBatchState, TargetId } from '../core/types';
+import { QueryTargetState } from './shared_client_state_syncer';
+import { assert } from '../util/assert';
+import { ClientId } from './shared_client_state';
+import { User } from '../auth/user';
+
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -0,0 +1,141 @@
...
+   */
+  readonly clientId: string;
+  readonly onlineState: string;
+}
+
+// The WebStorage key prefix for the key that stores the last sequence number allocated. The key
+// looks like 'firestore_sequence_number_<persistence_prefix>'.
+export const SEQUENCE_NUMBER_KEY_PREFIX = 'firestore_sequence_number';
+
+/** Assembles the key for the current sequence number. */
```
schmidt-sebastian: This function is new (was inlined before).

packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -0,0 +1,141 @@
...
+  updateTimeMs: number;
+}
+
+// The WebStorage prefix that stores the primary tab's online state. The
+// format of the key is:
+//     firestore_online_state_<persistence_prefix>
+export const ONLINE_STATE_KEY_PREFIX = 'firestore_online_state';
+
+/** Assembles the key for the online state of the primary tab. */
+export function createWebStorageOnlineStateKey(persistenceKey: string): string {
```
schmidt-sebastian: This function is new (was inlined before).

packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -0,0 +1,141 @@
...
+import { assert } from '../util/assert';
+import { ClientId } from './shared_client_state';
+import { User } from '../auth/user';
+
+// The format of the LocalStorage key that stores the client state is:
+//     firestore_clients_<persistence_prefix>_<instance_key>
+export const CLIENT_STATE_KEY_PREFIX = 'firestore_clients';
+
+/** Assembles the key for a client state in WebStorage */
+export function createWebStorageClientStateKey(
```
mikelehen: Out of curiosity (I don't care too much), does the fact that these functions are defined in this .d.ts file mean that they won't get minified?  Or is that not how it works?

schmidt-sebastian: The function name doesn't, but I thought it would be cleaner to have all schema-related information in this one file.

PR #2445, opened 2019-12-13
Fix Integration tests against 'minified' sources

schmidt-sebastian:
This PR fixes `npm run test:manual` in 'integration/firestore'.

It moves all tests that require Internal API to the respective "api_internal" tests. It also does the same for the helpers. For the timestampsInSnapshot tests, it unfortunately re-adds the logging to the console output (running a test that verifies that we don't minify our settings is probably a good enough justification).

Two follow up PRs for this:
- Replace Webpack with Rollup in the  'integration/firestore' build
- Make this part of CI, so it doesn't end up breaking again

PR #2448, opened 2019-12-14
Add minified integration tests to Firestore

schmidt-sebastian:
This changes the tests in integration/firestore to use the "minified" browser build and lets us verify our SDK once we turn on property name mangling.

This is pretty hacky and relies on a complete build of all SDKs, but short of converting our integration/firestore to Rollup, this might be the easiest way. Comments/suggestions more than welcome!


schmidt-sebastian:
https://github.com/firebase/firebase-js-sdk/pull/2445 makes them run again (or is supposed to - the PR currently fails to resolve its dependencies in CI as well as locally)

Issue #1491, opened 2019-01-22
request.auth is null in Firestore Rules when called from Firebase JS SDK in UWP


tomasszabo:
I've discovered a problem when querying Firestore from UWP Universal Javascript project with Firebase JS SDK.

**Problem description**

Problem occurs when querying Firestore with authenticated user and Firestore Security Rules set to:

````
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read, write: 
      	if request.auth.uid != null;        
    }
  }
}
````

I'm always receiving:

````
FirebaseError: [code=permission-denied]: Missing or insufficient permissions.
````

With try and fail method I was able to find out that 'request.auth' object is 'null' in Security Rules because with following rule I'm able to successfully query Firestore:

````
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read, write: 
      	if request.auth == null;   
    }
  }
}
````

However this Security Rule enables reading of data for everyone and stored data will not be secured. 

In my opinion, the problem is related to the fact that access token is not processed (therefore request.auth is null). The only difference I've found is that HTTP header is NOT 'Origin: file://' (Origin header is completely missing) and CORS headers are not set correctly set in Response (CORS header are missing, but I'm not sure if this is the reason why access token is not processed).

**Expected behavior**

Auth token should be processed and Firestore Security Rules applied accordingly.

**Attachements**

UWP Project to reproduce the error (need some Firebase Test project and Firebase user):

https://1drv.ms/u/s!AjL4mBIpdvWDnYZqNkla5hHjjguGVg

HTTP Request from UWP:

````

   Request URL: https://firestore.googleapis.com/google.firestore.v1beta1.Firestore/Listen/channel?database=projects%2__project__%2Fdatabases%2F(default)&VER=8&RID=21133&CVER=22&X-HTTP-Session-Id=gsessionid&%24httpHeaders=Authorization%3ABearer%20__access-token__%0D%0AX-Goog-Api-Client%3Agl-js%2F%20fire%2F5.7.2%0D%0A&zx=7qsckzfmr536&t=1
   Request Method: POST
   Status Code: 200 / 
 
 - Request Headers
   Accept: */*
   Accept-Encoding: gzip, deflate, br
   Accept-Language: en-US, en; q=0.7, cs; q=0.3
   Cache-Control: no-cache
   Content-Length: 662
   Content-Type: application/x-www-form-urlencoded
[...]

PR #2452, opened 2019-12-18
Add electron enviroment check to http headers overwrite

agray5:
Checks for Electron environment before overwriting HTTPS headers.

### DIFF
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -89,7 +89,7 @@ export function isReactNative(): boolean {
  * @return true if Electron environment is detected.
  */
 export function isElectron(): boolean {
-  const userAgent = navigator.userAgent.toLowerCase();
+  const userAgent = getUA();
   return (userAgent.indexOf(' electron/') > -1) 
 }
 
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -84,6 +84,16 @@ export function isReactNative(): boolean {
 }
 
 /**
+ * Detect Electron.
+ *
+ * @return true if Electron environment is detected.
+ */
+export function isElectron(): boolean {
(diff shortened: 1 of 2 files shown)


agray5:
@googlebot I signed it!

agray5:
@googlebot I fixed it.

agray5:
@googlebot I fixed it.

Feiyang1:
Hi @agray5 Can you please tell us what the motivation is for this change?
We added this check to workaround the header issues in react native. https://github.com/firebase/firebase-js-sdk/issues/703

Are you facing a similar issue in Electron?

mikelehen:
@Feiyang1 FYI- The context here is https://github.com/firebase/firebase-js-sdk/issues/1491.  It turns out this issue happens whenever the browser environment doesn't send an `Origin` header. React Native ran into this, but Electron/7 and other environments seem to also be hitting it.

packages/util/src/environment.ts
```diff
@@ -83,6 +83,16 @@ export function isReactNative(): boolean {
   );
 }
 
+/**
+ * Detect Electron.
+ *
+ * @return true if Electron environment is detected.
+ */
+export function isElectron(): boolean {
+  const userAgent = navigator.userAgent.toLowerCase();
```
mikelehen: `navigator` may not be defined in some environments (e.g. node.js) and so this could crash.  The easy solution is to use the existing `getUA()` helper.

agray5:
@googlebot I fixed it.

mikelehen:
Thanks @agray5 I ended up just rolling this up into https://github.com/firebase/firebase-js-sdk/pull/2464/files which I will try to get checked in and released sometime in January.  Appreciate the help!

PR #2460, opened 2019-12-19
Remove Logger from performance backend log use case

hsubox76:
In preparation for some changes to `Logger` that will open up some methods to users, we should remove it from this internal use case that doesn't make use of any of `Logger`'s functionality anyway.

### DIFF
--- a/packages/performance/src/services/cc_service.ts
+++ b/packages/performance/src/services/cc_service.ts
@@ -136,7 +136,6 @@ export function ccHandler(
   serializer: (...args: any[]) => string
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
 ): (...args: any[]) => void {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   return (...args) => {
     const message = serializer(...args);
     addToQueue({
--- a/packages/performance/src/services/cc_service.ts
+++ b/packages/performance/src/services/cc_service.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import { LogHandler, Logger, LogLevel } from '@firebase/logger';
 import { SettingsService } from './settings_service';
 import { ERROR_FACTORY, ErrorCode } from '../utils/errors';
 import { consoleLogger } from '../utils/console_logger';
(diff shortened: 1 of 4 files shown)


packages/performance/src/services/cc_service.ts
```diff
@@ -135,10 +134,9 @@ function addToQueue(evt: BatchEvent): void {
 export function ccHandler(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   serializer: (...args: any[]) => string
-): LogHandler {
-  // The underscores for loggerInstance and level parameters are added to avoid the
-  // noUnusedParameters related error.
-  return (_loggerInstance: Logger, _level: LogLevel, ...args) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): (...args: any[]) => void {
```
Feiyang1: `unknown[]` instead ?

packages/performance/src/services/perf_logger.ts
```diff
@@ -83,18 +82,18 @@ interface TraceMetric {
   counters?: Array<{ key: string; value: number }>;
   custom_attributes?: Array<{ key: string; value: string }>;
 }
+
 /* eslint-enble camelcase */
 
-let logger: Logger | undefined;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Feiyang1: The eslint suppression doesn't seem necessary.

packages/performance/src/services/perf_logger.ts
```diff
@@ -83,18 +82,18 @@ interface TraceMetric {
...
+let logger: (resource: {}, resourceType: ResourceType) => void | undefined;
 // This method is not called before initialization.
-function getLogger(): Logger {
-  if (logger) {
-    return logger;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Feiyang1: unnecessary eslint suppression. Is it possible to change the type fo `resource` to `unknown`?

packages/performance/src/services/cc_service.ts
hsubox76: Done.

packages/performance/src/services/perf_logger.ts
hsubox76: Done.

packages/performance/src/services/perf_logger.ts
hsubox76: Changed.

packages/performance/src/services/perf_logger.ts
```diff
@@ -83,18 +82,16 @@ interface TraceMetric {
   counters?: Array<{ key: string; value: number }>;
   custom_attributes?: Array<{ key: string; value: string }>;
 }
+
 /* eslint-enble camelcase */
 
-let logger: Logger | undefined;
+let logger: (resource: unknown, resourceType: ResourceType) => void | undefined;
```
alikn: A better type for resource is NetworkRequest | Trace. Same for serializer resource type.

hsubox76: Done!

PR #2464, opened 2019-12-20
Avoid using $httpHeaders in environments that don't send an Origin header.

mikelehen:
There is a backend bug (b/145624756) causing our $httpHeaders query param to be ignored if the request doesn't contain an Origin header.  Via https://github.com/firebase/firebase-js-sdk/issues/1491 we've found that several environments can run into this issue so until the backend issue is fixed we are excluding them from using $httpHeaders. This may incur an extra CORS preflight round-trip in some cases, but this is much better than having our Authorization header get ignored.

### DIFF
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -83,6 +83,22 @@ export function isReactNative(): boolean {
   );
 }
 
+/** Detects Electron apps. */
+export function isElectron(): boolean {
+  return getUA().indexOf('electron/') >= 0;
+}
+
+/** Detects Internet Explorer. */
+export function isIE(): boolean {
+  const ua = getUA();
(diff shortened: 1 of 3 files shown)


packages/util/src/environment.ts
```diff
@@ -83,6 +83,22 @@ export function isReactNative(): boolean {
   );
 }
 
+/** Detects Electron apps. */
+export function isElectron(): boolean {
+  return getUA().indexOf('Electron/') >= 0;
```
hsubox76: Looks good, just thought it was an interesting coincidence [this was posted today](https://groups.google.com/a/chromium.org/forum/m/#!msg/blink-dev/-2JIRNMWJ7s/yHe4tQNLCgAJ), definitely something we need to keep an eye on!

PR #2466, opened 2019-12-20
Fix flaky RTDB tests due to increment operations.

mikelehen:
The introduction of tests that try to send increment operations to the backend has resulted in test flakiness.  Basically the emulator rejects these as invalid operations and closes the connection.  The client automatically reconnects, so _usually_ this doesn't matter.  But the SDK intentionally doesn't tolerate disconnects in the middle of a transaction, so if the disconnect happens to happen during a transaction test, then the test will fail.

I'm avoiding the problem by isolating the increment tests better (use an isolated instance of RTDB and ensures it remains offline so it doesn't even try to send the requests to the emulator).

I've opened b/146657568 to track cleaning this up once the emulator supports increment operations.

### DIFF
--- a/packages/database/test/servervalues.test.ts
+++ b/packages/database/test/servervalues.test.ts
@@ -16,7 +16,7 @@
  */
 
 import { expect } from 'chai';
-import { getRandomNode } from './helpers/util';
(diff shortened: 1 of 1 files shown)


mikelehen:
Thanks Jeff!

PR #2484, opened 2019-12-31
Messaging rewrite

mmermerkaya:
I didn't want to leave the FCM SDK in it's current state, so I rewrote a large portion of it (and hopefully without any API changes this time :smile:). There are also some API changes that I'd really like to do, like removing `getToken` and `deleteToken` from `SWController` or making `vapidKey` a parameter of `getToken`, but I guess they'll have to wait.

**Changes:**
- Composition over inheritance
  - Removed `BaseController` that the two main controllers were extending and moved the common parts to "token manager" module.
  - Removed `DbInterface`, which was not actually an interface but an abstract class, and the two classes that were extending it.
- Database
  - Moved Push API related fields to the `subscriptionOptions` subfield. This is now an optional field, as preparation for the eventual Safari support, which will not use the Push API.
  - Dropped `swScope` as an index, for the same reason as before.
  - Database key changed to App ID, as there should only be one token for every App ID. This should prevent "multiple messages received" issues in topic sends. 
  - Created a new DB and added a migration function because the object store needed to change. This allowed removing the old DB code.
  - Started using the battle-tested IndexedDB library [idb](https://github.com/jakearchibald/idb/), which is already being used in Installations.
- Size reduction
  - Non minified ES5 ESM build is now 70.8kB. Old one was 96.8kB.
  - Minified and gzipped ES5 build is now 10.1kB. Old one was 12.5kB.
- Tests
  - Increased coverage from 93% to 99%. Coveralls reports a drop in coverage because I also changed the coverage config to stop including test files in coverage reports.
  - Made the tests more meaningful by removing all private member access from tests. Focusing on testing behavior, not implementation.
  - Added complete Service Worker fakes to use with every test. Removed messing with prototype objects in fakes.
  - Removed the uses of `any` type from tests. There used to be 83 instances of type `any` in tests.
- Structure and Naming:
  - Greatly reduced the number of classes. Not everything needs to be a class.
  - Combined src/ and test/ folders, following our style guide.
  - Removed the -model suffix. More meaningful names overall.


config/webpack.test.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
           options: { esModules: true }
         },
         enforce: 'post',
-        include: path.resolve(process.cwd(), 'src')
+        exclude: [/\.test\.ts$/, /test(ing)?\//]
```
dwoffinden: presumably this would also exclude a directory called "nottesting" :stuck_out_tongue: 

would a \b "word boundary" work?

/\btest(ing)?\//

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2019 Google Inc.
```
dwoffinden: go/copyright

It seems like this shouldn't be replaced, and you can add the current year so long as you preserve the original year? So 2017 or 2017-2019 but not 2019?

packages/messaging/src/core/api.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        new Response(JSON.stringify({ error: { message: 'error message' } }))
+      );
+      await expect(
+        requestUpdateToken(firebaseDependencies, tokenDetails)
+      ).to.be.rejectedWith('messaging/token-update-failed');
+
+      fetchStub.resolves(
+        new Response(
+          JSON.stringify({
+            /* no token */
```
dwoffinden: You could replace these four lines with the four characters "{}" but I guess consistency is important too :stuck_out_tongue:

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2019 Google Inc.
```
dwoffinden: Again with the copyright

I'll stop mentioning it but there are other files too :stuck_out_tongue:

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -86,254 +92,144 @@ export class WindowController extends BaseController {
...
+      this.onMessageCallback(payload);
+    }
+
+    const { data } = payload;
+    if (
+      data &&
+      FN_CAMPAIGN_ID in data &&
+      data[FN_CAMPAIGN_ANALYTICS_ENABLED] === '1'
+    ) {
+      // This message has a campaign id, meaning it was sent using the FN Console.
```
dwoffinden: The branding seems to be "Firebase Cloud Messaging" / "the Firebase Console" / "the Notifications composer"

https://firebase.google.com/docs/cloud-messaging/js/send-with-console

So if you want the comment to make sense to someone who doesn't know what "FN" means, I'd just say "the Firebase Console" or "the Notifications composer".

I guess the FN_... constants should just be COMPOSER or CONSOLE?

Also "id" -> "ID" (we aren't talking about https://en.wikipedia.org/wiki/Id,_ego_and_super-ego#Id)

packages/messaging/src/core/api.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
dwoffinden: ok if this is a new file then it *should* use 2019 :smile:

packages/messaging/src/core/token-management.ts
```diff
@@ -0,0 +1,179 @@
...
+import { dbGet, dbSet, dbRemove } from '../helpers/idb-manager';
+import { FirebaseInternalDependencies } from '../interfaces/internal-dependencies';
+import { SubscriptionOptions } from '../interfaces/subscription-options';
+import { TokenDetails } from '../interfaces/token-details';
+import { requestUpdateToken, requestGetToken, requestDeleteToken } from './api';
+import { arrayToBase64 } from '../helpers/array-to-base64';
+import { ERROR_FACTORY, ErrorCode } from '../util/errors';
+
+/** UpdateRegistration will be called once every week. */
+const TOKEN_EXPIRATION_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
```
dwoffinden: So much sadness from the lack of a ubiquitous date/time api :(

packages/messaging/src/helpers/array-to-base64.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { arrayToBase64 } from './array-to-base64';
+import { expect } from 'chai';
+import '../testing/setup';
+
+describe('arrayToBase64', () => {
+  it('returns expected value for p256dh example', () => {
+    // prettier-ignore
+    const buffer = new Uint8Array([
+      4, 181, 98, 240, 48, 62, 75, 119, 193, 227, 154, 69, 250, 216, 53, 110,
```
dwoffinden: Curious: any particular reason for these numbers? A valid test key? Lottery numbers?

For test code, if the exact values don't matter, it's usually an excuse to use joke data that's obviously unimportant, or at least exercises the edge cases :p

packages/messaging/src/helpers/idb-manager.ts
```diff
@@ -0,0 +1,106 @@
...
+// Exported for tests.
+export const DATABASE_NAME = 'firebase-messaging-database';
+const DATABASE_VERSION = 1;
+const OBJECT_STORE_NAME = 'firebase-messaging-store';
+
+let dbPromise: Promise<DB> | null = null;
+function getDbPromise(): Promise<DB> {
+  if (!dbPromise) {
+    dbPromise = openDb(DATABASE_NAME, DATABASE_VERSION, upgradeDB => {
+      // We don't use 'break' in this switch statement, the fall-through
```
dwoffinden: Solid logic, if a little confusing when you only have one case :smile:

packages/messaging/src/helpers/idb-manager.ts
```diff
@@ -0,0 +1,106 @@
...
+
+// Exported for tests.
+export const DATABASE_NAME = 'firebase-messaging-database';
+const DATABASE_VERSION = 1;
+const OBJECT_STORE_NAME = 'firebase-messaging-store';
+
+let dbPromise: Promise<DB> | null = null;
+function getDbPromise(): Promise<DB> {
+  if (!dbPromise) {
+    dbPromise = openDb(DATABASE_NAME, DATABASE_VERSION, upgradeDB => {
```
dwoffinden: upgradeDb

(else, ought the method not be getDBPromise?)

packages/messaging/src/helpers/migrate-old-database.ts
```diff
@@ -0,0 +1,193 @@
...
+  createTime: number;
+}
+
+const OLD_DB_NAME = 'fcm_token_details_db';
+/**
+ * The last DB version of 'fcm_token_details_db' was 4. This is one higher,
+ * so that the upgrade callback is called for all versions of the old DB.
+ */
+const OLD_DB_VERSION = 5;
+const OLD_OBJECT_STORE_NAME = 'fcm_token_object_Store';
```
dwoffinden: Why capital "S"?

packages/messaging/src/interfaces/message-payload.ts
```diff
@@ -29,5 +27,5 @@ export interface FcmOptions {
 export interface MessagePayload {
   fcmOptions?: FcmOptions;
   notification?: NotificationDetails;
-  data?: { [key: string]: unknown };
+  data?: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
dwoffinden: Isn't object better than any, in that it forces you to check whether anything is there rather than silently becoming anything?

packages/messaging/src/testing/compare-headers.ts
```diff
@@ -0,0 +1,38 @@
...
+// so we need to do that manually.
+export function compareHeaders(
+  expectedHeaders: HeadersWithEntries,
+  actualHeaders: HeadersWithEntries
+): void {
+  expect(expectedHeaders.entries).not.to.be.undefined;
+  expect(actualHeaders.entries).not.to.be.undefined;
+
+  const expected = new Map(Array.from(expectedHeaders.entries!()));
+  const actual = new Map(Array.from(actualHeaders.entries!()));
```
dwoffinden: ISTM like a little helper to assert that the entries are there and then turn them into a Map would be useful, you could then call it for both sets of headers :p

packages/messaging/src/testing/compare-headers.ts
```diff
@@ -0,0 +1,38 @@
...
+// Chai doesn't check if Headers objects contain the same entries,
+// so we need to do that manually.
+export function compareHeaders(
+  expectedHeaders: HeadersWithEntries,
+  actualHeaders: HeadersWithEntries
+): void {
+  expect(expectedHeaders.entries).not.to.be.undefined;
+  expect(actualHeaders.entries).not.to.be.undefined;
+
+  const expected = new Map(Array.from(expectedHeaders.entries!()));
```
dwoffinden: Doesn't the Map constructor take an iterable why do you need Array.from()?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

packages/messaging/src/testing/fakes/firebase-dependencies.ts
```diff
@@ -47,24 +53,26 @@ export function makeFakeApp(options: FirebaseOptions = {}): FirebaseApp {
...
-export function makeFakeAnalyticsProvider(): Provider<
-  FirebaseAnalyticsInternalName
-> {
-  return new Provider<FirebaseAnalyticsInternalName>(
-    'analytics-internal',
-    new ComponentContainer('test')
-  );
+function getFakeAnalyticsProvider(): Provider<FirebaseAnalyticsInternalName> {
+  const analytics: FirebaseAnalyticsInternal = {
+    logEvent: () => {}
```
dwoffinden: drop ":" and "=>", just "logEvent() {}"? :)

packages/messaging/src/util/constants.ts
```diff
@@ -15,81 +15,17 @@
...
-};
+export const DEFAULT_VAPID_KEY =
+  'BDOU99-h67HcA6JeFXHbSNMu7e2yNNu3RzoMj8TM4W88jITfq7ZmPvIM1Iv-4_l2LxQcYwhqby2xGpWwzjfAnG4';
 
 export const ENDPOINT = 'https://fcmregistrations.googleapis.com/v1';
 
+/** Key of FCM Payload in Notification's data field. */
+export const FCM_MSG = 'FCM_MSG';
+
 export const FN_CAMPAIGN_ID = 'google.c.a.c_id';
```
dwoffinden: Yeah... while we're here should these be CONSOLE or COMPOSER rather than FN?

config/webpack.test.js
mmermerkaya: > "nottesting"

Nice footgun. :smile: 

Thanks though, I didn't know `\b`. Changed.

packages/messaging/src/controllers/sw-controller.ts
mmermerkaya: It's a bot that adds these, I think I removed the original and it added a new one with the current (well, it was current then) year. :man_shrugging:

I'll fix them.

packages/messaging/src/core/api.test.ts
mmermerkaya: Yeah that's true, but I think this way better expresses what I am actually doing here. Plus consistency, of course. :sweat_smile:

packages/messaging/src/helpers/array-to-base64.test.ts
mmermerkaya: No idea, it was before my time. :smile:

https://github.com/firebase/firebase-js-sdk/blob/73a586c92afe3f39a844b2be86086fddb6877bb7/packages/messaging/test/array-buffer-to-base64.test.ts

packages/messaging/src/helpers/idb-manager.ts
mmermerkaya: I think it's still helpful to whoever's going to maintain this, telling them that in case they want to add a mutation, this is a good way to do so.

And definitely much better than having this comment, and actual break statements right afterwards. :sweat_smile:

https://github.com/firebase/firebase-js-sdk/blob/7250d25145f27ee099dc717a0130c2f8dc9841d2/packages/messaging/src/models/token-details-model.ts#L101-L102

packages/messaging/src/helpers/idb-manager.ts
mmermerkaya: Yup, changed.

packages/messaging/src/helpers/migrate-old-database.ts
mmermerkaya: No idea. It has always been like that and it always bothered me.

https://github.com/firebase/firebase-js-sdk/blob/73a586c92afe3f39a844b2be86086fddb6877bb7/packages/messaging/src/models/token-details-model.ts#L22

packages/messaging/src/interfaces/message-payload.ts
mmermerkaya: Good point. Made it `unknown`.

packages/messaging/src/testing/compare-headers.ts
mmermerkaya: Nice.

packages/messaging/src/testing/compare-headers.ts
mmermerkaya: It worked, so yes it does. :smile:

packages/messaging/src/testing/fakes/firebase-dependencies.ts
mmermerkaya: Sure.

packages/messaging/src/controllers/window-controller.ts
mmermerkaya: "Firebase Console" sounds good. Changed the constants as well.

packages/messaging/src/util/constants.ts
mmermerkaya: Changed to `CONSOLE`.

packages/messaging-types/index.d.ts
```diff
@@ -25,7 +25,10 @@ import {
 } from '@firebase/util';
 
 export class FirebaseMessaging {
-  private constructor();
+  private constructor(app: FirebaseApp);
+  readonly app: FirebaseApp;
```
hsubox76: I don't know Typescript or this codebase that well but I noticed both classes that implement this have a constructor that takes a `FirebaseInternalDependencies` and don't have a top level `app` property, is this type generalized for another use case I'm missing?

mmermerkaya: Thanks for taking a look!

My understanding was that all Firebase services need to have an `app` field (because [this test](https://github.com/firebase/firebase-js-sdk/blob/7250d25145f27ee099dc717a0130c2f8dc9841d2/integration/shared/namespaceDefinition.json#L212) exists), so I added it here, not realizing that it is already defined in the `FirebaseService` interface [here](https://github.com/firebase/firebase-js-sdk/blob/7250d25145f27ee099dc717a0130c2f8dc9841d2/packages/app-types/private.d.ts#L38).

I changed the two classes to implement that interface instead, and removed the `app` field from this one. Please let me know if `app` (and also `INTERNAL`) isn't actually required in the core packages (there's no type checking for it AFAICT), in which case I'll remove it, since it's not actually part of the FCM API.

And I have no idea what happened to that constructor, that's definitely a mistake. :sweat_smile:

hsubox76: As far as I can tell INTERNAL and app aren't required in the code anywhere, there is one place that `service.INTERNAL.delete()` is called on services, but it checks to see that `INTERNAL` exists first. I'm not sure if we should remove the `app` getter since it seems like a staple that maybe is useful for debugging or some future plan I don't know about, and it doesn't cost much.

mmermerkaya: Alright, thanks for checking, I'll keep it then. :slightly_smiling_face:

PR #2486, opened 2020-01-03
Add onIdChange function to Installations

mmermerkaya:


packages/installations/src/helpers/fid-changed.ts
```diff
@@ -0,0 +1,102 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { getKey } from '../util/get-key';
+import { AppConfig } from '../interfaces/app-config';
+import { IdChangeCallbackFn } from '../functions';
+
+const FID_CHANGE_CALLBACKS: Map<string, Set<IdChangeCallbackFn>> = new Map();
```
dwoffinden: "Constants: CONSTANT_CASE indicates that a value is intended to not be changed, and may be used for values that can technically be modified (i.e. values that are not deeply frozen) to indicate to users that they must not be modified."

Pretty sure you modify this :p

packages/installations/src/helpers/fid-changed.ts
```diff
@@ -0,0 +1,102 @@
...
+  // Open the broadcast channel if it's not already open,
+  // to be able to listen to change events from other tabs.
+  getBroadcastChannel();
+
+  const key = getKey(appConfig);
+
+  if (!FID_CHANGE_CALLBACKS.has(key)) {
+    FID_CHANGE_CALLBACKS.set(key, new Set());
+  }
+  FID_CHANGE_CALLBACKS.get(key)!.add(callback);
```
dwoffinden: You could probably avoid the double lookup in this idiom, not sure how much it actually helps but you needn't look up the thing you just inserted...

```
  let callbacks = FID_CHANGE_CALLBACKS.get(key);
  if (callbacks == null) {
    callbacks = new Set();
    FID_CHANGE_CALLBACKS.set(key, callbacks);
  }
  callbacks.add(callback);
```

packages/installations/src/helpers/fid-changed.ts
```diff
@@ -0,0 +1,102 @@
...
+
+function broadcastFidChange(key: string, fid: string): void {
+  const channel = getBroadcastChannel();
+  if (channel) {
+    channel.postMessage({ key, fid });
+  }
+  closeBroadcastChannel();
+}
+
+let BROADCAST_CHANNEL: BroadcastChannel | null = null;
```
dwoffinden: Again, definitely not a constant :stuck_out_tongue:

mmermerkaya: :man_facepalming:

packages/installations/src/helpers/fid-changed.ts
mmermerkaya: Done.

packages/installations/src/helpers/fid-changed.ts
mmermerkaya: Done.

packages/installations/src/helpers/fid-changed.ts
mmermerkaya: Done.

mmermerkaya:
@dwoffinden I changed this a bit, could you take another look?

PR #2490, opened 2020-01-06
Fix flaky Installations test

mmermerkaya:


packages/installations/src/helpers/get-installation-entry.test.ts
```diff
@@ -354,12 +354,19 @@ describe('getInstallationEntry', () => {
       const { registrationPromise } = await installationEntryPromise;
 
       // Let the new getInstallationEntry process start.
-      await sleep(10);
+      await sleep(50);
 
-      expect(await get(appConfig)).to.deep.equal({
+      const tokenDetails = (await get(
+        appConfig
```
dwoffinden: Is there a workaround for this formatting? :P

mmermerkaya: I could add a `prettier-ignore` comment but I'd rather not.

PR #2495, opened 2020-01-07
Add minified ("mangled") Firestore build

schmidt-sebastian:
Sending this one out for early review. Doc for more context: go/minification

This PR adds a new Firestore where most identifiers a mangled (a size saving of  9% gzipped). I don't know too much about the Rollup pipeline, so any suggestions there are more than appreciated.

The basic steps involved in this minification/mangling are:

- Gather a list of symbols that are used in any of Firestore's external dependencies (this includes the JavaScript library, other Firebase SDKs, the Protobuf files as well as our LocalStorage and IndexedDb schema).
- Using a Typescript Transformer, prefix all symbols that are not used in said public API with __PRIVATE_
- Use Terser's property mangling with to minify all renamed symbols.

The output of this process are a couple of new files:

- packages/firebase/firebase-firestore.min.js
- packages/firestore/dist/index.cjs.min.js
- packages/firestore/dist/index.esm.min.js
- packages/firestore/dist/index.esm2018.min.js

We discussed internally to only have a single build and make out default build use the new mangled property names. That might be coming before we release this, but we need to figure out how to run our integration test against the mangled build first.


packages/firestore/.npmignore
```diff
@@ -2,9 +2,10 @@
 /src
 test
 .idea
+.cache
```
hsubox76: So I think this is meant to go in `.gitignore` at the top level, at least, that is what fixed my IDE wanting to add 1000 files to the commit after building.

Fun fact, I believe `.npmignore` files do not do anything in our repo anymore because you can either whitelist with the `files` field in `package.json` OR blacklist with `.npmignore` and if you use `files` it completely ignores `.npmignore`.  We should probably delete this and set up the package.json `files` whitelist to be a bit more granular (it will pack everything in /dist into the npm package right now - we actually do need `src` for `d.ts` files but we don't need `test`).

packages/firestore/rollup.config.js
```diff
@@ -15,27 +15,72 @@
...
+};
+
 /**
  * ES5 Builds
  */
 const es5BuildPlugins = [
   typescriptPlugin({
-    typescript
+    typescript,
+    cacheRoot: './.cache/es5/'
```
hsubox76: Just curious about the reason for setting custom `cacheRoot` dirs for all the builds in this file.

packages/firestore/.npmignore
schmidt-sebastian: But if we check them into git, then we save everyone a bunch of time pre-building their cache! Just kidding - I added them to .gitignore. I wonder why my IDE didn't ask me to commit those 1000 files.

packages/firestore/rollup.config.js
schmidt-sebastian: This drove me nuts until I figured it out: The cache for the typescript plugin prevents any rebuild for any typescript file if it hasn't changed - regardless of transformers. So if you have two different builds that each touch the same file, the cache is re-used and the cached result from the first build is used for both builds. I solved this by using a different cache for each transformer setting.

Plantain:
>go/minification
Can you share this doc externally?

schmidt-sebastian:
> go/minification
> Can you share this doc externally?

Sorry - most of our design docs (including this one) are internal only for now. I can promise that it is not all that interesting :)

packages/firestore/package.json
```diff
@@ -48,8 +52,10 @@
     "protobufjs": "6.8.8",
     "rollup": "1.28.0",
     "rollup-plugin-copy-assets": "1.1.0",
+    "rollup-plugin-json": "4.0.0",
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "2.2.0",
+    "rollup-plugin-terser": "5.1.2",
```
hsubox76: Unless this version is important you could probably move up to terser 5.1.3 and not have 2 separate versions in yarn.lock

schmidt-sebastian: Updated to 5.1.3. Thanks!

stefanklokgieters:
Will this also be part of the Firebase package? Did not see a version bump yet 😄

schmidt-sebastian:
This hasn't been released yet. We are working on it!

Issue #2487, opened 2020-01-03
@firebase/database@0.5.13 ref.update is very slow/unusable


jmschae:
### [REQUIRED] Describe your environment

  * Operating System version: docker - node alpine 12
  * Browser version: n/a
  * Firebase SDK version: 0.5.13
  * Firebase Product: database (auth, database, storage, etc)

### [REQUIRED] Describe the problem

When updating from @firebase/database@0.5.12 to @firebase/database@0.5.13 updates became very slow. We operate in a relatively high write environment, and are streaming state updates into firebase. When upgrading from 0.5.12 to 0.5.13 we began seeing not just slow downs, but issues with ack'ing messages to pub/sub. We tracked this down to a change in the @firebase/database library, specifically commit 86971ea3c727084fa1a99322afe473d19068bac5. A change occurred in src/core/Repo.ts that added:

`this.serverSyncTree_.calcCompleteEventCache(path)`

to the Repo.update function. This update changed the time it took this loop:

```
each(childrenToMerge, (changedKey: string, changedValue: unknown) => {
  empty = false;
  const newNodeUnresolved = nodeFromJSON(changedValue);
  changedChildren[changedKey] = resolveDeferredValueSnapshot(
    newNodeUnresolved,
    this.serverSyncTree_.calcCompleteEventCache(path),
    serverValues
  );
});
```
to complete from an average of 0ms to 750ms. This average of 750ms spans a range of about 5ms to upwards of 10,000ms. We believe this slowdown to be the cause of both our increase in observed update time, and our issues ack'ing message to pub/sub, since this piece of code seemingly never releases the event loop.

#### Steps to reproduce:

Write ~100 updates continuously as fast as you can, and track the time it takes to complete an update.

katowulf:
@jmschae thanks for this, could I ask you to go a bit further and provide a working repro, with sample data and rules, we can run locally that shows the comparison? 

The code snippet is useful, but this is likely to be nuanced and hard to repro, so getting a complete [mcve](http://stackoverflow.com/help/mcve) will be a necessary starting point.

schmidt-sebastian:
@inlined - Do you have time to look at this?

https://github.com/firebase/firebase-js-sdk/commit/86971ea3c727084fa1a99322afe473d19068bac5 was merged in preparation for a new feature. Until this feature is released, I would recommend staying on 0.5.12 if the performance of 0.5.13 does not meet your needs.

jmschae:
[...]

PR #2499, opened 2020-01-07
Add an interface to ServerValues that lets us read synctrees just in time

inlined:
Addresses #2487 by creating a new interface that lets us defer reading existing values from a SyncTree until it's actually needed and restricted to the smallest possible scope.

### DIFF
--- a/packages/database/src/core/util/ServerValues.ts
+++ b/packages/database/src/core/util/ServerValues.ts
@@ -26,6 +26,53 @@ import { ChildrenNode } from '../snap/ChildrenNode';
 import { SyncTree } from '../SyncTree';
 import { Indexable } from './misc';
 
+/* It's critical for performance that we not calculate actual values from a SyncTree
+ * unless and until the value is needed. Because we expose both a SyncTree and Node
+ * version of deferred value resolution, we ned a wrapper class that will let us share
+ * code.
+ *
+ * @see https://github.com/firebase/firebase-js-sdk/issues/2487
+ */
+interface DeferredExistingValue {
+  getImmediateChild(childName: string): DeferredExistingValue;
+  node(): Node;
+}
+
+class ExistingSnapshotValue implements DeferredExistingValue {
+  private node_: Node;
+  constructor(node: Node) {
+    this.node_ = node;
+  }
+
+  getImmediateChild(childName: string): DeferredExistingValue {
+    const child = this.node_.getImmediateChild(childName);
+    return new ExistingSnapshotValue(child);
+  }
+
+  node(): Node {
+    return this.node_;
+  }
+}
+
+class DeferredSyncTreeValue implements DeferredExistingValue {
+  private syncTree_: SyncTree;
+  private path_: Path;
+
+  constructor(syncTree: SyncTree, path: Path) {
+    this.syncTree_ = syncTree;
+    this.path_ = path;
+  }
+
+  getImmediateChild(childName: string): DeferredExistingValue {
+    const childPath = this.path_.child(childName);
+    return new DeferredSyncTreeValue(this.syncTree_, childPath);
+  }
+
+  node(): Node {
+    return this.syncTree_.calcCompleteEventCache(this.path_);
+  }
+}
+
 /**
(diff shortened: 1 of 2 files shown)


packages/database/src/core/util/ServerValues.ts
```diff
@@ -26,6 +26,53 @@ import { ChildrenNode } from '../snap/ChildrenNode';
 import { SyncTree } from '../SyncTree';
 import { Indexable } from './misc';
 
+/* It's critical for performance that we not calculate actual values from a SyncTree
```
schmidt-sebastian: s/not/do not/

schmidt-sebastian:
@mikelehen won't be able to take a look until later today. Unfortunately, that means that we have to revert https://github.com/firebase/firebase-js-sdk/pull/2348. We can likely put it back into the the next release.

inlined:
I'm a bit confused; are you saying that the current proposal is wrong or an over-optimization? In the case that there's a `{".sv": {"increment": X}}` then the change is roughly equivalent (though we'd still need refactoring since the snapshot path also does child traversal). Isn't it best to avoid hitting the SyncTree altogether when it's avoidable? There's no reason why this customer should ever have calculated a compiled result since the increment feature isn't even released yet.


> And since we're applying `update()` increments to the wrong existing value, it presumably won't behave properly. So we'll want to add a regression test for that as well.

What do you mean we're applying update() to the wrong existing value? The previous and current code both traverse down the tree; it's just that the last iteration calculated the cache too soon because I needed a full snapshot to be pulled to reuse the snapshot helper function.

mikelehen:
I'm saying:
1. The initial ServerValue.increment() implementation had a bug in update() that caused it to calculate the complete cache at the root of the `.update()` call over and over, instead of for the actual paths being updated.  So `rootRef.update({'a/b/c': 1, 'd/e/f': 2}) is going to call calcCompleteEventCache() for the entire repo twice instead of once for a/b/c and once for d/e/f
2. In addition to being a performance problem, this means that ref.update({'a/b/c': ServerValue.increment(1)})` would use the value at `ref` instead of at `ref.child('a/b/c')` as the base value for the increment, which means it will calculate the wrong result and raise incorrect events.
3. Given that the user that complained was using `update()`, they were likely hitting this problem and weren't even exercising the SyncTree codepath that this PR changed, since the SyncTree codepath is only used by onDisconnect operations.
4. This PR is likely an over-optimization.  calcCompleteEventCache() is already relatively cheap since all the underlying data structures are immutable and so it can reuse their structures when computing the merged event cache.

inlined:
Modified based on @mikelehen's observation that `update()` wasn't using the synctree path. It couldn't because the synctree code was tied up in `SparseSnapshotTree` logic for the sake of `onDisconnect`. I moved the loop to `onDisconnect` so I could use the optimization in `update()` as well.

Using the mvp regression test provided by @jmschae in  #2487 (thanks!) I was able to confirm that `increment-master` reduced the 1200% slowdown (N^2 write calculations) to a 20% slowdown (N write calculations). With this change, (0 write calculations) we actually beat the baseline performance in master, though I can't explain why and will chalk it up to noise.

packages/database/src/core/util/ServerValues.ts
```diff
@@ -26,6 +26,53 @@ import { ChildrenNode } from '../snap/ChildrenNode';
...
+ * @see https://github.com/firebase/firebase-js-sdk/issues/2487
+ */
+interface DeferredExistingValue {
+  getImmediateChild(childName: string): DeferredExistingValue;
+  node(): Node;
+}
+
+class ExistingSnapshotValue implements DeferredExistingValue {
+  private node_: Node;
+  constructor(node: Node) {
```
schmidt-sebastian: Nit: You can drop the suffix for members of classes that are not part of the public API, which will allow you to use constructor property assignments (`constructor(readaonly node:Node)`).

Renaming to `node` also allows you to get rid of the getter here.

inlined: If I do that I fail to implement the protocol because `node` is a value instead of a method.

PR #2506, opened 2020-01-08
Initialize transport service in Performance Monitoring when performance() is called.

zijianjoy:
1. Rename cc_service to transport_service since we expect migration to Firelog service in the future.
2. transport_service.ts used to start setTimeout() when this file is loaded. This change wraps it into initialization function, which should be called when performance() is called (it is when developer activates Performance Monitoring component).
3. Update testing to call useFakeTimers() inside describe() instead of before loading transport_service.ts.


packages/performance/src/services/perf_logger.test.ts
```diff
@@ -17,7 +17,7 @@
 
 import { stub, SinonStub, useFakeTimers, SinonFakeTimers } from 'sinon';
 import { Trace } from '../resources/trace';
-import * as ccService from './cc_service';
+import * as ccService from './transport_service';
```
visumickey: Since we are replacing all the context of CC to Transport, should we import it as transportService instead of ccService?

zijianjoy: Updated, thank you!

packages/performance/src/services/oob_resources_service.ts
```diff
@@ -27,6 +28,7 @@ export function setupOobResources(): void {
   if (!getIid()) {
     return;
   }
+  setupTransportService();
```
alikn: setupOobResources is called when the call to get iid and RC is responsed. Starting the transport cycle here means that the initial time of sending logs is "iid call + RC call + 5.5s". I suggest not including those calls in the initial delay to avoid it from being undeterministic.

packages/performance/src/services/transport_service.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  beforeEach(() => {
+    fetchStub = stub(window, 'fetch');
+  });
+
+  afterEach(() => {
+    fetchStub.restore();
+  });
+
+  it('throws an error when logging an empty message', done => {
```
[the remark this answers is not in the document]
zijianjoy: Removed for not async test.

packages/performance/src/services/oob_resources_service.ts
zijianjoy: Moved initialization call to `controllers/perf.ts`.

PR #2509, opened 2020-01-09
Get closure-library from google-closure-library.

mikelehen:
Get closure-library from google-closure-library instead of via closure-builder so we can control the version directly.

### DIFF
--- a/packages/webchannel-wrapper/package.json
+++ b/packages/webchannel-wrapper/package.json
@@ -16,7 +16,7 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "closure-builder": "2.3.8",
-    "google-closure-library": "^20200101.0.0",
+    "google-closure-library": "20200101.0.0",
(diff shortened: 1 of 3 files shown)


packages/webchannel-wrapper/package.json
```diff
@@ -16,6 +16,7 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "closure-builder": "2.3.8",
+    "google-closure-library": "^20200101.0.0",
```
schmidt-sebastian: We don't use version ranges in the Web SDK. Not that it matters for this versioning scheme, but it might be better to be consistent and drop the `^`.

mikelehen: Oops!  I just did "npm install --save-dev" or some such...  Thanks. I'll remember this for next time!

mikelehen:
@hsubox76 Can you approve the yarn.lock change?

PR #2511, opened 2020-01-09
Decode remote doc in a promise.

wu-hui:


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -281,16 +281,25 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
 
-        const maybeDoc = this.serializer.fromDbRemoteDocument(dbRemoteDoc);
-        if (!query.path.isPrefixOf(maybeDoc.key.path)) {
-          control.done();
-        } else if (maybeDoc instanceof Document && query.matches(maybeDoc)) {
-          results = results.insert(maybeDoc.key, maybeDoc);
-        }
+        this.decodeDocAsync(dbRemoteDoc).next(maybeDoc => {
+          if (!query.path.isPrefixOf(maybeDoc.key.path)) {
+            control.done();
```
schmidt-sebastian: In my opinion, IndexedDb is the worst API that was ever designed. Everything in IndexedDb is scoped by transactions, but the lifetime of a transaction essentially depends on the "threading behavior" of your code (or however one would call this on the Web). A transaction is committed to disk when the Javascript event loop returns control back to the browser and there are no outstanding IndexedDb operations. To put this in code:

```
indexedDb.openTransaction();
indexedDb.put(...);
indexedDb.put(...);
Promise.resolve().then(() => {
  // This put may fail if the transaction is no longer active. It may have been committed because 
  // Promise.resolve() is an asynchronous operation that momentarily returns control back from the 
  // event loop
  indexedDb.put(...);  
});
```

We work around this with the best API that was ever designed: Our PersistencePromise implementation is synchronous, but matches most other Promise semantics.

Long story short:
- What you have here works, but it doesn't parallelize document decoding since PersistencePromises block the event loop.
- If you change it to use a Promise, `control.done()` will be invoked after the transaction is committed.

Instead, you should be able to mirror what Android is doing:
- Decode the path first. If the path doesn't match, call `control.done()`.
- If the path matches, create a new Promise-returning arrow-function that decodes the document and insert it into the result set if the document matches. This function cannot access IndexedDb anymore.
- Before returning the results, use `Promise.all()` to wait for all results from the previous step.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -281,14 +282,24 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
-        if (!query.path.isPrefixOf(maybeDoc.key.path)) {
-          control.done();
-        } else if (maybeDoc instanceof Document && query.matches(maybeDoc)) {
-          results = results.insert(maybeDoc.key, maybeDoc);
+        if (!query.path.isPrefixOf(new ResourcePath(key))) {
+            control.done();
+            return;
         }
+
+        const decodeDoc = new Promise<void>((resolve, reject) => {
```
schmidt-sebastian: This could be a tad simpler:

```
Promise.resolve().then(() => {
     const maybeDoc = this.serializer.fromDbRemoteDocument(dbRemoteDoc);
     if (maybeDoc instanceof Document && query.matches(maybeDoc)) {
            results = results.insert(maybeDoc.key, maybeDoc);
       }
});
```

Also, I am not 100% certain that the code in `new Promise() { .... }` is already async - but the code that runs after a resolve() definitely is. It might be worth checking what the behavior is for the former.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -281,14 +282,24 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+          if (maybeDoc instanceof Document && query.matches(maybeDoc)) {
+            results = results.insert(maybeDoc.key, maybeDoc);
+          }
+          resolve();
+        });
+        promises.push(decodeDoc);
       })
-      .next(() => results);
+      .next(() => {
+        Promise.all(promises);
```
schmidt-sebastian: You need to wait for the result here.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -270,6 +270,7 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       iterationOptions.index = DbRemoteDocument.collectionReadTimeIndex;
     }
 
+    let promises : Promise<void>[] = [];
```
schmidt-sebastian: s/let/const

packages/firestore/src/local/indexeddb_remote_document_cache.ts
wu-hui: Done.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
wu-hui: Done.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -281,14 +282,25 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
...
+        });
+        promises.push(decodeDoc);
       })
-      .next(() => results);
+      .next(async () => {
+        await Promise.all(promises);
+      })
+      .next(() => {
+        return results;
+      });
```
schmidt-sebastian: This could be condensed to:

```
.next(() => promise.all(promises))
.next(() => results);
```

or:
```
.next(async () => {
   await Promise.all(promises);
   return results;
});
```

PR #2534, opened 2020-01-14
Fix for #1491: Disable use of $httpHeaders in browser extensions too.

mikelehen:
This is a follow-up to https://github.com/firebase/firebase-js-sdk/pull/2464.

@wu-hui Can you test this with your repro extension?
@wilhuff Can you review?
@hsubox76 Can you review the util changes?

### DIFF
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -75,8 +75,8 @@ export function isBrowser(): boolean {
 /**
  * Detect browser extensions (Chrome and Firefox at least).
  */
-declare var chrome: any;
-declare var browser: any;
+declare const chrome: { runtime?: unknown };
+declare const browser: { runtime?: unknown };
 export function isBrowserExtension(): boolean {
   return (
     (typeof chrome === 'object' && chrome.runtime !== undefined) ||
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -73,6 +73,18 @@ export function isBrowser(): boolean {
 }
 
 /**
+ * Detect browser extensions (Chrome and Firefox at least).
+ */
+declare var chrome: any;
+declare var browser: any;
+export function isBrowserExtension(): boolean {
+  return (
+    (typeof chrome === 'object' && chrome.runtime !== undefined) ||
+    (typeof browser === 'object' && browser.runtime !== undefined)
+  );
+}
+
+/**
  * Detect React Native.
  *
  * @return true if ReactNative environment is detected.
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -75,13 +75,19 @@ export function isBrowser(): boolean {
 /**
  * Detect browser extensions (Chrome and Firefox at least).
  */
-declare const chrome: { runtime?: unknown };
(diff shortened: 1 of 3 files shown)


wu-hui:
I still see permission issue after applying this PR. 

FWIW, `isBrowserExtension` does return `true` correctly when run from extension.

wu-hui:
> I still see permission issue after applying this PR.
> 
> FWIW, `isBrowserExtension` does return `true` correctly when run from extension.

Turns out I did not signin first..after sign in it worked fine.

mikelehen:
FYI- Based on https://github.com/firebase/firebase-js-sdk/issues/1491#issuecomment-574172970 I tweaked the isBrowserExtension() implementation...  I'll try to verify it a bit better today.

packages/util/src/environment.ts
```diff
@@ -72,6 +72,18 @@ export function isBrowser(): boolean {
...
 }
 
+/**
+ * Detect browser extensions (Chrome and Firefox at least).
+ */
+declare const chrome: { runtime?: unknown };
+declare const browser: { runtime?: unknown };
+export function isBrowserExtension(): boolean {
+  return (
+    (typeof chrome === 'object' && chrome.runtime !== undefined) ||
```
hsubox76: Just tested this condition in Chrome console on a random webpage (not in an extension) and it evaluates true, `chrome.runtime` exists.  I read somewhere that if there are any extensions on the page, even if you are not in the extension's code, this will happen? I guess it needs to be available on the page to pass messages from the content script to the background script.  [This suggests](https://stackoverflow.com/questions/7507277/detecting-if-code-is-being-run-as-a-chrome-extension) checking for `chrome.runtime.id` which is indeed undefined in my browser.

hsubox76: Oh oops, saw you changed it already!

mikelehen: Yup.  Just tested the new version in and out of an extension on Chrome and Firefox, and it seems to work.  Sorry for the earlier mistake!

mikelehen:
Verified this returns the correct value in and out of a Chrome or Firefox extension.

packages/util/src/environment.ts
```diff
@@ -72,6 +72,24 @@ export function isBrowser(): boolean {
...
+/**
+ * Detect browser extensions (Chrome and Firefox at least).
+ */
+interface BrowserRuntime {
+  id?: unknown;
+}
+declare const chrome: { runtime?: BrowserRuntime };
+declare const browser: { runtime?: BrowserRuntime };
+export function isBrowserExtension(): boolean {
+  const runtime =
```
wilhuff: It looks like we're using TypeScript >= 3.7. Can't we use optional chaining? Something like this would be far more readable:

```ts
return chrome?.runtime?.id !== undefined || browser?.runtime?.id !== undefined;
```

mikelehen: I think TypeScript's optional chaining is less defensive than my code (it just does null/undefined checks instead of verifying it's of type object), and I'd like to be as defensive as possible to e.g. deal with the case that the developer just happens to have a `browser` variable but it's a `string` instead of `object` or whatever.

PR #2544, opened 2020-01-15
Move ignoreIfPrimaryLeaseLoss to local_store.ts

schmidt-sebastian:
This breaks the dependency of SyncEngine and RemoteStore on indexeddb_persistence.ts and makes it easier to have a clean separation between 'yet-to-be-named-component' and 'firebase/firestore/persistence'.


schmidt-sebastian:
@wilhuff Can you re-approve?

schmidt-sebastian:
> LGTM as-is.
> 
> However, these pieces now seem a little scattered around. The error message is in persistence.ts, and the function is in local_store.ts. Would it make sense to make a lease.ts and consolidate everything we can over there? For example there's also `PRIMARY_LEASE_EXCLUSIVE_ERROR_MSG`, which you didn't need to move but could.

`ignorePrimaryLeaseIfLost` is only ever used on the return value of LocalStore functions, so I thought it might be ok to move it to LocalStore. The error message however is very specific and to the Persistence API. I know this is not ideal, but the current state is the compromise that I came up with.

On top of that, I merged this before I saw your comment :) So for now I will leave as is.

PR #2546, opened 2020-01-16
Switch CI to Github Actions

hsubox76:
Switching CI from Travis to GitHub actions for slightly better speed and more reliability.

Has same functionality as Travis config, separated into 3 files:
- Standard tests (Node/Chrome) on push and PR
- Saucelabs cross-browser tests on push
- NPM publish canary build on push to master (PR merges)


.github/workflows/node-chrome-test.yml
```diff
@@ -0,0 +1,37 @@
+name: Main Test Flow
+
+on:
+  push:
+    branches:
+      - '*'
+      - '*/*'
+      - '!master'
```
hiranya911: You should probably include the master branch to be in safe side. Otherwise the merged state of some PRs will go untested.

.github/workflows/deploy-canary.yml
```diff
@@ -0,0 +1,31 @@
+name: Canary Release
+
+on:
+  push:
+    branches:
+      - master
+
```
hiranya911: Nit: Redundant whitespace

.github/workflows/deploy-canary.yml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        node-version: 10.x
+        registry-url: 'https://registry.npmjs.org/'
+    - name: Yarn install
+      run: yarn
+    - name: yarn build
+      run: yarn build
+    - name: Deploy canary
+      run: |
+        npm config set //registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN
```
hiranya911: Doesn't this need to go through wombot?

.github/workflows/cross-browser-test.yml
```diff
@@ -0,0 +1,34 @@
...
+      - '*/*'
+      - '!master'
+  pull_request:
+
+jobs:
+  cross-browser-test:
+    name: Cross-Browser (Saucelabs) Tests
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+    env:
```
hiranya911: I assume these have already been defined in the settings?

.github/workflows/deploy-canary.yml
hsubox76: It should eventually but the Travis config did not use it. Trying to change the minimum number of things at first in the move.

.github/workflows/cross-browser-test.yml
hsubox76: Yes, the env variables are in the repo's secrets already.

.github/workflows/deploy-canary.yml
hsubox76: Fixed.

.github/workflows/node-chrome-test.yml
hsubox76: Well, that makes the config line simpler.

.github/workflows/deploy-canary.yml
```diff
@@ -0,0 +1,30 @@
+name: Canary Release
+
+on:
+  push:
```
hiranya911: Does this really have to be on push? That means a release commit like this can also trigger it: https://github.com/firebase/firebase-js-sdk/commit/e50a4e906a135e12e582a53574abb4ab66c44521

I think you probably need something a bit more advanced like:

```
on:
  pull_request:
    branches:
      - master
    types:
      - closed

jobs:
  deploy:
    if: github.event.pull_request.merged
```

.github/workflows/cross-browser-test.yml
```diff
@@ -0,0 +1,34 @@
+name: Cross-Browser Test Flow
+
+on:
+  push:
```
hiranya911: You probably want `[push, pull_request]` here as well.

.github/workflows/deploy-canary.yml
hsubox76: Was copying the Travis config but this seems better, changed.

.github/workflows/cross-browser-test.yml
hsubox76: Fixed.

hsubox76:
Going to try to add wombot npm publish.

.github/workflows/cross-browser-test.yml
hsubox76: Oh wait, I think we only had it run on push because we have Saucelabs concurrency limitations right now that we are working to get fixed.  I'm switching it back to push.

PR #2548, opened 2020-01-16
Use approximate FieldValue size for MemoryLRU

schmidt-sebastian:


packages/firestore/src/model/field_value.ts
```diff
@@ -126,6 +126,12 @@ export abstract class FieldValue {
   abstract isEqual(other: FieldValue): boolean;
   abstract compareTo(other: FieldValue): number;
 
+  /**
+   * Returns an approximate (and wildly inaccurate) in-memory size for the field
+   * value.
+   */
+  abstract byteSize(): number;
```
mikelehen: Consider renaming `approximateByteSize()` ?

packages/firestore/src/model/field_value.ts
```diff
@@ -167,6 +173,10 @@ export class NullValue extends FieldValue {
     return this.defaultCompareTo(other);
   }
 
+  byteSize(): number {
+    return 1;
```
mikelehen: It's not clear to me why you chose `1`.  I'm not saying it's a bad value, but that it might be good for us to be deliberate about our approximations and document them to make it clearer the intention and allow some future person to improve them.  

It's tempting to try to be consistent with the old code... in which case, would this use `4` since the JSON-encoding is `null` which is 4 bytes?

Alternatively, if this is really meant to represent the in-memory representation, I would _guess_ (but could be wrong that at minimum, JavaScript stores values at 4-byte-aligned locations, so the minimum size of anything is probably 4 bytes.

packages/firestore/src/model/field_value.ts
```diff
@@ -221,6 +235,10 @@ export abstract class NumberValue extends FieldValue {
     }
     return this.defaultCompareTo(other);
   }
+
+  byteSize(): number {
+    return 4;
```
mikelehen: This should probably be 8?  (double / long are 8)

packages/firestore/src/model/field_value.ts
```diff
@@ -313,6 +331,13 @@ export class StringValue extends FieldValue {
     }
     return this.defaultCompareTo(other);
   }
+
+  byteSize(): number {
+    // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures:
+    // "JavaScript's String type is [...] a set of elements of 16-bit unsigned
+    // integer values"
+    return this.internalValue.length * 2;
```
mikelehen: Note that if we're going for the on-disk size, this will probably (not 100% sure about IndexedDb) be UTF-8 encoded which means most characters will be 1-byte, not 2.

packages/firestore/src/model/field_value.ts
```diff
@@ -436,6 +473,10 @@ export class BlobValue extends FieldValue {
     }
     return this.defaultCompareTo(other);
   }
+
+  byteSize(): number {
+    return this.internalValue.toUint8Array().byteLength;
```
mikelehen: All of the other calculations are cheap...  should we add a byteSize to Blob?

packages/firestore/src/model/field_value.ts
```diff
@@ -492,6 +541,11 @@ export class GeoPointValue extends FieldValue {
     }
     return this.defaultCompareTo(other);
   }
+
+  byteSize(): number {
+    // GeoPoints are made up of two distinct numbers (latitude + longitude)
+    return 8;
```
mikelehen: Should perhaps be 16 then, I think... although JavaScript may use some clever encoding of numbers that allows it to use 4-bytes for small numbers?

packages/firestore/src/model/field_value.ts
```diff
@@ -634,6 +688,14 @@ export class ObjectValue extends FieldValue {
     return FieldMask.fromSet(fields);
   }
 
+  byteSize(): number {
+    let size = 0;
+    this.internalValue.inorderTraversal((key, val) => {
+      size += key.length + val.byteSize();
```
mikelehen: For consistency with your other code, this would be 2* key.length, right?  But again, if we assume UTF-8 encoding, then this is okay.  

FWIW it is tempting to create a `Sizer` helper class that consistently deals with data types.  E.g. `Sizer.string(key.length) + val.byteSize()`

packages/firestore/src/model/field_value.ts
```diff
@@ -720,6 +782,13 @@ export class ArrayValue extends FieldValue {
     }
   }
 
+  byteSize(): number {
+    return this.internalValue.reduce(
+      (previousSize, value) => previousSize + value.byteSize(),
```
mikelehen: Would previousSize => totalSize or accumulatedSize be an improvement?

packages/firestore/test/unit/model/field_value.test.ts
```diff
@@ -485,4 +486,57 @@ describe('FieldValue', () => {
...
+      const expectedItemSize = group[0].byteSize();
+      for (const element of group) {
+        expect(element.byteSize()).to.equal(expectedItemSize);
+      }
+    }
+  });
+
+  it('estimates size correctly for relatively sized values', () => {
+    // This test verifies for each group that the estimated size increases
+    // as the size of the underlying data grows.
```
mikelehen: Clever!

packages/firestore/test/unit/model/field_value.test.ts
```diff
@@ -485,4 +486,57 @@ describe('FieldValue', () => {
...
+      [
+        new fieldValue.ServerTimestampValue(Timestamp.fromMillis(100), null),
+        new fieldValue.ServerTimestampValue(Timestamp.now(), null)
+      ]
+    ];
+
+    for (const group of equalityGroups) {
+      const expectedItemSize = group[0].byteSize();
+      for (const element of group) {
+        expect(element.byteSize()).to.equal(expectedItemSize);
```
wilhuff: Is it worth verifying/possible to verify that no size is zero?

I'm a little worried that these tests don't actually verify that we're aggregating the sizes of objects correctly.

packages/firestore/test/unit/model/field_value.test.ts
```diff
@@ -485,4 +486,57 @@ describe('FieldValue', () => {
...
+    const equalityGroups = [
+      [wrap(null), wrap(false), wrap(true)],
+      [wrap(blob(0, 1)), wrap(blob(128, 129))],
+      [wrap(NaN), wrap(Infinity), wrap(1), wrap(1.1)],
+      [wrap(new GeoPoint(0, 0)), wrap(new GeoPoint(0, 0))],
+      [wrap(Timestamp.fromMillis(100)), wrap(Timestamp.now())],
+      [
+        new fieldValue.ServerTimestampValue(Timestamp.fromMillis(100), null),
+        new fieldValue.ServerTimestampValue(Timestamp.now(), null)
+      ]
```
wilhuff: Missing coverage for strings, arrays, and reference values.

packages/firestore/test/unit/model/field_value.test.ts
```diff
@@ -485,4 +486,57 @@ describe('FieldValue', () => {
       }
     );
   });
+
+  it('estimates size correctly for fixed sized values', () => {
```
wilhuff: These tests are pretty clever in that they don't encode the actual sizes, but they don't directly verify the implementation. It would be more straightforward if the tests just listed out:

```ts
function sizeof(value) {
  return wrap(value).approximateByteSize();
}

it('estimates size correctly for fixed sized values', () => {
  expect(sizeof(null)).to.be(4);
  expect(sizeof('')).to.be(0);
  // etc ...

});

packages/firestore/test/unit/model/field_value.test.ts
schmidt-sebastian: This should now be taken care of by explicitly comparing the size to a provided value. I also added more types to this test case.

PR #2560, opened 2020-01-22
Exclude packages that fail cross-browser tests

hsubox76:
We're in the process of finding a better cross-browser testing solution but in the meantime, the Saucelabs tests are a guaranteed failing check and not very useful.  Excluding the packages that need their tests fixed will at least allow test failures to mean something, and not create an alarming failed check most of the time.


hsubox76:
Tests were successful but exiting with code 134, need to look into it further.

hsubox76:
Put this on hold until the rest of the new CI process is solid.

PR #2565, opened 2020-01-22
Run tests only on changed packages on PR/push

hsubox76:
Trying to run tests only on packages with changed files on PR/push.  Entire test suite will continue to run on merge, and twice a day.

Cross-browser tests moved to run only on merge until I can clean up our Saucelabs workflow to make results viewable and useful.


scripts/run_changed.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc.
```
hiranya911: Nit: 2020

scripts/run_changed.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+  await runTests(Object.keys(changedPackages));
+}
+
+/**
+ * Runs `yarn test` in all dirs in pathList.
+ * @param {Array<string>} pathList
+ */
+async function runTests(pathList) {
+  if (!pathList) return;
```
hiranya911: May be log this case: `No changes detected in any package`

scripts/run_changed.js
hsubox76: Thanks, also fixed this in the template that auto-adds the license headers.

scripts/run_changed.js
hsubox76: Good idea, added it to `runTestsOnChangedPackages` since this function is also used to run the integration tests.

PR #2566, opened 2020-01-23
Remove instanceof IndexedDbRemoteDocumentCache check

schmidt-sebastian:
Goal: remove `instanceof IndexedDbRemoteDocumentCache` check in LocalStore to allow me to drop the import.

Realized: We don't even use the document result from getLastDocumentChange. So I changed it to only return the read time.

### DIFF
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -291,11 +291,6 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       .next(() => results);
   }
 
-  /**
-   * Returns the set of documents that have been updated since the specified read
-   * time.
-   */
-  // PORTING NOTE: This is only used for multi-tab synchronization.
   getNewDocumentChanges(
     transaction: PersistenceTransaction,
     sinceReadTime: SnapshotVersion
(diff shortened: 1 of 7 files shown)


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -291,11 +291,6 @@ export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
       .next(() => results);
   }
 
-  /**
-   * Returns the set of documents that have been updated since the specified read
-   * time.
-   */
-  // PORTING NOTE: This is only used for multi-tab synchronization.
```
wilhuff: Drive by comment: don't the comments still apply? (i.e. isn't this still only used for multi-tab?)

schmidt-sebastian: The comment is already on the interface declaration.

schmidt-sebastian:
@var-const Do you have cycles for this today?

packages/firestore/src/local/remote_document_cache.ts
```diff
@@ -95,6 +95,15 @@ export interface RemoteDocumentCache {
     readTime: SnapshotVersion;
   }>;
 
+  /**
+   * Returns the read time of the most recently read document in the cache, or
+   * SnapshotVersion.MIN if not available.
+   */
+  // PORTING NOTE: This is only used for multi-tab synchronization.
+  getLastReadTime(
```
var-const: Question: why is this declaration here, whereas previously `getLastDocumentChange` wasn't?

schmidt-sebastian: It allows me to use this method in Persistence without checking `instanceof IndexedDbRemoteDocumentCache`

schmidt-sebastian:
Changed the PR title to "Remove instanceof IndexedDbRemoteDocumentCache check" to be more in line with the intention of the PR. Thanks for the review!

schmidt-sebastian:
Changed the PR title to "Remove instanceof IndexedDbRemoteDocumentCache check" to be more in line with the intention of the PR. Thanks for the review!

PR #2576, opened 2020-01-27
CI test workflow improvements

hsubox76:
Run all tests when files that affect the repo globally are changed, such as top-level dependencies and test/compile/lint config files.

Edit: Also now that some PRs only run partial tests, ensure canary deploy only runs after successful test run. `if` should ensure it only runs once on each PR merge and not after every full test run (such as the scheduled ones).


scripts/run_changed.js
```diff
@@ -17,19 +17,55 @@
...
+ * Identify modified packages that require tests.
+ */
+async function getChangedPackages() {
   const diff = await git.diff(['--name-only', 'origin/master...HEAD']);
   const changedFiles = diff.split('\n');
   const changedPackages = {};
   for (const filename of changedFiles) {
+    if (fullTestTriggerFiles.includes(filename)) {
+      console.log(
+        chalk`{blue Running all tests because ${filename} was modified.}`
```
hiranya911: I'm not familiar with this syntax. Hope it's correct.

scripts/run_changed.js
```diff
@@ -52,7 +89,7 @@ async function runTestsOnChangedPackages() {
  * @param {Array<string>} pathList
  */
 async function runTests(pathList) {
-  if (!pathList) return;
+  if (!pathList || pathList.length === 0) return;
```
hiranya911: Kind of redundant.

scripts/run_changed.js
hsubox76: Seems to work when run.

scripts/run_changed.js
hsubox76: Removed.

Issue #2555, opened 2020-01-20
IINTERNAL ASSERTION FAILED: AsyncQueue is already failed: FirebaseError: The current tab is not in the required state to perform this operation. It might be necessary to refresh the browser tab.


burtonator:
### [REQUIRED] Describe your environment

  * Operating System version: __ ubuntu
  * Browser version: _ electron 7.1.2
  * Firebase SDK version: __7.6.1
  * Firebase Product: __ firestore (auth, database, storage, etc)

#### Steps to reproduce:

We're getting this exception constantly in the wild, but I can't actually duplicate it locally.  

We have about 5k uses and receive this error about 9200 times per month.

#### Relevant Code:

I don't have a reduction unfortunately.  Is there a better way to reproduce this?

rommelpe:
Sorry to hear about your issues, @burtonator. I haven't encountered this error yet. Based from the error message, it's more likely related to (https://github.com/firebase/firebase-js-sdk/issues/1436) multi-tab synchronization support, and was addressed in v5.7.2.

 - Are you using `experimentalTabSynchronization`?
- Can you try running the app on multiple tabs to reproduce it (as mentioned in the issue)?

burtonator:
Thanks @rommelpe ... appreciate your help as this might be a huge bug for us.

> Are you using experimentalTabSynchronization?

I'm not but I am running synchronizeTabs.  

> Can you try running the app on multiple tabs to reproduce it (as mentioned in the issue)?

Electron doesn't really have 'tabs' but it does have windows.  I've tried to duplicate this and I can't ... I might try to disable synchronizeTabs.  

What would be the downside of using enablePersistence () without synchronizeTabs?  I assume this means the second 'tab' or window wouldn't have persistence?

thebrianchen:
@burtonator 
> What would be the downside of using enablePersistence () without synchronizeTabs? I assume this means the second 'tab' or window wouldn't have persistence?

Without `synchronizeTabs`, only the first tab that is opened will have persistence, so subsequently opened tabs will fall back to memory persistence. In web browsers, multiple tabs share a single instance of IndexedDB, and the `synchronizeTabs` setting allows multiple Firestore instances to share access to IndexedDB.

cmditch:
We've also been seeing this error on and off for several months. We are **not** on electron, just your typical web app.

We are using `firebase.firestore().enablePersistence({ synchronizeTabs:true });`

[...]

PR #2585, opened 2020-01-31
Ignore primary lease loss in maybeGarbageCollectMultiClientState

schmidt-sebastian:
This ignores the error if the primary lease is somehow lost between updateClientMetadataAndTryBecomePrimary() and maybeGarbageCollectMultiClientState(). There is no special recovery logic needed as the next primary will run the periodic GC.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2555

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -72,6 +72,7 @@ import { ReferenceSet } from './reference_set';
 import { ClientId } from './shared_client_state';
 import { TargetData } from './target_data';
 import { SimpleDb, SimpleDbStore, SimpleDbTransaction } from './simple_db';
+import {ignoreIfPrimaryLeaseLoss} from "./local_store";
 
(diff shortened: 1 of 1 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -489,7 +488,12 @@ export class IndexedDbPersistence implements Persistence {
             ).next(() => inactive);
           });
         }
-      );
+      ).catch(() => {
+        // Ignore primary lease violations or any other type of error.
```
wu-hui: I think it's worth adding why it's OK to ignore all the errors here. 

Something along the line where you put in PR description: `There is no special recovery logic needed as the next primary will run the periodic GC.`

schmidt-sebastian: Done.

schmidt-sebastian:
@wilhuff for approval

PR #2615, opened 2020-02-10
Improve logic for detecting what PR tests to run

hsubox76:
This adds detection of other tests that should run:
1) tests in packages dependent on the changed package(s)
2) special cases - firestore tests should run on firestore emulator file setup changes

### DIFF
--- a/scripts/run_changed.js
+++ b/scripts/run_changed.js
@@ -94,9 +94,11 @@ async function getChangedPackages() {
     const match = filename.match('^(packages/[a-zA-Z0-9-]+)/.*');
     if (match && match[1]) {
       const changedPackage = require(resolve(root, match[1], 'package.json'));
-      if (changedPackage && changedPackage.scripts.test) {
-        // Add the package itself.
-        changedPackages[match[1]] = 'direct';
+      if (changedPackage) {
+        if (changedPackage.scripts.test) {
+          // Add the package itself.
+          changedPackages[match[1]] = 'direct';
+        }
         // Add packages that depend on it.
         for (const package in depGraph) {
           if (depGraph[package].includes(changedPackage.name)) {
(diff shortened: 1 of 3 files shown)


scripts/run_changed.js
```diff
@@ -52,35 +52,82 @@ const alwaysRunTestPaths = [
...
+  'scripts/emulator-testing/firestore-test-runner.ts': ['packages/firestore'],
+  'scripts/emulator-testing/database-test-runner.ts': ['packages/database']
+};
+
 /**
  * Identify modified packages that require tests.
  */
 async function getChangedPackages() {
+  const packageInfo = JSON.parse(
+    (await exec('npx lerna ls --json', { cwd: root })).stdout
```
Feiyang1: Can use the already installed lerna at "node_modules/.bin/lerna" to remove the networking.

scripts/run_changed.js
```diff
@@ -52,35 +52,82 @@ const alwaysRunTestPaths = [
...
     if (match && match[1]) {
-      const pkg = require(resolve(root, match[1], 'package.json'));
-      if (pkg && pkg.scripts.test) {
-        changedPackages[match[1]] = true;
+      const changedPackage = require(resolve(root, match[1], 'package.json'));
+      if (changedPackage && changedPackage.scripts.test) {
+        // Add the package itself.
+        changedPackages[match[1]] = 'direct';
+        // Add packages that depend on it.
+        for (const package in depGraph) {
```
Feiyang1: Shouldn't we run tests for depending packages even if `changedPackage.scripts.test` evaluates to false? something like:
```
if (changedPackage) {
  if (changedPackage.scripts.test) {
      changedPackages[match[1]] = 'direct';
  }

  for (const package in depGraph) {
  ...
  }
}
```

scripts/run_changed.js
hsubox76: npx will run the local installed version if available.

scripts/run_changed.js
hsubox76: Yes, fixed.

Issue #2620, opened 2020-02-12
Severe performance regression in `onSnapshot` listeners in Cloud Firestore


djsweet:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10 Version 1909
  * Browser version: Google Chrome Version 80.0.3987.87 (Official Build) (64-bit)
  * Firebase SDK version: 7.8.1, since 7.3.0
  * Firebase Product: firestore (auth, database, storage, etc)

### [REQUIRED] Describe the problem

SDK 7.3.0 introduced a severe performance regression in the Firestore `onSnapshot` listener mechanism.

We listen to up to 3000 writes per second written in batches of 50 per transaction. As of SDK 7.2.3, each round of 3000 writes takes approximately 1 second, but often blocks the event loop for up to the entire second. Starting with SDK 7.3.0, this same amount of writes takes upwards of 30 seconds, also often blocking the event loop for the entirety of the duration. Our users were complaining that their browsers were recommending that they close their tabs due to unresponsiveness.

I have prepared [a sample project](https://github.com/djsweet/firestore-looplock-bug-report) to reproduce this problem. Here are some relevant screenshots with simple measurements. In the report columns, "RAF" is used as a placeholder for `requestAnimationFrame`. The sample project uses `window.requestAnimationFrame` to determine the severity of event loop blocking.

#### SDK 7.2.3

![Firebase 7 2 3 Update Times](https://user-images.githubusercontent.com/9657980/74306575-c1d88300-4d17-11ea-9bb6-8c66f4b14995.PNG)

#### SDK 7.3.0

![Firebase 7 3 0 Update Times](https://user-images.githubusercontent.com/9657980/74306579-c735cd80-4d17-11ea-959b-295a8d7be359.PNG)

#### SDK 7.8.0

![Firebase 7 8 0 Update Times](https://user-images.githubusercontent.com/9657980/74306583-cbfa8180-4d17-11ea-8268-0825c85471b3.PNG)

#### SDK 7.8.1

![Firebase 7 8 1 Update Times](https://user-images.githubusercontent.com/9657980/74306588-d157cc00-4d17-11ea-9fd2-f41915c8fdee.PNG)

It seems that there have been improvements since SDK 7.3.0, but stalling the event loop for 25 seconds as of SDK 7.8.1 is still unacceptable.

[...]

PR #2624, opened 2020-02-12
Add optimization to SortedSet.unionWith

wu-hui:
This is a fix to https://github.com/firebase/firebase-js-sdk/issues/2620

Basically `SortedSet.unionWith` does not take the size of the two sets that it is merging into consideration.

What happens is, if there are two set: A that is empty, and B that has size of `N`. The union would take `O(N*logN)` while it could have been a no-op.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
-- [fixed] Fixed a performance regression introduced in 1.7.0. (#2620)
+- [fixed] Fixed a performance regression introduced by the addition of
+  `Query.limitToLast(n: number)` in Firestore 1.7.0 (Firebase 7.3.0) (#2620).
 - [fixed] Fixed an issue where `CollectionReference.add()` would reject
   custom types when using `withConverter()`. (#2606)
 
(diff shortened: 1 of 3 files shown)


packages/testing/src/api/index.ts
```diff
@@ -134,7 +134,7 @@ function initializeApp(
         ({
           getToken: async () => ({ accessToken: accessToken }),
           getUid: () => null,
-          addAuthTokenListener: (listener) => {
+          addAuthTokenListener: listener => {
```
yuchenshi: I don't see why we need this change, but if it's an auto-formatting or linter thing, then you have my approval.

packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- [fixed] Fixed a performance regression introduced in 1.7.0. (#2620)
```
wilhuff: The public is largely unaware of the Firestore package's version (and our public release notes never mention them). I suggest tying this to "Firebase 7.3.0" somehow. You may also want to tie this to limitToLast to give users more of an anchor on exactly which version this was.

Maybe something like this:

> Fixed a performance regression introduced by the addition of `Query.limitToLast(n: number)` in Firestore 1.7.0 (Firebase 7.3.0) (#2620).

?

wu-hui: Done.

packages/testing/src/api/index.ts
wu-hui: Yeah, this is from auto-formatting.

Issue #2590, opened 2020-02-03
FCM Messaging breaks in version 7.0.0 or after


basvandenheuvel:
* Operating System version: Windows 10
  * Browser version: Chrome 79 
  * Firebase SDK version: 7.0.0+
  * Firebase Product: Messaging

#### Problem
Currently we are using Firebase SDK 6.5.0. Whenever we update to version 7.0.0 or later (we tried multiple versions including 7.0.0 and 7.8.0) Messaging stops working.
It looks like the background messaging keeps working when the website does not have focus. However when the website does have focus the FCM messaging does not work.

Ofcourse we followed the changelog and enabled the FCM Messaging API.

#### Relevant Code:
 The following code is used in version 6.5.0 and works fine, but breaks in 7.0.0 or after:

```
// Imports
import * as firebase from 'firebase/app';
import 'firebase/messaging';
```

```
// Messaging
this.messaging = firebase.messaging();
```

```
// Subscribing to messages 
this.messaging.onMessage((result: any) => {
    console.log(result);
});
```

rommelpe:
Thanks for reporting the issue, @basvandenheuvel. I can't seem to replicate it. 

- Were you able to reproduce it using [quickstart](https://github.com/firebase/quickstart-js/tree/master/messaging)?
- Have you noticed any significant logs from browser's DevTool?
- Could you share a [working repro](https://stackoverflow.com/help/minimal-reproducible-example) that we could run locally?

basvandenheuvel:
Hi @rommelpe , thanks for your reply.

1. I just tired the quickstart, also with the quickstart i do not get messages when the website has focus... i do get the messages in the background.

2. Haven't seen any logs.

3. I do not have anything for now...

rommelpe:
Sorry for the confusion, @basvandenheuvel. It's an intended behavior. If the page is open and focused, you will not see a notification and as a result, you'll receive a callback in onMessage in the page.

If the page is in the background (i.e. not focused) the background message handler will be called and will show a notification.

See [here](https://firebase.google.com/docs/cloud-messaging/js/receive):

> Firebase notifications behave differently depending on the foreground/background state of the receiving app. If you want foregrounded apps to receive notification messages or data messages, you’ll need to write code to handle the onMessageReceived callback. For an explanation of the difference between notification and data messages, see Message types.

basvandenheuvel:
@rommelpe 

I think i introduced the confusion.

[...]

PR #2637, opened 2020-02-14
Fix v7.6.2 sw response compatibility with v7.8.1 client

rommelpe:
https://github.com/firebase/firebase-js-sdk/issues/2590#issuecomment-585863435 We can reproduce it, the `sw` won't get the updated version (7.8.1) and retained the old ones (7.6.2). This is just a temporary fix while @Feiyang1 is looking for a way to update the `sw` syncly upon refreshing the page.


Feiyang1:
Thanks for the contribution! We will try https://github.com/firebase/firebase-js-sdk/pull/2638 and see if it fixes the issue.

PR #2638, opened 2020-02-15
reliably update service worker

Feiyang1:
fixes https://github.com/firebase/firebase-js-sdk/issues/2590
It should fix any sw update issue going forward.


packages/messaging/src/controllers/window-controller.ts
```diff
@@ -167,6 +167,12 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
             scope: DEFAULT_SW_SCOPE
           }
         );
+        
+        // The timing when browser updates sw when sw has an update is unreliable by my experiment.
```
mmermerkaya: > by my experiment.

This bit won't really matter to a reader in the future :smile:

PR #2639, opened 2020-02-15
Change `ProtoByteString` to use its own class rather than `String | Uint8Array`

thebrianchen:
We've been storing tokens internally as `ProtoByteString`, which is typed to `String | Uint8Array` in order to account for the fact that protos are sent as base64 strings and Uint8Arrays. However, casting between these 2 types is confusing and bug-prone, resulting in a latent bug where we were improperly decoding Uint8Arrays as storing them as strings. This PR separates creates a separate `ProtoByteString` class to provide a type-safe representation for tokens.


packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * @license
+ * Copyright 2020 Google Inc.
```
wilhuff: This should be `Copyright 2020 Google LLC`. See go/copyright.

However, see below--this file may not need to be added.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export function emptyByteString(): ProtoByteString {
+  return Blob.fromBase64String('');
+}
+
+/**
+ * Returns the representation of a "proto" byte string (in base64) for the
+ * platform from the given hexadecimal string.
+ */
+export function byteStringFromString(value: string): ProtoByteString {
```
wilhuff: `byteStringFromString` seems to only be used in tests. Move it to test/util/helpers.ts?

There's already a `blob` helper there too: https://github.com/firebase/firebase-js-sdk/blob/eaeed94a3092e81343129d95726a3d5fe173e57d/packages/firestore/test/util/helpers.ts#L198. Consider making this just an overload of that?

packages/firestore/src/core/types.ts
```diff
@@ -29,9 +31,7 @@ export type TargetId = number;
 
 export type ListenSequenceNumber = number;
 
-// TODO(b/35918695): In GRPC / node, tokens are Uint8Array. In WebChannel,
-// they're strings. We should probably (de-)serialize to a common internal type.
-export type ProtoByteString = Uint8Array | string;
+export type ProtoByteString = Blob;
```
wilhuff: This type declaration doesn't seem to be adding much. If anything it just adds to confusion.

Previously this existed as a separate type to indicate that it was this funky either-or type, but now it seems like it's mostly just a wart. There's no distance between these types and it's confusing to know when to pick which (as your PR shows).

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { ProtoByteString } from '../core/types';
+import { PlatformSupport } from '../platform/platform';
+import { Blob } from '../api/blob';
+
+/**
+ * Returns the representation of an empty "proto" byte string for the
+ * platform.
+ */
+export function emptyByteString(): ProtoByteString {
```
wilhuff: This seems like it could just a function in `api/blob.ts`.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ProtoByteString } from '../core/types';
+import { PlatformSupport } from '../platform/platform';
+import { Blob } from '../api/blob';
+
+/**
+ * Returns the representation of an empty "proto" byte string for the
```
wilhuff: This comment is inaccurate. This just returns an empty blob. There's nothing platform specific about it.

wilhuff:
Took a first pass.

@schmidt-sebastian See if you disagree with any of the feedback here. On Android we have a distinction between a `ByteString`, which is an internal type representing the bytes (supplied by the protobuf library), and a `Blob`, which is the public type for storing values in Firestore. There's a similar though more complicated distinction on iOS.

I'm a little wary of using the public type internally to represent what is logically a separate `ByteString` type but size concerns make me wary of recommending the extra type. WDYT?

schmidt-sebastian:
> I'm a little wary of using the public type internally to represent what is logically a separate `ByteString` type but size concerns make me wary of recommending the extra type. WDYT?

Blob is a pretty thin wrapper on top of the method exposed by [Platform](https://github.com/firebase/firebase-js-sdk/blob/df1b58816a5f502853f4b6fde59033abca92f3c9/packages/firestore/src/platform/platform.ts). We could address your concern by directly invoking the Platform API and representing Blobs internally as Strings. If we do that, I would advocate to still use the `ProtoByteString` type to keep some level of distinction (similar to what we do with `TargetId`).

packages/firestore/src/core/types.ts
thebrianchen: Removed from types.ts as we now use ProtoByteString as the default representation

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -134,7 +137,7 @@ export class IndexedDbMutationQueue implements MutationQueue {
     transaction: PersistenceTransaction
   ): PersistencePromise<ProtoByteString> {
     return this.getMutationQueueMetadata(transaction).next<ProtoByteString>(
-      metadata => metadata.lastStreamToken
+      metadata => Blob.fromBase64String(metadata.lastStreamToken)
```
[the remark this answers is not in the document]
thebrianchen: Standardized to ProtoByteString

packages/firestore/src/util/proto_byte_string.ts
thebrianchen: done.

packages/firestore/src/util/proto_byte_string.ts
thebrianchen: Now that it is typed to `ProtoByteString`, it no longer has anything to do with Blob.

packages/firestore/src/util/proto_byte_string.ts
thebrianchen: Moved to helpers, but now ProtoByteString is separated from Blob

wilhuff:
I’m definitely in favor of a separate type.

At this point though it’s not really tied to protos so much though. Previously the union type really represented “types that protos will represent byte sequences as”. Now this is independent of that representation. Is suggest renaming to just `ByteString`. Not only is this internally more clearly separate from protos, but also would match iOS and Android, where we also have a `ByteString` type.

packages/firestore/src/local/local_serializer.ts
```diff
@@ -247,18 +245,9 @@ export class LocalSerializer {
...
-      // TODO(b/78771403): Convert tokens to strings during deserialization
-      assert(
-        SimpleDb.isMockPersistence(),
-        'Persisting non-string stream tokens is only supported with mock persistence .'
-      );
-      resumeToken = targetData.resumeToken.toString();
-    } else {
-      resumeToken = targetData.resumeToken;
-    }
+    // We can't store the resumeToken as a ByteString in IndexedDb, so we
```
schmidt-sebastian: This is a much better comment than the ones in mutation_queue. The mutation queue comment explains what we do, but not really why we do it.

packages/firestore/src/local/memory_mutation_queue.ts
```diff
@@ -33,6 +32,7 @@ import { MutationQueue } from './mutation_queue';
 import { PersistenceTransaction, ReferenceDelegate } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { DocReference } from './reference_set';
+import { emptyByteString, ByteString } from '../util/proto_byte_string';
```
schmidt-sebastian: Here and everywhere: Please group your imports.

packages/firestore/src/remote/serializer.ts
```diff
@@ -278,13 +265,35 @@ export class JsonProtoSerializer {
...
+   * Visible for testing.
    */
-  private toBytes(bytes: Blob): string {
+  toBytes(bytes: Blob | ByteString): string {
     if (this.options.useProto3Json) {
       return bytes.toBase64();
     } else {
       // The typings say it's a string, but it needs to be a Uint8Array in Node.
-      return this.unsafeCastProtoByteString(bytes.toUint8Array());
+      // Cast as string to avoid type check failing.
```
schmidt-sebastian: This comment just describes the code below and can be dropped.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,93 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC.
```
schmidt-sebastian: Nit: Drop the period (Google LLC)

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,93 @@
...
+    .call(array, (char: number) => {
+      return String.fromCharCode(char);
+    })
+    .join('');
+}
+
+/**
+ * Helper function to convert a binary string to an Uint8Array.
+ */
+export function Uint8ArrayFromBinaryString(binaryString: string): Uint8Array {
```
schmidt-sebastian: Nit: Please use lower camel case for functions.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,93 @@
...
+    buffer[i] = binaryString.charCodeAt(i);
+  }
+  return buffer;
+}
+
+/**
+ * Returns an empty ByteString.
+ */
+export function emptyByteString(): ByteString {
+  return ByteString.fromBase64String('');
```
schmidt-sebastian: This could be a static member of ByteString (ByteString.EMPTY_BYTE_STRING) which would avoid object allocations.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export function binaryStringFromUint8Array(array: Uint8Array): string {
+  // We can't call array.map directly because it expects the return type to
+  // be a Uint8Array, whereas we can convert it to a regular array by invoking
+  // map on the Array prototype.
+  return Array.prototype.map
+    .call(array, (char: number) => {
+      return String.fromCharCode(char);
+    })
+    .join('');
```
schmidt-sebastian: If you want, this could be:

```return array.reduce((prev, current) => prev + String.fromCharCode(current), "");```

I am not sure whether this is more or less efficient than the current version. The most efficient version is probably:

```
  let result = '';
  for (let i = 0; i < array.length; ++i) {
    result += String.fromCharCode(array[i]);
  }
  return result;
```

File: packages/firestore/test/util/helpers.ts [hunk 14]
```diff
@@ -481,14 +485,23 @@ export function localViewChanges(
   return new LocalViewChanges(targetId, fromCache, addedKeys, removedKeys);
 }
 
+/**
+ * Returns a ByteString representation for the platform from the given hexadecimal
+ * string.
+ */
+export function byteStringFromString(value: string): ByteString {
+  const base64 = PlatformSupport.getPlatform().btoa(value);
```
schmidt-sebastian: Are we sure we are dealing with hex strings here? `btoa` takes a binary string.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1143,13 +1145,17 @@ abstract class TestRunner {
...
+          byteStringFromString(expected.resumeToken)
         )
       );
       expect(actualTarget.query).to.deep.equal(expectedTarget.query);
       expect(actualTarget.targetId).to.equal(expectedTarget.targetId);
       expect(actualTarget.readTime).to.equal(expectedTarget.readTime);
-      expect(actualTarget.resumeToken).to.equal(expectedTarget.resumeToken);
+      // actualTarget's resumeToken is a hexadecimal string, but the serialized
+      // resumeToken will be a base64 string, so we need to convert it back.
+      expect(actualTarget.resumeToken || '').to.equal(
```
schmidt-sebastian: See my comment below - can you verify that we are indeed using hexadecimal strings?

packages/firestore/src/remote/serializer.ts
```diff
@@ -278,13 +265,35 @@ export class JsonProtoSerializer {
...
+  }
+
+  /**
+   * Returns a ByteString based on the proto string value.
+   * DO NOT USE THIS FOR ANYTHING ELSE.
+   * This method cheats. Value is typed as "string" because that's what
+   * our generated proto interfaces say bytes must be, but it is actually
+   * an Uint8Array in Node.
+   */
+  fromBytes(value: string | undefined): ByteString {
```
schmidt-sebastian: This method could be properly typed and doesn't need to correlate to the Proto types (since `string` can be passed to `string | UInt8Array`). So this could just be:

```
  /**
   * Returns a ByteString based on the proto string value.
   */
  fromBytes(value: string | Uint8Array | undefined): ByteString {
    if (this.options.useProto3Json) {
      assert(typeof  value === 'string', "...");
      return ByteString.fromBase64String(value ? value : '');
    } else {
      assert(value === undefined || value instanceof Uint8Array, "...");
      return ByteString.fromUint8Array(
        value ? value: new Uint8Array()
      );
    }
  }
```

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -121,29 +122,33 @@ export class IndexedDbMutationQueue implements MutationQueue {
   acknowledgeBatch(
     transaction: PersistenceTransaction,
     batch: MutationBatch,
-    streamToken: ProtoByteString
+    streamToken: ByteString
   ): PersistencePromise<void> {
     return this.getMutationQueueMetadata(transaction).next(metadata => {
-      metadata.lastStreamToken = convertStreamToken(streamToken);
+      // We can't store the resumeToken as a ByteString in IndexedDB, so we
+      // convert it to a base64 string for storage.
```
schmidt-sebastian: Nit: s/base64/Base64/

Here and below.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -121,29 +122,33 @@ export class IndexedDbMutationQueue implements MutationQueue {
...
   setLastStreamToken(
     transaction: PersistenceTransaction,
-    streamToken: ProtoByteString
+    streamToken: ByteString
   ): PersistencePromise<void> {
     return this.getMutationQueueMetadata(transaction).next(metadata => {
-      metadata.lastStreamToken = convertStreamToken(streamToken);
+      // Convert the streamToken to base64 in order to store it as a string that can
+      // be reconstructed into a ByteString.
+      metadata.lastStreamToken = streamToken.toBase64();
```
schmidt-sebastian: Please use same comment as above.

packages/firestore/src/remote/serializer.ts
```diff
@@ -278,13 +265,37 @@ export class JsonProtoSerializer {
...
+    }
+  }
+
+  /**
+   * Returns a ByteString based on the proto string value.
+   * DO NOT USE THIS FOR ANYTHING ELSE.
+   * This method cheats. Value is typed as "string" because that's what
+   * our generated proto interfaces say bytes must be, but it is actually
+   * an Uint8Array in Node.
+   */
```
schmidt-sebastian: This method doesn't cheat anymore. You can remove line 281-284.

packages/firestore/src/remote/serializer.ts
```diff
@@ -718,8 +729,10 @@ export class JsonProtoSerializer {
         change.targetChange.targetChangeType || 'NO_CHANGE'
       );
       const targetIds: TargetId[] = change.targetChange.targetIds || [];
-      const resumeToken =
-        change.targetChange.resumeToken || this.emptyByteString();
+
+      // Depending on the platform, the resumeToken could be a Uint8Array or
+      // string, even though it is typed as "string".
+      const resumeToken = this.fromBytes(change.targetChange.resumeToken);
```
schmidt-sebastian: Consider removing this comment since `fromBytes` is now properly typed.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,85 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PlatformSupport } from '../platform/platform';
+
+/**
+ * Immutable class that represents a "proto" byte string.
+ *
+ * Proto byte strings can either be base64 strings or Uint8Arrays when sent
```
schmidt-sebastian: s/base64/Base64-encoded/

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,85 @@
...
+ * on the wire, but they are always typed to string. This class abstracts away
```
schmidt-sebastian: I think we can remove ", but they are always typed to string". This seems to not play a major role in this class.

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,85 @@
...
+ * this differentiation by holding the proto byte string in a common class that
+ * must be converted into a string before being sent as a proto.
+ */
+export class ByteString {
+  static readonly EMPTY_BYTE_STRING = new ByteString('');
+
+  private _binaryString: string;
```
schmidt-sebastian: readonly?

packages/firestore/src/util/proto_byte_string.ts
```diff
@@ -0,0 +1,85 @@
...
+    return new ByteString(binaryString);
+  }
+
+  toBase64(): string {
+    return PlatformSupport.getPlatform().btoa(this._binaryString);
+  }
+
+  toUint8Array(): Uint8Array {
+    const buffer = uint8ArrayFromBinaryString(this._binaryString);
+    return buffer;
```
schmidt-sebastian: Inline `buffer`.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -951,18 +952,21 @@ export class SpecBuilder {
         // `query` is not added yet.
         this.activeTargets[targetId] = {
           queries: [SpecBuilder.queryToSpec(query), ...activeQueries],
-          resumeToken: resumeToken || ''
+          // Convert to base64 string to be compatible with ByteString.
```
schmidt-sebastian: "so it can later be parsed into a ByteString"?

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: As discussed offline, these are all base64 encoded strings. Thanks for updating the comment.

PR #2640, opened 2020-02-18
Make Firestore-mangled build the only Firestore build

schmidt-sebastian:
This removes the `.min.js` files for Firestore and updates the default build to use the minified property names. 

It looks like the previous build did not include a mangled Node build - I added this here and because of that I also had to add externs for Node.


hsubox76:
`integration/firestore/gulpfile.js` still referring to `index.esm.min.js` which is causing the test fail I think.

packages/firestore/package.json
```diff
@@ -49,6 +46,7 @@
     "@firebase/app-types": "0.x"
   },
   "devDependencies": {
+    "@types/node": "8.10.59",
```
hsubox76: This is in `devDependencies` in the root `package.json`, is there a reason to include it separately here?

schmidt-sebastian: The paths in `externs.json` are relative to the package root, so I need this to be installed in packages/firestore/node_modules.

hsubox76: Doesn't line 38 of `firestore/rollup.config.js` set them to be relative to repo root? The other files in `externs.json` seem to be relative to repo root.

schmidt-sebastian: Yes! Updated.

schmidt-sebastian:
@Feiyang1 Let me know if you have time to check whether this works properly with SourceMaps.

schmidt-sebastian:
@Feiyang1 confirmed the source maps still work as is. Thanks!

PR #2646, opened 2020-02-19
Remove Closure build

schmidt-sebastian:
I have a feeling this is not used. 

Removing this simplifies my Memory-only Firestore development.


schmidt-sebastian:
> This may be used when exposed in google3. Please double-check.

I checked but also didn't see any references. FWIW, I created a dummy CL that removes it in Google3 and didn't see any issues: https://critique.corp.google.com/#review/296099840

PR #2647, opened 2020-02-19
Cleanup: Use Object.assign() to create shallow copies

schmidt-sebastian:


schmidt-sebastian:
Hmm. This breaks IE :/

Feiyang1:
We don't worry about breaking IE with new ES features. We ask developers to include polyfills for older browsers in https://firebase.google.com/support/guides/environments_js-sdk

Issue #2655, opened 2020-02-20
Error when upgrading from 7.8.2 to 7.9.0


tianhaoz95:
### Describe your environment

  * Operating System version: GitHub Actions ubuntu-latest
  * Browser version: headless
  * Firebase SDK version: 7.9.0
  * Firebase Product: core/firestore

### Describe the problem

The following error was thrown when upgrading firebase from `7.8.2` to `7.9.0` during jest tests with firebase emulator:

```bash
console.error node_modules/@firebase/firestore/node_modules/@firebase/logger/dist/index.cjs.js:94
    [2020-02-20T23:09:42.930Z]  @firebase/firestore: Firestore (7.8.1): FIRESTORE (7.8.1) INTERNAL ASSERTION FAILED: AsyncQueue is already failed: TypeError: __PRIVATE_protoLoader.HP is not a function
        at __PRIVATE_loadProtos (/home/runner/work/iwfp/iwfp/iwfpapi/functions/node_modules/@firebase/firestore/src/platform_node/load_protos.ts:48:29)
        at t.platform [as SE] (/home/runner/work/iwfp/iwfp/iwfpapi/functions/node_modules/@firebase/firestore/src/platform_node/node_platform.ts:47:20)
        at t._this [as bE] (/home/runner/work/iwfp/iwfp/iwfpapi/functions/node_modules/@firebase/firestore/src/core/firestore_client.ts:397:17)
        at /home/runner/work/iwfp/iwfp/iwfpapi/functions/node_modules/@firebase/firestore/src/core/firestore_client.ts:185:31
        at process._tickCallback (internal/process/next_tick.js:68:7)


TypeError: __PRIVATE_protoLoader.HP is not a function

      at __PRIVATE_loadProtos (node_modules/@firebase/firestore/src/platform_node/load_protos.ts:48:29)
      at t.platform [as SE] (node_modules/@firebase/firestore/src/platform_node/node_platform.ts:47:20)
      at t._this [as bE] (node_modules/@firebase/firestore/src/core/firestore_client.ts:397:17)
      at node_modules/@firebase/firestore/src/core/firestore_client.ts:185:31
```

here is a link to the GitHub Actions result page: https://github.com/tianhaoz95/iwfp/pull/330/checks?check_run_id=459036204

#### Steps to reproduce:
1. clone the repository https://github.com/tianhaoz95/iwfp
2. `cd iwfpapi/functions`
3. `npm install && npm run e2e`
4. see it works
[...]

PR #2658, opened 2020-02-21
De-mangle Node and CJS build

schmidt-sebastian:
Removes the mangling from the Node Firestore build, which is probably not as sensitive to size. The actual fix for Node is to just add externs for `GRPC` and `grpcloader`, but at this point I am kind of wary of doing this as we have no integration tests for the mangled Node build.

I also disabled Terser's whitespace and comment removal, since this was previously not applied to NPM builds. 
 
Fixes https://github.com/firebase/firebase-js-sdk/issues/2655

### DIFF
--- a/packages/firestore/rollup.config.js
+++ b/packages/firestore/rollup.config.js
@@ -51,7 +51,8 @@ const transformers = [
 
 const terserOptions = {
   output: {
-    comments: false
+    comments: "all",
+    beautify: true
   },
   mangle: {
     properties: {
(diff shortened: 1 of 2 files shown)


packages/firestore/rollup.config.js
```diff
@@ -96,15 +88,37 @@ const es5Builds = [
       )
   },
   /**
-   * Browser Builds
+   * Browser CJS Build
+   * 
+   * The Browser CJS build is not mangled as Terser's property name mangling
```
Feiyang1: Is CJS format the root of the issue? I thought only the Node build is affected.

schmidt-sebastian: The CJS build that we published contains a bunch of variables prefixed with __PRIVATE_, which are not renamed by Terser. As such, the file is bigger than before.
`firebase-firestore.js` uses the ESM build, before converting to CJS, which means it is not affected.

PR #2662, opened 2020-02-21
Implement Timestamp.valueOf().

### LINKED ISSUE #2632: FR: Add valueOf() to Timestamp instances

  * Operating System version: macOS 10.15.2 (19C57)
  * Browser version: Node v1.21.1
  * Firebase SDK version: 7.13.0
  * Firebase Product: Firestore

Unable to compare Timestamp instances with native comparators. Javascript calls `valueOf()` on the object and uses the results of that to do native comparisons.

#### Steps to reproduce:
Attempt to compare Timestamps directly.

#### Relevant Code:
```javascript
firebase = require('firebase-admin')
a = new firebase.firestore.Timestamp(1, 1)
b = new firebase.firestore.Timestamp(2, 2)
console.log(a < b) // false
console.log(b < a) // false
```

#### Workaround
Of course you can simply call `toMillis()` and compare those, but having it auto handled would be better.

dconeybe:
Implement Timestamp.valueOf() so that Timestamp objects can be compared for relative ordering using the arithmetic comparison operators (i.e. <, <=, >, and >=).

Fixes #2632

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,13 +1,13 @@
 # Unreleased
+- [feature] Implemented `Timestamp.valueOf()` so that `Timestamp` objects can be
+  compared for relative ordering using the JavaScript arithmetic comparison
+  operators (#2632).
 - [fixed] Fixed an issue where auth credentials were not respected in Cordova
   environments (#2626).
 - [fixed] Fixed a performance regression introduced by the addition of
   `Query.limitToLast(n: number)` in Firestore 1.7.0 (Firebase 7.3.0) (#2620).
 - [fixed] Fixed an issue where `CollectionReference.add()` would reject
   custom types when using `withConverter()` (#2606).
-- [feature] Implemented Timestamp.valueOf() so that Timestamp objects can be
-  compared for relative ordering using the JavaScript arithmetic comparison
-  operators (#2632).
 
 # 1.9.3
 - [fixed] Fixed an issue where auth credentials were not respected in some
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -5,6 +5,9 @@
   `Query.limitToLast(n: number)` in Firestore 1.7.0 (Firebase 7.3.0) (#2620).
 - [fixed] Fixed an issue where `CollectionReference.add()` would reject
   custom types when using `withConverter()` (#2606).
+- [feature] Implemented Timestamp.valueOf() so that Timestamp objects can be
+  compared for relative ordering using the JavaScript arithmetic comparison
+  operators (#2632).
 
 # 1.9.3
 - [fixed] Fixed an issue where auth credentials were not respected in some
--- a/packages/firestore/src/api/timestamp.ts
+++ b/packages/firestore/src/api/timestamp.ts
@@ -100,7 +100,7 @@ export class Timestamp {
    *
    * Overriding this method allows Timestamp objects to be compared in JavaScript using the
    * arithmetic comparison operators, such as `<` and `>`.
-
+   *
    * See https://github.com/firebase/firebase-js-sdk/issues/2632.
    */
   valueOf(): string {
--- a/packages/firestore/src/api/timestamp.ts
+++ b/packages/firestore/src/api/timestamp.ts
@@ -96,8 +96,8 @@ export class Timestamp {
   // character) in its string representation, which would affect its lexiographical ordering.
   valueOf(): string {
     const adjustedSeconds = this.seconds - MIN_SECONDS;
-    const formattedSeconds = adjustedSeconds.toString().padStart(12, '0');
-    const formattedNanoseconds = this.nanoseconds.toString().padStart(9, '0');
+    const formattedSeconds = String(adjustedSeconds).padStart(12, '0');
+    const formattedNanoseconds = String(this.nanoseconds).padStart(9, '0');
     return formattedSeconds + '.' + formattedNanoseconds;
   }
 }
--- a/packages/firestore/src/api/timestamp.ts
+++ b/packages/firestore/src/api/timestamp.ts
@@ -89,4 +89,31 @@ export class Timestamp {
       ')'
     );
   }
+
+  // Overriding valueOf() allows Timestamp objects to be compared in JavaScript using the
+  // arithmetic comparison operators, such as < and >.
+  // https://github.com/firebase/firebase-js-sdk/issues/2632
+  valueOf(): string {
+    const formattedSeconds = Timestamp.normalizeAndPad(
+      this.seconds,
+      Timestamp.MIN_SECONDS,
+      Timestamp.MAX_SECONDS
+    );
+    const formattedNanoseconds = Timestamp.normalizeAndPad(
+      this.nanoseconds,
+      Timestamp.MIN_NANOSECONDS,
+      Timestamp.MAX_NANOSECONDS
+    );
+    return formattedSeconds + '.' + formattedNanoseconds;
+  }
+
+  private static normalizeAndPad(
+    value: number,
+    minValue: number,
+    maxValue: number
+  ): string {
+    const padLength = Math.ceil(Math.log10(maxValue - minValue));
+    const normalizedValue = value - minValue;
+    return normalizedValue.toString().padStart(padLength, '0');
+  }
 }
--- a/packages/firestore/src/api/timestamp.ts
+++ b/packages/firestore/src/api/timestamp.ts
@@ -19,6 +19,13 @@ import { Code, FirestoreError } from '../util/error';
 import { primitiveComparator } from '../util/misc';
 
 export class Timestamp {
+  // Midnight at the beginning of 1/1/1 is the earliest Firestore supports.
+  private static readonly MIN_SECONDS = -62135596800;
+  // This will break in the year 10,000.
+  private static readonly MAX_SECONDS = 253402300799;
+  private static readonly MIN_NANOSECONDS = 0;
+  private static readonly MAX_NANOSECONDS = 1e9 - 1;
+
   static now(): Timestamp {
     return Timestamp.fromMillis(Date.now());
   }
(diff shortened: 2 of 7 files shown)


packages/firestore/src/api/timestamp.ts
```diff
@@ -19,6 +19,13 @@ import { Code, FirestoreError } from '../util/error';
 import { primitiveComparator } from '../util/misc';
 
 export class Timestamp {
+  // Midnight at the beginning of 1/1/1 is the earliest Firestore supports.
+  private static readonly MIN_SECONDS = -62135596800;
+  // This will break in the year 10,000.
+  private static readonly MAX_SECONDS = 253402300799;
+  private static readonly MIN_NANOSECONDS = 0;
+  private static readonly MAX_NANOSECONDS = 1e9 - 1;
```
schmidt-sebastian: Note that "private" is not a thing in the final JavaScript build. When TypeScript transpiles this class to JS, these static members essentially become public. To indicate that we want these member to be treated as private members, we generally prefix all private members of our API classes.

packages/firestore/src/api/timestamp.ts
```diff
@@ -19,6 +19,13 @@ import { Code, FirestoreError } from '../util/error';
 import { primitiveComparator } from '../util/misc';
 
 export class Timestamp {
+  // Midnight at the beginning of 1/1/1 is the earliest Firestore supports.
+  private static readonly MIN_SECONDS = -62135596800;
+  // This will break in the year 10,000.
```
schmidt-sebastian: Nit: This constant won't break. It will continue to point to the year 10,000 even in the year 10,001. 
You can probably just mention that the value corresponds to the year 10,000.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +89,31 @@ export class Timestamp {
...
+  }
+
+  private static normalizeAndPad(
+    value: number,
+    minValue: number,
+    maxValue: number
+  ): string {
+    const padLength = Math.ceil(Math.log10(maxValue - minValue));
+    const normalizedValue = value - minValue;
+    return normalizedValue.toString().padStart(padLength, '0');
```
schmidt-sebastian: FYI: I started a discussion in the JS room whether we can use this API. See https://caniuse.com/#feat=pad-start-end for context.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +89,31 @@ export class Timestamp {
...
+    );
+    return formattedSeconds + '.' + formattedNanoseconds;
+  }
+
+  private static normalizeAndPad(
+    value: number,
+    minValue: number,
+    maxValue: number
+  ): string {
+    const padLength = Math.ceil(Math.log10(maxValue - minValue));
```
schmidt-sebastian: Most `pad()` methods that I am aware of take as an argument the desired length to pad to. I would suggest you do the same thing here: The nanosecond length is always 9, and the MAX_SECONDS value is a constant that you could define as const or via `Math.ceil(Math.log10(MAX_SECONDS)))`.

packages/firestore/test/unit/api/timestamp.test.ts
```diff
@@ -16,12 +16,55 @@
...
 
 import { expect } from 'chai';
+import { Code } from '../../../src/util/error';
 import { Timestamp } from '../../../src/api/timestamp';
 import { addEqualityMatcher } from '../../util/equality_matcher';
 
 describe('Timestamp', () => {
   addEqualityMatcher();
 
+  it('constructor should validate the "seconds" argument and store it.', () => {
```
schmidt-sebastian: Nit: I think in general we don't use period in our test names.

schmidt-sebastian:
Uhm, I think GitHub ate my last comment....

Please also add a test that shows that `<`, `=` and `>` now work correctly. You should also now be able to use Timestamps in Map objects, and object equality should now be based on the Timestamp data rather than the address of the Timestamp object. You can test this like this:


```
Map m = new Map();
m.add(new Timestamp(1,1));
expect(m.has(new Timestamp(1,1)).to.be.true;
```

Furthermore, can you also add a Changelog entry: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/CHANGELOG.md

Your PR description should also include "Fixes: <link to PR>" which will auto-close the issue on merging.

packages/firestore/src/api/timestamp.ts
wilhuff: FWIW, you could declare these as constants within the file rather than the class and then they wouldn't be exported at all.

dconeybe: I have moved these variables and the normalizeAndPad() private method to outside the class declaration so that they will be truly private.

packages/firestore/src/api/timestamp.ts
dconeybe: Haha true.  I just copied this comment from the code in its old location.  I've updated these comments to state the date to which they correspond and removed the mention of "breaking".

packages/firestore/src/api/timestamp.ts
dconeybe: The padStart() method is covered by the "required" polyfills documented at https://firebase.google.com/support/guides/environments_js-sdk#polyfills.  So I think we're okay.

packages/firestore/src/api/timestamp.ts
dconeybe: Thanks for the suggestion.  I'll simply hardcode the pad lengths, since there is really no value in computing those constants over and over again.

packages/firestore/test/unit/api/timestamp.test.ts
dconeybe: Fixed.

dconeybe:
> Uhm, I think GitHub ate my last comment....
> 
> Please also add a test that shows that `<`, `=` and `>` now work correctly. You should also now be able to use Timestamps in Map objects, and object equality should now be based on the Timestamp data rather than the address of the Timestamp object. You can test this like this:
> 
> ```
> Map m = new Map();
> m.add(new Timestamp(1,1));
> expect(m.has(new Timestamp(1,1)).to.be.true;
> ```
> 
> Furthermore, can you also add a Changelog entry: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/CHANGELOG.md
> 
> Your PR description should also include "Fixes: " which will auto-close the issue on merging.

Implementing valueOf() does not affect the == or === operators when used on Timestamp objects.  These operators just test if two objects reference the exact same object and do not test the properties of the objects.  By extension, using Timestamp objects as keys in a Map will not change as a result of this PR.

I have updated the PR description as suggested and will update the changelog as well.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +94,21 @@ export class Timestamp {
...
+
+  // Overriding valueOf() allows Timestamp objects to be compared in JavaScript using the
+  // arithmetic comparison operators, such as < and >.
+  // https://github.com/firebase/firebase-js-sdk/issues/2632
+  //
+  // This method returns a string of the form <seconds>.<nanoseconds> where <seconds> is translated
+  // to have a non-negative value and both <seconds> and <nanoseconds> are left-padded with zeroes
+  // to be a consistent length. Strings with this format then have a lexiographical ordering that
+  // matches the relative ordering of the Timestamp objects whose valueOf() method returned them.
+  // The <seconds> translation is done to avoid having a leading negative sign (i.e. a leading '-'
```
schmidt-sebastian: s/relative ordering of the Timestamp objects whose valueOf() method returned them/expected ordering.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +94,21 @@ export class Timestamp {
       ')'
     );
   }
+
+  // Overriding valueOf() allows Timestamp objects to be compared in JavaScript using the
```
schmidt-sebastian: Nit: The top part should be a JSDoc-style method comment.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +94,21 @@ export class Timestamp {
...
+  // arithmetic comparison operators, such as < and >.
+  // https://github.com/firebase/firebase-js-sdk/issues/2632
+  //
+  // This method returns a string of the form <seconds>.<nanoseconds> where <seconds> is translated
```
schmidt-sebastian: Can you move this comment into the method? You are describing implementation details. The consumer of the method should only now that the `valueOf()` return type should sort naturally.

packages/firestore/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
   `Query.limitToLast(n: number)` in Firestore 1.7.0 (Firebase 7.3.0) (#2620).
 - [fixed] Fixed an issue where `CollectionReference.add()` would reject
   custom types when using `withConverter()` (#2606).
+- [feature] Implemented Timestamp.valueOf() so that Timestamp objects can be
```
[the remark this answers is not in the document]
dconeybe: Done.

packages/firestore/src/api/timestamp.ts
```diff
@@ -47,7 +49,7 @@ export class Timestamp {
       );
     }
     // Midnight at the beginning of 1/1/1 is the earliest Firestore supports.
```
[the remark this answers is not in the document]
dconeybe: Done

packages/firestore/src/api/timestamp.ts
dconeybe: Done

packages/firestore/src/api/timestamp.ts
dconeybe: Done

packages/firestore/src/api/timestamp.ts
dconeybe: Done

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +94,25 @@ export class Timestamp {
...
+   */
+  valueOf(): string {
+    // This method returns a string of the form <seconds>.<nanoseconds> where <seconds> is
+    // translated to have a non-negative value and both <seconds> and <nanoseconds> are left-padded
+    // with zeroes to be a consistent length. Strings with this format then have a lexiographical
+    // ordering that matches the expected ordering. The <seconds> translation is done to avoid
+    // having a leading negative sign (i.e. a leading '-' character) in its string representation,
+    // which would affect its lexiographical ordering.
+    const adjustedSeconds = this.seconds - MIN_SECONDS;
+    const formattedSeconds = String(adjustedSeconds).padStart(12, '0');
```
schmidt-sebastian: Maybe add comment that the year 10,000 as 12 digits.

packages/firestore/src/api/timestamp.ts
```diff
@@ -92,4 +94,25 @@ export class Timestamp {
       ')'
     );
   }
+
+  /**
```
schmidt-sebastian: How about:

```
/**
   * Converts this object to a primitive string, which allows Timestamp objects to be compared
   * using the `>`, `<=, `>=` and `>` operators.
   */
```

Rationale:
- You are providing the override. A user doesn't need to do further overriding.
- We generally don't link to GitHub issues in top-level method comments. Feel free to move the link to the implementation if you think it provides value.
- We are in JavaScript land, so I dropped the explicit mention of the language name.

You also need to copy this API to https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts and https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts

dconeybe: Done

packages/firestore/src/api/timestamp.ts
dconeybe: Done

PR #2665, opened 2020-02-23
Allow for copyrights to be assigned to Google LLC

wilhuff:
Current guidance at go/copyright calls for copyright to be assigned to Google LLC, like so:

```
Copyright 2020 Google LLC
```

This change makes this the default without changing existing files. Going forward, any changed files will have their copyright line changed to match.

Note that 1ef26d56 shows what this rewrite will look like.

### DIFF
--- a/tools/gitHooks/license.js
+++ b/tools/gitHooks/license.js
@@ -11,7 +11,7 @@ const root = resolve(__dirname, '../..');
 const git = simpleGit(root);
 const licenseHeader = `/**
  * @license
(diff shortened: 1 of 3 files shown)


wilhuff:
This also addresses the following warning that happens during `yarn prepush` under Node.js 12:

> [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead

PR #2670, opened 2020-02-24
Populate GMPID header

schmidt-sebastian:
This populates the GMPID header for Storage on Web. 

See https://github.com/firebase/firebase-android-sdk/blob/e95f91e64c79206549302e9f37341534c29f58af/firebase-storage/src/main/java/com/google/firebase/storage/network/NetworkRequest.java#L53 for reference.


schmidt-sebastian:
b/147740859

schmidt-sebastian:
Adding @Feiyang1 for codeowners.

schmidt-sebastian:
The backend already accepts this header (for Storage only).

PR #2671, opened 2020-02-24
Add an ObjectValue builder

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/1150

Many more PRs to come. See https://github.com/firebase/firebase-android-sdk/pull/1217 for reference.

### DIFF
--- a/packages/firestore/test/unit/model/field_value.test.ts
+++ b/packages/firestore/test/unit/model/field_value.test.ts
@@ -240,11 +240,11 @@ describe('FieldValue', () => {
       .build();
     objValue = objValue
       .toBuilder()
-      .set(field('b'), wrap('c'))
+      .set(field('b'), wrap('b'))
       .set(field('c'), wrap('c'))
       .build();
 
-    expect(objValue.value()).to.deep.equal({ a: 'a', b: 'a', c: 'c' });
+    expect(objValue.value()).to.deep.equal({ a: 'a', b: 'b', c: 'c' });
   });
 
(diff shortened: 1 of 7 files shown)


packages/firestore/test/unit/model/field_value.test.ts
```diff
@@ -304,15 +319,31 @@ describe('FieldValue', () => {
...
+    const objValue2 = deleteField(objValue, 'foo.baz');
+    const objValue3 = deleteField(objValue, 'foo.bar.baz');
+    const objValue4 = deleteField(objValue, 'a.b');
     expect(objValue.value()).to.deep.equal(expected);
     expect(objValue2.value()).to.deep.equal(expected);
     expect(objValue3.value()).to.deep.equal(expected);
     expect(objValue4.value()).to.deep.equal(expected);
   });
 
+  it('can delete multiple fields', () => {
```
var-const: Optional: add a test showing it's possible to set and immediately delete fields? (e.g., set `a`, `b`, `c`, delete `b`, check that the end result only has `a`, `c`)

schmidt-sebastian:
> The description of the PR refers to issue 115 which seems unrelated, can you please double-check?

It's #1150. I updated the description.

packages/firestore/test/unit/model/field_value.test.ts
schmidt-sebastian: Added a test.

PR #2673, opened 2020-02-24
Fix promises in Analytics tests

hsubox76:
More strongly tie promise `await`s in Analytics tests to the actual function being waited for, rather than generic ticks.  This will prevent test errors when upgrading `tslib` to a new version that handles promises more correctly.


hsubox76:
Reverting dependency updates. These tests still pass with or without them, and better to keep separate PRs.

PR #2676, opened 2020-02-25
Use ByteString in model types

schmidt-sebastian:
This is a simplification that removes the api Blob types from our FieldValue API and exposed ByteString.compareTo() which I would like to use in the FieldValue rewrite.

This PR is against the feature branch since it might break canonical IDs.

### DIFF
--- a/packages/firestore/test/unit/api/blob.test.ts
+++ b/packages/firestore/test/unit/api/blob.test.ts
@@ -17,12 +17,7 @@
 
 import { expect } from 'chai';
 import { Blob, PublicBlob } from '../../../src/api/blob';
-import {
-  blob,
-  expectCorrectComparisons,
-  expectEqual,
-  expectNotEqual
-} from '../../util/helpers';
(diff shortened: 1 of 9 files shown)


packages/firestore/test/unit/api/blob.test.ts
```diff
@@ -70,26 +65,6 @@ describe('Blob', () => {
     expect(Blob.fromBase64String('') instanceof PublicBlob).to.equal(true);
   });
 
-  it('compares correctly', () => {
```
schmidt-sebastian: FYI: The ordering is also tested in field_value.test.ts

schmidt-sebastian:
cc @thebrianchen

schmidt-sebastian:
The canonical tests will cover this type when it is ported, so we will have a test then: https://github.com/firebase/firebase-android-sdk/pull/1178/files#diff-439fbbe706d75bbdf003c18383525777R379

PR #2677, opened 2020-02-25
Remove dependency on global atob/btoa functions

### LINKED ISSUE #2667: 7.9.1 prevent the use of firestore in React Native
Hey,

Just quick FYI to tell you that it looks like 7.9.1 is preventing the use of firestore on React Native.
You can find the detail [here](https://stackoverflow.com/questions/60361519/cant-find-a-variable-atob/60369040#60369040).

Thanks!

thebrianchen:
Fixes #2667 

#2639 introduced use of `atob` in `ByteString`, which is causing the failure in environments where there is no global `atob` function. This PR removes the dependency on  global `atob` by using our own `atob` function when global `atob` is unavailable.

### DIFF
--- a/packages/firestore/externs.json
+++ b/packages/firestore/externs.json
@@ -20,6 +20,7 @@
     "packages/component/dist/src/component_container.d.ts",
     "packages/logger/dist/src/logger.d.ts",
     "packages/webchannel-wrapper/src/index.d.ts",
+    "packages/util/dist/src/crypt.d.ts",
     "packages/util/dist/src/environment.d.ts",
     "packages/firestore/src/protos/firestore_proto_api.d.ts",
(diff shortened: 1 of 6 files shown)


schmidt-sebastian:
@thebrianchen Can you provide some size stats for the new dependency? I would like to know what impact this has on the size of `firebase-firestore.js`.

thebrianchen:
@schmidt-sebastian It's going from 318,422B to 321,001B.

schmidt-sebastian:
As discussed offline, please see if our guidance on Polyfills applies here. Please see here: https://firebase.google.com/support/guides/environments_js-sdk#polyfills

thebrianchen:
Plan is to update docs with react-native polyfills to reduce SDK size.

PR #2679, opened 2020-02-25
Add mangled CJS build

schmidt-sebastian:
This adds a mangled CJS build for Firestore and works around the fact that Terser's mangling doesn't work well with CJS by relying on the ESM build.

Size of WebApp with Firebase App + Firestore CJS before:
./bundle.js  445 KiB       0  [emitted]  [big]  main

Size of WebApp with Firebase App + Firestore CJS after:
./bundle.js  384 KiB       0  [emitted]  [big]  main

I confirmed in a simple test app that the build works in browsers (but more testing is always appreciated).

### DIFF
--- a/packages/firestore/rollup.config.js
+++ b/packages/firestore/rollup.config.js
@@ -88,23 +88,6 @@ const es5Builds = [
       )
   },
   /**
-   * Browser CJS Build
-   *
-   * The Browser CJS build is not mangled as Terser's property name mangling
-   * does not work well with CommonJS-style files.
-   */
-  {
-    input: 'index.ts',
-    output: { file: pkg.browser, format: 'cjs', sourcemap: true },
-    plugins: [
-      typescriptPlugin({
-        typescript,
-        cacheRoot: './.cache/cjs/'
-      }),
-      json()
-    ]
-  },
-  /**
    * Browser ESM Build
    */
   {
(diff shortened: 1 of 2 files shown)


packages/firestore/rollup.config.js
```diff
@@ -120,7 +103,14 @@ const es5Builds = [
...
     ],
     external: id => deps.some(dep => id === dep || id.startsWith(`${dep}/`))
-  }
+  },
+  /**
+   * Browser CJS Build (based on the mangling in the ESM build above)
+   */
+  {
+    input: pkg.module,
+    output: { file: pkg.browser, format: 'cjs', sourcemap: true },
```
schmidt-sebastian: The sourcemap generated by this step now maps to ESM file, rather than the actual sources. @Feiyang1 do you know how to fix this?

schmidt-sebastian: Maybe something like this will help: https://webpack.js.org/loaders/source-map-loader/

Feiyang1: `rollup-plugin-sourcemaps` is what you need. :)
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/rollup.config.js#L64

I noticed that enums are not mangled and are still prefixed with `__PRIVATE_` in the built files. Can you skip enums when marking symbols to be mangled?

schmidt-sebastian: Nice. That worked right away.

Re: Enums. In my original implementation, I tried to be smart and detect what the underlying types were as I parsed the code. This is really difficult to get right. I concluded that a type safe pipeline is too complex and fragile, and most of all does not buy as much in terms of size. Hence, I won't be able to tell at the callsite whether we are referring to an enum constant.

To solve the Enum trouble, I actually filed b/150211410 ("Reduce enum usage") this morning.

Feiyang1: Can you please add this dependency in `firestore/package.json` ?

schmidt-sebastian: Done

PR #2680, opened 2020-02-25
Adding test utilities to create Value types

schmidt-sebastian:
This is a port of https://github.com/firebase/firebase-android-sdk/pull/1158

### DIFF
--- a/packages/firestore/test/util/values.ts
+++ b/packages/firestore/test/util/values.ts
@@ -1,0 +1,109 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


packages/firestore/test/util/values.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+
+/** Creates a MapValue from a list of key/value arguments. */
+export function mapOf(...entries: unknown[]): api.Value {
+  const result: api.Value = { mapValue: { fields: {} } };
+  for (let i = 0; i < entries.length; i += 2) {
+    result.mapValue!.fields![entries[i] as string] = valueOf(
+      entries[i + 1],
+      /* useProto3Json= */ false
```
wu-hui: Curious: why is it OK to set it to always false here?

schmidt-sebastian: For now, this is a test utility only. The best I can tell you that we will find out soon if it indeed is ok.

PR #2687, opened 2020-02-27
Add explicit FieldValue canonicalization

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/1178


packages/firestore/src/model/proto_values.ts
```diff
@@ -335,6 +335,88 @@ function compareMaps(left: api.MapValue, right: api.MapValue): number {
   return primitiveComparator(leftKeys.length, rightKeys.length);
 }
 
+/**
+ * Generate the canonical ID for the provided field value (as used in Target
```
wu-hui: Nit: `Generates the ...`

schmidt-sebastian: Done

PR #2689, opened 2020-02-27
Explicit byte size accounting

schmidt-sebastian:
This copies the existing byte size accounting from the FieldValue classes and exposes it via Proto-only methods.

### DIFF
--- a/packages/firestore/src/model/proto_values.ts
+++ b/packages/firestore/src/model/proto_values.ts
@@ -242,7 +242,7 @@ function compareTimestamps(
   if (typeof left === 'string' && typeof right === 'string') {
     // Use string ordering for ISO 8601 timestamps, but strip the timezone
     // suffix to ensure proper ordering for timestamps of different precision.
-    // The only supported timezone is UTC (i.e. 'Z') based on 
+    // The only supported timezone is UTC (i.e. 'Z') based on
     // ISO_TIMESTAMP_REG_EXP.
     return primitiveComparator(left.slice(0, -1), right.slice(0, -1));
   }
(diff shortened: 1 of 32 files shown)


packages/firestore/src/model/proto_values.ts
```diff
@@ -417,6 +417,63 @@ function canonifyArray(arrayValue: api.ArrayValue) : string {
...
+  forEach(mapValue.fields || {}, (key, val) => {
+    size += key.length + estimateByteSize(val);
+  });
+  return size;
+}
+
+function estimateArrayByteSize(arrayValue: api.ArrayValue): number {
+  let size = 0;
+  for (const value of arrayValue.values || []) {
+    size += estimateByteSize(value);
```
var-const: Optional: this looks like `reduce`, though it doesn't look like it would change things substantially.

packages/firestore/src/model/proto_values.ts
```diff
@@ -372,20 +372,20 @@ function canonifyValue(value: api.Value): string {
...
   return `time(${normalizedTimestamp.seconds},${normalizedTimestamp.nanos})`;
 }
 
-function canonifyGeoPoint(geoPoint: api.LatLng) : string {
+function canonifyGeoPoint(geoPoint: api.LatLng): string {
   return `geo(${geoPoint.latitude},${geoPoint.longitude})`;
 }
 
-function canoifyMap(mapValue: api.MapValue) : string {
+function canoifyMap(mapValue: api.MapValue): string {
```
var-const: Nit: typo (s/canoify/canonify/).

schmidt-sebastian: Fixed

packages/firestore/src/model/proto_values.ts
schmidt-sebastian: I like it. Rewritten.

PR #2693, opened 2020-02-28
Fix racy promises in SpecTestRunner

thebrianchen:
Context: An update to `tslib` ([link](https://github.com/microsoft/tslib/pull/70)) changed the promise resolution order. We previously had multiple race conditions in the SpecTestRunner that happened to resolve in the correct order, but now break. Although we drain the AsyncQueue after each step, operations scheduled onto the AQ as part of the step run after the `drain()` call with the `tslib` update, as opposed to running before `drain()` previously.

This PR makes the `SpecTestRunner` drain until the AQ is empty, which ensures that the `SpecTestRunner` is in a consistent state before moving onto the next step.


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -280,11 +288,16 @@ class MockConnection implements Connection {
...
-      });
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.queue
+        .enqueue(async () => {
+          if (this.writeStream === writeStream) {
+            writeStream.callOnOpen();
+          }
+        })
+        .then(() => {
+          this.writeOpen.resolve();
```
schmidt-sebastian: Drive-by comment: You can move line 299 into the previous block.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -549,6 +549,16 @@ abstract class TestRunner {
   async run(step: SpecStep): Promise<void> {
     await this.doStep(step);
     await this.queue.drain();
+    // Wait for any additional operations that the initial operations
+    // scheduled. We skip this iterative drain for steps that involve changing
```
wilhuff: If you can phrase it naturally, try to avoid "we" in comments about the system. Sometimes it can't be avoided--don't worry about it too much.

In this case using this form is placing unnecessary emphasis on us as the authors when the real actors here are the operations that might change the primary tab. Consider rephrasing to put more emphasis on that. Something like:

> However, when the current step changes the primary tab, the additional operations are expected by later steps, so avoid draining until empty in this case.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -549,6 +549,16 @@ abstract class TestRunner {
   async run(step: SpecStep): Promise<void> {
     await this.doStep(step);
     await this.queue.drain();
+    // Wait for any additional operations that the initial operations
```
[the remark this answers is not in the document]
thebrianchen: wow that is the elusive brevity I was searching for

packages/firestore/test/unit/specs/spec_test_runner.ts
thebrianchen: thanks for the explanation -- that makes a lot of sense

thebrianchen:
[link](https://github.com/firebase/firebase-js-sdk/blob/332346761a3e872592b44f06fbd89d15745a6186/packages/firestore/test/unit/specs/limbo_spec.test.ts#L481)

packages/firestore/src/util/async_queue.ts
```diff
@@ -219,6 +222,11 @@ export class AsyncQueue {
     return this._isShuttingDown;
   }
 
+  // Visible for testing
+  isEmpty(): boolean {
```
schmidt-sebastian: We should be able to make this private or remove it entirely.

packages/firestore/src/util/async_queue.ts
```diff
@@ -372,9 +382,13 @@ export class AsyncQueue {
    * Waits until all currently queued tasks are finished executing. Delayed
    * operations are not run.
    */
-  drain(): Promise<void> {
-    // It should still be possible to drain the queue after it is shutting down.
-    return this.enqueueEvenAfterShutdown(() => Promise.resolve());
+  async drain(): Promise<void> {
+    // Operations in the queue prior to draining may have enqueued additional
```
schmidt-sebastian: I don't know if this works, but if it does you could get rid of all size accounting:

```
let currentTail : Promise<void>;
do (
  currentTail = this.tail;
  await currentTail;
} while (currentTail !== this.tail);
```

thebrianchen: done.

PR #2697, opened 2020-02-29
Remove enum overhead

schmidt-sebastian:
This removes the overhead that Enums introduce in TypeScript by changing our enums to "const enums" which can be inlined: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

This also reverts some of the changes that were made to the OnlineState enum and uses the same solution.

Size before: 318393 bytes (90882 gzipped)
Size after: 315811 bytes (89863 gzipped - 1.2%)

### DIFF
--- a/packages/firestore/src/core/event_manager.ts
+++ b/packages/firestore/src/core/event_manager.ts
@@ -51,7 +51,7 @@ export class EventManager implements SyncEngineListener {
     q.canonicalId()
   );
 
-  private onlineState: OnlineState = 'Unknown';
+  private onlineState: OnlineState = OnlineState.Unknown;
 
   private snapshotsInSyncListeners: Set<Observer<void>> = new Set();
 
@@ -209,7 +209,7 @@ export class QueryListener {
 
   private snap: ViewSnapshot | null = null;
 
-  private onlineState: OnlineState = 'Unknown';
+  private onlineState: OnlineState = OnlineState.Unknown;
 
   constructor(
     readonly query: Query,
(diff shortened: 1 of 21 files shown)


schmidt-sebastian:
Note that the TypeScript style guides recommends against this, but the reasons do not apply to our codebase.

schmidt-sebastian:
FYI @Feiyang1

packages/firestore/src/core/event_manager.ts
```diff
@@ -51,7 +51,7 @@ export class EventManager implements SyncEngineListener {
     q.canonicalId()
   );
 
-  private onlineState: OnlineState = 'Unknown';
+  private onlineState: OnlineState = OnlineState.Unknown;
```
wilhuff: You can now probably move the explicit type here (i.e. restore it back to)

```ts
private onlineState = OnlineState.Unknown;
```

?

packages/firestore/src/core/types.ts
```diff
@@ -39,31 +39,32 @@ export type MutationBatchState = 'pending' | 'acknowledged' | 'rejected';
...
-export type OnlineState =
+export const enum OnlineState {
   /**
    * The Firestore client is in an unknown online state. This means the client
    * is either not actively trying to establish a connection or it is currently
    * trying to establish a connection, but it has not succeeded or failed yet.
    * Higher-level components should not operate in offline mode.
    */
-  | 'Unknown'
+  Unknown = 'Unknown',
```
wilhuff: Do we need the strings in here anymore? Can these just be numeric?

packages/firestore/src/core/event_manager.ts
schmidt-sebastian: Done. Note that this was part of a straight revert.

packages/firestore/src/core/types.ts
schmidt-sebastian: The string values are persisted in WebStorage (but it is mostly transient data). I added a comment to explain this.

PR #2707, opened 2020-03-02
Fix performance regression. Add tests against actual server behavior.

inlined:
Work in progress. I understand the bug fix that Mike wanted us to apply, though the result would still be pessimal (would require us to calculate resolved values always even if the there was no server values).

In trying to test the bug/fix I ran into some issues. It seems like the update command and racy sets are failing. The "handles sparse updates" test fails because the second {increment: X} is resolved as a set X. Some extra debug statements in the racy write case seem to suggest that the values are incrementing and re-decrementing with server-side results. The emulator itself seems to work, so I'm not sure what the problem is yet. Maybe the SyncTree?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
       "type": "node",
       "request": "launch",
       "name": "RTDB Unit Tests (Node)",
-      "program": "${workspaceRoot}/packages/firestore/node_modules/.bin/_mocha",
+      "program": "${workspaceRoot}/packages/firebase/node_modules/.bin/_mocha",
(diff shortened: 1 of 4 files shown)


inlined:
Confirmed that the bugs are in the server-side code, not the client-side. Tests pass with an emulator built on @jsdt 's [change](https://github.com/firebase/firebase-server/pull/3150).

This change addresses Mike's perf bugfix (though my other change would more aggressively avoid pessimization) and adds tests that revealed the serverside bugs.

inlined:
Sorry anyone who got spammed. Tried rebasing to get the testing framework to only touch db tests and it auto-added a bunch of reviewers.

PR #2708, opened 2020-03-02
Add comments explain why remote doc cache needs to be updated.

wu-hui:


packages/firestore/src/local/local_store.ts
```diff
@@ -606,6 +606,10 @@ export class LocalStore {
                 } else if (
                   existingDoc == null ||
                   doc.version.compareTo(existingDoc.version) > 0 ||
+                  // `existingDoc` might have `hasCommittedMutations` == true,
+                  // while `doc` might have some changes resulted from other users
+                  // change, which are not picked up by committed mutations from current user.
+                  // We need to update remoteDocumentCache when this happens.
                   (doc.version.compareTo(existingDoc.version) === 0 &&
```
schmidt-sebastian: How about:

If our local document version is based on a write acknowledged from the WriteStream, we overwrite it with the remote version from the WatchStream since the remote document may contain edits from other clients.

PR #2717, opened 2020-03-06
Avoid rebuilding the repo during the global test run

Feiyang1:
We don't want to rebuild the whole repo when running tests. It will slow down `yarn test` at the root of the repo and CI.


Feiyang1:
`firebase-firestore.js` is guaranteed to be created in every CI run because we always run `yarn build` at the root of the repo first. This PR will help us avoid one unnecessary rebuild.

PR #2720, opened 2020-03-06
size presubmit check

Feiyang1:


scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
+const { resolve } = require('path');
+const fs = require('fs');
+const { execSync } = require('child_process');
+
+const repoRoot = resolve(__dirname, '..');
+
+const commitHash = process.env.GITHUB_SHA || execSync('git rev-parse HEAD').toString();
```
yifanyang: We seem to use `simple-git` for git operations in [other places in this repo](https://github.com/firebase/firebase-js-sdk/blob/master/scripts/release/utils/git.js#L18).

I'm okay with both. Will leave it up to you whether we want to pursue any consistency.

scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
...
+    
+    const files = [
+        ...special_files.map(file => `${firebaseRoot}/${file}`), 
+        ...pkgJson.components.map(component => `${firebaseRoot}/firebase-${component}.js`)
+    ];
+    
+    for (const file of files) {
+        const { size } = fs.statSync(file);
+        const fileName = file.split('/').slice(-1)[0]
+        reports.push(makeReportObject('firebase', fileName, size))
```
yifanyang:
```suggestion
        const fileName = file.split('/').slice(-1)[0];
        reports.push(makeReportObject('firebase', fileName, size));
```

scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    return reports;
+}
+
+// @firebase/*
```
yifanyang:
```suggestion
// NPM packages
```

scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
...
+        ...generateReportForCDNScripts(),
+        ...generateReportForNPMPacakges()
+    ];
+
+    for (const r of reports) {
+        console.log(r.sdk, r.type, r.value);
+    }
+
+    return {
+        log: "https://www.goog.com",
```
yifanyang: I guess this is only applicable to CI runs. So maybe we want to configure it to be some value when on CI and empty when running locally.

Or we can leave this field empty. The only use of this field is to provide a link on the PR comment for developers to examine the original log (when they are surprised by the number.)

scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
...
+        metric: "BinarySize",
+        results: reports
+    };
+}
+
+generateSizeReport();
```
yifanyang: Can we write the result JSON to a file? The way to connect it to the metrics service to make a http post request with that file as the request body.

yifanyang: Oh, never mind. We are probably going to sent request programmatically anyway. So it is good now.

scripts/report_binary_size.js
```diff
@@ -0,0 +1,93 @@
...
+        const { size } = fs.statSync(file);
+        const fileName = file.split('/').slice(-1)[0]
+        reports.push(makeReportObject('firebase', fileName, size))
+    }
+
+    return reports;
+}
+
+// @firebase/*
+function generateReportForNPMPacakges() {
```
hsubox76: Typo ("Pacakges")

scripts/report_binary_size.js
Feiyang1: Done.

scripts/report_binary_size.js
Feiyang1: Done.

scripts/report_binary_size.js
Feiyang1: Corrected.

scripts/report_binary_size.js
Feiyang1: updated to report github action url.

Feiyang1:
replaced by https://github.com/firebase/firebase-js-sdk/pull/2744

PR #2727, opened 2020-03-09
Plumb handling of Proto3Json

schmidt-sebastian:
This PR removes the `FieldValue.proto` property in favor of a method that accepts `useProto3Json` as an argument. Since the SDK either uses Proto3 JSON or protobufJs throughout, there is no conversion logic. 

I also updated the tests to use the new flag, which introduces two different modes. As a side effect I was able to catch a regression: The String representation of `1` and NaN didn't compare correctly.

### DIFF
--- a/packages/firestore/src/model/field_value.ts
+++ b/packages/firestore/src/model/field_value.ts
@@ -256,7 +256,7 @@ export class DoubleValue extends NumberValue {
 
   toProto(useProto3Json: boolean): api.Value {
     if (useProto3Json) {
-      // Proto 3 let's us encode NaN and Infinity as string values as
+      // Proto 3 lets us encode NaN and Infinity as string values as
       // expected by the backend. This is currently not checked by our unit
       // tests because they rely on protobuf.js.
       if (isNaN(this.internalValue)) {
--- a/packages/firestore/src/model/field_value.ts
+++ b/packages/firestore/src/model/field_value.ts
@@ -685,7 +685,7 @@ export class ObjectValue extends FieldValue {
   }
 
   toProto(useProto3Json: boolean): api.Value {
-    let result: api.Value = {
+    const result: api.Value = {
       mapValue: { fields: {} }
     };
 
--- a/packages/firestore/src/model/field_value.ts
+++ b/packages/firestore/src/model/field_value.ts
@@ -15,10 +15,12 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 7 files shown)


schmidt-sebastian:
Whitespace-removed diff: https://github.com/firebase/firebase-js-sdk/pull/2727/files?w=1

packages/firestore/src/model/field_value.ts
```diff
@@ -235,6 +254,22 @@ export class DoubleValue extends NumberValue {
     }
   }
 
+  toProto(useProto3Json: boolean): api.Value {
+    if (useProto3Json) {
+      // Proto 3 let's us encode NaN and Infinity as string values as
```
dconeybe: Drive-by comment: "let's" should be "lets"

schmidt-sebastian: This was copied, so I got to apply this fixed twice. Thanks.

packages/firestore/src/model/field_value.ts
```diff
@@ -235,6 +254,22 @@ export class DoubleValue extends NumberValue {
...
 
+  toProto(useProto3Json: boolean): api.Value {
+    if (useProto3Json) {
+      // Proto 3 lets us encode NaN and Infinity as string values as
+      // expected by the backend. This is currently not checked by our unit
+      // tests because they rely on protobuf.js.
+      if (isNaN(this.internalValue)) {
+        return { doubleValue: 'NaN' };
+      } else if (this.internalValue === Infinity) {
+        return { doubleValue: 'Infinity' } as {};
```
wilhuff: Why are the `as {}` required here?

packages/firestore/src/model/field_value.ts
```diff
@@ -235,6 +254,22 @@ export class DoubleValue extends NumberValue {
     }
   }
 
+  toProto(useProto3Json: boolean): api.Value {
+    if (useProto3Json) {
+      // Proto 3 lets us encode NaN and Infinity as string values as
```
wilhuff: This comment reads backwards, as if it's describing the proto format, when really it's the JSON encoding of the proto that has this string form.  In other words, both encodings are proto 3 so we really need to distinguish between JSON or binary. It would be clearer if you wrote it something like this:

> In the JSON encoding, NaN and Infinity are string values ...

PR #2728, opened 2020-03-09
Update analytics ecommerce event and param types

hsubox76:
Update Analytics types and method signatures to reflect GA's enhanced ecommerce options.

- Some events and params have been added.
- Some have been deprecated.
- Some existing methods have the number of suggested params expanded.

The latter should only affect IDE autocomplete. All event params are always optional and any additional arbitrary param is always allowed.

This changes the public API and will require a documentation update.

Internal bug for this feature: b/151107871

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -4008,7 +4008,35 @@ declare namespace firebase.analytics {
      */
     logEvent(
       eventName: 'add_payment_info',
-      eventParams?: { [key: string]: any },
+      eventParams?: {
+        coupon?: EventParams['coupon'];
+        currency?: EventParams['currency'];
+        items?: EventParams['items'];
+        payment_type?: EventParams['payment_type'];
+        value?: EventParams['value'];
+        [key: string]: any;
+      },
+      options?: firebase.analytics.AnalyticsCallOptions
+    ): void;
+
+    /**
+     * Sends analytics event with given `eventParams`. This method
+     * automatically associates this logged event with this Firebase web
+     * app instance on this device.
+     * List of official event parameters can be found in
+     * {@link https://developers.google.com/gtagjs/reference/event
+     * the gtag.js reference documentation}.
+     */
+    logEvent(
+      eventName: 'add_shipping_info',
+      eventParams: {
+        coupon?: EventParams['coupon'];
+        currency?: EventParams['currency'];
+        items?: EventParams['items'];
+        shipping_tier?: EventParams['shipping_tier'];
+        value?: EventParams['value'];
+        [key: string]: any;
+      },
       options?: firebase.analytics.AnalyticsCallOptions
     ): void;
 
@@ -4155,7 +4183,7 @@ declare namespace firebase.analytics {
      * the gtag.js reference documentation}.
      */
     logEvent(
-      eventName: 'purchase',
+      eventName: 'purchase' | 'refund',
       eventParams: {
         value?: EventParams['value'];
         currency?: EventParams['currency'];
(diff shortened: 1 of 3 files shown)


packages/firebase/index.d.ts
```diff
@@ -4008,7 +4008,35 @@ declare namespace firebase.analytics {
...
+    /**
+     * Sends analytics event with given `eventParams`. This method
+     * automatically associates this logged event with this Firebase web
+     * app instance on this device.
+     * List of official event parameters can be found in
+     * {@link https://developers.google.com/gtagjs/reference/event
+     * the gtag.js reference documentation}.
+     */
+    logEvent(
+      eventName: 'add_shipping_info',
```
Feiyang1: Where are the new events and their associated params defined? I don't see them in the proposal.

packages/firebase/index.d.ts
```diff
@@ -4477,6 +4526,14 @@ declare namespace firebase.analytics {
     value?: number;
     event_label?: string;
     event_category: string;
+    shipping_tier?: string;
```
Feiyang1: The proposal only adds the following:
```
  // New suggested event parameters.
  shipping_tier?: string;
  item_list?: string;
  item_list_id?: string;
  payment_type?: string;
```
Where are the other ones defined?

packages/firebase/index.d.ts
```diff
@@ -4008,7 +4008,35 @@ declare namespace firebase.analytics {
...
+        [key: string]: any;
+      },
+      options?: firebase.analytics.AnalyticsCallOptions
+    ): void;
+
+    /**
+     * Sends analytics event with given `eventParams`. This method
+     * automatically associates this logged event with this Firebase web
+     * app instance on this device.
+     * List of official event parameters can be found in
```
egilmorez: Curious about the word "official."  Do we just mean "valid?"

I can see omitting an adjective entirely; "See <link> for a list of event parameters."

packages/firebase/index.d.ts
```diff
@@ -4238,12 +4265,30 @@ declare namespace firebase.analytics {
...
```
egilmorez: Same comment as above RE "official" and rephrasing.

packages/firebase/index.d.ts
hsubox76: No, now that I think about it, the term I've seen in other analytics docs might be "recommended".  Users can put any string they want and it will just be ingested as a custom value, but the "recommended"(?) ones are the ones that the analytics backend recognizes and can do some special things with. Is "recommended" the right word? It's actually usually used for event types, not event params: https://support.google.com/firebase/answer/6317498?hl=en&ref_topic=6317484

packages/firebase/index.d.ts
hsubox76: Same as above, I've also seen "default".  We definitely don't want users to think they can't use any other strings.

packages/firebase/index.d.ts
hsubox76: See spreadsheet linked in internal bug: b/151107871

packages/firebase/index.d.ts
egilmorez: Aha -- "recommended" definitely sounds best, if the value can be any string.

packages/firebase/index.d.ts
egilmorez: Yup, makes sense.  If you think "default" is better than "recommended," let's go with that, thanks!

packages/firebase/index.d.ts
hsubox76: Changed to "recommended".

packages/firebase/index.d.ts
hsubox76: Changed to "recommended".

packages/firebase/index.d.ts
Feiyang1: `item_list` is not present in the sheet. I think it has been replaced by `item_list_name`, see [comment](https://docs.google.com/document/d/1cJ0MsSbN6fzB36X9RHIZMuGJBQgqLxffXcgM0ofRIC4/edit?disco=AAAAJLaOVWQ).

hsubox76: Removed.

PR #2744, opened 2020-03-13
Report binary size metrics on GitHub pull requests.

yifanyang:


yifanyang:
I'm able to enable and configure it in my forked repo. See this [pull request comment](https://github.com/yifanyang/firebase-js-sdk/pull/1#issuecomment-598645446).

Github Actions are failing since the workflow requires adding new secrets in this repo, which I don't have permission to do. @Feiyang1 Can you help me here?

.github/workflows/test-changed.yml
```diff
@@ -28,3 +28,27 @@ jobs:
       run: yarn build
     - name: Run tests on changed packages
       run: xvfb-run yarn test:changed
+
+  health-metrics-test:
```
Feiyang1: Do we want to run it on push? I think we want to limit it to run on PR only, with a condition like
```
if: github.event_name == 'pull_request'
```

.github/workflows/test-changed.yml
```diff
@@ -28,3 +28,27 @@ jobs:
...
+    env:
+      METRICS_SERVICE_URL: ${{ secrets.METRICS_SERVICE_URL }}
+      GITHUB_PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
+      GITHUB_PULL_REQUEST_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10.x
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
Feiyang1: Is it for connecting to your service?

scripts/report_binary_size.js
```diff
@@ -0,0 +1,148 @@
...
+    console.log(`${report.metric} report:`, report);
+    console.log(`Posting to metrics service endpoint: ${path} ...`);
+
+    const request = https.request(METRICS_SERVICE_URL, options, response => {
+        response.setEncoding('utf8');
+        console.log(`Response status code: ${response.statusCode}`);
+        response.on('data', console.log);
+        response.on('end', () => {
+            if (response.statusCode !== 202) {
+                process.exit(1);
```
Feiyang1: Can you log the status code that's returned and any error message in case there is an error?

.github/workflows/test-changed.yml
yifanyang: It will need to be run on push for master branch. When we want to conduct the comparison for a pull request, the metrics for the PR head commit is from the pull request run, the metrics for the comparison base commit (usually a commit on master) actually comes from a push run for master. I'm okay will limiting it to pr runs and push runs for master only.

Also you mentioned that not all pull requests are to be merged to master. Let's call their merge targets "feature branch". If we also want to support diffing in those pull requests. We will also need to include the size test in push builds for those "feature branch"es. I'm not sure about the priority for this one.

Does it make sense?

.github/workflows/test-changed.yml
yifanyang: Yes. The metrics service, the one that receives size measurements, calculates the diff and posts reports back to pull requests as a comment, is hosted on GCP. Google cloud SDK is needed here to authenticate our size test so that it is allowed to call the metrics service.

scripts/report_binary_size.js
yifanyang: The status code is logged at line 135.

Line 136 logs the response body, which will be
* in case of success, an id generated by the backend service for the uploaded report
* in case of failure, various reason from unauthenticated to the error message of service internal exceptions

.github/workflows/test-changed.yml
Feiyang1: This trigger doesn't run when a PR is merged to master, but `test-all` workflow runs instead. I think it is because merging is not the same as directly pushing a commit.

"feature branch" diff will be nice to have. Will it require substantial change to enable?

yifanyang: Ahh, I'm not sure... For example this ["Push/PR" run](https://github.com/firebase/firebase-js-sdk/runs/506447993?check_suite_focus=true) seems to be from master branch. If we take a look at its [log](https://github.com/firebase/firebase-js-sdk/runs/506447993?check_suite_focus=true#step:2:21) for checking out source code, the commit being checked out and tested is actually the commit on master. The commit sha (4277e4c5fca2e95aabd80aef99b41c5fb06acb1e) is identical to the one in this ["Run All Tests" run](https://github.com/firebase/firebase-js-sdk/runs/506447974?check_suite_focus=true#step:2:21), which is for merging this [pull request](https://github.com/firebase/firebase-js-sdk/pull/2434). `git log origin/master` shows all commits on master, and it seems that they all can be found in this [history of "Push/PR" runs](https://github.com/firebase/firebase-js-sdk/actions?query=workflow%3APush%2FPR+branch%3Amaster).

I think here is what happened: after a pull request is merged to master, the new commit on master triggered both the "Push/PR" workflow and the "Run All Tests" workflow. The reason why there is no package specific test run in the "Push/PR" workflow is that the "Push/PR" workflow runs this [script](https://github.com/firebase/firebase-js-sdk/blob/master/.github/workflows/test-changed.yml#L30), in which we always compare the current checked out code (the `HEAD`) to the master (the `origin/master`) with this [line](https://github.com/firebase/firebase-js-sdk/blob/master/scripts/run_changed.js#L83). In case when the `HEAD` is actually the master branch, that `git` command returns no code change, thus no package specific test got run, as shown in this [line](https://github.com/firebase/firebase-js-sdk/runs/506447993?check_suite_focus=true#step:8:6).

In case of our size test here, since size test is an independent job in the workflow yaml file (not triggered by the `run_changed.js` script), as long as the "Push/PR" workflow is triggered, the size test will run. And I think it does get triggered by all new commits on master branch. In fact, our "Push/PR" workflow now is set to be triggered by new commit on all branches including master and "feature branches", thus the current code actually supports diffing for "feature branches".

(I'm not sure if I have been clear enough in this comment, as the reasoning process involved a lot of concepts. We can probably have a quick GVC to discuss it through.)

Feiyang1: Ah I see. I don't know if it is the intended behavior to run Push/PR on master because it does nothing. @hsubox76 Can you please clarify?

Commits to master should always run the full tests IMO. Right now I think our triggers are a little confusing, is it possible to add comment in yaml file or is there alternative triggers that we can use that makes the intent clearer?

hsubox76: Yes, so if I'm understanding correctly, the Push/PR trigger is `on: [push, pull_request]` which runs it as we intend, on every change to the PR and push to the PR's branch, but also unintentionally runs it on merge to master, which is useless.  I think we should try to exclude it.

The correct trigger for deliberately running tests ONLY on PR closed + merge to master (used for "Run All Tests") is
```
on: 
  pull_request:
    branches:
      - master
    types:
      - closed
```

If you want to trigger on a different combination like, both "pull request" AND "push to master"  but not "push to branch" maybe it's worth making a separate workflow file just for that workflow.

yifanyang: I think there is another problem with this trigger:
```
on: 
  pull_request:
    branches:
      - master
    types:
      - closed
```
For example, PR #2735 is closed without merging to master, but the closing event still triggered a [run](https://github.com/firebase/firebase-js-sdk/actions/runs/54609596). It can also be found by searching for text "2735" on the [history page for the "Run All Tests" workflow](https://github.com/firebase/firebase-js-sdk/actions?query=workflow%3A%22Run+All+Tests%22+event%3Apull_request). Is that expected?

hsubox76: I think it might be simpler to just change that trigger to
```
on:
  push:
    branches:
      - master
```

The only unwanted case there will be when we merge release branch to master every week, it will run, which we don't need, but that doesn't seem a big deal and in all the other cases will do what we want, I think, and avoid doing the above.

yifanyang: SGTM. Do you want me to include the trigger changes in this PR, or do you prefer to leave it in a separate PR?

hsubox76: The current plan is that we are trying to separate push and PR into different workflows because push can often be redundant and gets incorrect diff results on stale branches.  

- The PR workflow would only cover PR updates not including merge to master.  It will run tests on changed files and it will block PR submission.
- The push workflow would cover all pushes, including merge to master.  It will run all tests, and NOT block PR submission.

I could do it in a separate PR but I think you probably want to decide where to put your job based on this separation, or if neither trigger is appropriate, or if there is a flaw in this plan overall.  I think the triggers would be:

```
on: 
  pull_request:
    branches:
      - '*'
      - '!master'
```
and
```
on: push
```

yifanyang: This size test need to be run on all pull_requests and pushes (at least the ones from master).

How about we do this -- I'll hold off this PR for now, wait for your changes for workflows to merge in first, and then accommodate your change here. Does it make sense?

hsubox76: Sure, feel free to double check I am not missing something: https://github.com/firebase/firebase-js-sdk/pull/2763

yifanyang:
I'm making the size test in its own workflow for these reasons:
* Size test need to be run on both push and pull_request. This way I don't need to duplicate the job setup in two files (`test-all.yml` and `test-changed.yaml`).
* The displayed name seems clearer: "Health Metrics / Binary Size". Later I would like to surface startup times measurement in pull request as well, and it will be "Health Metrics / Startup Times".

What do you think?

Feiyang1:
LGTM

PR #2749, opened 2020-03-17
Add ProtoJS types to generated proto types

schmidt-sebastian:
See #2715 for the bigger picture.


packages/firestore/src/remote/serializer.ts
```diff
@@ -99,15 +99,13 @@ const OPERATORS = (() => {
...
 }
 
-// This is a supplement to the generated proto interfaces, which fail to account
-// for the fact that a timestamp may be encoded as either a string OR this.
-interface TimestampProto {
-  seconds?: string | number;
-  nanos?: number;
-}
+// Denotes the possible representations for timestamps in the Value type.
+export type TimestampValue =
```
wilhuff: The `Value` proto can contain a timestamp and the name `TimestampValue` suggests that this is one of those, when it's really just the timestamp itself. Also, this is somewhat confusing because historically FooValue has been one of our model types and not a proto. In this regard the old `TimestampProto` name was better.

packages/firestore/src/remote/serializer.ts
```diff
@@ -148,46 +146,33 @@ export class JsonProtoSerializer {
    * our generated proto interfaces say Int32Value must be. But GRPC actually
    * expects a { value: <number> } struct.
    */
-  private toInt32Value(val: number | null): number | null {
-    if (this.options.useProto3Json || typeUtils.isNullOrUndefined(val)) {
+  private toInt32Value(val: number | null): number | { value: number } | null {
```
wilhuff: Optional: `toInt32Value` doesn't really do much to indicate that it's converting to a proto message. What do you think of renaming these (and related items) to names like `toInt32Proto`?

The Android/iOS convention of naming these `encode`/`decode` helps somewhat.

packages/firestore/src/protos/firestore_proto_api.d.ts
```diff
@@ -180,7 +180,7 @@ export declare namespace firestoreV1ApiClientInterfaces {
     name?: string;
     fields?: ApiClientObjectMap<Value>;
     createTime?: string;
-    updateTime?: string;
+    updateTime?: string | { seconds?: string | number; nanos?: number };
```
wilhuff: Is this generated or hand edited?

If hand edited, would it be worthwhile to define the timestamp type in here and reuse it for all these declarations?

Then we could just call it `Timestamp` and refer to it as `api.Timestamp`.

packages/firestore/src/remote/serializer.ts
schmidt-sebastian: I renamed it to Timestamp, which aligns with LatLng and GeoPoint. I also moved it to the proto d.ts file.

packages/firestore/src/remote/serializer.ts
schmidt-sebastian: I renamed this instance to `toInt32Proto` but did not convert the methods to the encode/decode pattern. I am personally also confused by the naming in the file and constantly try to use the wrong methods (`toValue` instead of `fromValue` etc)

packages/firestore/src/protos/firestore_proto_api.d.ts
schmidt-sebastian: This used to be generated, but has been hand edited many times since then. As such, I added `api.Timestamp`.

PR #2750, opened 2020-03-17
Rename proto_values->values, create values.test.ts

schmidt-sebastian:
See #2715 for the bigger picture (and the source of values.test.ts, but it is mostly just a move of code)


schmidt-sebastian:
Note: The first commit just moves code around, the second is the diff.

packages/firestore/src/model/values.ts
```diff
@@ -385,6 +385,10 @@ function canonifyGeoPoint(geoPoint: api.LatLng): string {
   return `geo(${geoPoint.latitude},${geoPoint.longitude})`;
 }
 
+function canonifyReference(referenceValue: string): string {
+  return DocumentKey.fromName(referenceValue).toString();
```
wilhuff: Doesn't this break existing canonical IDs?

packages/firestore/test/unit/model/values.test.ts
```diff
@@ -0,0 +1,420 @@
...
+      [wrap({ bar: 1, foo: 1 })],
+      [wrap({ foo: 1 })]
+    ];
+    expectEqualitySets(values, (v1, v2) => equals(v1, v2));
+  });
+
+  it('normalizes values for equality', () => {
+    // Each subarray compares equal to each other and false to every other value
+    const values: api.Value[][] = [
+      [{ integerValue: '1' }, { integerValue: 1 }],
```
wilhuff: Isn't there also a case where there's a nested value object?

packages/firestore/src/model/values.ts
schmidt-sebastian: Yes, but the feature branch includes the schema migration.

packages/firestore/test/unit/model/values.test.ts
schmidt-sebastian: If you are referring to the `Int32Proto`, this seems to only be used in a target's `limit` specification, which allows us to encode a difference between `0` and unset. I don't see any usage for this patten in api.Value.

wilhuff: Ah, I see. Right.

PR #2751, opened 2020-03-17
Support -0.0 in Firestore

schmidt-sebastian:
We have a couple of unit tests that verify that we treat DoubleValue -0.0 distinct from 0.0, but we were never storing -0.0 as a DoubleValue.

### DIFF
--- a/packages/firestore/src/util/types.ts
+++ b/packages/firestore/src/util/types.ts
@@ -44,7 +44,7 @@ export const MAX_SAFE_INTEGER: number =
  * Added to not rely on ES6 features.
  * @param value The value to test for being an integer
  */
-export const isInteger: (value: unknown) => boolean =
+export const isInteger: (value: unknown) => value is number =
   NumberAsAny.isInteger ||
   (value =>
     typeof value === 'number' &&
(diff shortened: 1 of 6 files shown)


packages/firestore/src/util/types.ts
```diff
@@ -63,20 +63,17 @@ export function isNullOrUndefined(value: unknown): boolean {
...
-    isInteger(value) &&
-    (value as number) <= MAX_SAFE_INTEGER &&
-    (value as number) >= MIN_SAFE_INTEGER
-  );
-}
-
-/**
- * Safely checks if the number is NaN.
- */
-export function safeIsNaN(value: unknown): boolean {
```
schmidt-sebastian: Note this function was never used.

schmidt-sebastian:
This doesn't seem to work with persistence. I will take a look.

packages/firestore/test/util/equality_matcher.ts
```diff
@@ -46,6 +46,12 @@ function customDeepEqual(left: unknown, right: unknown): boolean {
   /**
    * END: Custom compare logic
    */
+  if (left === 0.0 && right === 0.0) {
```
wilhuff: You could probably move this inside the `left === right` case.

```ts
if (left === right) {
  if (left === 0) {
    // ...
    return 1 / left === 1 / right;
  } else {
    return true;
  }
}
```

Also, the comment above about "END: Custom compare logic" I can't say I really understand what it means. It seems like there's custom logic both above and below the comment. Could you strike it or make it clear what it's actually talking about?

schmidt-sebastian:
So it turns out that JSON serialization does not round-trip -0:

```
const obj = { d: -0 }
JSON.parse(JSON.stringify(obj)) => {d: 0}
```

I instead opted to use a string constant for -0, which seems to work (even with the backend).

packages/firestore/src/util/types.ts
```diff
@@ -58,25 +58,22 @@ export function isNullOrUndefined(value: unknown): boolean {
   return value === null || value === undefined;
 }
 
+/** Returns whether the value represents -0.. */
```
wilhuff: Extra period.

packages/firestore/test/integration/api/type.test.ts
```diff
@@ -47,6 +47,12 @@ apiDescribe('Firestore', (persistence: boolean) => {
     });
   });
 
+  it('can read and write number fields', () => {
+    return withTestDb(persistence, db => {
+      return expectRoundtrip(db, { a: 1, b: NaN, c: Infinity, d: -0.0 });
```
wilhuff: This doesn't actually make any assertion about the format of the `d` value. Is it worth separately asserting that in the forward direction `-0.0` renders as a string `'-0'`?

packages/firestore/src/util/types.ts
schmidt-sebastian: Removed

packages/firestore/test/integration/api/type.test.ts
schmidt-sebastian: `customDeepEqual` actually does assert that d is -0 with my changes below. If I just make those changes, this test fails.

packages/firestore/test/util/equality_matcher.ts
schmidt-sebastian: Done. I was debating doing this earlier anyways.

PR #2752, opened 2020-03-17
Remove custom implementations for Number methods

schmidt-sebastian:


schmidt-sebastian:
@Feiyang1 If I understand this correctly, we don't need these Polyfills on our end?

Feiyang1:
correct. LGTM

PR #2753, opened 2020-03-17
Add small Proto helpers

schmidt-sebastian:
Small PR with some helpers that will be used in NumericIncrementTransforms and other places.


packages/firestore/src/model/values.ts
```diff
@@ -539,3 +539,43 @@ export function normalizeByteString(blob: string | Uint8Array): ByteString {
...
+}
+
+/** Returns true if `value` is a NullValue. */
+export function isNullValue(value?: api.Value | null): boolean {
+  return !!value && 'nullValue' in value;
+}
+
+/** Returns true if `value` is NaN. */
+export function isNanValue(value?: api.Value | null): boolean {
+  return !!value && isNaN(Number(value.doubleValue));
```
wilhuff: This doesn't actually test that `value` is a double. Should this be `isDouble(value) && isNaN(Number(value.doubleValue));`?

packages/firestore/src/model/values.ts
```diff
@@ -539,3 +539,43 @@ export function normalizeByteString(blob: string | Uint8Array): ByteString {
     return ByteString.fromUint8Array(blob);
   }
 }
+
+/** Returns true if `value` is an IntegerValue. */
+export function isInteger(value?: api.Value | null): boolean {
```
wilhuff: If we defined an interface like

```ts
exporter interface IntegerValue {
  integerValue: number;
}
```

Then we could make this a type guard:

```ts
export function isInteger(value?: api.Value | null): value is IntegerValue {
  return !!value && 'integerValue' in value;
}
```

This PR doesn't include any uses of these functions, so it's not clear that this would necessarily be a win, but it seems like this could be helpful.

schmidt-sebastian: I updated the PR to allow us to make asserts on the input type, but I kept it in the method declarations for now and did not plumb through separate types. This should allow us to get rid of bangs for future usages.

packages/firestore/src/model/values.ts
schmidt-sebastian: Ugh JavaScript:

```
isNaN(Number(null)) => false
isNaN(Number(undefined)) => true
```

So it looks like we need to test here that `value.doubleValue` is set.

packages/firestore/src/model/values.ts
```diff
@@ -556,26 +560,36 @@ export function isNumber(value?: api.Value | null): boolean {
...
   return !!value && 'nullValue' in value;
 }
 
 /** Returns true if `value` is NaN. */
-export function isNanValue(value?: api.Value | null): boolean {
-  return !!value && isNaN(Number(value.doubleValue));
+export function isNanValue(
+  value?: api.Value | null
+): value is { doubleValue: 'NaN' | number } {
+  return !!value && isDouble(value) && isNaN(Number(value.doubleValue));
```
wilhuff: `isDouble` already tests `!!value` so you can drop it here.

schmidt-sebastian: Done

PR #2757, opened 2020-03-17
Make Serializer tests work in browser, remove test duplication

schmidt-sebastian:
This PR removes some of the duplicated serialization tests in FieldValue and changes the tests in serializer to run in the browser. My goal was to keep the test coverage the same, so on Node it still roundtrips through ProtobufJs.


packages/firestore/test/util/node_helpers.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
wilhuff: 2020? Also Google LLC

packages/firestore/test/util/node_helpers.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const protos = loadRawProtos();
+// tslint:disable-next-line:variable-name
+const ValueMessage = protos.lookupType('google.firestore.v1.Value');
+
+/**
+ * Verifies full round-trip of JSON protos through ProtobufJs.
+ */
+export function verifyProtobufJsRoundTrip(jsonValue: api.Value): void {
+  const protobufJsEncocdedProto = ValueMessage.fromObject(jsonValue);
```
wilhuff: Typos in Encocded and Decocded.

packages/firestore/test/unit/remote/serializer.test.ts
```diff
@@ -82,24 +75,26 @@ import {
...
   wrapObject
-} from '../../../util/helpers';
-import { ByteString } from '../../../../src/util/byte_string';
+} from '../../util/helpers';
+import { ByteString } from '../../../src/util/byte_string';
+import { isNode } from '../../util/test_platform';
+
+let verifyProtobufJsRoundTrip: (jsonValue: api.Value) => void = () => {};
+
+if (isNode()) {
```
wilhuff: It seems like you could just inline the definition of `verifyProtobufJsRoundTrip` here. It doesn't seem like it would be useful in any other context because it hard codes the use of `ValueMessage`.

schmidt-sebastian: I put in a separate file so we could import ProtobufJs and its types using "import" instead of "require". This gives us type safety. If I included it here, I would have to rely on the `d.t.s` files for these dependencies to use the types safely, without importing the actual sources (an import of `ProtobufJs` here would break the build for the browser tests)..

packages/firestore/test/util/node_helpers.ts
schmidt-sebastian: Fixed

PR #2758, opened 2020-03-17
Support storage.googleapis.com in refFromUrl

### LINKED ISSUE #2625: `storage.refFromURL()` is not working correctly
### Describe your environment

  * Operating System version: Windows 10 1909
  * Browser version: Google Chrome 80.0.3987.100 
  * Firebase SDK version: 7.8.1
  * Firebase Product: storage

### Describe the problem

`firebase.storage.refFromURL()` is looking for a different storage base url than the one I got from the File information on the backend, plus the [docs ](https://firebase.google.com/docs/storage/web/download-files#create_a_reference) are not correct.

#### Steps to reproduce:

I wanted to use the `firebase.storage.refFromURL()` method to get a reference from my URL that I put together from the File information I got from firebase-admin-sdk on the backend. 
The url looks like this: `https://storage.googleapis.com/storage/v1/b/<project-id>.appspot.com/o/<file-name>`. When I try to paste this URL to my browser it displays the metadata of my public file.

This should work in my opinion however it throws an error that the URL is incorrect. I see in the [implementation ](https://github.com/firebase/firebase-js-sdk/blob/eb614f0e7bcfee5b4bc1d390f894bea3d8f3867b/packages/storage/src/implementation/location.ts#L69) that it is looking for an URL like this: `https://firebasestorage.googleapis.com/v1/b/<project-id>.appspot.com/o/<file-name>`. When I format the URL like this and paste it to my browser I got a 404 error. But of course in this case the Firebase SDK is able to get the reference from it but just becuase these values are hardcoded to the SDK.

Additionally, there is a third type of URL in the [download URL docs](https://firebase.google.com/docs/storage/web/download-files#create_a_reference). As you can see, there is no version tag there and if I format my URL like this it is not working with the SDK either. Also, if I submit this URL to my browser it return 404.

schmidt-sebastian:
This modifies refFromUrl to accept URLs generated by Cloud Storage. 

Tested that this supports URLs from the Cloud Console as well as the format specified here:

https://cloud.google.com/storage/docs/access-control/signed-urls

Fixes #2625


packages/storage/test/service.test.ts
```diff
@@ -182,6 +182,25 @@ describe('Firebase Storage > Service', () => {
...
+    it('Works with storage.googleapis.com URLs with query params', () => {
+      const ref = service.refFromURL(
+        `https://storage.googleapis.com/mybucket/path%20with%20space/image.png?X-Goog-Algorithm=
+GOOG4-RSA-SHA256`
+      );
+      assert.equal(ref.toString(), 'gs://mybucket/path with space/image.png');
+    });
+    it('Works with storage.cloud.google.com URLs', () => {
+      const ref = service.refFromURL(
+        `https://storage.cloud.google.com/mybucket/path%20with%20space/image.png`
```
fredzqm: OOC, will this generate the same result? 
        `https://storage.cloud.google.com/mybucket/path%20with%20space%2Fimage.png`

Maybe add a test.

packages/storage/src/implementation/location.ts
```diff
@@ -76,23 +76,42 @@ export class Location {
...
 
     function httpModify(loc: Location): void {
       loc.path_ = decodeURIComponent(loc.path);
     }
     const version = 'v[A-Za-z0-9_]+';
-    const hostRegex = DEFAULT_HOST.replace(/[.]/g, '\\.');
-    const httpRegex = new RegExp(
-      `^https?://${hostRegex}/${version}/b/${bucketDomain}/o${path}`,
+    const firebaseStorageHost = DEFAULT_HOST.replace(/[.]/g, '\\.');
+    const firebaseStoragePath = '(/([^?#]*).*)?$';
```
fredzqm: What's special about `#`? Trying to understand the difference between `firebaseStoragePath` and `cloudStoragePath`.

schmidt-sebastian: I added a test.

packages/storage/test/service.test.ts
schmidt-sebastian: I added a test.

schmidt-sebastian:
@Feiyang1 for approval. Thanks.

PR #2762, opened 2020-03-18
Increment master

inlined:
Add back the increment operator.

### DIFF
--- a/packages/database/src/api/Database.ts
+++ b/packages/database/src/api/Database.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


packages/database/src/api/Database.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC
```
schmidt-sebastian: Per go/copyright, we are not supposed to update copyright notices in old files.

packages/database/test/servervalues.test.ts
```diff
@@ -0,0 +1,151 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
schmidt-sebastian: 2020?

packages/database/src/api/Database.ts
hsubox76: This is the PR that causes this update: https://github.com/firebase/firebase-js-sdk/pull/2665

From the description it sounds like it is intentional that it should update changed files and I think that's what go/copyright says (retain oldest year, change Inc. to LLC)

inlined:
@schmidt-sebastian Looks like this is ready to merge?

inlined:
@Feiyang1 can you approve the version bump of the emulator? It's necessary to get the server-side changes for increment operations.

PR #2763, opened 2020-03-18
Update CI Triggers

hsubox76:
Previously:
- "Run All Tests" ran when (1) a pull_request was closed and (2) the target branch was master. This left out some cases where we would want to run all tests, such as a direct merge to master.
- "Push/PR" ran tests on changed files, but the it runs 2 tests (1 each on the push trigger and PR trigger) that are redundant in most cases, and the "push" version may not diff correctly if the branch is stale, and may run too many tests.

Now:
- "Run All Tests" workflow is triggered on push to any branch. (Canary Deploy job only runs if the branch was master.)
- "Push/PR" renamed to "Test Modified Packages" and runs only on PR events (opened, synchronized, reopened) on any branch except master. This still may happen to run all tests if a key top level file was changed.
- "Cross Browser Test Flow" only runs on push to master.

Additionally, we will change the repo settings to ensure "Test Modified Packages" is a required check and "Run All Tests" is optional, so if it runs too many tests and fails a flaky test, it does not block merge.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -1,9 +1,6 @@
-name: Run All Tests
+name: Test All Packages
 
-on: 
-  push:
-  schedule:
-    - cron: '0 6,18 * * *'
+on: push
 
 jobs:
   test:
@@ -39,7 +36,7 @@ jobs:
   deploy:
     name: Canary Deploy
     runs-on: ubuntu-latest
-    if: github.event.pull_request.merged
+    if: github.ref == 'refs/heads/master'
     needs: test
 
     steps:
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -1,11 +1,7 @@
 name: Run All Tests
 
 on: 
-  pull_request:
-    branches:
-      - master
-    types:
-      - closed
+  push:
   schedule:
     - cron: '0 6,18 * * *'
 
--- a/.github/workflows/test-changed.yml
+++ b/.github/workflows/test-changed.yml
@@ -1,6 +1,10 @@
-name: Push/PR
+name: Test Modified Packages
 
-on: [push, pull_request]
+on: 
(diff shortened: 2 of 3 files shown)


.github/workflows/test-all.yml
```diff
@@ -1,11 +1,7 @@
 name: Run All Tests
```
yifanyang: Since the other workflow is renamed to "Test Modified Packages", how about also renaming this one to "Test All Packages"?

Feiyang1: +1

.github/workflows/test-all.yml
```diff
@@ -1,11 +1,7 @@
...
 
 on: 
-  pull_request:
-    branches:
-      - master
-    types:
-      - closed
+  push:
   schedule:
     - cron: '0 6,18 * * *'
```
Feiyang1: Do we still need the cron job?

.github/workflows/test-changed.yml
```diff
@@ -1,6 +1,10 @@
-name: Push/PR
+name: Test Modified Packages
 
-on: [push, pull_request]
+on: 
+  pull_request:
+    branches:
+      - '*'
+      - '!master'
```
Feiyang1: What does excluding master for PR mean? Don't we always create PR against master?

.github/workflows/test-all.yml
hsubox76: I guess it hasn't been very helpful. Removing.

.github/workflows/test-changed.yml
hsubox76: Yes, I think I was confused about what "branches" means for PR. Fixing.

.github/workflows/test-all.yml
hsubox76: Done.

PR #2765, opened 2020-03-19
exclude typedoc from dep update

Feiyang1:
Our doc generation process assumes the location and format where typedoc outputs files to. Updating typedoc has broken the process and has introduced a lot noises in the past, so we will disable its update.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1000,6 +1000,20 @@
     unique-filename "^1.1.1"
     which "^1.3.1"
 
+"@firebase/firestore@1.12.1":
+  version "1.12.1"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -1000,6 +1000,20 @@
     unique-filename "^1.1.1"
     which "^1.3.1"
 
+"@firebase/firestore@1.12.1":
```
hsubox76: Should `@firebase/firestore` be here?

Feiyang1: Good question. I don't know why it shows up here.
Yarn shows It's pulled in by `workspace-aggregator-ce563d83-8dd9-4c27-b9d6-f5e2a2c4a6bd`. Looks like a generated thing, but I couldn't figure out where it is coming from. I cleaned workspace and `yarn`, still the same thing. 

For now I manually removed firestore from yarn.lock.

PR #2766, opened 2020-03-19
Finish FieldValue migration

schmidt-sebastian:
This finishes up the field value migration.
- It ensures that we always only use one setting for `useProto3Json` by adding it to Platform and plumbing it through to the serializer.
- It removes the optimization to compare timestamps of any length as string, since it breaks equality (eg 12:00:00.000 should equal 12:00:00)
- It moves some of the numeric comparison code to values.ts since that's the only callsite.
- Many other things that I forgot to mention.

Note: This PR is against a branch that does nothing but delete the old field_value.ts and replace it with proto_field_value.ts

SDK Size Before:
317762 firebase-firestore.js
274464 firebase-firestore.memory.js

SDK Size After:
316174 firebase-firestore.js
272668 firebase-firestore.memory.js


packages/firestore/src/api/user_data_writer.ts
```diff
@@ -17,99 +17,112 @@
...
-      return this.convertTimestamp(value.value());
-    } else if (value instanceof ServerTimestampValue) {
-      return this.convertServerTimestamp(value);
+  convertValue(value: api.Value): unknown {
+    if ('nullValue' in value) {
+      return null;
+    } else if ('booleanValue' in value) {
+      return value.booleanValue!;
+    } else if ('integerValue' in value) {
+      return Number(value.integerValue!);
```
wilhuff: This seems like it should call `normalizeNumber` rather than just assuming it knows how to do this.

packages/firestore/src/api/user_data_writer.ts
```diff
@@ -17,99 +17,112 @@
...
     if (this.timestampsInSnapshots) {
       return timestamp;
     } else {
       return timestamp.toDate();
     }
   }
 
-  private convertReference(value: RefValue): DocumentReference<T> {
-    const key = value.value();
-    const database = this.firestore.ensureClientConfigured().databaseId();
```
wilhuff: Losing this error message seems like a bad idea. Developers won't have any indication that this isn't going to work without this, no?

packages/firestore/src/model/field_value.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2020 Google Inc.
+ * Copyright 2017 Google Inc.
```
wilhuff: Google LLC. See go/copyright

packages/firestore/src/model/transform_operation.ts
```diff
@@ -186,35 +183,32 @@ export class ArrayRemoveTransformOperation implements TransformOperation {
...
-    ) {
-      const sum = baseValue.internalValue + this.operand.internalValue;
-      return new IntegerValue(sum);
-    } else {
-      const sum = baseValue.internalValue + this.operand.internalValue;
-      return new DoubleValue(sum);
-    }
+    const baseValue = this.computeBaseValue(previousValue);
+    const sum = this.asNumber(baseValue) + this.asNumber(this.operand);
+    return this.serializer.toNumber(sum);
```
wilhuff: Does this give the same result for all possible integer values? It seems like if you add to a sum that's larger than `isSafeInteger(sum)` this will now return a `doubleValue` where it used to be an `integerValue`. Though the calculation lost precision, I'm not sure changing the type is appropriate.

packages/firestore/src/model/transform_operation.ts
```diff
@@ -224,27 +218,24 @@ export class NumericIncrementTransformOperation implements TransformOperation {
...
-  } else {
-    // coerce to empty array.
-    return [];
+  private asNumber(value: api.Value): number {
+    return normalizeNumber(value.integerValue || value.doubleValue);
   }
 }
+
+function coercedFieldValuesArray(value: api.Value | null): api.Value[] {
+  return value?.arrayValue?.values?.slice() || [];
```
wilhuff:
```ts
return isArray(value)
  ? value.arrayValue.values!.slice()
  : [];
```

Will likely result in less generated code.

packages/firestore/src/model/values.ts
```diff
@@ -201,7 +210,24 @@ export function compare(left: api.Value, right: api.Value): number {
...
-      return compareTimestamps(left.timestampValue!, right.timestampValue!);
+      if (isServerTimestamp(left)) {
+        if (isServerTimestamp(right)) {
+          return compareTimestamps(
+            getLocalWriteTime(left),
+            getLocalWriteTime(right)
+          );
+        } else {
+          // Server timestamps come after all concrete timestamps.
+          return 1;
```
wilhuff: You know, if we just made `ServerTimestampValue` its own entry in the `TypeOrder` enum we wouldn't have to special case this at all: any Timestamp to ServerTimestamp comparison would be handled automatically by the way we handle differing types.

packages/firestore/src/model/values.ts
```diff
@@ -185,6 +189,11 @@ function objectEquals(left: api.Value, right: api.Value): boolean {
   return true;
 }
 
+/** Returns true if the Value list contains the specified element. */
+export function contains(haystack: api.ArrayValue, needle: api.Value): boolean {
+  return (haystack.values || []).find(v => equals(v, needle)) !== undefined;
+}
+
 export function compare(left: api.Value, right: api.Value): number {
```
wilhuff: Reading through calling code I fear that names like `compare` and `equals` are too generic for free functions that don't operate on general arguments.

What do you think of naming these functions with a type-operation naming scheme? This would be `valueCompare`, etc.

WDYT?

packages/firestore/src/model/values.ts
```diff
@@ -137,34 +153,22 @@ function blobEquals(left: api.Value, right: api.Value): boolean {
...
-  }
-
-  for (let i = 0; i < leftArray.length; ++i) {
-    if (!equals(leftArray[i], rightArray[i])) {
-      return false;
+    if (n1 === n2) {
+      return isNegativeZero(n1) === isNegativeZero(n2);
+    } else {
+      // NaN == NaN
+      return n1 !== n1 && n2 !== n2;
```
wilhuff: Elsewhere in this file you're calling `isNaN`. Make these consistent?

packages/firestore/src/model/values.ts
schmidt-sebastian: Yeah, we can do that. I looked into removing the namespace import for our logging already, and `logWarn` was what I picked there as a temporary solution.

packages/firestore/src/api/user_data_writer.ts
schmidt-sebastian: I added this back. I was trying to limit the parsing code for references, since it is already pretty redundant in our code.

packages/firestore/src/model/values.ts
schmidt-sebastian: That's @mikelehen level genius. I love it.

packages/firestore/src/model/values.ts
```diff
@@ -578,9 +559,23 @@ export function refValue(databaseId: DatabaseId, key: DocumentKey): api.Value {
...
 
+/** Returns true if `value` is either an IntegerValue . */
+export function isInteger(
+  value?: api.Value | null
+): value is { integerValue: string | number } {
+  return !!value && 'integerValue' in value;
+}
+
+/** Returns true if `value` is either an IntegerValue . */
+export function isDobule(
```
wilhuff: s/isDobule/isDouble/

Also, comment is out of date.

packages/firestore/src/model/values.ts
```diff
@@ -578,9 +559,23 @@ export function refValue(databaseId: DatabaseId, key: DocumentKey): api.Value {
   };
 }
 
+/** Returns true if `value` is either an IntegerValue . */
```
wilhuff: Remove "either" from the comment.

packages/firestore/src/model/values.ts
schmidt-sebastian: Fixed (also renamed contains to arrayValueContains)

packages/firestore/src/model/values.ts
schmidt-sebastian: Fixed

packages/firestore/src/model/values.ts
mikelehen: WHAT DID I DO!?!?!?

Hi guys!

PR #2767, opened 2020-03-19
Make MemoryLru tree-shakeable

schmidt-sebastian:
Our hard-coded dependency means that tree-shaking doesn't remove the unused LRU code for memory persistence. This PR removes the dependency and should also completely remove the estimated byte size accounting in the field value rewrite.

Before:
317762  firebase-firestore.js
274464  firebase-firestore.memory.js

After:
315861  firebase-firestore.js
270591  firebase-firestore.memory.js

### DIFF
--- a/packages/firestore/src/local/memory_persistence.ts
+++ b/packages/firestore/src/local/memory_persistence.ts
@@ -87,34 +87,17 @@ export class MemoryPersistence implements Persistence {
 
   private _started = false;
 
-  readonly referenceDelegate: MemoryLruDelegate | MemoryEagerDelegate;
-
-  static createLruPersistence(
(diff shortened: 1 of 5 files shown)


schmidt-sebastian:
My five second summary of tree shaking: Our build pipeline drops unused imports and top-level exports. So if you have:

```
class A {}
class B {}
const b = B();
```

`A` won't be in the final bundle. If instead, you have:

```
class A {}
class B {  
 function unused() {
  return new A();
 }
}
const b = B();
```

RollUp doesn't know whether you are actually using `B.unused()` and hence `A` (since there is no guarantee for strong typings). In this snippet, `A` and `B` are included in the final bundle.

PR #2776, opened 2020-03-20
remove comments and whitespaces in builds

Feiyang1:
In preparation to enable size presubmit(https://github.com/firebase/firebase-js-sdk/pull/2744), this PR removes comments and whitespaces in our builds, so the size diffs are not skewed by comment changes.

Will wait for the CI to finish to see how much latency this change introduces.

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -38,7 +38,7 @@
     "rollup-plugin-json": "4.0.0",
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-typescript2": "0.26.0",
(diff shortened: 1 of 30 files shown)


schmidt-sebastian:
Would it be possible to do this a precursor only to the size computation? The comments and whitespace actually help a lot when debugging issues in these builds.

Feiyang1:
Possibly. I will look into that.

Feiyang1:
@schmidt-sebastian replaced by https://github.com/firebase/firebase-js-sdk/pull/2777

PR #2779, opened 2020-03-21
Remove mangling from Node build

schmidt-sebastian:
#2608 re-introduced mangling to the Node build, which wasn't mangled before. Since it uses CJS, the output is not minified very well. This reverts this change and makes sure the next release uses the same settings as the current release.


schmidt-sebastian:
See https://github.com/firebase/firebase-js-sdk/commit/fe06e91f5b1721db2e1fa8e0f001a988dbb0484f#diff-5701bafe0ac90809cab8ff16ba96ded1R72 for reference

PR #2799, opened 2020-03-25
Add a script to build only the SDK and its dependencies

Feiyang1:
Today, to work on a SDK, we usually run `yarn build` at the root folder first in order to build its dependencies. This step is quite slow because it is building the entire monorepo including repos that are not dependencies of the SDK.

This PR addresses the issue by adding a new npm script `build:deps` to each SDK, which builds the SDK and its dependencies in one command.

After this change, you don't need to run `yarn build` at the root folder any more. Instead, you would directly cd into your SDK directory and run `yarn build:deps`.

Build time improvement
`yarn build` at root -  234.25s
`yarn build:deps` in `pacakges/firestore` - 124.01s
`yarn build:deps` in `pacakges/database` - 33.34s

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "prebuild": "tsc -d --downlevelIteration --declarationDir dist/lib --emitDeclarationOnly && tsc -m es2015 --moduleResolution node scripts/*.ts ",
     "build": "rollup -c",
-    "build:deps": "lerna run --scope @firebase/firestore --include-dependencies build",
+    "build:deps": "lerna run --scope @firebase/'{app,firestore}' --include-dependencies build",
     "build:console": "node tools/console.build.js",
     "dev": "rollup -c -w",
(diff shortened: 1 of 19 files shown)


packages/firestore/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "prebuild": "tsc -d --downlevelIteration --declarationDir dist/lib --emitDeclarationOnly && tsc -m es2015 --moduleResolution node scripts/*.ts ",
     "build": "rollup -c",
+    "build:deps": "lerna run --scope @firebase/'{app,firestore}' --include-dependencies build",
```
wilhuff: I'm surprised that `app` isn't picked up by `--include-dependencies`. Does it not get picked up because it's a peer dependency?

Feiyang1: Yup. Will have to put it in the devDepedencies for lerna to pick it up, but it has an undesired side effect that it triggers Firestore version bump when `app` changes.

PR #2800, opened 2020-03-25
Use Dynamic Measurement ID in Analytics

hsubox76:
Switch Analytics to using dynamic measurement ID fetched from REST endpoint. Previously it was using  the ID provided in the user's config object. Measurement ID may change so we are switching to fetching it dynamically.


packages/analytics/src/helpers.ts
```diff
@@ -95,7 +92,9 @@ export function getOrCreateDataLayer(dataLayerName: string): DataLayer {
  */
 function wrapGtag(
```
Feiyang1: This function is getting pretty big and I found it a little hard to follow. An idea is to break it into 3 functions, one for each command type.

Also I think it will make the code more readable if you use await async instead of chaining promises. The change in return type should not matter as the caller can choose to ignore it.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '@firebase/app-types';
+import { DynamicConfig } from '@firebase/analytics-types';
+
+const DYNAMIC_CONFIG_URL =
+  'https://firebase.googleapis.com/v1alpha/projects/-/apps/{app-id}/webConfig';
+
+export function getHeaders(apiKey: string): Headers {
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/analytics/src/helpers.ts
hsubox76: Done.

packages/analytics/src/helpers.ts
```diff
@@ -95,33 +57,86 @@ export function getOrCreateDataLayer(dataLayerName: string): DataLayer {
...
-  return (
-    command: 'config' | 'set' | 'event',
-    idOrNameOrParams: string | ControlParams,
+  /**
+   * Wrapped gtag logic when gtag is called with 'config' command.
+   *
+   * @param measurementId GA Measurement ID.
+   * @param gtagParams Config params.
+   */
+  async function gtagOnConfig(
```
Feiyang1, line 70: Not a big fan of nested functions. They provides some level of code separation, but `wrapGtag()` is still big. Can you move them to the top level?

hsubox76: Done.

packages/analytics/src/get-config.test.ts
```diff
@@ -0,0 +1,192 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/get-config.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+      });
+      const app = getFakeApp(fakeAppParams);
+      let error;
+      try {
+        await fetchDynamicConfig(app);
+      } catch (e) {
+        error = e;
+      }
+      expect(error.message).to.include(AnalyticsError.CONFIG_FETCH_FAILED);
```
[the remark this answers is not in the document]
hsubox76: I think in that case if the catch block isn't reached, the test will "pass", since it does not hit the `expect` statement. I tried a more compact form, `to.be.rejectedWith`.

packages/analytics/src/factory.ts
```diff
@@ -28,21 +29,31 @@ import {
...
 import { FirebaseInstallations } from '@firebase/installations-types';
+import { initializeIds } from './initialize-ids';
+import { logger } from './logger';
 
 /**
- * Maps gaId to FID fetch promises.
+ * Maps appId to full initialization promise.
  */
-let initializedIdPromisesMap: { [gaId: string]: Promise<void> } = {};
+let initializationPromisesMap: { [appId: string]: Promise<string> } = {};
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/functions.ts
```diff
@@ -32,17 +32,24 @@ import { GtagCommand } from './constants';
...
-): void {
-  let params: EventParams | ControlParams = eventParams || {};
-  if (!options || !options.global) {
-    params = { ...eventParams, 'send_to': analyticsId };
+): Promise<void> {
+  if (options && options.global) {
+    gtagFunction(GtagCommand.EVENT, eventName, eventParams || {});
+    return Promise.resolve();
+  } else {
+    return initializationPromise.then(measurementId => {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/errors.ts
```diff
@@ -16,35 +16,50 @@
...
   [AnalyticsError.ALREADY_INITIALIZED]:
     'Firebase Analytics has already been initialized.' +
     'settings() must be called before initializing any Analytics instance' +
     'or it will have no effect.',
   [AnalyticsError.INTEROP_COMPONENT_REG_FAILED]:
-    'Firebase Analytics Interop Component failed to instantiate'
+    'Firebase Analytics Interop Component failed to instantiate',
+  [AnalyticsError.FETCH_THROTTLE]:
+    'The config fetch request timed out while in an exponential backoff state.' +
+    ' Configure timeout using "fetchTimeoutMillis" SDK setting.' +
```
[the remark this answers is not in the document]
hsubox76: Removed line.

packages/analytics/src/exponential_backoff.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
hsubox76: Deleted file, using shared file.

packages/analytics/index.test.ts
```diff
@@ -36,23 +36,40 @@ import { removeGtagScript } from './testing/gtag-script-util';
...
+const fakeMeasurementId = 'abcd-efgh';
+const fakeAppParams = { appId: 'abcdefgh12345:23405', apiKey: 'AAbbCCdd12345' };
 const gtagStub: SinonStub = stub();
+let fetchStub: SinonStub = stub();
 const customGtagName = 'customGtag';
 const customDataLayerName = 'customDataLayer';
 
+function stubFetch(status: number, body: object): void {
+  fetchStub = stub(window, 'fetch');
+  const mockResponse = new window.Response(JSON.stringify(body), {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/exponential_backoff.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** @fileoverview
+ * Copied from remote-config.
+ * TODO: Created shared utils to use between both packages.
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/remote-config/test/remote_config.test.ts
```diff
@@ -58,7 +58,7 @@ describe('RemoteConfig', () => {
 
   let getActiveConfigStub: sinon.SinonStub;
   let loggerDebugSpy: sinon.SinonSpy;
-  let loggerLogLevelSpy: sinon.SinonSpy;
+  let loggerLogLevelSpy: any;
```
Feiyang1, line 61: Are we not applying eslint in test files?

packages/analytics/src/get-config.ts
```diff
@@ -20,18 +20,41 @@
...
+  appId: string;
+  apiKey: string;
+  measurementId?: string;
+}
+
+/**
+ * Initial retry interval in seconds for 503 errors, which we want to be conservative about
+ * to avoid overloading servers.
+ */
+export const LONG_RETRY_FACTOR = 30;
```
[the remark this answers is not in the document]
hsubox76: It's not actually seconds, it just comes out to it because the default interval is 1 second. It's the base for the exponent in `calculateBackoffMillis`. I'll change the comment to be more accurate.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    // Increments backoff state.
+    const throttleMetadata = {
+      throttleEndTimeMillis: Date.now() + backoffMillis,
+      backoffCount: backoffCount + 1
+    };
+
+    // Persists state.
+    retryData.setThrottleMetadata(appId, throttleMetadata);
+    logger.debug(`Calling attemptFetch again in ${backoffMillis} millis`);
```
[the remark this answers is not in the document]
hsubox76: I didn't know that, I thought debug was something we could ask a user to turn on if we wanted to help them debug something. Or is that `info`?

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,319 @@
...
+
+// App config fields needed by analytics.
+export interface AppFields {
+  appId: string;
+  apiKey: string;
+  measurementId?: string;
+}
+
+/**
+ * Initial retry interval in seconds for 503 errors, which we want to be conservative about
```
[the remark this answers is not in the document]
hsubox76: Got it (https://github.com/firebase/firebase-js-sdk/pull/2800#discussion_r459649426)

packages/analytics/index.ts
```diff
@@ -130,3 +118,16 @@ async function isSupported(): Promise<boolean> {
...
   }
 }
+
+/**
+ * Define extension behavior of `registerAnalytics`
+ */
+declare module '@firebase/app-types' {
+  interface FirebaseNamespace {
+    analytics(app?: FirebaseApp): FirebaseAnalytics;
+    isSupported(): Promise<boolean>;
```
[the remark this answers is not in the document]
hsubox76: I think I overlooked this in the code review: https://github.com/firebase/firebase-js-sdk/pull/3165

Will remove.

packages/analytics/src/functions.ts
```diff
@@ -30,23 +30,27 @@ import { GtagCommand } from './constants';
...
+  if (options && options.global) {
+    gtagFunction(GtagCommand.EVENT, eventName, eventParams || {});
+    return;
+  } else {
+    const measurementId = await initializationPromise;
+    const params: EventParams | ControlParams = {
+      ...eventParams,
+      'send_to': measurementId
+    };
+    // Workaround for http://b/141370449 - third argument cannot be undefined.
```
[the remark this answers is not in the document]
hsubox76: Removed comment and both `|| {}`s

packages/analytics/src/functions.ts
```diff
@@ -30,23 +30,27 @@ import { GtagCommand } from './constants';
  * @param eventName Google Analytics event name, choose from standard list or use a custom string.
  * @param eventParams Analytics event parameters.
  */
-export function logEvent(
+export async function logEvent(
```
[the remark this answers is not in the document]
hsubox76: Does it make a difference as they all have to return a promise anyway? Changed them all for consistency in any case.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,319 @@
...
+    let errorMessage = '';
+    try {
+      // Try to get any error message text from server response.
+      const jsonResponse = (await response.json()) as {
+        error?: { message?: string };
+      };
+      if (jsonResponse.error?.message) {
+        errorMessage = jsonResponse.error.message;
+      }
+    } catch (ignored) {}
```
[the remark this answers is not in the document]
hsubox76: Added _, I don't think eslint allows removing the param.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,319 @@
...
+  return response.json();
+}
+
+/**
+ * Fetches dynamic config from backend, retrying if failed.
+ * @param app Firebase app to fetch config for.
+ */
+export async function fetchDynamicConfigWithRetry(
+  app: FirebaseApp,
+  retryData: RetryData = defaultRetryData, // for testing
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/helpers.ts
```diff
@@ -85,6 +49,113 @@ export function getOrCreateDataLayer(dataLayerName: string): DataLayer {
   }
   return dataLayer;
 }
+/**
+ * Wrapped gtag logic when gtag is called with 'config' command.
+ *
+ * @param measurementId GA Measurement ID.
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/helpers.ts
```diff
@@ -94,75 +165,53 @@ export function getOrCreateDataLayer(dataLayerName: string): DataLayer {
  */
 function wrapGtag(
   gtagCore: Gtag,
-  initializedIdPromisesMap: { [gaId: string]: Promise<void> }
+  initializationPromisesMap: { [appId: string]: Promise<string> },
+  dynamicConfigPromisesList: Array<
```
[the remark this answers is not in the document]
hsubox76: Updated JSdoc comments on all functions in this file.
Added param comments on `wrapGtag`.

packages/analytics/src/factory.ts
```diff
@@ -28,26 +30,40 @@ import {
...
- * Maps gaId to FID fetch promises.
+ * Maps appId to full initialization promise.
  */
-let initializedIdPromisesMap: { [gaId: string]: Promise<void> } = {};
+let initializationPromisesMap: {
+  [appId: string]: Promise<string>; // Promise contains measurement ID string.
+} = {};
+
+/**
+ * List of dynamic config fetch promises.
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/remote-config/test/remote_config.test.ts
hsubox76: `no-explicit-any` is always off in test files: https://github.com/firebase/firebase-js-sdk/blob/e1173544b2f63691e887622b827bcc819e743293/config/.eslintrc.js#L15

which was added in https://github.com/firebase/firebase-js-sdk/pull/1845

Which I definitely like because setting this type to `any` seemed like it was the only way I could make this spy work.

packages/analytics/src/get-config.ts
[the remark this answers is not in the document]
Feiyang1: Ah that makes sense.

packages/analytics/package.json
```diff
@@ -37,7 +39,7 @@
     "rollup-plugin-commonjs": "10.1.0",
     "rollup-plugin-json": "4.0.0",
     "rollup-plugin-node-resolve": "5.2.0",
-    "rollup-plugin-typescript2": "0.27.1",
+    "rollup-plugin-typescript2": "0.27.0",
```
Feiyang1, line 42: We should upgrade to 0.27.1 which is the current version.

packages/analytics/src/functions.ts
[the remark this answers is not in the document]
Feiyang1: No, it is only about consistency. You changed the signature of the functions to async but still use `then` instead of `await`. We should change them to use `await`

packages/analytics/src/get-config.ts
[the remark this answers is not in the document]
Feiyang1: Hmm yes, but I think we ought to stop doing that in the prod build and provide a special debug build for debugging.   Otherwise the debug messages are mostly dead code. It's probably worth discussing with other teams before changing it. So let's keep it for now.

packages/analytics/src/functions.ts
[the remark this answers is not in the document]
hsubox76: Updated.

packages/analytics/package.json
hsubox76: Something always seems to go wrong with the rebasing. Fixed.

packages/analytics/src/initialize-ids.ts
```diff
@@ -0,0 +1,92 @@
...
+      measurementIdToAppId[config.measurementId] = config.appId;
+      if (
+        app.options.measurementId &&
+        config.measurementId !== app.options.measurementId
+      ) {
+        logger.warn(
+          `Measurement ID in local firebase config (${app.options.measurementId})` +
+            ` does not match measurement ID fetched from server (${config.measurementId}).` +
+            ` To avoid analytics events being sent to the wrong measurement ID, update the` +
+            ` measurement ID field in the local config or remove it.`
```
rachelsaunders, line 64:
```suggestion
          `The measurement ID in the local Firebase config (${app.options.measurementId})` +
            ` does not match the measurement ID fetched from the server (${config.measurementId}).` +
            ` To avoid analytics events being sent to the wrong measurement ID, update the` +
            ` measurement ID field in the local config or remove it from the local config.`
```

packages/analytics/src/errors.ts
```diff
@@ -16,33 +16,42 @@
...
-    'Firebase Analytics Interop Component failed to instantiate',
+    'Firebase Analytics Interop Component failed to instantiate: {$reason}',
+  [AnalyticsError.FETCH_THROTTLE]:
+    'The config fetch request timed out while in an exponential backoff state.' +
+    ' Unix timestamp in milliseconds when fetch request throttling ends: {$throttleEndTimeMillis}.',
+  [AnalyticsError.CONFIG_FETCH_FAILED]:
+    'Dynamic config fetch failed: [{$httpStatus}] {$responseMessage}',
+  [AnalyticsError.NO_API_KEY]:
+    '"apiKey" field is empty in Firebase config. Firebase Analytics requires this field to' +
+    'contain a valid API key.',
```
rachelsaunders, line 51:
```suggestion
    'The "apiKey" field is empty in the local Firebase config. Firebase Analytics requires this field to' +
    'contain a valid API key.',
```

packages/analytics/src/errors.ts
```diff
@@ -16,33 +16,42 @@
...
+  [AnalyticsError.NO_APP_ID]:
+    '"appId" field is empty in Firebase config. Firebase Analytics requires this field to' +
+    'contain a valid app ID.',
```
rachelsaunders, line 54:
```suggestion
    'The "appId" field is empty in the local Firebase config. Firebase Analytics requires this field to' +
    'contain a valid app ID.',
```

packages/analytics/src/factory.ts
```diff
@@ -122,6 +151,26 @@ export function factory(
...
+  const appId = app.options.appId;
+  if (!appId) {
+    throw ERROR_FACTORY.create(AnalyticsError.NO_APP_ID);
+  }
+  if (!app.options.apiKey) {
+    if (app.options.measurementId) {
+      logger.warn(
+        `"apiKey" field is empty in Firebase config. This is needed to fetch the latest` +
+          ` measurement id for this Firebase app. Falling back to measurement id ${app.options.measurementId}` +
+          ` provided in "measurementId" field.`
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,322 @@
...
+  // Ensures the throttle end time is honored if the last attempt timed out.
+  // Note the SDK will never make a request if the fetch timeout expires at this point.
+  try {
+    await setAbortableTimeout(signal, throttleEndTimeMillis);
+  } catch (e) {
+    if (measurementId) {
+      logger.warn(
+        `Timed out fetching this Firebase app's measurement id from the server.` +
+          ` Falling back to measurement id ${measurementId}` +
+          ` provided in "measurementId" field. [${e.message}]`
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/get-config.ts
```diff
@@ -0,0 +1,322 @@
...
+
+    return response;
+  } catch (e) {
+    if (!isRetriableError(e)) {
+      retryData.deleteThrottleMetadata(appId);
+      if (measurementId) {
+        logger.warn(
+          `Failed to fetch this Firebase app's measurement id from the server.` +
+            ` Falling back to measurement id ${measurementId}` +
+            ` provided in "measurementId" field. [${e.message}]`
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/analytics/src/initialize-ids.ts
hsubox76: Done.

packages/analytics/src/errors.ts
hsubox76: Done.

packages/analytics/src/errors.ts
hsubox76: Done.

packages/analytics/src/initialize-ids.ts
```diff
@@ -0,0 +1,92 @@
...
+    .then(config => {
+      measurementIdToAppId[config.measurementId] = config.appId;
+      if (
+        app.options.measurementId &&
+        config.measurementId !== app.options.measurementId
+      ) {
+        logger.warn(
+          `The measurement ID in the local Firebase config (${app.options.measurementId})` +
+            ` does not match the measurement ID fetched from the server (${config.measurementId}).` +
+            ` To avoid analytics events being sent to the wrong measurement ID, update the` +
```
rachelsaunders, line 63: Question - is it that the events would be sent to the "wrong Analytics property", rather than the "wrong measurement ID".  IIUC, the measurement ID is how the Web App is associated with its Analytics property, i.e., the actual entity that logs the events is the "Analytics property".  Can anybody confirm this?

hsubox76: It would be the wrong property. I suppose it could also be no property if it's an invalid measurement ID so I guess I was just trying to be very literal and not go any farther than saying it would make the `gtag` call using the wrong measurement ID as a parameter, which could cause it to either be sent to the wrong property or no property at all if it's not a valid measurement ID (the previous expired one?).

I'm not sure if that distinction is important but I guess I imagined people not finding it in any of their properties and then saying they were misled or something.

rachelsaunders: Maybe we can flip this statement and say:

To ensure analytics events are always sent to the correct Analytics property, update....

hsubox76: Sounds good! Changed.

packages/analytics/index.test.ts
```diff
@@ -34,51 +34,144 @@ import { GtagCommand, EventName } from './src/constants';
...
+    });
+    it('Throws if no apiKey or measurementId in config', () => {
+      const app = getFakeApp({ appId: fakeAppParams.appId });
+      const installations = getFakeInstallations();
+      expect(() => analyticsFactory(app, installations)).to.throw(
+        AnalyticsError.NO_API_KEY
+      );
+    });
+    it('Warns if config has no apiKey but does have a measurementId', () => {
+      stubFetch(200, { measurementId: fakeMeasurementId });
```
Feiyang1, line 73: Do you need to stub fetch if there is no apiKey? 
In the expect statement later, can we match the full error string (or match a longer string) to make it easier to understand? e.g. `apiKey is not available, falling back to 'fakeMeasurementId'`

packages/analytics/index.test.ts
```diff
@@ -34,51 +34,144 @@ import { GtagCommand, EventName } from './src/constants';
...
+      await clock.runAllAsync();
+      expect(warnStub.args[0][1]).to.include(
+        AnalyticsError.INVALID_INDEXED_DB_CONTEXT
+      );
+      expect(warnStub.args[0][1]).to.include('idb open throw message');
+      warnStub.restore();
+      idbOpenStub.restore();
+      fetchStub.restore();
+    });
+    it('Throws if creating an instance with already-used analytics ID', () => {
```
Feiyang1, line 144: Is it more accurate to say `already-used appId`, since we are checking if the appId is already initialized?

packages/analytics/index.test.ts
```diff
@@ -179,20 +272,22 @@ describe('FirebaseAnalytics instance tests', () => {
   });
 
   describe('Page has no existing gtag script or dataLayer', () => {
-    before(() => {
+    before(async () => {
```
Feiyang1, line 275: What's the need for making it async? There doesn't seem to be any async code inside.

packages/analytics/src/factory.ts
```diff
@@ -95,10 +124,14 @@ export function resetGlobalVars(
  * For testing
  */
 export function getGlobalVars(): {
-  initializedIdPromisesMap: { [gaId: string]: Promise<void> };
+  initializationPromisesMap: { [gaId: string]: Promise<string> };
```
Feiyang1, line 127: /gaId/appId

packages/analytics/index.test.ts
```diff
@@ -179,20 +272,22 @@ describe('FirebaseAnalytics instance tests', () => {
...
       delete window['gtag'];
       delete window['dataLayer'];
       removeGtagScript();
+      fetchStub.restore();
     });
     it('Adds the script tag to the page', async () => {
-      const { initializedIdPromisesMap } = getGlobalVars();
-      await initializedIdPromisesMap[analyticsId];
+      const { initializationPromisesMap } = getGlobalVars();
+      await initializationPromisesMap[fakeMeasurementId];
```
Feiyang1, line 290: Isn't the key for `initializationPromisesMap` appId? Using `fakeMeasurementId` seems wrong if it's the case.

packages/analytics/src/factory.ts
```diff
@@ -44,11 +44,38 @@ import {
   areCookiesEnabled,
   isBrowserExtension
 } from '@firebase/util';
+import { initializeIds } from './initialize-ids';
+import { logger } from './logger';
+
+interface FirebaseAnalyticsInternal extends FirebaseAnalytics {
```
Feiyang1, line 50: You should implement this [interface](https://github.com/firebase/firebase-js-sdk/blob/master/packages/app-types/private.d.ts#L38).

Feiyang1 (changes requested):

packages/analytics/index.test.ts
hsubox76: Done.

packages/analytics/index.test.ts
hsubox76: Done.

packages/analytics/index.test.ts
hsubox76: Added a comment explaining why for each of these, let me know if that seems reasonable or if I should instead move `await clock.runAllAsync()` into the `before()` block. The only reason I didn't is I wanted `logEvent` to be called before the factory promise chain necessarily resolved, like in real usage.

packages/analytics/src/factory.ts
hsubox76: Done.

packages/analytics/index.test.ts
hsubox76: Ack, yes, fixed.

packages/analytics/src/factory.ts
hsubox76: Changed it to extend both, since it is implemented as a POJO and not a class, is that ok?

.changeset/witty-deers-study.md
```diff
@@ -0,0 +1,10 @@
+---
+'@firebase/analytics': minor
+'@firebase/analytics-types': minor
+'@firebase/remote-config': patch
```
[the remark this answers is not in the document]
hsubox76: Done.

hsubox76:
Oops, looks like I have to change the changeset checker to account for multiple changeset files.

Feiyang1 (approved):

PR #2806, opened 2020-03-26
Simplify logging

schmidt-sebastian:
Small cleanup for our logging.

The main reason for going down this path is that I am trying to remove namespace imports and I don't want to refer to `error` and `debug` as free-standing functions.

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -68,10 +68,7 @@ import {
   validateStringEnum,
   valueDescription
 } from '../util/input_validation';
-// eslint-disable-next-line import/no-duplicates
-import * as log from '../util/log';
-// eslint-disable-next-line import/no-duplicates
-import { LogLevel } from '../util/log';
+import { DEBUG, ERROR, log, setLogLevel, SILENT } from '../util/log';
 import { AutoId } from '../util/misc';
 import * as objUtils from '../util/obj';
 import { Deferred, Rejecter, Resolver } from '../util/promise';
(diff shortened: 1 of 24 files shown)


packages/firestore/src/util/log.ts
```diff
@@ -15,58 +15,27 @@
...
-      break;
-    default:
-      logClient.error(
-        `Firestore (${SDK_VERSION}): Invalid value passed to \`setLogLevel\``
-      );
-  }
+  return logClient.logLevel;
 }
 
-export function debug(tag: string, msg: string, ...obj: unknown[]): void {
```
wilhuff: This was better as two separate functions--I don't think importing `DEBUG` is any better than importing `debug`.

I do agree that we should prefer not to import `debug` and `error` but I think a better way would be to just rename these to `logDebug` and `logError`. This would also serve to match iOS (which has `LOG_DEBUG` and `LOG_ERROR` macros) and still loosely match Android (which has `log.debug` and `log.error`).

This means that in the common case there's only one thing to import from here as well, where now every call site needs two.

packages/firestore/src/util/log.ts
```diff
@@ -15,58 +15,27 @@
  * limitations under the License.
  */
 
-import { Logger, LogLevel as FirebaseLogLevel } from '@firebase/logger';
+import { Logger, LogLevel } from '@firebase/logger';
 import { SDK_VERSION } from '../core/version';
 import { PlatformSupport } from '../platform/platform';
 
-const logClient = new Logger('@firebase/firestore');
+export const DEBUG = LogLevel.DEBUG;
```
wilhuff: Could you could preserve the usability of a `LogLevel` enum by doing

```ts
import { Logger, LogLevel as FirebaseLogLevel } from '@firebase/logger';
// ...
export const LogLevel = FirebaseLogLevel;
```

packages/firestore/src/api/database.ts
```diff
@@ -653,31 +650,18 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     return new WriteBatch(this);
   }
 
-  static get logLevel(): firestore.LogLevel {
```
wilhuff: Isn't this a breaking change? I guess `logLevel` was never in the d.ts file so typescript users couldn't have counted on this. What about plain JavaScript users? Was this previously mangled?

schmidt-sebastian: It exists in the mangled build:

Object.defineProperty(t, "logLevel", {
        get: function() {
            switch (Hn()) {
              case 0 /* DEBUG */ :
                return "debug";

              case 1 /* ERROR */ :
                return "error";

              case 2 /* SILENT */ :
                return "silent";

              default:
                return ti("Unknown log level: " + Hn());
            }
        },

It was never part of our types and not in the public API, but it also wasn't prefixed with an underscore. Let me sleep on this.

schmidt-sebastian: I added it back, but also added an entry to the "breaking changes" doc to either remove it make it part of the public API.

PR #2807, opened 2020-03-26
Remove namespace import for objUtils

schmidt-sebastian:
This PR removes the namespace import for `objUtils` and replaces a bunch of usages of the helper functions with ES6 standard primitives (Objects -> Maps).

I verified that the object that I am now using Object.keys()/Object.values() on don't have any inherited properties.


schmidt-sebastian:
Better diff for viewing: https://github.com/firebase/firebase-js-sdk/pull/2807

packages/firestore/src/core/sync_engine.ts
```diff
@@ -912,7 +914,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       const activeTargets: TargetId[] = [];
 
       let p = Promise.resolve();
-      objUtils.forEachNumber(this.queriesByTarget, (targetId, _) => {
+      this.queriesByTarget.forEach((_, targetId) => {
```
dconeybe: Could these `forEach` loops be replaced with simpler and, arguable, more readable,`for` loops?

e.g.

```
this.queriesByTarget.forEach((_, targetId) => {
```

could be changed to:


```
for (targetId of this.queriesByTarget.keys()) {
```

packages/firestore/src/local/encoded_resource_path.ts
```diff
@@ -72,7 +72,7 @@ const encodedEscape = '\u0011';
 /**
  * Encodes a resource path into a IndexedDb-compatible string form.
  */
-export function encode(path: ResourcePath): EncodedResourcePath {
+export function encodeResourcePath(path: ResourcePath): EncodedResourcePath {
```
dconeybe: What is your rationale for renaming these functions?  It's not that I disagree, it's just that the motivation is unclear to me.

schmidt-sebastian: The are now used as top-level functions. I renamed them to be more descriptive than just "encode/decode".

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I tried making this work (I pushed two commits and one revert). The problem is that I would have to turn on "downlevelIteration" to make this work with ES5, which adds a lot of code (+4%).

The example you posted actually does not need downlevelIteration since you are converting the Map to an Array. That creates an extra copy though and hence I decided to leave this as is.

schmidt-sebastian:
@wilhuff for approval.

packages/firestore/src/core/sync_engine.ts
dconeybe: Cool.  Thanks for looking into this.

packages/firestore/src/api/database.ts
```diff
@@ -341,9 +338,7 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
     validateExactNumberOfArgs('Firestore.settings', arguments, 1);
     validateArgType('Firestore.settings', 'object', 1, settingsLiteral);
 
-    if (
-      objUtils.contains(settingsLiteral as objUtils.Dict<{}>, 'persistence')
-    ) {
+    if (settingsLiteral.hasOwnProperty('persistence')) {
```
wilhuff: If the user has overridden `hasOwnProperty` this validation result cannot be trusted. Also, if the object was created with `Object.create(null)` the object has no prototype and this invocation will throw. Calling `Object.prototype.hasOwnProperty.call` is the safer way to phrase this. Since we do this in a few places and the reference can't be minified, it's probably cheaper to keep the `contains` method.

Since this is really validation-related rather than general object manipulation it could make sense to move `contains` into `input_validation.ts`.

packages/firestore/test/unit/local/web_storage_shared_client_state.test.ts
```diff
@@ -103,7 +103,7 @@ class TestSharedClientSyncer implements SharedClientStateSyncer {
 
   get sharedClientState(): TestSharedClientState {
     return {
-      mutationCount: objUtils.size(this.mutationState),
+      mutationCount: size(this.mutationState),
```
wilhuff: This is a really ambiguous name now. Maybe `objectSize`?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -442,14 +438,14 @@ export class SpecBuilder {
 
     // Clear any preexisting limbo watch targets, which we'll re-create as
     // necessary from the provided keys below.
-    objUtils.forEach(this.limboMapping, (key, targetId) => {
+    forEach(this.limboMapping, (key, targetId) => {
```
wilhuff: You could map `limboMapping` a `Map` as well.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -408,9 +408,10 @@ abstract class TestRunner {
   );
 
   private expectedLimboDocs: DocumentKey[];
-  private expectedActiveTargets: {
-    [targetId: number]: { queries: SpecQuery[]; resumeToken: string };
-  };
+  private expectedActiveTargets: Map<
```
wilhuff: This type is repeated several times in here. Would it make sense to create a type alias for the map type or even just the value type within the map?

packages/firestore/src/platform_node/node_platform.ts
```diff
@@ -57,8 +57,8 @@ export class NodePlatform implements Platform {
   }
 
   formatJSON(value: unknown): string {
-    // util.inspect() results in much more readable output than JSON.stringify()
-    return util.inspect(value, { depth: 100 });
+    // util's inspect() results in much more readable output than JSON.stringify()
```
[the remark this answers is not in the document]
schmidt-sebastian: Okay, got it. Fixed.

PR #2813, opened 2020-03-27
Move test code to test helpers

schmidt-sebastian:
This PR moves test-only code out of our serializer in order to reduce SDK size.

It also pretty-prints the resume token in Spec tests since all good PRs need to have at least one unrelated change... :)


schmidt-sebastian:
@wilhuff for approval.

PR #2814, opened 2020-03-27
Remove asserts

schmidt-sebastian:
This PR removes all asserts from our minified builds. This results in 7-10% savings before GZIP (see below).

The non-minified integration tests continue to run against the build with asserts. The Node build is not minified/mangled and continues to contains the asserts (referred to as "main" build below).

This is based on prior work in https://github.com/firebase/firebase-js-sdk/compare/master...mikelehen:mikelehen/size-remove-asserts


schmidt-sebastian:
This is a branch with all messages removed: https://github.com/firebase/firebase-js-sdk/pull/2820 (for now this is just done via refactoring, rather than code transformation. We likely want to stick with code transformation to make testing and development easier).

What stands out though is that almost all of our asserts are for internal state (like calling start() before using instance methods, type checking and in general verifying that our own generated data is sane). There are very few asserts that center around the backend protocol (but there are some, e.g. we verify the number of returned documents from a BatchGetDocuments and the number of returned mutation responses).

The size savings for the uncompressed memory-only build are reduced by roughly 3% if we only strip out the messages. The IndexedDB build is 2% smaller with the asserts completely removed.
With gzip, the "asserts but no messages" build is 84571 KB / 72665 KB (IndexedDB/memory-only, while the "no assert" build is 82635 KB / 71080 KB (~2.2% savings).

Note that Google strips asserts in its TypeScript build if "recommended settings" are used:
https://cs.corp.google.com/piper///depot/google3/javascript/tools/jscompiler/builddefs/flags.bzl?l=150
-> https://cs.corp.google.com/piper///depot/google3/javascript/tools/jscompiler/builddefs/flags.bzl?rcl=303238226&g=0&l=69

With out current pipeline, we could also treat the memory-only build different from the IndexedDB-build and only strip asserts from the memory-only build, which would reduce the risk of corruption. We could also start differentiating between "debugAsserts" and "runtimeAsserts" and keep the few asserts that take into account external input. Or we could decided that 2.2% is not worth this effort and just strip out the messages....

wilhuff:
I favor differentiating debug asserts from runtime asserts, though maybe we could call those `hardAssert` to match Android and iOS naming?

Secondarily I wonder if we should replace hard asserts with proper error handling. That is: if we get an invalid value from watch, drop the whole message and proceed after logging rather than crashing?

I think we could proceed with this PR with only the first part and leave the latter suggestion for later.

schmidt-sebastian:
Updated the PR to remove the newly added "debugAsserts". With this, the savings are now 5-7% since we are not (yet) special-casing `fail`.

PR #2817, opened 2020-03-28
Remove unused code

schmidt-sebastian:
This PR removes unused code from the SDK (as shown by running the integration tests with coverage). It's only a small improvement, but it should not be harmful.


packages/firestore/src/util/async_queue.ts
```diff
@@ -289,15 +289,6 @@ export class AsyncQueue {
           const message = error.stack || error.message || '';
           logError('INTERNAL UNHANDLED ERROR: ', message);
 
-          // Escape the promise chain and throw the error globally so that
-          // e.g. any global crash reporting library detects and reports it.
-          // (but not for simulated errors in our tests since this breaks mocha)
-          if (message.indexOf('Firestore Test Simulated Error') < 0) {
```
dconeybe: This is still used, isn't it?

https://github.com/firebase/firebase-js-sdk/blob/bd85150b032ee360cb9e5e7c39b5a47de9b2bba6/packages/firestore/test/unit/util/async_queue.test.ts#L75

You mentioned in the PR description that you ran the integration tests... did you run the unit tests also?  Because it seems like the unit tests should fail with this line removed.

packages/firestore/src/local/encoded_resource_path.ts
```diff
@@ -175,25 +175,3 @@ export function decode(path: EncodedResourcePath): ResourcePath {
...
- * path. It's useful for defining the end of a prefix scan such that all keys
- * in the scan have the same prefix.
- *
- * Note that this is not a general prefix successor implementation, which is
- * tricky to get right with Strings, given that they encode down to UTF-8.
- * Instead this relies on the fact that all paths encoded by this class are
- * always terminated with a separator, and so a successor can always be
- * cheaply computed by incrementing the last character of the path.
- */
-export function prefixSuccessor(
```
dconeybe: Just for my own information, is it acceptable to delete exported functions as long as they are not "publicly exported" in https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts?  Otherwise, wouldn't this be a backwards-incompatible change?

packages/firestore/src/util/async_queue.ts
schmidt-sebastian: This is special handling of test code that is no longer needed, since we are not using `mocha` anymore (`karma` runs our browser tests). So the same test passes now without this special handling.

As for running the unit tests - all tests run as part of CI, and I usually run them locally as well. There are some tests that are "harder" to run locally, such as the minified integration tests (see /integration/firestore), but even those are run in CI.

packages/firestore/src/local/encoded_resource_path.ts
schmidt-sebastian: In general, if code is not in the api/ folder, it can't be used by our users.

In more technical terms, only code that is exported from our index.ts files is considered part of the public API:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/register-module.ts

Furthermore, the files under /api should match our types  (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts). All code in the exported classes above should either be part of the public API or prefixed with an underscore. We don't always get this right though - see https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/database.ts#L653. `logLevel` is not in our types, but can be accessed since it is exported as part of Firestore.

schmidt-sebastian:
@wilhuff for approval.

packages/firestore/src/local/encoded_resource_path.ts
schmidt-sebastian: The Tree-Shaking in the browser builds doesn't actually work very well. I am trying to fix that (currently not working though - https://github.com/firebase/firebase-js-sdk/pull/2829), but for the "older builds" only excludes code that in modules that aren't loaded at all. Only the ESM2017 build will not include functions, but also only if they are deemed "side-effect free".

packages/firestore/src/local/target_data.ts
```diff
@@ -118,18 +118,4 @@ export class TargetData {
       this.resumeToken
     );
   }
-
-  isEqual(other: TargetData): boolean {
```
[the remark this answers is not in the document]
schmidt-sebastian: TargetData was never passed to our custom "equals" implementation. I actually looked back at the usage and since there is only one call, I decided to inline it. So now we have no code that treats objects with isEqual and without in the same codepath.

PR #2818, opened 2020-03-28
Remove TargetIdGenerator

schmidt-sebastian:
This PR removes the TargetIdGenerator, which offered a fancy way to add +2 to integers.


packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -94,7 +92,13 @@ export class ClientMemoryState {
     this.queryMapping = new ObjectMap<Target, TargetId>(t => t.canonicalId());
     this.limboMapping = {};
     this.activeTargets = {};
-    this.limboIdGenerator = TargetIdGenerator.forSyncEngine();
+    this.nextLimboTarget = 1;
+  }
+
+  nextLimboTargetId(): TargetId {
```
dconeybe: nit: Naming the `nextLimboTargetId()` method the same as the variable name somewhat implies that it is a simple "getter" method.  Consider renaming to something that implies that it is "doing something" not just "getting something".  e.g. `generateLimboTargetId()`.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -170,8 +175,8 @@ export class SpecBuilder {
     return this.currentClientState;
   }
 
-  private get limboIdGenerator(): TargetIdGenerator {
-    return this.clientState.limboIdGenerator;
+  private nextLimboTargetId(): TargetId {
```
dconeybe: Same here; consider renaming the `nextLimboTargetId()` method.

packages/firestore/src/core/target_id_generator.ts
```diff
@@ -15,83 +15,43 @@
...
-    return this.next();
-  }
-
-  private seek(targetId: TargetId): void {
-    assert(
-      (targetId & RESERVED_BITS) === this.generatorId,
-      'Cannot supply target ID from different generator ID'
-    );
-    this.nextId = targetId;
+    this.lastId += OFFSET;
```
wilhuff, line 41: This is functionally different from what came before. This is now a pre-increment where before it was post-increment. Just double-checking: is this intended?

schmidt-sebastian: Yes. The target ID generator now gets initialized with 0 and -1 (shudder), which works around this inconsistency.

PR #2829, opened 2020-03-30
Tree-Shake ES5 classes

schmidt-sebastian:
Rollup doesn't tree-shake ES5 classes since it cannot detect that they are side-effect free. The recommended way to solve this is to not use Typescript for the ES5 conversion, but Babel. Unfortunately, Typescript's ES5 is much smaller than Babels.

Luckily, the Internet answers all questions:

Typescript emits "@class" comment when a class is side-effect free:
https://github.com/microsoft/TypeScript/issues/13721

This can be rewritten to __PURE__ as show here:
https://github.com/rollup/rollup/issues/2807#issuecomment-486267624

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -51,6 +51,7 @@
   "devDependencies": {
     "protobufjs": "6.8.9",
     "rollup": "2.0.6",
+    "rollup-plugin-babel": "4.4.0",
     "rollup-plugin-copy-assets": "1.1.0",
     "rollup-plugin-json": "4.0.0",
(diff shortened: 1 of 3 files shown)


Feiyang1:
Nice. Can you please check if source maps work correctly? I have used this workaround when doing size analysis. The size analysis tool told me the source maps are invalid where the number of lines mismatch, so I just switched to use es2017 build for analysis.

schmidt-sebastian:
Yeah, the source maps are not working... Looking into it.

Issue #2827, opened 2020-03-29
IE11 broken on 7.13.x


thesandlord:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Browser version: IE11
  * Firebase SDK version: 7.13.1
  * Firebase Product: Firestore

### [REQUIRED] Describe the problem

Firestore (both memory-only and standard versions) breaks on IE11 with version 7.13.0 and 7.13.1

Everything works fine on 7.12.0

The first error is:

`SCRIPT5009: 'crypto' is undefined`

Which is easy to fix with a pollyfill
`var crypto = window.crypto || window.msCrypto;`

After that, this error pops up:
`SCRIPT438: Object doesn't support property or method 'forEach'`

I think this is caused by this PR:
https://github.com/firebase/firebase-js-sdk/pull/2764

Specifically this line:
https://github.com/firebase/firebase-js-sdk/blob/97aada8c57d8f481433b8b3fb288c33dd91f931a/packages/firestore/src/util/misc.ts#L34

This is a Uint8Array, so it needs to be pollyfilled like this:
`Uint8Array.prototype.forEach = Array.prototype.forEach;`

I don't think these pollyfills are included in @babel/preset-env or the core-js all in one script

schmidt-sebastian:
This will be fixed in the next release.

PR #2832, opened 2020-03-30
Fix IE11 broken by Crypto change.

wu-hui:
Fix for #2827

### DIFF
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -75,6 +75,8 @@ export class BrowserPlatform implements Platform {
     }
 
     const v = new Uint8Array(nBytes);
+    // Polyfill for IE11.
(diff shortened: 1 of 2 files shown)


packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -75,6 +75,8 @@ export class BrowserPlatform implements Platform {
     }
 
     const v = new Uint8Array(nBytes);
+    // Polyfill for IE11.
+    const crypto = window.crypto || (window as any).msCrypto;
```
Feiyang1: You probably need to suppress eslint here to pass tests.

wu-hui: Done.

wilhuff: It seems like you could do this once at the top of the file (after imports before the class).

PR #2834, opened 2020-03-30
Use @firebase/firestore/memory for Memory-build

schmidt-sebastian:


packages/firebase/firestore/memory/index.ts
```diff
@@ -20,4 +20,4 @@
  * `firebase/firestore/memory`.
  */
 
-import '../../../firestore/dist/index.memory.esm';
+import '../../../firestore/dist/index.memory.esm2017';
```
Feiyang1: hmm, we should import from `@firebase/firestore` instead of from the local path. The build tool will choose the build according to the configuration.

```
import '@firebase/firestore/memory'
```

schmidt-sebastian:
Merging this since it is cleaner, but it doesn't actually change the size since Typescript imports the ES3 build.

PR #2840, opened 2020-03-31
Pre-process all ES5 builds

schmidt-sebastian:
@Feiyang1 I have to test these builds but this might enable tree-shaking in ES3/5 builds. I got this to work by using two different rollup phases.


Feiyang1:
LGTM. I'm glad (surprised?) to see that it reduces the memory only build significantly. It's also interesting to see that the size of `firebase-firestore.js` and `firebase.js` increased a tiny bit, probably because the mangling is done indirectly now for cjs and es5 builds? In that case, I don't think it is a problem.

schmidt-sebastian:
TODO Validation checklist:

 [  ] Browser builds are mangled
 [  ] Non-ES2017 builds are ES3.
 [  ] firebase.js works
 [  ] firebase-firestore.js works
 [  ] firebase-firestore.memory.works
 [  ] Node build works in Node 6
 [  ] Node memory-only build works in Node 6
 [  ] CJS build works in browser
 [  ] CJS memory-build works in browser
 [  ] ESM build works in browser
 [  ] ESM memory-build works in browser

schmidt-sebastian:
@Feiyang1  This should now be ready for an actual review. I manually tested all builds.

packages/firestore/package.json
```diff
@@ -26,6 +26,7 @@
     "prepare": "yarn build"
   },
   "main": "dist/index.node.cjs.js",
+  "mainES2017": "dist/index.node.esm2017.js",
```
Feiyang1, line 29: `main-esm2017` to be consistent with the existing convention? We have a `lite-esm2017` target in @firebase/app which is used to produce `firebase-performance-standalone.es2017.js`

packages/firestore/memory/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@firebase/firestore/memory",
   "description": "A memory-only build of the Cloud Firestore JS SDK.",
   "main": "../dist/index.memory.node.cjs.js",
+  "mainES2017": "../dist/index.memory.node.esm2017.js",
```
Feiyang1, line 5: `main-esm2017`?

packages/firestore/rollup.config.es2017.js
```diff
@@ -45,6 +48,11 @@ import {
 // The in-memory builds are roughly 130 KB smaller, but throw an exception
 // for calls to `enablePersistence()` or `clearPersistence()`.
 //
+// We use two different rollup pipelines to take advantage of tree shaking,
+// which Rollup only supports in pure ES2017 builds. The build pipeline in this
```
Feiyang1, line 52: This is not accurate. rollup can tree shake ES5 builds as long as they are in the ESM format. The limitation is specific to the Typescript classes compiled down to ES5. 

Also do you mean `post-tree shaken ES2017` builds? IIUC, treeshaking is done in this phase, the second phase just changes the module format.

packages/firestore/rollup.config.es.js
```diff
@@ -0,0 +1,128 @@
...
+const nodePlugins = [
+  typescriptPlugin({
+    typescript,
+    compilerOptions: {
+      allowJs: true,
+      importHelpers: true
+    },
+    include: ['dist/*.js']
+  }),
+  terser({
```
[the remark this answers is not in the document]
Feiyang1: IIUC, we are not doing property mangling here, and we don't care about the size saving by mangling variable names. Developers will do them when building their apps. 
Removing it here can save some build time.

packages/firestore/rollup.config.es2017.js
schmidt-sebastian: Updated the comment. I did mean "pre tree-shaken", as in "these builds are already tree shaken", which is the same as "post tree-shaken".  Hence, I dropped the term.

packages/firestore/rollup.config.es.js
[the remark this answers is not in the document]
schmidt-sebastian: Ok, removed.

schmidt-sebastian:
I also updated the PR to remove the error handling for memory externs, which was meant to ensure that our ES5 builds don't accidentally pull in sources they don't need.  If this happens now, Rollup will just tree shake those files.

PR #2843, opened 2020-03-31
Generate stable JSON from generate_spec_json.js

dconeybe:
Currently, the JSON generated by `generate_spec_json.js` uses `JSON.stringify()`, which generates JSON with the keys ordered by the order in which they are set in the objects.  As a result, code changes that alter the order in which the keys as added but have no other side effects cause different stringified JSON to be generated, resulting in unnecessarily-noisy diffs.

This PR changes `generate_spec_json.js` to use the third-party `json-stable-stringify` library: https://www.npmjs.com/package/json-stable-stringify.  This library produces JSON that has a stable key ordering, resulting in diffs when and only when keys are added, removed, or have their values modified.

I also implemented a custom comparator so that the generated JSON has a semblance of human readability, with the test "configuration" at the top and the "expected" blocks at the bottom.

I will submit follow-up PRs to the Android and iOS repositories with the new, stable JSON.  These PRs will be very noisy but will allow for cleaner future diffs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,5 +128,9 @@
     "hooks": {
       "pre-push": "node tools/gitHooks/prepush.js"
     }
+  },
+  "dependencies": {
+    "@types/json-stable-stringify": "^1.0.32",
+    "json-stable-stringify": "^1.0.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -128,9 +128,5 @@
     "hooks": {
       "pre-push": "node tools/gitHooks/prepush.js"
     }
-  },
-  "dependencies": {
-    "@types/json-stable-stringify": "^1.0.32",
-    "json-stable-stringify": "^1.0.1"
   }
 }
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -49,6 +49,8 @@
     "@firebase/app-types": "0.x"
   },
   "devDependencies": {
+    "@types/json-stable-stringify": "^1.0.32",
+    "json-stable-stringify": "^1.0.1",
     "protobufjs": "6.8.9",
     "rollup": "2.0.6",
     "rollup-plugin-copy-assets": "1.1.0",
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -49,8 +49,8 @@
     "@firebase/app-types": "0.x"
   },
   "devDependencies": {
-    "@types/json-stable-stringify": "^1.0.32",
-    "json-stable-stringify": "^1.0.1",
+    "@types/json-stable-stringify": "1.0.32",
+    "json-stable-stringify": "1.0.1",
     "protobufjs": "6.8.9",
     "rollup": "2.0.6",
     "rollup-plugin-copy-assets": "1.1.0",
--- a/packages/firestore/test/unit/specs/describe_spec.ts
+++ b/packages/firestore/test/unit/specs/describe_spec.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/packages/firestore/test/unit/specs/describe_spec.ts
+++ b/packages/firestore/test/unit/specs/describe_spec.ts
@@ -19,6 +19,7 @@ import { ExclusiveTestFunction, PendingTestFunction } from 'mocha';
 
 import { IndexedDbPersistence } from '../../../src/local/indexeddb_persistence';
 import { assert } from '../../../src/util/assert';
+import { primitiveComparator } from '../../../src/util/misc';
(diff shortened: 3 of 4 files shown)


packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -24,6 +24,8 @@ import { addEqualityMatcher } from '../../util/equality_matcher';
 import { SpecBuilder } from './spec_builder';
 import { SpecStep } from './spec_test_runner';
 
+import * as stringify from 'json-stable-stringify';
```
wilhuff: It's less important in tests but we should be getting in the habit of avoiding whole-module imports like this.

However, I'm not actually sure what the right way to do this import is. Does `import stringify from 'json-stable-stringify'` work?

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -242,7 +244,64 @@ export function describeSpec(
...
+/**
+ * A comparator function for stringify() that sorts keys in the JSON output.
+ *
+ * In order to produce semi-readable JSON that has an intuitive top-level key
+ * arrangement, some keys are sorted non-alphabetically; namely, the ordering
+ * defined in `stringifyCustomOrdering` is used, falling back to alphabetical
+ * ordering if one or both of the key are not overridden.
+ */
+function stringifyCmp(a: stringify.Element, b: stringify.Element): number {
+  // If the keys are equal, then avoid the costly computations below.
```
wilhuff: Performance doesn't matter here so I'd optimize for readability over anything else.

You can make this simpler by creating a function that determines the group number (use e.g. `stringifyCustomOrdering.length` for `expected` and `stringifyCustomOrdering.length + 1` for "everything else").

Then you can just do:

```ts
const aGroup = stringifyGroup(a.key);
const bGroup = stringifyGroup(b.key);

if (aGroup === bGroup) {
  return primitiveComparator(a.key, b.key);  // from util/misc.ts
} else {
  return primitiveComparator(aGroup, bGroup);
}
```

packages/firestore/test/unit/specs/describe_spec.ts
```diff
@@ -242,7 +244,64 @@ export function describeSpec(
...
+
+/**
+ * A comparator function for stringify() that sorts keys in the JSON output.
+ *
+ * In order to produce semi-readable JSON that has an intuitive top-level key
+ * arrangement, some keys are sorted non-alphabetically; namely, the ordering
+ * defined in `stringifyCustomOrdering` is used, falling back to alphabetical
+ * ordering if one or both of the key are not overridden.
+ */
+function stringifyCmp(a: stringify.Element, b: stringify.Element): number {
```
wilhuff: Avoid abbreviations like this. Elsewhere we fully spell out "comparator".

package.json
```diff
@@ -128,5 +128,9 @@
     "hooks": {
       "pre-push": "node tools/gitHooks/prepush.js"
     }
+  },
+  "dependencies": {
```
schmidt-sebastian: This should be a dev-dependency in packages/firestore.

packages/firestore/test/unit/specs/describe_spec.ts
dconeybe: Done.  Great suggestion.

package.json
dconeybe: Done.

packages/firestore/test/unit/specs/describe_spec.ts
wilhuff: Ah, it looks like we're missing the `esModuleInterop` compiler setting.

@Feiyang1 is there any reason not to enable this repo-wide? We're already using it in our [integration tests](https://github.com/firebase/firebase-js-sdk/blob/6668b59e2fa495fcea42a24a6b24821407ee703f/integration/typescript/tsconfig.json#L13).

dconeybe:
@Feiyang1 Could you review for the files changed in the root directory (package.json and yarn.lock)?

packages/firestore/test/unit/specs/describe_spec.ts
Feiyang1: We can enable it. I'm actually surprised that you can use `stringify` as a function directly. Thought typescript will give an error as it's not compliant with es6 modules spec.

packages/firestore/package.json
```diff
@@ -49,6 +49,8 @@
     "@firebase/app-types": "0.x"
   },
   "devDependencies": {
+    "@types/json-stable-stringify": "^1.0.32",
```
Feiyang1, line 52: please fix the versions.

packages/firestore/test/unit/specs/describe_spec.ts
Feiyang1: https://github.com/firebase/firebase-js-sdk/pull/2849

packages/firestore/package.json
dconeybe: Done.

PR #2845, opened 2020-03-31
Fix Saucelabs tests in analytics and functions

hsubox76:
Fix some failing IE tests in analytics and functions.
Also fixed error in workflow file so that cross-browser workflow will run again.

### DIFF
--- a/packages/analytics/index.test.ts
+++ b/packages/analytics/index.test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2019 Google Inc.
+ * Copyright 2019 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


packages/analytics/index.test.ts
```diff
@@ -88,7 +88,8 @@ describe('FirebaseAnalytics instance tests', () => {
         currency: 'USD'
       });
       // Clear event stack of async FID call.
-      await fidDeferred.promise;
+      // For IE: Need to wrap it in another function as FID fetch is wrapped in initializeGAId.
+      await (() => fidDeferred.promise);
```
Feiyang1, line 92: Does it really work? Are you meant to execute the function, `(async () => fidDeferred.promise)()`?

hsubox76: It does work on Saucelabs. I think the result is the same, if an async function returns a promise it doesn't wrap it with an additional promise. It only does if it's not a promise.

hsubox76: It did work but I don't know why, which seems bad. I changed it to something more readable but doesn't follow the logic as much. For some reason, only IE executes the `expect` assertions immediately after the mock FID promise resolves instead of executing the gtag config statements which should be first in line.

I also fixed another issue on the last test in `index.test.ts` where the config calls were resolving after the test was disposed of, which again seems to be IE only.

PR #2852, opened 2020-04-02
Make ServerValue.increment() public

schmidt-sebastian:
As per API proposal: go/servervalue-increment-api

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -6920,6 +6920,19 @@ declare namespace firebase.database.ServerValue {
    * ```
    */
   var TIMESTAMP: Object;
+  
+  /**
+   * Returns a placeholder value that can be used to atomically increment the 
+   * current database value by the provided delta.
+   *
+   * @param delta the amount to modify the current value atomically.
+   * @return a placeholder value for modifying data atomically server-side.
+   */
+  function increment(delta: number) : {
+    '.sv': {
+      'increment': number
+    }
+  };
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/firebase/index.d.ts
```diff
@@ -6920,6 +6920,15 @@ declare namespace firebase.database.ServerValue {
    * ```
    */
   var TIMESTAMP: Object;
+  
+  /**
+   * Returns a placeholder value that can be used to atomically increment the
```
inlined, line 6925: Consider the documentation from Firestore: https://googleapis.dev/nodejs/firestore/latest/FieldValue.html#.increment

inlined: To the degree that we follow the same rules of course. We overflow to double instead.

schmidt-sebastian: In the JS SDK itself, we don't do anything special though. Internally, we use floating point arithmetic (with its limited 2^53 precision) and once we round-trip through the server we gain land in JavaScript number land (please let me know if I am wrong here).

Firestore is a little different in that it can retain the actual value for 2^53 + 1 on the backend, whereas JavaScript cannot.

To me, it seems like the RTDB just uses JavaScript number semantics all throughout - is that correct? I can modify the comment to say that we are using floating point arithmetic throughout (similar to https://googleapis.dev/nodejs/firestore/latest/FieldValue.html#.increment), but I want to make sure that my understanding is correct.

inlined: I guess it just makes a difference that whole numbers would be interpreted as ints in other SDKs whereas a fractional number will cause other SDKs to see floats. OTOH that isn't different from any other field, so it doesn't really matter.

schmidt-sebastian:
@Feiyang1 I need someone else to approve this too. Can you take a look?

PR #2853, opened 2020-04-02
Request for ealy comments: IndexedDB recovery

schmidt-sebastian:
This PR is somewhere between exploration and hack. It is meant to provide a strategy for IndexedDB recovery.

So far, it recovers:
- When a Watch updates cannot be persisted. 
- When a mutation cannot be written to the mutation queue.
- When a mutation acknowledgement or rejection cannot be persisted.

Some notes: go/indexeddb-recovery

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -16,7 +16,12 @@
  */
 
 import { User } from '../auth/user';
-import { ignoreIfPrimaryLeaseLoss, LocalStore } from '../local/local_store';
+import {
+  handleOutdatedRemoteSnapshot,
+  handlePrimaryLeaseLoss,
+  LocalStore,
+  LocalWriteResult
+} from '../local/local_store';
 import { LocalViewChanges } from '../local/local_view_changes';
 import { ReferenceSet } from '../local/reference_set';
 import { TargetData, TargetPurpose } from '../local/target_data';
@@ -34,7 +39,7 @@ import { RemoteStore } from '../remote/remote_store';
 import { RemoteSyncer } from '../remote/remote_syncer';
 import { assert, fail } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
-import { logDebug } from '../util/log';
+import { logDebug, logError } from '../util/log';
 import { primitiveComparator } from '../util/misc';
 import { ObjectMap } from '../util/obj_map';
 import { Deferred } from '../util/promise';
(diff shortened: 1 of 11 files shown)


packages/firestore/src/local/local_store.ts
```diff
@@ -511,155 +513,155 @@ export class LocalStore {
 
     return this.persistence
       .runTransaction('Apply remote event', 'readwrite-primary', txn => {
-        const documentBuffer = this.remoteDocuments.newChangeBuffer({
-          trackRemovals: true // Make sure document removals show up in `getNewDocumentChanges()`
-        });
-
-        // Reset newTargetDataByTargetMap in case this transaction gets re-run.
-        newTargetDataByTargetMap = this.targetDataByTarget;
+        return this.targetCache
```
wilhuff, line 516: Could you explain a bit of the rationale behind this change? It seems mostly like this is just a rearrangement of what came before, but I'm having a hard time understanding both what changed and why.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -403,12 +403,9 @@ export class RemoteStore implements TargetMetadataProvider {
     }
 
     if (!snapshotVersion.isEqual(SnapshotVersion.MIN)) {
-      const lastRemoteSnapshotVersion = await this.localStore.getLastRemoteSnapshotVersion();
-      if (snapshotVersion.compareTo(lastRemoteSnapshotVersion) >= 0) {
-        // We have received a target change with a global snapshot if the snapshot
-        // version is not equal to SnapshotVersion.MIN.
-        await this.raiseWatchSnapshot(snapshotVersion);
-      }
+      // We have received a target change with a global snapshot if the snapshot
```
wilhuff, line 406: This seems like a big change. Is this check handled somewhere else?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -593,7 +600,9 @@ export class RemoteStore implements TargetMetadataProvider {
           this.writeStream.writeMutations(batch.mutations);
         }
       })
-      .catch(ignoreIfPrimaryLeaseLoss);
+      .catch(() => {
+        // TODO
```
wilhuff, line 604: Yes please :-).

packages/firestore/src/core/sync_engine.ts
```diff
@@ -596,7 +638,15 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       this.sharedClientState.updateMutationState(batchId, 'acknowledged');
       await this.emitNewSnapsAndNotifyLocalStore(changes);
     } catch (error) {
-      await ignoreIfPrimaryLeaseLoss(error);
+      logError(
+        LOG_TAG,
+        'Failed to persist write acknowledgment: ' + error.message
+      );
+      
+      // We close the write stream forcefully, which doesn't update the stream
```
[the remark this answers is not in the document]
schmidt-sebastian: We don't actually ever set the `streamToken`, but if we do change the code and send an outdated token I get the following: 

```
@firebase/firestore:', 'Firestore (7.13.2): Connection', 'WebChannel sending:', '{"database":"projects/its-april/databases/(default)","streamToken":{}}'
@firebase/firestore:', 'Firestore (7.13.2): Connection', 'WebChannel transport opened.'
@firebase/firestore:', 'Firestore (7.13.2): Connection', 'WebChannel received error:', '{"code":400,"message":"Request contains an invalid argument.","status":"INVALID_ARGUMENT"}'
```

The emulator doesn't error here - but it seems to only use a stream token of `[0]`.

PR #2856, opened 2020-04-02
Drop Node 6 / Replace grpc w/ @grpc/grpc-js

schmidt-sebastian:
- Adds engine field to match requirements for `@grpc/grpc-js`
- Removes `grpc`, replaces all usages with `@grpc/grpc-js`
- Updates `@types/node`, which is required to make `@grpc/grpc-js` compile. This also adds `finally` to the Promise API, which requires a small code change.

Note that there are currently two version of `@grpc/grpc-js` in our repo, which should go away once https://github.com/firebase/firebase-tools/pull/2093 is released.


package.json
```diff
@@ -55,7 +59,7 @@
     "@types/chai-as-promised": "7.1.2",
     "@types/long": "4.0.1",
     "@types/mocha": "7.0.2",
-    "@types/node": "8.10.59",
+    "@types/node": "13.11.0",
```
Feiyang1, line 62: I think we should reference a LTS version instead, like Node 12.

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -15,13 +15,18 @@
...
+  credentials as GrpcCredentials,
+  ServiceError
+} from '@grpc/grpc-js';
 
 import firebase from '@firebase/app';
 const SDK_VERSION = firebase.SDK_VERSION;
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
-const grpcVersion = require('grpc/package.json').version;
+const grpcVersion = require('@grpc/grpc-js/package.json').version;
```
Feiyang1, line 29: You can use import instead of require and we can remove the eslint suppress.

packages/testing/package.json
```diff
@@ -20,7 +23,7 @@
     "@firebase/logger": "0.2.1",
     "@firebase/util": "0.2.44",
     "@types/request": "2.48.4",
-    "grpc": "1.24.2",
+    "firebase": "7.13.1",
```
Feiyang1, line 26: It is redundant. `firebase` is already a dependency.
Was grpc a redundant dependency anyway regardless of this change?

packages/firestore/src/platform_node/grpc_connection.ts
```diff
@@ -25,8 +25,7 @@ import {
 import firebase from '@firebase/app';
 const SDK_VERSION = firebase.SDK_VERSION;
 
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-const grpcVersion = require('@grpc/grpc-js/package.json').version;
+const grpcVersion = import('@grpc/grpc-js/package.json').version;
```
Feiyang1: You could also do a normal import, then add a plugin in the [rollup.config.js](https://github.com/firebase/firebase-js-sdk/blob/master/packages/app/rollup.config.js#L94).
```
import { version } from '@grpc/grpc-js/package.json'
```

schmidt-sebastian: We already use this build setting, but since we don't use `rollup` for our unit tests, this breaks the unit test.

PR #2857, opened 2020-04-02
Remove error messages for deprecated APIs

schmidt-sebastian:
This PR removes error messages for APIs that are no longer supported and reduces the size for messages that warn about deprecated usage.


packages/firestore/src/api/database.ts
```diff
@@ -194,32 +194,15 @@ class FirestoreSettings {
...
-  const date = snapshot.get('created_at');
-  // New:
-  const timestamp = snapshot.get('created_at'); const date =
-  timestamp.toDate();
-
-  Please audit all existing usages of Date when you enable the new
-  behavior.`);
+      logError(
+        "Support for 'timestampsInSnapshots: false' will be removed soon. " +
+          'You must update your code to handle Timestamp objects.'
```
wilhuff, line 204: What do you think of putting the helpful text about how to handle this in a file like `packages/firestore/deprecation.md` and then linked to it? Then this could be

```ts
"Support for 'timestampsInSnapshots: false' will be removed. " +
  'See https://github.com/firebase/firebase-js-sdk/packages/firestore/deprecation.md#timestamps'
```

It's almost as short but remains just as helpful.

schmidt-sebastian: I expanded on the comment in our API docs and added a GIST with the comment (since we can't permalink to the API doc).

wilhuff: This is probably better linked to a file actually in the repo rather than a personal gist, similar to what we did here: https://github.com/firebase/firebase-android-sdk/blob/fa9a8c78ad8cae66dec02da9edb40c5303286cc8/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/AbstractStream.java#L282

schmidt-sebastian: FWIW I went back to the original error message (but left the updates to the API documentation). This error message is so old that the number of people that see this probably does not warrant any extra steps.

packages/firebase/index.d.ts
```diff
@@ -7682,11 +7682,23 @@ declare namespace firebase.firestore {
...
-     * this kind of problem.
-     *
-     * To opt into the old behavior of returning `Date` objects, you can
-     * temporarily set `timestampsInSnapshots` to false.
+     * Now, Firestore returns `Timestamp` values for all timestamp values stored 
+     * in Cloud Firestore instead of system `Date` objects, avoiding this kind 
+     * of problem. Consequently, you must update your code to handle `Timestamp`
+     * objects instead of `Date` objects.
+     *
+     * If you wish to TEMPORARILY opt into the old behavior of returning  `Date`
```
egilmorez, line 7690: WDYT about making these words bold instead of capitalizing them?  This feels a little shouty.

Also, suggest "want" instead of wish.

packages/firebase/index.d.ts
```diff
@@ -7682,11 +7682,23 @@ declare namespace firebase.firestore {
...
+     * objects, you may TEMPORARILY set `timestampsInSnapshots` to false.
```
egilmorez, line 7691: WDYT about making these words bold instead of capitalizing them?

packages/firebase/index.d.ts
```diff
@@ -7682,11 +7682,23 @@ declare namespace firebase.firestore {
...
+     * Now, Firestore returns `Timestamp` values for all timestamp values stored 
+     * in Cloud Firestore instead of system `Date` objects, avoiding this kind 
+     * of problem. Consequently, you must update your code to handle `Timestamp`
+     * objects instead of `Date` objects.
+     *
+     * If you wish to TEMPORARILY opt into the old behavior of returning  `Date`
+     * objects, you may TEMPORARILY set `timestampsInSnapshots` to false.  
+     * Opting into this behavior will no longer be possible in the next major 
+     * release of Firestore, after which code that expects Date objects WILL 
+     * BREAK.
```
egilmorez, line 7694: WDYT about making these words bold instead of capitalizing them?

schmidt-sebastian: OKAY FIXED

egilmorez: Hehe . . .

Issue #2870, opened 2020-04-05
"ReferenceError: window is not defined" when using firestore in a web worker.


universse:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Browser version: Firefox 74.0.1
  * Firebase SDK version: 7.13.2
  * Firebase Product: firestore

### [REQUIRED] Describe the problem
![image](https://user-images.githubusercontent.com/15166715/78465969-b8bfcd00-772e-11ea-9619-7ad39886d166.png)
![image](https://user-images.githubusercontent.com/15166715/78465953-8d3ce280-772e-11ea-9e5e-122e18446f1b.png)

laurentpayot:
@universse are you using the custom build from #983? I had the same error has you. The Firestore directory to overwrite has changed: you now have to copy your custom build into `node_modules/firebase/node_modules/@firebase/firestore`.

schmidt-sebastian:
Thanks for reporting this! For now, do you mind downgrading to 7.12.0? We will get this fixed in the next release.

universse:
Thanks for your response. Sure no worry.

merlinnot:
I'm also experiencing this issue when using the library in a Web Worker. In Web Workers, the API is actually available, but you need to access it through `self`, not `window`. So I'd suggest to fix it by checking both references if that's possible.

schmidt-sebastian:
We are going to revert this "crypto" change for now and will release a version that will support WebWorkers and IE11 shortly after.

Issue #2858, opened 2020-04-02
7.13.2 has unhandled promise rejection Wu.getRandomValues is undefined


pdugan20:
### Describe your environment
  * Operating System version: react native 0.61
  * Browser version: n/a
  * Firebase SDK version: 7.13.2
  * Firebase Product: firestore

### Describe the problem
Getting the following error when trying to initialize within react native:
```
[Unhandled promise rejection: TypeError: undefined is not an object (evaluating 'Wu.getRandomValues')]
```

Not sure what Wu is, is that referring to [Kai Wu](https://github.com/firebase/firebase-js-sdk/commit/aab78f9dfbf2138cf4c1d8dd8bae102b7a5bd0b1)? 

#### Steps to reproduce:
Try and initialize firebase within a react native environment.

#### Relevant Code:
```
import Constants from 'expo-constants';
import { decode, encode } from 'base-64';
import firebase from 'firebase';
import '@firebase/firestore';

global.btoa = encode;
global.atob = decode;
global.crypto = {};

global.crypto.getRandomValues = (byteArray) => {
    for (let i = 0; i < byteArray.length; i += 1) {
        byteArray[i] = Math.floor(256 * Math.random());
    }
};

class Fire {
    constructor() {
        if (!firebase.apps.length) {
            firebase.initializeApp(Constants.manifest.extra.firebaseConfig);
        }
    }
}

Fire.shared = new Fire();

export const db = firebase.firestore();
export const storage = firebase.storage();
export const auth = firebase.auth();

export default Fire;
```

schmidt-sebastian:
@pdugan20 Thanks for filing this issue.

@dconeybe The following line is likely the culprit:

```
const crypto = window.crypto || (window as any).msCrypto;
```
`msCrypto` is not in any of the .d.ts files we include here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/externs.json Hence, it gets mangled. 

Since we likely don't want to include an externs file just for this, your best bet may be to use object-literal access: `window["msCrypto"]` (please verify that this does work - the best build for these kind of verifications is the 'dist/index.esm2017.ts' build.

If I interpet this issue correctly. "crypto" is also not available in ReactNative. Otherwise, `msCrypto` should never get evaluated. Do we need to provide a fallback to Math.random()?

RWOverdijk:
I get this on @firebase/firestore@1.13.1.

IjzerenHein:
[...]

PR #2872, opened 2020-04-06
Revert: Use crypo RNG for auto ID generation to reduce conflicts

dconeybe:
This PR temporarily reverts the feature #2764: Use a cryptographically-strong random number generator for generating document IDs.  We received several bug reports from users prompting us to revisit the implementation.  In order to unblock those users, I am temporarily reverting the feature so that it can be picked up in tomorrow's release.

The issues in question are: #2832, #2858, and #2870.


schmidt-sebastian:
Nit: Can you change the PR title to “Revert: Use ...” before merging?

dconeybe:
Shoot... I forgot to update the change log.  I'll do that in a separate PR.

PR #2873, opened 2020-04-06
Add CHANGELOG.md entry for reverting the usage of 'crypto'

dconeybe:
This is the change log entry that was forgotten in https://github.com/firebase/firebase-js-sdk/pull/2872.

### DIFF
--- a/packages/firestore/CHANGELOG.md
+++ b/packages/firestore/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Unreleased
+- [fixed] Reverted the use of the "crypto" global since it caused some user
+  issues (#2872).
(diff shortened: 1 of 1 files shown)


packages/firestore/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Unreleased
+- [fixed] Reverted the use of the "crypto" global since it caused some user
```
schmidt-sebastian, line 2: How about: Temporarily reverted the use of `window.crypto` to generate document IDs to address compatibility issues with IE 11 and WebWorkers.

dconeybe: Done.

PR #2874, opened 2020-04-06
Discourage use of 'window' and 'document'

schmidt-sebastian:
This should help us find issues such as https://github.com/firebase/firebase-js-sdk/issues/2870 before they surface in ReactNative and WebWorkers.

### DIFF
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -20,11 +20,12 @@ import { Platform } from '../platform/platform';
 import { Connection } from '../remote/connection';
 import { JsonProtoSerializer } from '../remote/serializer';
 import { ConnectivityMonitor } from './../remote/connectivity_monitor';
-
 import { NoopConnectivityMonitor } from '../remote/connectivity_monitor_noop';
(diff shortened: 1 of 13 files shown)


packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -34,10 +35,14 @@ export class BrowserPlatform implements Platform {
   }
 
   get document(): Document | null {
+    // `document` is not always available, e.g. in ReactNative and WebWorkers.
+    // eslint-disable-next-line no-restricted-globals
     return typeof document !== 'undefined' ? document : null;
   }
 
   get window(): Window | null {
+    // `window` is not always available, e.g. in ReactNative and WebWorkers.
```
wilhuff, line 44: Drive-by comment: should this fall back on `self` in a web worker?

schmidt-sebastian: I don't know and we should be careful if we want to make this change. FWIW, this doc says to always use `self`: https://developer.mozilla.org/en-US/docs/Web/API/Window/self

This works because JS in browsers prefixes any unknown globals with `window` and `window.self` seems to be a thing.

PR #2875, opened 2020-04-06
Update @firebase/template to move tests next to source files

Feiyang1:


config/karma.base.js
```diff
@@ -57,7 +57,10 @@ const config = {
   // preprocess matching files before serving them to the browser
   // available preprocessors:
   // https://npmjs.org/browse/keyword/karma-preprocessor
-  preprocessors: { 'test/**/*.ts': ['webpack', 'sourcemap'] },
+  preprocessors: {
+    'test/**/*.ts': ['webpack', 'sourcemap'],
```
hsubox76, line 61: Should we get rid of this pattern since template doesn't have a `test/` dir anymore?

Feiyang1: This config file is shared with packages that still have tests in the `test/` dir, so we need to keep it for now.

PR #2879, opened 2020-04-07
Retry WebStorage operations

schmidt-sebastian:
This PR allows us to recover from IndexedDB failures when a Multi-Tab operation scheduled via a WebStorage notification fails. 

To do this:
- It introduces the concept of "retryable" operations. These are directly retried on the AsyncQueue.
- All retryable operations use the same retry mechanism. It is assumed that if the first retryable operation fails, the second would fail to.
- Retries use the same exponential backoff that we use for our streams (down to the delays).

Since a operation that is schedule via a `DelayedOperation` is always run after all other operations, the PR also changes the backoff handler to use `AsyncQueue.enqueue` directly if the backoff is zero. This is not required, but not doing this breaks all tests that rely on WebStorage since `AsyncQueue.drain()` does not drain delayed operations. 
I tried to make this change in the AsyncQueue directly, but it would require to return us to create a new way to create CancellableOperations from `enqueue()`.
A side effect of this is that the spec tests way to fast-forward the backoff timer doesn't work as it did before - the first backoff timer cannot be fast-forwarded and any subsequent attempts are not always scheduled. This causes some tests to time out as we sometimes wait for the backoff to expire. I changed the SpecTests to always skip the backoff delays, which speeds up all tests considerably (faster than on master).

Note: I wasn't able to figure out how to make the test "handles user changes while offline (b/74749605)" work, since the test relies on a backoff for the first attempt.

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -17,8 +17,12 @@
 
 import { debugAssert, fail } from './assert';
 import { Code, FirestoreError } from './error';
-import { logError } from './log';
+import { logDebug, logError } from './log';
 import { CancelablePromise, Deferred } from './promise';
+import { ExponentialBackoff } from '../remote/backoff';
+import { PlatformSupport } from '../platform/platform';
+
+const LOG_TAG = 'AsyncQueue';
 
 // Accept any return type from setTimeout().
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -66,7 +70,13 @@ export const enum TimerId {
    * A timer used to retry transactions. Since there can be multiple concurrent
    * transactions, multiple of these may be in the queue at a given time.
    */
-  RetryTransaction = 'retry_transaction'
+  RetryTransaction = 'retry_transaction',
+
+  /**
+   * A timer used to retry operations scheduled via retryable AsyncQueue
+   * operations.
+   */
+  AsyncQueueRetry = 'async_queue_retry'
 }
 
 /**
@@ -213,6 +223,29 @@ export class AsyncQueue {
   // List of TimerIds to fast-forward delays for.
   private timerIdsToSkip: TimerId[] = [];
 
+  // Backoff timer used to schedule retries for retryable operations
+  private backoff = new ExponentialBackoff(this, TimerId.AsyncQueueRetry);
+
+  // If set, points to the first retryable operation. The first retryable
+  // operation is the only operation that is retried with backoff. If this
+  // operation succeeds, all other retryable operations are run right away.
+  private firstRetryableOperation?: Promise<void>;
+
+  // Visibility handler that triggers an immediate retry of all retryable
+  // operations. Meant to speed up recovery when we regain file system access
+  // after page comes into foreground.
+  private visibilityHandler = (): void => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.runDelayedOperationsEarly(TimerId.AsyncQueueRetry);
+  };
+
+  constructor() {
+    const window = PlatformSupport.getPlatform().window;
+    if (window) {
+      window.addEventListener('visibilitychange', this.visibilityHandler);
+    }
+  }
+
   // Is this AsyncQueue being shut down? If true, this instance will not enqueue
   // any new operations, Promises from enqueue requests will not resolve.
   get isShuttingDown(): boolean {
@@ -262,6 +295,10 @@ export class AsyncQueue {
     this.verifyNotFailed();
     if (!this._isShuttingDown) {
       this._isShuttingDown = true;
+      const window = PlatformSupport.getPlatform().window;
+      if (window) {
+        window.removeEventListener('visibilitychange', this.visibilityHandler);
+      }
       await this.enqueueEvenAfterShutdown(op);
     }
   }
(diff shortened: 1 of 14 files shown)


packages/firestore/src/util/async_queue.ts
```diff
@@ -279,6 +316,41 @@ export class AsyncQueue {
...
+    } else {
+      const deferred = new Deferred<void>();
+      this.firstRetryableOperation = deferred.promise;
+
+      const retryingOp = async (): Promise<void> => {
+        try {
+          await op();
+          deferred.resolve();
+          this.backoff.reset();
+          this.firstRetryableOperation = undefined;
```
wilhuff, line 343: This chain from the first retryable operation creates a weird circumstance where we could have a train of retryable operations where we wake up, succeed in performing some operation, fail in some later one, and lose the rest of the chain.

It seems like maintaining an explicit queue of these (with a head and a tail) would make this more resistant to arbitrary hiccups like this.

Or maybe I'm not understand what this code is doing exactly, in which case please add comments about how this is supposed to work.

packages/firestore/src/util/async_queue.ts
```diff
@@ -279,6 +316,41 @@ export class AsyncQueue {
...
+        } catch (e) {
+          logDebug(LOG_TAG, 'Retryable operation failed: ' + e.message);
+          this.backoff.backoffAndRun(retryingOp);
+        }
+      };
+
+      this.backoff.backoffAndRun(retryingOp);
```
wilhuff, line 350: The streams try to perform their initial operation and only call `backoffAndRun` if there's been an initial failure. I wonder if your testing difficulties could be resolved if you called `this.enqueueAndForget` or similar here for the first iteration? That way you'd only be going through the backoff after a failure.

packages/firestore/src/util/async_queue.ts
```diff
@@ -66,7 +70,13 @@ export const enum TimerId {
...
    * transactions, multiple of these may be in the queue at a given time.
    */
-  RetryTransaction = 'retry_transaction'
+  RetryTransaction = 'retry_transaction',
+
+  /**
+   * A timer used to retry operations scheduled via retryable AsyncQueue
+   * operations.
+   */
+  AsyncQueueRetry = 'async_queue_retry'
```
wilhuff, line 79: nit: the inconsistency between "retry" at the beginning vs end of the enumeration entry is a little glaring. `RetryAsyncQueue` sounds a little weird. Maybe rename `RetryTransaction` to `TransactionRetry` to match?

packages/firestore/src/util/async_queue.ts
```diff
@@ -195,6 +205,10 @@ export class AsyncQueue {
   // The last promise in the queue.
   private tail: Promise<unknown> = Promise.resolve();
 
+  // The last retryable operation. Retryable operation are run in order and
+  // retried with backoff.
+  private retryableTail = Promise.resolve();
```
wilhuff, line 210: Does this need a type of `Promise<unknown>`, as on `tail`, above?

packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -203,6 +203,85 @@ describe('AsyncQueue', () => {
...
+  it('Retries retryable operations', async () => {
+    const queue = new AsyncQueue();
+    const completedSteps: number[] = [];
+    const doStep = (n: number): void => {
+      completedSteps.push(n);
+    };
+    queue.enqueueRetryable(async () => {
+      doStep(1);
+      if (completedSteps.length === 1) {
+        throw new Error("Let's retry");
```
wilhuff, line 215: When combing through through test logs, it's useful for simulated errors to clearly label themselves so they can be easily ignored. Consider prefixing this message with "Simulated error" or similar.

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    specTest(
+      'Write is acknowledged by primary client (with recovery)',
+      ['multi-client'],
+      () => {
+        return client(0)
+          .expectPrimaryState(true)
+          .client(1)
+          .expectPrimaryState(false)
+          .userSets('collection/a', { v: 1 })
+          .failDatabase()
```
wilhuff, line 37: Why are fail and recover instructions issued from client 1? Does it matter?

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -0,0 +1,71 @@
...
+          .client(0)
+          .writeAcks('collection/a', 1, { expectUserCallback: false })
+          .client(1)
+          .recoverDatabase()
+          .runTimer(TimerId.AsyncQueueRetry)
```
wilhuff, line 42: It would be useful to see that retry can happen multiple times (i.e. that running the timer before recovery does not raise an event but still does raise an event after recovery).

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -418,6 +418,22 @@ export class SpecBuilder {
     return this;
   }
 
+  failDatabase(): this {
```
wilhuff, line 421: Could you add comments here about how this works? I suspect this works by modifying all steps going forward until recoverDatabase is called, but the code here suggests this merely modifies the current step only.

packages/firestore/src/util/async_queue.ts
wilhuff: We made it `backoffAndRun` rather than `runWithBackoff` specifically because we were intending that the first invocation is after a failure, and the action taken is to backoff then run.

The backoff delay of zero for the first iteration is intentional, at least in the stream case. We found that e.g. connections could get in a stale state or watch could return spurious errors after a connection or database had been idle for a while.

TransactionRunner seems like it's using the API wrong, at least according to the design intent. We have the same problem there where connections can be stale and fail so it's worth avoiding delay on the first failure, even for transactions.

In the case of IndexedDB failures, the initial zero delay doesn't seem as useful, but it also doesn't seem harmful. If you wanted to make the zero delay on first retry optional that would be fine with me too.

schmidt-sebastian: I think for IndexedDB failures, if would be "feel better" if we retried with backoff right away.  We already retry aborted transactions immediately, since these are likely to succeed even without backoff. Thus, I would prefer if the secondary error-code unaware retry mechanism was different and used some sort of backoff. 
It does seem like this unnecessarily complicates our tests though. I would suggest we defer this behavior change until we have actual evidence that backing off is an improvement.

PR #2883, opened 2020-04-08
Fix Messaging test that fails in Firefox

hsubox76:
A step towards fixing all cross-browser tests.  This is the only failing test for Messaging.

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.test.ts
+++ b/packages/messaging/src/controllers/sw-controller.test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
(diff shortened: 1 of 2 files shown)


packages/messaging/src/controllers/sw-controller.test.ts
```diff
@@ -277,6 +277,10 @@ describe('SwController', () => {
     });
 
     it('warns if there are more action buttons than the browser limit', async () => {
+      // This doesn't exist on Firefox.
```
Feiyang1, line 280: Any link that we can reference here?

hsubox76: Added to both comments!

PR #2888, opened 2020-04-08
Fix flaky transaction tests

thebrianchen:
b/153595228


schmidt-sebastian:
Do you mind explaining in the bug you filed why your fix works?

PR #2889, opened 2020-04-09
ComponentProvider v2

schmidt-sebastian:
So... obviously the component provider design broke the other PR I had in flight. In https://github.com/firebase/firebase-js-sdk/pull/2879, I rely on the ability to provide a custom persistence implementation in the SpecTest that can inject failures. With the ComponentProvider as implemented, I would have to build up all components from scratch. It would be nice if I only needed to change the persistence layer.

I pushed three versions:
- e01a7c3 is cleaner, but increases size by 580b (full build) / 973b  (memory)
- 9725015 moves some shared code to ComponentBuilder. This only helps the full build (+241b instead of 580b)
- f1d4a49 gets rid of the abstract class and has a size increase of 311b (full) / 725b (memory)

I personally prefer the last option (the diff of this PR).

### DIFF
--- a/packages/firestore/src/core/component_provider.ts
+++ b/packages/firestore/src/core/component_provider.ts
@@ -118,9 +118,7 @@ export class IndexedDbComponentProvider implements ComponentProvider {
     const persistenceKey = IndexedDbPersistence.buildStoragePrefix(
       cfg.databaseInfo
     );
-    const serializer = cfg.platform.newSerializer(
-      cfg.databaseInfo.databaseId
-    );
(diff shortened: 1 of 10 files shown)


wilhuff:
I wonder if there's an approach in between the two--what if you made the base class the memory-only version and then made the IndexedDB configuration extend that? That way all the common stuff could stay common but the memory-only case wouldn't pay the penalty of the extra abstract class?

schmidt-sebastian:
Updated with your suggestion.

PR #2891, opened 2020-04-09
Migrate webchannel-wrapper to gulp + google-closure-compiler

hsubox76:
Webchannel wrapper was previously using `closure-builder` which bundles its specific version of `closure-library` and is not updated often.  This led to a bug where it was configured to be compiling to an ES5 target but ignored that option (`language_out`) and compiled to ES2017. This could cause a bug in IE if the app is using untransformed code directly from `webchannel-wrapper/dist`, such as in rxfire tests run in IE.

Moving to use gulp which directly uses `google-closure-compiler` and correctly recognizes the `language_out` attribute.


packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,136 @@
...
+        closureCompiler({
+          js_output_file: filename,
+          output_wrapper: `${prefix}%output%${suffix}`,
+          entry_point: 'firebase.webchannel.wrapper',
+          compilation_level: OPTIMIZATION_LEVEL,
+          externs: [
+            resolve(__dirname, './externs/overrides.js'),
+            resolve(__dirname, './externs/module.js')
+          ],
+          language_out: 'ES5',
```
schmidt-sebastian, line 86: Could this be an option? I wonder if we would get size savings if we used ES2017 for our ES2017 build.

hsubox76: Added an ES2017 build, size comparison:

```
47365 packages/webchannel-wrapper/dist/index.esm.js
41267 packages/webchannel-wrapper/dist/index.esm2017.js
47021 packages/webchannel-wrapper/dist/index.js
```

schmidt-sebastian: Thanks! I think that is worth the effort. Do you know why the SDK size went up, despite this change?

hsubox76: The problem was that previously all builds were actually ES2017, due to a bug in `closure-builder`.  This was great for size but not great for ES5 compatibility. Now the builds that are supposed to be ES5 actually are (making them bigger) and a correctly-labeled ES2017 build has been added.

schmidt-sebastian: Makes sens.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,155 @@
...
+const del = require('del');
+const path = require('path');
+const sourcemaps = require('gulp-sourcemaps');
+const { resolve } = require('path');
+const commonjs = require('rollup-plugin-commonjs');
+const rollupSourcemaps = require('rollup-plugin-sourcemaps');
+
+// The optimization level for the JS compiler.
+// Valid levels: WHITESPACE_ONLY, SIMPLE_OPTIMIZATIONS, ADVANCED_OPTIMIZATIONS.
+// TODO: Add ability to pass this in as a flag.
```
Feiyang1, line 30: Isn't it already passed in as a flag?

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,155 @@
...
+const OPTIMIZATION_LEVEL = 'ADVANCED_OPTIMIZATIONS';
+
+// For minified builds, wrap the output so we avoid leaking global variables.
+// For minified builds, wrap the output so we avoid leaking global variables.
```
Feiyang1, line 34: duplicate comments.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,155 @@
...
+      .pipe(sourcemaps.init())
+      .pipe(
+        closureCompiler({
+          js_output_file: filename,
+          output_wrapper: `${prefix}%output%${suffix}`,
+          entry_point: 'firebase.webchannel.wrapper',
+          compilation_level: OPTIMIZATION_LEVEL,
+          externs: [
+            resolve(__dirname, './externs/overrides.js'),
+            resolve(__dirname, './externs/module.js')
```
Feiyang1, line 84: Why is `module.js` needed?

packages/webchannel-wrapper/gulpfile.js
schmidt-sebastian: I actually thought more about this - the size increase for the NPM builds are explained by these changes. The size increase for the CDN build is not though. The CDN build should have post processed everything to ES5 already. Hence, I think what is going on here is that the new build pipeline is no longer tree shaking code that Firestore isn’t using. Is that a possibility?

hsubox76: So my explanation is incorrect for the ESM bundle which was already writing ES5, I think because of using an outdated library (see below).  The bundle size increase is actually due to a change in `closure-library` from versions 20200112 to 20200204. I don't know exactly which change yet, but this led to a 5K bundle size spike on the webchannel-wrapper ES5 builds between those two `closure-library` versions.

Because webchannel-wrapper was using `closure-builder`, the ESM build was fixed on a pretty old version bundled with `closure-builder`.  Once we upgraded it to use a newer version of google-closure-library (20200224, I copied the version auth was using) it saw a 5K size increase.

If we upgrade to the latest version of google-closure-library (20200406) the size increase is smaller (2.5K) but still there.

For now we might want to pin closure-library to 20200112 (before the change that caused this size increase) and try to figure out exactly where the increase is coming from.

packages/webchannel-wrapper/gulpfile.js
hsubox76: This prevents the compiler error `ERROR - [JSC_UNDEFINED_VARIABLE] variable module is undeclared` on the last line in /src/index.js

packages/webchannel-wrapper/gulpfile.js
hsubox76: Fixed.

packages/webchannel-wrapper/gulpfile.js
hsubox76: Leftover comment from copied auth code.

hsubox76:
Pinning to closure-library and closure-compiler versions 20200112. Recent closure-library versions starting at 0204 increase the bundle size, although the most recent version (0406) has a smaller size increase.  Latest closure-compiler doesn't cause problems but better to keep versions in sync?

Unfortunately the esm2017 build using 0112 doesn't seem to be much of a size savings, but it does decrease by 3K with 0406 so seems like  it's worth keeping in there for potential size savings when we're able to upgrade closure-library.

hsubox76:
Upgraded to latest closure deps. Using closure to compile to ES2017, then Typescript to compile to ES5.  Final bundle sizes should be smaller now.  "ES5" bundles do include `Set` but we already require developers to include their own polyfills, and Saucelabs tests will include polyfills as well.

Ideally auth should upgrade closure versions as well, but this currently causes auth tests to fail and would require more investigation.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+const cjsBuild = createBuildTask(
+  intermediateCjsFile,
+  CJS_WRAPPER_PREFIX,
+  CJS_WRAPPER_SUFFIX
+);
+
+// Closure-generated ES2017 intermediate file (no wrapper text)
+const intermediateEsmFile = 'temp/esm.js';
+const intermediateEsmPath = resolve(__dirname, 'dist/', intermediateEsmFile);
+const esmBuild = createBuildTask(intermediateEsmFile, '', '');
```
Feiyang1, line 152: Does sourcemap work correctly?
I think you need to add a suffix in order to pipe the sourcemap generated by closure into rollup. `//# sourceMappingURL=esm.js.map',`. You can take a look at how it is done currently.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+          typescript,
+          // Uncomment for development only. Prevents caching between builds.
+          // clean: true,
+          compilerOptions: { allowJs: true },
+          include: ['dist/**/*.js']
+        })
+      );
+    }
+    // Typescript step unminifies minified Closure output.
+    plugins.push(terser());
```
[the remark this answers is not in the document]
hsubox76: It looks bad in terms of package size increase and technically increases the size of the npm download.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+  format = 'es'
+}) {
+  return async function rollupBuild() {
+    const plugins = [rollupSourcemaps(), commonjs()];
+    if (compileToES5) {
+      plugins.push(
+        typescriptPlugin({
+          typescript,
+          // Uncomment for development only. Prevents caching between builds.
+          // clean: true,
```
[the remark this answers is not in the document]
hsubox76: Is Typescript needed or is clean needed? Lack of `clean` seems to cause problems when trying different Typescript config options during development and it seems good to remind myself/others to use it next time we try to tweak the config.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+// The optimization level for the JS compiler.
+// Valid levels: WHITESPACE_ONLY, SIMPLE_OPTIMIZATIONS, ADVANCED_OPTIMIZATIONS.
+const OPTIMIZATION_LEVEL = 'ADVANCED_OPTIMIZATIONS';
+
+// For minified builds, wrap the output so we avoid leaking global variables.
+const CJS_WRAPPER_PREFIX = `(function() {var firebase = require('@firebase/app').default;`;
+const CJS_WRAPPER_SUFFIX =
+  `}).apply(typeof global !== 'undefined' ? ` +
+  `global : typeof self !== 'undefined' ? ` +
+  `self : typeof window !== 'undefined' ? window : {});`;
```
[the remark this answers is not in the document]
hsubox76: 1. `@firebase/app` removed
2. Seems like `gulp-sourcemap` adds this line already. (see comment below)

packages/webchannel-wrapper/gulpfile.js
hsubox76: `gulp-sourcemaps` seems to add this line as part of its process.  If I add it to the suffix I see it twice in the closure output file.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+          include: ['dist/**/*.js']
+        })
+      );
+    }
+    // Typescript step unminifies minified Closure output.
+    plugins.push(terser());
+    const inputOptions = {
+      input: inputPath,
+      plugins,
+      external: ['@firebase/app']
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/webchannel-wrapper/gulpfile.js
```diff
@@ -0,0 +1,188 @@
...
+    };
+
+    const outputOptions = {
+      file: `dist/index${outputExtension ? '.' : ''}${outputExtension}.js`,
+      format,
+      sourcemap: true,
+      exports: 'named'
```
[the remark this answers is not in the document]
hsubox76: Avoids this warning:
```
Entry module "dist/temp/cjs.js" is using named and default exports together. Consumers of your bundle will have to use `chunk["default"]` to access the default export, which may not be what you want. Use `output.exports: "named"` to disable this warning
```
But I don't see where in that file there is a default export? It's a `module.exports = { ... }` where none of the keys in object are `default`.

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
Feiyang1: `clean`. Should we always enable it then?

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
Feiyang1: Is it just removing whitespaces, I guess Typescript just reformats the already mangled code?
Anyway I don't think keeping those numbers low are that meaningful to the SDK, and it will slow down the build.

packages/webchannel-wrapper/gulpfile.js
Feiyang1: SG

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
hsubox76: I don't think it's needed if you're not changing the Typescript config, and caching saves time.

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
hsubox76: Removed Terser.

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
Feiyang1: When should I enable it? Sorry that I'm nitpicking. If it is not really needed, I think we should just remove it.

test.js
```diff
@@ -0,0 +1,16 @@
+/**
```
Feiyang1, line 1: Should this file be removed?

hsubox76: Oops, removed.

packages/webchannel-wrapper/gulpfile.js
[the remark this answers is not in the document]
hsubox76: I want to leave a note for anyone in the future who has to tinker with the build pipeline so they don't forget to enable this when working, or they will get really frustrated when changes to tsconfig don't seem to do anything. But I don't want to leave it on for production and regular usage. I don't know if a comment is the best place to leave that note, but if it's outside the code I don't think anyone will remember to check.

PR #2892, opened 2020-04-09
Add signUp API method for auth-next

avolkovi:

### Discussion

### Testing

### API Changes

    can discuss it together.


packages-exp/auth-exp/test/api/authentication/sign_up.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    const mock = mockEndpoint(Endpoint.SIGN_UP, {
+      displayName: 'my-name',
+      email: 'test@foo.com'
+    });
+
+    const response = await signUp(mockAuth, request);
+    expect(response.displayName).to.eq('my-name');
+    expect(response.email).to.eq('test@foo.com');
+    expect(mock.calls[0]).to.eql({
+      request
```
avolkovi, line 45: @samhorlbeck what do you think about having this expose headers & method as well?

samhorlbeck: Sg, that's why I made it be an object containing a request rather than just request.. Figured we'd wanna put more stuff in there at some point

packages-exp/auth-exp/src/api/authentication/sign_up.test.ts
```diff
@@ -0,0 +1,78 @@
...
+import { signUp, SignUpRequest } from './sign_up';
+import { Endpoint } from '..';
+import { ServerError } from '../errors';
+import { FirebaseError } from '@firebase/util';
+import * as mockFetch from '../../../test/mock_fetch';
+import { mockEndpoint, mockAuth } from '../../../test/api/helper';
+
+use(chaiAsPromised);
+
+describe('signUp', () => {
```
avolkovi, line 29: @samhorlbeck do you think this is sufficient coverage per endpoint? or should we test more? most of the endpoints are parameterized so there isn't much going on

samhorlbeck: Yeah I think it's fine. There's almost no actual logic in these, it's mostly type scaffolding

packages-exp/auth-exp/test/api/helper.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { Auth } from '../../src/model/auth';
+import { Endpoint } from '../../src/api';
+import { mock, Route } from '../mock_fetch';
+
+const TEST_HOST = 'localhost';
+const TEST_SCHEME = 'mock';
+const TEST_KEY = 'test-api-key';
+
+export const mockAuth: Auth = {
```
samhorlbeck, line 26: Eventually I think we will want to have this in its own test helper so that we can configure different properties on the auth object. Up to you if you want to do that now or not

avolkovi: let's consolidate after we merge https://github.com/firebase/firebase-js-sdk/pull/2896

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      : {};
+    const response = await fetch(
+      `${auth.config.apiScheme}://${auth.config.apiHost}${path}?key=${auth.config.apiKey}`,
+      {
+        method,
+        headers: {
+          'Content-Type': 'application/json'
+        },
+        referrerPolicy: 'no-referrer',
+        ...body
```
bojeil-google, line 78: GET requests do not have body component. The parameters will be passed in the query string.

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      ? {
+          body: JSON.stringify(request)
+        }
+      : {};
+    const response = await fetch(
+      `${auth.config.apiScheme}://${auth.config.apiHost}${path}?key=${auth.config.apiKey}`,
+      {
+        method,
+        headers: {
+          'Content-Type': 'application/json'
```
bojeil-google, line 75: Not all APIs will have json content type. This is true for secure token endpoint.

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  customErrorMap: Partial<ServerErrorMap<ServerError>> = {}
+): Promise<V> {
+  const errorMap = { ...SERVER_ERROR_MAP, ...customErrorMap };
+  try {
+    const body = request
+      ? {
+          body: JSON.stringify(request)
+        }
+      : {};
+    const response = await fetch(
```
bojeil-google, line 70: You seem to hardcode the `fetch` call. I suppose you can polyfill this. But not sure this can be polyfilled for all environments (eg. Node.js, react-native). What is your plan regarding this.

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -0,0 +1,126 @@
...
+      `${auth.config.apiScheme}://${auth.config.apiHost}${path}?key=${auth.config.apiKey}`,
+      {
+        method,
+        headers: {
```
bojeil-google, line 74: Header should also contain the `X-Client-Version`. This is needed for our server to understand the source of the request. I assume you will add this later.

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -0,0 +1,126 @@
...
+  GET_ACCOUNT_INFO = '/v1/accounts:lookup',
+  GET_RECAPTCHA_PARAM = '/v1/recaptchaParams',
+  START_PHONE_MFA_ENROLLMENT = '/v2/accounts/mfaEnrollment:start',
+  FINALIZE_PHONE_MFA_ENROLLMENT = '/v2/accounts/mfaEnrollment:finalize',
+  START_PHONE_MFA_SIGN_IN = '/v2/accounts/mfaSignIn:start',
+  FINALIZE_PHONE_MFA_SIGN_IN = '/v2/accounts/mfaSignIn:finalize',
+  WITHDRAW_MFA = '/v2/accounts/mfaEnrollment:withdraw'
+}
+
+export async function performApiRequest<T, V>(
```
bojeil-google, line 56: Also I recommend you add a timeout for these calls and a way to catch it and map it to the correct error.

packages-exp/auth-exp/src/api/index.ts
avolkovi: I don't think our prototype ever required that endpoint, what content type does it use? I suppose we can allow this to be passed in

packages-exp/auth-exp/src/api/index.ts
avolkovi: Will fix in https://github.com/firebase/firebase-js-sdk/pull/2901

packages-exp/auth-exp/src/api/index.ts
avolkovi: We are investigating react-native right now, suppose this can be passed in from the Auth object like we do with popupRedirectResolver

packages-exp/auth-exp/src/api/index.ts
avolkovi: Will fix in #2901

packages-exp/auth-exp/src/api/index.ts
avolkovi: will do! thanks

PR #2896, opened 2020-04-10
Add initial user object implementation

samhorlbeck:


yarn.lock
```diff
@@ -10397,6 +10397,36 @@ mocha@7.1.0:
     yargs-parser "13.1.1"
     yargs-unparser "1.6.0"
 
+mocha@^7.1.1:
```
avolkovi, line 10400: what's up with all the changes here?

packages-exp/auth-exp/test/mock_auth.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AppName, ApiKey, Auth } from '../src/model/auth';
+
+export function mockAuth(name: AppName, apiKey: ApiKey): Auth {
```
avolkovi, line 20: I have something similar in my api_helper branch...

packages-exp/auth-exp/src/core/user_impl.ts
```diff
@@ -0,0 +1,70 @@
...
+  constructor({ uid, auth, ...opt }: UserParameters) {
+    this.uid = uid;
+    this.auth = auth;
+    this.displayName = opt.displayName || null;
+    this.email = opt.email || null;
+    this.phoneNumber = opt.phoneNumber || null;
+    this.photoURL = opt.photoURL || null;
+  }
+
+  getIdToken(forceRefresh?: boolean): Promise<string> {
```
avolkovi, line 55: should we go ahead and pull the implementations now? or do they have too many dependencies?

yarn.lock
samhorlbeck: Idk but apparently you're supposed to commit yarn.lock?
https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

packages-exp/auth-exp/test/mock_auth.ts
samhorlbeck: I'll wait for yours to merge and use it

packages-exp/auth-exp/src/core/user_impl.ts
samhorlbeck: These are gonna end up being used for just about everything.

I just added an token manager. This PR is pretty small so it seemed like the best option

yarn.lock
avolkovi: only if you change package.json though, right?

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  describe('#isExpired', () => {
+    it('is true if past expiration time', () => {
+      stsTokenManager.expirationTime = 1; // Ancient history
+      expect(stsTokenManager.isExpired).to.eq(true);
+    });
+
+    it('is true if exp is in future but within buffer', () => {
+      // Buffer is 30_000
+      stsTokenManager.expirationTime = now + 20_000;
```
avolkovi, line 48: is it possible to pull in the buffer as a variable here?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -0,0 +1,70 @@
...
+ * to refresh that token, to provide a buffer for RPCs to complete.
+ */
+const TOKEN_REFRESH_BUFFER_MS = 30_000;
+
+export interface Tokens {
+  accessToken: string;
+  refreshToken: string | null;
+}
+
+export const DATE_GENERATOR = {
```
avolkovi, line 31: is this a common pattern in JS to allow stubbing?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -0,0 +1,70 @@
...
+  now: () => Date.now()
+};
+
+export class StsTokenManager {
+  refreshToken: string | null = null;
```
avolkovi, line 36: do we have a standard pattern on when we use | null vs foo? ?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -0,0 +1,70 @@
...
+    return (
+      !this.expirationTime ||
+      DATE_GENERATOR.now() > this.expirationTime - TOKEN_REFRESH_BUFFER_MS
+    );
+  }
+
+  updateFromServerResponse({
+    idToken,
+    refreshToken,
+    expiresIn
```
avolkovi, line 50: can we name this expiresInSec to have the units be obvious?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -0,0 +1,70 @@
...
+      return {
+        accessToken: this.accessToken,
+        refreshToken: this.refreshToken
+      };
+    }
+
+    throw new Error('StsTokenManager: token refresh not implemented');
+  }
+
+  // TODO: There are a few more methods in here that need implemented
```
avolkovi, line 69: can you list out the methods that are missing?

packages-exp/auth-exp/src/core/user/user_impl.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      const user = new UserImpl({ uid: 'uid', auth, stsTokenManager });
+      const token = await user.getIdToken();
+      expect(token).to.eq('id-token-string');
+      expect(user.refreshToken).to.eq('refresh-token-string');
+    });
+
+    it('throws if refresh is required', async () => {
+      const user = new UserImpl({ uid: 'uid', auth, stsTokenManager });
+      await expect(user.getIdToken()).to.be.rejectedWith(Error);
```
avolkovi, line 84: can you add the error type & message expected so it's not just a generic error? that way we know it's failing as expected instead of due to some other issue

samhorlbeck: It's only a "temporary" error because it's a method stub. When I actually implement those there'll be actual tests for the legit errors that may occur. Do you still want the test to check for the "not implemented" error specifically?

avolkovi: oh yeah, I guess you're right, I thought this was was an actual expected throw.  Up to you whether we check for the 'not implemented' string, IMO it would be easier to grep through tests to see which ones need updating, but it's a nit at best

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: We don't but we should settle on one. Either is acceptable but we should be consistent.

I personally prefer null when a field _should always be present_ but *can* be null. There is a semantic difference between:
`const a = {b: undefined}`
and `const a = {}`

avolkovi: I like that, in that case we should probably use `| null` in most places

yarn.lock
samhorlbeck: Reverted it.

packages-exp/auth-exp/src/core/user/token_manager.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: Ehh it's a thing to do in last resort. I was dumb though it's fine to just stub the `Date` object.

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  updateFromServerResponse({
+    idToken,
+    refreshToken,
+    expiresIn: expiresInSec
+  }: IdTokenResponse): void {
+    this.refreshToken = refreshToken;
+    this.accessToken = idToken;
+    this.expirationTime = Date.now() + Number.parseInt(expiresInSec, 10) * 1000;
```
Feiyang1, line 50: We follow [go/tsstyle ](go/tsstyle#type-coercion)and use `Number()` instead of `parseInt` and `parseFloat` to parse numeric values.

samhorlbeck: Done. Lol I should know this given I'm a readability approver for ts

PR #2905, opened 2020-04-14
Prefix list results with correct bucket

schmidt-sebastian:
We need to use the same bucket in the list response as in the list request.

Fixes https://b.corp.google.com/issues/152053124


schmidt-sebastian:
@hsubox76 Do you mind approving this?

Issue #2755, opened 2020-03-17
Offline Mode Improvements


nVitius:
I wanted to propose a few improvements that might be made to the SDK's Firestore Offline Persistence functionality.

To provide a little context, I work for a small organization using Firebase across 5 separate web applications (also wrapped as native applications through WKWebView and Android WebView) servicing about 2000 users. One of the main driving factors behind our decision to implement Firebase was the ease of implementing offline data access; it is very important to us.

Currently, there is an issue across most devices/browsers where Firestore can lose connection unexpectedly to IndexedDB and cannot recover without a restart ( #2711 #2710 #1533 #1926 #2581 ).

A comment from @schmidt-sebastian on the latest issue:
> You should see this a lot less with SDK versions 7.2.1 and later. It does however seem that we were not able to completely solve this. If you do encounter this issue with the latest Firestore SDK, you will unfortunately have to reload your PWA/webpage or create a new Firestore instance.

> We are exploring ways to re-architect our SDK to be more robust agains IndexedDb failures in general, but this will take us a bit of time.

I understand that this issue is being worked on and it is rather difficult to address as it might require re-writing a large part of the SDK. That said, for a product that is no longer in beta, it is unacceptable ti have issues that cause data loss and are unrecoverable (even if it happens "a lot less"). For our applications, we use a session recording software that allows us to recover data for a user when this happens based on their logs. But for a lot of people that are not aware this is happening (there is no warnings on the documentation that offline persistence is broken) this means permanent data loss.

Of course, there is always the option to disable offline persistence. But I do not think that is an option for everyone. For us at least, it is one of the major selling points of our apps.

While a fix is being worked on, these are some improvements I think could be made to the system that might not require a lot of work and would make everyone's lives easier.

1) Error Handling
Currently, the only way to handle this error is to catch it at the Window level and then reload the page. The issue with this is that there is no way for me to know which of my `.set()` calls failed. When/if I reload the app, that data is still lost. Ideally, the error would (also?) be thrown to the original caller so that the application could easily identify which specific queries have failed. This would allow one to store the data manually and retry saving it when the app is reloaded.

2) Trigger Failure Manually
Given that this issue is very difficult to reproduce consistently in a real-world scenario, I think it is important to add tools to the SDK so that developers can test their solutions inside of their actual apps instead of hoping that it will work when the failure happens in produdtion.

3) Notify Client When Going Offline
[...]

PR #2916, opened 2020-04-16
Make 'handleClientStateEvent()/handleQueryTargetEvent()' idempotent

schmidt-sebastian:
This solves a problem where we were able to modify a in-memory state before the database transaction failed and also addresses issues where targets to allocate already existed or rejected targets were missing. 

I believe there are no other issues in the WebStorage callbacks.

Addresses #2755

### DIFF
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -879,11 +879,10 @@ export class LocalStore {
     keepPersistedTargetData: boolean
   ): Promise<void> {
     const targetData = this.targetDataByTarget.get(targetId);
-
-    if (targetData === null) {
-      logDebug(`Ignoring release of nonexistent target: ${targetId}`);
-      return Promise.resolve();
-    }
+    debugAssert(
+      targetData !== null,
+      `Tried to release nonexistent target: ${targetId}`
+    );
 
     const mode = keepPersistedTargetData ? 'readwrite' : 'readwrite-primary';
     return this.persistence
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -879,10 +879,11 @@ export class LocalStore {
     keepPersistedTargetData: boolean
   ): Promise<void> {
     const targetData = this.targetDataByTarget.get(targetId);
-    debugAssert(
-      targetData !== null,
-      `Tried to release nonexistent target: ${targetId}`
-    );
+
+    if (targetData === null) {
+      logDebug(`Ignoring release of nonexistent target: ${targetId}`);
+      return Promise.resolve();
+    }
 
     const mode = keepPersistedTargetData ? 'readwrite' : 'readwrite-primary';
     return this.persistence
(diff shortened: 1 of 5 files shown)


packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -75,5 +75,38 @@ describeSpec(
...
+            .failDatabase()
+            .client(1)
+            .userUnlistens(query)
+            .client(0)
+            // The primary client 0 receives a notification that the query can
+            // be released, but it can only process the change after we recover
+            // the database.
+            .recoverDatabase()
+            .runTimer(TimerId.AsyncQueueRetry)
+            .expectActiveTargets()
```
wilhuff, line 107: Should we expect that there is an active target before recoverDatabase?

packages/firestore/src/local/local_store.ts
```diff
@@ -879,10 +879,11 @@ export class LocalStore {
     keepPersistedTargetData: boolean
   ): Promise<void> {
     const targetData = this.targetDataByTarget.get(targetId);
-    debugAssert(
-      targetData !== null,
-      `Tried to release nonexistent target: ${targetId}`
-    );
+
+    if (targetData === null) {
```
wilhuff, line 883: How would this happen in practice that isn't a bug? I mean, how could we get in a state where the sync engine thought we were listening but local store didn't?

packages/firestore/test/unit/specs/recovery_spec.test.ts
schmidt-sebastian: The targets are checked at each step, but adding this here explicitly makes it much clearer what is going on. Done.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I was trying to guard against this line: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/sync_engine.ts#L1115

I did miss that this is only executed if the target is still active. I reverted this change.

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -75,5 +75,39 @@ describeSpec(
...
+            .runTimer(TimerId.AsyncQueueRetry)
+            .expectListen(query)
+            .failDatabase()
+            .client(1)
+            .userUnlistens(query)
+            .client(0)
+            // The primary client 0 receives a notification that the query can
+            // be released, but it can only process the change after we recover
+            // the database.
+            .expectActiveTargets({ query})
```
wilhuff, line 105: nit: should there be a space between `query` and `}`?

schmidt-sebastian: Yes. As you may have noticed, the `prettier` pre-commit hook doesn't work for me at all. I have been working with @hsubox76 to see if this can be fixed, but it seems like the behavior I am seeing is actually the indented behavior. Pre-commit hooks are only meant to push changes that existed prior to running.

PR #2918, opened 2020-04-16
React Native Integration

scottcrossen:


packages/auth-ts/src/core/persistence/react_local.ts
```diff
@@ -0,0 +1,52 @@
...
+  Instantiator
+} from '../persistence';
+
+import AsyncStorage from '@react-native-community/async-storage';
+
+class ReactNativeLocalPersistence implements Persistence {
+  type: PersistenceType = PersistenceType.LOCAL;
+
+  async isAvailable(): Promise<boolean> {
+    return true;
```
avolkovi, line 31: this should check if actually running in react native

packages/auth-ts/rollup.config.js
```diff
@@ -53,6 +53,19 @@ const es5Builds = [
     output: [{ file: pkg.main, format: 'cjs', sourcemap: true }],
     plugins: es5BuildPlugins,
     external: id => deps.some(dep => id === dep || id.startsWith(`${dep}/`))
+  },
+  /**
+   * App React Native Builds
+   */
+  {
+    input: 'src/legacy_polyfill.rn.ts',
```
avolkovi, line 61: old SDK only had one file for everything, so we should only have one legacy polyfill that includes everything

packages/auth-ts/package.json
```diff
@@ -66,6 +67,7 @@
   },
   "dependencies": {
     "@firebase/util": "^0.2.36",
+    "@react-native-community/async-storage": "1.9.0",
```
avolkovi, line 70: the original SDK doesn't have this as a dependency, can you see how they did it?

packages/auth-ts/package.json
```diff
@@ -8,6 +8,7 @@
   "browser": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "esm2017": "dist/index.esm2017.js",
+  "react-native": "dist/index.rn.cjs.js",
```
avolkovi, line 11: old SDK doesn't provide a react native build

packages/auth-ts/src/core/persistence/react_local.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  Persistence,
+  PersistenceType,
+  PersistenceValue,
+  Instantiator
+} from '../persistence';
+
+import AsyncStorage from '@react-native-community/async-storage';
+
+class ReactNativeLocalPersistence implements Persistence {
```
avolkovi, line 27: can you add unit tests?

packages/auth-ts/package.json
scottcrossen: They have the deprecated version of this as a dependency

packages/auth-ts/src/core/persistence/react_local.ts
scottcrossen: We dependency-inject this in for builds of react-native. If this file exists then it is running in react native. Would you me to make an additional check via the helper we've created?

avolkovi: for non react-native builds this function will still be available through the legacy polyfill, in those cases it should return false

packages/auth-ts/rollup.config.js
scottcrossen: Haven't we established that the old SDK did things wrong? I don't think that's good justification for making one entrypoint -- especially since the configuration of the other rollup files (app, firebase) do this as well for tree-shaking purposes. If you would like me to only have one file then we are going to have to access closure to access the dependency injected libraries we need (in extended namespace firebase.INTERNAL.reactNative.AsyncStorage). Is that what you want?

packages/auth-ts/package.json
avolkovi: is it specified somewhere other than https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/package.json?

packages/auth-ts/rollup.config.js
avolkovi: for the polyfill we need to support the old API surface which was "wrong", the new SDK API provides multiple entry points

packages/auth-ts/package.json
scottcrossen: That's because the old SDK doesn't have a rollup config and doesn't support tree shaking. packages/app and packages/firebase have the react native build that the old SDK detects and abuses

packages/auth-ts/src/core/persistence/react_local.ts
scottcrossen: I'll add in a followup PR once we establish how we would like the new SDK to be structured (this PR is finalized). I thought we were tracking these separately on the sheet so I didn't add unit tests.

packages/auth-ts/package.json
scottcrossen: Is what specified?

packages/auth-ts/src/core/persistence/react_local.ts
scottcrossen: So I suck at closure: Are you saying that `fireauth.storage.AsyncStorage.prototype.get` was part of the public API for the SDK?

packages/auth-ts/package.json
scottcrossen: Oh so this is my bad: This shouldn't be in package.json regardless. I didn't realize you were commenting on package.json

packages/auth-ts/rollup.config.js
Feiyang1: In the firebase polyfill/compat layer, I want to avoid recreating the legacy internal APIs(used only by auth today) unless absolutely necessary, so it would be great if you can handle the dependencies in auth itself.

For a rn specific entry point, I don't feel strongly one way or the other. The plan is to support polyfills for 1 major release to allow people to upgrade to the modular SDKs piecemeal, so they doesn't have to be super optimized.

Feiyang1: So I tried to bundle `@react-native-community/async-storage` and failed because its main field points to a [Flow](https://flow.org/en/) file. 
You might get it to work by compiling the Flow file first, but It clearly says to me that user shouldn't bundle `@react-native-community/async-storage`.

scottcrossen: For the compat layer we will be using the old 'react-native' dependency as before. The new PR (https://github.com/firebase/firebase-js-sdk/pull/2947) has this.

PR #2919, opened 2020-04-16
Only retry DOMException/DOMError

schmidt-sebastian:
Limit the number of errors that can be retried on the AsyncQueue to those thrown by browsers.

Addresses #2755


packages/firestore/package.json
```diff
@@ -20,8 +20,8 @@
     "test:all": "run-p test:browser test:travis test:minified",
     "test:browser": "karma start --single-run",
     "test:browser:debug": "karma start --browsers=Chrome --auto-watch",
-    "test:node": "FIRESTORE_EMULATOR_PORT=8080 FIRESTORE_EMULATOR_PROJECT_ID=test-emulator TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file index.node.ts --opts ../../config/mocha.node.opts",
-    "test:node:prod": "TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file index.node.ts --opts ../../config/mocha.node.opts",
+    "test:node": "FIRESTORE_EMULATOR_PORT=8080 FIRESTORE_EMULATOR_PROJECT_ID=test-emulator TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file index.node.ts --file test/util/node_persistence.ts --opts ../../config/mocha.node.opts",
```
schmidt-sebastian, line 23: While this seems wrong, this is needed to pull in `DOMException`. Since `USE_MOCK_PERSISTENCE` is not set, there are no other side effects. I chose not to rename this file since it is used in our test launchers.

packages/firestore/src/util/async_queue.ts
```diff
@@ -337,8 +338,19 @@ export class AsyncQueue {
           deferred.resolve();
           this.backoff.reset();
         } catch (e) {
-          logDebug(LOG_TAG, 'Retryable operation failed: ' + e.message);
-          this.backoff.backoffAndRun(retryingOp);
+          // We only retry errors that match the ones thrown by IndexedDB (see
+          // https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/error)
+          if (
+            (typeof DOMException !== 'undefined' &&
```
wilhuff, line 344: `DOMException` can mean just about any failure in the browser. A better way to handle this would be to take the [`transaction.error` property](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/simple_db.ts#L438) and wrap it in our own `PersistenceError` error type (or similar). That way it's unambiguous that it's an error that came from IndexedDB. This also sidesteps the issue of the availability of the `DOMException` type.

schmidt-sebastian: Done. I am now wrapping exceptions thrown by `abort` and `error` in their own types.

packages/firestore/src/local/simple_db.ts
```diff
@@ -406,6 +406,18 @@ export interface IterateOptions {
...
 }
 
+/** An error that wraps exceptions that thrown during IndexedDB execution. */
+export class IndexedDbTransactionError extends FirestoreError {
+  name = 'IndexedDbTransactionError';
+
+  constructor(cause: Error) {
+    super(
+      Code.UNKNOWN,
+      'IndexedDB transaction failed with environment error: ' + cause
```
wilhuff, line 416: Drop "with environment error". "IndexedDB transaction failed" is enough to indicate what's failing and the cause will carry the details.

packages/firestore/src/local/simple_db.ts
```diff
@@ -406,6 +406,18 @@ export interface IterateOptions {
   reverse?: boolean;
 }
 
+/** An error that wraps exceptions that thrown during IndexedDB execution. */
+export class IndexedDbTransactionError extends FirestoreError {
+  name = 'IndexedDbTransactionError';
+
+  constructor(cause: Error) {
+    super(
+      Code.UNKNOWN,
```
wilhuff, line 415: `UNAVAILABLE` could be a better code for this given what we know about when this happens.

packages/firestore/src/util/async_queue.ts
```diff
@@ -318,9 +319,10 @@ export class AsyncQueue {
   /**
    * Enqueue a retryable operation.
    *
-   * A retryable operation is rescheduled with backoff if it fails with any
-   * exception. All retryable operations are executed in order and only run
-   * if all prior operations were retried successfully.
+   * A retryable operation is rescheduled with backoff if it fails with a
+   * SimpleDbTransactionError (the error type used by SimpleDb). All
```
wilhuff, line 323: The error is actually IndexedDbTransactionError.

PR #2924, opened 2020-04-16
 Add fetchSignInMethodsForEmail to auth-next

avolkovi:


packages-exp/auth-exp/src/core/strategies/email.ts
```diff
@@ -0,0 +1,41 @@
...
+  // createAuthUri returns an error if continue URI is not http or https.
+  // For environments like Cordova, Chrome extensions, native frameworks, file
+  // systems, etc, use http://localhost as continue URL.
+  const continueUri = isHttpOrHttps() ? getCurrentUrl() : 'http://localhost';
+  const request: CreateAuthUriRequest = {
+    identifier: email,
+    continueUri
+  };
+
+  const response = await createAuthUri(auth, request);
```
samhorlbeck, line 38: You can just destructure this here
`const {signInMethods} = await ...`

packages-exp/auth-exp/src/core/util/location.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function getCurrentUrl(): string {
+  return window?.location?.href || self?.location?.href || '';
```
samhorlbeck, line 19: Isn't self just a direct alias to window?

avolkovi: apparently window doesn't exist in workers but self does, this should probably just use the self version though

avolkovi: https://developer.mozilla.org/en-US/docs/Web/API/Window/self

packages-exp/auth-exp/src/core/strategies/email.ts
avolkovi: done

PR #2925, opened 2020-04-16
Add a persistence manager class

samhorlbeck:


packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+
+export class PersistenceUserManager {
+  private constructor(
+    public persistence: Persistence,
+    private readonly auth: Auth,
+    private readonly userKey: string
+  ) {}
+
+  fullKeyName_(key: string): string {
+    const { config, name } = this.auth;
```
avolkovi, line 44: i feel like you can just one line this, you don't use the references for anything else

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+    return persistenceKeyName_(key, config.apiKey, name);
+  }
+
+  setCurrentUser(user: User): Promise<void> {
+    return this.persistence.set(this.fullKeyName_(this.userKey), user);
```
avolkovi, line 49: should we just store the fullKeyName on the object? then you don't have to keep calling this. it can't change anyway

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  getCurrentUser(): Promise<User | null> {
+    return this.persistence.get<User>(
+      this.fullKeyName_(this.userKey),
+      (o: { [key: string]: unknown }) => UserImpl.fromPlainObject(this.auth, o)
```
avolkovi, line 55: what's `o`?

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (this.persistence.type === newPersistence.type) {
+      return;
+    }
+
+    const currentUser = await this.getCurrentUser();
+    await this.removeCurrentUser();
+
+    this.persistence = newPersistence;
+
+    if (currentUser) {
```
avolkovi, line 80: what's the return type if currentUser is null?

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -110,4 +111,40 @@ describe('core/user/token_manager', () => {
       expect(tokens.refreshToken).to.eq('refresh');
     });
   });
+
+  describe('fromPlainObject', () => {
+    const fpo = (obj: { [key: string]: unknown }) =>
```
avolkovi, line 116: should we name this type? you use it in the implementation as well

avolkovi: also why do we need this alias?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -61,6 +62,38 @@ export class StsTokenManager {
...
+      { appName }
+    );
+    if (refreshToken && typeof refreshToken !== 'string') {
+      throw internalError;
+    }
+    if (accessToken && typeof accessToken !== 'string') {
+      throw internalError;
+    }
+    if (expirationTime && typeof expirationTime !== 'number') {
+      throw internalError;
```
avolkovi, line 89: should be assertions instead?

packages-exp/auth-exp/src/core/user/user_impl.test.ts
```diff
@@ -105,4 +106,46 @@ describe('core/user/user_impl', () => {
       expect(() => user.delete()).to.throw();
     });
   });
+
+  describe('fromPlainObject', () => {
+    const fpo = UserImpl.fromPlainObject;
```
avolkovi, line 111: i'm not sure I like these aliases, can we just call the function directly? or import it?

packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -80,4 +81,45 @@ export class UserImpl implements User {
...
+      return value;
+    }
+
+    if (typeof uid !== 'string' || !plainObjectTokenManager) {
+      throw internalError;
+    }
+
+    const stsTokenManager = StsTokenManager.fromPlainObject(
+      auth.name,
+      plainObjectTokenManager as { [key: string]: unknown }
```
avolkovi, line 112: here's that type again

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.test.ts
```diff
@@ -0,0 +1,182 @@
...
+import { UserImpl } from '../user/user_impl';
+
+chai.use(sinonChai);
+
+function makePersistence(
+  type = PersistenceType.NONE
+): {
+  persistence: Persistence;
+  stub: sinon.SinonStubbedInstance<Persistence>;
+} {
```
avolkovi, line 35: can we name this interface, even it is just for tests

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        'firebase:authUser:test-api-key:test-app'
+      );
+    });
+
+    it('#getCurrentUser calls with instantiator', async () => {
+      const rawObject = {};
+      const userImplStub = sinon.stub(UserImpl, 'fromPlainObject');
+      persistenceStub.get.callsFake(
+        (_: string, cb?: (o: { [key: string]: unknown }) => any) => {
+          return cb!(rawObject);
```
avolkovi, line 135: what's cb? why do we need to stub this other method?

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
samhorlbeck: setCurrentUser returns a Promise<void>. It needs to be returned or awaited, otherwise it's a hanging promise

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: The assertion in @firebase/util comes out as a database error which seems wrong

samhorlbeck: https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/assert.ts

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
samhorlbeck: Now it is used twice

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
samhorlbeck: Done, we do need two different versions but it makes sense to just store both rather than keep calling this

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
samhorlbeck: the blob. Anyway this is an instantiator. should be clear now

packages-exp/auth-exp/src/core/user/token_manager.test.ts
samhorlbeck: Dropped the alias, added that thing as a type: "PersistedBlob"

packages-exp/auth-exp/src/core/user/user_impl.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/user_impl.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.test.ts
samhorlbeck: Discussed offline, keeping it this way

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.test.ts
samhorlbeck: Done. Added a comment which should make this clear. The goal of this test is to ensure the instantiator is calling through to UserImpl.fromPlainObject

packages-exp/auth-exp/src/core/user/token_manager.ts
avolkovi: oh weird, maybe we should add an authAssert?

samhorlbeck: Done

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { AUTH_ERROR_FACTORY, AuthErrorCode } from '../errors';
+
+export function assert<T>(
```
samhorlbeck, line 20: @avolkovi mind giving a look at these?

avolkovi: looks ok... though it wonder if there's some way to have this handle the `string | undefined` case above

samhorlbeck: Done, now it can check multiple types.

avolkovi: 💯

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,110 @@
...
+import { User } from '../../model/user';
+import { ApiKey, AppName, Auth } from '../../model/auth';
+import { inMemoryPersistence } from './in_memory';
+import { UserImpl } from '../user/user_impl';
+
+export const AUTH_USER_KEY_NAME_ = 'authUser';
+export const PERSISTENCE_KEY_NAME_ = 'persistence';
+const PERSISTENCE_NAMESPACE_ = 'firebase';
+
+export function persistenceKeyName_(
```
Feiyang1, line 28: I don't think we have an official rule on prefix _ vs suffix _ for internal functions/constants, but I'd like to establish the convention of using prefix _. `@firebase/firestore` and `@firebase/app` are already using prefix _ today and the documentation tool we are going to use will [flag functions with `@internal` annotation without prefix _](https://api-extractor.com/pages/tsdoc/tag_internal/).  
WDYT?

samhorlbeck: Sg, done

packages-exp/auth-exp/src/core/persistence/index.ts
```diff
@@ -15,27 +15,26 @@
...
+  [key: string]: unknown;
+}
+
 export interface Instantiator<T> {
-  (blob: { [key: string]: unknown }): T;
+  (blob: PersistedBlob): T;
 }
 
-export type PersistenceValue = PersistenceType | User;
+export type PersistenceValue = PersistedBlob | string;
```
avolkovi, line 32: this is way better 👍

packages-exp/auth-exp/src/core/persistence/indexed_db.ts
```diff
@@ -16,20 +16,20 @@
...
   Persistence,
   PersistenceType,
-  PersistenceValue,
-  Instantiator
-} from '.';
+  PersistenceValue
+} from './';
 
-const STORAGE_AVAILABLE_KEY_ = '__sak';
+const _STORAGE_AVAILABLE_KEY = '__sak';
```
avolkovi, line 25: should this live in index? it's also declare in browser

packages-exp/auth-exp/src/core/persistence/indexed_db.ts
```diff
@@ -16,20 +16,20 @@
...
 
 export const DB_NAME = 'firebaseLocalStorageDb';
 const DB_VERSION = 1;
 const DB_OBJECTSTORE_NAME = 'firebaseLocalStorage';
 const DB_DATA_KEYPATH = 'fbase_key';
 
-type DBValue = { [key: string]: unknown } | string;
+type DBValue = PersistedBlob | string;
```
avolkovi, line 32: can we just use PersistenceValue here?

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const key = persistenceKeyName_(userKey, auth.config.apiKey, auth.name);
+    for (const persistence of persistenceHierarchy) {
+      if (await persistence.get(key)) {
+        return new PersistenceUserManager(persistence, auth, userKey);
+      }
+    }
+
+    // Check all the available storage options.
+    // TODO: Migrate from local storage to indexedDB
```
avolkovi, line 102: we should add these to the sheet so we don't forget

packages-exp/auth-exp/src/core/persistence/indexed_db.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/indexed_db.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -0,0 +1,109 @@
...
+import { User } from '../../model/user';
+import { PersistedBlob, Persistence } from '../persistence';
+import { UserImpl } from '../user/user_impl';
+import { inMemoryPersistence } from './in_memory';
+
+export const _AUTH_USER_KEY_NAME = 'authUser';
+export const _PERSISTENCE_KEY_NAME = 'persistence';
+const _PERSISTENCE_NAMESPACE = 'firebase';
+
+export function persistenceKeyName_(
```
Feiyang1, line 28: nit: change to prefix _

samhorlbeck: woops done

PR #2927, opened 2020-04-16
Add the umbrella firebase package for modular packages

Feiyang1:


packages-exp/firebase-exp/gulpfile.js
```diff
@@ -0,0 +1,35 @@
+/**
```
hsubox76, line 1: Is this the CDN build? Is it not going to include polyfills anymore? Should we still `registerVersion` as 'cdn' to track when this build is used?

Feiyang1: Yes, it's the `firebase.js` file. I don't think we need to include polyfills for several reasons:
1. This file is intended for development and most people would use newer browsers that don't need polyfills.
2. It discourages people to use it in production because it won't work in "all" browsers out of box.
3. People can add polyfills themselves if needed. And I want us to get out of the business of deciding what polyfills to include and maintaining them forever.

Yup, good idea that we should keep tracking its usage. Will update.

Feiyang1: Added `packages-exp/firebase-exp/app/index.cdn.ts`. PTAL.

scripts/exp/ts-transform-import-path.js
```diff
@@ -46,7 +46,7 @@ function visitNodeAndChildren(node, context) {
 
 function visitNode(node) {
   let importPath;
-  if (ts.isImportDeclaration(node)) {
+  if (ts.isImportDeclaration(node) || ts.isExportDeclaration(node)) {
     const importPathWithQuotes = node.moduleSpecifier.getText();
```
hsubox76, line 50: Seems to be causing a build error, need to do a null check on `moduleSpecifier`?

Feiyang1: yup, just fixed.

PR #2931, opened 2020-04-17
Strip messages for fail() and hardAssert()

schmidt-sebastian:
Goal: Reduce size of SDK by stripping error messages that should not show up in production builds. We can still trace the original error by looking at the stacktrace.

### DIFF
--- a/packages/firestore/rollup.shared.js
+++ b/packages/firestore/rollup.shared.js
@@ -44,8 +44,8 @@ const externsPaths = externs.map(p => path.resolve(__dirname, '../../', p));
 const publicIdentifiers = extractPublicIdentifiers(externsPaths);
 
 /**
- * Transformers that remove all debugAsserts and appends a __PRIVATE_ prefix to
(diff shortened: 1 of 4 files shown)


wilhuff:
Are you sure we'll be able to examine a stack trace to see this? In minified builds the stack is gibberish.

It's worth going through the exercise of triggering a failure (perhaps by inverting one of the conditions), doing a minified build, and then seeing what the error actually looks like.

schmidt-sebastian:
Unlike Android, the Web ecosystem is actually really well prepared to deal with mangled and minified sources. Since we ship sourcemaps, it is relatively straightforward to trace crashes back to their original location. To show this, I negated one of the assert in MemoryMutationQueue. This is the view I get by default in Chrome (no changes needed):

Screen Shot 2020-04-20 at 2 33 12 PM

While the stacktrace is mangled, it links back to the original source, which includes the original error message even though we have stripped it.

PR #2932, opened 2020-04-17
Initial auth object implementation + initializeAuth()

samhorlbeck:


packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  it('throws an API error if key not provided', () => {
+    expect(() =>
+      initializeAuth({
+        ...FAKE_APP,
+        options: {} // apiKey is missing
+      })
+    ).to.throw(
+      FirebaseError,
+      'Firebase: Your API key is invalid]: please check you have copied it correctly. (auth/invalid-api-key).'
```
avolkovi, line 134: we'll need to either fix the error messages or resolve during merge conflict later

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -0,0 +1,124 @@
...
+  constructor(
+    public readonly name: string,
+    public readonly config: Config,
+    persistenceHierarchy: Persistence[]
+  ) {
+    this.operations = Promise.resolve();
+
+    // This promise is intended to float; auth initialization happens in the
+    // background, meanwhile the auth object may be used by the app.
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
avolkovi, line 50: will this still call the promise? is there a way to test that?

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
samhorlbeck: Yeah I'll hold off and rebase onto your fix when it merges

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: It will, promises always execute. There's no concept of "hot" and "cold" promises like there are with observables. Since this.operations is initialized as `Promise.resolve()`, and `queue()` just puts the callback in a `.then()`, the initial promise is resolved and so the callback here will immediately execute.

There's not really a way to test that it runs on its own since the test has to `await` _something_, but my tests do cover this code by awaiting a later operation (but note it's _not_ the act of awaiting that triggers this code to run)

packages-exp/auth-exp/src/model/auth.d.ts
```diff
@@ -31,6 +31,15 @@ export interface Config {
 }
 
 export interface Auth {
```
Feiyang1, line 33: Will the public types eventually be moved to `@firebase/auth-types-exp` ?

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -0,0 +1,192 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect, use } from 'chai';
+import * as sinon from 'sinon';
+import * as sinonChai from 'sinon-chai';
+
+import { FirebaseApp } from '@firebase/app-types';
```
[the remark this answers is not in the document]
samhorlbeck: Oops, done

packages-exp/auth-exp/src/model/auth.d.ts
samhorlbeck: Yeah, the files in the model/ directory are meant to contain internal stuff as well. The types in the types package will expose only the public bits (and many of the fields will be marked as readonly, for example)

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
avolkovi: ready!

samhorlbeck: Already done :)

PR #2934, opened 2020-04-17
Add conditional delays to auth-next

avolkovi:


packages-exp/auth-exp/src/core/util/delay.ts
```diff
@@ -0,0 +1,54 @@
...
+  // The default value for the offline delay timeout in ms.
+
+  private readonly isMobile: boolean;
+  constructor(
+    private readonly shortDelay: number,
+    private readonly longDelay: number
+  ) {
+    // Internal error when improperly initialized.
+    if (shortDelay > longDelay) {
+      throw new Error('Short delay should be less than long delay!');
```
Feiyang1, line 38: I suppose this is a debug assertion that can be removed in prod build?

schmidt-sebastian: Not that I am looking at this code, but since you are adding a new code base: You should follow the model we added in Firstore and have "debug asserts" and "hard asserts". The debug asserts can then be stripped from the production build, which saves a lot of bytes.

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,28 +15,72 @@
...
+}
 
-  return expression;
+/**
+ * Verifies the given condition and fails if false, throwing a developer facing error
+ *
+ * @param assertion
+ * @param appName
+ */
+export function assert(assertion: boolean, appName: string): asserts assertion {
```
samhorlbeck, line 38: No need to constrain `assertion` to boolean. This check is good for truthy values as well

packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -104,23 +104,26 @@ export class UserImpl implements User {
       photoURL
     } = object;
 
-    assert(uid && plainObjectTokenManager, auth.name);
+    assert(uid && !!plainObjectTokenManager, auth.name);
```
samhorlbeck, line 107: With the other change I mentioned, no need to force this into a boolean

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,28 +15,72 @@
...
+  if (!assertion) {
+    fail(appName);
+  }
 }
 
-export function assertType<T>(
-  expression: unknown,
-  expected: string | string[],
+export function assertStringOrUndefined(
```
samhorlbeck, line 44: This is specific to the user object implementation. Should this assert just live in that file?

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,28 +15,72 @@
...
+ */
+export function debugFail(failure: string): never {
+  // Log the failure in addition to throw an exception, just in case the
+  // exception is swallowed.
+  const message = `AUTH (${SDK_VERSION}) INTERNAL ASSERTION FAILED: ` + failure;
+  logError(message);
 
-  assert(expected.includes(typeof expression), appName);
-  return expression as T;
+  // NOTE: We don't use FirestoreError here because these are internal failures
```
samhorlbeck, line 66: FirestoreError -> AuthError ?

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,28 +15,72 @@
...
+
+/**
+ * Fails if the given assertion condition is false, throwing an Error with the
+ * given message if it did.
+ *
+ * @param assertion
+ * @param message
+ */
+export function debugAssert(
+  assertion: boolean,
```
samhorlbeck, line 80: Same here re. truthy vs true

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,28 +15,72 @@
...
+/**
+ * Unconditionally fails, throwing an internal error with the given message.
+ *
+ * @param failure type of failure encountered
+ * @throws Error
+ */
+export function debugFail(failure: string): never {
+  // Log the failure in addition to throw an exception, just in case the
+  // exception is swallowed.
+  const message = `AUTH (${SDK_VERSION}) INTERNAL ASSERTION FAILED: ` + failure;
```
samhorlbeck, line 63: Looks like the logError function already includes Auth and the sdk version. Can probably remove those bits

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,29 +15,65 @@
...
+ * @param appName
+ */
+export function assert(
+  assertion: unknown,
   appName: string,
-  code = AuthErrorCode.INTERNAL_ERROR
-): T {
-  if (!expression) {
-    throw AUTH_ERROR_FACTORY.create(code, { appName });
+  errorCode: AuthErrorCode = AuthErrorCode.INTERNAL_ERROR
```
samhorlbeck, line 40: nit: no need to type this since it has a default value: go/tsstyle#type-inference

packages-exp/app-exp/package.json
```diff
@@ -51,7 +51,7 @@
   "bugs": {
     "url": "https://github.com/firebase/firebase-js-sdk/issues"
   },
-  "typings": "./dist/app-exp.d.ts",
```
samhorlbeck, line 54: Did you mean to do this?

avolkovi: weird, fixed

packages-exp/auth-exp/src/core/util/delay.ts
avolkovi: updated, ready for re-review

PR #2935, opened 2020-04-18
Don't include GRPC in browser build

schmidt-sebastian:
This addresses an OOM in Saucelabs, since  tries to make grpc-js work in IE11. We should not include any tests that depend on GRPC-JS


packages/firestore/test/unit/remote/serializer.helper.ts
```diff
@@ -0,0 +1,1487 @@
...
+const protoJsReader = testUserDataReader(/* useProto3Json= */ false);
+
+/**
+ * Runs the serializer test with an optional ProtobufJS verification step
+ * (only provided in Node).
+ *
+ * These tests are initialized in 'serializer.browser.test.ts' and
+ * 'serializer.node.test.ts'.
+ */
+export function serializerTest(
```
schmidt-sebastian, line 97: Note that this file is the same as `serializer.test.ts`, but wrapped in this function. I also pulled out `verifyProtobufJsRoundTrip`

PR #2936, opened 2020-04-18
use queryString helper in auth-next

avolkovi:
Can't believe you guys let this one slip past code review 🦀

### DIFF
--- a/packages-exp/auth-exp/src/api/index.ts
+++ b/packages-exp/auth-exp/src/api/index.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { FirebaseError } from '@firebase/util';
+import { FirebaseError, querystring } from '@firebase/util';
 import { AuthErrorCode, AUTH_ERROR_FACTORY } from '../core/errors';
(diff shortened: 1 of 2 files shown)


packages-exp/auth-exp/src/api/index.ts
```diff
@@ -65,28 +65,25 @@ export async function performApiRequest<T, V>(
...
-    const queryString = Object.keys(params)
-      .map(key => {
-        return `${encodeURIComponent(key)}=${encodeURIComponent(params[key])}`;
-      })
-      .join('&');
-
     const response: Response = await Promise.race<Promise<Response>>([
       fetch(
-        `${auth.config.apiScheme}://${auth.config.apiHost}${path}?${queryString}`,
+        `${auth.config.apiScheme}://${auth.config.apiHost}${path}?${querystring(
```
samhorlbeck, line 81: small nit: consider generating the query string first into a const then keeping this template string as a one-liner

avolkovi: done

PR #2937, opened 2020-04-18
Make @firebase/app-next have single entry point

Feiyang1:
Previously internal functions are exported in the submodule `@firebase/app-exp/internal`. This PR changes to directly export them in `@firebase/app-exp`. It removes the `internal` folder and the package.json inside it and simplifies the build configuration. It also makes it harder for developers to use our internal methods accidentally.

We generate 2 d.ts files as part of the build process.  One has only public API and is used to publish to npm, so developers won't be able to import internal methods. The other one has both public and internal APIs and is used during development, so other Firebase SDKs can import the internal methods.

### DIFF
--- a/common/api-review/app-exp.api.md
+++ b/common/api-review/app-exp.api.md
@@ -5,39 +5,29 @@
 ```ts
 
 import { Component } from '@firebase/component';
-import { FirebaseApp } from '@firebase/app-types';
-import { FirebaseAppConfig } from '@firebase/app-types';
-import { FirebaseOptions } from '@firebase/app-types';
+import { FirebaseApp } from '@firebase/app-types-exp';
+import { FirebaseAppConfig } from '@firebase/app-types-exp';
+import { FirebaseOptions } from '@firebase/app-types-exp';
 import { LogCallback } from '@firebase/logger';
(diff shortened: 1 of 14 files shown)


common/api-review/app-exp.api.md
```diff
@@ -4,12 +4,30 @@
...
+import { FirebaseApp } from '@firebase/app-types-exp';
+import { FirebaseAppConfig } from '@firebase/app-types-exp';
+import { FirebaseOptions } from '@firebase/app-types-exp';
 import { LogCallback } from '@firebase/logger';
 import { LogLevel } from '@firebase/logger';
 import { LogOptions } from '@firebase/logger';
+import { Name } from '@firebase/component';
+import { Provider } from '@firebase/component';
+
+// @internal (undocumented)
```
hsubox76, line 17: Looked at the docs for `api-extractor`/`api-documenter` and still not sure, how does this show up in the docs? Do either the `@internal` tag or the `(undocumented)` status prevent this from showing up in the docs? It seemed like `@internal` methods still show up in docs but they just say `@internal`? I assume we want it to not be in the docs at all?

PR #2938, opened 2020-04-18
Don't crash if write cannot be persisted

schmidt-sebastian:
Addresses https://github.com/firebase/firebase-js-sdk/issues/2755


nVitius:
Does this also address the issue if doing batched writes?

schmidt-sebastian:
> Does this also address the issue if doing batched writes?

Yes.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -373,7 +377,26 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
+      // If we can't persist the mutation, we reject the user callback and don't
+      // send the mutation. The user can then retry the write. We currently do
+      // not use `enqueueRetryable()` for writes since this would require us to
+      // move all view computation logic to `enqueueRetryable()`. Otherwise,
+      // this write should be surfaced in all subsequent operations.
+      logError(LOG_TAG, 'Dropping write that cannot be persisted: ' + e);
+      userCallback.reject(
+        new FirestoreError(
+          Code.UNKNOWN,
+          'Query allocation failed with environment error: ' + e
```
wilhuff, line 394: "Query allocation" seems like the wrong description of this error. Also, since this is catching everything, this isn't necessarily something to do with the environment. "Failed to enqueue a write in the local database"?

The catch here is also overbroad: this will catch internal assertion errors and the like as well. Taking an approach like the one I suggested in #2919 would make it possible to catch just errors related to IndexedDB.

Also, once we know the error is specifically about IndexedDB, I think the code can probably become `DATA_LOSS`.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -373,7 +377,26 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
...
   async write(batch: Mutation[], userCallback: Deferred<void>): Promise<void> {
     this.assertSubscribed('write()');
-    const result = await this.localStore.localWrite(batch);
+
+    let result: LocalWriteResult;
+    try {
+      result = await this.localStore.localWrite(batch);
+    } catch (e) {
+      // If we can't persist the mutation, we reject the user callback and don't
+      // send the mutation. The user can then retry the write. We currently do
```
wilhuff, line 386: This extra bit about not using `enqueueRetryable` isn't all that helpful. We're failing user writes because that's the developer experience we want users of the Firestore SDK to have. If we thought it was a better experience not to fail writes we'd make the changes required to make that happen.

schmidt-sebastian: I removed it.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Sorry, "Query allocation" was the error message for my next PR.

I updated the PR with the changes in #2919 and only throw this error for IndexedDbTransactionErrors.  I am a little uncertain whether we should just re-throw the error here or wrap it. For now, I decided to wrap it in a FirestoreError, as this makes it easier for our users to share error handling logic between the various backend errors and the new IndexedDB error.

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -75,5 +75,61 @@ describeSpec(
           .expectEvents(query, {});
       }
     );
+
+    specTest('Recovers when write cannot be persisted', [], () => {
```
wilhuff, line 79: Do these tests need a `no-ios` and `no-android` tag or some new kind of tag to indicate that failure recovery is required?

schmidt-sebastian: All tests in this file are marked `no-ios` and `no-android` at the moment: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/unit/specs/recovery_spec.test.ts#L26

schmidt-sebastian: FYI: I pushed a commit to this PR to remove the `durable-persistence` tags as the new recovery tests also work with memory persistence. Most of my changes in my next PR were due to the formatting caused by this change.

PR #2944, opened 2020-04-20
Don't crash if Target cannot be persisted

schmidt-sebastian:
This PR rejects a Query listener if its Target cannot be allocated.

Addresses #2755


packages/firestore/src/core/sync_engine.ts
```diff
@@ -246,8 +252,7 @@ export class SyncEngine implements RemoteSyncer, SharedClientStateSyncer {
       }
     }
 
-    this.syncEngineListener!.onWatchChange([viewSnapshot]);
```
schmidt-sebastian, line 249: I had to remove this indirection. `onWatchChange` doesn't fire if EventManager doesn't execute `this.queries.set(query, queryInfo)`. Executing `this.queries.set(query, queryInfo!)` before this method runs breaks listeners if the initial registration in IndexedDB failed.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -98,6 +98,12 @@ class QueryView {
   ) {}
 }
 
+/** Result type returned during initial query registration. */
+export interface QueryRegistration {
```
schmidt-sebastian, line 102: We now have three different versions of this interface:  QueryRegistration. ViewChange, QueryListenersInfo

They are all a bit different and i don't see a way to combine them. `ViewChange` would be the best candidate, but the View does not expose the targetId.

packages/firestore/src/core/event_manager.ts
```diff
@@ -59,16 +63,37 @@ export class EventManager implements SyncEngineListener {
...
-      this.queries.set(query, queryInfo);
     }
+
+    if (firstListen) {
+      try {
+        const { targetId, snapshot } = await this.syncEngine.listen(query);
+        queryInfo.targetId = targetId;
+        queryInfo.viewSnap = snapshot;
+      } catch (e) {
+        if (e.name === 'IndexedDbTransactionError') {
```
wilhuff, line 82: This may not be possible, but it seems like could benefit from handling these errors consistently in our API.

For example, we try to make FirestoreClient the place where we centrally handle dispatching onto the AsyncQueue. Should we try to similarly make handling retries (or error handling like this) a part of the SyncEngine?

This doesn't seem to work in this case because you're returning a successful promise in case of a known error, but that seems to be catering to the caller's use of this function in a way that makes this more complicated than it needs to be. Maybe the caller should handle this in the block submitted to the async queue?

packages/firestore/src/core/sync_engine.ts
wilhuff: At first I was concerned about this because it seemed like this would prevent calling `onViewSnapshot` on the listener, but actually it looks like code to do that was already explicitly in the `EventManager` (in `if (queryInfo.viewSnap)`, and I'm guessing that the explicit code was just being treated as a no-op because previously `queryInfo.viewSnap` would be undefined before?

packages/firestore/src/core/sync_engine.ts
wilhuff: The old arrangement where the sync engine was responsible for the first snapshot avoided this problem.

I wonder if would be possible for the sync engine to call back indicating the listen failed instead? That way EventManager.listen could proceed optimistically and then just surface the error/remove the listener as normal without any changes.

packages/firestore/src/core/event_manager.ts
```diff
@@ -59,15 +67,14 @@ export class EventManager implements SyncEngineListener {
     this.syncEngine.subscribe(this);
   }
 
-  listen(listener: QueryListener): Promise<TargetId> {
+  async listen(listener: QueryListener): Promise<void> {
```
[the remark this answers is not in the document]
schmidt-sebastian: If I reject the Promise here, then FirestoreClient has to handle the rejected Promise. This is actually pretty nice and keeps the code in EventManager much cleaner. The calling code is also still clean, since it was simple to begin with.

The problem I see is that this makes this code non-testable. The spec tests don't use FirestoreClient, and the integration cannot inject failures. I have made the change regardless and added some special handling in SyncEngine. I am not sure if this is an improvement though.

packages/firestore/src/core/event_manager.ts
schmidt-sebastian: I did end up updating this, but as stated this pushes the error handling to both FirestoreClient and SpecTestRunner.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: The issue here is actually the success case - the problem here is the ViewSnapshot for the initial listen needs to be raised. The current flow is:

- EventManager's listen() registers target in local data structure
- EventManager calls SyncEngine listen()
- SyncEngine's listen() calls back EventManager
- EventManager looks in local data structure to raise snapshot
- SyncEngine finishes listen()
- EventManager finishes listen()

This code seems fragile to me The callback in SyncEngine relies on an implementation detail of EventManager. I was very surprised that re-arranging the internal code of EventManager broke tests that were unrelated to my change. Because of this, I think it is better to return a ViewSnapshot.

Having said that, I was able to remove this type  after all because EventManager doesn't actually use the target ID (other than to pass it through).

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: I am not sure I fully understand this, but I hope that my previous comment reliefs your concerns. Let me know if it doesn't.

schmidt-sebastian:
Friendly bump :)

packages/firestore/src/core/firestore_client.ts
```diff
@@ -398,8 +398,22 @@ export class FirestoreClient {
...
+    this.asyncQueue.enqueueAndForget(async () => {
+      try {
+        await this.eventMgr.listen(listener);
+      } catch (e) {
+        logError(LOG_TAG, `Failed to register query: ${e}`);
+        if (e.name === 'IndexedDbTransactionError') {
+          listener.onError(
+            new FirestoreError(
+              Code.UNAVAILABLE,
+              `Failed to register query: ${e}`
```
wilhuff, line 410: A few thoughts on this:

  * From a user's point of view they listen to queries (or get them) but we don't have a public concept of "registration". I think it's probably fine to just say "failed to listen to query", "failed to start a listener to query" or maybe something more generic like "failed to start query".
  * We should include details of the query that failed (a toString or similar). Whenever people are filing issues about broken queries our first instructions are to have people go figure out which query is causing the problem and they're usually at a loss. We should just include that in the error message.

packages/firestore/src/core/event_manager.ts
```diff
@@ -59,16 +58,21 @@ export class EventManager implements SyncEngineListener {
...
       firstListen = true;
       queryInfo = new QueryListenersInfo();
-      this.queries.set(query, queryInfo);
     }
+
+    if (firstListen) {
+      queryInfo.viewSnap = await this.syncEngine.listen(query);
+    }
+
+    this.queries.set(query, queryInfo!);
```
wilhuff, line 75: I'm surprised the `!` is required here. Isn't `queryInfo` guaranteed to be set here? (What confuses me is you're not null-checking the assignment to e.g. `queryInfo.viewSnap`).

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -162,4 +163,37 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
...
+    const query1 = Query.atPath(path('collection1'));
+    const query2 = Query.atPath(path('collection2'));
+    const query3 = Query.atPath(path('collection3'));
+    return spec()
+      .userListens(query1)
+      .watchAcksFull(query1, 1)
+      .expectEvents(query1, {})
+      .failDatabase()
+      .userListens(query2)
+      .expectEvents(query2, { errorCode: Code.UNKNOWN })
```
wilhuff, line 177: What's the source of the `UNKNOWN` here? The spec test runner seems to be doing `new FirestoreError(Code.UNAVAILABLE, e.message)`. Is anything even handling `errorCode`?

packages/firestore/src/core/event_manager.ts
```diff
@@ -59,16 +58,21 @@ export class EventManager implements SyncEngineListener {
...
 
     let queryInfo = this.queries.get(query);
     if (!queryInfo) {
       firstListen = true;
       queryInfo = new QueryListenersInfo();
-      this.queries.set(query, queryInfo);
     }
+
+    if (firstListen) {
+      queryInfo.viewSnap = await this.syncEngine.listen(query);
```
wilhuff, line 72: Nothing seems to populate `queryInfo.targetId` anymore?

packages/firestore/src/core/event_manager.ts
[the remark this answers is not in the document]
wilhuff: I see what you mean; the spec test has to emulate this for itself. That's not completely awful but you're probably right that the other way was better because it was testable.

Would it help to put the error handling in `SyncEngine`?

packages/firestore/src/core/firestore_client.ts
schmidt-sebastian: Updated to use "initialize" and also included the Query in the message.

packages/firestore/src/core/event_manager.ts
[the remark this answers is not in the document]
schmidt-sebastian: I think SyncEngine is not the right place for this since it can only convey two states: Success or failure. We really need three states: Success, failure (but don't crash AsyncQueue) and failure (crash AsyncQueue). 
Unless we change `SyncEngine.listen()` return a tristate, we have to wrap any error thrown by SyncEngine in the same `if (error.name === '...')` check when we invoke the QueryListener.

Given this and the test ability, I moved the code back to EventListener. It's not as clean looking, but it will probably be easier to keep regression free.

PR #2960, opened 2020-04-21
Add sendSignInWithEmail to auth-next

avolkovi:


packages-exp/auth-exp/src/model/action_code_url.ts
```diff
@@ -0,0 +1,103 @@
...
+  CONTINUE_URL = 'continueUrl',
+  LANGUAGE_CODE = 'languageCode',
+  MODE = 'mode',
+  TENANT_ID = 'tenantId'
+}
+
+/**
+ * Map from mode string in action code URL to Action Code Info operation.
+ */
+const _ModeToOperationMap: { [key: string]: Operation } = {
```
samhorlbeck, line 38: So I _think_ we shouldn't be putting the `_` prefix on names that _aren't_ exported. These are all file-private. The `_` prefix is meant to denote exported (aka code public) names that shouldn't be used by other packages. See https://api-extractor.com/pages/tsdoc/tag_internal/ (Fei linked this to me in my PR)

Also file level consts should be UPPER_SNAKE_CASE

packages-exp/auth-exp/src/model/action_code_url.ts
```diff
@@ -0,0 +1,103 @@
...
+        appName: auth.name
+      });
+    }
+    this.apiKey = apiKey;
+    this.operation = operation;
+    this.code = code;
+    this.continueUrl = uri.searchParams.get(QueryField.CONTINUE_URL);
+    this.languageCode = uri.searchParams.get(QueryField.LANGUAGE_CODE);
+    this.tenantId = uri.searchParams.get(QueryField.TENANT_ID);
+    return this;
```
samhorlbeck, line 92: Don't need to do this

packages-exp/auth-exp/src/model/action_code_url.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  static _fromLink(auth: Auth, link: string): ActionCodeURL | null {
+    const actionLink = _parseDeepLink(link);
+    try {
+      return new ActionCodeURL(auth, actionLink);
+    } catch (e) {
```
samhorlbeck, line 99: Can just `} catch { ... }` since `e` isn't being used

packages-exp/auth-exp/src/model/action_code_url.ts
```diff
@@ -0,0 +1,102 @@
...
+  const doubleDeepLink = link ? new URL(link).searchParams.get('link') : null;
+  // iOS custom scheme links.
+  const iOSDeepLink = uri.searchParams.get('deep_link_id');
+  const iOSDoubleDeepLink = iOSDeepLink
+    ? new URL(iOSDeepLink).searchParams.get('link')
+    : null;
+  return iOSDoubleDeepLink || iOSDeepLink || doubleDeepLink || link || url;
+}
+
+export class ActionCodeURL {
```
avolkovi, line 67: @samhorlbeck should I split this into an interface & implementation like we have with user & auth?

packages-exp/auth-exp/src/model/action_code_url.ts
```diff
@@ -0,0 +1,102 @@
...
+  CONTINUE_URL = 'continueUrl',
+  LANGUAGE_CODE = 'languageCode',
+  MODE = 'mode',
+  TENANT_ID = 'tenantId'
+}
+
+/**
+ * Map from mode string in action code URL to Action Code Info operation.
+ */
+const ModeToOperationMap: { [key: string]: Operation } = {
```
samhorlbeck, line 38: `MODE_TO_OPERATION_MAP`

packages-exp/auth-exp/src/model/action_code_url.ts
samhorlbeck: Whoops yeah this whole thing should be moved to core. I don't think you'll need an interface for it (we need that for user and auth to avoid circular dependency issues... I don't anticipate that being a problem for this module)

PR #2961, opened 2020-04-21
Add auth listener implementation, add user.reload()

samhorlbeck:


packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -117,6 +118,102 @@ describe('AuthImpl', () => {
...
+  describe('change listeners', () => {
+    // // Helpers to convert auth state change results to promise
+    // function onAuthStateChange(callback: NextFn<User|null>)
+
+    it('immediately calls authStateChange if initialization finished', done => {
+      const user = testUser('uid');
+      auth.currentUser = user;
+      auth._isInitialized = true;
+      auth.onAuthStateChanged(user => {
+        expect(user).to.eq(user);
```
avolkovi, line 131: can you use  expect(callback).to.have.been.calledWith() instead?

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -117,6 +118,102 @@ describe('AuthImpl', () => {
...
+        auth.onIdTokenChange(callback);
+        await auth.updateCurrentUser(user);
+        expect(callback).to.have.been.calledTwice;
+      });
+
+      it('onIdTokenChange triggers on log out', async () => {
+        await auth.updateCurrentUser(user);
+        auth.onIdTokenChange(callback);
+        await auth.updateCurrentUser(null);
+        expect(callback).to.have.been.calledTwice;
```
avolkovi, line 190: can you move the first callback into setup so these are either have.been.calledOnce or not.have.been.called?

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -120,3 +203,18 @@ export function initializeAuth(
 
   return new AuthImpl(app.name, config, hierarchy);
 }
+
+/** Helper class to wrap subscriber logic */
+class Subscription<T> {
```
avolkovi, line 208: should this belong in util?

packages-exp/auth-exp/src/core/user/reload.test.ts
```diff
@@ -0,0 +1,158 @@
...
+      createdAt: 123,
+      lastLoginAt: 456
+    };
+
+    mockEndpoint(Endpoint.GET_ACCOUNT_INFO, {
+      users: [serverUser]
+    });
+
+    const user = testUser('abc', '', true);
+    await _reloadWithoutSaving(user);
```
avolkovi, line 78: can we test reload instead of _reloadWithoutSaving? Feels like an anti-pattern to test private helpers instead of the actual interface

packages-exp/auth-exp/src/core/user/reload.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    expect(user.tenantId).to.eq('tenantId');
+    expect(user.metadata).to.eql({
+      creationTime: '123',
+      lastSignInTime: '456'
+    });
+  });
+
+  it('adds missing provider data', async () => {
+    const user = testUser('abc', '', true);
+    user.providerData = [{ ...BASIC_USER_INFO }];
```
avolkovi, line 94: isn't this just [BASIC_USER_INFO]?

packages-exp/auth-exp/src/core/user/reload.ts
```diff
@@ -0,0 +1,83 @@
...
+  return [...deduped, ...newData];
+}
+
+function extractProviderData(providers: ProviderUserInfo[]): UserInfo[] {
+  return providers.map(provider => ({
+    uid: provider.rawId || '',
+    displayName: provider.displayName || null,
+    email: provider.email || null,
+    phoneNumber: provider.phoneNumber || null,
+    providerId: provider.providerId as ProviderId,
```
avolkovi, line 80: can this cast fail? should we have an assertion?

packages-exp/auth-exp/src/model/auth.d.ts
```diff
@@ -34,10 +43,22 @@ export interface Auth {
   currentUser: User | null;
   readonly name: AppName;
   readonly config: Config;
+  _isInitialized: boolean;
```
avolkovi, line 46: can this be awaited? I think we had this as a method in the prototype

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
samhorlbeck: Done. Yeah the problem is that the the callback is always immediately triggered. I've just switched to resetting the spy

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: I don't think so-- this is the only place it'll be used

packages-exp/auth-exp/src/core/user/reload.test.ts
samhorlbeck: `_reloadWithoutSaving()` gets used a ton throughout the rest of the SDK (more often than `reload()`). It's "private" in the sense that _other_ packages shouldn't be using it, but we'll be using it a lot

packages-exp/auth-exp/src/core/user/reload.test.ts
samhorlbeck: No, given that things in the user object do get mutated, we want to make sure we're using a copy. The objects in providerData _shouldn't_ be mutated, but then again that's why we have tests :)

packages-exp/auth-exp/src/core/user/reload.ts
samhorlbeck: Done.

packages-exp/auth-exp/src/model/auth.d.ts
samhorlbeck: What would be the use case for that? Anything that needs to await initialization should be putting tasks into `queue()`

avolkovi: we used it in signInWithCredential https://github.com/firebase/firebase-js-sdk/blob/bc990a8a2800669b53e0c6f158947071930bbef6/packages/auth-ts/src/core/strategies/index.ts#L60

samhorlbeck: Ah yeah so that isn't necessary -- that method then calls updateCurrentUser which tosses the update into the queue
https://github.com/firebase/firebase-js-sdk/blob/bc990a8a2800669b53e0c6f158947071930bbef6/packages/auth-ts/src/core/strategies/index.ts#L63

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -117,6 +118,138 @@ describe('AuthImpl', () => {
...
+      let user: User;
+      let callback: sinon.SinonSpy;
+
+      beforeEach(() => {
+        user = testUser('uid');
+        callback = sinon.spy();
+      });
+
+      it('onAuthStateChange triggers on log in', async () => {
+        auth.onAuthStateChanged(callback);
```
avolkovi, line 166: these first three lines of setup are re-used a ton, can you extract to a context() block with a beforeEeach()?

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
avolkovi: oh, maybe I misunderstood, it looks like you used calledWith() down below, but not in these test cases

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/model/auth.d.ts
```diff
@@ -15,12 +15,21 @@
...
+
 import { Persistence } from '../core/persistence';
 import { User } from './user';
 
 export type AppName = string;
 export type ApiKey = string;
 export type AuthDomain = string;
+export type NextOrObserver<T> = NextFn<T | null> | Observer<T | null>;
 
 export interface Config {
```
scottcrossen, line 34: What is the `packages-exp/auth-types-exp` package for? I thought it was for the type definitions that users interact with (like this).

avolkovi: <product>-types is for public types intended for clients, the types we have defined internally may be more permissive than what we expose to customers since there's no 'protected' keyword in TS.  Some fields that may exist/be writeable in auth-exp, may either not exist or be readonly in auth-type-exp

Issue #2965, opened 2020-04-22
Increase size of Auth 'signInWithPopup' window for Apple


jonpreececm:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS Catalina Latest
  * Browser version: Google Chrome v81
  * Firebase SDK version: 7.9.1
  * Firebase Product: auth

### [REQUIRED] Describe the problem]

When signing in using a popup and Apple oauth, the popup is too narrow.  The width is 600px, it should be 640px.

#### Steps to reproduce:
Trigger a sign in with popup flow using Apple oauth

#### Relevant Code:

```
  signUserInWithOAuth() {
    const provider = new Firebase.auth.OAuthProvider("apple.com");
    provider.addScope("email");
    provider.addScope("name");

    return Firebase.auth()
      .signInWithPopup(provider)
  },
```

This is the exact line of code that needs updating; https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/src/idp.js#L138

PR #2966, opened 2020-04-22
Fix Apple popup window window

jonpreececm:
This fixes the issue described here; https://github.com/firebase/firebase-js-sdk/issues/2965


jonpreececm:
@googlebot I signed it!

bojeil-google:
Thanks for the fix.

PR #2970, opened 2020-04-22
Untangle ObjectValue and ObjectValueBuilder

schmidt-sebastian:
If ObjectValue doesn't depend on ObjectValueBuilder and doesn't directly offer the ability to extract field masks, then we can use ObjectValue without pulling in SortedMap and SortedSet.


packages/firestore/src/model/field_value.ts
```diff
@@ -278,3 +238,32 @@ export class ObjectValueBuilder {
     return modified ? { mapValue: { fields: resultAtPath } } : null;
   }
 }
+
+/**
+ * Returns a FieldMask built from all fields in an MapValue.
+ */
+export function extractFieldMask(value: api.MapValue): FieldMask {
```
dconeybe, line 245: Consider changing the argument of `extractFieldMask()` from `api.MapValue` to `ObjectValue`. With this change, the call sites become more readable. For example,

```ts
extractFieldMask(objValue.proto.mapValue!);
```

becomes

```ts
extractFieldMask(objValue);
```

schmidt-sebastian: This was how the original call worked, but it used one level of indirection. The method is called recursively and only the first level uses an ObjectValue. I debated passing in just an api.Value, which would act as a sort of compromise. It however hides the fact that the method only works on a MapValue.

dconeybe: Ahh I see. Feel free to disregard this comment then. The original way you implemented it makes sense.

PR #2971, opened 2020-04-22
Untangle Datastore

schmidt-sebastian:
This removes the Datastore dependencies on SortedMap, AsyncQueue and the persistent streams, all of which are not needed in the "Rest Wrapper" client.


packages/firestore/src/remote/datastore.ts
```diff
@@ -51,36 +43,11 @@ interface CommitRequest extends api.CommitRequest {
  */
 export class Datastore {
   constructor(
-    private queue: AsyncQueue,
     private connection: Connection,
     private credentials: CredentialsProvider,
     private serializer: JsonProtoSerializer
   ) {}
 
-  newPersistentWriteStream(
```
wilhuff, line 60: Rather than changing the object models and completely rearranging the client (so that it doesn't match the other ports), a better strategy would be to adopt tree-shakeable APIs internally.

That is, make classes like `Datastore` more like just dumb structs, and make the methods that used to be in the class plain functions that take `Datastore` as an argument.

This will greatly reduce the amount of churn in the relative arrangement of components. Indeed, you could apply this to `commit` and other methods on datastore to ensure that all of it tree-shakes out if unused.

packages/firestore/test/integration/remote/stream.test.ts
```diff
@@ -173,68 +161,40 @@ class MockCredentialsProvider extends EmptyCredentialsProvider {
...
    */
   it('can be stopped before handshake', () => {
-    let writeStream: PersistentWriteStream;
-
-    return withTestDatastore(ds => {
-      writeStream = ds.newPersistentWriteStream(streamListener);
-      writeStream.start();
-      return streamListener.awaitCallback('open');
-    }).then(async () => {
+    return withTestWriteStream(async (writeStream, streamListener) => {
```
wilhuff, line 169: Despite the other comment on thinking the newPersistentStream methods should stay in `Datastore`, this change is a good one (though `withTestWriteStream` should call `newPersistentWriteStream(datastore)`.

packages/firestore/src/remote/datastore.ts
schmidt-sebastian: One of the downsides is that it makes every member public, and I kind of wanted to avoid this. Your argument that we should pull out all methods does win here though. I updated the PR, which also greatly reduces the diff (a large part are now just test cleanup).

packages/firestore/src/remote/datastore.ts
```diff
@@ -158,3 +91,79 @@ export class Datastore {
...
+
+export function invokeCommitRpc(
+  datastore: Datastore,
+  mutations: Mutation[]
+): Promise<MutationResult[]> {
+  const params: CommitRequest = {
+    database: datastore.serializer.encodedDatabaseId,
+    writes: mutations.map(m => datastore.serializer.toMutation(m))
+  };
+  return datastore
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/firestore/src/remote/datastore.ts
```diff
@@ -158,3 +91,79 @@ export class Datastore {
...
+    .invokeRPC<CommitRequest, api.CommitResponse>('Commit', params)
```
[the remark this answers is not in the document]
schmidt-sebastian: I'd rather keep this as is, but we can revisit if this pattern becomes a problem.

packages/firestore/src/remote/datastore.ts
```diff
@@ -158,3 +91,79 @@ export class Datastore {
...
+): Promise<MaybeDocument[]> {
+  const params: BatchGetDocumentsRequest = {
+    database: datastore.serializer.encodedDatabaseId,
+    documents: keys.map(k => datastore.serializer.toName(k))
+  };
+
+  return datastore
+    .invokeStreamingRPC<
+      BatchGetDocumentsRequest,
+      api.BatchGetDocumentsResponse
```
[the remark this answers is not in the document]
schmidt-sebastian: `BatchGetDocumentsRequest` is `api. BatchGetDocumentsRequest` with the `database` field set. I just added it in the Proto d.ts field since I/we have been editing this file manually lately.

packages/firestore/test/integration/remote/remote.test.ts
```diff
@@ -48,12 +47,9 @@ describe('Remote Storage', () => {
       const k = key('docs/1');
       const mutation = setMutation('docs/1', { sort: 10 });
 
-      return ds
-        .commit([mutation])
-        .then((result: MutationResult[]) => {
-          return ds.lookup([k]);
-        })
-        .then((docs: MaybeDocument[]) => {
+      return invokeCommitRpc(ds, [mutation])
```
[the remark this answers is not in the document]
schmidt-sebastian: Converted this file

packages/firestore/test/integration/remote/stream.test.ts
schmidt-sebastian: Done

packages/firestore/src/remote/datastore.ts
schmidt-sebastian: I pushed a commit with a TypeScript version of this hack: https://github.com/firebase/firebase-js-sdk/pull/2971/commits/0295cd5df3bd7ee7f30f8ecb18b174e1f35e8308

It should be pretty easy to understand, but I wanted to point out a downside: With this change, at least IntelliJ now consider all members of `DatastoreImpl` unused, which would breaks its automated refactor should we use it on classes like this.

packages/firestore/src/remote/datastore.ts
```diff
@@ -16,113 +16,44 @@
...
-}
-interface CommitRequest extends api.CommitRequest {
-  database?: string;
-}
+/**
+ * Datastore and its related methods are a wrapper around the external Google
+ * Cloud Datastore grpc API, which provides an interface that is more convenient
+ * for the rest of the client SDK architecture to consume.
+ */
+export class Datastore {}
```
wilhuff, line 40: If you made this an `interface` this wouldn't even materialize in the generated code.

The downside is that empty interfaces match any object so we'd effectively lose type safety. Maybe that's why you chose class?

packages/firestore/src/remote/datastore.ts
```diff
@@ -158,3 +89,100 @@ export class Datastore {
...
+  serializer: JsonProtoSerializer
+): Datastore {
+  return new DatastoreImpl(connection, credentials, serializer);
+}
+
+export async function invokeCommitRpc(
+  datastore: Datastore,
+  mutations: Mutation[]
+): Promise<MutationResult[]> {
+  debugAssert(
```
wilhuff, line 105: Is IntelliJ happier if you made this something like this here:

```ts
const datastoreImpl = cast(datastore);
```

and also added a function like this:

```ts
function cast(datastore: Datastore): DatastoreImpl {
  debugAssert(...);
  return datastore as DatastoreImpl;
}
```

packages/firestore/src/remote/datastore.ts
wilhuff: That's a lot better than my suggestion. If we can get IntelliJ happy, I think this is the way to go.

packages/firestore/src/remote/datastore.ts
schmidt-sebastian: Yes, an empty interface matches all objects in TypeScript. It is even commonly used to do exactly that.

packages/firestore/src/remote/datastore.ts
schmidt-sebastian: I just used an explicit cast to a new type, which makes IntelliJ happy.

wilhuff: A `cast` function also has the benefit of reducing the boilerplate at the top of every function from 5 lines to 1. The specific text of which function needed DatastoreImpl isn't important to preserve--should we run afoul of this assertion the stack will show with function was invoked improperly.

If you want to avoid the extra function in the emitted code (if that's a problem) we could add a rewrite rule to inline the function call (to just a cast). If `cast` is too ambiguous for the strategy, maybe name it `castDatastoreImpl`,  `recoverDatastoreImpl`, 'toDatastoreImpl`, or `asDatastoreImpl`? This way you could write dumb, unscoped global rules for this (again--if that's even worthwhile).

schmidt-sebastian: Not 100% convinced that this effort was worth it (since all these extra lines are stripped in production), but I added a `debugCast` function that not only taught me a lot about TypeScript and constructor signatures but also can be used everywhere else in the code.

wilhuff: Agreed that this makes little to no difference in the production code (after stripping). Where it makes a difference is in the readability of this required transformation. Having the same 5-line incantation at the top of nearly every function would have made my eyes bleed :-).

PR #2975, opened 2020-04-23
Add refresh token endpoint + implementation to token manager

samhorlbeck:

### DIFF
--- a/packages-exp/auth-exp/src/api/authentication/token.ts
+++ b/packages-exp/auth-exp/src/api/authentication/token.ts
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { querystring } from '@firebase/util';
+
+import { performFetchWithErrorHandling } from '../';
+import { Auth } from '../../model/auth';
+
+const ENDPOINT = 'https://securetoken.googleapis.com/v1/token';
+const GRANT_TYPE = 'refresh_token';
+
+export interface RequestStsTokenResponse {
+  access_token?: string;
+  expires_in?: string;
+  token_type?: string;
+  refresh_token?: string;
+  id_token?: string;
+  user_id?: string;
+  project_id?: string;
+}
+
+export async function requestStsToken(auth: Auth, refreshToken: string): Promise<RequestStsTokenResponse> {
+  return performFetchWithErrorHandling<RequestStsTokenResponse>(auth, {}, () => {
+    const body = querystring({
+      'grant_type': GRANT_TYPE,
+      'refresh_token': refreshToken,
+    }).slice(1);
+
+    return fetch(`${ENDPOINT}?key=${auth.config.apiKey}`, {
+      method: 'POST',
+      headers: {
+        'X-Client-Version': auth.config.sdkClientVersion,
+      },
+      body,
+    });
+  });
+}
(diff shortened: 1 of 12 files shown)


packages-exp/auth-exp/src/api/authentication/token.ts
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { querystring } from '@firebase/util';
+
+import { performFetchWithErrorHandling } from '../';
+import { Auth } from '../../model/auth';
+
+export const _ENDPOINT = 'https://securetoken.googleapis.com/v1/token';
```
avolkovi, line 23: can we pass this in on the auth object like we do with API_DOMAIN so that emulator can spoof it?

packages-exp/auth-exp/src/api/authentication/token.ts
```diff
@@ -0,0 +1,66 @@
...
+  expiresIn?: string;
+  refreshToken?: string;
+}
+
+export async function requestStsToken(
+  auth: Auth,
+  refreshToken: string
+): Promise<RequestStsTokenResponse> {
+  const response = await performFetchWithErrorHandling<{
+    [key: string]: string;
```
avolkovi, line 43: should this be a named type instead of just a string map?

packages-exp/auth-exp/src/api/authentication/token.ts
```diff
@@ -0,0 +1,66 @@
...
+  }>(auth, {}, () => {
+    const body = querystring({
+      'grant_type': GRANT_TYPE,
+      'refresh_token': refreshToken
+    }).slice(1);
+
+    return fetch(`${_ENDPOINT}?key=${auth.config.apiKey}`, {
+      method: 'POST',
```
avolkovi, line 51: HttpMethod.POST?

packages-exp/auth-exp/src/api/authentication/token.ts
```diff
@@ -0,0 +1,66 @@
...
+        'X-Client-Version': auth.config.sdkClientVersion,
+        'Content-Type': 'application/x-www-form-urlencoded'
+      },
+      body
+    });
+  });
+
+  // The response comes back in snake_case. Convert to camel:
+  return {
+    accessToken: response[ServerField.ACCESS_TOKEN],
```
avolkovi, line 62: if you type the response you won't need this map

packages-exp/auth-exp/src/api/authentication/token.ts
samhorlbeck: Can't-- ts lint stomps its feet and throws a tantrum if you have snake_case members/variables

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -71,6 +89,26 @@ export class StsTokenManager {
...
+      auth,
+      oldToken
+    );
+    this.updateTokensAndExpiration(accessToken, refreshToken, expiresIn);
+  }
+
+  private updateTokensAndExpiration(
+    accessToken: string | undefined,
+    refreshToken: string | undefined,
+    expiresInSec: string | undefined
```
avolkovi, line 103: can we push this | undefined up a level?

packages-exp/auth-exp/src/api/authentication/token.ts
avolkovi: you can ignore es-lint for that interface `eslint-disable-line`

packages-exp/auth-exp/src/api/authentication/token.ts
samhorlbeck: Done

packages-exp/auth-exp/src/api/authentication/token.ts
samhorlbeck: Ya it was gonna be like ten different places for that though, but no matter I figured out how to disable it for the whole file

packages-exp/auth-exp/src/api/authentication/token.ts
samhorlbeck: Done

packages-exp/auth-exp/src/api/authentication/token.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: Done, still needs to be `string | null` though

PR #2979, opened 2020-04-24
Add AdditionalUserInfo class to auth-exp

scottcrossen:
Sent with my hg client


packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+import { expect } from 'chai';
+import { IdTokenResponse } from '../../model/id_token';
+import { ProviderId } from '../providers';
+import { fromIdTokenResponse } from './additional_user_info';
+
+describe('core/user/additional_user_info', () => {
+
+  describe('fromIdTokenResponse', () => {
+    const profile = {login: 'scott', friends: [], netWorth: 5.00};
```
avolkovi, line 26: do we have a type for this? or is it just a map?

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    const rawUserInfo = JSON.stringify(profile);
+
+    describe('parses federated IDP response tokens', () => {
+
+      it('for FacebookAdditionalUserInfo', () => {
+        const additionalUserInfo = fromIdTokenResponse(idTokenResponse({providerId: ProviderId.FACEBOOK, rawUserInfo}))!;
```
avolkovi, line 32: nit: I would extract this into a variable to avoid nesting function calls

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        const {isNewUser, providerId, username, profile} = additionalUserInfo;
+        expect(isNewUser).to.be.false;
+        expect(providerId).to.eq(ProviderId.FACEBOOK);
+        expect(username).to.be.null;
+        expect(profile).to.eq(profile);
```
avolkovi, line 37: should this one be to.eql since you're doing value comparison?

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        const additionalUserInfo = fromIdTokenResponse(idTokenResponse({providerId: ProviderId.GOOGLE, rawUserInfo}))!;
+        const {isNewUser, providerId, username, profile} = additionalUserInfo;
+        expect(isNewUser).to.be.false;
+        expect(providerId).to.eq(ProviderId.GOOGLE);
+        expect(username).to.be.null;
+        expect(profile).to.eq(profile);
+      });
+
+      it('for TwitterAdditionalUserInfo', () => {
+        const additionalUserInfo = fromIdTokenResponse(idTokenResponse({providerId: ProviderId.TWITTER, rawUserInfo, screenName: 'scott'}))!;
```
avolkovi, line 59: it looks like this one contains both login & screenName, it's ambiguous in your tests which one is used to populate the username

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        expect(providerId).to.eq(ProviderId.TWITTER);
+        expect(username).to.eq('scott');
+        expect(profile).to.eq(profile);
+      });
+    });
+
+    describe('parses profile data', () => {
+
+      it('for valid JSON', () => {
+        expect(fromIdTokenResponse(idTokenResponse({providerId: ProviderId.FACEBOOK, rawUserInfo}))!.profile).to.deep.eq(profile);
```
avolkovi, line 71: nit: use variables to avoid deep function nesting, this one in particular is a bit hard to read

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      });
+
+      it('for missing JSON', () => {
+        expect(fromIdTokenResponse(idTokenResponse({providerId: ProviderId.FACEBOOK}))!.profile).to.deep.eq({});
```
avolkovi, line 75: prefer .to.be.empty

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+      it('for anonymous auth', () => {
+        const additionalUserInfo = fromIdTokenResponse(idTokenResponse({providerId: ProviderId.ANONYMOUS, rawUserInfo}))!;
+        const {isNewUser, providerId, username, profile} = additionalUserInfo;
+        expect(isNewUser).to.be.false;
+        expect(providerId).to.be.null;
+        expect(username).to.be.null;
+        expect(profile).to.eq(profile);
+      });
+      /*
```
avolkovi, line 113: why is this one commented out?

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    kind: 'Not a new user response ',
+    ...partial
+  };
+}
+
+interface PartialIdTokenResponse {
+  providerId?: ProviderId;
+  isNewUser?: boolean;
+  rawUserInfo?: string;
+  screenName?: string | null;
```
avolkovi, line 148: this becomes string | null | undefined, prefer just undefined  or null as the fallback

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,107 @@
...
+  const { providerId } = idTokenResponse;
+  const profile =
+    typeof idTokenResponse.rawUserInfo === 'string'
+      ? JSON.parse(idTokenResponse.rawUserInfo)
+      : {};
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === 'identitytoolkit#SignupNewUserResponse';
+  /*
+  if (!providerId && !!idTokenResponse) {
```
avolkovi, line 33: why is this one commented out?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,107 @@
...
+  idTokenResponse: IdTokenResponse
+): AdditionalUserInfo | null {
+  const { providerId } = idTokenResponse;
+  const profile =
+    typeof idTokenResponse.rawUserInfo === 'string'
+      ? JSON.parse(idTokenResponse.rawUserInfo)
+      : {};
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === 'identitytoolkit#SignupNewUserResponse';
```
avolkovi, line 31: if we rely on the kind, we should extract them to an enum/constant

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(
+    readonly isNewUser: boolean,
+    readonly providerId: ProviderId | null,
+    readonly username: string | null,
+    readonly profile: { [key: string]: unknown } | null
+  ) {}
+}
+
+class FacebookAdditionalUserInfo extends GenericAdditionalUserInfo {
+  constructor(isNewUser: boolean, profile: { [key: string]: unknown } | null) {
```
avolkovi, line 77: since you use this type for profile a bunch you should extract and name it

packages-exp/auth-exp/src/model/user.d.ts
```diff
@@ -20,6 +20,13 @@ import { ProviderId } from '../core/providers';
 import { Auth } from './auth';
 import { IdTokenResult } from './id_token';
 
+export interface AdditionalUserInfo {
+  readonly isNewUser: boolean;
+  readonly profile: { [key: string]: unknown } | null;
```
avolkovi, line 25: this should be a named type, like UserProfile

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
samhorlbeck: Also if you do need deep equality, use `.to.eql()` and for strict equality (aka `===`) use `.to.eq()`

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    describe('returns null', () => {
+      it('for missing provider IDs', () => {
+        expect(fromIdTokenResponse(idTokenResponse({}))).to.be.null;
+      });
+    });
+  });
+});
+
+function idTokenResponse(partial: PartialIdTokenResponse): IdTokenResponse {
```
samhorlbeck, line 133: Just use `Partial<IdTokenResponse>`

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    idToken: 'Parsing logic not implemented',
+    refreshToken: 'Doesn\'t matter',
+    expiresIn: 'Doesn\'t matter',
+    localId: 'Doesn\'t matter',
+    kind: 'Not a new user response ',
+    ...partial
+  };
+}
+
+interface PartialIdTokenResponse {
```
samhorlbeck, line 144: And remove this entirely

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
scottcrossen: It's any random map specific to the IDP

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
scottcrossen: haha oops... It also makes no sense to compare it with itself!

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
scottcrossen: That's because it's ambiguous in the code as well. Twitter uses screenName and github uses login

scottcrossen: I'll make the tests more specific

avolkovi: we should be explicit in the tests & code which one takes precedence, and make sure we match to existing behavior

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
scottcrossen: I've added comment inline

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: Do we know all the possible values? What are they?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: I've added a comment inline

scottcrossen: Now I've just removed the comment because sam merged a PR this was dependent on.

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    const userProfileNoLogin: UserProfile = { sample: 'data' };
+    const rawUserInfoNoLogin = JSON.stringify(userProfileNoLogin);
+
+    describe('parses federated IDP response tokens', () => {
+      it('for FacebookAdditionalUserInfo', () => {
+        const idResponse = idTokenResponse({
+          providerId: ProviderId.FACEBOOK,
+          rawUserInfo: rawUserInfoWithLogin
+        });
+        const additionalUserInfo = fromIdTokenResponse(idResponse)!;
```
avolkovi, line 41: this const is only used once, recommend you collapse these two lines into const { ... } = fromIdTokenResponse()

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,108 @@
...
+import { IdTokenResponse, IdTokenResponseKind } from '../../model/id_token';
+import { ProviderId } from '../providers';
+import { parseToken } from './id_token_result';
+
+export function fromIdTokenResponse(
+  idTokenResponse: IdTokenResponse,
+): AdditionalUserInfo | null {
+  const { providerId } = idTokenResponse;
+  const profile =
+    typeof idTokenResponse.rawUserInfo === 'string'
```
avolkovi, line 27: is rawUserInfo ever not a string? should we have an assertion here?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,108 @@
...
+      ? JSON.parse(idTokenResponse.rawUserInfo)
+      : {};
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === IdTokenResponseKind.SignupNewUser;
+  if (!providerId && !!idTokenResponse) {
+    const providerId = parseToken(idTokenResponse.idToken)?.firebase?.['sign_in_provider'] as ProviderId;
```
avolkovi, line 34: add an assertion that the provider is valid before casting

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (providerId && Object.values(ProviderId).includes(providerId)) {
```
avolkovi, line 35: this check should be an assertion, see above

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,108 @@
...
+      return new GenericAdditionalUserInfo(isNewUser, providerId, null, profile);
+    }
+  }
+  if (!providerId) {
```
avolkovi, line 39: is this reachable?

packages-exp/auth-exp/src/core/user/id_token_result.ts
```diff
@@ -79,7 +79,7 @@ function utcTimestampToDateString(timestamp: string | number): string | null {
   return null;
 }
 
-function parseToken(token: string): ParsedToken | null {
+export function parseToken(token: string): ParsedToken | null {
```
avolkovi, line 82: should this be _parseToken now?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,108 @@
...
+      return new GenericAdditionalUserInfo(
+        isNewUser,
+        providerId,
+        null,
+        profile
+      );
+  }
+}
+
+class GenericAdditionalUserInfo implements AdditionalUserInfo {
```
avolkovi, line 68: do these need to be classes or can they stay as interfaces?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: Thats fine yeah. We actually just need to do a null check.

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: Yes. providerId can still be null by this point

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: Classes enable the inheritance abstraction via the `super` keyword. I'm unaware of how to do this without that syntax with interfaces alone. This provides a clean method of instantiation.

scottcrossen:
The next step is to wire this into the actual SDK. In the legacy SDK this was done [here](https://github.com/firebase/firebase-js-sdk/blob/c4b7595935f1c0faa7e83d9ab9b831bcbc434631/packages/auth/src/auth.js#L1304) and [here](https://github.com/firebase/firebase-js-sdk/blob/c4b7595935f1c0faa7e83d9ab9b831bcbc434631/packages/auth/src/authuser.js#L1333). The API of this SDK will be of the form:

```
getAdditionalUserInfo(user: User)
```

Which is possible via `WeakMap` from User -> AdditionalUserInfo. However, I'll need to know where the portions of code equivalent to the above are so that I can update the weak map as needed. Do any of y'all know where those are?

packages-exp/auth-types-exp/index.d.ts
```diff
@@ -411,3 +411,18 @@ export interface UserMetadata {
...
+  readonly isNewUser: boolean;
+  readonly profile?: UserProfile;
+  readonly providerId: ProviderId | null;
+  readonly username: string | null;
+}
+
+/**
+ * User profile used in `AdditionalUserInfo`
+ */
+export type UserProfile = { [key: string]: unknown } | null;
```
avolkovi, line 428: Record<string, object>

packages-exp/auth-types-exp/index.d.ts
```diff
@@ -411,3 +411,18 @@ export interface UserMetadata {
...
   readonly lastSignInTime?: string;
 }
+
+/**
+ * Additional user information.
+ */
+export interface AdditionalUserInfo {
+  readonly isNewUser: boolean;
+  readonly profile?: UserProfile;
+  readonly providerId: ProviderId | null;
```
avolkovi, line 421: providerId can't be null https://firebase.google.com/docs/reference/js/firebase.auth#additionaluserinfo

packages-exp/auth-types-exp/index.d.ts
```diff
@@ -411,3 +411,18 @@ export interface UserMetadata {
...
+  readonly username: string | null;
```
avolkovi, line 422: looks like username should be defined as `username?: string | null` we should go through later and determine how we consolidate these optional | null properties

packages-exp/auth-types-exp/index.d.ts
```diff
@@ -411,3 +411,18 @@ export interface UserMetadata {
   readonly creationTime?: string;
   readonly lastSignInTime?: string;
 }
+
+/**
+ * Additional user information.
+ */
+export interface AdditionalUserInfo {
+  readonly isNewUser: boolean;
+  readonly profile?: UserProfile;
```
avolkovi, line 420: profile should be `UserProfile | null`, it's not optional https://firebase.google.com/docs/reference/js/firebase.auth#additionaluserinfo

yarn.lock
```diff
@@ -0,0 +1,15594 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5", "@babel/code-frame@^7.8.3":
```
avolkovi, line 5: don't check in changes to top level yarn.lock unless you are introducing a new dependency

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,137 @@
...
+    readonly isNewUser: boolean,
+    readonly providerId: ProviderId | null,
+    readonly username: string | null
+  ) {}
+}
+
+class FederatedAdditionalUserInfo extends GenericAdditionalUserInfo {
+  constructor(
+    isNewUser: boolean,
+    providerId: ProviderId | null,
```
avolkovi, line 98: your logic above ensures providerId is never null, can you update the signature to match?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,137 @@
...
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === IdTokenResponseKind.SignupNewUser;
+  if (!providerId && !!idTokenResponse) {
+    const providerId = _parseToken(idTokenResponse.idToken)?.firebase?.[
+      'sign_in_provider'
+    ];
+    assert(
+      // @ts-ignore - Check to see if string is castable to enum.
+      !providerId || Object.values(ProviderId).includes(providerId),
```
avolkovi, line 46: this won't work anymore since we're using const enums, you can probably remove this assertion now until we figure out a better way, also this check doesn't work with federated providers

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        expect(fromIdTokenResponse(idResponse)!).to.be.null;
+      });
+    });
+  });
+});
+
+function idTokenResponse(partial: Partial<IdTokenResponse>): IdTokenResponse {
+  return {
+    idToken: 'Parsing logic not implemented',
+    refreshToken: "Doesn't matter",
```
avolkovi, line 209: nit: we're choosing to use `property-name` in kebab case for the property values in tests, recommend you update to stay consistent with the other tests we wrote, ie: this one should read `refreshToken: 'refresh-token'`

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        expect(providerId).to.eq(ProviderId.FACEBOOK);
+        expect(username).to.be.null;
+        expect(profile).to.eq(profile);
+      });
+    });
+
+    describe('returns null', () => {
+      it('for missing provider IDs', () => {
+        const idResponse = idTokenResponse({});
+        expect(fromIdTokenResponse(idResponse)!).to.be.null;
```
avolkovi, line 200: how does this test work? you're asserting that it's non-null and then checking that it is null?

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,137 @@
...
+  idTokenResponse: IdTokenResponse,
+  appName: string
+): AdditionalUserInfo | null {
+  const { providerId } = idTokenResponse;
+  const profile = idTokenResponse.rawUserInfo
+    ? JSON.parse(idTokenResponse.rawUserInfo)
+    : {};
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === IdTokenResponseKind.SignupNewUser;
```
avolkovi, line 39: the more I look at this the more I feel like we should pass this in instead of depending on on the Kind which is deprecated, we know when we've called signUp

packages-exp/auth-types-exp/index.d.ts
scottcrossen: Well that just isn't true:

https://github.com/firebase/firebase-js-sdk/blob/799507101171bce76879d93ee135b12fb922ffcf/packages/auth/src/additionaluserinfo.js#L137

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
samhorlbeck: heh yeah the `!` "assertion" doesn't actually do anything but mollify the TS compiler. In any case @scottcrossen you can remove it in this case

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,137 @@
...
+  ProviderId
+} from '@firebase/auth-types-exp';
+import { IdTokenResponse, IdTokenResponseKind } from '../../model/id_token';
+import { _parseToken } from './id_token_result';
+import { assert } from '../util/assert';
+
+/**
+ * Parse the `AdditionalUserInfo` from the ID token response.
+ */
+export function fromIdTokenResponse(
```
samhorlbeck, line 29: We're using the `_prefixedWithUnderscore` convention for exported but internal methdods

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,137 @@
...
+  appName: string
+): AdditionalUserInfo | null {
+  const { providerId } = idTokenResponse;
+  const profile = idTokenResponse.rawUserInfo
+    ? JSON.parse(idTokenResponse.rawUserInfo)
+    : {};
+  const isNewUser =
+    !!idTokenResponse.isNewUser ||
+    idTokenResponse.kind === IdTokenResponseKind.SignupNewUser;
+  if (!providerId && !!idTokenResponse) {
```
samhorlbeck, line 40: Don't use `!!` unless you're explicitly casting the value to assign to a `bool` field/variable

go/tsstyle#type-coercion

packages-exp/auth-types-exp/index.d.ts
scottcrossen: Record<string, unknown> will work

packages-exp/auth-types-exp/index.d.ts
scottcrossen: Good catch!

packages-exp/auth-types-exp/index.d.ts
scottcrossen: The docs are wrong on this as well. It's never set for the base info:

https://github.com/firebase/firebase-js-sdk/blob/799507101171bce76879d93ee135b12fb922ffcf/packages/auth/src/additionaluserinfo.js#L114

packages-exp/auth-types-exp/index.d.ts
avolkovi: huh, i wonder if our docs need to be updated then

packages-exp/auth-exp/src/core/user/additional_user_info.test.ts
scottcrossen: Oops yeah my regex find-replace acted out

packages-exp/auth-exp/src/core/user/additional_user_info.ts
scottcrossen: Let's change this logic once we wire in this class. From looking at the legacy SDK this code isn't called in a convienent path from signUp (as far as I can tell)

packages-exp/auth-exp/src/core/user/additional_user_info.ts
```diff
@@ -0,0 +1,138 @@
...
+} from '@firebase/auth-types-exp';
+import { IdTokenResponse, IdTokenResponseKind } from '../../model/id_token';
+import { _parseToken } from './id_token_result';
+
+/**
+ * Parse the `AdditionalUserInfo` from the ID token response.
+ */
+export function _fromIdTokenResponse(
+  idTokenResponse: IdTokenResponse,
+): AdditionalUserInfo | null {
```
samhorlbeck, line 30: I just raised a big PR to make UserCredential accept generics properly. Not for this PR but eventually we'll probably want to make this function generic as well.

See for example: https://github.com/firebase/firebase-js-sdk/pull/3266/files#diff-4e1f7cb1a27ae1754eb03b583f5607ccL389

PR #2992, opened 2020-04-26
Minor deviation in the rxjs firestore examples

rossii123:
To make the code example more consistent with the other examples in the .md file  the function should print `changes` and not `users`. It's also a bit weird to print `users` and not `changes`.


rossii123:
@googlebot I signed it!

Issue #2991, opened 2020-04-26
TypeError: e.addEventListener is not a function - Firestore (Expo)


tomadimitrie:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 10.15.3
  * Browser version: - (Expo SDK 37)
  * Firebase SDK version: 7.14.2
  * Firebase Product: Firestore

### [REQUIRED] Describe the problem

Running any query to Firestore throws this error:

`[Unhandled promise rejection: TypeError: e.addEventListener is not a function. (In 'e.addEventListener("visibilitychange", this.ah)', 'e.addEventListener' is undefined)]`

#### Steps to reproduce:

Initialize Firebase, then run any query.

#### Relevant Code:

```ts
static subscribeToAvailableRides(callback: (rides: Ride[]) => void) {
    return firebase
      .firestore()
      .collection("rides")
      .where("driver.id", "==", null)
      .onSnapshot((docs) => {
        callback(
          docs.docs.map(doc => doc.data())
        );
      });
  }
```

snownoop:
Having same issue. Did not have solution right now.
There is also same topic on stackoverflow:
https://stackoverflow.com/questions/61429599/error-cloud-firestore-addeventlistener-is-not-a-function-react-native-firestor

Any ideas ?

tomadimitrie:
> Having same issue. Did not have solution right now.
> 
> There is also same topic on stackoverflow:
> 
> https://stackoverflow.com/questions/61429599/error-cloud-firestore-addeventlistener-is-not-a-function-react-native-firestor
> 
> 
> 
> Any ideas ?

I am using Expo so I ran `expo install firebase` and it installed 7.9.0 and the error was gone

schmidt-sebastian:
@tomadimitrie Thanks for this report. We added a `window. addEventListener()` call that is executed when `window` is defined. I will add a check to only call `addEventListener` when it `window.addEventListener` exists.

gwendall:
@tomadimitrie in the meantime, mock it by putting something like `window.addEventListener = x => x` at the top level of your code

tomadimitrie:
@gwendall thank you. One question, is the last version of the Firebase JS SDK fully compatible with Expo? For some reason running `expo install firebase` installs version 7.9.0 and I don't know whether to stick with that or upgrade to the latest version

edi:
[...]

PR #2994, opened 2020-04-27
Check for addEventListener() before calling

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-js-sdk/issues/2991

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -240,7 +240,7 @@ export class AsyncQueue {
 
   constructor() {
     const window = PlatformSupport.getPlatform().window;
-    if (window) {
+    if (window && typeof window.addEventListener === 'function') {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/util/async_queue.ts
```diff
@@ -240,7 +240,7 @@ export class AsyncQueue {
 
   constructor() {
     const window = PlatformSupport.getPlatform().window;
-    if (window) {
+    if (window && typeof window.addEventListener === 'function') {
```
wu-hui, line 243: Why checking for the type particularly? Are there platforms where this exists but not a function? Just curious, LGTM anyways.

schmidt-sebastian: I don't want to find out....

dichai1983: This check is needed for React Native dev.

PR #2995, opened 2020-04-28
Take WriteStream offline when IndexedDB is unavailable

schmidt-sebastian:
This PR does two things:

- In the first commit, it removes "setLastStreamToken()", "getLastStreamToken()" and "acknowledgeBadge", which manage unused state in IndexedDB and complicate the logic needed to recover the WriteStream from IndexedDB failures. When the backend implements proper support for these tokens, the logic in this PR should be entirely rewritten to rely on the `lastStreamToken`.

- In the second commit, the WriteStream now goes offline if a write acknowledgement cannot be persisted and remains offline until it can.

Addresses #2755


wilhuff:
I don't understand this is useful:

> * In the first commit, it removes "setLastSnapshotVersion", "getLastSnapshotVersion" and "acknowledgeBadge", which manage unused state in IndexedDB and complicate the logic needed to recover the WriteStream from IndexedDB failures. When the backend implements proper support for these tokens, the logic in this PR should be entirely rewritten to rely on the `lastSnapshotVersion`.

We've implemented the client portion of the protocol--why remove it? The protocol relies on stream tokens not `lastSnapshotVersion` so I don't see how these substitute for each other.

schmidt-sebastian:
> We've implemented the client portion of the protocol--why remove it? The protocol relies on stream tokens not `lastSnapshotVersion` so I don't see how these substitute for each other.

Sorry - I confused myself when I wrote the PR description (since updated). This removes `setLastStreamToken()` and `getLastStreamToken()` and doesn't interact with the 'lastRemoteSnapshotVersion. 

Keeping the stream token makes the retry logic much harder as we now also need to address three more areas:

- https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/remote_store.ts#L593
- https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/remote_store.ts#L675
- https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/remote/remote_store.ts#191

On top of the complication this creates for this PR, my reason for removing this code is:

- It is very easy code to add back. It just stores and reads values.
- I would be very surprised if the backend added support for these stream resumptions in any reasonable timeframe.
- If the backend does add support, the entire retry logic in this PR should be written on top of this behavior, and hence the entire PR needs to be rewritten.
- There is no actual test that we implemented the protocol as designed. Even without the removal, enabling support for stream tokens will be non trivial since we at the very least need to add spec tests that verify the behavior.

wilhuff:
I still don't follow how this is a good course.

Once we ship this series of PRs we're promising that we're going to handle IndexedDB being unavailable while in the background. A whole class of applications that currently can't work will come to depend upon this behavior, locking us into it.

In the meantime, you say it will be easy to add back later, but if that were true, we should just do it now. I suspect that if this is hard to do now it's actually going to be hard to re-add. To me this means we shouldn't commit to being able to handle IndexedDB background problems until we're sure we can load/store stream tokens correctly.

If you're not confident you can make changes to this without tests to support it (a reasonable supposition), the right thing to do is add the tests.

schmidt-sebastian:
What I said above is not a statement on the PR, but on the code that is removed in `2a6ee9c`.

With this code removed, there are only two failure cases left and the code in this PR _does_ address these.

wilhuff:
Right, I'm discussing the code removed in 2a6ee9c as well.

If it's too hard to keep that code working now I don't see why we should proceed with this approach. It means we'll commit ourselves to a public behavior (handling IndexedDB backgrounding errors) but can't be sure we actually make the necessary underlying stream token persistence work in combination with that.

schmidt-sebastian:
I updated the PR based on the decision that we can remove the `lastStreamToken`. This change was merged as https://github.com/firebase/firebase-js-sdk/pull/3132

packages/firestore/src/remote/remote_store.ts
```diff
@@ -661,11 +678,17 @@ export class RemoteStore implements TargetMetadataProvider {
     );
     const batch = this.writePipeline.shift()!;
     const success = MutationBatchResult.from(batch, commitVersion, results);
-    return this.syncEngine.applySuccessfulWrite(success).then(() => {
-      // It's possible that with the completion of this mutation another
-      // slot has freed up.
-      return this.fillWritePipeline();
-    });
+    try {
```
wilhuff, line 681: It almost seems like this try/catch blurb could be its own method that take the operation to perform, tries it, and then passes it to `disableNetworkUntilRecovery`.

schmidt-sebastian: Added a small wrapper. It is only used twice, but the wrapper itself is one one line, which is kind of nice.

PR #2999, opened 2020-04-29
Drop SortedSet from FieldMask

schmidt-sebastian:
I would like to not use SortedSet in FieldMask since the Rest Wrapper API ideally does not pull in our custom Red Black Tree. FieldMasks are used in any update/set(merge:true) call.

Right now, I have a debug assert that makes sure FieldMasks only contain unique fields. This means we have to validate this as we turn user data into FieldMasks. We could also strip duplicated fields in the FieldMask constructor.

### DIFF
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -20,7 +20,6 @@ import * as api from '../protos/firestore_proto_api';
 import { Timestamp } from '../api/timestamp';
 import { SnapshotVersion } from '../core/snapshot_version';
 import { debugAssert, fail, hardAssert } from '../util/assert';
-import { SortedSet } from '../util/sorted_set';
 
 import {
   Document,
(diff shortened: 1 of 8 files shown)


packages/firestore/src/model/mutation.ts
```diff
@@ -45,18 +44,14 @@ import { arrayEquals } from '../util/misc';
...
-
-  static fromArray(fields: FieldPath[]): FieldMask {
-    let fieldsAsSet = new SortedSet<FieldPath>(FieldPath.comparator);
-    fields.forEach(fieldPath => (fieldsAsSet = fieldsAsSet.add(fieldPath)));
-    return new FieldMask(fieldsAsSet);
+    // Sort the field mask to support `FieldMask.isEqual()` and assert below.
+    fields.sort(FieldPath.comparator);
+    debugAssert(
+      !fields.some((v, i) => i !== 0 && v.isEqual(fields[i - 1])),
+      'FieldMask contains fields that are not unique'
```
dconeybe, line 53: It may be helpful to include the name of the repeated field in the message. Consider adding this.

schmidt-sebastian: I couldn't think of a clean way to do this, so I added an ugly way. Done.

PR #3005, opened 2020-05-01
Add an integration test for calling clearPersistence() on a new Firestore instance

dconeybe:


packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1096,6 +1096,30 @@ apiDescribe('Database', (persistence: boolean) => {
...
+        const firestore = docRef.firestore;
+        await docRef.set({ foo: 'bar' });
+        const app = docRef.firestore.app;
+        const name = app.name;
+        const options = app.options;
+
+        await app.delete();
+        const app2 = firebase.initializeApp(options, name);
+        const firestore2 = firebase.firestore!(app2);
+        await firestore2.clearPersistence();
```
dconeybe, line 1113: Note that the _only_ difference between this test case and the one above it is that the location of the call to clearPersistence() has moved to be invoked on the newly-created `firestore2` object.

schmidt-sebastian: Just checking - is this test meaningfully different from:

```
// eslint-disable-next-line no-restricted-properties
(persistence ? it : it.skip)( 'can clear persistence if the client has not been initialized',
  () => withTestDb(db => db.clearPersistence())
)
```

You should also update your PR title. You are writing a super fancy integration test, and not a unit test :)

dconeybe: The "difference" between that 1-line test is that it doesn't verify the side effects of `clearPersistence()`. Unfortunately, the extra lines all exist to provide the verification that `clearPersistence()` does indeed clear the persistence. Also, I will update the PR description as suggested.

schmidt-sebastian: Makes sense. I did not quite catch that the first time around.

PR #3010, opened 2020-05-04
Take WatchStream offline when IndexedDB is unavailable

schmidt-sebastian:
This PR takes the network offline if we fail to talk to IndexedDB when applying a query snapshot or rejection.

Also cleans up some Promise handling in RemoteStore.

Addresses #2755

### DIFF
--- a/packages/firestore/src/remote/remote_store.ts
+++ b/packages/firestore/src/remote/remote_store.ts
@@ -124,10 +124,11 @@ export class RemoteStore implements TargetMetadataProvider {
   private isPrimary = false;
 
   /**
-   * A barrier that prevents the restart of the network streams. Incremented for
-   * each IndexedDB failure until access becomes available again.
+   * When set to `true`, the network was taken offline due to an IndexedDB
+   * failure. The state is flipped to `false` when access becomes available
+   * again.
    */
-  private indexedDbFailureBarrier = 0;
+  private indexedDbFailed = false;
 
   private onlineStateTracker: OnlineStateTracker;
 
@@ -217,9 +218,6 @@ export class RemoteStore implements TargetMetadataProvider {
   async disableNetwork(): Promise<void> {
     this.networkEnabled = false;
     await this.disableNetworkInternal();
-
-    // Set the OnlineState to Offline so get()s return from cache, etc.
-    this.onlineStateTracker.set(OnlineState.Offline);
   }
 
   private async disableNetworkInternal(): Promise<void> {
@@ -235,6 +233,9 @@ export class RemoteStore implements TargetMetadataProvider {
     }
 
     this.cleanUpWatchStreamState();
+
+    // Set the OnlineState to Offline so get()s return from cache, etc.
+    this.onlineStateTracker.set(OnlineState.Offline);
   }
 
   async shutdown(): Promise<void> {
@@ -348,11 +349,7 @@ export class RemoteStore implements TargetMetadataProvider {
   }
 
   canUseNetwork(): boolean {
-    return (
-      this.indexedDbFailureBarrier === 0 &&
-      this.isPrimary &&
-      this.networkEnabled
-    );
+    return !this.indexedDbFailed && this.isPrimary && this.networkEnabled;
   }
 
   private cleanUpWatchStreamState(): void {
@@ -407,7 +404,12 @@ export class RemoteStore implements TargetMetadataProvider {
       try {
         await this.handleTargetError(watchChange);
       } catch (e) {
-        logDebug(LOG_TAG, 'Failed to remove target: ' + e);
+        logDebug(
+          LOG_TAG,
+          'Failed to remove targets %s: %s ',
+          watchChange.targetIds.join(','),
+          e
+        );
         await this.disableNetworkUntilRecovery(e);
       }
       return;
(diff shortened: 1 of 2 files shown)


packages/firestore/src/remote/remote_store.ts
```diff
@@ -407,15 +426,43 @@ export class RemoteStore implements TargetMetadataProvider {
...
   }
 
+  /**
+   * Recovery logic for IndexedDB errors that takes the network offline until
+   * IndexedDb probing succeeds. Retries are scheduled with backoff using
+   * `enqueueRetryable()`.
+   */
+  private async disableNetworkUntilRecovery(e: FirestoreError): Promise<void> {
+    if (e.name === 'IndexedDbTransactionError') {
+      ++this.indexedDbFailureBarrier;
```
wilhuff, line 450: This mechanism seems more complicated/fiddly than it needs to be.

If we get multiple watch change events in a row, this is going to bump this `indexedDbFailureBarrier` multiple times and enqueue multiple retry tasks. When IndexedDB becomes available, each of them is going to succeed, decrement the count, and try to enable the network but only the last call to enableNetworkInternal will succeed.

It seems like instead of keeping a count, we could instead keep a boolean like `indexedDbFailed`. Then in this method:

  * If not `indexedDbFailed`, set it and enqueue the retry task
  * If `indexedDbFailed`, just log the error but do nothing.
  * In the retry task, if `getLastRemoteSnapshotVersion` succeeds, flip `indexedDbFailed` back to false and enable the network.

packages/firestore/src/remote/remote_store.ts
```diff
@@ -407,15 +426,43 @@ export class RemoteStore implements TargetMetadataProvider {
...
+      await this.disableNetworkInternal();
```
wilhuff, line 451: `disableNetworkInternal` does not change the `onlineStateTracker`, but it seems like while we've intentionally disabled the network we should mark ourselves offline, no?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -407,15 +426,43 @@ export class RemoteStore implements TargetMetadataProvider {
...
-        await this.raiseWatchSnapshot(snapshotVersion);
+      try {
+        const lastRemoteSnapshotVersion = await this.localStore.getLastRemoteSnapshotVersion();
+        if (snapshotVersion.compareTo(lastRemoteSnapshotVersion) >= 0) {
+          // We have received a target change with a global snapshot if the snapshot
+          // version is not equal to SnapshotVersion.min().
+          await this.raiseWatchSnapshot(snapshotVersion);
+        }
+      } catch (e) {
+        logDebug(LOG_TAG, 'Failed to raise snapshot: ' + e);
```
wilhuff, line 437: Is there any identifying information about the snapshot that we can log here?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -391,7 +404,13 @@ export class RemoteStore implements TargetMetadataProvider {
     ) {
       // There was an error on a target, don't wait for a consistent snapshot
       // to raise events
-      return this.handleTargetError(watchChange);
+      try {
+        await this.handleTargetError(watchChange);
+      } catch (e) {
+        logDebug(LOG_TAG, 'Failed to remove target: ' + e);
```
wilhuff, line 410: Can we log the query associated with the target here?

schmidt-sebastian: Unless we go out of our way, we don't have access to the query here. I logged the target ID instead. We should however already have this information from the Watch stream. 

FWIW, I didn't add this logging to begin with since this is a debug log. The other logs were error logs in response to user behavior. They are always visible, and often without context.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: I am not sure if it will help, but I added the snapshot version.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: You say "complicated/fiddly", I say "advanced" :) 

We should only ever get one of these events, since we go offline right away and discard all further messages from the stream. I realize that this is a good argument for making this a boolean and updated the code accordingly.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: That will create some flicker, but it is the right thing to do. Updated.

schmidt-sebastian: I had to add the state transition here since we don't raise offline events when `disableNetworkInternal` is called from `shutdown` or for primary state transitions.

packages/firestore/src/remote/remote_store.ts
wilhuff: The snapshot version probably isn't that useful, but the target ID would be. Anything to help associate this message with its context.

If the targetId isn't available going back to what you had before is probably better.

schmidt-sebastian: I changed this back to the original message. The target IDs are not easily available:

- DocumentWatchChange has `updatedTargetIds` and `removedTargetIds`
- ExistenceFilterChange has `targetId`
- WatchTargetChange has `targetIds`

If I understand the flow correctly, we should always be in a WatchTargetChange to raise a snapshot, but I don't want to bake this assumption into the code for better logging. I can invest some time here, but the targets are already in the Watch response that we log right before.

PR #3011, opened 2020-05-04
Update SourceMaps during transform

schmidt-sebastian:


schmidt-sebastian:
Example from before that failed to map renamed members:

Screen Shot 2020-05-04 at 1 09 53 PM


Example with PR:

Screen Shot 2020-05-04 at 1 07 50 PM

PR #3012, opened 2020-05-04
Reland "Use crypo RNG for auto ID generation to reduce conflicts (#2764)"

wu-hui:
Manually verified on IE11 and web-workers.

### DIFF
--- a/packages/firestore/src/platform/platform.ts
+++ b/packages/firestore/src/platform/platform.ts
@@ -43,6 +43,12 @@ export interface Platform {
   /** Converts a binary string to a Base64 encoded string. */
   btoa(raw: string): string;
 
+  /**
+   * Generates `nBytes` of random bytes. If `nBytes` is negative, an empty array
+   * will be returned.
+   */
+  randomBytes(nBytes: number): Uint8Array;
+
   /** The Platform's 'window' implementation or null if not available. */
   readonly window: Window | null;
 
--- a/packages/firestore/src/platform/platform.ts
+++ b/packages/firestore/src/platform/platform.ts
@@ -46,7 +46,7 @@ export interface Platform {
   /**
    * Generates `nBytes` of random bytes.
    *
-   * If `nBytes <= 0` , an error will be thrown.
+   * If `nBytes < 0` , an error will be thrown.
    */
   randomBytes(nBytes: number): Uint8Array;
 
--- a/packages/firestore/src/platform/platform.ts
+++ b/packages/firestore/src/platform/platform.ts
@@ -44,8 +44,9 @@ export interface Platform {
   btoa(raw: string): string;
 
   /**
-   * Generates `nBytes` of random bytes. If `nBytes` is negative, an empty array
-   * will be returned.
+   * Generates `nBytes` of random bytes.
+   *
+   * If `nBytes <= 0` , an error will be thrown.
    */
   randomBytes(nBytes: number): Uint8Array;
 
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -24,6 +24,7 @@ import { ConnectivityMonitor } from './../remote/connectivity_monitor';
 import { NoopConnectivityMonitor } from '../remote/connectivity_monitor_noop';
 import { BrowserConnectivityMonitor } from './browser_connectivity_monitor';
 import { WebChannelConnection } from './webchannel_connection';
+import { validatePositiveNumber } from '../util/input_validation';
 
 // Implements the Platform API for browsers and some browser-like environments
 // (including ReactNative).
@@ -75,18 +76,13 @@ export class BrowserPlatform implements Platform {
   }
 
   randomBytes(nBytes: number): Uint8Array {
-    if (nBytes <= 0) {
-      return new Uint8Array();
-    }
+    validatePositiveNumber('randomBytes', 1, nBytes);
 
     // Polyfill for IE and WebWorker
-    // eslint-disable-next-line no-restricted-globals
     const crypto =
-      (this.window &&
-        (this.window.crypto || (this.window as any)['msCrypto'])) ||
-      (self && self.crypto);
+      typeof self !== 'undefined' && (self.crypto || (self as any)['msCrypto']);
     const v = new Uint8Array(nBytes);
-    if (!!crypto) {
+    if (crypto) {
       crypto.getRandomValues(v);
     } else {
       // Falls back to Math.random
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -28,7 +28,6 @@ import { WebChannelConnection } from './webchannel_connection';
 // Implements the Platform API for browsers and some browser-like environments
 // (including ReactNative).
 export class BrowserPlatform implements Platform {
-  readonly useProto3Json = true;
   readonly base64Available: boolean;
 
   constructor() {
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -76,21 +76,21 @@ export class BrowserPlatform implements Platform {
   }
 
   randomBytes(nBytes: number): Uint8Array {
-    debugAssert(nBytes >= 0, 'Number of random bytes must not be negative');
+    debugAssert(nBytes >= 0, `Expecting non-negative nBytes, got: ${nBytes}`);
 
-    // Polyfill for IE and WebWorker
+    // Polyfills for IE and WebWorker by using `self` and `msCrypto` when `crypto` is not available.
     const crypto =
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       typeof self !== 'undefined' && (self.crypto || (self as any)['msCrypto']);
-    const v = new Uint8Array(nBytes);
+    const bytes = new Uint8Array(nBytes);
     if (crypto) {
-      crypto.getRandomValues(v);
+      crypto.getRandomValues(bytes);
     } else {
       // Falls back to Math.random
       for (let i = 0; i < nBytes; i++) {
-        v[i] = Math.floor(Math.random() * 256);
+        bytes[i] = Math.floor(Math.random() * 256);
       }
     }
-    return v;
+    return bytes;
   }
 }
--- a/packages/firestore/src/platform_browser/browser_platform.ts
+++ b/packages/firestore/src/platform_browser/browser_platform.ts
@@ -20,13 +20,19 @@ import { Platform } from '../platform/platform';
 import { Connection } from '../remote/connection';
 import { JsonProtoSerializer } from '../remote/serializer';
 import { ConnectivityMonitor } from './../remote/connectivity_monitor';
+
 import { NoopConnectivityMonitor } from '../remote/connectivity_monitor_noop';
 import { BrowserConnectivityMonitor } from './browser_connectivity_monitor';
 import { WebChannelConnection } from './webchannel_connection';
 
+// Polyfill for IE and WebWorker
+// eslint-disable-next-line no-restricted-globals
+const crypto = window['crypto'] || window['msCrypto'] || self['crypto'];
+
 // Implements the Platform API for browsers and some browser-like environments
 // (including ReactNative).
 export class BrowserPlatform implements Platform {
+  readonly useProto3Json = true;
   readonly base64Available: boolean;
 
   constructor() {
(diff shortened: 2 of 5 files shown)


packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +73,27 @@ export class BrowserPlatform implements Platform {
...
   }
+
+  randomBytes(nBytes: number): Uint8Array {
+    if (nBytes <= 0) {
+      return new Uint8Array();
+    }
+
+    // Polyfill for IE and WebWorker
+    // eslint-disable-next-line no-restricted-globals
+    const crypto =
```
schmidt-sebastian, line 84: Should we add support for "self" as well? 
https://stackoverflow.com/questions/11219775/global-variable-in-web-worker

We might also consider adding support for this shim on ReactNative: https://stackoverflow.com/questions/45301900/howto-patch-shim-crypto-getrandomvalues-for-react-native

While the fallback will make sure these environments don't break, we can possibly extend our support for `crypto`.

packages/firestore/src/platform/platform.ts
```diff
@@ -43,6 +43,12 @@ export interface Platform {
   /** Converts a binary string to a Base64 encoded string. */
   btoa(raw: string): string;
 
+  /**
+   * Generates `nBytes` of random bytes. If `nBytes` is negative, an empty array
+   * will be returned.
```
dconeybe, line 48: What is the rationale for accepting negative numbers? This seems like it may be hiding bugs in the calling code.

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +73,27 @@ export class BrowserPlatform implements Platform {
...
+      (this.window &&
+        (this.window.crypto || (this.window as any)['msCrypto'])) ||
+      (self && self.crypto);
+    const v = new Uint8Array(nBytes);
+    if (!!crypto) {
```
dconeybe, line 89: Is this double exclamation mark intentional? If not, you may want to add a unit test that would fail if Math.random() is used unintentionally.

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +73,27 @@ export class BrowserPlatform implements Platform {
...
+    if (nBytes <= 0) {
+      return new Uint8Array();
+    }
+
+    // Polyfill for IE and WebWorker
+    // eslint-disable-next-line no-restricted-globals
+    const crypto =
+      (this.window &&
+        (this.window.crypto || (this.window as any)['msCrypto'])) ||
+      (self && self.crypto);
```
schmidt-sebastian, line 87: According to https://developer.mozilla.org/en-US/docs/Web/API/Window/self, you can just write:

```
const crypto = typeof self !=== 'undefined' && (self.crypto || (self as any)['msCrypto']);
```

packages/firestore/src/platform_browser/browser_platform.ts
schmidt-sebastian: This can probably just be `if(crypto)` since an if statement just needs a "truthy" value and not a boolean.

packages/firestore/src/util/misc.ts
```diff
@@ -28,8 +29,17 @@ export class AutoId {
     const chars =
       'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
     let autoId = '';
-    for (let i = 0; i < 20; i++) {
-      autoId += chars.charAt(Math.floor(Math.random() * chars.length));
+    while (autoId.length < 20) {
+      const bytes = PlatformSupport.getPlatform().randomBytes(40);
+      for (const b of Array.from(bytes)) {
```
schmidt-sebastian, line 34: Nit: You can save one copy here if you do a `for(let i = 0; i < bytes.length; ++i)` loop.

packages/firestore/src/platform_browser/browser_platform.ts
wu-hui: Done.

packages/firestore/src/util/misc.ts
wu-hui: Done.

packages/firestore/src/platform_browser/browser_platform.ts
wu-hui: Done.

packages/firestore/src/platform/platform.ts
wu-hui: Changed it to validate input, and reject anything <=0.

packages/firestore/src/platform_browser/browser_platform.ts
wu-hui: So we don't need it, right?

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +74,22 @@ export class BrowserPlatform implements Platform {
   btoa(raw: string): string {
     return btoa(raw);
   }
+
+  randomBytes(nBytes: number): Uint8Array {
+    validatePositiveNumber('randomBytes', 1, nBytes);
```
schmidt-sebastian, line 79: This should be a `debugAssert` since we are not validating a public API call.

You should also allow calls with `nBytes: 0` (unless you have to write special code for this).

packages/firestore/src/platform_browser/browser_platform.ts
schmidt-sebastian: I am ok with not adding support for ReactNative until the ReactNative Firebase wrapper bundles "crypto", but note that without getting "crypto" from globals, we will always fall back to Math.random() for RN.

wu-hui: Yep, I am filing a ticket to get crypto added for react native firestore.

packages/firestore/src/platform_browser/browser_platform.ts
wu-hui: Done.

packages/firestore/src/platform_node/node_platform.ts
```diff
@@ -75,4 +77,10 @@ export class NodePlatform implements Platform {
   btoa(raw: string): string {
     return new Buffer(raw, 'binary').toString('base64');
   }
+
+  randomBytes(nBytes: number): Uint8Array {
+    validatePositiveNumber('randomBytes', 1, nBytes);
```
schmidt-sebastian, line 82: DebugAssert here as well.

packages/firestore/src/platform/platform.ts
```diff
@@ -43,6 +43,13 @@ export interface Platform {
   /** Converts a binary string to a Base64 encoded string. */
   btoa(raw: string): string;
 
+  /**
+   * Generates `nBytes` of random bytes.
+   *
+   * If `nBytes <= 0` , an error will be thrown.
```
dconeybe, line 49: nit: nBytes==0 is valid; please update this comment to state that it will only reject values that are *strictly* less than zero.

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +74,23 @@ export class BrowserPlatform implements Platform {
   btoa(raw: string): string {
     return btoa(raw);
   }
+
+  randomBytes(nBytes: number): Uint8Array {
+    debugAssert(nBytes >= 0, 'Number of random bytes must not be negative');
```
dconeybe, line 79: Consider including the invalid value in the failure message; this could assist in debugging should the assert ever get triggered and all we have to use for debugging is the stack track and error message.

e.g.
```js
debugAssert(nBytes >= 0, `invalid nBytes: ${nBytes}`);
```

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +74,23 @@ export class BrowserPlatform implements Platform {
...
+
+    // Polyfill for IE and WebWorker
```
dconeybe, line 81: Can you improve this comment to explain how it affects the reader? Is the polyfill being done by the code block that follows? Or is it instructing the reader to perform some polyfilling? Consider re-writing this comment as a sentence that would be either clearly informative or clearly actionable by the reader.

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -72,4 +74,23 @@ export class BrowserPlatform implements Platform {
...
+    const crypto =
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      typeof self !== 'undefined' && (self.crypto || (self as any)['msCrypto']);
+    const v = new Uint8Array(nBytes);
```
dconeybe, line 85: nit: Consider a more meaningful name than `v` (e.g. `generatedBytes`).

packages/firestore/src/util/misc.ts
```diff
@@ -28,8 +29,17 @@ export class AutoId {
...
       'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
     let autoId = '';
-    for (let i = 0; i < 20; i++) {
-      autoId += chars.charAt(Math.floor(Math.random() * chars.length));
+    while (autoId.length < 20) {
+      const bytes = PlatformSupport.getPlatform().randomBytes(40);
+      for (let i = 0; i < bytes.length; ++i) {
+        // Length of `chars` is 62. We only take bytes between 0 and 62*4-1
+        // (both inclusive). The value is then evenly mapped to indices of `char`
+        // via a modulo operation.
```
dconeybe, line 37: I'm assuming this logic is used to ensure an even distribution between all of the characters in `chars`. Please update the comment to explain _why_ this logic exists because it is non-obvious.

packages/firestore/src/util/misc.ts
```diff
@@ -28,8 +29,17 @@ export class AutoId {
     const chars =
       'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
     let autoId = '';
-    for (let i = 0; i < 20; i++) {
-      autoId += chars.charAt(Math.floor(Math.random() * chars.length));
+    while (autoId.length < 20) {
```
dconeybe, line 32: Consider re-writing this loop as `while (true)` and then breaking out of the loop with a `break` statement immediately after appending the 20th character to `autoId`. This would avoid hardcoding `20` in two different places.

packages/firestore/src/util/misc.ts
```diff
@@ -28,8 +29,17 @@ export class AutoId {
...
+      const bytes = PlatformSupport.getPlatform().randomBytes(40);
+      for (let i = 0; i < bytes.length; ++i) {
+        // Length of `chars` is 62. We only take bytes between 0 and 62*4-1
+        // (both inclusive). The value is then evenly mapped to indices of `char`
+        // via a modulo operation.
+        const maxValue = 62 * 4 - 1;
```
dconeybe, line 38: Consider moving the calculation of maxValue to immediately follow the declaration of `chars`. IMO, having it re-calculated on each loop iteration detracts from readability because it incorrectly implies that its value may change between iterations.

dconeybe, line 38: Hardcoding `62` in three different places (including in the comment) is a bit unnerving.  It would be preferable to use `chars.length`, for robustness. How about something like this:

```js
const chars = '...';
const maxValue = chars.length * 4 - 1;
debugAssert(maxValue < 256);
```

Then the hardcoded value of `62` in the modulo operation below can be safely replaced with `chars.length`.

packages/firestore/src/platform_node/node_platform.ts
dconeybe: Also, make sure that `0` is accepted as a valid value for `nBytes`.

packages/firestore/src/platform_browser/browser_platform.ts
mikehardy: react-native-firebase participant here, just cross-linking https://github.com/invertase/react-native-firebase/issues/3618

Salakar: Just thought I'd quickly summerize; React Native Firebase doesn't use the JS SDK, we're a thin JS wrapper around the native Android & iOS SDKs so this change has no direct affect on us. 

However our JS wrapper does use an identical random id generator - so I think we may need to port this change across in some form - we'll pick this up separately, thanks!

packages/firestore/src/util/misc.ts
wu-hui: Rewrote the section to reduce hardcoding numbers.

PR #3015, opened 2020-05-04
Ignore IndexedDB failures during garbage collection

schmidt-sebastian:
Rationale is that they the LRU task will retry its run when it is invoked next.

There are no tests since the GC scheduling code is not tested right now. Let me know if you would like me to add tests.

Note that there is a behavior change: If the primary lease is lost, the GC run is rescheduled. It will however get stoped in
https://github.com/firebase/firebase-js-sdk/blob/5a60243bf264967819fc5c3897892084c5eb4d43/packages/firestore/src/core/component_provider.ts#L203

Addresses #2755

### DIFF
--- a/packages/firestore/src/local/lru_garbage_collector.ts
+++ b/packages/firestore/src/local/lru_garbage_collector.ts
@@ -274,7 +274,7 @@ export class LruScheduler implements GarbageCollectionScheduler {
           await localStore.collectGarbage(this.garbageCollector);
         } catch (e) {
           if (e.name === 'IndexedDbTransactionError') {
-            logDebug(LOG_TAG, 'Ignoring IndexedDB error during LRU: ', e);
+            logDebug(LOG_TAG, 'Ignoring IndexedDB error during garbage collection: ', e);
           } else {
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/lru_garbage_collector.ts
```diff
@@ -265,13 +267,19 @@ export class LruScheduler implements GarbageCollectionScheduler {
...
         this.hasRun = true;
-        return localStore
-          .collectGarbage(this.garbageCollector)
-          .then(() => this.scheduleGC(localStore))
-          .catch(ignoreIfPrimaryLeaseLoss);
+        try {
+          await localStore.collectGarbage(this.garbageCollector);
+        } catch (e) {
+          if (e.name === 'IndexedDbTransactionError') {
+            logDebug(LOG_TAG, 'Ignoring IndexedDB error during LRU: ', e);
```
wilhuff, line 277: "LRU" is not a great descriptor of what's going on here. How about "garbage collection"?

Also update the PR description.

schmidt-sebastian: Done

PR #3020, opened 2020-05-05
Ignore IndexedDB failure during lease refresh

schmidt-sebastian:
Adresses #2755

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -101,7 +101,7 @@ const MAX_PRIMARY_ELIGIBLE_AGE_MS = 5000;
 const CLIENT_METADATA_REFRESH_INTERVAL_MS = 4000;
 /** User-facing error when the primary lease is required but not available. */
 const PRIMARY_LEASE_EXCLUSIVE_ERROR_MSG =
-  'Another tab has exclusive access to the persistence layer. ' +
+  'Failed to obtain exclusive access to the persistence layer. ' +
   'To allow shared access, make sure to invoke ' +
   '`enablePersistence()` with `synchronizeTabs:true` in all tabs.';
 const UNSUPPORTED_PLATFORM_ERROR_MSG =
@@ -191,11 +191,12 @@ export class IndexedDbPersistence implements Persistence {
   private readonly document: Document | null;
   private readonly window: Window;
 
-  // Technically these types should be `| undefined` because they are
+  // Technically `simpleDb` should be `| undefined` because it is
   // initialized asynchronously by start(), but that would be more misleading
   // than useful.
   private simpleDb!: SimpleDb;
-  private listenSequence!: ListenSequence;
+
+  private listenSequence: ListenSequence | null = null;
 
   private _started = false;
   private isPrimary = false;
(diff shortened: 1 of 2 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -412,8 +412,16 @@ export class IndexedDbPersistence implements Persistence {
       })
       .catch(e => {
         if (!this.allowTabSynchronization) {
-          throw e;
-        }
+          if (e.name === 'IndexedDbTransactionError') {
+            logDebug(LOG_TAG, "Failed to extend owner lease: ", e);
+            // Proceed in primary mode since the client was not initialized
```
wilhuff, line 417: Why is proceeding here assuming we're the primary a useful thing?

When this method is called in the course of `enablePersistence`, don't we want to report a failure to open the database?

schmidt-sebastian: The idea was to let start() proceed if this check failed. I have now changed the code to fail start() and hence enablePersistence() if we get an IndexedDbTransactionError and `synchronizeTabs` is false. That may however hide that we can never obtain the lease. The fix for this was pretty straightforward, but I ended up changing the callsites to use `this.runTransaction` instead of `this.simpleDb.runTransaction`, as this allows me to use MockPersistence to inject failures.

Note that the behavior is not all that different: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/indexeddb_persistence.ts#L808

We make one additional call to verify the lease (which succeeds if no other client has exclusive access). We also obtain access to all ObjectStores, but I think that is not all that different since there is no operation that doesn't also access the DbPrimaryStore.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -561,33 +580,10 @@ export class IndexedDbPersistence implements Persistence {
...
-          }
-
-          if (!this.isLocalClient(currentPrimary)) {
-            if (!currentPrimary!.allowTabSynchronization) {
-              // Fail the `canActAsPrimary` check if the current leaseholder has
-              // not opted into multi-tab synchronization. If this happens at
-              // client startup, we reject the Promise returned by
-              // `enablePersistence()` and the user can continue to use Firestore
-              // with in-memory persistence.
-              // If this fails during a lease refresh, we will instead block the
```
wilhuff, line 573: It seems like removing this error here is a regression. Consider:
  * If the client successfully starts
  * goes into the background and loses access to IndexedDB
  * stays in the background for longer than the lease duration

If device comes back online the user opens another tab before our retry has succeeded in refreshing the lease, there's a very real chance we'll lose this race. Users don't even have to be mixing and matching modes for this to fail now.

schmidt-sebastian: Summary of my last 5 minutes:
- Gil is wrong.
- Gil is right.
- Gil is wrong.
- Gil is right.
- Gil it right and wrong.

The idea behind the lease refresh was it would be a "best effort" only operation. If if fails or stalls, any operation to IndexedDB should verify the expected state. That is:

- "readwrite-primary" operations should verify that the tab has the lease or can become the leaseholder. This already happens.
- "readwrite", "readonly" operations should verify that they are allowed to access IndexedDB (i.e. there is no primary tab without `synchronizeTabs` turned on). This is not happening right now. Instead, this is periodically verified via the code that I removed.

So, in conclusion: We should move this verification and run it as part of every transaction. I will leave the code in `master` and work on this in a separate PR.

PR #3025, opened 2020-05-05
Add invokeRunQueryRpc()

schmidt-sebastian:
This adds a new `invokeRunQueryRpc` RPC that is used (and tested) in the Lite SDK to run queries via Rest.

This is similar to https://github.com/firebase/firebase-js-sdk/blob/96cd91dd7825f845ce860ffb98703dc48716b81b/packages/firestore/src/remote/datastore.ts#L124

Note: 
- The Query RPC needs a parent set. The types are not correct in `master`. See https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L442

- RunQueryResponse returns a stream of documents followed by an empty result which just has `readTime` set. See https://github.com/googleapis/googleapis/blob/master/google/firestore/v1/firestore.proto#L477

### DIFF
--- a/packages/firestore/src/remote/datastore.ts
+++ b/packages/firestore/src/remote/datastore.ts
@@ -171,7 +171,7 @@ export async function invokeRunQueryRpc(
 
   return (
     response
-      // Filter RunQueryResponses that only contain readTimes.
+      // Omit RunQueryResponses that only contain readTimes.
       .filter(proto => !!proto.document)
       .map(proto => datastoreImpl.serializer.fromDocument(proto.document!))
   );
--- a/packages/firestore/src/remote/datastore.ts
+++ b/packages/firestore/src/remote/datastore.ts
@@ -16,7 +16,7 @@
  */
 
 import { CredentialsProvider } from '../api/credentials';
-import { MaybeDocument } from '../model/document';
+import { MaybeDocument, Document } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { Mutation, MutationResult } from '../model/mutation';
 import * as api from '../protos/firestore_proto_api';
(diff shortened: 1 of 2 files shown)


packages/firestore/src/remote/datastore.ts
```diff
@@ -149,6 +150,33 @@ export async function invokeBatchGetDocumentsRpc(
...
+  };
+
+  const response = await datastoreImpl.invokeStreamingRPC<
+    api.RunQueryRequest,
+    api.RunQueryResponse
+  >('RunQuery', params);
+
+  return (
+    response
+      // Filter RunQueryResponses that only contain readTimes.
```
var-const, line 174: Ultranit: consider s/Filter/Omit/ ("filter" has the unfortunate ambiguity in that it might refer to either the part that is left in place or the part that gets removed. It's clear enough from the context, but I still think something like "omit" is slightly better).

packages/firestore/src/remote/datastore.ts
```diff
@@ -149,6 +150,33 @@ export async function invokeBatchGetDocumentsRpc(
...
+  const { structuredQuery, parent } = datastoreImpl.serializer.toQueryTarget(
+    query.toTarget()
+  );
+  const params = {
+    database: datastoreImpl.serializer.encodedDatabaseId,
+    parent,
+    structuredQuery
+  };
+
+  const response = await datastoreImpl.invokeStreamingRPC<
```
var-const, line 167: Optional: it appears that `response` is actually a container, perhaps rename it to `responses`, `streamingResponse`, etc.?

packages/firestore/src/remote/datastore.ts
var-const, line 174: Nit: from reading the PR description, it appears that only the last element may not contain a `document`, whereas the comment makes it seem as if there might be several such elements throughout the container.

schmidt-sebastian: Done

schmidt-sebastian: The PR description reflects the behavior that I have observed. The protocol state though that it is also valid to not provide a document mid-stream as a way to report status.

PR #3026, opened 2020-05-06
Move prepush scripts to precommit

hsubox76:
Moved prepush scripts (prettier, license, api-report) to run on precommit. Since they now run on staged and not committed files, they only `git add` and do not create commits.  Some changes have been made to properly diff staged files against master, or staged files against unstaged files, as appropriate.

Also attempts to fetch `origin/master` before diffing to reduce anomalies caused by stale clones.

### DIFF
--- a/tools/gitHooks/license.js
+++ b/tools/gitHooks/license.js
@@ -115,26 +115,24 @@ async function doLicenseCommit(changedFiles) {
     symbol: '✅'
   });
 
-  const hasDiff = await git.diff();
+  // Diff unstaged (prettier writes) against staged.
+  const stageDiff = await git.diff(['--name-only']);
 
-  if (!hasDiff) {
-    console.log(
-      chalk`\n{red License pass caused no changes.} Skipping commit.\n`
-    );
+  if (!stageDiff) {
+    console.log(chalk`\n{red License pass caused no changes.}\n`);
     return;
+  } else {
+    console.log(
+      `License script modified ${stageDiff.split('\n').length - 1} files.`
+    );
   }
 
-  const gitSpinner = ora(' Creating automated license commit').start();
+  const gitSpinner = ora(' Git staging license text modifications.').start();
   await git.add('.');
 
-  const commit = await git.commit('[AUTOMATED]: License Headers');
-
   gitSpinner.stopAndPersist({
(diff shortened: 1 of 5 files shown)


tools/gitHooks/license.js
```diff
@@ -115,26 +115,24 @@ async function doLicenseCommit(changedFiles) {
     symbol: '✅'
   });
 
-  const hasDiff = await git.diff();
+  // Diff unstaged (prettier writes) against staged.
```
Feiyang1, line 118: license writes?

tools/gitHooks/precommit.js
```diff
@@ -59,7 +82,18 @@ $ git stash pop
     await doLicenseCommit(changedFiles);
 
     // Generate API reports
-    await doApiReportsCommit();
+    await doApiReportsCommit(changedFiles);
```
Feiyang1, line 85: Remove `Commit` from the function names?

tools/gitHooks/precommit.js
```diff
@@ -59,7 +82,18 @@ $ git stash pop
...
+
+    // Diff staged changes against last commit. Don't do an empty commit.
+    const postDiff = await git.diff(['--cached']);
+    if (!postDiff) {
```
Feiyang1, line 89: It seems this condition will never be true since we are committing something from the staging area to begin with.

tools/gitHooks/license.js
hsubox76: Fixed.

tools/gitHooks/precommit.js
hsubox76: Done.

tools/gitHooks/precommit.js
hsubox76: There is a case I ran into while testing where if you only make formatting changes, and they are wrong, and Prettier reformats them back the way they were, there is no longer a diff between your changes and the last commit. I don't think this would ever happen on push but it could happen on a small commit if you did a lot of undos and don't realize the only remaining changes are formatting.

hsubox76:
Removed API report step because it has been removed in https://github.com/firebase/firebase-js-sdk/pull/3030

PR #3033, opened 2020-05-07
Add tooling to verify dependency chain

schmidt-sebastian:
This adds tooling to verify top-level dependencies of tree-shakeable APIs. It is hooked up a a very basic version of a tree-shakeable API.

The goal of this API is to let us verify the dependency chain. One example where I used this is here: https://github.com/firebase/firebase-js-sdk/pull/3024

I suspect this might be helpful for other APIs and as such I added it to `/scripts`.


packages/firestore/exp/test/deps/dependencies.json
```diff
@@ -0,0 +1,15 @@
+{
```
schmidt-sebastian, line 1: Note: This file is generated by the Node CLI tool and then used in the unit test as the source of truth. It is checked in and as such provides a nice diff as we make changes.

packages/firestore/exp/test/deps/dependencies.json
```diff
@@ -0,0 +1,15 @@
...
+            "bar",
+            "foo"
+        ],
+        "sizeInBytes": 67
+    },
+    "bar": {
+        "dependencies": [
+            "bar"
+        ],
+        "sizeInBytes": 39
```
avolkovi, line 13: this is per function? seems like a pain to always update

schmidt-sebastian: It is auto-updated via the script, but the script is not run automatically. I'll send this to @Feiyang1 and we can ponder how we can automate this step.

scripts/exp/extract-deps.ts
```diff
@@ -0,0 +1,61 @@
...
+
+async function buildJson(publicApi: string[], jsFile: string): Promise<string> {
+  const result: { [key: string]: ExportData } = {};
+  for (const exp of publicApi) {
+    result[exp] = await extractDependenciesAndSize(exp, jsFile);
+  }
+  return JSON.stringify(result, null, 4);
+}
+
+const publicApi = extractFunctionsAndClasses(path.resolve(argv.types));
```
Feiyang1, line 58: It doesn't handle reexports, e.g. `export * from 'src/impl'`. We can generate a rollup d.ts file to sidestep the problem, or we need to traverse the AST to get all exports.

scripts/exp/extract-deps.helpers.ts
```diff
@@ -0,0 +1,111 @@
...
+    throw new Error('Failed to parse file: ' + jsFile);
+  }
+
+  const exports: string[] = [];
+  ts.forEachChild(sourceFile, node => {
+    if (ts.isFunctionDeclaration(node)) {
+      exports.push(node.name!.text);
+    } else if (ts.isClassDeclaration(node)) {
+      exports.push(node.name!.text);
+    }
```
Feiyang1, line 108: It'd be interesting to see variable declarations too, e.g. the error message map.
It'd be nice to keep the types of the declarations and show them in the report, but it can be done in a separate PR.

schmidt-sebastian:
@Feiyang1 The quick and dirty fix for your build issue increased the size of the SDK. Instead, I updated `extract-api.ts` to perform a precompile step.

scripts/emulator-testing/emulators/emulator.ts
```diff
@@ -40,7 +40,7 @@ export abstract class Emulator {
 
   download(): Promise<void> {
     return new Promise<void>((resolve, reject) => {
-      tmp.dir((err: Error, dir: string) => {
+      tmp.dir((err: Error|null, dir: string) => {
```
schmidt-sebastian, line 43: This was needed to fix the CI.

scripts/exp/extract-deps.ts
Feiyang1: You are extracting APIs from the d.ts file, not the source file, right? Currently we don't bundle d.ts files into a single file, so there can be reexports from other d.ts files.

As we plan to have multiple entry points in firestore next, it will be easier to not have a single bundled source file so it's easier to share code between different entry points. Why would transitive dependencies be a problem though?

schmidt-sebastian: The public APIs are extracted from a handcrafted index.d.ts. Are we moving away from hand-crafted types? Note that this is not meant to work for all of Firebase, but only for each specific SDK.

FWIW, as more SDKs adopt this, we probably have to greatly expand the feature set of this tooling. I wonder if it makes sense to wait until there is a need, which can then happen together with CI integration?

Feiyang1: I leave that decision to teams. But generally we don't need to handcraft d.ts anymore in the modular world, because we control what gets exported in the source code, and we can generate d.ts files with only public APIs from it automatically. 

I generate index.d.ts from source code and use it for `@firebase/app-exp`. It allows me to keep documentation in the source code as well. I expect most of our SDKs to use automatically generated d.ts files in the modular world.

Sure. that sounds reasonable to me.

schmidt-sebastian: It would be interesting to see if we can hide SDK-public internals from the public typings without splitting up all Firestore classes into abstract Public-API-only base classes and implementation classes. I will keep an eye on this as we rewrite the client.

PR #3037, opened 2020-05-08
Make DeferredPromises Promises

schmidt-sebastian:
Goal: Clean up code.
Risk: Break everyone by essentially shipping our own Promise API.


Feiyang1:
What's the problem with the original way?

schmidt-sebastian:
I was writing a lot of code that accessed the Promise within a DeferredPromise and got a bit fed up with the repeating pattern. I think it is a non-issue though. Closing.

PR #3038, opened 2020-05-08
IndexedDB recovery for waitForPendingWrites()

schmidt-sebastian:
This:
- adda a new `executeWithIndexedDbRecovery` function that aims to centralize the error checking for IndexedDbTransactionErrors. I wasted a lot of time trying to come up with some clean API and clever API and  have concluded that the current approach is the most useable (unless we want to keep current duplication).
- fails the `waitForPendingWrites()` Promise if we cannot access IndexedDb. Note that we cannot retry here, since the retry might let other writes "slip through" which would then also erroneously be awaited.

Addresses #2755


packages/firestore/src/core/event_manager.ts
```diff
@@ -62,47 +63,45 @@ export class EventManager implements SyncEngineListener {
     this.syncEngine.subscribe(this);
   }
 
-  async listen(listener: QueryListener): Promise<void> {
-    const query = listener.query;
-    let firstListen = false;
+  listen(listener: QueryListener): Promise<void> {
+    return executeWithIndexedDbRecovery(
```
wilhuff, line 67: I can't say this pattern is really different from `try`/`catch`, except it's hiding that.

What do you think of the more direct:

```ts
try {
  const query = listener.query;
  //
} catch (e) {
  const msg = `Initialization of query '${query}' failed: ${e}`;	
  logError(LOG_TAG, msg);
  throwIfUnrecoverable(e);
  listener.onError(new FirestoreError(Code.UNAVAILABLE, msg));
}
```

As I see it:
  * It's not really more code than what you have here
  * It uses async/await try/cache directly and looks more natural/synchronous-like
  * On the flip side, there aren't a series of callbacks that you have to remember their meaning.
  * There isn't an extra log message about "Internal operation failed"
  * It still abstracts the handling of the `IndexedDbTransactionError` type

Building on this, you could actually cut the verbosity of each call down by making a `checkRecoverable` that takes the message, logs it, and then throws if unrecoverable or returns the `FirestoreError` if it is?

The catch blocks would look like:
```ts
const wrapped = checkRecoverable(e, `Initialization of query '${query}' failed: ${e}`);
listener.onError(new FirestoreError(Code.UNAVAILABLE, msg));
```

The implementation would look like:
```ts
checkRecoverable(e: Error, message: string): FirestoreError {
  logError(LOG_TAG, msg);
  if (e.name == ....) {
    return new FirestoreError(unavailable, message);
  } else {
    throw e;
  }
}
```

WDYT?

packages/firestore/src/local/lru_garbage_collector.ts
```diff
@@ -267,23 +269,15 @@ export class LruScheduler implements GarbageCollectionScheduler {
...
-              LOG_TAG,
-              'Ignoring IndexedDB error during garbage collection: ',
-              e
-            );
-          } else {
-            await ignoreIfPrimaryLeaseLoss(e);
-          }
-        }
-        await this.scheduleGC(localStore);
+        return executeWithIndexedDbRecovery(
```
wilhuff, line 275: In particular, this change has fairly significantly harmed this method:

  * Why is the extra `.then(() => {})` required?
  * Going back to `then`/`catch` from `async`/`await` is a bummer.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1144,7 +1144,7 @@ describe('IndexedDb: allowTabSynchronization', () => {
       'clientA',
       /* multiClient= */ false,
       async db => {
-        db.injectFailures = true;
+        db.injectFailures = { updateClientMetadataAndTryBecomePrimary: true };
```
schmidt-sebastian, line 1147: I think these tests are now broken in `master` because I merged two conflicting PRs that were both passing all tests on their own.

PR #3039, opened 2020-05-08
IndexedDB Recovery for Limbo documents

schmidt-sebastian:
Addresses #2755

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -483,14 +483,6 @@ export class SyncEngine implements RemoteSyncer {
     const limboResolution = this.activeLimboResolutionsByTarget.get(targetId);
     const limboKey = limboResolution && limboResolution.key;
     if (limboKey) {
-      // Since this query failed, we won't want to manually unlisten to it.
-      // So go ahead and remove it from bookkeeping.
-      this.activeLimboTargetsByKey = this.activeLimboTargetsByKey.remove(
-        limboKey
-      );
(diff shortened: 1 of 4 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -513,7 +505,16 @@ export class SyncEngine implements RemoteSyncer {
         documentUpdates,
         resolvedLimboDocuments
       );
-      return this.applyRemoteEvent(event);
+
+      await this.applyRemoteEvent(event);
+
+      // Since this query failed, we won't want to manually unlisten to it.
```
wilhuff, line 511: It took me a moment to think I understand why this ordering is better. If I'm understanding correctly, the idea is to avoid removing the limbo target until we successfully apply the remote event because if we fail, we'll shut down the listen stream, and then when we resume we'll resend the limbo targets, right?

A comment to this effect could help future readers understand why the watch start/stop lifecycle makes this a reasonable choice.

schmidt-sebastian: Yes. If we can't update the cache, it is  better for the documents to stay in limbo rather than go out of limbo temporarily. I modified the existing comment to say:

```
// Since this query failed, we won't want to manually unlisten to it.
// We only remove it from bookkeeping after we successfully applied the
// RemoteEvent. If `applyRemoteEvent()` throws, we want to re-listen to 
// this query when the RemoteStore restarts the Watch stream, which should 
// re-trigger the target failure.
```

PR #3044, opened 2020-05-10
Make notifyLocalViewChanges idempotent

schmidt-sebastian:
This changes `notifyLocalViewChanges` so that it has not side effects if the transaction fails. The main change is that I removed the indirection of the reference delegate to ask the LocalStore what documents are "pinned". Instead, the Local Store directly tells the reference delegate about every document that is added to a View.

Adresses #2755

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -682,8 +682,7 @@ export class SyncEngine implements RemoteSyncer {
     this.queriesByTarget.delete(targetId);
 
     if (this.isPrimaryClient) {
-      const limboKeys = this.limboDocumentRefs.referencesForId(targetId);
-      this.limboDocumentRefs.removeReferencesForId(targetId);
+      const limboKeys = this.limboDocumentRefs.removeReferencesForId(targetId);
       limboKeys.forEach(limboKey => {
         const isReferenced = this.limboDocumentRefs.containsKey(limboKey);
         if (!isReferenced) {
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -1070,8 +1070,6 @@ function clientMetadataStore(
 
 /** Provides LRU functionality for IndexedDB persistence. */
 export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
-  private inMemoryPins: ReferenceSet | null = null;
-
   readonly garbageCollector: LruGarbageCollector;
 
   constructor(private readonly db: IndexedDbPersistence, params: LruParams) {
@@ -1081,14 +1079,14 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
   getSequenceNumberCount(
     txn: PersistenceTransaction
   ): PersistencePromise<number> {
-    const docCountPromise = this.orphanedDocmentCount(txn);
+    const docCountPromise = this.orphanedDocumentCount(txn);
     const targetCountPromise = this.db.getTargetCache().getTargetCount(txn);
     return targetCountPromise.next(targetCount =>
       docCountPromise.next(docCount => targetCount + docCount)
     );
   }
 
-  private orphanedDocmentCount(
+  private orphanedDocumentCount(
     txn: PersistenceTransaction
   ): PersistencePromise<number> {
     let orphanedCount = 0;
@@ -1113,12 +1111,9 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
     );
   }
 
-  setInMemoryPins(inMemoryPins: ReferenceSet): void {
-    this.inMemoryPins = inMemoryPins;
-  }
-
   addReference(
     txn: PersistenceTransaction,
+    targetId: TargetId,
     key: DocumentKey
   ): PersistencePromise<void> {
     return writeSentinelKey(txn, key);
@@ -1126,6 +1121,7 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
 
   removeReference(
     txn: PersistenceTransaction,
+    targetId: TargetId,
     key: DocumentKey
   ): PersistencePromise<void> {
     return writeSentinelKey(txn, key);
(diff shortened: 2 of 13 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -682,8 +682,7 @@ export class SyncEngine implements RemoteSyncer {
     this.queriesByTarget.delete(targetId);
 
     if (this.isPrimaryClient) {
-      const limboKeys = this.limboDocumentRefs.referencesForId(targetId);
-      this.limboDocumentRefs.removeReferencesForId(targetId);
+      const limboKeys = this.limboDocumentRefs.removeReferencesForId(targetId);
```
schmidt-sebastian, line 685: Combined these lines since `removeReferencesForId` already returns the references for the target.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1141,7 +1137,7 @@ export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
       .removeTargets(txn, upperBound, activeTargetIds);
   }
 
-  removeMutationReference(
+  markPotentiallyOrphaned(
```
schmidt-sebastian, line 1139: Renamed this method as I need to use it to trigger `key` to be considered potentially orphaned. Only then with Eager GC pick it up during its GC run. Other names welcome, including `removeMutationReference`, which reveals less implementation details.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -1070,8 +1070,6 @@ function clientMetadataStore(
 
 /** Provides LRU functionality for IndexedDB persistence. */
 export class IndexedDbLruDelegate implements ReferenceDelegate, LruDelegate {
-  private inMemoryPins: ReferenceSet | null = null;
```
schmidt-sebastian, line 1073: I removed the `inMemoryPins` redirection to the LocalStore. This was possible because LocalStore now directly pins all documents in a View via `addReference`.

packages/firestore/src/local/local_store.ts
```diff
@@ -704,49 +696,56 @@ export class LocalStore {
...
-    for (const viewChange of viewChanges) {
-      const targetId = viewChange.targetId;
-
-      this.localViewReferences.addReferences(viewChange.addedKeys, targetId);
-      this.localViewReferences.removeReferences(
-        viewChange.removedKeys,
-        targetId
-      );
-
-      if (!viewChange.fromCache) {
```
schmidt-sebastian, line 716: Note that this part of the loop is just moved to below the transaction. The goal is that if the transaction fails, this code doesn't get executed.

packages/firestore/src/local/local_store.ts
```diff
@@ -704,49 +696,56 @@ export class LocalStore {
...
+      .runTransaction('notifyLocalViewChanges', 'readwrite', txn => {
         return PersistencePromise.forEach(
           viewChanges,
           (viewChange: LocalViewChanges) => {
             return PersistencePromise.forEach(
-              viewChange.removedKeys,
+              viewChange.addedKeys,
               (key: DocumentKey) =>
-                this.persistence.referenceDelegate.removeReference(txn, key)
+                this.persistence.referenceDelegate.addReference(
```
schmidt-sebastian, line 706: This is the call that makes it possible to remove the local pins.

packages/firestore/src/local/local_store.ts
```diff
@@ -869,26 +868,11 @@ export class LocalStore {
...
-          targetId
-        );
-
         if (!keepPersistedTargetData) {
-          return PersistencePromise.forEach(removed, (key: DocumentKey) =>
-            this.persistence.referenceDelegate.removeReference(txn, key)
-          ).next(() => {
-            this.persistence.referenceDelegate.removeTarget(txn, targetData!);
-          });
+          return this.persistence.referenceDelegate.removeTarget(
```
schmidt-sebastian, line 871: The reference delegate now knows internally what documents are orphaned in the view.

packages/firestore/src/local/memory_target_cache.ts
```diff
@@ -191,14 +191,7 @@ export class MemoryTargetCache implements TargetCache {
     targetId: TargetId
   ): PersistencePromise<void> {
     this.references.addReferences(keys, targetId);
-    const referenceDelegate = this.persistence.referenceDelegate;
-    const promises: Array<PersistencePromise<void>> = [];
-    if (referenceDelegate) {
-      keys.forEach(key => {
-        promises.push(referenceDelegate.addReference(txn, key));
```
schmidt-sebastian, line 198: We could keep this call... but all this does now is mark a document as non-orphaned if it was previously orphaned.  The keys now also added to the target as part of `notifyLocalViewChanges `.

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -184,7 +184,9 @@ export interface MemoryReferenceDelegate extends ReferenceDelegate {
 }
 
 export class MemoryEagerDelegate implements MemoryReferenceDelegate {
-  private inMemoryPins: ReferenceSet | null = null;
+  /** Manages all documents that are active in Query views. */
+  private localViewReferences: ReferenceSet = new ReferenceSet();
+  /** The list of documents that are potentially GCed after each transaction. */
   private _orphanedDocuments: Set<DocumentKey> | null = null;
```
var-const, line 190: Question: why is this variable prefixed with an underscore, while `localViewReferences` isn't?

packages/firestore/src/local/memory_persistence.ts
```diff
@@ -184,7 +184,9 @@ export interface MemoryReferenceDelegate extends ReferenceDelegate {
 }
 
 export class MemoryEagerDelegate implements MemoryReferenceDelegate {
-  private inMemoryPins: ReferenceSet | null = null;
+  /** Manages all documents that are active in Query views. */
```
var-const, line 187: Nit: I always find that `manage` is one of the words that can usually be replaced with something more specific. Even `keeps track of`, to me, seems more specific.

schmidt-sebastian: I hope this is not a general stance on management :) Updated to use "track".

packages/firestore/src/local/memory_persistence.ts
schmidt-sebastian: In general, we don't use underscore prefixes for classes that are not part of the private API. This is an exception as `orphanedDocuments` has a getter that verifies that is is set correctly. `orphanedDocuments` is cleared between each transaction run.

Issue #3045, opened 2020-05-10
Firestore - WriteBatch.Set ignoring {merge: true} when using FieldValue.increment


olivierkrener:
### [REQUIRED] Describe your environment

  * Operating System version: _____ Windows 10
  * Browser version: _____ Chrome (latest Version 81.0.4044.129) but this seems irrelevant as it is not working in the emulator either 
  * Firebase SDK version: _____ firebase@7.14.3, @firebase/testing@0.19.3
  * Firebase Product: _____ Firestore, Testing (auth, database, storage, etc)

### [REQUIRED] Describe the problem

Updating a document with FieldValue.increment using a batch via the set function and the {merge: true} option fails. Looking at the logs it seems it is ignoring the option and trying to overwrite the document.

`batch.update(somepath, {counter: firebase.firestore.FieldValue.increment(1)})`
=> works as expected
`batch.set(somepath, {counter: firebase.firestore.FieldValue.increment(1), {merge: true})`
=> will reset counter to 1 whatever its previous value

https://firebase.google.com/docs/reference/js/firebase.firestore.FieldValue#increment
https://firebase.google.com/docs/firestore/manage-data/add-data#set_a_document
https://firebase.google.com/docs/reference/js/firebase.firestore.SetOptions

This is neither working in the emulator nor in Cloud Firestore.

Note that with the previous minor version firebase@7.14.2, @firebase/testing@0.19.2 - it is all working fine

amad9720:
I have exactly the same issue... I noticed it yesterday.
An explanation of this behavior will be great...
Thanks

schmidt-sebastian:
This is indeed a regression. We will address this shortly.

PR #3048, opened 2020-05-11
Fix regression for merge with increment

schmidt-sebastian:
This reverts a behavior change in https://github.com/firebase/firebase-js-sdk/pull/3001

Before this PR, the parse contexts we used to parse ArrayTransform and Increment values was detached from the original parse context that was used to parse the update. Thus, setting the field path for the value that was being incremented did not end up modifying the context that was used to generate the Proto. The result of this was `{ updateMask: [] }`. By removing the copy, we ended up with `{ updateMask: ['sum'] }`, which ended up rewriting the value before applying the increment.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3045

### DIFF
--- a/packages/firestore/src/api/user_data_reader.ts
+++ b/packages/firestore/src/api/user_data_reader.ts
@@ -138,12 +138,12 @@ interface ContextSettings {
    * A path within the object being parsed. This could be an empty path (in
    * which case the context represents the root of the data being parsed), or a
    * nonempty path (indicating the context represents a nested location within
-   * the data). Defaults to null.
+   * the data).
    */
   readonly path?: FieldPath;
   /**
    * Whether or not this context corresponds to an element of an array.
-   * Defaults to false.
+   * If not set, elements are treated as if they were outside of arrays.
    */
   readonly arrayElement?: boolean;
 }
@@ -219,7 +219,7 @@ export class ParseContext {
 
   childContextForArray(index: number): ParseContext {
     // TODO(b/34871131): We don't support array paths right now; so make path
-    // null.
+    // undefined.
     return this.contextWith({ path: undefined, arrayElement: true });
   }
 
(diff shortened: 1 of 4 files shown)


packages/firestore/src/api/user_data_reader.ts
```diff
@@ -138,11 +138,14 @@ interface ContextSettings {
    * A path within the object being parsed. This could be an empty path (in
    * which case the context represents the root of the data being parsed), or a
    * nonempty path (indicating the context represents a nested location within
-   * the data).
+   * the data). Defaults to null.
```
dconeybe, line 141: Doesn't it default to "undefined"? Consider just omitting this statement because it is seems like it is just explaining how the "?" property modifier works.

packages/firestore/src/api/user_data_reader.ts
```diff
@@ -138,11 +138,14 @@ interface ContextSettings {
...
    */
-  readonly path: FieldPath | null;
-  /** Whether or not this context corresponds to an element of an array. */
-  readonly arrayElement: boolean;
+  readonly path?: FieldPath;
+  /**
+   * Whether or not this context corresponds to an element of an array.
+   * Defaults to false.
```
dconeybe, line 146: Consider re-wording this as something like "if undefined, it will be treated as false" because the way that it is worded suggests to me that it gets assigned a default value of false somehow, which I do not think is accurate.

packages/firestore/src/api/user_data_reader.ts
schmidt-sebastian: That's true. I dropped it.

packages/firestore/src/api/user_data_reader.ts
schmidt-sebastian: Changed to `If not set, elements are treated as if they were outside of arrays.`

packages/firestore/src/api/user_data_reader.ts
```diff
@@ -217,12 +220,12 @@ export class ParseContext {
   childContextForArray(index: number): ParseContext {
     // TODO(b/34871131): We don't support array paths right now; so make path
     // null.
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, good catch.

packages/firestore/src/api/user_data_reader.ts
dconeybe: Love it.

PR #3049, opened 2020-05-11
Transition to primary only if IndexedDB ops succeed

schmidt-sebastian:
Goal: Don't change local state before all IndexedDB transactions succeed.

Addresses #2755

### DIFF
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -1032,9 +1032,6 @@ export class MultiTabSyncEngine extends SyncEngine
 
   async applyPrimaryState(isPrimary: boolean): Promise<void> {
     if (isPrimary === true && this.isPrimary !== true) {
-      this.isPrimary = true;
-      await this.remoteStore.applyPrimaryState(true);
-
       // Secondary tabs only maintain Views for their local listeners and the
       // Views internal state may not be 100% populated (in particular
       // secondary tabs don't track syncedDocuments, the set of documents the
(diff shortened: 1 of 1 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -1032,9 +1032,6 @@ export class MultiTabSyncEngine extends SyncEngine
 
   async applyPrimaryState(isPrimary: boolean): Promise<void> {
     if (isPrimary === true && this.isPrimary !== true) {
-      this.isPrimary = true;
```
rafikhan, line 1035: I notice that sync_engine.ts uses a mixture of reading from `this.isPrimary` and `this.isPrimaryClient`. Shouldn't all reads be consolidated to the latter? If so, we should fix this in a separate PR.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1043,14 +1040,15 @@ export class MultiTabSyncEngine extends SyncEngine
       // match the state on disk.
       const activeTargets = this.sharedClientState.getAllActiveQueryTargets();
       const activeQueries = await this.synchronizeQueryViewsAndRaiseSnapshots(
-        activeTargets.toArray()
+        activeTargets.toArray(),
+        /*transitionToPrimary=*/ true
       );
+      this.isPrimary = true;
+      await this.remoteStore.applyPrimaryState(true);
```
rafikhan, line 1047: I want to verify something. These lines of code were moved down b/c the bit should only be set if a prior IndexDB operation succeeded correct?

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: `this.isPrimaryClient` is the getter in the API. `this.isPrimary` is the underlying member variable. A more common pattern is to use `this.isPrimary` and `this._isPrimary`. I will update this is in a follow-up.

packages/firestore/src/core/sync_engine.ts
schmidt-sebastian: Yes, exactly.

PR #3050, opened 2020-05-11
RTDB accepts multi region URL

fredzqm:
Allow SDK to parse url like `my-db.euw1.firebasedatabase.app`


packages/database/src/core/util/libs/parser.ts
```diff
@@ -174,17 +174,18 @@ export const parseDatabaseURL = function(
...
+    } else {
+      domain = host.substring(dotInd+1);
       // Normalize namespaces to lowercase to share storage / connection.
-      domain = parts[1];
-      subdomain = parts[0].toLowerCase();
-      // We interpret the subdomain of a 3 component URL as the namespace name.
+      subdomain = hostFirstPart.toLowerCase();
       namespace = subdomain;
-    } else if (parts.length === 2) {
-      domain = parts[0];
```
fredzqm, line 185: This case would be changed if we generalized it.
`foo.bar` ==> `namespace = foo, domain = bar`

packages/database/src/core/util/libs/parser.ts
```diff
@@ -72,7 +72,7 @@ export const parseRepoInfo = function(
   const parsedUrl = parseDatabaseURL(dataURL),
     namespace = parsedUrl.namespace;
 
-  if (parsedUrl.domain === 'firebase') {
+  if (parsedUrl.domain === 'firebase.com') {
```
schmidt-sebastian, line 75: Isn't this a breaking change for users that use `firebase.com?ns=foo`?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -174,17 +174,17 @@ export const parseDatabaseURL = function(
       colonInd = dataURL.length;
     }
 
-    const parts = host.split('.');
-    if (parts.length === 3) {
+    const hostWithoutPort = host.slice(0, colonInd);
```
schmidt-sebastian, line 177: From the `slice` docs:

"A negative index can be used, indicating an offset from the end of the sequence. slice(2,-1) extracts the third element through the second-to-last element in the sequence."

So, in this case, if no port is specified it seems like we would drop the last character, no?

packages/database/src/core/util/libs/parser.ts
```diff
@@ -174,17 +174,17 @@ export const parseDatabaseURL = function(
...
+    if (hostWithoutPort.toLowerCase() === 'localhost') {
+      domain = 'localhost';
+    } else if (hostWithoutPort.split('.').length <= 2) {
+      domain = hostWithoutPort; // Domain is at least 2 parts.
+    } else {
+      const dotInd = host.indexOf('.');
+      domain = host.substring(dotInd + 1);
+      subdomain = host.substring(0, dotInd).toLowerCase();
```
schmidt-sebastian, line 185: Nit: Flip two lines above to parse string consecutively.

packages/database/test/database.test.ts
```diff
@@ -62,6 +62,15 @@ describe('Database Tests', () => {
     expect(db.ref().toString()).to.equal('https://foo.bar.com/');
   });
 
+  it('Can get database with multi-region URL', () => {
```
schmidt-sebastian, line 65: Do you want to add a test with a non-lowercase namespace?

packages/database/src/core/util/libs/parser.ts
schmidt-sebastian: Never mind. This is only breaking for `firebase/ns?foo` which shouldn't affect anyone.

fredzqm: I believe in the past. `*.firebase.*` will hit this.

My guess is that RTDB was served in `firebase.com` instead of `firebaseio.com` in the past.

Added a test for this.

packages/database/src/core/util/libs/parser.ts
fredzqm: If no port is specified, `colonInd = host.length` above, so `host.slice(0, host.length) ==> host`

packages/database/test/database.test.ts
fredzqm: Sounds like a good idea.

packages/database/src/core/util/libs/parser.ts
fredzqm: Done

fredzqm:
Done

PR #3052, opened 2020-05-12
Allow failing of individual transactions

schmidt-sebastian:
Disclaimer: This is pretty ugly... feedback appreciated :)

The idea of the PR is to expand the new "failDatabase" spec test feature to allow the tests to specify which transaction action to fail. This allows individualized failures for operations that spawn multiple actions. Currently, we only test scenarios where the first transaction fails. This will allows us to test what happens if the first transaction succeeds and the second one fails.

### DIFF
--- a/packages/firestore/test/unit/specs/spec_test_components.ts
+++ b/packages/firestore/test/unit/specs/spec_test_components.ts
@@ -66,17 +66,7 @@ export class MockMemoryPersistence extends MemoryPersistence {
       transaction: PersistenceTransaction
     ) => PersistencePromise<T>
   ): Promise<T> {
-    if (this.injectFailures) {
-      if (this.injectFailures[action as PersistenceAction] === undefined) {
-        throw fail('Failure mode not specified for action: ' + action);
-      }
-      if (this.injectFailures[action as PersistenceAction]) {
-        return Promise.reject(
-          new IndexedDbTransactionError(new Error('Simulated retryable error'))
-        );
-      }
-    }
-
+    failTransactionIfNeeded(this.injectFailures, action);
     return super.runTransaction(action, mode, transactionOperation);
   }
 }
@@ -95,17 +85,28 @@ export class MockIndexedDbPersistence extends IndexedDbPersistence {
       transaction: PersistenceTransaction
     ) => PersistencePromise<T>
   ): Promise<T> {
-    if (this.injectFailures) {
-      if (this.injectFailures[action as PersistenceAction] === undefined) {
-        throw fail('Failure mode not specified for action: ' + action);
-      } else if (this.injectFailures[action as PersistenceAction]) {
-        return Promise.reject(
-          new IndexedDbTransactionError(new Error('Simulated retryable error'))
-        );
-      }
-    }
-
+    failTransactionIfNeeded(this.injectFailures, action);
     return super.runTransaction(action, mode, transactionOperation);
+  }
+}
+
+/**
+ * Shared failure handler between MockIndexedDbPersistence and
+ * MockMemoryPersistence that can inject transaction failures.
+ */
+function failTransactionIfNeeded(
+  config: SpecDatabaseFailures | undefined,
+  action: string
+): void {
+  if (config) {
+    const shouldFail = config[action as PersistenceAction];
+    if (shouldFail === undefined) {
+      throw fail('Failure mode not specified for action: ' + action);
+    } else if (shouldFail) {
+      throw new IndexedDbTransactionError(
+        new Error('Simulated retryable error' + action)
+      );
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/specs/spec_test_components.ts
```diff
@@ -66,12 +67,17 @@ export class MockMemoryPersistence extends MemoryPersistence {
     ) => PersistencePromise<T>
   ): Promise<T> {
     if (this.injectFailures) {
-      return Promise.reject(
-        new IndexedDbTransactionError(new Error('Simulated retryable error'))
-      );
-    } else {
-      return super.runTransaction(action, mode, transactionOperation);
+      if (this.injectFailures[action as PersistenceAction] === undefined) {
```
dconeybe, line 70: Consider saving the result of `this.injectFailures[action as PersistenceAction]` to a local variable instead of accessing it twice via this long dictionary access syntax.

packages/firestore/test/unit/specs/spec_test_components.ts
```diff
@@ -66,12 +67,17 @@ export class MockMemoryPersistence extends MemoryPersistence {
...
+        throw fail('Failure mode not specified for action: ' + action);
+      }
+      if (this.injectFailures[action as PersistenceAction]) {
+        return Promise.reject(
+          new IndexedDbTransactionError(new Error('Simulated retryable error'))
```
dconeybe, line 75: Consider including the value of `action` in the error message. This may assist with debugging to figure out which action caused the exception to be thrown.

packages/firestore/test/unit/specs/spec_test_components.ts
```diff
@@ -90,12 +96,16 @@ export class MockIndexedDbPersistence extends IndexedDbPersistence {
     ) => PersistencePromise<T>
   ): Promise<T> {
     if (this.injectFailures) {
-      return Promise.reject(
-        new IndexedDbTransactionError(new Error('Simulated retryable error'))
-      );
-    } else {
-      return super.runTransaction(action, mode, transactionOperation);
+      if (this.injectFailures[action as PersistenceAction] === undefined) {
```
dconeybe, line 99: If this block of code is an identical copy/paste from above, then consider moving it to a helper function.

packages/firestore/test/unit/specs/spec_test_components.ts
dconeybe, line 70: It looks like the Android and iOS spec test framework will need to be updated to handle this change in the generated JSON. Are you planning to do that as well?

schmidt-sebastian: Done.

schmidt-sebastian: Yes, we need to make those changes there too. For now, the "recovery" tests are marked `no-ios` and `no-android`. We won't need to update the runner until we remove these tags and implement LevelDB/SQLite recovery.

packages/firestore/test/unit/specs/spec_test_components.ts
schmidt-sebastian: Good idea. Future me will thank you as I am debugging test assertions.

packages/firestore/test/unit/specs/spec_test_components.ts
schmidt-sebastian: It's not identical because I only updated one of the two blocks to use "else if". But it should be identical... Moved it to a helper function.

packages/firestore/test/unit/specs/spec_test_components.ts
```diff
@@ -95,20 +85,31 @@ export class MockIndexedDbPersistence extends IndexedDbPersistence {
...
+  config: SpecDatabaseFailures | undefined,
+  action: string
+): void {
+  if (config) {
+    const shouldFail = config[action as PersistenceAction];
+    if (shouldFail === undefined) {
+      throw fail('Failure mode not specified for action: ' + action);
+    } else if (shouldFail) {
+      throw new IndexedDbTransactionError(
+        new Error('Simulated retryable error' + action)
```
dconeybe, line 107: nit: Add a space or colon after "error" so that the action text is not mashed into the word "error".

packages/firestore/test/unit/specs/spec_test_components.ts
dconeybe: Ahh you're right. I missed the `no-ios` and `no-android` tags... even though I was looking for them 🤔

packages/firestore/test/unit/specs/spec_test_components.ts
schmidt-sebastian: Not a nit but a crucial typo fix :) Done

PR #3064, opened 2020-05-14
Warn when stream is closing due to error, not debug.

wu-hui:
b/155897356


packages/firestore/src/api/database.ts
```diff
@@ -612,37 +618,14 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
-    switch (getLogLevel()) {
-      case LogLevel.DEBUG:
-        return 'debug';
-      case LogLevel.SILENT:
-        return 'silent';
-      default:
-        // The default log level is error
-        return 'error';
-    }
+  static get logLevel(): LogLevel {
```
schmidt-sebastian, line 621: Can you change this back to `firestore.LogLevel`, remove the use of the `LogLevel` type (we should only use LogLevelString) and change the typings for `firestore.LogLevel` to match the log types in Util?

packages/firestore/src/util/log.ts
```diff
@@ -15,11 +15,11 @@
  * limitations under the License.
  */
 
-import { Logger, LogLevel } from '@firebase/logger';
+import { Logger, LogLevel, LogLevelString } from '@firebase/logger';
 import { SDK_VERSION } from '../core/version';
 import { PlatformSupport } from '../platform/platform';
 
-export { LogLevel };
+export { LogLevel, LogLevelString };
```
schmidt-sebastian, line 22: You probably don't need to export these anymore. Instead, you can rely on the fact that our typings now match the types in `@firebase/logger`.

packages/firestore/src/api/database.ts
```diff
@@ -612,37 +618,14 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
+    return getLogLevel();
   }
 
-  static setLogLevel(level: firestore.LogLevel): void {
+  static setLogLevel(level: LogLevelString): void {
     validateExactNumberOfArgs('Firestore.setLogLevel', arguments, 1);
     validateArgType('Firestore.setLogLevel', 'non-empty string', 1, level);
```
schmidt-sebastian, line 627: Instead of this, you may want to call https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/firestore/src/util/input_validation.ts;l=303?q=validate%20enum&ss=firebase-sdk%2Ffirebase-js-sdk

If you don't validate the enum constant here, `setLogLevel` sets the log level to undefined.

packages/firestore-types/index.d.ts
```diff
@@ -37,7 +37,7 @@ export interface PersistenceSettings {
   experimentalTabSynchronization?: boolean;
 }
 
-export type LogLevel = 'debug' | 'error' | 'silent';
+export type LogLevel = 'debug' | 'error' | 'silent' | 'warn';
```
schmidt-sebastian, line 40: Do we not want to expose all levels? You are missing `verbose` and `info`.

packages/firestore/src/util/log.ts
```diff
@@ -28,8 +28,8 @@ export function getLogLevel(): LogLevel {
   return logClient.logLevel;
 }
 
-export function setLogLevel(newLevel: LogLevel): void {
-  logClient.logLevel = newLevel;
+export function setLogLevel(newLevel: LogLevelString | LogLevel): void {
```
schmidt-sebastian, line 31: You can probably drop `LogLevel` as an input type.

packages/firestore-types/index.d.ts
wu-hui: Done.

packages/firestore/src/util/log.ts
wu-hui: async_queue.test.ts calls `getLogLevel()` before changing it, then set the saved log level back when test is done. https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/unit/util/async_queue.test.ts#L84

It's only testing code, but i think it does no harm to keep it anyways?

packages/firestore/src/api/database.ts
```diff
@@ -626,23 +634,13 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
...
-        break;
-      default:
-        throw new FirestoreError(
-          Code.INVALID_ARGUMENT,
-          'Invalid log level: ' + level
-        );
-    }
+    validateStringEnum(
+      'setLogLevel',
+      ['debug', 'error', 'silent', 'warn'],
```
schmidt-sebastian, line 639: Please add 'verbose' and 'info' here as well.

packages/firestore/src/util/log.ts
schmidt-sebastian: SGTM.

This could also be fixed by returning the LogLevelString from `getLogLevel()` and calling `setLogLevel('silent')` instead of `setLogLevel(LogLevel.SILENT)`. We can leave as is though.

packages/firestore/src/api/database.ts

packages/firestore-types/index.d.ts
```diff
@@ -37,7 +37,13 @@ export interface PersistenceSettings {
...
 }
 
-export type LogLevel = 'debug' | 'error' | 'silent';
+export type LogLevel =
+  | 'debug'
+  | 'error'
+  | 'silent'
+  | 'warn'
+  | 'info'
+  | 'verbose';
```
Feiyang1, line 46: Are you planning to make it public? If so, you have to update `packages/firebase/index.d.ts`

Feiyang1: @wu-hui @schmidt-sebastian

schmidt-sebastian: Pinged on chat

PR #3065, opened 2020-05-14
Fix Emulator URL in database CI

schmidt-sebastian:
This should fix the CI failures in https://github.com/firebase/firebase-js-sdk/pull/3038/checks?check_run_id=675464771

cc @fredzqm - Just an FYI, it looks like we are no longer parsing `http://localhost:8080?ns=foo` URLs correctly, which should be fine since they are not valid. For reference: https://stackoverflow.com/questions/1617058/ok-to-skip-slash-before-query-string

### DIFF
--- a/packages/testing/src/api/index.ts
+++ b/packages/testing/src/api/index.ts
@@ -120,7 +120,9 @@ function initializeApp(
 ): firebase.app.App {
   let appOptions: { [key: string]: string } = {};
   if (databaseName) {
-    appOptions['databaseURL'] = `http://${DATABASE_ADDRESS}?ns=${databaseName}`;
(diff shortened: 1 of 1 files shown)


yuchenshi:
> it looks like we are no longer parsing `http://localhost:8080?ns=foo` URLs correctly, which should be fine since they are not valid. For reference: https://stackoverflow.com/questions/1617058/ok-to-skip-slash-before-query-string

Not true, see an updated answer at https://stackoverflow.com/a/42193734. I think we should instead fix our SDKs to allow omitting slashes, given the updated spec.

schmidt-sebastian:
I will revisit this.

PR #3066, opened 2020-05-14
Fix Firestore build

schmidt-sebastian:
This is already part of https://github.com/firebase/firebase-js-sdk/pull/3038, but it looks like it broke CI for everyone so I pulled in out into a separate PR.

This broke because https://github.com/firebase/firebase-js-sdk/pull/3052 and https://github.com/firebase/firebase-js-sdk/pull/3020 were both passing CI on their own but not combined.

### DIFF
--- a/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
+++ b/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
@@ -1144,7 +1144,7 @@ describe('IndexedDb: allowTabSynchronization', () => {
       'clientA',
       /* multiClient= */ false,
       async db => {
-        db.injectFailures = true;
+        db.injectFailures = { updateClientMetadataAndTryBecomePrimary: true };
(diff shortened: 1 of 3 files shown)


packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1144,7 +1144,7 @@ describe('IndexedDb: allowTabSynchronization', () => {
       'clientA',
       /* multiClient= */ false,
       async db => {
-        db.injectFailures = true;
+        db.injectFailures = { updateClientMetadataAndTryBecomePrimary: true };
```
wilhuff, line 1147: Are you planning on making the other clients have the same transaction names? (Obviously they won't have the multitab related transactions, but how will this failure mechanism work anywhere else?)

schmidt-sebastian: Yes. I am also planning to clean up our names before this happens.

PR #3068, opened 2020-05-15
Ensure run_changed errors on packages without test script

hsubox76:
For a more safety, we would like `run_changed.js` (the script behind the Test Changed Packages workflow) to fail if it finds a package without a "test" or "test:ci" script (whichever command it is running).

A maintainer can put placeholder stubs for those scripts if the package does not have tests to run, but this will prevent accidentally forgetting to add a test script.

Added stubs for these scripts to packages that did not have them.

### DIFF
--- a/scripts/run_changed.js
+++ b/scripts/run_changed.js
@@ -153,7 +153,7 @@ async function runTests(pathList) {
         stdio: 'inherit'
       });
     } catch (e) {
-      throw new Error(`Error running tests in ${testPath}.`);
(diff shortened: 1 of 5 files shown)


scripts/run_changed.js
```diff
@@ -25,6 +25,7 @@ const git = simpleGit(root);
 
 // use test:ci command in CI
 const testCommand = !!process.env.CI ? 'test:ci' : 'test';
+// const testCommand = 'print';
```
Feiyang1, line 28: remove

PR #3073, opened 2020-05-15
Add bundle.proto d.ts and introduce IDB object stores for bundles. 

wu-hui:


packages/firestore/src/protos/firestore/bundle.proto
```diff
@@ -0,0 +1,111 @@
...
+
+option csharp_namespace = "Firestore.Proto";
+option go_package = "google.golang.org/genproto/firestore/proto;firestore";
+option java_multiple_files = true;
+option java_outer_classname = "BundleProto";
+option java_package = "com.google.firestore.proto";
+option objc_class_prefix = "FSTPB";
+option php_namespace = "Firestore\\Proto";
+
+// Encodes a query saved in the bundle.
```
schmidt-sebastian, line 36: Nit: s/Encodes/Describes

Optional.

packages/firestore/src/protos/firestore/bundle.proto
```diff
@@ -0,0 +1,111 @@
...
+
+  // The schema version of the bundle.
+  uint32 version = 3;
+}
+
+// A Firestore bundle is a length-prefixed stream of JSON representations of 
+// `BundleElement`.
+// Only one `BundleMetadata` is expected, and it should be the first element.
+// The named queries follow after `metadata`. Every `document_metadata` is
+// immediately followed by a `document`.
```
schmidt-sebastian, line 100: The last part is not true anymore, is it? We skip document data for non-existing documents.

packages/firestore/src/protos/firestore/bundle.proto
wu-hui: Done.

packages/firestore/src/protos/firestore/bundle.proto
wu-hui: Updated.

PR #3075, opened 2020-05-15
Update internal types to work with external types

samhorlbeck:


packages-exp/auth-exp/src/core/user/reload.ts
```diff
@@ -52,19 +54,20 @@ export async function _reloadWithoutSaving(user: User): Promise<void> {
...
-  await _reloadWithoutSaving(user);
+export async function reload(user: externs.User): Promise<void> {
+  const userInternal: User = user as User;
+  await _reloadWithoutSaving(userInternal);
 
   // Even though the current user hasn't changed, update
   // current user will trigger a persistence update w/ the
   // new info.
-  return user.auth.updateCurrentUser(user);
+  return userInternal.auth.updateCurrentUser(userInternal);
```
avolkovi, line 64: maybe call the parameter as externalUser instead? that's more consistent with your naming

PR #3076, opened 2020-05-15
Make instance logLevel setter able to process string input

hsubox76:
When Firestore and other users of Logger set log level on their instance only, this change allows them to use string input in addition to LogLevel enums. Firestore has a public interface that takes string input (just as the top-level `setLogLevel()` does) and needs to use this input to set logLevel on the Firestore logger instance only.

### DIFF
--- a/packages/logger/src/logger.ts
+++ b/packages/logger/src/logger.ts
@@ -141,14 +141,15 @@ export class Logger {
    * The log level of the given Logger instance.
    */
   private _logLevel = defaultLogLevel;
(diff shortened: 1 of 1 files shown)


hsubox76:
Needed to make an additional setter method because built-in getters and setters are required to be the same type.

Feiyang1:
Do you have the link to the Firestore PR?

hsubox76:
Firestore PR: https://github.com/firebase/firebase-js-sdk/pull/3064/files

PR #3077, opened 2020-05-15
Add ignoreUndefinedProperties

schmidt-sebastian:
Port of https://github.com/googleapis/nodejs-firestore/pull/1062

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7746,10 +7746,9 @@ declare namespace firebase.firestore {
 
     /**
      * Whether to skip nested properties that are set to `undefined` during
-     * object serialization. If set to `true`, these properties will be skipped
-     * and are not be written to Firestore. If it is set `false` or omitted,
-     * the SDK will throw an exception when it encounters properties of type
-     * `undefined`.
+     * object serialization. If set to `true`, these properties are skipped
+     * and not written to Firestore. If set `false` or omitted, the SDK throws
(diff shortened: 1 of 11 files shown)


packages/firebase/index.d.ts
```diff
@@ -7743,6 +7743,15 @@ declare namespace firebase.firestore {
      * @webonly
      */
     experimentalForceLongPolling?: boolean;
+
+    /**
+     * Whether to skip nested properties that are set to `undefined` during
+     * object serialization. If set to `true`, these properties will be skipped
+     * and are not be written to Firestore. If the setting is set `false` or
```
wu-hui, line 7750: s/the setting is set/it is set to

schmidt-sebastian: Done

packages/firebase/index.d.ts
```diff
@@ -7743,6 +7743,15 @@ declare namespace firebase.firestore {
      * @webonly
      */
     experimentalForceLongPolling?: boolean;
+
+    /**
+     * Whether to skip nested properties that are set to `undefined` during
+     * object serialization. If set to `true`, these properties will be skipped
```
egilmorez, line 7749: Very much a nit, but we prefer present over future when possible.  Here it cold be "properties are skipped . . . . the SDK throws . . ."

schmidt-sebastian: Updated. Thanks for the review!

PR #3082, opened 2020-05-18
Public typings only swapped on build:exp:release

hsubox76:
Add/change scripts:

Root:
- `build:release` - For prod release builds, excludes `packages-exp`
- `buld:release:exp` - For exp releases, last step is to swap in public typings for `app-exp`.

These two scripts will only be run by our release scripts. Regular builds will no longer touch the typings field.

In app-exp:
- `build:release` - Modified to not swap public typings. Should be done at end of all exp package builds.
- `typings:public` - Swaps package.json field to point at public typings. Called by root build:release:exp.
- `typings:internal` - Swaps package.json field to point at internal typings. Scripts do not use this but it may be a useful shortcut for devs or future scripts.


package.json
```diff
@@ -24,6 +24,8 @@
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
     "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire prepare",
     "build:exp": "lerna run --scope @firebase/*-exp --scope firebase-exp build",
+    "build:release": "lerna run --scope @firebase/* --scope firebase --ignore @firebase/*-exp --scope firebase-exp build",
+    "build:exp:release": "yarn build:exp && yarn --cwd packages-exp/app-exp typings:public",
```
Feiyang1, line 28: I'd like to keep the swapping typings step in `@firebase/app-exp` because I want this step to run when I run `build:release` in `@firebase/app-exp`.
How about `"build:exp:release": "lerna run --scope @firebase/*-exp --scope firebase-exp prepare"`?

packages-exp/app-exp/use_public_typings.js
```diff
@@ -16,16 +17,23 @@
  */
 
 const { writeFileSync } = require('fs');
+const { argv } = require('yargs');
```
Feiyang1, line 20: Since the script takes an argument now, rename it to `use_typings.js`?

package.json
```diff
@@ -24,6 +24,8 @@
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
     "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire prepare",
```
Feiyang1, line 25: replace `prepare` with `build`, so it doesn't run swapping public typings in local builds.

package.json
hsubox76: Ok, app-exp `prepare`, which is only run by the root script, will not swap typings, but app-exp `build:release` which is only run locally, will.

packages-exp/app-exp/use_public_typings.js
hsubox76: Done.

package.json
hsubox76: It won't (see comment above) but it is better to run the expected command. Changed.

packages-exp/app-exp/package.json
```diff
@@ -15,19 +15,21 @@
     "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "lint:fix": "eslint --fix -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "build": "rollup -c && yarn api-report",
-    "build:release": "rollup -c rollup.config.release.js && node ./use_public_typings.js",
+    "build:release": "rollup -c rollup.config.release.js && yarn typings:public",
```
Feiyang1, line 18: Can you please add `yarn api-report` as the second step? It generates the rollup d.ts files.

hsubox76: Done!

PR #3083, opened 2020-05-18
Add test for getHighestListenSequenceNumber() recovery

schmidt-sebastian:
This adds a test that verify that we can bubble up the exception from `getHighestListenSequenceNumber` to `enablePersistence()`. The client can then proceed with MemoryPersistence: https://github.com/firebase/firebase-js-sdk/blob/3de5763b76d19314db1355072b9e0d9ccea70ccc/packages/firestore/src/core/firestore_client.ts#L325

Addresses #2755

### DIFF
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -45,7 +45,6 @@ import {
   DbPrimaryClient,
   DbPrimaryClientKey,
   DbTargetDocument,
-  DbTargetGlobal,
   SCHEMA_VERSION,
   SchemaConverter
 } from './indexeddb_schema';
@@ -306,10 +305,13 @@ export class IndexedDbPersistence implements Persistence {
 
         this.scheduleClientMetadataAndPrimaryLeaseRefreshes();
 
-        return this.simpleDb.runTransaction(
+        return this.runTransaction(
+          'getHighestListenSequenceNumber',
           'readonly',
-          [DbTargetGlobal.store],
-          txn => getHighestListenSequenceNumber(txn)
+          txn =>
+            getHighestListenSequenceNumber(
+              (txn as IndexedDbTransaction).simpleDbTransaction
+            )
         );
       })
       .then(highestListenSequenceNumber => {
--- a/packages/firestore/src/local/indexeddb_persistence.ts
+++ b/packages/firestore/src/local/indexeddb_persistence.ts
@@ -50,7 +50,6 @@ import {
 } from './indexeddb_schema';
 import {
   documentTargetStore,
-  getHighestListenSequenceNumber,
   IndexedDbTargetCache
 } from './indexeddb_target_cache';
 import { LocalSerializer } from './local_serializer';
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -306,10 +305,13 @@ export class IndexedDbPersistence implements Persistence {
...
         this.scheduleClientMetadataAndPrimaryLeaseRefreshes();
 
-        return this.simpleDb.runTransaction(
+        return this.runTransaction(
+          'getHighestListenSequenceNumber',
           'readonly',
-          [DbTargetGlobal.store],
-          txn => getHighestListenSequenceNumber(txn)
+          txn =>
+            getHighestListenSequenceNumber(
```
wilhuff, line 312: There's a `getHighestSequenceNumber` that does this translation for you, though I'm really dubious that we should really have names that differ in this way for the difference in transaction types.

It seems like there's a change warranted where we minimize the use of SimpleDbTransaction?

schmidt-sebastian: The original version of this PR changed `getHighestListenSequenceNumber` to take an IndedexDbTransaction. I had to duplicate a couple lines of code in the SimpleDb schema conversion to make this work, since the schema conversion needs the highest sequence number but doesn't use IndedexDbTransactions. 

The other callsite that does this conversion is part the TargetCache, which hasn't been started at this point.

I have no preference either way. The cleanest solution was probably the one that changed all sites to use IndedexDbTransaction and added three lines of duplicate code to the schema conversion.

wilhuff: It's a shame that we're ending up with two different transaction mechanisms and that migrations need to be written differently than the rest of the code. Is that the end state?

Is your intent to make everything mostly use IndedexDbTransaction except for the few places that can't? If so I'm not as concerned about the individual steps along the way.

schmidt-sebastian: I cleaned this up a little, and hope it is an improvement.
- I inlined two of the helper methods and changed them to only use IndedexDbTransactions/PersitenceTransactions.
- I changed the called in IndexedDbPersistence to use the method from the TargetCache. The TargetCache doesn't have a `start()`, so there is actually no issue here.
- The SimpleDbSchema class now has a couple lines of code duplication.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -231,16 +230,19 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+    return globalTargetStore.get(DbTargetGlobal.key).next(metadata => {
       const writeSentinelKey = (
         path: ResourcePath
       ): PersistencePromise<void> => {
         return documentTargetStore.put(
           new DbTargetDocument(
             0,
             encodeResourcePath(path),
-            currentSequenceNumber
+            metadata!.highestListenSequenceNumber!
```
schmidt-sebastian, line 245: We set the metadata earlier in the migration. It should exist at this point.

wilhuff: I'm wary of this style of commenting (review-only). The reviewer of this PR isn't the only one who's ever going to read this code.

Don't get me wrong--there's definitely a place for this kind of thing, but this comment in particular seems reasonable to just make in the code itself.

schmidt-sebastian: Added a debug assert that contains this message.

PR #3084, opened 2020-05-18
Add more Recovery spec tests

schmidt-sebastian:
Addresses #2755

### DIFF
--- a/packages/firestore/test/unit/specs/recovery_spec.test.ts
+++ b/packages/firestore/test/unit/specs/recovery_spec.test.ts
@@ -490,4 +490,41 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
         removed: [doc1]
       });
   });
+
+  specTest(
+    'User change handles transaction failures (with recovery)',
+    ['durable-persistence'],
+    () => {
+      const query = Query.atPath(path('collection'));
+      const doc1 = doc(
+        'collection/key1',
+        0,
+        { foo: 'a' },
+        { hasLocalMutations: true }
+      );
+      return spec()
+        .changeUser('user1')
+        .userSets('collection/key1', { foo: 'a' })
+        .userListens(query)
+        .expectEvents(query, {
+          added: [doc1],
+          fromCache: true,
+          hasPendingWrites: true
+        })
+        .failDatabaseTransactions('Handle user change')
+        .changeUser('user2')
+        .recoverDatabase()
+        .runTimer(TimerId.AsyncQueueRetry)
+        .expectEvents(query, { removed: [doc1], fromCache: true })
+        .failDatabaseTransactions('Handle user change')
+        .changeUser('user1')
+        .recoverDatabase()
+        .runTimer(TimerId.AsyncQueueRetry)
+        .expectEvents(query, {
+          added: [doc1],
+          fromCache: true,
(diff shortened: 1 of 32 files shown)


packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -114,6 +136,31 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
...
+    () => {
+      return client(0)
+        .expectPrimaryState(true)
+        .client(1)
+        .expectPrimaryState(false)
+        .client(0)
+        .failDatabaseTransactions('updateClientMetadataAndTryBecomePrimary')
+        .runTimer(TimerId.ClientMetadataRefresh)
+        .client(1)
+        .failDatabaseTransactions('updateClientMetadataAndTryBecomePrimary')
```
wilhuff, line 151: I'm sort of surprised we need to specify this on a per-client basis. My expectation would be that if the database is failing it would fail for all clients, no?

Do you find that you're writing tests where different clients see different failures? I wonder if we could make that the exception, rather than the rule?

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -114,6 +136,31 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
...
+        .runTimer(TimerId.ClientMetadataRefresh)
+        .client(0)
+        .recoverDatabase()
+        .runTimer(TimerId.ClientMetadataRefresh)
+        .expectPrimaryState(true)
+        .client(1)
+        .recoverDatabase()
+        .runTimer(TimerId.ClientMetadataRefresh)
```
wilhuff, line 159: What happens if client 1 happens to run their metadata refresh timer first?

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -57,6 +57,28 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
     }
   );
 
+  specTest(
+    'Clients fail to lookup mutations (with recovery)',
+    ['multi-client'],
+    () => {
+      return client(0)
```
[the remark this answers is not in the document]
schmidt-sebastian: Added top-level comment and comments inline to better explain this test.

packages/firestore/test/unit/specs/recovery_spec.test.ts
schmidt-sebastian: There are already a couple multi-tab tests in this file that use different failure modes per client, which also matches my expected behavior. A foreground tab is able to access IndexedDB, while a background tab lost filesystem access.

FWIW, I am not planning on adding further multi-tab tests for now.

packages/firestore/test/unit/specs/recovery_spec.test.ts
schmidt-sebastian: I added comment and a second section that verifies the reverse. The primary state doesn't transition during an IndexedDB failure and the order here should not affect which tab is primary.

Note that this is because of: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/indexeddb_persistence.ts#L433

PR #3085, opened 2020-05-18
Clean up recovery spec tests

schmidt-sebastian:
Almost all recovery spec tests only fail a single Transaction type. This PR cleans up the spec tests to make the calling code a lot prettier.

### DIFF
--- a/packages/firestore/test/unit/specs/recovery_spec.test.ts
+++ b/packages/firestore/test/unit/specs/recovery_spec.test.ts
@@ -23,6 +23,10 @@ import { Code } from '../../../src/util/error';
 import { deletedDoc, doc, filter, path } from '../../util/helpers';
 import { RpcError } from './spec_rpc_error';
 
+// The IndexedDB action that the Watch stream uses to detect if IndexedDB access
+// is available again.
+const ASYNC_QUEUE_PROBER = 'Get last remote snapshot version';
+
 describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
   specTest(
     'Write is acknowledged by primary client (with recovery)',
@@ -34,11 +38,7 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
           .client(1)
           .expectPrimaryState(false)
           .userSets('collection/a', { v: 1 })
-          .failDatabaseTransactions({
-            'Locally write mutations': true,
-            'Synchronize last document change read time': true,
-            'Lookup mutation documents': true
-          })
+          .failDatabaseTransactions('Lookup mutation documents')
           .client(0)
           .writeAcks('collection/a', 1, { expectUserCallback: false })
           .client(1)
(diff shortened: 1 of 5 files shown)


packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -34,11 +38,7 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
           .client(1)
           .expectPrimaryState(false)
           .userSets('collection/a', { v: 1 })
-          .failDatabaseTransactions({
-            'Locally write mutations': true,
-            'Synchronize last document change read time': true,
-            'Lookup mutation documents': true
-          })
+          .failDatabaseTransactions('Lookup mutation documents')
```
wilhuff, line 41: This seems like a different test than what was here before. Previously the write of the mutation to IndexedDB would fail (as would the lookup) but now this is failing just the lookup and allowing the write to succeed?

Alternatively, if this is working as intended, better names would help make it clearer just which transaction is failing. For example, if this were named `multitab.secondary.mutationLookup` it would be much, much clearer.

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -62,24 +62,21 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
...
         .expectPrimaryState(true)
         .client(1)
         .expectPrimaryState(false)
         .userListens(query)
-        .failDatabaseTransactions({
-          'Allocate target': true,
-          'Lookup mutation documents': true,
-          'Get new document changes': true
-        })
+        .failDatabaseTransactions('Get new document changes')
```
wilhuff, line 72: Same concern here: previously this failed to allocate a target at all, which should have caused the listen to fail. Now this is only failing much later.

packages/firestore/test/unit/specs/recovery_spec.test.ts
schmidt-sebastian: There is already a dedicated test that test if "Locally write mutations" fails ("Recovers when write cannot be persisted"). The fact that "Locally write mutations" is part of this list seems like an oversight to me - the mutation gets written right before and not after we start injecting failures.

There are certainly improvements to the transaction names that we should make before we port to Android/iOS, but I would like to point out that 'Lookup mutation documents' is also used by the primary tab to load a mutation that has been written by a secondary tab.

packages/firestore/test/unit/specs/recovery_spec.test.ts
schmidt-sebastian: Similar response: The target is already allocated and this seems like a copy/paste issue. Failed target allocations are tested in 'Fails targets that cannot be allocated'. 
I am not 100% convinced we need multi-tab specific tests for these initial failures, since these failures happen before we kick off any multi-tab specific failures. If we add a test, it would be a sanity check to ensure that we  don't write a Local Storage notification if we failed to persist the target/write.

wilhuff: OK. I see it now. Thanks for talking it through with me.

PR #3086, opened 2020-05-19
Remove the capability of sending events to clearcut from the Performance SDK.

visumickey:
Remove the capability of sending events to clearcut from the Performance SDK. This removes the need for fetching the RC configs to decide on the dispatch destination - so that is removed in the settings service and remote config service.

### DIFF
--- a/packages/performance/src/services/remote_config_service.test.ts
+++ b/packages/performance/src/services/remote_config_service.test.ts
@@ -236,5 +236,39 @@ describe('Performance Monitoring > remote_config_service', () => {
 
       expect(SettingsService.getInstance().loggingEnabled).to.be.true;
     });
+
+    it('gets the config from RC even with deprecated transport flag', async () => {
+      // Expired local config.
+      const EXPIRY_LOCAL_STORAGE_VALUE = '1556524895320';
+      const STRINGIFIED_CUSTOM_CONFIG = `{"entries":{\
+        "fpr_vc_network_request_sampling_rate":"0.250000",\
+        "fpr_log_transport_web_percent":"100.0",\
+        "fpr_vc_session_sampling_rate":"0.250000","fpr_vc_trace_sampling_rate":"0.500000"},\
+        "state":"UPDATE"}`;
+
+      const { storageGetItemStub: getItemStub } = setup(
+        {
+          expiry: EXPIRY_LOCAL_STORAGE_VALUE,
+          config: STRINGIFIED_CUSTOM_CONFIG
+        },
+        { reject: false, value: new Response(STRINGIFIED_CONFIG) }
+      );
+
+      await getConfig(IID);
+
+      expect(getItemStub).to.be.calledOnce;
+      expect(SettingsService.getInstance().loggingEnabled).to.be.true;
+      expect(SettingsService.getInstance().logEndPointUrl).to.equal(LOG_URL);
+      expect(SettingsService.getInstance().transportKey).to.equal(
+        TRANSPORT_KEY
+      );
+      expect(SettingsService.getInstance().logSource).to.equal(LOG_SOURCE);
+      expect(
+        SettingsService.getInstance().networkRequestsSamplingRate
+      ).to.equal(NETWORK_SAMPLIG_RATE);
+      expect(SettingsService.getInstance().tracesSamplingRate).to.equal(
+        TRACE_SAMPLING_RATE
+      );
+    });
   });
 });
--- a/packages/performance/src/services/remote_config_service.test.ts
+++ b/packages/performance/src/services/remote_config_service.test.ts
@@ -20,7 +20,7 @@ import { SettingsService } from './settings_service';
 import { CONFIG_EXPIRY_LOCAL_STORAGE_KEY } from '../constants';
 import { setupApi, Api } from './api_service';
 import * as iidService from './iid_service';
-import { getConfig, isDestFl } from './remote_config_service';
+import { getConfig } from './remote_config_service';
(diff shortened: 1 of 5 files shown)


ramanpreetSinghKhinda:
Quick thought: The codebase doesn't mention the name `clearcut` but the PR mentions. If the purpose is to hide the actual name than we need to create another PR and abandon this one.

ramanpreetSinghKhinda:
LGTM on a high level. I am not able to approve it though. May be only code owners can do that.

packages/performance/src/services/remote_config_service.test.ts
```diff
@@ -37,7 +37,6 @@ describe('Performance Monitoring > remote_config_service', () => {
   "fpr_log_endpoint_url":"https://firebaselogging.test.com",\
   "fpr_log_transport_key":"pseudo-transport-key",\
   "fpr_log_source":"2","fpr_vc_network_request_sampling_rate":"0.250000",\
-  "fpr_log_transport_web_percent":"100.0",\
```
zijianjoy, line 40: Can we keep one test which preserves `fpr_log_transport_web_percent` field in the config? Therefore we can validate that Web SDK doesn't fail while removing the definition of this flag.

packages/performance/src/services/remote_config_service.test.ts
```diff
@@ -20,7 +20,7 @@ import { SettingsService } from './settings_service';
 import { CONFIG_EXPIRY_LOCAL_STORAGE_KEY } from '../constants';
 import { setupApi, Api } from './api_service';
 import * as iidService from './iid_service';
-import { getConfig, isDestFl } from './remote_config_service';
```
zijianjoy, line 23: Let's also remove the definition of `isDestFl`: https://github.com/firebase/firebase-js-sdk/blob/0131e1fcade96853997fc04ba1fee0e8d40ed445/packages/performance/src/services/remote_config_service.ts#L271-L296

packages/performance/src/services/transport_service.ts
```diff
@@ -112,7 +112,7 @@ function dispatchQueueEvents(): void {
   };
   /* eslint-enable camelcase */
 
-  postToEndpoint(data, staged).catch(() => {
+  sendEventsToFl(data, staged).catch(() => {
```
zijianjoy, line 115: NIT: One thing we can do is to unwrap `sendEventsToFl` in the current function, since it doesn't require endpoint switching any more.

packages/performance/src/services/remote_config_service.ts
```diff
@@ -34,22 +34,14 @@ interface SecondaryConfig {
...
   logEndPointUrl?: string;
   transportKey?: string;
-  shouldSendToFl?: boolean;
   tracesSamplingRate?: number;
   networkRequestsSamplingRate?: number;
 }
 
 // These values will be used if the remote config object is successfully
 // retrieved, but the template does not have these fields.
 const DEFAULT_CONFIGS: SecondaryConfig = {
```
zijianjoy, line 43: Thought: It is nice to remove NO_TEMPLATE_CONFIGS in `remote_config_service.ts`. And I believe there is opportunity to improve even more around https://github.com/firebase/firebase-js-sdk/blob/acc545fc266340f6bb1136f2f6c3b43d68ff574c/packages/performance/src/services/remote_config_service.ts#L164-L225

visumickey: There is a lot of refactor/cleanup that can be done to make this more thinner and efficient. But I think that would be outside the context of this PR. Lets evaluate that separately.

packages/performance/src/services/remote_config_service.test.ts
visumickey: Good idea. Have added a test to validate that parsing of other relevant parameters are not impacted.

packages/performance/src/services/remote_config_service.test.ts
visumickey: Good catch, cleaned up!

packages/performance/src/services/transport_service.ts
visumickey: Unwrapping is a good idea. But that makes the function look too big. So I would prefer it to be as is to make it easy to understand.

packages/performance/src/services/remote_config_service.ts
```diff
@@ -75,12 +67,12 @@ const FIS_AUTH_PREFIX = 'FIREBASE_INSTALLATIONS_AUTH';
...
   const config = getStoredConfig();
   if (config) {
-    processConfig(iid, config);
+    processConfig(config);
     return Promise.resolve();
   }
 
   return getRemoteConfig(iid)
-    .then(config => processConfig(iid, config))
+    .then(config => processConfig(config))
```
alikn, line 75: This can be abbreviated as
.then(processConfig)

packages/performance/src/services/remote_config_service.ts
```diff
@@ -170,15 +162,13 @@ function getRemoteConfig(
  * is valid.
  */
 function processConfig(
-  iid: string,
   config: RemoteConfigResponse | undefined
```
alikn, line 165: can be written as 
function processConfig(config?: RemoteConfigResponse)

packages/performance/src/services/remote_config_service.ts
visumickey: Done.

packages/performance/src/services/remote_config_service.ts
visumickey: Done.

PR #3095, opened 2020-05-20
Fix Saucelabs tests for Firestore unit tests except IE

hsubox76:
Fix 1 error in Firefox + Safari and 2 errors in Safari only.

IE is excluded, to be fixed in a future PR, because:
1) it has 52 errors
2) It requires Babel compilation which causes OOM because it runs on the entire bundle of test files. This can be fixed by not using the bootstrap test bundle and just listing file globs in karma.conf.js the normal way.  This may be a little less performant (https://github.com/webpack-contrib/karma-webpack#alternative-usage) but need to test out. In the meantime, since we are postponing IE as a large task, why not skip Babel compilation temporarily.

TODO: Next step is to try to run `integration/firestore` tests instead of integration tests in `package/firestore`.  It is almost ready to run, but has 1 error common across all browsers, 1 exclusive to Firefox, and 12/13 in IE (memory/persistence).

### DIFF
--- a/packages/firestore/test/unit/util/async_queue.test.ts
+++ b/packages/firestore/test/unit/util/async_queue.test.ts
@@ -143,7 +143,8 @@ describe('AsyncQueue', () => {
     const doStep = (n: number): Promise<number> =>
       defer(() => completedSteps.push(n));
     queue.enqueueAndForget(() => doStep(1));
-    const last = queue.enqueueAfterDelay(timerId1, 5, () => doStep(4));
+    // Flaky on Safari, increasing delay to 1000ms to try to increase reliability.
+    const last = queue.enqueueAfterDelay(timerId1, 1000, () => doStep(4));
     // eslint-disable-next-line @typescript-eslint/no-floating-promises
     queue.enqueueAfterDelay(timerId2, 1, () => doStep(3));
     queue.enqueueAndForget(() => doStep(2));
--- a/packages/firestore/test/unit/util/async_queue.test.ts
+++ b/packages/firestore/test/unit/util/async_queue.test.ts
@@ -27,6 +27,13 @@ import { IndexedDbTransactionError } from '../../../src/local/simple_db';
 
 use(chaiAsPromised);
 
+function isSafari(): boolean {
+  return (
+    navigator.userAgent.includes('Safari') &&
+    !navigator.userAgent.includes('Chrome')
+  );
+}
+
 describe('AsyncQueue', () => {
   // We reuse these TimerIds for generic testing.
   const timerId1 = TimerId.ListenStreamConnectionBackoff;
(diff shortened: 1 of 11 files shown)


packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -143,7 +143,8 @@ describe('AsyncQueue', () => {
     const doStep = (n: number): Promise<number> =>
       defer(() => completedSteps.push(n));
     queue.enqueueAndForget(() => doStep(1));
-    const last = queue.enqueueAfterDelay(timerId1, 5, () => doStep(4));
+    // Flaky on Safari, increasing delay to 1000ms to try to increase reliability.
+    const last = queue.enqueueAfterDelay(timerId1, 1000, () => doStep(4));
```
schmidt-sebastian, line 147: This is too long - we cannot wait a full second for a single test. Can we skip this test in Safari?

hsubox76: Sure, adding a check.

packages/firestore/test/util/test_platform.ts
```diff
@@ -272,11 +273,11 @@ export class TestPlatform implements Platform {
   }
 }
 
-/** Returns true if we are running under Node. */
-export function isNode(): boolean {
```
hsubox76, line 276: Seems like this wasn't being used anywhere, and there's a better version in @firebase/util so I hope it's okay to remove?

schmidt-sebastian: @wu-hui might have just started using this. He would probably be delighted to switch to the better version :)

packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -137,7 +138,9 @@ describe('AsyncQueue', () => {
     });
   });
 
-  it('can schedule ops in the future', async () => {
+  // Flaky on Safari.
```
Feiyang1, line 141: Do we know why it is flaky on Safari?

packages/firestore/test/util/test_platform.ts
```diff
@@ -272,11 +273,11 @@ export class TestPlatform implements Platform {
...
+/** Returns true if we are running in Safari. */
+export function isSafari(): boolean {
```
Feiyang1, line 277: Should it be moved to `@firebase/util`? It looks like something that should be shared.

packages/firestore/test/unit/util/async_queue.test.ts
hsubox76: I haven't been able to find any reason. The test relies on two setTimeouts at 1ms and 5ms happening in the correct order, and sometimes Safari doesn't seem to get the order right. I can't find any known bug or issue that might cause that.

packages/firestore/test/util/test_platform.ts
hsubox76: Moved.

PR #3097, opened 2020-05-21
Implement a bundle reader for Web

wu-hui:
The implementation is based around Web's ReadableStream. This does not work on NodeJS unfortunately. I need to fix that, but want to send this out before it gets too big.

### DIFF
--- a/packages/firestore/src/util/bundle.ts
+++ b/packages/firestore/src/util/bundle.ts
@@ -1,0 +1,215 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as api from '../protos/firestore_proto_api';
+import {
+  BundledDocumentMetadata,
+  BundledQuery,
+  BundleElement,
+  BundleMetadata
+} from '../protos/firestore_bundle_proto';
+
+/**
+ * A complete element in the bundle stream, together with the byte length it
+ * occupies in the stream.
+ */
+export class SizedBundleElement {
+  constructor(
+    public payload: BundledQuery | api.Document | BundledDocumentMetadata,
+    public byteLength: number
+  ) {}
+}
+
+/**
+ * Create a `ReadableStream` from a underlying buffer.
+ *
+ * @param data: Underlying buffer.
+ * @param bytesPerRead: How many bytes to read from the underlying buffer from each read through the stream.
+ */
+export function toReadableStream(
+  data: Uint8Array | ArrayBuffer,
+  bytesPerRead = 10240
+): ReadableStream<Uint8Array | ArrayBuffer> {
+  let readFrom = 0;
+  return new ReadableStream({
+    start(controller) {},
+    async pull(controller): Promise<void> {
+      controller.enqueue(data.slice(readFrom, readFrom + bytesPerRead));
+      readFrom += bytesPerRead;
+      if (readFrom >= data.byteLength) {
+        controller.close();
+      }
+    }
+  });
+}
+
+/**
+ * A class representing a bundle.
+ *
+ * Takes a bundle stream or buffer, and presents abstractions to read bundled
+ * elements out of the underlying content.
+ */
+export class Bundle {
+  // Cached bundle metadata.
+  private metadata?: BundleMetadata | null;
+  // The reader instance of the given ReadableStream.
+  private reader: ReadableStreamDefaultReader;
+  // Internal buffer to hold bundle content, accumulating incomplete element content.
+  private buffer: Uint8Array = new Uint8Array();
+  private textDecoder = new TextDecoder('utf-8');
+
+  constructor(
+    private bundleStream:
+      | ReadableStream<Uint8Array | ArrayBuffer>
+      | Uint8Array
+      | ArrayBuffer
+  ) {
+    if (
+      bundleStream instanceof Uint8Array ||
+      bundleStream instanceof ArrayBuffer
+    ) {
+      this.bundleStream = toReadableStream(bundleStream);
+    }
+    this.reader = (this.bundleStream as ReadableStream).getReader();
+  }
+
+  /**
+   * Returns the metadata of the bundle.
+   */
+  async getMetadata(): Promise<BundleMetadata> {
+    if (!this.metadata) {
+      const result = await this.nextElement();
+      if (result === null || result instanceof SizedBundleElement) {
+        throw new Error(`The first element is not metadata, it is ${result}`);
+      }
+      this.metadata = (result as BundleElement).metadata;
+    }
+
+    return this.metadata!;
+  }
+
+  /**
+   * Asynchronously iterate through all bundle elements (except bundle metadata).
+   */
+  async *elements(): AsyncIterableIterator<SizedBundleElement> {
+    let element = await this.nextElement();
+    while (element !== null) {
+      if (element instanceof SizedBundleElement) {
+        yield element;
+      } else {
+        this.metadata = element.metadata;
+      }
+      element = await this.nextElement();
+    }
+  }
+
+  // Reads from the head of internal buffer, and pulling more data from underlying stream if a complete element
+  // cannot be found, until an element(including the prefixed length and the JSON string) is found.
+  //
+  // Once a complete element is read, it is dropped from internal buffer.
+  //
+  // Returns either the bundled element, or null if we have reached the end of the stream.
+  private async nextElement(): Promise<
+    BundleElement | SizedBundleElement | null
+  > {
+    const lengthBuffer = await this.readLength();
+    if (lengthBuffer === null) {
+      return null;
+    }
+
+    const lengthString = this.textDecoder.decode(lengthBuffer);
+    const length = parseInt(lengthString, 10);
+    if (isNaN(length)) {
+      throw new Error(`length string (${lengthString}) is not valid number`);
+    }
+
+    const jsonString = await this.readJsonString(lengthBuffer.length, length);
+    // Update the internal buffer to drop the read length and json string.
+    this.buffer = this.buffer.slice(lengthBuffer.length + length);
+
+    if (!this.metadata) {
+      const element = JSON.parse(jsonString) as BundleElement;
+      return element;
+    } else {
+      return new SizedBundleElement(
+        JSON.parse(jsonString),
+        lengthBuffer.length + length
+      );
+    }
+  }
+
+  // First index of '{' from the underlying buffer.
+  private indexOfOpenBracket(): number {
+    return this.buffer.findIndex(v => v === 123);
+  }
+
+  // Reads from the beginning of the inernal buffer, until the first '{', and return
+  // the content.
+  // If reached end of the stream, returns a null.
+  private async readLength(): Promise<Uint8Array | null> {
+    let position = this.indexOfOpenBracket();
+    while (position < 0) {
+      const done = await this.pullMoreDataToBuffer();
+      if (done) {
+        if (this.buffer.length === 0) {
+          return null;
+        }
+        position = this.indexOfOpenBracket();
+        // Underlying stream is closed, and we still cannot find a '{'.
+        if (position < 0) {
+          throw new Error(
+            'Reach to the end of bundle when a length string is expected.'
+          );
+        }
+      } else {
+        position = this.indexOfOpenBracket();
+      }
+    }
+
+    return this.buffer.slice(0, position);
+  }
+
+  // Reads from a specified position from the internal buffer, for a specified
+  // number of bytes, pulling more data from the underlying stream if needed.
+  //
+  // Returns a string decoded from the read bytes.
+  private async readJsonString(start: number, length: number): Promise<string> {
+    while (this.buffer.length < start + length) {
+      const done = await this.pullMoreDataToBuffer();
+      if (done) {
+        throw new Error('Reach to the end of bundle when more is expected.');
+      }
+    }
+
+    return this.textDecoder.decode(this.buffer.slice(start, start + length));
+  }
+
+  // Pulls more data from underlying stream to internal buffer.
+  // Returns a boolean indicating whether the stream is finished.
+  private async pullMoreDataToBuffer(): Promise<boolean> {
+    const result = await this.reader.read();
+    if (!result.done) {
+      const newBuffer = new Uint8Array(
+        this.buffer.length + result.value.length
+      );
+      newBuffer.set(this.buffer);
+      newBuffer.set(result.value, this.buffer.length);
+      this.buffer = newBuffer;
+    }
+    return result.done;
+  }
+}
--- a/packages/firestore/src/util/bundle_reader.ts
+++ b/packages/firestore/src/util/bundle_reader.ts
@@ -97,6 +97,7 @@ export class BundleReader {
     }
     this.reader = (this.bundleStream as ReadableStream).getReader();
 
+    // Read the metadata (which is the first element).
     this.nextElement().then(
       element => {
         if (element && element.isBundleMetadata()) {
@@ -109,9 +110,7 @@ export class BundleReader {
           );
         }
       },
-      error => {
-        this.metadata.reject(error);
-      }
+      error => this.metadata.reject(error)
     );
   }
 
@@ -175,23 +174,27 @@ export class BundleReader {
    * If reached end of the stream, returns a null.
    */
   private async readLength(): Promise<Uint8Array | null> {
-    let position = this.indexOfOpenBracket();
-    while (position < 0) {
+    let position: number;
+    while ((position = this.indexOfOpenBracket()) < 0) {
       const done = await this.pullMoreDataToBuffer();
       if (done) {
-        if (this.buffer.length === 0) {
-          return null;
-        }
-        position = this.indexOfOpenBracket();
-        // Underlying stream is closed, and we still cannot find a '{'.
-        if (position < 0) {
-          this.raiseError(
-            'Reached the end of bundle when a length string is expected.'
-          );
-        }
-      } else {
-        position = this.indexOfOpenBracket();
+        break;
       }
+    }
+
+    // Broke out of the loop because underlying stream is closed, and there
+    // happens to be no more data to process.
+    if (this.buffer.length === 0) {
+      return null;
+    }
+
+    position = this.indexOfOpenBracket();
+    // Broke out of the loop because underlying stream is closed, but still
+    // cannot find an open bracket.
+    if (position < 0) {
+      this.raiseError(
+        'Reached the end of bundle when a length string is expected.'
+      );
     }
 
     const result = this.buffer.slice(0, position);
--- a/packages/firestore/src/util/bundle_reader.ts
+++ b/packages/firestore/src/util/bundle_reader.ts
@@ -19,6 +19,8 @@ import {
   BundleElement,
   BundleMetadata
 } from '../protos/firestore_bundle_proto';
+import { Deferred } from './promise';
+import { AsyncQueue } from './async_queue';
 
 /**
  * A complete element in the bundle stream, together with the byte length it
@@ -40,7 +42,8 @@ export class SizedBundleElement {
  * Create a `ReadableStream` from a underlying buffer.
  *
  * @param data: Underlying buffer.
- * @param bytesPerRead: How many bytes to read from the underlying buffer from each read through the stream.
+ * @param bytesPerRead: How many bytes to read from the underlying buffer from
+ * each read through the stream.
  */
 export function toReadableStream(
   data: Uint8Array | ArrayBuffer,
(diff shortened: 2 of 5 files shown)


packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+  // Returns a boolean indicating whether the stream is finished.
+  private async pullMoreDataToBuffer(): Promise<boolean> {
+    const result = await this.reader.read();
+    if (!result.done) {
+      const newBuffer = new Uint8Array(
+        this.buffer.length + result.value.length
+      );
+      newBuffer.set(this.buffer);
+      newBuffer.set(result.value, this.buffer.length);
+      this.buffer = newBuffer;
```
schmidt-sebastian, line 211: This is a very expensive way of consuming the buffer that creates `O(n^2)` copies. 

Can we not take advantage of the fact that we are using length-prefixed JSON and remove the internal buffer?

- Read first n bytes to get length (let's say n = 10)
- Length is 123, we read 7 bytes extra
- Read next 116 bytes, store in nextBytes
- JSON.parse(length.slice(7), nextBytes)
- Return as bundle element

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+  BundledQuery,
+  BundleElement,
+  BundleMetadata
+} from '../protos/firestore_bundle_proto';
+
+/**
+ * A complete element in the bundle stream, together with the byte length it
+ * occupies in the stream.
+ */
+export class SizedBundleElement {
```
schmidt-sebastian, line 30: Would it be possible to replace this with `BundleElement`?

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  // Reads from a specified position from the internal buffer, for a specified
+  // number of bytes, pulling more data from the underlying stream if needed.
+  //
+  // Returns a string decoded from the read bytes.
+  private async readJsonString(start: number, length: number): Promise<string> {
+    while (this.buffer.length < start + length) {
+      const done = await this.pullMoreDataToBuffer();
+      if (done) {
+        throw new Error('Reach to the end of bundle when more is expected.');
```
schmidt-sebastian, line 194: s/Reach to the/Reached/

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+      const done = await this.pullMoreDataToBuffer();
+      if (done) {
+        if (this.buffer.length === 0) {
+          return null;
+        }
+        position = this.indexOfOpenBracket();
+        // Underlying stream is closed, and we still cannot find a '{'.
+        if (position < 0) {
+          throw new Error(
+            'Reach to the end of bundle when a length string is expected.'
```
schmidt-sebastian, line 175: s/Reach to the/Reached/

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+        throw new Error('Reach to the end of bundle when more is expected.');
+      }
+    }
+
+    return this.textDecoder.decode(this.buffer.slice(start, start + length));
+  }
+
+  // Pulls more data from underlying stream to internal buffer.
+  // Returns a boolean indicating whether the stream is finished.
+  private async pullMoreDataToBuffer(): Promise<boolean> {
```
schmidt-sebastian, line 203: FWIW a common pattern for this type of method is to return the number of bytes read or `-1` if the end of the stream is reached.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,354 @@
...
+(isNode() ? describe.skip : describe)('Bundle ', () => {
+  genericBundleReadingTests(1);
+  genericBundleReadingTests(4);
+  genericBundleReadingTests(64);
+  genericBundleReadingTests(1024);
+});
+
+function genericBundleReadingTests(bytesPerRead: number): void {
+  // Setting up test data.
+  const meta = {
```
schmidt-sebastian, line 69: Would it make sense to check this as a test resource and to read it from a file instead of setting this up here in this test? That makes the test much shorter (and more readable) and matches the expected usage pattern.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,354 @@
...
+        bytesPerRead
+      );
+      const bundle = new Bundle(bundleStream);
+
+      const actual = [];
+      for await (const sizedElement of bundle.elements()) {
+        actual.push(sizedElement);
+      }
+      expect(actual.length).to.equal(5);
+      expect(actual[0]).to.deep.equal({
```
schmidt-sebastian, line 238: Please add helper to get this pattern down to one line.

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+      this.metadata = (result as BundleElement).metadata;
+    }
+
+    return this.metadata!;
+  }
+
+  /**
+   * Asynchronously iterate through all bundle elements (except bundle metadata).
+   */
+  async *elements(): AsyncIterableIterator<SizedBundleElement> {
```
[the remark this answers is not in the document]
wu-hui: Thanks for the tip!

While it is very nice, I don't think it is required, how about I expose `nextElement` instead, also making it return `Promise<SizedBundleElement | null>` (no `BundleElement`). 

Implementation of `SizedBundleElement` will also have method to tell what payload it is storing.

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+    const length = parseInt(lengthString, 10);
+    if (isNaN(length)) {
+      throw new Error(`length string (${lengthString}) is not valid number`);
+    }
+
+    const jsonString = await this.readJsonString(lengthBuffer.length, length);
+    // Update the internal buffer to drop the read length and json string.
+    this.buffer = this.buffer.slice(lengthBuffer.length + length);
+
+    if (!this.metadata) {
```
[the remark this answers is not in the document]
wu-hui: Metadata being the first element of a stream is a feature, IMO. I cannot think of a use case where metadata being in the middle of a stream is useful, and I think it's better to reject bundles without metadata upfront, than try to cope with it.

I am willing to add it if someone actually find this useful, and request for that.

packages/firestore/src/util/bundle.ts
```diff
@@ -0,0 +1,215 @@
...
+  private indexOfOpenBracket(): number {
+    return this.buffer.findIndex(v => v === 123);
+  }
+
+  // Reads from the beginning of the inernal buffer, until the first '{', and return
+  // the content.
+  // If reached end of the stream, returns a null.
+  private async readLength(): Promise<Uint8Array | null> {
+    let position = this.indexOfOpenBracket();
+    while (position < 0) {
```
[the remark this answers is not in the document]
wu-hui: Do you mean we always encode length in a fixed length string, like instead of `123`, we always encode it as `00000123`?

This might be easier to implement (and less error-prone), but note when dealing with `ReadableStream` we cannot control how much bytes we read back. So we still need to pull more data from the stream if it just happens that our first read is not sufficient.

packages/firestore/src/util/bundle.ts
wu-hui: Note we cannot control how many bytes we read from `ReadableStream`, so an internal buffer cannot be avoided completely. 

Currently implementation only pulls more data when necessary, so the internal buffer is already kept as small as possible when there is a need to pull and concat.

Actually, I don't think it's `O(n^2)` copies in reality, because most of the data read from underlying stream is copied only once (into existing internal buffer), the part that will be copied again is the part that is the start of an element, which will be copied again when the next pull is called. 

The worst case is when one element requires multiple pulls, and each pulls will have to copy the old and new buffer. I feel this might not be very common though.

packages/firestore/test/unit/util/bundle.test.ts
wu-hui: It definitely makes it shorter, but might not be as more readable as the size reduction? Because readers now have to read the built bundle content to make sure the tests are doing the right thing?

packages/firestore/src/util/bundle.ts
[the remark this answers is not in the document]
schmidt-sebastian: I would suspect that even if Typescript transpiles generator functions down to ES5, the code size will be much larger than a simple `nextElement()` method. I am okay with the change you are proposing here.

packages/firestore/src/util/bundle.ts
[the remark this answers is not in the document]
schmidt-sebastian: How about we compromise on this one:
- The logic to parse the bundles will be agnostic of the position in the stream. I suspect that this will make the code much easier to understand.
- We throw/reject the user Promise if the bundle stream is invalid.  This should be secondary check that has a clear error (something like: "First element in bundle is not a metadata element")

packages/firestore/src/util/bundle.ts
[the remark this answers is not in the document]
schmidt-sebastian: 1) I don't think we should zero pad our length. While easy to implement, it seems a little hackish. We can however assume a maximum length of a number. The maximum number could be based on document size or integer boundaries (2^53, 2^64 or 2^128 - depending on other implementations).

2) Could we build a simple wrapper class that is used by both `readLength()` and `nextElement()`? This will essentially be the same as `pullMoreDataToBuffer()`, but it would allow us to implement a very simple version of `nextElement()`:

```
function nextElement() {
  const lengthBuffer = buffer.readInt(); // https://docs.oracle.com/javase/7/docs/api/java/io/DataInputStream.html#readInt()
  const json = buffer.read(lengthBuffer);
  return JSON.parse(buffer); // TODO: Validate contents
}
```

Having written this, it occurs to me that what you have is not all that different. Maybe if we can just change `readLength()` to call `readJsonString()`, which would unify the code to poll form `pullMoreDataToBuffer()`? If you added an option to not advance the buffer your calling code could be:

```
function nextElement() {
  const lengthBuffer = buffer.readJsonString(10, advance_no);  
  const length = decode(...);
  buffer.read(String(length).length, advance_yes)
  const json = buffer.read(length, advance_yes);
}
```

Sorry for my contradicting and non-coherent rambling.

packages/firestore/test/unit/util/bundle.test.ts
schmidt-sebastian: I see your point. However, I think this test is less readable precisely because of this setup. I can go either way here and defer to your judgement.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * Returns the metadata of the bundle.
+   */
+  async getMetadata(): Promise<BundleMetadata> {
+    if (!this.metadata) {
+      await this.nextElement();
+    }
+
+    return this.metadata!;
```
schmidt-sebastian, line 100: I still think this code is a bit surprising. It is not clear to me why `nextElement()` sets `this.metadata`.

I played around with your PR a bit. What do you think of these changes? https://paste.googleplex.com/5782882595373056

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+      } else {
+        this.raiseError(
+          `The first element of the bundle is not a metadata, it is ${JSON.stringify(
+            element.payload
+          )}`
+        );
+      }
+    }
+
+    return element;
```
schmidt-sebastian, line 128: The changes linked above simplify this and the related method above.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+          this.raiseError(
+            'Reached the end of bundle when a length string is expected.'
+          );
+        }
+      } else {
+        position = this.indexOfOpenBracket();
+      }
+    }
+
+    return this.buffer.slice(0, position);
```
schmidt-sebastian, line 189: Should this method advance the buffer as well and return the length as a number, rather than returning the encoded byte array (this is not part of my change)?

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+    // Update the internal buffer to drop the read length and json string.
+    this.buffer = this.buffer.slice(lengthBuffer.length + length);
+
+    return new SizedBundleElement(
+      JSON.parse(jsonString),
+      lengthBuffer.length + length
+    );
+  }
+
+  // First index of '{' from the underlying buffer.
```
schmidt-sebastian, line 161: JSDoc

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+  private indexOfOpenBracket(): number {
+    return this.buffer.findIndex(v => v === '{'.charCodeAt(0));
+  }
+
+  // Reads from the beginning of the internal buffer, until the first '{', and return
+  // the content.
+  // If reached end of the stream, returns a null.
+  private async readLength(): Promise<Uint8Array | null> {
```
schmidt-sebastian, line 169: JSDoc

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    return this.buffer.slice(0, position);
+  }
+
+  // Reads from a specified position from the internal buffer, for a specified
+  // number of bytes, pulling more data from the underlying stream if needed.
+  //
+  // Returns a string decoded from the read bytes.
+  private async readJsonString(start: number, length: number): Promise<string> {
```
schmidt-sebastian, line 196: JSDoc

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+    while (this.buffer.length < start + length) {
+      const bytesRead = await this.pullMoreDataToBuffer();
+      if (bytesRead < 0) {
+        this.raiseError('Reached the end of bundle when more is expected.');
+      }
+    }
+
+    return this.textDecoder.decode(this.buffer.slice(start, start + length));
```
schmidt-sebastian, line 204: Similarly, could we advance the buffer here? It might make the code simpler as `start` would always be zero now.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      id: 'test-bundle',
+      createTime: { seconds: 1577836805, nanos: 6 },
+      version: 1,
+      totalDocuments: 1,
+      totalBytes: 416
+    }
+  };
+  const metaString = lengthPrefixedString(meta);
+
+  const doc1Meta = {
```
schmidt-sebastian, line 85: One benefit of not using JSON is that you could encore types here. It might help if you assigned a strict type here:

```
const doc1Meta : BundleMetadata = {}
```

Your test is currently mostly untyped. If you change the schema, your test will still compile and may even pass depending on if you update the test expectations. By enforcing types manually, you can help us to diagnose problems in the future.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,332 @@
...
+      const actual = await getAllElement(bundle);
+      expect(actual.length).to.equal(0);
+    }
+  );
+
+  it(
+    'throws with ill-formatted bundle with bytesPerRead ' + bytesPerRead,
+    async () => {
+      await expect(
+        parseThroughBundle('metadata: "no length prefix"', bytesPerRead)
```
schmidt-sebastian, line 312: In the Web SDK, we generally assert on the actual error messages as this allows us to verify that the messages make sense. Do you mind doing this here? You can look at `validation.test.ts` for examples.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  private raiseError(message: string): void {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.reader.cancel('Invalid bundle format.');
+    throw new Error(message);
+  }
+
+  // Pulls more data from underlying stream to internal buffer.
+  // Returns a boolean indicating whether the stream is finished.
+  private async pullMoreDataToBuffer(): Promise<number> {
```
schmidt-sebastian, line 215: JSDoc

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  private raiseError(message: string): void {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.reader.cancel('Invalid bundle format.');
+    throw new Error(message);
+  }
+
+  // Pulls more data from underlying stream to internal buffer.
+  // Returns a boolean indicating whether the stream is finished.
```
schmidt-sebastian, line 214: If you keep this as is, please update the method comment to stay that it returns the length read. I did notice however that this information is not used, so we could keep the boolean.

packages/firestore/test/unit/util/bundle.test.ts
wu-hui: Done.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,338 @@
...
+    payload: unknown,
+    payloadString: string
+  ): void {
+    expect(element.payload).to.deep.equal(payload);
+    expect(element.byteLength).to.equal(
+      encoder.encode(payloadString).byteLength
+    );
+  }
+
+  async function parseThroughBundle(
```
schmidt-sebastian, line 197: This reads like "passthrough" but it's not. Maybe `generateBundleAndParse`?

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,245 @@
...
+  ) {
+    if (
+      bundleStream instanceof Uint8Array ||
+      bundleStream instanceof ArrayBuffer
+    ) {
+      this.bundleStream = toReadableStream(bundleStream);
+    }
+    this.reader = (this.bundleStream as ReadableStream).getReader();
+
+    this.nextElement().then(
```
schmidt-sebastian, line 100: Maybe add a small inline comment before this line:

// Read the metadata (which is the first element)

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,245 @@
...
+        // Underlying stream is closed, and we still cannot find a '{'.
+        if (position < 0) {
+          this.raiseError(
+            'Reached the end of bundle when a length string is expected.'
+          );
+        }
+      } else {
+        position = this.indexOfOpenBracket();
+      }
+    }
```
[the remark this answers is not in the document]
wu-hui: Thanks for the hint!

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -0,0 +1,338 @@
...
+          limit: { 'value': 1 }
+        },
+        limitType: 'LAST'
+      },
+      readTime: { 'seconds': 1590011379, 'nanos': 543063000 }
+    }
+  };
+  const limitToLastQueryString = lengthPrefixedString(limitToLastQuery);
+
+  async function getAllElement(
```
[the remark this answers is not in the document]
wu-hui: Done.

packages/firestore/test/unit/util/bundle.test.ts
wu-hui: Done.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -175,25 +174,29 @@ export class BundleReader {
...
       }
     }
 
+    // Broke out of the loop because underlying stream is closed, and there
+    // happens to be no more data to process.
+    if (this.buffer.length === 0) {
+      return null;
+    }
+
+    position = this.indexOfOpenBracket();
```
schmidt-sebastian, line 189: You can drop `position` from the loop above and assign it here as a const.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    this.nextElement().then(
+      element => {
+        if (element && element.isBundleMetadata()) {
+          this.metadata.resolve(element.payload.metadata!);
+        } else {
+          const payload = (element || { payload: null }).payload;
+          this.metadata.reject(
+            new Error(`The first element of the bundle is not a metadata, it is
+          ${JSON.stringify(payload)}`)
```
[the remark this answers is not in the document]
schmidt-sebastian: It works for me.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  /**
+   * Returns the next BundleElement (together with its byte size in the bundle)
+   * that has not been read from underlying ReadableStream. Returns null if we
+   * have reached the end of the stream.
+   */
+  async nextElement(): Promise<SizedBundleElement | null> {
+    // Ensures `nextElementImpl` calls are executed sequentially before they
+    // modifies internal buffer.
+    return this.asyncQueue.enqueue(() => this.nextElementImpl());
```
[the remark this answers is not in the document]
schmidt-sebastian: As discussed offline, the queue has been removed. This improves the error handling for the user as we can distinguish between different error messages and only raise the AsyncQueue errors for internal failures.

packages/firestore/src/util/bundle_reader.ts
wu-hui: Done.

packages/firestore/src/util/bundle_reader.ts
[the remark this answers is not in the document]
wu-hui: Hmm, my intellij keeps complaining, but compile from CLI works. I guess there is some config i need to fix for my intellij.

PR #3100, opened 2020-05-21
Add Firestore Lite API

schmidt-sebastian:
This adds the API for the Lite SDK as approved in go/firestore-next

### DIFF
--- a/packages/firestore/lite/index.d.ts
+++ b/packages/firestore/lite/index.d.ts
@@ -17,9 +17,15 @@
 
 import { FirebaseApp } from '@firebase/app-types';
 
-export type DocumentData = { [field: string]: any };
+/* eslint-disable @typescript-eslint/no-explicit-any */
 
-export type UpdateData = { [fieldPath: string]: any };
+export interface DocumentData {
+  [field: string]: any;
+}
+
+export interface UpdateData {
+  [fieldPath: string]: any;
+}
 
 export interface Settings {
   host?: string;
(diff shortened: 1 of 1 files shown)


packages/firestore/lite/index.d.ts
```diff
@@ -0,0 +1,338 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '@firebase/app-types';
+
+export type DocumentData = { [field: string]: any };
```
rafikhan, line 20: Should this be an interface vs a type?

schmidt-sebastian: This is mainly because it matches the type in the current API (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts#L20)

AFAIK there is no real difference, but this particular type should never be used in a class hierarchy.

The GitHub Actions seems to use different Lint settings from what we do. I only get lint violations for the usage of `any`.

rafikhan: As a separate work item we should make sure our local development and GHA match up.

schmidt-sebastian: @Feiyang1

schmidt-sebastian: FWIW, this is the first time I see this kind of discrepancy (but I am also not sure when we started Lint checking in this Rep, the nodejs-firestore repo has had this check on for a while).

schmidt-sebastian:
@rafikhan It sounds like you are satisfied with this. If so, can you approve? I can then start on some follow up work. Thanks!

rafikhan:
LGTM

PR #3101, opened 2020-05-22
Only run the provided delayed operation early

schmidt-sebastian:
The code to fast-forward the IndexedDB retries uses `runDelayedOperationsEarly()`, which runs not only the provided delayed operation, but all operations that were scheduled prior. This causes the OnlineStateTimer to run on page visibility events (see #2923).

Since Timer operations are likely not dependent on one another, it seems better to only run the provided operation. This also makes the API much less surprising.

Note that `op.skipDelay()` cannot be used directly in the visibility handler since we do not keep track of the DelayedOperation once it is scheduled via the backoff algorithm.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2923

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -463,23 +463,21 @@ export class AsyncQueue {
   }
 
   /**
-   * For Tests: Runs some or all delayed operations early.
+   * Runs some or all delayed operations early.
    *
-   * @param lastTimerId Delayed operations up to and including this TimerId will
-   *  be drained. Throws if no such operation exists. Pass TimerId.All to run
-   *  all delayed operations.
+   * @param timerId Delayed operations to run. Pass TimerId.All to run all
+   * delayed operations.
    * @returns a Promise that resolves once all operations have been run.
    */
-  runDelayedOperationsEarly(lastTimerId: TimerId): Promise<void> {
+  runDelayedOperationsEarly(timerId: TimerId): Promise<void> {
     // Note that draining may generate more delayed ops, so we do that first.
     return this.drain().then(() => {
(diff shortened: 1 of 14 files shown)


packages/firestore/src/util/async_queue.ts
```diff
@@ -463,23 +463,21 @@ export class AsyncQueue {
...
    *
-   * @param lastTimerId Delayed operations up to and including this TimerId will
-   *  be drained. Throws if no such operation exists. Pass TimerId.All to run
-   *  all delayed operations.
+   * @param timerId Delayed operations to run. Pass TimerId.All to run all
+   * delayed operations.
    * @returns a Promise that resolves once all operations have been run.
    */
-  runDelayedOperationsEarly(lastTimerId: TimerId): Promise<void> {
+  runDelayedOperationsEarly(timerId: TimerId): Promise<void> {
```
wilhuff, line 472: This breaks the intent of what this test-only method was for. The idea was that you could fast-forward time until the delayed operation ran, including anything scheduled prior to it. In this interface, now you have to know exactly what you want to run, which makes this a much less effective tool for testing. In particular, this makes it really hard to use this for spec tests because we're no longer simulating the actual passage of time, but merely jumping ahead to this specific event. This makes it far less likely that we'll catch scheduling-related problems with tests.

If we really need to allow a scheduled operation to run early, can't we add this as a method to the `DelayedOperation`?

schmidt-sebastian: I updated the code to return DelayedOperation instead of CancellablePromise, and added a skipDelay to ExponentialBackoff.

I feel that we might even get rid of the TimerIds with this change, since we now have direct access to the underlying operations - but that of course would remove the ability to run delayed operations in order. It might drastically simplify this code though.

PR #3109, opened 2020-05-26
Allow shared usage of UserDataReader and UserDataWriter

schmidt-sebastian:
Introduces a common base type for all DocumentReference APIS. DocumentKeyReference exposes all state that is common between the Lite, Full and Legacy SDKs.

Updates UserDataReader and UserDataWriter to use the DocumentKeyReference.

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -88,7 +88,11 @@ import {
   PartialObserver,
   Unsubscribe
 } from './observer';
-import { fieldPathFromArgument, UserDataReader } from './user_data_reader';
+import {
+  DocumentKeyReference,
+  fieldPathFromArgument,
+  UserDataReader
+} from './user_data_reader';
 import { UserDataWriter } from './user_data_writer';
 import { FirebaseAuthInternalName } from '@firebase/auth-interop-types';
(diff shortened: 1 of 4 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -1347,10 +1353,10 @@ export class DocumentSnapshot<T = firestore.DocumentData>
         return this._converter.fromFirestore(snapshot, options);
       } else {
         const userDataWriter = new UserDataWriter(
-          this._firestore,
+          this._firestore._databaseId,
           this._firestore._areTimestampsInSnapshotsEnabled(),
-          options.serverTimestamps,
-          /* converter= */ undefined
+          options.serverTimestamps || 'none',
```
rafikhan, line 1358: Should this be done inside of `validateSnapshotOptions`?

schmidt-sebastian: I like the idea behind this and played around with it: https://gist.github.com/schmidt-sebastian/f1e1432128d2e536b26b0007b07f0ddb

Unfortunately, I think we should not do this. We pass `options` to the converter (in line 1353) and it would surprising to the user if we filled in a default value.

PR #3110, opened 2020-05-26
Allow shared usage of WriteBatch

schmidt-sebastian:
This allows the Lite SDK to use the existing WriteBatch class by removing the dependency on Firestore. It allows removes some unused code in `invokeCommitRpc` that makes constructing WriteBatch from the Lite SDK more cumbersome since we would have to suppress the return type.

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -262,9 +262,18 @@ class FirestoreSettings {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
FYI: I dropped this PR as it may be more trouble than its worth. It might be better to not share code here, as the API input types are not well aligned.

PR #3114, opened 2020-05-27
Allow shared usage of Transaction

schmidt-sebastian:
This PR:

- Removes the need to round-trip through RemoteStore to create a Transaction.
- Removes the SortedMap dependency in Transaction to reduce code size in the Lite SDK.

### DIFF
--- a/packages/firestore/src/core/transaction.ts
+++ b/packages/firestore/src/core/transaction.ts
@@ -16,7 +16,6 @@
  */
 
 import { ParsedSetData, ParsedUpdateData } from '../api/user_data_reader';
-import { documentVersionMap } from '../model/collections';
 import { Document, MaybeDocument, NoDocument } from '../model/document';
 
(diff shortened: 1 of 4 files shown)


packages/firestore/src/core/transaction.ts
```diff
@@ -132,8 +133,8 @@ export class Transaction {
       throw fail('Document in a transaction was a ' + doc.constructor.name);
     }
 
-    const existingVersion = this.readVersions.get(doc.key);
-    if (existingVersion !== null) {
+    const existingVersion = this.readVersions.get(doc.key.toString());
+    if (existingVersion) {
```
rafikhan, line 137: Why did you drop the null check? Is `get` guaranteed to return an object or null?

schmidt-sebastian: Maps return `undefined` for non-existing values and I prefer `if(value)` over `if(value !=== undefined)`.

PR #3119, opened 2020-05-27
docs: updating initializeApp() TypeScript @example

EvanLovely:
Hi! Noticed that the `initializeApp()` TypeScript popups that have example code were using `applicationId` instead of `appId` so this updates them! Thanks for the awesome project!


EvanLovely:
@googlebot I signed it!

Feiyang1:
Thanks for the contribution!

PR #3122, opened 2020-05-27
Add updateProfile, updateEmail, updatePassword

samhorlbeck:
This change also updates the way the auth object handles notifications and persistence. This new way more closely matches the behavior of the JS sdk.

### DIFF
--- a/packages-exp/auth-exp/src/core/user/account_info.test.ts
+++ b/packages-exp/auth-exp/src/core/user/account_info.test.ts
@@ -24,13 +24,13 @@ import { UserInfo } from '@firebase/auth-types-exp';
 
 // import { UserInfo } from '@firebase/auth-types-exp';
 import { mockEndpoint } from '../../../test/api/helper';
-import { testUser } from '../../../test/mock_auth';
+import { testPersistence, testUser } from '../../../test/mock_auth';
 import * as fetch from '../../../test/mock_fetch';
 import { Endpoint } from '../../api';
+import { Auth } from '../../model/auth';
 import { User } from '../../model/user';
 import { ProviderId } from '../providers';
-// import { ProviderId } from '../providers';
-import { updateProfile } from './account_info';
+import { updateEmail, updatePassword, updateProfile } from './account_info';
 
 use(chaiAsPromised);
 use(sinonChai);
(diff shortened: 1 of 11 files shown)


packages-exp/auth-exp/src/core/user/account_info.test.ts
```diff
@@ -0,0 +1,261 @@
...
+ */
+
+import { expect, use } from 'chai';
+import * as chaiAsPromised from 'chai-as-promised';
+import * as sinon from 'sinon';
+import * as sinonChai from 'sinon-chai';
+
+import { ProviderId, UserInfo } from '@firebase/auth-types-exp';
+
+// import { UserInfo } from '@firebase/auth-types-exp';
```
avolkovi, line 25: i think you meant to delete this

packages-exp/auth-exp/src/core/user/account_info.ts
```diff
@@ -0,0 +1,107 @@
...
+interface Profile {
+  displayName?: string | null;
+  photoURL?: string | null;
+}
+
+export async function updateProfile(
+  externUser: externs.User,
+  { displayName, photoURL: photoUrl }: Profile
+): Promise<void> {
+  if (displayName === undefined && photoUrl === undefined) {
```
avolkovi, line 37: does this seem right? I can't clear both of them?

packages-exp/auth-exp/src/core/user/account_info.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (password) {
+    request.password = password;
+  }
+
+  const response = await apiUpdateEmailPassword(auth, request);
+
+  const tokensRefreshed = user._updateTokensIfNecessary(response);
+  await _reloadWithoutSaving(user);
+  await auth._persistUserIfCurrent(user);
```
avolkovi, line 103: is this reloadWithoutSaving && perististUserIfCurrent pattern common enough to warrant having it's own method?

packages-exp/auth-exp/src/model/user.d.ts
```diff
@@ -32,8 +37,9 @@ export interface User extends externs.User {
   refreshToken: string;
   emailVerified: boolean;
   tenantId: string | null;
-  providerData: externs.UserInfo[];
+  providerData: ModifiableUserInfo[];
```
avolkovi, line 40: 👍 but this should probably be called MutableUserInfo

packages-exp/auth-exp/src/core/user/account_info.ts
samhorlbeck: Undefined doesn't clear; undefined just leaves the field be (no-op). `null` specifically clears. See the comment in the code example here: https://firebase.google.com/docs/reference/js/firebase.User#updateprofile

packages-exp/auth-exp/src/core/user/account_info.ts
samhorlbeck: I don't *think* so.. In general that case is covered by the standard `reload()`. We're doing this like this here because we want to only do the notification at the very end

packages-exp/auth-exp/src/model/user.d.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/account_info.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/user/account_info.ts
avolkovi: let's keep an eye out for common patterns like this, we can always refactor them away later to by DRY

PR #3123, opened 2020-05-27
Add DocumentReference

schmidt-sebastian:
This adds the DocumentReference type to the Lite SDK. 

Inspired by https://github.com/firebase/firebase-js-sdk/blob/4a70e17f009ea6ab4949178b64c9bbeb68c88af5/packages/firestore/src/api/database.ts#L937

### DIFF
--- a/packages/firestore/src/api/user_data_reader.ts
+++ b/packages/firestore/src/api/user_data_reader.ts
@@ -50,6 +50,15 @@ import { PlatformSupport } from '../platform/platform';
 const RESERVED_FIELD_REGEX = /^__.*__$/;
 
 /**
+ * An untyped Firestore Data Converter interface that is shared between the
+ * lite, full and legacy SDK.
+ */
+export interface UntypedFirestoreDataConverter<T> {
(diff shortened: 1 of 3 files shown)


packages/firestore/src/api/user_data_reader.ts
```diff
@@ -50,6 +50,29 @@ import { DocumentReference } from './database';
...
+  fromFirestore(snapshot: unknown, options?: unknown): T;
+}
+
+/**
+ * A reference to a document in a Firebase project.
+ *
+ * This class serves as a common base class for the public DocumentReferences
+ * exposed in the lite, full and legacy SDK.
+ */
+export class DocumentKeyReference<T> {
```
schmidt-sebastian, line 68: This is the DocumentKeyReference type from the other PR, but it takes an `UntypedFirestoreDataConverter` so it can be shared. While that's a bit of a hack, it might be okay since it is purely a user type that we don't rely on internally.

schmidt-sebastian: I merged in master.

PR #3127, opened 2020-05-28
Run integration/firestore tests for Firestore code changes

schmidt-sebastian:
@thebrianchen noticed that one of our integration tests that only runs as part of integration/firestore broke. It looks like this was possible because we did not run those tests as part of Firestore changes.


schmidt-sebastian:
Tested with a change in the firestore package:

```
Running tests in:
integration/browserify (always runs)
integration/firebase-typings (always runs)
integration/typescript (always runs)
integration/webpack (always runs)
integration/firestore (depends on modified files)
packages/firestore (contains modified files)
packages/firebase (depends on modified files)
```

PR #3129, opened 2020-05-28
Refactor so every unit test gets its own Auth object

samhorlbeck:
This is just the way it should be. Every auth object also gets its own mock persistence, so that we can inspect interactions with the persistence layer.


samhorlbeck:
Note I also renamed everything from mockAuth -> auth, since now it's no longer actually a mock

packages-exp/auth-exp/src/core/persistence/in_memory.test.ts
```diff
@@ -37,7 +37,7 @@ describe('core/persistence/in_memory', () => {
 
   it('should work with user', async () => {
     const key = 'my-super-special-user';
-    const value = testUser('uid');
+    const value = testUser({} as any, 'uid');
```
avolkovi, line 40: can these by typed? why do we cast to 'as any'?

packages-exp/auth-exp/src/core/strategies/credential.test.ts
```diff
@@ -15,15 +15,18 @@
  * limitations under the License.
  */
 
-import { OperationType, ProviderId, SignInMethod } from '@firebase/auth-types-exp';
 import { expect, use } from 'chai';
 import * as chaiAsPromised from 'chai-as-promised';
+
+import { OperationType, ProviderId, SignInMethod } from '@firebase/auth-types-exp';
```
avolkovi, line 21: we should make sure we all use the same vscode settings, I feel like they are not consistent across the team

packages-exp/auth-exp/test/mock_auth.ts
```diff
@@ -27,8 +29,29 @@ export const TEST_AUTH_DOMAIN = 'localhost';
...
+    return super.set(key, object);
+  }
+
+  remove(key: string): Promise<void> {
+    this.lastObjectSet = null;
+    return super.remove(key);
+  }
+}
+
+export async function testEnvironment(): Promise<TestEnvironment> {
```
avolkovi, line 51: this could be a lot simpler if we just had a persistence accessor on auth, this was a FR anyway and not a bad time to implement, then you wouldn't need this whole testenvironment construct

packages-exp/auth-exp/src/core/strategies/credential.test.ts
samhorlbeck: We probably should but in this case it's because I forgot to push twice (after the formatter ran)

packages-exp/auth-exp/src/core/persistence/in_memory.test.ts
samhorlbeck: These values get immediately turned into plain objects. I explicitly don't want these to have an auth object. I've changed the type on testUser to avoid the cast to any

packages-exp/auth-exp/test/mock_auth.ts
samhorlbeck: Done-ish. I don't want to expose too many details of the persistence layer (there's the whole persistenceUserManager between the actual persistence and auth). Still, I've updated the auth object returned by the method to include the MockPersistence instance.

PR #3130, opened 2020-05-28
Add DocumentSnapshot

schmidt-sebastian:
This will be tested when getDoc() is added.


schmidt-sebastian:
I accidentally merged https://github.com/firebase/firebase-js-sdk/pull/3137 into this (I thought it was off of master). It's tiny so I will leave as is.

packages/firestore/lite/src/api/field_path.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    for (let i = 0; i < fieldNames.length; ++i) {
+      if (fieldNames[i].length === 0) {
+        throw new FirestoreError(
+          Code.INVALID_ARGUMENT,
+          `Invalid field name at argument $(i + 1). ` +
+            'Field names must not be empty.'
+        );
+      }
+    }
```
rafikhan, line 65: What version of ecmascript do we target? Seems like this could be done idiomatically with something like `if (fieldNames.some(fn => fn.length === 0)) { throw }`.

Also, out of curiosity I noticed the use of `++i` vs the more common `i++`. No need to change it but was wondering why this choice.

packages/firestore/lite/src/api/snapshot.ts
```diff
@@ -0,0 +1,125 @@
...
+    return this._document !== null;
+  }
+
+  data(): T | undefined {
+    if (!this._document) {
+      return undefined;
+    } else {
+      // We only want to use the converter and create a new DocumentSnapshot
+      // if a converter has been provided.
+      if (this._converter) {
```
rafikhan, line 66: Out of curiosity (i.e. no need to change), why create the nested block? Why not write it as `} else if (this._converter) {` ?

PR #3132, opened 2020-05-28
Remove support for lastStreamToken

schmidt-sebastian:
This extracts the Stream Token removal from https://github.com/firebase/firebase-js-sdk/pull/2995 into a separate PR, so it can be more easily reverted should the backend add full support for the stream tokens.

We are removing the Stream Tokens to ease IndexedDB recovery.

Proof of concept PR that implements IndexedDB recovery with Stream Tokens: https://github.com/firebase/firebase-js-sdk/pull/3131

### DIFF
--- a/packages/firestore/src/remote/remote_store.ts
+++ b/packages/firestore/src/remote/remote_store.ts
@@ -18,7 +18,7 @@
 import { SnapshotVersion } from '../core/snapshot_version';
 import { Transaction } from '../core/transaction';
 import { OnlineState, TargetId } from '../core/types';
-import { ignoreIfPrimaryLeaseLoss, LocalStore } from '../local/local_store';
+import { LocalStore } from '../local/local_store';
 import { TargetData, TargetPurpose } from '../local/target_data';
 import { MutationResult } from '../model/mutation';
 import {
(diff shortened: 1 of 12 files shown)


packages/firestore/src/remote/remote_store.ts
```diff
@@ -691,46 +678,17 @@ export class RemoteStore implements TargetMetadataProvider {
...
-      // The write stream might have been started by refilling the write
-      // pipeline for failed writes
-      if (this.shouldStartWriteStream()) {
-        this.startWriteStream();
-      }
+    // If the write stream closed after the write handshake completes, a write
+    // operation failed and we fail the pending operation.
+    if (error && this.writeStream.handshakeComplete) {
+      // This error affects the actual write.
+      await this.handleWriteError(error!);
```
wilhuff, line 685: We're still tracking the `lastStreamToken` and it's possible that if we're restarting the write stream after some successful writes, the server could reject the stream token we supply. This means we still need to handle the handshake error.

schmidt-sebastian: I'll add the handshake code back if you prefer (it might be safer to reduce the churn here), but what do you think about the fix in b530cd58d?

wilhuff: I have no strong preference on the general strategy, though it seems like clearing when the stream starts could be more bullet proof.

schmidt-sebastian: I modified it to match the assignment of `handshakeComplete_`, which is now true only if the stream token is non empty.

PR #3136, opened 2020-05-29
Add signInWithCustomToken implementation

samhorlbeck:
I've also added an internal UserCredential type so we can avoid some casts.


packages-exp/auth-exp/src/model/user.d.ts
```diff
@@ -47,3 +47,7 @@ export interface User extends externs.User {
   delete(): Promise<void>;
   toPlainObject(): PersistedBlob;
 }
+
+export interface UserCredential extends externs.UserCredential {
+  user: User;
```
avolkovi, line 52: 👍

PR #3138, opened 2020-05-29
Add deleteDoc()

schmidt-sebastian:


packages/firestore/src/remote/datastore.ts
```diff
@@ -106,20 +106,13 @@ export function newDatastore(
...
   const datastoreImpl = debugCast(datastore, DatastoreImpl);
   const params = {
     database: datastoreImpl.serializer.encodedDatabaseId,
     writes: mutations.map(m => datastoreImpl.serializer.toMutation(m))
   };
-  const response = await datastoreImpl.invokeRPC<
-    api.CommitRequest,
-    api.CommitResponse
-  >('Commit', params);
-  return datastoreImpl.serializer.fromWriteResults(
```
schmidt-sebastian, line 119: This result is never used - but it would force me to ignore it in every callsite if we do provide it.

PR #3139, opened 2020-05-29
Add setDoc()

schmidt-sebastian:
Also adds a lot of tests for existing code and a shared FieldValue class to be used by all SDK versions (comments inline that explain it).


packages/firestore/lite/src/api/reference.ts
```diff
@@ -300,6 +306,40 @@ export function getDoc<T>(
...
+  const dataReader = newUserDataReader(
+    ref.firestore._databaseId,
+    ref.firestore._settings!
+  );
+
+  const parsed = isMerge(options)
+    ? dataReader.parseMergeData('setDoc', convertedValue)
+    : isMergeFields(options)
+    ? dataReader.parseMergeData('setDoc', convertedValue, options.mergeFields)
+    : dataReader.parseSetData('setDoc', convertedValue);
```
rafikhan, line 333: Optional: I like the helper methods you made to interrogate the options but the ternary required effort to parse. Consider factoring this out into a helper function that includes the `toMutations()` call or perhaps convert this to a series of if/then/else.

schmidt-sebastian: I have been bothered by this code for a while, as it is repeated in a couple of places in the SDK. I changed UserDataReader.parseSetData() to perform this check on behalf of the caller.

PR #3148, opened 2020-06-02
Add updateDoc()

schmidt-sebastian:
See https://github.com/firebase/firebase-js-sdk/blob/554786165bf9289db122c5136de446882f4880b9/packages/firestore/src/api/database.ts#L1068


packages/firestore/lite/src/api/reference.ts
```diff
@@ -306,6 +307,15 @@ export function getDoc<T>(
...
 
+export function setDoc<T>(
+  reference: firestore.DocumentReference<T>,
+  data: T
+): Promise<void>;
+export function setDoc<T>(
+  reference: firestore.DocumentReference<T>,
+  data: Partial<T>,
+  options: firestore.SetOptions
+): Promise<void>;
```
schmidt-sebastian, line 318: Adding these overloads allows us to simulator more realtime usage in the integration tests as it removes the availability of the "catch all" API.

PR #3150, opened 2020-06-02
Add tslint for sorting imports.

wu-hui:
Two manually changed files are: tslint.json and package.json.


rafikhan:
LGTM but this should also be reviewed by @Feiyang1

Feiyang1:
I didn't realize it is just for `Firestore`. It's fine to merge as is, but If you don't mind, could you apply it to the entire repo (don't feel obliged to do it though, I can make another PR instead)?

packages/firestore/.eslintrc.js
```diff
@@ -27,6 +27,21 @@ module.exports = {
...
+        'groups': [
+          'builtin',
+          'external',
+          'internal',
+          'parent',
+          'sibling',
+          'index'
+        ],
+        'newlines-between': 'always',
+        'alphabetize': { 'order': 'asc', 'caseInsensitive': false }
```
schmidt-sebastian, line 42: My preference would be to make our import order case insensitive.

packages/firestore/exp/src/api/database.ts
```diff
@@ -15,21 +15,19 @@
...
+import { Firestore as LiteFirestore } from '../../../lite/src/api/database';
 import {
   ComponentProvider,
   MemoryComponentProvider
 } from '../../../src/core/component_provider';
-
-import { Firestore as LiteFirestore } from '../../../lite/src/api/database';
+import { FirestoreClient } from '../../../src/core/firestore_client';
+import { AsyncQueue } from '../../../src/util/async_queue';
+import * as firestore from '../../index';
```
schmidt-sebastian, line 30: Is it possible to move the namespace imports to the top? (Doesn't seem like it though).

packages/firestore/.eslintrc.js
wu-hui: Done.

packages/firestore/exp/src/api/database.ts
wu-hui: I could not find that config.

schmidt-sebastian:
I'm still looking at this, but the test failures reproduce for me locally:

```
cd firebase-js-sdk;
yarn build;
cd integration/firestore;
yarn test:persistence

 Uncaught TypeError: Object prototype may only be an Object or null: undefined
  at dist/test-harness.js:11520:9
  
  TypeError: Object prototype may only be an Object or null: undefined
      at setPrototypeOf (<anonymous>)
      at __extends (dist/test-harness.js:11520:9)
```

schmidt-sebastian:
It sounds like the error is due to our usage of circular imports: https://stackoverflow.com/questions/53122751/typeerror-object-prototype-may-only-be-an-object-or-null-undefined/53123468

The fix for this is probably out of scope for this PR.

PR #3151, opened 2020-06-02
Transaction/WriteBatch signature fix

schmidt-sebastian:
Updates types in firestore-next client.

Matches `setDoc`.


schmidt-sebastian:
Yes, we need to make the change here too, but I wanted to wait for the new signature to be supported by the current SDK. You could also update the new typings files in your PR.

PR #3156, opened 2020-06-03
Do not build firestore lite in build because it breaks regular release build

Feiyang1:
Script `build` should only build regular packages


schmidt-sebastian:
I added this because not having it breaks the binary size presubmit, since the presubmit runs based on the package.json in `/lite`.

PR #3157, opened 2020-06-03
Add terminate() to Lite SDK

schmidt-sebastian:
Added the terminate logic to Datastore instead of FIrestore to ensure that WriteBatches cannot be committed after terminate is called.


packages/firestore/lite/test/integration.test.ts
```diff
@@ -67,6 +71,26 @@ describe('Firestore', () => {
       'Firestore has already been started and its settings can no longer be changed.'
     );
   });
+
+  it('cannot use once terminated', () => {
```
wu-hui, line 75: I'd like to suggest to add a test to verify calls to `terminate()` after it is terminated are simply no-op. 

It's obvious now, but might get over looked when the code evolves.

schmidt-sebastian: Good idea. Added.

PR #3162, opened 2020-06-04
Handle IndexedDB errors in unsubscribe callback

schmidt-sebastian:
Addresses one of the three problems I found during randomized testing.

Addresses #2755


packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -399,8 +399,29 @@ abstract class TestRunner {
...
     debugAssert(!!eventEmitter, 'There must be a query to unlisten too!');
-    this.queryListeners.delete(query);
-    await this.queue.enqueue(() => this.eventManager.unlisten(eventEmitter!));
+
+    const deferred = new Deferred<void>();
+    await this.queue.enqueueRetryable(async () => {
+      try {
+        await this.eventManager.unlisten(eventEmitter!);
+        const currentEventEmitter = this.queryListeners.get(query);
+        // Before removing the listener, verify that the query hasn't been
```
wilhuff, line 408: I must be missing some context for this comment. If a user attempts to unlisten, that fails, gets in the retry queue, and then they listen to the same query while we're retrying, doesn't the event manager have two listeners for that query? The first query will stil succeed in unlistening.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -404,7 +404,7 @@ export class FirestoreClient {
     if (this.clientTerminated) {
       return;
     }
-    this.asyncQueue.enqueueAndForget(() => {
+    this.asyncQueue.enqueueRetryable(() => {
```
wilhuff, line 407: Why is this necessary? I thought that if we failed we tore down the streams? At that point haven't we effectively unlistened anyway? All that's missing is updating our in-memory bookkeeping to indicate that we don't want to restart the target when IndexedDB becomes available again.

One other concern is that if IndexedDB is failing we've likely been put in the background and may not come back for a while--possibly more than the 30 minutes for which a resume token means free query resumption. If we were asleep that long the user would be charged for the query anew--for a query they were trying to stop listening to. If all we're losing is the resume token and the sequence numbers for exact GC ordering, it seems better to let the unlisten succeed despite failing to write. 

So in the primary tab, if we fail trying to save metadata data about unlistening that's not really a big deal. Can a secondary tab fail trying to tell the primary that it wants to unlisten? It's that case we should probably retry.

If this squares with your understanding, it seems like this line may be accurate: we do want to retry unlistens but only when it's the secondary tab asking the primary tab to do it. Primary tabs should just succeed.

It seems like unlisten should succeed regardless of the indexeddb failure.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: `queryListeners` is a map of active listeners that the Spec tests care about. This code here was my first attempt at making sure that we don't remove a query from the Spec tests view if another operation has re-registered it. I was able to simplify this though by removing the query directly and outside of the queue, which should have no effect on the spec tests since this step would have previously executed in a very similar order.

packages/firestore/src/core/firestore_client.ts
schmidt-sebastian: This code path calls into LocalStore to invoke "Release Target" and then calls RemoteStore. We could change the plumbing here to drive this from RemoteStore, in which case RemoteStore could go offline and do its own retry. That sounds a like a fair bit of code churn that may lead to some unexpected side effects - the mapping between the `lastListen` state in EventManager and the active list of targets in SyncEngine/LocalStore may go out of sync.

One easy thing we could do here is just swallow IndexedDB failures if "Release Target" fails, which would even remove some of the newly added error handling in RemoteStore. Do you think that's worth exploring?

wilhuff: > just swallow IndexedDB failures if "Release Target" fails

This is what I was suggesting: releasing a target in the LocalStore is just about making sure the final updates are made for GC metadata (and the in-memory tracking of the target). It may be that even LocalStore itself may need to change so that even if the transaction fails it will remove the target from its own set of live things.

I'm not sure this undoes the error handling in RemoteStore, since it's really only unsubscribe can work this way.

schmidt-sebastian: I updated the PR to ignore errors in "Release target". Unfortunately, since a failed Target can generate a "fake" Remote Event (for Limbo resolutions), the code path that calls into this from RemoteStore still has to use the IndexedDB fallback.

packages/firestore/src/local/local_store.ts
```diff
@@ -922,21 +922,32 @@ class LocalStoreImpl implements LocalStore {
...
-      .then(() => {
-        this.targetDataByTarget = this.targetDataByTarget.remove(targetId);
-        this.targetIdByTarget.delete(targetData!.target);
-      });
+        });
+      }
+    } catch (e) {
+      if (isIndexedDbTransactionError(e)) {
+        // If `releaseTarget` fails, we did not advance the sequence number of
+        // the target. While the target might be deleted earlier than it
```
wilhuff, line 938: Optional: since the degree of "earlier" is very slight (we already update the sequence number when you start listening and during updates so this is only the final update on unlisten), consider softening the wording here or giving more context. Suggestion:

> All `releaseTarget` does is advance record final metadata state for the target, but we've been recording this periodically during target activity. If we lose this write this could cause a very slight difference in the order of target deletion during GC, but we don't define exact LRU semantics anyway so this is acceptable.

schmidt-sebastian: Done. Thanks for the rewrite :)

Issue #2393, opened 2019-11-30
Unhandled Rejection (SecurityError): IDBFactory.open() called in an invalid security context in Analytics


tomsun:
Application crashes when `firebase.analytics();` is called from within an iframe in macOS Safari.

Regression of or otherwise related to https://github.com/firebase/firebase-js-sdk/issues/631 ?

```
import * as firebase from 'firebase/app';
import 'firebase/analytics';
const app = firebase.initializeApp({...);
const analytics = firebase.analytics();
```

---

  * Operating System version: macOS Mojave
  * Browser version: Safari 13.0.3
  * Firebase SDK version: firebase@7.5.0
  * Firebase Product: @firebase/analytics@0.2.7

rommelpe:
Thanks for reporting the issue. Not sure if I'm missing something here as I can't replicate it on end. Can you please include a more [detailed repro](https://stackoverflow.com/help/minimal-reproducible-example) so we can help you debug the issue?

tomsun:
@rommelpe I see, it only happens in an *cross-domain*-loaded iframe apparently.

Here is a minimal failing example: https://github.com/tomsun/firebase-js-sdk-securityerror-example

It uses webpack-devserver to host a transpiled build of the example above plus firebase config for an analytics-enabled example project plus html files for parent and iframe. Served on localhost:8080.

To trigger the cross-domain scenario, the iframe is loaded from http://local.codemines.org:8080 instead, which resolves to localhost.

<img src="https://raw.githubusercontent.com/tomsun/firebase-js-sdk-securityerror-example/master/result.png" />

rommelpe:
Thanks for working on the reproducible code. It might be related to the one you mentioned (https://github.com/firebase/firebase-js-sdk/issues/631) as the issue only occurs in Safari, it works well in Firefox v70.0 & Chrome v78.0. I created an internal report (b/145587760) to confirm this behavior, and progress updates will be posted here as well.

tomsun:
thanks!

hsubox76:
Still taking a look but it seems like the issue is that analytics makes a call to the installations library for the Firebase instance ID (to associated the analytics data with the Firebase instance) and installations uses IndexedDB, which is restricted in the specific case of cross-origin iframes in Safari. I don't have an immediate solution or workaround unless you can change one of those criteria, but I think we've narrowed the problem down and will keep working on it.

amardeepsingh20:
[...]

PR #3165, opened 2020-06-04
Issue 2393 - Add environment check to Analytics Module

XuechunHou:
a fix for issue #2393 for analytics module. 

Throw runtime error (UNSUPPORTED_BROWSER) on initialization of the instance when IndexedDB is not supported in some browser situations.

The fix is tested on Safari and Firefox private window rendering iframe with https://github.com/tomsun/firebase-js-sdk-securityerror-example provided in the issue conversation.


packages/firebase/index.d.ts
```diff
@@ -5083,6 +5083,7 @@ declare namespace firebase.analytics {
     id?: string;
     name?: string;
   }
+  function isSupported(): boolean;
```
hsubox76, line 5086: Will need a comment here to document what this does. Comments in this file are the source of our official documentation, see other methods/properties for format.

packages/analytics/index.ts
```diff
@@ -97,8 +102,72 @@ registerAnalytics(firebase as _FirebaseNamespace);
...
+ * This method checks if indexedDB is supported by current browser and throws AnalyticsError.INDEXED_DB_UNSUPPORTED error if not.
+ *
+ * This method also validates browser context for indexedDB by opening a dummy indexedDB database and throws AnalyticsError.INVALID_INDEXED_DB_CONTEXT
+ * if errors occur during the database open operation.
+ */
+function validateIndexedDBSupport(): void {
+  if ('indexedDB' in window && indexedDB !== null) {
+    try {
+      let preExist: boolean = true;
+      const DUMMYDBNAME =
```
hsubox76, line 131: Maybe `DB_CHECK_NAME`

packages/analytics/src/errors.ts
```diff
@@ -39,12 +42,18 @@ const ERRORS: ErrorMap<AnalyticsError> = {
     'settings() must be called before initializing any Analytics instance' +
     'or it will have no effect.',
   [AnalyticsError.INTEROP_COMPONENT_REG_FAILED]:
-    'Firebase Analytics Interop Component failed to instantiate'
+    'Firebase Analytics Interop Component failed to instantiate',
+  [AnalyticsError.INDEXED_DB_UNSUPPORTED]:
+    'IndexedDB is not supported by current browswer',
+  [AnalyticsError.INVALID_INDEXED_DB_CONTEXT]:
+    "Environment doesn't support IndexedDB functionality: {$errorInfo}",
+  [AnalyticsError.COOKIE_NOT_ENABLED]: 'Browser cookie not enabled'
```
hsubox76, line 50: Suggest "Cookies not enabled in this browser environment. Analytics requires cookies to be enabled." or something similar?

packages/analytics/src/factory.ts
```diff
@@ -118,9 +122,19 @@ export function factory(
...
   installations: FirebaseInstallations
 ): FirebaseAnalytics {
+  if (!areCookiesEnabled()) {
+    throw ERROR_FACTORY.create(AnalyticsError.COOKIES_NOT_ENABLED);
+  }
+  if (!isIndexedDBAvailable()) {
+    throw ERROR_FACTORY.create(AnalyticsError.INDEXED_DB_UNSUPPORTED);
+  }
   // Async but non-blocking.
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
hsubox76, line 132: Is eslint ok now if we get rid of this comment?

PR #3167, opened 2020-06-05
Enable bundler reader for Node.

wu-hui:


packages/firestore/src/platform/platform.ts
```diff
@@ -60,6 +65,54 @@ export interface Platform {
   readonly base64Available: boolean;
 }
 
+/**
+ * An interface compatible with Web's ReadableStream.getReader() return type.
+ *
+ * This can be used as an abstraction to mimic `ReadableStream` where it is not
+ * available.
+ */
+export interface ByteStreamReader {
```
schmidt-sebastian, line 74: Did you take a look at using the type `ReadableStreamReader` directly?

```
interface ReadableStreamReader<R = any> {
    cancel(): Promise<void>;
    read(): Promise<ReadableStreamReadResult<R>>;
    releaseLock(): void;
}
```

(from lib.dom.d.ts)

packages/firestore/src/platform/platform.ts
```diff
@@ -60,6 +65,54 @@ export interface Platform {
...
+ * @param source The data source to use.
+ * @param bytesPerRead How many bytes each `read()` from the returned reader
+ *        will read.
+ */
+export function toByteStreamReader(
+  source: Uint8Array,
+  bytesPerRead = 10240
+): ByteStreamReader {
+  let readFrom = 0;
+  return new (class implements ByteStreamReader {
```
schmidt-sebastian, line 98: I assume you are using the `IIFE` pattern here to enforce the `ByteStreamReader` type?

You could just use:

const reader : ByteStreamReader = {
  ...
}
return reader;

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -93,4 +98,33 @@ export class BrowserPlatform implements Platform {
...
+  ): ByteStreamReader {
+    if (source instanceof Uint8Array) {
+      return toByteStreamReader(source);
+    }
+    if (source instanceof ArrayBuffer) {
+      return toByteStreamReader(new Uint8Array(source));
+    }
+    if (source instanceof ReadableStream) {
+      const reader = source.getReader();
+      return new (class implements ByteStreamReader {
```
schmidt-sebastian, line 116: Can we just return `reader` here?

packages/firestore/src/platform_browser/browser_platform.ts
```diff
@@ -93,4 +98,33 @@ export class BrowserPlatform implements Platform {
...
+          return reader.read();
+        }
+
+        cancel(reason?: string): Promise<void> {
+          return reader.cancel(reason);
+        }
+      })();
+    }
+    throw new Error(
+      'Source of `toByteStreamReader` has to be Uint8Array, ArrayBuffer or ReadableStream'
```
schmidt-sebastian, line 127: Missing `a` ("has to be a Uint8Array")

packages/firestore/src/platform/platform.ts
```diff
@@ -50,6 +50,11 @@ export interface Platform {
    */
   randomBytes(nBytes: number): Uint8Array;
 
+  /**
+   * Builds a `ByteStreamReader` from a data source.
+   */
+  toByteStreamReader(source: unknown): ByteStreamReader;
```
schmidt-sebastian, line 56: Can the source type be `Uint8Array | ArrayBuffer | ReadableStream<Uint8Array>`?

And since this type combination is used widely now, should we add a local type alias?

packages/firestore/src/platform_node/node_platform.ts
```diff
@@ -83,4 +87,14 @@ export class NodePlatform implements Platform {
 
     return randomBytes(nBytes);
   }
+
+  /**
+   * On Node, only supported data source is a `Uint8Array` for now.
+   */
+  toByteStreamReader(source: unknown): ByteStreamReader {
+    if (source instanceof Uint8Array) {
```
schmidt-sebastian, line 95: I suggest you use this pattern here: https://github.com/firebase/firebase-js-sdk/blob/e67affba53a53d28492587b2f60521a00166db60/packages/firestore/src/api/blob.ts#L81

If it doesn't apply here, then you should invoke it at the callsite (the API that the end user calls) and add a `debugAssert` here.

packages/firestore/test/unit/platform/platform.test.ts
```diff
@@ -16,10 +16,34 @@
...
+} from '../../../src/platform/platform';
 
 describe('Platform', () => {
   it('can load the platform at runtime', () => {
     expect(PlatformSupport.getPlatform()).to.exist;
   });
+
+  it('toByteStreamReader() steps underlying data', async () => {
+    const encoder = new TextEncoder();
+    const r = toByteStreamReader(encoder.encode('0123456789'), 4);
```
schmidt-sebastian, line 31: Add inline comment to explain what `4` is.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -17,17 +17,38 @@
...
+ * @param bytesPerRead: How many bytes to read from the underlying buffer from
+ * each read through the stream.
+ */
+export function readableStreamFromString(
   content: string,
-  bytesPerRead: number
-): ReadableStream {
-  return toReadableStream(new TextEncoder().encode(content), bytesPerRead);
+  bytesPerRead = 10240
+): ReadableStream<Uint8Array | ArrayBuffer> {
```
schmidt-sebastian, line 39: Would it be possible to convert the input to a `UInt8Array` and re-use toByteStreamReader?

packages/firestore/test/util/test_platform.ts
```diff
@@ -270,6 +270,10 @@ export class TestPlatform implements Platform {
   randomBytes(nBytes: number): Uint8Array {
     return this.basePlatform.randomBytes(nBytes);
   }
+
+  toByteStreamReader(source: unknown): ByteStreamReader {
+    return this.basePlatform.toByteStreamReader(source);
```
schmidt-sebastian, line 275: I suspect that if you move your Node-specifc `genericBundleReadingTests` logic here, then you can use the TestPlatform throughout and won't have to force override it.

packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -61,15 +84,31 @@ function lengthPrefixedString(o: {}): string {
...
 });
 
 function genericBundleReadingTests(bytesPerRead: number): void {
+  // On Node, we need to override `bytesPerRead` from it's platform's `toByteStreamReader` call.
+  if (isNode()) {
+    const platform = PlatformSupport.getPlatform();
+    platform.toByteStreamReader = source =>
+      toByteStreamReader(source as Uint8Array, bytesPerRead);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (PlatformSupport as any)._forceSetPlatform(platform);
```
schmidt-sebastian, line 101: I think you should try really hard to to have to do this. Options:

- Have a Node specific test file (bundle.node.test.ts similar to serializer.node.test.ts)
- Implement this logic in TestPlatform's `toByteStreamReader`.

A very strong argument against this pattern is that guaranteeing cleanup for this type of global state modification is hard. It even looks like you opted not do to it at all :)

packages/firestore/src/platform_node/node_platform.ts
wu-hui: Done.

packages/firestore/test/unit/platform/platform.test.ts
wu-hui: Done.

packages/firestore/test/unit/util/bundle.test.ts
wu-hui: Actually making `toByteStreamReader` take a `bytesPerRead` eliminates the problem.

packages/firestore/test/util/test_platform.ts
wu-hui: See above. It now works for both platforms without hacking.

PR #3168, opened 2020-06-05
Add log levels to type definition in firebase/index.d.ts

wu-hui:


Feiyang1:
@wu-hui Can I merge this PR?

PR #3170, opened 2020-06-05
Implement BundleCache for IDB and memory.

wu-hui:


packages/firestore/src/core/bundle.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Query } from './query';
+import { SnapshotVersion } from './snapshot_version';
+
+/**
+ * Represents a Firestore bundle saved by the SDK in its local storage.
+ */
+export interface Bundle {
+  readonly id: string;
+  readonly version: number;
+  // When the saved bundle is built from the server SDKs.
```
schmidt-sebastian, line 27: Can you point out in the documentation that this is to SnapshotVersion.MIN if the above is not true? In its current form, the comment raises the question why the type is not nullable.

packages/firestore/src/local/bundle_cache.ts
```diff
@@ -0,0 +1,60 @@
...
+   * Gets a saved `Bundle` for a given `bundleId`, returns undefined if
+   * no bundles are found under the given id.
+   */
+  getBundle(
+    transaction: PersistenceTransaction,
+    bundleId: string
+  ): PersistencePromise<Bundle | undefined>;
+
+  /**
+   * Saves a `BundleMetadata` from a bundle into local storage, using its id as the persistent key.
```
schmidt-sebastian, line 37: Nit: Wrap at 80 characters.

packages/firestore/src/local/indexeddb_bundle_cache.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  getBundle(
+    transaction: PersistenceTransaction,
+    bundleId: string
+  ): PersistencePromise<Bundle | undefined> {
+    return bundlesStore(transaction)
+      .get(bundleId)
+      .next(bundle => {
+        if (bundle) {
+          return this.serializer.fromDbBundle(bundle!);
```
schmidt-sebastian, line 44: You should be able to drop the bang.

packages/firestore/src/local/indexeddb_bundle_cache.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  getNamedQuery(
+    transaction: PersistenceTransaction,
+    queryName: string
+  ): PersistencePromise<NamedQuery | undefined> {
+    return namedQueriesStore(transaction)
+      .get(queryName)
+      .next(query => {
+        if (query) {
+          return this.serializer.fromDbNamedQuery(query!);
```
schmidt-sebastian, line 67: You should be able to drop the bang.

packages/firestore/src/local/local_serializer.ts
```diff
@@ -231,6 +236,81 @@ export class LocalSerializer {
...
+  }
+
+  /** Encodes a BundleMetadata proto object to a Bundle model object. */
+  fromBundleMetadata(metadata: bundleProto.BundleMetadata): Bundle {
+    return {
+      id: metadata.id!,
+      version: metadata.version!,
+      createTime: this.remoteSerializer.fromVersion(metadata.createTime!)
+    };
+  }
```
schmidt-sebastian, line 313: Ideally, we should not add any more code to the main SDK for "optional" features. Do you mind exploring what it would entail to make these functions free-standing? They would then look like:


function fromBundleMetadata(serializer: RemoteSerializer, metadata: bundleProto.BundleMetadata): Bundle {
    return {
      id: metadata.id!,
      version: metadata.version!,
      createTime: serializer.fromVersion(metadata.createTime!)
    };
  }
```

If you have to make a ton of state public, you can look at what we did here: https://github.com/firebase/firebase-js-sdk/blob/5cf39ab8cacd5f055d609acf79a859beda1032d3/packages/firestore/src/remote/datastore.ts#L51

DatastoreImp has public state, but in itself is private.

File: packages/firestore/src/core/bundle.ts [hunk 6]
```diff
@@ -0,0 +1,138 @@
...
+  readonly createTime: SnapshotVersion;
+}
+
+/**
+ * Represents a Query saved by the SDK in its local storage.
+ */
+export interface NamedQuery {
+  readonly name: string;
+  readonly query: Query;
+  // When the results for this query are read to the saved bundle.
```
schmidt-sebastian, line 40: `/** The time at which the results for this query were read. */`

packages/firestore/src/core/bundle.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * This reconstructs the original query used to build the bundle being loaded,
+ * including features exists only in SDKs (for example: limit-to-last).
+ */
+export function fromBundledQuery(
+  serializer: LocalSerializer,
+  bundledQuery: bundleProto.BundledQuery
+): Query {
+  const query = serializer.remoteSerializer.convertQueryTargetToQuery({
+    parent: bundledQuery.parent!,
```
schmidt-sebastian, line 107: Did you file a bug to validate the bundle contents? As part of this we need to validate that the parent matches the Firestore project that we are loading the SDK into.

packages/firestore/src/core/bundle.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/** Encodes a DbBundle to a Bundle. */
+export function fromDbBundle(
+  serializer: LocalSerializer,
+  dbBundle: DbBundle
+): Bundle {
+  return {
+    id: dbBundle.bundleId,
+    createTime: serializer.fromDbTimestamp(dbBundle.createTime),
+    version: dbBundle.version
```
schmidt-sebastian, line 52: What is our versioning strategy? I wonder if we should reject any bundle with a version greater than 1. If so, please file a bug.

packages/firestore/src/core/bundle.ts
```diff
@@ -0,0 +1,138 @@
...
+  };
+}
+
+/** Encodes a BundleMetadata to a DbBundle. */
+export function toDbBundle(
```
schmidt-sebastian, line 57: I don't have an actual opinion, but I am a bit surprised that you moved this function out of the serializer file. I think it makes sense that even the free-standing functions are next to the main serializer, but I can also see why you moved them to. Just wanted to know whether you considered this? 

No need to move them.

packages/firestore/test/unit/local/bundle_cache.test.ts
```diff
@@ -0,0 +1,207 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect } from 'chai';
+import { IndexedDbPersistence } from '../../../src/local/indexeddb_persistence';
+import { filter, orderBy, path } from '../../util/helpers';
+import * as persistenceHelpers from './persistence_test_helpers';
```
schmidt-sebastian, line 21: In general, I would suggest to avoid namespace imports (`import * as `).

packages/firestore/test/unit/local/bundle_cache.test.ts
```diff
@@ -0,0 +1,207 @@
...
+  if (!IndexedDbPersistence.isAvailable()) {
+    console.warn('No IndexedDB. Skipping IndexedDbBundleCache tests.');
+    return;
+  }
+
+  let cache: TestBundleCache;
+  let persistence: IndexedDbPersistence;
+  beforeEach(async () => {
+    persistence = await persistenceHelpers.testIndexedDbPersistence({
+      synchronizeTabs: true
```
schmidt-sebastian, line 52: I would omit this setting. It poses questions as to why you added it here, but doesn't (yet) at much benefit - other than remind me that I forgot to reply to your question about multi-tab sync...

packages/firestore/test/unit/local/bundle_cache.test.ts
wu-hui: Done.

packages/firestore/test/unit/local/bundle_cache.test.ts
wu-hui: Done.

PR #3171, opened 2020-06-05
Make Converter passing explicit

schmidt-sebastian:
One of the comments in the Transaction PR was about a missing converter argument. This PR should make these problems go away by no longer allowing `undefined`.

I also stumbled across a test Firestore that I can use instead of `fakeFirestore`.

### DIFF
--- a/packages/firestore/test/util/api_helpers.ts
+++ b/packages/firestore/test/util/api_helpers.ts
@@ -45,8 +45,8 @@ import { Provider, ComponentContainer } from '@firebase/component';
  */
 export const FIRESTORE = new Firestore(
   {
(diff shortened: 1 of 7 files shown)


packages/firestore/test/util/api_helpers.ts
```diff
@@ -45,8 +45,8 @@ import { Provider, ComponentContainer } from '@firebase/component';
  */
 export const FIRESTORE = new Firestore(
   {
-    projectId: 'projectid',
-    database: 'database'
+    projectId: 'test-project',
```
schmidt-sebastian, line 48: Changed so I can get rid of `fakeFirestore`.

PR #3173, opened 2020-06-05
add _removeServiceInstance() to app-exp

Feiyang1:


packages-exp/app-types-exp/index.d.ts
```diff
@@ -116,9 +116,15 @@ export interface _FirebaseService {
   delete(): Promise<void>;
 }
 
+export interface VersionService {
+  library: string;
+  version: string;
+}
```
schmidt-sebastian, line 122: Not sure what this is but rest LGTM.

Feiyang1: Some files was importing `@firebase/app-types` initially. Now I have changed them to `@firebase/app-types-exp`, so I need to define the type that was missing in `@firebase/app-types-exp`

PR #3176, opened 2020-06-06
Add equals methods

schmidt-sebastian:


packages/firestore/lite/test/integration.test.ts
```diff
@@ -785,3 +789,96 @@ describe('Query', () => {
     );
   });
 });
+
+describe('equality', () => {
+  it('for collection references', () => {
```
dconeybe, line 794: Consider adding a test that compares a `DocumentReference` to a `CollectionReference`, which should evaluate to `false`.

schmidt-sebastian: Done

packages/firestore/lite/src/api/reference.ts
```diff
@@ -302,6 +303,28 @@ export function collection(
   }
 }
 
+export function collectionGroup(
+  firestore: firestore.FirebaseFirestore,
+  collectionId: string
+): Query<firestore.DocumentData> {
+  const firestoreClient = cast(firestore, Firestore);
```
avolkovi, line 310: @samhorlbeck we should look into using this cast() helper

packages/firestore/lite/src/api/reference.ts
```diff
@@ -302,6 +303,28 @@ export function collection(
...
+
+  validateArgType('collectionGroup', 'non-empty string', 1, collectionId);
+  if (collectionId.indexOf('/') >= 0) {
+    throw new FirestoreError(
```
avolkovi, line 314: recommend using an ErrorFactory https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/errors.ts#L106

packages/firestore/lite/src/api/reference.ts
```diff
@@ -533,6 +556,38 @@ export function addDoc<T>(
     .then(() => docRef);
 }
 
+export function refEqual<T>(
+  left: firestore.DocumentReference<T> | firestore.CollectionReference<T>,
+  right: firestore.DocumentReference<T> | firestore.CollectionReference<T>
+): boolean {
+  if (
```
avolkovi, line 563: is this if() just to guard against invalid types from JS?

packages/firestore/lite/src/api/snapshot.ts
```diff
@@ -112,8 +113,8 @@ export class QueryDocumentSnapshot<T = firestore.DocumentData>
 export class QuerySnapshot<T = firestore.DocumentData>
   implements firestore.QuerySnapshot<T> {
   constructor(
-    readonly query: firestore.Query<T>,
-    private readonly _docs: Array<QueryDocumentSnapshot<T>>
+    public readonly query: firestore.Query<T>,
```
avolkovi, line 116: use `readonly` over `public readonly` http://go/ts-style#visibility

packages/firestore/lite/test/integration.test.ts
```diff
@@ -832,4 +836,134 @@ describe('Query', () => {
...
+      );
+      expect(() => collectionGroup(firestore, '/')).to.throw(
+        "Invalid collection ID '/' passed to function collectionGroup(). Collection IDs must not contain '/'."
+      );
+    });
+  });
+});
+
+describe('equality', () => {
+  it('for collection references', () => {
```
avolkovi, line 875: consider using `context()` blocks to setup test context, and then use `it()` blocks for expectations

```
describe('equality', () => {
  context('for collection references', () => {
   it("returns true when they are equal", () => {} );
   it("returns false when they are not equal", () => {});
  });
});
```

packages/firestore/lite/test/integration.test.ts
```diff
@@ -832,4 +836,134 @@ describe('Query', () => {
...
+    return withTestDb(firestore => {
+      const coll1a = collection(firestore, 'a');
+      const coll1b = parent(doc(firestore, 'a/b'));
+      const coll2 = collection(firestore, 'c');
+
+      expect(refEqual(coll1a, coll1b)).to.be.true;
```
avolkovi, line 881: this is really 4 test cases that you are conflating into one it() block, consider splitting them out

packages/firestore/lite/test/integration.test.ts
```diff
@@ -832,4 +836,134 @@ describe('Query', () => {
...
+        fromFirestore: snap => snap.data()
+      });
+      expect(refEqual(doc1a, doc1c)).to.be.false;
+
+      expect(refEqual(doc1a, collection(firestore, 'a'))).to.be.false;
+    });
+  });
+
+  it('for queries', () => {
+    return withTestCollectionAndInitialData(
```
avolkovi, line 914: is there a reason these test helpers can't live in a beforeEach() block?

packages/firestore/lite/test/integration.test.ts
schmidt-sebastian: I understand, but these test cases are very related and overall still small. Having a very large number of small tests can also become a nuisance. Fortunately, no one on our team has TypeScript readability, so we can decide on a case by case basis. Some of this might go away as we add "context" sections.

FYI We will probably rewrite all these tests once we have a version of the full Firestore API that uses tree-shaking. We should then use the same tests against all codebases.

packages/firestore/lite/test/integration.test.ts
schmidt-sebastian: These helper set up different state each time and make the state visible in the test code. This also matches existing Firestore integration tests.

packages/firestore/lite/test/integration.test.ts
schmidt-sebastian: I have never seen this context before. I will add a TODO. 

FYI I am planning to address all TODOs before releasing, but I want to avoid huge merge conflicts.

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: I was unaware of this Added TODO.

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: It also does the type casting which exposes the private API.

packages/firestore/lite/src/api/snapshot.ts
schmidt-sebastian: Done (here and everywhere)

PR #3177, opened 2020-06-06
Add collectionGroup()

schmidt-sebastian:


packages/firestore/lite/src/api/reference.ts
```diff
@@ -318,6 +318,30 @@ export function collection(
...
+  const firestoreClient = cast(firestore, Firestore);
+
+  validateArgType('collectionGroup', 'non-empty string', 1, collectionId);
+  if (collectionId.indexOf('/') >= 0) {
+    throw new FirestoreError(
+      Code.INVALID_ARGUMENT,
+      `Invalid collection ID '${collectionId}' passed to function ` +
+        `collectionGroup(). Collection IDs must not contain '/'.`
+    );
+  }
```
rafikhan, line 334: Suggestion: Factor into a separate assertion method.

schmidt-sebastian: This is a special error that only applies to collection IDs. The `collection` and `doc` methods can be called with `/`-separated paths.

rafikhan:
LGTM

Issue #3172, opened 2020-06-05
Can't use FieldValues inside an Array: AsyncQueue Failed to persist write: TypeError: Cannot read property 'M' of undefined


thesandlord:
# [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Browser version: Chrome 83.0.4103.61 (Official Build) (64-bit)
  * Firebase SDK version: 7.15.0
  * Firebase Product: Firestore

### [REQUIRED] Describe the problem

When using a Firestore `FieldValue` in an array, for example a `serverTimestamp` or `delete`, Firestore crashes with the error:

firestore-memory:
```
AsyncQueue Failed to persist write: TypeError: Cannot read property 'M' of undefined
```

or

firstore:
```
AsyncQueue Failed to persist write: TypeError: Cannot read property 'B' of undefined
```

#### Steps to reproduce:

```javascript
firebase.firestore().collection("test").doc("a").set({ 
  thing: [firebase.firestore.FieldValue.serverTimestamp()] 
})
```

PTAL at the codesandbox. Click "Bad Data" and it will crash Firestore.

#### Relevant Code:

https://codesandbox.io/s/simple-firestore-javascript-example-gmygz

As far as I know, this used to work

wilhuff:
Hi! Thanks for writing in.

This has never worked. In prior versions (e.g. 7.14.0) we threw an exception like this:

```
Failure: FirebaseError: [code=invalid-argument]: Function DocumentReference.set() called with invalid data. FieldValue.serverTimestamp() is not currently supported inside arrays
```

It looks like recent changes we've made to minification have broken this validation logic.

PR #3183, opened 2020-06-07
Throw proper error for FieldValue transforms in arrays

schmidt-sebastian:
Fixes a regression introduced in #3048 which changed the context.path variable from `Path|null` to `undefined|null`.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3172

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -214,7 +214,6 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
       // outstanding so the response is potentially for a previous user (which
       // user, we can't be sure).
       if (this.tokenCounter !== initialTokenCounter) {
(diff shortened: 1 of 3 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -214,6 +214,7 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
       // outstanding so the response is potentially for a previous user (which
       // user, we can't be sure).
       if (this.tokenCounter !== initialTokenCounter) {
+        this.forceRefresh = true;
```
dconeybe, line 217: Is this change related to the fix at hand?

schmidt-sebastian: No, this is meant for a different PR. Removed.

Issue #3169, opened 2020-06-05
Time consuming debug: Function DocumentReference.update() called with invalid data. Unsupported field value: undefined


arthabus:
### Describe your environment

  * Operating System version: macOs 10.14.5
  * Browser version: Chrome, Version 83.0.4103.61 (Official Build) (64-bit)
  * Firebase SDK version: 7.9.1
  * Firebase Product: database (firestore)

###  Describe the problem

Is it possible to include "key" name and the failed database path in the error message?

Since our application is growing, it becomes harder and harder to debug this error - our objects became quite complex and often times we are doing multiple requests in parallel (Promise.all(...)) 

It started to take quite a chunk of time to just figure out where exactly we've got an `undefined` field.

#### Steps to reproduce:

1. Create  a complex JSON object with multiple nesting.
2. Assign an `undefined` to a random field in one of the objects.
3. Save it to firestore database

Bonus step:

4. Create several objects, randomly assign `undefined` to one of their fields and save each of them to a different location in firestore in parallel.

**Problem**: Trying to figure out what object failed and which field is `undefined` using the error log takes quite some time as it currently just says: `Unsupported field value: undefined`

**Expected**: The log specifies the name of the key with the `undefined` value and the database path where the update/insert was suppose to happen so it's clear from the log where to look at.

wilhuff:
Hi! Thanks for writing in and sorry for the difficulty. I've put together #3189 that should address this.

PR #3189, opened 2020-06-08
Provide document path context for input validation errors.

wilhuff:
Potentially addresses #3169.


packages/firestore/src/api/field_value.ts
```diff
@@ -107,11 +107,12 @@ export class ArrayUnionFieldValueImpl extends SerializableFieldValue {
 
   _toFieldTransform(context: ParseContext): FieldTransform {
     // Although array transforms are used with writes, the actual elements
-    // being uniomed or removed are not considered writes since they cannot
+    // being unioned or removed are not considered writes since they cannot
     // contain any FieldValue sentinels, etc.
     const parseContext = new ParseContext(
```
schmidt-sebastian, line 112: It might be time to add a helper method to ParseContext (or a tree-shakeable function in this file) that simplifies the creation of an argument parser. If `arrayElement` remains configurable, this code would replace line 112-122, 145-155 and 173-181.

Optional.

packages/firestore/src/api/user_data_reader.ts
```diff
@@ -761,22 +775,49 @@ export function fieldPathFromArgument(
...
+): Error {
+  const hasPath = path && !path.isEmpty();
+  const hasDocument = targetDoc !== undefined;
+
+  let description = '';
+  if (hasPath || hasDocument) {
+    description += ' (found';
+
+    if (hasPath) {
+      description += ` in field ${path!.toString()}`;
```
schmidt-sebastian, line 807: Nit: `${path!.toString()}` is the same as `${path}` (here and below).

PR #3190, opened 2020-06-08
Add GPMID header to Firebase Database

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/1610/files


packages/database/CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+- [changed] Added internal HTTP header to the WebChannel connection.
```
Feiyang1, line 2: I don't think database uses WebChannel? or are you using it as a general term for realtime connection?

schmidt-sebastian: WebSockets. Will fix.

PR #3191, opened 2020-06-08
Add signInWithPhoneNumber implementation

samhorlbeck:


packages-exp/auth-exp/src/core/providers/phone.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    sinon.restore();
+  });
+
+  context('#verifyPhoneNumber', () => {
+    it('calls verify on the appVerifier and then calls the server', async () => {
+      const route = mockEndpoint(Endpoint.SEND_VERIFICATION_CODE, {
+        sessionInfo: 'verification-id'
+      });
+
+      auth.settings.appVerificationDisabledForTesting = true;
```
avolkovi, line 48: this setup pattern with appVerificationDisabledForTesting and recaptachaverifier is used a couple times, do you want to extract to testAuth()?

packages-exp/auth-exp/src/core/strategies/phone.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    it('calls verify phone number', async () => {
+      await signInWithPhoneNumber(auth, '+15105550000', verifier);
+
+      expect(sendCodeEndpoint.calls[0].request).to.eql({
+        recaptchaToken: 'recaptcha-token',
+        phoneNumber: '+15105550000'
+      });
+    });
+
+    context('ConfirmationResult', () => {
```
avolkovi, line 71: can you test the callback getting called?

packages-exp/auth-exp/src/core/strategies/phone.ts
```diff
@@ -0,0 +1,103 @@
...
+  return new ConfirmationResult(verificationId, cred =>
+    signInWithCredential(auth, cred)
+  );
+}
+
+/**
+ *  Returns a verification ID to be used in conjunction with the SMS code that
+ *  is sent.
+ */
+export async function _verifyPhoneNumber(
```
avolkovi, line 67: this is nice and clean 👍

packages-exp/auth-exp/src/core/providers/phone.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  static credentialFromError(
+    error: FirebaseError
+  ): externs.AuthCredential | null {
+    void error;
+    return debugFail('not implemented');
+  }
+
+  static credentialFromJSON(json: string | object): externs.AuthCredential {
```
avolkovi, line 67: isn't this one trivial to implement in this PR?

packages-exp/auth-types-exp/index.d.ts
```diff
@@ -200,7 +203,6 @@ export abstract class OAuthCredential implements AuthCredential {
   constructor();
 
   toJSON(): object;
-  fromJSON(json: object | string): OAuthCredential | null;
```
avolkovi, line 203: what about these static methods?

packages-exp/auth-exp/src/core/providers/phone.test.ts
```diff
@@ -0,0 +1,79 @@
...
+      });
+    });
+  });
+
+  context('.credential', () => {
+    it('creates a phone auth credential', () => {
+      const credential = PhoneAuthProvider.credential('id', 'code');
+
+      // Allows us to inspect the object
+      const blob = credential.toJSON() as { [key: string]: string };
```
yuchenshi, line 73: nit
```suggestion
      const blob = credential.toJSON() as Record<string, string>;
```

packages-exp/auth-exp/src/core/strategies/phone_credential.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    return {
+      sessionInfo: verificationId,
+      code: verificationCode
+    };
+  }
+
+  toJSON(): object {
+    const obj: { [key: string]: string } = {
```
yuchenshi, line 76: nit
```suggestion
    const obj: Record<string, string> = {
```

packages-exp/auth-exp/src/core/strategies/phone_credential.ts
```diff
@@ -0,0 +1,122 @@
...
+      phoneNumber,
+      temporaryProof
+    } = json as { [key: string]: string };
+    if (
+      !verificationCode &&
+      !verificationId &&
+      !phoneNumber &&
+      !temporaryProof
+    ) {
+      return null;
```
yuchenshi, line 112: Shall we throw instead of return null here?

samhorlbeck: [No, the spec is that it just returns `null`](https://firebase.google.com/docs/reference/js/firebase.auth.AuthCredential#static-fromjson)

packages-exp/auth-exp/src/core/strategies/phone.test.ts
samhorlbeck: That's covered in the test "calling #confirm finishes the sign in flow"

The callback calls `signInWithCredential()` which itself calls the `SIGN_IN_WITH_PHONE_NUMBER` endpoint

packages-exp/auth-exp/src/core/providers/phone.ts
samhorlbeck: I'm actually not sure the purpose of that one... It's already a static method on the PhoneCredential

packages-exp/auth-types-exp/index.d.ts
samhorlbeck: They just need to be defined on every `AuthCredential` implementation. Unfortunately there's no way in TS to define a static method that's always available. This kinda makes sense though-- you always need a reference to the actual class (not just an interface) in order to call a static method.

packages-exp/auth-exp/src/core/providers/phone.test.ts
samhorlbeck: Done for `appVerificationDisabledForTesting`. I think it makes sense to keep the verifier in each individual test

packages-exp/auth-exp/src/core/providers/phone.test.ts
samhorlbeck: TIL

packages-exp/auth-exp/src/core/strategies/phone_credential.ts
samhorlbeck: Done

PR #3198, opened 2020-06-10
Add Firestore ReactNative build

schmidt-sebastian:
This builds work out of the box in RN since it relies on js-base64 for base64 encoding.

Fixes https://github.com/firebase/firebase-js-sdk/issues/2667

### DIFF
--- a/packages/firestore/index.rn.ts
+++ b/packages/firestore/index.rn.ts
@@ -14,28 +14,28 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 import firebase from '@firebase/app';
 import { FirebaseNamespace } from '@firebase/app-types';
 
 import { Firestore } from './src/api/database';
-import { MemoryComponentProvider } from './src/core/component_provider';
+import { IndexedDbComponentProvider } from './src/core/component_provider';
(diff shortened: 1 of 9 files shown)


packages/firestore/index.rn.ts
```diff
@@ -0,0 +1,41 @@
...
+import { name, version } from './package.json';
+
+/**
+ * Registers the main Firestore ReactNative build with the components framework.
+ * Persistence can be enabled via `firebase.firestore().enablePersistence()`.
+ */
+export function registerFirestore(instance: FirebaseNamespace): void {
+  configureForFirebase(
+    instance,
+    (app, auth) => new Firestore(app, auth, new IndexedDbComponentProvider())
```
Feiyang1, line 36: RN doesn't support indexeddb, so we should use `MemoryComponentProvider` here. Probably just remove this file, and point `react-native` in the root package.json to the memory build.

schmidt-sebastian: We have some users that are using our IndexedDb Shim support. I am not sure if we should break them in this PR. FWIW, our main Node build also ships with IndexedDb support.

PR #3200, opened 2020-06-10
Implement bundle features in local store.

wu-hui:


packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+   *
+   * @param updatedKeys Keys of the documents to be applied.
+   * @param documents Documents to be applied.
+   * @param remoteVersion The read time of the documents to be applied, it is
+   * a `SnapshotVersion` if all documents have the same read time, or a `DocumentVersionMap` if
+   * they have different read times.
+   */
+  private applyDocuments(
+    documentBuffer: RemoteDocumentChangeBuffer,
+    txn: PersistenceTransaction,
```
schmidt-sebastian, line 684: Nit: All of our methods take `txn` as the first argument.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+    updatedKeys: DocumentKeySet,
+    documents: MaybeDocumentMap,
+    remoteVersion: SnapshotVersion | DocumentVersionMap
```
schmidt-sebastian, line 687: Please split this up into two arguments: `gloablSnapshotVersion` and `doucmentVersions`. It will make the code below a lot easier to follow. With my advice above, you now even get to keep the same number of arguments for this method :)

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+  ): PersistencePromise<MaybeDocumentMap> {
+    const universalReadTime = remoteVersion instanceof SnapshotVersion;
```
schmidt-sebastian, line 689: If you were to keep this, this should be `hasGlobalSnapshotVersion`.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+   * @param documents Documents to be applied.
+   * @param remoteVersion The read time of the documents to be applied, it is
+   * a `SnapshotVersion` if all documents have the same read time, or a `DocumentVersionMap` if
+   * they have different read times.
+   */
+  private applyDocuments(
+    documentBuffer: RemoteDocumentChangeBuffer,
+    txn: PersistenceTransaction,
+    updatedKeys: DocumentKeySet,
+    documents: MaybeDocumentMap,
```
schmidt-sebastian, line 686: It seems like `updatedKeys` always matches the keys of the MaybeDocumentMap. If so, we probably don't need to pass two arguments.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
   }
 
+  /**
+   * Applies the documents from a bundle to the "ground-state" (remote)
+   * documents.
+   *
+   * LocalDocuments are re-calculated if there are remaining mutations in the
+   * queue.
+   */
+  applyBundledDocuments(
```
schmidt-sebastian, line 625: I'd prefer if we kept referring to the feature as "Bundle". `applyBundleDocuments` makes it clear that the documents are from a Firestore Bundle, whereas `applyBundledDocuments` just suggests that the documents were somehow bundled (possibly in a generic fashion).

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+  ): Promise<boolean> {
+    const currentReadTime = this.bundleConverter.toSnapshotVersion(
+      bundleMetadata.createTime!
+    );
+    return this.persistence
+      .runTransaction('isNewerBundleLoaded', 'readonly', transaction => {
+        return this.bundleCache.getBundle(transaction, bundleMetadata.id!);
+      })
+      .then(cached => {
+        return !!cached && cached!.createTime!.compareTo(currentReadTime) > 0;
```
schmidt-sebastian, line 758: You only need to do the `!!` if you are assigning to a boolean. If statements except "truthy" values directly, so `!!cached` is the same as `cached`. 

You can also drop the bang on `cached!`.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+  /**
+   * Saves the given `BundleMetadata` to local persistence.
+   * @param bundleMetadata
+   */
+  saveBundle(bundleMetadata: bundleProto.BundleMetadata): Promise<void> {
+    return this.persistence.runTransaction(
+      'Save bundle',
+      'readwrite',
+      transaction => {
+        return this.bundleCache.saveBundleMetadata(transaction, bundleMetadata);
```
schmidt-sebastian, line 771: Should we also update `getBundle` to be `getBundleMetdadata`?. I just had to look up what `getBundle` does (to verify that it doesn't look a bunch of data in `isNewerBundleLoaded`).

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +615,191 @@ export class LocalStore {
...
+   * Returns a promise of a `NamedQuery` associated with given query name. Promise
+   * resolves to undefined if no persisted data can be found.
+   */
+  getNamedQuery(queryName: string): Promise<NamedQuery | undefined> {
+    return this.persistence.runTransaction(
+      'Get named query',
+      'readonly',
+      transaction => {
+        return this.bundleCache.getNamedQuery(transaction, queryName);
+      }
```
schmidt-sebastian, line 786: Optional (here and all over): 

```
     transaction => {
        return this.bundleCache.getNamedQuery(transaction, queryName);
      }
```

is the same as

```
transaction => this.bundleCache.getNamedQuery(transaction, queryName)
```

which is ore concise. What you prefer is purely a style question and your opinion might differ here.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -66,23 +75,16 @@ export abstract class RemoteDocumentChangeBuffer {
...
-    );
-    return this._readTime;
+  protected getReadTime(key: DocumentKey): SnapshotVersion {
+    const change = this.changes.get(key);
+    if (change) {
+      debugAssert(
+        !!change.readTime,
+        `Read time is not set for ${key}. All removeEntry() calls must include a readTime if 'trackRemovals' is used.`
+      );
+      return change.readTime!;
```
schmidt-sebastian, line 85: You can remove the bang.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -66,23 +75,16 @@ export abstract class RemoteDocumentChangeBuffer {
...
-    this._readTime = value;
-  }
-
-  protected get readTime(): SnapshotVersion {
-    debugAssert(
-      this._readTime !== undefined,
-      'Read time is not set. All removeEntry() calls must include a readTime if `trackRemovals` is used.'
-    );
-    return this._readTime;
+  protected getReadTime(key: DocumentKey): SnapshotVersion {
```
schmidt-sebastian, line 78: Is there a reason you changed this from a property to a method?

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -153,6 +166,19 @@ class LocalStoreTester {
     return this;
   }
 
+  afterBundleDocuments(documents: BundledDocuments): LocalStoreTester {
+    this.prepareNextStep();
+
+    this.promiseChain = this.promiseChain
+      .then(() => {
+        return this.localStore.applyBundledDocuments(documents);
```
schmidt-sebastian, line 174: Note that you used both `afterBundleDocuments` and `applyBundledDocuments` here.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -387,6 +413,25 @@ class LocalStoreTester {
     return this;
   }
 
+  toHaveNewerBundle(
+    metadata: BundleMetadata,
+    expected: boolean
+  ): LocalStoreTester {
+    this.promiseChain = this.promiseChain.then(() => {
+      return this.localStore.isNewerBundleLoaded(metadata).then(actual => {
```
schmidt-sebastian, line 421: Name is also inconsistent here. FWIW, `toHaveNewerBundle` matches my name suggestion from above.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1486,6 +1535,115 @@ function genericLocalStoreTests(
...
+  it('handles saving bundled documents with older existing version', () => {
+    const query = Query.atPath(path('foo'));
+    return expectLocalStore()
+      .afterAllocatingQuery(query)
+      .toReturnTargetId(2)
+      .after(docAddedRemoteEvent(doc('foo/bar1', 1, { val: 'to-delete' }), [2]))
+      .toContain(doc('foo/bar1', 1, { val: 'to-delete' }))
+      .after(
+        bundledDocuments([
+          doc('foo/bar', 1, { sum: 1336 }),
```
schmidt-sebastian, line 1583: Should this be version 2? 

We may also want to skip bundle document loading for documents that we already have cached at that version. The local documents seem safer than replacing them with a bundled document that could be from an arbitrary source.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1486,6 +1535,115 @@ function genericLocalStoreTests(
...
+      .toReturnTargetId(2)
+      .afterMutations([
+        patchMutation('foo/bar', {}, Precondition.none()),
+        transformMutation('foo/bar', { sum: FieldValue.increment(1) })
+      ])
+      .toReturnChanged(
+        doc('foo/bar', 0, { sum: 1 }, { hasLocalMutations: true })
+      )
+      .toContain(doc('foo/bar', 0, { sum: 1 }, { hasLocalMutations: true }))
+      .after(bundledDocuments([doc('foo/bar', 1, { sum: 1337 })]))
```
schmidt-sebastian, line 1609: This should also be a new version.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1486,6 +1535,115 @@ function genericLocalStoreTests(
...
+      .toReturnChanged(
+        doc('foo/bar', 1, { sum: 1 }, { hasLocalMutations: true })
+      )
+      .toContain(doc('foo/bar', 1, { sum: 1 }, { hasLocalMutations: true }))
+      .finish();
+  });
+
+  it('handles PatchMutation with Transform -> BundledDocuments', () => {
+    // Note: This test reflects the current behavior, but it may be preferable
+    // to replay the mutation once we receive the first value from the backend.
```
schmidt-sebastian, line 1619: I think what is implemented is the better behavior, since the document will likely end up looking like the local mutation. What do you think?

packages/firestore/test/util/helpers.ts
```diff
@@ -401,6 +404,55 @@ export function docUpdateRemoteEvent(
   return aggregator.createRemoteEvent(doc.version);
 }
 
+export class TestBundledDocuments {
```
schmidt-sebastian, line 407: Can this be inlined? `bundledDocuments` would just return `BundledDocuments`.

packages/firestore/src/local/local_store.ts
wu-hui: Here `!!` is used to remove the `| undefined` part of the return type, without it the return type mismatches.

packages/firestore/src/local/local_store.ts
wu-hui: Yeah, I was trying to avoid using the same type name there. Changed to use `getBundleMetadata` now.

packages/firestore/src/local/local_store.ts
wu-hui: Updated.

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -25,6 +25,18 @@ import { PersistenceTransaction } from './persistence';
 import { PersistencePromise } from './persistence_promise';
 import { SnapshotVersion } from '../core/snapshot_version';
 
+/**
+ * Represents a document change to be applied to remote document cache.
+ */
+class RemoteDocumentChange {
```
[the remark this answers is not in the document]
wu-hui: Done.

packages/firestore/src/local/remote_document_change_buffer.ts
wu-hui: Done.

packages/firestore/src/local/remote_document_change_buffer.ts
wu-hui: Because it takes a parameter now?

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: You were right!

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: Agree!

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: Not there were no `foo/bar` when loading the bundle. I changed the name to make it less confusing.

Also added a test to check when the existing version is the same.

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: It is. Note the existing version is 0.

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: This is copied from the remote event tests. I am actually not sure why it was there and is it still valid.

I think we can discuss this more with the team in one of the standups.

packages/firestore/test/util/helpers.ts
wu-hui: I added this because it hard to add code to `after(op)` method otherwise, suggestions welcome.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +618,187 @@ export class LocalStore {
...
+  /**
+   * Applies documents to remote document cache, returns the document changes
+   * resulting from applying those documents.
+   *
+   * @param updatedKeys Keys of the documents to be applied.
+   * @param documents Documents to be applied.
+   * @param remoteVersion The read time of the documents to be applied, it is
+   * a `SnapshotVersion` if all documents have the same read time, or a `DocumentVersionMap` if
+   * they have different read times.
+   */
```
schmidt-sebastian, line 682: Do you mind adding a short description for all documents?

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +618,187 @@ export class LocalStore {
...
+  private applyDocuments(
+    txn: PersistenceTransaction,
+    documentBuffer: RemoteDocumentChangeBuffer,
+    documents: MaybeDocumentMap,
+    globalVersion: SnapshotVersion | undefined,
+    documentVersions: DocumentVersionMap | undefined
```
schmidt-sebastian, line 688: Looking at this more, would it be possible to use the readTime from the bundle as the `globalVersion` and drop `documentVersions`? AFAIK the only real guarantee that we have to make is that a changed document has a changed read time that is higher than the new document version. 

If that is not possible, I think we should explore adding the read time to MaybeDocument (it's already part of DbRemoteDocument, so we do have the data). That might be a relatively simple way to improve the code flow here.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +618,187 @@ export class LocalStore {
...
+  /**
+   * Returns a promise of a boolean to indicate if the given bundle has already
+   * been loaded and the create time is newer than the current loading bundle.
+   */
+  hasNewerBundle(bundleMetadata: bundleProto.BundleMetadata): Promise<boolean> {
+    const currentReadTime = this.bundleConverter.toSnapshotVersion(
+      bundleMetadata.createTime!
+    );
+    return this.persistence
+      .runTransaction('isNewerBundleLoaded', 'readonly', transaction => {
```
schmidt-sebastian, line 754:
```suggestion
      .runTransaction('hasNewerBundle', 'readonly', transaction => {
```

packages/firestore/src/local/remote_document_change_buffer.ts
```diff
@@ -25,6 +25,16 @@ import { PersistenceTransaction } from './persistence';
...
 import { SnapshotVersion } from '../core/snapshot_version';
 
+/**
+ * Represents a document change to be applied to remote document cache.
+ */
+interface RemoteDocumentChange {
+  // The document in this change, null if it is a removal from the cache.
+  readonly maybeDocument: MaybeDocument | null;
+  // The timestamp when this change is read.
+  readonly readTime?: SnapshotVersion;
```
schmidt-sebastian, line 35: The use of "null" and "undefined" for "not set" is a bit inconsistent here. 

FWIW, we could drop readTime if we add it to MaybeDocument.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -153,6 +166,19 @@ class LocalStoreTester {
     return this;
   }
 
+  afterBundleDocuments(documents: BundledDocuments): LocalStoreTester {
+    this.prepareNextStep();
+
+    this.promiseChain = this.promiseChain
+      .then(() => {
+        return this.localStore.applyBundleDocuments(documents);
+      })
```
schmidt-sebastian, line 175:
```suggestion
      .then(() => this.localStore.applyBundleDocuments(documents))
```

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -387,6 +413,25 @@ class LocalStoreTester {
...
+        expect(actual).to.equal(expected);
+      });
+    });
+    return this;
+  }
+
+  afterSavingBundle(metadata: BundleMetadata): LocalStoreTester {
+    this.promiseChain = this.promiseChain.then(() => {
+      return this.localStore.saveBundle(metadata);
+    });
```
schmidt-sebastian, line 431:
```suggestion
    this.promiseChain = this.promiseChain.then(() => this.localStore.saveBundle(metadata));
```

packages/firestore/test/util/helpers.ts
```diff
@@ -401,6 +404,55 @@ export function docUpdateRemoteEvent(
...
+        metadata: {
+          name: JSON_SERIALIZER.toName(d.key),
+          readTime: JSON_SERIALIZER.toVersion(d.version),
+          exists: true
+        },
+        document: JSON_SERIALIZER.toDocument(d)
+      });
+    }
+  }
+
```
schmidt-sebastian, line 436: Maybe:

```suggestion
  const result = documents.map(d => {
    return {
        metadata: {
          name: JSON_SERIALIZER.toName(d.key),
          readTime: JSON_SERIALIZER.toVersion(d.version),
          exists: d instanceof Document
        },
        document: d instanceof Document ? JSON_SERIALIZER.toDocument(d) : undefined
      }); 
  });

```

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +618,187 @@ export class LocalStore {
...
   }
 
+  /**
+   * Applies the documents from a bundle to the "ground-state" (remote)
+   * documents.
+   *
+   * LocalDocuments are re-calculated if there are remaining mutations in the
+   * queue.
+   */
+  applyBundleDocuments(documents: BundledDocuments): Promise<MaybeDocumentMap> {
```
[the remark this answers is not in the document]
wu-hui: Happy to make this tree-shakeable, not sure I did everything right though.

I'll update my other PRs to make it tree-shakeable as well.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: I think it would be possible to add a readTime to `MaybeDocument`. It looks like we would need to change a lot of our plumbing, so I am ok with leaving as is. We could add a TODO if this is something that we think would clean up our code in general.

packages/firestore/src/local/local_store.ts
```diff
@@ -89,6 +95,40 @@ export interface QueryResult {
   readonly remoteKeys: DocumentKeySet;
 }
 
+/**
+ * Base of local store implementations. This is mainly used to define fields
```
schmidt-sebastian, line 99: s/local store/

packages/firestore/src/local/local_store.ts
```diff
@@ -89,6 +95,40 @@ export interface QueryResult {
...
+ * that are used by tree-shakeable free functions that implement optional
+ * local store features.
+ */
+class LocalStoreBase {
```
schmidt-sebastian, line 103: So the idea behind the pattern that is used in Datastore/DatastoreImpl (and just as of today, also in Target/TargetImpl) is to have a main class (in this case LocalStore) that doesn't change its public API.  

So you would keep LocalStore as is (and continue to export it), and then you would create LocalStoreImpl that makes fields public that you only need to be public in this module. Then you would add a `newLocalStore` function that creates a LocalStoreImpl, but is typed to return a LocalStore. You then end up with a module that has a bunch of public members that are not leaked externally (in our code base).

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +628,84 @@ export class LocalStore {
...
+   * @param documents Documents to be applied.
+   * @param globalVersion A `SnapshotVersion` representing the read time if all
+   *        documents have the same read time.
+   * @param documentVersions A DocumentKey-to-SnapshotVersion map if documents
+   *        have their own read time.
+   *
+   * Note: this function will use `documentVersions` if it is defined;
+   * when it is not defined, resorts to `globalVersion`.
+   */
+  prepareApplyingDocuments(
```
schmidt-sebastian, line 647: s/prepareApplyingDocuments/prepareToApplyDocuments/ (or back to `applyDocuments`)

Or rather: `populateDocumentChangeBuffer`?

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +628,84 @@ export class LocalStore {
       });
   }
 
+  /**
+   * Prepares document change buffer resulted from documents to be applied,
```
schmidt-sebastian, line 632: Nit: This line needs a rewrite.

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +628,84 @@ export class LocalStore {
...
+   *        have their own read time.
+   *
+   * Note: this function will use `documentVersions` if it is defined;
+   * when it is not defined, resorts to `globalVersion`.
+   */
+  prepareApplyingDocuments(
+    txn: PersistenceTransaction,
+    documentBuffer: RemoteDocumentChangeBuffer,
+    documents: MaybeDocumentMap,
+    globalVersion: SnapshotVersion | undefined,
```
schmidt-sebastian, line 651: Can we make this non-optional and use SnapshotVersion.min() for the bundle loader?

packages/firestore/src/local/local_store.ts
```diff
@@ -642,6 +628,84 @@ export class LocalStore {
...
+      let changedDocs = maybeDocumentMap();
+      documents.forEach((key, doc) => {
+        const existingDoc = existingDocs.get(key);
+        let docReadVersion: SnapshotVersion | null = null;
+        if (documentVersions) {
+          docReadVersion = documentVersions.get(key);
+        } else if (globalVersion) {
+          docReadVersion = globalVersion;
+        }
+        debugAssert(!!docReadVersion, 'Document read version must exist');
```
schmidt-sebastian, line 666:
```suggestion
const readTime = documentVersions?.get(key) ?? globalVersion
```

packages/firestore/src/local/local_store.ts
```diff
@@ -1049,11 +1114,13 @@ export class MultiTabLocalStore extends LocalStore {
   }
 
   setNetworkEnabled(networkEnabled: boolean): void {
-    this.persistence.setNetworkEnabled(networkEnabled);
+    (this.persistence as IndexedDbPersistence).setNetworkEnabled(
```
schmidt-sebastian, line 1117: If you add a `persistence: IndexedDbPersistence` field and assign it manually (similar to mutationQueue), this cast will go away.

packages/firestore/src/local/local_store.ts
```diff
@@ -1130,3 +1197,131 @@ export async function ignoreIfPrimaryLeaseLoss(
...
+ * documents.
+ *
+ * LocalDocuments are re-calculated if there are remaining mutations in the
+ * queue.
+ */
+export function applyBundleDocuments(
+  localStore: LocalStore,
+  documents: BundledDocuments
+): Promise<MaybeDocumentMap> {
+  const bundleConverter = new BundleConverter(localStore.serializer);
```
schmidt-sebastian, line 1212: If you follow the Datastore pattern, you would start this off with:

```
const localStoreImpl = debugCast(localStore, LocalStoreImpl);
```

Otherwise, this functions are tree-shakeable as is. Thanks!

packages/firestore/src/local/local_store.ts
wu-hui: Sorry I misunderstood earlier.

As discussed offline, we will need to make LocalStore and SyncEngine interfaces to achieve this because of existing MultiTab* implementations, which is best done from master.

packages/firestore/src/local/local_store.ts
wu-hui: `populateDocumentChangeBuffer` sounds good.

packages/firestore/src/local/local_store.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Done, thanks!

packages/firestore/src/local/local_store.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Will do once we have addressed inheritance issue in `master`.

packages/firestore/src/core/bundle.ts
```diff
@@ -40,3 +51,50 @@ export interface NamedQuery {
   /** The time at which the results for this query were read. */
   readonly readTime: SnapshotVersion;
 }
+
+/**
+ * Represents a bundled document, including the metadata and the document
+ * itself, if exists.
```
schmidt-sebastian, line 57: missing "it"

packages/firestore/src/core/bundle.ts
```diff
@@ -40,3 +51,50 @@ export interface NamedQuery {
...
+ */
+export class BundleConverter {
+  constructor(private serializer: JsonProtoSerializer) {}
+
+  toDocumentKey(name: string): DocumentKey {
+    return fromName(this.serializer, name);
+  }
+
+  /**
+   * Converts a [metadata, document] pair to a MaybeDocument.
```
schmidt-sebastian, line 80:
```suggestion
   * Converts a BundledDocument to a MaybeDocument.
```

packages/firestore/src/local/local_store.ts
```diff
@@ -347,6 +357,7 @@ class LocalStoreImpl implements LocalStore {
       this.mutationQueue,
       this.persistence.getIndexManager()
     );
+    this.bundleCache = persistence.getBundleCache();
```
schmidt-sebastian, line 360: I think we have done enough iterations on this PR, but ideally this would not be a field on LocalStore (as it breaks tree-shaking). Could `BundleCache` take LocalStore as an argument? Please explore in a follow up.

packages/firestore/src/local/local_store.ts
```diff
@@ -708,6 +680,80 @@ class LocalStoreImpl implements LocalStore {
...
+    // this parameter.
+    documentVersions: DocumentVersionMap | undefined
+  ): PersistencePromise<MaybeDocumentMap> {
+    let updatedKeys = documentKeySet();
+    documents.forEach(k => (updatedKeys = updatedKeys.add(k)));
+    return documentBuffer.getEntries(txn, updatedKeys).next(existingDocs => {
+      let changedDocs = maybeDocumentMap();
+      documents.forEach((key, doc) => {
+        const existingDoc = existingDocs.get(key);
+        const docReadTime = documentVersions?.get(key) ?? globalVersion;
```
schmidt-sebastian, line 714: Nit: I probably suggested this, but this is functionally equivalent to `documentVersions?.get(key) || globalVersion`, which generates a lot less code. `??` is only needed if you only want to check for undefined or null and explicitly want to not fallback on  "" (empty string), `0` or false.

```
if (doc?.version || 'foo')

=> 

if ((doc === null || doc === void 0 ? void 0 : doc.version) || 'foo')
 

if (doc?.version ?? 'foo')

=>

if ((_a = doc === null || doc === void 0 ? void 0 : doc.version) !== null && _a !== void 0 ? _a : 'foo')
```

As an example for where `??` works but `||` doesn't, consider:

```
class Settings {
   enableAmazingFeature?: boolean;
}

 // This is now always true
if (settings.enableAmazingFeature || true) {}
// This is true iff enableAmazingFeature=undefined or enableAmazingFeature=true
if (settings.enableAmazingFeature ?? true) {}
```

packages/firestore/src/core/bundle.ts
wu-hui: Done.

packages/firestore/src/core/bundle.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: I see. Will do a followup PR.

packages/firestore/src/local/local_store.ts
wu-hui: I see, applied. Thanks!

PR #3201, opened 2020-06-10
Implement bundle loading.

wu-hui:


packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -0,0 +1,257 @@
...
+        verifySuccessProgress(progress);
+
+        // Read from cache. These documents do not exist in backend, so they can
+        // only be read from cache.
+        const snap = await otherDb
+          .collection('coll-1')
+          .get({ source: 'cache' });
+        expect(toDataArray(snap)).to.deep.equal([
+          { k: 'a', bar: 1 },
+          { k: 'b', bar: 2 }
```
schmidt-sebastian, line 252: Does this add any additional coverage for this test? Wouldn't `otherDb.loadBundle` succeeding just be enough?

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -0,0 +1,257 @@
...
+      verifyInProgress(progresses[1], 1);
+      verifyInProgress(progresses[2], 2);
+
+      // Read from cache. These documents do not exist in backend, so they can
+      // only be read from cache.
+      const snap = await db.collection('coll-1').get({ source: 'cache' });
+      expect(toDataArray(snap)).to.deep.equal([
+        { k: 'a', bar: 1 },
+        { k: 'b', bar: 2 }
+      ]);
```
schmidt-sebastian, line 124: Since the document creation is hidden away in a helper, it might make sense to also move the verification step into a helper. We already cannot really tell where these values come from.

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -0,0 +1,257 @@
...
+        expect(
+          otherDb.loadBundle(
+            builder.build('test-bundle', { seconds: 1001, nanos: 9999 })
+          )
+        ).to.be.rejectedWith('Tried to deserialize key from different project');
+
+        // Verify otherDb still functions, despite loaded a problematic bundle.
+        builder = bundleWithTestDocs(otherDb);
+        const progress = await otherDb.loadBundle(
+          builder.build('test-bundle', { seconds: 1001, nanos: 9999 })
```
schmidt-sebastian, line 241: This is not the progress, is it? It is just the final result?

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+    );
+  }
+  return builder.build(
+    'test-bundle',
+    toVersion(JSON_SERIALIZER, version(testDoc.readTime))
+  );
+}
+
+describeSpec('Bundles:', [], () => {
+  specTest('Newer docs from bundles should overwrite cache.', [], () => {
```
schmidt-sebastian, line 62: Same nit regarding period at end of test name.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+      testDoc.content!
+    );
+  }
+  return builder.build(
+    'test-bundle',
+    toVersion(JSON_SERIALIZER, version(testDoc.readTime))
+  );
+}
+
+describeSpec('Bundles:', [], () => {
```
schmidt-sebastian, line 61: Add "no-ios", "no-android" while this is pending work.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+      updateTime: 2999,
+      content: { key: { stringValue: 'b' } }
+    });
+
+    return spec()
+      .userListens(query1)
+      .watchAcksFull(query1, 1000, docA)
+      .expectEvents(query1, { added: [docA] })
+      .loadBundle(bundleString)
+      .expectEvents(query1, { modified: [docAChanged] });
```
schmidt-sebastian, line 80: Should this be raised with `fromCache: true`?

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+                { key: 'patched' },
+                { hasLocalMutations: true }
+              )
+            ],
+            hasPendingWrites: true
+          })
+          .writeAcks('collection/a', 1000)
+          // loading bundleString1 will not raise snapshots, because it is before
+          // the acknowledged mutation.
+          .loadBundle(bundleString1)
```
schmidt-sebastian, line 164: Nit: Please use descriptive name for `bundleString1` and `bundleString2`. You might even be able to get rid of this comment.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+              doc(
+                'collection/a',
+                250,
+                { key: 'patched' },
+                { hasLocalMutations: true }
+              )
+            ],
+            hasPendingWrites: true
+          })
+          // Loading both bundles will not raise snapshots, because of the
```
schmidt-sebastian, line 218: Drop "of"

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+          // mutation is not acknowledged.
+          .loadBundle(bundleString1)
```
schmidt-sebastian, line 220: This bundle is ignored (as tested above), so we probably don't need to test it here. We could also add a small separate test that just verifies this code path.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+
+      return (
+        client(0)
+          .userListens(query)
+          .watchAcksFull(query, 250, docA)
+          .expectEvents(query, {
+            added: [docA]
+          })
+          .client(1)
+          // Bundle tells otherwise, leads to limbo resolution.
```
schmidt-sebastian, line 276: Please be more descriptive. What is the conflicting message in the bundle?

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -448,6 +453,18 @@ abstract class TestRunner {
...
+  private async doLoadBundle(bundle: string): Promise<void> {
+    const reader = new BundleReader(
+      toByteStreamReader(new TextEncoder().encode(bundle))
+    );
+    const task = new LoadBundleTaskImpl();
+    this.queue.enqueueAndForget(() => {
+      return loadBundle(this.syncEngine, reader, task);
+    });
+
+    await task;
```
schmidt-sebastian, line 465:
```suggestion
    const task = new LoadBundleTaskImpl();
    return this.queue.enqueue(() => loadBundle(this.syncEngine, reader, task));
```

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1371,3 +1380,57 @@ export function newMultiTabSyncEngine(
     maxConcurrentLimboResolutions
   );
 }
+
+export function loadBundle(
+  syncEngine: SyncEngine,
+  bundleReader: BundleReader,
+  task: LoadBundleTaskImpl
+): Promise<void> {
```
schmidt-sebastian, line 1388: Did you consider returning the LoadBundleTask instead of having it passed in? You are now mixing two async patterns in one callsite.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,357 @@
...
+          })
+          .writeAcks('collection/a', 1000)
+          // loading bundleString1 will not raise snapshots, because it is before
+          // the acknowledged mutation.
+          .loadBundle(bundleString1)
+          // loading bundleString2 will raise a snapshot, because it is after
+          // the acknowledged mutation.
+          .loadBundle(bundleString2)
+          .expectEvents(query, {
+            modified: [doc('collection/a', 1001, { key: 'fromBundle' })]
```
schmidt-sebastian, line 169: Ideally, this should also be `fromCache`.

packages/firestore/src/core/bundle.ts
```diff
@@ -98,3 +107,147 @@ export class BundleConverter {
...
+      this.progress.documentsLoaded += 1;
+      this.unpairedDocumentMetadata = null;
+    }
+
+    // Loading a document metadata will not update progress.
+    if (this.unpairedDocumentMetadata) {
+      return null;
+    }
+
+    return { ...this.progress };
```
schmidt-sebastian, line 229: Note that I changed this in my diff to just retun this.progress, but I think you are correct and we need to return an immutable copy.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1371,3 +1378,70 @@ export function newMultiTabSyncEngine(
     maxConcurrentLimboResolutions
   );
 }
+
+/**
+ * Loads a Firestore bundle into the SDK, the returned promise resolves when
```
schmidt-sebastian, line 1383:
```suggestion
 * Loads a Firestore bundle into the SDK. The returned promise resolves when
```

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1371,3 +1378,70 @@ export function newMultiTabSyncEngine(
...
+      task._updateProgress(progress);
+    }
+
+    element = await reader.nextElement();
+  }
+
+  const result = await loader.complete();
+  if (result.changedDocs) {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    syncEngine.emitNewSnapsAndNotifyLocalStore(
```
schmidt-sebastian, line 1437: This method updates local state and needs to be waited on.

packages/firestore/src/core/view.ts
```diff
@@ -280,13 +280,15 @@ export class View {
...
+   * @param fromBundle Whether the changes are from applying a bundle file.
    * @return A new ViewChange with the given docs, changes, and sync state.
    */
   // PORTING NOTE: The iOS/Android clients always compute limbo document changes.
   applyChanges(
     docChanges: ViewDocumentChanges,
     updateLimboDocuments: boolean,
-    targetChange?: TargetChange
+    targetChange?: TargetChange,
+    fromBundle: boolean = false
```
schmidt-sebastian, line 291: It might make sense to move this to `ViewDocumentChanges`, which already has similar state (`needsRefill`).

packages/firestore/src/core/view.ts
```diff
@@ -323,7 +325,7 @@ export class View {
         oldDocs,
         changes,
         docChanges.mutatedKeys,
-        newSyncState === SyncState.Local,
+        newSyncState === SyncState.Local || fromBundle,
```
schmidt-sebastian, line 328: How do we propagate this state? Let's say I have an active query with Doc A. I then load bundle with Doc B and Doc C, which raises a "fromCache" snapshot. The backend then sends us Doc B and a global snapshot. That snapshot still needs to be fromCache. Do the documents themselves rather than the ViewSnapshot need to carry this state?

packages/firestore/src/util/byte_stream.ts
```diff
@@ -17,6 +17,13 @@
 
 import { debugAssert } from './assert';
 
+/**
+ * For the byte streams where we have control (like backed by a UInt8Array),
+ * how many bytes to read each time when `ReadableStreamReader.read()` is
+ * called.
```
schmidt-sebastian, line 23:
```suggestion
 * How many bytes to read each time when `ReadableStreamReader.read()` is
 * called. Only applicable for byte streams that we control (e.g. those backed by 
 * an UInt8Array).
```

packages/firestore/test/integration/api_internal/bundle.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      ]);
+    });
+  });
+
+  it('load with documents from other projects fails', () => {
+    return withTestDb(persistence, async db => {
+      let builder = bundleWithTestDocs(db);
+      return withAlternateTestDb(persistence, async otherDb => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        expect(
```
schmidt-sebastian, line 218: This needs to be awaited, otherwise the assert can fire after the test completed.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+import { toVersion } from '../../../src/remote/serializer';
+
+interface TestBundleDocument {
+  key: DocumentKey;
+  readTime: TestSnapshotVersion;
+  createTime?: TestSnapshotVersion;
+  updateTime?: TestSnapshotVersion;
+  content?: api.ApiClientObjectMap<Value>;
+}
+function bundleWithDocument(testDoc: TestBundleDocument): string {
```
schmidt-sebastian, line 40: Nit: Empty line.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+import * as api from '../../../src/protos/firestore_proto_api';
+import { Value } from '../../../src/protos/firestore_proto_api';
+import { toVersion } from '../../../src/remote/serializer';
+
+interface TestBundleDocument {
+  key: DocumentKey;
+  readTime: TestSnapshotVersion;
+  createTime?: TestSnapshotVersion;
+  updateTime?: TestSnapshotVersion;
+  content?: api.ApiClientObjectMap<Value>;
```
schmidt-sebastian, line 38: The existing spec tests use the "user facing types" rather than the Value proto to encode the types. This means that we would store `{ foo: 1 }` rather than `{ foo: { integerValue: 1 }}`. I think that you will find that following this pattern will greatly simplify replaying these tests on Android and iOS.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+    return spec()
+      .userListens(query1)
+      .watchAcksFull(query1, 1000, docA)
+      .expectEvents(query1, { added: [docA] })
+      .loadBundle(bundleString)
+      .expectEvents(query1, { modified: [docAChanged], fromCache: true });
+  });
+
+  specTest('Newer deleted docs from bundles should delete cache', [], () => {
```
schmidt-sebastian, line 83:
```suggestion
  specTest('Newer deleted docs from bundles should delete cached docs', [], () => {
```

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+        .userListens(query1)
+        .watchAcksFull(query1, 1000, docA)
+        .expectEvents(query1, { added: [docA] })
+        // No events are expected here.
+        .loadBundle(bundleString)
+    );
+  });
+
+  specTest(
+    'Newer docs from bundles should raise snapshot only when watch catches up with acknowledged writes',
```
schmidt-sebastian, line 120:
```suggestion
    'Newer docs from bundles should raise snapshot only when Watch catches up with acknowledged writes',
```

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+      const bundleAfterMutationAck = bundleWithDocument({
+        key: docA.key,
+        readTime: 1001,
+        createTime: 250,
+        updateTime: 1001,
+        content: { key: { stringValue: 'fromBundle' } }
+      });
+      return (
+        spec()
+          .withGCEnabled(false)
```
schmidt-sebastian, line 143: Can you file a bug to define the behavior bundles when `withGCEnabled` is not set to false? Note that this flag is spec test only and that this not actually a mode that users can enable with memory persistence.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,349 @@
...
+        .watchAcksFull(query, 250)
+        // Backend tells there is no such doc.
+        .expectEvents(query, {})
+        // Bundle tells otherwise, leads to limbo.
+        .loadBundle(bundleString1)
+        .expectEvents(query, {
+          added: [doc('collection/a', 500, { key: 'b' })],
+          fromCache: true
+        })
+        .expectLimboDocs(docA.key)
```
schmidt-sebastian, line 242: Can you expand this test: We should verify that when the Limbo doc is resolved we go back to `fromCache: false`.

wu-hui: Done.

packages/firestore/test/unit/specs/bundle_spec.test.ts
wu-hui: Done.

packages/firestore/src/api/bundle.ts
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Notifies all observers that bundle loading has failed, with a provided
+   * `Error` as the reason.
+   */
+  _failWith(error: Error): void {
+    this._lastProgress.taskState = 'Error';
+
+    if (this._progressObserver && this._progressObserver.next) {
+      this._progressObserver.next(this._lastProgress);
+    }
```
schmidt-sebastian, line 87: Please try to match this behavior: https://github.com/firebase/firebase-js-sdk/blob/0131e1fcade96853997fc04ba1fee0e8d40ed445/packages/storage/src/task.ts#L626

Basically, `next` doesn't get called for the `Complete` of `Error` state.

packages/firestore/src/api/bundle.ts
```diff
@@ -0,0 +1,111 @@
...
+  ): Promise<T | R> {
+    return this._taskCompletionResolver.promise.then(onFulfilled, onRejected);
+  }
+
+  /**
+   * Notifies all observers that bundle loading has completed, with a provided
+   * `LoadBundleTaskProgress` object.
+   */
+  _completeWith(progress: firestore.LoadBundleTaskProgress): void {
+    this._updateProgress(progress);
```
schmidt-sebastian, line 70: Do we not need to set the taskState here (like you did below)?

packages/firestore/src/api/bundle.ts
```diff
@@ -0,0 +1,111 @@
...
+    this._taskCompletionResolver.reject(error);
+  }
+
+  /**
+   * Notifies a progress update of loading a bundle.
+   * @param progress The new progress.
+   */
+  _updateProgress(progress: firestore.LoadBundleTaskProgress): void {
+    debugAssert(
+      this._lastProgress.taskState !== 'Error',
```
schmidt-sebastian, line 102: You may want to add 'complete' to this check as well.

packages/firestore/src/core/bundle.ts
```diff
@@ -98,3 +107,132 @@ export class BundleConverter {
...
+    totalBytes: metadata.totalBytes!
+  };
+}
+
+export class LoadResult {
+  constructor(
+    readonly progress: firestore.LoadBundleTaskProgress,
+    readonly changedDocs?: MaybeDocumentMap
+  ) {}
+}
```
schmidt-sebastian, line 148: Since this is an exported class, it should have a less generic name.

packages/firestore/src/core/bundle.ts
```diff
@@ -98,3 +107,132 @@ export class BundleConverter {
...
+    const lastDocument =
+      this.documents.length === 0
+        ? null
+        : this.documents[this.documents.length - 1];
+    debugAssert(
+      !!lastDocument ||
+        !lastDocument!.metadata.exists ||
+        (!!lastDocument!.metadata.exists && !!lastDocument!.document),
+      'Bundled documents ends with a document metadata.'
+    );
```
schmidt-sebastian, line 224:
```suggestion
debugAssert(
        this.documents[this.documents.length - 1]?.metadata.exists !== true || 
        this.documents[this.documents.length - 1].document),
      'Bundled documents ends with a document metadata and missing document.'
    );
```

packages/firestore/src/core/bundle.ts
```diff
@@ -98,3 +107,132 @@ export class BundleConverter {
...
+
+    for (const q of this.queries) {
+      await saveNamedQuery(this.localStore, q);
+    }
+
+    let changedDocs;
+    if (this.documents.length > 0) {
+      changedDocs = await applyBundleDocuments(this.localStore, this.documents);
```
schmidt-sebastian, line 232: How often do you think bundles will be empty? If this is rare, then we could just call `applyBundleDocuments`, which only has a little bit of overhead but removes two if statements from your code. You already don't call into this code for skipped bundles, so I would think `changedDocs` would rarely be empty.

packages/firestore/test/unit/specs/bundle_spec.test.ts
schmidt-sebastian: Do you want to add a comment here that links to the bug you filed? Makes it easier to parse this test.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,368 @@
...
+    return spec()
+      .userListens(query1)
+      .watchAcksFull(query1, 1000, docA)
+      .expectEvents(query1, { added: [docA] })
+      .loadBundle(bundleString)
+      .expectEvents(query1, { modified: [docAChanged] });
+  });
+
+  specTest(
+    'Newer deleted docs from bundles should delete cache docs',
```
schmidt-sebastian, line 90: s/cache/cached

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,368 @@
...
+                'collection/a',
+                250,
+                { key: 'patched' },
+                { hasLocalMutations: true }
+              )
+            ],
+            hasPendingWrites: true
+          })
+          .writeAcks('collection/a', 1000)
+          // loading bundleBeforeMutationAck will not raise snapshots, because it is before
```
schmidt-sebastian, line 174: s/it is before/its snapshot version is older than/

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,368 @@
...
+        updateTime: 500,
+        content: { key: 'b' }
+      });
+
+      const bundleAfterMutationAck = bundleWithDocument({
+        key: docA.key,
+        readTime: 1001,
+        createTime: 250,
+        updateTime: 1001,
+        content: { key: 'fromBundle' }
```
schmidt-sebastian, line 149: s/key/value since the `key` is 'a'. You need to update all document contents.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,368 @@
...
+                'collection/a',
+                250,
+                { key: 'patched' },
+                { hasLocalMutations: true }
+              )
+            ],
+            hasPendingWrites: true
+          })
+          // Loading the bundle will not raise snapshots, because the
+          // mutation is not acknowledged.
```
schmidt-sebastian, line 223: s/is not/has not been/

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -0,0 +1,368 @@
...
+      content: { key: 'b' }
+    });
+    const limboQuery = Query.atPath(docA.key.path);
+
+    return (
+      spec()
+        .withGCEnabled(false)
+        .userListens(query)
+        .watchAcksFull(query, 250)
+        // Backend tells there is no such doc.
```
schmidt-sebastian, line 246: s/tells/tells us

packages/firestore/src/core/firestore_client.ts
```diff
@@ -512,4 +516,27 @@ export class FirestoreClient {
...
+      content = new TextEncoder().encode(data);
+    } else {
+      content = data;
+    }
+    const reader = new BundleReader(toByteStreamReader(content));
+    const task = new LoadBundleTask();
+    this.asyncQueue.enqueueAndForget(async () => {
+      loadBundle(this.syncEngine, reader, task);
+      return task.catch(e => {
+        logWarn(`Loading bundle failed with ${e}`);
```
schmidt-sebastian, line 536: You should add a LOG_TAG (see most other log lines).

packages/firestore/src/api/bundle.ts
```diff
@@ -0,0 +1,111 @@
...
+import * as firestore from '@firebase/firestore-types';
+import { Deferred } from '../util/promise';
+import { PartialObserver } from './observer';
+import { debugAssert } from '../util/assert';
+
+export class LoadBundleTask
+  implements
+    firestore.LoadBundleTask,
+    PromiseLike<firestore.LoadBundleTaskProgress> {
+  private _progressObserver?: PartialObserver<firestore.LoadBundleTaskProgress>;
```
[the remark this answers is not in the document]
wu-hui: Done.

packages/firestore/src/api/bundle.ts
wu-hui: As discussed, we would keep current behaviour because these progress updates can be considered part of the progress update series.

packages/firestore/src/core/bundle.ts
```diff
@@ -98,3 +107,127 @@ export class BundleConverter {
...
+    bytesLoaded: metadata.totalBytes!,
+    totalDocuments: metadata.totalDocuments!,
+    totalBytes: metadata.totalBytes!
+  };
+}
+
+export class BundleLoadResult {
+  constructor(
+    readonly progress: firestore.LoadBundleTaskProgress,
+    readonly changedDocs?: MaybeDocumentMap
```
schmidt-sebastian, line 146: This doesn't need to be optional anymore, which I think leads to one more simplification.

packages/firestore-types/index.d.ts
```diff
@@ -93,9 +93,40 @@ export class FirebaseFirestore {
...
 
+  loadBundle(
+    bundleData: ArrayBuffer | ReadableStream<ArrayBuffer> | string
+  ): LoadBundleTask;
+
   INTERNAL: { delete: () => Promise<void> };
 }
 
+export interface LoadBundleTask {
+  onProgress(
```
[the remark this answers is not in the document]
wu-hui: Yeah, it's actually in the API doc. I will add it in a followup PR.

packages/firestore/src/core/bundle.ts
wu-hui: Done.

PR #3205, opened 2020-06-11
Remove usages of goog.isFunction

kjin:
goog.isFunction has been deprecated since Q2 2020.


kjin:
Ping @bojeil-google @wti806 -- could you take a look? Thanks!

PR #3206, opened 2020-06-11
Replace Platform injection with source preprocessor

schmidt-sebastian:
So this will either get my commit access revoked or result in lot of external job offers.

One of the problems with our client initialization is that we use a single Platform class to inject all platform specific dependencies. In the Lite and the Tree-Shakeable SDK this means that pulling in any code at all will bring in the platform specific dependencies.

This PR removes the Platform class and instead replaces each API with a top-level function. It then uses rollup and webpack plugins to replace the a generic platform with the target platform. 
Due to some dependency conflict with ts-node, I had to extract out "formatJSON" and "randomBytes" into separate files.


schmidt-sebastian:
I just realized that we might also be able to do this if we publish an internal `@firebase/firestore/platform` package that exposes different builds for node, browser and react-native. I can explore that next week if that seems cleaner.

packages/firestore/src/platform_node/node_platform.ts
```diff
@@ -15,72 +15,71 @@
...
-  randomBytes(nBytes: number): Uint8Array {
-    debugAssert(nBytes >= 0, `Expecting non-negative nBytes, got: ${nBytes}`);
+/** The Platform's 'document' implementation or null if not available. */
+export function getDocument(): Document | null {
+  return null;
+}
 
-    return randomBytes(nBytes);
-  }
+/** True if and only if the Base64 conversion functions are available. */
```
wilhuff, line 82: This seems related to the other base64 methods--maybe group these together?

packages/firestore/src/platform_rn/rn_platform.ts
```diff
@@ -15,22 +15,36 @@
...
-import { BrowserPlatform } from '../platform_browser/browser_platform';
 import { base64 } from '@firebase/util';
 
-// Implements the Platform API for ReactNative.
-export class ReactNativePlatform extends BrowserPlatform {
-  readonly base64Available = true;
+// ReactNative-specific platform implementation that provides its own Base64
+// encoding.
+// The exports in this class must match the exports in '../platform/platform' as
+// they are bundled with the browser build during the Rollup build.
```
wilhuff, line 23: These aren't bundled with the browser build--it's the react native, build, right?

packages/firestore/src/util/types.ts
```diff
@@ -47,3 +47,18 @@ export function isSafeInteger(value: unknown): boolean {
     value >= Number.MIN_SAFE_INTEGER
   );
 }
+
+/** The subset of the Window interface used by the SDK. */
```
wilhuff, line 51: Is `Window` universally available? I thought it was specific to the Browser? If so, minor nit: this is a subset of the browser's Window class. Same nit below.

packages/firestore/src/platform/format_json.ts
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as node from '../platform_node/node_format_json';
+import * as browser from '../platform_browser/browser_format_json';
+import { isNode } from '@firebase/util';
+
+// This file provides additional platform specific APIs. It acts an add on to
```
Feiyang1, line 22: missing "as" after acts

packages/firestore/src/platform/platform.ts
```diff
@@ -15,69 +15,64 @@
...
-export interface Platform {
-  loadConnection(databaseInfo: DatabaseInfo): Promise<Connection>;
-  newConnectivityMonitor(): ConnectivityMonitor;
-  newSerializer(databaseId: DatabaseId): JsonProtoSerializer;
+import * as node from '../platform_node/node_platform';
+import * as browser from '../platform_browser/browser_platform';
 
-  /** Formats an object as a JSON string, suitable for logging. */
-  formatJSON(value: unknown): string;
+// NOTE: This class is replaced in Firestore release builds by either
```
Feiyang1, line 28: s/class/module

packages/firestore/src/platform/random_bytes.ts
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as node from '../platform_node/node_random_bytes';
+import * as browser from '../platform_browser/browser_random_bytes';
+import { isNode } from '@firebase/util';
+
+// This file provides additional platform specific APIs. It acts an add on to
```
Feiyang1, line 22: missing "as" after acts

config/webpack.test.js
```diff
@@ -18,6 +18,9 @@
 const path = require('path');
 const webpack = require('webpack');
 
+/** A regular expression used to replace Firestore' platform specific modules. */
```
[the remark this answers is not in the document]
schmidt-sebastian: I expanded the comment and changed the RegExp to match all files. This is cleaner, but I initially wanted to keep it more restricted because I am afraid that it will start matching files in other platforms. Unfortunately, I haven't found a way to restrict it to just "packages/firestore" since the paths that are used are the relative import paths we specify in our sources. 

As a general heads up - I debated getting rid of the platform.ts file before, but kept it as is during the first review. I have since moved all implementations to specific "category" files (base64.ts, serializer.ts, dom.ts, etc). These seems cleaner that having two files that are outliers.

packages/firestore/lite/src/api/database.ts
```diff
@@ -83,19 +83,11 @@ export class Firestore implements firestore.FirebaseFirestore {
...
-          databaseInfo.databaseId
-        );
-        const datastore = newDatastore(
-          connection,
-          this._credentials,
-          serializer
-        );
-        this._datastoreDeferred.resolve(datastore);
-      });
+    loadConnection(databaseInfo).then(connection => {
```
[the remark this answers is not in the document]
schmidt-sebastian: Updated to `newConnection`.

packages/firestore/src/platform/format_json.ts
schmidt-sebastian: Moved this to src/platform/README.md, which contains different text.

packages/firestore/src/platform/platform.ts
schmidt-sebastian: Addressed in README.md

packages/firestore/src/platform/platform.ts
```diff
@@ -15,69 +15,64 @@
...
+// "node_platform" or "browser_platform". The implementations below are only
+// used by ts-node to run tests.
 
-  /** Converts a Base64 encoded string to a binary string. */
-  atob(encoded: string): string;
+/** Loads the network stack (GRPC or WebChannel). */
+export function loadConnection(
+  databaseInfo: DatabaseInfo
+): Promise<Connection> {
+  return isNode()
```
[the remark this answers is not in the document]
schmidt-sebastian: I don't think this would be possible unless we define an interface as an intermediate layer. It did however dawn on me that this file is never used in browser builds. I ended up replacing the `isNode` check with an assert.

packages/firestore/src/platform/random_bytes.ts
schmidt-sebastian: Removed

packages/firestore/src/platform_node/node_platform.ts
schmidt-sebastian: They are now group in the "base64" module.

packages/firestore/src/platform_rn/rn_platform.ts
schmidt-sebastian: That's correct. The comment has been removed.

packages/firestore/src/util/types.ts
schmidt-sebastian: Fixed

packages/firestore/src/platform/README.md
```diff
@@ -0,0 +1,4 @@
+The modules in this directory provide implementations for platform-specific features that are replaced at build time
```
wilhuff, line 1: Nit: use the same line length as our sources?

packages/firestore/src/platform/connection.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
wilhuff, line 3: 2020? Here and throughout.

packages/firestore/src/platform/connection.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isNode, isReactNative } from '@firebase/util';
+import {
```
wilhuff, line 19: This is pretty hefty amount of code to make all these distinct imports. If this is what it takes I guess this is what we're stuck with, but it doesn't seem like this file needs to be tree-shakeable so whole module imports seem like they'd work and would cost way less.

schmidt-sebastian: It is pretty ugly indeed. I reverted back to namespace imports.

packages/firestore/src/platform/connection.ts
schmidt-sebastian: Fixed

PR #3212, opened 2020-06-12
Add createUserWithEmailAndPassword to auth-next

avolkovi:


packages-exp/auth-exp/src/core/strategies/email_and_password.test.ts
```diff
@@ -324,6 +323,45 @@ describe('core/strategies/verifyPasswordResetCode', () => {
...
+    const {
+      credential,
+      user,
+      operationType
+    } = await createUserWithEmailAndPassword(
+      auth,
+      'some-email',
+      'some-password'
+    );
+    expect(credential?.providerId).to.eq(ProviderId.PASSWORD);
```
yuchenshi, line 357: nit: Please use `credential!.providerId` if you are asserting that `credential` is not null and has a field. In this way, when `credential` is null you'll get a proper runtime error message saying "cannot read property `providerId` of null", instead of "expected null to equal some-password". The latter doesn't differentiate whether `credential` is null or `credential.providerId` is.

yuchenshi: Ditto for similar assertions below.

avolkovi: done

samhorlbeck:
Looks good, just add the methods to `src/index.ts`

avolkovi:
updated with index.ts

PR #3213, opened 2020-06-12
Add linkWithCredential(), linkWithPhoneNumber(), unlink()

samhorlbeck:


samhorlbeck:
cc: @yuchenshi

packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -100,16 +101,27 @@ export class UserImpl implements User {
     return reload(this);
   }
 
-  _updateTokensIfNecessary(response: IdTokenResponse): boolean {
+  async _updateTokensIfNecessary(
+    response: IdTokenResponse,
+    reload = false
+  ): Promise<void> {
```
avolkovi, line 107: we don't need the return value anymore?

packages-exp/auth-exp/src/core/user/unlink.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import * as externs from '@firebase/auth-types-exp';
+
+import { deleteLinkedAccounts } from '../../api/account_management/account';
+import { User } from '../../model/user';
+import { _assertLinkedStatus } from '../strategies/credential';
+
+export async function unlink(
+  userExtern: externs.User,
+  providerId: externs.ProviderId
```
avolkovi, line 26: should we have a version of this that takes multiple providers to match the API?

packages-exp/auth-exp/src/core/user/unlink.ts
```diff
@@ -0,0 +1,46 @@
...
+  const user = userExtern as User;
+  await _assertLinkedStatus(true, user, providerId);
+  const { providerUserInfo } = await deleteLinkedAccounts(user.auth, {
+    idToken: await user.getIdToken(),
+    deleteProvider: [providerId]
+  });
+
+  const providersLeft = (providerUserInfo || []).map(i => i.providerId);
+
+  user.providerData = user.providerData.filter(pd =>
```
avolkovi, line 37: wouldn't a reload be easier and more consistent here?

packages-exp/auth-exp/src/core/user/unlink.ts
```diff
@@ -0,0 +1,46 @@
...
+    providersLeft.includes(pd.providerId || undefined)
+  );
+  if (!providersLeft.includes(externs.ProviderId.PHONE)) {
+    user.phoneNumber = null;
+  }
+
+  await user.auth._persistUserIfCurrent(user);
```
avolkovi, line 44: do we need to notify? feels like we should notify listeners

packages-exp/auth-exp/src/core/user/unlink.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      endpoint = mockEndpoint(Endpoint.SET_ACCOUNT_INFO, {
+        providerUserInfo: [
+          {
+            providerId: ProviderId.GOOGLE
+          }
+        ]
+      });
+    });
+
+    it('removes the provider from the list and persists', async () => {
```
avolkovi, line 84: can you add a separate test for removing phone vs removing another provider?

packages-exp/auth-exp/src/core/strategies/phone.ts
```diff
@@ -37,10 +41,10 @@ class ConfirmationResult implements externs.ConfirmationResult {
   ) {}
 
   confirm(verificationCode: string): Promise<externs.UserCredential> {
-    const authCredential = PhoneAuthProvider.credential(
-      this.verificationId,
+    const authCredential = new PhoneAuthCredential({
```
avolkovi, line 44: why use the constructor directly vs the static factory?

packages-exp/auth-exp/src/core/strategies/credential.ts
```diff
@@ -39,3 +46,52 @@ export async function signInWithCredential(
...
+  // TODO: Handle Oauth cases
+  return null;
+}
+
+export async function _assertLinkedStatus(
+  expected: boolean,
+  user: User,
+  provider: externs.ProviderId
+): Promise<void> {
+  await _reloadWithoutSaving(user);
```
avolkovi, line 88: there's a potential race condition if this gets called twice, do we care?

packages-exp/auth-exp/src/core/strategies/credential.ts
```diff
@@ -39,3 +46,52 @@ export async function signInWithCredential(
...
+  const providerIds = user.providerData.map(({ providerId }) => providerId);
+  if (providerIds.includes(provider) !== expected) {
```
avolkovi, line 90: this is similar to the filter you have below, should we provide an accessor method on user that fetches a particular provider?

avolkovi, line 90: would this `if` be simpler as an `assert`?

packages-exp/auth-exp/src/core/strategies/credential.ts
```diff
@@ -39,3 +46,52 @@ export async function signInWithCredential(
...
+  await _assertLinkedStatus(false, user, credential.providerId);
+
+  const response = await credential._linkToIdToken(
+    user.auth,
+    await user.getIdToken()
+  );
+
+  const newCred = _authCredentialFromTokenResponse(response);
+  await user._updateTokensIfNecessary(response, /* reload */ true);
+  return new UserCredentialImpl(user, newCred, OperationType.LINK);
```
avolkovi, line 65: shouldn't this be `newCred || credential`?

packages-exp/auth-exp/src/core/strategies/credential.test.ts
```diff
@@ -63,32 +69,115 @@ describe('core/strategies/signInWithCredential', () => {
...
+      expect(operationType).to.eq(OperationType.LINK);
+      expect(newUser).to.eq(user);
+      expect(credential).to.be.null;
+    });
   });
 
-  it('should update the current user', async () => {
-    const { user } = await signInWithCredential(auth, authCredential);
-    expect(auth.currentUser).to.eq(user);
+  describe('_assertLinkedStatus', () => {
```
avolkovi, line 137: since all the phone stuff behaves so differently, can you add a separate set of tests for linking a non-phone cred?

packages-exp/auth-exp/src/core/credentials/phone.ts
```diff
@@ -46,9 +47,10 @@ export class PhoneAuthCredential
   }
 
   _linkToIdToken(auth: Auth, idToken: string): Promise<IdTokenResponse> {
-    void auth;
-    void idToken;
-    return debugFail('not implemented');
+    return linkWithPhoneNumber(auth, {
+      idToken,
+      ...this.makeVerificationRequest()
```
avolkovi, line 52: 💯

packages-exp/auth-exp/src/core/user/user_impl.ts
samhorlbeck: No I refactored all of this so that it handles the persistence + notification if necessary

packages-exp/auth-exp/src/core/user/unlink.ts
samhorlbeck: Maybe? Not sure if we'd need to go though a review pass for that

packages-exp/auth-exp/src/core/user/unlink.ts
samhorlbeck: This is what the current sdk does. I'm not sure _why_ exactly but I'm inclined to keep it that way

packages-exp/auth-exp/src/core/user/unlink.ts
samhorlbeck: This doesn't update tokens. We only notify if tokens get updated

packages-exp/auth-exp/src/core/strategies/credential.ts
samhorlbeck: Discussed offline, it should be fine

packages-exp/auth-exp/src/core/strategies/credential.ts
samhorlbeck: No:
https://github.com/firebase/firebase-js-sdk/blob/712dfde546a89af6778ab2f6c787bfd47abbd3b4/packages/auth/src/authuser.js#L1330

packages-exp/auth-exp/src/core/strategies/credential.test.ts
samhorlbeck: Done. I just switched these over to GOOGLE.. The provider in question doesn't actually matter for this set of tests

packages-exp/auth-exp/src/core/strategies/credential.ts
samhorlbeck: Added a utility method. Doesn't live on user though since down below the source isn't user

samhorlbeck: Done

packages-exp/auth-exp/src/core/user/unlink.test.ts
samhorlbeck: Done

PR #3217, opened 2020-06-16
Cleanup public types in auth-next

avolkovi:


packages-exp/auth-exp/src/core/providers/phone.ts
```diff
@@ -41,7 +42,11 @@ export class PhoneAuthProvider implements externs.PhoneAuthProvider {
     applicationVerifier: externs.ApplicationVerifier
     /* multiFactorSession?: MultiFactorSession, */
```
samhorlbeck, line 43: This can be removed

packages-exp/auth-exp/src/core/providers/phone.ts
```diff
@@ -41,7 +42,11 @@ export class PhoneAuthProvider implements externs.PhoneAuthProvider {
     applicationVerifier: externs.ApplicationVerifier
```
samhorlbeck, line 42: This should be
`verifyPhoneNumber(phoneNumber: externs.PhoneInfoOptions | string, ...`

packages-exp/auth-exp/src/index.ts
```diff
@@ -15,9 +15,42 @@
  * limitations under the License.
  */
 
+import * as externs from '@firebase/auth-types-exp';
+import { ErrorFn, CompleteFn, Unsubscribe } from '@firebase/util';
+
 // core/auth
 export { initializeAuth } from './core/auth/auth_impl';
 
+// Non-optional auth methods.
```
samhorlbeck, line 24: I liked having the index.ts file here be 100% re-exported functions. WDYT about moving these methods into `src/auth/util.ts` or something along those lines?

packages-exp/auth-exp/src/index.ts
```diff
@@ -67,11 +99,10 @@ export {
...
 export { unlink } from './core/user/unlink';
 
-// model
-export { Operation as ActionCodeOperationType } from './model/action_code_info';
-
-// platform-browser/recaptcha
-export { RecaptchaVerifier } from './platform_browser/recaptcha/recaptcha_verifier';
+// Non-optional user methods.
+export { reload } from './core/user/reload';
+export async function deleteUser(user: externs.User): Promise<void> {
```
samhorlbeck, line 106: Same here, this could probably live under `src/core/user/`

packages-exp/auth-exp/src/core/providers/phone.ts
avolkovi: done

packages-exp/auth-exp/src/core/providers/phone.ts
avolkovi: done

packages-exp/auth-exp/src/index.ts
avolkovi: possibly! let's hold off until we write the polyfill, I imagine we will have a lot of refactoring then

PR #3221, opened 2020-06-16
Fix FCM rxjs Incompatibility 

### LINKED ISSUE #3142: FCM rxjs compatibility broken?
FCM API suggests that it offers rxjs compatibility out of box. Specifically, you could do
```
let foregroundMessages$ = rxjs.Observable.create(function subscribe(observer) {
  messaging.onMessage(observer);
});

foregroundMessages$.subscribe({
  next: (msg) => handleForegroundMessage(msg),
  error: (err) => console.log("err: " + err),
});
```

However, an error is thrown:
```
Uncaught (in promise) TypeError: this._next is not a function
    at rt.Subscriber.next [as onMessageCallback] (rxjs.umd.js:472)
    at rt.<anonymous> (window-controller.ts:197)
    at tslib.es6.js:100
    at Object.next (tslib.es6.js:81)
    at tslib.es6.js:74
    at new Promise (<anonymous>)
    at v (tslib.es6.js:70)
    at rt.messageEventListener (window-controller.ts:188)
    at ServiceWorkerContainer.<anonymous> (window-controller.ts:47)
```

Came across this error here as well (https://github.com/angular/angularfire/issues/2299)

zwu52:
Fixes #3142 

Adding a field variable to hold rxjs observer or subscriber. It enables onMessage(nextFn) and onMessage(observer) to both execute  if both are set.


packages/messaging/src/controllers/window-controller.ts
```diff
@@ -133,13 +136,15 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
   onMessage(nextOrObserver: NextFn<object> | Observer<object>): Unsubscribe {
-    this.onMessageCallback =
-      typeof nextOrObserver === 'function'
-        ? nextOrObserver
-        : nextOrObserver.next;
+    if (typeof nextOrObserver === 'function') {
+      this.onMessageCallback = nextOrObserver;
+    } else {
+      this.onMessageObserver = nextOrObserver;
+    }
```
Feiyang1, line 143: This allows registering 2 listeners and is possibly a breaking change. I don't think we should change the current behavior of single listener. On way to do it can be remove `onMessageCallback` field, convert the callback to an observer and assign it to `onMessageObserver`.

zwu52: I can't think of a case where this could break an application. Maybe some unexpected behaviors? But I agree that this can be done more elegantly with one callback since there is really no need to have two.

.changeset/tough-rings-bake.md
```diff
@@ -0,0 +1,6 @@
+---
+'firebase': minor
```
Feiyang1, line 2: I believe it should be a patch release, since there is no API change.

Issue #3158, opened 2020-06-03
Firebase storage UploadTask callback type requires all callbacks


nicholaschiang:
<!-- Step 1 [READ THIS] -->
<!--
Are you in the right place?
  * For issues or feature requests related to __the code in this repository__
    file a Github issue.
    * If this is a __feature request__ make sure the issue title starts with "FR:".
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).
-->

<!-- Step 2 -->

### [REQUIRED] Describe your environment

  * Operating System version: Ubuntu 18.04.2
  * Browser version: N/A
  * Firebase SDK version: 7.14.2
  * Firebase Product: storage

<!-- Step 3 -->

### [REQUIRED] Describe the problem

#### Steps to reproduce:
<!--
  What happened? How can we make the problem occur?
  This could be a description, log/console output, etc.
-->
The `UploadTask` callback type doesn't match what is [documented here](https://firebase.google.com/docs/reference/js/firebase.storage.UploadTask#on):
> Callbacks can be passed either as three separate arguments or as the next, error, and complete properties of an object. **Any of the three callbacks is optional, as long as at least one is specified.** In addition, when you add your callbacks, you get a function back. You can call this function to unregister the associated callbacks.

But the `UploadTask` callback parameter type seems to require all three callbacks (i.e. it needs an object with `next`, `complete`, and `error` fields).

#### Relevant Code:

So I've called the `UploadTask` `on()` method using an object containing the `error` and `complete` callbacks as follows:
```typescript
 ref.put(file).on(firebase.storage.TaskEvent.STATE_CHANGED, {                                                                                                                                                                 
   async complete() {                                                                                                                                                                                                         
     onChange((await ref.getDownloadURL()));                                                                                                                                                                                  
     setValue(`Uploaded ${filename}.`);                                                                                                                                                                                       
   },                                                                                                                                                                                                                         
   error(error: Error) {                                                                                                                                                                                                      
     setErrored(true);                                                                                                                                                                                                        
     setValue(`An error occurred while uploading ${filename}. ${error.message}`);                                                                                                                                             
[...]

PR #3224, opened 2020-06-16
Allow undefined callbacks in Storage Observer

schmidt-sebastian:
Matches both the documentation as well as the internal type: https://github.com/firebase/firebase-js-sdk/blob/master/packages/storage/src/implementation/observer.ts#L25

Fixes https://github.com/firebase/firebase-js-sdk/issues/3158


schmidt-sebastian:
Note: This requires our users to be on TypeScript 2.1, which might be an existing requirement.

schmidt-sebastian:
@hsubox76 Do you have time to take a look?

Issue #3144, opened 2020-05-26
Database SDK tries to access credential even when FIREBASE_DATABASE_EMULATOR_HOST is set


filipesilva:
### [REQUIRED] Environment info

**firebase-tools:** 8.4.0

**Platform:** Windows

### [REQUIRED] Test case

I have a test case with minimal cloud function code and emulator at https://github.com/filipesilva/firebase-emulator-credentials.
```
const functions = require('firebase-functions');
const admin = require('firebase-admin');
admin.initializeApp();

exports.helloWorld = functions.https.onRequest(async (request, response) => {
  const db = admin.database();
  const ref = await db.ref('auth-test').set({
    date: new Date().toISOString()
  });
  response.send(`Added: ${ref.id}`);
});
```

### [REQUIRED] Steps to reproduce

```
git clone https://github.com/filipesilva/firebase-emulator-credentials
cd firebase-emulator-credentials
yarn
yarn firebase logout
yarn emulators
# open http://localhost:5001/emulator-test-1/us-central1/helloWorld in the browser
```

You should see the following error log in the console:
```
kamik@RED-X1C6 MINGW64 /d/sandbox/firebase-emulator-credentials
$ yarn emulators
yarn run v1.21.1
$ firebase emulators:start
!  emulators: You are not currently authenticated so some features may not work correctly. Please run firebase login to authenticate the CLI.
i  emulators: Starting emulators: functions, database
!  Your requested "node" version "8" doesn't match your global version "12"
i  database: Database Emulator logging to database-debug.log
!  hosting: The hosting emulator is configured but there is no hosting configuration. Have you run firebase init hosting?
i  ui: Emulator UI logging to ui-debug.log
i  functions: Watching "D:\sandbox\firebase-emulator-credentials\functions" for Cloud Functions...
+  functions[helloWorld]: http function initialized (http://localhost:5001/emulator-test-1/us-central1/helloWorld).

┌───────────────────────────────────────────────────────────────────────┐
│ ✔  All emulators ready! View status and logs at http://localhost:4000 │
└───────────────────────────────────────────────────────────────────────┘

┌───────────┬────────────────┬─────────────────────────────────┐
│ Emulator  │ Host:Port      │ View in Emulator UI             │
├───────────┼────────────────┼─────────────────────────────────┤
│ Functions │ localhost:5001 │ http://localhost:4000/functions │
├───────────┼────────────────┼─────────────────────────────────┤
[...]

PR #3228, opened 2020-06-17
Use Admin AuthTokenProvider when targeting Emulator

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-js-sdk/issues/3144


yuchenshi:
I'll be brief here. I believe this will break two use cases:

1) Admin SDK authenticating as Admin (where `Authorization: Bearer owner` must be specified, instead of empty.)
2) Developers using production Firebase Auth to login a production user and then write to local RTDB emulator using the production JWT.

Feel free to correct me if I'm wrong, but I may be overly cautious since we had a similar change before that broke admin and firebase-testing. If you can correct me by testing admin, production auth, and `@firebase/testing`, that will be even better. Bonus points if those tests can be automated.

schmidt-sebastian:
@yuchenshi Thanks for your comment. I am kind of stuck here now, but will try to come up with a way to address your concerns. I hope we can do more than just silence the logs.

filipesilva:
@schmidt-sebastian did you perhaps come across a workaround that could be used meanwhile? It looks like this fix ended up being a bit more complicated than it looked like. I'm happy enough working around it on my side for the time being, if that's an option.

schmidt-sebastian:
I haven't been able to come up with a way to solve this - as far as I can tell, we might just have to disable the log output. Does anything actually break on your end, or are you mostly (and rightfully) concerned about the log output?

filipesilva:
Anything that involves using recent emulator version with RTDB and without having GOOGLE_APPLICATION_CREDENTIALS set will fail. 

In the original repro in https://github.com/firebase/firebase-js-sdk/issues/3144, the cloud fn times out with RTDB, but not with Firestore, because of this problem. On our setup we're stuck on `firebase-tools@7.12.1` and can't update because of it. For some reason older emulator versions are ok.

packages/database/src/core/AuthTokenProvider.ts
```diff
@@ -104,3 +111,13 @@ export class AuthTokenProvider {
     warn(errorMessage);
   }
 }
+
+/* Auth token provider used to connect to the Emulator. */
+export class EmptyAuthTokenProvider implements AuthTokenProvider {
+  getToken(forceRefresh: boolean): Promise<FirebaseAuthTokenData> {
+    return Promise.resolve({ accessToken: '' });
```
samtstern, line 118: @yuchenshi can comment but I think the Admin SDKs are supposed to send the special token `owner` when talking to the emulators, rather than blank.

samtstern:
@schmidt-sebastian do we have any way to detect if we're in an Admin or Client context?  If so the behavior matrix looks like this:

- In client context...
   - Host is emulator: use Firebase Auth
   - Host is prod: use Firebase Auth
- In admin context...
  - Host is emulator: use `EmptyTokenProvider` that always returns exactly `owner`
  - Host is prod: use standard cloud auth

packages/database/src/core/RepoManager.ts
```diff
@@ -116,6 +123,11 @@ export class RepoManager {
       dbUrl = `http://${dbEmulatorHost}?ns=${repoInfo.namespace}`;
```
samtstern, line 123: Does this branch `if (dbEmulatorHost) { }` trigger only if the emulator connection is made via env var?  What about if the user does this:

```js
firebase.initializeApp({
  databaseURL: 'http://localhost:9000?ns=namespace'
})
``` 

You might consider using the `EmulatorAdminTokenProvider` any time `Constants.NODE_ADMIN` is true and the database is `http://` (not `https://`).  That's what we do in the Firebase CLI, it catches more cases and it also has the nice side benefit of not allowing someone's admin credentials to ever go to an insecure host.

packages/database/src/core/RepoManager.ts
```diff
@@ -108,16 +113,30 @@ export class RepoManager {
...
     }
+
     if (dbEmulatorHost) {
+      isEmulator = true;
       dbUrl = `http://${dbEmulatorHost}?ns=${repoInfo.namespace}`;
       parsedUrl = parseRepoInfo(dbUrl);
       repoInfo = parsedUrl.repoInfo;
+    } else {
+      isEmulator =
+        parsedUrl.repoInfo.host === 'localhost' && !parsedUrl.repoInfo.secure;
```
samtstern, line 132: We do see people running their emulator on non-localhost hosts.  Like `0.0.0.0`.  I think just anything non-secure (`http://`) is the right condition.

schmidt-sebastian: Done. I will verify this manually on Monday, which should allow us to release this next week.

filipesilva:
@schmidt-sebastian if there's a snapshot build somewhere that I can pull in via npm I can test on my setup too.

schmidt-sebastian:
I manually verified using Sam's testing matrix.

@filipesilva We will publish a prerelease version on NPM, likely tomorrow.

samtstern:
So it looks like `firebase-admin-node` `9.0.0` went out yesterday ... does that mean this bugfix missed the last `8.x` release?

schmidt-sebastian:
@samtstern `firebase-admin-node` 8.x should pick this up on re-install since we don't pin versions.

PR #3231, opened 2020-06-17
Tree-Shakeable Serialization

schmidt-sebastian:
This PR replaces the class methods in JSONSerializer with free standing functions, which allows for a lot of tree shaking (see binary size report).

This PR is split in three parts:
- fad9e64 is mostly a mechanical Refactor in IntelliJ
- ~~`005abb3` is a manual change that uses an options interface in lieu of the empty serializer class.~~
-  2e9c758 is a small cleanup.
- 9c4663a is an auto-generated update to the dependency file in the Lite SDK.

This PR is a bit easier to review by appending "?w=1" to the URL.


packages/firestore/lite/src/api/database.ts
```diff
@@ -93,10 +93,14 @@ export class Firestore implements firestore.FirebaseFirestore {
       this._datastorePromise = PlatformSupport.getPlatform()
         .loadConnection(databaseInfo)
         .then(connection => {
-          const serializer = PlatformSupport.getPlatform().newSerializer(
+          const serializationOptions = PlatformSupport.getPlatform().newSerializationOptions(
```
wilhuff, line 96: In tree-shakeable style, the structure containing the data we pass through that used to be a class is now just a dumb struct, but it still has the same identity. All that's changed is that it's a dumb struct with accompanying functions instead of a class with member functions.

For parity with the other platforms I think it still makes sense to call that thing a `Serializer`.

schmidt-sebastian: The parity argument makes sense. I updated the PR accordingly.

packages/firestore/src/remote/serializer.ts
```diff
@@ -99,995 +103,1083 @@ function assertPresent(value: unknown, description: string): asserts value {
...
-    } else {
-      result = val;
-    }
-    return isNullOrUndefined(result) ? null : result;
-  }
+/**
+ * Returns a number (or null) from a google.protobuf.Int32Value proto.
+ */
+function fromInt32Proto(
+  serializer: JsonProtoSerializer,
```
wu-hui, line 156: `serializer` is not needed here.

schmidt-sebastian: Fixed (also in a couple other places)

PR #3234, opened 2020-06-18
FCM Pre Modularization

zwu52:
[ariane/4026631 IS APPROVED]

This PR implement "Milestone I" in [section: API Change Plan: Deprecation, Removal, Addition](go/fcm-web-modularization). Specially, this PR:

### Mark Deprecate:
- setBackgroundHandler (instead divert developer to use onBackgroundMessage)
- onTokenRefresh (currently no-op and we have no plan to implement in foreseeable future)
- useVapidKey (divert developer to use getToken(options:{serviceWorkerRegistration, vapidKey})
- useServiceWorker (divert developer to use getToken(options:{serviceWorkerRegistration, vapidKey})
- getToken() (divert developer to use getToken(options:{serviceWorkerRegistration, vapidKey})
### Add:
- getToken(options:{serviceWorkerRegistration, vapidKey}
- onBackgroundMessage
- Typing <MessagePayload>  for onBackgroundMessage: this would make easy for user to know what parameters are gettable.


integration/messaging/test/static/helpers.js
```diff
@@ -0,0 +1,51 @@
...
+    addPayloadToDbInternal(db, {
+      ...payload,
+      // ndx is required as the primary key of the store. It doesn't have any other testing purpose
+      ndx: BACKGROUND_MESSAGES_OBJECT_STORE_DEFAULT_NDX
+    });
+  };
+}
+
+async function addPayloadToDbInternal(db, payload) {
+  // onsuccess might race with onupgradeneeded. Consequently causing "object stores was not found" error. Therefore, wait briefly for db.createObjectStore to complete
```
kroikie, line 43:
```suggestion
  // onsuccess might race with onupgradeneeded. Consequently causing "object stores was not found" error.
  // Therefore, wait briefly for db.createObjectStore to complete
```

integration/messaging/test/static/valid-vapid-key-modern-sw/sw.js
```diff
@@ -15,16 +15,23 @@
...
-    payload: MessagePayload;
-  };
-}
+firebase.initializeApp(FIREBASE_CONFIG);
+const messaging = firebase.messaging();
+
+messaging.onBackgroundMessage(payload => {
+  console.log(
+    TAG +
+      'a background message is received in the onBackgroundMessage hook: ' +
```
kroikie, line 31: Should there be a space after `TAG` or before `a background ...`?

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,49 @@ declare namespace firebase.messaging {
    *
    */
   interface Messaging {
+    /**
+     * Deletes the only registration token associated with this messaging instance and unsubscribes
```
kroikie, line 6953: Remove the word `only`, it is unnecessary.

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,49 @@ declare namespace firebase.messaging {
...
      * @return The promise resolves when the token has been
      *   successfully deleted.
+     *
+     * @deprecated Use deleteToken() instead.
      */
     deleteToken(token: string): Promise<boolean>;
+
+    /**
+     * Subscribes the user to push notifications. Returns an FCM registration
+     * token that can be used to send push messages to the user.
```
kroikie, line 6974: I would avoid using the term "user" here since the token identifies the instance and not the user in particular.

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,49 @@ declare namespace firebase.messaging {
...
+     *
+     * If notification permission isn't already granted, this method asks the
+     * user for permission. The returned promise rejects if the user does not
+     * allow the app to show notifications.
+     *
+     * @param options.vapidKey the public server key provided to push services. It is used to authenticate
+     * the push subscribers to receive push messages only from sending servers that holds the corresponding private key. If it is not provided, a default VAPID key will be be used. Note that some push services (Chrome Push Service) require a non-default VAPID key. Therefore, it is recommended to to generate and import a VAPID key for your project with {@link https://firebase.google.com/docs/cloud-messaging/js/client#configure_web_credentials_with_fcm Configure Web Credentials with FCM}. Also See {@link https://developers.google.com/web/fundamentals/push-notifications/web-push-protocol The Web Push Protocol} for details on web push services.}
```
kroikie, line 6981: Please wrap this line.

packages/firebase/index.d.ts
```diff
@@ -6988,6 +7025,25 @@ declare namespace firebase.messaging {
       error?: firebase.ErrorFn,
       completed?: firebase.CompleteFn
     ): firebase.Unsubscribe;
+
+    /**
+     * Called when a message is received while the app is in the background. An app is considered as a background app if no active window is displayed.
```
kroikie, line 7030: Looks good, please wrap.

packages/firebase/index.d.ts
```diff
@@ -6998,12 +7054,15 @@ declare namespace firebase.messaging {
      *     is called when a token refresh has occurred.
      * @return To stop listening for token
      *   refresh events execute this returned function.
+     *
+     * @deprecated There is no need to handle token rotation.
```
kroikie, line 7058: This makes me so happy!!!

packages/firebase/index.d.ts
```diff
@@ -7030,17 +7089,81 @@ declare namespace firebase.messaging {
...
+
+    /**
+     * See {@link firebase.Messaging.FcmOptions}.
+     */
+
+    fcmOptions?: FcmOptions;
+  }
+
+  /**
+   * Options for features provided by the FCM SDK for Web. See more {@link https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#webpushfcmoptions}.
```
kroikie, line 7140: Could we add a bit more context here as to why these might be necessary. A `for example` may be useful.

integration/messaging/test/static/helpers.js
kroikie: Seems like the wrapping was undone in several places below so similar fix will be needed there.

integration/messaging/test/static/valid-vapid-key-modern-sw/sw.js
zwu52: I have the TAG be set as "const TAG = 'FCM_INTEGRATION_TEST: " with an trailing space so  all reference to this constant should be correctly formatted

packages/firebase/index.d.ts
zwu52: good point. I changed it to "messaging instance" for accuracy

packages/firebase/index.d.ts
egilmorez: +1

packages/firebase/index.d.ts
egilmorez: I'm not currently seeing the updates Kai.  Do you need to push your latest?

packages/firebase/index.d.ts
zwu52: Got it. I added to the comment for the specific parameter

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,56 @@ declare namespace firebase.messaging {
...
+
+    /**
+     * Subscribes the messaging instance to push notifications. Returns an FCM registration
+     * token that can be used to send push messages to the user.
+     *
+     * If notification permission isn't already granted, this method asks the
+     * user for permission. The returned promise rejects if the user does not
+     * allow the app to show notifications.
+     *
+     * @param options.vapidKey the public server key provided to push services. It is used to
```
egilmorez, line 6980: Should this be a cap "T" ?

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,56 @@ declare namespace firebase.messaging {
...
+     * authenticate the push subscribers to receive push messages only from sending servers that
+     * holds the corresponding private key. If it is not provided, a default VAPID key will be used.
```
egilmorez, line 6982: Suggest "hold" plural, if in fact there can be multiple sending servers.

egilmorez, line 6982: Suggest "is used"

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,56 @@ declare namespace firebase.messaging {
...
+     * Note that some push services (Chrome Push Service) require a non-default VAPID key.
+     * Therefore, it is recommended to generate and import a VAPID key for your project with
+     * {@link https://firebase.google.com/docs/cloud-messaging/js/client#configure_web_credentials_with_fcm Configure Web Credentials with FCM}.
+     * See {@link https://developers.google.com/web/fundamentals/push-notifications/web-push-protocol The Web Push Protocol}
+     * for details on web push services.}
+     *
+     * @param options.serviceWorkerRegistration the service worker registration for receiving push messaging.
```
egilmorez, line 6989: I'm thinking that these are new sentences following the param literal . . . if so, they should start with caps.

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,56 @@ declare namespace firebase.messaging {
...
+     * If it is not provided explicitly, you need to have a `firebase-messaging-sw.js` at your root location.
```
egilmorez, line 6990: Suggest specifying "the registration" instead of "it"

packages/firebase/index.d.ts
```diff
@@ -6949,15 +6949,56 @@ declare namespace firebase.messaging {
...
+     * See {@link https://firebase.google.com/docs/cloud-messaging/js/client#retrieve-the-current-registration-token Retrieve the current registration token}
+     * for more details.
+     *
+     * @return The promise resolves with the FCM token string.
```
egilmorez, line 6994: Is the "FCM token string" the registration token?

It might be confusing in this context, but we prefer "registration token" as described here.  Your call.

https://firebase.google.com/docs/cloud-messaging/concept-options#credentials

packages/firebase/index.d.ts
```diff
@@ -6988,6 +7032,26 @@ declare namespace firebase.messaging {
       error?: firebase.ErrorFn,
       completed?: firebase.CompleteFn
     ): firebase.Unsubscribe;
+
+    /**
+     * Called when a message is received while the app is in the background. An app is considered
+     * as a background app if no active window is displayed.
```
egilmorez, line 7038: Suggest "as a background app" >> "to be in the background"

packages/firebase/index.d.ts
```diff
@@ -7030,17 +7097,101 @@ declare namespace firebase.messaging {
...
+    collapseKey: string;
+  }
+
+  /**
+   * Options for features provided by the FCM SDK for Web. See more {@link https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#webpushfcmoptions}.
+   */
+  export interface FcmOptions {
+    /**
+     * The link to open when the user clicks on the notification. For all URL values, HTTPS is required.
+     * For example, by setting this value to your app's URL, a notification click click event will
```
egilmorez, line 7164: Is this a typo?  Probably just one "click" works.

packages/firebase/index.d.ts
```diff
@@ -7030,17 +7097,101 @@ declare namespace firebase.messaging {
...
+     * put your app in focus for the user.
+     */
+    link?: string;
+
+    /**
+     * Label associated with the message's analytics data.
```
egilmorez, line 7170: Do you think an @link would be good here?  Your call.  It could be https://firebase.google.com/docs/cloud-messaging/understand-delivery#adding-analytics-labels-to-messages.

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -53,30 +67,137 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
+    }
+
+    return getToken(
+      this.firebaseDependencies,
+      this.swRegistration!,
+      this.vapidKey!
+    );
+  }
+
+  async updateVapidKey(vapidKey: string | undefined): Promise<void> {
```
Feiyang1, line 136: `vapidKey?: string`

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -53,30 +67,137 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
+      console.debug(
+        TAG +
+          'no VapidKey is provided. Using the default VapidKey. Note that Push will NOT work in Chrome without a non-default VapidKey.'
+      );
+      this.vapidKey = DEFAULT_VAPID_KEY;
+    }
+  }
+
+  async updateSwReg(
+    swRegistration: ServiceWorkerRegistration | undefined
```
Feiyang1, line 159: `swRegistration?: ServiceWorkerRegistration`

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -53,30 +67,137 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
-    return getToken(this.firebaseDependencies, swRegistration, this.vapidKey);
+    await this.updateVapidKey(options?.vapidKey);
+    await this.updateSwReg(options?.serviceWorkerRegistration);
+
+    if (!this.swRegistration) {
+      console.debug(
+        TAG +
+          'no sw has been provided explicitly. Attempting to find firebase-messaging-sw.js in default directory.'
+      );
+      await this.registerDefaultSw();
```
Feiyang1, line 126: Should we move it into `this.updateSwReg()` to be consistent with `updateVapidKey()`?
We assign the default vapid key to `this.vapidKey` inside `updateVapidKey()`. Similarly can we do the same for `this.swRegistration`?

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -53,30 +67,137 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
+    } catch (e) {
+      throw ERROR_FACTORY.create(ErrorCode.FAILED_DEFAULT_REGISTRATION, {
+        browserErrorMessage: e.message
+      });
+    }
   }
 
   async deleteToken(): Promise<boolean> {
-    const swRegistration = await this.getServiceWorkerRegistration();
+    if (!this.swRegistration) {
```
Feiyang1, line 196: Should it be no-op if there is no swRegistration?

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -102,7 +223,6 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
     }
   }
 
-  // TODO: Deprecate this and make VAPID key a parameter in getToken.
   usePublicVapidKey(vapidKey: string): void {
```
Feiyang1, line 226: Can you please annotate the deprecated methods with `@deprecated` in the source code as well?

packages/messaging/src/helpers/externalizePayload.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  return payload;
+}
+
+function propagateNotificationPayload(
+  payload: MessagePayload,
+  messagePayloadInternal: MessagePayloadInternal
+): void {
+  if (!messagePayloadInternal.notification) {
+    delete payload.notification;
```
Feiyang1, line 47: Instead of defining an empty object for `payload.notification`, then deleting it, it would be more efficient(marginally though) to only create an object and assign it to `payload.notification` if `messagePayloadInternal.notification` exists.

The same applies to `payload.data` and `payload.fcmOptions` as well.

packages/messaging/src/helpers/externalizePayload.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const body = messagePayloadInternal.notification!.body;
+  if (!!body) {
+    payload.notification!.body = body;
+  }
+
+  const image = messagePayloadInternal.notification!.image;
+  if (!!image) {
+    payload.notification!.image = image;
+  }
```
Feiyang1, line 64: Is it possible that all 3 properties are undefined? In that case, you could still get an empty object for `payload.notification`

packages/messaging/src/controllers/window-controller.test.ts
```diff
@@ -104,6 +108,134 @@ describe('WindowController', () => {
   });
 
   describe('getToken', () => {
+    it('uses default sw if none was registered nor provided', async () => {
+      assert.isUndefined(windowController.getSwReg());
```
Feiyang1, line 112: We don't use `assert` style in tests. Can you please convert `assert`s to `expect`s to be consistent with our convention?

packages/messaging/src/controllers/window-controller.test.ts
```diff
@@ -36,12 +33,19 @@ import { WindowController } from './window-controller';
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
Feiyang1, line 35: License should be at the top of the file.

packages/messaging/src/controllers/window-controller.test.ts
```diff
@@ -104,6 +108,134 @@ describe('WindowController', () => {
...
+      });
+
+      expect(getTokenStub).to.have.been.calledOnceWith(
+        firebaseDependencies,
+        otherSwReg,
+        DEFAULT_VAPID_KEY
+      );
+    });
+
+    it('uses default VAPID if: a) no VAPID was stored and b) non iss provided in option', async () => {
```
Feiyang1, line 171: s/iss/is

packages/firebase/index.d.ts
zwu52: got you. I am updating all the other parameters comments

packages/firebase/index.d.ts
zwu52: It is referring the same subject. I think keeping it consistence across the board make sense. Changing accordingly.

packages/firebase/index.d.ts
zwu52: nice catch thanks!

packages/firebase/index.d.ts
zwu52: yay. would be nice to link a @link here

packages/messaging/src/controllers/window-controller.test.ts
zwu52: hmm. I think I have a buggy format extension that moved it down. nice catch

packages/firebase/index.d.ts
```diff
@@ -6941,106 +6941,236 @@ declare namespace firebase.messaging {
...
      */
     onMessage(
       nextOrObserver: firebase.NextFn<any> | firebase.Observer<any>,
       error?: firebase.ErrorFn,
       completed?: firebase.CompleteFn
     ): firebase.Unsubscribe;
+
+    /**
+     * Called when a message is received while the app is in the background. An app is to be in the
+     * background if no active window is displayed.
```
egilmorez, line 7017: Missing one word here "considered"

"An app is considered to be in the
     * background"

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -139,34 +159,53 @@ export class SwController implements FirebaseMessaging, FirebaseService {
...
+   * If there is no notification data in the payload then no notification will be shown.
    */
   async onPush(event: PushEvent): Promise<void> {
-    const payload = getMessagePayload(event);
-    if (!payload) {
+    const internalPayload = getMessagePayloadInternal(event);
+    if (!internalPayload) {
+      console.debug(
+        TAG +
+          'failed to get parse MessagePayload from the PushEvent. Skip handling the push.'
```
Feiyang1, line 176: parsed?

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -139,34 +159,53 @@ export class SwController implements FirebaseMessaging, FirebaseService {
...
+    // background handling: display and pass to onBackgroundMessage hook
+    let isNotificationShown = false;
+    if (!!internalPayload.notification) {
+      await showNotification(wrapInternalPayload(internalPayload));
+      isNotificationShown = true;
+    }
+
+    if (
+      isNotificationShown === true &&
+      this.isOnBackgroundMessageUsed === false
```
Feiyang1, line 196: Can you put the comment for `isOnBackgroundMessageUsed` here as well? It will make it easier to understand without jumping around in the code.

packages/messaging/src/controllers/window-controller.ts
```diff
@@ -53,30 +56,131 @@ export class WindowController implements FirebaseMessaging, FirebaseService {
...
+      this.vapidKey = vapidKey;
+    }
+
+    if (!this.vapidKey && !!vapidKey) {
+      this.vapidKey = vapidKey;
+    }
+
+    if (!this.vapidKey && !vapidKey) {
+      this.vapidKey = DEFAULT_VAPID_KEY;
+    }
```
Feiyang1, line 133: Is this equivalent to
```
if (!!vapidKey) {
  this.vapidKey = vapidKey;
} else if (!this.vapidKey) {
  this.vapidKey = DEFAULT_VAPID_KEY;
}
```

packages/messaging/src/controllers/window-controller.ts
Feiyang1: It seems we can make `swRegistration` an optional parameter for `deleteToken()` (the imported function), so it will only try to unsubscribe if the registration is available. This way we don't have to register the default SW here.

packages/installations/package.json
```diff
@@ -44,6 +44,8 @@
     "@firebase/app-types": "0.x"
   },
   "dependencies": {
+    "@firebase/app": "^0.6.7",
```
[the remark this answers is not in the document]
zwu52: thanks for catching. this cripped in while I was doing some other debugging work on FIS

packages/messaging/src/controllers/window-controller.ts
zwu52: I think that's correct

PR #3236, opened 2020-06-18
Remove gRPC and protos from @firebase/testing

samtstern:
Removes the protos and `@grpc/grpc-js` from `@firebase/testing` and replaces it with simple REST calls.

Also fun fact: this module never actually listed `grpc-js` as a dependency so we were just lucky.


packages/testing/test/database.test.ts
```diff
@@ -131,14 +131,14 @@ describe('Testing Module Tests', function() {
         }
       }`
     });
-    await expect(promise).to.be.rejectedWith(/UNAVAILABLE/);
+    await expect(promise).to.be.rejectedWith(/ECONNREFUSED/);
```
yuchenshi, line 134: Could you please fix this test case while we're at it? The test case name says `succeeds on valid input` but actually just asserts connection to be refused. I think we should introduce http mocks here (like using `nock` like other JS SDK packages) and assert it sends out the right request instead.

packages/testing/test/database.test.ts
```diff
@@ -131,14 +131,14 @@ describe('Testing Module Tests', function() {
...
   });
 
   it('clearFirestoreData() succeeds on valid input', async function() {
     let promise = firebase.clearFirestoreData({
       projectId: 'foo'
     });
-    await expect(promise).to.be.rejectedWith(/UNAVAILABLE/);
+    await expect(promise).to.be.rejectedWith(/ECONNREFUSED/);
```
yuchenshi, line 141: ditto

packages/testing/test/database.test.ts
yuchenshi: Or maybe use emulators for this test?

samtstern: Good idea, moved tests to the emulators.

samtstern:
```
@firebase/messaging: npm ERR! code ELIFECYCLE
@firebase/messaging: npm ERR! errno 1
@firebase/messaging: npm ERR! firebase-messaging-selenium-test@0.2.1 test: `mocha --exit`
@firebase/messaging: npm ERR! Exit status 1
@firebase/messaging: npm ERR! 
@firebase/messaging: npm ERR! Failed at the firebase-messaging-selenium-test@0.2.1 test script.
@firebase/messaging: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
@firebase/messaging: npm ERR! A complete log of this run can be found in:
@firebase/messaging: npm ERR!     /home/runner/.npm/_logs/2020-06-19T17_43_30_344Z-debug.log
``` 

These keep getting in my way.

yuchenshi:
Please keep trying -- I doubt it's related. It might be just flaky tests. If this keep happening let's pull in some folks who might be familiar with it.

PR #3242, opened 2020-06-19
Big refactor of reauth / link + basic idp tasks

samhorlbeck:


packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+    auth,
+    credential,
+    OperationType.SIGN_IN,
+    response
+  );
+
+  return userCredential;
+}
+
+export async function _reauth(params: IdpTaskParams): Promise<UserCredential> {
```
avolkovi, line 75: nit: _reauthenticateWithIdp

packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+    requestUri,
+    sessionId,
+    postBody: postBody || null,
+    tenantId,
+    pendingToken,
+    returnSecureToken: true
+  };
+}
+
+export async function _signIn(params: IdpTaskParams): Promise<UserCredential> {
```
avolkovi, line 58: nit: _signInWithIdp

packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export async function _reauth(params: IdpTaskParams): Promise<UserCredential> {
+  const { auth, user } = params;
+  assert(user, auth.name);
+  const requestPromise = signInWithIdp(auth, paramsToRequest(params));
+  return _reauthenticate(user, requestPromise);
+}
+
+export async function _link(params: IdpTaskParams): Promise<UserCredential> {
```
avolkovi, line 82: nit: _linkWithIdp

packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+  sessionId?: string;
+  tenantId?: string;
+  postBody?: string;
+  pendingToken?: string;
+  user?: User;
+}
+
+export type IdpTask = (params: IdpTaskParams) => Promise<UserCredential>;
+
+function paramsToRequest({
```
avolkovi, line 41: nit: imo this would be a bit simpler if you added `auth` and `passIdToken: boolean` into this and just had one place where you called the API

packages-exp/auth-exp/src/core/user/user_credential_impl.ts
```diff
@@ -46,4 +48,10 @@ export class UserCredentialImpl implements UserCredential {
     // updateAdditionalUserInfoFromIdTokenResponse(userCred, idTokenResponse);
     return userCred;
   }
+
+  static async _forOperation(user: User, operationType: externs.OperationType, response: PhoneOrOauthTokenResponse): Promise<UserCredentialImpl> {
```
avolkovi, line 52: should we also use this for SIGN_IN?

packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export async function _signIn(params: IdpTaskParams): Promise<UserCredential> {
+  const request = paramsToRequest(params);
+  const auth = params.auth;
+
+  const response = await signInWithIdp(auth, request);
+
+  const credential = _authCredentialFromTokenResponse(response);
+  const userCredential = await UserCredentialImpl._fromIdTokenResponse(
```
avolkovi, line 65: why not use _forOperation here? then you can fold the logic for _authCredentialFromTokenResponse and only have it used in one place

packages-exp/auth-exp/src/core/user/link_unlink.ts
```diff
@@ -45,3 +52,25 @@ export async function unlink(
...
 }
+
+/**
+ * Internal-only link helper
+ */
+export async function _link(user: User, linkAction: Promise<IdTokenResponse>): Promise<UserCredentialImpl> {
+  return UserCredentialImpl._forOperation(user, externs.OperationType.LINK, await linkAction);
+}
+
+export async function _assertLinkedStatus(
```
avolkovi, line 63: does this need to be exported? can we have it as a private method only available from _link()?

packages-exp/auth-exp/src/core/strategies/idp.ts
```diff
@@ -0,0 +1,90 @@
...
+import { signInWithIdp, SignInWithIdpRequest } from '../../api/authentication/idp';
+import { Auth } from '../../model/auth';
+import { User, UserCredential } from '../../model/user';
+import { _authCredentialFromTokenResponse } from '../credentials/inferred';
+import { _link as _linkUser } from '../user/link_unlink';
+import { _reauthenticate } from '../user/reauthenticate';
+import { UserCredentialImpl } from '../user/user_credential_impl';
+import { assert } from '../util/assert';
+
+export interface IdpTaskParams {
```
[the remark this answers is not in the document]
samhorlbeck: No, this will only be used internally

packages-exp/auth-exp/src/core/strategies/idp.ts
samhorlbeck: I'm breaking all of this up into smaller bits, but the way these functions are used is grouped:
```ts
import * as idpTasks from '../strategies/idp';
...
{
  switch(something) {
    case A:
      execIdpTask(idpTasks._signIn);
    case B:
      execIdpTask(idpTasks._reauth);
  }
}

function execIdpTask(task: idpTasks.IdpTask, context ...) {
  return task(some other stuff here)
}
etc
```

Do you still want these renamed?

packages-exp/auth-exp/src/core/user/user_credential_impl.ts
samhorlbeck: Let me clean this up in a separate PR. The `credential` param in `_fromIdTokenResponse` shouldn't be set as the `credential` field in the `UserCredential` object. Instead, it is just used to determine "isAnonymous". I will consolidate all of this into one method later

packages-exp/auth-exp/src/core/strategies/idp.ts
samhorlbeck: One thing too is that _forOperation doesn't init a user object, which we do need for sign in.

packages-exp/auth-exp/src/core/user/link_unlink.ts
samhorlbeck: Yeah, it needs to be exported. It's used in `linkWithPhoneNumber()` to check the link status before doing the SMS flow (whereas `_link()` is called _after_ the SMS flow)

packages-exp/auth-exp/src/core/strategies/credential.test.ts
```diff
@@ -18,11 +18,7 @@
 import { expect, use } from 'chai';
 import * as chaiAsPromised from 'chai-as-promised';
 
-import {
-  OperationType,
-  ProviderId,
-  SignInMethod
-} from '@firebase/auth-types-exp';
+import { OperationType, ProviderId, SignInMethod } from '@firebase/auth-types-exp';
```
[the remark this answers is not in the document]
samhorlbeck: Done

packages-exp/auth-exp/src/core/credentials/inferred.ts
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { PhoneOrOauthTokenResponse } from '../../api/authentication/mfa';
+import { SignInWithPhoneNumberResponse } from '../../api/authentication/sms';
+import { AuthCredential } from './';
+import { PhoneAuthCredential } from './phone';
+
+export function _authCredentialFromTokenResponse(
```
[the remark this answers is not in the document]
samhorlbeck: Can't live in index-- the index exports an abstract class. If you switch it from `index.d.ts` to `index.ts`, that blows up (even with `export declare abstract class ...`). I've changed the name to `from_token_response.ts`

packages-exp/auth-exp/src/core/strategies/idp.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/strategies/idp.ts
avolkovi: seems fine, we already reuse names between api/implementation

samhorlbeck:
Didn't mean to close it -_-

PR #3245, opened 2020-06-21
Add getJSON method

jaxx2104:
## Context: 

- There is a method called [jsonValue](https://firebase.google.com/docs/reference/swift/firebaseremoteconfig/api/reference/Classes/RemoteConfigValue?/c:objc(cs)FIRRemoteConfigValue(py)JSONValue) in the iOS and Android SDKs. ([FIRConfigValue](https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Sources/FIRConfigValue.m#L63-L75))
- The method is not available on the Web, but the Firebase console has a JSON Editor, so it would be nice to have a method to get JSON.

nikuman_Firebase_コンソール

## This PR did the following:

- [x] Add getJSON method ✍️ 
- [x] Add unit test ✅

### DIFF
--- a/packages/remote-config-types/index.d.ts
+++ b/packages/remote-config-types/index.d.ts
@@ -75,6 +75,13 @@ export interface RemoteConfig {
   getBoolean(key: string): boolean;
 
   /**
+   * Gets the value for the given key as a json.
+   *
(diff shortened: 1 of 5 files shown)


jaxx2104:
@googlebot I signed it!

PR #3249, opened 2020-06-22
Update docs for UserCredential

samtstern:
When you do an operation like `signInWithPopup` there's no pointer in the docs that tells you that you can cast `UserCredential#credential` as a subclass of `AuthCredential` so you can get the `accessToken`

### DIFF
--- a/packages/firebase/externs/firebase-auth-externs.js
+++ b/packages/firebase/externs/firebase-auth-externs.js
@@ -2036,6 +2036,20 @@ firebase.auth.Auth.prototype.updateCurrentUser = function(user) {};
  * operationType could be 'signIn' for a sign-in operation, 'link' for a linking
  * operation and 'reauthenticate' for a reauthentication operation.
  *
+ * Depending on the original operation, the <code>credential</code> field may contain
+ * one of the following subclasses of <code>AuthCredential</code>:
+ *
+ * <ul>
+ *   <li>
+ *    {@link firebase.auth.PhoneAuthCredential} for phone auth operations such as
+ *    <code>signInWithPhoneNumber</code>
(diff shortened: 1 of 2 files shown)


packages/firebase/externs/firebase-auth-externs.js
```diff
@@ -2036,6 +2036,20 @@ firebase.auth.Auth.prototype.updateCurrentUser = function(user) {};
...
+ *
+ * <ul>
+ *   <li>
+ *    {@link firebase.auth.PhoneAuthCredential} for phone auth operations such as
+ *    <code>signInWithPhoneNumber</code>
+ *   </li>
+ *   <li>
+ *    {@link firebase.auth.OAuthCredential} for OAuth operations such as
+ *    <code>signInWithPopup</code>
+ *   </li>
```
samhorlbeck, line 2050: Should we note that this field will be null if the sign in is any other type (like email, for example).

Or is that noted somewhere else already?

samtstern: It's already optional but not a bad note!  I'll add it.

hsubox76:
Let me know if I'm misunderstanding the purpose of this PR but our documentation (https://firebase.google.com/docs) is currently generated from this file: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts

(After modularization it will be generated from the source code.)

samtstern:
@hsubox76 thanks!  I updated both files now.

samtstern:
@samhorlbeck should I continue with this PR or will your generic PR make it obsolete?

samtstern:
I think this is obsolete ...

PR #3250, opened 2020-06-22
Tree-shakeable Bound

schmidt-sebastian:
This is a tiny follow up to  https://github.com/firebase/firebase-js-sdk/pull/3248


packages/firestore/src/core/query.ts
```diff
@@ -732,73 +732,80 @@ export const enum Direction {
...
+  }
+
+  if (
+    left.before !== right.before ||
+    left.position.length !== right.position.length
+  ) {
+    return false;
+  }
+  for (let i = 0; i < left.position.length; i++) {
+    const thisPosition = left.position[i];
```
dconeybe, line 802: nit: Consider renaming `thisPosition` to `leftPosition` and `otherPosition` to `rightPosition`.

schmidt-sebastian: Done

PR #3251, opened 2020-06-22
Add ReactNative build back, fix base64 serialization

schmidt-sebastian:
Manually tested on iOS:
- Mutation can be written
- Document listeners work
- A Firestore Blob that is written in RN can be read from the native iOS SDK and contains the same value

Fixes: https://github.com/firebase/firebase-js-sdk/issues/2667


schmidt-sebastian:
We should fix the static ESM build before proceeding with this change.

schmidt-sebastian:
@Feiyang1 Do you have time to look at this?

PR #3252, opened 2020-06-22
Tree-shakeable FieldFilter

schmidt-sebastian:
This PR makes the Proto-consuming code in FieldFilter tree-shakeable. Once this is merged, we can make Query tree-shakeable.


wilhuff:
Could you explain how this results in any net savings for users?

AFAICT, you've changed this from a set of classes with polymorphic behavior to a big if-then ladder, but so long as you run queries at all, you still get all the code. There's been a very slight savings in package size at a massive cost in readability and maintainability.

FWIW, the other kind of filter we had mind were conjunctions and disjunctions.

schmidt-sebastian:
Sorry, I missed your comment. The idea was to make both `FieldFilter.equals()` and `FieldFilter.canonicalId()` optional, which leads to FieldFilter classes that only have a single method ("matches()"). This, combined with the existing if-then code in `FieldFilter.create()` led me down this path. I'll explore adding the classes back, but they will only have this one method.

packages/firestore/src/core/target.ts
```diff
@@ -22,11 +22,14 @@ import {
...
   canonifyBound,
-  canonifyOrderBy,
-  Filter,
+  canonifyFilter,
+  filterEquals,
+  stringifyFilter,
   OrderBy,
   orderByEquals,
-  stringifyOrderBy
+  stringifyOrderBy,
```
wilhuff, line 30: I think we've discussed this before, so I won't push, but I find the difference in ordering (subject-operation vs operation-subject) between `orderByEquals` and `stringifyOrderBy` to be jarring. Autocompletion and muscle memory would be much easier if there was just one way to do it.

If these were all in the form subject-operation they would group nicely in imports as well.

What do you think of these?
```
orderByEquals
orderByToString
orderByCanonicalId
```

schmidt-sebastian: The naming in this PR is inspired by the names in the Values helper. I am also not a big fan how "equals" is an outlier, but _on its own_ I do think that each name  looks ok. For that reason, I am kind of reluctant to change it right now, but I am not opposed to overhauling all stringify/canonicalize methods in a follow up.

PR #3256, opened 2020-06-23
Export some unchanged members for the full Firestore "exp" SDK

schmidt-sebastian:
This adds some exports to the full tree-shakeable Firestore SDK for some of the existing APIs that can be exported without modification.


rafikhan:
LGTM

PR #3259, opened 2020-06-23
Expose static members of public types in ESM2017 build

schmidt-sebastian:
`makeConstructorPrivate` currently does not re-assign static members in the ESM2017 build, which essentially breaks the static methods on Blob, FieldValue and FieldPath.

### DIFF
--- a/packages/firestore/src/util/api.ts
+++ b/packages/firestore/src/util/api.ts
@@ -17,6 +17,9 @@
 
 import { Code, FirestoreError } from './error';
 
+/** List of JavaScript built-in members that cannot be reassigned. */
+const RESERVED_READONLY_PROPS = ['length', 'name'];
+
 /**
  * Helper function to prevent instantiation through the constructor.
  *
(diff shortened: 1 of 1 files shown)


packages/firestore/src/util/api.ts
```diff
@@ -41,11 +44,15 @@ export function makeConstructorPrivate<T extends Function>(
...
+  for (const staticProp of Object.getOwnPropertyNames(cls)) {
+    if (RESERVED_READONLY_PROPS.indexOf(staticProp) === -1) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (PublicConstructor as any)[staticProp] = (cls as any)[staticProp];
+    }
+  }
 
-  // Copy any static methods/members
+  // Copy instance methods/members
   Object.assign(PublicConstructor, cls);
```
Feiyang1, line 56: We don't need it anymore since instance methods/members are copied in `prototype`.

Issue #3222, opened 2020-06-16
auth.onAuthStateChanged + firestore.onSnapshot cause infinite loop of calls to /token.


notVitaliy:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Browser version: Chrome && Firefox
  * Firebase SDK version: >=7.10.0
  * Firebase Product: Auth

### [REQUIRED] Describe the problem

Calls to `onAuthStateChanged` and `onIdTokenChanged` send off a flurry of requests to `securetoken.googleapis.com/v1/token` until it responds with `QUOTA_EXCEEDED`. This only happens when the user is authenticated.

#### Steps to reproduce:

This error does not occur on Linux or macOS. I was not able to replicate it in a windows VM on Linux. It's only on windows in both firefox and chrome. However, I've only had success replicating with 2 windows machines and failed to replicate on another windows laptop.

0. Only windows (maybe?)
1. go to the stackblitz link
2. authenticate with email (random email + password should work)
3. uncomment line 42
4. refresh the app stackblitz preview window
5. check network tab, there should be 100's of requests

https://stackblitz.com/edit/firebase-issue-sandbox-vzdnt7

This is the strangest bug that I've run across and it took quite a bit to isolate the exact sequence of events that was causing it.

My initial thoughts are that calls to the firestore document snapshot are causing the `onAuthStateChanged` observer to trigger.

edit: I just did some more digging and this bug does not happen in 7.9.3 but does in 7.10.0

samhorlbeck:
Interesting.. I'm not sure but I suspect this is more related to the Firestore SDK than Auth directly. I'm rerouting this to Firestore.

var-const:
Thanks for reporting the issue. I will try to reproduce it and will post an update when I have one.

lymeo:
same here

var-const:
I'm really sorry, but I had no luck trying to reproduce this so far. I'm using a Windows 10 laptop and Chrome. The only significant change I made to the repro case is to change the project configuration. Can you please clarify a few things? (I have added numbers to the reproduction steps in your original message to make it easier to refer to them, I hope you don't mind)
* regarding step 4 -- after reloading the preview window, it's not necessary to sign in again to see the error, right?
* regarding step 2 -- do you create a new user or sign in as an existing user? (I tried both, though)
* regarding step 3 -- what is the significance of this step? Does the issue not reproduce if line 42 isn't commented out during the first run?
[...]

PR #3260, opened 2020-06-23
Do not use recursive getToken() loop

schmidt-sebastian:
There has been another user affected by https://github.com/firebase/firebase-js-sdk/issues/3222, so I am sending this out for review now. This reduced the change of a possible `getToken` loop if our token is invalid by just using the last received token.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3222


schmidt-sebastian:
FWIW, the newest user report is from a user that doesn't use Firestore.

packages/firestore/src/api/credentials.ts
```diff
@@ -199,38 +197,35 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
     }
 
+    this.lastToken = null;
     return this.auth.getToken(forceRefresh).then(tokenData => {
-      // Cancel the request since the token changed while the request was
-      // outstanding so the response is potentially for a previous user (which
-      // user, we can't be sure).
-      if (this.tokenCounter !== initialTokenCounter) {
+      if (this.lastToken) {
+        // If we received a new token while this callback was pending (via
```
var-const, line 210: Question: what is the reason to prefer `lastToken` to `tokenData` here?

schmidt-sebastian: I would like you to verify this - but I think this matches the previous behavior. If we receive a token update via our callback listener, we ignore the token that is returned from the `getToken` call.

FWIW, while I think this is the cleaner solution to our token handling, I think the risk of it probably means we should not merge it. After all, we now have evidence that the problem we are trying to solve also exists in at least one other place (and maybe only that place).

Maybe @wilhuff wants a distraction from his current work and take a peak as well.

wilhuff: The reason this is complicated is that we need to subscribe to auth to get notified when the identity changes, so that we can trigger user changes in the local store.

The reason we ignored the token in that callback was because when remote store wants to restart streams it it's going to trigger a call to getToken. This restart might happen significantly later, because there may not have been active streams when the identity changed.

The central question we need to answer is whether or not the token listener will call us before the current access token expires. If it does, it seems like we could just always use the token supplied by the token listener, no? If we did that this could be much simpler.

schmidt-sebastian:
Closing this for now.

PR #3266, opened 2020-06-23
Update the UserCredential type to be generic

samhorlbeck:
This PR also fixes up the behavior for UserCredential; it should only have a `credential` field if the sign in method is Phone or OAuth. Otherwise, this field is always `null`.


packages-exp/auth-types-exp/index.d.ts
```diff
@@ -386,9 +388,9 @@ export interface User extends UserInfo {
 /**
  * https://firebase.google.com/docs/reference/js/firebase.auth#usercredential
  */
-export interface UserCredential {
+export interface UserCredential<T extends AuthCredential | null> {
```
samhorlbeck, line 391: @samtstern This whole PR for this one little change.

This will (eventually) address the issue you raised the other day

packages-exp/auth-exp/src/core/strategies/anonymous.test.ts
```diff
@@ -53,8 +51,7 @@ describe('core/strategies/anonymous', () => {
   describe('signInAnonymously', () => {
     it('should sign in an anonymous user', async () => {
       const { credential, user, operationType } = await signInAnonymously(auth);
-      expect(credential?.providerId).to.eq(ProviderId.ANONYMOUS);
```
avolkovi, line 56: there's no credential for anonymous?

packages-exp/auth-exp/src/core/strategies/credential.test.ts
```diff
@@ -86,8 +92,7 @@ describe('core/strategies/credential', () => {
         auth,
         authCredential
       );
-      expect(credential!.providerId).to.eq(ProviderId.FIREBASE);
```
avolkovi, line 89: you should have the test ensure that the credential is preserved, maybe do UserCredential<MockAuthCredential> or whatevenr

packages-exp/auth-exp/src/core/strategies/credential.ts
```diff
@@ -27,15 +27,17 @@ import { _reauthenticate } from '../user/reauthenticate';
 import { _reloadWithoutSaving } from '../user/reload';
 import { UserCredentialImpl } from '../user/user_credential_impl';
 
-export async function signInWithCredential(
+export async function signInWithCredential<
```
avolkovi, line 30: this is a public API change... right?

packages-exp/auth-exp/src/core/strategies/email_and_password.test.ts
```diff
@@ -430,8 +426,7 @@ describe('core/strategies/email_and_password/signInWithEmailAndPassword', () =>
       user,
       operationType
     } = await signInWithEmailAndPassword(auth, 'some-email', 'some-password');
-    expect(credential!.providerId).to.eq(ProviderId.PASSWORD);
```
avolkovi, line 433: really? these don't get set?

packages-exp/auth-types-exp/index.d.ts
avolkovi: should we run this change past a larger audience? seem like a pretty big signature change to me...

Feiyang1:
@samhorlbeck Is this PR still being worked on?

PR #3268, opened 2020-06-24
Enable changesets

Feiyang1:
Once merged, we will need to install the [changeset Github App](https://github.com/apps/changeset-bot).


scripts/release/utils/workspace.ts
```diff
@@ -39,21 +45,20 @@ function mapWorkspaceToPackages(workspaces) {
...
 }
-exports.mapWorkspaceToPackages = mapWorkspaceToPackages;
 
-function mapPackagestoPkgJson(packagePaths) {
+function mapPackagestoPkgJson(packagePaths: string[]) {
   return packagePaths
     .map(path => {
       try {
-        return require(`${path}/package.json`);
+        return JSON.parse(readFileSync(`${path}/package.json`, 'utf8'));
```
hsubox76, line 53: Just out of curiosity is it because ts-node can't require json files?

Feiyang1: `require` caches the content of the package.json, so you will read stale `package.json` if you read it first, update it, then read it again. using `readFile` will always get the latest version from disk.

Issue #3218, opened 2020-06-16
Promise never resolves when performing an action on a Firestore DB after calling auth.createUserWithEmailAndPassword


nefarioustim:
* Operating System version: MacOS Mojave 10.14.6
  * Browser version: Firefox 78, Chrome 84, Safari 13.1.1
  * Firebase SDK version: 7.15
  * Firebase Product: auth, firestore

When attempting to follow a call to `auth.createUserWithEmailAndPassword` with code that updates or creates a Firestore document (in my initial case it was a `firestore.collection().doc().set()`, but in testing it was the same for an `.update()`), the returning Promise never resolves and simply hangs.

Here is a simplified example of the Vuex actions I am calling in my code that exhibit this issue:

```javascript
    signUserUp ({ dispatch, commit }, payload) {
      const $app = this.$app
      commit('setLoading', true)
      commit('clearError')

      return fb.auth
        .createUserWithEmailAndPassword(payload.email, payload.password)
        .then(userRef => {
          $app.logInfo('Created auth user', {'email': payload.email})

          return dispatch("createAccount", payload)
        })
        .catch(error => {
          commit('setLoading', false)
          commit('setError', error)
          $app.logError('Error signing up user', {error: error.message})
        })
    },

    createAccount({ commit, dispatch }, payload) {
      const $app = this.$app
      return fb.db.collection(COLLECTION)
        .doc(payload.account)
        .set({
          'created': new Date().getTime(),
          'updated': new Date().getTime(),
          'active': true,
          'account': payload.account,
          'legalName': payload.legalName,
          'country': payload.country,
          'currency': payload.currency,
          'industry': payload.industry,
          'email': payload.busemail || payload.email, // Allow for signup form
          'about': payload.about,
          'tags': payload.tags,
          'type': payload.type
        })
        .catch(error => {
          commit('setLoading', false)
          commit('setError', error)
          $app.logError('Error creating account', {error: error.message})
        })
    }
```

I have managed to temporarily resolve this issue by rolling back to version 7.14 of the SDK, but obviously I'm interested in a more longterm fix that allows me to continue upgrading the SDK.

Possible related to #3120 and raised following comments there.

schmidt-sebastian:
@nefarioustim Thanks for reporting. 

This is likely fixed in v17.15.1, which includes https://github.com/firebase/firebase-js-sdk/pull/3184. Do you mind verifying your issue with this version?

gragland:
[...]

PR #3270, opened 2020-06-24
Don't delay retryable operations

schmidt-sebastian:
One of the behavior changes of `enqueueRetryable` (https://github.com/firebase/firebase-js-sdk/pull/2879) was that it only scheduled the next operation when the first operation succeeds. This means that any operation that was scheduled after the second retryable operation will run before this run:

```
enqueue(console.log(1))
enqueueRetryable(console.log(2))
enqueueRetryable(console.log(3))
enqueue(console.log(4))

-> prints 1,2,4,3 since 3 is only enqueued once 2 finishes
```

While theoretically okay, this still seems to cause some issues with the user change. This PR changes the retry behavior so that operations run in the original order if no failure occurred. 

Addresses https://github.com/firebase/firebase-js-sdk/issues/3218

### DIFF
--- a/packages/firestore/src/util/async_queue.ts
+++ b/packages/firestore/src/util/async_queue.ts
@@ -207,7 +207,7 @@ export class AsyncQueue {
   // The last promise in the queue.
   private tail: Promise<unknown> = Promise.resolve();
 
-  // A list of retryable operations Retryable operation are run in order and
+  // A list of retryable operations. Retryable operations are run in order and
   // retried with backoff.
   private retryableOps: Array<() => Promise<void>> = [];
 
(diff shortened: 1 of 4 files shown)


packages/firestore/test/unit/util/async_queue.test.ts
```diff
@@ -313,6 +312,65 @@ describe('AsyncQueue', () => {
     expect(completedSteps).to.deep.equal([1, 1, 2]);
   });
 
+  it('Doesn not delay retryable operations that succeed', async () => {
```
var-const, line 315: Nit: s/Doesn/Does/.

packages/firestore/src/util/async_queue.ts
```diff
@@ -323,32 +323,32 @@ export class AsyncQueue {
...
-            throw e; // Failure will be handled by AsyncQueue
-          }
+  /**
+   * Runs the next operation from the retryable queue. If the operation fails,
+   * reschedules with backoff.
+   */
+  private async retryNextOp(): Promise<void> {
+    const op = this.retryableOps.shift();
+
+    if (op) {
```
var-const, line 337: Can you use an early return instead to reduce nestedness?

packages/firestore/src/util/async_queue.ts
```diff
@@ -323,32 +323,32 @@ export class AsyncQueue {
...
+        } else {
+          throw e; // Failure will be handled by AsyncQueue
         }
-      };
-      this.enqueueAndForget(retryingOp);
-      return deferred.promise;
-    });
+      }
+
+      this.backoff.backoffAndRun(() => this.retryNextOp());
```
var-const, line 350: Question: why is backoff invoked in the successful case?

packages/firestore/src/util/async_queue.ts
```diff
@@ -207,9 +207,9 @@ export class AsyncQueue {
   // The last promise in the queue.
   private tail: Promise<unknown> = Promise.resolve();
 
-  // The last retryable operation. Retryable operation are run in order and
+  // A list of retryable operations Retryable operation are run in order and
```
var-const, line 210: Nit: missing full stop.

var-const, line 210: Nit: should be `Retryable operations are run in order...` (i.e. plural).

packages/firestore/test/unit/specs/recovery_spec.test.ts
```diff
@@ -787,6 +787,7 @@ describeSpec('Persistence Recovery', ['no-ios', 'no-android'], () => {
           .expectActiveTargets({ query })
           // We are now user 2
           .expectEvents(query, { removed: [doc1], fromCache: true })
+          .runTimer(TimerId.AsyncQueueRetry)
```
var-const, line 790: Question: why is this change necessary?

PR #3274, opened 2020-06-24
Add getDoc() to firestore-exp

schmidt-sebastian:
This adds getDoc() to the firestore-exp SDK. It re-uses most code from the Lite SDK where possible (e.g. DocumentReferences, CollectionReferences) but includes new full featured implementation for DocumentSnapshot and QueryDocumentSnapshot (tests for those will come when we add setDoc). 

The new `getDoc()` uses FirestoreClient to fetch a latency compensated snapshot, and uses some factored out methods from the current SDK to convert them to a ViewSnapshot. This is the reason for the changes in `src/api/database.ts`, which is refactored so that the existing methods to use generic ViewSnapshots (which can be consumed by both the current and the firestore-exp SDK).


packages/firestore/exp/index.node.ts
```diff
@@ -15,8 +15,37 @@
...
+  DocumentReference,
+  CollectionReference,
+  Query,
+  doc,
+  collection,
+  collectionGroup,
+  parent
+} from '../lite/src/api/reference';
+
+export { runTransaction, Transaction } from '../lite/src/api/transaction';
```
schmidt-sebastian, line 45: All these exports can be re-used without modification since they don't rely on latency-compensation.

packages/firestore/src/api/database.ts
```diff
@@ -1177,150 +1186,179 @@ export class DocumentReference<T = firestore.DocumentData>
...
-          // user actions affecting the now stale query.
-          unlisten();
+  /**
+   * Converts a ViewSnapshot that contains the current document to a
+   * DocumentSnapshot.
+   */
+  private _convertToDocSnapshot(snapshot: ViewSnapshot): DocumentSnapshot<T> {
+    debugAssert(
+      snapshot.docs.size <= 1,
+      'Too many documents returned on a document query'
```
rafikhan, line 1249: This message is odd relative to the check being done. Can you add a comment with some context?

packages/firestore/lite/src/api/database.ts
```diff
@@ -41,6 +41,7 @@ import { Settings } from '../../';
 // settings() defaults:
 const DEFAULT_HOST = 'firestore.googleapis.com';
 const DEFAULT_SSL = true;
+const DEFAULT_FORCE_LONG_POLLING = false; // Used by full SDK
```
rafikhan, line 44: Is this necessary given the fixes we have in the pipeline? Will you pull this out later?

packages/firestore/src/api/database.ts
schmidt-sebastian: Changed to `Expected zero or a single result on a document-only query`.

PR #3277, opened 2020-06-24
Address (some) of the tree shaking issues w/ ReCaptcha

samhorlbeck:
* Remove global exported consts
 * Remove `get` property, replaced w/ traditional method


samhorlbeck:
No, these changes are 100% internal workings of the SDK.

Now the _other_ updates to ReCaptcha that I outlined in the doc I shared with you _will_ require an API review

PR #3280, opened 2020-06-25
Modular Export Binary Size Analysis - WIP

XuechunHou:
Stage a PR to track progress of modular export binary size calculator project


repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+  return { dependencies, sizeInBytes: Buffer.byteLength(code!, 'utf-8') };
+}
+
+/**
+ * Extracts all function, class and variable declarations using the TypeScript
+ * compiler API.
+ */
+export function extractDeclarations(
+  jsFile: string,
+  map?: Map<string, string>
```
Feiyang1, line 113: `map` is only available for analyzing js bundle, right?
Can you also comment on what the map is, and under what condition it will be conditional?

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  let declarations: MemberList = {
+    functions: [],
+    classes: [],
+    variables: [],
+    enums: []
+  };
+
+  ts.forEachChild(sourceFile, node => {
+    if (ts.isFunctionDeclaration(node)) {
```
Feiyang1, line 131: Can you add comment for each if statement on what it actually checks, preferably with examples.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+      const variableDeclarations = node.declarationList.declarations;
+      variableDeclarations.forEach(variableDeclaration => {
+        if (ts.isIdentifier(variableDeclaration.name)) {
+          declarations.variables.push(
+            (variableDeclaration.name as ts.Identifier).getText(sourceFile)
+          );
+        }
+        // TODO: variableDeclaration.name is an union type (Identifier | BindingPattern)
+        // need Identifier type, not sure in what case BindingPattern type is for.
+        else {
```
Feiyang1, line 149: what is the difference between Identifier and BindingPattern?

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+  return newMemberList;
+}
+
+function isReExported(symbol: string, reExportedSymbols: string[]): boolean {
+  return reExportedSymbols.includes(symbol);
+}
+
+function extractRealSymbolName(exportSpecifier: ts.ExportSpecifier): string {
+  // if property name is not null -> export is renamed
+  if (exportSpecifier.propertyName) {
```
Feiyang1, line 271: is the propertName the renamed name, or the original name? can use a comment here.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+              exportSpecifier
+            );
+            // if export is renamed, replace with new name
+            if (isExportRenamed(exportSpecifier)) {
+              actualExports.push(exportSpecifier.name.escapedText.toString());
+              replaceAll(
+                reExportsMember,
+                reExportedSymbol,
+                exportSpecifier.name.escapedText.toString()
+              );
```
Feiyang1, line 176: This part is confusing.  I guess `exportSpecifier.name.escapedText.toString()` is the renamed name, and `reExportedSymbol ` is the original name?

Can you add some comment to explain that, and also add comment to explain what `replaceAll()` tries to do.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,316 @@
...
+        }
+        // TODO: variableDeclaration.name is an union type (Identifier | BindingPattern)
+        // need Identifier type, not sure in what case BindingPattern type is for.
+        else {
+          console.log(
+            'this VariableDeclaration.name object is of BindingPattern type !'
+          );
+        }
+      });
+    } else if (ts.isExportDeclaration(node)) {
```
Feiyang1, line 155: This block has some complex logic that will be hard to understand for people who is not familiar with ts compiler and the nuances in exports. We should add more comments explaining the code assuming the reader doesn't have these knowledge, and make it clear to them.

repo-scripts/size-analysis/analysis.ts
```diff
@@ -0,0 +1,271 @@
...
+  } else if (!argv.inputDtsFile && !argv.inputBundleFile) {
+    // retrieve All Module Names
+    let allModulesLocation = await mapWorkspaceToPackages([
+      `${projectRoot}/packages-exp/*`
+    ]);
+    if (argv.inputModule) {
+      allModulesLocation = allModulesLocation.filter(path => {
+        try {
+          const json = require(`${path}/package.json`);
+          return argv.inputModule.includes(json.name);
```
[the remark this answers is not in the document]
XuechunHou: argv.inputModule is an array type, so it's checking if json.name is an element of the array.....?

package.json
```diff
@@ -22,8 +22,8 @@
   ],
   "scripts": {
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
-    "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire build",
-    "build:exp": "lerna run --scope @firebase/*-exp --scope firebase-exp build",
+    "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire --scope size-analysis-firebase build ",
+    "build:exp": "lerna run --scope @firebase/*-exp --scope firebase-exp   build",
```
hsubox76, line 26: stray whitespace?

packages-exp/app-exp/src/index.ts
```diff
@@ -27,4 +27,4 @@ import { registerCoreComponents } from './registerCoreComponents';
 export * from './api';
 export * from './internal';
 
-registerCoreComponents();
+//egisterCoreComponents();
```
hsubox76, line 30: typo?

packages/analytics/index.ts
```diff
@@ -45,6 +45,7 @@ declare global {
  * Type constant for Firebase Analytics.
```
hsubox76, line 45: revert this file? (whitespace changes only)

packages-exp/functions-exp/package.json
```diff
@@ -64,4 +64,4 @@
     ],
     "reportDir": "./coverage/node"
   }
-}
```
hsubox76, line 67: revert

repo-scripts/size-analysis/index.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
hsubox76, line 1: I'm sure there's a reason, but just asking for my own clarification, what is the purpose of this? If I read correctly, this isn't code used for the actual analysis script, it's something that builds as part of a full repo build, and functions as a sort of test to make sure the package builds without breaking? And there's also a separate karma test file `index.test.ts` that runs with `yarn test`. Do these cover different things?

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -366,3 +387,25 @@ export function writeReportToDirectory(
...
+/**
+ *
+ * @param minimizedDependencies The dependant symbols extracted from a bundle file that doesn't resolve external firebase modules
+ *
+ * @param fullyResolvedDependencies The dependant symbols extracted from a bundle file that resolves external firebase modules
+ *
+ * This function calculates the difference between two dependency sets, the difference is the symbols of external firebase modules
+ * that the current modular export depends on.
+ */
+export function extractExternalDependencies(
```
Feiyang1, line 481: It is going to include symbols we don't care about from within external dependencies. We should only include the top level external dependencies.

For example, for `import { someFunction } from 'externalModule'`, we only want to include `someFunction`, not anything that `someFunction` depends on from within `externalModule`.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -79,7 +79,8 @@ export async function extractDependenciesAndSize(
   map: Map<string, string>
 ): Promise<ExportData> {
   const input = tmp.fileSync().name + '.js';
-  const output = tmp.fileSync().name + '.js';
+  const fulllyResolvedOutput = tmp.fileSync().name + '.js';
+  const minimizedBundleOutput = tmp.fileSync().name + '.js';
```
Feiyang1, line 86: s/fullly/fully

You are using `fullyResolvedOutput` for bundle that resolves all dependencies. Should we change `minimizedBundleOutput` to  something that contrast with it to make the difference apparent for readers, for example, `externalDepsNotResolvedOutput`?

common/api-review/dummy-exp.api.md
```diff
@@ -4,18 +4,149 @@
 
 ```ts
 
+import { LogLevel } from '@firebase/logger';
```
Feiyang1, line 7: Remove this file and remove api-report related scripts from dummy-exp package. You don't need it for dummy-exp or size-analysis packages

repo-scripts/size-analysis/analysis.ts
```diff
@@ -66,10 +70,16 @@ const argv = yargs
       alias: 'ib',
       desc: 'support for adhoc analysis. requires a path to a bundle file'
     },
+    ci: {
+      type: 'boolean',
+      alias: 'ci',
+      default: false,
+      desc:
+        "when enabled, the binary size report is not persisted on file system; Instead, it's uploaded to CI backend"
+    },
```
Feiyang1, line 79: Uploading to backend should be handled by a different script, not the analyzer.
We should make it possible to use the size-analysis tool programmatically by allowing another script to import it, execute a function and get a result object back.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -225,52 +226,117 @@ export function extractDeclarations(
...
-      if (node.moduleSpecifier && ts.isStringLiteral(node.moduleSpecifier)) {
-        const symbol = checker.getSymbolAtLocation(node.moduleSpecifier);
-        const reExportFullPath = symbol.valueDeclaration.getSourceFile()
-          .fileName;
-
-        // first step: always retrieve all exported symbols from the source location of the re-export.
-        const reExportsMember = extractDeclarations(reExportFullPath);
-
-        // named exports: eg: export {foo, bar} from '...'; and export {foo as foo1, bar} from '...';
+      if (node.moduleSpecifier) {
```
Feiyang1, line 267: This block has grown too big and too complex with nested if else. It's very hard for me to review. Can you extract it into a few smaller functions each of which handles a single use case (e.g. one for directly export, one for reexport, one for import then export,  etc). Also please add comments to help people understand.

repo-scripts/size-analysis/size-analysis.test.ts
```diff
@@ -0,0 +1,365 @@
...
+import { retrieveTestModuleDtsFile } from './test-utils';
+import { exec } from 'child_process';
+import * as fs from 'fs';
+import { resolve } from 'path';
+
+describe('extractDeclarations', () => {
+  let testModuleDtsFile: string;
+  let extractedDeclarations: MemberList;
+  before(() => {
+    // this.timeout(300000);
```
Feiyang1, line 41: Did you fix it?

package.json
```diff
@@ -22,7 +22,7 @@
   ],
   "scripts": {
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
-    "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire build",
+    "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire --scope size-analysis-firebase build",
```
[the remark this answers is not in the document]
XuechunHou: tests I wrote require the module to be built .. so..

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,779 @@
...
+          variableDeclaration.name.elements.forEach(node => {
+            declarations.variables.push(node.name.getText(sourceFile));
+          });
+        }
+      });
+    } else if (ts.isImportDeclaration(node) && node.importClause) {
+      const symbol = checker.getSymbolAtLocation(node.moduleSpecifier);
+      if (symbol && symbol.valueDeclaration) {
+        const importFilePath = symbol.valueDeclaration.getSourceFile().fileName;
+        // import { a, b } from '@firebase/dummy-exp'
```
Feiyang1, line 242: It doesn't seem we are handling renamed imports, e.g. `import { a as AA} from ...`. You may need to store both original name and renamed name in order to handle exports without imports correctly.

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,779 @@
...
+ */
+function handleExportStatementsWithoutFromClause(
+  node: ts.ExportDeclaration,
+  symbolLocation: Map<string, string>,
+  declarations: MemberList
+): void {
+  if (node.exportClause && ts.isNamedExports(node.exportClause)) {
+    node.exportClause.elements.forEach(exportSpecifier => {
+      // export symbol could be renamed, we retrieve its current/renamed name
+      const exportedSymbolCurrentName = exportSpecifier.name.escapedText.toString();
```
Feiyang1, line 420: It sounds like you are using `myA` and `myB` in the following case, so that you won't match with the imports. Shouldn't we use the original name.
```
     import {a, b} from '...'
     export {a as myA, b as myB};
```

Further, we need to handle renamed imports as I mentioned above. We should be handle the following case

```
     import {a as AA, b as BB} from '...'
     export {AA as myA, BB as myB};
```

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,779 @@
...
+  if (!fs.existsSync(directoryPath)) {
+    fs.mkdirSync(directoryPath, { recursive: true });
+  }
+  fs.writeFileSync(outputFile, report);
+}
+/**
+ *
+ * This functions writes generated json report(s) to a file of given directory
+ */
+export function writeReportToDirectory(
```
Feiyang1, line 556: Can you generate the filename and reuse `writeReportToFile()`?

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -0,0 +1,779 @@
...
+          }
+        });
+        // import * as fs from 'fs'
+      } else if (
+        node.importClause.namedBindings &&
+        ts.isNamespaceImport(node.importClause.namedBindings)
+      ) {
+        externals[moduleName].push('*');
+        // import a from '@firebase/dummy-exp'
+      } else if (
```
Feiyang1, line 616: We should use a special string for default exports too. e.g. `default export`

repo-scripts/size-analysis/size-analysis.test.ts
```diff
@@ -0,0 +1,378 @@
...
+  });
+
+  it('test re-exported function extractions from same module', () => {
+    expect(extractedDeclarations.functions).to.include.members([
+      'basicFuncExportNoDependenciesFar',
+      'basicFuncExportVarDependenciesFar',
+      'basicFuncExportFuncDependenciesFar',
+      'basicFuncExportEnumDependenciesFar',
+      'basicFuncExternalDependenciesFar'
+      // 'basicFuncExportNoDependenciesBar',
```
Feiyang1, line 104: Remove comments

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -240,30 +257,47 @@ export function extractDeclarations(
...
+          // import * as fs from 'fs
         } else if (
           node.importClause.namedBindings &&
           ts.isNamespaceImport(node.importClause.namedBindings)
         ) {
           const symbolName: string = node.importClause.namedBindings.name.getText(
             sourceFile
           );
-          symbolLocation.set(symbolName, importFilePath);
+          declarations.variables.push(symbolName);
```
Feiyang1, line 289: IIUC, this assumes `fs` is part of the public API, but it might not be. Consider the following case:
```
import * as fs from 'fs';

export declare const aVariable: fs.someType;
```

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -348,8 +387,12 @@ function handleExportStatementsWithFromClause(
     node.exportClause.elements.forEach(exportSpecifier => {
       declarations.variables.push(exportSpecifier.name.escapedText.toString());
     });
-  } else {
-    console.log(`${moduleName}: ${Warning.MODULE_NOT_RESOLVED}`);
+  }
+  // handles the case when exporting * from a module whose location can't be resolved
+  else {
+    console.log(
+      `export statement: export * from ${moduleName}: ${Warning.MODULE_NOT_RESOLVED}`
```
Feiyang1, line 394: Can you add some context in the message? something like, `the public API of moduleA is not complete because it reexports from 'moduleName' using * export, but we couldn't resolve 'moduleName' `

repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -437,21 +530,36 @@ function handleExportStatementsWithoutFromClause(
...
     });
   }
+  return declarations;
+}
+/**
+ * This functions returns a copy of source MemberList. It deep copies the functions, classes, variables field, and
+ * shallow copies externals field.
+ * @param source source MeberList to make a copy of
+ */
+function copyOf(source: MemberList): MemberList {
```
Feiyang1, line 550: You can use `import {deepCopy} from '@firebase/utils'`

repo-scripts/size-analysis/size-analysis.test.ts
```diff
@@ -0,0 +1,401 @@
...
+    const aFile = `a-file`;
+    const aPathToFile = `${aDir}/${aFile}`;
+    fs.writeFileSync(aPathToFile, 'content');
+    expect(() =>
+      writeReportToDirectory('content', aFile, aPathToFile)
+    ).to.throw(ErrorCode.OUTPUT_DIRECTORY_REQUIRED);
+  });
+
+  it('should not throw error when given path does not pre-exist', () => {
+    const aDir = resolve('./a-dir/a-sub-dir');
```
Feiyang1, line 375: This relies on the assumption that the tests run in order where the previous test creates the dir, which may not be true. This test should create the folder in its own body. The previous test should clean up once it's finished.

repo-scripts/size-analysis/size-analysis.test.ts
```diff
@@ -0,0 +1,401 @@
...
+import { retrieveTestModuleDtsFile } from './test-utils';
+import * as fs from 'fs';
+import { resolve } from 'path';
+
+describe('extractDeclarations', () => {
+  let testModuleDtsFile: string;
+  let extractedDeclarations: MemberList;
+  before(() => {
+    testModuleDtsFile = retrieveTestModuleDtsFile();
+    extractedDeclarations = extractDeclarations(testModuleDtsFile);
```
Feiyang1, line 41: Can you add tests for extractDeclarations where the input is a js bundle?

PR #3288, opened 2020-06-25
Tree shakeable persistence

samhorlbeck:
Exporting `const`s that are instantiated using a function call (like a constructor) can't be tree shaken since the tools don't know whether or not the function side-effects. To get around this while keeping the appearance of exported `const`s that the user can use, the classes themselves take on the shape of the `externs.Persistence` interface, and provide internal static methods for instantiation.

I've verified in the demo build that this will indeed allow all of these persistence classes to be tree-shaken. This also does not change the API surface at all as far as the user is concerned.


packages-exp/auth-exp/src/core/persistence/browser.test.ts
```diff
@@ -31,7 +36,7 @@ describe('core/persistence/browser', () => {
   afterEach(() => sinon.restore());
 
   describe('browserLocalPersistence', () => {
-    const persistence: Persistence = browserLocalPersistence as Persistence;
+    const persistence: Persistence = (browserLocalPersistence as PersistenceInstantiator)._getInstance();
```
avolkovi, line 39: you do this cast all over the place, why not make this an importable function like `_getPeristenceInstance(persistence: externs.Persistence): Persistence`

packages-exp/auth-exp/src/core/persistence/browser.ts
```diff
@@ -56,9 +56,28 @@ class BrowserPersistence implements Persistence {
...
+  static instance: Persistence | null = null;
+  static _getInstance(): Persistence {
+    if (!this.instance) {
+      this.instance = new BrowserPersistence(localStorage);
+    }
+    return this.instance;
+  }
+}
+
+class BrowserSessionPersistence {
```
avolkovi, line 70: shouldn't these implement `PersistenceInstantiator` to make the cast safe?

packages-exp/auth-exp/src/core/persistence/browser.ts
```diff
@@ -56,9 +56,28 @@ class BrowserPersistence implements Persistence {
...
-export const browserLocalPersistence: externs.Persistence = new BrowserPersistence(
-  localStorage
-);
-export const browserSessionPersistence: externs.Persistence = new BrowserPersistence(
-  sessionStorage
-);
+class BrowserLocalPersistence {
+  static type: 'LOCAL' = 'LOCAL';
+  static instance: Persistence | null = null;
+  static _getInstance(): Persistence {
```
avolkovi, line 62: can you make this a generic function and just implement once on PersistenceInstantiator? `_getInstance<T>(): T`

packages-exp/auth-exp/src/core/persistence/browser.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/browser.ts
samhorlbeck: Obsolete; there's no way to tell TS "hey, this class's static instance implements these methods" unfortunately

packages-exp/auth-exp/src/core/persistence/browser.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/persistence/react_native.ts
```diff
@@ -25,36 +25,50 @@ import {
...
+ * be able to be instantiated (as a class) using "new". That function also
+ * expects the constructor to be empty. Since ReactNativeStorage requires the
+ * underlying storage layer, we need to be able to create subclasses
+ * (closures, esentially) that have the storage layer but empty constructor.
  */
-export class ReactNativePersistence implements Persistence {
-  readonly type: PersistenceType = PersistenceType.LOCAL;
 
-  constructor(private readonly storage: ReactNativeAsyncStorage) {}
+export function makeReactNativePersistence(
```
avolkovi, line 41: nit: we should call this reactNativePersistenceGenerator or Provider to match the other places we have providers, alternatively call "getReactNativePersistence" to match getMultiFactorResolver naming

samhorlbeck: Done

PR #3254, opened 2020-06-22
add support for set() with SetOptions when using FirestoreDataConverter

thebrianchen:
Porting over from [node](https://github.com/googleapis/nodejs-firestore/pull/1087).

PR #3291, opened 2020-06-25
Add set() overrides to lite sdk

thebrianchen:
Setting the base as #3254 since it requires the changes to `applyFirestoreDataConverter`. Since this branch sits on top of #3254, I will diff this against master once #3254 goes in.


packages/firestore/lite/index.d.ts
```diff
@@ -45,6 +45,7 @@ export function setLogLevel(logLevel: LogLevel): void;
 
 export interface FirestoreDataConverter<T> {
   toFirestore(modelObject: T): DocumentData;
+  toFirestore(modelObject: Partial<T>, options: SetOptions): DocumentData;
   fromFirestore(snapshot: QueryDocumentSnapshot): T;
```
schmidt-sebastian, line 49: Should we explicitly type this as QueryDocumentSnapshot<DocumentData> to make it easier to understand?

packages/firestore/lite/test/integration.test.ts
```diff
@@ -1045,4 +1076,111 @@ describe('withConverter() support', () => {
...
+        batch.set(ref, { title: 'olive' }, { merge: true })
+      ).to.throw(
+        'Function WriteBatch.set() called with invalid data ' +
+          '(via `toFirestore()`). Unsupported field value: undefined ' +
+          '(found in field author in document posts/post)'
+      );
+    });
+  });
+
+  it('WriteBatch.set() supports partials with merge', async () => {
```
schmidt-sebastian, line 1095: Can you move these tests to https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/lite/test/integration.test.ts#L253, which allows you to write one test for WriteBatch, Transaction and set().

packages/firestore/lite/index.d.ts
```diff
@@ -45,6 +45,7 @@ export function setLogLevel(logLevel: LogLevel): void;
 
 export interface FirestoreDataConverter<T> {
   toFirestore(modelObject: T): DocumentData;
+  toFirestore(modelObject: Partial<T>, options: SetOptions): DocumentData;
```
schmidt-sebastian, line 48: Note that this change also need to be made in exp/index.d.ts

packages/firestore/lite/index.d.ts
thebrianchen: done.

packages/firestore/lite/index.d.ts
thebrianchen: done for converter and `SetOptions`.

packages/firestore/lite/test/integration.test.ts
thebrianchen: done.

PR #3294, opened 2020-06-25
Add getQuery(), getQueryFromCache() and getQueryFromServer()

schmidt-sebastian:
This adds getQuery(), getQueryFromCache() and getQueryFromServer() as well as a QuerySnapshot that is heavily inspired by the existing SDK's QuerySnapshot.

Most code churn is to allow code reuse.


packages/firestore/exp/src/api/reference.ts
```diff
@@ -83,6 +84,63 @@ export function getDocFromServer<T>(
   });
 }
 
+export function getQuery<T>(
+  query: firestore.Query<T>
+): Promise<QuerySnapshot<T>> {
+  const internalQuery = cast<Query<T>>(query, Query);
+  const firestore = cast<Firestore>(query.firestore, Firestore);
+  return firestore._getFirestoreClient().then(async firestoreClient => {
```
rafikhan, line 92: Can you make methods like this fully async? I'm not seeing why they need mixed notation.  This applies to the other functions below.

schmidt-sebastian: The outermost code needs to be async to ensure synchronous function execution, which impacts error propagation and execution order. The `cast` calls here, for example, would not throw but cause rejected Promises.

PR #3306, opened 2020-06-26
Add setDoc, updateDoc, deleteDoc and addDoc

schmidt-sebastian:
While this is a large PR, this is mostly a copy of code from the Lite SDK, but instead of Datastore it uses FirestoreClient to commit the write (which adds it to LocalStore first).


packages/firestore/exp/src/api/reference.ts
```diff
@@ -42,6 +53,131 @@ export function getDoc<T>(
...
+    parsed = dataReader.parseUpdateData(
+      'updateDoc',
+      ref._key,
+      fieldOrUpdateData
+    );
+  }
+
+  return firestore
+    ._getFirestoreClient()
+    .then(firestoreClient =>
```
rafikhan, line 135: Same comment as the previous PR. Why do we mix promise chaining like this vs using async/await?

rafikhan, line 135: Same comment from previous PR's. Why is there chaining instead of just using async/await?

schmidt-sebastian: Same response: )

"The outermost code needs to be async to ensure synchronous function execution, which impacts error propagation and execution order. The cast calls here, for example, would not throw but cause rejected Promises."

In this case, this also impact user input validation, which we want to throw as it is likely caused by invalid SDK usage.

Issue #3305, opened 2020-06-26
IE11 error message on Firebase SDK 7.15.5


tandalesc:
### Environment
  * Operating System version: Windows 10
  * Browser version: Internet Explorer 11
  * Firebase SDK version: 7.15.5
  * Firebase Product: core (auth, database, storage, etc)

### Problem
Hello,

I was working on a new project and everything was going smoothly until I tried to view the webpage in IE11. On Firebase 7.15.5, the site does not load at all. I tried earlier versions of the Firebase JS SDK and they all work -- including 7.15.4. On other browsers and operating systems, the site works without issue.

I'm getting the following error message on 7.15.5 in IE11:

> Accessing the 'caller' property of a function or arguments object is not allowed in strict mode

I believe I've traced the problem to this file: `packages/firestore/src/util/api.ts`. Specifically, this section (around line 50):
```
for (const staticProp of Object.getOwnPropertyNames(cls)) {
    if (RESERVED_READONLY_PROPS.indexOf(staticProp) === -1) {
      // eslint-disable-next-line @typescript-eslint/no-explicit-any
      (PublicConstructor as any)[staticProp] = (cls as any)[staticProp];
    }
}
```
Unfortunately, I'm not super familiar with the inner workings of the Firebase SDK so I'm not sure how to best address this issue.

I didn't include any code samples because the problem occurred during the initial startup process. I don't even have authentication or firestore collections setup yet. Please let me know if I can help in any way!

JFGHT:
Can confirm.

kbi-daniel:
Just happened to a project I'm working on.  Reverted to 7.15.4 and everything works fine.

schmidt-sebastian:
We will remove the code that caused this breakage altogether, as we are currently not able to run our tests in IE11.

vandres:
@schmidt-sebastian Sorry to bother you but when will this be released?

schmidt-sebastian:
End of this week, likely Thursday the 11th.

PR #3309, opened 2020-06-26
Remove makeConstructorPrivate

schmidt-sebastian:
The makeConstructorPrivate pattern is used in the SDK to prevent JS SDK users from constructing our internal types. This pattern is unique to Firestore, but seems to break in subtle and not so subtle ways that we currently cannot catch in CI (due to missing Saucelabs tests). Instead of iterating on this pattern, we decided to remove it altogether, which also adds a little bit of tree-shakeability to our current SKD (but probably only for users of `Blob`).

Fixes https://github.com/firebase/firebase-js-sdk/issues/3305


.changeset/small-ghosts-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/firestore": patch
```
Feiyang1, line 2: Please add `"firebase": patch`

schmidt-sebastian: Done

PR #3317, opened 2020-06-29
Add Snapshot Listeners to firestore-exp

schmidt-sebastian:
This PR adds the Snapshot Listeners to the firestore-exp SDK. It does not add tests, since I want these functions to be tested using the main integration tests. 

The PR also adds the call to `validateHasExplicitOrderByForLimitToLast` to all Query evaluations. This is needed to make sure that `limitToLast()` is not invoked on an empty Query order, since the backend does not support a reverse index on the document ID.

Source for all this code is https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/database.ts#L1136 and https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/database.ts#L2100


packages/firestore/exp/src/api/snapshot.ts
```diff
@@ -121,15 +121,21 @@ export class QueryDocumentSnapshot<T = firestore.DocumentData>
...
 
   constructor(
     readonly _firestore: Firestore,
     readonly query: Query<T>,
-    readonly _snapshot: ViewSnapshot,
-    readonly metadata: SnapshotMetadata
-  ) {}
+    readonly _snapshot: ViewSnapshot
+  ) {
+    this.metadata = new SnapshotMetadata(
```
schmidt-sebastian, line 134: The metadata can be constructed from the view snapshot.

packages/firestore/src/api/database.ts
```diff
@@ -1879,6 +1868,17 @@ export class BaseQuery {
   }
 }
 
+export function validateHasExplicitOrderByForLimitToLast(
```
rafikhan, line 1871: Nice name!

packages/firestore/exp/src/api/reference.ts
```diff
@@ -280,6 +278,159 @@ export function addDoc<T>(
...
 
+// TODO(firestorexp): Make sure these overloads are testing via the Firestore
+// integration tests
+export function onSnapshot<T>(
+  reference: firestore.DocumentReference<T>,
+  observer: {
+    next?: (snapshot: firestore.DocumentSnapshot<T>) => void;
+    error?: (error: firestore.FirestoreError) => void;
+    complete?: () => void;
+  }
```
rafikhan, line 289: I've been meaning to ask this for some time but what's the history behind using 3 separate callbacks in web vs what we do in other languages?

schmidt-sebastian: They are meant to make it easy to plug in RxObservables.

packages/firestore/src/api/database.ts
schmidt-sebastian: Credit goes to @wu-hui :)

packages/firestore/exp/src/api/snapshot.ts
schmidt-sebastian: I noticed that we don't pass the SnapshotMetadata to the QuerySnapshot in the main SDK. QuerySnapshot is basically a view on top of the ViewSnapshot, so the metadata matches that of the ViewSnapshot.  In the ViewSnapshot, in can be arbitrarily constructed.

PR #3320, opened 2020-06-29
Add onSnapshotsInSync to firestore-exp

schmidt-sebastian:
This adds the code for onSnapshotsInSync, tests are coming when we can target the existing integration tests against firestore-exp


packages/firestore/src/api/database.ts
```diff
@@ -675,6 +657,29 @@ export class Firestore implements firestore.FirebaseFirestore, FirebaseService {
   }
 }
 
+/** Register the listener for onSnapshotsInSync() */
```
wu-hui, line 660: `Registers`.

schmidt-sebastian: Done

PR #3321, opened 2020-06-29
Add remaining firestore-exp functionality

schmidt-sebastian:
This adds some barebones support for the remaining API surface for the firestore-exp client. This is meant to unblock adding the Dependencies tests as well as the integration tests. TODOs added for functionality that is missing.


packages/firestore/exp/src/api/database.ts
```diff
@@ -80,16 +90,71 @@ export class Firestore extends LiteFirestore
...
+
   delete(): Promise<void> {
     return terminate(this);
   }
+
+  _clearPersistence(): Promise<void> {
+    if (this._firestoreClientPromise !== undefined && !this._terminated) {
+      throw new FirestoreError(
+        Code.FAILED_PRECONDITION,
+        'Persistence cannot be cleared after this Firestore instance is initialized.'
```
dconeybe, line 137: nit: This statement is not entirely correct. Consider this rewording:

*Persistence can only be cleared before the Firestore instance is initialized or after it is terminated.*

schmidt-sebastian: Udpated

Issue #3289, opened 2020-06-25
Callable Function throw Deadline Exceeded Error even on successful response


stevecastaneda:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS Catalina 10.15.5
  * Browser version: Chrome 83.0.4103.116
  * Firebase SDK version: 7.15.5
  * Firebase Product: Functions
  * Issue persists with Node 12.18.1 and 14.4.0

https://github.com/firebase/firebase-js-sdk/blob/ee130b56b2813fb1c01fd9430059814ece45a738/packages/functions/src/api/service.ts#L187

#### Steps to reproduce:
This is a continuation from issue #2746, and I'm experiencing the same `deadline-exceeded` error even after Promise.race successfully returns a promise (as evidenced by the callable function returning data as I expect).

@minism

#### Function Relevant Code: [Typescript]

```
import * as functions from "firebase-functions";
import * as admin from "firebase-admin";
admin.initializeApp();

export const helloUser = functions.https.onCall((data, context) => {
  if (context.auth) {
    return { response: "There's a user!", data };
  } else {
    throw new functions.https.HttpsError("permission-denied", "You must be logged in.");
  }
});
```
### Frontend (React)
```
async function testFunction() {
  try {
        let functionTest = functions.httpsCallable("helloUser");
        let response = await functionTest();
        console.log(response);
      } catch (error) {
        console.log(error);
      }
}
```

stevecastaneda:
For reference, could this be a proposed solution so that the timer is cleared when Promise.race is satisfied?

https://stackoverflow.com/questions/30936824/what-is-the-best-general-practice-to-timeout-a-function-in-promise

```
export function race({promise, timeout, error}) {
  let timer = null;

  return Promise.race([
    new Promise((resolve, reject) => {
      timer = setTimeout(reject, timeout, error);
      return timer;
    }),
    promise.then((value) => {
      clearTimeout(timer);
      return value;
    })
  ]);
}
```

Feiyang1:
@stevecastaneda issue #2746 mentions the issue only manifests in Node. Is your issue also happening in Nodejs (server rendering)? Just want to clarify because you include React code, also that I can't reproduce it in browser.

stevecastaneda:
[...]

PR #3330, opened 2020-07-01
clear failAfter timeout after request resolves

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/3289


Feiyang1:
Awesome! added the test.

PR #3335, opened 2020-07-01
Add popup util and bare-bones OAuthProvider

samhorlbeck:


packages-exp/auth-exp/src/core/providers/oauth.ts
```diff
@@ -0,0 +1,82 @@
...
+ * limitations under the License.
+ */
+
+import * as externs from '@firebase/auth-types-exp';
+
+import { AuthErrorCode } from '../errors';
+import { debugFail } from '../util/assert';
+
+export interface CustomParameters {
+  [key: string]: string;
```
avolkovi, line 24: Record<string, string>?

packages-exp/auth-exp/src/core/providers/oauth.ts
```diff
@@ -0,0 +1,82 @@
...
+  idToken?: string;
+  accessToken?: string;
+  rawNonce?: string;
+}
+
+export class OAuthProvider implements externs.AuthProvider {
+  defaultLanguageCode: string | null = null;
+  private scopes: string[] = [];
+  private customParameters: CustomParameters = {};
+  constructor(readonly providerId: externs.ProviderId) {}
```
avolkovi, line 37: make constructor private pls

packages-exp/auth-exp/src/core/providers/oauth.ts
```diff
@@ -0,0 +1,82 @@
...
+  static credentialFromJSON(_json: object): externs.OAuthCredential {
+    debugFail('not implemented');
+  }
+
+  credential(_params: CredentialParameters): externs.OAuthCredential {
+    debugFail('not implemented');
+  }
+
+  setDefaultLanguage(languageCode: string | null): void {
+    this.defaultLanguageCode = languageCode;
```
avolkovi, line 57: should this pull from navigator language like Auth does?

packages-exp/auth-exp/src/core/providers/oauth.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  setCustomParameters(
+    customOAuthParameters: CustomParameters
+  ): externs.AuthProvider {
+    this.customParameters = customOAuthParameters;
+    return this;
+  }
+
+  getCustomParameters(): CustomParameters {
```
avolkovi, line 67: get/set?

packages-exp/auth-exp/src/core/util/popup.ts
```diff
@@ -0,0 +1,108 @@
...
+    ...BASE_POPUP_OPTIONS,
+    width: width.toString(),
+    height: height.toString(),
+    top,
+    left
+  };
+
+  // Chrome iOS 7 and 8 is returning an undefined popup win when target is
+  // specified, even though the popup is not necessarily blocked.
+  const ua = _uaGetter.getUA().toLowerCase();
```
avolkovi, line 75: use the helpers in firebase/util please

packages-exp/auth-exp/src/core/util/popup.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  if (name) {
+    target = ua.includes(CHROME_IOS_UA) ? '_blank' : name;
+  }
+
+  if (ua.includes(FIREFOX_UA)) {
```
avolkovi, line 81: make an isFirefox() helper in util

packages-exp/auth-exp/src/core/util/popup.ts
```diff
@@ -0,0 +1,108 @@
...
+    top,
+    left
+  };
+
+  // Chrome iOS 7 and 8 is returning an undefined popup win when target is
+  // specified, even though the popup is not necessarily blocked.
+  const ua = _uaGetter.getUA().toLowerCase();
+
+  if (name) {
+    target = ua.includes(CHROME_IOS_UA) ? '_blank' : name;
```
avolkovi, line 78: same here, please make an isChromeIOS() util helper

packages-exp/auth-exp/src/core/util/popup.ts
```diff
@@ -0,0 +1,108 @@
...
+    // issues when the user has many Google accounts, for instance.
+    options.scrollbars = 'yes';
+  }
+
+  const optionsString = Object.entries(options).reduce(
+    (accum, [key, value]) => `${accum}${key}=${value},`,
+    ''
+  );
+
+  // TODO: Plain-old window.open isn't going to work for iOS, need to fix this
```
avolkovi, line 94: is there a polyfill we can just depend on that would make this easier?

packages-exp/auth-exp/src/core/providers/oauth.ts
samhorlbeck: It's meant to be public: https://firebase.google.com/docs/reference/js/firebase.auth.OAuthProvider#constructor

packages-exp/auth-exp/src/core/providers/oauth.ts
samhorlbeck: Can't use those in tree-shaking world. See go/guidelines-tree-shaking

packages-exp/auth-exp/src/core/providers/oauth.ts
samhorlbeck: Not in this method: https://github.com/firebase/firebase-js-sdk/blob/992432e07ca3b7594b860c0d50da23596311ea56/packages/auth/src/authcredential.js#L547

packages-exp/auth-exp/src/core/util/popup.ts
samhorlbeck: This is just so that we can stub this out in test without importing all of firebase/util.

The `import * as <something> from '...'` works for stubbing BUT it also will break tree shaking (the entire module will be imported).

This is fine when there's only one or two functions in that module, but for something like firebase/util, it's a no-go

packages-exp/auth-exp/src/core/providers/oauth.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/util/popup.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/util/popup.ts
samhorlbeck: No, but I just added the implementation

packages-exp/auth-exp/src/core/providers/oauth.ts
avolkovi: boo

packages-exp/auth-exp/src/core/util/popup.ts
avolkovi: we only need to do the import * as something in the tests, righty? in the implementation you can just pull in the method you want, either way we pull the UA helpers elsewhere

PR #3336, opened 2020-07-01
Add browser popup resolver class, some utils

samhorlbeck:


packages-exp/auth-exp/src/model/popup_redirect.d.ts
```diff
@@ -0,0 +1,83 @@
...
+  LINK_VIA_REDIRECT = 'linkViaRedirect',
+  REAUTH_VIA_POPUP = 'reauthViaPopup',
+  REAUTH_VIA_REDIRECT = 'reauthViaRedirect',
+  SIGN_IN_VIA_POPUP = 'signInViaPopup',
+  SIGN_IN_VIA_REDIRECT = 'signInViaRedirect',
+  UNKNOWN = 'unknown',
+  VERIFY_APP = 'verifyApp'
+}
+
+// TODO: convert from these to FirebaseError
```
avolkovi, line 48: can we do that as part of this PR?

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -0,0 +1,143 @@
...
+    auth: Auth,
+    provider: externs.AuthProvider,
+    authType: AuthEventType,
+    eventId?: string
+  ): Promise<AuthPopup> {
+    const url = getRedirectUrl(auth, provider, authType, eventId);
+    return _open(auth.name, url, _generateEventId());
+  }
+
+  async initialize(auth: Auth): Promise<EventManager> {
```
avolkovi, line 59: does this need that promise chaining behavior to prevent interleaving like we have on auth/user?

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -0,0 +1,143 @@
...
+    // TODO set additionalParams?
+    // let additionalParams = provider.getAdditionalParams();
+    // for (let key in additionalParams) {
+    //   if (!params.hasOwnProperty(key)) {
+    //     params[key] = additionalParams[key]
+    //   }
+    // }
+  }
+
+  // TODO: maybe set tid as tenantId
```
avolkovi, line 132: we should definitely do this at least now that we have it on auth

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -0,0 +1,143 @@
...
+    provider.setDefaultLanguage(auth.languageCode);
+    params.providerId = provider.providerId || '';
+    if (!isEmpty(provider.getCustomParameters())) {
+      params.customParameters = JSON.stringify(provider.getCustomParameters());
+    }
+    const scopes = provider.getScopes();
+    if (scopes.length > 0) {
+      params.scopes = scopes.join(',');
+    }
+    // TODO set additionalParams?
```
avolkovi, line 123: can you see if we have to do this? I forget why I left it off on the auth-ts branch

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -0,0 +1,143 @@
...
+  PopupRedirectResolver
+} from '../model/popup_redirect';
+import { _openIframe } from './iframe/iframe';
+
+/**
+ * URL for Authentication widget which will initiate the OAuth handshake
+ */
+const WIDGET_URL = '__/auth/handler';
+
+export class BrowserPopupRedirectResolver implements PopupRedirectResolver {
```
avolkovi, line 44: shouldn't there be a redirect method on this?

avolkovi: also add this to index.ts

packages-exp/auth-exp/src/model/popup_redirect.d.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export interface EventManager {
+  registerConsumer(authEventConsumer: AuthEventConsumer): void;
+  unregisterConsumer(authEventConsumer: AuthEventConsumer): void;
+}
+
+export interface PopupRedirectResolver extends externs.PopupRedirectResolver {
+  initialize(auth: Auth): Promise<EventManager>;
+  openPopup(
```
avolkovi, line 77: both this and initialize should be _packageprivate

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -0,0 +1,143 @@
...
+  private eventManager: EventManager | null = null;
+
+  // Wrapping in async even though we don't await anywhere in order
+  // to make sure errors are raised as promise rejections
+  async openPopup(
```
avolkovi, line 49: don't we need to await initialize()?

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: Re. redirect method: coming soon to a PR near you!

samhorlbeck: I'm going to also push the index.ts bit off for a bit. We'll need to do an instantiation thing like we do with the persistence classes, and I'd like to try to make all of that more generic

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: No, the way this new architecture works is callers are required to call _initialize() in order to get the event manager.

I've added a `debugAssert` for this

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: It doesn't come from the provider. Only place I've found it used is in the cordova handler. I think we should take care of this when we do cordova

packages-exp/auth-exp/src/model/popup_redirect.d.ts
samhorlbeck: Done

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: Doesn't need chaining but it should be cached as an init promise; good catch

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: Done

packages-exp/auth-exp/src/model/popup_redirect.d.ts
samhorlbeck: Done

PR #3358, opened 2020-07-06
Misc fixes for tree-shakeable API

schmidt-sebastian:
Contains a couple of bug fixes for the tree-shakeable clients, which align them more with the regular client (as shown by the integration tests).

- Added override for `Firestore._makeDatabaseInfo` to the firestore-exp client since firestore-exp needs access to the persistence key.
- Changed terminate handling so that firestore-exp is able to throw exceptions for method that are invoked after termination, rather than rejecting a Promise.
- Cleaned up the code around `Firestore._firestoreClient`. Since only `DatabaseInfo` cannot be initialized in the constructor, I moved this argument to `FirestoreClient.start()`. This allows us to instantiate the FirestoreClient directly.
- Added missing validation for `cacheSizeBytes` setting.
- Some other random one-line fixes.


packages/firestore/exp/src/api/database.ts
```diff
@@ -131,11 +143,10 @@ export class Firestore extends LiteFirestore
   }
 
   _clearPersistence(): Promise<void> {
-    if (this._firestoreClientPromise !== undefined && !this._terminated) {
+    if (this._deferredInitialization !== undefined && !this._terminated) {
       throw new FirestoreError(
         Code.FAILED_PRECONDITION,
-        'Persistence can only be cleared before the Firestore instance is ' +
-          'initialized or after it is terminated.'
+        'Persistence cannot be cleared after this Firestore instance is initialized.'
```
dconeybe, line 149: This statement is not entirely correct, since you *can* clear persistence after terminating the client. I recommend using the original wording.

schmidt-sebastian: I changed the error here to match the error message in the existing SDK. I reverted this change and changed the error message in the main SDK.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1149,7 +1149,8 @@ apiDescribe('Database', (persistence: boolean) => {
         await expect(
           firestore.clearPersistence()
         ).to.eventually.be.rejectedWith(
-          'Persistence cannot be cleared after this Firestore instance is initialized.'
+          'Persistence cannot be cleared after this Firestore instance is ' +
+            'initialized or after it is terminated.'
```
dconeybe, line 1153: Also fix this message.

packages/firestore/exp/src/api/database.ts
```diff
@@ -146,7 +146,8 @@ export class Firestore extends LiteFirestore
     if (this._deferredInitialization !== undefined && !this._terminated) {
       throw new FirestoreError(
         Code.FAILED_PRECONDITION,
-        'Persistence cannot be cleared after this Firestore instance is initialized.'
+        'Persistence cannot be cleared after this Firestore instance is ' +
+          'initialized or after it is terminated.'
```
dconeybe, line 150: This message is now more incorrect. Persistence *can* be cleared after it is terminated. The error message says that it cannot.

Suggested rewording: Persistence cannot be cleared after this Firestore instance is initialized and before it is terminated.

schmidt-sebastian: "more incorrect" almost sounds like an improvement :) Thanks for catching.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: Fixed

PR #3361, opened 2020-07-06
Add test-only API shim (firestore-exp)

schmidt-sebastian:
This PR adds the API shim that I am using to run the existing integration tests agains the tree-shakeable API. It essentially exposes the tree-shakeable client in terms of the old API.

The code is unused right now, but works as shown here: https://github.com/firebase/firebase-js-sdk/pull/3338


packages/firestore/exp/test/shim.ts
```diff
@@ -0,0 +1,768 @@
...
+      }
+    }
+    return obj;
+  } else {
+    return value;
+  }
+}
+
+/**
+ * Takes user data that uses API types from this shim and replaces them
```
thebrianchen, line 745: ultra-uber-nit (so that I can say I left a comment): Move "with" to line above.

packages/firestore/exp/test/shim.ts
```diff
@@ -0,0 +1,768 @@
...
+    if (options) {
+      this._delegate.set(documentRef._delegate, unwrap(data), options);
+    } else {
+      this._delegate.set(documentRef._delegate, unwrap(data));
+    }
+    return this;
+  }
+
+  update(
+    documentRef: DocumentReference<any>,
```
thebrianchen, line 168: Is there a reason why we're using `<any>` here as opposed to `<unknown>`?

packages/firestore/exp/test/shim.ts
```diff
@@ -0,0 +1,768 @@
...
+    options: legacy.SnapshotListenOptions,
+    onNext: (snapshot: DocumentSnapshot<T>) => void,
+    onError?: (error: Error) => void,
+    onCompletion?: () => void
+  ): () => void;
+  onSnapshot(...args: any): () => void {
+    let options: legacy.SnapshotListenOptions = {};
+    let userObserver: PartialObserver<DocumentSnapshot<T>>;
+
+    if (isPartialObserver(args[0])) {
```
thebrianchen, line 356: optional: There seems to be some code overlap between here and L547. You could probably extract the logic into a shared method, but might not be worth it considering it's just shim testing code.

packages/firestore/exp/test/shim.ts
```diff
@@ -0,0 +1,768 @@
...
+    return new FieldValue(increment(n));
+  }
+
+  isEqual(other: FieldValue): boolean {
+    return this._delegate.isEqual(other._delegate);
+  }
+}
+
+export class FieldPath implements legacy.FieldPath {
+  private fieldNames: string[];
```
thebrianchen, line 701: nit: (courtesy of intellij) s/private/private readonly

packages/firestore/exp/test/shim.ts
schmidt-sebastian: Good idea! Done.

packages/firestore/exp/test/shim.ts
schmidt-sebastian: Done

schmidt-sebastian: (Sorry - I just realized I forgot to do this :/)

PR #3362, opened 2020-07-06
Remove Multi-Tab overhead for non-multi-tab persistence

schmidt-sebastian:
This PR changes the name of existing IndexedDbComponentProvider to MultiTabComponentProvider and adds a new IndexedDbComponentProvider, which is used in firestore-exp to reduce the code size for users that use `enabledIndexedDbPersistence` instead of `enableMultiTabIndexedDbPersistence`. Unfortunately, we have to use MultiTabComponentProvider in the old SDK since we don't know which mode the SDK will end up picking.


packages/firestore/exp/test/deps/dependencies.json
```diff
@@ -9243,12 +9228,11 @@
...
                 "ViewSnapshot",
                 "WatchChangeAggregator",
-                "WatchTargetChange",
-                "WebStorageSharedClientState"
+                "WatchTargetChange"
             ],
             "variables": []
         },
-        "sizeInBytes": 335737
+        "sizeInBytes": 311847
```
schmidt-sebastian, line 9235: This change drops about 24KB from `enableIndexedDbPersistence`.

PR #3364, opened 2020-07-07
Add test-only API shim (Lite SDK)

schmidt-sebastian:
This PR adds the API shim that I am using to run the existing integration tests agains the tree-shakeable API. It essentially exposes the tree-shakeable client in terms of the old API.

Same logic as https://github.com/firebase/firebase-js-sdk/pull/3361


packages/firestore/lite/test/shim.ts
```diff
@@ -0,0 +1,649 @@
...
+    return new FieldValue(increment(n));
+  }
+
+  isEqual(other: FieldValue): boolean {
+    return this._delegate.isEqual(other._delegate);
+  }
+}
+
+export class FieldPath implements legacy.FieldPath {
+  private fieldNames: string[];
```
thebrianchen, line 582: s/private/private readonly

PR #3369, opened 2020-07-08
Make the _getInstance() instantiator a util for the whole SDK

samhorlbeck:


packages-exp/auth-exp/src/core/util/instantiator.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Unfortunately we can't export these as classes instantiated directly since
+ * the constructor may side effect and therefore can't be proven to be safely
+ * culled. Instead, we export these classes themselves as a lowerCamelCase
+ * constant, and instantiate them under the hood.
+ */
+export interface SingletonInstantiator<T> {
+  new (): T;
+}
+
+const persistenceCache: Map<unknown, unknown> = new Map();
```
avolkovi, line 29: this is no longer persistence specific, maybe call it instanceCache?

packages-exp/auth-exp/src/core/util/instantiator.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function _getInstance<T>(cls: unknown): T {
+  if (persistenceCache.has(cls)) {
+    return persistenceCache.get(cls)! as T;
+  }
+
+  const persistence = new (cls as SingletonInstantiator<T>)();
```
avolkovi, line 36: same here, `const instance`

packages-exp/auth-exp/src/core/util/instantiator.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/util/instantiator.ts
samhorlbeck: Done

PR #3372, opened 2020-07-08
Upgrade to latest version of closure

rafikhan:
This PR upgrades to the latest version of closure (20200628.0.0)


.changeset/calm-apes-collect.md
```diff
@@ -0,0 +1,7 @@
+---
+'@firebase/webchannel-wrapper': minor
```
schmidt-sebastian, line 2: @Feiyang1 Do we also need to explicitly add '@firebase/firestore' and 'firebase'? i assume this is needed to make sure they get released, but wanted to double check.

.changeset/calm-apes-collect.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Upgrade to the latest version of google closure library and compiler. This dependency will be
```
schmidt-sebastian, line 5: Google stylizes this as "Closure Library", so this should be "Google Closure Library and Compiler": https://developers.google.com/closure/library

rafikhan: Done

.changeset/calm-apes-collect.md
Feiyang1: No. It will trigger `patch` release for `@firebase/firestore` and `firebase` automatically which I guess is what you want.

Feiyang1: Why is it a minor release? looks like a patch release to me.

Feiyang1: Note that there is 5% increase in the library size. We didn't upgrade closure to the latest yet exactly because of it. @hsubox76 investigated it and found it's because the usage of Set is no longer stripped in the closure library - https://github.com/google/closure-library/issues/1070. However the closure team has not done anything to fix it.

schmidt-sebastian: We settled on a "minor" release to mitigate some of the risks associated with jumping between major versions in google-closure-library. Have we previously just released patch releases for these updates?

Feiyang1: Yes, we have done patch releases for these updates in the past. As long as the Firestore node tests pass, we should be confident that it works, right?

In case you actually want to bump minor for `@firebase/firestore` and thus `firebase`, you should include them explicitly in the changeset.

PR #3373, opened 2020-07-08
Make clearIndexedDbPersistence() work without enableIndexedDbPersistence()

schmidt-sebastian:
In the current firestore-exp client, clearIndexedDbPersistence() cannot be called on an uninitialized instance since the IndexedDbComponentProvider is not registered. This PR changes the plumbing so that clearIndexedDbPersistence() brings all logic that it needs to perform the delete without pulling in IndexedDbPersistence or the IndexedDbComponentProvider.

Note that I did update the dependencies file, but the firestore-exp client is currently so bad at being "tree-shaken" that the dependencies don't make much sense.


packages/auth/src/args.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC
```
wilhuff, line 3: These changes seem unrelated. Revert?

packages/firestore/exp/src/api/database.ts
```diff
@@ -247,7 +254,9 @@ export function clearIndexedDbPersistence(
   firestore: firestore.FirebaseFirestore
 ): Promise<void> {
   const firestoreImpl = cast(firestore, Firestore);
-  return firestoreImpl._clearPersistence();
+  return firestoreImpl._clearPersistence((databaseId, persistenceKey) => {
+    return clearPersistence(buildStoragePrefix(databaseId, persistenceKey));
```
wilhuff, line 258: This code reads as if it's circular unless you pay careful attention. If `clearPersistence` were named `indexedDbClearPersistence` it would be clearer that the higher level API was delegating to the lower-level implementation.

I realize this is getting at that type-operation naming discussion we started earlier so feel free to defer if you must, but I think it's really hurting the ability to understand what's going on here while reading the code directly. Sure you can go look at the imports but that breaks flow. Reading code is hard enough and we should strive to make it easy to understand.

packages/auth/src/args.js
schmidt-sebastian: Removed. I actually tried to remove them earlier, but the pre commit hook brought them back in. I had to disable the precommit hook for this commit.

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: I stumbled across this earlier too, but didn't act upon my confusion. I think a minimum bar for code quality should be that at the very least the person writing it shouldn't be confused...

Renamed to `indexedDbClearPersistence` and `indexedDbStoragePrefix`.

PR #3374, opened 2020-07-08
Share Integration Tests between firestore-exp and the main SDK

schmidt-sebastian:
This PR removes the custom integration tests for firestore-exp and changes the integration tests under `test/integration/api/` to be able to run against the firestore-exp SDK (which I started calling "modular SDK" in the code, but maybe I shouldn't do that).

The logic is based on the FirestoreShim added in https://github.com/firebase/firebase-js-sdk/pull/3361

Couple things:
- I verified that we didn't lose any test coverage.
- I updated two error messages in the firestore-exp SDK to match those in the main SDK.
- I added the new tests to the the main test rule in package.json, making sure they run on CI.
- I added test runners for IntelliJ.
- The update to "firebase_export" is to make sure that the "minified" integration tests continue to work as is.


integration/firestore/firebase_export.ts
```diff
@@ -0,0 +1,56 @@
...
+      : nameOrApp;
+
+  const firestore = firebase.firestore(app);
+  if (settings) {
+    firestore.settings(settings);
+  }
+  return firestore;
+}
+
+export function usesModularApi(): false {
```
wilhuff, line 45: "Modular" doesn't really cleanly describe the difference between the the Firestore-API-with-classes vs the tree-shakeable one.

The core of the difference seems to be class-based vs function-based. To me this suggests that this should be the basis of this kind of property.

wilhuff:
> (No code updates yet).
> 
> The future plan for clients that are rewritten from scratch rather than replaced inline was to build a shim layer that implements the new API in terms of the old API. I essentially implemented this for our tests. There are some areas where the Shim does less work (e.g. validation), but it is pretty close to the model that @Feiyang1 had originally proposed for all SDKs.

Ultimately though, the model that was proposed was that the classic API would delegate to the new exp API though. As proposed this PR doesn't do that; instead it leaves two parallel implementations that just happen to be testable via the same integration tests.

I suppose what you're saying right now is that you don't have enough confidence in the exp API to make this swap, and that's fair. If the intent behind this PR is that this arrangement is temporary I'm more OK with it. I still think it's a lot of plumbing though.

> I can add some of the validation to the Shim layer, which would mean we would run more tests. Even with that, it still feels really scary to me to replace our existing code with the shim (at least at this point in time).

Isn't this validation logic something we'd need in firestore-exp? If it's not needed there, why is it needed in the classic Firestore?

> I also have to rewrite all user data as the enter and leave the Shim, which we would need to optimize before our users are using the Shim.

What is the source of this problem? Aren't things like Timestamps and GeoPoints the same in both implementations?

> Furthermore, I can reduce the amount of plumbing by not adding the test files to exp/test and instead just changing the test rules. Is that something I should explore?

Re test rules, that seems reasonable.

Part of my concern though is that test duration in the JS repo is already quite long, so essentially doubling the size of the integration test suite seems painful. Moving to a model where the classic API delegated to the functional API would allow you to test them both together (because the classic API would be the shim).

schmidt-sebastian:
I updated the name of `useModularApi` to `useFunctionalApi` and removed the new test files.

> Ultimately though, the model that was proposed was that the classic API would delegate to the new exp API. As proposed this PR doesn't do that; instead it leaves two parallel implementations that just happen to be testable via the same integration tests.
> 
> I suppose what you're saying right now is that you don't have enough confidence in the exp API to make this swap, and that's fair. If the intent behind this PR is that this arrangement is temporary I'm more OK with it. I still think it's a lot of plumbing though.

The goal of this PR is for now just to share the tests, not the surface. If we did share the surface, we wouldn't need to share the tests and wouldn't have to run them twice (which is something I hadn't considered). 
I agree it is a good goal to get into a place where we only have one test run and one API surface, but I would like to do this over a matter of weeks and ideally only transform a small set of API types per release. This allows us to slowly roll out this change  and make it small increments, which will be much easier to review. It also gives us production testing of the experimental API. 

> Isn't this validation logic something we'd need in firestore-exp? If it's not needed there, why is it needed in the classic Firestore?

@Feiyang1 and I have been talking about our input validation, as it adds a lot of overhead to the SDK. At the moment, I have only implemented the kind of validation we have in the other Firestore SDKs, which should prevent invalid state for syntactically valid input. There is an ongoing effort to come to a conclusion on our strategy for Web SDK input validation: go/jscore-input-validation

If we added the input validation to the Shim layer and also implemented `timestampsInSnapshots`, almost no test changes would be needed. If we want to productionize the Shim layer, that would be a good first step (but it would likely increase the size of this PR, rather than decrease it).

>> I also have to rewrite all user data as the enter and leave the Shim, which we would need to optimize before our users are using the Shim.
> What is the source of this problem? Aren't things like Timestamps and GeoPoints the same in both implementations?

FieldPath, FieldValue, and DocumentReference. FieldPath and FieldValue would be relatively trivial to accept without rewriting, since they are input-only types.

schmidt-sebastian:
PTAL :) Thanks.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -925,13 +926,29 @@ apiDescribe('Database', (persistence: boolean) => {
 
   it('exposes "firestore" on document references.', () => {
     return withTestDb(persistence, async db => {
-      expect(db.doc('foo/bar').firestore).to.equal(db);
+      if (usesModularApi()) {
+        // FirestoreShim returns a new instance of FirebaseFirestore that is
+        // semantically identical to the originating instance, but does not
+        // compare equal using the default JavaScript semantics.
+        expect(db.doc('foo/bar').firestore).to.be.an.instanceof(Firestore);
```
wilhuff, line 933: Is there a way we could define equality of Firestore instances based on e.g. the project, database, and app name?

I don't mean necessarily adding equals methods or something like that, but just an assertion we could use in tests? That way we could assert the Firestore instances are equivalent and use the same assertion for both cases.

As it stands, verifying that the thing returned is an instance of Firestore ... isn't much of a test.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1140,12 +1155,18 @@ apiDescribe('Database', (persistence: boolean) => {
...
-        ).to.eventually.be.rejectedWith(
+        const expectedError =
           'Persistence can only be cleared before a Firestore instance is ' +
-            'initialized or after it is terminated.'
-        );
+          'initialized or after it is terminated.';
+        if (usesModularApi()) {
+          // The modular API throws an exception rather than rejecting the
+          // Promise, which matches our overall handling of API call violations.
+          expect(() => firestore.clearPersistence()).to.throw(expectedError);
```
wilhuff, line 1164: If this is the behavior we want, why not change the classic version to match?

packages/firestore/test/integration/api/fields.test.ts
```diff
@@ -353,33 +354,43 @@ apiDescribe('Timestamp Fields in snapshots', (persistence: boolean) => {
...
-          expect(docSnap.get('nested.timestamp2'))
-            .to.be.a('date')
-            .that.deep.equals(timestamp.toDate());
-          expect(docSnap.data()!['nested']['timestamp2'])
-            .to.be.a('date')
-            .that.deep.equals(timestamp.toDate());
-        });
-    });
-  });
+  // timestampInSnapshots is not support in the modular API.
```
wilhuff, line 357: Off topic: are we removing this option at Fireconf?

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -117,45 +120,61 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
-        'Firestore has already been started and its settings can no ' +
-          'longer be changed. You can only call settings() before ' +
-          'calling any other methods on a Firestore object.'
-      );
+      ).to.throw(errorMsg);
     });
 
-    validationIt(persistence, 'enforces minimum cache size', db => {
+    validationIt(persistence, 'enforces minimum cache size', () => {
+      const db = newTestFirestore('testProjectId');
```
wilhuff, line 170: This is an invalid project ID. Only lowercase, numbers, and dashes are allowed.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -271,30 +301,41 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
-      );
-      expect(() => baseCollectionRef.doc(undefined as any)).to.throw(
-        'Function CollectionReference.doc() requires its first ' +
-          'argument to be of type non-empty string, but it was: undefined'
-      );
-      expect(() => (baseCollectionRef.doc as any)('foo', 'bar')).to.throw(
-        'Function CollectionReference.doc() requires between 0 and ' +
-          '1 arguments, but was called with 2 arguments.'
-      );
+      if (usesModularApi()) {
```
wilhuff, line 304: Since the only differences in the error messages are the function name and argument position, it seems like we could redefine these tests as the common set and those that are additional for the classic API. As it stands there's nothing requiring any of the equivalent validation checks to be identical. Given their superficial dissimilarity it seems possible for someone to overlook that a message needs to be updated in both places.

I'm not actually sure this is a good idea, but I thought I'd mention the idea to see if it resonates with you at all.

packages/firestore/test/integration/api/fields.test.ts
```diff
@@ -27,9 +27,10 @@ import {
 } from '../util/helpers';
 import { DEFAULT_SETTINGS } from '../util/settings';
 
-const FieldPath = firebase.firestore!.FieldPath;
-const FieldValue = firebase.firestore!.FieldValue;
-const Timestamp = firebase.firestore!.Timestamp;
+const FieldPath = firebaseExport.FieldPath;
+const FieldValue = firebaseExport.FieldValue;
+const Timestamp = firebaseExport.Timestamp;
+const usesModularApi = firebaseExport.usesFunctionalApi;
```
wilhuff, line 33: Could we make this consistent throughout?

packages/firestore/package.json
```diff
@@ -22,13 +22,13 @@
...
     "gendeps:exp": "../../scripts/exp/extract-deps.sh --types ./exp/index.d.ts --bundle ./dist/exp/index.node.esm2017.js --output ./exp/test/deps/dependencies.json",
-    "pretest:exp": "yarn build:exp",
     "pregendeps:lite": "yarn build:lite",
     "gendeps:lite": "../../scripts/exp/extract-deps.sh --types ./lite/index.d.ts --bundle ./dist/lite/index.node.esm2017.js --output ./lite/test/dependencies.json",
-    "test:exp": "TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'exp/test/**/*.test.ts' --file exp/index.node.ts --config ../../config/mocharc.node.js",
+    "test:exp": "TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/integration/api/*.test.ts' --file exp/index.node.ts --config ../../config/mocharc.node.js",
+    "test:exp:persistence": "USE_MOCK_PERSISTENCE=YES TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/integration/api/*.test.ts' --require ts-node/register --require exp/index.node.ts --require test/util/node_persistence.ts  --config ../../config/mocharc.node.js",
```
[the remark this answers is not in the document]
schmidt-sebastian: I would say you are being nice here :) There are 100% illegible. I tried to create a script this morning, but I failed so I am going to tackle this later.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: I changed the shim to pass the Firestore instance between types, which is likely what we need to to once we productize this code path. This change removes the need to special-case this here.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: I think this meets the bar for "breaking change" since it throws instead of returning a rejected Promise.

packages/firestore/test/integration/api/fields.test.ts
schmidt-sebastian: Oops. Done.

packages/firestore/test/integration/api/fields.test.ts
schmidt-sebastian: I am planning to do so: go/firestore-breaking-changes

packages/firestore/test/integration/api/validation.test.ts
schmidt-sebastian: Uses "test-project" here and elsewhere (via a new TEST_PROJECT export).

packages/firestore/test/integration/api/validation.test.ts
schmidt-sebastian: Done. This should also go away as we productionize the shim.

PR #3378, opened 2020-07-09
Migrate testing to rules-unit-testing

samtstern:

### Discussion

Googlers: [see change rationale here](https://docs.google.com/document/d/15iRiWQ43X1zJkIan3qcjKohygAgRqpCvv8ToC0k4r1A/edit#)

TODO:

- [x] Un-delete the `testing` package (see https://github.com/firebase/firebase-js-sdk/pull/3556)
- [x] We need https://github.com/firebase/firebase-js-sdk/pull/3228 to be submitted and released, otherwise we can't use `firebase-admin` to create mock admin apps without setting `GOOGLE_APPLICATION_CREDENTIALS`.
- [x] Wait for resolution of https://github.com/googleapis/nodejs-firestore/issues/1265 which is causing test failures
- [x] Figure out a meaningful test for `initializeAdminApp()`
- [x] Add change set

### Testing

  * Added unit tests
  * Tested locally with `quickstart-testing` although it was made difficult by the fact that `file:` deps with `peerDependencies` [don't play nice](https://npm.community/t/local-package-fails-to-find-package-installed-in-the-project/6970/7).

### API Changes

  * Rename `@firebase/testing` to `@firebase/rules-unit-testing`


packages/rules-unit-testing/package.json
```diff
@@ -32,8 +33,11 @@
     "rollup": "2.7.6",
     "rollup-plugin-typescript2": "0.27.0"
   },
+  "peerDependencies": {
+    "firebase-admin": "^8.13.0"
```
samtstern, line 37: This will need to be updated when https://github.com/firebase/firebase-js-sdk/pull/3228 is released.

packages/rules-unit-testing/src/api/index.ts
```diff
@@ -261,9 +286,23 @@ export function clearFirestoreData(
...
   return pr.then(
-    v =>
-      Promise.reject(new Error('Expected request to fail, but it succeeded.')),
-    err => err
+    (v: any) => {
+      return Promise.reject(
+        new Error('Expected request to fail, but it succeeded.')
+      );
+    },
+    (err: any) => {
```
samtstern, line 294: We now make sure the error is a `FirebaseError` with a `message` that contains `PERMISSION_DENIED`.  This prevents connect failures from giving the illusion of security.

yuchenshi:
One question: How would you update the old package once this PR gets merged? For example, if we'd like to add a deprecation notice or have the methods print out a warning, how would we track and review those changes?

samtstern:
@yuchenshi I believe that deprecation message is entirely managed by running `npm deprecate` on `@firebase/testing` ... it's not something we do here on GitHub (correct me if I am wrong @Feiyang1)

yuchenshi:
> @yuchenshi I believe that deprecation message is entirely managed by running `npm deprecate` on `@firebase/testing` ... it's not something we do here on GitHub (correct me if I am wrong @Feiyang1)

What if we want to print a warning message when developers call the `initializeAdminApp()` method to notify them of the caveats and the upcoming change?

Also, although unlikely, what if we need to apply security patches? e.g. updating a dependency due to an open vulnerability report?

samtstern:
@yuchenshi I would be fine to do a copy instead of a fork, I just want to know if @Feiyang1 has any problem with "dead" code sitting around in `packages/testing`

samtstern:
@Feiyang1 can you weigh in on the copy vs fork thing?  Would it be alright if we had `@firebase/testing` and `@firebase/rules-unit-testing` side by side for at least one release?

Feiyang1:
I'm fine with that. We can remove `@firebase/testing` in the repo when we feel comfortable.

.github/CODEOWNERS
```diff
@@ -59,7 +59,7 @@ packages/auth  @bojeil-google @avolkovi @samhorlbeck @scottcrossen @firebase/jss
 packages/auth-types  @bojeil-google @avolkovi @samhorlbeck @scottcrossen @firebase/jssdk-global-approvers
 
 # Testing Code
-packages/testing  @avolkovi @samhorlbeck @scottcrossen @yuchenshi @firebase/jssdk-global-approvers
```
samtstern, line 62: TODO: Restore

packages/rules-unit-testing/test/database.test.ts
```diff
@@ -42,18 +42,29 @@ describe('Testing Module Tests', function () {
       .catch(() => {});
   });
 
-  it('assertFails() iff failure', async function () {
+  it('assertFails() iff PERMISSION_DENIED', async function () {
```
samtstern, line 45: @yuchenshi do you have a suggestion for what a meaningful test of `initializeAdminApp` would look like?  The current one checks the internal token provider for `owner` but that's no longer relevant.

yuchenshi: We should just call one of the admin-only operations (list collections?) and assert that succeeds.

samtstern:
@Feiyang1 after you approve this and https://github.com/firebase/firebase-js-sdk/pull/3556 I wil merge it into here and then merge this when the tests pass.

packages/rules-unit-testing/package.json
```diff
@@ -1,16 +1,18 @@
 {
-  "name": "@firebase/testing",
-  "version": "0.20.9",
+  "name": "@firebase/rules-unit-testing",
+  "version": "0.30.0",
```
Feiyang1, line 3: Should the new package start at `0.1.0`? Also the changeset will auto bump the minor version during the release, so you want to use a version number that is one minor version smaller.

packages/rules-unit-testing/package.json
```diff
@@ -25,13 +27,19 @@
     "request": "2.88.2"
   },
   "devDependencies": {
+    "@google-cloud/firestore": "^4.2.0",
```
Feiyang1, line 30: Please fix the version for devDependencies.

packages/rules-unit-testing/package.json
yuchenshi: Or should it start at `1.x` to signify that we're breaking old APIs AND stabilizing into new APIs? Just throwing ideas out there

samtstern: I'm cool with `1.x` ... so does that mean I should make it `0.1.0` and then do a major changeset?

samtstern: Going with 1.0

Feiyang1: Yes

packages/testing/CHANGELOG.md
```diff
@@ -1,12 +1,5 @@
 # @firebase/testing
 
-## 0.20.10
```
Feiyang1, line 3: This file and `packages/testing/package.json` should be reverted.

PR #3380, opened 2020-07-09
Cleanup demo code & get it working with MFA flows

avolkovi:
- lint the demo
- fix some error handling code
- fix verifyPhoneNumber when called with MFA context
- add to/fromJSON to EmailAuthCredential


packages-exp/auth-exp/src/core/errors.ts
```diff
@@ -54,6 +55,7 @@ export const enum AuthErrorCode {
   INVALID_EMAIL = 'invalid-email',
   INVALID_IDP_RESPONSE = 'invalid-credential',
   INVALID_MESSAGE_PAYLOAD = 'invalid-message-payload',
+  INVALID_MFA_SESSION = 'invalid-multi-factor-session',
```
avolkovi, line 58: verified against public docs: https://cloud.google.com/identity-platform/docs/error-codes

packages-exp/auth-exp/index.ts
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is the file that people using Browsers will actually import. You should
+ * only include this file if you have something specific about your
+ * implementation that mandates having a separate entrypoint. Otherwise you can
+ * just use index.ts
```
samhorlbeck, line 22: ? this is itself `packages-exp/auth-exp/index.ts`

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -125,7 +125,8 @@ export async function _performFetchWithErrorHandling<V>(
       return response.json();
     } else {
       const json: JsonError = await response.json();
-      const authError = errorMap[json.error.message];
+      const serverErrorCode = json.error.message.split(' : ')[0] as ServerError;
```
samhorlbeck, line 128: consider using a regex to make this potentially-less-brittle
`/\s*:/` should do it

yarn.lock
```diff
@@ -2175,6 +2175,24 @@
   dependencies:
     slash "^3.0.0"
 
+"@rollup/plugin-strip@^1.3.2":
```
samhorlbeck, line 2178: Are the changes to `yarn.lock` necessary?

packages-exp/auth-exp/index.ts
avolkovi: i think i'll remove this for now, i added it because of a linter warning in package.json

yarn.lock
avolkovi: they're specific to our branch, i reverted the yarn.lock when I rebased against master yesterday

packages-exp/auth-exp/src/api/index.ts
avolkovi: looks like the existing SDK does it a bit different... https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/src/rpchandler.js#L2916 they use everything to the right of the colon as the message, overriding whatever is in the error map

PR #3381, opened 2020-07-09
Temporarily remove firebase devDependency from rxfire

hsubox76:
Rxfire has a dependency on firebase in both peerDependencies and devDependencies, and when changesets script sees the peerDependency version requirement (a broad range) is met, it skips updating the devDependency version. This causes problems in our release process. Temporarily removing until we can make a PR to the Changesets project that may help fix this.

### DIFF
--- a/packages/rxfire/package.json
+++ b/packages/rxfire/package.json
@@ -44,7 +44,7 @@
   },
   "devDependencies": {
     "firebase": "7.16.0",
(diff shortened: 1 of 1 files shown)


Feiyang1:
Will be fixed by https://github.com/atlassian/changesets/pull/412

@hsubox76 can you please resolve the conflict?

PR #3382, opened 2020-07-09
Optional Datastore/FirestoreClient

schmidt-sebastian:
This implements a proposal for optional dependencies discussed in the `firebase-js-sdk` chatroom. We currently have:

```
class LiteFirestore {
  datastore: Datastore;
}

class Firestore extends LiteFirestore {
  firestoreClient: FirestoreClient;
}
```

This means that any user of `Firestore` will pull in Datastore and FirestoreClient, even if they were to just use `Timestamp` or (more realistically) the new methods that only read from cache.

To solve this problem, I added a `components.ts` file for both the Lite SDK and the firestore-exp SDK that manages the relationship between Firestore and its components. Each Firestore instance has up to one component, and the Firestore class is responsible for its cleanup (via the `removeMethods`).

To make this worthwhile I also move the `ComponentProvider` initialization into `components.ts`, which has some impact  on `enableIndexedDbPersistence` and `clearIndexdDbPersistence`.

The first commit contains the code changes, the second commit contains the updated dependency file (auto-generated).


schmidt-sebastian:
@rafikhan Do you have time to review this?

schmidt-sebastian:
I updated the PR to use a more generic mechanism for component removal - see 2bac522f2. Note that for now only one component exists so the difference is minimal.

packages/firestore/lite/src/api/components.ts
```diff
@@ -0,0 +1,74 @@
...
+      settings.ssl ?? DEFAULT_SSL,
+      /* forceLongPolling= */ false
+    );
+    const datastorePromise = newConnection(databaseInfo).then(connection => {
+      const serializer = newSerializer(databaseInfo.databaseId);
+      const datastore = newDatastore(firestore._credentials, serializer);
+      datastore.start(connection);
+      return datastore;
+    });
+    datastoreInstances.set(firestore, datastorePromise);
```
rafikhan, line 59: Add appropriate logging during instantiation and removal.

packages/firestore/exp/src/api/database.ts
```diff
@@ -234,64 +107,98 @@ export function enableIndexedDbPersistence(
...
-  return firestoreImpl._clearPersistence((databaseId, persistenceKey) => {
-    return indexedDbClearPersistence(
-      indexedDbStoragePrefix(databaseId, persistenceKey)
+  if (firestoreImpl._initialized && !firestoreImpl._terminated) {
+    throw new FirestoreError(
+      Code.FAILED_PRECONDITION,
+      'Persistence can only be cleared before a Firestore instance is ' +
+        'initialized or after it is terminated.'
     );
+  }
```
rafikhan, line 152: Suggestion: ConsiderPut this into an assert method like `assertIsNotInitialized()`.

packages/firestore/lite/src/api/components.ts
schmidt-sebastian: Done

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: Kept as is for now since this is the only callsite and the only method that I am aware of that has this contract (plus I would have to come up with a good name - assertIsNotInitialized is only 90% correct...)

rafikhan: Okay.

PR #3391, opened 2020-07-10
Add verifyBeforeUpdate to auth-next

avolkovi:
Also add a couple edge cases that I missed first pass around

### DIFF
--- a/packages-exp/auth-exp/src/core/strategies/email.ts
+++ b/packages-exp/auth-exp/src/core/strategies/email.ts
@@ -65,3 +65,28 @@ export async function sendEmailVerification(
     await user.reload();
   }
 }
+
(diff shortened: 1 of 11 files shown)


packages-exp/auth-exp/src/core/strategies/email.ts
```diff
@@ -65,3 +65,28 @@ export async function sendEmailVerification(
     await user.reload();
   }
 }
+
+export async function verifyBeforeUpdateEmail(
```
samhorlbeck, line 69: This should be reexported in `index.ts` right?

avolkovi: good catch! thanks

PR #3393, opened 2020-07-11
Bundles load from secondary tabs

wu-hui:

### Discussion

### Testing

### API Changes

    can discuss it together.


packages/firestore/src/core/sync_engine.ts
```diff
@@ -1100,6 +1100,11 @@ class MultiTabSyncEngineImpl extends SyncEngineImpl {
     }
   }
 
+  async synchronizeWithChangedDocuments(): Promise<void> {
+    const changes = await this.localStore.getNewDocumentChanges();
+    await this.emitNewSnapsAndNotifyLocalStore(changes);
+  }
```
schmidt-sebastian, line 1106: This is the style of method where I would advocate to drop the await keyword and simply return the Promise. See here for the code overhead of using `await` in ES5: https://www.typescriptlang.org/play/?target=1#code/IYZwngdgxgBAZgV2gFwJYHsL3egFAShgG8AoGcmYAd2FWRggFMqYAFAJ3QFtURHcCMALwA+YgF98AbhLigA

Note that __awaiter and __generator are already part of our code base, so we are only looking at the overhead in `foo`.

packages/firestore/src/platform/browser/dom.ts
```diff
@@ -28,3 +28,17 @@ export function getDocument(): Document | null {
   // eslint-disable-next-line no-restricted-globals
   return typeof document !== 'undefined' ? document : null;
 }
+
+/**
+ * An instance of the Platform's 'TextEncoder' implementation.
+ */
+export function newTextEncoder(): TextEncoder {
```
schmidt-sebastian, line 35: This should only be in platform is there is some divergence between platforms. This doesn't seem to be the case?

packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -115,6 +115,18 @@ export function createWebStorageOnlineStateKey(persistenceKey: string): string {
   return `${ONLINE_STATE_KEY_PREFIX}_${persistenceKey}`;
 }
 
+// The WebStorage prefix that plays as a event to indicate the remote documents
+// might have changed due to some actions secondary tabs did.
+// format of the key is:
+//     firestore_remote_documents_changed_<persistence_prefix>
```
schmidt-sebastian, line 121:
```suggestion
//     firestore_remote_documents_changed_<persistenceKey>
```

packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -115,6 +115,18 @@ export function createWebStorageOnlineStateKey(persistenceKey: string): string {
...
+export const REMOTE_DOCUMENTS_CHANGED_KEY_PREFIX =
+  'firestore_remote_documents_changed';
```
schmidt-sebastian, line 123: So the remote documents change when queries get updated and even when mutations are committed, but this even only fires when a bundle is loaded. Should this be a bundle specific notification?

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -711,6 +719,10 @@ export class WebStorageSharedClientState implements SharedClientState {
     this.persistOnlineState(onlineState);
   }
 
+  remoteDocumentsChanged(): void {
```
schmidt-sebastian, line 722: `notifyRemoteDocumentsChanged()`? But also see my comment below: I think this should be bundle specific.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1412,7 +1417,11 @@ export function loadBundle(
   syncEngineImpl.assertSubscribed('loadBundle()');
 
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
-  loadBundleImpl(syncEngineImpl, bundleReader, task);
+  loadBundleImpl(syncEngineImpl, bundleReader, task).then(() => {
+    if (!syncEngineImpl.isPrimaryClient) {
```
schmidt-sebastian, line 1421: Is there a downside to invoking this on all tabs? Seems like we should issue this notification even on the primary tabs.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -173,6 +174,8 @@ export interface SharedClientState {
   setOnlineState(onlineState: OnlineState): void;
 
   writeSequenceNumber(sequenceNumber: ListenSequenceNumber): void;
+
+  remoteDocumentsChanged(): void;
```
schmidt-sebastian, line 178: Please add JSDoc.

packages/firestore/src/local/shared_client_state_syncer.ts
```diff
@@ -49,4 +49,10 @@ export interface SharedClientStateSyncer {
 
   /** Returns the IDs of the clients that are currently active. */
   getActiveClients(): Promise<ClientId[]>;
+
+  /**
+   * Query for changed documents from remote document cache and raise snapshots
```
schmidt-sebastian, line 54:
```suggestion
   * Retrieves newly changed documents from remote document cache and raises snapshots
```

packages/firestore/src/platform/dom.ts
```diff
@@ -41,3 +41,30 @@ export function getDocument(): Document | null {
     return browser.getDocument();
   }
 }
+
+/**
+ * An instance of the Platform's 'TextEncoder' implementation or null
```
schmidt-sebastian, line 46: This never returns null.

packages/firestore/src/core/sync_engine.ts
wu-hui: You are right, this can be called on all tabs.

packages/firestore/src/local/shared_client_state_schema.ts
schmidt-sebastian: Please also update the key prefix. For sake of brevity, you could consider dropping "remote documents" altogether and just make it something like "bundle loaded".

packages/firestore/src/platform/browser/dom.ts
schmidt-sebastian: Sad :/ 

Should we move this to the ./serializer.ts file though? It seems much more fitting.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -77,6 +78,8 @@ export class BundleReader {
     /** The reader to read from underlying binary bundle data source. */
     private reader: ReadableStreamReader<Uint8Array>
   ) {
+    this.textDecoder = newTextDecoder();
+    debugAssert(!!this.textDecoder, 'Cannot create a valid text decoder.');
```
schmidt-sebastian, line 82: This would never fire, would it?

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -131,7 +134,7 @@ export class BundleReader {
       return null;
     }
 
-    const lengthString = this.textDecoder.decode(lengthBuffer);
+    const lengthString = this.textDecoder!.decode(lengthBuffer);
```
schmidt-sebastian, line 137: Drop bang.

packages/firestore/src/util/bundle_reader.ts
```diff
@@ -199,7 +202,7 @@ export class BundleReader {
       }
     }
 
-    const result = this.textDecoder.decode(this.buffer.slice(0, length));
+    const result = this.textDecoder!.decode(this.buffer.slice(0, length));
```
schmidt-sebastian, line 205: Drop bang.

packages/firestore/src/local/shared_client_state_schema.ts
wu-hui: Done.

packages/firestore/src/platform/browser/dom.ts
wu-hui: Hmm, but it is not used in `serializer.ts`. The only place it is used in non-testing code is `util/bundle_reader.ts`, I think it makes slightly more sense to keep it here.

packages/firestore/src/util/bundle_reader.ts
wu-hui: Right, fixed.

packages/firestore/src/util/bundle_reader.ts
wu-hui: Done.

packages/firestore/src/util/bundle_reader.ts
wu-hui: Done.

packages/firestore/src/local/shared_client_state.ts
```diff
@@ -532,6 +540,10 @@ export class WebStorageSharedClientState implements SharedClientState {
 
     this.onlineStateKey = createWebStorageOnlineStateKey(this.persistenceKey);
 
+    this.bundleChangedRemoteDocumentsKey = createRemoteDocumentsLoadFromBundleKey(
```
schmidt-sebastian, line 543: `createRemoteDocumentsLoadFromBundleKey` seems like it needs a name update now as well. `createBundleLoadedKey`?

For sake of brevity, I would also change bundleChangedRemoteDocumentsKey to bundleLoadedKey. I seems shorter and more specific.

packages/firestore/src/platform/browser/dom.ts
schmidt-sebastian: I mean to move it from packages/firestore/src/platform/browser/dom.ts to packages/firestore/src/platform/browser/serializer.ts. It has little to do with DOM, mostly because it exists in Node which doesn't have DOM.

packages/firestore/src/local/shared_client_state.ts
wu-hui: Done.

packages/firestore/src/platform/browser/dom.ts
wu-hui: Ah, sorry. Done.

PR #3395, opened 2020-07-12
Bundle's named query resume.

wu-hui:


packages/firestore/src/local/local_store.ts
```diff
@@ -909,7 +915,8 @@ class LocalStoreImpl implements LocalStore {
                   target,
                   targetId,
                   TargetPurpose.Listen,
-                  txn.currentSequenceNumber
+                  txn.currentSequenceNumber,
+                  readFrom || SnapshotVersion.min()
```
schmidt-sebastian, line 919: What happens when we don't go into this branch? Should we not still update the target data, even if we already have a (potentially outdated) local copy?

While I am here, is "targetData.snapshotVersion" the same as "readTime"? Do we need both?

wu-hui: Added code to update local copy. And removed `|| SnapshotVersion.min()`(that is what you meant right?).

packages/firestore/src/local/local_store.ts
```diff
@@ -239,7 +239,10 @@ export interface LocalStore {
    * Allocating an already allocated `Target` will return the existing `TargetData`
    * for that `Target`.
    */
-  allocateTarget(target: Target): Promise<TargetData>;
+  allocateTarget(
+    target: Target,
+    readFrom?: SnapshotVersion
```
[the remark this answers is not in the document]
wu-hui: Done.

packages/firestore/exp/src/api/database.ts
```diff
@@ -310,3 +310,19 @@ export function loadBundle(
...
+  name: string
+): Promise<firestore.Query | null> {
+  const firestoreImpl = cast(firestore, Firestore);
+  const client = await firestoreImpl._getFirestoreClient();
+  const namedQuery = await client.getNamedQuery(name);
+  if (!namedQuery) {
+    return null;
+  }
+
+  // TODO(wuandy): make this work with exp build.
```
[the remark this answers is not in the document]
schmidt-sebastian: You simply imported the wrong Query:


```
diff --git a/packages/firestore/exp/src/api/database.ts b/packages/firestore/exp/src/api/database.ts
index 52f80f80..e613e8a1 100644
--- a/packages/firestore/exp/src/api/database.ts
+++ b/packages/firestore/exp/src/api/database.ts
@@ -44,13 +44,14 @@ import { cast } from '../../../lite/src/api/util';
 import { Code, FirestoreError } from '../../../src/util/error';
 import { Deferred } from '../../../src/util/promise';
 import { LruParams } from '../../../src/local/lru_garbage_collector';
-import { CACHE_SIZE_UNLIMITED, Query } from '../../../src/api/database';
+import { CACHE_SIZE_UNLIMITED } from '../../../src/api/database';
 import { DatabaseId, DatabaseInfo } from '../../../src/core/database_info';
 import {
   indexedDbStoragePrefix,
   indexedDbClearPersistence
 } from '../../../src/local/indexeddb_persistence';
 import { LoadBundleTask } from '../../../src/api/bundle';
+import {Query} from "../../../lite";
 
 /**
  * The root reference to the Firestore database and the entry point for the
@@ -322,7 +323,5 @@ export async function namedQuery(
     return null;
   }
 
-  // TODO(wuandy): make this work with exp build.
-  return null;
-  // return new Query(namedQuery.query, firestoreImpl, null, namedQuery.readTime);
+  return new Query(firestoreImpl,namedQuery.query,  null, namedQuery.readTime);
 }
```

packages/firestore/src/api/database.ts
```diff
@@ -1925,7 +1936,8 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
   constructor(
     public _query: InternalQuery,
     readonly firestore: Firestore,
-    protected readonly _converter: firestore.FirestoreDataConverter<T> | null
+    protected readonly _converter: firestore.FirestoreDataConverter<T> | null,
+    private _readTime?: SnapshotVersion
```
schmidt-sebastian, line 1940: `_readTime` should only be part of TargetData and not be used in the API surface.

packages/firestore/src/api/database.ts
```diff
@@ -2223,7 +2235,10 @@ export class Query<T = firestore.DocumentData> implements firestore.Query<T> {
     return addQuerySnapshotListener(
       firestoreClient,
       this._query,
-      options,
+      {
+        readFrom: this._readTime,
+        includeMetadataChanges: options.includeMetadataChanges
+      },
```
schmidt-sebastian, line 2241: This shouldn't be needed.

packages/firestore/src/core/bundle.ts
```diff
@@ -180,9 +180,11 @@ export class BundleLoader {
     this.progress.bytesLoaded += element.byteLength;
 
     let documentsLoaded = this.progress.documentsLoaded;
+    let loadedNamedQuery = false;
 
     if (element.payload.namedQuery) {
       this.queries.push(element.payload.namedQuery);
+      loadedNamedQuery = true;
```
schmidt-sebastian, line 187: This task should take very little time. We should weigh the benefits of not raising a progress event for this amount of time over the confusion that might arise if we raise two progress events with the same values.

packages/firestore/src/core/event_manager.ts
```diff
@@ -60,7 +61,10 @@ export class EventManager implements SyncEngineListener {
     this.syncEngine.subscribe(this);
   }
 
-  async listen(listener: QueryListener): Promise<void> {
+  async listen(
+    listener: QueryListener,
+    readFrom?: SnapshotVersion
```
schmidt-sebastian, line 66: This should go away.

packages/firestore/src/core/event_manager.ts
```diff
@@ -196,6 +200,16 @@ export interface ListenOptions {
...
+
+  /**
+   * Tells the backend whether the client already has the query results up to
+   * a point in time, and the backend will only send deltas from that point
+   * on if applicable (for example, backend might already lose track of this
+   * particular query, and has to restart and send everything anyways).
+   *
+   * When not set, backend simply assumes the client does not have anything.
+   */
+  readFrom?: SnapshotVersion | undefined;
```
schmidt-sebastian, line 212: This should go away.

packages/firestore/src/local/local_store.ts
```diff
@@ -890,7 +897,10 @@ class LocalStoreImpl implements LocalStore {
     });
   }
 
-  allocateTarget(target: Target): Promise<TargetData> {
+  allocateTarget(
+    target: Target,
+    readFrom?: SnapshotVersion
```
schmidt-sebastian, line 902: s/readFrom/readTime (if still needed)

packages/firestore/src/local/local_store.ts
```diff
@@ -899,8 +909,16 @@ class LocalStoreImpl implements LocalStore {
           .next((cached: TargetData | null) => {
             if (cached) {
               // This target has been listened to previously, so reuse the
-              // previous targetID.
+              // previous targetID if no `readFrom` is provided.
+              // If `readFrom` is provided, the cached target data needs to be
+              // compared and updated if `readFrom` is newer.
               // TODO(mcg): freshen last accessed date?
+              if (readFrom && cached.snapshotVersion.compareTo(readFrom) < 0) {
```
schmidt-sebastian, line 916: On second thought, it seems unlikely that a bundle has a newer readTime than the target we have in cache. Maybe we can just go back to the old code.

packages/firestore/src/local/target_data.ts
```diff
@@ -69,6 +69,19 @@ export class TargetData {
     readonly resumeToken: ByteString = ByteString.EMPTY_BYTE_STRING
   ) {}
 
+  /** Creates a new target data instance with an updated sequence number. */
```
schmidt-sebastian, line 72: Comment is outdated.

packages/firestore/src/local/target_data.ts
```diff
@@ -69,6 +69,19 @@ export class TargetData {
...
+  withSnapshotVersion(version: SnapshotVersion): TargetData {
```
schmidt-sebastian, line 73: Should this be `withReadTime`?

packages/firestore/src/remote/serializer.ts
```diff
@@ -975,6 +975,11 @@ export function toTarget(
 
   if (targetData.resumeToken.approximateByteSize() > 0) {
     result.resumeToken = toBytes(serializer, targetData.resumeToken);
+  } else if (targetData.snapshotVersion.compareTo(SnapshotVersion.min()) > 0) {
```
schmidt-sebastian, line 978: Can we:
- Always update the read time for each TargetData?
- Always persist it?

packages/firestore/test/integration/api_internal/bundle.test.ts
```diff
@@ -62,6 +63,25 @@ apiDescribe('Bundles', (persistence: boolean) => {
...
     );
+
+    builder.addNamedQuery(
+      'limit',
+      { seconds: 1000, nanos: 9999 },
+      (collectionReference('coll-1')
+        .orderBy('bar', 'desc')
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        .limit(1) as any)._query
+    );
```
schmidt-sebastian, line 74: It would be so nice if these tests lived in "api", which would mean that they get tested via firestore-exp and the minified builds. Can we use pre-encoded bundle and load it from JSON:?

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -369,6 +369,17 @@ export class SpecBuilder {
     return this;
   }
 
+  userListensToNamedQuery(
+    name: string,
+    query: Query,
+    readFrom: TestSnapshotVersion
+  ): this {
+    // Note we do not call this.nextStep() here because we reuse userListens().
```
schmidt-sebastian, line 377: I am not sure this needs to be spec tested, and if anything it seems like it should be two steps (one that loads the query, one that listens to it). To me, restoring a Query by name is separate from query execution. It doesn't need to interact with the rest of the system (SyncEngine, RemoteStore). 


The test could be simple unit tests for the Query class:

```
const query1 = query.where('foo', > 1);
storeInBundle('fooQuery',  'querye.where('foo', > 1)}');
const query2 = namedQuery('fooQuery');
expect(query1).to.equal(query2);
```

packages/firestore/test/unit/util/bundle_data.ts
```diff
@@ -73,14 +78,28 @@ export class TestBundleBuilder {
     });
     return this;
   }
+
   addNamedQuery(
     name: string,
     readTime: api.Timestamp,
-    bundledQuery: BundledQuery
+    query: Query,
+    limitType?: LimitType
```
schmidt-sebastian, line 86: Query already has the LimitType.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1360,8 +1377,11 @@ export interface SpecStep {
   expectedSnapshotsInSyncEvents?: number;
 }
 
-/** [<target-id>, <query-path>] */
-export type SpecUserListen = [TargetId, string | SpecQuery];
+export interface SpecUserListen {
+  targetId: TargetId;
+  query: string | SpecQuery;
+  fromName?: string;
```
schmidt-sebastian, line 1383: This should ideally be removed (or moved to a different step).

packages/firestore/src/core/bundle.ts
wu-hui: Done.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1654,6 +1688,40 @@ function genericLocalStoreTests(
       .finish();
   });
 
+  it('handles saving and loading named queries', async () => {
+    return expectLocalStore()
+      .after(namedQuery('test', query('coll'), 'FIRST', SnapshotVersion.min()))
```
schmidt-sebastian, line 1693: It is not clear what 'test' and 'FIRST' are. Can you make it clear (possible by using the enum constants)?

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -270,13 +292,93 @@ describeSpec('Bundles:', ['no-ios', 'no-android'], () => {
...
+        createTime: 250,
+        updateTime: 500,
+        content: { value: 'b' }
+      },
+      { name: 'bundled-query', readTime: 400, query: query1 }
+    );
+
+    return spec()
+      .loadBundle(bundleString1)
+      .userListens(query1, 400)
```
schmidt-sebastian, line 311: Please add a comment for non-descriptive arguments. I was just writing a comment on how the target here should have a read time when I realized that this is already provided.

packages/firestore/test/unit/specs/bundle_spec.test.ts
```diff
@@ -270,13 +292,93 @@ describeSpec('Bundles:', ['no-ios', 'no-android'], () => {
...
+            added: [doc('collection/a', 500, { value: 'b' })],
+            fromCache: true
+          })
+          // Loads a newer bundle.
+          .loadBundle(bundleString2)
+          .expectEvents(query1, {
+            modified: [doc('collection/a', 550, { value: 'c' })],
+            fromCache: true
+          })
+          .userUnlistens(query1)
```
schmidt-sebastian, line 363: This test would be more interesting if you didn't unlisten here, since you then have to provide an `expectActiveTarget` with a read time for `client(0)`.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -73,7 +73,7 @@ export interface LimboMap {
 
 export interface ActiveTargetSpec {
   queries: SpecQuery[];
-  resumeToken: string;
+  resumeToken: string | TestSnapshotVersion;
```
schmidt-sebastian, line 76: This should probably be split into two distinct fields, since TestSnapshotVersion is not a resumeToken.

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -253,7 +253,7 @@ export class SpecBuilder {
     return this;
   }
 
-  userListens(query: Query, resumeToken?: string): this {
+  userListens(query: Query, resumeFrom?: string | TestSnapshotVersion): this {
```
schmidt-sebastian, line 256: FWIW, the test above would read super cleanly if you made this
```suggestion
  userListens(query: Query, resumeFrom?: { resumeToken?: string; readTime?: TestSnapshotVersion}}): this {
```

packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -1029,7 +1029,7 @@ export class SpecBuilder {
   private addQueryToActiveTargets(
     targetId: number,
     query: Query,
-    resumeToken?: string
+    resumeToken?: string | TestSnapshotVersion
```
schmidt-sebastian, line 1032: Make this `resumeFrom` as well, since TestSnapshotVersion is not a resumeToken.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1360,8 +1367,10 @@ export interface SpecStep {
   expectedSnapshotsInSyncEvents?: number;
 }
 
-/** [<target-id>, <query-path>] */
-export type SpecUserListen = [TargetId, string | SpecQuery];
+export interface SpecUserListen {
```
schmidt-sebastian, line 1370: HOLY SMOKES. THIS IS THE BEST THING I HAVE READ ALL YEAR.

Thanks for cleaning this up.

packages/firestore/test/unit/util/bundle_data.ts
schmidt-sebastian: Would it at all be possible to do the Query translation (flip the order) here? Not sure if feasible, but then we could drop the LimitType and reduce the confusion.

packages/firestore/exp/src/api/database.ts
```diff
@@ -310,3 +311,17 @@ export function loadBundle(
...
   return resultTask;
 }
+
+export async function namedQuery(
+  firestore: firestore.FirebaseFirestore,
+  name: string
+): Promise<firestore.Query | null> {
+  const firestoreImpl = cast(firestore, Firestore);
+  const client = await firestoreImpl._getFirestoreClient();
+  const namedQuery = await client.getNamedQuery(name);
```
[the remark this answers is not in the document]
wu-hui: Yeah, I'd like to defer to my a follow up PR.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Done.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Done.

packages/firestore/test/unit/specs/spec_builder.ts
wu-hui: Separated.

packages/firestore/test/unit/util/bundle_data.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
```diff
@@ -1379,6 +1419,14 @@ export function saveNamedQuery(
...
     'readwrite',
-    transaction => localStoreImpl.bundleCache.saveNamedQuery(transaction, query)
+    transaction => {
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      localStore.allocateTarget(
+        queryToTarget(fromBundledQuery(query.bundledQuery!)),
+        fromVersion(query.readTime!),
+        transaction
+      );
+      return localStoreImpl.bundleCache.saveNamedQuery(transaction, query);
```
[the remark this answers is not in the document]
schmidt-sebastian: They will get collected in IndexedDB, but the local target mapping doesn't get cleared. `this.targetDataByTarget` and `this.targetIdByTarget` seems to stay around. I wonder if this makes it possible for LRU GC to delete from disk even though it is still considered active in LocalStore? Can you verify what the implications would be?

PR #3399, opened 2020-07-13
Remove Storage AuthWrapper layer

hsubox76:
Remove `AuthWrapper` class wrapping Storage service. It no longer seems to be necessary and will simplify the code in preparation for modularization.


packages/storage/src/service.ts
```diff
@@ -17,55 +17,138 @@
...
+  private authProvider_: Provider<FirebaseAuthInternalName>;
+  private appId_: string | null = null;
+
+  private storageRefMaker_: (loc: Location) => Reference;
+  private requestMaker_: requestMaker;
+  private pool_: XhrIoPool;
+  private requestMap_: RequestMap;
+  private deleted_: boolean = false;
+  private maxOperationRetryTime_: number;
+  private maxUploadRetryTime_: number;
```
schmidt-sebastian, line 56: Should we mark these as readonly (where possible)?

packages/storage/src/service.ts
```diff
@@ -17,55 +17,138 @@
...
-    );
     this.app_ = app;
+    this.authProvider_ = authProvider;
+    this.storageRefMaker_ = (loc: Location): Reference => {
+      return new Reference(this, loc);
+    };
+    this.requestMaker_ = makeRequest;
+    this.maxOperationRetryTime_ = constants.DEFAULT_MAX_OPERATION_RETRY_TIME;
+    this.maxUploadRetryTime_ = constants.DEFAULT_MAX_UPLOAD_RETRY_TIME;
+    this.requestMap_ = new RequestMap();
```
schmidt-sebastian, line 72: Could we remove the `RequestMap` class and just make it a Set? It looks like it is only used in this class.

packages/storage/src/service.ts
```diff
@@ -17,55 +17,138 @@
...
   }
 
+  private static extractBucket_(config: FirebaseOptions): string | null {
+    const bucketString = config[constants.CONFIG_STORAGE_BUCKET_KEY] || null;
+    if (bucketString == null) {
+      return null;
+    }
+    const loc: Location = Location.makeFromBucketSpec(bucketString);
+    return loc.bucket;
+  }
```
schmidt-sebastian, line 95:
```suggestion
      const bucketFromOptions =StorageService.extractBucket_(this.app_?.options);
      if (bucketFromOptions != null) {
        this.bucket_ = bucketFromOptions;
    }
    this.internals_ = new ServiceInternals(this);
  }

  private static extractBucket_(config?: FirebaseOptions): Location | null {
    const bucketString = config?[constants.CONFIG_STORAGE_BUCKET_KEY];
    if (bucketString == null) {
      return null;
    }
    return Location.makeFromBucketSpec(bucketString);
  }
```

packages/storage/src/service.ts
```diff
@@ -17,55 +17,138 @@
...
+          } else {
+            return null;
+          }
+        },
+        () => null
+      );
+    } else {
+      return Promise.resolve(null);
+    }
+  }
```
schmidt-sebastian, line 113:
```suggestion
  async getAuthToken(): Promise<string | null> {
    const auth = this.authProvider_.getImmediate({ optional: true });
    if (auth) {
      const token = await; auth.getToken();
          if (token !== null) {
            return response.accessToken;
          } 
    }
    return null;
  }
```

Since we are already rewriting it :) Optional.

packages/storage/src/service.ts
```diff
@@ -17,55 +17,138 @@
...
-      app,
-      authProvider,
-      maker,
-      RequestExports.makeRequest,
-      this,
-      pool
-    );
     this.app_ = app;
+    this.authProvider_ = authProvider;
+    this.storageRefMaker_ = (loc: Location): Reference => {
```
schmidt-sebastian, line 66: I think you can just inline this now and remove storageRefMaker_.

packages/storage/src/service.ts
hsubox76: Sure, done.

packages/storage/src/service.ts
hsubox76: Done.

packages/storage/src/service.ts
hsubox76: Did this, added lines for self-removal on complete and cancellations on clear.

packages/storage/src/service.ts
hsubox76: Thanks, changed.

packages/storage/src/service.ts
hsubox76: Thanks, changed.

packages/storage/src/service.ts
```diff
@@ -42,15 +38,14 @@ import * as errorsExports from './implementation/error';
  */
 export class StorageService {
   private app_: FirebaseApp | null;
```
schmidt-sebastian, line 40: This can be readonly as well.

packages/storage/.eslintrc.js
```diff
@@ -14,6 +33,13 @@ module.exports = {
         varsIgnorePattern: '^_',
         args: 'none'
       }
+    ],
+    'import/no-extraneous-dependencies': [
```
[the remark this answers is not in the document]
hsubox76: We're not, accidental copy-paste artifact. I only put this here because vscode for some reason doesn't process the top-level `packageDir` path correctly, only if it's in the package's local `eslintrc.js` file.

packages/storage/src/service.ts
```diff
@@ -17,55 +17,122 @@
...
-      app,
-      authProvider,
-      maker,
-      RequestExports.makeRequest,
-      this,
-      pool
-    );
     this.app_ = app;
+    this.authProvider_ = authProvider;
+    this.requestMaker_ = makeRequest;
```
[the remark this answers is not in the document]
hsubox76: Yes, makes sense. Just leftover from AuthWrapper.

packages/storage/.eslintrc.js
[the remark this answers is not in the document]
Feiyang1: Do you mean we need it here to make the eslint plugin happy?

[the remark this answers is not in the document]
hsubox76: Yes.

packages/storage/src/service.ts
hsubox76: It seems like it can't because `deleteApp()` sets it to null.

PR #3403, opened 2020-07-14
Add support for collection(coll) and doc(doc)

schmidt-sebastian:
This adds new methods, which are pending review in go/firestore-next-amendment.

The new collection methods allows direct access to subcollections. Instead of `collection(doc(coll, 'a'), 'b')` we now support `collection(coll, 'a/b')`. This matches the behavior of this method when call with `db` (Firestore) already.

The same change is made for `doc`.

### DIFF
--- a/packages/firestore/lite/src/api/reference.ts
+++ b/packages/firestore/lite/src/api/reference.ts
@@ -284,24 +284,42 @@ export function collection(
   collectionPath: string
 ): CollectionReference<firestore.DocumentData>;
 export function collection(
+  reference: firestore.CollectionReference<unknown>,
+  collectionPath: string
+): CollectionReference<firestore.DocumentData>;
+export function collection(
   reference: firestore.DocumentReference,
   collectionPath: string
 ): CollectionReference<firestore.DocumentData>;
 export function collection(
-  parent: firestore.FirebaseFirestore | firestore.DocumentReference<unknown>,
+  parent:
+    | firestore.FirebaseFirestore
+    | firestore.DocumentReference<unknown>
+    | firestore.CollectionReference<unknown>,
   relativePath: string
 ): CollectionReference<firestore.DocumentData> {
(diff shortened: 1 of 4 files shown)


packages/firestore/lite/src/api/reference.ts
```diff
@@ -350,22 +375,34 @@ export function doc<T>(
...
+    const absolutePath = ResourcePath.fromString(
+      `${parent.path}/${relativePath}`
+    );
     validateDocumentPath(absolutePath);
     return new DocumentReference(
-      coll.firestore,
+      parent.firestore,
       new DocumentKey(absolutePath),
-      coll._converter
+      parent instanceof CollectionReference ? parent._converter : null
```
thebrianchen, line 405: Can you add a test to make sure the converter is properly passed/omitted?

packages/firestore/lite/src/api/reference.ts
```diff
@@ -283,25 +283,43 @@ export function collection(
...
   if (parent instanceof Firestore) {
-    validateCollectionPath(path);
-    return new CollectionReference(parent, path, /* converter= */ null);
+    const absolutePath = ResourcePath.fromString(relativePath);
+    validateCollectionPath(absolutePath);
+    return new CollectionReference(parent, absolutePath, /* converter= */ null);
   } else {
-    const doc = cast(parent, DocumentReference);
-    const absolutePath = doc._key.path.child(path);
+    if (
```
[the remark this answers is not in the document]
schmidt-sebastian: This would be hit if someone passes a DocumentReference-looking type into our SDK. Our external types only validate that type passed in looks like a DocumentReference (has all the same properties) but they cannot enforce that the provided type is an actual instance of DocumentReference.

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: Done

PR #3406, opened 2020-07-14
Populate enrolled factors on reload()

avolkovi:


packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -102,6 +106,28 @@ export class UserImpl implements User {
...
+    this.reloadListener = callback;
+    if (this.reloadUserInfo) {
+      this._notifyReloadListener(this.reloadUserInfo);
+      this.reloadUserInfo = null;
+    }
+  }
+
+  _notifyReloadListener(userInfo: APIUserInfo): void {
+    if (this.reloadListener) {
+      this.reloadListener(userInfo);
```
samhorlbeck, line 124: why not set `this.reloadUserInfo = null` here instead of on line 118?

avolkovi: I prefer the check next to the use, that way the logic is check, use, null-out, we don't actually use reloadUserInfo in _notifyReloadListener so it seems a bit off to null it out without using it

PR #3412, opened 2020-07-15
Update exp docgen script and add functions-exp docs

hsubox76:
Converted scripts/exp/docgen.js to TS and ran it to generate functions-exp docs.


Feiyang1:
We need to generate docs for `functions-types-exp` as well since it has types that `functions-exp` references.

PR #3414, opened 2020-07-15
Add Storage Integration tests

schmidt-sebastian:
This adds a barebones integration test file for Storage. Requires anonymous auth in the test project.

### DIFF
--- a/packages/firestore/src/platform/rn/base64.ts
+++ b/packages/firestore/src/platform/rn/base64.ts
@@ -16,6 +16,7 @@
  */
 
 import { base64 } from '@firebase/util';
+import { uint8ArrayFromBinaryString } from '../../util/byte_string';
 
 // WebSafe uses a different URL-encoding safe alphabet that doesn't match
 // the encoding used on the backend.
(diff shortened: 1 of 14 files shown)


packages/storage/src/implementation/error.ts
```diff
@@ -54,7 +54,11 @@ export class FirebaseStorageError implements Error {
   }
 
   get message(): string {
-    return this.message_;
+    if (this.serverResponse_) {
+      return this.message_ + '\n' + this.serverResponse_;
```
schmidt-sebastian, line 58: Added this so that the server response shows up in the error message for "unknown" errors.

packages/firestore/src/platform/rn/base64.ts
```diff
@@ -16,6 +16,7 @@
  */
 
 import { base64 } from '@firebase/util';
+import { uint8ArrayFromBinaryString } from '../../util/byte_string';
```
hsubox76, line 19: I see Github Actions has inlined this ts error from the build log ('uint8ArrayFromBinaryString' is defined but never used) so that's cool.

schmidt-sebastian: Removed this line.

PR #3417, opened 2020-07-16
Move Datastore to ComponentProvider

schmidt-sebastian:
This is the first step in my two step plan to split up the ComponentProvider into one that provides only the components that are needed to reach from cache and into one that provides the RemoteStore and other components needed for connectivity.

### DIFF
--- a/packages/firestore/test/integration/util/internal_helpers.ts
+++ b/packages/firestore/test/integration/util/internal_helpers.ts
@@ -53,9 +53,10 @@ export function withTestDatastore(
   credentialsProvider: CredentialsProvider = new EmptyCredentialsProvider()
 ): Promise<void> {
   const databaseInfo = getDefaultDatabaseInfo();
-  return newConnection(databaseInfo).then(conn => {
+  return newConnection(databaseInfo).then(async conn => {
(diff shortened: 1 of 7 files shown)


packages/firestore/test/integration/util/internal_helpers.ts
```diff
@@ -53,9 +53,10 @@ export function withTestDatastore(
   credentialsProvider: CredentialsProvider = new EmptyCredentialsProvider()
 ): Promise<void> {
   const databaseInfo = getDefaultDatabaseInfo();
-  return newConnection(databaseInfo).then(conn => {
+  return newConnection(databaseInfo).then(async conn => {
     const serializer = newSerializer(databaseInfo.databaseId);
-    const datastore = newDatastore(conn, credentialsProvider, serializer);
+    const datastore = newDatastore(credentialsProvider, serializer);
+    await datastore.start(conn);
```
var-const, line 59: Hmm, how does this work with `await`? `start` doesn't seem to be `async`.

schmidt-sebastian: "await" returns the value that is being awaited even if it is not a Promise, so it should work. That being said, we should still drop it here. Thanks for catching.

PR #3418, opened 2020-07-16
Change Error to FirestoreError

### LINKED ISSUE #1515: Wrong type information for `firebase.storage.UploadTask.on(.., .., error, ..)`
### Describe your environment

  * Operating System version: win10
  * Browser version: Chrome 71.0.3578.98
  * Firebase SDK version: 5.8.0
  * Firebase Product: Storage

### Describe the problem
The `firebase.storage.UploadTask` interface has the method `on( .., .., error?: ((a: Error) => any) | null, ..)`.
Checking the error code Error interface doesn't have the code property.

#### Steps to reproduce:

Using the example provided [here](https://firebase.google.com/docs/storage/web/upload-files#full_example), in the callback function for error a `switch-case` statement is used to switch among the error codes. The problem is that code is not recognized as property.   

```javascript
...
switch (error.code) {
    case 'storage/unauthorized':
      // User doesn't have permission to access the object
      break;
...
````
#### Possible solution
I suppose the solution is to use `FirebaseError` instead of `Error` line 720 of type definition.
https://github.com/firebase/firebase-js-sdk/blob/6b53e0058483c9002d2fe56119f86fc9fb96b56c/packages/storage-types/index.d.ts#L84

Thanks

dconeybe:
Change the signature of `onSnapshot()` and `onSnapshotsInSync()` methods to report `FirestoreError` instead of a generic `Error`. This allows users to extract the error code from the error without having to do a questionable cast.

This will address #1515 for Firestore. A separate PR will be needed for Storage.


dconeybe:
I'm going to merge despite the "Test All Packages" check failing. Its failure is due to a timeout and is not critical to this feature.

PR #3420, opened 2020-07-16
Add auth isolation for oauth flows

samhorlbeck:


packages-exp/auth-exp/src/core/strategies/redirect.ts
```diff
@@ -172,3 +172,7 @@ class RedirectAction extends AbstractPopupRedirectOperation {
 
   cleanUp(): void {}
 }
+
+export function _clearOutcomes(): void {
+  redirectOutcomeMap = {};
```
avolkovi, line 177: Can you call https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/clear instead and keep this is a const?

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -57,8 +61,11 @@ class BrowserPopupRedirectResolver implements PopupRedirectResolver {
     authType: AuthEventType,
     eventId?: string
   ): Promise<AuthPopup> {
+    console.log(this.eventManagers);
```
avolkovi, line 64: stray logs

samhorlbeck: Done

packages-exp/auth-exp/src/core/strategies/redirect.ts
samhorlbeck: Done

PR #3424, opened 2020-07-16
Issue 2393 - Add environment check to Performance Module

XuechunHou:
Issue 2393 fix for Performance module.


packages/performance/index.ts
```diff
@@ -76,9 +83,28 @@ declare module '@firebase/app-types' {
...
       (app?: FirebaseApp): FirebasePerformance;
+      isSupprted(): Promise<boolean>;
     };
   }
   interface FirebaseApp {
     performance?(): FirebasePerformance;
   }
 }
+
+async function isSupported(): Promise<boolean> {
```
alikn, line 94: Performance is a monitoring sdk. It is easier for the users to initialize it in all environments, but performance will not start internally if all the apis it needs are not available.
Because of that, we probably do not need the isSupported function so the user can check before initializing.

hsubox76: Analytics is in a similar situation but it throws if the environment conditions aren't met, on the assumption the developer might want to be alerted if they're not getting analytics data from a certain type of environment. `isSupported()` is a way to prevent initialization and avoid users having to see those errors. Maybe analytics could make some changes in the future?

But for now, since Performance is designed to surface this through `console.info` which won't appear for most users, I guess we don't need a mechanism to suppress it. Also performance internal initialization can be blocked by an async check (unlike analytics) so it seems like all the checks and blocking can be handled without developers or users seeing anything.

XuechunHou: I removed the `isSupported` function

packages/performance/src/services/api_service.ts
```diff
@@ -109,11 +113,30 @@ export class Api {
     );
   }
 
-  requiredApisAvailable(): boolean {
-    if (fetch && Promise && this.navigator && this.navigator.cookieEnabled) {
+  async requiredApisAvailable(): Promise<boolean> {
```
alikn, line 116: The change looks good!
Can you please add a test for this function as it is not trivial anymore?

packages/performance/src/services/api_service.test.ts
```diff
@@ -47,9 +49,45 @@ describe('Firebase Performance > api_service', () => {
...
+      });
+    });
+
+    it('call logger when validateIndexedDBOpenable throws an exception', () => {
+      stub(consoleLogger, 'info');
+      stub(FirebaseUtil, 'isIndexedDBAvailable').returns(true);
+      stub(FirebaseUtil, 'validateIndexedDBOpenable').throws();
+      stub(api.navigator, 'cookieEnabled').value(true);
+      return api.requiredApisAvailable().then(() => {
+        expect(consoleLogger.info).to.to.be.called;
```
alikn, line 87: nit: one "to" can be removed.

packages/performance/src/services/api_service.test.ts
```diff
@@ -47,9 +49,45 @@ describe('Firebase Performance > api_service', () => {
...
+
     setupApi(mockWindow);
     api = Api.getInstance();
   });
+  describe('requiredApisAvailable', () => {
+    it('call logger when navigator is not available', () => {
+      stub(consoleLogger, 'info');
+      stub(api, 'navigator').value(null);
+      return api.requiredApisAvailable().then(() => {
+        expect(consoleLogger.info).to.be.called;
```
alikn, line 61: Can we check for the return value as well?

PR #3434, opened 2020-07-17
First pass at service workers for auth-next

avolkovi:


packages-exp/auth-exp/demo/rollup.config.js
```diff
@@ -37,28 +63,28 @@ const es5Builds = [
...
     ]
+  },
+  {
+    input: 'src/worker/web-worker.ts',
+    output: [{ file: pkg.webworker, format: 'esm', sourcemap: true }],
+    plugins: workerPlugins
+  },
+  {
+    input: 'src/worker/service-worker.ts',
+    output: [{ file: pkg.serviceworker, format: 'esm', sourcemap: true }],
```
Feiyang1, line 86: How does it work? I don't think esm works in worker context.

avolkovi: it works cause we don't have actually have any external dependencies, all the import statements are resolved by the time the output is generated, should I use something else here?

Feiyang1: ah I see. LGTM

PR #3436, opened 2020-07-17
Remove MultiTabLocalStore

schmidt-sebastian:
This PR is part of my effort to create a cache-only entry point into Firestore (for `getDocFromLocalCache()`). My plan is to split up the component providers and extract one that only provides `Persistence`, `SharedClientState` and `LocalStore` . The "WIP end result" is here: https://github.com/firebase/firebase-js-sdk/pull/3425. This PR is a version of the first commit in that PR.

To avoid the combinatorial explosion of `ComponentProvider`s, this PR removes `MultiTabLocalStore` and instead makes all methods in `MultiTabLocalStore` tree-shakeable.

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -202,7 +202,8 @@ export class FirestoreClient {
   enableNetwork(): Promise<void> {
     this.verifyNotTerminated();
     return this.asyncQueue.enqueue(() => {
-      return this.syncEngine.enableNetwork();
+      this.persistence.setNetworkEnabled(true);
+      return this.remoteStore.enableNetwork();
     });
   }
 
@@ -334,7 +335,8 @@ export class FirestoreClient {
   disableNetwork(): Promise<void> {
     this.verifyNotTerminated();
     return this.asyncQueue.enqueue(() => {
-      return this.syncEngine.disableNetwork();
+      this.persistence.setNetworkEnabled(false);
+      return this.remoteStore.disableNetwork();
     });
   }
 
--- a/packages/firestore/src/local/local_store.ts
+++ b/packages/firestore/src/local/local_store.ts
@@ -1054,9 +1054,10 @@ export function lookupMutationDocuments(
   localStore: LocalStore,
   batchId: BatchId
 ): Promise<MaybeDocumentMap | null> {
-  const localStoreImpl = debugCast(localStore, LocalStoreImpl);
-  const mutationQueueImpl = debugCast(
-    localStoreImpl.mutationQueue,
+  const [localStoreImpl, mutationQueueImpl] = debugCast(
+    localStore,
+    LocalStoreImpl,
+    (localStore as LocalStoreImpl).mutationQueue,
     IndexedDbMutationQueue // We only support IndexedDb in multi-tab mode.
   );
   return localStoreImpl.persistence.runTransaction(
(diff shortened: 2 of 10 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -202,7 +202,8 @@ export class FirestoreClient {
   enableNetwork(): Promise<void> {
     this.verifyNotTerminated();
     return this.asyncQueue.enqueue(() => {
-      return this.syncEngine.enableNetwork();
```
schmidt-sebastian, line 205: I removed the FirestoreClient -> SyncEngine -> LocalStore -> Persistence indirection for `enableNetwork` and `disableNetwork` since I didn't want to come up with non-conflicting global names for the new free-standing functions.

packages/firestore/src/local/local_store.ts
```diff
@@ -1057,152 +1048,137 @@ export function newLocalStore(
...
+// PORTING NOTE: Multi-Tab only.
+export function lookupMutationDocuments(
+  localStore: LocalStore,
+  batchId: BatchId
+): Promise<MaybeDocumentMap | null> {
+  const localStoreImpl = debugCast(localStore, LocalStoreImpl);
+  const mutationQueueImpl = debugCast(
+    localStoreImpl.mutationQueue,
+    IndexedDbMutationQueue // We only support IndexedDb in multi-tab mode.
+  );
```
var-const, line 1061: This setup is being done in every reworked method, would it be possible to refactor it out into a helper function returning two values?

packages/firestore/src/local/local_store.ts
```diff
@@ -1057,152 +1048,137 @@ export function newLocalStore(
...
-      .runTransaction('Get new document changes', 'readonly', txn =>
-        this.remoteDocuments.getNewDocumentChanges(
-          txn,
-          this.lastDocumentChangeReadTime
-        )
+/**
+ * Returns the set of documents that have been updated since the last call.
+ * If this is the first call, returns the set of changes since client
+ * initialization. Further invocations will return document changes since
+ * the point of rejection.
```
var-const, line 1135: Can you clarify what is meant by `rejection`?

packages/firestore/src/local/local_store.ts
```diff
@@ -1057,152 +1048,137 @@ export function newLocalStore(
...
+    .then(({ changedDocs, readTime }) => {
+      localStoreImpl.lastDocumentChangeReadTime = readTime;
+      return changedDocs;
+    });
+}
 
-  async synchronizeLastDocumentChangeReadTime(): Promise<void> {
-    this.lastDocumentChangeReadTime = await this.persistence.runTransaction(
+/**
+ * Reads the newest document change from persistence and forwards the internal
```
var-const, line 1160: Nit: can you clarify where/to what it's being forwarded?

packages/firestore/src/local/local_store.ts
schmidt-sebastian: Changed to `will return document that have changed since the prior call`.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: It's not that pretty since it needs a cast in the method call, but I changed it:

```
const [localStoreImpl, mutationQueueImpl] = 
    debugCast(localStore, LocalStoreImpl,  (localStoreImpl as LocalStoreImpl).mutationQueue, IndexedDbMutationQueue);
```

packages/firestore/src/local/local_store.ts
schmidt-sebastian: The "forward" does not mean "send something along" but rather "move forward". Rewritten.

packages/firestore/src/local/local_store.ts
schmidt-sebastian: Hm, I undid this since this pattern actually uses one more line for each invocation.

var-const: Hmm, what I had in mind was just
```typescript
const [localStoreImpl, mutationQueueImpl] = foo(localStore);
```
with the helper function taking care of everything else.

schmidt-sebastian: I can add a `multiTabLocalStore` function that returns LocalStoreImpl, IndexedDbRemoteDocumentCache, IndexedDbMutationQueue and IndexedDbTargetCache. Does that sound good to you?

PR #3437, opened 2020-07-17
Fix imports for Console builds

schmidt-sebastian:
I stumbled across this today: When I removed `makeConstructorPrivate` (https://github.com/firebase/firebase-js-sdk/pull/3309), I didn't change the imports in the `index.console.ts` file. This file is not part of our build and only used in Google3, hence nothing blew up (yet).


schmidt-sebastian:
@wilhuff - Can you rubber stamp this? This is now blocking some of my browser test work :/

PR #3440, opened 2020-07-20
Add new build rule for gendeps script

schmidt-sebastian:
This is required since the Node build now uses UMD and the Browser builds are mangled. As such, we no longer have a nice build to look at to generate the list of dependencies.

Unfortunately, using the `rollup.shared.js` file in the new script meant that I had to convert it into CommonJS.

I also checked in the script/*.js file and removed the prebuilt step (for development speed). Furthermore, I removed the `verify-depedencies.test.ts` tests since I only ever looked at the JSON files itself (these tests are pretty slow).


schmidt-sebastian:
Hm, I need to look at the size increase. Very strange.

schmidt-sebastian:
> Hm, I need to look at the size increase. Very strange.

Fixed

Feiyang1:
Instead of checking in js files, can you use `ts-node` to run the ts files directly to avoid the compilation step? then you also don't need to make `rollup.shared` cjs.

You can create `tsconfig.json` in the `script` folder, setting the target module to `cjs`, then run the scripts using `ts-node-script` which will use the correct `tsconfig.json` file.

schmidt-sebastian:
These scripts are using in transformers, as such I don't think we can use `ts-node`.

PR #3443, opened 2020-07-20
Use the SDK version from Firebase package.json

schmidt-sebastian:
This change is needed since the firebase-exp SDK is not meant to depend on @firebase/app

### DIFF
--- a/packages/firestore/src/core/version.ts
+++ b/packages/firestore/src/core/version.ts
@@ -15,7 +15,5 @@
  * limitations under the License.
  */
 
-import firebase from '@firebase/app';
-
 /** The semver (www.semver.org) version of the SDK. */
-export const SDK_VERSION = firebase.SDK_VERSION;
+export { version as SDK_VERSION } from '../../../firebase/package.json';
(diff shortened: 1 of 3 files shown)


packages/firestore/src/core/version.ts
```diff
@@ -15,7 +15,5 @@
  * limitations under the License.
  */
 
-import firebase from '@firebase/app';
-
 /** The semver (www.semver.org) version of the SDK. */
-export const SDK_VERSION = firebase.SDK_VERSION;
+export { version as SDK_VERSION } from '../../../firebase/package.json';
```
Feiyang1, line 19: You will need to use 'rollup-plugin-json' to resolve it in all build tasks.
You will want to set `preferConst: true` for es2017 builds.

schmidt-sebastian: I am already using it in exp and lite, and we have been using it in the legacy build already. Is there some other target that is missing?

Feiyang1: My mistake. I did see it in `rollup.config.es5.js`, but now I realize es5 builds are based on esm2017 builds which uses it.

PR #3444, opened 2020-07-21
Firestore exp release script

Feiyang1:


packages/firestore/rollup.config.lite.release.js
```diff
@@ -0,0 +1,114 @@
...
+    output: {
+      file: path.resolve('./lite', pkg.main),
+      format: 'umd',
+      name: 'firebase.firestore'
+    },
+    plugins: [alias(generateAliasConfig('node')), ...nodePlugins],
+    external: resolveNodeExterns,
+    treeshake: {
+      moduleSideEffects: false
+    }
```
schmidt-sebastian, line 84: This made no difference in my size testing so I removed it.

packages/firestore/rollup.config.exp.release.js
```diff
@@ -0,0 +1,114 @@
+/**
```
schmidt-sebastian, line 1: What do we need the non-release rollup configs for?  I am getting rid of the  one dependency I have on the non-mangled builds (https://github.com/firebase/firebase-js-sdk/pull/3440). Are there others?

packages/firestore/rollup.config.lite.release.js
Feiyang1: You wouldn't need it normally.
I use it to remove an empty import (`import '@firebase/app-types'`) caused by`importPathTransformer`.

packages/firestore/rollup.config.exp.release.js
Feiyang1: Replaced the non-release config with the release config and removed the release rollup config files.

packages/firestore/exp/package.json
```diff
@@ -1,8 +1,10 @@
 {
   "name": "@firebase/firestore/exp",
   "description": "A tree-shakeable version of the Firestore SDK",
-  "main": "../dist/exp/index.node.umd.js",
+  "main": "../dist/exp/index.node.esm2017.js",
```
schmidt-sebastian, line 4: Should we update these names to match the ones you defined in firestore-exp? Alternatively, can we just drop this?

Feiyang1: I need to revert this. What do you mean `the ones you defined in firestore-exp`?

schmidt-sebastian: We should use the same names in packages-exp/firestore/package.json and packages/firestore/exp/package.json

But it seems like we could just remove ` packages/firestore/exp/package.json`

Feiyang1: There is no `packages-exp/firestore`. Do you mean `packages-exp/firebase-exp/firestore/package.json`?
I need `packages/firestore/exp/package.json` for the location of the binary files for exp, so I can use them to update the fields in `packages/firestore/package.json`.

The exp release process is hacky and might look confusing. Hopefully we just need it for a short period of time.

packages-exp/firebase-exp/firestore/lite/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "firebase-exp/firestore/lite",
+  "main": "dist/index.cjs.js",
+  "browser": "dist/index.esm.js",
+  "module": "dist/index.esm.js",
+  "typings": "dist/firestore/lite/index.d.ts"
```
schmidt-sebastian, line 6: Are these values used at all? Looks like the release script replaces them. Might make sense to use dummy values if that is indeed the case.

scripts/utils.ts
```diff
@@ -59,3 +63,11 @@ export async function getPackageInfo(
 
   return packageInfo;
 }
+
+export async function readPackageJson(packagePath: string) {
+  /**
+   * Nout using require here because require caches the file
```
schmidt-sebastian, line 69: s/Nout/Not

scripts/exp/prepare-firestore-for-exp-release.ts
```diff
@@ -0,0 +1,80 @@
...
+import path from 'path';
+
+const writeFile = promisify(_writeFile);
+const readFile = promisify(_readFile);
+const packagePath = `${projectRoot}/packages/firestore`;
+
+// Prepare @firebase/firestore, so scripts/exp/release.ts can be used to release it
+export async function prepare() {
+  // Update package.json
+  const packageJson = await readPackageJson(packagePath);
```
schmidt-sebastian, line 30: I don't fully follow this. Maybe that's ok, maybe we need some comments.

This seems to create a new package.json based on  `expPackageJson`, but it seems to use the dependencies and the name from the main Firestore package. If that understanding is correct, then I think we are not using the right name for firestore/lite. We should also consider using different dependencies so that we don't force webchannel-wrapper on users that don't use a bundler for their code.

packages-exp/firebase-exp/firestore/lite/package.json
Feiyang1: Yes. They are used when you do `import {...} from 'firebase/firestore/lite'`

scripts/utils.ts
Feiyang1: done.

scripts/exp/prepare-firestore-for-exp-release.ts
Feiyang1: I added some comments. PTAL

Why do you think we are not using the right name for firestore/lite? It will be accessible by `@firebase/firestore/lite` and `firebase/firestore/lite`. Is it incorrect?

How do we force  `webchannel-wrapper` on users that don't use a bundler for their code?
Are you talking about people using CDN scripts? We will build `firebase-firestore-lite.js` in `firebase`, which should not contain webchannel-wrapper.

I believe firebase/exp has the same dependencies as we currently have, what different dependencies should it have?

schmidt-sebastian: "@firebase/firestore/lite" and "firebase/firestore/lite" is correct.

Re: wenchannel-wrapper. It would be nice if users can develop against "firebase/firestore/lite" without depending on WebChannelWrapper. If we continue to ask users to just install the "firebase" package then this point is moot though.

Thanks for the comment update.

Feiyang1: I see. We won't be able to do that unless we put `firestore/lite` into its own package. IIUC, it's only an issue for Nodejs apps, right? E.g. the ones running in serverless environments that has limited storage. I don't think it matters to Web apps since unused deps will just sit in `node_modules`, not in the compiled app code.

packages-exp/firebase-exp/gulpfile.js
```diff
@@ -22,7 +22,10 @@ var sourcemaps = require('gulp-sourcemaps');
 const OUTPUT_FILE = 'firebase.js';
 const pkgJson = require('./package.json');
 const files = [
-  ...pkgJson.components.map(component => `firebase-${component}.js`)
+  ...pkgJson.components.map(component => {
```
hsubox76, line 25: I guess it was there before and it's not very important but [...array] seems unnecessary when there's only one item, it could just be array

packages/firestore/rollup.config.exp.js
```diff
@@ -42,7 +43,8 @@ const nodePlugins = [
       }
     },
     clean: true,
-    transformers: removeAssertTransformer
+    abortOnError: false,
+    transformers: [...removeAssertTransformer, importPathTransformer]
```
hsubox76, line 47: Can this and `removeAssertAndPrefixInternalTransformer` be plural ("transformers") so the spread makes sense?

scripts/exp/prepare-firestore-for-exp-release.ts
schmidt-sebastian: Correct.

PR #3448, opened 2020-07-21
Add Browser tests for Lite and Exp

schmidt-sebastian:


packages/firestore/exp/test/deps/verify_dependencies.test.ts
```diff
@@ -1,50 +0,0 @@
-/**
```
schmidt-sebastian, line 1: This is https://github.com/firebase/firebase-js-sdk/pull/3440. I cannot include this test since it uses `path`

PR #3453, opened 2020-07-22
Foundation for integration tests for auth-exp

samhorlbeck:
Based on https://github.com/firebase/firebase-js-sdk/pull/3414/files

### DIFF
--- a/packages-exp/auth-exp/package.json
+++ b/packages-exp/auth-exp/package.json
@@ -23,7 +23,7 @@
     "test:browser": "karma start --single-run",
     "test:browser:unit": "karma start --single-run --unit",
     "test:browser:integration": "karma start --single-run --integration",
-    "test:browser:debug": "karma start",
+    "test:browser:debug": "karma start --auto-watch",
     "test:node": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha src/**/*.test.* --opts ../../config/mocha.node.opts",
     "type-check": "tsc -p . --noEmit",
     "prepare": "yarn build",
--- a/packages-exp/auth-exp/package.json
+++ b/packages-exp/auth-exp/package.json
@@ -21,6 +21,8 @@
     "dev": "rollup -c -w",
     "test": "yarn type-check && run-p lint test:browser",
     "test:browser": "karma start --single-run",
+    "test:browser:unit": "karma start --single-run --unit",
+    "test:browser:integration": "karma start --single-run --integration",
     "test:browser:debug": "karma start",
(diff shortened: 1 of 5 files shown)


packages-exp/auth-exp/package.json
```diff
@@ -21,6 +21,8 @@
     "dev": "rollup -c -w",
     "test": "yarn type-check && run-p lint test:browser",
     "test:browser": "karma start --single-run",
+    "test:browser:unit": "karma start --single-run --unit",
+    "test:browser:integration": "karma start --single-run --integration",
     "test:browser:debug": "karma start",
```
schmidt-sebastian, line 26: You probably want "--auto-watch" for the debug rule, which performs hot code reloading.

packages-exp/auth-exp/test/helpers/integration/with_test_instance.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { initializeApp } from '@firebase/app-exp';
+import { initializeAuth } from '@firebase/auth-exp/index.browser';
```
schmidt-sebastian, line 19: Is this the import path that customers will use? Just curious - it seems like it might take some time getting used to.

packages-exp/auth-exp/package.json
samhorlbeck: Done

packages-exp/auth-exp/test/helpers/integration/with_test_instance.ts
samhorlbeck: @avolkovi has a better idea about this but I think so :/ The node package is the default index

packages-exp/auth-exp/test/integration/smoke_test.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import { expect } from 'chai';
+
+import { signInAnonymously } from '@firebase/auth-exp/index.browser';
+import { OperationType } from '@firebase/auth-types-exp';
+
+import { describeIntegration } from '../helpers/integration/helpers';
+
+describeIntegration('smoke test', auth => {
```
schmidt-sebastian, line 25: Future you will probably regret doing this. It breaks all UI test runners and certain lint rules (such as prohibiting describe.only).

Can you follow the pattern we use in Firestore? https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/lite/test/integration.test.ts#L145

samhorlbeck: So the issue with that pattern is we need the cleanup code to always run, even if the test fails. So to do that, each `describe` needs an `after`. I was trying to avoid code duplication as much as possible.

I've switched this up to have a bit of a compromise.

PR #3456, opened 2020-07-22
build app-exp for release, so firestore test works

Feiyang1:


package.json
```diff
@@ -24,7 +24,7 @@
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
     "build": "lerna run --scope @firebase/* --scope firebase --scope rxfire build",
     "build:exp": "lerna run --scope @firebase/*-exp --scope firebase-exp build",
-    "build:release": "lerna run --scope @firebase/* --scope firebase --ignore @firebase/*-exp --ignore firebase-exp prepare",
+    "build:release": "lerna run --scope @firebase/app-exp build:deps && lerna run --scope @firebase/* --scope firebase --ignore @firebase/*-exp --ignore firebase-exp prepare",
```
hsubox76, line 27: If it's just one package, why a `lerna run` instead of `yarn --cwd` ?

Feiyang1: Yeah, it's the same. I find using package name a bit easier than using a path.

PR #3458, opened 2020-07-22
Clean up old changelogs

Feiyang1:


packages/messaging/CHANGELOG.md
```diff
@@ -15,5 +15,6 @@
   - @firebase/component@0.1.16
   - @firebase/installations@0.4.14
 
+## 0.6.11
 - [fixed] Fixed an issue introduced in 7.7.0, when FCM switched to provide base64-encoded VAPID
```
hsubox76, line 19: Should this say Firebase 7.7.0 or something to make it clear it's the overall SDK version and not the messaging version? Just realized it looks confusing next to the messaging version numbers.

PR #3459, opened 2020-07-23
Remove duplicate test run

schmidt-sebastian:
The tests with persistence run all tests twice already - once with and once without persistence. Thus, we can skip the setting test run.

### DIFF
--- a/scripts/emulator-testing/firestore-test-runner.ts
+++ b/scripts/emulator-testing/firestore-test-runner.ts
@@ -48,6 +48,8 @@ async function run(): Promise<void> {
   try {
     await emulator.download();
     await emulator.setUp();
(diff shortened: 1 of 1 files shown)


scripts/emulator-testing/firestore-test-runner.ts
```diff
@@ -49,7 +49,6 @@ async function run(): Promise<void> {
     await emulator.download();
     await emulator.setUp();
     await runTest(emulator.port, emulator.projectId, true);
```
Feiyang1, line 51: Can you copy the PR comment here?

schmidt-sebastian: Done

PR #3460, opened 2020-07-23
Extract Offline-only component provider

schmidt-sebastian:
This is the next PR in the infinite series that will eventually be https://github.com/firebase/firebase-js-sdk/pull/3425

It changes the component providers so that it is possible to only instantiate the components that are needed for cache-only queries.

There are now:

- MemoryOfflineComponentProvider
- IndexedDbOfflineComponentProvider
- MultiTabOfflineComponentProvider
- OnlineComponentProvider

The offline providers provide LocalStore, Persistence and related classes. OnlineComponentProvider provides SyncEngine and RemoteStore (among others).

MultiTabOfflineComponentProvider depends on OnlineComponentProvider. That means that in Multi-Tab mode, even a client that only uses `getDocsFromCache` will bundle RemoteStore and SyncEngine. This is probably what we want though, since it may have to listen to queries from secondary tabs.


packages/firestore/package.json
```diff
@@ -18,7 +18,7 @@
     "dev": "rollup -c -w",
     "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "lint:fix": "eslint --fix -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
-    "prettier": "prettier --write '*.js' '{,!(dist)}/**/*.ts'",
+    "prettier": "prettier --write '*.js' '*.ts' '{,!(dist)}/**/*.ts'",
```
var-const, line 21: Question: what's the reason for this addition? Shouldn't the glob on the right cover it?

schmidt-sebastian: I tried to figure this out with https://globster.xyz/, but it seems to slightly disagree with what we have.  I think `'{,!(dist)}/**/*.ts'` doesn't match since it requires at least one level of nesting.

PR #3472, opened 2020-07-24
Util version bump

hsubox76:
I approved https://github.com/firebase/firebase-js-sdk/pull/3165/files without realizing that it required a util bump because util code was changed.

I'm not sure if the bump should be minor instead because it added 3 public util methods.

Hope to cherrypick and re-release.

### DIFF
--- a/.changeset/short-cars-scream.md
+++ b/.changeset/short-cars-scream.md
@@ -1,7 +1,7 @@
 ---
 '@firebase/analytics': patch
 '@firebase/util': minor
-'firebase': minor
+'firebase': patch
 ---
 
 - Fix an error where an analytics PR included a change to `@firebase/util`, but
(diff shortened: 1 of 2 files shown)


.changeset/short-cars-scream.md
```diff
@@ -0,0 +1,8 @@
+---
+'@firebase/analytics': patch
+'@firebase/util': patch
```
Feiyang1, line 3: `util` should be minor. Can you also add what new functions are added to `util`?

hsubox76: Done, I suppose firebase should be minor as well then?

Feiyang1: no, because it doesn't cause any API change in `firebase`

PR #3480, opened 2020-07-24
Make sure the value that is set for a metric is an integer.

visumickey:
Firebase Performance BE just allows the value of counter to be an integer. Currently the counter value could be set as a float and this would fail in processing.

### DIFF
--- a/packages/performance/src/resources/trace.test.ts
+++ b/packages/performance/src/resources/trace.test.ts
@@ -117,6 +117,13 @@ describe('Firebase Performance > trace', () => {
       expect(trace.getMetric('cacheHits')).to.eql(600);
     });
 
+    it('increments the metric value as an integer even if the value is provided in float.', () => {
+      trace.incrementMetric('cacheHits', 200);
+      trace.incrementMetric('cacheHits', 400.38);
+
+      expect(trace.getMetric('cacheHits')).to.eql(600);
+    });
+
     it('throws error if metric doesnt exist and has invalid name', () => {
       expect(() => trace.incrementMetric('_invalidMetric', 1)).to.throw();
     });
--- a/packages/performance/src/resources/trace.test.ts
+++ b/packages/performance/src/resources/trace.test.ts
@@ -117,7 +117,7 @@ describe('Firebase Performance > trace', () => {
       expect(trace.getMetric('cacheHits')).to.eql(600);
     });
 
-    it('increments the metric value as an integer even if the value is provided in float.', () => {
+    it('increments metric value as an integer even if the value is provided in float.', () => {
       trace.incrementMetric('cacheHits', 200);
       trace.incrementMetric('cacheHits', 400.38);
 
--- a/packages/performance/src/resources/trace.test.ts
+++ b/packages/performance/src/resources/trace.test.ts
@@ -129,6 +129,12 @@ describe('Firebase Performance > trace', () => {
       expect(trace.getMetric('cacheHits')).to.eql(200);
     });
 
+    it('sets the metric value as an integer even if the value is provided in float.', () => {
+      trace.putMetric('timelapse', 200.48);
+
+      expect(trace.getMetric('timelapse')).to.eql(200);
+    });
+
     it('replaces metric if it already exists.', () => {
       trace.putMetric('cacheHits', 200);
       trace.putMetric('cacheHits', 400);
--- a/packages/performance/src/resources/trace.test.ts
+++ b/packages/performance/src/resources/trace.test.ts
@@ -124,6 +124,13 @@ describe('Firebase Performance > trace', () => {
       expect(trace.getMetric('cacheHits')).to.eql(600);
     });
 
+    it('increments metric value with a negative float.', () => {
+      trace.incrementMetric('cacheHits', 200);
+      trace.incrementMetric('cacheHits', -230.38);
+
+      expect(trace.getMetric('cacheHits')).to.eql(-31);
+    });
+
     it('throws error if metric doesnt exist and has invalid name', () => {
       expect(() => trace.incrementMetric('_invalidMetric', 1)).to.throw();
     });
--- a/packages/performance/src/resources/trace.ts
+++ b/packages/performance/src/resources/trace.ts
@@ -31,9 +31,11 @@ import {
   isValidCustomAttributeName,
   isValidCustomAttributeValue
 } from '../utils/attributes_utils';
-import { isValidMetricName } from '../utils/metric_utils';
+import {
+  isValidMetricName,
+  convertMetricValueToInteger
+} from '../utils/metric_utils';
 import { PerformanceTrace } from '@firebase/performance-types';
-import { consoleLogger } from '../utils/console_logger';
 
 const enum TraceState {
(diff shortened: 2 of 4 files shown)


packages/performance/src/resources/trace.ts
```diff
@@ -158,13 +158,13 @@ export class Trace implements PerformanceTrace {
...
    * Sets a custom metric to a specified value. Will create a new custom metric if one with the
-   * given name does not exist.
+   * given name does not exist. The value will be floored down to an integer.
    * @param counter Name of the custom metric
    * @param num Set custom metric to this value
    */
   putMetric(counter: string, num: number): void {
     if (isValidMetricName(counter, this.name)) {
-      this.counters[counter] = num;
+      this.counters[counter] = Math.floor(num);
```
alikn, line 167: If we truncate the metric to the integer part, it would be nice to let the developer know. Can we log a message at info level in such a situation?

packages/performance/src/resources/trace.ts
```diff
@@ -158,13 +158,13 @@ export class Trace implements PerformanceTrace {
 
   /**
    * Sets a custom metric to a specified value. Will create a new custom metric if one with the
-   * given name does not exist.
+   * given name does not exist. The value will be floored down to an integer.
    * @param counter Name of the custom metric
    * @param num Set custom metric to this value
    */
   putMetric(counter: string, num: number): void {
```
alikn, line 165: Can we update the function description to mention that only respects the integer part in the type definition please?
https://github.com/firebase/firebase-js-sdk/blob/c1cc428139fa0a934ead91aa955903792d605e23/packages/performance-types/index.d.ts#L74

packages/performance/src/resources/trace.ts
```diff
@@ -158,13 +158,13 @@ export class Trace implements PerformanceTrace {
 
   /**
    * Sets a custom metric to a specified value. Will create a new custom metric if one with the
-   * given name does not exist.
+   * given name does not exist. The value will be floored down to an integer.
```
alikn, line 161: We need to do the same thing for incrementMetric method.

packages/performance/src/resources/trace.ts
visumickey: Good idea, done.

packages/performance/src/resources/trace.ts
visumickey: Done.

packages/performance/src/resources/trace.ts
visumickey: Good idea! Added it.

packages/performance/src/resources/trace.ts
```diff
@@ -145,26 +146,39 @@ export class Trace implements PerformanceTrace {
...
-   * @param num Increment by value
+   * @param numAsInteger Increment by value
    */
-  incrementMetric(counter: string, num = 1): void {
+  incrementMetric(counter: string, numAsInteger = 1): void {
     if (this.counters[counter] === undefined) {
       this.putMetric(counter, 0);
     }
-    this.counters[counter] += num;
+    const valueAsInteger: number = Math.floor(numAsInteger);
```
alikn, line 158: Can we put the shared part in a function at the end of this file?

packages/performance/src/resources/trace.ts
```diff
@@ -145,26 +146,39 @@ export class Trace implements PerformanceTrace {
...
+    if (valueAsInteger < numAsInteger) {
+      consoleLogger.info(
+        `Metric value should be an Integer, setting the value as : ${valueAsInteger}.`
+      );
+    }
+    this.counters[counter] += valueAsInteger;
```
zijianjoy, line 164: Can the method be like the following?

```
incrementMetric(counter: string, numAsInteger = 1): void {
    if (this.counters[counter] === undefined) {
      this.putMetric(counter, numAsInteger);
    } else {
      this.putMetric(counter, this.counters[counter] + numAsInteger);
    }
}
```

packages/performance/src/resources/trace.test.ts
```diff
@@ -117,6 +117,13 @@ describe('Firebase Performance > trace', () => {
       expect(trace.getMetric('cacheHits')).to.eql(600);
     });
 
+    it('increments metric value as an integer even if the value is provided in float.', () => {
+      trace.incrementMetric('cacheHits', 200);
```
zijianjoy, line 121: There are two scenarios I think worth checking:

1. Can the value be negative? If so how does the flooring look like?
2. What if the integer excess number's range limit?

packages/performance/src/resources/trace.ts
visumickey: Good idea. Done.

packages/performance/src/resources/trace.ts
visumickey: For the sake of readability, I would prefer to stick with the older format. Any key benefits with this optimization?

packages/performance/src/resources/trace.test.ts
visumickey: I don't think that would be the scope of this PR. For both of the above, I would prefer to retain the current functionality. Let me know what you think.

visumickey:
> Can we update the function descriptions in d.ts file as well?

Can you reference me to that file?

packages/performance/src/resources/trace.ts
zijianjoy: Main benefit:

**Consistency and Accuracy**: We don't need to remember updating two APIs if the logic here has changed. Plus that consoleLogger message, validation logic, etc. won't be duplicated.

**Size**: Consolidated logic produces smaller SDK size.

**Readability**: A simplified version has better readability. (No need to worry about actual implementation of putMetric())

packages/performance/src/resources/trace.test.ts
zijianjoy: Since we added `Math.floor()` logic, it comes to my question that how will the negative value be changed if we provide negative value, unless we don't allow negative value to be provided.

packages/performance/src/resources/trace.ts
```diff
@@ -262,6 +250,22 @@ export class Trace implements PerformanceTrace {
     }
   }
 
+  /**
+   * Converts the provided value to an integer value to be used in case of a metric.
+   * @param providedValue Provided number value of the metric that needs to be converted to an integer.
+   *
+   * @returns Converted integer number to be set for the metric.
+   */
+  private convertMetricValueToInteger(providedValue: number): number {
```
alikn, line 259: This does not have to be a member of the class and can be added to metrics_utils file.

visumickey: Makes sense Ali. Thanks for the suggestion, done.

packages/performance/src/resources/trace.test.ts
visumickey: Added a test for negative values provided as inputs.

packages/performance/src/resources/trace.ts
visumickey: Thanks James for clarifying, makes sense. Done.

PR #3487, opened 2020-07-27
Add check if crypto.getRandomValues is available

neelance:
On some IE11 the randomBytes helper crashes with "TypeError: getRandomValues is not defined". The reason is not clear, maybe something pollutes self.crypto with a bad polyfill. Nevertheless it makes sense to fall back to Math.random if crypto.getRandomValues is not available.

### DIFF
--- a/packages/firestore/src/platform/browser/random_bytes.ts
+++ b/packages/firestore/src/platform/browser/random_bytes.ts
@@ -30,7 +30,7 @@ export function randomBytes(nBytes: number): Uint8Array {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     typeof self !== 'undefined' && (self.crypto || (self as any)['msCrypto']);
   const bytes = new Uint8Array(nBytes);
-  if (crypto && crypto.getRandomValues) {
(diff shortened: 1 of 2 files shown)


neelance:
@googlebot I signed it!

neelance:
@schmidt-sebastian Mind taking a look? :)

packages/firestore/src/platform/browser/random_bytes.ts
```diff
@@ -30,7 +30,7 @@ export function randomBytes(nBytes: number): Uint8Array {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     typeof self !== 'undefined' && (self.crypto || (self as any)['msCrypto']);
   const bytes = new Uint8Array(nBytes);
-  if (crypto) {
+  if (crypto && crypto.getRandomValues) {
```
schmidt-sebastian, line 33: Do you mind doing this?

```suggestion
  if (crypto && typeof crypto.getRandomValues === 'function') {
```

neelance: Sure, good idea.

PR #3488, opened 2020-07-27
Make loadBundle work with exp build

wu-hui:
Note in `exp/src/api/database.ts` and `shim.ts` the diff is based on some hacks I pushed to `wuandy/Bundles`.


packages/firestore/exp/test/shim.ts
```diff
@@ -168,7 +169,7 @@ export class FirebaseFirestore implements legacy.FirebaseFirestore {
 
   loadBundle(
     bundleData: ArrayBuffer | ReadableStream<Uint8Array> | string
-  ): legacy.LoadBundleTask {
+  ): LoadBundleTask {
```
schmidt-sebastian, line 172: This should return the legacy Task (all APIs return the legacy type), even though it is for now the same.

wu-hui: Done.

PR #3489, opened 2020-07-27
Refactor the way UserCredentials work in anticipation of mudularization

samhorlbeck:
Basically:
Instead of interpreting the result of the IdTokenResponse as soon as it comes in, we delay that until the user directly requests it.

Before:
```ts
const userCred = await signInWithPopup(...);
const accessToken = userCred.credential.accessToken;
```

After:
```ts
const userCred = await signInWithPopup(...);
const credential = GoogleAuthProvider.credentialFromResult(userCred);
const accessToken = credential.accessToken;
```

This saves pulling in at least four different unnecessary classes


packages-exp/auth-types-exp/index.d.ts
```diff
@@ -446,7 +446,7 @@ export interface User extends UserInfo {
  */
 export interface UserCredential {
```
avolkovi, line 447: woof. big api change, no?

samhorlbeck: No this was actually in the plan all along; we had been following the old way instead of the new one. Go double check the doc :stuck_out_tongue:

avolkovi: oh that is excellent

packages-exp/auth-exp/src/core/providers/phone.ts
```diff
@@ -58,11 +61,27 @@ export class PhoneAuthProvider implements externs.PhoneAuthProvider {
...
+      credential._tokenResponse,
+      credential.user.auth.name,
+      AuthErrorCode.ARGUMENT_ERROR
+    );
+    const {
+      phoneNumber,
+      temporaryProof
+    } = credential._tokenResponse as SignInWithPhoneNumberResponse;
+    if (phoneNumber && temporaryProof) {
+      return PhoneAuthCredential._fromTokenResponse(
```
avolkovi, line 78: should we just change the signature here to take the response instead of the two parts?

packages-exp/auth-exp/src/core/strategies/anonymous.ts
```diff
@@ -30,7 +31,8 @@ export async function signInAnonymously(
     // If an anonymous user is already signed in, no need to sign them in again.
     return new UserCredentialImpl(
       auth.currentUser,
-      credential,
+      null,
```
avolkovi, line 34: this signature is hard to read, should we pass a usercredentialparams in instead?

packages-exp/auth-exp/src/core/strategies/email_and_password.test.ts
```diff
@@ -391,16 +392,20 @@ describe('core/strategies/email_and_password/createUserWithEmailAndPassword', ()
 
   it('should sign in the user', async () => {
     const {
-      credential,
+      _tokenResponse,
       user,
       operationType
-    } = await createUserWithEmailAndPassword(
+    } = (await createUserWithEmailAndPassword(
```
avolkovi, line 398: nit: why the extra paren?

packages-exp/auth-exp/src/core/strategies/anonymous.ts
avolkovi: alternatively maybe reorder the params so the optional ones go last?

packages-exp/auth-exp/src/core/providers/phone.ts
samhorlbeck: I don't think so. We have to break into those properties here anyway to know whether or not to fail with argument error

packages-exp/auth-exp/src/core/strategies/email_and_password.test.ts
samhorlbeck: The formatter put it in. I think because it gets cast now

packages-exp/auth-exp/src/core/strategies/anonymous.ts
samhorlbeck: Done.

PR #3491, opened 2020-07-28
Simplify Listen API

schmidt-sebastian:
This is a cleanup PR before making FirestoreClient tree-shakeable.

- It combines listen/unlisten into the same interface in FirestoreClient and inlines some APIs that no longer do anything with that change
- It moves the PartialObserver -> AsyncObserver conversion into one central place

Note that the "unsubscribe" callbacks now enqueue a callback on the async queue even after termination. This is a no-op and so we don't need to check for termination: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/util/async_queue.ts#L312

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -47,6 +47,7 @@ import {
   MemoryOfflineComponentProvider,
   OfflineComponentProvider
 } from './component_provider';
+import { AsyncObserver } from '../util/async_observer';
 
 const LOG_TAG = 'FirestoreClient';
 const MAX_CONCURRENT_LIMBO_RESOLUTIONS = 100;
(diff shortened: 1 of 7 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -391,24 +393,16 @@ export class FirestoreClient {
...
-    // termination to be a no-op.
-    if (this.clientTerminated) {
-      return;
-    }
-    this.asyncQueue.enqueueAndForget(() => {
-      return this.eventMgr.unlisten(listener);
-    });
+    return () => {
+      observer.mute();
+      this.asyncQueue.enqueueAndForget(() => this.eventMgr.unlisten(listener));
```
dconeybe, line 404: @schmidt-sebastian Can you just confirm that it is indeed unnecessary to check for `if (this.clientTerminated)` before calling `this.asyncQueue.enqueueAndForget()`? I'm not familiar enough with `AsyncQueue` to know, but the old code avoids calling `this.asyncQueue.enqueueAndForget()` in the case that `this.clientTerminated` is false. The new code will call it regardless of the value of `this.clientTerminated`.

This comment also applies to `addSnapshotsInSyncListener()` below.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -486,24 +480,17 @@ export class FirestoreClient {
...
-    // Checks for shutdown but does not raise error, allowing remove after
-    // shutdown to be a no-op.
-    if (this.clientTerminated) {
-      return;
-    }
-    this.asyncQueue.enqueueAndForget(() => {
-      this.eventMgr.removeSnapshotsInSyncListener(observer);
-      return Promise.resolve();
-    });
+    this.asyncQueue.enqueueAndForget(async () =>
```
dconeybe, line 485: Why is it necessary to specify an async function here to `this.asyncQueue.enqueueAndForget()`? In the `listen()` method it just specifies a normal, non-async function. I see that this is consistent with the old code but the reason for the inconsistency within this file is non-obvious to me. Should all calls to `enqueueAndForget()` be async functions? It would appear so based on the signature of `enqueueAndForget()`.

schmidt-sebastian: There is a test already: https://github.com/firebase/firebase-js-sdk/blob/ea699fac1220f6d1de3f719c00d96f5039e388ed/packages/firestore/test/integration/api/database.test.ts#L1235

dconeybe: Ahh perfect.

schmidt-sebastian: Sorry - I replied to the wrong comment (but looks like you managed to restore the context). `async` makes the function return a Promise, even if the return of the value being returned is not a Promise. In the case above, "eventManager.listen" already returns a Promise.

This change is not needed, it just felt right to me. Manually returning `Promise.resolve` works as well.

dconeybe: That makes sense. I forgot to check the return type of `eventManager.listen()`. With this explanation, I'd recommend to keep it as is, with the more modern `async` keyword.

PR #3497, opened 2020-07-28
return FirebaseData from initStandalone

Feiyang1:


Feiyang1:
What other callsites? We are changing the the return type of the method. The return type is inferred from the returned object.

schmidt-sebastian:
You could change https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/database/src/core/RepoManager.ts;l=103?q=databaseFromApp&ss=firebase-sdk%2Ffirebase-js-sdk,
which would also affect https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/database/index.ts;l=52?q=databaseFromApp&ss=firebase-sdk%2Ffirebase-js-sdk and https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/database/index.node.ts;l=112?q=databaseFromApp&ss=firebase-sdk%2Ffirebase-js-sdk. These changes are technically no-ops, but feel like the right thing to do. Optional.

PR #3498, opened 2020-07-28
Add a standard OAuthCredential and a GoogleAuthProvider

samhorlbeck:


packages-exp/auth-exp/src/core/credentials/oauth.ts
```diff
@@ -0,0 +1,161 @@
...
+
+      if (params.pendingToken) {
+        cred.pendingToken = params.pendingToken;
+      }
+    } else if (params.oauthToken && params.oauthTokenSecret) {
+      // OAuth 1 and OAuth token with token secret
+      cred.accessToken = params.oauthToken;
+      cred.secret = params.oauthTokenSecret;
+    } else {
+      fail('todo', AuthErrorCode.ARGUMENT_ERROR);
```
samhorlbeck, line 84: To get the appName here we'd have to pipe an auth object through all of these methods which seems like a major bummer

packages-exp/auth-exp/src/core/credentials/oauth.ts
```diff
@@ -0,0 +1,161 @@
...
+const IDP_REQUEST_URI = 'http://localhost';
+
+export interface OAuthCredentialParams {
+  idToken?: string | null;
+  accessToken?: string | null;
+  oauthToken?: string;
+  secret?: string;
+  oauthTokenSecret?: string;
+  nonce?: string;
+  pendingToken?: string;
```
avolkovi, line 40: can you add some comments around which token is used in which case? we hve at least 4 different types of tokens in these params

packages-exp/auth-exp/src/core/credentials/oauth.ts
avolkovi: doesn't seem that bad, it's just about injecting Auth into *Provider.credential()

packages-exp/auth-exp/src/core/credentials/oauth.ts
```diff
@@ -0,0 +1,161 @@
...
+    return signInWithIdp(auth, request);
+  }
+
+  _getReauthenticationResolver(auth: Auth): Promise<IdTokenResponse> {
+    const request = this.makeRequest();
+    request.autoCreate = false;
+    return signInWithIdp(auth, request);
+  }
+
+  private makeRequest(): SignInWithIdpRequest {
```
avolkovi, line 130: nit: buildRequest, make is ambiguous around whether or not the request is performed

packages-exp/auth-exp/src/core/credentials/oauth.ts
```diff
@@ -0,0 +1,161 @@
...
+    const request: SignInWithIdpRequest = {
+      requestUri: IDP_REQUEST_URI,
+      returnSecureToken: true,
+      postBody: null
+    };
+
+    const postBody: { [key: string]: string } = {};
```
avolkovi, line 137: nit: consider creating a type here so you don't have to ['field'] all over the place, alternatively this can be a Record

packages-exp/auth-exp/src/core/credentials/oauth.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  _getReauthenticationResolver(auth: Auth): Promise<IdTokenResponse> {
+    const request = this.makeRequest();
+    request.autoCreate = false;
+    return signInWithIdp(auth, request);
+  }
+
+  private makeRequest(): SignInWithIdpRequest {
+    const request: SignInWithIdpRequest = {
```
avolkovi, line 131: nit: you should just create this and not mutate it, figure out the postBody before you jump into creating the request, then postBody can never be null

packages-exp/auth-exp/src/core/providers/google.ts
```diff
@@ -0,0 +1,75 @@
...
+    error: FirebaseError
+  ): externs.OAuthCredential | null {
+    return GoogleAuthProvider.credentialFromTaggedObject(
+      error as TaggedWithTokenResponse
+    );
+  }
+
+  private static credentialFromTaggedObject({
+    _tokenResponse: tokenResponse
+  }: TaggedWithTokenResponse): externs.OAuthCredential | null {
```
avolkovi, line 61: why pass this object containing a field instead of the field by itself?  you could change this to credentialFromOauthResponse and just pass a PhoneOrOauthTokenResponse, or better yet, scope to just OauthTokenResponse

packages-exp/auth-exp/src/core/credentials/oauth.ts
samhorlbeck: I'm not convinced Provider.credential needs an auth. The credential can be used with any auth.

packages-exp/auth-exp/src/core/providers/google.ts
samhorlbeck: Because the tag could not be on the object after all, and then I'd have to check for that in two places rather than one. Plus, with this, I can pass in an internal userCredential and it just works.

It would be improper to scope the type down as well. The user could call this with any userCredential including one that was passed back from signInWithPhoneNumber.

packages-exp/auth-exp/src/core/credentials/oauth.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/credentials/oauth.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/credentials/oauth.ts
samhorlbeck: Using `Record<>`

PR #3500, opened 2020-07-28
Add the remaining oauth providers

samhorlbeck:


packages-exp/auth-exp/src/core/providers/facebook.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  static credential(accessToken: string): externs.OAuthCredential {
+    return OAuthCredential._fromParams({
+      providerId: FacebookAuthProvider.PROVIDER_ID,
+      signInMethod: FacebookAuthProvider.FACEBOOK_SIGN_IN_METHOD,
+      accessToken
+    });
+  }
+
+  static credentialFromResult(
```
avolkovi, line 39: can these be defined in OAuthProvider?

samhorlbeck: Unfortunately no, since they're all subtly different and TS has no concept of `static abstract` functions

PR #3507, opened 2020-07-29
Refactor some FirestoreClient methods

schmidt-sebastian:
The next PR is going to have a larger diff so this one extracts the change out that I made for two methods. I moved this methods to `firestore_client.ts` to match the other methods of https://github.com/firebase/firebase-js-sdk/pull/3492.

In the follow up PR, I will then drop FirestoreClient from firestore-exp and will directly call the new methods.

### DIFF
--- a/packages/firestore/exp/src/api/reference.ts
+++ b/packages/firestore/exp/src/api/reference.ts
@@ -63,8 +63,7 @@ export function getDoc<T>(
   const ref = cast<DocumentReference<T>>(reference, DocumentReference);
   const firestore = cast<Firestore>(ref.firestore, Firestore);
   return getFirestoreClient(firestore).then(async firestoreClient => {
-    const viewSnapshot = await firestoreClient.getDocumentFromLocalCache(
-      firestoreClient,
+    const viewSnapshot = await firestoreClient.getDocumentViaSnapshotListener(
       ref._key
     );
     return convertToDocSnapshot(firestore, ref, viewSnapshot);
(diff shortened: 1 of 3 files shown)


packages/firestore/exp/src/api/reference.ts
```diff
@@ -65,7 +63,7 @@ export function getDoc<T>(
   const ref = cast<DocumentReference<T>>(reference, DocumentReference);
   const firestore = cast<Firestore>(ref.firestore, Firestore);
   return getFirestoreClient(firestore).then(async firestoreClient => {
-    const viewSnapshot = await getDocViaSnapshotListener(
+    const viewSnapshot = await firestoreClient.getDocumentFromLocalCache(
```
var-const, line 66: Why does this use `getDocumentFromLocalCache` and not `getDocumentViaSnapshotListener` like the other refactorings in this file?

packages/firestore/src/core/firestore_client.ts
```diff
@@ -415,6 +418,20 @@ export class FirestoreClient {
     );
   }
 
+  async getDocumentViaSnapshotListener(
```
var-const, line 421: Optional: consider something like `getSingleDocumentViaSnapshotListener`. As it stands, these function names are really similar to each other.

schmidt-sebastian: This is actually the name that was used previously and matches other SDKs, such as Android. I am leaving as is for consistency.

packages/firestore/exp/src/api/reference.ts
schmidt-sebastian: This is a bug. Thanks for catching.

PR #3511, opened 2020-07-29
Add an interface Database

Feiyang1:
The admin SDK wants to export the type of the instance of the `FirebaseDatabase` class in their namespace, but encountered an issue where Typescript thinks they are exporting the actual class from `@firebase/database-types` and compiles it to code that leads to runtime error:

```
var database_types_1 = require("@firebase/database-types");
```
Adding an interface that's implemented by `FirebaseDatabase` to work around the issue.

I tried to use `InstanceType<typeof FirebaseDatabase>` to get the type of the instance without creating a new type, but it doesn't work because the constructor is private 🤷 .

### DIFF
--- a/.changeset/thirty-flies-flow.md
+++ b/.changeset/thirty-flies-flow.md
@@ -1,0 +1,7 @@
+---
+"@firebase/database-types": patch
+"@firebase/database": patch
+---
+
+Added interface `Database` which is implemented by `FirebaseDatabase` to allow the use of interface instead of class for the instance type.
+It solves an issue admin SDK encountered where they only want to export the type, but Typescript thinks they are exporting the actual class.
--- a/.changeset/thirty-flies-flow.md
+++ b/.changeset/thirty-flies-flow.md
@@ -3,4 +3,4 @@
 "@firebase/database": patch
 ---
 
-Added interface `Database` which is implemented by `FirebaseDatabase. This allows consumer SDKs (such as the Firebase Admin SDK) to export the database types as an interface.
+Added interface `Database` which is implemented by `FirebaseDatabase`. This allows consumer SDKs (such as the Firebase Admin SDK) to export the database types as an interface.
--- a/.changeset/thirty-flies-flow.md
+++ b/.changeset/thirty-flies-flow.md
@@ -3,5 +3,4 @@
 "@firebase/database": patch
 ---
 
-Added interface `Database` which is implemented by `FirebaseDatabase` to allow the use of interface instead of class for the instance type.
-It solves an issue admin SDK encountered where they only want to export the type, but Typescript thinks they are exporting the actual class.
+Added interface `Database` which is implemented by `FirebaseDatabase. This allows consumer SDKs (such as the Firebase Admin SDK) to export the database types as an interface.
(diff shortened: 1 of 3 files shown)


.changeset/thirty-flies-flow.md
```diff
@@ -0,0 +1,7 @@
+---
+"@firebase/database-types": patch
+"@firebase/database": patch
+---
+
+Added interface `Database` which is implemented by `FirebaseDatabase` to allow the use of interface instead of class for the instance type.
+It solves an issue admin SDK encountered where they only want to export the type, but Typescript thinks they are exporting the actual class.
```
schmidt-sebastian, line 7:
```suggestion
Added interface `Database` which is implemented by `FirebaseDatabase. This allows consumer SDKs (such as the Firebase Admin SDK) to export the database types as an interface.
```

packages/database-types/index.d.ts
```diff
@@ -32,9 +32,16 @@ export interface DataSnapshot {
...
-  private constructor();
+export interface Database {
+  app: FirebaseApp;
+  goOffline(): void;
+  goOnline(): void;
+  ref(path?: string | Reference): Reference;
+  refFromURL(url: string): Reference;
+}
 
+export class FirebaseDatabase implements Database {
```
schmidt-sebastian, line 43: Since this is a .d.ts file, isn't "class" already considered an interface? Could we just drop `private constructor`?

Feiyang1: IIUC, we use it to type the actual class that we export and we don't want people to instantiate it, so `private constructor` is important. Please correct me if i'm wrong.

Feiyang1: You can use d.ts file to annotate a JS library. It tells Typescript that the JS library exports a class that people can import and use. People can actually do this and Typescript allows it:

```
import {FirebaseDatabase} from '@firebase/database-types';

new FirebaseDatabase(); // if not having the private constructor
```

For this reason, I consider it's a bad practice to use class instead of interface in our types packages, if you are just going for the instance type.

schmidt-sebastian: If I understand this correctly, it seems like the following two are equivalent:

```
export class Foo {
  private constructor();
  foo: string;
}

export interface Foo {
 foo: string;
}
```

If so, can we drop the existing class declaration? If not, this LGTM.

Feiyang1: They are the same for the instance type, but we need the constructor type which requires to define it as a class. See https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/index.ts#L97

I believe it is to allow people to do instanceof, but disallow newing an object.

schmidt-sebastian: Yeah, the `instanceof` check was the only thing that I could think of, but it seemed a bit unlikely. Better not to break anyone though.

.changeset/thirty-flies-flow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@firebase/database-types": patch
+"@firebase/database": patch
+---
+
+Added interface `Database` which is implemented by `FirebaseDatabase. This allows consumer SDKs (such as the Firebase Admin SDK) to export the database types as an interface.
```
schmidt-sebastian, line 6: Missing ` after FirebaseDatabase.

Feiyang1: Updated.

PR #3513, opened 2020-07-30
Updating the FirebasePerformance Controller and index to support tree-shaking.

jposuna:
This is the first step. The changes to Traces (and other things) still need to be done in a follow-up PR.

### DIFF
--- a/packages-exp/performance-exp/index.ts
+++ b/packages-exp/performance-exp/index.ts
@@ -17,14 +17,13 @@
 
 import '@firebase/installations';
 
-import firebase from '@firebase/app';
-import { FirebaseApp, FirebaseNamespace } from '@firebase/app-types';
+import { FirebaseApp } from '@firebase/app-types';
 import { _FirebaseNamespace } from '@firebase/app-types/private';
-import { Component, ComponentType } from '@firebase/component';
-import { FirebaseInstallations } from '@firebase/installations-types';
-import { FirebasePerformance } from '@firebase/performance-types-exp';
+import {
+  FirebasePerformance,
+  PerformanceSettings
+} from '@firebase/performance-types-exp';
 
-import { name, version } from './package.json';
 import { PerformanceController } from './src/controllers/perf';
 import { setupApi } from './src/services/api_service';
 import { SettingsService } from './src/services/settings_service';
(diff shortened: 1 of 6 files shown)


packages-exp/performance-exp/index.ts
```diff
@@ -17,68 +17,32 @@
...
-
-  instance.registerVersion(name, version);
-}
-
-registerPerformance(firebase);
-
-declare module '@firebase/app-types' {
-  interface FirebaseNamespace {
-    performance?: { (app?: FirebaseApp): FirebasePerformance };
+export function getPerformance(
```
alikn, line 34: Seems like we still need to register the component with the firebaseApp. Here is an example:
https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/functions-exp/src/config.ts

packages-exp/performance-exp/index.ts
```diff
@@ -17,68 +17,32 @@
 
 import '@firebase/installations';
 
-import firebase from '@firebase/app';
-import { FirebaseApp, FirebaseNamespace } from '@firebase/app-types';
+import { FirebaseApp } from '@firebase/app-types';
```
alikn, line 20: app-types-exp

jposuna: Do we want to migrate these now, or once all other packages have moved to exp? So far I only updated the internal references to exp, not for our dependencies.

packages-exp/performance-exp/index.ts
jposuna: We need to have installations available before we can do this, and we need to verify whether we need to actually register performance or not. Will leave out until we have more clarity.

packages-exp/performance-exp/index.ts
jposuna: Since there are some missing packages right now, this may not be viable. Once the rest of our dependencies are in -exp we can make this change.

Feiyang1: You should be able to use `@firebase/app-types-exp`. I think the problem is you are not using the component framework which we will continue to use in vNext.

You should still register Perf with a factory method, like you are doing currently. It would look something like:
```
import { _registerComponent, _getProvider } from '@firebase/app-exp';

_registerComponent(
  new Component(
      'performance',
      container => {
        const app = container.getProvider('app-exp').getImmediate();
        const installations = container
          .getProvider('installations')
          .getImmediate();

        return factoryMethod(app, installations);
      },
      ComponentType.PUBLIC
));

export getPerformance(): Performance {
  const provider: Provider<'performance'> = _getProvider(
    app,
    'performance'
  );
  const perfInstance = provider.getImmediate();
  return perfInstance;
};
```

jposuna: Thanks for the information! I will do this in a follow up PR since I've got quite a few changes built on top of this PR and want to avoid the merge conflict (and since this is not usable yet, there's no danger of things not working).

PR #3517, opened 2020-07-30
First pass at polyfill

avolkovi:
I had to refactor our AuthImpl a lot to get this to work, TLDR we have:

types:
- `externs.Auth` (new public types)
- `compat.FirebaseAuth` (old public types)
- `auth-exp/AuthCore` (at most `externs.Auth | compat.FirebaseAuth`, but in reality it's just pared down to the core fields)
- `auth-exp/Auth extends auth-exp/AuthCore` (`auth-exp/AuthCore` + private methods)

implementations:
- `AuthImplCompat implements auth-exp/Auth`
- `AuthImpl extends AuthImplCompat implements externs.Auth`
- `auth-compat-exp/Auth extends AuthImplCompat implements compat.FirebaseAuth`

notes:
- I changed most usage of `auth-exp/Auth` -> `auth-exp/AuthCore`, since it's now a strict subset of `externs.Auth` we can drop a ton of casts
- Tests should only use `TestAuth` since it implements both `auth-exp/Auth` and `externs.Auth` and doesn't require casting to either
- The few places we have that actually depend on private methods in `auth-exp/Auth` should use the unsafe `_castAuth` and ensure they never call `setPersistence`, `onAuthStateChanged` and `onIdTokenStateChanged` since those methods could have different signatures depending on whether an instance of `AuthImpl` or `auth-compat-exp/Auth` was passed in.
- We need to perform an unsafe cast of `auth-compat-exp/Auth` -> `externs.Auth` before calling from the polyfill to the new implementation, this cast is unsafe since the signature on the 3 above mentioned changes, hence we must avoid ever calling those methods directly
- `AuthImplCompat` is generic so `currentUser` types correctly to the right implementation
- `AuthImplCompat` constructor takes a `UserProvider` so it can instantiate users of the right class

heuristics to follow:
- If you are taking input from a user, use `externs.Auth`
- If you are a private method, use `auth-exp/AuthCore` (true for 99% cases that only need auth for `auth.name`), otherwise use `auth-exp/Auth` if you or a dependency call one of the `_` methods on auth
- If you are a test, use `TestAuth`
- Converting from `externs.Auth` -> `auth-exp/AuthCore` is free
- Converting from `externs.Auth` -> `auth-exp/Auth` is achieved through an unsafe cast by calling `_castAuth()`
- Converting from `auth-compat-exp/Auth` -> `externs.Auth` is achieved through an unsafe cast by calling `this.asExtern()`
- Never ever ever call `setPersistence`, `onAuthStateChanged` or `onIdTokenStateChanged` from our code since you have no idea which implementation you're holding on to, use the equivalent `_` prefixed methods if you need to call them.  Only the user (or integration tests) get to call these directly.


packages-exp/auth-compat-exp/src/user_credential.ts
```diff
@@ -0,0 +1,118 @@
...
+  const { operationType, user } = await credential;
+
+  return {
+    operationType,
+    credential: credentialFromResponse(credential as UserCredential),
+    user: user as User
+  };
+}
+
+export async function convertComfirmationResult(
```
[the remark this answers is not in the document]
avolkovi: opting to keep them in one place, I only ended up need these two and they both deal with userCredential

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -51,16 +52,17 @@ export const DEFAULT_TOKEN_API_HOST = 'securetoken.googleapis.com';
...
+export class AuthImplCompat implements Auth {
   currentUser: User | null = null;
   private operations = Promise.resolve();
-  private persistenceManager?: PersistenceUserManager;
-  private redirectPersistenceManager?: PersistenceUserManager;
-  private authStateSubscription = new Subscription<User>(this);
-  private idTokenSubscription = new Subscription<User>(this);
+  private persistenceManager?: PersistenceUserManager<User>;
+  private redirectPersistenceManager?: PersistenceUserManager<User>;
+  protected authStateSubscription = new Subscription<User>(this);
```
avolkovi, line 60: these can go back to being private

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -35,7 +35,7 @@ function _persistenceKeyName(
   return `${PERSISTENCE_NAMESPACE}:${key}:${apiKey}:${appName}`;
 }
 
-export class PersistenceUserManager {
+export class PersistenceUserManager<T extends User> {
```
avolkovi, line 38: TODO: see if this refactor actually helps or not, if not roll it back

packages-exp/auth-exp/src/core/user/reload.test.ts
```diff
@@ -159,7 +159,7 @@ describe('core/user/reload', () => {
     user.auth.currentUser = user;
 
     const cb = sinon.stub();
-    user.auth.onIdTokenChanged(cb);
+    user.auth._onIdTokenChanged(cb);
```
avolkovi, line 162: this shouldn't be necessary, try rolling back

packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -204,7 +196,7 @@ export class UserImpl implements User {
     assertStringOrUndefined(phoneNumber, auth.name);
     assertStringOrUndefined(photoURL, auth.name);
     assertStringOrUndefined(_redirectEventId, auth.name);
-    const user = new UserImpl({
+    const user = new auth._userProvider({
```
avolkovi, line 199: maybe we can do this by parameterizing AuthImpl instead?

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  async createUserWithEmailAndPassword(
+    email: string,
+    password: string
+  ): Promise<compat.UserCredential> {
+    return convertCredential(
+      impl.createUserWithEmailAndPassword(this._asExtern(), email, password)
+    );
+  }
+  // currentUser: User | null;
```
samhorlbeck, line 104: What's up with this?

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -0,0 +1,269 @@
...
+      this._asExtern(),
+      impl.browserPopupRedirectResolver
+    );
+    if (!credential) {
+      return {
+        credential: null,
+        user: null
+      };
+    }
+    return convertCredential(Promise.resolve(credential!));
```
samhorlbeck, line 122: `!` is unnecessary. The if statement above this guarantees credential is defined

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+  // languageCode: string | null;
+  //       settings: AuthSettings;
+  onAuthStateChanged(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    nextOrObserver: Observer<any> | ((a: compat.User | null) => any),
```
samhorlbeck, line 128: you should probably use `unknown` rather than `any`

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -0,0 +1,269 @@
...
+    actionCodeSettings?: compat.ActionCodeSettings | null
+  ): Promise<void> {
+    return impl.sendPasswordResetEmail(
+      this._asExtern(),
+      email,
+      actionCodeSettings || undefined
+    );
+  }
+  async setPersistence(persistence: string): Promise<void> {
+    function convertPersistence(
```
samhorlbeck, line 173: why does this need to be a separate function?

packages-exp/auth-compat-exp/src/user.ts
```diff
@@ -0,0 +1,155 @@
...
+  getIdTokenResult(forceRefresh?: boolean): Promise<compat.IdTokenResult> {
+    return impl.getIdTokenResult(this, forceRefresh);
+  }
+  getIdToken(forceRefresh?: boolean): Promise<string> {
+    return impl.getIdToken(this, forceRefresh);
+  }
+  linkAndRetrieveDataWithCredential(
+    credential: compat.AuthCredential
+  ): Promise<compat.UserCredential> {
+    return this.linkWithCredential(credential);
```
samhorlbeck, line 46: why is this a method?

packages-exp/auth-compat-exp/src/user_credential.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  const {
+    oauthIdToken,
+    oauthAccessToken,
+    oauthTokenSecret,
+    pendingToken,
+    nonce
+  } = _tokenResponse as SignInWithIdpResponse;
+  switch (providerId) {
```
samhorlbeck, line 51: Why not just have all of these in the switch statements call Provider.credentialFromResult() like you do for phone auth

packages-exp/auth-compat-exp/src/auth.ts
avolkovi: these are the old types https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth-types/index.d.ts#L413 do you still think we should change them?

packages-exp/auth-compat-exp/src/auth.ts
avolkovi: switch statements don't return their value in TS, this allows me to be functional and avoid mutable variables, happy to change if you have strong opinions

packages-exp/auth-compat-exp/src/user.ts
avolkovi: Defined (albeit deprecated) in the old SDK https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth-types/index.d.ts#L27

packages-exp/auth-compat-exp/src/auth.ts
avolkovi: I'll remove this, it's no longer necessary now that I have generic types in AuthImplCompat

packages-exp/auth-compat-exp/src/auth.ts
avolkovi: done

packages-exp/auth-compat-exp/src/user_credential.ts
avolkovi: good idea, I guess we didn't have these before and I was just copying the old implementation... we should add one to OAuthProvider for the last bit here

packages-exp/auth-compat-exp/src/auth.ts
samhorlbeck: yes. `uknown` is equivalent to `any` but will through a compiler error if you attempt to use the value

packages-exp/auth-compat-exp/src/auth.ts
samhorlbeck: ah that's fine

packages-exp/auth-compat-exp/src/auth.ts
avolkovi: done

PR #3532, opened 2020-07-31
Fix bug in terminate()

schmidt-sebastian:
Found this in https://github.com/firebase/firebase-js-sdk/blob/8a8e60ada7f012b198c2c26089daff85c8270ee2/packages/firestore/src/core/firestore_client.ts#L375


schmidt-sebastian:
This is for firestore-exp

Issue #3495, opened 2020-07-28
Firestore breaks on mobile safari if network is lost while app is in the background.


aidanas:
### [REQUIRED] Describe your environment

  * Operating System version: `iOS 13.6`
  * Browser version: Mobile Safari. (Agent string `"Mozilla/5.0 (iPad; CPU OS 13_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"`)
  * Firebase SDK version: `7.16.0` (also present on `7.17.1` )
  * Firebase Product: `firestore`

### [REQUIRED] Describe the problem
We have a multi page Cordova app which uses Firestore with persistence enabled. It issue happens when users navigate about a few pages and then put the app in the background and toggle their wifi on and off. Then when users bring the app to the foreground firebase finds itself in some inconsistent state and cannot be used until users reload the page.

This could potentially be related to :
https://github.com/firebase/firebase-js-sdk/issues/2710

And this
https://github.com/firebase/firebase-js-sdk/issues/2755
Even though all PRs in @schmidt-sebastian comment in the link above have been merged we're still experiencing the issue.

#### Steps to reproduce:

With our production app we can consistently recreate it with the following steps:

1. Open the app and navigate around a few screens.
   This seems to be important as we could not recreate it while being on the same screen. However it is not important which few screens we navigate to as we could reproduce the same error navigating between any two screens. I will add that that before leaving each of the screens (before runtime is destroyed) we destroy firebase app to release resources and then reinitialise it on the following screen.
2. Put app to background
3. Go to settings and toggle wifi off back on
4. Resume the app
5. Observe firestore errors in console and none of firebase features work.

Debug enabled log output shows starts at the time just before app is put to background, then toggling wifi on and off in the settings and ends after the app has been resumed:
```
[Log] [2020-07-28T08:42:31.200Z]  @firebase/firestore: – "Firestore (7.16.0): IndexedDbPersistence" – "Client '5NhcKwxvsOrbLAAu0fIU' is not zombied in LocalStorage" (cordova.js, line 1540)
[Log] [2020-07-28T08:42:31.204Z]  @firebase/firestore: – "Firestore (7.16.0): SimpleDb" – "PUT" – "owner" – "owner" – "{\"ownerId\":\"5NhcKwxvsOrbLAAu0fIU\",\"allowTabSynchronization\":true,\"leaseTimestampMs\":1595925751203}" (cordova.js, line 1540)
[...]

PR #3535, opened 2020-07-31
Re-open IndexedDB if closed

schmidt-sebastian:
This PR addresses an IndexedDb error that may occur when a transaction is first opened. The error `The database connection is closing` can be reproduced locally by invoking SimpleDb.close() and then calling `db.transaction` and this PR fixes at least this case by extending the IndexedDB retry code to include the code around `db.transaction` and by closing and re-opening the IndexedDB instance.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3495

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -20,7 +20,6 @@ import { debugAssert } from '../util/assert';
 import { Code, FirestoreError } from '../util/error';
 import { logDebug, logError } from '../util/log';
 import { Deferred } from '../util/promise';
-import { SCHEMA_VERSION } from './indexeddb_schema';
 import { PersistencePromise } from './persistence_promise';
 
 // References to `window` are guarded by SimpleDb.isAvailable()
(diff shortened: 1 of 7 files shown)


packages/firestore/src/local/simple_db.ts
```diff
@@ -249,12 +186,91 @@ export class SimpleDb {
...
+        const request = indexedDB.open(this.name, this.version);
+
+        request.onsuccess = (event: Event) => {
+          const db = (event.target as IDBOpenDBRequest).result;
+          resolve(db);
+        };
+
+        request.onblocked = () => {
+          reject(
+            new IndexedDbTransactionError(
```
schmidt-sebastian, line 210: This is mostly a copy of the code above, but this line changed from FirestoreError to IndexedDbTransactionError.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1261,3 +1215,19 @@ describe('IndexedDb: allowTabSynchronization', () => {
...
+    console.warn('No IndexedDB. Skipping tests.');
+    return;
+  }
+
+  it('can re-open after close', async () => {
+    return withDb(2, async db => {
+      db.close();
+      await db.runTransaction('readwrite', V1_STORES, () =>
+        PersistencePromise.resolve()
+      );
```
wilhuff, line 1230: This test doesn't make an assertion, which is usually a sign that something's missing. If this would implicitly fail somehow it's worth adding a comment about that or perhaps adding a does not throw expectation.

schmidt-sebastian: I am not sure how `expect().doesNotThrow()` interacts with Promises, so I added a comment instead.

Issue #3514, opened 2020-07-30
Usage instructions are inconsistent/conflicting/confusing


Download:
In [one place](https://www.npmjs.com/package/@firebase/firestore#usage) it says:

```js
import firebase from '@firebase/app';
import '@firebase/firestore'
```
Note the use of the namespace `@firebase`.

In [another place](https://www.npmjs.com/package/firebase#get-the-code-nodejs---server-and-command-line) it says:

```js
// This import loads the firebase namespace.
import firebase from 'firebase/app';
 
// These imports load individual services into the firebase namespace.
import 'firebase/auth';
import 'firebase/database';
```

Note how this basically all uses the same package `firebase`, without a namespace.

I think the first form, with `@firebase` is the new/preferred form or not?

Looking at the first example, it uses [@firebase/app](https://www.npmjs.com/package/@firebase/app). If we open that up, we see that it points us to [firebase](https://www.npmjs.com/package/firebase)

> This package is not intended for direct usage, and should only be used via the officially supported firebase package.

...and then we come to the second example. This is confusing. Which form should I use on my new project?

All packages point to this Github Repo, which is why I am making this issue here.

### EDIT
I read more docs and now I'm more confused... 

It seems the [docs on GCP](https://firebase.google.com/docs/web/setup#using-module-bundlers) use yet another form:

```js
// Firebase App (the core Firebase SDK) is always required and must be listed first
import * as firebase from "firebase/app";

// If you enabled Analytics in your project, add the Firebase SDK for Analytics
import "firebase/analytics";

// Add the Firebase products that you want to use
import "firebase/auth";
import "firebase/firestore";
```

I don't see the namespace used here. So in that sense it looks like the second example. But this one uses `import * as firebase from` instead of `import firebase from`.

I know many big projects (e.g. Babel, Webpack) have been moving over to using namespaces/scopes. So they have been introducing the `@` symbol in their package names and have been [recommending their users](https://babeljs.io/docs/en/v7-migration#scoped-packages) to use the new form. Is the same true for Firebase? Is the first form actually the recommended form? And so are the second example and the docs on GCP outdated? Or is the opposite true and is the example on the Firestore package the one that is outdated? And if we need to use the non-scoped package, do we use it like `import firebase from` or like `import * as firebase from`? Please advice.

Download:
Thanks Google OSS Bot, but when I spend time creating an issue, I sorta expect a human to look at it...

hsubox76:
[...]

PR #3551, opened 2020-08-03
Update READMEs

hsubox76:
Since READMEs show up on the npm website, we should make sure they are updated and consistent. Some example import patterns were inconsistent (e.g. `import firebase from 'firebase/app'` vs` import * as firebase from 'firebase/app'`).

- Removed installation and usage instructions from package READMEs since they are not intended to be installed independently, and have a link to the main Firebase package, which has correct instructions.
- Removed installation instructions from internal packages (util, component, etc.)
- Update all import examples to`import * as firebase` pattern.
- Add a note to all-in-one import instructions to look at section below for importing individual packages.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3514

### DIFF
--- a/packages/component/README.md
+++ b/packages/component/README.md
@@ -2,3 +2,11 @@
 
 _NOTE: This is specifically tailored for Firebase JS SDK usage, if you are not a
 member of the Firebase team, please avoid using this package_
+
+## Usage
+
+**ES Modules**
+
+```javascript
+import { Component } from '@firebase/component';
+```
--- a/packages/component/README.md
+++ b/packages/component/README.md
@@ -2,19 +2,3 @@
 
 _NOTE: This is specifically tailored for Firebase JS SDK usage, if you are not a
 member of the Firebase team, please avoid using this package_
-
-## Installation
-
-You can install this wrapper by running the following in your project:
-
-```bash
-$ npm install @firebase/component
-```
-
-## Usage
-
-**ES Modules**
-
-```javascript
-import { Component } from '@firebase/component';
-```
--- a/packages/template/README.md
+++ b/packages/template/README.md
@@ -7,7 +7,4 @@ Firebase JS SDK. It will give you a couple things OOTB:
   with the rest of the SDK.
 - **Isomorphic Testing/Coverage:** Your tests will be run in both Node.js and
   Browser environments and coverage from both, collected.
-- **Links to all of the other packages:** Should your new package need to take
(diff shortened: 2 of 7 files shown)


packages/component/README.md
```diff
@@ -2,19 +2,3 @@
 
 _NOTE: This is specifically tailored for Firebase JS SDK usage, if you are not a
 member of the Firebase team, please avoid using this package_
-
-## Installation
```
Feiyang1, line 6: I think it's okay to keep this section for internal packages in order to provide instruction for Googlers. We have clearly stated that this package is intended to be only used by the Firebase team, so I don't think there is any confusion.

hsubox76: I restored Usage but kept Installation removed because I don't think any internal user would (or should) npm install it?

packages/util/README.md
```diff
@@ -4,29 +4,3 @@ _NOTE: This is specifically tailored for Firebase JS SDK usage, if you are not a
...
-
-## Installation
-
-You can install this wrapper by running the following in your project:
-
-```bash
-$ npm install @firebase/util
-```
-
-## Usage
```
Feiyang1, line 16: restore the usage?

packages/template/README.md
```diff
@@ -7,7 +7,4 @@ Firebase JS SDK. It will give you a couple things OOTB:
   with the rest of the SDK.
 - **Isomorphic Testing/Coverage:** Your tests will be run in both Node.js and
   Browser environments and coverage from both, collected.
-- **Links to all of the other packages:** Should your new package need to take
```
Feiyang1, line 10: This seems useful and should be kept.

packages/util/README.md
hsubox76: Restored.

packages/template/README.md
hsubox76: It's not true, though, they're not included in `package.json` and they probably shouldn't be because the versions would have to be constantly updated?

It does have app and app-types as peerDependencies but that's all. I looked at the original commit where this line was first written (https://github.com/firebase/firebase-js-sdk/commit/9fcdd378fdd9253248bb385ce87161de64b54592) and it doesn't even include `@firebase/util` then, or maybe I'm confused about what he's saying.

Feiyang1: Thanks for digging into it. I'm fine as is.

PR #3555, opened 2020-08-03
Browser Extension Check for Analytics Module

XuechunHou:
added browser extension check to analytics module; prevent initialization if not in a browser context.

### DIFF
--- a/.changeset/spicy-masks-sort.md
+++ b/.changeset/spicy-masks-sort.md
@@ -1,5 +1,5 @@
 ---
-"@firebase/analytics": minor
+"@firebase/analytics": patch 
 ---
 
 Browser Extension Check for Analytics Module
--- a/.changeset/spicy-masks-sort.md
+++ b/.changeset/spicy-masks-sort.md
@@ -1,5 +1,6 @@
 ---
+"@firebase/firebase": patch
 "@firebase/analytics": patch 
 ---
 
-Browser Extension Check for Analytics Module
+Added Browser Extension check for Firebase Analytics. analytics.isSupported() will now return Promise<false> for extension environments. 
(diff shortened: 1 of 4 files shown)


packages/analytics/src/errors.ts
```diff
@@ -50,7 +51,9 @@ const ERRORS: ErrorMap<AnalyticsError> = {
     'Wrap initialization of analytics in analytics.isSupported() ' +
     'to prevent initialization in unsupported environments',
   [AnalyticsError.COOKIES_NOT_ENABLED]:
-    'Cookies are not enabled in this browser environment. Analytics requires cookies to be enabled.'
+    'Cookies are not enabled in this browser environment. Analytics requires cookies to be enabled.',
+  [AnalyticsError.INVALID_ANALYTICS_CONTEXT]:
+    'Analytics module is not supported in browser extensions environemnt.'
```
hsubox76, line 56: How about `Firebase Analytics is not supported in browser extensions.`

.changeset/spicy-masks-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/analytics": minor
```
hsubox76, line 2: I think this is a patch change since we are not changing the API surface.

.changeset/spicy-masks-sort.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/analytics": patch 
+---
+
+Browser Extension Check for Analytics Module
```
Feiyang1, line 5: Can you make it a bit more specific? Something like: `analytics.isSupport() will now return false for extension environments`

Feiyang1: Can you add `firebase: patch` too, so we will generate changelog for `firebase`?

.changeset/spicy-masks-sort.md
```diff
@@ -0,0 +1,6 @@
+---
+"@firebase/firebase": patch
+"@firebase/analytics": patch 
+---
+
+Added Browser Extension check for Firebase Analytics. analytics.isSupported() will now return Promise<false> for extension environments.
```
hsubox76, line 6: Nit: backticks around `analytics.isSupported()` and `Promise<false>` will make it look nicer in markdown.

PR #3557, opened 2020-08-04
Add NOT_IN and != queries (not public)

thebrianchen:


packages/firestore/src/core/query.ts
```diff
@@ -643,8 +645,40 @@ export class FieldFilter extends Filter {
     }
   }
 
+  private static createKeyFieldInFilter(
+    field: FieldPath,
+    op: Operator,
+    value: api.Value
+  ): FieldFilter {
+    debugAssert(
```
wilhuff, line 653: You can avoid this assertion by making this a part of the type of the parameter. I believe you can declare `op` like this:

```ts
op: Operator.IN | Operator.NOT_IN
```

and then the caller will have to be in a context where this is true.

packages/firestore/src/core/query.ts
```diff
@@ -762,6 +799,27 @@ export class KeyFieldInFilter extends FieldFilter {
...
+      debugAssert(
+        isReferenceValue(v),
+        'Comparing on key with NOT_IN, but an array value was not a ReferenceValue'
+      );
+      return DocumentKey.fromName(v.referenceValue);
+    });
+  }
+
+  matches(doc: Document): boolean {
+    return !this.keys.some(key => key.isEqual(doc.key));
```
wilhuff, line 819: The difference between these types is the operator name and this exclamation point. It seems like you could either:

  * extract a common class for this that takes the operator as a parameter;
  * extract a helper function for creating the keys array;
  * make `KeyFiieldInFilter` the base class and have it take the operator as a parameter;
  * move the construction of the keys array into the caller (passing it as an extra argument) since you now have a dedicated `createKeyFieldInFilter`.

(listed in order of my decreasing preference, but my preference isn't strong.)

Just duplicating the code seems size unfriendly.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1117,6 +1125,28 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
+    validationIt(
+      persistence,
+      'with != and inequality queries on different fields fail',
+      db => {
+        const collection = db.collection('test');
+        expect(() =>
+          collection.where('y', '>', 32).where('x', notEqualOp, 33)
+        ).to.throw(
+          'Invalid query. All where filters with an ' +
+            'inequality (<, <=, >, or >=) must be on the same field.' +
```
wilhuff, line 1144: Should `!=` be in this list of inequalities?

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1171,6 +1204,36 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
+          .collection('test')
+          .where('foo', notInOp, [2, 3])
+          .where('foo', 'array-contains', 1)
+      ).to.throw(
+        "Invalid query. You cannot use 'array-contains' filters with " +
+          "'not-in' filters."
+      );
+    });
+
+    validationIt(persistence, 'with != and NOT_IN filters fail', db => {
```
wilhuff, line 1219: Since the user-visible spelling for NOT_IN is not-in, maybe describe this in terms of "not-in" rather than the proto name for this operator?

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -1171,6 +1204,36 @@ apiDescribe('Validation:', (persistence: boolean) => {
...
+      expect(() =>
+        db
+          .collection('test')
+          .where('foo', notInOp, [2, 3])
+          .where('foo', notEqualOp, 4)
+      ).to.throw(
+        "Invalid query. You cannot use '!=' filters with " + "'not-in' filters."
```
wilhuff, line 1226: What's with the extra plus sign in here? There's a couple in this file. Search for `" + "`.

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -276,3 +277,17 @@ export function withTestCollectionSettings(
     }
   );
 }
+
+// TODO(ne-queries): This exists just so we don't have to do the cast
+// repeatedly. Once we expose '!=' publicly we can remove it and
+// just use 'array-contains-any' in all the tests.
```
wilhuff, line 283: Typo? Surely you don't mean "array-contains-any" here (or "in" in the next one).

packages/firestore/test/unit/core/query.test.ts
```diff
@@ -218,6 +218,64 @@ describe('Query', () => {
     expect(queryMatches(query1, document)).to.be.true;
   });
 
+  it('matches NOT_IN filters', () => {
```
wilhuff, line 221: Same comment re NOT_IN vs not-in. NOT_IN is a proto representation detail while `not-in` is the public API value.

packages/firestore/src/core/query.ts
thebrianchen: Extracted into helper.

packages/firestore/src/api/database.ts
```diff
@@ -1700,14 +1707,41 @@ function validateDisjunctiveFilterElements(
   }
 }
 
+function conflictingOps(op: Operator): Operator[] {
```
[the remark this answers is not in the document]
wilhuff: These comments are mostly just repeating what the code is doing. What I was hoping for was a higher-level comment (outside the switch statement) that gave the collective rules together. For example, as you've implemented this now, you've repeated "Only one ARRAY operator can be used" in multiple places and I'm prompting for it to be in one place.

Basically: when you constructed these lists, you had to evaluate a list of rules. Then for each operator pair you evaluated those rules to determine if they were incompatible. What I want you to document is the list of rules you used to build this structure.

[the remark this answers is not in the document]
thebrianchen: Makes sense. Documented at the top of the function.

PR #3562, opened 2020-08-04
Add precondition format checks against api-key and app-id

ChaoqunCHEN:


packages/installations/src/util/precondition-checks.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import { assert } from 'chai';
+import {
+  isValidAppIdFormat,
+  isValidApiKeyFormat
+} from '../util/precondition-checks';
+
+describe('precondition checks', () => {
+  it('returns true if App ID is valid', () => {
+    assert.isTrue(isValidAppIdFormat('1:123456789:android:abcdef'));
+    assert.isTrue(
```
zwu52, line 27: (For all other asserts) I think core folks want to have a consistence testing style by using behavior testing (i.e. expect). Please consider swapping

packages/installations/src/testing/fake-generators.ts
```diff
@@ -29,7 +29,7 @@ export function getFakeApp(): FirebaseApp {
   return {
     name: 'appName',
     options: {
-      apiKey: 'apiKey',
+      apiKey: 'AIzaSyabcdefghijklmnopqrstuvwxyz1234567',
```
zwu52, line 32: This looks like a legit api key value. Would it be nicer to continue  just use a dummy value for testing purpose? or a least comment on the fakeness of this value everywhere (also making this a constant would be ideal)

packages/installations/src/util/precondition-checks.ts
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const APP_ID_IDENTIFICATION_SUBSTRING = ':';
+const API_KEY_FORMAT = new RegExp('^A[\\w-]{38}$');
+
+export function isValidAppIdFormat(appId: string): boolean {
```
zwu52, line 21: Does FIS currently support these "non valid" API ID and API Key. If that's the case, this change will be breaking for those users. And such change want to be delayed tilled 2020 Firecon which is in September

ChaoqunCHEN: No, FIS doesn't support invalid API-ID and API-key. Any requests with options that cannot pass these precondition checks will be rejected by FIS backend.

packages/installations/src/testing/fake-generators.ts
ChaoqunCHEN: It's a fake api key if you look closely. yeah, I extracted them into a constant. Thanks for the advice.

packages/installations/src/util/precondition-checks.test.ts
ChaoqunCHEN: done

packages/installations/src/util/errors.ts
```diff
@@ -30,6 +32,14 @@ export const enum ErrorCode {
 const ERROR_DESCRIPTION_MAP: { readonly [key in ErrorCode]: string } = {
   [ErrorCode.MISSING_APP_CONFIG_VALUES]:
     'Missing App configuration value: "{$valueName}"',
+  [ErrorCode.INVALID_APP_ID_FORMAT]:
+    'Please set your Application ID. A valid Firebase App ID is required to communicate ' +
```
Feiyang1, line 36: The message sounds like the app id is missing, but the error is actually a malformed appId. I would also just use `appId` which matches the name in the config object.

packages/installations/src/util/errors.ts
```diff
@@ -30,6 +32,14 @@ export const enum ErrorCode {
...
+    'with Firebase server APIs: It identifies your application with Firebase.' +
+    'Please refer to https://firebase.google.com/support/privacy/init-options.',
+  [ErrorCode.INVALID_API_KEY_FORMAT]:
+    'Please set a valid API key. A Firebase API key is required to communicate with ' +
```
Feiyang1, line 40: Similarly I would just use `apiKey` to match the name in the config object.

integration/firestore/firebase_export.ts
```diff
@@ -33,7 +33,10 @@ export function newTestFirestore(
   }
   const app =
     typeof nameOrApp === 'string'
-      ? firebase.initializeApp({ apiKey: 'fake-api-key', projectId }, nameOrApp)
+      ? firebase.initializeApp(
+          { apiKey: 'A-test-api-key-jklmnopqrstuvwxyz1234567', projectId },
```
Feiyang1, line 37: Does the test fail without the change? I think you only need the correct format in installation tests, specifically the tests that involves `extractAppConfig()`. We can leave other places unchanged, right?

ChaoqunCHEN: Yeah, I saw many tests were failing so I added these changes.

packages/installations/src/util/errors.ts
ChaoqunCHEN: makes sense. done

packages/installations/src/util/errors.ts
ChaoqunCHEN: done

integration/firestore/firebase_export.ts
Feiyang1: That is weird. I don't think your change affects firestore. Are you sure the failure comes from firestore? It will be really inconvenient if we are forced use exact 38 characters apiKey in tests.

andirayo (commented):
Removing this from my list of review-requests.

PR #3567, opened 2020-08-05
Fix more edge cases around auth-compat-exp polyfill (auth-next)

avolkovi:
- Add environment checks to signInWithPopup/Redirect
- Add tenantId check to updateCurrentUser
- Rejigger PhoneMultiFactorGenerator and PhoneMultiFactorAssertion to not require an auth object, we'll use the one provided to getMultiFactorResolver by getting it from the closure


packages-exp/auth-compat-exp/src/platform.ts
```diff
@@ -0,0 +1,159 @@
...
+  isIE
+} from '@firebase/util';
+
+declare global {
+  interface Document {
+    documentMode?: number;
+  }
+}
+
+// TODO(avolkovi): Remove this after rebase and use the version in @firebase/util
```
samhorlbeck, line 33: can you do this now?

avolkovi: sure, i just need to rebase auth-next on master

avolkovi: done

PR #3568, opened 2020-08-05
Tree-Shake RemoteStore Streams

schmidt-sebastian:
This PR changes the remote store to make the streams optional. In the callsites that require the streams, the streams are loaded and registered with the RemoteStore. The RemoteStore then tracks these streams via `networkStatusHandlers`, which take care of forwarding the calls to disable and enable the network.

This PR also ensures that the mutation queue is drained at startup if persistence is enabled. Unfortunately, this means that clients that use IndexedDb now also always depend on SyncEngine and RemoteStore.


packages/firestore/src/remote/remote_store.ts
```diff
@@ -924,3 +895,111 @@ export async function remoteStoreApplyPrimaryState(
     remoteStoreImpl.onlineStateTracker.set(OnlineState.Unknown);
   }
 }
+
+/**
+ * Registers a callback function that is invoked when the network status
+ * changes. Multiple callbacks are supported.
+ */
+function addNetworkStatusChangedHandler(
```
wilhuff, line 903: Should this just be `onNetworkStatusChange` to match other event handlers we have?

packages/firestore/src/remote/remote_store.ts
```diff
@@ -924,3 +895,111 @@ export async function remoteStoreApplyPrimaryState(
...
+
+/**
+ * If not yet initialized, registers the WatchStream and its network state
+ * callback with `remoteStoreImpl`. Returns the existing stream if one is
+ * already available.
+ *
+ * PORTING NOTE: On iOS and Android, the WatchStream gets registered on startup.
+ * This is not done on Web to allow it to be tree-shaken.
+ */
+function ensureWatchStream(
```
wilhuff, line 919: Optional: since this is essentially a lazy-initializing getter, you might want to just call this `watchStream` to cut verbosity in all the call sites.

wilhuff (approved):
LGTM with some naming ideas.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: I inlined this and changed the internal variable name to `onNetworkStatusChange`.

packages/firestore/src/remote/remote_store.ts
schmidt-sebastian: I left this as is since this it would require me to rename all occurrences of `writeStream` and `watchStream` (to avoid name clashes at the callsite), which might make the code even more verbose.

PR #3569, opened 2020-08-05
Make newConnection() synchronous

schmidt-sebastian:
`newConnection()` doesn't need to return a Promise, which means that Datastore can be constructed synchronously.


packages/firestore/test/unit/specs/spec_test_components.ts
```diff
@@ -121,19 +121,16 @@ function failTransactionIfNeeded(
...
-  connection!: MockConnection;
-
-  async loadConnection(cfg: ComponentConfiguration): Promise<Connection> {
-    this.connection = new MockConnection(cfg.asyncQueue);
-    return this.connection;
+  constructor(private readonly connection: MockConnection) {
+    super();
   }
-
+
```
thebrianchen, line 127: nit: delete extra whitespace

packages/firestore/src/platform/node_lite/connection.ts
```diff
@@ -24,9 +24,9 @@ import { Connection } from '../../remote/connection';
 export { newConnectivityMonitor } from '../browser/connection';
 
 /** Initializes the HTTP connection for the REST API. */
-export function newConnection(databaseInfo: DatabaseInfo): Promise<Connection> {
+export function newConnection(databaseInfo: DatabaseInfo): Connection {
   // node-fetch is meant to be API compatible with `fetch`, but its type don't
```
thebrianchen, line 28: bonus nit: s/don't/doesn't

packages/firestore/src/platform/browser_lite/connection.ts
```diff
@@ -22,6 +22,6 @@ import { FetchConnection } from './fetch_connection';
 export { newConnectivityMonitor } from '../browser/connection';
 
 /** Initializes the HTTP connection for the REST API. */
-export function newConnection(databaseInfo: DatabaseInfo): Promise<Connection> {
-  return Promise.resolve(new FetchConnection(databaseInfo, fetch.bind(null)));
+export function newConnection(databaseInfo: DatabaseInfo): Connection {
+  return new FetchConnection(databaseInfo, fetch.bind(null));
```
thebrianchen, line 26: Can you explain why we need to bind `fetch` to `null` here?

packages/firestore/lite/src/api/components.ts
```diff
@@ -81,6 +84,6 @@ export async function removeComponents(firestore: Firestore): Promise<void> {
   if (datastorePromise) {
     logDebug(LOG_TAG, 'Removing Datastore');
     datastoreInstances.delete(firestore);
-    return (await datastorePromise).termiate();
+    return (await datastorePromise).terminate();
```
thebrianchen, line 87: s/dataPromise/datastoreInstance?

You don't need the `await` anymore since, it's no longer a promise.

packages/firestore/src/remote/datastore.ts
```diff
@@ -131,7 +125,7 @@ class DatastoreImpl extends Datastore {
       });
   }
 
-  async termiate(): Promise<void> {
+  async terminate(): Promise<void> {
```
thebrianchen, line 128: Unrelated question to this PR, but why are we making `DatastoreImpl.terminate()` async if the method isn't? Why does the base `Datastore` need to have an abstract `terminate()` that returns `Promise<void>`, considering the only implementation of the class is `DatastoreImpl`?

packages/firestore/lite/src/api/components.ts
schmidt-sebastian: Done

packages/firestore/src/platform/browser_lite/connection.ts
schmidt-sebastian: The first argument to any JavaScript function is the "this context", which is this case is not used. It still has to be explicitly set as we are using "bind". If we just pass "bind" as this then calling this function pointer later would use the first argument as the "this" context... and essentially drop it. 
Note that we could also do `fetch.bind(this)`, which would be equivalent to using an array function here.

As an example, if we have a function like this:

```
class Foo {
bar = 'bar';

function foo() {
  console.log(this.bar);
}

function test() {
   this.foo(); // Prints 'bar';
   (() => this.foo())() // Prints 'bar', since arrow functions retain `this` context.
   this.foo().bind({bar:'foo'})(); // Different `this`. Prints 'foo'.
}
```

```

packages/firestore/src/platform/node_lite/connection.ts
schmidt-sebastian: Done

packages/firestore/src/remote/datastore.ts
schmidt-sebastian: I don't have an answer for you, so I changed this as well. In general, returning a Promise even when it is not needed can be seen as "future proofing", but this is only really needed for Public APIs that we don't want to break if there is ever a need to make them async.

packages/firestore/test/unit/specs/spec_test_components.ts
schmidt-sebastian: Done

PR #3575, opened 2020-08-06
Make shutdown idempotent

### LINKED ISSUE #2755: Offline Mode Improvements

schmidt-sebastian:
The CI doesn't seem to work right now, but I still wanted to send this out for review. This replaces https://github.com/firebase/firebase-js-sdk/pull/3164 and makes it possible to call terminate() again if it failed with an IndexedDB exception. 

This is the last PR for IndexedDB retries (at least for the current Firestore SDK, firestore-exp has one TODO to use a retryable operation).

Fixes #2755


packages/firestore/exp/src/api/database.ts
```diff
@@ -131,16 +133,28 @@ export class Firestore extends LiteFirestore
...
-    debugAssert(!this._terminated, 'Cannot invoke _terminate() more than once');
-    return this._queue.enqueueAndInitiateShutdown(async () => {
-      await super._terminate();
-      await removeComponents(this);
-
-      // `removeChangeListener` must be called after shutting down the
-      // RemoteStore as it will prevent the RemoteStore from retrieving
-      // auth tokens.
-      this._credentials.removeChangeListener();
+    this._queue.initiateShutdown();
```
wilhuff, line 136: This is essentially a parallel change to the one in the iOS repo (which I hadn't yet ported everywhere else). During that change discussion arose that "shutdown" didn't really capture what was happening anymore: the queue is really still running, but there's just a limited set of operations you can perform on it. This method became `enterRestrictedMode` and the corresponding enqueue would be `enqueueAndForgetEvenWhileRestricted`.

Would you be willing to make that renaming change while you're in here? Usage ends up being the same as here: https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/core/firestore_client.cc#L230, so structurally this is already a move toward convergence.

schmidt-sebastian: Done. This is a good reminder that I need to port the RemoteStore modes to the other platforms.

PR #3578, opened 2020-08-06
Add an assertTypes utility

samhorlbeck:
See the comment and the tests for usage.

There are unrelated prettier changes

### DIFF
--- a/packages-exp/auth-exp/src/core/util/assert.ts
+++ b/packages-exp/auth-exp/src/core/util/assert.ts
@@ -44,13 +44,12 @@ export function assert(
   }
 }
 
-type TypeExpectation = Function|string|MapType;
+type TypeExpectation = Function | string | MapType;
 
-interface MapType extends Record<string, TypeExpectation|Optional> {}
+interface MapType extends Record<string, TypeExpectation | Optional> {}
 
 class Optional {
-  constructor(readonly type: TypeExpectation) {
-  }
+  constructor(readonly type: TypeExpectation) {}
 }
 
 export function opt(type: TypeExpectation): Optional {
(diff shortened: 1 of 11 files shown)


packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -44,6 +44,92 @@ export function assert(
...
+ *
+ * @param args
+ * @param expected
+ */
+export function assertTypes(
+  args: Omit<IArguments, 'callee'>,
+  ...expected: Array<TypeExpectation | Optional>
+): void {
+  if (args.length > expected.length) {
+    fail('TODO', AuthErrorCode.ARGUMENT_ERROR);
```
avolkovi, line 89: can this be debugFail? We should never hit this based on user input, right?

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -44,6 +44,92 @@ export function assert(
...
+      const record = arg as Record<string, unknown>;
+      const map = expect as MapType;
+      const keys = Object.keys(expect);
+
+      assertTypes(
+        keys.map(k => record[k]),
+        ...keys.map(k => map[k])
+      );
+    } else {
+      assert(arg instanceof expect, 'app', AuthErrorCode.ARGUMENT_ERROR);
```
avolkovi, line 128: same here, we should probably use debugAssert and not require an app

packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -44,6 +44,92 @@ export function assert(
...
+        'TODO',
+        AuthErrorCode.ARGUMENT_ERROR
+      );
+    } else if (typeof expect === 'object') {
+      // Recursively check record arguments
+      const record = arg as Record<string, unknown>;
+      const map = expect as MapType;
+      const keys = Object.keys(expect);
+
+      assertTypes(
```
avolkovi, line 123: will the recursion here work with cross referenced types like Auth & User that point at each other via this.auth and this.currentUser?

packages-exp/auth-exp/src/core/util/assert.ts
samhorlbeck: Nope, for example:

```ts
function foo(auth: Auth) {
  assertTypes(arguments, AuthImpl);
}
```
should fail with this call (which is perfectly valid in javascript)
```js
foo(auth, 'some-other-parameter');
```

packages-exp/auth-exp/src/core/util/assert.ts
samhorlbeck: These are runtime checks and _should_ fail in the production build

packages-exp/auth-exp/src/core/util/assert.ts
samhorlbeck: The recursion only happens for record types (not for instance checks). So
```ts
function foo(profile: {displayName?: string}) {
  assertTypes(arguments, {displayName: opt(string)});
}
```
will recurse, whereas
```
function foo(auth: Auth) {
  assertTypes(arguments, AuthImpl);
}
```
will not

PR #3585, opened 2020-08-07
Extended Usage of `isIndexedDBAvailable` to Service Worker

XuechunHou:
change from `'indexedDB' in window` to `'indexedDB' in self`

### DIFF
--- a/.changeset/silly-moles-rush.md
+++ b/.changeset/silly-moles-rush.md
@@ -1,0 +1,6 @@
+---
+"@firebase/firebase": minor
+"@firebase/util": minor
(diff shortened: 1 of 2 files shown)


.changeset/silly-moles-rush.md
```diff
@@ -0,0 +1,6 @@
+---
+"@firebase/firebase": minor
```
hsubox76, line 2: I didn't catch this in the previous PR but the package name is `firebase`, there's no `@firebase/firebase`. Also I think these should be patch because there's no API change.

PR #3588, opened 2020-08-07
Size Analysis Tool - Phase 1

XuechunHou:
- [x] completed the implementation of size analysis tool

- [x] provide a command line interface for using the tool

### DIFF
--- a/.changeset/shy-bears-heal.md
+++ b/.changeset/shy-bears-heal.md
@@ -1,5 +1,0 @@
----
-"size-analysis-firebase": patch
----
-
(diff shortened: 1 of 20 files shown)


.changeset/shy-bears-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"size-analysis-firebase": patch
+---
+
+Size Analysis Tool - Phase 1
```
Feiyang1, line 5: We don't need changeset for private packages. please delete

PR #3594, opened 2020-08-10
Merge firestore bundles implementation without exposing public APIs

wu-hui:


.changeset/lemon-steaks-draw.md
```diff
@@ -0,0 +1,3 @@
+---
```
schmidt-sebastian, line 1: Please add a changeset entry to describe the size increase.

packages/firestore-types/index.d.ts
```diff
@@ -97,6 +97,33 @@ export class FirebaseFirestore {
...
+
+export interface LoadBundleTaskProgress {
+  documentsLoaded: number;
+  totalDocuments: number;
+  bytesLoaded: number;
+  totalBytes: number;
+  taskState: TaskState;
+}
+
+export type TaskState = 'Error' | 'Running' | 'Success';
```
schmidt-sebastian, line 125: For now, we should remove this from this file.

packages/firestore/exp-types/index.d.ts
```diff
@@ -516,6 +516,33 @@ export function snapshotEqual<T>(
...
+export interface LoadBundleTaskProgress {
+  documentsLoaded: number;
+  totalDocuments: number;
+  bytesLoaded: number;
+  totalBytes: number;
+  taskState: TaskState;
+}
+
+export type TaskState = 'Error' | 'Running' | 'Success';
+
```
schmidt-sebastian, line 545: These can stay. You can also just make this feature public in firestore-exp.

packages/firestore/exp/index.ts
```diff
@@ -31,8 +31,9 @@ export {
   waitForPendingWrites,
   disableNetwork,
   enableNetwork,
-  terminate,
-  Settings
+  _namedQuery,
```
schmidt-sebastian, line 34: Feel free to make these APIs public in firestore-exp.

packages/firestore/exp/src/api/database.ts
```diff
@@ -479,3 +489,41 @@ function verifyNotInitialized(firestore: FirebaseFirestore): void {
...
+
+  const resultTask = new LoadBundleTask();
+
+  firestore._queue.enqueueAndForget(async () => {
+    const databaseId = (await firestore._getConfiguration()).databaseInfo
+      .databaseId;
+    const reader = createBundleReader(bundleData, newSerializer(databaseId));
+    const syncEngine = await getSyncEngine(firestore);
+
+    loadBundleSyncEngine(syncEngine, reader, resultTask);
```
schmidt-sebastian, line 507: FWIW, I have lately followed Gil's suggestions more strictly and prefixed the free standing APIs with the class name.

That would mean s/loadBundleSyncEngine/syncEngineLoadBundle/

packages/firestore/exp/src/api/database.ts
```diff
@@ -479,3 +489,41 @@ function verifyNotInitialized(firestore: FirebaseFirestore): void {
...
+    return resultTask.catch((e: Error) => {
+      logWarn(LOG_TAG, `Loading bundle failed with ${e}`);
+    });
```
schmidt-sebastian, line 510: This pattern means that the resulting user task never fails.  If you want to log an error, you should catch and rethrow here. A better way might be to log an error when the error occurs.
Note that you also should not return this task to the AsyncQueue, as it will cause it to fail. Instead, just return an empty Promise (which is the default behavior for async functions, so you can just remove `return`).

packages/firestore/exp/src/api/database.ts
```diff
@@ -479,3 +489,41 @@ function verifyNotInitialized(firestore: FirebaseFirestore): void {
...
+  name: string
+): Promise<Query | null> {
+  return getLocalStore(firestore).then(localStore => {
+    return getNamedQuery(localStore, name).then(namedQuery => {
+      if (!namedQuery) {
+        return null;
+      }
+
+      return null;
+      // return new Query(namedQuery.query, firestore, null);
```
schmidt-sebastian, line 526: Let's fix this before merging.

packages/firestore/exp/test/shim.ts
```diff
@@ -162,6 +164,17 @@ export class FirebaseFirestore
...
 
+  loadBundle(
+    bundleData: ArrayBuffer | ReadableStream<Uint8Array> | string
+  ): LoadBundleTask {
+    return _loadBundle(this._delegate, bundleData)!;
+  }
+
+  async namedQuery(name: string): Promise<Query | null> {
+    return null;
+    // return namedQuery(this._delegate, name);
```
schmidt-sebastian, line 175: Let's fix this before merging.

packages/firestore/rollup.shared.js
```diff
@@ -129,7 +129,8 @@ const manglePrivatePropertiesOptions = {
   },
   mangle: {
     properties: {
-      regex: /^__PRIVATE_/
+      regex: /^__PRIVATE_/,
+      reserved: ['do']
```
schmidt-sebastian, line 133: Should we just add all JS keywords here? This will fail with `if` and `of` next and it is really tough to figure out these failures.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -573,6 +581,29 @@ export class FirestoreClient {
...
+
+    const reader = createBundleReader(
+      data,
+      newSerializer(this.databaseInfo.databaseId)
+    );
+    this.asyncQueue.enqueueAndForget(async () => {
+      loadBundle(this.syncEngine, reader, resultTask);
+      return resultTask.catch(e => {
+        logWarn(LOG_TAG, `Loading bundle failed with ${e}`);
+      });
```
schmidt-sebastian, line 599: Same comment as above.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -573,6 +581,29 @@ export class FirestoreClient {
...
+    });
+  }
+
+  getNamedQuery(queryName: string): Promise<NamedQuery | undefined> {
+    this.verifyNotTerminated();
+    return getNamedQuery(this.localStore, queryName);
```
schmidt-sebastian, line 605: This should happen on the AsyncQueue.

packages/firestore/src/local/local_store.ts
```diff
@@ -1218,3 +1271,219 @@ export async function ignoreIfPrimaryLeaseLoss(
...
   }
 }
+
+/**
+ * Creates a new target using the given bundle name, which will be used to
+ * hold the keys of all documents from the bundle in query-document mappings.
+ * This ensures that the loaded documents do not get garbage collected
+ * right away.
+ */
+export function umbrellaTarget(bundleName: string): Target {
```
schmidt-sebastian, line 1281: This doesn't need to be exported.

packages/firestore/src/local/shared_client_state_schema.ts
```diff
@@ -115,6 +115,15 @@ export function createWebStorageOnlineStateKey(persistenceKey: string): string {
   return `${ONLINE_STATE_KEY_PREFIX}_${persistenceKey}`;
 }
 
+// The WebStorage prefix that plays as a event to indicate the remote documents
+// might have changed due to some secondary tabs loading a bundle.
+// format of the key is:
+//     firestore_remote_documents_changed_<persistenceKey>
+export const BUNDLE_LOADED_KEY_PREFIX = 'firestore_bundle_loaded';
```
schmidt-sebastian, line 122:
```suggestion
//     firestore_bundle_loaded_<persistenceKey>
export const BUNDLE_LOADED_KEY_PREFIX = 'firestore_bundle_loaded';
```

packages/firestore/test/integration/api_internal/bundle.test.ts
```diff
@@ -0,0 +1,309 @@
...
+} from '../util/helpers';
+import { DatabaseId } from '../../../src/core/database_info';
+import { key } from '../../util/helpers';
+import { EventsAccumulator } from '../util/events_accumulator';
+import { TestBundleBuilder } from '../../unit/util/bundle_data';
+import { newTextEncoder } from '../../../src/platform/serializer';
+import { collectionReference } from '../../util/api_helpers';
+import { LoadBundleTask } from '../../../src/api/bundle';
+
+// TODO(b/162594908): Move this to api/ instead of api_internal.
```
schmidt-sebastian, line 34: I would consider this blocking as this change will ensure that this feature works in the minified Browser build and is also required to ensure that the firestore-exp code you added is tested.

packages/firestore/src/core/firestore_client.ts
wu-hui: This looks different now, but should be fixed.

packages/firestore/src/core/firestore_client.ts

packages/firestore/src/local/shared_client_state_schema.ts
wu-hui: Done.

packages/firestore/test/integration/api_internal/bundle.test.ts
wu-hui: Done.

packages/firestore/exp/src/api/database.ts
wu-hui: Done.

packages/firestore/exp/src/api/database.ts
wu-hui: I think this is fixed, although this particular piece of code no longer exists.

packages/firestore/exp/src/api/database.ts
wu-hui: Done.

packages/firestore/exp/test/shim.ts
wu-hui: Done.

packages/firestore/rollup.shared.js
wu-hui: Done.

wu-hui (commented):
This has changed quite a bit on the api/firestore_client level since last time you reviewed it. 

Also, this is no longer release but hide it from public API, it is release bundles as free functions. I will hold it until we got all approvals and ready to launch.

packages/firestore/src/api/database.ts
```diff
@@ -427,36 +424,6 @@ export function setLogLevel(level: PublicLogLevel): void {
   setClientLogLevel(level);
 }
 
-export function loadBundle(
```
schmidt-sebastian: You can keep these functions as long as they are not referenced.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -648,3 +656,42 @@ export function executeQueryViaSnapshotListener(
   });
   return eventManagerListen(eventManager, listener);
 }
+
+export async function firestoreClientLoadBundle(
+  client: FirestoreClient,
+  data: ReadableStream<Uint8Array> | ArrayBuffer | string,
+  resultTask: LoadBundleTask
+): Promise<void> {
+  client.verifyNotTerminated();
```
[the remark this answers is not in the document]
schmidt-sebastian: Same just below.

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1508,3 +1532,77 @@ export function ensureWriteCallbacks(syncEngine: SyncEngine): SyncEngineImpl {
...
+): void {
+  const syncEngineImpl = debugCast(syncEngine, SyncEngineImpl);
+
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  loadBundleImpl(syncEngineImpl, bundleReader, task).then(() => {
+    syncEngineImpl.sharedClientState.notifyBundleLoaded();
+  });
+}
+
+async function loadBundleImpl(
```
schmidt-sebastian, line 1556: Could this potentially be inlined? Or did we not want to make `syncEngineLoadBundle` async?

schmidt-sebastian (commented):
Looks there there are still some stray API bits, but otherwise good. Since the size tooling seems to be broken at the moment, can you provide a rough number on the size increase?

wu-hui (commented):
firebase-firestore.js 283361 -> 286235
firebase-firestore.memory.js 221571 -> 224766

packages/firestore/src/api/database.ts
```diff
@@ -424,6 +427,36 @@ export function setLogLevel(level: PublicLogLevel): void {
   setClientLogLevel(level);
 }
 
+export function loadBundle(
+  db: Firestore,
+  bundleData: ArrayBuffer | ReadableStream<Uint8Array> | string
+): ApiLoadBundleTask {
+  const resultTask = new LoadBundleTask();
```
schmidt-sebastian, line 434: We need to figure out where to call "ensureClientConfigured". It might depend on how we expose this function in the end.

schmidt-sebastian (approved):
Wheee!

PR #3595, opened 2020-08-10
Expose additionalUserInfo in auth-exp and auth-compat-exp

avolkovi:


packages/auth-types/index.d.ts
```diff
@@ -135,7 +135,7 @@ export type ActionCodeSettings = {
 export type AdditionalUserInfo = {
   isNewUser: boolean;
   profile: Object | null;
-  providerId: string;
+  providerId: string | null;
```
avolkovi, line 138: this was a mistake in the old public types https://github.com/firebase/firebase-js-sdk/blob/6b53e0058483c9002d2fe56119f86fc9fb96b56c/packages/auth/src/additionaluserinfo.js#L137

PR #3596, opened 2020-08-10
Add lite/package.json to 'files'

schmidt-sebastian:


schmidt-sebastian:
@Feiyang1 Note that "lite/package.json" still has `private: true`. Do we need to change that?

Feiyang1:
You can remove the `private` field in "lite/package.json" because it doesn't mean anything for submodules.

PR #3600, opened 2020-08-10
Add runtime type checking to public-facing methods

samhorlbeck:


samhorlbeck:
This introduced a circular dependency which I need to fix before merging

packages-exp/auth-exp/src/core/action_code_url.ts
```diff
@@ -108,5 +111,6 @@ export function parseActionCodeURL(
   auth: externs.Auth,
   link: string
 ): externs.ActionCodeURL | null {
+  assertTypes(arguments, AuthImplCompat, 'string');
```
Feiyang1, line 114: Runtime type checks are dead code for Typescript developers, which I believe a large part if not majority of our developers are. Do we know what is the size impact of adding the `assertTypes`? 

Given the trend that Typescript is gaining more popularity yoy,  I feel strongly that we don't need runtime type check, and should just let Typescript handle it, even if the size impact is moderate.

@jamesdaniels @davideast, do you have an opinion on it?

samhorlbeck: This PR increases the size of the umd build by 3k (from 135k to 138k)

Feiyang1: Thanks for getting the size diff. It looks small, but I still don't think we should include it unless it is absolutely necessary (which I don't think is). I created a [doc](go/jscore-input-validation) a while ago on this topic. It's probably a good time to revive it to gain consensus across teams. Would you mind putting comment in the doc so we can continue discussion there?

PR #3601, opened 2020-08-11
delete services that implement the new FirebaseService interface 

Feiyang1:
Replaces https://github.com/firebase/firebase-js-sdk/pull/3454 which for some reason keeps failing CI test.


packages/firestore/scripts/run-tests.js
```diff
@@ -1,4 +1,4 @@
-"use strict";
+'use strict';
```
schmidt-sebastian, line 1: I know this is a pain, but I am trying not to run these generated sources through formatting to make it obvious that they are generated.

PR #3620, opened 2020-08-12
Update query-document mapping from bundles.

wu-hui:


packages/firestore/src/protos/firestore_bundle_proto.ts
```diff
@@ -54,6 +54,9 @@ export interface BundledDocumentMetadata {
 
   /** BundledDocumentMetadata exists */
   exists?: boolean | null;
+
+  /** The names of the queries in this bundle that this document matches to. */
+  queries?: string[] | null;
```
schmidt-sebastian, line 59: In Proto3, is not possible to have a missing array. Missing arrays are represented as empty lists, so this should be `queries: string[]`.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1706,6 +1758,69 @@ function genericLocalStoreTests(
       .finish();
   });
 
+  it('handles loading named queries allocates targets and updates target document mapping', async () => {
```
schmidt-sebastian, line 1761:
```suggestion
  it('loading named queries allocates targets and updates target document mapping', async () => {
```

packages/firestore/src/core/bundle.ts
```diff
@@ -218,16 +218,17 @@ export class BundleLoader {
       'Bundled documents ends with a document metadata and missing document.'
     );
 
+    const result = await applyBundleDocuments(this.localStore, this.documents);
+
```
schmidt-sebastian, line 222: As far as I can tell, returning the map of matching documents from `applyBundleDocuments` saves one iteration. This code would be less complex if you extracted the mapping in this function (ideally, via a call to a helper function). This would remove your changes from `applyBundleDocuments` and some of the changes from LocalStoreTester.

packages/firestore/src/local/local_store.ts
```diff
@@ -1259,11 +1264,12 @@ export async function ignoreIfPrimaryLeaseLoss(
 export function applyBundleDocuments(
   localStore: LocalStore,
   documents: BundledDocuments
-): Promise<MaybeDocumentMap> {
+): Promise<ApplyBundleDocumentsResult> {
   const localStoreImpl = debugCast(localStore, LocalStoreImpl);
   const bundleConverter = new BundleConverter(localStoreImpl.serializer);
   let documentMap = maybeDocumentMap();
   let versionMap = documentVersionMap();
+  const queryDocumentMap: Map<string, DocumentKeySet> = new Map();
```
schmidt-sebastian, line 1272:
```suggestion
  const queryDocumentMap = new Map<string, DocumentKeySet>();
```

packages/firestore/src/local/local_store.ts
```diff
@@ -1400,10 +1421,22 @@ export async function saveNamedQuery(
...
       }
-      return updateReadTime.next(() =>
-        localStoreImpl.bundleCache.saveNamedQuery(transaction, query)
-      );
+      return updateReadTime
+        .next(() =>
+          localStoreImpl.bundleCache.saveNamedQuery(transaction, query)
+        )
+        .next(() =>
+          localStoreImpl.targetCache.addMatchingKeys(
```
schmidt-sebastian, line 1434: I think you should only do this if you updated the readTime above. On top of that, you need to also remove all existing keys, otherwise your mapping will not match the backend's mapping.

packages/firestore/src/core/bundle.ts
wu-hui: Done. It IS nicer this way!

packages/firestore/src/local/local_store.ts
wu-hui: Done.

packages/firestore/src/protos/firestore_bundle_proto.ts
wu-hui: Done.

packages/firestore/src/local/local_store.ts
wu-hui: Agree to both.

packages/firestore/test/unit/local/local_store.test.ts
wu-hui: Done.

packages/firestore/src/core/bundle.ts
```diff
@@ -218,17 +223,20 @@ export class BundleLoader {
...
-    const result = await applyBundleDocuments(this.localStore, this.documents);
+    const changedDocuments = await applyBundleDocuments(
+      this.localStore,
+      this.documents
+    );
 
+    const queryDocumentMap = getQueryDocumentMapping(
+      this.localStore,
+      this.documents
+    );
```
schmidt-sebastian, line 230: Nit: Empty line below.

packages/firestore/src/local/local_store.ts
```diff
@@ -1249,9 +1249,28 @@ export async function ignoreIfPrimaryLeaseLoss(
...
-export interface ApplyBundleDocumentsResult {
-  changedDocuments: MaybeDocumentMap;
-  queryDocumentMap: Map<string, DocumentKeySet>;
+export function getQueryDocumentMapping(
+  localStore: LocalStore,
+  documents: BundledDocuments
+): Map<string, DocumentKeySet> {
+  const localStoreImpl = debugCast(localStore, LocalStoreImpl);
+  const queryDocumentMap = new Map<string, DocumentKeySet>();
+  const bundleConverter = new BundleConverter(localStoreImpl.serializer);
```
schmidt-sebastian, line 1258: Is it possible to get the serializer from the callsite of this function, drop its dependency on "LocalStore" and move it to bundle.ts?

packages/firestore/src/local/local_store.ts
```diff
@@ -1300,6 +1324,9 @@ export function applyBundleDocuments(
             txn,
             changedDocs
           );
+        })
+        .next(changedDocuments => {
+          return PersistencePromise.resolve(changedDocuments);
```
schmidt-sebastian, line 1329: This looks like a no-op.

packages/firestore/src/local/local_store.ts
```diff
@@ -1389,21 +1417,41 @@ export async function saveNamedQuery(
     'readwrite',
     transaction => {
       // Update allocated target's read time, if the bundle's read time is newer.
-      let updateReadTime = PersistencePromise.resolve();
+      let updateReadTime = PersistencePromise.resolve(false);
```
schmidt-sebastian, line 1420: If still needed: s/updateReadTime/updatedReadTime (or readTimeUpdated)

packages/firestore/src/local/local_store.ts
```diff
@@ -1389,21 +1417,41 @@ export async function saveNamedQuery(
...
+          if (updated) {
+            return localStoreImpl.targetCache
+              .removeMatchingKeysForTargetId(transaction, allocated.targetId)
+              .next(() =>
+                localStoreImpl.targetCache.addMatchingKeys(
+                  transaction,
+                  documents,
+                  allocated.targetId
+                )
+              );
```
schmidt-sebastian, line 1449: Could this not just be chained off the call to "updateTargetData"?

packages/firestore/src/local/local_store.ts
```diff
@@ -1389,21 +1417,41 @@ export async function saveNamedQuery(
...
           readTime
         );
-        updateReadTime = localStoreImpl.targetCache.updateTargetData(
-          transaction,
+        updateReadTime = localStoreImpl.targetCache
+          .updateTargetData(transaction, newTargetData)
+          .next(
+            () => true,
+            () => false
+          );
```
schmidt-sebastian, line 1432: This needs explanation (but I think it could be removed if you update the target mapping right here)

packages/firestore/src/local/local_store.ts
wu-hui: I had to add a `getSerializer` to localstore interface, but that still seems better than current.

wu-hui:
Applied with some minor fixes.

packages/firestore/exp/src/api/database.ts
```diff
@@ -332,9 +331,15 @@ export function loadBundle(
   const firestoreImpl = cast(firestore, Firestore);
   const resultTask = new LoadBundleTask();
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
-  getSyncEngine(firestoreImpl).then(syncEngine =>
-    enqueueLoadBundle(firestoreImpl._queue, syncEngine, bundleData, resultTask)
-  );
+  getSyncEngine(firestoreImpl).then(async syncEngine => {
+    enqueueLoadBundle(
+      (await firestoreImpl._getConfiguration()).databaseInfo.databaseId,
```
schmidt-sebastian, line 336: This style of code is pretty hard to debug (a lot of the magic happens inline). Do you mind extracting a local variable for this?

wu-hui: Done.

PR #3622, opened 2020-08-12
Create @firebase/app-compat and firebase/compat

Feiyang1:


packages-exp/app-compat/src/firebaseNamespaceCore.ts
```diff
@@ -0,0 +1,208 @@
...
+    return apps[name];
+  }
+
+  // @ts-ignore
+  app['App'] = firebaseAppImpl;
+
+  /**
+   * Create a new App instance (name must be unique).
+   */
+  function initializeAppCompat(
```
hsubox76, line 121: Doesn't accept a string as a second arg anymore?

packages-exp/app-compat/src/firebaseNamespaceCore.ts
```diff
@@ -0,0 +1,208 @@
...
+  // A namespace is a plain JavaScript Object.
+  const namespace: FirebaseNamespace = {
+    // Hack to prevent Babel from modifying the object returned
+    // as the firebase namespace.
+    // @ts-ignore
+    __esModule: true,
+    initializeApp: initializeAppCompat,
+    // @ts-ignore
+    app,
+    registerVersion,
```
hsubox76, line 67: No more `onLog()` and `setLogLevel()`?

packages-exp/app-compat/src/lite/firebaseAppLite.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  get name(): string {
+    return this.app.name;
+  }
+
+  get options(): FirebaseOptions {
+    return this.app.options;
+  }
+
+  delete(): Promise<void> {
```
hsubox76, line 61: Why is this one 2 lines while the one in firebaseApp has a 2 step Promise with a checkDestroyed check?

packages-exp/app-compat/test/firebaseCompat.test.ts
```diff
@@ -0,0 +1,365 @@
+/**
```
hsubox76, line 1: name: firebaseAppCompat.test.ts ?

packages-exp/firebase-exp/compat/index.cdn.ts
```diff
@@ -0,0 +1,45 @@
...
+(replace <PACKAGE> with the name of a component - i.e. auth, database, etc):
+
+https://www.gstatic.com/firebasejs/5.0.0/firebase-<PACKAGE>.js
+`);
+
+import '@firebase/polyfill';
+import firebase from './app';
+import { name, version } from '../package.json';
+
+// import './auth';
```
hsubox76, line 33: What's going to happen with these? No more all-in-one bundles? If not, do we need the warning anymore? Or is this just temporary during development?

packages-exp/app-compat/src/firebaseNamespaceCore.ts
Feiyang1: It does. `FirebaseNamespace` defines the signature for `firebase.initializeApp` which I didn't change.

packages-exp/app-compat/src/lite/firebaseAppLite.ts
Feiyang1: Just to make the light version smaller by sacrificing a little safety. I'm actually not sure if we still need a lite build for compat layer. I will keep it for now, but might delete it in the future.

packages-exp/app-compat/test/firebaseCompat.test.ts
Feiyang1: Renamed.

packages-exp/firebase-exp/compat/index.cdn.ts
Feiyang1: We will add them as their compat packages become available. We will have a compat all-in-one bundle.

packages/firestore/exp/src/api/components.ts
```diff
@@ -23,7 +23,7 @@ import {
   OfflineComponentProvider,
   OnlineComponentProvider
 } from '../../../src/core/component_provider';
-import {handleUserChange, LocalStore} from '../../../src/local/local_store';
+import { handleUserChange, LocalStore } from '../../../src/local/local_store';
```
hsubox76, line 26: Looks like this PR includes some prettier fixes not caught during firestore or rules-unit-testing commits somehow, do you want to include them here or in a separate PR?

PR #3623, opened 2020-08-12
Add environment check to Remote-Config Module

XuechunHou:
Issue 2393 fix for Remote-Config module.

Added `isSupported` method where users can call to check if indexedDB is supported before initializing the module. 

Fixes https://github.com/firebase/firebase-js-sdk/issues/5502

### DIFF
--- a/packages/firebase/compat/index.d.ts
+++ b/packages/firebase/compat/index.d.ts
@@ -2031,6 +2031,13 @@ declare namespace firebase.remoteConfig {
    * Defines levels of Remote Config logging.
    */
   export type LogLevel = 'debug' | 'error' | 'silent';
+  /**
+   * An async function that returns true if current browser context supports
+   * initialization of `RemoteConfig` instance (`firebase.remoteConfig()`).
+   *
+   * Returns false otherwise.
+   */
+  export function isSupported(): Promise<boolean>;
 }
 
 declare namespace firebase.functions {
--- a/packages/firebase/compat/index.d.ts
+++ b/packages/firebase/compat/index.d.ts
@@ -2032,10 +2032,13 @@ declare namespace firebase.remoteConfig {
    */
   export type LogLevel = 'debug' | 'error' | 'silent';
   /**
-   * Returns true if current browser context supports
-   * initialization of a `RemoteConfig` instance (`firebase.remoteConfig()`).
+   * This method provides two different checks:
    *
-   * Returns false otherwise.
+   * 1. Check if IndexedDB exists in the browser environment.
+   * 2. Check if the current browser context allows IndexedDB `open()` calls.
+   *
+   * It returns a `Promise` which resolves to true if a {@link RemoteConfig} instance
+   * can be initialized in this environment, or false if it cannot.
    */
   export function isSupported(): Promise<boolean>;
 }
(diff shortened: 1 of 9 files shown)


packages/remote-config/index.ts
```diff
@@ -139,3 +146,22 @@ declare module '@firebase/app-types' {
...
   }
 }
+/**
+ * this is a public static method provided to users that wraps two different checks:
+ *
+ * 1. check if IndexedDB is supported by the browser environment.
+ * 2. check if the current browser context is valid for using IndexedDB.
+ *
+ */
+async function isSupported(): Promise<boolean> {
```
erikeldridge, line 156: Thinking (no action req'd): the change to `registerRemoteConfig` will throw if IndexedDB is unavailable, in which case we'd need a method to call prior to registration, which is why `isSupported` is static, ie:

```
import * as remoteconfig from 'firebase/remote-config';
import * as firebase from 'firebase/app';

if (remoteconfig.isSupported()) {
  const rc = firebase.remoteconfig();
  ...
}
```

packages/remote-config/src/storage/storage.ts
```diff
@@ -75,28 +75,36 @@ type ProjectNamespaceKeyFieldValue =
...
+        // that correspond to those versions to run, not only the last one.
+        // eslint-disable-next-line default-case
+        switch (event.oldVersion) {
+          case 0:
+            db.createObjectStore(APP_NAMESPACE_STORE, {
+              keyPath: 'compositeKey'
+            });
+        }
+      };
+    } catch (error) {
```
erikeldridge, line 101: Thinking (no action req'd): we call `isIndexedDBAvailable` during registration and catch on-open errors here. `isSupported` provides a way to check both.

packages/firebase/index.d.ts
```diff
@@ -1772,6 +1772,15 @@ declare namespace firebase.remoteConfig {
    * Defines levels of Remote Config logging.
    */
   export type LogLevel = 'debug' | 'error' | 'silent';
+  /**
+   * An async function that returns true if current browser context supports initialization of remote config module
```
erikeldridge, line 1776: Minor: we know it's an async function from the return type; something like "Returns true if current browser context supports initialization of remote config module (`firebase.remoteConfig()`)" seems sufficient.

packages/firebase/index.d.ts
```diff
@@ -1772,6 +1772,15 @@ declare namespace firebase.remoteConfig {
    * Defines levels of Remote Config logging.
    */
   export type LogLevel = 'debug' | 'error' | 'silent';
+  /**
+   * Returns true if current browser context supports initialization of remote config module
```
Feiyang1, line 1776: @egilmorez Can you please take a look at it?

packages/firebase/compat/index.d.ts
```diff
@@ -2031,6 +2031,16 @@ declare namespace firebase.remoteConfig {
    * Defines levels of Remote Config logging.
    */
   export type LogLevel = 'debug' | 'error' | 'silent';
+  /**
```
hsubox76, line 2034: @egilmorez This is the only new block here, the rest is the automatic formatter cleaning up.

packages/remote-config/src/api2.ts
```diff
@@ -39,3 +43,26 @@ export async function fetchAndActivate(
   await fetchConfig(remoteConfig);
   return activate(remoteConfig);
 }
+
+/**
```
hsubox76, line 47: @eglimorez This is the other documentation block, it should be identical to the text in the compat/index.d.ts file.

packages/remote-config/index.ts
```diff
@@ -139,3 +146,22 @@ declare module '@firebase/app-types' {
...
+  if (!isIndexedDBAvailable()) {
+    return false;
+  }
+
+  try {
+    const isDBOpenable: boolean = await validateIndexedDBOpenable();
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/firebase/compat/index.d.ts
egilmorez: LG!

egilmorez (approved):
LG, thanks!

hsubox76:
I've rebased and updated this since it was last updated a year ago.

Feiyang1 (approved):

PR #3629, opened 2020-08-13
Parallelized builds

schmidt-sebastian:
This changes the Firestore build pipeline to run in parallel for most operations.

- First commit is the dist/ directory with the old files
- Second commit is the rollup changes
- Third commit is after re-running the build (with minimal changes as the Node ESM2017 build now uses 'const')
- Fourth commit is removal of the dist/ files

Dedicated to @samtstern


schmidt-sebastian:
The build doesn't seem faster on GA, but it is much faster for me locally. @samtstern - is it a lot of work to pull this into your branch that runs on GAE?

packages/firestore/rollup.config.exp.js
```diff
@@ -36,7 +37,7 @@ const nodePlugins = [
         target: 'es2017'
       }
     },
-    clean: true,
+    cacheDir: tmp.dirSync(),
```
Feiyang1, line 40: Does it work? seems it creates a new dir every time, so the cache is not shared across builds.

schmidt-sebastian: I believe we already didn't take advantage of cached build because we had to use "clean". IIRC, the build cache didn't work for us since we do not always use the same transformers, and the cache was not aware of what transformers were used to build it. If I leave "clean: true" but use the same cache dir, the builds fail as the clobber one directory.

PR #3637, opened 2020-08-14
Register auth with the app component. Add the getAuth() methods for each platform

samhorlbeck:


packages-exp/auth-exp/index.node.ts
```diff
@@ -40,8 +42,8 @@ FetchProvider.initialize(
 // Core functionality shared by all clients
 export * from './src';
 
-export const initializeAuth = _initializeAuthForClientPlatform(
-  ClientPlatform.NODE
-);
+export function getAuth(app?: FirebaseApp): Auth {
+  return initializeAuth(app);
```
avolkovi, line 46: what about deps?

packages-exp/auth-exp/index.rn.ts
```diff
@@ -36,8 +39,10 @@ export const reactNativeLocalPersistence = getReactNativePersistence(
   AsyncStorage
 );
 
-export const initializeAuth = _initializeAuthForClientPlatform(
-  ClientPlatform.REACT_NATIVE
-);
+export function getAuth(app?: FirebaseApp): Auth {
```
avolkovi, line 42: deps should be specified by user

packages-exp/auth-exp/demo/src/index.js
```diff
@@ -1640,10 +1641,7 @@ function onApplyAuthSettingsChange() {
 function initApp() {
   log('Initializing app...');
   app = initializeApp(config);
-  auth = initializeAuth(app, {
-    persistence: browserSessionPersistence,
-    popupRedirectResolver: browserPopupRedirectResolver
-  });
+  auth = getAuth(app);
```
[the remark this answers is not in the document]
avolkovi: nvm

packages-exp/auth-exp/index.node.ts
avolkovi: apparently you get both

packages-exp/auth-exp/index.ts
```diff
@@ -60,8 +65,11 @@ export { browserPopupRedirectResolver } from './src/platform_browser/popup_redir
 // MFA
 export { PhoneMultiFactorGenerator } from './src/platform_browser/mfa/assertions/phone';
 
-export const initializeAuth = _initializeAuthForClientPlatform(
-  ClientPlatform.BROWSER
-);
+export function getAuth(app?: FirebaseApp): Auth {
+  return initializeAuth(app, {
+    popupRedirectResolver: browserPopupRedirectResolver,
+    persistence: [indexedDBLocalPersistence, browserLocalPersistence]
```
avolkovi, line 71: why can't we do the same style of fallback for workers?

packages-exp/auth-exp/src/core/auth/register.ts
```diff
@@ -0,0 +1,72 @@
...
+            apiScheme: DEFAULT_API_SCHEME,
+            sdkClientVersion: _getClientVersion(clientPlatform)
+          };
+          return new AuthImpl(app.name, config);
+        })(app);
+      },
+      ComponentType.PUBLIC
+    )
+  );
+  registerVersion('auth-exp', version, getVersionForPlatform(clientPlatform));
```
avolkovi, line 71: the `auth-exp` should be a constant or pulled from packages.json

avolkovi: you use it above as well

packages-exp/auth-exp/index.ts
samhorlbeck: The "fallback" is doing different things. Here, it's saying "check these persistences in this order," whereas for workers, it's literally checking to see if it's available (if you tried to pass indexedDB into it when it wasn't available, the instantiation would just error out)

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -15,24 +15,26 @@
  * limitations under the License.
  */
 
+import '@firebase/installations';
```
Feiyang1, line 18: You don't need installations.

packages-exp/auth-exp/index.ts
Feiyang1: Wouldn't auth fallback to memory persistence when indexedDB is not available. Why is it different for workers?

packages-exp/auth-exp/package.json
```diff
@@ -39,6 +39,7 @@
     "@firebase/app-types-exp": "0.x",
     "@firebase/auth-types-exp": "0.x",
     "@firebase/auth-exp": "0.x",
+    "@firebase/component": "0.x",
```
Feiyang1, line 42: These shouldn't be a peerDepenencies except for `app-exp` and `app-types-exp`. And they should use exact versions .

packages-exp/auth-exp/src/core/index.ts
```diff
@@ -54,6 +54,8 @@ export function signOut(auth: externs.Auth): Promise<void> {
   return auth.signOut();
 }
 
+export { initializeAuth } from './auth/initialize';
```
Feiyang1, line 57: What is the entry point for it? `@firebase/auth/core`?

packages-exp/auth-exp/index.ts
samhorlbeck: Currently it would error. It probably shouldn't but we need to clarify that separately. For now I'd like to move ahead and deal with that issue in a separate PR. I'm tracking this in b/165376704

packages-exp/auth-exp/src/core/index.ts
samhorlbeck: No entry point for it, it's only used as a building block for the other entry points

packages-exp/auth-compat-exp/src/auth.ts
samhorlbeck: Done

packages-exp/auth-exp/package.json
samhorlbeck: Added component but I think https://github.com/firebase/firebase-js-sdk/pull/3654 does the rest

packages-exp/auth-exp/src/core/index.ts
Feiyang1: `initializeAuth` is a public API in the original proposal, which allows people to configure things, e.g. persistence and resolvers. Is that changed?

samhorlbeck: No. It's just the way our exports go:

`src/core/index.ts` re-exports all of the things in core
`src/index.ts` just does `export * from './core'`;
`index.ts` does `export * from './src';` as well as the browser-specific APIs

packages-exp/auth-exp/src/core/auth/register.ts
samhorlbeck: Done

PR #3654, opened 2020-08-18
Auth next build 

Feiyang1:
1. Removed unused dependencies and files
2. Moved most peerDependencies to dependencies
3. Updated the build to share code between the internal and the public entry points.


packages-exp/auth-compat-exp/src/user.ts
```diff
@@ -47,7 +47,7 @@ export class User extends impl.UserImpl implements compat.User {
     credential: compat.AuthCredential
   ): Promise<compat.UserCredential> {
     return convertCredential(
-      this.auth,
+      (this.auth as unknown) as externs.Auth,
```
Feiyang1, line 50: @avolkovi Are these castings safe? I think you mentioned we never use the public APIs internally.

avolkovi: there are three methods that have a changed signature in the new API vs the old one, we don't ever call those directly internally so it should be safe, I do a similar cast in auth-compat-exp

PR #3655, opened 2020-08-18
Add the internal auth interface including proactive refresh

samhorlbeck:


packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -271,6 +275,20 @@ export class AuthImplCompat<T extends User> implements Auth, _FirebaseService {
     return `${this.config.authDomain}:${this.config.apiKey}:${this.name}`;
   }
 
+  _startProactiveRefresh(): void {
+    this.isProactiveRefreshEnabled = true;
```
avolkovi, line 279: do we need this on the auth object? or should it just live exclusively on the user?

packages-exp/auth-exp/src/core/auth/firebase_internal.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      sinon.stub(user.stsTokenManager, 'getToken').returns(
+        Promise.resolve({
+          accessToken: 'access-token',
+          refreshToken: 'refresh-tken',
+          wasRefreshed: true
+        })
+      );
+      sinon
+        .stub(user, '_startProactiveRefresh')
+        .callsFake(() => (isProactiveRefresh = true));
```
avolkovi, line 82: can these just be sinon.spy()'s and you check to make sure they were called?

packages-exp/auth-exp/src/core/auth/register.ts
```diff
@@ -24,13 +24,16 @@ import { AuthErrorCode } from '../errors';
...
+  _castAuth,
   AuthImpl,
   DEFAULT_API_HOST,
   DEFAULT_API_SCHEME,
   DEFAULT_TOKEN_API_HOST
 } from './auth_impl';
+import { AuthInternal } from './firebase_internal';
 
 export const _AUTH_COMPONENT_NAME = 'auth-exp';
+export const _AUTH_INTERNAL_COMPONENT_NAME = 'auth-internal-exp';
```
avolkovi, line 36: is this right? or should it be auth-interop?

packages-exp/auth-exp/src/core/auth/register.ts
```diff
@@ -70,6 +73,20 @@ export function registerAuth(clientPlatform: ClientPlatform): void {
       ComponentType.PUBLIC
     )
   );
+
+  _registerComponent(
+    new Component(
+      _AUTH_INTERNAL_COMPONENT_NAME,
+      container => {
+        const auth = _castAuth(
+          container.getProvider('auth-exp').getImmediate()!
```
avolkovi, line 82: here's that auth-exp again, we should put it in a variable

packages-exp/auth-exp/src/core/user/proactive_refresh.ts
```diff
@@ -0,0 +1,99 @@
...
+// Refresh the token five minutes before expiration
+export const _OFFSET_DURATION = 5 * 1000 * 60;
+
+export const _RETRY_BACKOFF_MIN = 30 * 1000;
+export const _RETRY_BACKOFF_MAX = 16 * 60 * 1000;
+
+export class ProactiveRefresh {
+  private isRunning = false;
+
+  // Node timers and browser timers return fundamentally different types.
```
avolkovi, line 30: i have a similar issue in my PR, do you think it's worth extracting the setInterval logic into a common util?

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: Auth needs to know about it because start/stop are tightly coupled with the current user lifecycle.

packages-exp/auth-exp/src/core/auth/firebase_internal.test.ts
samhorlbeck: I like this better since I'm testing as it gets toggled back and forth. I suppose you could do it by counting the number of times start was called and how many stop was called, but that seems more complicated

packages-exp/auth-exp/src/core/auth/register.ts
samhorlbeck: @Feiyang1 for guidance. I believe internal is correct since that's what we do now, but maybe we're changing our language?
https://github.com/firebase/firebase-js-sdk/search?q=%27auth-internal%27&unscoped_q=%27auth-internal%27

packages-exp/auth-exp/src/core/auth/register.ts
samhorlbeck: If only it was already defined in a variable...

Done

packages-exp/auth-exp/src/core/user/proactive_refresh.ts
samhorlbeck: Eh that would mean adding more code that is compiled deadweight just to circumvent the type system.

packages-exp/auth-exp/src/core/auth/firebase_internal.ts
```diff
@@ -0,0 +1,86 @@
...
+      return;
+    }
+
+    this.internalListeners.delete(listener);
+    unsubscribe();
+    this.updateProactiveRefresh();
+  }
+
+  private updateProactiveRefresh(): void {
+    if (this.internalListeners.size > 0) {
```
Feiyang1, line 80: Should we actively refresh for external listeners as well?

packages-exp/auth-exp/src/core/auth/register.ts
Feiyang1: `auth-interna` is correct

Feiyang1:
The registration part for auth-internal looks good to me.

packages-exp/auth-exp/src/core/auth/firebase_internal.ts
samhorlbeck: We've definitely had feature requests for it. I think it should still be an optional thing you turn on, exposed through a method on user or something

Feiyang1: Are we going to do it, or this is something for the future?

samhorlbeck: I can add it to the new API proposal and see what kind of feedback we get

PR #3657, opened 2020-08-18
Add API Extractor for @firestore/lite + exp

schmidt-sebastian:
This adds API Extractor to firestore lite. Fingers crossed that we can get the generated and the automated types to match (via follow-up cleanup).


packages/firestore/lite/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@firebase/firestore/lite",
+  "name": "@firebase/firestore-lite",
```
schmidt-sebastian, line 2: API Extractor considers this name invalid - looks like one slash is the maximum.

package.json
```diff
@@ -136,6 +135,7 @@
     "ora": "4.0.5",
     "prettier": "2.0.5",
     "protractor": "5.4.2",
+    "rollup-plugin-copy": "3.3.0",
```
schmidt-sebastian, line 138: rollup-plugin-copy-asset doesn't allow me to move the proto files to dist/lite/src/proto

package.json
```diff
@@ -66,8 +66,7 @@
     "@changesets/changelog-github": "0.2.6",
     "@changesets/cli": "2.9.2",
     "@microsoft/api-documenter": "7.8.21",
-    "@microsoft/api-extractor": "7.9.2",
-    "@types/mz": "2.7.1",
+    "@microsoft/api-extractor": "7.8.2-pr1796.0",
```
schmidt-sebastian, line 69: This pulls in https://github.com/microsoft/rushstack/issues/1029, which makes this diff about 500 lines smaller (see the revert https://github.com/firebase/firebase-js-sdk/pull/3657/commits/609b2b3395b8b6988aed6996e45dc1f87a12e842)

package.json
```diff
@@ -66,8 +66,7 @@
     "@changesets/changelog-github": "0.2.6",
     "@changesets/cli": "2.9.2",
     "@microsoft/api-documenter": "7.8.21",
-    "@microsoft/api-extractor": "7.9.2",
-    "@types/mz": "2.7.1",
```
schmidt-sebastian, line 70: This is in here twice.

packages/firestore/lite/api-extractor.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../../config/api-extractor.json",
+  "mainEntryPointFilePath": "../dist/lite/firestore/lite/index.d.ts",
+  "dtsRollup": {
+    "enabled": true,
```
Feiyang1, line 5: Do you think you will ever need main and lite to share declarations? There is a risk where the same type from main and lite becomes incompatible if they use separate rollup d.ts files - see discussion [here](https://github.com/microsoft/rushstack/issues/1596#issuecomment-546790721). It's probably not a problem for us, since we are already using separate d.ts files for lite and main, but it's good to keep that in mind.

package.json
Feiyang1: eh, the commit shows `445 additions and 477 deletions`, so the net gain seems to be 32 lines?

packages/firestore/lite/package.json
Feiyang1: Api-extractor doesn't support multiple entry points, so you need to trick it to think that lite is a package instead of a submodule. 

(I have a [PR](https://github.com/microsoft/rushstack/pull/1932) to support multiple entry points, but only for documentation and api report)

packages/firestore/package.json
```diff
@@ -7,14 +7,15 @@
   "description": "The Cloud Firestore component of the Firebase JS SDK.",
   "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
   "scripts": {
+    "api-report:lite": "(cd lite; api-extractor run --local --verbose)",
```
Feiyang1, line 10: Once this [PR](https://github.com/microsoft/rushstack/pull/1932) is merged, we can generate api reports for multiple entry points together, so you won't generate report for lite and main separately.
It seems it might take a long time to merge though, so I might publish a forked version to npm which we will use. For now, this is okay.

packages/firestore/rollup.config.lite.js
```diff
@@ -41,7 +43,23 @@ const nodePlugins = [
...
-  json({ preferConst: true })
+  json({ preferConst: true }),
+  copy({
+    targets: [
+      {
+        src: 'src/protos',
+        dest: 'dist/lite/src'
+      },
+      {
+        // Copy into generated source files to support API Extractor
```
Feiyang1, line 54: Why does API extractor need a different path?

schmidt-sebastian (commented):
Some follow up questions.

I planned to do this in a follow-up to for /exp. I realized that we are not actually blocked because of the behavior differences that I would like to clarify in the API documentation, since the APIs that have different behavior are already implemented twice (e.g. getDoc). I pushed the changes to this PR now since without them the build was broken.

packages/firestore/lite-types/index.d.ts
```diff
@@ -365,31 +365,6 @@ export function snapshotEqual<T>(
   right: DocumentSnapshot<T> | QuerySnapshot<T>
 ): boolean;
 
-export type FirestoreErrorCode =
```
schmidt-sebastian, line 368: The Lite SDK doesn't use FirestoreError.

package.json
schmidt-sebastian: FWIW, I created a new PR to merge just this: https://github.com/firebase/firebase-js-sdk/pull/3660

packages/firestore/lite/package.json
Feiyang1: The name actually doesn't matter, so it can be whatever that's easier for you. We use`/` to be consistent with how developer would use the submodules, so I prefer to stay this way. Does it give you any issue other than unpleasant validation error?

schmidt-sebastian: I get the following error:

```
Error: Error parsing /Users/mrschmidt/GitHub/firebase/firebase-js-sdk/packages/firestore/lite/api-extractor.json:
The package name "@firebase/firestore/lite" contains an invalid character: "/"
```

The build stops after this message.

Feiyang1: oh right, api-extractor definitely doesn't like it. You can change it to `-` to work around it for now. I'm planning to release a forked version of api-extractor with support for multiple entry points. We will come back to this once that's done.

schmidt-sebastian:
PR updated to undo the api-extractor upgrade.

schmidt-sebastian:
Abandoning.

PR #3660, opened 2020-08-19
Remove all namespace imports

schmidt-sebastian:
This would unblock https://github.com/firebase/firebase-js-sdk/pull/3657 without the API Extractor update.


packages/firestore/src/protos/firestore_proto_api.js
```diff
@@ -0,0 +1,18 @@
+/**
```
Feiyang1, line 1: Do you still need this file?

schmidt-sebastian: Yup, it breaks the TypeScript compile even without the API Extractor.

PR #3663, opened 2020-08-19
Fix canary release flow

hsubox76:
Add new rules-unit-testing token for canary release and separate canary workflow from test workflow.


.github/workflows/canary-deploy.yml
```diff
@@ -0,0 +1,61 @@
...
+        # Canary release script requires git history and tags.
+        fetch-depth: 0
+    - name: Set up Node (10)
+      uses: actions/setup-node@v1
+      with:
+        node-version: 10.x
+    - name: Yarn install
+      run: yarn
+    - name: Deploy canary
+      run: yarn release --canary
```
Feiyang1, line 26: skipTest?

hsubox76: It already skips tests, cli script branches to canary release flow before tests are run.

PR #3665, opened 2020-08-19
Show CI logs if run is longer than 1 hour

hsubox76:


scripts/run_tests_in_ci.js
```diff
@@ -52,10 +57,14 @@ const argv = yargs.options({
 
     await testProcess;
     console.log('Success: ' + name);
+    appendFileSync(LOGFILE, stdout + '\n' + stderr, { encoding: 'utf8' });
```
Feiyang1, line 60: Will there be any race condition ?

hsubox76: There will absolutely be race conditions. I changed it to write multiple files.

PR #3673, opened 2020-08-24
Remove the duplicate dependency.

yifanyang:


package.json
```diff
@@ -68,7 +68,6 @@
     "@changesets/cli": "2.9.2",
     "@microsoft/api-documenter": "7.8.21",
     "@microsoft/api-extractor": "7.9.2",
-    "@types/mz": "2.7.1",
```
yifanyang, line 71: Duplicate with line 79.

PR #3677, opened 2020-08-25
Rename Blob to Bytes

schmidt-sebastian:
As part of go/firestore-next-amendment-amendment, we decided to rename Blob to Bytes to avoid name clashes with the JavaScript Blob type.

Couple notes:
- I created a new type instead of renaming the export since the error messages would have been incorrect. 
- I removed the now incorrect error messages in the new type.
- I believe all of our platforms now support byte conversion (since we have the Polyfill for ReactNative, which was required to make stream resumption work). Hence, I did not port the validation.
- The old Blob type now extends Bytes to make it easier for the serializer to use common logic.

### DIFF
--- a/packages/firestore/src/api/blob.ts
+++ b/packages/firestore/src/api/blob.ts
@@ -23,6 +23,7 @@ import {
   validateExactNumberOfArgs
 } from '../util/input_validation';
 import { ByteString } from '../util/byte_string';
+import { Bytes } from '../../lite/src/api/bytes';
 
 /** Helper function to assert Uint8Array is available at runtime. */
 function assertUint8ArrayAvailable(): void {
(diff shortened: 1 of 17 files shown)


packages/firestore/src/api/blob.ts
```diff
@@ -51,16 +52,7 @@ function assertBase64Available(): void {
...
-  // Prefix with underscore to signal that we consider this not part of the
-  // public API and to prevent it from showing up for autocompletion.
-  _byteString: ByteString;
-
-  constructor(byteString: ByteString) {
-    assertBase64Available();
-    this._byteString = byteString;
-  }
-
+export class Blob extends Bytes {
```
thebrianchen, line 55: optional: It might be helpful to add some context in comments about why the main SDK's `Blob` is extending the `Bytes` class in the Lite SDK. It's clear to me what's happening when I look at the PR description, but without it, I would be confused.

schmidt-sebastian: Done (but missed this before merging - coming in the next PR).

PR #3679, opened 2020-08-25
Move parent/doc/collection back

schmidt-sebastian:
As part of go/firestore-next-amendment-amendment, we decided to make `parent` once again a property of CollectionReference and DocumentReference and also to re-add `doc()` and `collection()` method to these classes (with the old methods staying as is).

Tree-shaking will be improved with https://github.com/firebase/firebase-js-sdk/pull/3678


packages/firestore/lite/src/api/reference.ts
```diff
@@ -382,6 +402,31 @@ export class CollectionReference<T = firestore.DocumentData> extends Query<T>
...
   }
 
+  get parent(): DocumentReference<firestore.DocumentData> | null {
+    const parentPath = this._path.popLast();
+    if (parentPath.isEmpty()) {
+      return null;
+    } else {
+      return new DocumentReference(
+        this.firestore,
+        /* converter= */ null,
```
thebrianchen, line 412: nit: Can you standardize the constructor argument order for `DocumentReference` and `CollectionReference`?

thebrianchen: I missed this in an earlier review when you made the converter nullable, but the `converter` in the `CollectionReference` constructor can be `readonly`.

packages/firestore/lite/src/api/reference.ts
```diff
@@ -753,3 +768,16 @@ export function newUserDataReader(firestore: Firestore): UserDataReader {
     serializer
   );
 }
+
+function validateNonEmptyArgument(
+  functionName: string,
+  argumentName: string,
+  argument?: string
+): asserts argument is string {
```
thebrianchen, line 776: ooooooh this is fancy :)

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: Done (changed CollectionReference to match). As for `readonly`, that is done in the parent class Query and we rely on the Query's converter property here.

PR #3687, opened 2020-08-26
Analytics integration tests

hsubox76:
Add some simple integration smoke tests to analytics dynamic measurement ID PR. Should this run in CI? Which workflow?

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -18,7 +18,7 @@
     "test": "run-p lint test:browser",
     "test:ci": "node ../../scripts/run_tests_in_ci.js -s test:browser",
     "test:browser": "karma start --single-run --nocache",
-    "test:integration": "karma start ./testing/e2e/karma.conf.js --single-run --nocache",
+    "test:integration": "karma start ./karma.integration.conf.js --single-run --nocache",
     "prepare": "yarn build"
   },
   "peerDependencies": {
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -15,8 +15,9 @@
     "build": "rollup -c",
     "build:deps": "lerna run --scope @firebase/analytics --include-dependencies build",
     "dev": "rollup -c -w",
-    "test": "run-p lint test:browser test:integration",
-    "test:ci": "node ../../scripts/run_tests_in_ci.js -s test",
+    "test": "run-p lint test:all",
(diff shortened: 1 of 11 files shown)


packages/analytics/testing/integration-tests/integration.ts
```diff
@@ -0,0 +1,52 @@
...
+    firebase.initializeApp(config);
+    firebase.analytics().logEvent('login');
+    await new Promise(resolve => setTimeout(resolve, 10000));
+    const resources = performance.getEntriesByType('resource');
+    const callsWithEvent = resources.filter(
+      resource =>
+        resource.name.includes('google-analytics.com') &&
+        resource.name.includes('en=login')
+    );
+    expect(callsWithEvent.length).to.be.at.least(1);
```
Feiyang1, line 39: Should it be exactly once?

hsubox76: Changed.

hsubox76:
> Can you try to include integration tests in the `yarn test` script? I think we can just add them to the normal test run if they runs reasonably fast.

Do you think there's any circumstances where it's bad that the test makes a live network request? Or is it assumed anyone running the test will be online?

Feiyang1:
> > Can you try to include integration tests in the `yarn test` script? I think we can just add them to the normal test run if they runs reasonably fast.
> 
> Do you think there's any circumstances where it's bad that the test makes a live network request? Or is it assumed anyone running the test will be online?

I think it's okay. We can add a script that only runs unit tests to mitigate the concern.

packages/analytics/package.json
```diff
@@ -15,9 +15,10 @@
     "build": "rollup -c",
     "build:deps": "lerna run --scope @firebase/analytics --include-dependencies build",
     "dev": "rollup -c -w",
-    "test": "run-p lint test:browser",
-    "test:ci": "node ../../scripts/run_tests_in_ci.js -s test:browser",
+    "test": "run-p lint test:browser test:integration",
+    "test:ci": "node ../../scripts/run_tests_in_ci.js -s test",
```
Feiyang1, line 19: `test:ci` should not run lint. You can create a `test:all` script that doesn't do lint and to be used by `test` and `test:ci`

hsubox76: Done.

PR #3700, opened 2020-08-28
Don't switch network status to UNKNOWN during user change

schmidt-sebastian:
The user change handling on Web resets the OnlineStatus to unknown even if we are offline. This breaks offline behavior for empty snapshots from cache, as we will stop raising empty snapshots.

This fixes a regression in the Web client. The Android and iOS client should have the correct behavior (see Android: https://github.com/firebase/firebase-android-sdk/blob/751684d3073329adef4cbca8865efa22c953fa5d/firebase-firestore/src/main/java/com/google/firebase/firestore/remote/RemoteStore.java#L319)

### DIFF
--- a/.changeset/cold-impalas-drop.md
+++ b/.changeset/cold-impalas-drop.md
@@ -1,0 +1,5 @@
+---
+"@firebase/firestore": patch
+---
+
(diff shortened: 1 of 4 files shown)


schmidt-sebastian:
@dconeybe I had to change the code because it caused the IndexedDB retry to fail.

PR #3701, opened 2020-08-28
Schedule everything on the AsyncQueue

schmidt-sebastian:
This PR fixes a scheduling problem in the firestore-exp client. Before this change, most methods looked like this:

```
getComponent().then(() => enqueueOp())
```

That meant that `getComponent()` did not run on the AsyncQueue. During a test I noticed that the following did not issue a write:

```
setDoc(docRef, {});
terminate(firestore);
```

This was because `terminate` did not use `getComponent()` and hence was able to schedule its task on the async queue before `setDoc()`.

This is a large PR but I am out next week so take your time :)


packages/firestore/exp/src/api/database.ts
```diff
@@ -325,7 +337,7 @@ export function terminate(
 }
 
 function verifyNotInitialized(firestore: Firestore): void {
-  if (firestore._initialized) {
+  if (firestore._initialized || firestore._terminated) {
```
var-const, line 340: Question: what's the reason to check for `_terminated`? Can an instance be terminated without being initialized? If yes, consider adding this bit to the error message below.

var-const (approved):

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: If a user calls "terminate()" on an uninitialized Firestore, we should also not initialize it. The error message is actually shared between the new SDK and the old SDK and hence, I would rather not touch it.

hsubox76 (approved):

PR #3705, opened 2020-08-28
Tree-Shake RemoteStore

schmidt-sebastian:
This is purely mechanical change with no logic changes.

Goal is to make diff of https://github.com/firebase/firebase-js-sdk/pull/3568 smaller.


packages/firestore/src/core/component_provider.ts
```diff
@@ -37,7 +37,13 @@ import {
   newSyncEngine,
   SyncEngine
 } from './sync_engine';
-import { RemoteStore } from '../remote/remote_store';
+import {
+  fillWritePipeline,
```
wu-hui, line 41: Remove if this lint error is correct.

PR #3709, opened 2020-08-31
console build for rtdb

Feiyang1:


packages/database/src/api/internal.ts
```diff
@@ -67,3 +81,52 @@ export const interceptServerData = function (
...
+}: {
+  app: FirebaseApp;
+  url: string;
+  version: string;
+  customAuthImpl: FirebaseAuthInternal;
+  namespace: T;
+}): {
+  instance: types.Database;
+  namespace: T;
+} {
```
schmidt-sebastian, line 109: I am not married to the original way this was implemented, but I don't see much benefit of splitting this code across two different functions that are both named the same. Reducing this back down to one method would also remove the options object, which marginally improves readability but adds some duplication to ensure type safety. What do you think of moving this code back?

Feiyang1: I moved it here to make it available via `firebase.database.INTERNAL.initStandalone`, so console can consume the normal cdn script without needing a special build target. 

https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/index.node.ts#L139

schmidt-sebastian (approved):

PR #3710, opened 2020-08-31
Stop redirecting to non-origin sites

zwu52:
FM should not open non-origin links.  

for more context: go/fcm-security-issue

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -246,15 +246,25 @@ export class SwController implements FirebaseMessaging, FirebaseService {
     event.stopImmediatePropagation();
     event.notification.close();
 
+    // Note clicking on a notification with no link set will focus the Chrome's current tab.
     const link = getLink(internalPayload);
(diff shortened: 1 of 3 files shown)


packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -246,15 +246,25 @@ export class SwController implements FirebaseMessaging, FirebaseService {
...
     if (!link) {
       return;
     }
 
-    let client = await getWindowClient(link);
+    // FM should only open/focus links from app's origin.
+    const url = new URL(link, self.location.href);
+    const originUrl = new URL(self.location.origin);
+
+    if (url.host !== originUrl.host) {
```
Feiyang1, line 259: Wouldn't it be a breaking change

Feiyang1: I guess it's not if you consider it as a security bug.

zwu52: Yay 🐛

PR #3714, opened 2020-09-01
Add '@override' annotation to RecaptchaVerifier.Verify method in AuthClient externs.

richieforeman:
Adds an `@override` annotation on the RecaptchaVerifier.Verify in AuthClient externs

--

This is needed because this interface implements the `firebase.auth.ApplicationVerifier` type.

This resolves a Closure compiler warning message

tag @hsubox76

### DIFF
--- a/packages/firebase/externs/firebase-firestore-externs.js
+++ b/packages/firebase/externs/firebase-firestore-externs.js
@@ -755,7 +755,7 @@ firebase.firestore.DocumentSnapshot.prototype.metadata;
 
 /**
  * An object containing all the data in a document.
- * @type {Object}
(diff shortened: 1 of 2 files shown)


packages/firebase/externs/firebase-firestore-externs.js
```diff
@@ -755,7 +755,7 @@ firebase.firestore.DocumentSnapshot.prototype.metadata;
 
 /**
  * An object containing all the data in a document.
- * @type {Object}
+ * @typedef {Object}
```
hsubox76, line 758: I think this was also done in https://github.com/firebase/firebase-js-sdk/pull/3713

PR #3716, opened 2020-09-01
Add hooks into auth-next for emulator config

samhorlbeck:


yuchenshi:
@samhorlbeck This PR has 93 commits and 273 changed files. Did you mean to rebase?

packages-exp/auth-exp/src/core/auth/initialize.ts
```diff
@@ -34,6 +36,24 @@ export function initializeAuth(
...
 }
 
+export function useEmulator(
+  authExtern: externs.Auth,
+  hostname: string,
+  port: number
+): void {
+  const auth = _castAuth(authExtern);
+  if (!auth._canInitEmulator) {
+    fail(AuthErrorCode.EMULATOR_CONFIG_FAILED, {
```
avolkovi, line 46: why not assert instead of fail since you have the if block?

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -182,6 +187,16 @@ export async function _performSignInRequest<T, V extends IdTokenResponse>(
   return serverResponse;
 }
 
+export function _getFinalTarget(auth: AuthCore, url: string): string {
+  const { emulator } = auth.config;
+  if (!emulator) {
+    return url;
+  }
+
+  const urlWithoutScheme = url.replace(/^.*:\/\//, '');
```
avolkovi, line 196: instead of doing string regex, can you just pass all the parts to this function and have it combine everything once figuring out the correct host?

packages-exp/auth-exp/src/core/auth/initialize.ts
samhorlbeck: Because I'm silly

packages-exp/auth-exp/src/api/index.ts
samhorlbeck: Done

samhorlbeck:
Right: the API proposal doesn't cover the new modular design so I'm just going on what we've done for the other non-core-functionality (namely putting the target object as the first parameter).

On the other hand, all of these are going to be exporting a "useEmulator" so it probably makes sense to put it on the auth object directly. I will discuss with the core SDK team.

Compat will come in a later PR. We need to do some refactoring there

PR #3717, opened 2020-09-01
Fix the broken build.

samhorlbeck:


packages-exp/auth-exp/demo/public/index.html
```diff
@@ -1,763 +0,0 @@
-<!DOCTYPE html>
```
avolkovi, line 1: you're removing the demo?

samhorlbeck: woops, just meant to get rid of the index.js and index.js.map

PR #3724, opened 2020-09-01
Adding experimentalAutoDetectLongPolling

rafikhan:
Adding support for `experimentalAutoDetectLongPolling`.  This PR includes validation to ensure mutual exclusive use with `experimentalForceLongPolling` flag and a unit test to ensure that behavior.

In this PR the default value is `true` so that the code path gets hit during test runs; however, we don't want this in release builds yet. Need feedback on how to configure this correctly.


packages/firestore/src/api/database.ts
```diff
@@ -146,6 +147,7 @@ const DEFAULT_HOST = 'firestore.googleapis.com';
 const DEFAULT_SSL = true;
 const DEFAULT_TIMESTAMPS_IN_SNAPSHOTS = true;
 const DEFAULT_FORCE_LONG_POLLING = false;
+const DEFAULT_FORCE_AUTO_DETECT_LONG_POLLING = true;
```
schmidt-sebastian, line 150: Please set this to false and use https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/integration/util/helpers.ts#L160 to pass custom settings.

packages/firestore/src/util/input_validation.ts
```diff
@@ -199,6 +199,28 @@ export function validateNamedOptionalType(
...
+): void {
+  if (
+    argument1 !== undefined &&
+    argument2 !== undefined &&
+    argument1 === true &&
+    argument2 === true
+  ) {
+    throw new FirestoreError(
+      Code.INVALID_ARGUMENT,
+      `${optionName1} and ${optionName2} can not be used together.`
```
schmidt-sebastian, line 219: Suggestion: s/can not/cannot/

packages/firestore/src/util/input_validation.ts
```diff
@@ -199,6 +199,28 @@ export function validateNamedOptionalType(
...
+  argument1: boolean | undefined,
+  optionName2: string,
+  argument2: boolean | undefined
+): void {
+  if (
+    argument1 !== undefined &&
+    argument2 !== undefined &&
+    argument1 === true &&
+    argument2 === true
+  ) {
```
schmidt-sebastian, line 216:
```suggestion
  if (
    argument1 === true &&
    argument2 === true
  ) {
```

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -231,7 +231,8 @@ abstract class TestRunner {
       TEST_PERSISTENCE_KEY,
       'host',
       /*ssl=*/ false,
-      /*forceLongPolling=*/ false
+      /*forceLongPolling=*/ false,
+      /*autoDetectLongPolling=*/ true
```
schmidt-sebastian, line 235: I would use the default value here, since it matches the expected usage for most users.

schmidt-sebastian (commented):
This core logic looks good.

Couple more homework items: 
- This should have an integration test that uses this flag.
- There should be a chanelog entry.
- Please undo all unrelated formatting changes.

packages/firebase/index.d.ts
```diff
@@ -7879,15 +7879,27 @@ declare namespace firebase.firestore {
...
      */
     experimentalForceLongPolling?: boolean;
 
+    /**
+     * Configures the SDK's underlying transport (WebChannel) to automatically detect if
+     * long-polling should be used and also update as the networking conditions change. This is
+     * very similar to experimentalForceLongPolling but does automatic detection.
+     *
+     * This setting cannot be used with experimentalForceLongPolling and may be removed in a future
+     * release.
```
schmidt-sebastian, line 7897: Maybe:

```suggestion
     * Configures the SDK's underlying transport (WebChannel) to automatically detect if
     * long-polling should be used. This is very similar to `experimentalForceLongPolling`,
     * but only uses long-polling if required.
     *
     * This setting will likely be enabled by default in future releases and cannot be 
     * combined with `experimentalForceLongPolling`.
```

packages/firestore/test/unit/specs/spec_test_runner.ts
rafikhan: The constant I declared in `database.ts` isn't exported. Does that mean that single variable should be in order to reference it here?

packages/firestore/src/api/database.ts
rafikhan: Should I set this for both the emulator and prod settings in https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/test/integration/util/settings.ts#L35 or should I be passing this in differently?

rafikhan:
> Please undo all unrelated formatting changes.

@schmidt-sebastian - How do I do this? This seems to happen as a pre-commit hook.

packages/firestore/test/unit/specs/spec_test_runner.ts
schmidt-sebastian: You can just hardcode "false" as we did for "forceLongPolling" above. Ideally, we should remove both settings pretty soon.

packages/firestore/src/api/database.ts
schmidt-sebastian: This is the only place you need to set it to false. What you have now applies to all settings, unless it is overridden by the user.

schmidt-sebastian (commented):
Thanks for the updates. Please do remember to add an integration test that sets this value to `true` and does some basic operations (in test/api/integration). This will then ensure that we test this setting in Browser environments.

As for undoing the formatting changes. I would sync `master` into this branch. This will get you a different version of `prettier`, so you have to go to repo root and run `yarn`. Once that is done, most of the formatting changes should be gone in the next commit. You will likely still have changes in `packages/firestore/script` so you might have to do `git checkout master packages/firestore/script`. These files should then be ignored by prettier (Christina just added them to .prettierignore). 

By the way, if you want to opt out of formatting because something screwed up you can skip the pre-commit hook by passing `--no-verify`.

packages/firestore/src/core/database_info.ts
```diff
@@ -29,13 +29,16 @@ export class DatabaseInfo {
    * @param ssl Whether to use SSL when connecting.
    * @param forceLongPolling Whether to use the forceLongPolling option
    * when using WebChannel as the network transport.
+   * @param autoDetectLongPolling Whether to use the autoDetectLongPolling
```
schmidt-sebastian, line 32:
```suggestion
   * @param autoDetectLongPolling Whether to use the detectBufferingProxy
```

This matches the name of the WebChannel option.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1600,4 +1601,31 @@ apiDescribe('Database', (persistence: boolean) => {
...
+
+    return withTestDbsSettings(
+      persistence,
+      DEFAULT_PROJECT_ID,
+      settings,
+      1,
+      async ([db]) => {
+        const doc = await db.collection('users').doc();
+
+        const data = { name: 'Rafi', email: 'abc@xyz.com' };
```
schmidt-sebastian, line 1619:
```suggestion
        const data = { name: 'Rafi', email: 'abc@xyz.com' };
        const doc = await db.collection('users').doc();
```

... but this is just a personal preference that you may ignore :)

schmidt-sebastian (approved):
Very nice! Landing 2 days before the quarter cutoff too :) 

Can you follow the link that "changeset-bot" posted and add a short changelog?

rafikhan:
@schmidt-sebastian - Thanks for the reminder on changesets. I've added the file but am not sure if this is a patch release vs minor.

.changeset/yellow-turkeys-lay.md
```diff
@@ -0,0 +1,11 @@
+---
+"firebase": patch
+"@firebase/firestore-types": patch
+"@firebase/firestore": patch
+"@firebase/webchannel-wrapper": patch
```
schmidt-sebastian, line 5:
```suggestion
"@firebase/firestore-types": minor
"@firebase/firestore": minor
"@firebase/webchannel-wrapper": minor
```

We generally no longer include the `firebase` package in this list, since the version bump for the main package will be determined based on the individual version bumps. 

As for whether this is a `patch` or `minor` bump: Any additions to public API surface (or any new feature work that is more than a bug fix) warrant a `minor` release. I think it is safe to say that if the public interface is touched, we always have to bump at least the `minor` version.

schmidt-sebastian (approved):

.changeset/yellow-turkeys-lay.md
```diff
@@ -0,0 +1,10 @@
+---
+"@firebase/firestore-types": minor
+"@firebase/firestore": minor
+"@firebase/webchannel-wrapper": minor
```
hsubox76, line 4: Can you add `"firebase": minor` here to minor bump the main package?

hsubox76:
Added @egilmorez as reviewer for documentation change to packages/firebase/index.d.ts

packages/firebase/index.d.ts
```diff
@@ -7879,15 +7879,27 @@ declare namespace firebase.firestore {
      * buffer traffic indefinitely. Use of this option will cause some
      * performance degradation though.
      *
-     * This setting may be removed in a future release. If you find yourself
-     * using it to work around a specific network reliability issue, please
-     * tell us about it in
-     * https://github.com/firebase/firebase-js-sdk/issues/1674.
+     * This setting cannot be used with experimentalAutoDetectLongPolling and
```
egilmorez, line 7882: Suggest backticks for this literal.

egilmorez (approved):
One nit, otherwise everything in index.d.ts looks fine, thanks!

zwu52 (approved):

.changeset/yellow-turkeys-lay.md
rafikhan: Done

hsubox76 (approved):

PR #3726, opened 2020-09-02
[Deprecated] Messaging onPush should await onBackgroundMessage handler (in favor of https://github.com/firebase/firebase-js-sdk/pull/3780)

### LINKED ISSUE #3725: onPush not awaiting onBackgroundMessage event handler
* Operating System version: ArchLinux 5.8.5
  * Browser version: Chromium 85
  * Firebase SDK version: 7.19.1
  * Firebase Product: messaging

There's always additional notification when there is background push message coming, saying something like "The site has been updated in the background". I suspect the culprit is code below

```
      if (typeof this.bgMessageHandler === 'function') {
        this.bgMessageHandler(payload);
      } else {
        this.bgMessageHandler.next(payload);
      }
```

As you can see, `this.bgMessageHandler` is called without `await`, so when I return promise from `onBackgroundMessage` callback to display custom notification, it is not await-ed.

niyoko:
Await onBackgroundMessage handler on onPush method. This PR fixes #3725

### DIFF
--- a/packages/messaging/src/controllers/sw-controller.ts
+++ b/packages/messaging/src/controllers/sw-controller.ts
@@ -206,9 +206,9 @@ export class SwController implements FirebaseMessaging, FirebaseService {
       const payload = externalizePayload(internalPayload);
 
       if (typeof this.bgMessageHandler === 'function') {
-        this.bgMessageHandler(payload);
+        await this.bgMessageHandler(payload);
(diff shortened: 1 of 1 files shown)


niyoko:
@googlebot I signed it!

zwu52 (approved):

PR #3729, opened 2020-09-02
Fix externals for packages/firebase builds

hsubox76:
Externals should include tslib and use `startsWith` pattern.

### DIFF
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -70,7 +70,8 @@ const plugins = [
   commonjs()
 ];
 
-const external = Object.keys(pkg.dependencies || {});
+const deps = [...Object.keys(pkg.dependencies || {}), 'tslib'];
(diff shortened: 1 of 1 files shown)


packages/firebase/rollup.config.js
```diff
@@ -70,7 +70,8 @@ const plugins = [
   commonjs()
 ];
 
-const external = Object.keys(pkg.dependencies || {});
+const deps = [...Object.keys(pkg.dependencies || {}), 'tslib'];
```
Feiyang1, line 73: You don't need to add `tslib` which is a not a direct dependency of `firebase/firestore/memory`, etc. We are good as long as the scoped packages e.g. `@firebase/firestore/memory` are excluded.

Feiyang1 (changes requested):

packages/firebase/rollup.config.js
hsubox76: Fixed.

Feiyang1 (approved):

PR #3730, opened 2020-09-02
Add origin validation to popup/redirect flows

samhorlbeck:
(b/166287563)


packages-exp/auth-exp/src/core/util/validate_origin.ts
```diff
@@ -0,0 +1,81 @@
...
+      // Do nothing if there's a URL error; just continue searching
+    }
+  }
+
+  // In the old SDK, this error also provides helpful messages.
+  fail(AuthErrorCode.INVALID_ORIGIN, { appName: auth.name });
+}
+
+function matchDomain(expected: string): boolean {
+  const currentUrl = _getCurrentUrl();
```
avolkovi, line 45: doesn't this already do something funny on non-http? like force localhost?

packages-exp/auth-exp/src/core/util/validate_origin.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return protocol === 'chrome-extension:' && ceUrl.hostname === hostname;
+  }
+
+  if (!HTTP_REGEX.test(protocol)) {
+    return false;
+  }
+
+  if (IP_ADDRESS_REGEX.test(expected)) {
```
avolkovi, line 66: what about port specification?

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
```diff
@@ -19,14 +19,14 @@ import { SDK_VERSION } from '@firebase/app-exp';
 import * as externs from '@firebase/auth-types-exp';
 import { isEmpty, querystring } from '@firebase/util';
 
+import { _getProjectConfig } from '../api/project_config/get_project_config';
```
avolkovi, line 22: unused import, run your linter

packages-exp/auth-exp/src/core/util/validate_origin.ts
samhorlbeck: Nah, you're thinking of something else. This function is defined as:
```js
return (typeof self !== 'undefined' && self.location?.href) || '';
```

packages-exp/auth-exp/src/core/util/validate_origin.ts
samhorlbeck: Not sure-- this is copied directly from the old SDK fwiw

packages-exp/auth-exp/src/platform_browser/popup_redirect.ts
samhorlbeck: Linter not picking it up for some reason

PR #3735, opened 2020-09-03
run size reports in parallel in CI

Feiyang1:
1. run the regular size report and modular export size report in parallel
2. fix an issue where the regular size report exit before the report is created

### DIFF
--- a/.github/workflows/health-metrics-test.yml
+++ b/.github/workflows/health-metrics-test.yml
@@ -3,7 +3,7 @@ name: Health Metrics
 on: [push, pull_request]
 
 jobs:
-  binary-size-test:
+  binary-size:
     name: Binary Size
     if: github.event_name == 'push' || !(github.event.pull_request.head.repo.fork)
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/health-metrics-test.yml
```diff
@@ -19,12 +19,28 @@ jobs:
...
       - name: Run health-metrics/binary-size test
         run: yarn size-report
+  modular-export-size:
+    name: Binary Size For Modular Exports
+    if: github.event_name == 'push' || !(github.event.pull_request.head.repo.fork)
+    runs-on: ubuntu-latest
+    env:
+      METRICS_SERVICE_URL: ${{ secrets.METRICS_SERVICE_URL }}
+      GITHUB_PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
+      GITHUB_PULL_REQUEST_BASE_SHA: ${{ github.event.pull_request.base.sha }}
```
yifanyang, line 33: It's allowed to set the common ones at the workflow level ([doc](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#env)). It's up to you.

yifanyang (approved):

hsubox76 (approved):

.github/workflows/health-metrics-test.yml
Feiyang1: Nice! Changed.

PR #3736, opened 2020-09-03
Remove usage of the NODE_ADMIN global in RTDB

samtstern:
### Discussion

The `NODE_ADMIN` global means that the RTDB Admin and Client SDKs cannot coexist, whichever is initialized first will control the behavior.

This is a problem for testing, where clients want to use the Admin SDK to do privileged actions but then test the behavior of the Client SDK with low privileges.

### Testing

- [x] New unit test in `rules-unit-testing`

### API Changes

N/A

### DIFF
--- a/packages/database/index.node.ts
+++ b/packages/database/index.node.ts
@@ -51,13 +51,19 @@ const ServerValue = Database.ServerValue;
  * @param app A valid FirebaseApp-like object
  * @param url A valid Firebase databaseURL
  * @param version custom version e.g. firebase-admin version
+ * @param nodeAdmin true if the SDK is being initialized from Firebase Admin.
  */
-export function initStandalone(app: FirebaseApp, url: string, version: string) {
+export function initStandalone(
+  app: FirebaseApp,
+  url: string,
+  version: string,
+  nodeAdmin: boolean = true
+) {
   /**
    * This should allow the firebase-admin package to provide a custom version
    * to the backend
    */
-  CONSTANTS.NODE_ADMIN = true;
+  CONSTANTS.NODE_ADMIN = nodeAdmin;
   setSDKVersion(version);
 
   /**
@@ -84,7 +90,8 @@ export function initStandalone(app: FirebaseApp, url: string, version: string) {
     instance: RepoManager.getInstance().databaseFromApp(
       app,
       authProvider,
-      url
+      url,
+      nodeAdmin
     ) as types.Database,
     namespace: {
       Reference,
--- a/packages/database/src/realtime/WebSocketConnection.ts
+++ b/packages/database/src/realtime/WebSocketConnection.ts
@@ -87,7 +87,7 @@ export class WebSocketConnection implements Transport {
    */
   constructor(
     public connId: string,
-    repoInfo: RepoInfo,
+    private repoInfo: RepoInfo,
     private applicationId?: string,
     transportSessionId?: string,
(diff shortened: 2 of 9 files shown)


packages/database/src/realtime/WebSocketConnection.ts
```diff
@@ -87,7 +87,7 @@ export class WebSocketConnection implements Transport {
    */
   constructor(
     public connId: string,
-    repoInfo: RepoInfo,
+    private repoInfo: RepoInfo,
```
yuchenshi, line 90: I don't think we need a full reference to repoInfo. Can we just grab `nodeAdmin` as a field?

yuchenshi (approved):

packages/database/src/realtime/WebSocketConnection.ts
samtstern: You caught me being lazy

packages/database/index.node.ts
```diff
@@ -51,13 +51,19 @@ const ServerValue = Database.ServerValue;
...
+  url: string,
+  version: string,
+  nodeAdmin: boolean = true
+) {
   /**
    * This should allow the firebase-admin package to provide a custom version
    * to the backend
    */
-  CONSTANTS.NODE_ADMIN = true;
+  CONSTANTS.NODE_ADMIN = nodeAdmin;
```
Feiyang1, line 66: We can remove this constant since RTDB is the only consumer of it.  If any other JS SDKs were used by the admin SDK in the future, they should not rely on this constant either.

yuchenshi: We should hold off on this. `firebase-admin` does not pin its dependency against `@firebase/database`, so we can only remove this in a major version bump to avoid breaking customers.

yuchenshi: Oh I think I misunderstood -- this is all internal in this package and utils

samtstern: @Feiyang1 so I still want `isNodeSdk()` to work ... where does `NODE_ADMIN` actually come from when it's not set in `initStandalone`?  Is it set in part of the build process?

Feiyang1: It defaults to `false` unless set by `initStandalone`. https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/constants.ts#L30

samtstern: I'm not comfortable removing this at the moment ... it's used within `isNodeSdk()` which is used in many places including one place where `module.exports` is redefined.  I think removing this is likely to break something I don't understand.  Maybe someone from JSCore feels confident enough to do it as a follow up.

Feiyang1: The check for CONSTANTS.NODE_ADMIN is unnecessary in `isNodeSdk()` because `CONSTANTS.NODE_CLIENT` will be always true when `CONSTANTS.NODE_ADMIN` is true. Sure, I will do a follow up PR for this.

samtstern:
@schmidt-sebastian I need your codeowners review on this!

schmidt-sebastian (approved):
Basically LGTM, though I wonder if we can simplify this a bit: https://gist.github.com/schmidt-sebastian/08c74a0160809165fed81e31088cf8ce

If not, we can merge as is.

samtstern:
@hsubox76 says I can merge over this error, so merging!

PR #3740, opened 2020-09-04
Prepare auth-next for merging to master

Feiyang1:


packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
```diff
@@ -33,6 +33,10 @@ export class RecaptchaVerifier
     impl.assertFn(app.options?.apiKey, impl.AuthErrorCode.INVALID_API_KEY, {
       appName: app.name
     });
-    super(container, parameters, (app.auth!() as unknown) as externs.Auth);
+    super(
+      container,
+      parameters as any,
```
Feiyang1, line 38: I cast it to `any` to workaround the TS error - 
```
Argument of type 'object | null | undefined' is not assignable to parameter of type 'Parameters'.
  Type 'undefined' is not assignable to type 'Parameters'.
```
The error seems legit. Is the casting safe here?

packages-exp/auth-exp/test/integration/flows/phone.test.ts
```diff
@@ -69,7 +70,11 @@ describe('Integration test: phone auth', () => {
     auth = getTestInstance();
     fakeRecaptchaContainer = document.createElement('div');
     document.body.appendChild(fakeRecaptchaContainer);
-    verifier = new RecaptchaVerifier(fakeRecaptchaContainer, undefined, auth);
+    verifier = new RecaptchaVerifier(
+      fakeRecaptchaContainer,
+      undefined as any,
```
Feiyang1, line 75: Has to cast to `any` to workaround TS error here and below. `RecaptchaVerifier` doesn't take `undefined` as the second argument, but I guess it's okay if the test doesn't touch the second argument.

Feiyang1:
Renamed `recaptcha.d.ts` to `recaptcha.ts`, otherwise `recpatcha.d.ts` is not created in the `/dist` folder causing broken typings. We should always use `.ts` files in the implementation package, even just for defining types.

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
samhorlbeck: Yeah the typing in the parent class is wrong. I'll need to go in and fix it there

samhorlbeck (approved):
Thanks so much for doing this Fei!

avolkovi (approved):

PR #3750, opened 2020-09-09
Increase timeout for flaky test

hsubox76:
Test is consistently exceeding timeout of 60s. Extending `before` timeout to 120s.


hiranya911 (approved):
LGTM. You can also just give an increased timeout for this test case, if that's more appropriate.

```
describe(...).timeout(120000);
```

PR #3752, opened 2020-09-09
Fix NotInFilter matches

thebrianchen:
Porting over from [Android](https://github.com/firebase/firebase-android-sdk/pull/1947).

Making integration tests run against prod since protos are now public. Tests are currently failing b/c the client proxy has not been fully released yet, but they should be passing by EOW.


wilhuff (approved):
LGTM, but don't submit until tests are passing.

PR #3753, opened 2020-09-09
Add forceOwnership to firestore-exp

schmidt-sebastian:
This adds "experimentalForceOwningTab" to the experimental SDK. Exposed as "forceOwnership" - comments welcome.


thebrianchen (approved):
lgtm, I like forceOwnership more than forceOwningTab. This is a good way to save on another round of naming discussions :P

PR #3773, opened 2020-09-13
fix type

wilf312:


wilf312:
@googlebot I signed it!

packages/auth-types/index.d.ts
```diff
@@ -313,7 +313,7 @@ export class TwitterAuthProvider_Instance implements AuthProvider {
 
 export type UserCredential = {
   additionalUserInfo?: AdditionalUserInfo | null;
-  credential: AuthCredential | null;
```
avolkovi, line 316: the existing typing is correct, in the case of signInWithPhoneNumber this would be a https://firebase.google.com/docs/reference/js/firebase.auth.phoneauthcredential

wilf312: @avolkovi 
Thank you for reviewing and comment!

[My problem](https://github.com/wilf312/github-notification-ui/blob/task/okada/addLogin/pages/index.hooks.ts#L37) is when I do github authentication and the signInWithPopup returns an AuthCredential instead of an OAuthCredential.

> signInWithPopup(provider: AuthProvider): Promise<UserCredential>;

the value returned by the Promise<UserCredential> would change depending on the provider passed in [signInWithPopup](https://github.com/wilf312/firebase-js-sdk/blob/wilf/fixAuthType/packages/auth-types/index.d.ts#L450).


I think the value returned by the promise will change when using an authentication provider such as github, so the
in some auth cases  need the type information such as accessToken provided by [OAuthCredential](https://firebase.google.com/docs/reference/js/firebase.auth.OAuthCredential).

avolkovi: Is it possible for you to cast the response to an OAuthCredential once you have it? This is unfortunately not a trivial thing for us to fix given the existing API, but something we can look into fixing as we rewrite this SDK into TypeScript 

@samhorlbeck FYI

wilf312: Thank you.  I will respond in the way you suggested.
If SDK rewritten to TS that would be great, for me!

PR #3780, opened 2020-09-15
Set 1s timeout for onBackgroundMessage Hook

zwu52:
(credit to niyoko for reporting [issue](https://github.com/firebase/firebase-js-sdk/issues/3725) with [PR](https://github.com/firebase/firebase-js-sdk/pull/3726). 

set a 1s onBackgroundMessage hook so that the silent-push warnings wouldn't show if showNotification is called inside the hook. Currently, if the logic inside the hook is "long-running", `showNotification` message will show up as well as the warning. 

1s is derived from experiments of 1000 runs to at allow iterating through at least 1000 message with the notification merge practice we advocate.


packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -206,9 +206,9 @@ export class SwController implements FirebaseMessaging, FirebaseService {
       const payload = externalizePayload(internalPayload);
 
       if (typeof this.bgMessageHandler === 'function') {
-        this.bgMessageHandler(payload);
+        await this.bgMessageHandler(payload);
       } else {
-        this.bgMessageHandler.next(payload);
+        await this.bgMessageHandler.next(payload);
```
Feiyang1, line 211: This is not compatible with `Observer` signature where `next()` should return `void`. Perhaps we should only take a callback instead of an observer in `onBackgroundMessage()` ?

Feiyang1, line 211: This is not compatible with `Observer` signature where `next()` should return `void`. Perhaps we should only take a callback instead of an observer in `onBackgroundMessage()` ?

Feiyang1 (changes requested):

packages/messaging/src/controllers/sw-controller.ts
Feiyang1: @zxu123

zwu52: I think the issue is caused by the fact that the the the silent push check races the onBackgroundMessage. The only thing we are during differently is the wrapping and transformation operations on the payload (which is necessary for format and style).

Feiyang1: Do you think it's a bug on Chrome side? Are we supposed to wait on the promise returned by `ServiceWorkerRegistration.showNotification()` ?

zwu52: I am not sure if blocking ServiceWorkerRegistration.showNotification would work. I tried a await sleep(10s) at the end of `onPush`. The update in background would still show up. So I am suspecting there is a timer on this. I am going to verify it in Chrome code and possibly reach out to them to figure a solution if that's the case

Feiyang1: Does waiting on the promise returned by `ServiceWorkerRegistration.showNotification()` work? Chrome will always show a notification if we don't show a notification ourselves.

zwu52: `onBackgroundMessage` is a non-async function so can't `await` in that hook. 

However, I think I as mistaken. The `showNotification` check is ran after the onPush is completed with `event.waitUntil(onPush)`.   

The solution I can think of is to block n seconds at the end of `onPush` ([here](https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/src/controllers/sw-controller.ts#L213).  pros and cons of doing so:
__pro__: - limit how much time a piece of code can run in the background. Imagine a malicious code snippet just keeps running in the background and never completes. The `site has been updated in background` message will never show (but there is probably a max execution time enforcer somewhere). 
__con_): - blocking messages for every n seconds even if the onBackgroundMessage logic runs less than that. - It's kind of arbitrary on how to pick that n seconds. 

WDYT of this approach or is there a better way to put one an execution time limit in js?

Feiyang1: As you noted, I'm not sure how to pick the `n` seconds. And I'm not sure if we should make the decision on developer's behalf as there might be valid use cases that needs longer running time. 

`await onBackgroundMessage()` is the right thing to do IMO. Can we change its signature to `void | Promise<void>`, so that it's backward compatible? WDTY

zwu52: I think changing it to `Promise<void>` is reasonable.

Feiyang1: I don't think we can avoid the API review, but we can try to get it approved by email.

packages/messaging/src/controllers/sw-controller.ts
```diff
@@ -211,6 +217,9 @@ export class SwController implements FirebaseMessaging, FirebaseService {
         this.bgMessageHandler.next(payload);
       }
     }
+
+    // wait briefly to allow onBackgroundMessage to complete
+    await sleep(BACKGROUND_HANDLE_EXECUTION_TIME_LIMIT_MS);
```
Feiyang1, line 222: I'm sure you tested it. Just want to double check that we actually tested it in browser that it will prevent the warning to show.

Feiyang1: Also worth mentioning how we reached the number.

.changeset/fluffy-panthers-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/messaging': patch
+---
+
+Awaits onBackgroundMessage hook to avoid false-positive silent pushes warnings.
```
Feiyang1, line 5: `await` has a specially meaning in JS. It sounds like we are directly awaiting on the hook, but we are actually setting a timeout in parallel and await that timeout. Can you please make the changelog more specific about what we are doing and why we do it? It's probably worth mentioning the 1s limit and how we reached the number.

Feiyang1 (approved):

Issue #3768, opened 2020-09-11
Node-Fetch downstream security vulnerability


bradcypert:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Catalina 10.15.4
  * Browser version: 76.0.1
  * Firebase SDK version: 7.20.0
  * Firebase Product: firestore (auth, database, storage, etc)

### [REQUIRED] Describe the problem
There is a known security vulnerability with Node-Fetch that has been fixed in version 2.6.1+
https://www.npmjs.com/advisories/1556

#### Steps to reproduce:

Create a new project using `yarn init`.
Add firebase via `yarn add firebase`.
Run `yarn audit`
Note the security vulnerability on downstream dependency node-fetch.

#### Relevant Code:
As this is a downstream dependency issue, I dont have any relevant code to provide.

hsubox76:
For my info:
The two sources of older versions of node-fetch are

- in Firestore's package.json (I see a PR for that, thanks)
- in Functions as a dependency of [`isomorphic-fetch`](https://github.com/matthew-andrews/isomorphic-fetch) which seems to be no longer maintained. Perhaps this should be replaced with `node-fetch` directly.

wilhuff:
The issue in Firestore was resolved with #3759. The issue with isomorphic-fetch remains.

PR #3782, opened 2020-09-15
Remove isomorphic-fetch dependency

hsubox76:
The `isomorphic-fetch` dependency has not been updated in years and depends on an older version of `node-fetch` which has now been discovered to have security issues. Removing isomorphic-fetch and consuming node-fetch directly.

I tried another approach that involved passing the node-fetch implementation directly into the functions service, but it involved passing through many layers, and since it is passed into the Service constructor, the `createTestService` test util function needs to detect the environment and pass the correct fetch implementation as well which seems like a lot of places. This seems like the simplest approach and only touches the `index.node.ts` file.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3768

### DIFF
--- a/packages-exp/functions-exp/src/config.ts
+++ b/packages-exp/functions-exp/src/config.ts
@@ -27,23 +27,29 @@ import { FUNCTIONS_TYPE } from './constants';
 
 export const DEFAULT_REGION = 'us-central1';
 
-const factory: InstanceFactory<'functions'> = (
-  container: ComponentContainer,
-  region?: string
-) => {
-  // Dependencies
-  const app = container.getProvider('app-exp').getImmediate();
-  const authProvider = container.getProvider('auth-internal');
-  const messagingProvider = container.getProvider('messaging');
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return new FunctionsService(app, authProvider, messagingProvider, region);
-};
-
-export function registerFunctions(): void {
+export function registerFunctions(fetchImpl: typeof fetch): void {
   const namespaceExports = {
     // no-inline
     Functions: FunctionsService
+  };
+
+  const factory: InstanceFactory<'functions'> = (
+    container: ComponentContainer,
+    region?: string
+  ) => {
+    // Dependencies
+    const app = container.getProvider('app-exp').getImmediate();
+    const authProvider = container.getProvider('auth-internal');
+    const messagingProvider = container.getProvider('messaging');
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return new FunctionsService(
+      app,
+      authProvider,
+      messagingProvider,
+      region,
+      fetchImpl
+    );
   };
 
   _registerComponent(
--- a/packages-exp/functions-exp/src/config.ts
+++ b/packages-exp/functions-exp/src/config.ts
@@ -28,11 +28,6 @@ import { FUNCTIONS_TYPE } from './constants';
 export const DEFAULT_REGION = 'us-central1';
 
 export function registerFunctions(fetchImpl: typeof fetch): void {
-  const namespaceExports = {
-    // no-inline
-    Functions: FunctionsService
-  };
-
   const factory: InstanceFactory<'functions'> = (
     container: ComponentContainer,
     region?: string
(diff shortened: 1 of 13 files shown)


packages-exp/functions-exp/src/index.node.ts
```diff
@@ -16,10 +16,22 @@
  */
 import { registerVersion } from '@firebase/app-exp';
 import { registerFunctions } from './config';
-import 'isomorphic-fetch';
+import 'whatwg-fetch';
```
Feiyang1, line 19: Instead of including a new polyfill, we can replace `Header` with a string map. I prefer to not include a new polyfill that only provides `Header` and given that the workaround is really simple.

packages-exp/functions-exp/src/index.node.ts
```diff
@@ -16,10 +16,22 @@
...
+/**
+ * Patch global object with node-fetch. whatwg-fetch polyfill patches other global
+ * fetch objects such as Headers. Then we override the implemenation of `fetch()`
+ * itself with node-fetch.
+ * https://github.com/node-fetch/node-fetch#loading-and-configuring-the-module
+ * node-fetch type deviates somewhat from fetch spec:
+ * https://github.com/apollographql/apollo-link/issues/513#issuecomment-548219023
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+(global as any).fetch = (nodeFetch as unknown) as typeof fetch;
```
Feiyang1, line 33: I'm worried about introducing hard to debug issues in developers' app by patching globals. For example, they might use a different fetch polyfill which can be overwritten by our `node-fetch` polyfill depending on the import order. node-fetch might have a bug that the other polyfill doesn't have. It will be really hard for developers to track down the issue.

Though the chance of this kind of bug is slim, but it's not impossible, I think it's a good practice to not patch globals to future proof the SDK and developers' apps.

packages-exp/functions-exp/src/index.node.ts
hsubox76: Fixed (it is a string-keyed object, `fetch` signature does not accept Maps).

packages-exp/functions-exp/src/index.node.ts
hsubox76: Fixed, it is passed in now.

packages-exp/functions-exp/src/config.ts
```diff
@@ -27,25 +27,31 @@ import { FUNCTIONS_TYPE } from './constants';
...
-  const authProvider = container.getProvider('auth-internal');
-  const messagingProvider = container.getProvider('messaging');
-
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return new FunctionsService(app, authProvider, messagingProvider, region);
-};
-
-export function registerFunctions(): void {
+export function registerFunctions(fetchImpl: typeof fetch): void {
   const namespaceExports = {
```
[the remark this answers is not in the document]
hsubox76: Removed.

Feiyang1 (approved):

joandvgv:
Hi! sorry to bother. Any ideas on when this one will be released? Thanks!

hsubox76:
It should be in this week's release, which will likely go out today sometime. I'm not sure what issue you are hoping to fix but it shouldn't affect functionality, it just fixes a potential vulnerability and will avoid an ugly warning message on npm install.

joandvgv:
Thanks for the answer! Yeah I had the same vulnerability and had to use node-fetch as well but it still remains for this package, I followed the issues and discussions related to the vuln so just wanted to know when it'll be merged as my boss is always aiming for the 0 vulnerabilities when doing npm i or npm audit lol

inakie:
Hi,

I'm trying to build a website with Gridsome, and the build process gives me the following error, which I believe is related with this commit:

```
ReferenceError: fetch is not defined
    at Object.iTTW (node_modules/@firebase/functions/dist/index.esm.js:652:26)
```

If I edit the file `node_modules/@firebase/functions/dist/index.esm.js` and add at the beginning of the file `const fetch = require("node-fetch");` everything works well.

But I don't want to edit that file in node_modules... do you know if there is another way to solve this problem?

Thanks!

hsubox76:
See issue https://github.com/firebase/firebase-js-sdk/issues/3890

Just released a fix in 7.22.1

PR #3790, opened 2020-09-16
Add API Extractor for Firestore + API Pruning

schmidt-sebastian:
This adds support for API extractor to Firestore Exp and Firestore Lite and a script to prune its output to match our previous manual API layer. 

I manually verified that the new and the old type files are equivalent (which was quite painful, but very necessary).

There are some things we could still improve:
- Remove unused imports
- Go back to `export` instead of `export declare`
- Whitespace between lines.


packages/firestore/scripts/prune-dts.ts
```diff
@@ -0,0 +1,339 @@
...
+          type.expression
+        );
+        if (publicName && publicName != currentName) {
+          // If there is a public type that we can refer to, update the import
+          // statement to refer to the public type.
+          exportedTypes.push(
+            ts.updateExpressionWithTypeArguments(
+              type,
+              type.typeArguments,
+              ts.createIdentifier(publicName)
```
Feiyang1, line 142: IIUC, this can potentially make the class/interface has more members than it really has. Considering this:
```
class A {
   data: string;
}

export class C extends A {
   additionalData: number;
}

export class B  extends A {
}

```

 will become 

```
export class C {
  data: string;
  additionalData: number;
}

export class B  extends C {
}
```

Now `B` has `additionalData` it didn't have before.

I think the safest strategy is always copy the members from the private symbols instead of replacing it with a public one that extends it.

Feiyang1:
Nit: I saw some usage of `==` and `!=` which should be replaced by `===` and `!==` instead.

schmidt-sebastian (commented):
Here it is again.

Sorry, the commit history is a bit messy, but I wanted to keep it in case I need to go back to an earlier revision. 

I addressed your main two points as follows:
- If a private API has different generic arguments from a public API, I will no longer simply substitute their names. Instead, I merge the declarations of the private API into the public type.
- To merge the declarations, I now use TypeScript's `codefix` package. At first, I tried to just port the code that we need, but it quickly got out of hand. So for now, this script relies on the same private API that VS Code uses to auto-fill missing imports. If this private API ever breaks, we can move the required parts into our repository (a hacky attempt is here: https://github.com/firebase/firebase-js-sdk/pull/3790/commits/ffc5ba1847bbe5f37f7b63d0ebce94a288705575#diff-6db0153ee9f6f29fa6fc10782448b200R204, but it still has some issues)

packages/firestore/scripts/prune-dts.ts
Feiyang1: I'm confused. Is B2 B and C2 C from my example? Or are you giving a different example?

packages/firestore/scripts/prune-dts.ts
```diff
@@ -0,0 +1,420 @@
...
+      } else {
+        const publicSymbol = extractPublicSymbol(
+          typeChecker,
+          sourceFile,
+          type.expression
+        );
+        if (
+          publicSymbol &&
+          publicSymbol.name !== currentName &&
+          verifyMatchingTypeArguments(
```
Feiyang1, line 174: I'm not sure it's the sufficient condition for a safe replacement. Considering:
```
declare class A<T> {
   aData: T;
}

export declare class B<T> extends A<string> {
    bData: T; 
}

export declare class C<T> extends A<T> {
}
```
Will it become?
```
export declare class B<T> {
    aData: string;
    bData: T; 
}
export declare class C<T> extends B<T> { // it is incorrect, because the type of `aData` becomes `string` which was `T` in the original definition.
}
```

packages/firestore/scripts/prune-dts.ts
```diff
@@ -0,0 +1,420 @@
...
+  sourceFile: ts.SourceFile,
+  parentClassSymbols: ts.Symbol[],
+  currentClass: ts.ClassDeclaration | ts.InterfaceDeclaration
+): ts.NamedDeclaration[] {
+  const newMembers: ts.NamedDeclaration[] = [];
+  // The `codefix` package is not public but it does exactly what we want. We
+  // can explore adding a slimmed down version of this package to our repository
+  // if this dependency should ever break.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (ts as any).codefix.createMissingMemberNodes(
```
Feiyang1, line 259: Does it handle generics with different names? For example:
```
declare class A<T> {
  aData: T;
}

export declare class B<K> extends A<K> {
  bData:K;
}
```
Does it become:
```
export declare class B<K> {
  aData: K; // instead of aData: T
  bData: K;
}
```

packages/firestore/scripts/prune-dts.ts
```diff
@@ -0,0 +1,339 @@
...
+ * Input:
+ * class PrivateFoo {}
+ * export class PublicFoo extends PrivateFoo {}
+ * export function doFoo(foo: PrivateFoo);
+ *
+ * Output:
+ * export class PublicFoo {}
+ * export function doFoo(foo: PublicFoo);
+ */
+function extractPublicName(
```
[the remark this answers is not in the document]
Feiyang1: Can you show me an example in a snippet? Don't lite and exp have separate d.ts files, why would there be 2 versions of FirebaseDataConverter? and what do you mean by leaking invisible types?

schmidt-sebastian:
@Feiyang1 - Can you take another look at this?

Feiyang1:
@schmidt-sebastian I don't see any new changes since my last review. Did you mean to push something, but didn't?

PR #3791, opened 2020-09-16
Remove CHANGELOG.md

zwu52:
Remove package CHANGELOG as we have moved to using `changeset`


zwu52:
realized this was autogened

PR #3794, opened 2020-09-16
Hide converter from Public API

schmidt-sebastian:
The converter property is currently not part of our public API, and as such it should be prefixed with an underscore

This is firestore-exp/lite only.

### DIFF
--- a/packages/firestore/lite/src/api/reference.ts
+++ b/packages/firestore/lite/src/api/reference.ts
@@ -108,10 +108,10 @@ export class DocumentReference<T = DocumentData> extends _DocumentKeyReference<
 
   constructor(
     readonly firestore: FirebaseFirestore,
-    readonly converter: FirestoreDataConverter<T> | null,
+    _converter: FirestoreDataConverter<T> | null,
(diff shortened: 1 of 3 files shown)


packages/firestore/lite/src/api/reference.ts
```diff
@@ -108,10 +108,10 @@ export class DocumentReference<T = DocumentData> extends _DocumentKeyReference<
 
   constructor(
     readonly firestore: FirebaseFirestore,
-    readonly converter: FirestoreDataConverter<T> | null,
+    _converter: FirestoreDataConverter<T> | null,
```
thebrianchen, line 111: Why does this need to have its `readonly` property removed? I don't see the converter value being assigned to a new value anywhere.

thebrianchen (approved):
lgtm w/ question

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: The readonly is already part of the parent class, which already uses the underscore naming.

Issue #414, opened 2018-01-07
Encoding is not UTF-8 error


ardmn:
### [REQUIRED] Describe your environment

  * Operating System version: mac os
  * Firebase SDK version: 4.8.1
  * Firebase Product: database

### [REQUIRED] Describe the problem

browser show error When I try add firebase.js to chrome extension: 
Can't load file  /../firebase.js ... . Encoding is not UTF-8.

I try convert it to utf-8 but browser show error anyway 

#### Steps to reproduce:
1) download firebase.js using https://www.gstatic.com/firebasejs/4.8.1/firebase.js
2) add firebase.js to chrome extension package
3) enable developer mode in chrome browser
4) load extension to chrome browser

ardmn:
Only avalible solution what I find it is using backgrounf.html, wich load firebase.js (<script> tag) and with messaging betwin background and conten scripts. 

But if u know how download firebase.js or how transform firebase.js in order to use the js file in chrome extension - please tell me

barna-n:
Same problem...

davemart-in:
Seeing the same thing...

winterlightning:
same problem

davemart-in:
I found a work-around. It doesn't look pretty, but by recompiling the second library that is included in firebase.js with [Google closure compiler](https://closure-compiler.appspot.com/home) I was able to make it work: https://pastebin.com/QPDqVimD

falcondai:
Another workaround by *avoiding* to import `firebase-auth.js`  https://stackoverflow.com/a/49496631/1121123. The issue is likely the same as https://github.com/pouchdb/pouchdb/issues/5798

protonhs:
@davemart-in  your solution's awesome ! it solves it perfectly! Thanks!

davemart-in:
Awesome!  Glad to hear!

aseire:
@davemart-in Brilliant! it solves it perfectly! Thank You!

lukejacksonn:
So I don't know what this approach might break.. but I wasn't happy with recompiling or using background.html and sending messages. So I searched the source for `\u` and eventually you come across the character `￿` it is used in this snippet of code `xb=/\uffff/.test("￿")`. If you remove all instances of this character so that the code reads `xb=/\uffff/.test("")` then the error no longer appears but everything seems to work as normal (database and auth at least).

laocoi:
I have the same problem with PHP when I extract text from HTML string by substr function.
[...]

PR #3805, opened 2020-09-18
Minified Console Build

schmidt-sebastian:
Size before: 360299 kB
Size after: 228092 kB
Down by 37%.
You are welcome.


packages/firestore/tools/console.build.js
```diff
@@ -19,34 +19,44 @@
...
+
+const es2017OutputOptions = {
+  file: tmpFile,
+  format: 'es'
 };
-const outputOptions = {
+
+const es2017toEs5InputOptions = {
+  input: tmpFile,
+  plugins: rollupUtil.es2017ToEs5Plugins(/* mangled= */ true),
```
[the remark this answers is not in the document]
Feiyang1: Tree-shaking seems to be working as I don't see `IndexedDbOfflineComponentProvider` in the output.

packages/firestore/tools/console.build.js
```diff
@@ -19,34 +19,44 @@
...
 
 const EXPORTNAME = '__firestore_exports__';
 
-const inputOptions = {
+const tmpFile = tmp.fileSync().name;
+
+const es2017InputOptions = {
   input: 'index.console.ts',
-  plugins
+  // If I set mangled to true the build breaks, but all other build pipelines
```
Feiyang1, line 36: The problem seems to come from the 2 interface exports - `FirestoreDatabase` and `FirstPartyCredentialsSettings`. Once they are removed, the mangling works again.

The issue seems to be that they are not included in the public externs.

Feiyang1: I don't think the interfaces provide any value here since console uses the js output, so they can probably be removed.

schmidt-sebastian (commented):
Thanks for all the pointers Fei. I have addressed your feedback and it seems to work now. Savings are quite astonishing.

schmidt-sebastian:
FYI Laura confirmed that this build works in the Console.

packages/firestore/src/api/credentials.ts
```diff
@@ -29,14 +29,16 @@ import { logDebug } from '../util/log';
...
+  ['client']: unknown;
+  ['sessionIndex']: string;
 }
 
 export interface ProviderCredentialsSettings {
-  type: 'provider';
-  client: CredentialsProvider;
+  // These are external types. Prevent minification.
+  ['type']: 'provider';
+  ['client']: CredentialsProvider;
```
Feiyang1, line 41: Does it make any difference? These are just types which won't be minified anyway.

Feiyang1: If you are making the change to be able to do `credentials['type']`, it's not necessary. Typescript will allow you to do indexing as expected.

packages/firestore/tools/console.build.js
```diff
@@ -20,40 +20,51 @@
...
+};
+
+const es2017toEs5InputOptions = {
+  input: esm2017OutputFile,
+  plugins: [
+    resolve(),
+    ...rollupUtil.es2017ToEs5Plugins(/* mangled= */ true),
+    uglify({
+      output: {
+        ascii_only: true // escape unicode chars
```
Feiyang1, line 57: nice! Found the solution for https://github.com/firebase/firebase-js-sdk/issues/414

packages/firestore/src/api/credentials.ts
schmidt-sebastian: This blocks minification.

Feiyang1: I see. that makes sense

Feiyang1 (approved):

PR #3810, opened 2020-09-18
Add a way to specify emulator host/port for Auth

avolkovi:


MeghaB (approved):
lgtm - thanks for adding the console warning!

packages/auth/src/auth.js
```diff
@@ -272,6 +292,63 @@ fireauth.Auth.prototype.useDeviceLanguage = function() {
...
+
+/**
+ * Emits a console warning and a visual banner if emulator integration is
+ * enabled.
+ */
+fireauth.Auth.prototype.emitEmulatorWarning_ = function() {
+  fireauth.util.consoleWarn('WARNING: You are using the Auth Emulator,' +
+    ' which is intended for local testing only.  Do not use with' +
+    ' production credentials.');
+  if (goog.global.document) {
```
yuchenshi, line 324:
```suggestion
  if (goog.global.document && goog.global.document.body) {
```

packages/auth/src/auth.js
```diff
@@ -272,6 +292,63 @@ fireauth.Auth.prototype.useDeviceLanguage = function() {
...
+/**
+ * Emits a console warning and a visual banner if emulator integration is
+ * enabled.
+ */
+fireauth.Auth.prototype.emitEmulatorWarning_ = function() {
+  fireauth.util.consoleWarn('WARNING: You are using the Auth Emulator,' +
+    ' which is intended for local testing only.  Do not use with' +
+    ' production credentials.');
+  if (goog.global.document && goog.global.document.body) {
+    const ele = goog.global.document.createElement('p');
```
samhorlbeck, line 325: This should probably be a `<div>`instead but no need to change it now

samhorlbeck (approved):

Feiyang1 (changes requested):
Can you please add the corresponding typings to `packages/firebase/index.d.ts`?

Feiyang1:
Actually are you planning to make this API public now, or later?

packages/firebase/index.d.ts
```diff
@@ -3119,6 +3119,14 @@ declare namespace firebase.auth {
      * Sets the current language to the default device/browser preference.
      */
     useDeviceLanguage(): void;
+    /**
+     * Modify this Auth instance to communicate with the Firebase Auth emulator.  This must be 
+     * called synchronously immediately following the first call to `firebase.auth()`.  Do not use
+     * with production credentials as emulator traffic is not encrypted.
+     * 
+     * @param url - The URL at which the emulator is running (eg, 'http://localhost:9099')
```
Feiyang1, line 3127:
```suggestion
     * @param url The URL at which the emulator is running (eg, 'http://localhost:9099')
```

We are using JSDoc currently, which doesn't requires the `-` after the  parameter name

Feiyang1 (approved):

Issue #3811, opened 2020-09-18
FR: Query get() method for RTDB


jmwski:
### Describe your environment

  * Firebase Product: database (auth, database, storage, etc)

### Describe the problem

Per the contributor guidelines, this issue serves as a discussion forum for the implementation of [go/rtdb-get](go/rtdb-get) in the firebase database javascript/web sdk.

get() should be implemented as a method on the database `Query` class and use the RTDB backend `Get` wire protocol message. Once a new value is retrieved, the SDK should update any relevant caches that could use the value.

Here's a WIP [implementation](https://github.com/firebase/firebase-js-sdk/pull/3812).

PR #3812, opened 2020-09-18
Query get() method for RTDB

IanWyszynski:
Firebase API Proposal: [go/rtdb-get](go/rtdb-get)

Github Issue: https://github.com/firebase/firebase-js-sdk/issues/3811

NOTE: CI will fail until https://github.com/firebase/firebase-server/pull/3298 is merged and the emulator is released.


packages/database/test/query.test.ts
```diff
@@ -2191,6 +2191,49 @@ describe('Query Tests', () => {
       });
   });
 
+  it('get() at empty root returns null', done => {
```
IanWyszynski, line 2194: @schmidt-sebastian Can you take a look at these tests? We discussed specing out the behavior of get() as it relates to listeners, but I don't think get() should interact with them directly in the JS sdk since they should be up to date anyway.

packages/database/src/api/Query.ts
```diff
@@ -295,6 +299,17 @@ export class Query {
...
   }
 
+  get(): Promise<DataSnapshot> {
+    const deferred = new Deferred<DataSnapshot>();
+    const callback = (snapshot: DataSnapshot) => {
+      deferred.resolve(snapshot);
+    };
+    this.repo.get(this, callback, err => {
+      deferred.reject(err);
+    });
```
schmidt-sebastian, line 309: Do you think you could push the callback to Promise conversion further down? The existing code uses a lot of callbacks since Promises can't drive Event-based logic, but the code that you are adding in this PR should be able to rely mostly on Promises.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -75,6 +75,13 @@ interface OutstandingPut {
   onComplete: (a: string, b?: string) => void;
 }
 
+interface OutstandingGet {
+  action: string;
+  request: object;
+  queued?: boolean;
+  onComplete: (a: string, b?: string) => void;
```
schmidt-sebastian, line 82: What are you thoughts on breaking some of the coding conventions in this repo and used descriptive names for `a` (value?)  and `b` (error?),

This might also go away if you follow my advice above.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -184,6 +193,29 @@ export class PersistentConnection extends ServerActions {
     }
   }
 
+  get(query: Query, onComplete: (a: string, b: unknown) => void) {
```
schmidt-sebastian, line 196: This should:
- Specify a return type
- Ideally return a Promise
- If not, use descriptive names for the options argument (and `b` seems to always be a string?)

packages/database/test/query.test.ts
schmidt-sebastian, line 2194:
```suggestion
  it('get() at empty root returns null', async () => {
```

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2191,49 @@ describe('Query Tests', () => {
...
+    const node = getRandomNode() as Reference;
+    node.get().then(snapshot => {
+      const val = snapshot.val();
+      expect(val).to.be.null;
+      done();
+    });
```
schmidt-sebastian, line 2200:
```suggestion
    const node = getRandomNode() as Reference;
    const snapshot = await node.get();
    const val = snapshot.val();
    expect(val).to.be.null;
    });
```

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2191,49 @@ describe('Query Tests', () => {
...
+  });
+
+  it('get() at non-empty root returns correct value', done => {
```
schmidt-sebastian, line 2203:
```suggestion
  it('get() at non-empty root returns correct value', async () => {
```

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2191,49 @@ describe('Query Tests', () => {
...
+    const reader = nodes[0];
+    const writer = nodes[1];
+    writer.set({ foo: 'a', bar: 'b' }, (err, dummy) => {
+      reader.get().then(snapshot => {
+        const val = snapshot.val();
+        expect(val['foo']).to.equal('a');
+        expect(val['bar']).to.equal('b');
+        done();
+      });
+    });
```
schmidt-sebastian, line 2214:
```suggestion
    await writer.set({ foo: 'a', bar: 'b' });
    const snapshot =  reader.get();
    const val = snapshot.val();
    expect(val['foo']).to.equal('a');
    expect(val['bar']).to.equal('b');
    });
```

schmidt-sebastian: (You should also rewrite the test below)

packages/database/test/query.test.ts
schmidt-sebastian: Got it. I think the main reason I asked for these tests is to make sure that values returned via `get` are cached. What you can do is maybe follow the pattern in https://github.com/firebase/firebase-js-sdk/blob/55f09a2642043d9fe9415c069e1dc82e1bcb612c/packages/database/test/info.test.ts#L142 and create two database instances. The first instance is just for writing a new value, the second value would call `get()` followed by `once()` to ensure that `once()` returns the correct value.

schmidt-sebastian (commented):
This is pretty good. Some nits. Thanks!

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,111 @@ describe('Query Tests', () => {
...
+    reader.child('foo/bar').once('value', snap => {
+      expect(snap.val()).to.equal(snapshot.val());
+    });
+  });
+
+  it('get does not cache sibling data', async () => {
+    const reader = getFreshRepo('db');
+    const writer = getFreshRepo('db');
+    await writer.set({ foo: { bar: { data: '1' }, baz: { data: '2' } } });
+    const snapshot = await reader.child('foo/bar').get();
```
schmidt-sebastian, line 2280:
```suggestion
    await writer.set({ foo: { cached: { data: '1' }, notCached: { data: '2' } } });
    const snapshot = await reader.child('foo/cached').get();
```

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,111 @@ describe('Query Tests', () => {
...
+        done = true;
+        reject();
+      }, 3000);
+      reader.child('foo/baz').once('value', snapshot => {
+        if (done) {
+          return;
+        }
+        done = true;
+        resolve(snapshot);
+      });
```
schmidt-sebastian, line 2298: Couldn't you just use `get()` here again since we are offline and it falls back to `once`?

schmidt-sebastian (commented):
18 nits, but overall this is great. Please also add a changelog (link via a bot-message in the main "conversation" view).

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,111 @@ describe('Query Tests', () => {
...
+    const node = getRandomNode() as Reference;
+    node.database.goOffline();
+    const snapshot = await node.get();
+    expect(snapshot.val()).to.be.null;
+    // Required for test isolation.
+    node.database.goOnline();
+  });
+
+  it('get caches results at path', async () => {
+    const reader = getFreshRepo('db');
```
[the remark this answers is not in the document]
IanWyszynski: How would that happen? I think `getFreshRepo` uses a global counter:

https://github.com/firebase/firebase-js-sdk/blob/b07f822bdb4cf628d53d337dc0df3db7f865a5aa/packages/database/test/helpers/util.ts#L143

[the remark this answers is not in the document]
IanWyszynski: I think `getFreshRepo` uses a global counter. Isn't everything single threaded here?

https://github.com/firebase/firebase-js-sdk/blob/b07f822bdb4cf628d53d337dc0df3db7f865a5aa/packages/database/test/helpers/util.ts#L143

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,111 @@ describe('Query Tests', () => {
...
+      });
+    });
+    const onSnapshot = await prom;
+    node.database.goOffline();
+    const getSnapshot = await node.get();
+    node.off();
+    expect(getSnapshot.val()).to.deep.equal((onSnapshot as DataSnapshot).val());
+  });
+
+  it('get above path caches data', async () => {
```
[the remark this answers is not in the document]
IanWyszynski: I see now it's not. Will remove.

packages/database/src/core/Repo.ts
```diff
@@ -296,6 +297,42 @@ export class Repo {
...
+        this.eventQueue_.raiseEventsAtPath(query.path, events);
+        return Promise.resolve(
+          new DataSnapshot(
+            node,
+            query.getRef(),
+            query.getQueryParams().getIndex()
+          )
+        );
+      },
+      err => {
```
schmidt-sebastian, line 317: Do we log the `err` anywhere? I assume this will be in the WebChannel logs, but if not, we may want to log it here as well.

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
       });
   });
 
+  it('get at empty node is rejected', async () => {
+    const node = getRandomNode() as Reference;
+    expect(node.get()).to.eventually.be.rejected;
```
schmidt-sebastian, line 2200: This is a "Promise assertion". You need to return the Promise or add `await`.

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
...
+    expect(snapshot.val()).to.deep.equal(expected);
+  });
+
+  it('get for removed node returns correct value', async () => {
+    const node = getRandomNode() as Reference;
+    const expected = { foo: 'a', bar: 'b' };
+    await node.set(expected);
+    let snapshot = await node.get();
+    let val = snapshot.val();
+    expect(val).to.deep.equal(expected);
```
schmidt-sebastian, line 2217: Please fix lint error.

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
...
+    await node.remove();
+    snapshot = await node.get();
+    expect(snapshot.val()).to.be.null;
+  });
+
+  it('get while offline is rejected', async () => {
+    const node = getRandomNode() as Reference;
+    node.database.goOffline();
+    try {
+      expect(node.get()).to.eventually.be.rejected;
```
schmidt-sebastian, line 2227: Missing `await` or `return`.

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
...
+    } finally {
+      node.database.goOnline();
+    }
+  });
+
+  it('get reads from cache if database is not connected', async () => {
+    const node = getRandomNode() as Reference;
+    const node2 = getFreshRepo((node as Reference).path);
```
schmidt-sebastian, line 2235: Is `as Reference` needed?

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
...
+        (onSnapshot as DataSnapshot).val()
+      );
+    } finally {
+      node.database.goOnline();
+    }
+  });
+
+  it('get does not cache sibling data', async () => {
+    const reader = getRandomNode() as Reference;
+    const writer = getFreshRepo((reader as Reference).path);
```
schmidt-sebastian, line 2257: Is `as Reference` needed?

packages/database/test/query.test.ts
```diff
@@ -2191,6 +2195,79 @@ describe('Query Tests', () => {
...
+    await writer.set({
+      foo: { cached: { data: '1' }, notCached: { data: '2' } }
+    });
+    const snapshot = await reader.child('foo/cached').get();
+    expect(snapshot.val()).to.deep.equal({ data: '1' });
+    reader.database.goOffline();
+    try {
+      expect(reader.child('foo/notCached').get()).to.eventually.be.rejected;
```
schmidt-sebastian, line 2265: Missing `await` or `return`.

schmidt-sebastian (commented):
Please remember to add Changelog.

packages/database/test/query.test.ts
```diff
@@ -15,11 +15,15 @@
  * limitations under the License.
  */
 
-import { expect } from 'chai';
+import * as chai from 'chai';
+const expect = chai.expect;
```
schmidt-sebastian, line 19: Can you clean up the order here and move this to the bottom:

```
const expect = chai.expect;
chai.use(chaiAsPromised);
```

You could also just do:

```
import {use, expect} from 'chai';
```

packages/database/src/core/Repo.ts
```diff
@@ -296,6 +297,45 @@ export class Repo {
...
+        this.log_(
+          'get for query ' +
+            stringify(query) +
+            ' falling back to cache after error ' +
```
schmidt-sebastian, line 321:
```suggestion
            ' falling back to cache after error: ' +
```

schmidt-sebastian (approved):

schmidt-sebastian (approved):

.changeset/many-snails-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
```
schmidt-sebastian, line 2: s/patch/minor

packages/database/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [changed] Added support for query get operation.
```
schmidt-sebastian, line 3: Please remove this/ This will be auto-populated by the SDK tooling.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -186,15 +195,55 @@ export class PersistentConnection extends ServerActions {
...
+          deferred.reject(payload);
+        }
+      }
+    };
+    this.outstandingGets_.push(outstandingGet);
+    this.outstandingGetCount_++;
+    const index = this.outstandingGets_.length - 1;
+
+    if (!this.connected_) {
+      const self = this;
```
schmidt-sebastian, line 226: This is very old style JavaScript. If you use a lambda below (`() => {...}`) you won't need to rewrite `this` and will also remove the lint warning.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -186,15 +195,55 @@ export class PersistentConnection extends ServerActions {
...
+        } else {
+          deferred.reject(payload);
+        }
+      }
+    };
+    this.outstandingGets_.push(outstandingGet);
+    this.outstandingGetCount_++;
+    const index = this.outstandingGets_.length - 1;
+
+    if (!this.connected_) {
```
schmidt-sebastian, line 225: Is there a reason we shouldn't always have this timeout?

packages/database/src/core/PersistentConnection.ts
```diff
@@ -186,15 +195,55 @@ export class PersistentConnection extends ServerActions {
...
+        if (get === undefined || outstandingGet !== get) {
+          return;
+        }
+        delete self.outstandingGets_[index];
+        self.outstandingGetCount_--;
+        if (self.outstandingGetCount_ === 0) {
+          self.outstandingGets_ = [];
+        }
+        self.log_('get ' + index + ' timed out on connection');
+        deferred.reject(new Error('Client is offline.'));
```
schmidt-sebastian, line 238: Should we have logic that falls back to cache here? This would only trigger if there was an active listener that covers this location, but for some reason we cannot send the `get()` (e.g. we were able to connect to the backend at some point but are no longer able to). If you decide against this, you should probably add a comment to explain the difference.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -76,6 +77,12 @@ interface OutstandingPut {
   onComplete: (a: string, b?: string) => void;
 }
 
+interface OutstandingGet {
+  action: string;
```
schmidt-sebastian, line 81: The action is always `g` here. Maybe we can drop it.

packages/database/src/core/PersistentConnection.ts
IanWyszynski: We sort of allow for false negatives if we have this timeout unconditionally.

packages/database/src/core/PersistentConnection.ts
IanWyszynski: Is this the right abstraction layer to add caching logic to? I don't think the serverSyncTree is in scope here. I'd prefer to add a comment.

schmidt-sebastian: Sounds good, please do add the comment.

packages/database/src/core/PersistentConnection.ts
```diff
@@ -184,6 +193,57 @@ export class PersistentConnection extends ServerActions {
...
+          deferred.reject(payload);
+        }
+      }
+    };
+    this.outstandingGets_.push(outstandingGet);
+    this.outstandingGetCount_++;
+    const index = this.outstandingGets_.length - 1;
+
+    if (!this.connected_) {
+      setTimeout(() => {
```
schmidt-sebastian, line 225: We should consider cancelling this timeout if the `get` succeeded (since it may keep the Node process running). We can do this in a follow-up.

schmidt-sebastian (approved):

packages/database/src/core/PersistentConnection.ts
IanWyszynski: Will do this in a followup.

IanWyszynski:
@Feiyang1 Adding you because of the database emulator update.

.changeset/many-snails-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": minor
```
Feiyang1, line 2:
```suggestion
"@firebase/database": minor
"firebase": minor
```

packages/database/src/core/Repo.ts
```diff
@@ -296,6 +297,71 @@ export class Repo {
...
+   *    this repo is handling.
+   *
+   * In case (1), it's possible that the client still has an active
+   * listener, with cached data. Since this is the latest known
+   * value satisfying the query, that's what getValue will return.
+   * If there is no cached data, `getValue` surfaces an "offline"
+   * error.
+   *
+   * In case (2), `getValue` will trigger a time-limited connection
+   * attempt. If the client is unable to connect to the server, the
```
Feiyang1, line 318:
```suggestion
   * attempt. If the client is unable to connect to the server, it
```

Feiyang1 (changes requested):

Feiyang1 (approved):

PR #3827, opened 2020-09-22
Fix generate_spec_json script

schmidt-sebastian:
This drops the old Platform specification in favor of the default (which is Node)


wilhuff (approved):
LGTM

PR #3830, opened 2020-09-22
Adding performance to firebase-exp

jposuna:
Also adding a missing param doc.


alikn (approved):

packages-exp/firebase-exp/installations/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
hsubox76, line 3: 2020

packages-exp/firebase-exp/installations/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
hsubox76, line 1: I think that to ensure these bundles get built, looking at the build config in https://github.com/firebase/firebase-js-sdk/blob/0c607c992527982666d604bc445a1b857f7da66f/packages-exp/firebase-exp/rollup.config.js#L114 it looks like installations and performance need to be added to the `components` list in https://github.com/firebase/firebase-js-sdk/blob/0c607c992527982666d604bc445a1b857f7da66f/packages-exp/firebase-exp/package.json#L59

Can you test out that `yarn build` in `packages-exp/firebase-exp` generates these files?

Feiyang1 (changes requested):
LGTM except that `installations` doesn't need to be added to `firebase-exp` as it is not intended for public usage. We don't expose it in `firebase` today.

jposuna:
> LGTM except that `installations` doesn't need to be added to `firebase-exp` as it is not intended for public usage. We don't expose it in `firebase` today.

Thanks for the info. Reverted the changes to installations.

hsubox76 (approved):

packages-exp/firebase-exp/performance/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "firebase-exp/performance",
+  "main": "dist/index.cjs.js",
+  "browser": "dist/index.cjs.js",
```
Feiyang1, line 4: `browser` should point to `dist/index.esm.js` the same as `module`

packages-exp/firebase-exp/performance/package.json
```diff
@@ -0,0 +1,8 @@
...
+  "module": "dist/index.esm.js",
+  "esm2017": "dist/index.esm2017.js",
```
Feiyang1, line 6: Remove. We don't need an esm2017 build in `firebase-exp` which will be just the same as the esm build, because we just reexport everything from `@firebase` packages

Feiyang1 (approved):
LGTM. Just need some updates in the package.json file

Issue #3504, opened 2020-07-29
chrome://flags/#cookies-without-same-site-must-be-secure and from Firefox 69 in about:config by setting network.cookie.sameSite.noneRequiresSecure.


markymarkfunky:
### [REQUIRED] Describe your environment

  * Operating System version: _____
  * Browser version: _____
  * Firebase SDK version: _____
  * Firebase Product: _____ (auth, database, storage, etc)

#### Relevant Code:

https://stackblitz.com/fork/firebase-issue-sandbox

```javascript
// TODO(you): code here to reproduce the problem
```

Feiyang1:
Closing the issue since it provides no information. Feel free to open a new issue with details.

Issue #2284, opened 2019-10-20
FR : Fix chrome cookies warning (`SameSite` attribute)


deka:
### [REQUIRED] Describe your environment

  * Operating System version:  Windows 10
  * Browser version:  Chrome (Version 77.0.3865.120)
  * Firebase SDK version:  6.6.1
  * Firebase Product: auth

### [REQUIRED] Describe the problem

Warning with cookies : 

> A cookie associated with a cross-site resource at http://google.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

#### Steps to reproduce:

- A domain (like https://my-domain.com)
- A single page application
- Use firebase.auth lib to signin with google account

bojeil-google:
The Auth SDK does not set any cookies. Please provide actual details on the source of this issue. This unlikely to be from our library. However, we do pull in other external dependencies. If you can help pinpoint that, we could route you to the right owners to fix this.

Morbden:
A cookie associated with a cross-site resource at http://cloudflare.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
v2.mapee.app/:1 A cookie associated with a resource at http://gstatic.com/ was set with `SameSite=None` but without `Secure`. A future release of Chrome will only deliver cookies marked `SameSite=None` if they are also marked `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5633521622188032.
A cookie associated with a cross-site resource at https://smartlock.google.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.
[...]

PR #3839, opened 2020-09-24
firebase/analytics => added setCookieFlags method

hobbydevs:

### Discussion

Related: #2284
Related: #3504

### Testing

Added a single method and a test to it.

### API Changes

    can discuss it together.

**There are no breaking changes.**

With the addition you are able to pass the `cookie_flags` to gtag. This is especially helpful when you need to use analytics inside an iframe (third party) and need the cookies to be set properly.
Without this method the cookies cannot be set (see screenshot below)

See [here](https://developers.google.com/analytics/devguides/collection/gtagjs/cookies-user-id#cookie_flags) for more information about the `cookie_flags`.

gtag

### DIFF
--- a/packages/analytics-types/index.d.ts
+++ b/packages/analytics-types/index.d.ts
@@ -73,6 +73,11 @@ export interface FirebaseAnalytics {
    * window['ga-disable-analyticsId'] = true;
    */
   setAnalyticsCollectionEnabled(enabled: boolean): void;
+
+  /**
+   * Sets cookie flags on gtag
+   * */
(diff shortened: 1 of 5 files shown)


hobbydevs:
@googlebot I signed it!

Feiyang1:
Hi @hobbydevs, Thanks for the contribution! This seems to be a useful thing to support, but we have to go through an API review process before updating/adding any public APIs. I will take it to the team and discuss it internally. Will report back once we make a decision about adding or not adding this API.

As a workaround for now, you can use the raw `gtag()` command to set the `cookie_flags` (Firebase Analytics and gtag.js are fully interoperable). Thanks!

PR #3850, opened 2020-09-28
Fix custom trace event dispatch before initialization for Performance Monitoring

zijianjoy:
Developer reported that `custom trace` right after initializing perf object was not able to be recorded. That is because initialization didn't finish, thus was not able to flip the bit of `SettingsService.loggingEnabled` from false to true. Therefore custom trace was discarded.

### Fix

Check `SettingsService.loggingEnabled` after initialization finishes, thus avoid dropping events before reading RemoteConfig value.

Internal tracking bug: b/169126072

### DIFF
--- a/packages-exp/performance-exp/src/services/perf_logger.test.ts
+++ b/packages-exp/performance-exp/src/services/perf_logger.test.ts
@@ -96,7 +96,7 @@ describe('Performance Monitoring > perf_logger', () => {
   });
 
   describe('logTrace', () => {
-    it('will not drop event before initialization finishes', async () => {
+    it('will not drop custom events sent before initialization finishes', async () => {
       getIidStub.returns(IID);
       stub(attributeUtils, 'getVisibilityState').returns(VISIBILITY_STATE);
       stub(initializationService, 'isPerfInitialized').returns(false);
--- a/packages-exp/performance-exp/src/services/perf_logger.test.ts
+++ b/packages-exp/performance-exp/src/services/perf_logger.test.ts
@@ -70,7 +70,6 @@ describe('Performance Monitoring > perf_logger', () => {
   }
 
   setupApi(self);
-
   const fakeFirebaseApp = ({
     options: { appId: APP_ID }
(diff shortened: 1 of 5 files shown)


packages-exp/performance-exp/src/services/perf_logger.test.ts
```diff
@@ -98,6 +96,29 @@ describe('Performance Monitoring > perf_logger', () => {
   });
 
   describe('logTrace', () => {
+    it('will not drop event before initialization finishes', async () => {
```
alikn, line 99: How about something like "will not drop custom events sent before initialization finishes" to make it more specific?

alikn (approved):

jposuna (approved):

packages-exp/performance-exp/src/services/perf_logger.test.ts
zijianjoy: Done.

packages/performance/src/services/perf_logger.ts
```diff
@@ -120,18 +120,12 @@ export function logTrace(trace: Trace): void {
...
   }
 
-  if (
-    !settingsService.loggingEnabled ||
-    !settingsService.logTraceAfterSampling
-  ) {
-    return;
-  }
-
   if (isPerfInitialized()) {
```
Feiyang1, line 123: Nit: We can always use the promise chain without checking `isPerfInitialized()`

Feiyang1 (approved):

Feiyang1 (changes requested):
LGTM. Can you please add changeset? Is there a GitHub issue we can reference?

packages/performance/src/services/perf_logger.ts
zijianjoy: That is true, however, Fireperf initialization state will only change once and will stay true after initialization. Check for initialization state in synchronized way is more straightforward and efficient.

Feiyang1 (approved):

Issue #3849, opened 2020-09-26
Callable functions internal error


kubanac95:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10 Pro - 1909
  * Browser version: Chrome - 85.0.4183.121
  * Firebase SDK version: 7.21.1
  * Firebase Product: functions

### [REQUIRED] Describe the problem

I have been working on a chrome extension trying to integrate firebase js SDK for authentication alongside other services.

I made the authentication working fine as well as firestore for reading out documents, but when I tried calling a fireabse callable function I tun into some strange issues.

SDK kept returning an error with `{ "code": "internal" }`

![image](https://user-images.githubusercontent.com/16191931/94349689-319d4300-0047-11eb-9d9f-ce6728b91e4f.png)

At first, I taught it was an issue with my implementation, that I'm missing some permissions or CSP since it was working in my web app. After hours of scratching my head, I decided to try a different firebase SDK version.

Strangely enough, that fixed the problem. Going down from `7.21.1` to `7.21.0` resolved the issues and the internal error lo longer occurs.

I think this will be enough to pinpoint the issue since it seems to be a transition between those two versions that caused the issue.

Code to 

#### Relevant Code:
    firebase
      .functions()
      .httpsCallable("onCreateSearch")({
        url: "https://github.com/firebase/firebase-js-sdk",
      })
      .then(console.log)
      .catch((error) => {
        console.log(error);
        console.log(JSON.stringify(error));
      });

hsubox76:
Seems likely to be related to PR https://github.com/firebase/firebase-js-sdk/pull/3782

Will take a look.

hsubox76:
For my own notes - because I did not add a changeset to #3782, @firebase/functions was not released with 7.21.1, but the yarn.lock file with isomorphic-fetch removed would have been. So the dep was removed but the code using it was not updated. 7.22.0 did publish the functions package because it included a changeset for the customDomain change, and consequently introduced the bind error.

TLDR, both 7.21.1 and 7.22.0 will have functions callable errors, for different reasons.

PR #3853, opened 2020-09-28
Functions: Bind passed fetch function

hsubox76:
Fix a bug caused by PR https://github.com/firebase/firebase-js-sdk/pull/3782 to replace isomorphic-fetch dependency in functions. Bug occurs in Chrome extensions and possibly other environments. When passing `fetch` function as an argument it loses `window` context. Binding `self` so that it will work in service workers as well.

See https://stackoverflow.com/questions/44720448/fetch-typeerror-failed-to-execute-fetch-on-window-illegal-invocation for a similar issue.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3849

### DIFF
--- a/.changeset/dirty-rings-cry.md
+++ b/.changeset/dirty-rings-cry.md
@@ -1,0 +1,6 @@
+---
+'@firebase/functions-exp': patch
+'@firebase/functions': patch
+---
+
+Fixes a bug introduced in #3782 that causes callable functions to throw an error in browser extensions.
--- a/.changeset/dirty-rings-cry.md
+++ b/.changeset/dirty-rings-cry.md
@@ -1,5 +1,4 @@
 ---
-'@firebase/functions-exp': patch
 '@firebase/functions': patch
 ---
 
--- a/packages-exp/functions-exp/src/index.ts
+++ b/packages-exp/functions-exp/src/index.ts
@@ -21,5 +21,5 @@ import { name, version } from '../package.json';
 
 export * from './api';
 
-registerFunctions(window.fetch.bind(window));
+registerFunctions(fetch.bind(self));
 registerVersion(name, version);
--- a/packages-exp/functions-exp/src/index.ts
+++ b/packages-exp/functions-exp/src/index.ts
@@ -21,5 +21,5 @@ import { name, version } from '../package.json';
 
 export * from './api';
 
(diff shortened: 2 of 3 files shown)


Feiyang1 (changes requested):
Can you please add changeset?

hsubox76:
> Can you please add changeset?

Changeset added!

.changeset/dirty-rings-cry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@firebase/functions-exp': patch
```
Feiyang1, line 2: Please remove `@firebase/functions-exp`. Exp packages should not be included in changeset.

Feiyang1 (approved):

zaripych:
@hsubox76  +1 for fixing the security issue with `node-fetch` earlier, just a bit of relevant information here that the bug actually used to happen not just in Chrome extensions. I have a CRA app (built with webpack) and was hitting this one. It was a little tricky to diagnose because exception message was hidden in here: 

https://github.com/firebase/firebase-js-sdk/blob/master/packages/functions/src/api/service.ts#L200

I managed to dig-out the error to be `"TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation"`.

Anyway, with new `firebase` package release the issue is fixed. Just a little feedback - it would probably be great if that `e` variable is somehow "converted" to a proper error JSON object so that information is not lost and it's easy to diagnose what is happening. 

Thanks!

hsubox76:
Thanks for the additional information that it isn't just Chrome extensions.

As for throwing the error instead of hiding it, I did look into that, but it's tricky. Most of the time, the error that comes from the try/catch around the "fetch" is simply "Failed to fetch" (on Chrome) which will happen if it's a network error, but also if there is an error in the actual Cloud Function code, which would be misleading and cause developers to waste time trying to debug network issues instead of checking the function code.  So when it was originally coded, the decision was made to hide the error because it was potentially more misleading than helpful.

If there was a way to detect and hide "Failed to fetch" errors, we could throw all other errors. However it seems brittle to detect it based on message text, which could change, and the type is TypeError, same as the one we would want to show, above. Additionally, on Firefox the error text is "NetworkError when attempting to fetch resource" and who knows what it is in other browsers.

PR #3857, opened 2020-09-28
Remove cyclic references in RemoteDocumentCache

schmidt-sebastian:
This PR removes a static inheritance that WebPack cannot tree-shake. 

Before, our RemoteDocumentCaches would look like this:

```
class XYZRemoteDocumentCache {
static class XYZRemoteDocumentChangeBuffer extends RemoteDocumentChangeBuffer {...}
}
```

This patterns breaks code optimization in Webpack (as per @Feiyang1's research), but it does allow us to access "private" members within the enclosing class while also extending from another class. To address the Webpack issue while keeping some of the encapsulation, I:

- Extracted the XYZRemoteDocumentChangeBuffers into top-level classes.
- Changed XYZRemoteDocumentCaches to XYZRemoteDocumentCacheImpls
- Made XYZRemoteDocumentCacheImpls methods public, so they can be used by the newly extracted types

The rest of the repo still refers to the old types, which are now interfaces and hide the new public methods. This is the same pattern that I applied in lots of other places already (see LocalStore), and it also allowed me to make two functions tree-shakeable that are only used in Multi-Tab (`getNewDocumentChanges()` and `getReadTime()`).

There are no logic changes.

### DIFF
--- a/packages/firestore/lite/src/api/transaction.ts
+++ b/packages/firestore/lite/src/api/transaction.ts
@@ -123,9 +123,14 @@ export class Transaction {
    *
    * @param documentRef A reference to the document to be set.
    * @param data An object of the fields and values for the document.
+   * @param options An object to configure the set behavior.
    * @return This `Transaction` instance. Used for chaining method calls.
    */
-  set<T>(documentRef: DocumentReference<T>, data: Partial<T>): this;
+  set<T>(
+    documentRef: DocumentReference<T>,
+    data: Partial<T>,
+    options: SetOptions
+  ): this;
   set<T>(
     documentRef: DocumentReference<T>,
     value: T,
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -35,7 +35,7 @@ import { SortedMap } from '../util/sorted_map';
 import { SortedSet } from '../util/sorted_set';
 
 import { SnapshotVersion } from '../core/snapshot_version';
-import { debugAssert, fail, hardAssert } from '../util/assert';
+import { debugAssert, debugCast, fail, hardAssert } from '../util/assert';
 import { IndexManager } from './index_manager';
 import { IndexedDbPersistence } from './indexeddb_persistence';
 import {
@@ -58,14 +58,23 @@ import { RemoteDocumentChangeBuffer } from './remote_document_change_buffer';
 import { IterateOptions, SimpleDbStore } from './simple_db';
 import { ObjectMap } from '../util/obj_map';
 
-export class IndexedDbRemoteDocumentCache implements RemoteDocumentCache {
+export interface IndexedDbRemoteDocumentCache extends RemoteDocumentCache {
+  // The IndexedDbRemoteDocumentCache doesn't implement any methods on top
+  // of RemoteDocumentCache. This class exists for consistency.
+}
+
+/**
+ * The RemoteDocumentCache for IndexedDb. To construct, invoke
+ * `newIndexedDbRemoteDocumentCache()`.
+ */
+export class IndexedDbRemoteDocumentCacheImpl implements RemoteDocumentCache {
   /**
-   * @param {LocalSerializer} serializer The document serializer.
-   * @param {IndexManager} indexManager The query indexes that need to be maintained.
+   * @param serializer The document serializer.
+   * @param indexManager The query indexes that need to be maintained.
    */
   constructor(
     readonly serializer: LocalSerializer,
-    private readonly indexManager: IndexManager
+    readonly indexManager: IndexManager
   ) {}
 
   /**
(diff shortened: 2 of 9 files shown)


packages/firestore/lite/src/api/transaction.ts
```diff
@@ -123,9 +123,14 @@ export class Transaction {
...
    * @param documentRef A reference to the document to be set.
    * @param data An object of the fields and values for the document.
+   * @param options An object to configure the set behavior.
    * @return This `Transaction` instance. Used for chaining method calls.
    */
-  set<T>(documentRef: DocumentReference<T>, data: Partial<T>): this;
+  set<T>(
+    documentRef: DocumentReference<T>,
+    data: Partial<T>,
+    options: SetOptions
```
schmidt-sebastian, line 132: This somehow doesn't compile in this PR, so I had to fix this. This likely broke when I merged https://github.com/firebase/firebase-js-sdk/pull/3843, but I don't know how I was able to merge it.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -58,14 +58,23 @@ import { RemoteDocumentChangeBuffer } from './remote_document_change_buffer';
...
+export interface IndexedDbRemoteDocumentCache extends RemoteDocumentCache {
+  // The IndexedDbRemoteDocumentCache doesn't implement any methods on top
+  // of RemoteDocumentCache. This class exists for consistency.
+}
+
+/**
+ * The RemoteDocumentCache for IndexedDb. To construct, invoke
+ * `newIndexedDbRemoteDocumentCache()`.
+ */
+export class IndexedDbRemoteDocumentCacheImpl implements RemoteDocumentCache {
```
dconeybe, line 70: Can "export" be removed from the declaration of `IndexedDbRemoteDocumentCacheImpl`? It doesn't appear to be used outside of this file and its name implies that it is an implementation detail that should not be exposed.

dconeybe, line 70: Should `IndexedDbRemoteDocumentCache` implement `IndexedDbRemoteDocumentCache` instead of `RemoteDocumentCache`?

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -50,7 +50,18 @@ function documentEntryMap(): DocumentEntryMap {
...
+    transaction: PersistenceTransaction,
+    f: (key: DocumentKey) => PersistencePromise<void>
+  ): PersistencePromise<void>;
+}
+
+/**
+ * The memory-only RemoteDocumentCache for IndexedDb. To construct, invoke
+ * `newMemoryRemoteDocumentCache()`.
+ */
+class MemoryRemoteDocumentCacheImpl implements RemoteDocumentCache {
```
dconeybe, line 64: Should `MemoryRemoteDocumentCacheImpl` implement `MemoryRemoteDocumentCache` instead of `RemoteDocumentCache`?

packages/firestore/src/local/memory_remote_document_cache.ts
```diff
@@ -183,49 +194,63 @@ export class MemoryRemoteDocumentCache implements RemoteDocumentCache {
...
+ *
+ * @param indexManager A class that manages collection group indices.
+ * @param sizer Used to assess the size of a document. For eager GC, this is expected to just
+ * return 0 to avoid unnecessarily doing the work of calculating the size.
+ */
+export function newMemoryRemoteDocumentCache(
+  indexManager: IndexManager,
+  sizer: DocumentSizer
+): MemoryRemoteDocumentCache {
+  return new MemoryRemoteDocumentCacheImpl(indexManager, sizer);
```
dconeybe, line 216: TypeScript noobie question: How does this compile since `MemoryRemoteDocumentCacheImpl` does not implement `MemoryRemoteDocumentCache`?

dconeybe (approved):

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: Yes. Thanks for catching.

packages/firestore/src/local/memory_remote_document_cache.ts
schmidt-sebastian: TypeScript interfaces are enforced by duck-typing - if a type happens to implement all members of an interface, it is considered identical to a type that explicitly implements the interface. Explicitly implementing comes in handy when the methods are renamed or added as it will produce much saner errors.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: Yes. Thanks for catching.

schmidt-sebastian: The entire idea behind the XYZ and XYZImpl is that XYZImpl is not exported, so that the public fields in XYZImpl are only visible within the module. So, yes, this should not be exported.

PR #3866, opened 2020-09-30
Throw exception when Performance custom trace API is provided with non-positive startTime or duration

zijianjoy:
It is common for developer to provide 0 or invalid value for `startTime` and `duration` when calling `record(startTime, duration, options)` to create custom trace. This doesn't create error on the client side, but will be filtered silently during backend processing. Add basic validation to check developer input of both variables on the client side.

### DIFF
--- a/packages-exp/performance-exp/src/utils/errors.ts
+++ b/packages-exp/performance-exp/src/utils/errors.ts
@@ -21,6 +21,8 @@ import { SERVICE, SERVICE_NAME } from '../constants';
 export const enum ErrorCode {
   TRACE_STARTED_BEFORE = 'trace started',
   TRACE_STOPPED_BEFORE = 'trace stopped',
+  NONPOSITIVE_TRACE_START_TIME = '',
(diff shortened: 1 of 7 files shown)


packages-exp/performance-exp/src/utils/errors.ts
```diff
@@ -21,6 +21,8 @@ import { SERVICE, SERVICE_NAME } from '../constants';
 export const enum ErrorCode {
   TRACE_STARTED_BEFORE = 'trace started',
   TRACE_STOPPED_BEFORE = 'trace stopped',
+  NONPOSITIVE_TRACE_START_TIME = '',
```
alikn, line 24: Please add non-empty unique strings as enum values.

zijianjoy: Thank you Ali! Updated.

Feiyang1 (approved):

jposuna (approved):

alikn (approved):

Issue #3862, opened 2020-09-29
FieldValue.serverTimestamp() no longer serializable


cmditch:
* Operating System version: macOS 10.15.6
  * Browser version: Chrome 85.0.4183.121
  * Firebase SDK version: 7.21.1
  * Firebase Product: firestore

This change appeared after 7.15.0 (worked in 7.14.6).

```
console.log(JSON.stringify(firebase.firestore.FieldValue.serverTimestamp()));

> Uncaught TypeError: Converting circular structure to JSON
>    --> starting at object with constructor 'n'
>    --- property 's_' closes the circle
>    at JSON.stringify
```

The value references itself.
![Screen Shot 2020-09-29 at 10 45 30 AM](https://user-images.githubusercontent.com/15849320/94607071-aa271e00-0258-11eb-98d8-60f608047476.png)

schmidt-sebastian:
@cmditch Thanks for filing this. Almost all of our types will have this problem and I am trying to understand what the user impact is here. Can you describe your use case?

schmidt-sebastian:
@cmditch Never mind. This is already fixed in the release that is going out this week. Thanks for catching it.

cmditch:
In short, we [granularly track](https://github.com/RealKinetic/elm-firestore/blob/master/src/index.d.ts#L96) the state of a document, and need to serialize this latest version of the doc to pass off to our runtime (Elm).

Our stack is [elm-lang](https://elm-lang.org) + firebase. Elm is a compile-to-js language. If you want to talk to native javascript (e.g. firebase-js-sdk) from Elm you must use [ports](https://guide.elm-lang.org/interop/ports.html). Ports rely on `JSON.stringify` which allow [decoders](https://guide.elm-lang.org/effects/json.html#json-decoders) to work their magic. Long story short, since Elm <-> JS interop is not callback based, the cleanest way we've found to update the state of our Elm app is to pass the entire firestore document each time an alteration occurs.

The library we've written ([elm-firestore](https://github.com/RealKinetic/elm-firestore/blob/master/src/index.d.ts#L62)) also uses [hooks](url) which allow the user to [massage document data](https://github.com/RealKinetic/elm-firestore/blob/master/example/index.ts#L56) in ways that Elm can't, e.g. adding `FieldValue.serverTimestamp()` to a field.

schmidt-sebastian:
This has already been fixed by https://github.com/firebase/firebase-js-sdk/pull/3819

PR #3869, opened 2020-09-30
Add tests to ensure that FieldValue has no cyclic references

schmidt-sebastian:
Addresses https://github.com/firebase/firebase-js-sdk/issues/3862


packages/firestore/test/unit/api/field_value.test.ts
```diff
@@ -33,4 +33,16 @@ describe('FieldValue', () => {
...
+
+  it('supports JSON.stringify()', () => {
+    try {
+      JSON.stringify(FieldValue.delete());
+      JSON.stringify(FieldValue.serverTimestamp());
+      JSON.stringify(FieldValue.increment(1));
+      JSON.stringify(FieldValue.arrayUnion(2));
+      JSON.stringify(FieldValue.arrayRemove(3));
+    } catch (e) {
+      expect.fail('Unexpected error: ' + e.message);
```
dconeybe, line 45: Can you just let the exception propagate to fail the test instead of catching it explicitly? As it is, the fail message will not indicate which call threw an exception but if the exception just propagates then I assume that a full stack trace would be provided, indicating which call failed.

dconeybe (approved):

packages/firestore/test/unit/api/field_value.test.ts
schmidt-sebastian: I wrote the test this way to ensure that it is obvious that I am only verifying that the API calls do not throw. Let me do this with a comment instead and remove the try/catch block.

Issue #3852, opened 2020-09-28
New Errors Logged after Enabling Persistence


sanaze:
* Operating System version: Multiple
  * Browser version: Multiple
  * Firebase SDK version: 7.21.1
  * Firebase Product: firestore

### Describe the problem

This weekend we enabled persistence with tab synchronization on a web app where we did not have persistence enabled before.

After enabling it, a number of new errors started showing up in our production error reporting logs:

`Firestore (7.21.1): Failed to obtain primary lease for action 'Release target'.`

`Firestore (7.21.1): Failed to obtain primary lease for action 'Apply remote event'.`

`FIRESTORE (7.21.1) INTERNAL ASSERTION FAILED: Unexpected state`

`Firestore (7.21.1): Detected an update time that is in the future: YYY > ZZZ [where YYY and ZZZ are numeric timestamps]`

`IndexedDB transaction failed: UnknownError: Internal error opening backing store for indexedDB.open`

`IndexedDB transaction failed: AbortError: The transaction was aborted, so the request cannot be fulfilled.`

These are infrequent but affect multiple users. We haven't observed the errors in our own testing and development. We also didn't receive any bug reports from users that appear to be related to them.

I can't find description on a lot of these, so it is hard to assess their severity. It would be helpful to get additional documentation on them.

Are they something that only affect the newly enabled persistence layer (so the worse case is the app works fine but without persistence)?

Benny739:
Same here, we're seeing these errors in sentry but it looks like they are not affecting the app.

schmidt-sebastian:
@sanaze / @Benny739: Thanks for filing this issue.

> Firestore (7.21.1): Failed to obtain primary lease for action 'Release target'.
> Firestore (7.21.1): Failed to obtain primary lease for action 'Apply remote event'.

Multi-tab elects a primary tab. When the primary tab becomes unresponsive, another tab may take over and "steal" the primary lease. These error messages are likely related to this. The SDK should automatically recover from these errors.

> FIRESTORE (7.21.1) INTERNAL ASSERTION FAILED: Unexpected state

I would need a stack trace to figure out what this is. The SDK will not recover from this and will no longer function.

> Firestore (7.21.1): Detected an update time that is in the future: YYY > ZZZ [where YYY and ZZZ are numeric timestamps]

[...]

PR #3871, opened 2020-09-30
Add IndexedDB action to error message

schmidt-sebastian:
This propagates the action name that we already use in IndexedDbPersistence to SimpleDb, which should give us more insights when users provide IndexedDB stacktraces that do not have a cause associated.

Addresses https://github.com/firebase/firebase-js-sdk/issues/3852

### DIFF
--- a/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
+++ b/packages/firestore/test/unit/local/indexeddb_persistence.test.ts
@@ -1282,7 +1282,7 @@ describe('IndexedDb: allowTabSynchronization', () => {
       async db1 => {
         db1.injectFailures = ['getHighestListenSequenceNumber'];
         await expect(db1.start()).to.eventually.be.rejectedWith(
-          'IndexedDB transaction failed'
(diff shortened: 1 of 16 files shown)


packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -262,6 +264,7 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
 
     return withDb(2, db => {
       return db.runTransaction(
+        TEST_ACTION,
```
wilhuff, line 267: Is it possible to use the current test name, description, or similar here? This would benefit test failures similarly to what you're proposing for end users.

wilhuff (approved):
LGTM

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
schmidt-sebastian: We have to convert the tests to use proper functions, since they will be associated with Mocha's `this` context, which then provides the test name. I updated this file and SimpleDbTest.

PR #3876, opened 2020-10-01
Improve typing for options.auth

samtstern:
### Discussion

This is a non-breaking change to improve the typing for `options.auth` in `initializeTestApp`.  Note that this removes the `uid` property of the token which should never have been there in the first place (`sub` is the canonical `uid` field).

Related issues: 
  - https://github.com/firebase/quickstart-testing/issues/2
  - https://github.com/firebase/firebase-tools/issues/2405
  - http://b/153398707 (internal)

### Testing

Improved existing unit tests

### API Changes

N/A

### DIFF
--- a/.changeset/chilled-ways-promise.md
+++ b/.changeset/chilled-ways-promise.md
@@ -1,0 +1,5 @@
+---
+'@firebase/rules-unit-testing': patch
+---
+
+Add stronger types to the 'options.auth' option for initializeTestApp
--- a/packages/rules-unit-testing/src/api/index.ts
+++ b/packages/rules-unit-testing/src/api/index.ts
@@ -41,25 +41,118 @@ let _databaseHost: string | undefined = undefined;
 /** The actual address for the Firestore emulator */
 let _firestoreHost: string | undefined = undefined;
 
-/** Create an unsecured JWT for the given auth payload. See https://tools.ietf.org/html/rfc7519#section-6. */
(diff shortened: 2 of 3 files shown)


yuchenshi (approved):

mdirolf:
When using custom claims in firestore rules I grab them off the top-level token object, as documented here https://firebase.google.com/docs/auth/admin/custom-claims:
```
    function isAdmin() {
      return request.auth.token.admin;
    }
```

In testing I was able to create an app instance that passed those rules like this:
```
firebaseTesting.initializeTestApp({
    projectId,
    auth: {
      uid: 'mike', admin: true, firebase: {
        sign_in_provider: 'google.com'
      }
    }
  })
```

This change makes that `admin: true` a type error, so I tried moving it into a claims object like:
```
firebaseTesting.initializeTestApp({
    projectId,
    auth: {
      uid: 'mike', claims: {admin: true}, firebase: {
        sign_in_provider: 'google.com'
      }
    }
  })
```

That passes type checking but doesn't seem to work when testing those security rules. Am I doing something wrong or do custom claims need to get hoisted up a level at some point?

samtstern:
@mdirolf hmmm that's interesting, would you mind opening a new issue rather than discussing it here on this closed PR?

samtstern:
@mdirolf fixed in #3915  !

mdirolf:
@samtstern thanks! sorry for not opening a new issue to begin with

PR #3878, opened 2020-10-01
Fix a couple of parity issues for auth-exp

samhorlbeck:
* State change listeners before initialization completes now fire after initialization is done
 * Deserializing user objects now does not care if the a value is null or undefined

b/169870894  
b/169870246

### DIFF
--- a/packages-exp/auth-exp/src/core/user/user_impl.ts
+++ b/packages-exp/auth-exp/src/core/user/user_impl.ts
@@ -219,20 +219,20 @@ export class UserImpl implements User {
   }
 
   static _fromJSON(auth: Auth, object: PersistedBlob): User {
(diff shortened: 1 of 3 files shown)


packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -219,20 +219,20 @@ export class UserImpl implements User {
   }
 
   static _fromJSON(auth: Auth, object: PersistedBlob): User {
+    const displayName = object.displayName ?? undefined;
```
avolkovi, line 222: do you wanna test at least one of these with null/undefined?

avolkovi (approved):
nice and clean ✨

PR #3883, opened 2020-10-01
Reference docs for Auth next

Feiyang1:
1. Integrate api-extractor into auth-next
2. Generate API report for auth-next
3. Generate reference docs for auth-next

There is a bunch of warning in the API report, which we should fix, e.g. adding documentation and release tag.


samhorlbeck:
Is it possible to have it ignore `_prefixed` methods and properties?

hsubox76 (approved):
Wow, that is a lot of files.

Feiyang1:
@samhorlbeck Yes! You can annotate those methods with `@internal` like [this](https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/app-exp/src/internal.ts#L39), then `api-extractor` can generate a rollup d.ts file with them removed.

@schmidt-sebastian is working on an [utility](https://github.com/firebase/firebase-js-sdk/pull/3790) that strips away `_prefixed` things automatically without the need to annotating them. It seems to be close to be ready, so you might want to wait for it if adding `@internal`s feels too tedious.

samhorlbeck (approved):

PR #3888, opened 2020-10-02
Add GMPID Header to Firestore

schmidt-sebastian:


packages/firestore/src/platform/node/grpc_connection.ts
```diff
@@ -51,6 +55,7 @@ function createMetadata(databasePath: string, token: Token | null): Metadata {
       }
     }
   }
+  metadata.set('x-firebase-gmpid', appId);
```
var-const, line 58: Question: the plumbing accounts for the possibility of `appId` being empty. What do you think about not setting the header at all if `appId` is empty rather than setting an "empty" header?

packages/firestore/src/remote/rest_connection.ts
```diff
@@ -118,6 +118,7 @@ export abstract class RestConnection implements Connection {
     token: Token | null
   ): void {
     headers['X-Goog-Api-Client'] = X_GOOG_API_CLIENT_VALUE;
+    headers['X-Firebase-GMPID'] = this.databaseInfo.appId;
```
var-const, line 121: Nit: capitalization of the header name is different between this file and `grpc_connection`. Is this intentional?

packages/firestore/src/platform/node/grpc_connection.ts
schmidt-sebastian: I looked at the RTDB code and it uses an empty app ID and sends the header. I wanted to match that behavior.

packages/firestore/src/remote/rest_connection.ts
schmidt-sebastian: The goal was to match the surrounding code. Let me fix the GRPC headers.

var-const (approved):

schmidt-sebastian:
This is blocked on https://critique-ng.corp.google.com/cl/335516871 (hence the failing tests)

Feiyang1:
Seems we are ready to merge this PR, since https://critique-ng.corp.google.com/cl/335516871 has been submitted.

schmidt-sebastian:
We have been waiting for the CL to be rolled out, but this has now been done.

PR #3896, opened 2020-10-05
Control the maximum amount of events in single logRequest to be 1000 in Performance Monitoring

zijianjoy:
Backend starts rejecting logRequest with more than 1000 events in batch. In order for client event to be accepted and avoid any unintentional flood, put a cap for each logRequest dispatch to be 1000, extra events will stay in queue until next dispatch.

### DIFF
--- a/packages-exp/performance-exp/src/services/transport_service.test.ts
+++ b/packages-exp/performance-exp/src/services/transport_service.test.ts
@@ -57,7 +57,7 @@ describe('Firebase Performance > transport_service', () => {
     }).to.throw;
   });
 
-  it('does not attempt to log an event to cc after INITIAL_SEND_TIME_DELAY_MS if queue is empty', () => {
+  it('does not attempt to log an event after INITIAL_SEND_TIME_DELAY_MS if queue is empty', () => {
     fetchStub.resolves(
       new Response('', {
         status: 200,
@@ -68,7 +68,8 @@ describe('Firebase Performance > transport_service', () => {
     clock.tick(INITIAL_SEND_TIME_DELAY_MS);
     expect(fetchStub).to.not.have.been.called;
   });
-  it('attempts to log an event to cc after DEFAULT_SEND_INTERVAL_MS if queue not empty', async () => {
+
+  it('attempts to log an event after DEFAULT_SEND_INTERVAL_MS if queue not empty', async () => {
     fetchStub.resolves(
       new Response('', {
         status: 200,
@@ -104,8 +105,8 @@ describe('Firebase Performance > transport_service', () => {
 
     // Returns successful response from fl for logRequests.
     const response = generateSuccessResponse();
-    fetchStub.resolves(response);
     stub(response, 'json').resolves(JSON.parse(generateSuccessResponseBody()));
+    fetchStub.resolves(response);
 
     // Act
     // Generate 1020 events, which should be dispatched in two batches (1000 events and 20 events).
@@ -114,6 +115,7 @@ describe('Firebase Performance > transport_service', () => {
     }
     // Wait for first and second event dispatch to happen.
     clock.tick(INITIAL_SEND_TIME_DELAY_MS);
+    // This is to resolve the floating promise chain in transport service.
     await Promise.resolve().then().then().then();
     clock.tick(DEFAULT_SEND_INTERVAL_MS);
 
--- a/packages-exp/performance-exp/src/services/transport_service.test.ts
+++ b/packages-exp/performance-exp/src/services/transport_service.test.ts
@@ -31,6 +31,8 @@ describe('Firebase Performance > transport_service', () => {
   let fetchStub: SinonStub<[RequestInfo, RequestInit?], Promise<Response>>;
   const INITIAL_SEND_TIME_DELAY_MS = 5.5 * 1000;
   const DEFAULT_SEND_INTERVAL_MS = 10 * 1000;
+  const MAX_EVENT_COUNT_PER_REQUEST = 1000;
+  const TRANSPORT_DELAY_INTERVAL = 30000;
   // Starts date at timestamp 1 instead of 0, otherwise it causes validation errors.
   let clock: SinonFakeTimers;
   const testTransportHandler = transportHandler((...args) => {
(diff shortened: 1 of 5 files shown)


jposuna (approved):

packages-exp/performance-exp/src/services/transport_service.ts
```diff
@@ -90,9 +91,11 @@ function processQueue(timeOffset: number): void {
 }
 
 function dispatchQueueEvents(): void {
-  // Capture a snapshot of the queue and empty the "official queue".
-  const staged = [...queue];
-  queue = [];
+  // Extract events up to the maximum cap of single logRequest from top of "official queue".
+  // The staged events will be used for current logRequest attempt, remaining events will be kept
+  // for next attempt.
+  const staged = queue.slice(0, MAX_EVENT_COUNT_PER_REQUEST);
```
alikn, line 97: use splice instead.

packages/performance/src/services/transport_service.test.ts
```diff
@@ -78,36 +80,128 @@ describe('Firebase Performance > transport_service', () => {
     clock.tick(INITIAL_SEND_TIME_DELAY_MS);
     testTransportHandler('someEvent');
     clock.tick(DEFAULT_SEND_INTERVAL_MS);
+    await Promise.resolve();
```
alikn, line 83: Why do we need this?

packages/performance/src/services/transport_service.test.ts
```diff
@@ -78,36 +80,128 @@ describe('Firebase Performance > transport_service', () => {
...
-      )
+      generateSuccessResponse()
     );
 
     testTransportHandler('event1');
     clock.tick(INITIAL_SEND_TIME_DELAY_MS);
     expect(fetchStub).to.have.been.calledOnce;
   });
+
+  it('sends up to the maximum event limit in one request', async () => {
```
alikn, line 101: This unit test is hard to follow. Can we instead have one in which there is more than 1000 logs in the queue and check the number of logs received by fetch stub to be equal to 1000?

packages-exp/performance-exp/src/services/transport_service.ts
zijianjoy: Good suggestion! Applied `splice()` to split the array.

packages/performance/src/services/transport_service.test.ts
zijianjoy: Removed.

packages/performance/src/services/transport_service.test.ts
zijianjoy: Simplified unit test to capture only the first dispatch to be equal to 1000 events. Since we also have the splice logic of keeping the remaining events in the queue, added validation on remaining event count is kept.

packages-exp/performance-exp/src/services/transport_service.test.ts
```diff
@@ -82,32 +84,102 @@ describe('Firebase Performance > transport_service', () => {
...
+    // Returns successful response from fl for logRequests.
+    const response = generateSuccessResponse();
+    fetchStub
+      .withArgs(flTransportFullUrl, {
+        method: 'POST',
+        body: JSON.stringify(logRequest)
+      })
+      .resolves(response);
+    stub(response, 'json').returns(
+      Promise.resolve(JSON.parse(generateSuccessResponseBody()))
```
alikn, line 123: Here we can use stub(...).resolve instead of stub(...).returns(Promise.resolve(...))

packages-exp/performance-exp/src/services/transport_service.test.ts
```diff
@@ -82,32 +84,102 @@ describe('Firebase Performance > transport_service', () => {
...
+    // Assert
+    clock.tick(INITIAL_SEND_TIME_DELAY_MS);
+    // First logRequest has been called.
+    expect(fetchStub).to.have.been.calledWith(flTransportFullUrl, {
+      method: 'POST',
+      body: JSON.stringify(logRequest)
+    });
+    // Wait for async action to call for next dispatch cycle.
+    await Promise.resolve()
+      .then(() => {
```
alikn, line 141: I think this is equivalent to 
Promise.resolve()
.then(() => {
clock.tick(1);
clock.tick(1);
}

packages-exp/performance-exp/src/services/transport_service.test.ts
```diff
@@ -82,32 +84,102 @@ describe('Firebase Performance > transport_service', () => {
...
+        clock.tick(1);
+      })
+      .then(() => {
+        clock.tick(1);
+      });
+    clock.tick(DEFAULT_SEND_INTERVAL_MS * 3);
+    // Remaining events stay in queue.
+    expect(readQueue().length).to.be.equal(20);
```
alikn, line 149: Shouldn't the 20 logs be sent in the next batch?

packages-exp/performance-exp/src/services/transport_service.ts
```diff
@@ -65,6 +66,10 @@ export function setupTransportService(): void {
   }
 }
 
+export function readQueue(): BatchEvent[] {
```
alikn, line 69: Let's avoid exposing the internals of the service just for testing.

packages-exp/performance-exp/src/services/transport_service.test.ts
zijianjoy: Updated, thank you Ali!

packages-exp/performance-exp/src/services/transport_service.test.ts
zijianjoy: Thank you Ali! After investigation, I found that `clock.tick()` is not required in this promise chain, removing them.

packages-exp/performance-exp/src/services/transport_service.test.ts
zijianjoy: That is correct, switched test logic to validate two batches with 1000 and 20 events each.

packages-exp/performance-exp/src/services/transport_service.ts
zijianjoy: I agree, I removed this exposure, and validate the `fetch` calls with expected parameter instead.

alikn (approved):

hsubox76 (approved):

PR #3900, opened 2020-10-05
Update auth compat polyfill to use composition instead of inheritance

samhorlbeck:
Big change to switch to composition over inheritance.

The changes in `auth-exp` are mostly just cleaning up the complexity that was required for inheritance. The important bits are in the `auth-compat-exp` folder.

Also switched `auth-compat-exp` to depend on `app-compat` instead of `app`

### DIFF
--- a/common/api-review/auth-exp.api.md
+++ b/common/api-review/auth-exp.api.md
@@ -10,6 +10,7 @@ import { ErrorFn } from '@firebase/util';
 import * as externs from '@firebase/auth-types-exp';
 import { FirebaseApp } from '@firebase/app-types-exp';
 import { FirebaseError } from '@firebase/util';
+import { NextFn } from '@firebase/util';
 import { ProviderId } from '@firebase/auth-types-exp';
 import { Unsubscribe } from '@firebase/util';
 import { UserCredential } from '@firebase/auth-types-exp';
@@ -39,17 +40,17 @@ export function applyActionCode(auth: externs.Auth, oobCode: string): Promise<vo
 // @public (undocumented)
 export class AuthCredential {
     protected constructor(providerId: string, signInMethod: string);
-    // Warning: (ae-forgotten-export) The symbol "AuthCore" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "Auth" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "PhoneOrOauthTokenResponse" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
-    _getIdTokenResponse(_auth: AuthCore): Promise<PhoneOrOauthTokenResponse>;
+    _getIdTokenResponse(_auth: Auth_2): Promise<PhoneOrOauthTokenResponse>;
     // (undocumented)
-    _getReauthenticationResolver(_auth: AuthCore): Promise<IdTokenResponse>;
+    _getReauthenticationResolver(_auth: Auth_2): Promise<IdTokenResponse>;
     // Warning: (ae-forgotten-export) The symbol "IdTokenResponse" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
-    _linkToIdToken(_auth: AuthCore, _idToken: string): Promise<IdTokenResponse>;
+    _linkToIdToken(_auth: Auth_2, _idToken: string): Promise<IdTokenResponse>;
     // (undocumented)
     readonly providerId: string;
     // (undocumented)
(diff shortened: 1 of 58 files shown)


common/api-review/auth-exp.api.md
```diff
@@ -39,17 +40,17 @@ export function applyActionCode(auth: externs.Auth, oobCode: string): Promise<vo
 // @public (undocumented)
 export class AuthCredential {
     protected constructor(providerId: string, signInMethod: string);
-    // Warning: (ae-forgotten-export) The symbol "AuthCore" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "Auth" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "PhoneOrOauthTokenResponse" needs to be exported by the entry point index.d.ts
     //
     // (undocumented)
-    _getIdTokenResponse(_auth: AuthCore): Promise<PhoneOrOauthTokenResponse>;
+    _getIdTokenResponse(_auth: Auth_2): Promise<PhoneOrOauthTokenResponse>;
```
avolkovi, line 47: Auth_2?

packages-exp/auth-compat-exp/src/user.ts
```diff
@@ -22,20 +22,41 @@ import {
   convertConfirmationResult,
   convertCredential
 } from './user_credential';
+import { unwrap, Wrapper } from './wrap';
+
+export class User implements compat.User, Wrapper<externs.User> {
+  // Maintain a map so that there's always a 1:1 mapping between new User and
+  // legacy compat users
+  private static readonly USER_MAP = new Map<externs.User, User>();
```
avolkovi, line 30: should this be a weak map? what happens when a user is deleted?

avolkovi (approved):
thanks for doing this

common/api-review/auth-exp.api.md
samhorlbeck: Yeah that's weird, but I think it references model.Auth instsead of externs.Auth. It should only be visible to internal methods so it should go away once we signal the documentation to ignore symbols prefixed with `_`.

packages-exp/auth-compat-exp/src/user.ts
samhorlbeck: Switched to weak map.

If the developer's code calls `user.delete()`, but retains a reference to the user object, they will still be able to see all the fields on the user-- we don't do any cleanup. So this implementation is still consistent with that

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -60,53 +59,69 @@ export class Auth
       sdkClientVersion: impl._getClientVersion(_getClientPlatform())
     };
 
-    super(app, config, User);
-    this.app = app;
+    this.auth = new impl.AuthImpl(app, config);
```
Feiyang1, line 62: The modular auth instance should be retrieved from the component framework and passed into the compat auth constructor instead of being created manually. Otherwise, auth-exp and auth-compat won't share the same state.

In compat auth factory:
```
        const app = container.getProvider('app').getImmediate();
        const auth = container.getProvider('auth-exp').getImmediate();
        return new Auth(app, auth);
```

It probably also means you need to register `auth-exp` with the component framework in '@firebase/auth-exp/internal'. You can move the registration code to a separate file, e.g. `register.ts` and make both `@firebase/auth-exp` and `@firebase/auth-exp/internal` import it.

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
```diff
@@ -15,15 +15,17 @@
...
 
-import firebase from '@firebase/app';
+import firebase from '@firebase/app-compat';
 import { FirebaseApp } from '@firebase/app-types';
 import * as impl from '@firebase/auth-exp/internal';
 import * as compat from '@firebase/auth-types';
 import * as externs from '@firebase/auth-types-exp';
+import { unwrap, Wrapper } from './wrap';
 
 export class RecaptchaVerifier
```
Feiyang1, line 25: Is it any different than the `RecaptchaVerifier` from `auth-exp`. Can we just reexport the one from `auth-exp`?

Feiyang1 (changes requested):

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
samhorlbeck: The auth-exp one needs an unwrapped auth object, which is why this has to be this way

packages-exp/auth-compat-exp/src/auth.ts
samhorlbeck: I tried this and I kept getting the error that it couldn't find the provider:

```
Uncaught Error: Service auth-exp is not available
    at Provider.getImmediate (provider.ts:101)
```

I've pushed up this version so you can take a look. Not sure what's going wrong

Feiyang1: Ah I see. It's the script issue with compat packages we chatted about before. Will report back once I get the compat script work.

samhorlbeck: Would you be willing to let me merge this with a TODO? Many of the remaining parity issues on my plate deal with the compat layer.

Feiyang1: https://github.com/firebase/firebase-js-sdk/pull/3916
It should work with this PR. I tested it locally. 

Sure, if this PR doesn't solve the problem completely, we can merge it as this with a TODO.

Feiyang1 (approved):

PR #3904, opened 2020-10-07
Implement useEmulator for Database

samtstern:
### Discussion

In order for `useEmulator()` to work the `Repo` has to be changed or replaced.  In Android this was easier because the database instance lazily creates the `Repo` (via `ensureRepo()`) but in JS the `Repo` creates the database instance.

In this change I move the `repo_` and `root_` fields to have getters/setters so that we can track their usage and allow `useEmulator()` to replace the `Repo` only if no database operations have been performed yet.

Android implementation:

This follows Proposal 2 at:

### Testing

  * Added new unit tests

### API Changes

  * Approved, see above

### DIFF
--- a/packages/database/src/api/Database.ts
+++ b/packages/database/src/api/Database.ts
@@ -33,13 +33,11 @@ import { FirebaseDatabase } from '@firebase/database-types';
  * @implements {FirebaseService}
  */
 export class Database implements FirebaseService {
-  INTERNAL: DatabaseInternals;
-
   /** Track if the instance has been used (root or repo accessed) */
   private instanceStarted_: boolean = false;
 
   /** Backing state for root_ */
-  private rootInternal_: Reference;
+  private rootInternal_?: Reference;
 
   static readonly ServerValue = {
     TIMESTAMP: {
@@ -64,18 +62,20 @@ export class Database implements FirebaseService {
         "Don't call new Database() directly - please use firebase.database()."
       );
     }
-
-    this.repo_ = repoInternal_;
-    this.INTERNAL = new DatabaseInternals(this);
   }
+
+  INTERNAL = {
+    delete: async () => {
+      this.checkDeleted_('delete');
+      RepoManager.getInstance().deleteRepo(this.repo_);
+      this.repoInternal_ = null;
+      this.rootInternal_ = null;
+    }
+  };
 
   private get repo_(): Repo {
     this.instanceStarted_ = true;
     return this.repoInternal_;
-  }
-
-  private set repo_(repo: Repo) {
-    this.repoInternal_ = repo;
   }
 
   get root_(): Reference {
@@ -84,10 +84,6 @@ export class Database implements FirebaseService {
     }
 
     return this.rootInternal_;
-  }
-
-  set root_(root: Reference) {
-    this.rootInternal_ = root;
   }
 
   get app(): FirebaseApp {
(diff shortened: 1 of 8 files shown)


schmidt-sebastian (commented):
Do you have thoughts on this: https://gist.github.com/schmidt-sebastian/b84902bf402a887ee2b3bc6591a51068

This brings this repo a little close in line with Android (or does it - but it seems simpler).

samtstern:
Quoting @schmidt-sebastian from chat:

> Reserving seems like a good idea, but I  am not sure how complex this would be - but I am sure that it will be more complex than your initial commit
> Let's go back to that with some minor tweaks:
> s/instanceUsed/started
> and can you use the same applyEmulatorSettings naming and logic as you did on Android?
> so no clone/delete but rather a mutable Repo class
>

packages/database/src/core/RepoManager.ts
```diff
@@ -87,6 +87,25 @@ export class RepoManager {
     }
   }
 
+  /**
+   * Update an existing repo in place to point to a new host/port.
+   */
+  applyEmulatorSettings(repo: Repo, host: string, port: number): void {
+    repo.repoInfo_ = new RepoInfo(
+      `${host}:${port}`,
```
schmidt-sebastian, line 95: FWIW, I think we need to pick up on our discussion on Friday and figure out whether the "public facing" database URLs (refFromUrl/toString) should use the emulator URLs (as they will with this PR) or if they should expose the production database URLs, which will allow the SDK to work with existing data that hardcodes these URLs.

samtstern: Just pushed a change to address `refFromUrl()`  it's a bit strange at first but I think it's correct:

- `refFromUrl()` must always take an argument that looks like a production URL, not an emulated URL
- If `refFromUrl()` is called after `useEmulator` you get an emulated ref, we just use the path
- If `ref.toString()` is called on a ref that points to the emulator it shows the emulated host, even if it was created with `refFromUrl`

Unless your app relies on `toString()`, this makes emulation a drop-in change.

packages/database/src/api/Database.ts
```diff
@@ -51,25 +54,66 @@ export class Database implements FirebaseService {
...
-    /** @type {Reference} */
-    this.root_ = new Reference(repo_, Path.Empty);
+  INTERNAL = {
+    delete: async () => {
+      this.checkDeleted_('delete');
+      RepoManager.getInstance().deleteRepo(this.repo_);
+      this.repoInternal_ = null;
+      this.rootInternal_ = null;
+    }
+  };
```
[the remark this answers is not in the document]
samtstern: Lol this is your code.

packages/database/src/core/RepoManager.ts
schmidt-sebastian: After talking to @yuchenshi, let's update this behavior a bit:

- `toString()` returns `localhost` for the Emulator (same as today)
- `refFromUrl`  only validates if hosts ends in `.firebaseio.com`.

I think that means going back to commits and adding a single if-check. Then we merge :)

samtstern: @schmidt-sebastian ok that's done in my latest commit, PTAL

schmidt-sebastian (approved):

schmidt-sebastian:
@Feiyang1 - Looks like we need more approvals.

Feiyang1 (approved):

PR #3906, opened 2020-10-08
Implement useEmulator for Functions

samtstern:
### Discussion

Android implementation:

This follows Proposal 2 at:

### Testing

  * Added new unit tests

### API Changes

  * Approved, see above

### DIFF
--- a/packages-exp/functions-exp/src/api.ts
+++ b/packages-exp/functions-exp/src/api.ts
@@ -29,6 +29,7 @@ import {
   FunctionsService,
   DEFAULT_REGION,
   useFunctionsEmulator as _useFunctionsEmulator,
+  useEmulator as _useEmulator,
   httpsCallable as _httpsCallable
 } from './service';
 
@@ -53,6 +54,23 @@ export function getFunctions(
     identifier: regionOrCustomDomain
   });
   return functionsInstance;
+}
+
+/**
+ * Modify this instance to communicate with the Cloudd Functions emulator.
+ *
+ * <p> Note: this must be called before this instance has been used to do any operations.
+ *
+ * @param host the emulator host (ex: localhost)
+ * @param port the emulator port (ex: 5001)
+ * @public
+ */
+export function useEmulator(
+  functionsInstance: Functions,
+  host: string,
+  port: number
+): void {
+  _useEmulator(functionsInstance as FunctionsService, host, port);
 }
 
 /**
--- a/packages-exp/functions-exp/src/api.ts
+++ b/packages-exp/functions-exp/src/api.ts
@@ -29,7 +29,6 @@ import {
   FunctionsService,
   DEFAULT_REGION,
   useFunctionsEmulator as _useFunctionsEmulator,
-  useEmulator as _useEmulator,
   httpsCallable as _httpsCallable
 } from './service';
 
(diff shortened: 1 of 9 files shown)


packages-exp/functions-exp/src/api.ts
```diff
@@ -55,6 +56,23 @@ export function getFunctions(
   return functionsInstance;
 }
 
+/**
+ * Modify this instance to communicate with the Cloudd Functions emulator.
```
hsubox76, line 60: typo on "Cloudd", also in `functions-types/index.d.ts` and `firebase/index.d.ts` and both `service.ts` files.

packages-exp/functions-exp/src/api.ts
```diff
@@ -55,6 +56,23 @@ export function getFunctions(
...
+ *
+ * <p> Note: this must be called before this instance has been used to do any operations.
+ *
+ * @param host the emulator host (ex: localhost)
```
hsubox76, line 64: Capitalize first letter of param descriptions ("The"), sorry for being pedantic, but it will go into docs. Not sure if devsite will automatically capitalize but exp docs don't go into devsite yet: https://github.com/firebase/firebase-js-sdk/blob/master/docs-exp/functions.getfunctions.md

packages-exp/functions-exp/src/api.ts
```diff
@@ -55,6 +56,23 @@ export function getFunctions(
   return functionsInstance;
 }
 
+/**
+ * Modify this instance to communicate with the Cloudd Functions emulator.
+ *
+ * <p> Note: this must be called before this instance has been used to do any operations.
```
hsubox76, line 62: Does the <p> tag do something on devsite?

packages-exp/functions-exp/src/service.ts
```diff
@@ -132,10 +132,28 @@ export class FunctionsService implements _FirebaseService {
...
+  port: number
+): void {
+  functionsInstance.emulatorOrigin = `http://${host}:${port}`;
+}
+
 /**
  * Changes this instance to point to a Cloud Functions emulator running
  * locally. See https://firebase.google.com/docs/functions/local-emulator
  *
+ * @deprecated prefer the useEmulator(host, port) method.
```
hsubox76, line 156: Capitalize "Prefer", same in other instances.

packages-exp/functions-exp/src/api.ts
```diff
@@ -55,6 +56,23 @@ export function getFunctions(
...
+ *
+ * @param host the emulator host (ex: localhost)
+ * @param port the emulator port (ex: 5001)
+ * @public
+ */
+export function useEmulator(
```
hsubox76, line 68: Is it intended for the function to have the same name, `useEmulator`, in exp for Functions, Firestore, and Database?  Sorry if I missed it in the proposal, I don't think I saw anything on it. This would require the user to rename it on import if they're planning to use more than one, correct? e.g.
```
import { useEmulator as useFunctionsEmulator } from 'firebase/functions'
import { useEmulator as useDatabaseEmulator } from 'firebase/database'
```

samtstern: Ah yeah the proposal didn't cover `exp` since I didn't know about that at the time.  I see the issue.  I could either:

a) Omit `exp` for now
b) Go with `useFunctionsEmulator` as I think that's reasonable.

packages-exp/functions-exp/src/api.ts
samtstern: Nope, just JavaDoc habits.

packages-exp/functions-exp/src/api.ts
samtstern: And actually now that I think about there's no reason to carry forward a `@deprecated` method to the `exp` API at all, right?

packages-exp/functions-exp/src/api.ts
hsubox76: I don't think this change got in?

packages-exp/functions-exp/src/api.ts
hsubox76: Oh right, makes sense, much simpler.

packages-exp/functions-exp/src/api.ts
samtstern: Ohh sorry yeah I did miss this one.

samtstern: Now it's in.

hsubox76 (approved):
LGTM, I think the CI is having some Chrome install problems right now, hopefully temporary, but good to go as soon as that is sorted out.

PR #3909, opened 2020-10-08
Implement useEmulator for Firestore

samtstern:
### Discussion

Note: I did not feel comfortable adding this to `firestore-exp` because I couldn't quite understand the internals and it seems like we'll need to amend the API proposal to add something like `EmulatorSettings` since it looks like there is no longer a global settings object.  Hoping @schmidt-sebastian can do that in a follow-up PR?

Android implementation:

This follows Proposal 2 at:

### Testing

  * Added new unit tests

### API Changes

  * Approved, see above


packages/firestore/exp/test/shim.ts
```diff
@@ -99,6 +99,10 @@ export class FirebaseFirestore
     initializeFirestore(this.app._delegate, settings);
   }
 
+  useEmulator(host: string, port: number) {
+    this.settings({ host: `${host}:${port}`, ssl: false, merge: true });
```
samtstern, line 103: This feels a little sketch but I think it's fine for the shim?

schmidt-sebastian: Works for me. I do think we can add it to firestore-exp with not a whole lot of trouble, but we can do this a  follow up for sure:

```
function useEmulator(db: Firestore, host: string, port: number): 
```

The only tricky part is that this will be the only API that modifies the settings after startup.

packages/firestore/src/api/database.ts
```diff
@@ -413,6 +419,32 @@ export class Firestore implements PublicFirestore, FirebaseService {
...
+    validateArgType('Firestore.useEmulator', 'string', 1, host);
+    validateArgType('Firestore.useEmulator', 'number', 2, port);
+
+    if (this._firestoreClient) {
+      throw new FirestoreError(
+        Code.FAILED_PRECONDITION,
+        'Firestore has already been started and its settings can no longer be changed. ' +
+          'You can only call useEmulator() before calling any other methods on a Firestore object.'
+      );
+    }
```
schmidt-sebastian, line 433: Do we need to have this check both here and in `settings`? Seems like we could get away with one check if we make the error message a little bit more vague (or just dropping the second sentence).

schmidt-sebastian (approved):
Basically LGTM. I think we should merge the two error checks that verify that the settings cannot be changed after instantiation.

packages/firestore/exp/test/shim.ts
samtstern: Ok sgtm.  I think it will have to be called `useFirestoreEmulator` so that we don't conflict between vNext packages.

Should I file a bug to track this TODO?

packages/firestore/src/api/database.ts
samtstern: No I was just trying to have distinct error messages, I'll merge it.

packages/firestore/exp/test/shim.ts
schmidt-sebastian: Yes a bug would help (and `useFirestoreEmulator` make sense)

samtstern: http://b/170936261

Feiyang1 (approved):

PR #3911, opened 2020-10-09
Add extra asserts around action code settings (auth-next)

avolkovi:


packages-exp/auth-exp/src/core/strategies/action_code_settings.ts
```diff
@@ -18,20 +18,41 @@
 import { ActionCodeSettings } from '@firebase/auth-types-exp';
 
 import { GetOobCodeRequest } from '../../api/authentication/email_and_password';
+import { AuthErrorCode } from '../errors';
+import { assert } from '../util/assert';
+import { AuthCore } from '../../model/auth'; 
 
-export function setActionCodeSettingsOnRequest(
+export function setActionCodeSettingsOnRequest_(
```
samhorlbeck, line 25: Internal methods should be prefixed with `_`, not suffixed

packages-exp/auth-exp/src/core/strategies/action_code_settings.ts
```diff
@@ -18,20 +18,41 @@
...
+  auth: AuthCore,
   request: GetOobCodeRequest,
   actionCodeSettings: ActionCodeSettings
 ): void {
+  assert(typeof actionCodeSettings.url !== 'undefined', AuthErrorCode.MISSING_CONTINUE_URI, {
```
samhorlbeck, line 30: Shouldn't this just be `assert(actionCodeSettings.url)`?

Aren't actonCodeSettings.url = null and actionCodeSettings.url = '' both incorrect as well?

packages-exp/auth-exp/src/core/strategies/action_code_settings.ts
avolkovi: mybad, I've been in the closure SDK too long

packages-exp/auth-exp/src/core/strategies/action_code_settings.ts
avolkovi: we type url to be `string` instead of `string | undefined`, so technically this case should not be necessary, in the old SDK there were two separate errors for undefined (MISSING) and empty (INVALID), for the other fields I dropped the extra undefined checks by having our type signature stricter, not sure what to do here... we do have the dedicated error string, but this should also never happen due to typing....

samhorlbeck: I see. Up to you-- the rationale for not doing runtime type checking is code size, but `typeof actionCodeSettings.url !== 'undefined'` is actually heavier on the code size because `typeof` and `'undefined'` can't be minified.

`t(a.c, 'MISSING_CONTINUE_URL', {...})`  
vs  
`t(typeof a.c !== 'undefined', 'MISSING_CONTINUE_URL', {...})`

avolkovi: let's drop this check, it's not consistent with the rest of our new API and avoidance of run time checks for things that should be covered by types

samhorlbeck (approved):

Feiyang1 (approved):
Please fix the lint error

PR #3915, opened 2020-10-09
Fix custom claims

samtstern:

### Discussion

See https://github.com/firebase/firebase-js-sdk/pull/3876#issuecomment-706241099

### Testing

  * New unit tests

### API Changes

  * Reverting a bug


yuchenshi (approved):

samtstern:
cc @mdirolf

yuchenshi:
Well actually... Shouldn't we use intersection like `MyInterface & `{[claim: string]: any}` instead? Let's see which one works better

PR #3916, opened 2020-10-09
Make compat script work

Feiyang1:


samhorlbeck (approved):
Does this mean that the I will need to update the main auth register function to also import as modularAPIs?
https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/auth-exp/src/core/auth/register.ts#L54

PR #3917, opened 2020-10-09
Handle app deletion a little more gracefully in auth-exp and auth-compat-exp

samhorlbeck:
b/166287154


packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -193,7 +212,7 @@ export class AuthImpl implements Auth, _FirebaseService {
   }
 
   async _delete(): Promise<void> {
-    // TODO: Determine what we want to do in this case
+    this._deleted = true;
```
avolkovi, line 215: do we want/need tests?

samhorlbeck: Done

avolkovi (approved):

PR #3919, opened 2020-10-09
Fix out-of-order execution for enqueueRetryable

schmidt-sebastian:
We currently do follow the global order of AsyncQueue operations when we insert items into the retryable queue. This seems to allow operations that inserted into the retryable queue to skip ahead of items that are inserted into the regular queue. This PR ties the insertion of new items to the retryable queue to the order in which all items are enqueue.

I still am not able to reproduce the original failure in a unit test (tried various delays, various order of operations, scheduling operations while other operations are running), but this code change addresses the consistent Integration tests failures in a follow-up PR.

### DIFF
--- a/.changeset/long-hornets-scream.md
+++ b/.changeset/long-hornets-scream.md
@@ -1,0 +1,5 @@
+---
+"@firebase/firestore": patch
+---
+
+Fixed a potential issue in our internal queue that could have allowed API calls to be executed out of order.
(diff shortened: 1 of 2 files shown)


.changeset/long-hornets-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/firestore": patch
```
wilhuff, line 2: Do we also need a patch for the firebase-wide package?

wilhuff (approved):
LGTM

.changeset/long-hornets-scream.md
schmidt-sebastian: I believe that we don't need to for a patch release, since every release is at least a minimum patch version bump.

PR #3924, opened 2020-10-12
`lerna` field is deprecated since v3

n0mer:
Remove deprecated field. Anyway, currenly project uses learna v3.22.1, not 3.10.x


hsubox76 (approved):
Thanks!

PR #3925, opened 2020-10-12
Make the emulator API match for auth-exp

samhorlbeck:
b/170304240

### DIFF
--- a/packages-exp/auth-exp/src/api/index.test.ts
+++ b/packages-exp/auth-exp/src/api/index.test.ts
@@ -358,7 +358,7 @@ describe('api/_performApiRequest', () => {
 
     it('works properly with an emulated environment', () => {
       (auth.config as ConfigInternal).emulator = {
-        host: 'http://localhost:5000'
+        url: 'http://localhost:5000'
       };
       expect(_getFinalTarget(auth, 'host', '/path', 'query=test')).to.eq(
         'http://localhost:5000/host/path?query=test'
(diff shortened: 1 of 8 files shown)


packages-exp/auth-exp/src/api/index.test.ts
```diff
@@ -358,8 +358,7 @@ describe('api/_performApiRequest', () => {
 
     it('works properly with an emulated environment', () => {
       (auth.config as ConfigInternal).emulator = {
-        hostname: 'localhost',
-        port: 5000
+        host: 'http://localhost:5000'
```
avolkovi, line 361: this should probably be `url` to match JS SDK

packages-exp/auth-exp/src/core/util/emulator.ts
```diff
@@ -20,13 +20,12 @@ import { debugAssert } from './assert';
 
 export function _emulatorUrl(config: ConfigInternal, path?: string): string {
   debugAssert(config.emulator, 'Emulator should always be set here');
-  const { hostname, port } = config.emulator;
+  const { host: emulatorHost } = config.emulator;
```
avolkovi, line 23: nit: we should probably do some sort of URI.parse() here to standardize the URL

samhorlbeck: Done

avolkovi (approved):

avolkovi (approved):

PR #3927, opened 2020-10-12
forbid prereleasing to npm tag @latest

Feiyang1:
prevent accidentally releasing prereleases to firebase@latest

### DIFF
--- a/scripts/release/prerelease.ts
+++ b/scripts/release/prerelease.ts
@@ -103,7 +103,7 @@ export async function runPrerelease({
   }
 }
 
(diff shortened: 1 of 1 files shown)


scripts/release/prerelease.ts
```diff
@@ -97,6 +103,8 @@ export async function runPrerelease({
   }
 }
 
+const FORBIDDEN_TAGS = ['latest'];
```
hsubox76, line 106: Should we add "next", "exp", and/or "canary"? They're not as important but, I mean, since the code is all set up, why not?

hsubox76 (approved):

PR #3932, opened 2020-10-13
Point browser field to esm build

Feiyang1:
Rationale

- Make Firebase work nicely with Angular10.
   Angular10 complains loudly if the browser field points to cjs build. And it's hard for developer to change build configuration which is encapsulated by ngcc
   See https://github.com/firebase/firebase-js-sdk/issues/3315

- Address the incompatibility issue between our typings and esm builds
  Our typings are modeled after the cjs module format, using `export = firebase`, so we ask developers to use namespace import to import firebase, like so: `import * as firebase from ‘firebase/app’`, but it’s not compatible with our esm builds which only have a default export. 
  
  If you manually configure the bundler to use our esm build (e.g. using the `module` field or the` esm2017` field), you will get an error. For example, you will get the following error with webpack:
`"export 'initializeApp' (imported as 'firebase') was not found in 'firebase/app'`

  This change allows developers to use any of our builds without resorting to hacks.


packages/rxfire/package.json
```diff
@@ -33,7 +33,7 @@
     "test:browser:debug": "karma start --browsers=Chrome --auto-watch"
   },
   "main": "dist/index.node.cjs.js",
-  "browser": "dist/index.cjs.js",
+  "browser": "dist/index.esm.js",
```
hsubox76, line 36: Does rxfire have a top-level dist/ dir? I couldn't find one after a build. It does have a dist/ and a package.json inside each component subdir.

hsubox76:
I think you have to change this line:
https://github.com/firebase/firebase-js-sdk/blob/eeb1dfa4f629dc5cf328e4b4a224369c0670c312/packages/remote-config/rollup.config.js#L44
to point to `pkg.main` or else you make a weird corrupted esm file which is causing the CI build to fail.

packages/component/rollup.config.js
```diff
@@ -38,10 +38,7 @@ const es5Builds = [
    */
   {
     input: 'index.ts',
-    output: [
-      { file: pkg.browser, format: 'cjs', sourcemap: true },
```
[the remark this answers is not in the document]
Feiyang1: Good catch. reverted.

packages/rxfire/package.json
Feiyang1: hmm, we don't have a top level one. We should remove them all.

Feiyang1:
> I think you have to change this line:
> https://github.com/firebase/firebase-js-sdk/blob/eeb1dfa4f629dc5cf328e4b4a224369c0670c312/packages/remote-config/rollup.config.js#L44
> 
> 
> to point to `pkg.main` or else you make a weird corrupted esm file which is causing the CI build to fail.
> There could possibly be a similar issue in some of the other rollup config files.

Good catch. Updated

hsubox76 (approved):
Approved pending CI tests.

packages/rules-unit-testing/src/api/index.ts
```diff
@@ -15,15 +15,16 @@
...
 import { _FirebaseApp } from '@firebase/app-types/private';
 import { FirebaseAuthInternal } from '@firebase/auth-interop-types';
 import * as request from 'request';
 import { base64 } from '@firebase/util';
 import { setLogLevel, LogLevel } from '@firebase/logger';
 import { Component, ComponentType } from '@firebase/component';
 
-export { database, firestore } from 'firebase';
+const { firestore, database } = firebase;
+export { firestore, database };
```
Feiyang1, line 27: @samtstern @yuchenshi 
Meant to ask you before merging the PR - How does developer use these two exports? Do they use the typings from them?
They won't get typings after this change because `firestore` and `database` are not namespaces anymore.

samtstern: I believe they're used to provide access to Firestore / RTDB classes like this:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/rules-unit-testing/test/database.test.ts#L313

If those tests still pass I think we're fine?  To be honest I don't quite understand namespaces.

Feiyang1: Yeah, that will continue to work. We are good. Thanks for confirming.

PR #3933, opened 2020-10-13
Add documentation for ActionCodeURL

avolkovi:


docs-exp/auth.actioncodeurl.operation.md
```diff
@@ -4,6 +4,8 @@
 
 ## ActionCodeURL.operation property
 
+The action performed by the email action link. It returns from one of the types from
```
avolkovi, line 7: looks like the @link got truncated

docs-exp/auth-types.actioncodeurl.operation.md
```diff
@@ -4,6 +4,8 @@
 
 ## ActionCodeURL.operation property
 
+The action performed by the email action link. It returns from one of the types from [ActionCodeInfo](./auth-types.actioncodeinfo.md)
```
samhorlbeck, line 7: does {@link ActionCodeInfo} work?

docs-exp/auth.actioncodeurl.operation.md
samhorlbeck: ah that answers it

docs-exp/auth-types.actioncodeurl.operation.md
avolkovi: looks like it works in the auth-types docs but not the auth docs, probably will self resolve once I add documentation for ActionCodeInfo in both

packages-exp/auth-exp/src/core/action_code_url.ts
```diff
@@ -45,11 +49,20 @@ const MODE_TO_OPERATION_MAP: { [key: string]: externs.Operation } = {
 
 /**
  * Maps the mode string in action code URL to Action Code Info operation.
+ *
+ * @param mode
+ * @internal
```
samhorlbeck, line 54: You don't need to do this right? Since these are not exported?

samhorlbeck (approved):

packages-exp/auth-exp/src/core/action_code_url.ts
avolkovi: I think you're right, doing it for consistency here anyway, no promises I'll do it everywhere

Feiyang1 (approved):

PR #3935, opened 2020-10-13
Simplify ComponentProviders

schmidt-sebastian:
This PR is meant to reduce the diff for https://github.com/firebase/firebase-js-sdk/pull/3901

#3901 changes the legacy Firestore SDK to use `enableIndexedDbPersistence` or `enableMultiTabIndexedDbPersistence` for any call to `Firestore.enablePersistence()`. This allows us to only include the IndexedDbOfflineComponentProvider or MultiTabOfflineComponentProvider when needed, which results in some simplifications.

We now know that  `MultiTabOfflineComponentProvider` is only used with multi-tab persistence. Before, `MultiTabOfflineComponentProvider` also needed to support non-multi-tab persistence. This already simplified the component provider, but this PR goes a bit further and removes all persistence related settings from the Component Provider configuration and instead configures each persistence provider directly at instantiation. 

This PR also adds the existing `canFallback` logic for persistence-related errors to firestore-exp.


packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -105,15 +105,11 @@ const MAX_PRIMARY_ELIGIBLE_AGE_MS = 5000;
 const CLIENT_METADATA_REFRESH_INTERVAL_MS = 4000;
 /** User-facing error when the primary lease is required but not available. */
 const PRIMARY_LEASE_EXCLUSIVE_ERROR_MSG =
-  'Failed to obtain exclusive access to the persistence layer. ' +
-  'To allow shared access, make sure to invoke ' +
-  '`enablePersistence()` with `synchronizeTabs:true` in all tabs. ' +
-  'If you are using `experimentalForceOwningTab:true`, make sure that only ' +
-  'one tab has persistence enabled at any given time.';
+  'Failed to obtain exclusive access to the persistence layer. To allow ' +
+  'shared access, multi-tab synchronization has to be enabled in all tabs.';
```
wilhuff, line 109: It seems a shame to remove the helpful error message regarding `experimentalForceOwningTab:true`, though it seems like we should know if that's enabled and give that helpful message only when applicable.

wilhuff (approved):
LGTM

packages/firestore/src/local/indexeddb_persistence.ts
schmidt-sebastian: I was hoping to not use the "experimental" name in firestore-exp and dropped it due to the two different names. I will come up with a way to add it back.

Feiyang1 (approved):
Approve for prettier changes

PR #3939, opened 2020-10-13
Remove JS Input Validation

schmidt-sebastian:
This PR removes all input validation that should be caught by any JavaScript environments that supports strict type checking, such as TypeScript or Google Closure. It is merely meant to reduce binary size.

Validations that remain:
- Empty strings are still rejected as collection ID/document IDs (note that Android & iOS don't seem to do this
- Limits have to be positive
- All query constraints and restrictions that are not expressed via the API
- The `firestore-exp` API still calls `cast` to verify that the passed in objects are the right Firestore objects. This should catch almost all cases where a user forgets to change the arguments as they transition to the functional API.


schmidt-sebastian:
I sat on this for too long since I was hoping this would shave off even more bytes, but I was just informed that code freeze for Fireconf is tomorrow.

packages/firestore/src/api/database.ts
```diff
@@ -275,21 +230,9 @@ class FirestoreSettings {
...
       settings.experimentalForceLongPolling ?? DEFAULT_FORCE_LONG_POLLING;
 
-    validateNamedOptionalType(
-      'settings',
-      'boolean',
-      'experimentalAutoDetectLongPolling',
-      settings.experimentalAutoDetectLongPolling
-    );
     this.experimentalAutoDetectLongPolling =
       settings.experimentalAutoDetectLongPolling ??
```
wilhuff, line 237: So given that we're not validating that `settings.experimentalAutoDetectLongPolling` is actually a boolean value, should we be accessing it with a forced coercion (i.e. as `!!settings.experimentalAutoDetectLongPolling`)?

Or is the idea that we've declared the type in typescript and all bets are off for users who aren't using our d.ts files?

packages/firestore/src/api/database.ts
```diff
@@ -1478,10 +1346,8 @@ export class DocumentSnapshot<T = DocumentData>
 
   get(
     fieldPath: string | ExternalFieldPath,
-    options?: PublicSnapshotOptions
+    options: PublicSnapshotOptions = {}
```
wilhuff, line 1349: Does this require an API change to the d.ts file, or is this rendered by tsc as a local assignment if not defined? Stated alternatively: does the caller supply the default argument or does the callee?

packages/firestore/src/api/database.ts
```diff
@@ -2487,7 +2274,7 @@ export class CollectionReference<T = DocumentData>
   }
 }
 
-function validateSetOptions(
+export function validateSetOptions(
```
wilhuff, line 2277: Doesn't this potentially create a cycle? Better to move this into `input_validation.ts`.

wilhuff (commented):
Basically LGTM, though with some questions.

packages/firestore/src/api/database.ts
schmidt-sebastian: I originally decided against this approach, because it makes it impossible to turn a feature off it its default value is `true`. Since we have only ever been using opt-in features, this has not been a problem yet. I updated this code, which ensures that we have a boolean value once we exit this function (it also removes some code).

packages/firestore/src/api/database.ts
schmidt-sebastian: The default argument is provided by the callee. An example in ES3:

```
function foo(bar = {}) {
}
```

transpiles to:

```
"use strict";
function foo(bar) {
    if (bar === void 0) { bar = {}; }
}
 ```

ES5+JavaScript supports default arguments, but the semantics match the code snippet above.

packages/firestore/src/api/database.ts
schmidt-sebastian: Done

wilhuff (approved):
LGTM

PR #3942, opened 2020-10-14
Add assertion for emulator url scheme in auth

samhorlbeck:


avolkovi (approved):

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
```diff
@@ -489,6 +489,17 @@ describe('core/auth/auth_impl useEmulator', () => {
       expect(normalEndpoint.calls.length).to.eq(0);
       expect(emulatorEndpoint.calls.length).to.eq(1);
     });
+
+    it('checks the scheme properly', () => {
+      expect(() => auth.useEmulator('http://localhost:2020')).not.to.throw;
+      delete auth.config.emulator;
+      expect(() => auth.useEmulator('https://localhost:2020')).not.to.throw;
+      delete auth.config.emulator;
+      expect(() => auth.useEmulator('ssh://localhost:2020')).to.throw(
```
yuchenshi, line 498: Let's add a test case for `localhost:5000` (with no `://`), a somewhat common misuse

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -203,6 +203,10 @@ export class AuthImpl implements Auth, _FirebaseService {
       appName: this.name
     });
 
+    assert(/^https?:\/\//.test(url), AuthErrorCode.INVALID_EMULATOR_SCHEME, {
```
yuchenshi, line 206: FWIW This just throws `ARGUMENT_ERROR` in the JS SDK. I don't think we need a dedicated error code for it.

yuchenshi (changes requested):

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: We don't do message overrides to errors in the new SDK, so in order to have that text it needs to be its own error code (otherwise it's just going to be a plain argument error with no explanation)

packages-exp/auth-exp/src/core/auth/auth_impl.test.ts
samhorlbeck: Done

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: Resolving this

yuchenshi (approved):

PR #3944, opened 2020-10-14
Remove undocumented getLogLevel

schmidt-sebastian:
This removes the undocumented Firestore.logLevel API and makes `setLogLevel` a tree-shakeable function (even in the old SDK).


wu-hui (approved):
LGTM, let's see if I have power now.

PR #3945, opened 2020-10-14
Fix storage-types

hsubox76:
Separated out from storage modularization PR: https://github.com/firebase/firebase-js-sdk/pull/3499

Should fix https://github.com/firebase/firebase-js-sdk/issues/1515

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7620,12 +7620,7 @@ declare namespace firebase.storage {
   /**
    * An error returned by the Firebase Storage SDK.
    */
-  interface FirebaseStorageError {
-    name: string;
-    code: string;
-    message: string;
-    serverResponse: null | string;
-  }
+  interface FirebaseStorageError extends FirebaseError {}
 
   interface StorageObserver<T> {
     next?: NextFn<T> | null;
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7618,6 +7618,22 @@ declare namespace firebase.storage {
   }
 
   /**
+   * An error returned by the Firebase Storage SDK.
+   */
+  interface FirebaseStorageError {
+    name: string;
+    code: string;
+    message: string;
+    serverResponse: null | string;
+  }
+
+  interface StorageObserver<T> {
+    next?: NextFn<T> | null;
+    error?: (error: FirebaseStorageError) => void | null;
+    complete?: CompleteFn | null;
+  }
+
+  /**
    * Represents the process of uploading an object. Allows you to monitor and
    * manage the upload.
    */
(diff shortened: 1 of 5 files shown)


packages/storage-types/index.d.ts
```diff
@@ -85,24 +84,41 @@ export interface UploadMetadata extends SettableMetadata {
...
+  catch(onRejected: (a: FirebaseStorageError) => any): Promise<any>;
   on(
     event: TaskEvent,
     nextOrObserver?:
-      | Partial<Observer<UploadTaskSnapshot>>
+      | Partial<StorageObserver<UploadTaskSnapshot>>
       | null
-      | ((a: UploadTaskSnapshot) => any),
-    error?: ((a: Error) => any) | null,
+      | ((a: UploadTaskSnapshot) => unknown),
```
schmidt-sebastian, line 112: This `unknown` seems inconsistent with the other return types.

Since this is going to land in a breaking change release, do you think it makes sense to start using more descriptive names than "a"? (here and everywhere)

schmidt-sebastian (commented):
Thanks for doing this!

We also need to update packages/firebase/index.d.ts

packages/storage-types/index.d.ts
hsubox76: Sure, seems like a good idea.

schmidt-sebastian (approved):

packages/firebase/index.d.ts
```diff
@@ -7617,6 +7617,22 @@ declare namespace firebase.storage {
...
   }
 
+  /**
+   * An error returned by the Firebase Storage SDK.
+   */
+  interface FirebaseStorageError {
+    name: string;
+    code: string;
+    message: string;
+    serverResponse: null | string;
```
Feiyang1, line 7627: Who adds `serverResponse` to the error object?
In https://github.com/firebase/firebase-js-sdk/pull/3946, I'm moving template data under a new field `customData` to avoid overwriting existing fields, should we do the same for `serverResponse`?

hsubox76:
Depends on https://github.com/firebase/firebase-js-sdk/pull/3946, rebase after that is merged.

packages/firebase/index.d.ts
```diff
@@ -7617,6 +7617,19 @@ declare namespace firebase.storage {
     md5Hash?: string | null;
   }
 
+  /**
+   * An error returned by the Firebase Storage SDK.
+   */
+  interface FirebaseStorageError extends FirebaseError {
+    customData: { serverResponse: string | null };
```
Feiyang1, line 7624: Since you have a getter `serverResponse`, `customData` feels like the internal implementation and you can just expose the getter.

hsubox76: Done.

Feiyang1 (approved):

Issue #1891, opened 2019-06-18
Getting error after hot module reloading: Overwriting FirebaseError base field "name" can cause unexpected behavior.


MillouwMills:
### Environment

  * Operating System version: macOS 10.13.6
  * Browser version: 75.0.3770.90 (Official Build) (64-bit)
  * Firebase SDK version: 6.2.0
  * Firebase Product: app

### Problem
Get a warning after hot reloading my app w/  `react-hot-loader` and `hot(App)`
Extract of console output:

- (index.cjs.js:652) Overwriting FirebaseError base field "name" can cause unexpected behavior.
-  ./node_modules/@firebase/util/dist/index.cjs.js.ErrorFactory.create @ index.cjs.js:652
- initializeApp @ index.cjs.js:327
- ./node_modules/@firebase/app/dist/index.cjs.js.firebaseNamespace.initializeApp @ index.cjs.js:495
- initFirebaseApp @ index.js:6
- App @ App.js:16
- safeReactConstructor...

#### Relevant Code:

`./src/firebase/index.js` (loader)
```javascript
import { initializeApp } from "firebase/app";
import firebaseConfig from "./config";

// load and apply firebase config
export default function initFirebaseApp(appName) {
  let app =  initializeApp(firebaseConfig, appName);
}

export { firebaseConfig };
```

`./src/App.js` (extract) (loader caller)`
```javascript
import { hot } from "react-hot-loader/root";
import React from "react";
import debug from "debug";
import ...

import AppContext from "./AppContext";

class App extends React.Component {
  constructor(props) {
    super(props);

    this.firebaseApp = initFirebaseApp(firebaseConfig.projectId);

    this.logger = debug(firebaseConfig.projectId);
    this.logger.log = console.log.bind(console);

    this.state = {
      ...
    };
  }

  componentWillUnmount() {
    // TODO: look up app.database.ref(...).off()

    firebase
      .delete(this.firebaseApp)
      .then(() => {
        console.log("App deleted successfully");
      })
      .catch(error => {
        console.log("Error deleting firebase app:", error);
      });
  }

  render() {
    return (
      <div>
        <AppContext.Provider
          value={...}
        >
          <Header />
          <main>
            ...
          </main>
        </AppContext.Provider>
      </div>
    );
  }
}

export default hot(App);
```

mmermerkaya:
[...]

PR #3946, opened 2020-10-14
Move template data under a new `customData` field

Feiyang1:
Write template data to a new `customData` field in `FirebaseError` instead of writing to the error object itself to avoid overwriting existing fields.

Solves https://github.com/firebase/firebase-js-sdk/issues/1891#issuecomment-503189696


samhorlbeck (approved):
Looks good for auth

packages-exp/auth-exp/src/core/providers/facebook.test.ts
```diff
@@ -59,7 +59,7 @@ describe('src/core/providers/facebook', () => {
     const error = AUTH_ERROR_FACTORY.create(AuthErrorCode.NEED_CONFIRMATION, {
       appName: 'foo'
     });
-    (error as TaggedWithTokenResponse)._tokenResponse = {
+    (error.customData! as TaggedWithTokenResponse)._tokenResponse = {
```
hsubox76, line 62: Always having to use the ! seems weird. In storage, all errors are FirebaseStorageError which extends FirebaseError and makes customData non-optional. But I see auth-exp uses the shared ErrorFactory so I don't know if it would be really messy to get it to return an extended error type instead of FirebaseError.

hsubox76: Also should a property of customData be underscore-prefixed? Not sure what that would mean.

hsubox76 (approved):

packages-exp/auth-exp/src/core/providers/facebook.test.ts
Feiyang1: Valid point. What other SDKs have done is do a `is[Product]Error()` check, which narrows the type to the specific error type, then you can access `customData` safely. See [here](https://github.com/firebase/firebase-js-sdk/pull/3946/files#diff-d37d76c3ba7e573c7326dc7b8eb7ee3b7d72f6ca34b8bbc4285a64f0acfc6dcbR287-R292) as an example.

That said, I'm not too worried about them since they are all in test files. 

I think the underscore means it's for internal usage only, which is fine. @samhorlbeck

PR #3947, opened 2020-10-14
Copy the auth user on updateCurrentUser(), also assert the app name matches

samhorlbeck:
This PR creates a differentiation between the public `updateCurrentUser()` and the internal `_updateCurrentUser()`. The latter does not make a copy because it is called by the SDK itself when the SDK signs someone in or updates something.

b/170886559,
b/170854629

### DIFF
--- a/packages-exp/auth-exp/src/core/auth/auth_impl.ts
+++ b/packages-exp/auth-exp/src/core/auth/auth_impl.ts
@@ -214,10 +214,14 @@ export class AuthImpl implements Auth, _FirebaseService {
   async updateCurrentUser(userExtern: externs.User | null): Promise<void> {
     // The public updateCurrentUser method needs to make a copy of the user,
     // and also needs to verify that the app matches
-    const user = userExtern as User|null;
-    assert(!user || user.auth.name === this.name, AuthErrorCode.ARGUMENT_ERROR, {
-      appName: this.name
-    });
+    const user = userExtern as User | null;
+    assert(
+      !user || user.auth.name === this.name,
+      AuthErrorCode.ARGUMENT_ERROR,
+      {
+        appName: this.name
+      }
+    );
 
     return this._updateCurrentUser(user && user._clone());
   }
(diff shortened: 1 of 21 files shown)


packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -211,7 +211,22 @@ export class AuthImpl implements Auth, _FirebaseService {
...
+    const user = userExtern as User | null;
+    assert(
+      !user || user.auth.name === this.name,
+      AuthErrorCode.ARGUMENT_ERROR,
+      {
+        appName: this.name
+      }
+    );
+
+    return this._updateCurrentUser(user && user._clone());
```
avolkovi, line 226: user?._clone()

packages-exp/auth-exp/src/model/user.ts
```diff
@@ -34,16 +34,16 @@ export interface UserParameters {
...
   stsTokenManager: StsTokenManager;
 
-  displayName?: string;
-  email?: string;
-  phoneNumber?: string;
-  photoURL?: string;
-  isAnonymous?: boolean;
-  emailVerified?: boolean;
-  tenantId?: string;
+  displayName?: string | null;
```
avolkovi, line 37: we really should figure out this undefined | null situation at some point...

avolkovi (approved):

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: That types out to `User|undefined` but the method wants `User|null`. So it's either `user?._clone() || null` or `user && user._clone()`. The latter seems cleaner to me so I'm keeping it the way it's written

packages-exp/auth-exp/src/model/user.ts
samhorlbeck: Agreed :P

PR #3955, opened 2020-10-16
[Abandoned] A new Node script to build testing bundles for given project IDs

wu-hui:
The "script" has basically the entire SDK in it. My plan is to have bundles integration test run this script every time, giving it project ids currently available, and use the JSON output to get the bundles string. The built json is manually copied to api/ for now, but not used yet. I could not figure out a way to automatically run the built, given browser restrictions.

I originally wanted this to be in `packages/firestore` but that leads to cyclic dependency. 

I want to get this merged before I attack moving the integration test, and make bundles tree-shakeable only.


schmidt-sebastian (commented):
Hello :) I think I need some explanation for how this will be used. Can you explain how the generated bundle will be consumed?

In general, I wonder though:
- Could we just use the generated bundle JSON and do a string replace on the project ID?
- If we can't, does it make sense to commit the generated bundle file? Should it instead be re-generated before every test run?
- Could we remove the rollup config and run the script through `ts-node`?

Lastly, this repo should only have a single `yarn.lock` at the root. We need to figure out why your install created a second lock file.

wu-hui:
> Hello :) I think I need some explanation for how this will be used. Can you explain how the generated bundle will be consumed?

See https://github.com/firebase/firebase-js-sdk/pull/3962 which is intended as the next PR. It passes all tests locally, I just need to figure out why it is failing on CI. But basically the generate json is checked in.

> 
> In general, I wonder though:
> 
> * Could we just use the generated bundle JSON and do a string replace on the project ID?

We need to figure out the "length" for each element that has project ID, which means we need to know the structure, which means we need access to bundler_reader.ts at least, which will lead us to needed many more.

> * If we can't, does it make sense to commit the generated bundle file? Should it instead be re-generated before every test run?

Yes. The bundle file is checked in.

> * Could we remove the rollup config and run the script through `ts-node`?

`ts-node` won't work without having all the internal deps ready.

> 
> Lastly, this repo should only have a single `yarn.lock` at the root. We need to figure out why your install created a second lock file.

I deleted. I think it was added in my many trials to make it work.

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * The main function for this script.
+ * @param {array} args The command line arguments.
+ */
```
schmidt-sebastian, line 4: This seems misplaced.

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+  PROJECT_CONFIG.projectId
+];
+
+function main(args: string[]) {
+  // If arguments are given, assume they are project IDs used to generate bundles,
+  // otherwise generate using default project ids used by integration tests.
+  let projectIds = PROJECT_IDS;
+  if (args.length >= 3) {
+    projectIds = args.slice(2);
+  }
```
schmidt-sebastian, line 77: Please use `yargs`, which allows you to provide descriptive argument names and type-safe argument conversions. See also here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/scripts/build-bundle.ts#L27

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+import * as firestore from '@firebase/firestore-types';
+import { TestBundleBuilder } from '@firebase/firestore/test/unit/util/bundle_data';
+import { DatabaseId } from '@firebase/firestore/src/core/database_info';
+import { key } from '@firebase/firestore/test/util/helpers';
+import { collectionReference } from '@firebase/firestore/test/util/api_helpers';
+import {
+  EMULATOR_PROJECT_ID,
+  ALT_PROJECT_ID
+} from '@firebase/firestore/test/integration/util/emulator_settings';
```
schmidt-sebastian, line 13: Your import make a pretty strong case that this file should live under packages/firestore (probably `packages/firestore/scripts`. Once you do that, you should be able to use `ts-node` as well and remove the Rollup config.

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+  const b = key('coll-1/b');
+  const builder = new TestBundleBuilder(new DatabaseId(projectId));
+
+  builder.addNamedQuery(
+    'limit',
+    { seconds: 1000, nanos: 9999 },
+    (collectionReference('coll-1')
+      .orderBy('bar', 'desc')
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .limit(1) as any)._query
```
schmidt-sebastian, line 36: Should `TestBundleBuilder` take a public API query? It should then be able to cast internally to the internal query.

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Returns a bundle builder for the given projectId. The builder will build with the
+ * same documents and queries, but with the given projectId.
+ */
+function bundleWithTestDocsAndQueries(projectId: string): TestBundleBuilder {
+  const testDocs: { [key: string]: firestore.DocumentData } = {
+    a: { k: { stringValue: 'a' }, bar: { integerValue: 1 } },
+    b: { k: { stringValue: 'b' }, bar: { integerValue: 2 } }
+  };
+
+  const a = key('coll-1/a');
```
schmidt-sebastian, line 26: Maybe s/a/docA

integration/firestore/bundle_test_data/generate_test_bundle.ts
```diff
@@ -0,0 +1,88 @@
...
+    projectIds = args.slice(2);
+  }
+  const result: { [k: string]: string } = {};
+  for (const projecId of projectIds) {
+    result[projecId] = bundleWithTestDocsAndQueries(
+      projecId
+    ).build('test-bundle', { seconds: 1001, nanos: 9999 });
+  }
+  console.log(JSON.stringify(result));
+  process.exit();
```
schmidt-sebastian, line 85: Is this needed? This process should exit here automatically.

packages/firestore/test/integration/api/test_bundles.json
```diff
@@ -0,0 +1 @@
...undledQuery\":{\"parent\":\"projects/fir-testing-field/databases/(default)/documents\",\"structuredQuery\":{\"from\":[{\"collectionId\":\"coll-1\"}],\"orderBy\":[{\"field\":{\"fieldPath\":\"bar\"},\"direction\":\"DESCENDING\"},{\"field\":{\"fieldPath\":\"__name__\"},\"direction\":\"DESCENDING\"}],\"limit\":{\"value\":1}},\"limitType\":\"LAST\"}}}152{\"documentMetadata\":{\"name\":\"projects/fir-testing-field/databases/(default)/documents/coll-1/a\",\"readTime\":{\"seconds\":1000,\"nanos\":9999},\"exists\":true}}223{\"document\":{\"name\":\"projects/fir-testing-field/databases/(default)/documents/coll-1/a\",\"createTime\":{\"seconds\":1,\"nanos\":9},\"updateTime\":{\"seconds\":1,\"nanos\":9},\"fields\":{\"k\":{\"stringValue\":\"a\"},\"bar\":{\"integerValue\":1}}}}152{\"documentMetadata\":{\"name\":\"projects/fir-testing-field/databases/(default)/documents/coll-1/b\",\"readTime\":{\"seconds\":1000,\"nanos\":9999},\"exists\":true}}223{\"document\":{\"name\":\"projects/fir-testing-field/databases/(default)/documents/coll-1/b\",\"createTime\":{\"seconds\":1,\"nanos\":9},\"updateTime\":{\"seconds\":1,\"nanos\":9},\"fields\":{\"k\":{\"stringValue\":\"b\"},\"bar\":{\"integerValue\":2}}}}"}
```
schmidt-sebastian, line 1: Are the quotes meant to be escaped? Seems like they shouldn't be.

integration/firestore/bundle_test_data/package.json
```diff
@@ -0,0 +1,44 @@
+{
```
schmidt-sebastian, line 1: You should really be able to run this under ts-node or compile this once and check in the JS sources (as done in packages/firestore/scripts). Running this via ts-node will make your life much easier, as it makes debugging painless.

integration/firestore/bundle_test_data/yarn.lock
```diff
@@ -0,0 +1,1219 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
schmidt-sebastian, line 1: Please remove.

packages/firestore/scripts/run-tests.ts
```diff
@@ -57,8 +58,9 @@ let args = [
 
 if (argv.emulator) {
   process.env.FIRESTORE_EMULATOR_PORT = '8080';
-  process.env.FIRESTORE_EMULATOR_PROJECT_ID = 'test-emulator';
+  process.env.FIRESTORE_EMULATOR_PROJECT_ID = EMULATOR_PROJECT_ID;
 }
+export const ALT_PROJECT_ID = 'test-db2';
```
schmidt-sebastian, line 63: This seems slightly off. Who is importing this file?

packages/firestore/test/integration/util/emulator_settings.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Projects used in emulator integration tests.
+export const EMULATOR_PROJECT_ID = 'test-emulator';
+export const ALT_PROJECT_ID = 'test-db2';
```
schmidt-sebastian, line 20: Could this be merged with `settings.ts`?

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -26,7 +26,8 @@ import {
   withTestCollection,
   withTestDb
 } from '../util/helpers';
-import { ALT_PROJECT_ID, DEFAULT_PROJECT_ID } from '../util/settings';
+import { DEFAULT_PROJECT_ID } from '../util/settings';
+import {ALT_PROJECT_ID} from '../util/emulator_settings';
```
schmidt-sebastian, line 30: Formatting seems inconsistent here.

integration/firestore/bundle_test_data/package.json
wu-hui: Let's address this before other comments.

I tried to move it there and use tsc/node as other scripts, the issue I have is the generated js has 

`var bundle_data_1 = require("../test/unit/util/bundle_data");`

Which is expecting a bundle_data.js to exist, any idea how I can fix this? This is the first script to actually access the internals of the SDK, however, I think moving it here makes sense, and I want to make it happen now.

schmidt-sebastian: To make this work in JS, you need to bundle all JS dependencies with the final script. Unfortunately, it looks like you may need a Rollup configuration for this after all. This will allow you define "externs" and decide which externs are inlined.

You can probably use your existing rollup configuration and put it in packages/firestore/rollup.script.js and then change https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/package.json#L12 to invoke this instead.

Let me know if you need more help with this.

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: Done.

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: Done.

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: Done.

integration/firestore/bundle_test_data/package.json
wu-hui: Thanks for the help.

packages/firestore/test/integration/api/test_bundles.json
wu-hui: The script is generating bundle strings, which are then put in a string-to-string map and serialize to json, to support multiple project ids. That is why the quotes are escaped here.

packages/firestore/scripts/run-tests.ts
wu-hui: Removed.

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: It is shared with spec tests, and I think it would help those to keep using the internal one?

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: Yes. The SDK keeps running without this, and generates lots of errors. It won't break the generation, but it's annoying.

integration/firestore/bundle_test_data/generate_test_bundle.ts
wu-hui: Done.

packages/firestore/test/integration/util/emulator_settings.ts
wu-hui: After I moved them to `settings.ts`, the failure reminded me why they are in a separate file: `settings.ts` itself imports relative path `../../../../../config/project.json`, which will not work in the rollup build.

I had to do this, and let the tree-shaking help to remove it.

PR #3960, opened 2020-10-16
Remove null and NaN checks in filters

thebrianchen:


.changeset/wicked-cups-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/firestore': minor
+---
+
+Removed validation of `null` and `NaN` values in filters. However, note that only `==` and `!=` filters match against `null` and `NaN`.
```
wilhuff, line 5: > However, note that only `==` and `!=` filters match against `null` and `NaN`.

This reads to me in a way that suggests we're removing guard rails but still expecting users to remember this. I'd suggest something more like this:

> Removed excess validation of `null` and `NaN` values in query filters. This more closely aligns the SDK with the Firestore backend, which has always accepted `null` and `NaN` for all operators, even though this isn't necessarily useful.

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -800,6 +813,14 @@ apiDescribe('Queries', (persistence: boolean) => {
       // With objects.
       const snapshot2 = await coll.where('zip', 'in', [{ code: 500 }]).get();
       expect(toDataArray(snapshot2)).to.deep.equal([{ zip: { code: 500 } }]);
+
+      // With null.
+      const snapshot3 = await coll.where('zip', 'in', [null]).get();
```
wilhuff, line 818: Do we need a test that shows that "in" with null/nan and something else still finds the something else?

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -932,6 +955,18 @@ apiDescribe('Queries', (persistence: boolean) => {
         .where('array', 'array-contains-any', [{ a: 42 }])
         .get();
       expect(toDataArray(snapshot2)).to.deep.equal([{ array: [{ a: 42 }] }]);
+
+      // With null.
+      const snapshot3 = await coll
+        .where('zip', 'array-contains-any', [null])
```
wilhuff, line 961: Do we need a test that shows that "array-contains-any" with null/nan and something else still finds the something else?

wilhuff (approved):
LGTM

packages/firestore/test/integration/api/query.test.ts
thebrianchen: done.

packages/firestore/test/integration/api/query.test.ts
thebrianchen: done.

PR #3962, opened 2020-10-18
Move bundle integration tests under api/ and complete exp API

wu-hui:
Add code to generate testing bundle string from a template, by replacing project id and re-calculating lengths. It is fragile, but should suffice for our needs.


packages/firestore/src/remote/serializer.ts
```diff
@@ -339,21 +339,24 @@ export function fromName(
...
-    (!resource.get(3) && !serializer.databaseId.database) ||
-      resource.get(3) === serializer.databaseId.database,
-    'Tried to deserialize key from different database: ' +
-      resource.get(3) +
-      ' vs ' +
-      serializer.databaseId.database
-  );
+
+  if (resource.get(1) !== serializer.databaseId.projectId) {
+    throw new Error(
```
schmidt-sebastian, line 344: Can we throw a FirestoreError here (with Code.INVALID_ARGUMENT)?. We usually throw FirestoreErrors from the SDK.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -543,7 +543,9 @@ export class FirestoreClient {
 
   getNamedQuery(queryName: string): Promise<NamedQuery | undefined> {
     this.verifyNotTerminated();
-    return getNamedQuery(this.localStore, queryName);
+    return this.asyncQueue.enqueue(() =>
```
schmidt-sebastian, line 546: Small comment on `FirestoreClient.loadBundle` above. Instead of logging the exception in FirestoreClient, could we log it in `loadBundle()`?. That would remove the catch statement, and also bubble up an UnhandledPromiseRejection for the user if they don't handle the failure.

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -51,74 +45,133 @@ function verifyInProgress(
...
+ * The function is commented out, but kept for documentation purpose. It accesses SDK
+ * internals, which is not available in test:minified.
+ *
+ * The tests uses `BUNDLE_TEMPLATE` as test data instead, which is generated from this function
+ * and replaced with different project IDs when required. To update `BUNDLE_TEMPALTE`, you
+ * need to uncomment the function, and copy/paste from the console output to `BUNDLE_TEMPALTE`.
+ * It is manual, but should be required only rarely.
+ */
+/*
+function bundleWithTestDocsAndQueries(projectId: string): string {
```
schmidt-sebastian, line 60: Can you move this method to one of our helper/util classes and uncomment it? If you leave it here and leave it commented out, the function will break as we refactor the surrounding code.

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -51,74 +45,133 @@ function verifyInProgress(
...
+      if (char === '}') {
+        depth -= 1;
+        if (depth === 0) {
+          // Obviously assuming TEMPLATE_PROJECT_ID only appear as project id in
+          // the bundle, not as document id or filed value.
+          elements.push(runningElement.replace(TEMPLATE_PROJECT_ID, projectId));
+          runningElement = '';
+        }
+      }
+    }
```
schmidt-sebastian, line 152: This mostly performs the same processing on the same static string. Have you considered storing the bundle as an array of template strings?

```
const elements = [
     "abc/{0}/def",
      ...
].map(val => val.format(projectID));
```

You can then calculate the lengths as you are doing right now.

schmidt-sebastian (commented):
Thanks! This is much better (and shorter!)

packages/firestore/test/integration/api/bundle.test.ts
schmidt-sebastian: Manually, yes.

wu-hui: Done. Only `String.format` does not seem to exist, and `replace` should be enough for our purpose here?

packages/firestore/src/core/firestore_client.ts
schmidt-sebastian: I don't see this change. Can you push it?

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -51,74 +45,57 @@ function verifyInProgress(
...
+  '{"documentMetadata":{"name":"projects/{0}/databases/(default)/documents/coll-1/a","readTime":{"seconds":1000,"nanos":9999},"exists":true}}',
+  '{"document":{"name":"projects/{0}/databases/(default)/documents/coll-1/a","createTime":{"seconds":1,"nanos":9},"updateTime":{"seconds":1,"nanos":9},"fields":{"k":{"stringValue":"a"},"bar":{"integerValue":1}}}}',
+  '{"documentMetadata":{"name":"projects/{0}/databases/(default)/documents/coll-1/b","readTime":{"seconds":1000,"nanos":9999},"exists":true}}',
+  '{"document":{"name":"projects/{0}/databases/(default)/documents/coll-1/b","createTime":{"seconds":1,"nanos":9},"updateTime":{"seconds":1,"nanos":9},"fields":{"k":{"stringValue":"b"},"bar":{"integerValue":2}}}}'
+];
 
+apiDescribe('Bundles', (persistence: boolean) => {
   function verifySnapEqualTestDocs(snap: firestore.QuerySnapshot): void {
```
schmidt-sebastian, line 61: s verifySnapEqualTestDocs/verifySnapEqualsTestDocs ?

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -51,74 +45,57 @@ function verifyInProgress(
...
   }
 
+  /**
+   * Returns a valid bundle string from replacing project id in `BUNDLE_TEMPLATE` with the given
+   * db project id (also recalculate length prefixes).
+   */
+  function bundleString(db: firestore.FirebaseFirestore): string {
+    const projectId: string = db.app.options.projectId;
+
+    // Extract elements from BUNDLE_TEMPLATE, and replace project ID.
```
schmidt-sebastian, line 75:
```suggestion
    // Extract elements from BUNDLE_TEMPLATE and replace the project ID.
```

packages/firestore/src/core/firestore_client.ts
wu-hui: As discussed, the log is moved into `sync_engine.ts`

packages/firestore/test/integration/api/bundle.test.ts
wu-hui: Done.

packages/firestore/test/integration/api/bundle.test.ts
wu-hui: Done.

schmidt-sebastian (approved):

PR #3965, opened 2020-10-19
Have user provide React Native AsyncStorage object

hsubox76:
This is a breaking change.

Instead of importing the `AsyncStorage` bundled with 'react-native', which is now deprecated in favor of getting it from `@react-native-community/async-storage`, we let the user provide it with a required `firebase.useReactNativeAsyncStorage()` method which must be called before initializing auth if the user is using Firebase Auth and React Native.

We cannot move to importing the new one by default because it will break Expo users. We cannot conditionally import one way or the other because the React Native bundler doesn't support conditional `require`s and the `require` for the one that doesn't exist will break the build even if it's in a fork that isn't reached.

Tested manually on iOS, Android, and Expo. Will test further before release.

### DIFF
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -112,6 +112,12 @@ export interface FirebaseNamespace {
   // Sets log handler for all Firebase components.
   onLog(logCallback: LogCallback, options?: LogOptions): void;
 
+  /**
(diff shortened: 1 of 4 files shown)


hsubox76:
Will hold off until modularized package. May put it in the compat package.

PR #3966, opened 2020-10-20
Support generating ref docs for Firestore exp & lite

Feiyang1:
- Added support for generating ref docs for firestore exp and lite using api-extractor fork
- Updated the ref docs


schmidt-sebastian (commented):
Do you know why the inline documentation does not show up? We have added JSDoc to all public methods.

Feiyang1:
@schmidt-sebastian It's using the manually created d.ts files. There is no comments in them.
I think you have an open PR that changes to generate d.ts files instead. We should get comments in the doc once it is merged.

schmidt-sebastian:
> I think you have an open PR that changes to generate d.ts files instead. We should get comments in the doc once it is merged.

That's still a WIP :(

schmidt-sebastian (approved):

hsubox76 (approved):

PR #3967, opened 2020-10-20
Remove JS input validation for Storage

schmidt-sebastian:
Removes all input validation that can be caught by TypeScript transpiler.


packages/storage/src/implementation/metadata.ts
```diff
@@ -206,20 +207,23 @@ export function toResourceString(
   return JSON.stringify(resource);
 }
 
-export function metadataValidator(p: unknown): void {
-  if (!type.isObject(p) || !p) {
-    throw 'Expected Metadata object.';
-  }
+export function validateMetadata(p: { [key: string]: unknown }): void {
```
hsubox76, line 210: I actually got rid of this entirely in storage-exp because it's checking types of Metadata fields which should be caught by typescript check? Or did I make a mistake?

packages/storage/src/reference.ts
```diff
@@ -353,3 +338,23 @@ export class Reference {
     }
   }
 }
+
+export function validateNumber(
```
hsubox76, line 342: Should this go in a separate file? type.ts?

hsubox76 (approved):

Feiyang1:
wow, impressive 10% saving!

PR #3968, opened 2020-10-20
Fix a typing error in database-types

Feiyang1:
`push()` should return a `ThenableReference`

### DIFF
--- a/.changeset/tame-knives-change.md
+++ b/.changeset/tame-knives-change.md
@@ -4,4 +4,4 @@
 "firebase": patch
 ---
 
-Updated `ThenableReference` to only implement `then` and `catch` to match the implementation.
+Updated the type definition for `ThenableReference` to only implement `then` and `catch`, which matches the implementation.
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
LGTM. I tried to update the implementation as well, but we would need to handle `finally` and `toStringTag` as well. This PR is a good first step.

schmidt-sebastian (approved):

Feiyang1:
That reminds me that our public typing is wrong as well. I updated the thenable to only implement `then` and `catch`. Can you please take another look?
@schmidt-sebastian

.changeset/tame-knives-change.md
```diff
@@ -0,0 +1,7 @@
+---
+"@firebase/database-types": patch
+"@firebase/database": patch
+"firebase": patch
+---
+
+Updated `ThenableReference` to only implement `then` and `catch` to match the implementation.
```
schmidt-sebastian, line 7:
```suggestion
Updated the type definition for `ThenableReference` to only implement `then` and `catch`, which matches the implementation.
```

schmidt-sebastian (approved):

hsubox76 (approved):

PR #3974, opened 2020-10-21
Using Compat-Layer for IndexedDB persistence (v2)

schmidt-sebastian:
This is a new version of https://github.com/firebase/firebase-js-sdk/pull/3974. The main goal is to keep the code similar to the other SDKs.

Instead of using https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/exp/src/api/components.ts to manage optional dependencies, it transistions firestore-exp to use `FirestoreClient`. As a a result, the code from `components.ts` is moved to FirestoreClient and `firestore-exp` calls into the tree-shakeable FirestoreClient APIs for the first batch of methods.

One change from #3974 remains: The credential registration happens in two layers. FirestoreClient registers a listener with FirebaseAuth. This allows us to swap the logic for credential changes when RemoteStore is instantiated (via the OnlineComponentProvider).

Let me know if you want to chat before reviewing.


packages/firestore/exp/src/api/database.ts
```diff
@@ -326,12 +266,12 @@ export function enableMultiTabIndexedDbPersistence(
  * but the client remains usable.
  */
 function setPersistenceProviders(
-  firestore: FirebaseFirestore,
+  firestore: FirestoreClient,
```
wilhuff, line 269: This instance is no longer a Firestore or FirestoreCompat: rename to `client` or similar?

packages/firestore/exp/src/api/database.ts
```diff
@@ -250,15 +196,12 @@ export function getFirestore(app: FirebaseApp): FirebaseFirestore {
...
+  firestore: FirestoreCompat,
   persistenceSettings?: PersistenceSettings
 ): Promise<void> {
   verifyNotInitialized(firestore);
 
-  // `_getSettings()` freezes the client settings and prevents further changes
-  // to the components (as `verifyNotInitialized()` would fail). Components can
-  // then be accessed via `getOfflineComponentProvider()` and
-  // `getOnlineComponentProvider()`
+  const firestoreClient = firestore._ensureClientConfigured();
```
wilhuff, line 204: There aren't that many different kinds of clients in this code base, so you can probably abbreviate this to just `client` if you wanted to. Here and throughout if you choose to.

packages/firestore/lite/src/api/components.ts
```diff
@@ -88,3 +83,18 @@ export function removeComponents(firestore: FirebaseFirestore): void {
...
+  persistenceKey: string,
+  settings: Settings
+): DatabaseInfo {
+  return new DatabaseInfo(
+    databaseId,
+    persistenceKey,
+    settings.host ?? DEFAULT_HOST,
+    settings.ssl ?? DEFAULT_SSL,
+    /* forceLongPolling= */ false,
+    /* forceAutoDetectLongPolling= */ true
```
wilhuff, line 98: Why not consult settings for this?

packages/firestore/src/api/database.ts
```diff
@@ -314,55 +310,19 @@ export class MemoryPersistenceProvider implements PersistenceProvider {
...
-        );
-        deferred.resolve();
-      } catch (e) {
-        deferred.reject(e);
-      }
-    });
-    return deferred.promise;
+    return enableIndexedDbPersistence(firestore, { forceOwnership });
   }
+  enableMultiTabIndexedDbPersistence = enableMultiTabIndexedDbPersistence;
```
wilhuff, line 318: This seems like it's a no-op. Did you mean to change a member of `this`? Here and the next line.

packages/firestore/exp/src/api/database.ts
```diff
@@ -76,106 +60,68 @@ export interface Settings extends LiteSettings {
...
-  _setCredentialChangeListener(
-    credentialListener: CredentialChangeListener
-  ): void {
-    logDebug(LOG_TAG, 'Registering credential change listener');
-    this._credentialListener = credentialListener;
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this._receivedInitialUser.promise.then(() =>
-      this._credentialListener(this._user)
-    );
+  _ensureClientConfigured(): FirestoreClient {
```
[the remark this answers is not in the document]
schmidt-sebastian: I implemented this. It allows me to merge some of the code from Firestore-Exp and Firestore. I went back and forth on the naming for these functions. The last commit renames them to "ensureFirestoreConfigured" and "configureFirestore".

packages/firestore/exp/src/api/database.ts
```diff
@@ -76,106 +60,68 @@ export interface Settings extends LiteSettings {
...
+    if (!this._firestoreClient) {
+      this._configureClient();
+    }
+    this._firestoreClient!.verifyNotTerminated();
```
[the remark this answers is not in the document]
schmidt-sebastian: I opted not to do this since this is the only callsite.

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: Mass-renamed `firestoreClient` to `client`.

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: Done

packages/firestore/lite/src/api/components.ts
schmidt-sebastian: I originally chose not to pull in FirestoreSettings into firestore-exp since it included a lot of validation code. This is no longer the case so I was able to use FirestoreStettings here.

packages/firestore/src/api/database.ts
schmidt-sebastian: In this instance, the `this` member doesn't matter since these new functions should not rely on an associated `this` context. To make this more readable, I assigned `null` using `bind`.

wilhuff: The thing I'm missing is that this looks like an assignment of a value to itself, which should do nothing and could therefore be removed. What am I missing?

schmidt-sebastian: The left part points to an instance member, while the right part is the free-standing function of the same name. JavaScript can differentiate because access to instance members requires `this`.

wilhuff (approved):
LGTM

PR #3975, opened 2020-10-21
Remote Config Modularization

Feiyang1:
The internal implementation is unchanged (clients and storage). They are created when you initialize remote config and are members of the remote config object, unchanged from the current behavior. It's possible to modularize them, so that they are only included when you import functions that use them, e.g. `fetchConfig`, but it probably doesn't provide any real benefits since everyone who uses RC uses `fetchConfig`. 

There are lots of files in this PR, but most of them are just copied and unchanged. Only the following files are changed/added/deleted:

Changed: 
- 
- `src/remote_config.ts`
- `test/remote_config.test.ts`
- `package.json`
- `index.d.ts` in `remote-config-types-exp`

Added:
-
- `src/api.ts`
- `src/api2.ts`
- `src/index.ts`
- `src/register.ts`
- `api-extractor.json`

TODOs (will address in separate PRs to keep this PR reasonable in terms of size):
-
- Update demo app using modular APIs
- Add documentation
- Compat package


packages-exp/remote-config-exp/CHANGELOG.md
```diff
@@ -0,0 +1,39 @@
+# @firebase/remote-config
+
+## 0.1.28
```
erikeldridge, line 3: I see this file matches master: https://github.com/firebase/firebase-js-sdk/blob/master/packages/remote-config/CHANGELOG.md

When/where will we add a changelog entry for modularization?

packages-exp/remote-config-exp/src/api.ts
```diff
@@ -0,0 +1,176 @@
...
+    case 'silent':
+      rc._logger.logLevel = FirebaseLogLevel.SILENT;
+      break;
+    default:
+      rc._logger.logLevel = FirebaseLogLevel.ERROR;
+  }
+}
+
+function asRemoteConfigImpl(remoteConfig: RemoteConfig): RemoteConfigImpl {
+  return remoteConfig as RemoteConfigImpl;
```
erikeldridge, line 168: Minor: I'm guessing it's a best-practice to wrap the cast operation, but it's new to me. Is there a doc you'd recommend for the rationale?

packages-exp/remote-config-exp/src/api2.ts
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { RemoteConfig } from '@firebase/remote-config-types-exp';
+import { activate, fetchConfig } from './api';
+
+// This API is put in a separate file, so we can stub fetchConfig and activate in tests.
```
erikeldridge, line 21: Helpful comment

erikeldridge (approved):
Thx, Fei 🙏

common/api-review/remote-config-exp.api.md
```diff
@@ -0,0 +1,48 @@
...
+export function ensureInitialized(remoteConfig: RemoteConfig): Promise<void>;
+
+// @public (undocumented)
+export function fetchAndActivate(remoteConfig: RemoteConfig): Promise<boolean>;
+
+// @public (undocumented)
+export function fetchConfig(remoteConfig: RemoteConfig): Promise<void>;
+
+// @public (undocumented)
+export function getAll(remoteConfig: RemoteConfig): Record<string, Value>;
```
erikeldridge, line 25: I'm unfamiliar w `Record`. The return type before was `object`:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/remote-config-types/index.d.ts#L68 -> https://firebase.google.com/docs/reference/js/firebase.remoteconfig.RemoteConfig#getall

Using `{ [key: string]: Value }` still seems idiomatic: 
https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types

What's the value of using `Record`?

Feiyang1: It's an alias of `{ [key:string]: Value}`. The only reason to use it is because it is shorter.

packages-exp/remote-config-exp/src/api.ts
Feiyang1: Yeah, I don't think it's a best-practice. I didn't want to do casting(using `as`) in every API, so I moved it to a function. On reflection, I think it's probably not necessary, and will waste cycles by introducing a no-op function call. 

Changed to do casting directly in individual functions.

packages-exp/remote-config-exp/CHANGELOG.md
Feiyang1: We will add it before the official release. Actually I shouldn't have included the CHANGELOG.md now. We will create it before the official release.

Deleted.

hsubox76 (approved):

common/api-review/remote-config-exp.api.md
erikeldridge: 👍  Good to know. Thx

PR #3981, opened 2020-10-22
Add compat to test commands

hsubox76:
@Feiyang1 Should `@firebase/*-compat` be added to "api-report" too?


Feiyang1:
No, they should not. It won't work for them because they don't export the API.

Feiyang1 (approved):

PR #3983, opened 2020-10-22
Make FirestoreClient tree-shakeable

schmidt-sebastian:
This is a mechanical refactor. The next PR will use these change to introduce the firestore-exp components to FirestoreClient.


packages/firestore/src/api/database.ts
```diff
@@ -676,7 +693,9 @@ export class Firestore implements PublicFirestore, FirebaseService {
   runTransaction<T>(
     updateFunction: (transaction: PublicTransaction) => Promise<T>
   ): Promise<T> {
-    return this.ensureClientConfigured().transaction(
+    this.ensureClientConfigured();
```
wilhuff, line 696: Assign to a temporary rather than referring to `this._firestoreClient!`?

wilhuff (approved):
LGTM

packages/firestore/src/api/database.ts
schmidt-sebastian: Done

PR #3986, opened 2020-10-22
Inline constants in auth-exp

avolkovi:


packages-exp/auth-exp/src/core/action_code_url.ts
```diff
@@ -33,28 +33,29 @@ enum QueryField {
...
-
 /**
  * Maps the mode string in action code URL to Action Code Info operation.
  *
  * @param mode
  * @internal
  */
 function parseMode(mode: string | null): externs.Operation | null {
-  return mode ? MODE_TO_OPERATION_MAP[mode] || null : null;
+  switch (mode) {
```
samhorlbeck, line 43: Is this really smaller?

samhorlbeck (approved):

packages-exp/auth-exp/src/core/action_code_url.ts
avolkovi: no, but it's shakeable

PR #3991, opened 2020-10-23
Add auth-exp to firebase-exp

avolkovi:


Feiyang1 (approved):
Thanks for doing this. Verified locally that script and bundle both work correctly.

PR #4001, opened 2020-10-27
Prevent __proto__ pollution in util.deepExtend

Feiyang1:


packages/util/src/deepCopy.ts
```diff
@@ -62,14 +64,19 @@ export function deepExtend(target: unknown, source: unknown): unknown {
...
   for (const prop in source) {
-    if (!source.hasOwnProperty(prop)) {
+    // use isValidKey to guard against prototype pollution. See https://snyk.io/vuln/SNYK-JS-LODASH-450202
+    if (!source.hasOwnProperty(prop) || !isValidKey(prop)) {
       continue;
     }
-    (target as { [key: string]: unknown })[prop] = deepExtend(
-      (target as { [key: string]: unknown })[prop],
-      (source as { [key: string]: unknown })[prop]
+    (target as Record<string, unknown>)[prop] = deepExtend(
```
schmidt-sebastian, line 71: Maybe a little cleaner if you create a type alias for `Record<string, unknown>` to reduce the repetitive cast.

packages/util/test/deepCopy.test.ts
```diff
@@ -100,7 +99,19 @@ describe('deepExtend', () => {
...
+    expect({ a: source }).to.deep.equal(target);
+    expect(source).to.equal(target.a);
+  });
+
+  it('does not extend property __proto__', () => {
+    const src = JSON.parse('{ "__proto__": { "polluted": "polluted" } }');
+    const a = {};
+    deepExtend(a, src);
+
+    // @ts-expect-error
```
schmidt-sebastian, line 111: In my opinion, this usage of "ts-expect-error" doesn't quite match the spirit. It seems geared towards explicit invalid usage, and would for example allow to test input validation. The usage here is not testing an error case. I would recommend just declaring `a` as an object with arbitrary keys.

schmidt-sebastian (approved):
2 nits but LGTM.

packages/util/test/deepCopy.test.ts
Feiyang1: Good point. I got lazy.

PR #4006, opened 2020-10-28
Bundles as free functions

wu-hui:


avolkovi (approved):
LGTM for the auth change, looks like just whitespace

PR #4008, opened 2020-10-28
Refactor auth error maps to make the verbose error map an optional dependency

samhorlbeck:


packages-exp/auth-exp/src/core/util/assert.ts
```diff
@@ -15,35 +15,112 @@
...
+  AuthErrorCode,
+  AuthErrorParams
+} from '../errors';
 import { _logError } from './log';
 
+type AuthErrorListParams<K> = K extends keyof AuthErrorParams
+  ? [AuthErrorParams[K]]
+  : [];
+type LessAppName<K extends AuthErrorCode> = Omit<AuthErrorParams[K], 'appName'>;
+
```
samhorlbeck, line 32: @avolkovi this is the interesting bit

common/api-review/auth-exp.api.md
```diff
@@ -309,6 +310,11 @@ export class PhoneMultiFactorGenerator implements externs.PhoneMultiFactorGenera
     static assertion(credential: externs.PhoneAuthCredential): externs.PhoneMultiFactorAssertion;
 }
 
+// Warning: (ae-forgotten-export) The symbol "AuthErrorMap" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export const prodErrorMap: AuthErrorMap;
```
avolkovi, line 316: why is this getting referenced in the docs?

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
```diff
@@ -22,6 +22,8 @@ import * as compat from '@firebase/auth-types';
 import * as externs from '@firebase/auth-types-exp';
 import { unwrap, Wrapper } from './wrap';
 
+const _assert: typeof impl._assert = impl._assert;
```
avolkovi, line 25: why do you need these as consts?

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
```diff
@@ -32,7 +34,7 @@ export class RecaptchaVerifier
     app: FirebaseApp = firebase.app()
   ) {
     // API key is required for web client RPC calls.
-    impl.assertFn(app.options?.apiKey, impl.AuthErrorCode.INVALID_API_KEY, {
+    _assert(app.options?.apiKey, impl.AuthErrorCode.INVALID_API_KEY, {
```
avolkovi, line 37: shouldn't the signature be updated here?

packages-exp/auth-exp/src/api/index.test.ts
```diff
@@ -173,7 +173,7 @@ describe('api/_performApiRequest', () => {
       );
       await expect(promise).to.be.rejectedWith(
         FirebaseError,
-        'auth/internal-error'
+        'auth/awesome error'
```
avolkovi, line 176: we should change spaces to dashes in these errors

packages-exp/auth-exp/src/api/index.ts
```diff
@@ -187,8 +175,7 @@ export async function _performSignInRequest<T, V extends IdTokenResponse>(
     customErrorMap
   )) as V;
   if ('mfaPendingCredential' in serverResponse) {
-    throw AUTH_ERROR_FACTORY.create(AuthErrorCode.MFA_REQUIRED, {
-      appName: auth.name,
+    throw _createError(auth, AuthErrorCode.MFA_REQUIRED, {
```
avolkovi, line 178: should this be _fail()?

packages-exp/auth-exp/src/core/errors.ts
```diff
@@ -123,223 +123,240 @@ export const enum AuthErrorCode {
...
+
+export interface ErrorMapRetriever extends AuthErrorMap {
+  (): ErrorMap<AuthErrorCode>;
+}
+
+function _prodErrorMap(): ErrorMap<AuthErrorCode> {
+  return {} as ErrorMap<AuthErrorCode>;
+}
+
+export const verboseErrorMap: AuthErrorMap = _verboseErrorMap;
```
avolkovi, line 358: nit: debugErrorMap? debug vs prod makes more senes to me than verbose vs prod, alternatively you can do long vs short, or verbose vs concise

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -100,21 +98,25 @@ export class StsTokenManager {
...
+      _assert(typeof accessToken === 'string', AuthErrorCode.INTERNAL_ERROR, {
         appName
       });
       manager.accessToken = accessToken;
     }
     if (expirationTime) {
-      assert(typeof expirationTime === 'number', AuthErrorCode.INTERNAL_ERROR, {
-        appName
-      });
+      _assert(
```
avolkovi, line 113: shouldn't this be _assert(assert, auth, code)?

packages-exp/auth-compat-exp/src/recaptcha_verifier.ts
samhorlbeck: We don't have an auth object yet here

packages-exp/auth-exp/src/core/errors.ts
```diff
@@ -373,7 +390,7 @@ export interface AuthErrorParams extends GenericAuthErrorParams {
   };
 }
 
-export const AUTH_ERROR_FACTORY = new ErrorFactory<
+export const FALLBACK_AUTH_ERROR_FACTORY = new ErrorFactory<
```
avolkovi, line 393: nit: DEFAULT_AUTH_ERROR_FACTORY or PROD_ERROR_FACTORY to match prodErrorMap

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -268,6 +275,14 @@ export class AuthImpl implements Auth, _FirebaseService {
     return this.assertedPersistence.persistence.type;
   }
 
+  _updateErrorMap(errorMap: AuthErrorMap): void {
```
avolkovi, line 278: since the dep is required at initialization, can we make this immutable and in the constructor?

avolkovi (approved):
Looks good overall, thanks for doing this.

Couple minor nits around naming of maps/functions

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: We don't have an auth object in the static context

packages-exp/auth-exp/src/core/auth/auth_impl.ts
samhorlbeck: No; the instance is actually constructed here: https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/auth-exp/src/core/auth/register.ts#L67

We don't have access to any of the specific auth deps at that point.

packages-exp/auth-exp/src/core/user/token_manager.ts
samhorlbeck: We don't have an auth instance here

common/api-review/auth-exp.api.md
samhorlbeck: Should be better now

packages-exp/auth-exp/src/core/auth/initialize.ts
```diff
@@ -44,6 +44,9 @@ export function _initializeAuthInstance(
     ? persistence
     : [persistence]
   ).map<Persistence>(_getInstance);
+  if (deps?.errorMap) {
+    auth._updateErrorMap(deps.errorMap);
```
Feiyang1, line 48: I didn't bring it up in the API review, but I wonder if setting the error map should be a global operation instead of a per-instance one because I think if people want to use the debug map, they want it to use it for all instances. No action needed now, we can wait and see if we got any feedback.

Feiyang1 (approved):

packages-exp/auth-exp/src/core/auth/initialize.ts
samhorlbeck: Yeah I mentioned that yesterday in the biweekly check in as well. I definitely think that might be a good longer-term solution if it turns out people like this change

PR #4016, opened 2020-10-29
Remote Config compat package

Feiyang1:


packages-exp/remote-config-compat/src/remoteConfig.ts
```diff
@@ -0,0 +1,115 @@
...
+  getValue
+} from '@firebase/remote-config-exp';
+import { FirebaseService } from '@firebase/app-types/private';
+
+export class RemoteConfigCompatImpl
+  implements RemoteConfigCompat, FirebaseService {
+  constructor(public app: FirebaseApp, private _remoteConfig: RemoteConfig) {}
+
+  get defaultConfig(): { [key: string]: string | number | boolean } {
+    return this._remoteConfig.defaultConfig;
```
erikeldridge, line 46: Thinking: the plan to modularize attribute accessors is to have ppl call `getRemoteConfig()` and then calling the accessor on the returned `RemoteConfig` instance.

packages-exp/remote-config-compat/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function remoteConfigFactory(
+  container: ComponentContainer,
+  namespace?: string
+): RemoteConfigCompatImpl {
+  // TODO: change 'app' to 'app-compat' before the official release
+  const app = container.getProvider('app').getImmediate();
+  // The following call will always succeed because rc `import {...} from '@firebase/remote-config-exp'`
+  const remoteConfig = container.getProvider('remote-config-exp').getImmediate({
```
erikeldridge, line 57: Thinking: if we had a `getRemoteConfig(app, namespace?)` modular method, we could just call that here, but it doesn't make sense to add one until there's customer demand.

erikeldridge (approved):
Thx, @Feiyang1 !

hsubox76 (approved):

PR #4025, opened 2020-11-02
Add remote-config-exp to firebaes-exp

Feiyang1:
- Add remote-config-exp to firebaes-exp
- Make RC test app work with remote-config-exp


erikeldridge (approved):
Thx @Feiyang1 !

packages-exp/remote-config-exp/test_app/index.js
```diff
@@ -48,8 +48,8 @@ window.onload = function () {
 
 function initializeFirebase() {
   const val = document.querySelector('#firebase-config').value;
-  firebase.initializeApp(parseObjFromStr(val));
-  remoteConfig = firebase.remoteConfig();
+  const app = firebase.app.initializeApp(parseObjFromStr(val));
+  remoteConfig = firebase.remoteConfig.getRemoteConfig(app);
```
erikeldridge, line 52: @Feiyang1 since we assign RC to a global var here, do we need to prefix RC references w `firebase` below? 
If we no longer need the global var, mind removing it?

Feiyang1: `remoteConfig` is the remote config instance. `firebase.remoteConfig` is the namespace where rc functions are attached. We still need it and to avoid confusion, I renamed `remoteConfig` to `rcInstance`.

hsubox76 (approved):

PR #4027, opened 2020-11-03
use the firebase-exp version for `SDK_VERSION` in app-exp

Feiyang1:


Feiyang1:
@avolkovi  FYI
@schmidt-sebastian Does Firestore use this value to track usage?

schmidt-sebastian:
> @schmidt-sebastian Does Firestore use this value to track usage?

I don't think so. We use our own header: https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/firestore/src/remote/rest_connection.ts;l=41

Feiyang1:
@schmidt-sebastian Do you want to track the exp SDK usage? If so, I think we need to find a way to inject the version number to Firestore.

hsubox76 (approved):

schmidt-sebastian:
> @schmidt-sebastian Do you want to track the exp SDK usage? If so, I think we need to find a way to inject the version number to Firestore.

Yes.

PR #4028, opened 2020-11-03
Update versions for exp packages

Feiyang1:
Change from `0.0.800` to `0.0.900` to reflect next major version (v9)


hsubox76 (changes requested):
I think I may have merged functions-compat after you started this PR, can you include it?

hsubox76 (approved):

PR #4038, opened 2020-11-04
Compat class for FieldPath

schmidt-sebastian:
This PR removes the FieldPath class in the legacy class in favor of a Compat class that delegates to firestore-exp. It changes all callsites to accept either the Compat class or the firestore-exp FieldPath.

### DIFF
--- a/packages/firestore/lite/src/api/field_path.ts
+++ b/packages/firestore/lite/src/api/field_path.ts
@@ -15,8 +15,11 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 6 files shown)


schmidt-sebastian:
Assigning to Andy since Denver doesn't have time.

schmidt-sebastian:
Note: this is similar to https://github.com/firebase/firebase-js-sdk/pull/3819

wu-hui (approved):

PR #4039, opened 2020-11-04
Rename all public API types to PublicX

schmidt-sebastian:


wilhuff (approved):
LGTM

packages/firestore/src/api/database.ts
```diff
@@ -635,7 +635,7 @@ export class WriteBatch implements PublicWriteBatch {
   ): WriteBatch;
   update(
     documentRef: PublicDocumentReference<unknown>,
-    fieldOrUpdateData: string | ExternalFieldPath | UpdateData,
+    fieldOrUpdateData: string | ExternalFieldPath | PublicUpdateData,
```
schmidt-sebastian, line 638: Note that this will get cleaned up by https://github.com/firebase/firebase-js-sdk/pull/4038 (which uses PublicFieldPath here)

PR #4043, opened 2020-11-05
Compat class for DocumentReference

schmidt-sebastian:
This adds the Compat class for DocumentReference, along with the nasty hack "UntypedDataConverter" that treats the different FirestoreUserConverter (for all three SDKs) the same by ignoring the type of QuerySnapshot.

Note that this PR pulls in a lot of new code - I hope that I can get rid of most of this when I we add the QuerySnapshot Compat class.


packages/firestore/src/api/database.ts
```diff
@@ -138,9 +149,10 @@ import {
   WhereFilterOp as PublicWhereFilterOp,
   WriteBatch as PublicWriteBatch
 } from '@firebase/firestore-types';
-import { newUserDataReader } from '../../lite/src/api/reference';
+
 import { makeDatabaseInfo } from '../../lite/src/api/database';
 import { DEFAULT_HOST } from '../../lite/src/api/components';
+import * as exp from '../../exp/index';
```
wilhuff, line 155: I thought `*` imports were to be avoided?

packages/firestore/src/api/database.ts
```diff
@@ -893,102 +877,129 @@ export class DocumentReference<T = PublicDocumentData>
...
-      complete: args[currArg + 2] as CompleteFn
-    };
-
-    return firestoreClientListen(
-      this._firestoreClient,
-      newQueryForPath(this._key.path),
-      internalOptions,
-      observer
+    const options = extractSnapshotOptions(args);
+    const observer = wrapObserver<DocumentSnapshot<T>, exp.DocumentSnapshot<T>>(
```
wilhuff, line 881: Above you're using import renaming to use `ExpDocumentReference` but here you're using a star import to get `exp.DocumentSnapshot`. We should do this consistently.

packages/firestore/src/api/database.ts
```diff
@@ -893,102 +877,129 @@ export class DocumentReference<T = PublicDocumentData>
...
-    debugAssert(
-      snapshot.docs.size <= 1,
-      'Too many documents returned on a document query'
-    );
-    const doc = snapshot.docs.get(this._key);
+/**
+ * Replaces the function name in an error thrown by the firestore-exp API
+ * with the function names used in the classic API.
+ */
+function rewriteError(
```
wilhuff, line 935: Since this is specific to method references maybe call this `rewriteErrorMethod` or maybe even just `replaceMethod`? Alternatively, leave it as a general `rewriteError` and make the caller supply the parens?

packages/firestore/src/api/database.ts
```diff
@@ -893,102 +877,129 @@ export class DocumentReference<T = PublicDocumentData>
...
+    if (typeof arg === 'object' && !isPartialObserver(arg)) {
+      return arg as exp.SnapshotListenOptions;
+    }
+  }
+  return {};
+}
+
+/**
+ * Creates an observer that can be passed to the firestore-exp SDK. The
+ * observer converts all observed values into the format expected by the shim.
```
wilhuff, line 965: This is no longer a "shim", right?

wilhuff (approved):
LGTM with some nits.

packages-exp/auth-exp/index.webworker.ts
```diff
@@ -37,7 +37,10 @@ registerAuth(ClientPlatform.WORKER);
 export function getAuth(app = getApp()): Auth {
   // Unlike the other environments, we need to explicitly check if indexedDb is
   // available. That means doing the whole rigamarole
-  const auth = _getProvider(app, _ComponentName.AUTH).getImmediate() as AuthImpl;
+  const auth = _getProvider(
```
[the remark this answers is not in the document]
schmidt-sebastian: No - they keep showing up though due to the Presubmit hook. I will revert them before merging.

packages/firestore/src/api/database.ts
schmidt-sebastian: Yeah, I should have caught this. Removed this import and aliased DocumentReference to ExpDocumentReference.

packages/firestore/src/api/database.ts
schmidt-sebastian: Used `ExpDocumentReference`

packages/firestore/src/api/database.ts
schmidt-sebastian: Renamed to `replaceFunctionName`.

packages/firestore/src/api/database.ts
schmidt-sebastian: Fixed

PR #4047, opened 2020-11-06
Add retry logic around logic IndexedDB connections.

avolkovi:
Similar to https://github.com/firebase/firebase-js-sdk/blob/a10c18f8996fc35942779f5fea5690ae5d102bb0/packages/firestore/src/local/simple_db.ts

### DIFF
--- a/packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
+++ b/packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
@@ -164,9 +164,10 @@ function deleteObject(db: IDBDatabase, key: string): Promise<void> {
 
 /** @internal */
 export const _POLLING_INTERVAL_MS = 800;
+/** @internal */
+export const _TRANSACTION_RETRY_COUNT = 3;
 
 class IndexedDBLocalPersistence implements Persistence {
-  private static readonly TRANSACTION_RETRY_COUNT = 3;
   static type: 'LOCAL' = 'LOCAL';
 
   type = PersistenceType.LOCAL;
@@ -210,7 +211,7 @@ class IndexedDBLocalPersistence implements Persistence {
         const db = await this._openDb();
         return await op(db);
       } catch (e) {
-        if (numAttempts++ > IndexedDBLocalPersistence.TRANSACTION_RETRY_COUNT) {
+        if (numAttempts++ > _TRANSACTION_RETRY_COUNT) {
           throw e;
         }
         if (this.db) {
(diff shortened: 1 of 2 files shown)


packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
```diff
@@ -166,6 +166,7 @@ function deleteObject(db: IDBDatabase, key: string): Promise<void> {
 export const _POLLING_INTERVAL_MS = 800;
 
 class IndexedDBLocalPersistence implements Persistence {
+  private static readonly TRANSACTION_RETRY_COUNT = 3;
```
samhorlbeck, line 169: Don't use static variables (per Fei's tree shaking work)

Use this as a file-local `const` instead

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
```diff
@@ -193,14 +194,34 @@ class IndexedDBLocalPersistence implements Persistence {
...
+        return await op(db);
+      } catch (e) {
+        if (numAttempts++ > IndexedDBLocalPersistence.TRANSACTION_RETRY_COUNT) {
+          throw e;
+        }
+        if (this.db) {
+          this.db.close();
+          this.db = undefined;
+        }
+        // TODO: consider adding exponential backoff
```
samhorlbeck, line 220: Is it worth adding any sort of delay here?

samhorlbeck (approved):

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
avolkovi: question for @schmidt-sebastian I think he mentioned they had this in Firestore but I couldn't find proof

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
avolkovi: done

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
schmidt-sebastian: We retry immediately at this scope, but at the higher level we retry with backoff: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/util/async_queue.ts#L361

schmidt-sebastian (approved):

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
avolkovi: thanks, will consider adding that in a followup if this doesn't fix it

PR #4051, opened 2020-11-09
Compat class for DocumentSnapshot

schmidt-sebastian:
This is the Compat class for DocumentSnapshot. The only real challenge here is that it we now either generate firestore-exp or Classic DocumentReferences or Bytes/Blob types. To facilitate this, I pass the UserDataWriter as an argument to the DocumentSnapshot.


packages/firestore/exp/src/api/reference.ts
```diff
@@ -233,14 +237,15 @@ export function getDocsFromCache<T>(
 ): Promise<QuerySnapshot<T>> {
   const firestore = cast(query.firestore, FirebaseFirestore);
   const client = ensureFirestoreConfigured(firestore);
+  const userDataWriter = newExpUserDataWriter(firestore, query._converter);
```
wilhuff, line 240: Optional: you could save a little bit of memory by deferring the creation of the `userDataWriter` until you need it just as you're creating the `QuerySnapshot`.

packages/firestore/lite/src/api/reference.ts
```diff
@@ -1233,3 +1247,14 @@ export function newUserDataReader(
     serializer
   );
 }
+
+export function newExpUserDataWriter<T>(
```
wilhuff, line 1251: I'm surprised this doesn't live closer to the UserDataWriter. It doesn't seem like it belongs in `reference.ts`.

Also, it's confusing that this is named `newExpUserDataWriter` but it lives in the lite SDK?

packages/firestore/lite/src/api/reference.ts
```diff
@@ -1233,3 +1247,14 @@ export function newUserDataReader(
...
+  firestore: FirebaseFirestore,
+  converter: UntypedFirestoreDataConverter<T> | null
+): UserDataWriter {
+  return new UserDataWriter(
```
wilhuff, line 1255: A more traditional arrangement would be to have `UserDataWriter` be an abstract class and have different subclasses supply the conversion routines by implementing methods that are missing. Then you wouldn't need a `newExpUserDataWriter` function, you'd just invoke the constructor of the subtype instead.

packages/firestore/src/api/user_data_writer.ts
```diff
@@ -54,14 +54,23 @@ export type ServerTimestampBehavior = 'estimate' | 'previous' | 'none';
...
     private readonly databaseId: DatabaseId,
-    private readonly serverTimestampBehavior: ServerTimestampBehavior,
     private readonly referenceFactory: (key: DocumentKey) => unknown,
     private readonly bytesFactory: (bytes: ByteString) => Bytes
   ) {}
 
-  convertValue(value: ProtoValue): unknown {
+  convertValue(
+    value: ProtoValue,
+    serverTimestampBehavior: ServerTimestampBehavior = 'none'
```
wilhuff, line 72: Why can't this remain a constructor parameter? That seemed like a cleaner arrangement.

packages/firestore/exp/src/api/reference.ts
schmidt-sebastian: I slightly prefer the current arrangement as I can use the same pattern across all functions. It also keeps the assignment in the Lambda shorter, which makes it fit one in line.

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: Huh, you do catch all the weirdness here :) I was hoping I could get away with this weird name. I ended up fixing this by applying your suggestion below. There is now a LiteUserDataWriter, an ExpUserDataWriter and an UserDataWriter (which we could name ClassicUserDataWriter, but I don't yet use this term in our class names). Note that LiteUserDataWriter and ExpUserDataWriter are essentially the same, but I kept them separate for clarity. I also re-exported the DocumentReference types from `/exp/src/api/reference` for additional clarity.

packages/firestore/lite/src/api/reference.ts
schmidt-sebastian: I updated the PR as suggested (which would match what we discussed last week). I was a bit lazy in the first iteration, as the factory functions were already implemented.

packages/firestore/src/api/user_data_writer.ts
schmidt-sebastian: Since I am now "injecting" the UserDataWriter when the DocumentSnapshot is constructed, I don't yet know what the setting for `serverTimestampBehavior` will be.

packages/firestore/src/api/user_data_writer.ts
```diff
@@ -158,18 +154,70 @@ export class UserDataWriter {
...
+    super();
+  }
+
+  protected convertBytes(bytes: ByteString): Blob {
+    return new Blob(bytes);
+  }
+
+  protected convertReference(name: string): DocumentReference {
+    const key = this.convertDocumentKey(name, this.firestore._databaseId);
+    return DocumentReference.forKey(key, this.firestore, /* converter= */ null);
```
wilhuff, line 187: This seems like it now unconditionally creates a reference without a converter, where previously the converter was propagated in from the API construct that was used to create the reference/snapshot. Why is this OK?

schmidt-sebastian: I thought about this problem, but I think this new behavior actually makes sense. We don't know anything about the document that is referenced here, so it seems like we shouldn't assume that the original converter applies here. I think this is an edge case that probably no one noticed, but it would be pretty challenging to continue exposing the old behavior if we deemed this a breaking change.

wilhuff: I think it's reasonable to call the old behavior a bug and call this a fix :-).

We should probably call this out in the release notes.

wilhuff (approved):
LGTM

PR #4053, opened 2020-11-10
Compat class for WriteBatch

schmidt-sebastian:
This is a simple compat layer for WriteBatch.


wu-hui (approved):
LGTM, sorry for the delay.

PR #4055, opened 2020-11-10
Update firestore types to allow specifying DocumentData type

kentcdodds:
I thought this would be more helpful than an issue. Not 100% certain that this is correct, but it seems to work for me :) Thanks!

Before these changes, I was unable to specify the type of `.collection('id').get()`.

### DIFF
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -65,11 +65,11 @@ export class FirebaseFirestore {
 
   enablePersistence(settings?: PersistenceSettings): Promise<void>;
 
-  collection(collectionPath: string): CollectionReference<DocumentData>;
+  collection<T = DocumentData>(collectionPath: string): CollectionReference<T>;
 
-  doc(documentPath: string): DocumentReference<DocumentData>;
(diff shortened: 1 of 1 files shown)


gugahoi:
I believe the correct way to specify the return value for a collection or document interaction is via [withConverter](https://firebase.google.com/docs/reference/js/firebase.firestore.CollectionReference#withconverter).

kentcdodds:
Hmmm... I'm not sure I understand. What's the purpose of `collection` then?

gugahoi:
Super contrived example:

```typescript
interface User {
  id:string
  name: string
}

const converter = {
  toFirestore(user: User): firebase.firestore.DocumentData {
    return {userId: user.id, name: user.name}
  },
  fromFirestore(data: firebase.firestore.DocumentData): User {
    return {id: data.userId, name: data.name}
  },
}

db.collection('users').withConverter(converter).get()
```

kentcdodds:
Thank you! I appreciate you taking the time to explain this to me. This makes sense 👍

gugahoi:
Glad I could help @kentcdodds! Thanks for all your work, I'm a fan 😄

PR #4056, opened 2020-11-10
Accept Compat types in firestore-exp API

schmidt-sebastian:
This PR allows all firestore-exp APIs to transparently accept Compat types, which makes migration for the user easier.


packages/firestore/exp/src/api/database.ts
```diff
@@ -169,6 +170,7 @@ export function enableIndexedDbPersistence(
   firestore: FirebaseFirestore,
```
Feiyang1, line 170: Is compat `FirebaseFirestore` compatible with the exp `FirebaseFirestore`? Or should the type be something like`FirebaseFirestore | FirebaseFirestoreCompat`?

packages/firestore/src/util/input_validation.ts
```diff
@@ -166,7 +166,8 @@ export function tryGetCustomObjectType(input: object): string | null {
 }
 
 /**
- * Casts `obj` to `T`. Throws if  `obj` is not an instance of `T`.
+ * Casts `obj` to `T`, optionally unwrapping Compat types to expose the
```
Feiyang1, line 169: Can we just check `obj._delegate` instead of relying on `instanceof`ing the abstract class, so that I can put it into `@firebase/util` and share with other SDKs?

packages/firestore/exp/src/api/reference.ts
```diff
@@ -659,6 +667,10 @@ export function onSnapshot<T>(
   reference: Query<T> | DocumentReference<T>,
   ...args: unknown[]
 ): Unsubscribe {
+  if (reference instanceof Compat) {
```
Feiyang1, line 670: Any reason you don't use `cast` here and in other places where you directly use `instanceof`?

Feiyang1 (commented):
I really like the approach as it makes it seamless to use compat and exp side by side. My only question is (asked in the comment too) - do people need to cast the compat type to the exp type in order to use the exp API?

packages/firestore/exp/src/api/database.ts
schmidt-sebastian: I don't want to pollute the generated types if we do ever merge https://github.com/firebase/firebase-js-sdk/pull/3790

packages/firestore/exp/src/api/reference.ts
schmidt-sebastian: I am unable to use `cast` in a couple of places where the input type is a union of two distinct class types.

packages/firestore/src/util/input_validation.ts
schmidt-sebastian: Updated. This function also does a name check and provides a different error message if there it detects that two classes share the same name but are not compatible. I was wondering if we need to strip suffixes here, since Rollup adds "$1" and counting during the build phase?

schmidt-sebastian:
To answer your question -I am envisioning a world where all types are fully exchangeable. The user will not need to do any casting.

schmidt-sebastian:
@Feiyang1 Can you take another look?

packages/firestore/src/util/input_validation.ts
Feiyang1: Thanks. Yes, rollup will append `$x` to things with the same name if they end up in the same bundle. I'm not sure if it's the case with webpack which creates many closures instead of one big top level scope. I think it very much depends on the format of the bundle, thus is hard for us to predict/optimize for. I would just leave it as is for now and see if developer reports that the error is not helpful.

Feiyang1 (approved):

PR #4059, opened 2020-11-12
[auth] Retry IndexedDB errors a fixed number of times

### LINKED ISSUE #1926: auth: Failed to execute 'transaction' on 'IDBDatabase'

  * Operating System version: Android/Windows
  * Browser version: Chrome
  * Firebase SDK version: 6.2.2
  * Firebase Product: auth

Migrating this issue on behalf of @alexplumb from #1533. We've seen this indexdb issue in Firestore on Safari 12.2 (both in #1533 and the canonical one for that issue, #1670). 

However, Alex doesn't seem to have persistence enabled, so firestore shouldn't even be looking at indexdb. The name of the indexdb in question is "firebaseLocalStorageDb" which looks like it's the auth db (https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/src/storage/indexeddb.js#L191)

#### Steps to reproduce:
Happens inconsistently; unable to reproduce. :(

avolkovi:
There appears to be an issue with webkit where it periodically closes the connections such as when the tab is backgrounded (https://bugs.webkit.org/show_bug.cgi?id=197050).  We can't repro, but Firestore did something similar and have stopped getting error reports.

Closes #1926

### DIFF
--- a/packages/auth/src/storage/indexeddb.js
+++ b/packages/auth/src/storage/indexeddb.js
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -227,6 +227,14 @@ fireauth.storage.IndexedDB.POLLING_DELAY_ = 800;
 
 
 /**
+ * Maximum number of times to retry a transaction in the event the connection is
+ * closed.
+ * @private @const {number}
+ */
+fireauth.storage.IndexedDB.TRANSACTION_RETRY_COUNT_ = 3;
+
+
+/**
  * The indexedDB polling stop error.
  * @private @const {string}
  */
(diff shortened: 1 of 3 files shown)


packages/auth/src/storage/indexeddb.js
```diff
@@ -416,43 +455,42 @@ fireauth.storage.IndexedDB.prototype.onIDBRequest_ =
...
-        var objectStore = self.getDataObjectStore_(
-            self.getTransaction_(dbTemp, true));
-        return self.onIDBRequest_(objectStore.get(key));
-      })
-      .then(function(data) {
-        var objectStore = self.getDataObjectStore_(
-            self.getTransaction_(dbTemp, true));
+  let isLocked = false;
+  return this
+    .withRetry_((db) => {
```
Feiyang1, line 460: Does closure compile it to regular functions? I wasn't sure if you can use arrow functions in closure.

Feiyang1 (approved):

packages/auth/src/storage/indexeddb.js
avolkovi: seems to compile to `function() {}.bind(this)` https://github.com/google/closure-compiler/issues/2853

trevorhere (commented):
looks good to me!

PR #4065, opened 2020-11-13
Compat class for Query and CollectionReference

schmidt-sebastian:
This series of PRs is coming to an end.
This PR adds Compat classes for Query and CollectionReference and is split across two commits. The first one creates the Compat classes, the second one moves code from src/api/database.ts to lite/src/api/reference.ts as the helper functions are now no longer needed by the classic API.


packages/firestore/src/api/database.ts
```diff
@@ -1170,590 +1125,119 @@ export class QueryDocumentSnapshot<T = PublicDocumentData>
...
-        'Invalid query. All where filters with an inequality' +
-          ' (<, <=, >, or >=) must be on the same field. But you have' +
-          ` inequality filters on '${existingField.toString()}'` +
-          ` and '${filter.field.toString()}'`
+  orderBy(
+    fieldPath: string | FieldPath,
+    directionStr?: PublicOrderByDirection
+  ): Query<T> {
+    try {
+      // The "as string" cast is a little bit of a hack. `orderBy` accepts the
```
wilhuff, line 1158: I'm surprised that we can't just make this different in our public typings vs are internal actual types. It seems like we could declare this in the .ts file and just omit it from the .d.ts?

packages/firestore/src/api/database.ts
```diff
@@ -1942,77 +1361,59 @@ export class CollectionReference<T = PublicDocumentData>
...
 
-  doc(pathString?: string): PublicDocumentReference<T> {
-    // We allow omission of 'pathString' but explicitly prohibit passing in both
-    // 'undefined' and 'null'.
-    if (arguments.length === 0) {
-      pathString = AutoId.newId();
+  doc(documentPath?: string): DocumentReference<T> {
+    try {
+      if (documentPath !== undefined) {
+        return new DocumentReference(
```
wilhuff, line 1386: Why doesn't this case work if `documentPath` is undefined? I thought `doc` can't tell if you called it as `doc(this._delegate, undefined)` vs `doc(this._delegate)`.

packages/firestore/test/integration/api/validation.test.ts
```diff
@@ -847,8 +847,8 @@ apiDescribe('Validation:', (persistence: boolean) => {
         const reason =
           `Invalid query. You have a where filter with an ` +
           `inequality (<, <=, >, or >=) on field 'x' and so you must also ` +
-          `use 'x' as your first orderBy(), but your first orderBy() is on ` +
-          `field 'y' instead.`;
+          `use 'x' as your first argument to Query.orderBy(), but your first ` +
+          `orderBy() is on field 'y' instead.`;
```
wilhuff, line 851: It's somewhat surprising that we replace the first `orderBy` but not the second. In other cases where they're not spelled the same this seems like it would allow the implementation to leak out.

wilhuff (commented):
Basically LGTM

packages/firestore/src/api/database.ts
schmidt-sebastian: I believe we owe this to the genius of @mikelehen.

I added this comment:

```
        // Call `doc` without `documentPath` if `documentPath` is `undefined`
        // as `doc` validates the number of arguments to prevent users from
        // accidentally passing `undefined`.
```

packages/firestore/src/api/database.ts
schmidt-sebastian: We are moving away from manually crafted typings. I still need to iron out some final details in https://github.com/firebase/firebase-js-sdk/pull/3790, but once that is merged, we will generate typings from source. All other firebase-exp SDKs are doing the same.

packages/firestore/test/integration/api/validation.test.ts
schmidt-sebastian: JavaScript reg-exp only replace one instance by default. I could use `/g` here but then the message will be harder to parse (or has to be rewritten even more).

schmidt-sebastian (commented):
Feedback addressed (unfortunately as part of the merge commit: b5f4490, which merges the WriteBatch removal)

wilhuff:
LGTM

wilhuff (approved):
LGTM for real this time.

packages/firestore/src/api/database.ts
mikelehen: 😱

PR #4070, opened 2020-11-13
Compat class for Transaction

schmidt-sebastian:


packages/firestore/exp/src/api/transaction.ts
```diff
@@ -103,7 +103,7 @@ export function runTransaction<T>(
   firestore._queue.enqueueAndForget(async () => {
     const datastore = await getDatastore(firestore);
     new TransactionRunner<T>(
-      new AsyncQueue(),
+      firestore._queue,
```
schmidt-sebastian, line 106: Note: This is a bug in the firestore-exp code, which breaks a test. See how `firestoreClientTransaction` using the main AsyncQueue here.

wu-hui (approved):

PR #4078, opened 2020-11-17
Fix experimentalAutoDetectLongPolling and add logging

schmidt-sebastian:
This fixes a bug in experimentalAutoDetectLongPolling, as the WebChannel flag was not added to the closure externs and as such mangled. It also adds logging for each mode.

Verified with a test app in Chrome (albeit only for the "no buffering proxy" scenario).

### DIFF
--- a/packages/firestore/src/platform/browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform/browser/webchannel_connection.ts
@@ -22,7 +22,12 @@ import {
   WebChannel,
   WebChannelError,
   WebChannelOptions,
-  XhrIo
+  XhrIo,
+  getStatEventTarget,
+  EventTarget,
+  StatEvent,
+  Event,
+  Stat
 } from '@firebase/webchannel-wrapper';
 
 import {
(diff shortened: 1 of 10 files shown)


packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -348,6 +356,16 @@ export class WebChannelConnection extends RestConnection {
       }
     );
 
+    unguardedEventListen<StatEvent>(requestStats, Event.STAT_EVENT, event => {
+      if (!event) return;
```
rafikhan, line 360: Is this line necessary? If so, under what conditions would it happen. Might be worth adding a comment.

rafikhan (approved):
LGTM

packages/firestore/src/platform/browser/webchannel_connection.ts
schmidt-sebastian: It seems like `unguardedEventListen` can be modified safely to assume that there is always an event. Fixed.

Feiyang1 (approved):

PR #4081, opened 2020-11-17
Fix issue with auto-redirect signin in auth-exp

samhorlbeck:
`getRedirectResult()` was hanging previously if `config.authDomain` wasn't set. This PR makes `getRedirectResult()` error if `config.authDomain` isn't available, and also skips the eager redirect path entirely if authDomain isn't available at initialization


avolkovi (approved):

Feiyang1:
@hsubox76 FYI

Feiyang1 (approved):

PR #4084, opened 2020-11-17
Add nonresumable uploadBytes to modular Storage

hsubox76:
Adding `uploadBytes` to storage-exp. This is a new method that does a non-resumable upload which requires less code than the resumable upload. We expect people to use this by default and only use the resumable version when needed, so it gets the default-sounding name and the resumable upload is named `uploadBytesResumable`.

`uploadBytes` has to generate an `UploadTaskSnapshot` with less properties (no `task`, `state`, `bytesTransferred`...do we still need `totalBytes`? not sure) so I created a separate class, and again gave it the default sounding name, while the old one is named `UploadTaskResumableSnapshot`.


packages/storage/src/reference.ts
```diff
@@ -118,7 +120,36 @@ export class Reference {
...
+ * @public
+ * @param ref - Storage Reference where data should be uploaded.
+ * @param data - The data to upload.
+ * @param metadata - Metadata for the newly uploaded string.
+ * @returns An UploadTaskNonResumableSnapshot
+ */
+export async function uploadBytes(
+  ref: Reference,
+  data: Blob | Uint8Array | ArrayBuffer,
+  metadata: Metadata | null = null
```
Feiyang1, line 134: the signature should be `metadata?: Metadata` according to the API proposal. It applies to `uploadBytesResumable` and `uploadString` as well.

packages/storage/src/reference.ts
```diff
@@ -118,7 +120,36 @@ export class Reference {
...
+  const authToken = await ref.storage.getAuthToken();
+  const requestInfo = fbsRequests.simpleUpload(
+    ref.storage,
+    ref._location,
+    getMappings(),
+    new FbsBlob(data),
+    metadata
+  );
+  const request = ref.storage.makeRequest(requestInfo, authToken);
+  const finalMetadata = metadata || (await request.getPromise());
```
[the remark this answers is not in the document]
Feiyang1: Shouldn't we always use the metadata resolved from the promise? Will it be any different than the metadata passed in?

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+  });
+
+  it('can delete object ', async () => {
+    const reference = ref(storage, 'public/exp-delete');
+    await uploadString(reference, 'foo');
+    await getDownloadURL(reference);
+    await deleteObject(reference);
+    try {
+      await getDownloadURL(reference);
```
Feiyang1, line 107:
```suggestion
      expect(getDownloadURL(reference)).to.eventually.rejectedWith(Error, /Object 'public\/exp-delete' does not exist/ );
```

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('can list files', async () => {
+    const referenceA = ref(storage, 'public/exp-list/a');
+    const referenceB = ref(storage, 'public/exp-list/b');
+    const referenceCD = ref(storage, 'public/exp-list/c/d');
+    await uploadString(referenceA, '');
+    await uploadString(referenceB, '');
+    await uploadString(referenceCD, '');
+    const listResult = await listAll(await ref(storage, 'public/exp-list'));
```
Feiyang1, line 150:
```suggestion
    const listResult = await listAll(ref(storage, 'public/exp-list'));
```

packages/storage/src/tasksnapshot.ts
```diff
@@ -19,7 +19,7 @@ import { Metadata } from './metadata';
 import { Reference } from './reference';
 import { UploadTask } from './task';
 
-export class UploadTaskSnapshot {
+export class UploadTaskResumableSnapshot {
```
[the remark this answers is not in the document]
hsubox76: Think so, done.

packages/storage/src/reference.ts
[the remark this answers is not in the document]
hsubox76: I'm not really sure what I was thinking there, I think it should always be the returned metadata.

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/src/reference.ts
```diff
@@ -151,7 +182,7 @@ export function uploadString(
   format: StringFormat = StringFormat.RAW,
```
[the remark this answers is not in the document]
hsubox76: Ok, done, so now I moved the resumable implementation of this inside `putString()` in compat since it's for compat only.

packages/storage/src/implementation/requests.ts
```diff
@@ -279,6 +279,41 @@ export function metadataForUpload_(
...
+export function simpleUpload(
+  service: StorageService,
+  location: Location,
+  mappings: MetadataUtils.Mappings,
+  blob: FbsBlob,
+  metadata?: Metadata | null
+): RequestInfo<Metadata> {
+  const urlPart = location.bucketOnlyServerUrl();
+  const headers: { [prop: string]: string } = {
+    'Content-Type': 'application/octet-stream'
```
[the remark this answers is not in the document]
hsubox76: Normally (in the old resumable setup) the content type in the header doesn't actually match the metadata.contentType field because the metadata has a specific content type as seen in the logic above, and the request header Content-Type is always `multipart`. Since we're doing away with `multipart` I guess it would be best to make the header match the metadata.

packages/storage/src/implementation/metadata.ts
```diff
@@ -187,7 +187,7 @@ export function downloadUrlFromResourceString(
 }
 
 export function toResourceString(
-  metadata: Metadata,
+  metadata: { [key: string]: unknown },
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/storage/src/implementation/requests.ts
```diff
@@ -279,6 +279,41 @@ export function metadataForUpload_(
...
+  };
+
+  const metadata_ = metadataForUpload_(location, blob, metadata);
+  const urlParams: UrlParams = { name: metadata_['fullPath']! };
+  const url = UrlUtils.makeUrl(urlPart);
```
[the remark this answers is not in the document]
hsubox76: Did this in files touched by this PR but I would like to go through and do another PR for files not touched by this PR.

packages/storage/src/reference.ts
```diff
@@ -37,6 +37,8 @@ import { ListOptions, ListResult } from './list';
 import { UploadTask } from './task';
 import { invalidRootOperation, noDownloadURL } from './implementation/error';
 import { validateNumber } from './implementation/type';
+import { UploadTaskSnapshot } from './tasksnapshot';
+import * as fbsRequests from './implementation/requests';
```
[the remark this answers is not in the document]
hsubox76: Done, will sweep all storage files in a separate PR.

hsubox76:
Whoops, I misunderstood what `multipartUpload` does (it sends metadata alongside the data and is not related to resumableness), rewriting.

yarn.lock
```diff
@@ -13648,6 +13648,17 @@ rollup-plugin-terser@7.0.2:
     serialize-javascript "^4.0.0"
     terser "^5.0.0"
 
+rollup-plugin-typescript2@0.27.3:
```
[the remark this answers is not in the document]
hsubox76: Rebased.

packages/storage/src/implementation/requests.ts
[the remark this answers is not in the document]
schmidt-sebastian: FWIW, I have gone back on forth on this. Is this actually true in builds that are bundled? I would assume that only the import that make it into the final bundle matter, and most of these internal imports should be removed. This is just a guess though, and in general I am avoiding namespace imports.

packages/storage/exp/index.ts
```diff
@@ -44,6 +52,16 @@ export {
  */
 const STORAGE_TYPE = 'storage-exp';
 
+export function getStorage(app: FirebaseApp): StorageService {
+  // Dependencies
```
schmidt-sebastian, line 56: This comment confuses me more than it helps :)

packages/storage/src/reference.ts
```diff
@@ -118,7 +120,39 @@ export class Reference {
 }
 
 /**
- * Uploads a blob to this object's location.
+ * Uploads data to this object's location.
+ * The upload is not resumable.
+ * @public
+ * @param ref - Storage Reference where data should be uploaded.
```
schmidt-sebastian, line 126: "Storage reference" or "StorageReference" ?

packages/storage/src/reference.ts
```diff
@@ -118,7 +120,39 @@ export class Reference {
...
+ * @param data - The data to upload.
+ * @param metadata - Metadata for the newly uploaded string.
```
schmidt-sebastian, line 128: s/string/data

packages/storage/src/reference.ts
```diff
@@ -145,19 +180,32 @@ export function uploadBytesResumable(
  * @returns An UploadTask that lets you control and
  *     observe the upload.
  */
-export function uploadString(
+export async function uploadString(
```
schmidt-sebastian, line 183: We shouldn't make public API functions  `async` as it may alter their timing and exception behavior.

packages/storage/src/tasksnapshot.ts
```diff
@@ -19,13 +19,15 @@ import { Metadata } from './metadata';
...
-  ) {}
+export interface UploadTaskResumableSnapshot {
+  readonly bytesTransferred: number;
+  readonly totalBytes: number;
+  readonly state: TaskState;
+  readonly metadata: Metadata;
+  readonly task: UploadTask;
+  readonly ref: Reference;
+}
+export interface UploadTaskSnapshot {
```
schmidt-sebastian, line 30: Missing space. Should we also flip the declaration order?

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export const PROJECT_ID = PROJECT_CONFIG.projectId;
+export const STORAGE_BUCKET = PROJECT_CONFIG.storageBucket;
+export const API_KEY = PROJECT_CONFIG.apiKey;
+export const AUTH_DOMAIN = PROJECT_CONFIG.authDomain;
+
+describe('FirebaseStorage Exp', () => {
+  let app: FirebaseApp;
+  let storage: StorageService;
+  beforeEach(async () => {
```
schmidt-sebastian, line 56: Missing space

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+    await signInAnonymously(getAuth(app));
+    storage = getStorage(app);
+  });
+  afterEach(async () => {
+    await deleteApp(app);
+  });
+  it('can upload bytes', async () => {
+    const reference = ref(storage, 'public/exp-bytes');
+    const snap = await uploadBytes(reference, new Uint8Array([0, 1, 3]));
```
schmidt-sebastian, line 71: Missing space

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    expect(snap.metadata.timeCreated).to.exist;
+  });
+  it('can upload bytes (resumable)', async () => {
```
schmidt-sebastian, line 74: Missing space

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,160 @@
...
+      /Object 'public\/exp-delete' does not exist/
+    );
+    // try {
+    //   await getDownloadURL(reference);
+    //   expect.fail();
+    // } catch (e) {
+    //   expect(e.message).to.satisfy((v: string) =>
+    //     v.match(/Object 'public\/exp-delete' does not exist/)
+    //   );
+    // }
```
schmidt-sebastian, line 119: Please remove.

schmidt-sebastian (commented):
Looks pretty good (but somewhat hard to review due to the import/formatting changes). I wonder if you have been able to compute the size difference?

packages/storage/exp/index.ts
hsubox76: I think that's a copy/paste artifact! Removed.

packages/storage/src/reference.ts
hsubox76: Hmm, exported type name is just `Reference`. I guess I should rename it since it's publicly exposed. Compat stays the same since it's namespaced and to remain compat.

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/src/reference.ts
hsubox76: Removed (shared code between uploadString and uploadBytes should be moved out to a shared function anyway I think).

packages/storage/src/tasksnapshot.ts
hsubox76: Done, makes sense.

packages/storage/test/integration/integration.exp.test.ts
hsubox76: Done.

packages/storage/src/implementation/requests.ts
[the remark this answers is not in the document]
Feiyang1: Rollup will remove unused functions even when using namespace imports. So to your point, it doesn't matter whether or not namespace imports are used in bundled builds. I still think we should use namespace imports as the best practice because they are likely to work well with any build tool we might use. 

index.ts
```
import * as impl from './impl';

console.log(impl.foo());
```

impl.ts
```
export function foo() {}

// removed in the bundle
export function bar() {}
```

bundle.js produced by rollup
```
function foo() {
}

// import * as firebase from 'firebase/app';
console.log(foo());

```

packages/storage/compat/reference.ts
```diff
@@ -114,8 +115,17 @@ export class ReferenceCompat implements types.Reference {
     metadata?: Metadata
   ): types.UploadTask {
     this._throwIfRoot('putString');
+    const data = dataFromString(format, value);
+    const metadataClone = { ...metadata } as Metadata;
+    if (metadataClone['contentType'] == null && data.contentType != null) {
+      metadataClone['contentType'] = data.contentType!;
```
schmidt-sebastian, line 121: Is the bang needed here? Doesn't the `!=` check catch both `null` and `undefined`?

packages/storage/package.json
```diff
@@ -19,8 +19,12 @@
...
     "test": "run-p test:browser lint",
     "test:ci": "node ../../scripts/run_tests_in_ci.js -s test:browser",
-    "test:browser:unit": "karma start --single-run --unit",
-    "test:browser:integration": "karma start --single-run --integration",
+    "test:browser:compat:unit": "karma start --single-run --compat --unit",
+    "test:browser:exp:unit": "karma start --single-run --exp --unit",
+    "test:browser:compat:integration": "karma start --single-run --compat --integration",
+    "test:browser:exp:integration": "karma start --single-run --exp --integration",
+    "test:browser:compat": "karma start --single-run --compat",
+    "test:browser:exp": "karma start --single-run --exp",
```
schmidt-sebastian, line 27: Should we add these new test targets to the CI rule?

packages/storage/src/reference.ts
```diff
@@ -118,46 +123,90 @@ export class Reference {
...
- * Uploads a blob to this object's location.
+ * Uploads data to this object's location.
+ * The upload is not resumable.
  * @public
- * @param ref - Storage Reference where data should be uploaded.
+ * @param ref - StorageReference where data should be uploaded.
  * @param data - The data to upload.
- * @param metadata - Metadata for the newly uploaded string.
+ * @param metadata - Metadata for the newly uploaded data.
+ * @returns An UploadTaskNonResumableSnapshot
```
schmidt-sebastian, line 132:
```suggestion
 * @returns An UploadTaskSnapshot
```

packages/storage/src/reference.ts
```diff
@@ -118,46 +123,90 @@ export class Reference {
...
  * Uploads a string to this object's location.
+ * The upload is not resumable.
  * @public
- * @param ref - Storage Reference where string should be uploaded.
+ * @param ref - StorageReference where string should be uploaded.
  * @param value - The string to upload.
  * @param format - The format of the string to upload.
- * @param metadata - Metadata for the newly uploaded object.
+ * @param metadata - Metadata for the newly uploaded string.
  * @returns An UploadTask that lets you control and
```
schmidt-sebastian, line 170:
```suggestion
 * @returns UploadTaskSnapshot
```

packages/storage/src/reference.ts
```diff
@@ -118,46 +123,90 @@ export class Reference {
...
   const metadataClone = { ...metadata } as Metadata;
   if (metadataClone['contentType'] == null && data.contentType != null) {
     metadataClone['contentType'] = data.contentType!;
   }
-  return new UploadTask(ref, new FbsBlob(data.data, true), metadataClone);
+  return _nonResumableUpload(ref, data.data, metadataClone);
+}
+
+/**
+ * Base code for nonresumable upload, used by uploadString and uploadBytes.
```
schmidt-sebastian, line 189:
```suggestion
 * Shared code for nonresumable upload, used by uploadString and uploadBytes.
```

Optional.

packages/storage/src/reference.ts
```diff
@@ -289,8 +338,8 @@ export async function getMetadata(ref: Reference): Promise<Metadata> {
  *     See `firebaseStorage.Reference.prototype.getMetadata`
  */
 export async function updateMetadata(
-  ref: Reference,
-  metadata: Metadata
+  ref: StorageReference,
+  metadata: Record<string, unknown>
```
schmidt-sebastian, line 342: Isn't Metadata more than just a `Record<string, unknown>` since some keys are pre-defined? It seems like changing this to a Record loses some information, even if it doesn't make a huge difference as an input type.

packages/storage/src/tasksnapshot.ts
```diff
@@ -16,16 +16,19 @@
...
+  readonly ref: StorageReference;
+}
+
+export interface UploadTaskResumableSnapshot {
+  readonly bytesTransferred: number;
+  readonly totalBytes: number;
+  readonly state: TaskState;
+  readonly metadata: Metadata;
+  readonly task: UploadTask;
+  readonly ref: StorageReference;
```
schmidt-sebastian, line 33: I wonder if these are the best names. Have you consider `UploadTaskSnapshot` and `UploadResult` or similar? The reason I ask is:
- `UploadTaskSnapshot` isn't really a snapshot anymore, it's the end state. Plus, the new upload format doesn't actually involve UploadTasks anymore,
- `UploadTaskResumableSnapshot` doesn't offer any APIs or method that let you pause or resume an Upload. Instead, these methods live on UploadTask and don't depend on this snapshot.

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  uploadString,
+  deleteObject,
+  getMetadata,
+  updateMetadata,
+  listAll
+} from '../../exp/index';
+
+// See https://github.com/typescript-eslint/typescript-eslint/issues/363
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+// import * as storage from '@firebase/storage-types';
```
schmidt-sebastian, line 37: Please remove

packages/storage/test/integration/integration.exp.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+import { use, expect } from 'chai';
+import * as chaiAsPromised from 'chai-as-promised';
+use(chaiAsPromised);
+import { FirebaseApp } from '@firebase/app-types-exp';
+import { StorageService } from '../../src/service';
```
schmidt-sebastian, line 43:
```suggestion
import { use, expect } from 'chai';
import * as chaiAsPromised from 'chai-as-promised';
import { FirebaseApp } from '@firebase/app-types-exp';
import { StorageService } from '../../src/service';

use(chaiAsPromised);
```

packages/storage/src/reference.ts
```diff
@@ -118,46 +123,90 @@ export class Reference {
...
+  data: Blob | Uint8Array | ArrayBuffer,
+  metadata?: Metadata
+): Promise<UploadTaskSnapshot> {
+  ref._throwIfRoot('uploadBytes');
+  return _nonResumableUpload(ref, data, metadata);
+}
+
+/**
+ * Uploads data to this object's location.
+ * The upload is resumable and exposes progress updates.
```
schmidt-sebastian, line 145:
```suggestion
 * The upload can be paused, resumed and exposes progress updates.
```

Suggesting this because the Web SDK doesn't actually allow you to resume an upload with an Upload URL (as far as I can tell). Instead, the user can only resume uploads with an UploadTask that they paused.

Android let's users actually resume an upload with an Upload URL: https://firebase.google.com/docs/reference/android/com/google/firebase/storage/StorageReference#putFile(android.net.Uri,%20com.google.firebase.storage.StorageMetadata,%20android.net.Uri)

packages/storage/src/task.ts
```diff
@@ -36,19 +36,26 @@ import {
...
 import { async as fbsAsync } from './implementation/async';
-import * as fbsMetadata from './implementation/metadata';
-import * as fbsRequests from './implementation/requests';
-import { Reference } from './reference';
-import { getMappings } from './implementation/metadata';
+import { Mappings, getMappings } from './implementation/metadata';
+import {
+  createResumableUpload,
+  getResumableUploadStatus,
+  resumableUploadChunkSize,
```
schmidt-sebastian, line 45:
```suggestion
  RESUMABLE_UPLOAD_CHUNK_SIZE,
```

packages/storage/test/unit/reference.exp.test.ts
```diff
@@ -277,11 +331,16 @@ describe('Firebase Storage > Reference', () => {
...
+    it('uploadString throws', async () => {
       expect(() => uploadString(root, 'raw', StringFormat.RAW)).to.throw(
         'storage/invalid-root-operation'
       );
     });
+    it('uploadBytes throws', async () => {
+      expect(() => uploadBytes(root, new Blob(['a']))).to.throw(
+        'storage/invalid-root-operation'
+      );
+    });
```
schmidt-sebastian, line 343:
```suggestion
    it('uploadString throws', () => {
      expect(() => uploadString(root, 'raw', StringFormat.RAW)).to.throw(
        'storage/invalid-root-operation'
      );
    });

    it('uploadBytes throws', () => {
      expect(() => uploadBytes(root, new Blob(['a']))).to.throw(
        'storage/invalid-root-operation'
      );
    });
```

packages/storage/compat/reference.ts
```diff
@@ -114,8 +115,17 @@ export class ReferenceCompat implements types.Reference {
     metadata?: Metadata
   ): types.UploadTask {
     this._throwIfRoot('putString');
+    const data = dataFromString(format, value);
```
schmidt-sebastian, line 118: Just curious - could this method just call `put` with `new FbsBlob(data.data, true)`?

packages/storage/src/implementation/requests.ts
```diff
@@ -279,10 +277,13 @@ export function metadataForUpload_(
   return metadataClone;
 }
 
+/**
+ * Prepare RequestInfo for resumable uploads as Content-Type: multipart.
```
Feiyang1, line 281:
```suggestion
 * Prepare RequestInfo for uploads as Content-Type: multipart.
```
It is used for both resumable and non-resumable uploads.

packages/storage/src/service.ts
```diff
@@ -151,7 +151,11 @@ export class StorageService implements _FirebaseService {
     /**
      * @internal
      */
-    readonly _url?: string
+    readonly _url?: string,
+    /**
+     * @internal
+     */
+    readonly _firebaseVersion?: string
```
Feiyang1, line 158: Should it be required?

packages/storage/src/tasksnapshot.ts
Feiyang1: +1

res
```diff
@@ -0,0 +1,998 @@
+{
```
Feiyang1, line 1: delete

scripts/ci-test/build_changed.ts
```diff
@@ -60,9 +59,13 @@ if (!allTestConfigNames.includes(inputTestConfigName)) {
...
 const config = testConfig[inputTestConfigName]!;
 
-buildForTests(config, buildAppExp);
+buildForTests(config, buildAppExp, buildAppCompat);
 
-async function buildForTests(config: TestConfig, buildAppExp = false) {
+async function buildForTests(
+  config: TestConfig,
+  buildAppExp = false,
+  buildAppCompat = false
```
Feiyang1, line 67: Was the script broken because of the missing argument until now? but the workflow seems to work fine?

packages/storage/package.json
hsubox76: These were for development testing convenience, CI tests everything.

packages/storage/src/reference.ts
hsubox76: Metadata requires all its fields, the param that can be supplied here only needs to have some of the fields, or even just 1. Should it be named something else? `metadataUpdates`?

packages/storage/src/service.ts
hsubox76: Easier to leave the logic intact in `addVersionHeader_` in `request.ts` that replaces it with "AppManager" (or some other default string if that's obsolete) if it is undefined, and not have to change the argument order in StorageService constructor (it would have to come before _url and all the calls to StorageService() have to be rewritten). Just trying not to change too much code at once.

scripts/ci-test/build_changed.ts
hsubox76: It used to be a global variable so it worked but maybe in an unexpected way.

packages/storage/compat/reference.ts
hsubox76: Removed, seems like it was left over from back when it was using `type.isDef` instead of `!= null`.

packages/storage/src/tasksnapshot.ts
hsubox76: Sounds like a good plan.

packages/storage/src/reference.ts
hsubox76: Changed.

packages/storage/src/task.ts
hsubox76: Done.

packages/storage/test/unit/reference.exp.test.ts
hsubox76: Done.

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/src/reference.ts
hsubox76: Done.

packages/storage/test/integration/integration.exp.test.ts
hsubox76: Done.

packages/storage/test/integration/integration.exp.test.ts
hsubox76: Done.

packages/storage/compat/reference.ts
hsubox76: You mean call `uploadBytes` instead of having a shared _nonResumableUpload function? I guess the only drawback would be an extra _throwIfRoot call that won't do anything.

common/api-review/storage.api.md
```diff
@@ -0,0 +1,68 @@
...
+// @public
+export function getMetadata(ref: StorageReference): Promise<Metadata>;
+
+// Warning: (ae-forgotten-export) The symbol "StorageService" needs to be exported by the entry point index.d.ts
+//
+// @public
+export function getStorage(app: FirebaseApp): StorageService;
+
+// Warning: (ae-forgotten-export) The symbol "ListOptions" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "ListResult" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
hsubox76: Partially addressed. There are a lot of files to fix so I will do a full pass on all files in another PR.

packages/storage/src/implementation/requests.ts
hsubox76: Thanks, fixed.

res
hsubox76: Oops

packages/storage/compat/reference.ts
schmidt-sebastian: Yes

packages/storage/package.json
schmidt-sebastian: Please leave them. It makes everyone's life easier.

packages/storage/src/reference.ts
```diff
@@ -118,46 +124,88 @@ export class Reference {
...
+}
+
+/**
+ * Uploads data to this object's location.
+ * The upload can be paused and resumed, and exposes progress updates.
+ * @public
+ * @param ref - StorageReference where data should be uploaded.
+ * @param data - The data to upload.
+ * @param metadata - Metadata for the newly uploaded data.
+ * @returns An UploadTaskSnapshot
```
schmidt-sebastian, line 151:
```suggestion
 * @returns An UploadTask
```

packages/storage/src/reference.ts
schmidt-sebastian: What about `Partial<Metadata>`? I think that makes the intent very clear.

schmidt-sebastian (approved):
LGTM from my end (please wait for Fei's LGTM as well). Consider calling `uploadBytes` from `uploadString`.

packages/storage/test/integration/integration.exp.test.ts
hsubox76: Done.

packages/storage/test/integration/integration.exp.test.ts
hsubox76: Done.

packages/storage/src/reference.ts
hsubox76: Oh, perfect, done.

packages/storage/src/reference.ts
hsubox76: Done.

Feiyang1 (approved):

PR #4093, opened 2020-11-19
Fix assertFails not recognising database permission denied error

mikkelvp:
```assertFails``` does not recognise the permission denied error from database ```once('value')```

example
```
const app = firebase.initializeTestApp({
  databaseName,
  auth: { uid: 'alice' },
})
app
  .database()
  .ref('secrets/bob/id')
  .once('value')
  .then((snapshot) => {
    console.log('read succeeded')
  })
  .catch((err) => {
    console.log('read failed')
    // current implementation of isPermissionDenied 
    // https://github.com/firebase/firebase-js-sdk/blob/d934f74d6750dc31cca480e126757d53935cefcd/packages/rules-unit-testing/src/api/index.ts#L457
    const isPermissionDenied =
      (err && err.message && err.message.indexOf('PERMISSION_DENIED') >= 0) ||
      (err && err.code === 'permission-denied')
    console.log(`isPermissionDenied ${isPermissionDenied}`)
    console.log(JSON.stringify({ message: err.message, ...err }, null, 2))
  })
```
output:
```
read failed
isPermissionDenied false
{
  "message": "permission_denied at /secrets/bob/id: Client doesn't have permission to access the desired data.",
  "code": "PERMISSION_DENIED"
}
```

### DIFF
--- a/.changeset/early-dots-peel.md
+++ b/.changeset/early-dots-peel.md
@@ -1,0 +1,5 @@
+---
+"@firebase/rules-unit-testing": patch
+---
+
+Fix assertFails not recognising database permission denied error
--- a/packages/rules-unit-testing/src/api/index.ts
+++ b/packages/rules-unit-testing/src/api/index.ts
@@ -454,9 +454,14 @@ export function assertFails(pr: Promise<any>): any {
       );
     },
     (err: any) => {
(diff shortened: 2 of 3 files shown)


yuchenshi (changes requested):
Thanks for the PR! Would you please also add a test for the new case?

yuchenshi (approved):

PR #4096, opened 2020-11-19
Add type definitions for get()

IanWyszynski:


packages/firebase/index.d.ts
```diff
@@ -6171,6 +6171,16 @@ declare namespace firebase.database {
       context?: Object | null
     ): void;
 
+    /**
+     * Gets the most up-to-date result for this query.
+     *
+     * @return A promse which resolves to the resulting DataSnapshot if
```
Feiyang1, line 6177:
```suggestion
     * @return A promise which resolves to the resulting DataSnapshot if
```

.changeset/many-snails-kneel.md
```diff
@@ -0,0 +1,6 @@
+---
+"@firebase/database": patch
```
Feiyang1, line 2:
```suggestion

```

Feiyang1: Since there is no change in `@firebase/database`, we don't need to bump its version

Feiyang1 (approved):

PR #4105, opened 2020-11-20
API Extractor Doc fixes

schmidt-sebastian:
API Extractor will be used to generate the reference documentation for firestore-exp. It currently throws some validation errors, and this fixes them by doing the following:

- Use `@returns` instead of `@return`
- Add "-" for `@param`
- HTML escape special characters
- Fix `@link` tags
- Add `@hideconstructor` for non-public (internal) constructors

I don't think this needs a thorough review, but a spot check would be nice.

### DIFF
--- a/packages/firestore/exp/src/api/snapshot.ts
+++ b/packages/firestore/exp/src/api/snapshot.ts
@@ -83,7 +83,7 @@ import { newQueryComparator } from '../../../src/core/query';
 export interface FirestoreDataConverter<T> {
   /**
    * Called by the Firestore SDK to convert a custom model object of type `T`
-   * into a plain Javascript object (suitable for writing directly to the
+   * into a plain JavaScript object (suitable for writing directly to the
    * Firestore database). To use `set()` with `merge` and `mergeFields`,
    * `toFirestore()` must be defined with `Partial<T>`.
    */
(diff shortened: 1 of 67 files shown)


packages/firestore/src/local/index_manager.ts
```diff
@@ -22,7 +22,7 @@ import { PersistencePromise } from './persistence_promise';
 /**
  * Represents a set of indexes that are used to execute queries efficiently.
  *
- * Currently the only index is a [collection id] => [parent path] index, used
+ * Currently the only index is a [collection id] =&gt; [parent path] index, used
```
Feiyang1, line 25: Why do we need to escape it? Does it not work in markdown?

dconeybe (approved):
LGTM

packages/firestore/src/local/index_manager.ts
schmidt-sebastian: API Extractor complains about all these HTML-like characters (at least with default settings).

packages/firestore/exp/src/api/snapshot.ts
```diff
@@ -88,14 +88,21 @@ export interface FirestoreDataConverter<T> {
    * `toFirestore()` must be defined with `Partial<T>`.
    */
   toFirestore(modelObject: T): DocumentData;
+
+  /**
+   * Called by the Firestore SDK to convert a custom model object of type `T`
+   * into a plain Javascript object (suitable for writing directly to the
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

PR #4107, opened 2020-11-21
Whee! Auto-remove unused imports

schmidt-sebastian:
This works with `yarn lint:fix`.


Feiyang1 (approved):
Do you mind moving it to the shared eslint config?

PR #4116, opened 2020-11-24
Create modularized Analytics package.

hsubox76:
Created "analytics-exp" modularized analytics package.

One question: `logEvent` has a lot of signature overloads in the public interface (convenient recommended param autocompletes for all the standard eventNames). I haven't added them here yet (they can be seen in the old packages/firebase/index.d.ts monolith). Is there a good place for me to put them that doesn't clutter up the code?


packages-exp/analytics-exp/src/helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+  const correspondingAppId = measurementIdToAppId[measurementId as string];
+  try {
+    if (correspondingAppId) {
+      await initializationPromisesMap[correspondingAppId];
+    } else {
+      // If config is not fetched yet, wait for all configs (we don't know which one we need) and
+      // find the appId (if any) corresponding to this measurementId. If there is one, wait on
+      // that appId's initialization promise. If there is none, promise resolves and gtag
+      // call goes through.
+      const dynamicConfigResults = await Promise.all(dynamicConfigPromisesList);
```
Feiyang1, line 89: We should use `Promise.allSettled()` instead of `Promise.all()`, because we should proceed to try to find a match even if some promises rejected.

`Promise.allSettled()` is a pretty new addition to the es spec, so we need to use a polyfill. We can create one ourselves:

```
const a = Promise.resolve(1);
const b = Promise.reject(new Error(2));
const c = Promise.resolve(3);

// allSettled polyfill code
Promise.all([a, b, c].map(p => p.catch(e => e)))
  .then(results => console.log(results)) // 1,Error: 2,3
```

packages-exp/analytics-exp/src/helpers.ts
```diff
@@ -0,0 +1,308 @@
...
+    // an initializeIds() promise we are waiting for.
+    if (gtagParams && gtagParams['send_to']) {
+      let gaSendToList: string | string[] = gtagParams['send_to'];
+      // Make it an array if is isn't, so it can be dealt with the same way.
+      if (!Array.isArray(gaSendToList)) {
+        gaSendToList = [gaSendToList];
+      }
+      // Checking 'send_to' fields requires having all measurement ID results back from
+      // the dynamic config fetch.
+      const dynamicConfigResults = await Promise.all(dynamicConfigPromisesList);
```
Feiyang1, line 134: Use Promise.allSettled()

packages-exp/analytics-exp/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+
+function registerAnalytics(): void {
+  _registerComponent(
+    new Component(
+      ANALYTICS_TYPE,
+      container => {
+        // getImmediate for FirebaseApp will always succeed
+        const app = container.getProvider('app-exp').getImmediate();
+        const installations = container
+          .getProvider('installations-exp')
```
Feiyang1, line 48:
```suggestion
          .getProvider('installations-exp-internal')
```

packages-exp/analytics-exp/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+          .getImmediate();
+
+        return factory(app, installations);
+      },
+      ComponentType.PUBLIC
+    ).setServiceProps({
```
Feiyang1, line 54: no need to set service props anymore.

packages-exp/analytics-exp/src/initialize-ids.ts
```diff
@@ -0,0 +1,137 @@
...
+ * part of the `page_view` event that's sent during the initialization
+ * @param app Firebase app
+ * @param gtagCore The gtag function that's not wrapped.
+ * @param dynamicConfigPromisesList Array of all dynamic config promises.
+ * @param measurementIdToAppId Maps measurementID to appID.
+ * @param installations FirebaseInstallations instance.
+ *
+ * @returns Measurement ID.
+ */
+export async function initializeIds(
```
Feiyang1, line 71: Can we give the function a more descriptive name, e.g. describing what specific things are being initialized, or maybe give it a more high level name, e.g. initializeAnalytics()?

packages-exp/analytics-exp/src/factory.ts
```diff
@@ -0,0 +1,248 @@
...
+  Analytics
+} from '@firebase/analytics-types-exp';
+import {
+  insertScriptTag,
+  getOrCreateDataLayer,
+  wrapOrCreateGtag,
+  findGtagScriptOnPage
+} from './helpers';
+import { AnalyticsError, ERROR_FACTORY } from './errors';
+import { FirebaseInstallations } from '@firebase/installations-types-exp';
```
Feiyang1, line 32:
```suggestion
import { _FirebaseInstallationsInternal } from '@firebase/installations-types-exp';
```

packages-exp/analytics-exp/src/initialize-ids.ts
```diff
@@ -0,0 +1,137 @@
...
+  app: FirebaseApp,
+  dynamicConfigPromisesList: Array<
+    Promise<DynamicConfig | MinimalDynamicConfig>
+  >,
+  measurementIdToAppId: { [key: string]: string },
+  installations: FirebaseInstallations,
```
Feiyang1, line 77:
```suggestion
  installations: _FirebaseInstallationsInternal,
```

packages-exp/analytics-exp/src/initialize-ids.ts
```diff
@@ -0,0 +1,137 @@
...
+      }
+    })
+    .catch(e => logger.error(e));
+  // Add to list to track state of all dynamic config promises.
+  dynamicConfigPromisesList.push(dynamicConfigPromise);
+
+  const fidPromise: Promise<string | undefined> = validateIndexedDB().then(
+    envIsValid => {
+      if (envIsValid) {
+        return getId(installations);
```
Feiyang1, line 105: The `installations` here is an internal installations instance where you can do `installations.getId()`. We never import/use functions from other Firebase SDKs directly. We rely entirely on the component framework for inter-SDK communication.

packages-exp/analytics-exp/src/initialize-ids.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+
+import {
+  DynamicConfig,
+  Gtag,
+  MinimalDynamicConfig
+} from '@firebase/analytics-types-exp';
+import { GtagCommand, GA_FID_KEY, ORIGIN_KEY } from './constants';
+import { FirebaseInstallations } from '@firebase/installations-types-exp';
+import { getId } from '@firebase/installations-exp';
```
Feiyang1, line 25:
```suggestion

```

packages-exp/analytics-exp/package.json
```diff
@@ -0,0 +1,65 @@
...
+  },
+  "repository": {
+    "directory": "packages/analytics",
+    "type": "git",
+    "url": "https://github.com/firebase/firebase-js-sdk.git"
+  },
+  "bugs": {
+    "url": "https://github.com/firebase/firebase-js-sdk/issues"
+  },
+  "typings": "dist/index.d.ts",
```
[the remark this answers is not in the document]
hsubox76: Fixed.

packages-exp/analytics-exp/src/factory.ts
```diff
@@ -0,0 +1,248 @@
...
+ * this page independently of Firebase Analytics, and is using non-default
+ * names for either the `gtag` function or for `dataLayer`.
+ * Must be called before calling `firebase.analytics()` or it won't
+ * have any effect.
+ *
+ * @public
+ *
+ * @param options - Custom gtag and dataLayer names.
+ */
+export function settings(options: SettingsOptions): void {
```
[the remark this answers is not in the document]
hsubox76: I think I neglected to include it in the API proposal. I changed it to `analyticsSettings`. Should I amend the API proposal?

packages-exp/analytics-exp/src/index.ts
hsubox76: Oops, removed.

packages-exp/analytics-exp/src/constants.ts
```diff
@@ -0,0 +1,74 @@
...
+  EVENT = 'event',
+  SET = 'set',
+  CONFIG = 'config'
+}
+
+/*
+ * Officially recommended event names for gtag.js
+ * Any other string is also allowed.
+ */
+export enum EventName {
```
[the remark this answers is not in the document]
hsubox76: Changed.

packages-exp/analytics-exp/src/factory.ts
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * Name for window global gtag function used by GA: defaults to 'gtag'.
+ */
+let gtagName: string = 'gtag';
+
+/**
+ * Reproduction of standard gtag function or reference to existing
+ * gtag function on window object.
+ */
+let gtagCoreFunction: Gtag;
```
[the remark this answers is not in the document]
hsubox76: It is created once in the `if (!globalInitDone)` block and passed to any subsequent `initializeIds` calls for the first and any future instances, we want to store the reference to the one made in the initial wrapOrCreateGtag call and make sure we only make that call once.

packages-exp/analytics-exp/src/index.ts
hsubox76: Done.

packages-exp/analytics-exp/src/initialize-ids.ts
hsubox76: Done.

packages-exp/analytics-exp/src/factory.ts
hsubox76: Done.

packages-exp/analytics-exp/src/initialize-ids.ts
hsubox76: Ah, I was confused about how it works.

packages-exp/analytics-exp/src/initialize-ids.ts
hsubox76: Done.

packages-exp/analytics-exp/src/initialize-ids.ts
hsubox76: `initializeAnalytics` seems good.

packages-exp/analytics-exp/src/helpers.ts
hsubox76: I made a polyfill function. It adds an extra tick to the async stack so this same function has to be used in tests (see helpers.test.ts).

packages-exp/analytics-exp/src/helpers.ts
hsubox76: Done.

packages-exp/analytics-exp/src/factory.ts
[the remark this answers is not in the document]
Feiyang1: I actually prefer `settings` to `analyticsSettings` since everything is already namespaced by the package name by default. I just thought a name with "verb + thing" pattern like `updateSettings()` feels more natural for a free function.

Yes, we should amend the API proposal. Let's start with `settings()`. We should be able to do everything via emails.

packages-exp/analytics-exp/src/initialize-ids.ts
```diff
@@ -0,0 +1,136 @@
+/**
```
Feiyang1, line 1: Should the file name also be changed?

Feiyang1 (approved):
LGTM except for a few minor comments

packages-exp/analytics-exp/src/constants.ts
```diff
@@ -0,0 +1,74 @@
...
+  EVENT = 'event',
+  SET = 'set',
+  CONFIG = 'config'
+}
+
+/*
+ * Officially recommended event names for gtag.js
+ * Any other string is also allowed.
+ */
+export const enum EventName {
```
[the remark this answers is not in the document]
hsubox76: I think we wanted to make the enum available to users so they could easily get a list of official EventNames that they can then put into `logEvent`. The string literals are already available as autocompletes because of the overloads but maybe they want to save it to a variable? Feel safer from typos?

It wasn't exported so I added it to exports from `api.ts`.

Basically, if you import the `EventName` enum you can do:

`logEvent(analytics, EventName.LOGIN, { method: 'email' });`

or

`logEvent(analytics, 'login', { method: 'email' });`

packages-exp/analytics-exp/src/initialize-ids.ts
hsubox76: Changed name of file and corresponding test file.

packages-exp/analytics-exp/src/constants.ts
```diff
@@ -0,0 +1,74 @@
...
+export const ORIGIN_KEY = 'origin';
+
+export const FETCH_TIMEOUT_MILLIS = 60 * 1000;
+
+export const DYNAMIC_CONFIG_URL =
+  'https://firebase.googleapis.com/v1alpha/projects/-/apps/{app-id}/webConfig';
+
+export const GTAG_URL = 'https://www.googletagmanager.com/gtag/js';
+
+export enum GtagCommand {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages-exp/analytics-exp/src/constants.ts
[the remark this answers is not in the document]
Feiyang1: If we want to export it, it has to be a normal enum instead of const enum because const enum only exists in compile time. Is it part of the API proposal? I'm okay exporting it since it's useful, but I don't think enums are tree-shakable, so it will be included in bundle even if not used.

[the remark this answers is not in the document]
hsubox76: I think we should keep it especially as it's available in the current API. I'll add it to the API proposal and send an email.

PR #4117, opened 2020-11-24
Remove SimpleQueryEngine

schmidt-sebastian:
This PR addresses a long outstanding TODO and removes `SimpleQueryEngine`, which existed to make IndexFreeQueryEngine optional during development. This PR makes IndexFreeQueryEngine the only QueryEngine.


schmidt-sebastian:
Note: The diff here is kinda bad, but I renamed index_free_query_engine.ts to query_engine.ts and edited some comments. The diff is clearer in IntelliJ/Git directly.

packages/firestore/src/local/query_engine.ts
```diff
@@ -15,26 +15,202 @@
...
+ * the same results as full collection scans. In these cases, query
+ * processing falls back to full scans. These cases are:
+ *
+ * - Limit queries where a document that matched the query previously no longer
+ *   matches the query.
+ *
+ * - Limit queries where a document edit may cause the document to sort below
+ *   another document that is in the local cache.
+ *
+ * - Queries that have never been CURRENT or free of Limbo documents.
```
dconeybe, line 58: ultranit: I think that the "L" in "Limbo" can be lowercase, since "limbo" is not a proper noun.

packages/firestore/test/unit/local/query_engine.test.ts
```diff
@@ -67,25 +66,27 @@ const MISSING_LAST_LIMBO_FREE_SNAPSHOT = SnapshotVersion.min();
  * `getDocumentsMatchingQuery()` to detect index-free execution.
  */
 class TestLocalDocumentsView extends LocalDocumentsView {
-  expectIndexFreeExecution: boolean | undefined;
+  exceptFullCollectionScan: boolean | undefined;
```
dconeybe, line 69: Typo: "exceptFullCollectionScan" should be "expectFullCollectionScan".

dconeybe (approved):
LGTM, with two ultra-minor nits.

packages/firestore/src/local/query_engine.ts
schmidt-sebastian: Fixed. I do consider "Limbo" more of a concept, rather than just the state that it refers to. Fixed either way though.

packages/firestore/test/unit/local/query_engine.test.ts
schmidt-sebastian: Fixed

PR #4143, opened 2020-12-01
Separate memory build further.

wu-hui:


packages/firebase/firestore/memory/index.ts
```diff
@@ -21,4 +21,4 @@
  */
 
 import '@firebase/firestore/memory';
-import '@firebase/firestore/bundle';
+import '@firebase/firestore/memory-bundle';
```
schmidt-sebastian, line 24: Can this be `@firebase/firestore/memory/bundle`?

wu-hui: It cannot be done gracefully. 

The reason is relative paths in `packages/firestore/memory-bundle/package.json` are interpreted in different context (in rollup or in Typescript import from packages/firebase/firestore/...) and therefore some of them need to be prefixed `../..` and some of them just `../`, which is confusing.

wu-hui:
I will merge this into `wuandy/RollBundlesAsPatchedModule` and let's review that PR.

PR #4147, opened 2020-12-01
Fix docs formatting issues.

hsubox76:
Some `@link` tags were incorrectly formatted or the generated html file had the wrong case, leading to broken links in generated docs.

Sort of hard to avoid in Node docs, because comments sometimes reference some pages that only exist in (client) JS docs. Decided to remove the markup and leave the words as plain text in this case.

Also added 2 missing TOC entries from Storage.

### DIFF
--- a/scripts/docgen/content-sources/js/toc.yaml
+++ b/scripts/docgen/content-sources/js/toc.yaml
@@ -67,31 +67,31 @@ toc:
   - title: "IdTokenResult"
     path: /docs/reference/js/firebase.auth.IDTokenResult
   - title: "MultiFactorAssertion"
-    path: /docs/reference/js/firebase.auth.multifactorassertion
+    path: /docs/reference/js/firebase.auth.MultiFactorAssertion
   - title: "MultiFactorError"
-    path: /docs/reference/js/firebase.auth.multifactorerror
+    path: /docs/reference/js/firebase.auth.MultiFactorError
   - title: "MultiFactorInfo"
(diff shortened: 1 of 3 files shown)


scripts/docgen/content-sources/js/toc.yaml
```diff
@@ -67,31 +67,31 @@ toc:
   - title: "IdTokenResult"
     path: /docs/reference/js/firebase.auth.IDTokenResult
   - title: "MultiFactorAssertion"
-    path: /docs/reference/js/firebase.auth.multifactorassertion
+    path: /docs/reference/js/firebase.auth.MultiFactorAssertion
```
egilmorez, line 70: I think these changes might actually break the links, won't they?

The current lowercase is working; but when I manually change a URL to camel case, I get 404s.

egilmorez (commented):
Thanks Christina! One question for you.

scripts/docgen/content-sources/js/toc.yaml
hsubox76: This will change the case of the generated filename, when the docs are published, they will work. This is the result of running docgen locally after this change:

Screen Shot 2020-12-01 at 12 05 08 PM

Feiyang1 (approved):

scripts/docgen/content-sources/js/toc.yaml
egilmorez: Got it, thanks!

egilmorez (approved):

PR #4148, opened 2020-12-01
Properly escape field paths

schmidt-sebastian:
JavaScript's `replace()` by default only replaces the first occurrence of a string, which breaks our field path escaping.


wilhuff (approved):
LGTM

PR #4153, opened 2020-12-02
Replace usage of transform with updateTransforms

thebrianchen:
Context: b/174694687

### DIFF
--- a/.changeset/rich-birds-wink.md
+++ b/.changeset/rich-birds-wink.md
@@ -1,5 +1,4 @@
 ---
 '@firebase/firestore': minor
 ---
-
-Field tranforms performed within a write no longer count as additional operations. However, a field transform on its own still counts as an operation.
+A write to a document that contains field transforms is no longer split into separate operations. This reduces the number of writes that the backend performs.
--- a/.changeset/rich-birds-wink.md
+++ b/.changeset/rich-birds-wink.md
@@ -1,0 +1,5 @@
+---
+'@firebase/firestore': minor
+---
+
+Field tranforms performed within a write no longer count as additional operations. However, a field transform on its own still counts as an operation.
--- a/.changeset/rich-birds-wink.md
+++ b/.changeset/rich-birds-wink.md
@@ -1,4 +1,4 @@
 ---
 '@firebase/firestore': minor
 ---
-A write to a document that contains field transforms is no longer split into separate operations. This reduces the number of writes that the backend performs.
+A write to a document that contains FieldValue transforms is no longer split up into two separate operations. This reduces the number of writes the backend performs and allows each WriteBatch to hold 500 writes regardless of how many FieldValue transformations are attached.
--- a/packages/firestore/src/local/local_serializer.ts
+++ b/packages/firestore/src/local/local_serializer.ts
@@ -200,8 +200,7 @@ export function fromDbMutationBatch(
   // on the SDK means that old `transform` mutations stored in IndexedDB need
   // to be updated to `update_transforms`.
   // TODO(b/174608374): Remove this code once we perform a schema migration.
-  let i = dbBatch.mutations.length - 1;
-  while (i >= 0) {
+  for (let i = dbBatch.mutations.length - 1; i >= 0; --i) {
     const mutationProto = dbBatch.mutations[i];
     if (mutationProto?.transform !== undefined) {
       debugAssert(
@@ -215,7 +214,6 @@ export function fromDbMutationBatch(
       dbBatch.mutations.splice(i, 1);
       --i;
     }
-    --i;
   }
 
   const mutations = dbBatch.mutations.map(m =>
--- a/packages/firestore/src/local/local_serializer.ts
+++ b/packages/firestore/src/local/local_serializer.ts
@@ -194,6 +194,30 @@ export function fromDbMutationBatch(
   const baseMutations = (dbBatch.baseMutations || []).map(m =>
     fromMutation(localSerializer.remoteSerializer, m)
   );
+
+  // Squash old transform mutations into existing patch or set mutations.
+  // The replacement of representing `transforms` with `update_transforms`
+  // on the SDK means that old `transform` mutations stored in IndexedDB need
+  // to be updated to `update_transforms`.
+  // TODO(b/174608374): Remove this code once we perform a schema migration.
+  let i = dbBatch.mutations.length - 1;
+  while (i >= 0) {
+    const mutationProto = dbBatch.mutations[i];
+    if (mutationProto?.transform !== undefined) {
+      debugAssert(
+        i >= 1 &&
+          dbBatch.mutations[i - 1].transform === undefined &&
+          dbBatch.mutations[i - 1].update !== undefined,
+        'TransformMutation should be preceded by a patch or set mutation'
+      );
+      const mutationToJoin = dbBatch.mutations[i - 1];
+      mutationToJoin.updateTransforms = mutationProto.transform.fieldTransforms;
+      dbBatch.mutations.splice(i, 1);
+      --i;
+    }
+    --i;
+  }
+
   const mutations = dbBatch.mutations.map(m =>
     fromMutation(localSerializer.remoteSerializer, m)
   );
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -107,7 +107,7 @@ export function fieldTransformsAreEqual(
     return true;
   }
 
-  if (!!left && !!right) {
+  if (left && right) {
     return arrayEquals(left, right, (l, r) => fieldTransformEquals(l, r));
   }
 
@@ -266,16 +266,7 @@ export abstract class Mutation {
   abstract readonly type: MutationType;
   abstract readonly key: DocumentKey;
   abstract readonly precondition: Precondition;
-}
-
-/**
- * Used to represent Mutation classes that have a `fieldTransforms` field,
- * namely SetMutation and PatchMutation. This intermediate class makes it
- * easier to access the `fieldTransforms` field directly from a Mutation
- * object.
- */
-export abstract class MutationWithTransforms extends Mutation {
-  abstract readonly fieldTransforms?: FieldTransform[];
+  abstract readonly fieldTransforms: FieldTransform[];
 }
 
 /**
@@ -386,12 +377,9 @@ export function extractMutationBaseValue(
   mutation: Mutation,
   maybeDoc: MaybeDocument | null
 ): ObjectValue | null {
-  if (
-    mutation instanceof MutationWithTransforms &&
-    (mutation as MutationWithTransforms).fieldTransforms !== undefined
-  ) {
+  if (mutation.fieldTransforms !== undefined) {
     return extractTransformMutationBaseValue(
-      (mutation as MutationWithTransforms).fieldTransforms!,
+      mutation.fieldTransforms,
       maybeDoc
     );
   }
@@ -441,24 +429,19 @@ export function mutationEquals(left: Mutation, right: Mutation): boolean {
   }
 
   if (left.type === MutationType.Set) {
-    const setLeft: SetMutation = left as SetMutation;
-    const setRight: SetMutation = right as SetMutation;
     return (
-      setLeft.value.isEqual(setRight.value) &&
-      fieldTransformsAreEqual(setLeft.fieldTransforms, setRight.fieldTransforms)
+      (left as SetMutation).value.isEqual((right as SetMutation).value) &&
+      fieldTransformsAreEqual(left.fieldTransforms, right.fieldTransforms)
     );
   }
 
   if (left.type === MutationType.Patch) {
-    const patchLeft: PatchMutation = left as PatchMutation;
-    const patchRight: PatchMutation = right as PatchMutation;
     return (
-      patchLeft.data.isEqual(patchRight.data) &&
-      patchLeft.fieldMask.isEqual(patchRight.fieldMask) &&
-      fieldTransformsAreEqual(
-        patchLeft.fieldTransforms,
-        patchRight.fieldTransforms
-      )
+      (left as PatchMutation).data.isEqual((right as PatchMutation).data) &&
+      (left as PatchMutation).fieldMask.isEqual(
+        (right as PatchMutation).fieldMask
+      ) &&
+      fieldTransformsAreEqual(left.fieldTransforms, right.fieldTransforms)
     );
   }
 
@@ -497,12 +480,12 @@ function getPostMutationVersion(
  * A mutation that creates or replaces the document at the given key with the
  * object value contents.
  */
-export class SetMutation extends MutationWithTransforms {
+export class SetMutation extends Mutation {
   constructor(
     readonly key: DocumentKey,
     readonly value: ObjectValue,
     readonly precondition: Precondition,
-    readonly fieldTransforms?: FieldTransform[]
+    readonly fieldTransforms: FieldTransform[] = []
   ) {
     super();
   }
@@ -581,13 +564,13 @@ function applySetMutationToLocalView(
  *  * When a field is not in the mask but is in the values, the values map is
  *    ignored.
  */
-export class PatchMutation extends MutationWithTransforms {
+export class PatchMutation extends Mutation {
   constructor(
     readonly key: DocumentKey,
     readonly data: ObjectValue,
     readonly fieldMask: FieldMask,
     readonly precondition: Precondition,
-    readonly fieldTransforms?: FieldTransform[]
+    readonly fieldTransforms: FieldTransform[] = []
   ) {
     super();
   }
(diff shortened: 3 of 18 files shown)


packages/firestore/src/local/local_serializer.ts
```diff
@@ -194,6 +194,30 @@ export function fromDbMutationBatch(
...
     fromMutation(localSerializer.remoteSerializer, m)
   );
+
+  // Squash old transform mutations into existing patch or set mutations.
+  // The replacement of representing `transforms` with `update_transforms`
+  // on the SDK means that old `transform` mutations stored in IndexedDB need
+  // to be updated to `update_transforms`.
+  // TODO(b/174608374): Remove this code once we perform a schema migration.
+  let i = dbBatch.mutations.length - 1;
+  while (i >= 0) {
```
schmidt-sebastian, line 204: I think this might be easier to parse as a `for` loop.

packages/firestore/src/model/mutation.ts
```diff
@@ -99,6 +99,21 @@ export function fieldTransformEquals(
...
 
+export function fieldTransformsAreEqual(
+  left?: FieldTransform[],
+  right?: FieldTransform[]
+): boolean {
+  if (left === undefined && right === undefined) {
+    return true;
+  }
+
+  if (!!left && !!right) {
```
schmidt-sebastian, line 110: I don't think you need the double bang here.

packages/firestore/src/model/mutation.ts
```diff
@@ -254,6 +268,16 @@ export abstract class Mutation {
   abstract readonly precondition: Precondition;
 }
 
+/**
+ * Used to represent Mutation classes that have a `fieldTransforms` field,
+ * namely SetMutation and PatchMutation. This intermediate class makes it
+ * easier to access the `fieldTransforms` field directly from a Mutation
+ * object.
+ */
+export abstract class MutationWithTransforms extends Mutation {
```
schmidt-sebastian, line 277: Can this just be an interface? Interfaces are neutral in terms of code size and disappear once transpiled to JS.

I also think this should be called `TransformMutation` (sounds familiar? :)) or something. Before you tackle this take a look at the comment below - I think we might be able to simplify the code if Mutation itself has `fieldTransforms`.

packages/firestore/src/model/mutation.ts
```diff
@@ -365,12 +386,47 @@ export function extractMutationBaseValue(
   mutation: Mutation,
```
schmidt-sebastian, line 386: Should this `mutation` be a TranformMutation? It might be cleaner to move the type check to the callsite. You can probably then inline the inner function again (since the method is now simpler).

packages/firestore/src/model/mutation.ts
```diff
@@ -385,26 +441,27 @@ export function mutationEquals(left: Mutation, right: Mutation): boolean {
   }
 
   if (left.type === MutationType.Set) {
-    return (left as SetMutation).value.isEqual((right as SetMutation).value);
+    const setLeft: SetMutation = left as SetMutation;
+    const setRight: SetMutation = right as SetMutation;
+    return (
+      setLeft.value.isEqual(setRight.value) &&
+      fieldTransformsAreEqual(setLeft.fieldTransforms, setRight.fieldTransforms)
```
schmidt-sebastian, line 448: Can we move `fieldTransforms` up to the Mutation base class? I think the PR will become a lot simpler then. You can then check `fieldTransformsAreEqual` regardless of type and `MutationWithTransforms` might even go away.

packages/firestore/src/model/mutation.ts
```diff
@@ -254,6 +268,16 @@ export abstract class Mutation {
...
 }
 
+/**
+ * Used to represent Mutation classes that have a `fieldTransforms` field,
+ * namely SetMutation and PatchMutation. This intermediate class makes it
+ * easier to access the `fieldTransforms` field directly from a Mutation
+ * object.
+ */
+export abstract class MutationWithTransforms extends Mutation {
+  abstract readonly fieldTransforms?: FieldTransform[];
```
schmidt-sebastian, line 278: Would the code be simpler if we made this non-optional and defaulted to `[]`?

packages/firestore/src/model/mutation.ts
```diff
@@ -457,29 +515,55 @@ function applySetMutationToRemoteDocument(
...
-    mutationResult.transformResults == null,
-    'Transform results received by SetMutation.'
-  );
-
   // Unlike applySetMutationToLocalView, if we're applying a mutation to a
   // remote document the server has accepted the mutation so the precondition
   // must have held.
-  return new Document(mutation.key, mutationResult.version, mutation.value, {
+  let newData = mutation.value;
+  if (mutation.fieldTransforms && mutationResult.transformResults) {
```
schmidt-sebastian, line 522: Shouldn't we always either have or not have `mutation.fieldTransforms` and `mutationResult.transformResults`? The transform results for some of the operations are only partial but the entry still exists (e.g array union) .

FYI, if you initialize `mutation.fieldTransforms` to an empty array then this if-check goes away.

packages/firestore/src/model/mutation.ts
```diff
@@ -515,35 +600,57 @@ function applyPatchMutationToRemoteDocument(
...
+    const transformResults = serverTransformResults(
+      mutation.fieldTransforms,
+      maybeDoc,
+      mutationResult.transformResults
+    );
+    newData = transformObject(
+      mutation.fieldTransforms,
+      newData,
+      transformResults
+    );
```
schmidt-sebastian, line 622: We might be able to reduce code duplication if you pass `transformResults` to `patchDocument ` and let it before the duties of `transformObject` as well.

packages/firestore/src/model/mutation.ts
```diff
@@ -582,136 +689,6 @@ function patchObject(mutation: PatchMutation, data: ObjectValue): ObjectValue {
...
-/**
- * A mutation that modifies specific fields of the document with transform
- * operations. Currently the only supported transform is a server timestamp, but
- * IP Address, increment(n), etc. could be supported in the future.
- *
- * It is somewhat similar to a PatchMutation in that it patches specific fields
- * and has no effect when applied to a null or NoDocument (see comment on
- * Mutation for rationale).
- */
-export class TransformMutation extends Mutation {
```
schmidt-sebastian, line 594: Byeees!

packages/firestore/src/remote/serializer.ts
```diff
@@ -644,6 +635,16 @@ export function toMutation(
     return fail('Unknown mutation type ' + mutation.type);
   }
 
+  if (
+    mutation instanceof MutationWithTransforms &&
+    mutation.fieldTransforms &&
+    mutation.fieldTransforms.length > 0
```
schmidt-sebastian, line 641: Again, this could be simplified a bit. We should make sure though to not send an empty array if we no transforms exist (instead, we should just not specify the field to save on bandwidth).

packages/firestore/test/unit/local/local_serializer.test.ts
```diff
@@ -0,0 +1,328 @@
...
+      // Legacy transform proto.
+      {
+        transform: {
+          document: toName(s, mutationSet.key),
+          fieldTransforms: [
+            { fieldPath: 'integer', increment: { integerValue: '42' } },
+            { fieldPath: 'double', increment: { doubleValue: 13.37 } }
+          ]
+        },
+        currentDocument: { exists: true }
```
schmidt-sebastian, line 59: Can you extract this into a helper function or `const` variable?

packages/firestore/test/unit/local/local_serializer.test.ts
```diff
@@ -0,0 +1,328 @@
...
+      [],
+      mutations
+    );
+
+    const expected = {
+      update: toMutationDocument(s, mutationSet.key, mutationSet.value),
+      updateTransforms: [
+        { fieldPath: 'integer', increment: { integerValue: '42' } },
+        { fieldPath: 'double', increment: { doubleValue: 13.37 } }
+      ]
```
schmidt-sebastian, line 75: Maybe extract this part as well.

packages/firestore/test/unit/model/mutation.test.ts
```diff
@@ -244,19 +243,20 @@ describe('Mutation', () => {
...
   // test once we have integration tests.
   it('can create arrayUnion() transform.', () => {
-    const transform = transformMutation('collection/key', {
+    const transform = patchMutation('collection/key', {
       foo: FieldValue.arrayUnion('tag'),
       'bar.baz': FieldValue.arrayUnion(true, { nested: { a: [1, 2] } })
     });
-    expect(transform.fieldTransforms.length).to.equal(2);
+    expect(transform.fieldTransforms).to.not.be.undefined;
+    expect(transform.fieldTransforms!.length).to.equal(2);
```
schmidt-sebastian, line 251: https://www.chaijs.com/api/bdd/#method_lengthof

packages/firestore/test/unit/model/mutation.test.ts
```diff
@@ -270,12 +270,13 @@ describe('Mutation', () => {
   // we don't have unit tests for it currently. We could consider removing this
   // test once we have integration tests.
   it('can create arrayRemove() transform.', () => {
-    const transform = transformMutation('collection/key', {
+    const transform = patchMutation('collection/key', {
       foo: FieldValue.arrayRemove('tag')
     });
-    expect(transform.fieldTransforms.length).to.equal(1);
+    expect(transform.fieldTransforms).to.not.be.undefined;
+    expect(transform.fieldTransforms!.length).to.equal(1);
```
schmidt-sebastian, line 277: Same as above - see https://www.chaijs.com/api/bdd/#method_lengthof

packages/firestore/test/unit/remote/serializer.helper.ts
```diff
@@ -661,66 +660,108 @@ export function serializerTest(
         verifyMutation(mutation, proto);
       });
 
-      it('TransformMutation (ServerTimestamp transform)', () => {
-        const mutation = transformMutation('baz/quux', {
+      it('ServerTimestamp transform', () => {
+        const mutation = patchMutation('baz/quux', {
```
schmidt-sebastian, line 664: Nit: These tests would be a tad shorter if you used a `set` mutation.

packages/firestore/test/unit/remote/serializer.helper.ts
```diff
@@ -661,66 +660,108 @@ export function serializerTest(
...
         const proto = {
-          transform: {
-            document: toName(s, mutation.key),
-            fieldTransforms: [
-              { fieldPath: 'a', setToServerValue: 'REQUEST_TIME' },
-              { fieldPath: 'bar.baz', setToServerValue: 'REQUEST_TIME' }
-            ]
-          },
+          update: toMutationDocument(s, mutation.key, mutation.data),
+          updateMask: toDocumentMask(mutation.fieldMask),
```
schmidt-sebastian, line 670: You should add a comment to the `fieldMask` field that states that the mask does not include document transforms.

packages/firestore/test/util/helpers.ts
```diff
@@ -230,7 +230,20 @@ export function setMutation(
...
+    'setMutation',
+    setKey,
+    json,
+    false
+  );
+  return new SetMutation(
+    setKey,
+    parsed.data,
+    Precondition.none(),
+    parsed.fieldTransforms.length > 0 ? parsed.fieldTransforms : undefined
```
schmidt-sebastian, line 245: Again, this would be simpler too if you passed an empty array.

schmidt-sebastian (commented):
This looks great! All my comments are basically just nits.

packages/firestore/test/unit/remote/serializer.helper.ts
schmidt-sebastian: https://github.com/firebase/firebase-js-sdk/blob/80d19e2f62ef1e43d7ea059415d9b93804deb025/packages/firestore/src/api/user_data_reader.ts#L102

.changeset/rich-birds-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/firestore': minor
+---
+
+Field tranforms performed within a write no longer count as additional operations. However, a field transform on its own still counts as an operation.
```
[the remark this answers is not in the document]
schmidt-sebastian: Most users access Firestore through our SDK. They are not aware of the terminology that is used in the Proto, and we should use the SDK terms in our release notes (at least in addition to the Proto terms).

BulkWriter will not come to the Mobile SDKs. WriteBatches on Mobile are meant for atomic commits and I don't expect that people will use them for large bulk insertions.

packages/firestore/src/model/mutation.ts
```diff
@@ -385,23 +429,19 @@ export function mutationEquals(left: Mutation, right: Mutation): boolean {
   }
 
   if (left.type === MutationType.Set) {
-    return (left as SetMutation).value.isEqual((right as SetMutation).value);
+    return (
+      (left as SetMutation).value.isEqual((right as SetMutation).value) &&
+      fieldTransformsAreEqual(left.fieldTransforms, right.fieldTransforms)
```
schmidt-sebastian, line 434: This check can now be extracted and performed before going into any of the if statements.

packages/firestore/src/model/mutation.ts
schmidt-sebastian: The `mutation.fieldTransforms` check still shows up for me.

packages/firestore/src/model/mutation.ts
```diff
@@ -556,15 +634,20 @@ function applyPatchMutationToLocalView(
...
 ): ObjectValue {
   let data: ObjectValue;
   if (maybeDoc instanceof Document) {
     data = maybeDoc.data();
   } else {
     data = ObjectValue.empty();
   }
-  return patchObject(mutation, data);
+  data = patchObject(mutation, data);
+  if (transformResults) {
```
schmidt-sebastian, line 647: This is never false since you pass in `[]`.

packages/firestore/src/remote/serializer.ts
```diff
@@ -644,6 +634,12 @@ export function toMutation(
     return fail('Unknown mutation type ' + mutation.type);
   }
 
+  if (mutation.fieldTransforms && mutation.fieldTransforms.length > 0) {
```
schmidt-sebastian, line 637: `mutation.fieldTransforms` is never false.

packages/firestore/src/remote/serializer.ts
```diff
@@ -659,31 +655,34 @@ export function fromMutation(
     ? fromPrecondition(proto.currentDocument)
     : Precondition.none();
 
+  const fieldTransforms = proto.updateTransforms
+    ? proto.updateTransforms.map(transform =>
+        fromFieldTransform(serializer, transform)
+      )
+    : undefined;
```
schmidt-sebastian, line 662: Doesn't this need be initialized to `[]`?

packages/firestore/test/unit/local/local_serializer.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      document: toName(s, mutationSet.key),
+      fieldTransforms: [
+        { fieldPath: 'integer', increment: { integerValue: '42' } },
+        { fieldPath: 'double', increment: { doubleValue: 13.37 } }
+      ]
+    },
+    currentDocument: { exists: true }
+  };
+  // Updated transform proto representation.
+  const updateTransformWrite = {
```
schmidt-sebastian, line 72: s/updateTransformWrite/updateTransforms ?

packages/firestore/test/unit/model/mutation.test.ts
```diff
@@ -244,19 +243,20 @@ describe('Mutation', () => {
...
+    const transform = patchMutation('collection/key', {
       foo: FieldValue.arrayUnion('tag'),
       'bar.baz': FieldValue.arrayUnion(true, { nested: { a: [1, 2] } })
     });
-    expect(transform.fieldTransforms.length).to.equal(2);
+    expect(transform.fieldTransforms).to.not.be.undefined;
+    expect(transform.fieldTransforms).to.have.lengthOf(2);
 
-    const first = transform.fieldTransforms[0];
+    const first = transform.fieldTransforms![0];
```
schmidt-sebastian, line 253: Is this bang needed? (Same below)

packages/firestore/test/unit/model/mutation.test.ts
```diff
@@ -270,12 +270,13 @@ describe('Mutation', () => {
   // we don't have unit tests for it currently. We could consider removing this
   // test once we have integration tests.
   it('can create arrayRemove() transform.', () => {
-    const transform = transformMutation('collection/key', {
+    const transform = patchMutation('collection/key', {
       foo: FieldValue.arrayRemove('tag')
     });
-    expect(transform.fieldTransforms.length).to.equal(1);
+    expect(transform.fieldTransforms).to.not.be.undefined;
```
schmidt-sebastian, line 276: This should always be true, shouldn't it?

schmidt-sebastian (commented):
Whee. Found more nits, but this should only need a couple more minutes.

packages/firestore/src/model/mutation.ts
thebrianchen: Extracted.

packages/firestore/src/model/mutation.ts
thebrianchen: Oops, removed them everywhere but here.

packages/firestore/src/model/mutation.ts
thebrianchen: Removed the if statement.

packages/firestore/src/remote/serializer.ts
thebrianchen: Removed.

packages/firestore/src/remote/serializer.ts
thebrianchen: Changed.

packages/firestore/test/unit/local/local_serializer.test.ts
thebrianchen: Done.

packages/firestore/test/unit/model/mutation.test.ts
thebrianchen: Removed here and throughout.

.changeset/rich-birds-wink.md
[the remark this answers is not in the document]
thebrianchen: Thanks for explaining. I changed it to use your wording and added a bit about WriteBatch limits.

schmidt-sebastian (approved):

PR #4154, opened 2020-12-02
Update Console domains for RTDB SDK

schmidt-sebastian:
Updates the console detection to use new URLs (*.corp.google.com and console.firebase.google.com). This will ensure that we append the "r=f" URL param.

### DIFF
--- a/packages/database/src/realtime/Constants.ts
+++ b/packages/database/src/realtime/Constants.ts
@@ -25,7 +25,8 @@ export const REFERER_PARAM = 'r';
 
 export const FORGE_REF = 'f';
 
-// Matches console.firebase.google.com and *.corp.google.com.
+// Matches console.firebase.google.com, firebase-console-*.corp.google.com and
+// firebase.corp.google.com
 export const FORGE_DOMAIN_RE = /(console\.firebase|firebase-console-\w+\.corp|firebase\.corp)\.google\.com/;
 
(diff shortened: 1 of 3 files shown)


packages/database/src/realtime/Constants.ts
```diff
@@ -25,7 +25,8 @@ export const REFERER_PARAM = 'r';
 
 export const FORGE_REF = 'f';
 
-export const FORGE_DOMAIN = 'firebaseio.com';
+// Matches console.firebase.google.com and *.corp.google.com.
+export const FORGE_DOMAIN_RE = /(console\.firebase|.*\.corp)\.google\.com/;
```
lepatryk, line 30: Can we make the regexp stricter to match firebase console under corp domain only? I'm worried it may brake some internal apps.

schmidt-sebastian: Updated to use firebase-console-\w*.corp as per Laura's suggestion.

lepatryk: Now it doesn't match the prod firebase corp URL: "firebase.corp.google.com" :-).

schmidt-sebastian: It's a regular expression, so of course it is going to be buggy. Fixed (and manually tested)

lepatryk (approved):

jsdt (approved):

PR #4159, opened 2020-12-03
Add scripts to build auth-compat for release

hsubox76:
Add a release build config for auth-compat, and npm script.

Also bump firebase-exp version because I was in the middle of an npm publish when debugging this and the publish completed.


samhorlbeck (approved):
Thanks!

PR #4160, opened 2020-12-03
Remove Cyclic Dependencies

schmidt-sebastian:
This PR should unblock the Console build and should also allow us to revive our old goal to sort our imports.

It also adds a build rule change to block us from introducing circular dependencies in the future (https://github.com/firebase/firebase-js-sdk/pull/4160/commits/ec5ac56afc95a374672bb041b2a722c6ba8eb374)


schmidt-sebastian:
FYI @dannhauser and @hsubox76 confirmed that this solves their respective build problems.

packages/firestore/src/local/indexeddb_persistence.ts
```diff
@@ -261,7 +228,7 @@ export class IndexedDbPersistence implements Persistence {
       );
     }
 
-    this.referenceDelegate = new IndexedDbLruDelegate(this, lruParams);
+    this.referenceDelegate = new IndexedDbLruDelegateImpl(this, lruParams);
```
wilhuff, line 218: Should there be a `newIndexedDbLruDelegate` to hide this and mirror other splits like this?

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -583,10 +582,9 @@ class IndexedDbRemoteDocumentChangeBuffer extends RemoteDocumentChangeBuffer {
 function documentGlobalStore(
   txn: PersistenceTransaction
 ): SimpleDbStore<DbRemoteDocumentGlobalKey, DbRemoteDocumentGlobal> {
-  return IndexedDbPersistence.getStore<
-    DbRemoteDocumentGlobalKey,
-    DbRemoteDocumentGlobal
-  >(txn, DbRemoteDocumentGlobal.store);
+  return txn.getStore<DbRemoteDocumentGlobalKey, DbRemoteDocumentGlobal>(
```
wilhuff, line 584: This seems like a regression in terms of tree-shakeability. Should this be a free function?

packages/firestore/src/local/persistence_transaction.ts
```diff
@@ -47,4 +50,12 @@ export abstract class PersistenceTransaction {
   raiseOnCommittedEvent(): void {
     this.onCommittedListeners.forEach(listener => listener());
   }
+
+  getStore<Key extends IDBValidKey, Value>(
```
wilhuff, line 53: Putting this in PersistenceTransaction seems off. Only the IndexedDb implementation actually uses SimpleDb.

As commented elsewhere, a free function in `indexeddb_transaction.ts` that takes `PersistenceTransaction` as a type and downcasts would make this cleaner.

packages/firestore/src/local/persistence_transaction.ts
```diff
@@ -16,6 +16,9 @@
  */
 
 import { ListenSequenceNumber } from '../core/types';
+import { SimpleDb, SimpleDbStore } from './simple_db';
+import { fail } from '../util/assert';
+import { IndexedDbTransaction } from './indexeddb_transaction';
```
wilhuff: This dependency seems backwards cyclically.

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -207,7 +203,7 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
     return queuesStore.loadAll().next(queues => {
       return PersistencePromise.forEach(queues, (queue: DbMutationQueue) => {
         const range = IDBKeyRange.bound(
-          [queue.userId, BATCHID_UNKNOWN],
+          [queue.userId, Number.MIN_SAFE_INTEGER],
```
wilhuff: Maybe put BATCHID_UNKNOWN in core/types so that this can remain a reference to that constant?

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -350,6 +346,71 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
   }
 }
 
+/**
+ * Delete a mutation batch and the associated document mutations.
+ * @returns A PersistencePromise of the document mutations that were removed.
+ */
+export function removeMutationBatch(
```
wilhuff: This doesn't seem like it belongs logically with indexedb_schema.

Would it make sense to have something like a mutation_batch.ts that holds this and related functions?

packages/firestore/src/model/path.ts
```diff
@@ -366,3 +366,57 @@ export class FieldPath extends BasePath<FieldPath> {
     return new FieldPath([]);
   }
 }
+
+export class DocumentKey {
```
wilhuff, line 370: DocumentKey doesn't seem to have any circular references with paths--notably this didn't result in any import changes in `path.ts`. This class is separate on other platforms and it seems like it should be separate here as well. What's the reason to move this?

packages/firestore/src/api/field_value.ts
```diff
@@ -24,9 +24,23 @@ import {
...
 import { ParseContext, parseData, UserDataSource } from './user_data_reader';
 import { debugAssert } from '../util/assert';
-import { FieldValue } from '../../lite/src/api/field_value';
 import { toNumber } from '../remote/value_serializer';
 
+/**
+ * Sentinel values that can be used when writing document fields with `set()`
+ * or `update()`.
+ */
+export abstract class FieldValue {
```
wilhuff: I can't say I really understand the layering model we have here and why this move is desirable.

It seems like conceptually the classic API should only ever point to the lite API and not the other way around. Allowing the other order encourages cycles and the possibility of polluting the lite implementation with classes that can't be tree shaken.

packages/firestore/src/api/user_data_reader.ts
```diff
@@ -398,6 +405,148 @@ export function parseSetData(
   );
 }
 
+export class DeleteFieldValueImpl extends FieldValue {
```
wilhuff, line 425: A possible thought to follow up on layering: would it make sense to explicitly break out common code and the classic API as separate paths? That way you could impose layering checks, e.g.

  * lite -> common
  * exp -> (lite, common)
  * classic -> (exp, common)

or similar?

packages/firestore/lite/src/api/components.ts
```diff
@@ -18,32 +18,141 @@
...
+      this.ignoreUndefinedProperties === other.ignoreUndefinedProperties
+    );
+  }
+}
+
 // The components module manages the lifetime of dependencies of the Firestore
 // client. Dependencies can be lazily constructed and only one exists per
 // Firestore instance.
 
+export interface FirestoreService extends _FirebaseService {
```
wilhuff, line 34: Comments, please. In particular what's a FirestoreService and how does it differ from a FirebaseFirestore instance?

packages/firestore/exp/src/api/reference.ts
```diff
@@ -823,3 +816,12 @@ function convertToDocSnapshot<T>(
     ref._converter
   );
 }
+export { endAt } from '../../../lite/src/api/query';
+export { endBefore } from '../../../lite/src/api/query';
+export { startAfter } from '../../../lite/src/api/query';
```
wilhuff: Sort these?

Edit: it seems like this may have been addressed during cleanup.

packages/firestore/lite/src/api/query.ts
```diff
@@ -0,0 +1,1247 @@
...
+ * @returns A Promise resolved once the data has been successfully written
+ * to the backend.
+ */
+export function setDoc<T>(
+  reference: DocumentReference<T>,
+  data: Partial<T>,
+  options: SetOptions
+): Promise<void>;
+
+export function setDoc<T>(
```
wilhuff: Setting, updating, and adding documents doesn't seem to have anything to do with "queries". Did you mean to move this here given that `reference.ts` still exists?

Alternatively, if this can't exist in `reference.ts` to solve circularity issues, maybe create `reference_impl.ts` to match the structure you've established elsewhere (e.g. with LocalStoreImpl)?

wilhuff (commented):
Very impressive!

One high level concern I have with the PR as is, is that it's not necessarily obvious where to look for the implementation of a given type. For example, in some cases it's the _impl but in others it's _methods. Also, whether or not this split even exists seems arbitrary.

It also seems that if you squint just right, this is a very similar change to the forward declarations change we made in iOS C++: declare the exported types separately from the actual implementations to drastically reduce the number of files imported by those type declarations. In the iOS implementation, we created a single file that "exported" the types for a package, e.g. model in `[model_fwd.h](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/src/model/model_fwd.h)`. I wonder if a similar approach here could make this feel more regular.

packages/firestore/src/local/persistence_transaction.ts
schmidt-sebastian: Done

packages/firestore/src/local/persistence_transaction.ts
schmidt-sebastian: Removed in later commit.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Done

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Moved to indexeddb_mutation_batch_impl.ts

packages/firestore/src/model/path.ts
schmidt-sebastian: I swear it did when I looked at it, but it seems to have magically disappeared. Reverted this change.

packages/firestore/src/api/field_value.ts
schmidt-sebastian: Sorry, this is also something that was fixed in a later commit. It is indeed in lite/src/api/field_value

packages/firestore/src/api/user_data_reader.ts
schmidt-sebastian: Once all the dust has settled, I would like to move the /exp and /lite files under /src. They only contain API code and the nested imports are not pretty. This is something I will consider then.

packages/firestore/lite/src/api/components.ts
schmidt-sebastian: I added the following:

```
/**
 * An interface implemented by FirebaseFirestore that provides
 * compatibility with the usage in this file. 
 * 
 * This interface mainly exists to remove a cyclic dependency.
 */
```

I will explore doing the same FirebaseFirestore/FirebaseFirestoreImpl pattern in a follow-up PR.

packages/firestore/lite/src/api/query.ts
schmidt-sebastian: Sorry, this was also cleaned up in a later commit, which I know is impossible to detect with a PR this size.

 I did rename the file to reference_impl.ts as well.

schmidt-sebastian (commented):
Feedback addressed (or punted to follow up work for next week). Sorry about this PR and thanks for the review!

package/packages/firebase/firebase-messaging.js
```diff
@@ -2560,14 +2560,12 @@
                   );
                 case 1:
                   return (
-                    e
```
schmidt-sebastian: FYI: This is removed in a commit later.

packages/firestore/exp/src/api/database.ts
```diff
@@ -68,7 +69,7 @@ export const CACHE_SIZE_UNLIMITED = LRU_COLLECTION_DISABLED;
  * Do not call this constructor directly. Instead, use {@link getFirestore}.
  */
 export class FirebaseFirestore extends LiteFirestore {
-  readonly _queue = newAsyncQueue();
+  readonly _queue: AsyncQueue = newAsyncQueue();
```
schmidt-sebastian, line 72: Somehow the `AsyncQueue` type is required to not break the APIExtractor toolchain.

packages/firestore/src/model/document_key.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google LLC
+ * Copyright 2020 Google LLC
```
wilhuff, line 3: If you change this back to 2017 and add a newline back after `import { debugAssert }` this file we be unchanged in this PR.

packages/firestore/src/remote/backoff.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google LLC
+ * Copyright 2020 Google LLC
```
wilhuff, line 3: If you change this back to 2017 and revert the import reordering, this file could now be unchanged in this PR.

wilhuff (approved):
LGTM

packages/firestore/src/model/document_key.ts
schmidt-sebastian: Done

packages/firestore/src/remote/backoff.ts
schmidt-sebastian: Done

Issue #4158, opened 2020-12-03
Firebase storage uploadTask snapshots not progressing


jamesdaniels:
It appears there's a bug with the way the new `Compat` delegates are setup in storage, a good reproduction is to cancel an upload and then check the snapshot state.

`task.snapshot.state` still equals `running`, whereas `task._delegate.snapshot.state` appropriately reads `canceled`.

You can see this on success too. I noticed this while debugging AngularFire/RxFire issues that users were reporting, while I worked around this by using [the promise for now](https://github.com/angular/angularfire/pull/2704/files#diff-44011a9aed62d316df06a7472095a3e7b6d817b99fc525bc17505663b4f46557R23) developers won't be able to `takeWhile(snap => snap.state === 'running')` etc. to catch errors, leading to broken observables and general mayhem in their apps.

Attached some screenshots of states seen in my console.logs. Note the conflict between the _delegate and the compat states.

<img width="416" alt="Screen Shot 2020-12-03 at 4 44 39 PM" src="https://user-images.githubusercontent.com/44975/101092010-f78c7800-3586-11eb-9a34-8a309e079cda.png">

<img width="386" alt="Screen Shot 2020-12-03 at 4 44 26 PM" src="https://user-images.githubusercontent.com/44975/101092002-f4918780-3586-11eb-9370-f6f35d76600a.png">

Issue #1939, opened 2019-07-03
Lazy (on first subscription rather than construction) rxfire storage observables


mattgibb:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS Mojave 10.14.5
  * Browser version: Chrome 74.0.3729.169
  * Firebase SDK version: SDK v0.1.0, @firebase/storage v0.3.4
  * Firebase Product: storage

Hey guys, thanks for a super useful library! Apologies if this is already discussed/resolved, but I thought I may as well give this a whirl.
As a beginner, I had assumed that the `put` observable was cold, since you pass in the ref and the data as separate arguments rather than construct and initialise the Task yourself and pass that in. The task is actually created when you call `put`, and the download begins eagerly, when the observable is created. This means that if subscription happens at any time after observable creation, some of the `UploadTaskSnapshots` have been missed.

In my particular application, I came across this when assigning a maximum concurrency to `mergeMap()`, so that only 10 of say 100 files can be uploaded at any given time. As the files arrived from the previous step (data being anonymised for a medical application in my case), they would sit in the internal `mergeMap` queue having been mapped to a task via `fromTask`, but not yet subscribed to (hopefully I have understood/diagnosed this correctly :-S). This meant that I was then missing smaller files or files that happen to have zero bytes downstream, and progress was never reported.

What do you think about a behaviour more like `shareReplay(/*bufferSize=*/1)`, where the download is started on the first subscription, and only the most recent `UploadTaskSnapshot` is cached to be fed to any subsequent or late subscribers?

I'd be very interested to hear your feedback on this, and if you were happy, I would be delighted to put together a little PR for this.

#### Steps to reproduce:
```javascript
const file = new File([""], 'hello')
const immediatelySubscribed  = put(ref, f)
immediatelySubscribed.subscribe(
  console.log,
  console.error,
  () => console.warn("done")
)
# outputs:
# * UploadTaskSnapshot
# * complete
const subscribedLater = put(ref, f)
concat(timer(500), subscribedLater).subscribe(
  console.log,
  console.error,
  () => console.warn("done")
)
# outputs (with no TaskUpload Snapshot):
# * complete
```

jamesdaniels:
Rather than have a lazy observable for `put`/`putString` I've just reworked how Storage observables work in RxFire; this should be much much less surprising. https://github.com/FirebaseExtended/rxfire/pull/8

Issue #1659, opened 2019-04-04
rxFire's storage `fromTask` can cancel uploads unexpectedly


jhuleatt:
* Firebase SDK version: rxFire, latest
  * Firebase Product: storage

rxfire/storage's [`fromTask`](https://github.com/firebase/firebase-js-sdk/blob/master/packages/rxfire/docs/storage.md#fromtask) cancels uploads unexpectedly

### expected behavior

using `pipe(first())` or `pipe(take(1))` on the `fromTask` observable lets me get the initial snapshot returned from the Storage upload task, but doesn't affect the upload itself.

### actual behavior

using `pipe(first())` or `pipe(take(1))` on the `fromTask` [cancels the UploadTask](https://github.com/firebase/firebase-js-sdk/blob/master/packages/rxfire/storage/index.ts#L29) in its unsubscribe. This is problematic when using `take` or `first` on the observable, which automatically call the unsubscribe and therefore cancel the upload.

jhuleatt:
cc @jamesdaniels @davideast

SaliZumberi:
Any updates on this?

PR #4170, opened 2020-12-04
Various fixes for RxFire storage/fromTask

jamesdaniels:
Bringing in the storage fixes from AngularFire 6.1.4 into RxFire.

* Don't cancel the upload if the task is unsubscribed from, this is unexpected behavior #1659. `fromTask` is actually a hot-observable and the task controls it (having the cancel, pause, and resume functions) thus we should lean on the task to control itself to not surprise the developer
* As such `fromTask` should be idempotent and immediately emit the latest snapshot, as the task is effective a Subject
* Before completion `fromTask` was not emitting a `success` state, move to the promise form of the task to get that, as such `progress` was never reaching 100%
* After #4158 is addressed, `fromTask` should emit an `error` or `canceled` snapshot, before the observable errors out; making for easier handling of errors. `takeWhile(snap => snap.state !== 'canceled')` for instance.
* `put` and `putString` should be cold, replay, and cancel on unsubscribe however #1939 

FYI @jhuleatt 

TODO we really need some storage tests, I'll work on those.

### DIFF
--- a/.changeset/spicy-garlics-switch.md
+++ b/.changeset/spicy-garlics-switch.md
@@ -1,0 +1,12 @@
+---
+"rxfire": patch
+---
+
+Addressing several bugs with `rxfire/storage` and cleaning up some poorly defined behavior:
+
+* `fromTask` should not cancel the task on unsubscribe
+* `fromTask` should immediately emit the current snapshot
+* `fromTask` should return a `success` snapshot before completion
+* `fromTask` should emit a `canceled` or `error` snapshot before erroring
+* `progress` should now correctly report 100% uploads as it uses `fromTask`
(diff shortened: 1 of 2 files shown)


.changeset/spicy-garlics-switch.md
```diff
@@ -0,0 +1,12 @@
+---
+"rxfire": patch
```
jamesdaniels, line 2: IMO this is a patch since it's cleaning up poorly defined behavior and bugs and this library is in beta.

.changeset/spicy-garlics-switch.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Addressing several bugs with `rxfire/storage` and cleaning up some poorly defined behavior:
```
jamesdaniels, line 5: TODO after reading up on changesets, clear I should break this up into multiple

PR #4177, opened 2020-12-07
A few fixes for Firestore bundle sdk builds

wu-hui:
* Add missing `firebase/firestore/bundle` and `firebase/firestore/memory/bundle`.
* Add bundle to firebase complete build
* Add missing firestore.proto for Firestore memory build for Node.

### DIFF
--- a/packages/firebase/rollup.config.js
+++ b/packages/firebase/rollup.config.js
@@ -111,7 +111,7 @@ const appBuilds = [
 
 const componentBuilds = pkg.components
   // The "app" component is treated differently because it doesn't depend on itself.
-  .filter(component => component !== 'app')
+  .filter(component => component !== 'app' && component !== 'firestore')
   .map(component => {
     const pkg = require(`./${component}/package.json`);
     return [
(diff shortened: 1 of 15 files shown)


packages/firebase/rollup.config.js
```diff
@@ -111,7 +117,7 @@ const appBuilds = [
 
 const componentBuilds = pkg.components
   // The "app" component is treated differently because it doesn't depend on itself.
-  .filter(component => component !== 'app')
+  .filter(component => component !== 'app' && component !== 'firestore')
```
schmidt-sebastian, line 120: Please update comment.

packages/firebase/rollup.config.js
```diff
@@ -149,17 +155,67 @@ const componentBuilds = pkg.components
...
+        format: 'cjs',
+        sourcemap: true
+      },
+      {
+        file: resolve('firestore', firestorePkg.module),
+        format: 'es',
+        sourcemap: true
+      }
+    ],
+    plugins,
```
schmidt-sebastian, line 173: Do we not need

```
uglify({
           output: {
             ascii_only: true // escape unicode chars
            }	               
})
```

packages/firebase/src/index.rn.ts
```diff
@@ -24,6 +24,7 @@ import '../database';
 //  `atob`). We should provide a RN build that works out of the box.
 import '../storage';
 import '../firestore';
+import '../firestore/bundle';
```
schmidt-sebastian, line 27: The React Nativce build of the main Firestore package splits up bundles and non-bundled code. I think we should do the same here.

packages/firebase/rollup.config.js
wu-hui: Done.

packages/firebase/rollup.config.js
wu-hui: The index.ts is just `import '@firebase/firestore';` (so is the output), I think we can skip it here.

packages/firebase/src/index.rn.ts
wu-hui: This is for "complete" build, which include everything (except memory only builds) on other platforms.

schmidt-sebastian (approved):

PR #4181, opened 2020-12-08
Sort imports

schmidt-sebastian:
This is the same as https://github.com/firebase/firebase-js-sdk/pull/3150, but now without cyclic dependencies.

### DIFF
--- a/packages/firestore/.eslintrc.js
+++ b/packages/firestore/.eslintrc.js
@@ -33,6 +33,21 @@ module.exports = {
         args: 'none'
       }
     ],
+    'import/order': [
+      'error',
+      {
+        'groups': [
+          'builtin',
+          'external',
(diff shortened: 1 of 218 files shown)


packages/firestore/.eslintrc.js
```diff
@@ -33,6 +33,21 @@ module.exports = {
...
+          'external',
+          'internal',
+          'parent',
+          'sibling',
+          'index'
+        ],
+        'newlines-between': 'always',
+        'alphabetize': { 'order': 'asc', 'caseInsensitive': true }
+      }
+    ],
```
schmidt-sebastian, line 50: This is the only manual change and just a copy  from the original PR.

wu-hui (approved):
Yeah!

PR #4183, opened 2020-12-08
Clean up source location

schmidt-sebastian:
This commit cleans up some of the code structure around firestore-exp/firestore-lite.

The changes are:
- Code for firestore-exp is moved from exp/src/api to src/exp. There is no code left in /exp.
- Code for firestore-lite is moved from lite/src/api to src/lite. Also moved firestore-lite test to test/lite.  There is no code left in /lite.
- Removed old IntelliJ runner for firestore-exp. Removed the old `exp` support from the Karma runner.
- Deleted the `dependencies.json` files that were used during development. Updated the build rules to instead create this these in a temporary location.
- Moved `Timestamp` and `GeoPoint` to src/lite and added some export-only files to src/exp. The goal is to have a clear dependency chain where /api can only depend on /exp and /exp can only depend on /lite (the next step would be to move the `UserDataReader` to `src/lite`).


common/api-review/firestore-lite.api.md
```diff
@@ -168,8 +168,10 @@ export abstract class FieldValue {
     abstract _toFieldTransform(context: ParseContext): FieldTransform | null;
 }
 
+// Warning: (ae-forgotten-export) The symbol "FirestoreService" needs to be exported by the entry point index.d.ts
```
wilhuff, line 171: I don't know how to interpret this comment. Is this a TODO? Is this suppressing a warning?

packages/firestore/src/exp/database.ts
```diff
@@ -15,40 +15,41 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
wilhuff, line 18: Is it worth separating this group of imports from the rest and adding a comment about why this is necessary?

wilhuff (approved):
LGTM

common/api-review/firestore-lite.api.md
schmidt-sebastian: This is an artifact from our API generation pipeline, which produces these auto-generated files. My goal is to fix this problem in the current sprint.

packages/firestore/src/exp/database.ts
schmidt-sebastian: We are now using eslint to sort our imports (possible since we no longer have cycled dependencies). With that, we lost all control over formatting and empty lines.

hsubox76 (approved):

PR #4184, opened 2020-12-08
WebChannel Connection Type Callback

rafikhan:
WIP

### DIFF
--- a/packages/firestore/exp/src/api/database.ts
+++ b/packages/firestore/exp/src/api/database.ts
@@ -53,6 +53,7 @@ import {
   indexedDbStoragePrefix
 } from '../../../src/local/indexeddb_persistence';
 import { cast } from '../../../src/util/input_validation';
+import { TimeToFirstByteCallback } from '../../../src/remote/stream_bridge';
 
 /** DOMException error code constants. */
 const DOM_EXCEPTION_INVALID_STATE = 11;
@@ -81,7 +82,8 @@ export class FirebaseFirestore extends LiteFirestore {
   /** @hideconstructor */
   constructor(
     databaseIdOrApp: DatabaseId | FirebaseApp,
-    authProvider: Provider<FirebaseAuthInternalName>
+    authProvider: Provider<FirebaseAuthInternalName>,
+    public onTimeToFirstByteCallback: TimeToFirstByteCallback
   ) {
     super(databaseIdOrApp, authProvider);
     this._persistenceKey =
--- a/packages/firestore/src/core/component_provider.ts
+++ b/packages/firestore/src/core/component_provider.ts
@@ -69,6 +69,7 @@ import { newSerializer } from '../platform/serializer';
 import { getDocument, getWindow } from '../platform/dom';
 import { CredentialsProvider } from '../api/credentials';
 import { JsonProtoSerializer } from '../remote/serializer';
+import { TimeToFirstByteCallback } from '../remote/stream_bridge';
 
 export interface ComponentConfiguration {
   asyncQueue: AsyncQueue;
@@ -317,6 +318,7 @@ export class OnlineComponentProvider {
   syncEngine!: SyncEngine;
 
   async initialize(
+    timeToFirstByte: TimeToFirstByteCallback,
     offlineComponentProvider: OfflineComponentProvider,
     cfg: ComponentConfiguration
   ): Promise<void> {
@@ -332,6 +334,7 @@ export class OnlineComponentProvider {
     this.remoteStore = this.createRemoteStore(cfg);
     this.eventManager = this.createEventManager(cfg);
     this.syncEngine = this.createSyncEngine(
+      timeToFirstByte,
       cfg,
       /* startAsPrimary=*/ !offlineComponentProvider.synchronizeTabs
     );
@@ -380,10 +383,12 @@ export class OnlineComponentProvider {
   }
 
   createSyncEngine(
+    timeToFirstByte: TimeToFirstByteCallback,
     cfg: ComponentConfiguration,
     startAsPrimary: boolean
   ): SyncEngine {
     return newSyncEngine(
+      timeToFirstByte,
       this.localStore,
       this.remoteStore,
       this.eventManager,
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -83,6 +83,7 @@ import { newSerializer, newTextEncoder } from '../platform/serializer';
 import { toByteStreamReader } from '../platform/byte_stream_reader';
 import { NamedQuery } from './bundle';
 import { JsonProtoSerializer } from '../remote/serializer';
+import { TimeToFirstByteCallback } from '../remote/stream_bridge';
 
 const LOG_TAG = 'FirestoreClient';
 export const MAX_CONCURRENT_LIMBO_RESOLUTIONS = 100;
@@ -102,12 +103,13 @@ export class FirestoreClient {
   onlineComponents?: OnlineComponentProvider;
 
   constructor(
+    public readonly onTimeToFirstByte: TimeToFirstByteCallback,
     private credentials: CredentialsProvider,
     /**
      * Asynchronous queue responsible for all of our internal processing. When
      * we get incoming work from the user (via public API) or the network
      * (incoming GRPC messages), we should always schedule onto this queue.
-     * This ensures all of our work is properly serialized (e.g. we don't
+     * This Fes all of our work is properly serialized (e.g. we don't
      * start processing a new operation while the previous one is waiting for
      * an async I/O to complete).
      */
@@ -223,6 +225,7 @@ export async function setOnlineComponentProvider(
   logDebug(LOG_TAG, 'Initializing OnlineComponentProvider');
   const configuration = await client.getConfiguration();
   await onlineComponentProvider.initialize(
+    client.onTimeToFirstByte,
     offlineComponentProvider,
     configuration
   );
--- a/packages/firestore/src/core/sync_engine.ts
+++ b/packages/firestore/src/core/sync_engine.ts
@@ -109,8 +109,10 @@ import {
   EventManager,
   eventManagerOnOnlineStateChange,
   eventManagerOnWatchChange,
-  eventManagerOnWatchError
+  eventManagerOnWatchError,
+  onTimeToFirstByte
 } from './event_manager';
+import { TimeToFirstByteCallback } from '../remote/stream_bridge';
 
 const LOG_TAG = 'SyncEngine';
 
@@ -260,6 +262,7 @@ class SyncEngineImpl implements SyncEngine {
   _isPrimaryClient: undefined | boolean = undefined;
 
   constructor(
+    readonly timeToFirstByte: TimeToFirstByteCallback,
     readonly localStore: LocalStore,
     readonly remoteStore: RemoteStore,
     readonly eventManager: EventManager,
@@ -275,6 +278,7 @@ class SyncEngineImpl implements SyncEngine {
 }
 
 export function newSyncEngine(
+  timeToFirstByte: TimeToFirstByteCallback,
   localStore: LocalStore,
   remoteStore: RemoteStore,
   eventManager: EventManager,
@@ -285,6 +289,7 @@ export function newSyncEngine(
   isPrimary: boolean
 ): SyncEngine {
   const syncEngine = new SyncEngineImpl(
+    timeToFirstByte,
     localStore,
     remoteStore,
     eventManager,
@@ -1509,6 +1514,16 @@ function ensureWatchCallbacks(syncEngine: SyncEngine): SyncEngineImpl {
     null,
     syncEngineImpl
   );
+
+  let timeToFirstByteRaised = false;
+
+  syncEngineImpl.remoteStore.remoteSyncer.handleTimeToFirstByte = data => {
+    if (!timeToFirstByteRaised) {
+      timeToFirstByteRaised = true;
+      syncEngineImpl.timeToFirstByte(data);
+    }
+  };
+
   syncEngineImpl.syncEngineListener.onWatchChange = eventManagerOnWatchChange.bind(
     null,
     syncEngineImpl.eventManager
--- a/packages/firestore/src/platform/browser/webchannel_connection.ts
+++ b/packages/firestore/src/platform/browser/webchannel_connection.ts
@@ -241,6 +241,9 @@ export class WebChannelConnection extends RestConnection {
     // on a closed stream
     let closed = false;
 
+    let connectionStartTime: Date | null = null;
+    let hasRaisedTimeToFirstByteEvent = false;
+
     const streamBridge = new StreamBridge<Req, Resp>({
       sendFn: (msg: Req) => {
         if (!closed) {
@@ -284,6 +287,8 @@ export class WebChannelConnection extends RestConnection {
       if (!closed) {
         logDebug(LOG_TAG, 'WebChannel transport opened.');
       }
+
+      connectionStartTime = new Date();
     });
 
     unguardedEventListen(channel, WebChannel.EventType.CLOSE, () => {
@@ -361,6 +366,18 @@ export class WebChannelConnection extends RestConnection {
         logDebug(LOG_TAG, 'Detected buffering proxy');
       } else if (event.stat === Stat.NOPROXY) {
         logDebug(LOG_TAG, 'Detected no buffering proxy');
+      }
+
+      if (!hasRaisedTimeToFirstByteEvent) {
+        hasRaisedTimeToFirstByteEvent = true;
+
+        const connectionStopTime = new Date();
+        const elapsedMs = +connectionStopTime - +connectionStartTime!;
+
+        streamBridge.callOnTimeToFirstByte({
+          type: event.stat,
+          timeToFirstByte: elapsedMs
+        });
       }
     });
 
--- a/packages/firestore/src/remote/persistent_stream.ts
+++ b/packages/firestore/src/remote/persistent_stream.ts
@@ -45,6 +45,7 @@ import {
   toTarget,
   versionFromListenResponse
 } from './serializer';
+import { TimeToFirstByteArgs, TimeToFirstByteCallback } from './stream_bridge';
 import { WatchChange } from './watch_change';
 
 const LOG_TAG = 'PersistentStream';
@@ -127,6 +128,8 @@ export interface PersistentStreamListener {
    * FirestoreError will be set.
    */
   onClose: (err?: FirestoreError) => Promise<void>;
+
+  onTimeToFirstByte: TimeToFirstByteCallback;
 }
 
 /** The time a stream stays open after it is marked idle. */
@@ -394,6 +397,10 @@ export abstract class PersistentStream<
    */
   protected abstract onMessage(message: ReceiveType): Promise<void>;
 
+  public onTimeToFirstByte(data: TimeToFirstByteArgs): void {
+    this.listener.onTimeToFirstByte(data);
+  }
+
   private auth(): void {
     debugAssert(
       this.state === PersistentStreamState.Initial,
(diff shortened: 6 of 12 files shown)


packages/firestore/src/core/sync_engine.ts
```diff
@@ -1509,6 +1514,16 @@ function ensureWatchCallbacks(syncEngine: SyncEngine): SyncEngineImpl {
     null,
     syncEngineImpl
   );
+
+  let timeToFirstByteRaised = false;
```
rafikhan, line 1518: Note to self: Make this a member on syncEngine

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -284,6 +287,8 @@ export class WebChannelConnection extends RestConnection {
       if (!closed) {
         logDebug(LOG_TAG, 'WebChannel transport opened.');
       }
+
+      connectionStartTime = new Date();
```
schmidt-sebastian, line 291: This should probably be happen when the call `.open()`, which shows our intent.

packages/firestore/src/core/firestore_client.ts
```diff
@@ -102,12 +103,13 @@ export class FirestoreClient {
...
 
   constructor(
+    public readonly onTimeToFirstByte: TimeToFirstByteCallback,
     private credentials: CredentialsProvider,
     /**
      * Asynchronous queue responsible for all of our internal processing. When
      * we get incoming work from the user (via public API) or the network
      * (incoming GRPC messages), we should always schedule onto this queue.
-     * This ensures all of our work is properly serialized (e.g. we don't
+     * This Fes all of our work is properly serialized (e.g. we don't
```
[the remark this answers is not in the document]
rafikhan: Whoops!

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -32,6 +32,54 @@ const describeFn =
...
+    );
+
+    // Test data
+    const payload = {
+      database: 'projects/' + projectId + '/databases/(default)',
+      addTarget: {
+        query: {
+          parent: 'projects/' + projectId + '/databases/(default)',
+          structuredQuery: {
+            from: [{ collectionId: 'WHAT_COLLECTION_ID_SHOULD_THIS_USE' }]
```
rafikhan, line 51: @schmidt-sebastian - The stats event doesn't get raised if there's isn't a successful message after open. I need a payload that will succeed. Can you recommend anything?

packages/firestore/test/integration/api_internal/standalone.browser.test.ts
```diff
@@ -0,0 +1,37 @@
...
+        });
+
+        try
+        {
+            await fs.collection('users').doc('foo@bar.com').get();
+        }
+        catch(e)
+        {
+            console.error(e);
+        }
```
schmidt-sebastian, line 31: Please remove.

packages/firestore/test/integration/api_internal/standalone.browser.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        const stats = await deferred.promise;
+
+        expect(stats).to.be.ok;
+    });
+});
```
schmidt-sebastian, line 37: Please add newline.

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -32,6 +32,54 @@ const describeFn =
       xdescribe;
 
 describeFn('WebChannel', () => {
+  it.only('receives error messages', done => {
+    const projectId = DEFAULT_PROJECT_ID;
```
schmidt-sebastian, line 36: Please address.

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -32,6 +32,54 @@ const describeFn =
...
+    stream.onMessage(() => {
+      // Shut down the stream since we don't care about processing any more messages
+      stream.close();
+    });
+
+    // Ensure Time to first byte was raised 
+    let timeToFirstByteData:{ type: number, timeToFirstByteMs: number } |null = null;
+
+    stream.onTimeToFirstByte((type: number, timeToFirstByteMs: number) => {
+      // console.log(JSON.stringify(data, null, 3));
```
schmidt-sebastian, line 71: Please remove.

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -32,6 +32,54 @@ const describeFn =
...
+      stream.send(payload);
+    });
+
+    stream.onMessage(() => {
+      // Shut down the stream since we don't care about processing any more messages
+      stream.close();
+    });
+
+    // Ensure Time to first byte was raised 
+    let timeToFirstByteData:{ type: number, timeToFirstByteMs: number } |null = null;
```
schmidt-sebastian, line 68:
```suggestion
    let timeToFirstByteData:{ type: number, timeToFirstByteMs: number };
```

For brevity (if it compiles - I am not sure right now).

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -60,6 +108,7 @@ describeFn('WebChannel', () => {
       stream.send(payload);
     });
 
+
```
schmidt-sebastian, line 111: Please remove.

packages/firestore/test/integration/util/settings.ts
```diff
@@ -39,7 +39,8 @@ const EMULATOR_FIRESTORE_SETTING = {
 
 const PROD_FIRESTORE_SETTING = {
   host: 'firestore.googleapis.com',
-  ssl: true
+  ssl: true,
+  experimentalAutoDetectLongPolling: true
```
schmidt-sebastian, line 43: Please revert until we make this the default setting.

packages/firestore/test/integration/browser/webchannel.test.ts
schmidt-sebastian: The correct response to this message is https://source.corp.google.com/piper///depot/google3/google/firestore/v1beta1/firestore.proto;l=932?q=firestore.proto

This test should be mocking the WebChannel connection though, so I am not convinced that we need any actual message here. You should be able to invoke the callbacks in the right order even without sending a message.

packages/firestore/exp/src/api/database.ts
```diff
@@ -81,7 +82,8 @@ export class FirebaseFirestore extends LiteFirestore {
   /** @hideconstructor */
   constructor(
     databaseIdOrApp: DatabaseId | FirebaseApp,
-    authProvider: Provider<FirebaseAuthInternalName>
+    authProvider: Provider<FirebaseAuthInternalName>,
+    public onTimeToFirstByteCallback: TimeToFirstByteCallback | null = null
```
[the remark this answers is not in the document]
rafikhan: I cleaned up the names, made it readonly internal and went with undefined.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -394,6 +397,10 @@ export abstract class PersistentStream<
    */
   protected abstract onMessage(message: ReceiveType): Promise<void>;
 
+  public onTimeToFirstByte(type: number, timeToFirstByteMs: number): void {
+    this.listener.onTimeToFirstByte(type, timeToFirstByteMs);
```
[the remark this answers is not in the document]
rafikhan: Done

packages/firestore/src/remote/remote_store.ts
```diff
@@ -744,6 +744,14 @@ async function onWriteStreamOpen(
   ensureWriteStream(remoteStoreImpl).writeHandshake();
 }
 
+function onTimeToFirstByteAvailable(
```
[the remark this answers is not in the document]
rafikhan: Done

packages/firestore/src/remote/remote_syncer.ts
```diff
@@ -72,4 +72,6 @@ export interface RemoteSyncer {
    * May be called repeatedly for the same user.
    */
   handleCredentialChange?(user: User): Promise<void>;
+
+  handleTimeToFirstByte?(type: number, timeToFirstByteMs: number): void;
```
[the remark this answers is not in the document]
rafikhan: Done

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -356,11 +359,28 @@ export class WebChannelConnection extends RestConnection {
...
       if (event.stat === Stat.PROXY) {
         logDebug(LOG_TAG, 'Detected buffering proxy');
+        statEvent = event;
       } else if (event.stat === Stat.NOPROXY) {
         logDebug(LOG_TAG, 'Detected no buffering proxy');
+        statEvent = event;
+      }
+
+      if (statEvent !== null && !hasRaisedTimeToFirstByteEvent) {
+        hasRaisedTimeToFirstByteEvent = true;
```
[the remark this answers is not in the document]
rafikhan: Technically we can get the desired result exclusively with the if check in sync engine but then it looks weird b/c this component is frequently calling the onTimeToFirstByte and the elapsed time is going to keep getting bigger and bigger. It would get swallowed after the first invocation in sync engine but could make a future maintainer wonder what the intent was. Perhaps I should leave a comment instead?

packages/firestore/src/core/sync_engine.ts
```diff
@@ -1509,6 +1514,14 @@ function ensureWatchCallbacks(syncEngine: SyncEngine): SyncEngineImpl {
     null,
     syncEngineImpl
   );
+
+  syncEngineImpl.remoteStore.remoteSyncer.handleTimeToFirstByte = (type: number, timeToFirstByteMs: number) => {
+    if (!syncEngineImpl.timeToFirstByteRaised && !!syncEngineImpl.timeToFirstByte) {
+      syncEngineImpl.timeToFirstByteRaised = true;
```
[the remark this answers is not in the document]
rafikhan: Noted.

rafikhan (commented):
@schmidt-sebastian - I've addressed some of the feedback and still have a bunch of stuff to update. I have some questions for you.

packages/firestore/test/integration/browser/webchannel.test.ts
rafikhan: Originally it didn't compile which is why I have it there.

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -356,11 +359,28 @@ export class WebChannelConnection extends RestConnection {
...
+
+        const onStatEventRaised = +new Date();
+        const elapsedMs = onStatEventRaised! - onOpenCalled!;
+ 
+        streamBridge.callOnTimeToFirstByte(
+          statEvent.stat,
```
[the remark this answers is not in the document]
rafikhan: I've switched it to use a boolean.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -461,6 +468,15 @@ export abstract class PersistentStream<
...
       });
     });
+    this.stream.onTimeToFirstByte((type: number, timeToFirstByteMs: number) => {
+      dispatchIfNotClosed(
+        () =>
+          new Promise(resolve => {
+            this.onTimeToFirstByte(type, timeToFirstByteMs);
+            resolve();
+          })
+      );
```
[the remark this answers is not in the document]
schmidt-sebastian: It's definitely worth discussing, but in this case we are saving quite a bit of written code.

packages/firestore/test/integration/api_internal/standalone.browser.test.ts
schmidt-sebastian: The console statements and the catch block (which should not be needed)

packages/firestore/src/core/component_provider.ts
```diff
@@ -77,6 +78,7 @@ export interface ComponentConfiguration {
   clientId: ClientId;
   initialUser: User;
   maxConcurrentLimboResolutions: number;
+  timeToFirstByte?:TimeToFirstByteCallback,
```
schmidt-sebastian, line 81: I think your `prettier` commit hook is not firing. There should be a space after the `:` and a semicolon instead of the comma.

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -241,11 +241,15 @@ export class WebChannelConnection extends RestConnection {
     // on a closed stream
     let closed = false;
 
+    let onOpenTimestamp: Date | null = null;
+    onOpenTimestamp = new Date();
```
schmidt-sebastian, line 245: Can we make `onOpenTimestamp` non-nullable now by combining these two lines?

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -241,11 +241,15 @@ export class WebChannelConnection extends RestConnection {
...
+
     const streamBridge = new StreamBridge<Req, Resp>({
       sendFn: (msg: Req) => {
         if (!closed) {
           if (!opened) {
             logDebug(LOG_TAG, 'Opening WebChannel transport.');
+            // onOpenTimestamp = new Date();  // This isn't working anymore. Hmm...
```
schmidt-sebastian, line 252: I think the difference in timing here is negligible. Line 252 would fire very quickly after line 245. We always send a message one we open the stream. The only exception is possibly a user adding and removing a snapshot listener in quick succession, which might cause us to open a stream and then deciding that we don't need it. I would have to convince myself that this is an actual problem and am mostly indifferent on whether we set the timer here or earlier.

packages/firestore/src/platform/browser/webchannel_connection.ts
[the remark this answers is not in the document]
schmidt-sebastian: You could solve this by moving the call of `callOnTimeToFirstByte` to the if statements below. You will no longer need to check for `hasRaisedTimeToFirstByteEvent` and won't need to set it either. You can also remove `isLongPollingConnection:boolean `. Plus, if you initialize onOpenTimestamp to `new Date()` it will also not be nullable and you don't need to perform a debug check.

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -356,11 +360,32 @@ export class WebChannelConnection extends RestConnection {
...
+        isLongPollingConnection = true;
       } else if (event.stat === Stat.NOPROXY) {
         logDebug(LOG_TAG, 'Detected no buffering proxy');
+        isLongPollingConnection = false;
+      }
+
+      if (isLongPollingConnection !== null && !hasRaisedTimeToFirstByteEvent) {
+        hasRaisedTimeToFirstByteEvent = true;
+
+        if (onOpenTimestamp === null) {
```
schmidt-sebastian, line 378: If you anticipate that this check never fires, then you should make it a `debugAssert`, which removes overhead from the production build. If you anticipate that it might crash, then you need to handle it in a way that doesn't cause the SDK to crash.

FWIW, I think this check breaks your code in 252, since it will be validated for a CONNECT_ATTEMPT, which fires before any other stat event.

packages/firestore/src/remote/persistent_stream.ts
```diff
@@ -127,6 +128,8 @@ export interface PersistentStreamListener {
    * FirestoreError will be set.
    */
   onClose: (err?: FirestoreError) => Promise<void>;
+
+  onTimeToFirstByte: TimeToFirstByteCallback;
```
schmidt-sebastian, line 132: Please add JSDoc similar to the lines above.

schmidt-sebastian (commented):
Much cleaner. Thanks for the initial round of updates

packages/firestore/test/integration/browser/webchannel.test.ts
rafikhan: > The correct response to this message is...

I'm not sure what to do w/ the proto. If there's an error in web channel the stats event under time to first byte never gets raised. I either need to pass in the correct collection id (which I don't know how to get) or make a request that is guaranteed to succeed. If I'm not being clear perhaps we can clear this up over GVC?

> This test should be mocking the WebChannel connection ...

Are you sure about that? The file is listed as a browser integration test and when I turn on debug logging I'm seeing network traffic. If this is supposed to be mocked I can't tell where.

rafikhan (commented):
Updated with additional cleanup and more questions.

packages/firestore/src/core/component_provider.ts
schmidt-sebastian: It's supposed to run as a pre-commit hook. It would suggest merging master and running `yarn` at the repo root.

packages/firestore/test/integration/browser/webchannel.test.ts
schmidt-sebastian: Every time I look at this test I think it is a unit tests. But it turns out I am wrong every time. 

You should be able to use any collection ID - they don't have to exist to be queried. Do you get any responses? My bet would be that the project ID is wrong.

rafikhan: I pushed my code with the default project id and settings and it's failing w/ an internal assertion error on CI. Not sure if this is related or not.

packages/firestore/src/platform/browser/webchannel_connection.ts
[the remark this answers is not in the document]
schmidt-sebastian: I suggest something like this: https://gist.github.com/schmidt-sebastian/8faadec621a59b214b6cd01247de605a

packages/firestore/test/integration/browser/webchannel.test.ts
schmidt-sebastian: Here is a patch: https://gist.github.com/schmidt-sebastian/302d74c42d81117eb129fa34731caddf (includes the diff above).

The main issues were:
- The 'detectBufferingProxy' flag wasn't set
- The onMessage/onStat event were racing. If onMessage was executed before, it would invoke `close()` which suppressed the stat event.

With those changes, I can get the test to pass.

schmidt-sebastian (commented):
This looks very clean now. I played a bit with the test and was able to get it to pass.

packages/firestore/src/platform/browser/webchannel_connection.ts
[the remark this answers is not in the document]
rafikhan: Got it. This makes sense and is more compact. Thank you.

packages/firestore/test/integration/browser/webchannel.test.ts
rafikhan: Thanks!

packages/firestore/test/integration/browser/webchannel.test.ts
```diff
@@ -33,41 +45,64 @@ const describeFn =
...
     // Wait until we receive data, then send a bad "addTarget" request, causing
     // the stream to be closed with an error. In this case, bad means having a
     // different database ID.
     stream.onMessage(msg => {
       if (msg.targetChange) {
-        payload.database = 'projects/some-other-project-id/databases/(default)';
         didSendBadPayload = true;
+        const payload = { ...ADD_TARGET };
+        payload.database =
+          ' projects/some-other-project-id/databases/(default)';
```
schmidt-sebastian, line 105: Remove space

schmidt-sebastian (commented):
Whee! Looks great!

packages/firestore/test/integration/browser/webchannel.test.ts
rafikhan: Sure. FYI, your original code snippet w/ the diff had this in there.

schmidt-sebastian: It was a test to see if your standards are higher than mine.

schmidt-sebastian (approved):

PR #4185, opened 2020-12-08
Storage-exp: fix doc comments and remove namespace imports.

hsubox76:
- Remove namespace imports in all src files and in some test files (not needed but for consistency).
- Fix up comment tags for api-extractor.

Was not sure about the correct way to resolve some of the warnings from API Extractor, such as, it asked to export a lot of internal symbols from index.ts. They are marked `@internal` and won't show up in the documentation or public d.ts, but is it correct to export them at all? Is there another way to get around that?

Will run the actual doc generation in a separate PR.


packages/storage/exp/index.ts
```diff
@@ -53,7 +53,36 @@ export { ListOptions, ListResult } from '../src/list';
...
+  RequestInfo as _RequestInfo,
+  UrlParams as _UrlParams
+} from '../src/implementation/requestinfo';
+export { Request as _Request } from '../src/implementation/request';
+export { Location as _Location } from '../src/implementation/location';
+export { XhrIoPool as _XhrIoPool } from '../src/implementation/xhriopool';
+export {
+  TaskEvent,
+  TaskState,
+  InternalTaskState as _InternalTaskState
```
Feiyang1, line 66: We should not export internal things. I commented some of them out and api extractor continues to work without giving any warnings. 

If I remember correctly, api extractor only asks you to export an symbol if it is used in a exported function/class/etc., but itself not being exported. It means we are referencing internal symbols in the public interface, which we should fix.
What is the specific warning that asks you to export internal symbols?

hsubox76: I removed all the internal exports and ran api-report again, you can view the md file now to see the warning comments, an example is:

```
// Warning: (ae-forgotten-export) The symbol "Request" needs to be exported by the entry point index.d.ts
```

common/api-review/storage.api.md
```diff
@@ -35,85 +55,110 @@ export interface ListOptions extends types.ListOptions {
...
         [key: string]: string;
     } | undefined;
-    // (undocumented)
     downloadTokens: string[] | undefined;
-    // (undocumented)
     md5Hash: string | undefined;
-    // (undocumented)
     ref: StorageReference | undefined;
-    // (undocumented)
+    // @internal (undocumented)
```
schmidt-sebastian, line 78: FWIW, this property doesn't seem to be used in the SDK.

common/api-review/storage.api.md
```diff
@@ -35,85 +55,110 @@ export interface ListOptions extends types.ListOptions {
...
     get fullPath(): string;
     // Warning: (ae-forgotten-export) The symbol "Location" needs to be exported by the entry point index.d.ts
     //
     // @internal (undocumented)
     _location: Location_2;
-    // (undocumented)
     get name(): string;
-    // (undocumented)
+    // @internal (undocumented)
     protected newRef(service: StorageService, location: Location_2): StorageReference;
```
schmidt-sebastian, line 141: s/newRef/_newRef

common/api-review/storage.api.md
```diff
@@ -123,18 +168,17 @@ export class StorageService implements _FirebaseService {
     _delete(): Promise<void>;
     // @internal (undocumented)
     readonly _firebaseVersion?: string | undefined;
-    // (undocumented)
+    // @internal (undocumented)
     getAuthToken(): Promise<string | null>;
```
schmidt-sebastian, line 172: s/getAuthToken/_ getAuthToken

common/api-review/storage.api.md
```diff
@@ -123,18 +168,17 @@ export class StorageService implements _FirebaseService {
...
     // @internal (undocumented)
     readonly _firebaseVersion?: string | undefined;
-    // (undocumented)
+    // @internal (undocumented)
     getAuthToken(): Promise<string | null>;
     // Warning: (ae-forgotten-export) The symbol "RequestInfo" needs to be exported by the entry point index.d.ts
     // Warning: (ae-forgotten-export) The symbol "Request" needs to be exported by the entry point index.d.ts
     //
     // @internal (undocumented)
     makeRequest<T>(requestInfo: RequestInfo_2<T>, authToken: string | null): Request_2<T>;
```
schmidt-sebastian, line 177: s/makeRequest/_makeRequest and so on.

BTW, the "prune-dts" script enforces the underscore prefix convention and gets rid of the `@internal` tags.

packages/storage/package.json
```diff
@@ -15,6 +15,7 @@
     "lint:fix": "eslint --fix -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "build": "rollup -c rollup.config.compat.js",
     "build:exp": "rollup -c rollup.config.exp.js",
+    "build:exp:release": "yarn build:exp && yarn api-report",
```
schmidt-sebastian, line 18: Don't we always want to build the API report to ensure that we don't accidentally change the public API?

packages/storage/src/implementation/error.ts
```diff
@@ -42,6 +62,9 @@ export class FirebaseStorageError extends FirebaseError {
     }
   }
 
+  /**
+   * Optional serverResponse string that can be added to a server error.
```
schmidt-sebastian, line 66:
```suggestion
   * Optional response message that was added by the server.
```

packages/storage/src/implementation/observer.ts
```diff
@@ -14,26 +14,66 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as type from './type';
+import { isFunction } from './type';
 import { FirebaseStorageError } from './error';
 
+/**
+ * @public
+ * Called once for each value in a stream of values.
```
schmidt-sebastian, line 22: Suggestion: Function that is called once ...

Same for ErrorFn and CompleteFn below.

packages/storage/src/implementation/observer.ts
```diff
@@ -14,26 +14,66 @@
...
+
+/**
+ * @public
+ * No events will be sent to next() once complete() is called.
+ */
 export type CompleteFn = () => void;
+
+/**
+ * @public
+ * Unsubscribes to a stream.
```
schmidt-sebastian, line 40: s/to/from

packages/storage/src/implementation/taskenums.ts
```diff
@@ -20,18 +20,40 @@
  */
 
 /**
- * Enum for task events.
+ * An event that is triggered on a task.
+ * @public
  */
 export type TaskEvent = string;
+/**
```
schmidt-sebastian, line 27: Missing empty line above.

packages/storage/src/implementation/taskenums.ts
```diff
@@ -20,18 +20,40 @@
...
+ */
 export const TaskEvent = {
-  /** Triggered whenever the task changes or progress is updated. */
+  /**
+   * For this event,
+   * <ul>
+   *   <li>The `next` function is triggered on progress updates and when the
+   *       task is paused/resumed with a
+   *       `UploadTaskSnapshot` as the first
+   *       argument.</li>
```
schmidt-sebastian, line 38: Nit: remove extra line breaks.

packages/storage/src/implementation/taskenums.ts
```diff
@@ -20,18 +20,40 @@
...
+   */
   STATE_CHANGED: 'state_changed'
 };
 
 /**
  * Internal enum for task state.
+ * @internal
  */
 export type InternalTaskState = string;
+/**
```
schmidt-sebastian, line 53: Nit: Missing empty line above.

packages/storage/src/implementation/taskenums.ts
```diff
@@ -43,9 +65,14 @@ export const InternalTaskState = {
 };
 
 /**
- * External (API-surfaced) enum for task state.
+ * Represents the current state of a running upload.
+ * @public
  */
 export type TaskState = string;
+/**
```
schmidt-sebastian, line 72: Nit: Missing empty line above.

packages/storage/src/list.ts
```diff
@@ -33,7 +33,22 @@ export interface ListOptions extends types.ListOptions {}
...
+  /**
+   * References to prefixes (sub-folders). You can call list() on them to
+   * get its contents.
+   *
+   * Folders are implicit based on '/' in the object paths.
+   * For example, if a bucket has two objects '/a/b/1' and '/a/b/2', list('/a')
+   * will return '/a/b' as a prefix.
+   */
   prefixes: StorageReference[];
+  /**
```
schmidt-sebastian, line 45: Nit: Missing empty line above.

packages/storage/src/list.ts
```diff
@@ -33,7 +33,22 @@ export interface ListOptions extends types.ListOptions {}
...
+   * Objects in this directory.
+   * You can call getMetadata() and getDownloadUrl() on them.
+   */
   items: StorageReference[];
+  /**
```
schmidt-sebastian, line 50: Nit: Missing empty line above.

packages/storage/src/metadata.ts
```diff
@@ -27,15 +27,45 @@ import { StorageReference } from './reference';
  * @public
  */
 interface Metadata extends types.FullMetadata {
+  /**
+   * @internal
+   */
   type: string | undefined;
+  /**
```
schmidt-sebastian, line 34: Nit: Missing empty line above (here and everywhere)

packages/storage/src/metadata.ts
```diff
@@ -27,15 +27,45 @@ import { StorageReference } from './reference';
  * @public
  */
 interface Metadata extends types.FullMetadata {
+  /**
+   * @internal
+   */
   type: string | undefined;
```
schmidt-sebastian, line 33: As stated above, this doesn't seem to be used.

packages/storage/src/tasksnapshot.ts
```diff
@@ -24,7 +24,13 @@ import { UploadTask } from './task';
  * @public
  */
 export interface UploadResult {
+  /**
+   * Contains the metadata sent back from the server.
+   */
   readonly metadata: Metadata;
+  /**
```
schmidt-sebastian, line 31: This file is also missing some empty lines. You can push back if you want :)

packages/storage/src/implementation/error.ts
hsubox76: Done.

common/api-review/storage.api.md
hsubox76: It gets set here. I'm not sure what the backend does or doesn't do with it but I don't want to mess with it.

https://github.com/firebase/firebase-js-sdk/blob/b95734c175ce87dc704e55466f852601b972f799/packages/storage/src/implementation/metadata.ts#L130

packages/storage/src/implementation/observer.ts
hsubox76: Done.

packages/storage/package.json
hsubox76: Probably, changed.

common/api-review/storage.api.md
hsubox76: Done.

common/api-review/storage.api.md
hsubox76: Done.

packages/storage/src/tasksnapshot.ts
hsubox76: I was going to say if we are going to enforce it we should probably set Prettier to do it but it seems like it can't.

packages/storage/src/implementation/observer.ts
hsubox76: Done.

common/api-review/storage.api.md
hsubox76: Added it as an extra step after api-extractor creates storage-public.d.ts, but there seems to be a little bug that cuts off the first comment (on CompleteFn). Can probably fix it later.

packages/storage/exp/index.ts
Feiyang1: Right. These symbols are referenced in the public classes/interfaces/functions, so they need to be exported even if the method that references them are annotated as `internal`. Here, `internal` has a slightly different meaning than just "hidden from public". It is designed to be used in a project with many packages that depend on each other. When a symbol is annotated as `internal` in a package, it means the symbol will be used in other packages in the same project, but not for public use. 

It means that api-extractor will generate 2 different d.ts files:
- a dev version of the d.ts for local development, which has all public APIs +  internal APIs/types (annotated as `internal`), which other internal packages can see.
- a public version of the d.ts, where all `internal` typings are stripped
`@firebase/app-exp` uses this feature to make, for example, `_registerComponent()` available for other firebase packages, but hide it from developers.

Since both d.ts are generated from the same source file, you need to export everything including the internal ones, so the warning makes sense.
In your case, you don't really need the `internal` feature because no other internal packages need to access any non-public APIs from storage.

What we should do is create a public interface for every implementation classes (e.g. StorageService, Reference, etc) in the storage-types package, and only reference these public interfaces in the public APIs. This way, you don't have to annotate things with `internal` in the implementation classes and there should be no warning to ask you to export internal things.

schmidt-sebastian: > What we should do is create a public interface for every implementation classes (e.g. StorageService, Reference, etc) in the storage-types package, and only reference these public interfaces in the public APIs. This way, you don't have to annotate things with `internal` in the implementation classes and there should be no warning to ask you to export internal things.

Does that not contradict our plan to move away from manual types? I thought of one the goals of API Extractor was to not have to do this.

Feiyang1: Not really. We want to have consistent typings by getting rid of the manually curated `firebase/index.d.ts`, so there is a single source of truth for the public APIs for every SDK, which is `@firebase/foo` + `@firebase/foo-types`. `@firebase/foo` defines what gets exported and `@firebase/foo-types` defines the public types, e.g. parameter types and return types.

API extractor can help us get rid of `@firebase/foo-types` as well as manually defined public types via its release tag feature, but I think we want to keep `@firebase/foo-types` to allow people to depend on our public types without depending on the actual implementation, which is, IIUC, why we structured our packages this way back in the day.

hsubox76: Just to be clear before I go through all the methods, I should be doing something like this for every public method? (doing it for `uploadBytes()` in `packages/storage/exp/index.ts`).

```
import * as types from '@firebase/storage-types/exp';
import { Metadata } from '../src/metadata';
import { UploadResult } from '../src/tasksnapshot';
import {
  uploadBytes as uploadBytesInternal,
  Reference
} from '../src/reference';

export function uploadBytes(
  ref: types.StorageReference,
  data: Blob | Uint8Array | ArrayBuffer,
  metadata: types.Metadata
): Promise<UploadResult> {
  return uploadBytesInternal(ref as Reference, data, metadata as Metadata);
}
```

Seems like a lot of boilerplate to repeat, but if it's the right way to do things, I'm sure it needs to be done.

Also is packages/storage-types/exp/ folder the best place to put it? Or should I make a packages-exp/storage-types-exp? In the first case, will I have to do something special to make sure exp users get the package since @firebase/storage-types doesn't normally get published as part of exp release?

Feiyang1: Exactly. Please make sure the implementations actually implement these interfaces to keep them in sync. Storage-types/exp sounds good to me, just make sure the new file will be included when publishing.

hsubox76: I've done this and everything seems fine on the code side, the api-extractor md file looks clean, but the problem is that when I generate the reference docs, all the types from index.d.ts (like types.Metadata) are dead-ends and do not have a link with details. For example, the reference doc page for updateMetadata(), a function which has types.Metadata as a param and return type, should be linking to a separate page documenting the Metadata interface, I think. But if the type was imported from @firebase/storage-types/exp it's just a text link, the user cannot find any info on Metadata's param. Same with UploadTask which is even more important I think.

Feiyang1: These types should link to @firebase/storage-types/exp correctly when we generate reference docs for all packages together. See the return type of `getApp()` which links to a page in `@firebase/app-types`.

You need to create a api-extractor script in `@firebase/storage-types`, then run `yarn docgen:exp` at repo root. `docgen:exp` will call this script in all projects and copy their output to a folder and generate reference docs for all of them with linking across packages. You also need to make sure the package names match in the api-extractor output, so linking works. There might be some special handling you need to do to get matching package names, since you are rewriting it in place.

packages/storage/src/reference.ts
```diff
@@ -52,12 +53,15 @@ import { UploadResult } from './tasksnapshot';
...
  */
-export class StorageReference {
+export class Reference {
   /**
    * @internal
    */
   _location: Location;
 
+  /**
+   * @internal
```
Feiyang1, line 63: Do we need any of these annotations for all implementation classes? I think we don't since our public API references only types instead of these concrete classes.

common/api-review/storage.api.md
```diff
@@ -5,144 +5,92 @@
 ```ts
 
-import { FirebaseAuthInternalName } from '@firebase/auth-interop-types';
 import { FirebaseError } from '@firebase/util';
-import { _FirebaseService } from '@firebase/app-types';
-import { Provider } from '@firebase/component';
-import * as types from '@firebase/storage-types';
+import * as types from '@firebase/storage-types/exp';
```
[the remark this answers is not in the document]
hsubox76: Fixed.

packages/storage-types/exp/index.d.ts
```diff
@@ -0,0 +1,157 @@
...
+  task: UploadTask;
+  totalBytes: number;
+}
+
+export interface UploadResult {
+  readonly metadata: Metadata;
+  readonly ref: StorageReference;
+}
+
+export class FirebaseStorage {
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/storage-types/exp/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@firebase/storage-types-exp",
```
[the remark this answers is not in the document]
hsubox76: Oh, removed.

packages/storage/src/implementation/error.ts
```diff
@@ -53,8 +76,17 @@ export class FirebaseStorageError extends FirebaseError {
...
+ * @public
+ * Error codes that can be attached to `FirebaseStorageError`s.
+ */
+export type StorageErrorCode = string;
+
+/**
+ * @public
+ * Error codes that can be attached to `FirebaseStorageError`s.
+ */
+export const StorageErrorCode = {
```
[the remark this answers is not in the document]
hsubox76: Doesn't have to be public, changed exported `FirebaseStorageError` to export the public type to avoid requiring this. I made it an enum but if it's a const enum I can't import it into other files, and it's used a constant in equality checks in task.ts for example.

packages/storage/src/implementation/observer.ts
```diff
@@ -14,26 +14,67 @@
...
+
+/**
+ * @public
+ * Unsubscribes from a stream.
+ */
 export type Unsubscribe = () => void;
 
+/**
+ * @public
+ * An observer that wraps any errors in `FirebaseStorageError`.
```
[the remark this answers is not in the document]
hsubox76: I changed exp/index to export the public `Observer` type which should have the public comment. This is now an internal comment just for us to explain the only reason we need a special StorageObserver is that it wraps errors in `FirebaseStorageError` instead of a generic `Error`.

packages/storage/src/implementation/observer.ts
```diff
@@ -14,26 +14,67 @@
...
+ * Subscribes to an event stream.
+ */
 export type Subscribe<T> = (
   next?: NextFn<T> | StorageObserver<T>,
   error?: ErrorFn,
   complete?: CompleteFn
 ) => Unsubscribe;
 
+/**
+ * @internal
```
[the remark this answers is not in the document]
hsubox76: Removed all tags from this file since we are exporting the public types instead.

packages/storage/src/implementation/taskenums.ts
```diff
@@ -20,18 +20,41 @@
...
 
 /**
  * Internal enum for task state.
+ * @internal
  */
 export type InternalTaskState = string;
+
+/**
+ * Internal enum for task state.
+ * @internal
```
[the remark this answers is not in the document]
hsubox76: It could be an enum but not const enum because it's imported elsewhere. It seems like you can't export const enums for use in another file but let me know if I'm just missing something.

I removed all tags in this file since there are public types for them.

packages/storage/src/reference.ts
hsubox76: Removed all the `@internal` tags that are not needed anymore after switching to public types.

packages/storage/src/implementation/error.ts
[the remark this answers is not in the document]
Feiyang1: You can import const enum in other files in the same package and use it as values. it's that you can't import it from another package. The equality check should also work with const enum.

packages/storage/src/implementation/observer.ts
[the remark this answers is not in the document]
Feiyang1: I still don't understand it. What does it mean to have observer wrap errors in `FirebaseStorageError`?

packages/storage/src/implementation/taskenums.ts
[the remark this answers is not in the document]
Feiyang1: See my comment earlier.

packages/storage/src/implementation/observer.ts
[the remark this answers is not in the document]
hsubox76: I tried to update the comment to be more clear.

General Observer type that you can get from @firebase/util is roughly
```
Observer<T> {
  next?: (value: T) => void;
  error?: (error: Error) => void;
  complete?: () => void;
}
```
In a `StorageObserver` type the error function parameter is a `FirebaseStorageError` instead of `Error`.

packages/storage/src/implementation/error.ts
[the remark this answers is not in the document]
hsubox76: I was getting build errors and finally figured it out, you can export const enums but not from the same file as code? Functions? I'm not sure why. So I moved it to `constants.ts` and it works now. Is that the right solution?

packages/storage/src/implementation/taskenums.ts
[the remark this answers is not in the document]
hsubox76: I was able to convert all 3 enums to const enum (the public ones have a string literal type in storage-types/exp so we don't need to export the ones here) but I had to move the function `taskStateFromInternalTaskState` out of the file (I moved it to `task.ts`, the only file that uses it).

packages/storage/src/implementation/observer.ts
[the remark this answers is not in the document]
Feiyang1: Ah I see

packages/storage/src/implementation/error.ts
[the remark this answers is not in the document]
Feiyang1: Weird. I moved it back to error.ts, and I didn't get any build error.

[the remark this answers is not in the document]
hsubox76: Works in its original place after clean build, restored.

packages/storage/src/implementation/taskenums.ts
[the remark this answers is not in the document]
hsubox76: Restored, same as StorageErrorCode.

Feiyang1 (approved):

Issue #4175, opened 2020-12-06
Firestore query uses stale token after forced refresh


jadengis:
### [REQUIRED] Describe your environment

  * Operating System version: Ubuntu 20.04
  * Browser version: Chrome, Firefox both latest
  * Firebase SDK version: 8.1.1
  * Firebase Product: auth, firestore

#### Steps to reproduce:

In v8.0.0, I am able to perform the following sequence of operations successfully

- Add a custom claim to a user's token in the backend using admin sdk
- Force refresh token on the client side using `user.getIdTokenResult(true)`
- Immediately perform a query that requires that custom claim in the security rules 

In v8.1.1, the above sequence fails when trying to make the query that requires the new claim. I logged the result of `user.getIdTokenResult(true)` to the console and confirmed that it did in fact has the new claim, which lead me to believe that the firestore sdk is using a stale token. Also after I refresh the page, the query starts working.

#### Relevant Code:

I'm not sure how to create a complete reproduction, as this sequence involves the admin sdk in the backend to create the custom claim, but I will include some code snippets.

I have a firestore rule that looks like this:
```
match /listings/{listingId} {
      // store_id is a custom claim added to the users token
      allow read: if return resource.data.state == "available" || resource.data.store.id == request.auth.token.store_id;
}
```

In the backend I am running the follow code in an http request to create a custom claim on my user:
```javascript
import admin from "firebase-admin";

...

admin.auth().setCustomUserClaims(userId, {store_id: storeId});
```

In the front end I have code that does the following in response to the above request.
```javascript
import firebase from "firebase/app";

...

addStoreIdToUserToken(storeId)
    .then(() => firebase.auth().currentUser)
    .then(user => user.getIdTokenResult(true))
    .then(token => console.log(token.claims)); // The log here shows that the new token has the new claim
```

After the above chain of promises I make a query like this:
```javascript
import firebase from "firebase/app";

// storeId is the same as the above code snippet
firebase.firestore().collection("listings").where("store.id", "==", storeId).get();
```

In version 8.0.0, the above code works, but in 8.1.1 is get a rules violation for the rule snippet above. This leads me to believe firestore is using a stale token without the new claim, even though I forced a refresh.

Thanks for looking into these, please let me know if you require any additional information.

scottfr:
We're experience the same issue in 8.1.2. Code using this pattern worked in 7.24.0, but fails in 8.1.2.

[...]

PR #4189, opened 2020-12-10
Invoke RemoteStore's credential change listener even when user didn't change

schmidt-sebastian:
This fixes an issue in the 8.0.2+ Firestore SDK, which only propagates user changes but not credential changes. This means that our WebChannel connection did not get reinitialized if a user ID didn't change, but the user's permission did.

If we get a user change notification, the correct behavior is:
- Invoke the LocalStore's user change handler if the user changed. We should not invoke this handler if the user didn't change, because it is costly to reload the MutationQueue with IndexedDB and results in dataloss with memory persistence/
- Always invoke the RemoteStore's credential change handler, since custom claims could have changed.

Verified using Sam's repro. Unfortunately, we don't have any testing infrastructure that automatically exercises this code.

Fixes https://github.com/firebase/firebase-js-sdk/issues/4175


packages/firestore/src/core/firestore_client.ts
```diff
@@ -199,11 +197,15 @@ export async function setOfflineComponentProvider(
...
-    client.asyncQueue.enqueueRetryable(async () => {
-      await handleUserChange(offlineComponentProvider.localStore, user);
-    })
-  );
+  let currentUser = configuration.initialUser;
+  client.setCredentialChangeListener(user => {
+    if (!currentUser.isEqual(user)) {
+      currentUser = user;
+      client.asyncQueue.enqueueRetryable(async () => {
+        await handleUserChange(offlineComponentProvider.localStore, user);
```
schmidt-sebastian, line 205: I moved the user change guard for LocalStore's `handleUserChange` here. It has always been part of the SyncEngine's call into LocalStore (https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/sync_engine.ts#L1084). This part of the code handles the user change when SyncEngine is not yet loaded (for a fully offline client).

I do however think that the guard here doesn't make a huge difference and mostly added it to be technically correct. This line of code is only executed when memory persistence is used without network access. The only sensible use case for this is an offline Firestore experience that is backed by Bundles. Bundles don't contain any mutations, so the user change would always be a no-op here. Once the user adds the first mutation, SyncEngine gets loaded and overwrites this listener.

I probably didn't describe this very well. But at least I tried, right? :)

rafikhan: It's not unreasonable to have here. The change is queued vs being immediately applied and also provides a tiny bit more predictability to the system.

rafikhan (approved):
@schmidt-sebastian and I discussed the nature of the change offline.  LGTM.

PR #4194, opened 2020-12-11
Fix remote-config-exp installations import

hsubox76:
If this import is not in the same file that uses the code (that has the `getProvider()` call), this import gets tree-shaken out during the rollup build and installations fails to register. (The error won't happen if there's another library, like performance, bringing in installations.)

### DIFF
--- a/packages-exp/remote-config-exp/src/index.ts
+++ b/packages-exp/remote-config-exp/src/index.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import '@firebase/installations-exp';
 import { registerRemoteConfig } from './register';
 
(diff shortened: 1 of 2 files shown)


Feiyang1 (approved):
Interesting, but I'm not sure `getProvider()` decides whether or not the import is tree shaken because it's a side effect import and I don't think rollup is capable of knowing whether things in `@firebase/installations-exp` is used or not in the importing module. 

Actually I think `import @firebase/installations-exp` should always be removed regardless of its location because we set `moduleSideEffect: true` in the rollup config. It may be a bug in rollup, or something I don't fully understand. For now, the PR is good to go to unblock the releases.

PR #4197, opened 2020-12-11
Add database-exp

schmidt-sebastian:
This adds `getDatabase` and `initializeDatabase` to firebase-exp.


packages/database/src/exp/Database.ts
```diff
@@ -0,0 +1,155 @@
...
+ * with default settings if no instance exists or if the existing instance uses
+ * a custom database URL.
+ *
+ * @param app - The {@link FirebaseApp} instance that the returned Firestore
+ * instance is associated with.
+ * @returns The `Firestore` instance of the provided app.
+ */
+export function getDatabase(app: FirebaseApp): FirebaseDatabase {
+  return _getProvider(app, 'database-exp').getImmediate() as FirebaseDatabase;
+}
```
Feiyang1, line 155: Should we just have `getDatabase(app: FirebaseApp, url?: string)`? This also removes the potential confusion you mentioned in the chat. Having a second optional parameter in `getFoo()` is the standard way to support multiple instances. see functions - https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/functions-exp/src/api.ts#L43

schmidt-sebastian: That would work for me, but it seems inconsistent with Firebase Auth and Firestore. I can update this PR, but we should still talk about how we can consistently pass settings to our various SDKs. For example, I don't see us changing the `getFirestore()` call to accept setting since we wouldn't want users to specify the same settings over and over again. We also would likely not want to return a different instance if the settings are different.

Feiyang1: Agreed we need to standardize passing settings. My immediate proposal is use `getFoo(app, identifier?)` for getting instances with identifier and use `initializeFoo(app, identifier, settings)` for passing settings, which will throw if you try to initialize the same app + identifier twice.

schmidt-sebastian: For this to work, should I add the `identifier` as an argument to `_getProvider()` or do you have a different solution in mind?

Feiyang1: No need. `provider.get()` and `provider.getImmeidate()` already support `identifier`.

schmidt-sebastian: Awesome, thanks for the information. I updated the PR.

schmidt-sebastian:
@Feiyang1 Do you have time to look at this again? We would like to get this merged before the next EAP release.

packages/database/.idea/runConfigurations/All_Tests.xml
```diff
@@ -0,0 +1,17 @@
+<component name="ProjectRunConfigurationManager">
```
Feiyang1, line 1: What is it for?

packages/database/exp-types/index.d.ts
```diff
@@ -0,0 +1,147 @@
...
+export interface Database {
+  app: FirebaseApp;
+  useEmulator(host: string, port: number): void;
+  goOffline(): void;
+  goOnline(): void;
+  ref(path?: string | Reference): Reference;
+  refFromURL(url: string): Reference;
+}
+
+export class FirebaseDatabase implements Database {
```
Feiyang1, line 46: Why do we export the class?

packages/database/exp/index.ts
```diff
@@ -0,0 +1,61 @@
...
+function registerDatabase(): void {
+  _registerComponent(
+    new Component(
+      'database-exp',
+      (container, url) => {
+        const app = container.getProvider('app-exp').getImmediate()!;
+        return ((app, auth) => new FirebaseDatabase(app, auth, url))(
+          app,
+          container.getProvider('auth-internal')
+        );
```
Feiyang1, line 53: This looks weird. It reads much cleaner this way:
```
const app = container.getProvider('app-exp').getImmediate()!;
const authProvider = container.getProvider('auth-internal');
return new FirebaseDatabase(app, authProvider, url);
```

packages/database/package.json
```diff
@@ -7,11 +7,13 @@
   "browser": "dist/index.esm.js",
   "module": "dist/index.esm.js",
   "esm2017": "dist/index.esm2017.js",
-  "files": ["dist"],
+  "files": ["dist", "exp/package.json", "exp-types"],
```
Feiyang1, line 10: I don't think we want to include exp artifacts when publishing our prod package. If there is any special handling needed for the exp release, we will handle that in the exp release script similar to how firestore is handled.
```suggestion
  "files": ["dist"],
```

packages/database/src/exp/Database.ts
```diff
@@ -0,0 +1,143 @@
...
+ * is specified, the URL overrides the database URL provided by the Firebase
+ * App.
+ *
+ * @param app - The {@link FirebaseApp} with which the Realtime Database
+ * instance will be associated with.
+ * @param url - The URL of the Realtime Database instance to connect to. If not
+ * provided, the SDK connects to the default instance of the Firebase App.
+ * @returns A newly initialized `FirebaseDatabase` instance.
+ */
+export function initializeDatabase(
```
Feiyang1, line 117: Do we need it anymore? It's identical to `getDatabase()`

packages/database/.idea/runConfigurations/All_Tests.xml
schmidt-sebastian: This is a unit test run configuration for IntelliJ. We have these for Firestore too (https://github.com/firebase/firebase-js-sdk/tree/master/packages/firestore/.idea/runConfigurations)

packages/database/exp-types/index.d.ts
schmidt-sebastian: It matches the existing types, but we can delete this export.

packages/database/exp/index.ts
schmidt-sebastian: Done

packages/database/src/exp/Database.ts
schmidt-sebastian: I am okay with removing it, but I added it to allow consistency between the different APIs. Ideally, we would throw if `initializeDatabase` gets called multiple times or called after `getDatabase`.

packages/database/.idea/runConfigurations/All_Tests.xml
Feiyang1: Interesting.  Does anyone use IntelliJ to develop the JS SDK? isn't it a Java IDE?

packages/database/src/exp/Database.ts
Feiyang1: `initilizeFoo()` is only needed if you can pass configuration when initializing a SDK, which is not the case for Database. I'd remove it. We can add it back if there is a need in the future. WDYT?

packages/database/.idea/runConfigurations/All_Tests.xml
schmidt-sebastian: My entire team does. It's really good for TypeScript development and debugging.

packages/database/src/exp/Database.ts
schmidt-sebastian: Removed for now.

Feiyang1:
https://github.com/firebase/firebase-js-sdk/pull/4317 should fix the CI failure and includes a minor build fix.

Feiyang1 (approved):

PR #4199, opened 2020-12-11
Fix refFromUrl when used with emulator

schmidt-sebastian:
The unit test `refFromURL returns an emulated ref with useEmulator` fails for me locally. I think that we don't want to enforce the host check when the database URL is emulated.

Note that this test probably passes on CI since `localhost` gets used as the default host.

### DIFF
--- a/.changeset/clever-kangaroos-hug.md
+++ b/.changeset/clever-kangaroos-hug.md
@@ -1,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fixes an issue that caused `refFromUrl()` to reject production database URLs when `useEmulator()` was used.
--- a/packages/database/src/api/Database.ts
+++ b/packages/database/src/api/Database.ts
@@ -156,7 +156,7 @@ export class Database implements FirebaseService {
 
     const repoInfo = parsedURL.repoInfo;
     if (
-      !repoInfo.isCustomHost() &&
+      !this.repo_.repoInfo_.isCustomHost() &&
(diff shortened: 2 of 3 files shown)


schmidt-sebastian:
FYI It looks like this may be missing from Android - https://github.com/firebase/firebase-android-sdk/blob/master/firebase-database/src/main/java/com/google/firebase/database/FirebaseDatabase.java#L208

.changeset/clever-kangaroos-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fixex an issue that caused `refFromUrl()` to reject production database URLs when `useEmulator()` was used.
```
samtstern, line 5:
```suggestion
Fixes an issue that caused `refFromUrl()` to reject production database URLs when `useEmulator()` was used.
```

samtstern (approved):

jsdt (approved):

schmidt-sebastian:
@hsubox76 Do you know why I cannot merge?

hsubox76 (approved):

hsubox76:
Changeset changes need to be approved by global approvers, changeset approvers, or firestore-js-team which is defined here: https://github.com/firebase/firebase-js-sdk/blob/34b4e3004ecd3e69a60893cfa0d1801890064a1a/.github/CODEOWNERS#L26

I approved.

PR #4206, opened 2020-12-14
Issue template for alpha

samtstern:
This PR makes a few changes to issue templates:

  1. Adds a separate issue template for Alpha bugs. It has a slightly different preamble and also auto-applies the `alpha` label.
  1. Adds a new "choice" screen to direct people to the right template and also offer a link to Support.

Hard to preview in a PR but here's how the iOS SDK chooser looks:
Screen Shot 2020-12-14 at 1 21 25 PM

### DIFF
--- a/.github/ISSUE_TEMPLATE/alpha_bug_report.md
+++ b/.github/ISSUE_TEMPLATE/alpha_bug_report.md
@@ -1,16 +1,20 @@
+---
+name: 🧪 Alpha SDK bug report
+about: Have you found a bug in one of our pre-release or alpha SDKs? File it here.
+title: ""
+labels: "alpha"
+assignees: ""
+---
+
 <!-- Step 1 [READ THIS] -->
 <!--
 Are you in the right place?
   * For issues or feature requests related to __the code in this repository__
     file a Github issue.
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/alpha_bug_report.md
```diff
@@ -0,0 +1,48 @@
...
+  What happened? How can we make the problem occur?
+  This could be a description, log/console output, etc.
+-->
+#### Relevant Code:
+
+<!--
+  Reproduce the issue on StackBlitz and provide your forked URL
+  or give us some sample code below
+-->
+https://stackblitz.com/fork/firebase-issue-sandbox
```
hsubox76, line 44: The stackblitz link isn't going to work for exp and I guess nobody even uses it for regular bug reports, which is understandable since the project template seems to be set on firebase 4.9.0. Maybe we should just remove it from both for now?

samtstern: Hah ok will do!

hsubox76 (approved):

PR #4207, opened 2020-12-14
Use pruned DTS file for API report

schmidt-sebastian:
This PR hooks the "prune-dts" script into Firestor's API Extractor pipeline. It runs three phases:

- Use API Extractor to generate a rollup DTS
- Prune the rollup DTS to strip private symbols
- Use API Extractor to generate the API report based on the pruned symbols

As you can see in the output, this removes all sort of weird private types from the API report.

Also changed the `.eslint.json` in `prune-dts` to be more efficient by not including all `.ts` files in the entire repo.

Fixes https://github.com/firebase/firebase-js-sdk/issues/4441


schmidt-sebastian:
@Feiyang1 I was wondering if you had an opinion here. Should we carry this forward?

packages/firestore/package.json
```diff
@@ -79,7 +78,8 @@
     "@firebase/app-types": "0.x"
   },
   "devDependencies": {
-    "@firebase/app": "0.6.14",
+    "@firebase/app": "0.6.13",
```
hsubox76: bump back up to 14

packages/firestore/package.json
```diff
@@ -27,9 +25,9 @@
     "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "lint:fix": "eslint --fix -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "prettier": "prettier --write '*.js' '*.ts' '@(lite|exp|src|test)/**/*.ts'",
-    "pregendeps:exp": "yarn api-report:exp && node scripts/build-bundle.js --input ./exp/index.ts --output ./dist/exp/tmp.js",
+    "pregendeps:exp": "yarn api-report&& node scripts/build-bundle.js --input ./exp/index.ts --output ./dist/exp/tmp.js",
```
hsubox76, line 28: Space before && (don't think it NEEDS it but for formatting)

schmidt-sebastian:
Added bug fix to Prune DTS so that it no longer replaces PromiseLike with LoadBundleTask.

packages/firestore/package.json
schmidt-sebastian: Yeah, I think this was causing all the build failures earlier.

packages/firestore/package.json
schmidt-sebastian: Done

packages/firestore/package.json
```diff
@@ -46,7 +44,8 @@
     "test:node:persistence:prod": "node ./scripts/run-tests.js --main=index.node.ts --persistence 'test/{,!(browser|lite)/**/}*.test.ts'",
     "test:travis": "ts-node --compiler-options='{\"module\":\"commonjs\"}' ../../scripts/emulator-testing/firestore-test-runner.ts",
     "test:minified": "(cd ../../integration/firestore ; yarn test)",
-    "api-report": "rm -rf temp && api-extractor run --local --verbose",
+    "prepare": "yarn build:release",
```
Feiyang1, line 47:
```suggestion
```
prepare scripts were removed in https://github.com/firebase/firebase-js-sdk/pull/4296 to speed up releases.

packages/firestore/scripts/api-report.ts
```diff
@@ -0,0 +1,167 @@
...
+  pruneDts,
+  removeUnusedImports
+} from '../../../repo-scripts/prune-dts/prune-dts';
+
+/* eslint-disable no-console */
+
+// This script takes the output of the API Extractor, post-processes it using
+// the pruned-dts script and then invokes API report to generate a report
+// that only includes exported symbols. This is all done in temporary folders,
+// all all configuration is auto-generated for each run.
```
Feiyang1, line 34:
```suggestion
// all configuration is auto-generated for each run.
```

hsubox76 (approved):
LGTM, I don't know if @Feiyang1 wants to take a second glance at the api-report script.

packages/firestore/scripts/api-report.ts
```diff
@@ -0,0 +1,167 @@
...
+    /* dtsRollupEnabled= */ false,
+    /* apiReportEnabled= */ true
+  );
+  Extractor.invoke(extractorConfig, { localBuild: true });
+  console.log(
+    `API report for firestore-${sdkVariant} written to ${reportFolder}`
+  );
+}
+
+void generateApi('lite').then(() => generateApi('exp'));
```
Feiyang1, line 167: What does void do?

packages/firestore/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "outDir": "dist"
   },
   "exclude": [
+    "scripts/**/*",
```
Feiyang1, line 7: Is it for performance or something else?

packages/firestore/package.json
```diff
@@ -80,6 +79,7 @@
   },
   "devDependencies": {
     "@firebase/app": "0.6.14",
+    "@microsoft/api-extractor": "7.13.1",
```
Feiyang1, line 82:
```suggestion
```

Currently we are using a fork of api-extractor that supports multiple entry points and it's already part of the root package.json.

packages/firestore/scripts/api-report.ts
```diff
@@ -0,0 +1,167 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { Extractor, ExtractorConfig } from '@microsoft/api-extractor';
```
Feiyang1, line 21:
```suggestion
import { Extractor, ExtractorConfig } from 'api-extractor-me';
```

`api-extractor-me` is the package name of the fork

packages/firestore/package.json
```diff
@@ -46,7 +44,8 @@
     "test:node:persistence:prod": "node ./scripts/run-tests.js --main=index.node.ts --persistence 'test/{,!(browser|lite)/**/}*.test.ts'",
     "test:travis": "ts-node --compiler-options='{\"module\":\"commonjs\"}' ../../scripts/emulator-testing/firestore-test-runner.ts",
     "test:minified": "(cd ../../integration/firestore ; yarn test)",
-    "api-report": "rm -rf temp && api-extractor run --local --verbose",
+    "prepare": "yarn build:release",
+    "api-report": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' ts-node ./scripts/api-report.js",
```
Feiyang1, line 48: Should we execute `api-report.ts` directly since we are using ts-node. Then we won't need to commit the compiled js.

Feiyang1 (changes requested):

packages/firestore/package.json
schmidt-sebastian: The compiled JS gets committed because of the `yarn build:scripts` rule which is needed because of `rollup`. I fixed this line though since using `ts-node` and the `.js` script is pretty... awkward.

packages/firestore/scripts/api-report.ts
schmidt-sebastian: It suppresses `no-floating-promises`: https://github.com/palantir/tslint/issues/3020

packages/firestore/tsconfig.json
schmidt-sebastian: Yes, otherwise it parses all the TS files in this tree.

schmidt-sebastian (commented):
Addressed feedback. Thanks for making most of it clickable :)

Feiyang1 (approved):

PR #4215, opened 2020-12-17
Fis compat

ChaoqunCHEN:
Implement Installations compat layer.


packages-exp/installations-compat/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { name, version } from '../package.json';
+import { _FirebaseNamespace } from '@firebase/app-types/private';
+import { Component, ComponentType } from '@firebase/component';
+import { FirebaseInstallations as FirebaseInstallationsCompat } from '@firebase/installations-types';
+import { FirebaseApp } from '@firebase/app-types';
+import { InstallationsCompat } from './installationsCompat';
+
+declare module '@firebase/component' {
+  interface NameServiceMapping {
+    'app-compat': FirebaseApp;
```
Feiyang1, line 28:
```suggestion
```

Feiyang1 (approved):
Awesome! The PR looks great, thank you!

packages-exp/installations-compat/src/index.ts
ChaoqunCHEN: Thank you for reviewing! 
Removal of this line causes build failure. So I merged this PR without the removal.

PR #4225, opened 2020-12-18
Backport Android updateTransform changes

thebrianchen:
After porting update transforms to [Android](https://github.com/firebase/firebase-android-sdk/pull/2262), I found some changes that needed to be ported back. I also replaced usages of TransformMutation.

### DIFF
--- a/packages/firestore/src/model/mutation.ts
+++ b/packages/firestore/src/model/mutation.ts
@@ -87,11 +87,11 @@ export class MutationResult {
      */
     readonly version: SnapshotVersion,
     /**
-     * The resulting fields returned from the backend after a
-     * TransformMutation has been committed. Contains one FieldValue for each
-     * FieldTransform that was in the mutation.
+     * The resulting fields returned from the backend after a mutation
+     * containing field transforms has been committed. Contains one FieldValue
+     * for each FieldTransform that was in the mutation.
      *
-     * Will be null if the mutation was not a TransformMutation.
+     * Will be null if the mutation did not contain any field transforms.
      */
     readonly transformResults: Array<ProtoValue | null> | null
   ) {}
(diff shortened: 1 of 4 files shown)


packages/firestore/src/model/mutation.ts
```diff
@@ -191,28 +191,22 @@ export function preconditionIsValidForDocument(
...
  * locally, we return an `UnknownDocument` and rely on Watch to send us the
  * updated version.
  *
- * Note that TransformMutations don't create Documents (in the case of being
- * applied to a NoDocument), even though they would on the backend. This is
- * because the client always combines the TransformMutation with a SetMutation
- * or PatchMutation and we only want to apply the transform if the prior
- * mutation resulted in a Document (always true for a SetMutation, but not
- * necessarily for a PatchMutation).
+ * Field transforms are used only with Patcdh and Set Mutations. We use the
```
wilhuff, line 207:
```suggestion
 * Field transforms are used only with Patch and Set Mutations. We use the
```

packages/firestore/src/model/mutation.ts
```diff
@@ -689,14 +675,6 @@ function localTransformResults(
       previousValue = maybeDoc.field(fieldTransform.field);
     }
 
-    if (previousValue === null && baseDoc instanceof Document) {
```
wilhuff, line 692: Why is this logic no longer needed? This seems pretty critical. Or is it handled elsewhere now?

thebrianchen: My understanding is that this if-statement checks for the edge case where a PatchMutation clears the values before the TransformMutation can be applied. With the addition of updateTransforms, field transformation are encoded on the Patch or Set Mutation, rather than in a separate TransformMutation. This means that we won't bump into this edge case anymore since the update and fieldTransform are on the same mutation.

I've also removed the now obsolete server timestamp test that tested this edge case.

wilhuff (approved):
LGTM

PR #4232, opened 2020-12-21
Implement `startAfter` and `endBefore` for RTDB queries

IanWyszynski:
### Discussion

API proposal: [go/rtdb-exclusive-query-filters](go/rtdb-exclusive-query-filters)

### Testing

The integration tests in `query.tests.ts` and `orderby.test.ts` have been extended to cover `startAfter`.

### API Changes

This adds `startAfter()` to the query interface.

`startAfter(value: any, key?: string | null)`

Returns values that are greater than `(value, key)`. If `key` null or undefined, this is equivalent to `startAt(value, MAX_NAME)`, where `MAX_NAME` compares higher than any other key. 

If `key` is not null and defined, it is equivalent to `startAt(value, key + 1)`, where `key + 1` means `'' + (parseInt(key) + 1)"` if `key` is parse-able as an integer, and the lexicographically next key if the value is not parseable as an integer.


packages/database/src/core/util/util.ts
```diff
@@ -566,6 +566,42 @@ export const tryParseInt = function (str: string): number | null {
   return null;
 };
 
+const MIN_PRINTABLE_CHAR: number = 0x20;
```
IanWyszynski, line 569: This should be changed to only allow the blessed characters: https://stackoverflow.com/questions/36086317/firebase-push-id-what-characters-can-be-inside

IanWyszynski:
@schmidt-sebastian This is ready for review.

packages/database/src/api/Query.ts
```diff
@@ -526,6 +529,28 @@ export class Query {
...
+
+    return new Query(this.repo, this.path, newParams, this.orderByCalled_);
+  }
+
+  startAfter(
+    value: number | string | boolean | null = null,
+    name?: string | null
+  ): Query {
+    validateArgCount('Query.startAfter', 0, 2, arguments.length);
+    validateFirebaseDataArg('Query.startAfter', 1, value, this.path, true);
```
schmidt-sebastian, line 541: This argument isn't optional (it can be null, but not undefined). Not sure why it is marked optional in startAt.

schmidt-sebastian (approved):

IanWyszynski:
@schmidt-sebastian Do you think I should merge https://github.com/firebase/firebase-js-sdk/pull/4270 into this before merging with master?

schmidt-sebastian:
You can merge this any way you want :)

.changeset/hip-glasses-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/database': minor
+---
+
+Add `startAfter` for RTDB queries
```
schmidt-sebastian, line 5: Do you want to add a bit of an explanation on what this is used for? This is the space for you to sell this to our users, especially those that struggle with pagination.

Issue #4226, opened 2020-12-18
[Bug] Field type reference in DocumentSnapshot obtained in transaction is not an instance of DocumentReference


rulamon:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10 Home 1909
  * Browser version:  Chrome 87
  * Firebase SDK version: 8.1.2
  * Firebase Product: firestore

### [REQUIRED] Describe the problem
When getting a document through a transaction (Transaction.get()) the field type reference does not does not return an instance of DocumentReference anymore. Similar to https://github.com/firebase/firebase-js-sdk/issues/4125 but with transactions instead of queries.

schmidt-sebastian:
Thanks for reporting this. We will get this fixed in the next release, which unfortunately is not slated to happen until the beginning of January.

PR #4233, opened 2020-12-22
Return Firestore Classic types from Transaction API

schmidt-sebastian:
We need to use the "classic" UserDataWriter in the old Transaction API, as only the classic UserDataWriter returns the old API types. This is the same fix as https://github.com/firebase/firebase-js-sdk/pull/4136 (and the last!)

Fixes https://github.com/firebase/firebase-js-sdk/issues/4226

### DIFF
--- a/.changeset/slimy-mugs-type.md
+++ b/.changeset/slimy-mugs-type.md
@@ -2,4 +2,4 @@
 "@firebase/firestore": patch
 ---
 
-Fixes an issue in the Transaction API that caused the SDK to return invalid DocumentReferences when returned from `DocumentSnapshot.data()`.
+Fixes an issue in the Transaction API that caused the SDK to return invalid DocumentReferences through `DocumentSnapshot.data()` calls.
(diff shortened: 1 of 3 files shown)


.changeset/slimy-mugs-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/firestore": patch
+---
+
+Fixes an issue in the Transaction API that caused the SDK to return invalid DocumentReferences when returned from `DocumentSnapshot.data()`.
```
thebrianchen, line 5: How about "caused `DocumentSnapshot.data()` to return valid DocumentReferences"? The mixing of present and past tense for "return" makes it a bit hard to understand.

thebrianchen (approved):
LGTM. I restarted the CI b/c the failure looked flaky. Not sure what the changeset error is though.

.changeset/slimy-mugs-type.md
schmidt-sebastian: I changed it up a bit.

PR #4262, opened 2021-01-05
Update outdated error message to include '!='

### LINKED ISSUE #4253: Outdated Error Message

  * Operating System version: Windows 10
  * Browser version: Chrome
  * Firebase SDK version: 6.14.10
  * Firebase Product: Firestore

### [REQUIRED] Describe the problem/steps to reproduce:

I get an error message when running a query with the not equal comparison combined with an order by on another field:
```
Invalid query.... You have a where filter with an inequality (<, <=, >, or >=) on field 'myDocField' and so you must also use 'myDocField' as your first argument to Query.orderBy(), but your first orderBy() is on field 'someOtherDocField' instead.
```
I believe the error message need to be updated to include the "!=" symbol.

```
#### Relevant Code:

 My query is like this:
```
myQuery = myReference.where("myDocField", "!=", 0).orderBy("someOtherDocField").get(); //throws error
```
while this doesn't throw the error:
```
myQuery = myReference.where("myDocField", "!=", 0).orderBy("myDocField").orderBy("someOtherDocField").get(); //allowed

thebrianchen:
Fixes #4253.

### DIFF
--- a/packages/firestore/src/lite/query.ts
+++ b/packages/firestore/src/lite/query.ts
@@ -795,7 +795,7 @@ function validateNewFilter(query: InternalQuery, filter: Filter): void {
       throw new FirestoreError(
         Code.INVALID_ARGUMENT,
         'Invalid query. All where filters with an inequality' +
-          ' (<, <=, !=, >, or >=) must be on the same field. But you have' +
+          ' (<, <=, !=, not-in, >, or >=) must be on the same field. But you have' +
           ` inequality filters on '${existingField.toString()}'` +
           ` and '${filter.field.toString()}'`
       );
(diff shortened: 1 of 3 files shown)


packages/firestore/src/lite/query.ts
```diff
@@ -795,7 +795,7 @@ function validateNewFilter(query: InternalQuery, filter: Filter): void {
       throw new FirestoreError(
         Code.INVALID_ARGUMENT,
         'Invalid query. All where filters with an inequality' +
-          ' (<, <=, >, or >=) must be on the same field. But you have' +
+          ' (<, <=, !=, >, or >=) must be on the same field. But you have' +
```
schmidt-sebastian, line 798: I think this also needs to include `in` and `not-in`, which is rejected by the backend as well.

thebrianchen: Good catch! Added.

schmidt-sebastian (approved):

PR #4263, opened 2021-01-05
Add import of TextEncoder

thebrianchen:
Not having the import caused the integration tests to fail on my local machine.


schmidt-sebastian (approved):

thebrianchen:
Turns out I was using v10.x.x of node, which was causing this error. Updating to v12.x.x removed the import error.

Issue #18, opened 2017-05-21
FR: Storage for node.js


MatiasConTilde:
There is a way to use it with Google Cloud, but it is a bit complicated and firebase storage directly for web is much easier and convenient to use.

sphippen:
This is definitely on our radar. There's stuff that many developers would want to do in a Node.js environment that doesn't even exist in the current Storage JS API, like file downloads.

As you alluded to, https://github.com/GoogleCloudPlatform/google-cloud-node *generally* works as a replacement for server-side use cases where you can use service account credentials. The API is different, but it lets you do things the JS API doesn't support right now (file downloads!).

schmidt-sebastian:
The Firebase Admin SDK for Node has recently added support for Storage by wrapping the Google Cloud Storage libraries: https://firebase.google.com/docs/storage/admin/start. This makes it a lot easier to integrate GCS with Firebase (as it supports Firebase Authentication). While this integration does not unify the API surface, it simplifies the initial setup.

mikamaunula:
Firebase js sdk has now support for nodejs #221. Thank you for that!! 
We have a need for storage api also. Is there any possibility to add nodejs support?

ricsmania:
@MatiasConTilde could you please provide an example on how to use Google Cloud to access Firebase Storage? I tried it but I'm getting an error: 

> Anonymous users does not have storage.objects.get access to mybucket.appspot.com/object.txt

My Firebase Storage is public and I'm not authenticating, so I think it should work.

I have never used Google Cloud directly, do I have to create an account there and configure it too?

I tried setting an incorrect bucket and I got a different error, so I guess at least that part is correct.

IStep95:
Why it is not possible to upload with this configuration to any bucket? 
Like on frontend(Angular2+) application it is possible, why is it different on backend(node.js)?
var firebaseConfig = {
    apiKey: "YOUR-API-KEY",
    authDomain: "YOUR-AUTH-DOMAIN",
    databaseURL: "YOUR-DATABASE-URL",
    projectId: "YOUR-PROJECT-ID",
    storageBucket: "YOUR-STORAGE-BUCKET",
    messagingSenderId: "YOUR-MESS-SENDER-ID",
    appId: "YOUR-APP-ID"
  };

I know I can accomplish upload with google cloud storage api but than I need public and private key.
I am very interested will it be implemented soon, or it will not?

schmidt-sebastian:
[...]

PR #4266, opened 2021-01-06
Allow storage processes to exit before timeout in node

theprojectsomething:
Currently the storage API is unsupported for node (see #18). This means storage rules are unavailable from a node environment (even using the admin interface). This can be resolved with some very minor updates to the storage package (for later versions of node at least):

- [x] Resolve any ongoing processes (in this case uncleared timeouts) to allow node to exit as expected once a script is complete
- [ ] Include an XHR polyfill (via dynamic import or node-specific build) ... successfully tested with [xmlhttprequest-ssl](https://www.npmjs.com/package/xmlhttprequest-ssl)

This PR does half the work. It clears a timeout that keeps a storage process running for a minimum 10mins in node (waiting on the 600,000ms function timeout to complete). This fix should also be useful in other environments as it clears up resources.

This PR doesn't include the XHR polyfill as that seems like a fairly opinionated piece of work. The `firebase/app` package actually already includes an XHR polyfill (incompatible here), but gives some direction for implementation.

That said, given the current changes in place, users needing storage functionality in node can easily install the XHR module in their project directly and patch the compiled storage script `@firebase/storage/dist/index.cjs.js` for a workable (if not full) level of functionality.

### DIFF
--- a/packages/storage/src/implementation/backoff.ts
+++ b/packages/storage/src/implementation/backoff.ts
@@ -42,6 +42,7 @@ export function start(
   // TODO: find a way to exclude Node type definition for storage because storage only works in browser
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let timeoutId: any = null;
+  let hitTimeoutId: any = null;
   let hitTimeout = false;
   let cancelState = 0;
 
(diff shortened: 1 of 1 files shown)


remorses:
Hey @schmidt-sebastian can you take a look?

PR #4270, opened 2021-01-07
Add `endBefore` to RTDB queries

IanWyszynski:
[go/rtdb-exclusive-query-filters](go/rtdb-exclusive-query-filters)


IanWyszynski:
@schmidt-sebastian When you're looking at https://github.com/firebase/firebase-js-sdk/pull/4232 could you also have a look here?

IanWyszynski:
@schmidt-sebastian This is ready for another review round.

packages/database/src/api/Query.ts
```diff
@@ -572,7 +575,28 @@ export class Query {
...
+
+    return new Query(this.repo, this.path, newParams, this.orderByCalled_);
+  }
+
+  endBefore(
+    value: number | string | boolean | null = null,
+    name?: string | null
+  ): Query {
+    validateArgCount('Query.endBefore', 0, 2, arguments.length);
+    validateFirebaseDataArg('Query.endBefore', 1, value, this.path, true);
```
schmidt-sebastian, line 591: This is not optional. Do `null` values still work if you pass `false` as the last argument?

packages/database/test/query.test.ts
```diff
@@ -1667,6 +1891,64 @@ describe('Query Tests', () => {
     );
   });
 
+  it('Ensure startAt / endBefore with priority and name works (2).', () => {
```
schmidt-sebastian, line 1894: If you need more things on your plate, can you also use a different suffix for this test name?

packages/database/test/query.test.ts
```diff
@@ -3766,6 +4192,90 @@ describe('Query Tests', () => {
...
+    ref.set(
+      {
+        1: true,
+        50: true,
+        550: true,
+        6: true,
+        600: true,
+        70: true,
+        8: true,
+        80: true
```
schmidt-sebastian, line 4232: You don't need these extra values.

packages/database/test/query.test.ts
```diff
@@ -3766,6 +4192,90 @@ describe('Query Tests', () => {
...
+        1: true,
+        50: true,
+        550: true,
+        6: true,
+        600: true,
+        70: true,
+        8: true,
+        80: true,
+        'a': true
+      },
```
schmidt-sebastian, line 4256: It might make sense to extract a local variable here so you can use the same set twice.

schmidt-sebastian (approved):

PR #4280, opened 2021-01-12
Expose emulator config for the Auth SDK

samhorlbeck:
Also add the relevant bits to the externs in the firebase package, so that other Closure libraries can consume the new features.


avolkovi (approved):

yuchenshi (approved):
Rubber stamping since this is already reviewed internally

Feiyang1 (approved):

PR #4284, opened 2021-01-13
Fix FirestoreDataConverter.fromFirestore() being called with QueryDocumentSnapshot from exp

### LINKED ISSUE #4278: DocumentReference from QueryDocumentSnapshot in model converter not typed correctly.

  * Operating System version: Windows
  * Browser version: Chrome  (87.0.4280.141)
  * Firebase SDK version: 9.1.0
  * Firebase Product: firestore

When trying to save the DocumentReference from the QueryDocumentSnapshot in the ModelConverter.fromFirestore function. The resulting document reference object is missing several fields.

#### Steps to reproduce:

See code below:

#### Relevant Code:

```
interface MyModel {
  ref: firebase.firestore.DocumentReference;
}

const modelConverter = {
  toFirestore: function (model: MyModel) {
    return _.omit(model, "ref");
  },
  fromFirestore: function (
    snapshot: firebase.firestore.QueryDocumentSnapshot
  ): MyModel {
    const model = snapshot.data();
    // Some kind of validation, etc.
    model.ref = snapshot.ref;
    return model as MyModel;
  },
};

const modelDocumentRef = firebase.firestore()
  .collection("/models")
  .doc("some_id")
  .withConverter(modelConverter);

  modelDocumentRef.onSnapshot((snapshot) => {
  const model = snapshot.data();
  const modelSubcollectionRef = model?.ref.collection("/subcollection");
});
```

Runtime output:

```
Uncaught (in promise) TypeError: model.collection is not a function
```

Instead I have to do:

```
  fromFirestore: function (
    snapshot: firebase.firestore.QueryDocumentSnapshot
  ): MyModel {
    const model = snapshot.data();
    // Some kind of validation, etc.
>   model.ref = firebase.firestore().doc(snapshot.ref.path);
    return model as MyModel;
  },

```

dconeybe:
`FirestoreDataConverter.fromFirestore()` was being invoked with an instance of `QueryDocumentSnapshot` from the experimental SDK; however, it was expected to be invoked with an instance of `QueryDocumentSnapshot` from the Classic SDK. This worked okay, except that the object was lacking attributes like `ref`, which are only provided by the Classic SDK.

The root cause is that the `FirestoreDataConverter` provided by the user was registered directly with the `QueryDocumentSnapshot` from the experimental SDK, and therefore was invoked with types from the experimental SDK. The fix is to instead register a "wrapper" data converter with the `QueryDocumentSnapshot` that calls the underlying data converter with the equivalent objects from the Classic SDK.

There was, however, a side effect of this change that caused `isEqual()` methods (e.g. `DocumentReference.isEqual()`) to incorrectly return `false` if compared with another `DocumentReference` created with the same converter instance. This is because the implementation of `isEqual()` compares the converters using the `===` operator. The wrapper instances that wrapped the same converters were distinct objects, and thus did not compare equal using `===`. The fix was to store each converter in a `WeakMap` mapping it to its wrapper instance so that the same wrapper instance can be used for each given converter, causing `isEqual()` to compare correctly.

Fixes #4278

### DIFF
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -553,18 +553,34 @@ export class WriteBatch
   }
 }
 
-class PublicFirestoreDataConverterAdapter<U> implements UntypedFirestoreDataConverter<U> {
-  constructor(readonly firestore: Firestore, readonly delegate: PublicFirestoreDataConverter<U>) {}
+class PublicFirestoreDataConverterAdapter<U>
+  implements UntypedFirestoreDataConverter<U> {
+  constructor(
+    readonly firestore: Firestore,
+    readonly delegate: PublicFirestoreDataConverter<U>
+  ) {}
 
   fromFirestore(snapshot: unknown, options?: unknown): U {
     const expSnapshot = snapshot as ExpQueryDocumentSnapshot<U>;
-    const querySnapshot = new QueryDocumentSnapshot<U>(this.firestore, expSnapshot);
-    return this.delegate.fromFirestore(querySnapshot, options as PublicSnapshotOptions);
+    const querySnapshot = new QueryDocumentSnapshot<U>(
+      this.firestore,
+      expSnapshot
+    );
+    return this.delegate.fromFirestore(
+      querySnapshot,
+      options as PublicSnapshotOptions
+    );
   }
 
   toFirestore(modelObject: U): PublicDocumentData;
-  toFirestore(modelObject: Partial<U>, options: PublicSetOptions): PublicDocumentData;
-  toFirestore(modelObject: U | Partial<U>, options?: PublicSetOptions): PublicDocumentData {
+  toFirestore(
+    modelObject: Partial<U>,
+    options: PublicSetOptions
+  ): PublicDocumentData;
+  toFirestore(
+    modelObject: U | Partial<U>,
+    options?: PublicSetOptions
+  ): PublicDocumentData {
     if (options === undefined) {
       return this.delegate.toFirestore(modelObject as U);
     } else {
@@ -773,7 +789,9 @@ export class DocumentReference<T = PublicDocumentData>
   ): PublicDocumentReference<U> {
     return new DocumentReference<U>(
       this.firestore,
-      this._delegate.withConverter(new PublicFirestoreDataConverterAdapter(this.firestore, converter)),
+      this._delegate.withConverter(
+        new PublicFirestoreDataConverterAdapter(this.firestore, converter)
+      )
     );
   }
 }
@@ -1080,7 +1098,9 @@ export class Query<T = PublicDocumentData>
   withConverter<U>(converter: PublicFirestoreDataConverter<U>): Query<U> {
     return new Query<U>(
       this.firestore,
-      this._delegate.withConverter(new PublicFirestoreDataConverterAdapter(this.firestore, converter)),
+      this._delegate.withConverter(
+        new PublicFirestoreDataConverterAdapter(this.firestore, converter)
+      )
     );
   }
 }
@@ -1222,7 +1242,9 @@ export class CollectionReference<T = PublicDocumentData>
   ): CollectionReference<U> {
     return new CollectionReference<U>(
       this.firestore,
-      this._delegate.withConverter(new PublicFirestoreDataConverterAdapter(this.firestore, converter)),
+      this._delegate.withConverter(
+        new PublicFirestoreDataConverterAdapter(this.firestore, converter)
+      )
     );
   }
 }
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -575,18 +575,16 @@ class FirestoreDataConverter<U>
     snapshot: ExpQueryDocumentSnapshot,
     options?: PublicSnapshotOptions
   ): U {
+    const expSnapshot = new ExpQueryDocumentSnapshot(
+      this._firestore._delegate,
+      this._userDataWriter,
+      snapshot._key,
+      snapshot._document,
+      snapshot.metadata,
+      /* converter= */ null
+    );
     return this._delegate.fromFirestore(
-      new QueryDocumentSnapshot<U>(
-        this._firestore,
-        new ExpQueryDocumentSnapshot<U>(
-          this._firestore._delegate,
-          this._userDataWriter,
-          snapshot._key,
-          snapshot._document,
-          snapshot.metadata,
-          /* converter= */ null
-        )
-      ),
+      new QueryDocumentSnapshot(this._firestore, expSnapshot),
       options ?? {}
     );
   }
--- a/packages/firestore/src/api/database.ts
+++ b/packages/firestore/src/api/database.ts
@@ -553,26 +553,25 @@ export class WriteBatch
   }
 }
 
-class UntypedFirestoreDataConverterAdapter<U>
-  extends Compat<PublicFirestoreDataConverter<U>>
-  implements UntypedFirestoreDataConverter<U> {
+/**
+ * Wraps a `PublicFirestoreDataConverter` translating the types from the
+ * experimental SDK into corresponding types from the Classic SDK before passing
+ * them to the wrapped converter.
+ */
+class FirestoreDataConverter<U> extends Compat<PublicFirestoreDataConverter<U>> implements UntypedFirestoreDataConverter<U> {
+
   private static readonly INSTANCES = new WeakMap();
 
   private constructor(
     private readonly _firestore: Firestore,
+    private readonly _userDataWriter: UserDataWriter,
     delegate: PublicFirestoreDataConverter<U>
   ) {
     super(delegate);
   }
 
-  fromFirestore(snapshot: unknown, options?: unknown): U {
-    return this._delegate.fromFirestore(
-      new QueryDocumentSnapshot<U>(
-        this._firestore,
-        snapshot as ExpQueryDocumentSnapshot<U>
-      ),
-      options as PublicSnapshotOptions
-    );
+  fromFirestore(snapshot: ExpQueryDocumentSnapshot, options?: PublicSnapshotOptions): U {
+    return this._delegate.fromFirestore(new QueryDocumentSnapshot<U>(this._firestore, new ExpQueryDocumentSnapshot<U>(this._firestore._delegate, this._userDataWriter, snapshot._key, snapshot._document, snapshot.metadata,/* converter= */ null)), options ?? {});
   }
 
   toFirestore(modelObject: U): PublicDocumentData;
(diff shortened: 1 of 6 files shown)


packages/firestore/src/api/database.ts
```diff
@@ -552,6 +553,41 @@ export class WriteBatch
   }
 }
 
+class UntypedFirestoreDataConverterAdapter<U>
```
schmidt-sebastian, line 556: This now looks like the rest of our "Compat" classes. You could follow the existing pattern and start this off as such:

```
class FirestoreDataConverter<U>
  extends Compat<ExpFirestoreDataConverter<U>>
  implements FirestoreDataConverter<T> {
```

I also wonder if thanks to this adapter, we may now be able to stop using the `UntypedFirestoreDataConverterAdapter`.

packages/firestore/src/api/database.ts
```diff
@@ -753,7 +789,7 @@ export class DocumentReference<T = PublicDocumentData>
     return new DocumentReference<U>(
       this.firestore,
       this._delegate.withConverter(
-        converter as UntypedFirestoreDataConverter<U>
+        new UntypedFirestoreDataConverterAdapter(this.firestore, converter)
```
schmidt-sebastian, line 792: Since we cannot force our users to implement a custom `isEqual` function, the problem you mentioned in chat can likely only be solved by looking at the type of data converter and "unwrapping" the adapter when needed.

This would look something like this:

```
const c1 = this.converter instanceof Compat ? this.converter._delegate : this.converter;
const c2 = that.converter instanceof Compat ? that.converter._delegate : that.converter;

if (c1 == c2) console.log("Woohoo!");
```

FWIW, a more straightforward fix may be to not call https://github.com/firebase/firebase-js-sdk/blob/566a62dd4d54ca6ca8bef8bb4409023050bb9f90/packages/firestore/src/exp/snapshot.ts#L264 when a converter is used, but to instead do something like this:


````
// In the compat class

data() {
  if (!this.converter) return this.delegate.data(); 
  else {
  const snapshot = new ExpQueryDocumentSnapshot(this._firestore, ...,
     new QueryDocumentSnapshot(
        this._firestore,
        this._userDataWriter,
        this._key,
        this._document,
        this.metadata,
        /* converter= */ null
      ));
      return this._converter.fromFirestore(snapshot, options);
  }
}


````

File: packages/firestore/src/api/database.ts (hunk 1 above)
Comment (dconeybe) [reply packages/firestore/src/api/database.ts:556]: Good point about using the `Compat` class.

I was also thinking the same thing about `UntypedFirestoreDataConverterAdapter` not being needed anymore. I think that you're right, if we use this "compat" adapter strategy then we can get rid of `UntypedFirestoreDataConverterAdapter`.

File: packages/firestore/src/api/database.ts (hunk 2 above)
Comment (dconeybe) [reply packages/firestore/src/api/database.ts:792]: I like this more "straightforward" fix. We would just have to be careful to implement `isEqual` to check equality of `this._converter` instead of relying on comparing the `_converter` of the delegate (since it will always be null). I think this is the best approach though so I'm going to try it out.

Comment (schmidt-sebastian) [reply packages/firestore/src/api/database.ts:792]: I believe the `isEqual` problem goes away if we use the "straightforward fix.

Comment (dconeybe) [reply packages/firestore/src/api/database.ts:792]: If we use the "straightforward fix" then the delegate's converter will always be null. Since `refEqual()` checks the converter for equality, that logic would need to be moved into the `isEqual()` methods to maintain the same semantics of `isEqual()`. Otherwise, the converters would not be considered when checking for equality, and the current logic *does* check the converters for equality.

File: packages/firestore/src/api/database.ts [hunk 3]
```diff
@@ -552,6 +553,61 @@ export class WriteBatch
...
+    delegate: PublicFirestoreDataConverter<U>
+  ) {
+    super(delegate);
+  }
+
+  fromFirestore(snapshot: unknown, options?: unknown): U {
+    return this._delegate.fromFirestore(
+      new QueryDocumentSnapshot<U>(
+        this._firestore,
+        snapshot as ExpQueryDocumentSnapshot<U>
```
schmidt-sebastian, line 572: This unfortunately causes any data (such as DocumentReferences) to be returned as exp-types. You need to follow the pattern here: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api/database.ts#L417

packages/firestore/src/api/database.ts
```diff
@@ -552,6 +553,61 @@ export class WriteBatch
...
+  private static readonly INSTANCES = new WeakMap();
+
+  private constructor(
+    private readonly _firestore: Firestore,
+    delegate: PublicFirestoreDataConverter<U>
+  ) {
+    super(delegate);
+  }
+
+  fromFirestore(snapshot: unknown, options?: unknown): U {
```
schmidt-sebastian, line 568: Can you use `ExpQueryDocumentSnapshot` and `PublicSnapshotOptions` here (instead of unknown)?

packages/firestore/src/api/database.ts
```diff
@@ -552,6 +553,61 @@ export class WriteBatch
   }
 }
 
+class UntypedFirestoreDataConverterAdapter<U>
```
schmidt-sebastian, line 556: Can you add some JSDoc for this class? You can likely summarize your PR description.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1373,6 +1373,25 @@ apiDescribe('Database', (persistence: boolean) => {
       }
     };
 
+    class ModelWithRef {
```
schmidt-sebastian, line 1376: Could you re-use the Post-Converter? That might cut down on the extra lines in this file.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1373,6 +1373,25 @@ apiDescribe('Database', (persistence: boolean) => {
...
+      constructor(readonly ref: firestore.DocumentReference<ModelWithRef>) {}
+    }
+
+    const modelWithRefConverter = {
+      toFirestore(model: ModelWithRef): firestore.DocumentData {
+        // Remove the "ref" attribute, since it's not meant to be stored in
+        // the Firestore database.
+        const { ['ref']: _, ...rest } = model;
+        return rest;
```
schmidt-sebastian, line 1385: Doesn't this remove all the data since model only contains `ref`?

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1590,6 +1609,51 @@ apiDescribe('Database', (persistence: boolean) => {
...
+    });
+
+    it('Correct snapshot specified to fromFirestore() when registered with CollectionReference', () => {
+      return withTestDb(persistence, async db => {
+        const untypedCollection = db
+          .collection('/models')
+          .doc()
+          .collection('sub');
+        const collection = untypedCollection.withConverter(
+          modelWithRefConverter
```
schmidt-sebastian, line 1630: I would expect that this converter does an instanceof check.

packages/firestore/test/integration/api/database.test.ts
```diff
@@ -1590,6 +1609,51 @@ apiDescribe('Database', (persistence: boolean) => {
...
       });
     });
+
+    it('Correct snapshot specified to fromFirestore() when registered with DocumentReference', () => {
+      return withTestDb(persistence, async db => {
+        const untypedDocRef = db.collection('/models').doc();
+        const docRef = untypedDocRef.withConverter(modelWithRefConverter);
+        await docRef.set(new ModelWithRef(docRef));
+        const docSnapshot = await docRef.get();
+        expect(untypedDocRef.isEqual(docSnapshot.data()!.ref)).to.be.true;
```
schmidt-sebastian, line 1619: This uses `refsEqual` which compares the path and the converter, but doesn't perform an instanceof check. You should add an instanceof check here and make sure that it fails before your changes.

packages/firestore/src/api/database.ts
```diff
@@ -753,7 +809,10 @@ export class DocumentReference<T = PublicDocumentData>
     return new DocumentReference<U>(
       this.firestore,
       this._delegate.withConverter(
-        converter as UntypedFirestoreDataConverter<U>
+        UntypedFirestoreDataConverterAdapter.getInstance(
+          this.firestore,
+          converter
```
schmidt-sebastian, line 814: If we store this as an instance member (i.e. rather than storing the original converter we store the `UntypedFirestoreDataConverterAdapter`), do we still need the `getInstance` method and its map? The converter would then be the same across calls. You can even pass the UntypedFirestoreDataConverterAdapter down from the CollectionReference to the DocumentReference if you make sure that you never wrap a converter twice.

packages/firestore/src/api/database.ts
schmidt-sebastian, line 556: Also the obligatory comment on naming: All classes in this file are essentially adapters. I would just use the public API name for this, which follows the usage across this file (e.g. DocumentReference, DocumentSnapshot and the newly introduced FirestoreDataConverter).

schmidt-sebastian (commented):
Thank you for this fix! I have a couple of lame suggestions for you, but I am impressed. Especially with the PR description :)

packages/firestore/test/integration/api/database.test.ts
dconeybe: I changed this to:

```
const ref = docSnapshot.data()!.ref!;
expect(ref).to.be.an.instanceof(firestore.DocumentReference);
expect(untypedDocRef.isEqual(ref)).to.be.true;
```

However, at runtime it fails with that error we were talking about a few days ago:

```
Error: Cannot find module '@firebase/firestore-types'
Require stack:
- packages/firestore/test/integration/api/database.test.ts
...
```

Do you know how I can work around this problem? I'm not sure what the "correct" way is to import `DocumentReference` from `database.ts`.

packages/firestore/test/integration/api/database.test.ts
dconeybe: I'm not sure what you are suggesting. Can you clarify?

packages/firestore/src/api/database.ts
dconeybe: Done.

dconeybe: Done.

packages/firestore/src/api/database.ts
dconeybe: I also realized that we need to key the WeakMap on both Firestore in addition to the converter instance, so that the object specified to the wrapper converter will be associated with the correct Firestore instance.

packages/firestore/test/integration/api/database.test.ts
schmidt-sebastian: Sorry, this is a bit more complicated than I anticipated. You need to get the actual instance of DocumentReference from "firebase_export" rather than the type. "firebase_export" is used so that it can be replaced in the minified test. Here is a patch with the changes: https://gist.github.com/schmidt-sebastian/7201fe874a6f463e0ad2256780386107

Note that you must also edit ../../integration/firestore since this test file is part of the minified integration tests that runs on CI.

packages/firestore/test/integration/api/database.test.ts
dconeybe: Ahh I see what you mean (from the Gist you posted: https://gist.github.com/schmidt-sebastian/7201fe874a6f463e0ad2256780386107). You were asking for an instanceof check in the fromFirestore() method. I've included this.

packages/firestore/test/integration/api/database.test.ts
dconeybe: Thanks for the gist! That's perfect. I've incorporated your suggestion.

schmidt-sebastian (approved):

PR #4286, opened 2021-01-13
Remove analytics prefix from settings()

hsubox76:
(This PR was originally "add initializeAnalytics()")

Remove `analytics` prefix from `settings()` method.

### DIFF
--- a/packages-exp/analytics-exp/src/api.ts
+++ b/packages-exp/analytics-exp/src/api.ts
@@ -24,8 +24,7 @@ import {
   AnalyticsCallOptions,
   CustomParams,
   EventNameString,
-  EventParams,
-  SettingsOptions
+  EventParams
 } from '@firebase/analytics-types-exp';
 import { Provider } from '@firebase/component';
 import {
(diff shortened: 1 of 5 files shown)


packages-exp/analytics-exp/src/api.ts
```diff
@@ -57,15 +58,38 @@ declare module '@firebase/component' {
   }
 }
 
+/**
+ * Firebase Analytics initialization function that allows custom
+ * settings. Can be called at most once. If no custom settings
```
Feiyang1, line 63: Should it throw if it's called more than once, or after getAnalytics() is called for the same app instance?

Also, `settings()` manages a global state that should not change from instance to instance. Does it even make sense to call it as part of `initializeAnalytics()` which is creating an instance. It seems to make more sense to let people call `settings()` manually once before calling any `getAnalytics()`. So perhaps we don't need `initializeAnalytics()` ?

hsubox76: Changed PR to just remove analytics prefix from analyticsSettings.

Feiyang1 (approved):

Issue #462, opened 2018-01-23
FR: please add function Auth.getCurrentUser() as a promise that will return current user or null


Thaina:
* Firebase SDK version: 4.9.0
  * Firebase Product: auth

`Auth.currentUser` object is unreliable to get user from persistence storage. If the authentication system could be pending and don't really have ensure finishing endpoint then you should have a promise for it

So I want to have promise function `Auth.getCurrentUser()` in addition to `onAuthStateChange` that would eventually return user but would return null if no user exist in persistence storage or user require sign in from expired token

bojeil-google:
You can easily implement that on your own with a couple of lines:
```
function getCurrentUser(auth) {
  return new Promise((resolve, reject) => {
     const unsubscribe = auth.onAuthStateChanged(user => {
        unsubscribe();
        resolve(user);
     }, reject);
  });
}
```

Thaina:
@bojeil-google If the user does not signed in will that promise would ever fired for null?

And I think this function is really common enough to included in the SDK

bojeil-google:
Yes that promise will fire. `onAuthStateChanged` always triggers with the initial state even when it is null. That is what you care for here, the current state. After that is determined, you unsubscribe.

Thaina:
I see but anyway if it really just a couple line and will be the same for everyone so why it not be inside SDK. Did you not consider it useful?

And also that function would fire only one time at first call and will not fire again in second call am I right? What I want from `getCurrentUser()` function is consistently return null if it already checking in persistence storage, or return the actual current user after that

At least it should have a function like `getPreparedAuthStated()` or `isPreparedAuthFinished`

bojeil-google:
I have no idea what you mean by: " consistently return null if it already checking in persistence storage, or return the actual current user after that".

Thaina:
@bojeil-google If I don't misunderstand, the `onAuthStateChanged in promise` function you present will work only on the first call. If I call that function on the second time it will never return because the auth state never change again after that

This is not what I want from `getCurrentUser()`. I want this function to actively checking that user is not really signed in before. I want it to be promise just in case that firebase auth does not finish initializing yet, in that case the user was signed in before but firebase itself are not ready. So return promise is better solution

What I want is
- If firebase auth preparation is not finished : return promise to wait
- If firebase auth preparation is finished but user is not signed in : resolve null from the promise
- If firebase auth preparation is finished and user is signed in : resolve currentUser from the promise

[...]

PR #4288, opened 2021-01-13
Issue #462 fix : Add new method that returns the logged-in user as Promise : 

harshshredding:
Issue https://github.com/firebase/firebase-js-sdk/issues/462
`onAuthStateChanged` has the following caveats that seem to be making it difficult for new-comers to get started with `firebase-auth` : 
- The method name doesn't clearly indicate a very common use-case: getting the current user info. Also, the method name implies that some state has to be changed in order for the callback to be triggered, but the callback is also triggered when auth is not pending (no user is logged in).
- Callbacks have caveats when compared to promises, which can be chained and utilized with the `async` keyword.

I was asked to tag @avolkovi and @bojeil-google.

### DIFF
--- a/packages/auth/src/auth.js
+++ b/packages/auth/src/auth.js
@@ -1596,6 +1596,21 @@ fireauth.Auth.prototype.onAuthStateChanged = function(
       opt_completed);
 };
 
+/**
+ * @return {Promise} The fireauth.AuthUser if a user successfully logged in.
+ * Null if no user is logged in.
+ */
+fireauth.Auth.prototype.getSignedInUser = function() {
+  return new Promise((resolve, reject) => {
+    if (this.currentUser_()) {
+      resolve(this.currentUser_());
+    }
+    const unsubscribe = this.onAuthStateChanged(user => {
+      unsubscribe();
+      resolve(user);
+    }, reject);
+  });
+}
 
 /**
  * Returns an STS token. If the cached one is unexpired it is directly returned.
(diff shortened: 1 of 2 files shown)


samhorlbeck:
Hi @harshshredding, thank you very much for this pull request! API changes of this nature go through an internal process for approval. I'm sponsoring this and will get back to you in a few weeks!

sam-gc:
Hi @harshshredding, thanks so much for your patience. After an internal review, we've concluded that this function should **not** itself return a user object. Instead, the function should return a `Promise<void>` that resolves when the auth state is ready and settled (i.e. when reading `auth().currentUser` is valid).

That function should be called `authStateReady()` and have the following signature:
```ts
interface Auth {
  authStateReady(): Promise<void>;
}
```

The function could then be used like:
```js
async function doStuff() {
  await firebase.auth().authStateReady();
  nowSafeToRead(firebase.auth().currentUser);
}
```

I believe your function in this PR can be easily updated to do the same thing but simply not resolve with any value.

harshshredding:
@sam-gc  I will update the PR very soon. Thanks a lot for the feedback.

Deliana90:
wow

Malix-Off:
@harshshredding
> https://github.com/firebase/firebase-js-sdk/pull/4288#event-8542325400

???
What happened?

harshshredding:
Sorry guys, I don't have the bandwidth. @sam-gc can you assign this to someone else?

prameshj:
> Sorry guys, I don't have the bandwidth. Sam can you assign this to someone else?

We got API review feedback to expose this slightly differently. We are working on it in https://github.com/firebase/firebase-js-sdk/pull/7384

PR #4296, opened 2021-01-15
Remove prepare script in our projects

Feiyang1:
`prepare` script runs automatically before `npm publish`, but it is redundant in our setup because we build everything before running `npm publish`. Removing it will speed up our release greatly.


package.json
```diff
@@ -24,8 +24,7 @@
     "dev": "lerna run --parallel --scope @firebase/* --scope firebase --scope rxfire dev",
     "build": "lerna run --scope @firebase/app-exp build:deps && lerna run --scope @firebase/* --scope firebase --scope rxfire --ignore @firebase/app-exp build",
     "build:exp": "lerna run --scope @firebase/*-exp --scope @firebase/*-compat --scope firebase-exp build",
-    "build:release": "lerna run --scope @firebase/app-exp build:deps && lerna run --scope @firebase/* --scope firebase --scope rxfire --ignore @firebase/*-exp --ignore @firebase/*-compat prepare",
-    "build:exp:release": "yarn --cwd packages/app build:deps && lerna run --scope @firebase/*-exp --scope @firebase/*-compat --scope firebase-exp prepare && yarn --cwd packages-exp/app-exp typings:public",
+    "build:release": "lerna run --scope @firebase/app-exp build:deps && lerna run --scope @firebase/* --scope firebase --scope rxfire --ignore @firebase/*-exp --ignore @firebase/*-compat build",
```
hsubox76, line 27: We no longer need a build:exp:release script to build @firebase/app first?

Feiyang1: I don't think we use this npm script as we use the exp release script when publishing. Is it useful for you?

hsubox76: Seems fine to remove, just checking the history and it seems it was added in June last year because something was breaking: https://github.com/firebase/firebase-js-sdk/pull/3301/files Was this before exp release was built with the script?

hsubox76 (approved):

PR #4298, opened 2021-01-15
Add toJSON() support for Firestore classes

### LINKED ISSUE #4258: Enumerable circular references in many firestore objects

  * Operating System version: Win 10
  * Browser version: 87.0.4280.88
  * Firebase SDK version: 8.2.1
  * Firebase Product: firestore+core

Many of the firestore objects have enumerable/public properties that result in circular references when deeply scanning through these objects, for example when `JSON.stringify` a DocumentReference. The JSON example is just the most common deep scan method that might be applied to the object, there are other cases where deep scanning is used that can run foul of these circular references.

In our specific application we're using Vue + Vuex + Firestore. We store the firestore documents in our vuex store which exposes these objects to the Vue reactivity system. While this works most of the time there are a few standard corner cases where cyclic references cause issues for us

*Deep watchers*
If we have watchers that observe any object that contains a firestore reference then this causes the Vue reactivity system to stack overflow while it follows the infinite loop.

*Vue(x) Dev Tools*
When passing data between the page context and the dev tools/plugin context the Vue dev tools plugin attempts to serialise the Vuex state to a string using `JSON.stringify`. If the state contains any firestore objects with circular references then this fails and the plugin falls back to a slower, more memory intensive, json process that deals with this. This causes the vue dev tools to crash frequently with out-of-memory. It's totally understandable that this could be a Vue dev tools issue, but it could also be solved by removing circular references in firebase.

#### Steps to reproduce:

Try to `JSON.stringify` any of the following directly, or objects that have properties that reference these:
- firebase.firestore.DocumentReference
- firebase.firestore.CollectionReference
- firebase.firestore.Query

There are probably plenty of others, but these are the ones that have been spotted in our app.

#### Relevant Code:

```javascript
const app = firebase.initialiseApp();
const db = app.firestore();
const docRef = db.doc('docs/doc');
JSON.stringify(docRef); // an example of walking the object tree
// Outputs: Uncaught TypeError: Converting circular structure to JSON
//    --> starting at object with constructor 'FirebaseAppImpl'
//    |     property 'firebase_' -> object with constructor 'Object'
//    |     property 'apps' -> object with constructor 'Array'
//    --- index 0 closes the circle
//    at JSON.stringify (<anonymous>)
```

### Places I've noticed circular references

- The `apps` list in `FirebaseAppImpl.firebase_.apps` has a reference to the enclosing `FirebaseAppImpl`
- The `providers` map in `FirebaseAppImpl.container.providers` contains a provider for the app that keeps track of the instance that contains the providers.

There might be others.

### Working around the issue

In our app we've had to work around the issue in some very specific ways.

1. We deep scan all firestore objects before they get put into our store and `Object.defineProperty(..., {enumerable: false})` any `DocumentReference.*` we find. This fixes the vast majority of cases but has a performance impact and is flaky
2. We are careful to avoid exposing non-required firestore objects to vue or vuex. This is harder than you'd think as even arguments to mutators in vuex count as exposure

thebrianchen:
Fixes #4258.

Adding` toJSON()` on the Firebase instance breaks the circular reference between `firebase.app.container` and the Firestore instance. Similarly, adding `toJSON()` to the `AsyncQueue` and `ExponentialBackoff` allows them to be serialized despite the circular references. 

Some questions:
1. Should I include my Firebase changes in `app-compat`?
2. Should the firebase packages also get the same patch message, or should there be a separate patch message for them?
3. Should we add more precise return types for `toJSON()`? I'm afraid of locking them in and causing a breaking change if we ever want to change them in the future.

### DIFF
--- a/.changeset/clean-meals-double.md
+++ b/.changeset/clean-meals-double.md
@@ -1,7 +1,6 @@
 ---
 '@firebase/app-compat': patch
 '@firebase/app': patch
-'@firebase/app-types': patch
 '@firebase/firestore': patch
 ---
 
--- a/.changeset/clean-meals-double.md
+++ b/.changeset/clean-meals-double.md
@@ -1,0 +1,8 @@
+---
+'@firebase/app-compat': patch
+'@firebase/app': patch
+'@firebase/app-types': patch
+'@firebase/firestore': patch
+---
+
+Firestore classes like DocumentReference and Query can now be serialized to JSON (#4258)
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -112,6 +112,8 @@ export interface FirebaseNamespace {
   // Sets log handler for all Firebase components.
   onLog(logCallback: LogCallback, options?: LogOptions): void;
 
+  toJSON(): object;
+
   // The current SDK version.
   SDK_VERSION: string;
 }
--- a/packages/app-types/index.d.ts
+++ b/packages/app-types/index.d.ts
@@ -112,8 +112,6 @@ export interface FirebaseNamespace {
   // Sets log handler for all Firebase components.
   onLog(logCallback: LogCallback, options?: LogOptions): void;
 
-  toJSON(): object;
-
   // The current SDK version.
   SDK_VERSION: string;
 }
--- a/packages/firestore/src/lite/database.ts
+++ b/packages/firestore/src/lite/database.ts
@@ -136,7 +136,7 @@ export class FirebaseFirestore implements FirestoreService {
 
   toJSON(): object {
     return {
-      app: this.app.name,
+      app: this._app,
       databaseId: this._databaseId,
       settings: this._settings
     };
--- a/packages/firestore/src/lite/database.ts
+++ b/packages/firestore/src/lite/database.ts
@@ -134,6 +134,13 @@ export class FirebaseFirestore implements FirestoreService {
     return this._terminateTask;
   }
 
+  toJSON(): object {
+    return {
+      databaseId: this._databaseId,
+      settings: this._settings
+    };
+  }
+
   /**
    * Terminates all components used by this client. Subclasses can override
    * this method to clean up their own dependencies, but must also call this
--- a/packages/firestore/src/lite/database.ts
+++ b/packages/firestore/src/lite/database.ts
@@ -136,6 +136,7 @@ export class FirebaseFirestore implements FirestoreService {
 
   toJSON(): object {
     return {
+      app: this.app.name,
       databaseId: this._databaseId,
       settings: this._settings
     };
--- a/packages/firestore/test/unit/api/database.test.ts
+++ b/packages/firestore/test/unit/api/database.test.ts
@@ -32,6 +32,10 @@ describe('CollectionReference', () => {
     expectEqual(collectionReference('foo'), collectionReference('foo'));
     expectNotEqual(collectionReference('foo'), collectionReference('bar'));
   });
+
+  it('JSON.stringify() does not throw', () => {
+    JSON.stringify(collectionReference('foo'));
+  });
 });
 
 describe('DocumentReference', () => {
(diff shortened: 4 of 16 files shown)


schmidt-sebastian (commented):
I think the code change is fine but I disagree a bit with the direction (but can certainly be convinced). We have our own internal data structures, and we shouldn't make an effort to make the "consumable" by our users. This makes our life more complicated and gives off the impression that we will support the output format going forward. It also convolutes the returned data, which very little additional benefit to the user.

Consider this: What does the user want when they call toJSON on a DocumentReference? They probably would prefer something like:

```
{ 
 path: 'coll/doc',
 firestore: {project: 'foo', database: '(default)'}
 converter: null
 }
```

over 
```
{ 
 path: 'coll/doc',
 firestore: {project: 'foo', database: '(default)', asyncQueue: {...}, firestoreClient: {...} }
 converter: null
 }
```


The first one is more concise, provides the information the user likely needs, and is less likely to change over time.

In case you are wondering where my argument is going (and if it is going anywhere): I think we should just implement toJSON on FirebaseFirestore and call it a day. The toJSON should have enough information to help the developer distinguish it from other instances, that is - database ID and FirebaseApp (and maybe others, but select instance members).

And yes, this should also work for Firbease@exp.

Let me know if you want to talk about this more.

thebrianchen:
@schmidt-sebastian Thanks for the suggestion -- your proposal creates a much more simplified API surface and doesn't require us to maintain a bunch of toJSON() fields. I ended up only including the databaseId and settings, since the project/database fields are not available by default. I'm not sure if I should expand the individual fields or add a return type. WDYT?

packages/firestore/test/unit/api/database.test.ts
```diff
@@ -72,13 +80,24 @@ describe('DocumentSnapshot', () => {
       documentSnapshot('rooms/bar', { a: 1 }, false)
     );
   });
+
+  it('JSON.stringify() does not throw', () => {
+    JSON.stringify(documentSnapshot('foo/bar', { a: 1 }, true));
+    JSON.stringify(documentSnapshot('foo/bar', { a: 1 }, false));
+    JSON.stringify(documentSnapshot('foo/bar', null, true));
+    JSON.stringify(documentSnapshot('foo/bar', null, false));
```
schmidt-sebastian, line 88: Do these four cases test different paths in a toJSON call?

packages/firestore/src/lite/database.ts
```diff
@@ -134,6 +134,13 @@ export class FirebaseFirestore implements FirestoreService {
     return this._terminateTask;
   }
 
+  toJSON(): object {
+    return {
+      databaseId: this._databaseId,
```
schmidt-sebastian, line 139: Should we include the app as well? I think the developers might want to see the app name here, since it acts as the main identifier for all Firebase components.

schmidt-sebastian (approved):
I think we should add `app` to Firebase.toJSON as well, but otherwise this LGTM. You might want to verify that you are on the right prettier before submitting (but you might be and the code style may be outdated)

packages/app-types/index.d.ts
```diff
@@ -112,6 +112,8 @@ export interface FirebaseNamespace {
   // Sets log handler for all Firebase components.
   onLog(logCallback: LogCallback, options?: LogOptions): void;
 
+  toJSON(): object;
```
schmidt-sebastian, line 115: You should also modify the Firestore and the Firebase types if you modify this (firestore-types/index.d.ts and firebase/index.d.ts)

thebrianchen: done.

Feiyang1 (commented):
LGTM. Can you please add a test for `toJSON()` in @firebase/app?

thebrianchen:
> LGTM. Can you please add a test for `toJSON()` in @firebase/app?

@Feiyang1 Done. Also added support for toJSON() in FirebaseAppLite.

.changeset/clean-meals-double.md
```diff
@@ -0,0 +1,8 @@
+---
+'@firebase/app-compat': patch
+'@firebase/app': patch
+'@firebase/app-types': patch
```
Feiyang1, line 4:
```suggestion
```

Feiyang1 (approved):

PR #4299, opened 2021-01-15
RTDB get shouldn't send requests for listened-to data

IanWyszynski:
https://b.corp.google.com/issues/177263312

### DIFF
--- a/.changeset/hip-turtles-travel.md
+++ b/.changeset/hip-turtles-travel.md
@@ -2,4 +2,4 @@
 '@firebase/database': patch
 ---
 
-RTDB query gets shouldn't send server requests for listened-to data
+get()s issued for queries that are being listened to no longer send backend requests.
--- a/.changeset/hip-turtles-travel.md
+++ b/.changeset/hip-turtles-travel.md
@@ -1,0 +1,5 @@
+---
+'@firebase/database': patch
+---
+
+RTDB query gets shouldn't send server requests for listened-to data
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
> https://b.corp.google.com/issues/177263312

FYI - we usually don't post links to Corp on GitHub, since they error our for external users. I usually just post b/xxxxxx or go/ links.

.changeset/hip-turtles-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/database': patch
+---
+
+RTDB query gets shouldn't send server requests for listened-to data
```
schmidt-sebastian, line 5: This might need a little rewrite. Maybe something like

`A `get()` issued for a query that is already being listened to no longer sends a request to the backend."

packages/database/src/core/Repo.ts
```diff
@@ -307,29 +307,28 @@ export class Repo {
...
-   * exists.
+   * If the client is not connected, this method will try to
+   * establish a connection and request the value for `query`. If
+   * the client is not able to retrieve the query result, it reports
+   * an error.
    *
    * @param query - The query to surface a value for.
    */
   getValue(query: Query): Promise<DataSnapshot> {
+    const cached = this.serverSyncTree_.calcCompleteEventCache(query.path);
```
schmidt-sebastian, line 322: Maybe add a comment that says that this work on Web since we only cache active queries.

schmidt-sebastian (approved):

PR #4308, opened 2021-01-19
Make Firestore collection and doc getters generic

btoo:
This enables the ability to define the `ReturnType` of Firestore's `collection` and `doc` getters


btoo:
@googlebot I signed it!

PR #4310, opened 2021-01-19
Package analytics-exp

hsubox76:
- Create firebase-exp/analytics-exp entry point.
- Make sure to export public types from main package.


packages-exp/analytics-exp/src/constants.ts
```diff
@@ -43,7 +43,7 @@ export const enum GtagCommand {
  *
  * @public
  */
-export const enum EventName {
+export enum EventName {
```
Feiyang1, line 46: enum is not treeshakeable. It's helpful for developers to see what events are available, but we also define function overloads for all of them, which should provide the same convenience. Since it's kinda of a big object and will likely grow in size and we have event name overloads for logEvent(), I don't think we should export it.

I also notice that it is only used in tests internal, so perhaps we can remove it completely or move it to a test only file?

packages-exp/analytics-exp/src/index.ts
```diff
@@ -80,6 +78,18 @@ function registerAnalytics(): void {
...
 }
 
-export { factory, resetGlobalVars, getGlobalVars };
-
 registerAnalytics();
+
+export * from './api';
+
+export {
+  DataLayer,
```
Feiyang1, line 86: What is the use case for DataLayer? It feels like an internal type that's only useful for us.

packages-exp/analytics-exp/src/constants.ts
hsubox76: Ok I guess they can rely on the function overloads. Removed. Removed from test files too since the tests should probably call the functions as users would - with string literals.

packages-exp/analytics-exp/src/index.ts
hsubox76: You're right, removed.

Feiyang1 (approved):

PR #4315, opened 2021-01-20
Create storage-exp packaging scripts

hsubox76:
Set up scripts to publish storage-exp, similar to Firestore's.

I have a question:
I set up the script based on Firestore's pattern. Based on this, the `typings` field in `packages-exp/firebase-exp/storage/package.json` points to "dist/storage/index.d.ts". If I look at that generated file, it contains one line `export * from '@firebase/storage';`. The same thing happens in Firestore, with `firestore` instead of `storage` of course. Is that right? I get a little confused about the firebase-exp entry points.

### DIFF
--- a/scripts/exp/prepare-storage-for-exp-release.ts
+++ b/scripts/exp/prepare-storage-for-exp-release.ts
@@ -45,15 +45,15 @@ export async function prepare() {
     '@firebase/app-types-exp': '0.x'
   };
 
-  packageJson.main = expPackageJson.main.replace('../', './exp');
-  packageJson.module = expPackageJson.module.replace('../', './exp');
-  packageJson.browser = expPackageJson.browser.replace('../', './exp');
+  packageJson.main = expPackageJson.main.replace('./', 'exp/');
+  packageJson.module = expPackageJson.module.replace('./', 'exp/');
+  packageJson.browser = expPackageJson.browser.replace('./', 'exp/');
   delete packageJson['esm2017'];
 
-  packageJson.typings = expPackageJson.typings.replace('../', './exp');
+  packageJson.typings = expPackageJson.typings.replace('./', 'exp/');
 
   // include files to be published
-  packageJson.files = [...packageJson.files, packageJson.typings];
+  packageJson.files = ['exp/dist'];
 
   // update package.json files
   await writeFile(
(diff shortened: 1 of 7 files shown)


Feiyang1:
Yes, it looks correct
> I have a question:
I set up the script based on Firestore's pattern. Based on this, the typings field in packages-exp/firebase-exp/storage/package.json points to "dist/storage/index.d.ts". If I look at that generated file, it contains one line export * from '@firebase/storage';. The same thing happens in Firestore, with firestore instead of storage of course. Is that right? I get a little confused about the firebase-exp entry points.

scripts/exp/prepare-storage-for-exp-release.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+import { projectRoot, readPackageJson } from '../utils';
+import { writeFile as _writeFile, readFile as _readFile } from 'fs';
+import { promisify } from 'util';
+
+const writeFile = promisify(_writeFile);
+const packagePath = `${projectRoot}/packages/storage`;
+
+//
```
Feiyang1, line 25:
```suggestion
```

scripts/exp/prepare-storage-for-exp-release.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  packageJson.peerDependencies = {
+    '@firebase/app-exp': '0.x',
+    '@firebase/app-types-exp': '0.x'
+  };
+
+  packageJson.main = expPackageJson.main.replace('./', 'exp/');
+  packageJson.module = expPackageJson.module.replace('./', 'exp/');
+  packageJson.browser = expPackageJson.browser.replace('./', 'exp/');
+  delete packageJson['esm2017'];
```
Feiyang1, line 51: Is there no esm2017 build?

Feiyang1 (approved):
LGTM. Could you run `yarn release:exp --dry-run` to verify the end state looks good before merging the PR?

scripts/exp/prepare-storage-for-exp-release.ts
hsubox76: I copied this line from the Firestore script but I assume the intent is that the only ES build is esm2017 now, which is pointed to by both `module` and `browser`, so there's no need for a separate `esm2017` field.

PR #4327, opened 2021-01-22
Fix analytics-exp type exports

hsubox76:
Separate type import and export statements. Transformer is causing some trouble with `export type`.


scripts/exp/release.ts
```diff
@@ -216,6 +216,26 @@ async function buildPackages() {
     }
   );
 
+  // Build app-exp but revert typings
+  await spawn(
```
Feiyang1: The changes in this file doesn't seem necessary. Lerna should correctly build app-exp first since it's a dependency of other packages.
Reverting to internal types is also not needed, since we set `abortOnError: false` in the release build for every packages.

I did a dry run and everything works fine without these changes.

Feiyang1 (approved):

PR #4337, opened 2021-01-24
Add bundle conformance tests to verify documents are loaded correctly.

wu-hui:


packages/firestore/src/core/bundle_impl.ts
```diff
@@ -207,7 +207,7 @@ export function bundleInitialProgress(
     documentsLoaded: 0,
     bytesLoaded: 0,
     totalDocuments: metadata.totalDocuments!,
-    totalBytes: metadata.totalBytes!
+    totalBytes: parseInt(metadata.totalBytes!.toString(), 10)
```
schmidt-sebastian, line 210: Please use `normailizeNumber` for this: https://osscs.corp.google.com/firebase-sdk/firebase-js-sdk/+/master:packages/firestore/src/model/normalize.ts;l=73?q=normalizeNumber&ss=firebase-sdk%2Ffirebase-js-sdk

This is the function that we use in other places to deserialize large numbers.

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -252,3 +264,178 @@ apiDescribe('Bundles', (persistence: boolean) => {
...
+  ];
+
+  // This is built by Java(proto3 json formatter) with above document and one additional field of `pathValue: db.doc('col1/ref1')`.
+  const BUNDLES_PROTOB3_CONFORMANCE_TEMPLATE = [
```
schmidt-sebastian, line 298: What does the `B` stand for?

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -252,3 +264,178 @@ apiDescribe('Bundles', (persistence: boolean) => {
...
+    zeroDateValue: new Timestamp(0, 0),
+    arrayValue: ['foo', 42, 'bar'],
+    emptyArray: [],
+    nilValue: null,
+    geoPointValue: new GeoPoint(50.1430847, -122.947778),
+    zeroGeoPointValue: new GeoPoint(0, 0),
+    bytesValue: Blob.fromUint8Array(new Uint8Array([0x01, 0x02]))
+  };
+
+  // This is built by Node(protobuf.js) with above document and one additional field of `pathValue: db.doc('col1/ref1')`.
```
schmidt-sebastian, line 290: Note that this is now a strange mix of ProtobufJS and Proto3 JSON.

packages/firestore/test/integration/api/bundle.test.ts
```diff
@@ -252,3 +264,178 @@ apiDescribe('Bundles', (persistence: boolean) => {
     });
   });
 });
+
+apiDescribe('Bundles conformance', (persistence: boolean) => {
+  const CONFORMANCE_DOC = {
```
schmidt-sebastian, line 269: The approach of "hardcoding" the JSON strings from Node and Java has some merits, but since it is Monday morning and we have a full week ahead of us, I would like to point out that you could also use the Web serializer to built both Proto3 JSON and ProtobufJS documents in the tests themselves. This approach has some benefits (less code to copy/paste, less chance that the code will go out of sync) and some downsides (less portable, may not catch bugs in the Node/Java clients) so I will leave the decision to you.

If we do decide to only support Proto3 JSON, we can also get rid of one of the tests here.

packages/firestore/src/core/bundle_impl.ts
wu-hui: Cool!

packages/firestore/test/integration/api/bundle.test.ts
wu-hui: Stands for `I-cannot-type` i guess.

packages/firestore/test/integration/api/bundle.test.ts
wu-hui: I added a todo instead, my 3 minute research did not tell me clearly how to do it..

schmidt-sebastian: We might want to spend some more time on this since adding this and then removing it right away just adds code churn. You should be able to call [wrap()](https://github.com/firebase/firebase-js-sdk/blob/4090271bb71023c3e6587d8bd8315ebf99b3ccd7/packages/firestore/test/util/helpers.ts#L178) with `CONFORMANCE_DOC` and get a JSON encoded Value proto. You may have to add an overload so you can control the `useProto3JSON` flag (and as discussed, this only works in api_internal)

packages/firestore/test/integration/api/bundle.test.ts
schmidt-sebastian: Please fix though :)

packages/firestore/test/integration/api/bundle.test.ts
wu-hui: As discussed offline, this is harder than I expected, and does not bring much value because we still need to hard code elements that are not documents. Let's not do this for now?

PR #4340, opened 2021-01-25
Add url param to getStorage and update public Metadata types

hsubox76:
Add url option to `getStorage()` (should have been there, existed in previous API, factory is set up to accommodate it).

Split public `Metadata` type into 3, as it was in the previous API
- `SettableMetadata` - can be set using `updateMetadata()`
- `UploadMetadata` - can be set as part of an upload of a storage object (seems to be same as `SettableMetadata` but with additional `md5Hash`).
- `FullMetadata` - metadata returned from server after various operations

### DIFF
--- a/packages/storage-types/exp/index.d.ts
+++ b/packages/storage-types/exp/index.d.ts
@@ -178,50 +178,50 @@ export interface FullMetadata extends UploadMetadata {
   /**
    * The bucket this object is contained in.
    */
-  bucket?: string;
+  bucket: string;
 
   /**
    * The full path of this object.
    */
-  fullPath?: string;
+  fullPath: string;
 
   /**
    * The object's generation.
    * {@link https://cloud.google.com/storage/docs/generations-preconditions}
    */
-  generation?: string;
+  generation: string;
 
   /**
    * The object's metageneration.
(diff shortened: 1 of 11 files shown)


packages/storage-types/exp/index.d.ts
```diff
@@ -121,102 +121,112 @@ export interface ListResult {
...
+ * The full set of object metadata, including read-only properties.
+ * @public
+ */
+export interface FullMetadata extends UploadMetadata {
   /**
-   * A date string representing when this object was last updated.
+   * The bucket this object is contained in.
    */
-  updated: string;
+  bucket?: string;
```
schmidt-sebastian, line 181: Since this is not used for metadata that is returned from the backed, it should no longer be optional.

hsubox76: Is this guaranteed to always be returned from the backend? Is there a list of metadata properties that always come back from the backend?

schmidt-sebastian: Looks like there is a list here: https://cloud.google.com/storage/docs/json_api/v1/objects#resource

Feiyang1 (approved):

PR #4345, opened 2021-01-26
Remove Prototype-Patching in Repo_transaction

schmidt-sebastian:
This should allow us to build the SDK without relying on module side-effects.


Feiyang1 (approved):
Weird we wrote it this way before.

schmidt-sebastian:
Thanks for helping me notice that CI finally passed :)

PR #4346, opened 2021-01-26
Add useEmulator() to Storage

hsubox76:
Add `storage().useEmulator()` to current storage package and `useStorageEmulator()` to storage-exp.

Wait for emulator to be available before merging.


packages/storage/src/service.ts
```diff
@@ -133,7 +144,8 @@ function extractBucket(config?: FirebaseOptions): Location | null {
  * @param opt_url - gs:// url to a custom Storage Bucket
  */
 export class StorageService implements _FirebaseService {
-  readonly _bucket: Location | null = null;
+  _bucket: Location | null = null;
+  private _emulatorOrigin?: string;
```
schmidt-sebastian, line 148: Could this just be `host` and be initialized to DEFAULT_HOST? I think this would get rid of most "is host undefined" checks.

packages/storage/src/service.ts
```diff
@@ -160,6 +172,23 @@ export class StorageService implements _FirebaseService {
...
 
+  get emulatorOrigin(): string | undefined {
+    return this._emulatorOrigin;
+  }
+
+  /**
+   * Set emulator origin string for this service.
+   * @param origin Origin string in the form of http://[host]:[port]
+   */
+  set emulatorOrigin(origin: string | undefined) {
```
schmidt-sebastian, line 183: Do we need to make this callable with `undefined`?

.changeset/tricky-seahorses-look.md
```diff
@@ -0,0 +1,6 @@
+---
+'firebase': minor
+'@firebase/storage': minor
+---
+
+Add `storage().useEmulator()` method allowing users to set a storage emulator host and port.
```
Feiyang1, line 6: How about `Add `storage().useEmulator()` method to enable the storage emulator mode, allowing users to set a storage emulator host and port` ?

packages/storage/src/service.ts
```diff
@@ -160,6 +172,23 @@ export class StorageService implements _FirebaseService {
     }
   }
 
+  get emulatorOrigin(): string | undefined {
+    return this._emulatorOrigin;
+  }
+
+  /**
+   * Set emulator origin string for this service.
+   * @param origin Origin string in the form of http://[host]:[port]
```
Feiyang1, line 181:
```suggestion
   * @param origin - Origin string in the form of http://[host]:[port]
```

Feiyang1 (approved):

packages/storage/src/service.ts
hsubox76: Doesn't seem so, removed.

packages/storage/src/service.ts
hsubox76: Makes sense, done.

.changeset/tricky-seahorses-look.md
hsubox76: Done.

packages/storage/src/service.ts
hsubox76: Fixed.

packages/storage/src/implementation/url.ts
```diff
@@ -18,11 +18,16 @@
 /**
  * @fileoverview Functions to create and manipulate URLs for the server API.
  */
-import { DEFAULT_HOST } from './constants';
 import { UrlParams } from './requestinfo';
 
-export function makeUrl(urlPart: string): string {
-  return `https://${DEFAULT_HOST}/v0${urlPart}`;
+export function makeUrl(urlPart: string, host: string): string {
+  const protocolMatch = host.match(/^(\w+):\/\/.+/);
```
hsubox76, line 24: Added this because emulator uses http when production use defaults to https. Seems cleaner to let the host string contain a protocol or not, and if so, parse it in this one function, than to store protocol as a separate property on StorageService and get it everywhere service.host is also used.

NothingEverHappens (approved):

.changeset/tricky-seahorses-look.md
```diff
@@ -0,0 +1,8 @@
+---
+'firebase': minor
+'@firebase/storage': minor
+'@firebase/storage-types': minor
+---
+
+Add `storage().useEmulator()` method to enable emulator mode for storage, allowing users
```
NothingEverHappens, line 7: `.changeset/tricky-seahorses-look.md` what is this file?

packages/storage/src/implementation/url.ts
NothingEverHappens, line 24: nit/optional, I think using replace with callback would save some lines and be more readable

NothingEverHappens (approved):
Is this safe to merge now?

.changeset/tricky-seahorses-look.md
NothingEverHappens: nvm, figured it out

packages/storage/src/implementation/url.ts
hsubox76: I couldn't figure out how to easily set up a regex that would call the callback if the protocol wasn't there so I left it alone for now. We can always optimize any time so let me know if you have a snippet you want to put in.

vbeffa:
@hsubox76 is this available for use yet? From the docs it seems we need to download/reference `https://www.gstatic.com/firebasejs/${JSCORE_VERSION}/firebase.js` but I don't see any new tags on this project.

hsubox76:
It's merged and should go out in the release next Thursday.

vbeffa:
Thanks!

jmcdl:
Is this feature actually ready for firebase developers? I ask because as of v9.10.0 the firebase-tools CLI still does not include storage in the list of emulators that can be initialised.

vbeffa:
@jmcdl see firebase/firebase-tools#3256.

jmcdl:
@vbeffa thank you!

PR #4347, opened 2021-01-26
handle ignoreUndefinedProperties in set(merge: true)

wilhuff:
Previously any field in the document would be propagated into the
FieldMask regardless of whether or not it had an undefined value, which
led to the client acting as if the user had passed FieldValue.delete(),
which wasn't intended.

Fixes googleapis/nodejs-firestore#1392 for the JS SDK.

### DIFF
--- a/packages/firestore/src/api/user_data_reader.ts
+++ b/packages/firestore/src/api/user_data_reader.ts
@@ -741,7 +741,11 @@ export function parseData(
   } else {
     // If context.path is null we are inside an array and we don't support
     // field mask paths more granular than the top-level array.
-    if (context.path) {
+    //
+    // If the input is undefined it can never participate in the fieldMask. With
+    // `ignoreUndefinedProperties` set to false, `parseScalarValue` will reject
+    // an undefined value.
+    if (context.path && input !== undefined) {
       context.fieldMask.push(context.path);
     }
 
--- a/packages/firestore/src/api/user_data_reader.ts
+++ b/packages/firestore/src/api/user_data_reader.ts
@@ -738,14 +738,15 @@ export function parseData(
     // context.fieldMask and we return null as our parsing result.
     parseSentinelFieldValue(input, context);
     return null;
+  } else if (input === undefined && context.ignoreUndefinedProperties) {
+    // If the input is undefined it can never participate in the fieldMask, so
+    // don't handle this below. If `ignoreUndefinedProperties` is false,
+    // `parseScalarValue` will reject an undefined value.
+    return null;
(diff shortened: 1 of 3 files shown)


packages/firestore/test/integration/api/fields.test.ts
```diff
@@ -398,6 +398,15 @@ apiDescribe('`undefined` properties', (persistence: boolean) => {
     });
   });
 
+  it('are ignored in set({ merge: true })', () => {
+    return withTestDocAndSettings(persistence, settings, async doc => {
+      await doc.set({ foo: 'foo', 'bar': 'unchanged' });
```
schmidt-sebastian, line 403: I am surprised the pre-commit hook didn't drop the quotes around `bar`.

packages/firestore/src/api/user_data_reader.ts
```diff
@@ -741,7 +741,11 @@ export function parseData(
   } else {
     // If context.path is null we are inside an array and we don't support
     // field mask paths more granular than the top-level array.
-    if (context.path) {
+    //
+    // If the input is undefined it can never participate in the fieldMask. With
+    // `ignoreUndefinedProperties` set to false, `parseScalarValue` will reject
+    // an undefined value.
+    if (context.path && input !== undefined) {
```
schmidt-sebastian, line 748: It might be cleaner to pull the `undefined` check from `parseScalarValue` into this function. Something like:

```
if (lookslikeJson) { ... }
else if (input instanceof FieldValue) { ... }
else if (input == undefined && context.ignoreUndefinedProperties) {
    return null;
}  else { ... }
```

Optional.

schmidt-sebastian (approved):

packages/firestore/test/integration/api/fields.test.ts
wilhuff: Fixed.

packages/firestore/src/api/user_data_reader.ts
wilhuff: Good idea. Done.

schmidt-sebastian (approved):

brcarey:
Hi folks, should this work in firebase-admin@9.5.0?

schmidt-sebastian:
It should, but you may need to delete your package lock file before re-installing your dependencies. The fix for `@google-cloud/firestore` (which, by the way, was https://github.com/googleapis/nodejs-firestore/pull/1396) was released with version 4.9.1

brcarey:
Awesome, works - much appreciated

PR #4349, opened 2021-01-26
Update size analysis cli to support analyzing default export, side effect export and namespace export

Feiyang1:


repo-scripts/size-analysis/bundle-analysis.ts
```diff
@@ -410,8 +427,64 @@ function createEntryFileContent(bundleDefinition: BundleDefinition): string {
...
+
+/**
+ * In case a symbol with the same name is already imported from another module, we need to give this symbol another name
+ * using "originalname as anothername" syntax, otherwise it returns the original symbol name.
+ */
+function createSymbolName(symbol: string, symbolsCache: Set<string>): string {
+  let nameToUse = symbol;
+  const max = 100;
+  while (symbolsCache.has(nameToUse)) {
+    nameToUse = `${symbol}_${Math.floor(Math.random() * max)}`;
```
hsubox76, line 477: Super edge case but is it safer to increment instead of random?

hsubox76 (approved):

repo-scripts/size-analysis/bundle-analysis.ts
Feiyang1: If there are more than 100 symbols with the same name, yes, but we likely will never encounter such scenario.
I chose to use random because it's faster than incrementing without maintaining additional cache. We could make incrementing O(1) by creating another cache that remembers the last number used for every symbol, but I think random is good enough.

PR #4362, opened 2021-01-29
Export public types from auth-exp

hsubox76:
Export public types from auth-exp main package entry point.

### DIFF
--- a/packages-exp/auth-exp/index.ts
+++ b/packages-exp/auth-exp/index.ts
@@ -16,7 +16,6 @@
  */
 
 import { FirebaseApp } from '@firebase/app-types-exp';
-import { Auth } from '@firebase/auth-types-exp';
 
 import { initializeAuth } from './src';
 import { registerAuth } from './src/core/auth/register';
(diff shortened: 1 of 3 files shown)


packages-exp/auth-exp/index.ts
```diff
@@ -25,6 +24,103 @@ import { browserLocalPersistence } from './src/platform_browser/persistence/loca
...
 import { browserPopupRedirectResolver } from './src/platform_browser/popup_redirect';
 
+import {
+  // Classes
+  ActionCodeURL,
+  AuthCredential,
+  EmailAuthProvider,
+  MultiFactorResolver,
+  OAuthCredential,
+  PhoneAuthCredential,
```
Feiyang1, line 34: We can't export classes as types because it allows people to new them. Should they be interface instead? If we want to export the actual classes, the implementation should be exported instead.

hsubox76: Seems like MultiFactorResolver can be an interface. The other ones are real classes that are already exported, so I am removing those type exports. They could maybe be removed from auth-types-exp/index.d.ts entirely but that's where the documentation comments are and the code class is linking to them using `@inheritDoc`.

Feiyang1 (approved):

PR #4363, opened 2021-01-29
Fix startAfter/endBefore for orderByKeys queries

IanWyszynski:
We should not be passing a fallback key to startAfter if we're querying on a key index. The indexValue is the key in this case.

### DIFF
--- a/packages/database/src/api/Query.ts
+++ b/packages/database/src/api/Query.ts
@@ -101,10 +101,7 @@ export class Query {
         'or equalTo() must be a string.';
       if (params.hasStart()) {
         const startName = params.getIndexStartName();
-        if (
-          startName !== MIN_NAME &&
-          !(params.hasStartAfter() && startName === MAX_NAME)
-        ) {
+        if (startName !== MIN_NAME) {
           throw new Error(tooManyArgsError);
         } else if (typeof startNode !== 'string') {
           throw new Error(wrongArgTypeError);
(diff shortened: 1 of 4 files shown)


packages/database/src/api/Query.ts
```diff
@@ -101,21 +101,15 @@ export class Query {
         'or equalTo() must be a string.';
       if (params.hasStart()) {
         const startName = params.getIndexStartName();
-        if (
```
IanWyszynski, line 104: These just revert a previous change.

packages/database/src/core/view/QueryParams.ts
```diff
@@ -289,13 +289,21 @@ export class QueryParams {
   }
 
   startAfter(indexValue: unknown, key?: string | null): QueryParams {
-    let childKey: string;
-    if (key == null) {
-      childKey = MAX_NAME;
+    let params: QueryParams;
+    if (this.index_ === KEY_INDEX) {
+      if (typeof indexValue === 'string') {
```
IanWyszynski, line 294: This check makes it so that startAfter still throws exceptions for incorrectly typed keys.

schmidt-sebastian: Where does it throw? Should we throw at the callsite to preserve the stacktrace of the original mistake?

IanWyszynski: It throws in validateQueryEndpoints_: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/api/Query.ts#L110, which is the same place that startAt() and endAt() throw for this case. 

I think that the way it's done now the original mistake, passing a non-string key to `startAfter()` should still be clear. I'll also update the error message there.

schmidt-sebastian (approved):
Please add change log entry.

PR #4378, opened 2021-02-01
Fix (silently) broken Auth-exp tests

samhorlbeck:


packages-exp/auth-exp/src/platform_browser/popup_redirect.test.ts
```diff
@@ -74,10 +72,17 @@ describe('platform_browser/popup_redirect', () => {
           })
       } as unknown) as gapi.iframes.Context)
     );
+
+    authWindow._window().gapi = ({
```
yuchenshi, line 76: Can we use `sinon` for this?

yuchenshi (approved):

packages-exp/auth-exp/src/platform_browser/popup_redirect.test.ts
samhorlbeck: Done.

PR #4379, opened 2021-02-01
Add initial bare-bones Cordova support / popup redirect

samhorlbeck:


packages-exp/auth-exp/index.cordova.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is the file that people using React Native will actually import. You
```
yuchenshi, line 19:
```suggestion
 * This is the file that people using Cordova will actually import. You
```

packages-exp/auth-exp/index.cordova.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Core functionality shared by all clients
+export * from './src';
+
+// Cordova also supports indexedDB / browserSession / browserLocal
+export { indexedDBLocalPersistence } from './src/platform_browser/persistence/indexed_db';
+export { browserLocalPersistence } from './src/platform_browser/persistence/local_storage';
+export { browserSessionPersistence } from './src/platform_browser/persistence/session_storage';
+
+export { cordovaPopupRedirectResolver } from './src/platform_cordova/popup_redirect';
```
yuchenshi, line 41: Do we need to export this?

packages-exp/auth-exp/src/platform_cordova/popup_redirect.test.ts
```diff
@@ -0,0 +1,131 @@
...
+    },
+    InAppBrowser: {
+      open: () => {}
+    }
+  };
+  win.universalLinks = {
+    subscribe: () => {}
+  };
+  win.BuildInfo = {
+    packageName: 'com.name.package'
```
yuchenshi, line 118:
```suggestion
    packageName: 'com.example.name.package'
```

packages-exp/auth-exp/src/platform_cordova/popup_redirect.test.ts
```diff
@@ -0,0 +1,131 @@
...
+      expect(() =>
+        resolver._openRedirect(
+          auth,
+          new GoogleAuthProvider(),
+          AuthEventType.SIGN_IN_VIA_REDIRECT
+        )
+      ).to.throw(FirebaseError, 'auth/invalid-cordova-configuration');
+    });
+
+    it('rejects if browsertab openUrl', () => {
```
yuchenshi, line 75:
```suggestion
    it('rejects if browsertab openUrl is missing', () => {
```

packages-exp/auth-exp/src/platform_cordova/popup_redirect.test.ts
```diff
@@ -0,0 +1,131 @@
...
+          auth,
+          new GoogleAuthProvider(),
+          AuthEventType.SIGN_IN_VIA_REDIRECT
+        )
+      ).to.throw(FirebaseError, 'auth/invalid-cordova-configuration');
+    });
+  });
+});
+
+function setExpectedPluginsButMissing(missingPlugins: string[]): void {
```
yuchenshi, line 99: IMO it's probably easier to just have this function mock the whole thing (with all plugins), and then write things like `delete window.universalLinks.subscribe` in individual tests. It's more straightforward and don't need the awkward object traversing below.

packages-exp/auth-exp/src/platform_cordova/popup_redirect.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    it('rejects if universal links is missing', () => {
+      setExpectedPluginsButMissing(['universalLinks.subscribe']);
+      expect(() =>
+        resolver._openRedirect(
+          auth,
+          new GoogleAuthProvider(),
+          AuthEventType.SIGN_IN_VIA_REDIRECT
+        )
+      ).to.throw(FirebaseError, 'auth/invalid-cordova-configuration');
```
yuchenshi, line 61: We should also test for the `missingPlugin` property. Same for tests below.

packages-exp/auth-exp/src/platform_cordova/popup_redirect.ts
```diff
@@ -0,0 +1,97 @@
...
+    return new Promise(() => {});
+  }
+  _isIframeWebStorageSupported(
+    _auth: Auth,
+    _cb: (support: boolean) => unknown
+  ): void {
+    throw new Error('Method not implemented.');
+  }
+  readonly _redirectPersistence = browserSessionPersistence;
+  _completeRedirectFn: () => Promise<null> = async () => null;
```
yuchenshi, line 54: Please move fields to the top

packages-exp/auth-exp/src/platform_cordova/popup_redirect.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function checkCordovaConfiguration(auth: Auth): void {
+  // Check all dependencies installed.
+  // https://github.com/nordnet/cordova-universal-links-plugin
+  // Note that cordova-universal-links-plugin has been abandoned.
+  // A fork with latest fixes is available at:
+  // https://www.npmjs.com/package/cordova-universal-links-plugin-fix
+  if (typeof window?.universalLinks?.subscribe !== 'function') {
+    _fail(auth, AuthErrorCode.INVALID_CORDOVA_CONFIGURATION, {
```
yuchenshi, line 64: These should use `_assert(condition, ...)` instead of `if-fail`.

packages-exp/auth-exp/src/platform_cordova/popup_redirect.ts
```diff
@@ -0,0 +1,97 @@
...
+  // https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-inappbrowser/
+  if (typeof window?.cordova?.InAppBrowser?.open !== 'function') {
+    _fail(auth, AuthErrorCode.INVALID_CORDOVA_CONFIGURATION, {
+      missingPlugin: 'cordova-plugin-inappbrowser'
+    });
+  }
+}
+
+/**
+ * An implementation of {@link @firebase/auth-types#PopupRedirectResolver} suitable for browser
```
yuchenshi, line 92:
```suggestion
 * An implementation of {@link @firebase/auth-types#PopupRedirectResolver} suitable for Cordova
```

yuchenshi (approved):

packages-exp/auth-exp/index.cordova.ts
samhorlbeck: Yes. It can be used independently (if you use `initializeAuth()` for example instead of `getAuth()`)

packages-exp/auth-exp/src/platform_cordova/popup_redirect.test.ts
samhorlbeck: Done. Still have to do a bit of a dance because the package doesn't mark those as optional. Still cleaner :+1:

PR #4380, opened 2021-02-02
Fix size analysis script issues

hsubox76:
Size analysis CI bugs started happening after we changed auth-exp to export const enum types from the main package.

I hope this is the most correct way to find and filter out const enums.
Also added 'tslib' as an external - otherwise rollup will do it anyway but give you a lot of warnings.

### DIFF
--- a/repo-scripts/size-analysis/analysis-helper.ts
+++ b/repo-scripts/size-analysis/analysis-helper.ts
@@ -225,7 +225,10 @@ export function extractDeclarations(
     } else if (ts.isVariableDeclaration(node)) {
       declarations.variables.push(node.name!.getText());
     } else if (ts.isEnumDeclaration(node)) {
+      // `const enum`s should not be analyzed. They do not add to bundle size and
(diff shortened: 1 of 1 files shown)


repo-scripts/size-analysis/analysis-helper.ts
```diff
@@ -224,7 +225,11 @@ export function extractDeclarations(
     } else if (ts.isVariableDeclaration(node)) {
       declarations.variables.push(node.name!.getText());
     } else if (ts.isEnumDeclaration(node)) {
-      declarations.enums.push(node.name.escapedText.toString());
+      if (
+        !node.modifiers?.some(mod => mod.kind === ts.SyntaxKind.ConstKeyword)
```
Feiyang1, line 229: Can we add some comments ?

Feiyang1 (approved):
LGTM.

repo-scripts/size-analysis/analysis-helper.ts
hsubox76: Added.

PR #4395, opened 2021-02-02
Dequeue limbo resolutions when their respective queries are stopped

dconeybe:
Fix a bug where enqueued limbo resolutions are left in the queue even after all targets that care about their resolutions are stopped. This erroneous behavior was found in and reported against the Android SDK (https://github.com/firebase/firebase-android-sdk/issues/2311) but also applies to the Web and iOS SDKs. This fix will be ported to those other SDKs. This bug was introduced when limbo resolution throttling was implemented almost a year ago (https://github.com/firebase/firebase-js-sdk/pull/2790).

### DIFF
--- a/.changeset/eight-bananas-nail.md
+++ b/.changeset/eight-bananas-nail.md
@@ -2,4 +2,4 @@
 '@firebase/firestore': patch
 ---
 
-Fix a bug where local cache inconsistencies were unnecessarily being resolved.
+Fixes a bug where local cache inconsistencies were unnecessarily being resolved.
--- a/.changeset/eight-bananas-nail.md
+++ b/.changeset/eight-bananas-nail.md
@@ -1,0 +1,5 @@
+---
+'@firebase/firestore': patch
+---
+
+Fix a bug where local cache inconsistencies were unnecessarily being resolved.
--- a/packages/firestore/src/core/sync_engine_impl.ts
+++ b/packages/firestore/src/core/sync_engine_impl.ts
@@ -113,6 +113,7 @@ import {
   ViewChange
 } from './view';
 import { ViewSnapshot } from './view_snapshot';
+import { ResourcePath } from '../model/path';
 
 const LOG_TAG = 'SyncEngine';
 
@@ -208,9 +209,10 @@ class SyncEngineImpl implements SyncEngine {
   queriesByTarget = new Map<TargetId, Query[]>();
   /**
    * The keys of documents that are in limbo for which we haven't yet started a
-   * limbo resolution query.
+   * limbo resolution query. The strings in this set are the result of calling
+   * `key.path.canonicalString()` where `key` is a `DocumentKey` object.
    */
-  enqueuedLimboResolutions: DocumentKey[] = [];
+  enqueuedLimboResolutions = new Set<string>();
   /**
    * Keeps track of the target ID for each document that is in limbo with an
    * active target.
@@ -876,6 +878,8 @@ function removeLimboTarget(
   syncEngineImpl: SyncEngineImpl,
   key: DocumentKey
 ): void {
+  syncEngineImpl.enqueuedLimboResolutions.delete(key.path.canonicalString());
+
   // It's possible that the target already got removed because the query failed. In that case,
   // the key won't exist in `limboTargetsByKey`. Only do the cleanup if we still have the target.
   const limboTargetId = syncEngineImpl.activeLimboTargetsByKey.get(key);
@@ -927,7 +931,7 @@ function trackLimboChange(
   const key = limboChange.key;
   if (!syncEngineImpl.activeLimboTargetsByKey.get(key)) {
     logDebug(LOG_TAG, 'New document in limbo: ' + key);
-    syncEngineImpl.enqueuedLimboResolutions.push(key);
+    syncEngineImpl.enqueuedLimboResolutions.add(key.path.canonicalString());
     pumpEnqueuedLimboResolutions(syncEngineImpl);
   }
 }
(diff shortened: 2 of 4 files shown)


.changeset/eight-bananas-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/firestore': patch
+---
+
+Fix a bug where local cache inconsistencies were unnecessarily being resolved.
```
schmidt-sebastian, line 5: Fixes or Fixed

packages/firestore/src/core/sync_engine_impl.ts
```diff
@@ -208,9 +209,10 @@ class SyncEngineImpl implements SyncEngine {
   queriesByTarget = new Map<TargetId, Query[]>();
   /**
    * The keys of documents that are in limbo for which we haven't yet started a
-   * limbo resolution query.
+   * limbo resolution query. The strings in this set are the result of calling
+   * `key.path.canonicalString()` where `key` is a `DocumentKey` object.
    */
-  enqueuedLimboResolutions: DocumentKey[] = [];
+  enqueuedLimboResolutions = new Set<string>();
```
schmidt-sebastian, line 215: We should maybe mention that we can use this datastructure in JS because it preserves insertion order.

packages/firestore/test/unit/specs/spec_test_runner.ts
```diff
@@ -1008,31 +1007,13 @@ abstract class TestRunner {
...
-          `but it was not in the queue (${actualLimboDocs.toString()})`
-      );
-    });
+    const actualLimboDocs = Array.from(
+      syncEngineGetEnqueuedLimboDocumentResolutions(this.syncEngine)
+    ).sort();
+    const expectedLimboDocs = Array.from(this.expectedEnqueuedLimboDocs, key =>
+      key.path.canonicalString()
+    ).sort();
+    expect(actualLimboDocs).to.deep.equal(expectedLimboDocs);
```
schmidt-sebastian, line 1016: To keep some of the descriptiveness of the original error, you should add a message argument (to the last function call) that mentions that this codepath validates limbo resolutions. Otherwise, an error will show up, but it will not be all that helpful.

Furthermore, you should use `to.have.members` and remove `sort()`.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -922,4 +922,149 @@ describeSpec('Limbo Documents:', [], () => {
...
+    [],
+    () => {
+      const doc1 = doc('collection1/doc', 1000, { key: 1 });
+      const query1 = query('collection1');
+      const filteredQuery1 = query('collection1', filter('key', '==', 1));
+
+      const doc2 = doc('collection2/doc', 1000, { key: 2 });
+      const query2 = query('collection2');
+      const filteredQuery2 = query('collection2', filter('key', '==', 2));
+      const filteredQuery3 = query('collection2', filter('key', '>=', 2));
```
schmidt-sebastian, line 937: Maybe:

s/filteredQuery2/filteredQuery2a
s/filteredQuery3/filteredQuery2b

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -922,4 +922,149 @@ describeSpec('Limbo Documents:', [], () => {
...
+          .expectLimboDocs(doc1.key)
+          .expectEnqueuedLimboDocs()
+
+          // Put doc1 into limbo in a different query; verify that another limbo
+          // resolution is neither started nor enqueued.
+          .userListens(filteredQuery2)
+          .expectEvents(filteredQuery2, { added: [doc1], fromCache: true })
+          .watchAcksFull(filteredQuery2, 1002)
+          .expectLimboDocs(doc1.key)
+          .expectEnqueuedLimboDocs()
```
schmidt-sebastian, line 1017: Should we "finish" this test and resolve the first limbo document, enqueue the second one and then resolve this one too?

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -922,4 +922,149 @@ describeSpec('Limbo Documents:', [], () => {
...
+          // verify that doc2 is not removed from the limbo resolution queue.
+          .userUnlistens(filteredQuery3)
+          .expectLimboDocs(doc1.key)
+          .expectEnqueuedLimboDocs(doc2.key)
+
+          // Stop the other query that enqueued a limbo resolution for doc2;
+          // verify that doc2 *is* removed from the limbo resolution queue.
+          .userUnlistens(filteredQuery2)
+          .expectLimboDocs(doc1.key)
+          .expectEnqueuedLimboDocs()
```
schmidt-sebastian, line 982: Nit: Both of the tests below seem to be subsets of this test. I would therefore move this test under the other tests to first verify the basic implementation before going into the more completed cases.

schmidt-sebastian (approved):

schmidt-sebastian:
Let me know if you change the spec tests so much that you want another review.

packages/firestore/test/unit/specs/spec_test_runner.ts
dconeybe: Done.

dconeybe:
> Let me know if you change the spec tests so much that you want another review.

I had a few questions based on the comments you left.

packages/firestore/test/unit/specs/limbo_spec.test.ts
```diff
@@ -922,4 +922,149 @@ describeSpec('Limbo Documents:', [], () => {
...
+
+          // Max out the number of active limbo resolutions.
+          .userListens(query1)
+          .watchAcksFull(query1, 1000, doc1)
+          .expectEvents(query1, { added: [doc1] })
+          .userUnlistens(query1)
+          .userListens(filteredQuery1)
+          .expectEvents(filteredQuery1, { added: [doc1], fromCache: true })
+          .watchAcksFull(filteredQuery1, 1001)
+          .expectLimboDocs(doc1.key)
```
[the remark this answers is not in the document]
schmidt-sebastian: I am going with "one of the above seems ok". I think the argument for consistency makes sense, and the test can generally be followed, as long as the person looking at it also looks at the actual contents of the docs/queries.

packages/firestore/test/unit/specs/limbo_spec.test.ts
schmidt-sebastian: Oh, I am not sure how I missed that this test is different in this way. I thought the second query added another limbo doc. Never mind.

schmidt-sebastian (approved):

PR #4409, opened 2021-02-03
Syntax fixes in 'app' packages

hsubox76:
Fix some things that will cause errors if upgrading to Typescript 4.1.+ (the upgrade isn't being done in this PR, other packages need to be fixed as well).

I think referencing `name` when it hadn't been passed in didn't show up as an error before because it's also a global variable?

### DIFF
--- a/packages-exp/app-exp/src/api.ts
+++ b/packages-exp/app-exp/src/api.ts
@@ -284,10 +284,7 @@ export function onLog(
   options?: LogOptions
 ): void {
   if (logCallback !== null && typeof logCallback !== 'function') {
-    throw ERROR_FACTORY.create(AppError.INVALID_LOG_ARGUMENT, {
-      // appName isn't used in INVALID_LOG_ARGUMENT error but ErrorFactory requires it
-      appName: ''
-    });
+    throw ERROR_FACTORY.create(AppError.INVALID_LOG_ARGUMENT);
   }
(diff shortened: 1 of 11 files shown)


packages-exp/app-exp/src/api.ts
```diff
@@ -285,7 +285,8 @@ export function onLog(
 ): void {
   if (logCallback !== null && typeof logCallback !== 'function') {
     throw ERROR_FACTORY.create(AppError.INVALID_LOG_ARGUMENT, {
-      appName: name
+      // appName isn't used in INVALID_LOG_ARGUMENT error but ErrorFactory requires it
+      appName: ''
```
Feiyang1, line 289: The correct fix should be change `type ErrorParams` in `errors.ts` to

```
type ErrorParams = {
  [AppError.NO_APP]: { appName: string },
  [AppError.BAD_APP_NAME]: { appName: string },
  [AppError.DUPLICATE_APP]: { appName: string },
  [AppError.APP_DELETED]: { appName: string },
  [AppError.INVALID_APP_ARGUMENT]: { appName: string }
};
```

Feiyang1 (changes requested):

packages-exp/app-exp/src/api.ts
hsubox76: Done.

Feiyang1 (approved):

Issue #4407, opened 2021-02-03
Storage SDK missing methods to reference multiple buckets


samtstern:
### [REQUIRED] Describe your environment

  * Operating System version: macOS
  * Browser version: N/A
  * Firebase SDK version: `exp`
  * Firebase Product: storage

### [REQUIRED] Describe the problem

Originally reported here:
https://github.com/firebase/firebase-js-sdk/discussions/4402

#### Relevant Code:

There seems to be no way to reproduce either of these snippets in the new modular SDK:

```js
const storage = firebase.app().storage("gs://my-custom-bucket");
```

```js
  const customApp = firebase.initializeApp({
    // ... custom stuff
  });

  // Get the default bucket from a custom firebase.app.App
  const storage = customApp.storage();

  // Get a non-default bucket from a custom firebase.app.App
  const storage = customApp.storage("gs://my-custom-bucket");
```

hsubox76:
You can do:
```
const storage = getStorage(app, "gs://my-custom-bucket")
```

The second param is named `url` in the definition so I guess it should probably be named "bucket,"  and it doesn't have an accompanying jsdoc description so that should be added. Is that enough or am I missing anything else that would be needed?

PR #4410, opened 2021-02-03
Fill out storage bucket url param

hsubox76:
Add jsdoc description and change name from `url` to `bucketUrl` to make it more clear. (FYI the param name in the original API is `url`).

Fixes https://github.com/firebase/firebase-js-sdk/issues/4407

### DIFF
--- a/packages/storage/exp/index.ts
+++ b/packages/storage/exp/index.ts
@@ -294,15 +294,22 @@ const STORAGE_TYPE = 'storage-exp';
  * Gets a Firebase StorageService instance for the given Firebase app.
  * @public
  * @param app - Firebase app to get Storage instance for.
+ * @param bucketUrl - The gs:// url to your Firebase Storage Bucket.
+ * If not passed, uses the app's default Storage Bucket.
  * @returns A Firebase StorageService instance.
  */
-export function getStorage(app: FirebaseApp, url?: string): StorageService {
+export function getStorage(
+  app: FirebaseApp,
+  bucketUrl?: string
+): StorageService {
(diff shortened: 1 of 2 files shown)


samtstern:
@hsubox76 to be clear this is what I see in my types today:
```
/**
 * Gets a Firebase StorageService instance for the given Firebase app.
 * @public
 * @param app - Firebase app to get Storage instance for.
 * @returns A Firebase StorageService instance.
 */
export declare function getStorage(app: FirebaseApp): StorageService;
```

So in #4407 I wasn't saying that the `url` parameter wasn't clear, I was saying that I don't have it at all.  Let me see if I can remove `node_modules` and see if maybe I missed an update to `exp`

samtstern (approved):

packages/storage/exp/index.ts
```diff
@@ -294,15 +294,22 @@ const STORAGE_TYPE = 'storage-exp';
...
  * @public
  * @param app - Firebase app to get Storage instance for.
+ * @param bucketUrl - The gs:// url to your Firebase Storage Bucket.
+ * If not passed, uses the app's default Storage Bucket.
  * @returns A Firebase StorageService instance.
  */
-export function getStorage(app: FirebaseApp, url?: string): StorageService {
+export function getStorage(
+  app: FirebaseApp,
+  bucketUrl?: string
```
schmidt-sebastian, line 303: gsUrl?

Feel free to keep as is.

schmidt-sebastian (approved):

Feiyang1 (approved):

PR #4413, opened 2021-02-04
Add the outward-bound part of the Cordova redirect flow to auth@exp

samhorlbeck:
This change handles the generation of the redirect URL including lots of platform-specific handling. This change allows `signInWithRedirect()` to successfully redirect to the oauth handler / IdP page. The event is not yet handled on return, however.


packages-exp/auth-exp/src/core/util/handler.ts
```diff
@@ -0,0 +1,120 @@
...
+  redirectUrl?: string;
+  v: string;
+  providerId?: string;
+  scopes?: string;
+  customParameters?: string;
+  eventId?: string;
+  tid?: string;
+} & { [key: string]: string | undefined };
+
+export function _getRedirectUrl(
```
samhorlbeck, line 56: This is moved from `platform_browser/popup_redirect`. The only change is that it now has an "additionalParams" parameter

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+  return chars.join('');
+}
+
+/**
+ * Computes the SHA-256 of a session ID. The SubtleCrypto interface is only
+ * available in "secure" contexts, which covers Cordova (which is served on a file
+ * protocol).
+ */
+async function computeSha256(sessionId: string): Promise<string> {
+  const encoder = new TextEncoder();
```
yuchenshi, line 185: What do we do for IE11? https://caniuse.com/textencoder says no support

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+  const buf = await crypto.subtle.digest('SHA-256', encoder.encode(sessionId));
```
yuchenshi, line 186: In IE11, this returns `CryptoOperation` (not Promise) and `await`-ing this won't give the right result.

See: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+  const arr = new Uint8Array(buf);
+
+  let hexString = '';
```
yuchenshi, line 189: nit: MDN has a more concise implementation here: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string

It involves `Array.from` but I think performance shouldn't be a concern here -- code is not on a critical path and Array size is constant.

yuchenshi (approved):

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
yuchenshi: FYI: `CryptoOperation` is from an earlier draft here: https://www.w3.org/TR/2013/WD-WebCryptoAPI-20130108/#cryptooperation-interface

tl;dr: we need to attach `oncomplete` / `onerror` handlers instead of `await`.

packages-exp/auth-exp/karma.conf.js
```diff
@@ -35,6 +35,8 @@ function getTestFiles(argv) {
     return ['src/**/*.test.ts', 'test/helpers/**/*.test.ts'];
   } else if (argv.integration) {
     return ['test/integration/**/*.test.ts'];
+  } else if (argv.cordova) {
+    return ['src/platform_cordova/**/*.test.ts'];
```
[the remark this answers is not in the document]
samhorlbeck: Which other tests?

packages-exp/auth-exp/src/model/popup_redirect.ts
```diff
@@ -86,7 +86,7 @@ export interface PopupRedirectResolver extends externs.PopupRedirectResolver {
     provider: externs.AuthProvider,
     authType: AuthEventType,
     eventId?: string
-  ): Promise<never>;
+  ): Promise<unknown>;
```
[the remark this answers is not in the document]
samhorlbeck: In Cordova environments, the redirect does not necessarily end the activity, so a call to `signInWithRedirect()` can actually resolve

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.test.ts
```diff
@@ -0,0 +1,274 @@
...
+  _checkCordovaConfiguration,
+  _generateHandlerUrl,
+  _generateNewEvent,
+  _performRedirect
+} from './utils';
+import { AuthEvent, AuthEventType } from '../../model/popup_redirect';
+import { GoogleAuthProvider } from '../../core/providers/google';
+import { AuthProvider } from '../../../internal';
+
+enum UA {
```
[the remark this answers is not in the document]
samhorlbeck: I had that originally but I liked this better lol

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.test.ts
```diff
@@ -0,0 +1,274 @@
...
+      (BuildInfo as Record<string, string>).displayName = 'This is my app';
+      const params = getParams(
+        await _generateHandlerUrl(auth, event, provider)
+      );
+      expect(params.get('appDisplayName')).to.eq('This is my app');
+    });
+  });
+
+  describe('_performRedirect', () => {
+    let isBrowsertabAvailable: boolean;
```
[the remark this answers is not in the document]
samhorlbeck: Only one of these has it set to true, the rest to false

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
samhorlbeck: I didn't realize Cordova supported IE11 :(

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
samhorlbeck: Ya I didn't realize Cordova supported IE11. I'm going to add this as a TODO but since I can't actually test this at the moment anyway I don't want to block on IE11 support here

packages-exp/auth-exp/src/core/util/handler.ts
```diff
@@ -0,0 +1,120 @@
...
+    provider.setDefaultLanguage(auth.languageCode);
+    params.providerId = provider.providerId || '';
+    if (!isEmpty(provider.getCustomParameters())) {
+      params.customParameters = JSON.stringify(provider.getCustomParameters());
+    }
+    const scopes = provider.getScopes().filter(scope => scope !== '');
+    if (scopes.length > 0) {
+      params.scopes = scopes.join(',');
+    }
+    // TODO set additionalParams?
```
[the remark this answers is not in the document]
samhorlbeck: Keeping this around to refer to the provider

packages-exp/auth-exp/src/platform_cordova/popup_redirect/utils.ts
samhorlbeck: Added a function to fallback to a simple encoding function

PR #4424, opened 2021-02-05
Add README.md about spec tests

dconeybe:


packages/firestore/test/unit/specs/README.md
```diff
@@ -0,0 +1,58 @@
...
+SDK when they change.
+
+## Running the Tests
+
+The spec tests are part of the unit test suite of each SDK. Therefore, to run
+the spec tests simply run the entire unit test suite and the spec tests will be
+executed as a part of it.
+
+For example, in the Web SDK (this SDK), the best way to run the spec tests is
+via the Intellij IDEA IDE. After following the instructions to set up IntelliJ
```
schmidt-sebastian, line 26: 99% of JS developers now use VS Code. The other 1% works on Firestore. I think IntelliJ is not that good of a Web IDE unless users pay for the professional edition. I would state that the tests can be run via VS Code and IntelliJ, but mostly we should point out how to use the Node test runner.

I would also give a shout out to the fact that we fake IndexedDb persistence under Node.

packages/firestore/test/unit/specs/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+To export the spec tests to the iOS SDK, run the following command:
+
+```
+cd ~/firebase-js-sdk/packages/firestore/test/unit &&
+./generate_spec_json.sh ~/firebase-ios-sdk/Firestore/Example/Tests/SpecTests/json
+```
+
+This command assumes that this Git repository is cloned into `~/firebase-js-sdk`
+and the iOS SDK is cloned into `~/firebase-ios-sdk`.
```
schmidt-sebastian, line 58: In general, this is a good starting point for a spec test introduction. I think the complications arise when someone wants to go past these setup steps, but I am honestly not quite sure how we could teach someone how to write a spec test over a README. The actual tests require an insane amount of technical knowledge, and I would be wary to put all this in writing.

Even if we want to focus on the high level nature of these tests, we may still want to expand on this and write about the tags. We should mention "exclusive", "no-ios" and "no-android" (and maybe "no-web", but I don't know when that would make sense).

packages/firestore/test/unit/specs/README.md
dconeybe: I ended up just removing this paragraph about IntelliJ since the instructions for running the tests are documented in each repository and don't need to be duplicated here. I also mentioned the IndexedDb persistence faking above.

packages/firestore/test/unit/specs/README.md
dconeybe: That's a good point about the tags. I've added information about the tags. I agree that trying to document the spec tests in their entirety is not feasible. I am mostly writing this so that the next time I need to work on the spec tests I have something to remind me of the basics, and this will hopefully be helpful to future authors as well.

schmidt-sebastian (approved):
Thanks!

PR #4439, opened 2021-02-09
Tree-shake Change

schmidt-sebastian:


hsubox76 (approved):
LG, just one question

packages/database/src/core/view/Change.ts
```diff
@@ -32,44 +47,33 @@ export class Change {
     public oldSnap?: Node,
     public prevName?: string | null
   ) {}
+}
```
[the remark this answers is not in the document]
schmidt-sebastian: We can make it an interface, which completely removes this type from the final bundle.

packages/database/src/core/view/Change.ts
```diff
@@ -17,59 +17,68 @@
...
 
-  static childChangedChange(
-    childKey: string,
-    newSnapshot: Node,
-    oldSnapshot: Node
-  ): Change {
-    return new Change(Change.CHILD_CHANGED, newSnapshot, childKey, oldSnapshot);
-  }
+/** Event type for a value change */
+export const CHANGE_TYPE_VALUE = 'value';
```
Feiyang1, line 33: Can these `const` variables be `const enum`? Can shave a little bit of size.

Feiyang1: Also will be a bit easier to use.

schmidt-sebastian: Done. Thanks for the suggestion

schmidt-sebastian: -500 bytes by the way

PR #4442, opened 2021-02-09
Big refactor of a bunch of redirect code that will be used by Cordova as well as Browser

samhorlbeck:
This change doesn't modify any behavior or add anything new


yuchenshi (approved):
rubber-stamped assuming it's just moving code around

PR #4444, opened 2021-02-09
Database tree-shake step: modularize CompoundWrite

hsubox76:
Extract methods from `CompoundWrite`.

### DIFF
--- a/packages/database/src/core/CompoundWrite.ts
+++ b/packages/database/src/core/CompoundWrite.ts
@@ -35,9 +35,6 @@ export class CompoundWrite {
   static empty(): CompoundWrite {
     return new CompoundWrite(new ImmutableTree(null));
   }
-
-  addWrite() {}
-  addWrites() {}
 }
 
 export function addWrite(
--- a/packages/database/src/core/CompoundWrite.ts
+++ b/packages/database/src/core/CompoundWrite.ts
@@ -30,145 +30,173 @@ import { each } from './util/util';
  * to reflect the write added.
  */
 export class CompoundWrite {
-  constructor(private writeTree_: ImmutableTree<Node>) {}
+  constructor(public writeTree_: ImmutableTree<Node>) {}
 
   static empty(): CompoundWrite {
     return new CompoundWrite(new ImmutableTree(null));
   }
 
-  addWrite(path: Path, node: Node): CompoundWrite {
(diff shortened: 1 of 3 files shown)


packages/database/src/core/CompoundWrite.ts
```diff
@@ -30,145 +30,170 @@ import { each } from './util/util';
...
 
   static empty(): CompoundWrite {
     return new CompoundWrite(new ImmutableTree(null));
   }
+}
 
-  addWrite(path: Path, node: Node): CompoundWrite {
-    if (path.isEmpty()) {
-      return new CompoundWrite(new ImmutableTree(node));
+export function addWrite(
```
schmidt-sebastian, line 40: Can you prefix all these method names with `compoundWrite`? This has some benefits:
- All imports sort in order
- It is still possible to translate Android and iOS APIs to Web APIs
- There are no clashes and we don't have to think about how to deal with clashes on a case-by-case basis.
- 
I initially did not like this method naming either, but we updated all of Firestore to match this pattern. This created a very consistent method-naming convention across the SDK. See https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/local/local_store_impl.ts#L216

packages/database/src/core/WriteTree.ts
```diff
@@ -215,34 +228,34 @@ export class WriteTree {
...
     if (!writeIdsToExclude && !includeHiddenWrites) {
-      const shadowingNode = this.visibleWrites_.getCompleteNode(treePath);
+      const shadowingNode = getCompleteNode(this.visibleWrites_, treePath);
       if (shadowingNode != null) {
         return shadowingNode;
       } else {
-        const subMerge = this.visibleWrites_.childCompoundWrite(treePath);
-        if (subMerge.isEmpty()) {
+        const subMerge = childCompoundWrite(this.visibleWrites_, treePath);
+        if (isEmpty(subMerge)) {
```
schmidt-sebastian, line 236: Especially this name is a bit overly generic.

packages/database/src/core/CompoundWrite.ts
hsubox76: Done. It's starting to look a little like Java now!

schmidt-sebastian (approved):
Thanks!

PR #4445, opened 2021-02-09
Add a few new event helpers to the Cordova redirect handler

samhorlbeck:


packages-exp/auth-exp/src/platform_cordova/popup_redirect/events.ts
```diff
@@ -41,6 +49,67 @@ export function _generateNewEvent(
...
+  url: string
+): AuthEvent | null {
+  // Parse the deep link within the dynamic link URL.
+  const callbackUrl = _getDeepLinkFromCallback(url);
+  // Confirm it is actually a callback URL.
+  // Currently the universal link will be of this format:
+  // https://<AUTH_DOMAIN>/__/auth/callback<OAUTH_RESPONSE>
+  // This is a fake URL but is not intended to take the user anywhere
+  // and just redirect to the app.
+  if (callbackUrl.includes('/__/auth/callback')) {
```
yuchenshi, line 77: This will have a false positive with `http://foo/#/__/auth/callback` (note the hash fragment). Considering using `new URL(callbackUrl).pathname` for proper URL parsing (with caveat of browser compatibility).

packages-exp/auth-exp/src/platform_cordova/popup_redirect/events.ts
```diff
@@ -41,6 +49,67 @@ export function _generateNewEvent(
...
+    // Check if there is an error in the URL.
+    // This mechanism is also used to pass errors back to the app:
+    // https://<AUTH_DOMAIN>/__/auth/callback?firebaseError=<STRINGIFIED_ERROR>
+    const params = searchParamsOrEmpty(callbackUrl);
+    // Get the error object corresponding to the stringified error if found.
+    const errorObject = params['firebaseError']
+      ? JSON.parse(decodeURIComponent(params['firebaseError']))
```
yuchenshi, line 84: nit: is it a concern that `JSON.parse` may throw?

packages-exp/auth-exp/src/platform_cordova/popup_redirect/events.ts
```diff
@@ -51,3 +120,38 @@ function generateSessionId(): string {
   }
   return chars.join('');
 }
+
+function storage(): Persistence {
+  return _getInstance(browserLocalPersistence);
+}
+
+function key(auth: Auth): string {
```
yuchenshi, line 128: nit: rename to `persistenceKey`

yuchenshi (approved):

packages-exp/auth-exp/src/platform_cordova/popup_redirect/events.ts
samhorlbeck: This is copied more or less directly from the old SDK so I'm inclined to keep it: https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/src/cordovahandler.js#L707

This URL comes back from the handler so we can be confident there won't be this edge case in any case

yuchenshi: sg, let's keep it that way

packages-exp/auth-exp/src/platform_cordova/popup_redirect/events.ts
samhorlbeck: Good point, done. I've added a helper that tries to parse or returns null

yuchenshi (approved):

PR #4446, opened 2021-02-09
[Deprecated] Separate window/sw API checks

zwu52:
Also renames the checkers to make it more readable (since we no longer reference window and sw flows as `controllers`)


Feiyang1 (approved):

Feiyang1:
Seems messaging-exp is failing build.

Issue #683, opened 2018-04-13
Date handling in functions serializer


abrgr:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Firebase SDK version: 4.12.1
  * Firebase Product: functions

### [REQUIRED] Describe the problem
Javascript dates passed as arguments to firebase functions are converted to empty objects.  The code in https://github.com/firebase/firebase-js-sdk/blob/master/packages/functions/src/serializer.ts does not account for dates as distinct from objects.  I believe the more idiomatic method would either be to serialize dates as iso8601 and expect that functions convert to a javascript date themselves or to make use of the '@type' properties used in decode (though, obviously, this would require some backend changes).  Also, despite the comment to the contrary, the decode method does not appear to handle dates.

#### Steps to reproduce:
Invoke a function via the callable interface.
```
const func = functions.httpsCallable('someFunc');
func({
  date: new Date() // this will appear as { date: {} } in someFunc
}).then(({ returnedDate }) => {
  // notice that returnedDate is a string
});

milesingrams:
Dealing with the same issue. Ive had to implement my own serialize/deserialize shim for dates for all my cloud functions. Interesting that the serializer.ts file specifically says its built to handle dates but has no functionality for dates at all. Either the documentation should change to reflect the current state or the function should include date handling. Furthermore the cloud functions should also handle serializing and deserializing firebase classes such as Timestamps and GeoPoints as it probably is not uncommon to be sending those documents in and out of cloud functions.

mohshraim:
any chance for this??

bklimt:
Hi. We hear you. We are still looking into the best way to support this. Thank you for your feedback and your patience.

milesingrams:
Glad you’re looking into it! I know serialization/deserialization comes with a lot of caveats. 
Perhaps you can use a reserved property that specifies the instance type of a complex object. For example:
{
  _instanceType: “Date”,
  milliseconds: 1234567890
}
{
  _instanceType: “firebase.firestore.Timestamp”,
  seconds: 1234567890,
  nanoseconds: 1234567890
}
Of course this would break if someone already has a property with the key _instanceType. Another option is to send a separate object alongside the data object that has a map of paths with the types to deserialize into. The benefit being that there are no reserved properties.

And finally you can just leave it up to the user and provide a way to extend the serialize/deserialize functions arbitrarily which would allow the user to easily use any class type they please. 

[...]

PR #4450, opened 2021-02-10
(de)serialize dates w/ ISO8601 (#683)

nVitius:
There's been an issue (#683) open for a few years asking for serialization of dates for httpsCallables.
It seemed like something really easy to do, so I created this PR as a proposal for how we might consider doing it.

Encoding is done with `Date.toISOString`. JavaScript's `JSON.stringify` also uses `Date.toISOString` to encode Dates ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON), [ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description)). Decoding tests Strings against a RegEx that matches ISO 8601 strings.

It would be nice to support a more robust serialization that also supports Firestore data types (Timestamp, GeoPoint, Reference). Perhaps we could eventually use the same serialization format that the [Firestore REST API](https://firebase.google.com/docs/firestore/reference/rest/v1/Value) uses. I do think that would require more coordination efforts across both libs to deploy without breaking any compatibility.

Ideally, this change would be accompanied by a similar change on the `firebase-functions` repo.
However, it looks like the actual code for encoding/decoding `onCall` functions is actually private.
https://github.com/firebase/firebase-functions/blob/master/src/providers/https.ts#L377
https://github.com/firebase/firebase-functions/blob/master/src/providers/https.ts#L339
At least, that's what I understood from the comments on those functions.

For what it's worth, the existing implementation doesn't work with Date types at all. It converts them to `{ date: {} }`. I don't believe that merging this will break anything even if the `firebase-functions` isn't updated at the same time.

### DIFF
--- a/packages/functions/src/serializer.ts
+++ b/packages/functions/src/serializer.ts
@@ -51,6 +51,9 @@ export class Serializer {
     if (data === true || data === false) {
       return data;
     }
+    if (data instanceof Date) {
+      return data.toISOString();
+    }
     if (Object.prototype.toString.call(data) === '[object String]') {
       return data;
     }
@@ -94,6 +97,14 @@ export class Serializer {
     }
     if (typeof json === 'function' || typeof json === 'object') {
       return mapValues(json!, x => this.decode(x));
+    }
+    if (
+      typeof json === 'string' &&
+      /^(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(.[0-9]+)?(Z)?$/.test(
+        json
+      )
+    ) {
+      return new Date(json);
     }
     // Anything else is safe to return.
     return json;
--- a/packages/functions/test/serializer.test.ts
+++ b/packages/functions/test/serializer.test.ts
@@ -80,6 +80,18 @@ describe('Serializer', () => {
     expect(serializer.decode(1.2)).to.equal(1.2);
   });
 
+  it('encodes date', () => {
+    expect(serializer.encode(new Date('2021-02-10T13:26:37.977Z'))).to.equal(
+      '2021-02-10T13:26:37.977Z'
+    );
+  });
+
+  it('decodes date', () => {
+    expect(serializer.decode('2021-02-10T13:26:37.977Z')).to.deep.equal(
+      new Date('2021-02-10T13:26:37.977Z')
+    );
+  });
+
   it('encodes string', () => {
     expect(serializer.encode('hello')).to.equal('hello');
   });


nVitius:
@googlebot I signed it!

nVitius:
@Feiyang1 @hiranya911 @hsubox76 
Can get some feedback on this PR/Issue?

nVitius:
@Feiyang1 @hiranya911 @hsubox76 
Just pinging you guys again. Would appreciate some feedback here.

The underlying issue has been open for almost three years now. It seems to be an easy fix (the change in this PR is non-breaking btw). If there's something blocking this, it would be great to know about it.

hsubox76:
Thanks, I will discuss this with other teams (functions backend, other platforms such as iOS and Android) to try to get background and make sure we're not missing any context and it won't have any unintended effects. We also probably want changes to be made in sync with the other platforms (iOS, Android).

nVitius:
@hsubox76 Following up on this:
Were you able to bring this issue up with the other Firebase teams? Can we get some feedback on the status of this issue?

I mentioned it before, but the current implementation doesn't work with dates at all. Presumably, anyone using this to send dates is using their own workaround. The change proposed in this PR is non-breaking.

hsubox76:
Yes, I did bring it up with the other teams and the larger issue was too complex to reach a conclusion on, but we decided to fix the immediate bug we want to `JSON.stringify()` the entire data object on the way to the server, so that all data is at least serialized in a predictable, consistent way, instead of the surprise of having a Date be an empty object.

We don't want to automatically deserialize ISO strings returned from the server into Date objects, in case someone intentionally is sending an ISO formatting string back and wants to keep it as a string.

I was planning to make these changes this week but have not had time yet due to a number of pressing beta releases. If you have time to change this PR to reflect the plan above, we can merge it.

A full solution for functions handling special data types on the server side and passing type data back to the deserializer, etc. will probably be a long term feature I'm not sure I can promise any timeline or prioritization on yet.

nVitius:
Thanks for your reply.

I can make the changes to the PR to remove the decoding, but I do think that it would be valuable to have.

I understand the reasoning for not wanting to automatically deserialize ISO strings. I'm obviously not familiar with the implications this would have to the larger issue, but I think it would make sense to encode them to an object like this:

```
{ _firebaseSerializedDate: `the iso string` }
```

This way, we can almost guarantee that only `Date` objects would be serialized/deserialized.

hsubox76:
That would definitely be a long-term solution as the server-side function would have to send some kind of standardized wrapper back to indicate the type, and the other platform SDKs (Android, iOS) would have to recognize this wrapper. There was an effort to do this in the past but it was never completed. Would like to see what we can do now but can't make any promises.

We can and should quickly fix the serialization error where a Date ends up as an empty object by JSON.stringifying the entire data object, since that can be entirely handled in the JS SDK. I hope to get to it soon if you don't have time. By the way, if you want to go through with it, this PR will also need a changeset (`yarn changeset` to add).

nVitius:
Not at all. I'll close this one then.

PR #4460, opened 2021-02-11
Implement analytics-compat package

hsubox76:
Create analytics-compat package wrapping analytics-exp. Did not add to `firebase` package - doesn't seem like most compat packages have been added to it yet.

### DIFF
--- a/packages-exp/analytics-compat/src/constants.ts
+++ b/packages-exp/analytics-compat/src/constants.ts
@@ -1,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Officially recommended event names for gtag.js
+ * Any other string is also allowed.
+ *
+ * @public
+ */
+export enum EventName {
+  ADD_SHIPPING_INFO = 'add_shipping_info',
+  ADD_PAYMENT_INFO = 'add_payment_info',
+  ADD_TO_CART = 'add_to_cart',
+  ADD_TO_WISHLIST = 'add_to_wishlist',
+  BEGIN_CHECKOUT = 'begin_checkout',
+  /**
+   * @deprecated
+   * This event name is deprecated and is unsupported in updated
+   * Enhanced Ecommerce reports.
+   */
+  CHECKOUT_PROGRESS = 'checkout_progress',
+  EXCEPTION = 'exception',
+  GENERATE_LEAD = 'generate_lead',
+  LOGIN = 'login',
+  PAGE_VIEW = 'page_view',
+  PURCHASE = 'purchase',
+  REFUND = 'refund',
+  REMOVE_FROM_CART = 'remove_from_cart',
+  SCREEN_VIEW = 'screen_view',
+  SEARCH = 'search',
+  SELECT_CONTENT = 'select_content',
+  SELECT_ITEM = 'select_item',
+  SELECT_PROMOTION = 'select_promotion',
+  /** @deprecated */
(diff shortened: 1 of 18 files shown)


packages-exp/analytics-compat/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { firebase } from '@firebase/app-compat';
+import { _FirebaseNamespace } from '@firebase/app-types/private';
```
Feiyang1, line 19: FYI, we may have merge conflict with https://github.com/firebase/firebase-js-sdk/pull/4461 depending on which one got merged first.

packages-exp/analytics-compat/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    .getImmediate({
+      identifier: regionOrCustomDomain
+    });
+
+  return new AnalyticsService(app as FirebaseApp, analyticsServiceExp);
+};
+
+export function registerAnalytics(): void {
+  const namespaceExports = {
+    // no-inline
```
Feiyang1, line 61: what does the comment mean?

packages/analytics-types/index.d.ts
```diff
@@ -40,13 +40,421 @@ export interface FirebaseAnalytics {
...
    * app instance on this device.
-   * List of official event parameters can be found in
+   * List of recommended event parameters can be found in
    * {@link https://developers.google.com/gtagjs/reference/event
    * the gtag.js reference documentation}.
    */
   logEvent(
-    eventName: EventNameString,
-    eventParams?: EventParams,
+    eventName: 'add_payment_info',
```
Feiyang1, line 48: Did you miss the overloads before?

packages/analytics/src/factory.ts
```diff
@@ -226,7 +229,11 @@ export function factory(
     app,
     // Public methods return void for API simplicity and to better match gtag,
     // while internal implementations return promises.
-    logEvent: (eventName, eventParams, options) => {
+    logEvent: (
+      eventName: string,
```
Feiyang1, line 233: Is it necessary to type them explicitly? Are they not inferred by `FirebaseAnalyticsInternal ` ?

packages/firebase/index.d.ts
```diff
@@ -4479,7 +4479,7 @@ declare namespace firebase.analytics {
      */
     logEvent(
       eventName: 'add_shipping_info',
-      eventParams: {
+      eventParams?: {
```
Feiyang1, line 4482: These changes are unrelated to `-compat` and should have a changeset. We should create a separate PR for them.

Feiyang1 (changes requested):

packages/analytics-types/index.d.ts
hsubox76: It didn't seem necessary to put them in analytics-types because it's only for user convenience, so it was only in firebase/index.d.ts. But analytics-compat won't have access to firebase/index.d.ts.

packages/analytics/src/factory.ts
hsubox76: It used to be fine but after adding the logEvent overloads to analytics-types now it complains if the params aren't typed?

packages/firebase/index.d.ts
hsubox76: Done: https://github.com/firebase/firebase-js-sdk/pull/4469

packages-exp/analytics-compat/src/index.ts
hsubox76: It's left over from copying from another file. I see it in functions and storage and I think it might have been an annotation for Prettier to keep the following line on its own line even when there is only one item. But it doesn't seem to do anything now even in files where there's only one item. Removed.

packages-exp/analytics-compat/src/constants.ts
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Officially recommended event names for gtag.js
+ * Any other string is also allowed.
+ *
+ * @public
```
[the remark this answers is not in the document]
hsubox76: Removed.

Feiyang1 (approved):

PR #4461, opened 2021-02-11
Remove @firebase/app-types-exp

Feiyang1:


common/api-review/app-exp.api.md
```diff
@@ -35,11 +33,57 @@ export const _DEFAULT_ENTRY_NAME = "[DEFAULT]";
...
+
+// @public (undocumented)
+export interface FirebaseAppConfig {
+    // (undocumented)
+    automaticDataCollectionEnabled?: boolean;
+    // (undocumented)
+    name?: string;
+}
+
+// @public (undocumented)
```
hsubox76, line 51: I guess this has to be public so app-compat is able to import it from app-exp? Will it end up in the generated documentation?

hsubox76 (commented):
I think the changeset warning means you have to remove it from https://github.com/firebase/firebase-js-sdk/blob/c0a1f6bf515413fb9f2ac082fa42f5ef2018c40d/.changeset/config.json#L17

common/api-review/app-exp.api.md
Feiyang1: Yes, it should have a `@internal` annotation so it doesn't show up in the documentation.
Fixed.

common/api-review/app-exp.api.md
```diff
@@ -35,11 +33,57 @@ export const _DEFAULT_ENTRY_NAME = "[DEFAULT]";
...
-
-export { FirebaseAppConfig }
+// @public
+export interface FirebaseApp {
+    automaticDataCollectionEnabled: boolean;
+    readonly name: string;
+    readonly options: FirebaseOptions;
+}
+
+// @public (undocumented)
```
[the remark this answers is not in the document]
Feiyang1: Yup. Added comments for all except for the members of `FirebaseOptions` since they are supposed to be opaque to developers and are kinda self explanator

packages/database/src/core/RepoManager.ts
```diff
@@ -16,7 +16,7 @@
  */
 
 import { FirebaseApp } from '@firebase/app-types';
-import { FirebaseApp as FirebaseAppExp } from '@firebase/app-types-exp';
+import { FirebaseApp as FirebaseAppExp } from '@firebase/app-exp';
```
hsubox76, line 19: I guess this should be added to database's deps (according to the failing check)

hsubox76 (approved):
LG, just need to fix database deps issue to address CI failure.

PR #4463, opened 2021-02-11
throw if initializeFirestore is called more than once

Feiyang1:
Throw if `initializeFirestore()` is called after a Firestore instance has already been created for the app.

### DIFF
--- a/packages/firestore/src/lite/database.ts
+++ b/packages/firestore/src/lite/database.ts
@@ -181,11 +181,11 @@ export function initializeFirestore(
   app: FirebaseApp,
   settings: Settings
 ): FirebaseFirestore {
-  const provider = _getProvider(app, 'firestore-exp');
+  const provider = _getProvider(app, 'firestore/lite');
 
   if (provider.isInitialized()) {
     throw new FirestoreError(
-      Code.ALREADY_INITIALIZED,
+      Code.FAILED_PRECONDITION,
       'Firestore can only be initialized once per app.'
     );
   }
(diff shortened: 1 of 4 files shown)


packages/firestore/src/lite/database.ts
```diff
@@ -181,10 +181,16 @@ export function initializeFirestore(
   app: FirebaseApp,
   settings: Settings
 ): FirebaseFirestore {
-  const firestore = _getProvider(
-    app,
-    'firestore/lite'
-  ).getImmediate() as FirebaseFirestore;
+  const provider = _getProvider(app, 'firestore-exp');
```
schmidt-sebastian, line 184: This should be firestore/lite

packages/firestore/src/util/error.ts
```diff
@@ -116,6 +116,12 @@ export const Code = {
    */
   ALREADY_EXISTS: 'already-exists' as FirestoreErrorCode,
 
+  /**
+   * Calling initializeFirestore() after a Firestore instance has already been created for the app
+   * It can happen when initializeFirestore() is called more than once or after calling getFirestore().
+   */
+  ALREADY_INITIALIZED: 'already-initialized' as FirestoreErrorCode,
```
schmidt-sebastian, line 123: These are GRPC error codes. They cannot be modified. You should use `FAILED_PRECONDITION`.

packages/firestore/src/lite/database.ts
Feiyang1: done

packages/firestore/src/util/error.ts
Feiyang1: done

schmidt-sebastian (approved):

PR #4466, opened 2021-02-12
Add generics to httpsCallable

hsubox76:
Add `RequestData` and `ResponseData` generics to `httpsCallable`. `RequestData` is the arg provided to the callable function (only one allowed). The callable function returns a promise with a `data` field, and `RequestData` describes whatever is in that data field.

Used the generics in a couple of `httpsCallable` tests as a check to make sure the typing works, and left them out of some tests to make sure it also works if they are omitted.

### DIFF
--- a/packages-exp/functions-exp/src/callable.test.ts
+++ b/packages-exp/functions-exp/src/callable.test.ts
@@ -86,7 +86,10 @@ describe('Firebase Functions > Call', () => {
       null: null
     };
 
-    const func = httpsCallable(functions, 'dataTest');
+    const func = httpsCallable<
+      Record<string, any>,
+      { message: string; code: number; long: number }
+    >(functions, 'dataTest');
     const result = await func(data);
 
     expect(result.data).to.deep.equal({
@@ -98,7 +101,7 @@ describe('Firebase Functions > Call', () => {
 
   it('scalars', async () => {
     const functions = createTestService(app, region);
-    const func = httpsCallable(functions, 'scalarTest');
+    const func = httpsCallable<number, number>(functions, 'scalarTest');
     const result = await func(17);
     expect(result.data).to.equal(76);
   });
--- a/common/api-review/functions-exp.api.md
+++ b/common/api-review/functions-exp.api.md
@@ -33,7 +33,7 @@ export interface HttpsCallable {
 }
 
 // @public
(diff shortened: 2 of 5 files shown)


packages-exp/functions-types-exp/index.d.ts
```diff
@@ -20,16 +20,16 @@ import { FirebaseError } from '@firebase/util';
...
+  readonly data: ResponseData;
 }
 
 /**
  * An HttpsCallable is a reference to a "callable" http trigger in
  * Google Cloud Functions.
  */
-export interface HttpsCallable {
-  (data?: {} | null): Promise<HttpsCallableResult>;
+export interface HttpsCallable<RequestParams = any, ResponseData = any> {
```
Feiyang1, line 31: Should the generics default to `unknown` instead?

hsubox76: Before when I tried it I had some problems with the types not getting inferred in usage like if I did
```
const callTest = httpsCallable<string, string>(functions, "callTest");
const result = await callTest();
```
both `callTest` and `result` would be `any`. But maybe I just neglected to cast the return value in `httpsCallable` and I did that now and it seems to work.

Feiyang1 (approved):

packages-exp/functions-exp/src/callable.test.ts
```diff
@@ -98,7 +101,7 @@ describe('Firebase Functions > Call', () => {
 
   it('scalars', async () => {
     const functions = createTestService(app, region);
-    const func = httpsCallable(functions, 'scalarTest');
+    const func = httpsCallable<number, number>(functions, 'scalarTest');
```
mbleigh, line 104: I believe at least the request (and maybe the response as well) *must* be objects. Scalars may not be allowed by the server-side implementation.

hsubox76: This is a live test that sends a number to the server and successfully gets a number response so it seems to work, unless I'm misunderstanding. To make sure, I also created a new `onCall` function that logged a number sent to it and returned a number and tried it from a test app and it seems to work:
```
exports.numberTest = functions.https.onCall((data) => {
    console.log(data, typeof data);
    return(44);
});
```
Let me know if I misunderstood.

PR #4469, opened 2021-02-12
Fix analytics logEvent public signatures

hsubox76:
`eventParams` should be an optional argument.

### DIFF
--- a/.changeset/little-suns-train.md
+++ b/.changeset/little-suns-train.md
@@ -1,5 +1,4 @@
 ---
-'@firebase/analytics': patch
 'firebase': patch
 ---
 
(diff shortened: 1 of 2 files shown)


Feiyang1 (changes requested):
It should have a changeset.

hsubox76:
I added a changeset. By default it only wanted to add the `firebase` package because only `packages/firebase/index.d.ts` changed but I thought `@firebase/analytics` should be added so it goes under "analytics" in the changelog.

Feiyang1:
I don't think we should bump `@firebase/analytics` since it has no changes. Is the typing correct in `@firebase/analytics-types`?

Feiyang1 (approved):

PR #4470, opened 2021-02-12
Tree-Shake OperationSource

schmidt-sebastian:


packages/database/src/core/operation/Operation.ts
```diff
@@ -42,40 +41,38 @@ export interface Operation {
...
 
-  static user() {
-    return new OperationSource(
-      /*fromUser=*/ true,
-      false,
-      null,
-      /*tagged=*/ false
-    );
-  }
+export function newOperationSourceUser() {
```
Feiyang1, line 51:
```suggestion
export function newOperationSourceUser(): OperationSource {
```

packages/database/src/core/operation/Operation.ts
```diff
@@ -42,40 +41,38 @@ export interface Operation {
...
 
-  static server() {
-    return new OperationSource(
-      false,
-      /*fromServer=*/ true,
-      null,
-      /*tagged=*/ false
-    );
-  }
+export function newOperationSourceServer() {
```
Feiyang1, line 60:
```suggestion
export function newOperationSourceServer(): OperationSource {
```

Feiyang1 (changes requested):
Perhaps we can start turning on eslint rules that we turned off for database as part of the rewrite?

schmidt-sebastian:
> Perhaps we can start turning on eslint rules that we turned off for database as part of the rewrite?

Is it possible to do this only for modified files?

packages/database/src/core/operation/Operation.ts
```diff
@@ -42,40 +41,38 @@ export interface Operation {
...
 
-  static user() {
-    return new OperationSource(
-      /*fromUser=*/ true,
-      false,
-      null,
-      /*tagged=*/ false
-    );
-  }
+export function newOperationSourceUser(): OperationSource {
```
hsubox76, line 51: Why are these all prefixed with "new"?

schmidt-sebastian: They are factory functions. I can remove the prefix if you think we shouldn't do this for consistency - let me know,

hsubox76: That makes sense.

hsubox76 (approved):

PR #4477, opened 2021-02-16
Tree-Shake Path

schmidt-sebastian:


packages/database/src/core/util/validation.ts
```diff
@@ -215,11 +223,11 @@ export const validateFirebaseMergePaths = function (
   // Check that update keys are not descendants of each other.
   // We rely on the property that sorting guarantees that ancestors come
   // right before descendants.
-  mergePaths.sort(Path.comparePaths);
+  mergePaths.sort((l, r) => pathCompare(l, r));
```
hsubox76, line 226: does `pathCompare` not work by itself?

hsubox76 (approved):

packages/database/src/core/util/validation.ts
schmidt-sebastian: It should, but it might be a bit of an anti-pattern, as it might introduce some future breakages if the callsite ends up passing arguments that `pathCompare` does not support (e.g. a "position" argument that is passed `pathCompare`, which might expect an optional string description instead). 

I did update this location though since it matches the original implementation.

Issue #1874, opened 2019-06-12
Can I provide a custom persistence implementation for auth in node?


nbransby:
I need to persist the user across sessions in node - is there to do this?

bojeil-google:
That is not currently available. We will consider it in our [modularization](https://github.com/firebase/firebase-js-sdk/issues/332) efforts.

nbransby:
Can you suggest a workaround? Maybe use onIdTokenChanged to save the user and updateCurrentUser on startup to restore? Problem with that is deserializing the user with the correct prototype

JakeHedman:
Bit late to the party, but this workaround worked for me:

Save user:
```js
const userJson = JSON.stringify(currentUser.toJSON())
// Write userJson to disk
```

Load user:
```js
// Read userJson from disk
const userData = JSON.parse(userJson)
const user = firebase.User(userData, userData.stsTokenManager, userData)
firebase.auth().updateCurrentUser(user)
```

dfernandezm:
> Bit late to the party, but this workaround worked for me:
> 
> Save user:
> 
> ```js
> const userJson = JSON.stringify(currentUser.toJSON())
> // Write userJson to disk
> ```
> 
> Load user:
> 
> ```js
> // Read userJson from disk
> const userData = JSON.parse(userJson)
> const user = new firebase.User(userData, userData.stsTokenManager, userData)
> firebase.auth().updateCurrentUser(user)
> ```

@JakeHedman I would love to implement this, but at least in my Typescript setup I get the following:

```
const user = new firebase.User(userData, userData.stsTokenManager, userData);
```

```
Property 'User' does not exist on type 'typeof firebase'.ts(2339)
```

Seems like I can reference the type but not instantiate it. I am using:

```
"firebase": "^7.14.3",
"firebase-admin": "^8.12.1",
```

Any workarounds on how to instantiate `firebase.User` that way with Typescript?

billybraga:
> Bit late to the party, but this workaround worked for me:
> 
> Save user:
> 
> ```js
> const userJson = JSON.stringify(currentUser.toJSON())
> // Write userJson to disk
> ```
> 
> Load user:
> 
> ```js
> // Read userJson from disk
> const userData = JSON.parse(userJson)
> const user = new firebase.User(userData, userData.stsTokenManager, userData)
> firebase.auth().updateCurrentUser(user)
> ```

Could this be documented as a public API so we can rely on it not changing in the future?

[...]

PR #4484, opened 2021-02-17
Add a resaturation function for auth User instances

samhorlbeck:
This allows symmetry for the user object:

```ts
const user = (await signInAnonymously(auth)).user;
const userString = JSON.stringify(user);
const resaturated = /* new: */ userFromJSON(auth, userString);
```

This also allows more flexibility in environments whose storage we don't always fully support (see https://github.com/firebase/firebase-js-sdk/issues/1874, https://github.com/firebase/firebase-js-sdk/discussions/4468)

**Pending API review.**


yuchenshi (commented):
Code LGTM, will approve after API proposal is reviewed and finalized

Feiyang1 (commented):
LGTM. Lint will pass once you merge the branch with master.

alexcastillo:
This is exactly what I need. Any workarounds I can use until this is merged?

smashah:
Any movement on this PR? I would really like to keep my client-side node apps running in stateless containers authenticated with a first-party solution

jacksteamdev:
This would be extremely useful for MV3 Chrome Extensions, where the background service worker has a short lifespan.

packages-exp/auth-exp/src/core/user/user_impl.ts
```diff
@@ -317,3 +318,7 @@ export class UserImpl implements User {
     return user;
   }
 }
+
+export function userFromJSON(auth: externs.Auth, json: string): externs.User {
+  return UserImpl._fromJSON(_castAuth(auth), JSON.parse(json));
```
jacksteamdev, line 323: Would be useful if this could take a plain object as well (some storage solutions parse the JSON automatically):

```javascript
const userObj = typeof json === "string" ? JSON.parse(json) : json;
return UserImpl._fromJSON(_castAuth(auth), userObj);
```

bashimr:
Was this function 'userFromJSON' ever released? I can't seem to find it in any documentation. As mentioned above, Chrome Extension firebase google authentication is very hard to achieve in the main extension context. The only viable option is to do the authentication offscreen and then use 'updateCurrentUser' to populate the Auth of the main context. But that requires creating a new user from the stingified creds passed to the main context via messaging.

PR #4485, opened 2021-02-17
Refactor QueryParams

Feiyang1:
I kept simple getters and `copy()` within the class. Other methods are changed to functions.

### DIFF
--- a/packages/database/src/core/view/QueryParams.ts
+++ b/packages/database/src/core/view/QueryParams.ts
@@ -203,7 +203,7 @@ export class QueryParams {
   }
 }
 
-function getNodeFilter(queryParams: QueryParams): NodeFilter {
+export function getNodeFilter(queryParams: QueryParams): NodeFilter {
   if (queryParams.loadsAllData()) {
     return new IndexedFilter(queryParams.getIndex());
   } else if (queryParams.hasLimit()) {
@@ -333,7 +333,7 @@ export function endBefore(
   return params;
 }
 
-function orderBy(queryParams: QueryParams, index: Index): QueryParams {
+export function orderBy(queryParams: QueryParams, index: Index): QueryParams {
   const newParams = queryParams.copy();
   newParams.index_ = index;
   return newParams;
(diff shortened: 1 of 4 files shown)


packages/database/src/core/view/QueryParams.ts
```diff
@@ -178,7 +174,18 @@ export class QueryParams {
     return this.index_;
   }
 
-  private copy_(): QueryParams {
+  loadsAllData(): boolean {
+    return !(this.startSet_ || this.endSet_ || this.limitSet_);
+  }
+
+  /**
+   * What does isDefault() mean?
```
hsubox76, line 182: Were you planning to get an answer before checking in the code?

Feiyang1: Removed the comment. But what does it mean @schmidt-sebastian ?

packages/database/src/core/view/QueryParams.ts
```diff
@@ -194,211 +198,237 @@ export class QueryParams {
...
-    }
+  return qs;
+}
 
-    if (this.endSet_) {
-      qs[REST_CONSTANTS.END_AT] = stringify(this.indexEndValue_);
-      if (this.endNameSet_) {
-        qs[REST_CONSTANTS.END_AT] += ',' + stringify(this.indexEndName_);
-      }
+export function queryParamsGetQueryObject(
```
schmidt-sebastian, line 399: FYI It looks like you re-ordered some of the methods, which makes this a internals impossible to review. I trust that you updated the methods correctly.

packages/database/src/core/view/QueryParams.ts
schmidt-sebastian: It's a query with no filters and no constraints with the default ordering (by priority)

schmidt-sebastian (approved):

packages/database/src/core/view/QueryParams.ts
Feiyang1: Tests are passing!

PR #4494, opened 2021-02-19
Run integration tests on Node.js.

yuchenshi:
If CI passes then all should be good.


packages-exp/auth-exp/test/helpers/integration/helpers.ts
```diff
@@ -18,7 +18,7 @@
 import { deleteApp, initializeApp } from '@firebase/app-exp';
 import { Auth, User } from '@firebase/auth-types-exp';
 
-import { getAuth } from '../../../index';
+import { getAuth } from '../../../'; // Use browser OR node dist entrypoint depending on test env.
```
samhorlbeck, line 21: nit: two spaces between code and comment

samhorlbeck (approved):

packages-exp/auth-exp/test/helpers/integration/helpers.ts
yuchenshi:
```suggestion
import { getAuth } from '../../../';  // Use browser OR node dist entrypoint depending on test env.
```

PR #4496, opened 2021-02-19
Remove remote-config-types-exp

Feiyang1:


packages-exp/remote-config-compat/src/remoteConfig.test.ts
```diff
@@ -21,7 +21,6 @@ import { stub } from 'sinon';
 import { RemoteConfigCompatImpl } from './remoteConfig';
 import { getFakeApp, getFakeModularRemoteConfig } from '../test/util';
 import * as modularApi from '@firebase/remote-config-exp';
```
hsubox76, line 23: I know this isn't in the scope of the PR but this isn't a potential tree-shaking problem is it?

hsubox76 (approved):

packages-exp/remote-config-compat/src/remoteConfig.test.ts
Feiyang1: It's a test, so no tree-shaking concerns. It's imported this way to allow stubbing.

PR #4497, opened 2021-02-19
Remove installations-types-exp

Feiyang1:


ChaoqunCHEN (approved):

packages-exp/installations-compat/package.json
```diff
@@ -30,7 +30,9 @@
     "api-report": "api-extractor run --local --verbose",
     "predoc": "node ../../scripts/exp/remove-exp.js temp",
     "doc": "api-documenter markdown --input temp --output docs",
-    "build:doc": "yarn build && yarn doc"
+    "build:doc": "yarn build && yarn doc",
+    "typings:public": "node ../../scripts/exp/use_typings.js ./dist/app-exp-public.d.ts",
```
hsubox76, line 34: Is it supposed to be `installations-exp-public.d.ts` and on the line below?

Feiyang1: Removed. These scripts are only used for exp packages, it was an mistake to have them in the compat packages.

hsubox76 (approved):

hsubox76 (approved):

PR #4507, opened 2021-02-22
Remove storage-types/exp

hsubox76:
Remove exp types for storage as part of the effort to remove all -types packages from the exp SDK.

### DIFF
--- a/packages/storage/exp/index.ts
+++ b/packages/storage/exp/index.ts
@@ -52,7 +52,7 @@ import {
   SettableMetadata,
   UploadMetadata,
   FullMetadata
-} from '@firebase/storage-types/exp';
+} from './public-types';
 import { Metadata as MetadataInternal } from '../src/metadata';
 import {
   uploadBytes as uploadBytesInternal,
--- a/packages/storage/exp/index.ts
+++ b/packages/storage/exp/index.ts
@@ -51,7 +51,8 @@ import {
   StorageObserver,
   SettableMetadata,
   UploadMetadata,
-  FullMetadata
+  FullMetadata,
+  UploadTaskSnapshot
 } from './public-types';
 import { Metadata as MetadataInternal } from '../src/metadata';
 import {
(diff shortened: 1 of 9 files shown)


packages/storage/exp/index.ts
```diff
@@ -51,8 +51,9 @@ import {
   StorageObserver,
   SettableMetadata,
   UploadMetadata,
-  FullMetadata
-} from '@firebase/storage-types/exp';
+  FullMetadata,
+  UploadTaskSnapshot
+} from './public-types';
```
Feiyang1, line 56: We can directly export them instead of import, then export since `./public-types` is part of the project, so we won't have the typings in the js file issues.

Feiyang1: Also can we just export every thing from `./public-types`. All types in `./public-types` are supposed to be public/exported.

packages/storage/package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "eslint -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "lint:fix": "eslint --fix -c .eslintrc.js '**/*.ts' --ignore-path '../../.gitignore'",
     "build": "rollup -c rollup.config.compat.js && yarn build:exp",
-    "build:exp": "rollup -c rollup.config.exp.js ; yarn api-report",
+    "build:exp": "rollup -c rollup.config.exp.js && yarn api-report",
```
Feiyang1, line 18: I think we want `;` which executes regardless of the return status of the previous command. I believe `rollup -c rollup.config.exp.js` will return an error status in the exp release build, though build will go through.

Feiyang1 (changes requested):

packages/storage/package.json
hsubox76: Oh, ok, restored.

packages/storage/exp/index.ts
hsubox76: Done. A number of these types are used in this file so I kept those imports.

Feiyang1 (approved):

PR #4509, opened 2021-02-23
Add protobuf license needed for internal use.

hsubox76:
See cl/358448976 (internal).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -207,7 +207,7 @@ which is licensed under BSD-3-Clause, both licenses given below.
 
 For Google protobuf source.
 
-file: src/packages/firestore/src/protos/google/protobuf/any.proto
+file: [source code root]/packages/firestore/src/protos/google/protobuf/any.proto
 
 // Protocol Buffers - Google's data interchange format
 // Copyright 2008 Google Inc.  All rights reserved.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -200,3 +202,39 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+------------------------------------------------------------------------------
+
+For Google protobuf source.
+
+file: src/packages/firestore/src/protos/google/protobuf/any.proto
```
Feiyang1, line 210: What should the path here?

hsubox76: It's a relative path in google3, it's copied verbatim from the CL.

Feiyang1: But this license is for the public version of the SDK, should it use the path in Github?

hsubox76: Unless we want to write a special script to reformat it on google3 import, it's probably best to make it applicable to wherever it is. I tried 
```
[source code root]/packages/firestore/src/protos/google/protobuf/any.proto
```

Feiyang1 (approved):
Do we need to update a script to change the path for google3?

hsubox76:
Discussed with CL author, the path does not need to be strict, it is just for providing context to any human readers of the license.

PR #4512, opened 2021-02-23
Fix the OAuth credential logic, and add support for OIDC

samhorlbeck:
The OIDC stuff was pretty much there already, but the type for several things internally needed to be changed from strictly `ProviderId` to `ProviderId | string` as OIDC provider IDs are arbitrary strings


yuchenshi (approved):
Neat!

I remember doing a similar `ProviderId | string` change for Auth Emulator as well. 😸

PR #4515, opened 2021-02-23
Tree-shake RepoInfo

hsubox76:
It seems like most of the methods should stay on the class, I only extracted `connectionURL` and its helper method `needsQueryParam` but let me know if that's incorrect.

### DIFF
--- a/packages/database/src/core/RepoInfo.ts
+++ b/packages/database/src/core/RepoInfo.ts
@@ -92,7 +92,7 @@ export class RepoInfo {
   }
 }
 
-function needsQueryParam(repoInfo: RepoInfo): boolean {
+function repoInfoNeedsQueryParam(repoInfo: RepoInfo): boolean {
   return (
     repoInfo.host !== repoInfo.internalHost ||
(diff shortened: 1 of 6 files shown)


packages/database/src/core/RepoInfo.ts
```diff
@@ -133,3 +91,51 @@ export class RepoInfo {
     return `${protocol}${this.host}/${query}`;
   }
 }
+
+function needsQueryParam(repoInfo: RepoInfo): boolean {
```
schmidt-sebastian, line 95: Nit: I would also prefix this with "repoInfo" for consistency.

schmidt-sebastian (approved):
I think it is pretty hard to determine which methods actually make a difference in terms of bundle size until we have the API implemented. I think your change makes sense. If there is anything that stands out once we have a tree-shakeable API we can always change it.

packages/database/src/core/RepoInfo.ts
hsubox76: Done.

Feiyang1 (approved):

PR #4520, opened 2021-02-24
Remove RepoManager

schmidt-sebastian:
Removes the RepManager, which was only used as a singleton. It is replaced with two const variables. All functions are moved to be tree-shakeable (interrupt() and resume() were removed since they are not used).

### DIFF
--- a/packages/database/src/core/RepoManager.ts
+++ b/packages/database/src/core/RepoManager.ts
@@ -19,7 +19,7 @@ import { FirebaseApp } from '@firebase/app-types';
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { FirebaseApp as FirebaseAppExp } from '@firebase/app-exp';
 import { safeGet } from '@firebase/util';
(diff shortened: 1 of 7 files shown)


packages/database/src/core/RepoManager.ts
```diff
@@ -55,171 +53,149 @@ export type FirebaseAppLike = FirebaseApp | FirebaseAppExp;
...
-  resume() {
-    for (const appName of Object.keys(this.repos_)) {
-      for (const dbUrl of Object.keys(this.repos_[appName])) {
-        repoResume(this.repos_[appName][dbUrl]);
-      }
-    }
+/**
+ * Update an existing repo in place to point to a new host/port.
+ */
+export function repoManagerApplyEmulatorSettings(
```
Feiyang1, line 70: Since repoManager is removed, do we still want the prefix?

schmidt-sebastian: Yeah, I want to keep it to be able to translate these API calls between iOS, Android and Web.

Feiyang1 (approved):

PR #4543, opened 2021-02-26
Modify URL parsing for React Native compat.

yuchenshi:
I've tried to keep the regex short and to the point, but let's see.

Once this is merged, there should be no `new URL` or `new URLSearchParams` floating in `auth-exp`.

### DIFF
--- a/packages-exp/auth-exp/src/core/auth/emulator.ts
+++ b/packages-exp/auth-exp/src/core/auth/emulator.ts
@@ -58,20 +58,58 @@ export function useAuthEmulator(
     AuthErrorCode.INVALID_EMULATOR_SCHEME
   );
 
-  const parsedUrl = new URL(url);
   const disableWarnings = !!options?.disableWarnings;
 
-  // Store the normalized URL whose path is always nonempty (i.e. containing at least a single '/').
-  authInternal.config.emulator = { url: parsedUrl.toString() };
+  const protocol = extractProtocol(url);
+  const { host, port } = extractHostAndPort(url);
+  const portStr = port === null ? '' : `:${port}`;
+
+  // Always replace path with "/" (even if input url had no path at all, or had a different one).
+  authInternal.config.emulator = { url: `${protocol}//${host}${portStr}/` };
   authInternal.settings.appVerificationDisabledForTesting = true;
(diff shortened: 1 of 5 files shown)


packages-exp/auth-exp/src/core/auth/emulator.ts
```diff
@@ -58,22 +58,60 @@ export function useAuthEmulator(
...
 
   emitEmulatorWarning(disableWarnings);
 }
 
+function extractProtocol(url: string): string {
+  const protocolEnd = url.indexOf(':');
+  return protocolEnd < 0 ? '' : url.substr(0, protocolEnd + 1);
+}
+
+function extractHostAndPort(
```
samhorlbeck, line 85: I'm definitely glad this will be tree shaken :+1:

packages-exp/auth-exp/src/core/auth/emulator.ts
```diff
@@ -58,22 +58,60 @@ export function useAuthEmulator(
...
+  url: string
+): { host: string; port: number | null } {
+  const protocol = extractProtocol(url);
```
samhorlbeck, line 88: Should we always try to use URL() first in the happy path? That way this code is only being exercised on a few platforms

yuchenshi: Impl standpoint: `new URL(input)` works fine in React Native, it just throws on `urlInstance.hostname` the getter. This may change at any time and some methods are stubs or half-baked implementation and won't throw. This makes it annoying to feature-detect and I'd rather not work with such fragile assumptions. It will lead to hard to debug issues with combinations of React Native versions and TS SDK versions (`O(M*N)`).

Testing standpoint: I'd rather just have one code path that is not platform dependent and always tested rather than two. Our implementation is not a proper polyfill and let's not pretend it is. If something breaks here we can just fix it for all platforms.

samhorlbeck: Fair enough. I'm fine with that since this is just for enabling the emulator, not in any code that is meant to reach production

samhorlbeck (approved):

Feiyang1 (approved):

Feiyang1:
Please fix lint errors.

PR #4550, opened 2021-02-26
[Auth] set key in storage before redirect, check this key before checking for redirect events

samhorlbeck:
This change prevents checking for redirects unless a key is set in storage. This brings the new SDK closer in line with the old SDK.

This change also similarly adds a new field to the resolver called `_shouldInitProactively`. This field is true for specific environments where opening the iframe is costly or can break the popup flow (i.e. Safari).

### DIFF
--- a/packages-exp/auth-exp/src/platform_browser/auth.test.ts
+++ b/packages-exp/auth-exp/src/platform_browser/auth.test.ts
@@ -46,6 +46,7 @@ import { PopupRedirectResolverInternal } from '../model/popup_redirect';
 import { UserCredentialImpl } from '../core/user/user_credential_impl';
 import { UserInternal } from '../model/user';
 import { _createError } from '../core/util/assert';
+import { makeMockPopupRedirectResolver } from '../../test/helpers/mock_popup_redirect_resolver';
 
 use(sinonChai);
 use(chaiAsPromised);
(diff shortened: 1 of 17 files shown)


packages-exp/auth-exp/src/platform_browser/auth.test.ts
```diff
@@ -191,6 +192,28 @@ describe('core/auth/initializeAuth', () => {
       expect(reload._reloadWithoutSaving).not.to.have.been.called;
     });
 
+    it('does not early-initialize the resolver', async () => {
```
yuchenshi, line 195:
```suggestion
    it('does not early-initialize the resolver if _shouldInitProactively is false', async () => {
```

packages-exp/auth-exp/src/platform_browser/auth.test.ts
```diff
@@ -191,6 +192,28 @@ describe('core/auth/initializeAuth', () => {
...
+      const resolverInternal: PopupRedirectResolverInternal = _getInstance(
+        popupRedirectResolver
+      );
+      sinon.stub(resolverInternal, '_shouldInitProactively').value(false);
+      sinon.spy(resolverInternal, '_initialize');
+      await initAndWait(inMemoryPersistence, popupRedirectResolver);
+      expect(resolverInternal._initialize).not.to.have.been.called;
+    });
+
+    it('does early-initialize the resolver', async () => {
```
yuchenshi, line 206:
```suggestion
    it('early-initializes the resolver if _shouldInitProactively is true', async () => {
```

packages-exp/auth-exp/test/helpers/redirect_persistence.ts
```diff
@@ -0,0 +1,33 @@
...
+export class RedirectPersistence extends InMemoryPersistence {
+  hasPendingRedirect = false;
+  redirectUser: object | null = null;
+
+  async _get<T extends PersistenceValue>(key: string): Promise<T | null> {
+    if (key.includes('pendingRedirect')) {
+      return this.hasPendingRedirect.toString() as T;
+    }
+
+    return this.redirectUser as T | null;
```
yuchenshi, line 31: nit: consider checking the key here for redirectUser and throw if the key is neither pendingRedirect or redirectUser to prevent misuse (and false positives in tests)

yuchenshi (approved):

packages-exp/auth-exp/test/helpers/redirect_persistence.ts
samhorlbeck: Done

PR #4565, opened 2021-03-02
Treeshake SyncPoint

schmidt-sebastian:


packages/database/src/core/SyncPoint.ts
```diff
@@ -41,239 +39,254 @@ let __referenceConstructor: ReferenceConstructor;
...
+  );
+  referenceConstructor = val;
+}
+
+function syncPointGetReferenceConstructor(): ReferenceConstructor {
+  assert(referenceConstructor, 'Reference.ts has not been loaded');
+  return referenceConstructor;
+}
+
+export class SyncPoint {
```
hsubox76, line 60: Does this still need to be a class?

schmidt-sebastian: We can make it an interface that just contains the one member, but that means that some callsites need to be aware of the internals, which used to be private. That's why I left it as is.

Feiyang1 (approved):

PR #4574, opened 2021-03-02
Update the phone tests to work with the emulator

samhorlbeck:
This change introduces a new mechanism for detecting emulation and then pulling the OTP from the emulator API


yuchenshi (approved):
Nice!

PR #4575, opened 2021-03-02
Add `initializeAnalytics()` to analytics-exp

hsubox76:
Create an `initializeAnalytics()` method that can only be called once, which can take an `AnalyticsOptions` object with an initial config. Allows users to set useful custom gtag config parameters such as `send_page_view` on initialization of analytics.

Updated some documentation comments.

Design doc (internal link):
go/firebase-analytics-initial-config

### DIFF
--- a/packages-exp/analytics-exp/src/factory.ts
+++ b/packages-exp/analytics-exp/src/factory.ts
@@ -21,7 +21,7 @@ import { getOrCreateDataLayer, wrapOrCreateGtag } from './helpers';
 import { AnalyticsError, ERROR_FACTORY } from './errors';
 import { _FirebaseInstallationsInternal } from '@firebase/installations-exp';
 import { areCookiesEnabled, isBrowserExtension } from '@firebase/util';
-import { _initializeAnalytics } from './initialize-analytics';
+import { initializeAnalytics } from './initialize-analytics';
 import { logger } from './logger';
 import { FirebaseApp, _FirebaseService } from '@firebase/app-exp';
 
@@ -136,7 +136,7 @@ export function getGlobalVars(): {
  */
 export function settings(options: SettingsOptions): void {
   if (globalInitDone) {
-    throw ERROR_FACTORY.create(AnalyticsError.ALREADY_INITIALIZED_SETTINGS);
+    throw ERROR_FACTORY.create(AnalyticsError.ALREADY_INITIALIZED);
   }
   if (options.dataLayerName) {
     dataLayerName = options.dataLayerName;
@@ -174,7 +174,11 @@ function warnOnBrowserContextMismatch(): void {
  * Analytics instance factory.
  * @internal
  */
-export function factory(app: FirebaseApp): AnalyticsService {
+export function factory(
+  app: FirebaseApp,
+  installations: _FirebaseInstallationsInternal,
+  options?: AnalyticsOptions
+): AnalyticsService {
   warnOnBrowserContextMismatch();
   const appId = app.options.appId;
   if (!appId) {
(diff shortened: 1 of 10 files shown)


packages-exp/analytics-exp/src/factory.ts
```diff
@@ -218,18 +215,29 @@ export function factory(
...
+export function _initializeAnalyticsForApp(
+  app: FirebaseApp,
+  installations: _FirebaseInstallationsInternal,
+  options?: AnalyticsOptions
+): void {
   // Async but non-blocking.
   // This map reflects the completion state of all promises for each appId.
-  initializationPromisesMap[appId] = initializeAnalytics(
+  // Factory will have thrown if options has no appId.
+  initializationPromisesMap[app.options.appId!] = _initializeAnalytics(
```
Feiyang1, line 234: After giving it some thought, I realize we can't do initialization outside the factory, otherwise FCM, which depends on Analytics, might try to use an uninitialized Analytics service, as shown in the below example:

```
import { getMessaging, onMessage } from 'firebase/messaging';
import { getAnalytics } from 'firebase/analytics';

const app = initializeApp({...});

/**
 * Messaging calls `const analytics = await messaging.firebaseDependencies.analyticsProvider.get();` 
 * which will create and return an uninitialized Analytics service.
 * And it will probably throw when calling `analytics.logEvent(...)`
 */
const msg = getMessaging(app);
onMessage(msg, () => { 
    // process messages
});

// getAnalytics async after a timeout
setTimeout(() => getAnalytics(), 1000);
```

We should let analytics initialize with the default options in this situation.
To support this,  some changes to the component framework need to be made, e.g. allow it to take an option and pass it to the factory.

I need to think more about it and let's chat offline.  This PR will need to be put on hold for now, Sorry!

Feiyang1: https://github.com/firebase/firebase-js-sdk/pull/4595 is merged. We can use `Provider.initialize()` to initialize Auth in the factory function.

packages-exp/analytics-exp/src/initialize-analytics.ts
```diff
@@ -121,12 +123,13 @@ export async function initializeAnalytics(
   // We keep it together with other initialization logic for better code structure.
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   (gtagCore as any)('js', new Date());
+  // User config added first. We don't want users to accidentally overwrite
+  // base Firebase config properties.
+  const configProperties: Record<string, unknown> = options?.config ?? {};
```
Feiyang1, line 128: nice! a lot of ?s

packages-exp/analytics-exp/testing/integration-tests/integration.ts
```diff
@@ -36,36 +37,71 @@ const RETRY_INTERVAL = 1000;
...
+      async function checkForEventCalls(): Promise<PerformanceEntry[]> {
+        await new Promise(resolve => setTimeout(resolve, RETRY_INTERVAL));
+        const resources = performance.getEntriesByType('resource');
+        const callsWithEvent = resources.filter(
+          resource =>
+            resource.name.includes('google-analytics.com') &&
+            resource.name.includes('en=login')
+        );
+        if (callsWithEvent.length === 0) {
+          return checkForEventCalls();
```
Feiyang1, line 87: We should add a limit on the number of retries, otherwise it may run forever.

Feiyang1 (changes requested):

packages-exp/analytics-exp/testing/integration-tests/integration.ts
hsubox76: I thought the karma timeout will catch it but I added a timeout and extracted the function, which helped me realize the second test was not correct (performance entries were not cleared from the browser between tests), so I fixed it.

Feiyang1 (approved):

PR #4578, opened 2021-03-03
[Auth] Add a bunch of integration tests for the various OOB Code flows

samhorlbeck:
These tests are purely local; emulator required.

### DIFF
--- a/packages-exp/auth-exp/test/helpers/integration/emulator_rest_helpers.ts
+++ b/packages-exp/auth-exp/test/helpers/integration/emulator_rest_helpers.ts
@@ -18,8 +18,20 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { Auth } from '@firebase/auth-exp';
 import { getApps } from '@firebase/app-exp';
+import { FetchProvider } from '../../../src/core/util/fetch_provider';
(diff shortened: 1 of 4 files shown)


packages-exp/auth-exp/test/helpers/integration/emulator_rest_helpers.ts
```diff
@@ -29,19 +41,47 @@ interface VerificationCodesResponse {
...
+  auth: Auth
+): Promise<Record<string, OobCodeSession>> {
+  assertEmulator(auth);
+  const url = getEmulatorUrl(auth, 'oobCodes');
+  const response: OobCodesResponse = await (
+    await FetchProvider.fetch()(url)
+  ).json();
+
+  return response.oobCodes.reduce((accum, session) => {
+    accum[session.email] = session;
```
yuchenshi, line 80: This will be flaky if more than one email is sent to the same address (e.g. the recover OOB when changing emails). I'd advise keeping an array and asserting the array length.

samhorlbeck: Done.

yuchenshi (approved):

PR #4579, opened 2021-03-03
Publicly export auth error codes

hsubox76:
This is for the current package only, we'll have to think of a different approach for exp.


packages/firebase/index.d.ts
```diff
@@ -5277,6 +5277,103 @@ declare namespace firebase.User {
   }
 }
 
+/**
+ * Developer-facing Firebase Auth error codes.
+ */
+declare namespace firebase.auth.ErrorCode {
+  var ADMIN_ONLY_OPERATION: string;
```
yuchenshi, line 5284: Shall we use the enum value as type instead of `string`?

yuchenshi (approved):

samhorlbeck:
I'm a little concerned about the size increase, but this package is already fairly large as it is

laurentpayot:
Maybe a simple `const E = fireauth.authenum.Error` or something like that to replace the 91 occurences of `fireauth.authenum.Error` in [exports_auth.js ](https://github.com/firebase/firebase-js-sdk/pull/4579/files#diff-4bcb4558bec4ba2fa4140cccb5efde7777166a5df6ca313b3c5d1905dad17d3f) could save us some bandwidth? :wink:

packages/firebase/index.d.ts
hsubox76: Done.

hsubox76:
Going to abandon this due to the size increase, which users cannot opt out of in the current version of the SDK. However, we plan to provide an error code object map in the upcoming modularized version of auth, that users can choose to import or not, depending on size concerns.

PR #4580, opened 2021-03-03
[Auth] Add some initial WebDriver tests against the emulator

samhorlbeck:
This change lays the groundwork for the suite of Selenium WebDriver tests. As a proof of concept I've added a couple simple anonymous auth tests. I've also updated the README to explain the various kinds of tests in the library and the commands to use.


packages-exp/auth-exp/README.md
```diff
@@ -2,4 +2,46 @@
...
+are not broken into such granular details. Check out `package.json` for more.
+
+### Integration testing with the emulator
+
+To test against the emulator, set up the Auth emulator
+([instructions](https://firebase.google.com/docs/emulator-suite/connect_and_prototype)).
+When starting the emulators, make sure to specify the project ID manually
+when issuing the command (the local integration tests use a fake project ID):
+`firebase emulators:start --project test-emulator`). Once the emulator is
+started and waiting, you can execute the tests. In addition to the npm commands
```
yuchenshi, line 31: These tests should use `emulators:exec` which starts the emulator and runs the script. More importantly, it sets the `GCLOUD_PROJECT` and `FIREBASE_AUTH_EMULATOR` env vars which should be used in the test script. Don't hard code `localhost:9099` or `test-emulator` please.

packages-exp/auth-exp/README.md
```diff
@@ -2,4 +2,55 @@
...
+you would simply execute the following command:
+
+```sh
+firebase emulators:exec --project foo-bar --only auth "yarn test:integration:local"
+```
+
+### Selenium Webdriver tests
+
+These tests assume that you have both Firefox and Chrome installed on your
+computer and in your `PATH`. The tests will error out if this is not the case.
```
yuchenshi, line 52:
```suggestion
computer and in your `$PATH`. The tests will error out if this is not the case.
```

packages-exp/auth-exp/test/integration/webdriver/static/index.html
```diff
@@ -0,0 +1,3 @@
+<body>
```
yuchenshi, line 1: Add `<!DOCTYPE html>`, drop `<body>`

yuchenshi (approved):

PR #4582, opened 2021-03-04
Treeshake Tree

schmidt-sebastian:


packages/database/src/core/util/Tree.ts
```diff
@@ -15,20 +15,20 @@
...
-export class TreeNode<T> {
+export interface TreeNode<T> {
   // TODO: Consider making accessors that create children and value lazily or
   // separate Internal / Leaf 'types'.
-  children: { [name: string]: TreeNode<T> } = {};
-  childCount = 0;
-  value: T | null = null;
+  children: Record<string, TreeNode<T>>;
+  childCount: number;
+  value?: T;
```
schmidt-sebastian, line 31: Made this undefined so it doesn't have to be initialized in its callsites.

hsubox76 (approved):

PR #4589, opened 2021-03-05
Add some redirect webdriver tests

samhorlbeck:
There are still a lot more tests to write, but this is a proof of concept

### DIFF
--- a/packages-exp/auth-exp/test/integration/webdriver/redirect.test.ts
+++ b/packages-exp/auth-exp/test/integration/webdriver/redirect.test.ts
@@ -18,13 +18,13 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { OperationType, UserCredential } from '@firebase/auth-exp';
 import { expect } from 'chai';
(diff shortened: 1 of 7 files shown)


packages-exp/auth-exp/test/integration/webdriver/redirect.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    await driver.callNoWait(TestFunction.IDP_REDIRECT);
+    const widget = new IdPPage(driver.webDriver);
+
+    // We're now on the widget page; wait for load
+    await widget.pageLoad();
+    await widget.clickAddAccount();
+    await widget.fillEmail('bob@bob.test');
+    await widget.fillDisplayName('Bob Test');
+    await widget.fillScreenName('bob.test');
+    await widget.fillProfilePhoto('bob.test/bob');
```
yuchenshi, line 37: Please use a full URL. It's probably an Auth Emulator bug for not requiring a full URL and it may get fixed one day.

yuchenshi (approved):

PR #4596, opened 2021-03-09
Update timeout for starting browsers.

yuchenshi:
It times out most of the time (4 out of 5) on my laptop.

### DIFF
--- a/packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
+++ b/packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
@@ -57,7 +57,8 @@ export function browserDescribe(
 setTimeout(() => {
   for (const browser of BROWSERS) {
     describe(`Testing in browser "${browser}"`, () => {
-      before(async () => {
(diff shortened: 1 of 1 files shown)


packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
```diff
@@ -57,7 +57,8 @@ export function browserDescribe(
 setTimeout(() => {
   for (const browser of BROWSERS) {
     describe(`Testing in browser "${browser}"`, () => {
-      before(async () => {
+      before(async function () {
+        this.timeout(20000); // Starting browsers can be slow.
```
samhorlbeck, line 61: Don't you have to type the this variable?

samhorlbeck (approved):

packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
yuchenshi: Nope :) automatic

PR #4602, opened 2021-03-09
[Auth] Add popup WebDriver tests (duplicated from the redirect tests)

samhorlbeck:


packages-exp/auth-exp/test/integration/webdriver/static/index.js
```diff
@@ -18,12 +18,14 @@
...
 import * as anonymous from './anonymous';
 import * as core from './core';
+import * as popup from './popup';
 import { initializeApp } from '@firebase/app-exp';
 import { getAuth, useAuthEmulator } from '@firebase/auth-exp';
 
-window.core = core;
-window.anonymous = anonymous;
-window.redirect = redirect;
+window.core = { ...core };
```
yuchenshi, line 25: what does this do?

yuchenshi (approved):

packages-exp/auth-exp/test/integration/webdriver/static/index.js
samhorlbeck: In `popup.js` and `redirect.js`, I was previously assigning state to the window directly (things like `errorCred`). I didn't want that to clash by using it in multiple files (it applies to both popup and redirect) so I started putting it on these objects (`window.redirect.errorCred` and `window.popup.errorCred`). That doesn't work though if you make `window.popup` the imported module directly.

PR #4614, opened 2021-03-10
Add database@exp API

schmidt-sebastian:
This adds API stubs (see go/rtdb-next)

### DIFF
--- a/packages/database/test/exp/integration.test.ts
+++ b/packages/database/test/exp/integration.test.ts
@@ -20,13 +20,19 @@ import { initializeApp, deleteApp } from '@firebase/app-exp';
 import { expect } from 'chai';
 import { DATABASE_ADDRESS, DATABASE_URL } from '../helpers/util';
 
-import { getDatabase } from '../../exp/index';
+import {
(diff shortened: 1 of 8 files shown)


packages/database/test/exp/integration.test.ts
```diff
@@ -20,13 +20,19 @@ import { initializeApp, deleteApp } from '@firebase/app-exp';
...
+  ref,
+  refFromURL
+} from '../../exp/index';
 
 export function createTestApp() {
   return initializeApp({ databaseURL: DATABASE_URL });
 }
 
-describe('Database Tests', () => {
+describe.skip('Database Tests', () => {
```
hsubox76, line 35: Do you want to put a todo to make sure this is removed when the code is ready or will this be obvious enough to remember?

hsubox76 (approved):

packages/database/test/exp/integration.test.ts
schmidt-sebastian: Better safe than sorry :)

PR #4616, opened 2021-03-11
Build and release Firestore compat 

Feiyang1:
Changes to facilitate build and publish Firestore compat to npm.


packages/firestore/compat/bundle/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
schmidt-sebastian, line 1: I am wondering if we need this since Firestore Compat is meant to be temporary. I wouldn't be opposed to just including Bundle support by default.

And I also think we should include it in the main API in the near future (once all the size-conscious users have migrated).

packages/firestore/compat/index.memory.ts
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import firebase from '@firebase/app-compat';
+import { FirebaseNamespace } from '@firebase/app-types';
+
+import { Firestore, MemoryPersistenceProvider } from '../src/api/database';
```
schmidt-sebastian, line 21: Same thought here - I don't think we need to provide a memory build.

packages/firestore/exp/index.ts
```diff
@@ -130,3 +131,5 @@ export { Timestamp } from '../src/exp/timestamp';
 export { CACHE_SIZE_UNLIMITED } from '../src/exp/database';
 
 export { FirestoreErrorCode, FirestoreError } from '../src/util/error';
+
+export { AbstractUserDataWriter } from '../src/lite/user_data_writer';
```
schmidt-sebastian, line 135: Does this not break our public API generation? I would think that this ends up showing up in `public.d.ts` (but it should be an easy fix that I can provide as a follow up)

schmidt-sebastian (commented):
Thanks for doing this. Some thoughts on simplification.

packages/firestore/exp/index.ts
Feiyang1: It's removed from `public.d.ts` because it's `@internal.`

packages/firestore/compat/bundle/package.json
Feiyang1: Do you remember how large the bundle code is? I'm a little concerned if people migrate to compat and find the size of their app increases significantly (though it could be a motivation for them to migrate to the modular API).

What do you mean by the main API, in v8? I don't think we are going to touch v8 once v9 is published.

schmidt-sebastian: Main API - v8.

I think bundles was 30Kb (before gzip).

packages/firestore/compat/config.ts
```diff
@@ -0,0 +1,92 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
hsubox76, line 3: 2021

packages/firestore/compat/config.ts
```diff
@@ -0,0 +1,92 @@
...
+  QuerySnapshot,
+  CollectionReference,
+  FieldPath,
+  FieldValue,
+  setLogLevel,
+  CACHE_SIZE_UNLIMITED
+};
+
+/**
+ * Configures Firestore as part of the Firebase SDK by calling registerService.
```
hsubox76, line 66: registerComponent?

packages/firestore/compat/index.bundle.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
hsubox76, line 3: All 2020 => 2021

scripts/exp/prepare-firestore-for-exp-release.ts
```diff
@@ -65,9 +73,110 @@ export async function prepare() {
...
+
+  // update memory/bundle/packages.json
+  await transformFile(
+    resolve(FIRESTORE_COMPAT_DEST, 'memory/bundle/package.json'),
+    async content =>
+      content.replace(/\.\.\/\.\.\/\.\.\/dist\/compat/g, '../../dist')
+  );
+}
+
+function copyRescursiveSync(
```
hsubox76, line 149: typo I think, "Rescursive" => "Recursive"

hsubox76 (approved):

packages/firestore/compat/bundle/package.json
Feiyang1: Removed bundle and memory builds. I think it's reasonable to assume size conscious customers will migrate to the modular SDK, and thus providing bundle/memory builds in the compat layer provides limited value while complicates the build.

PR #4620, opened 2021-03-12
Add persistence tests and fix some issues.

yuchenshi:


packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -113,19 +113,53 @@ export class PersistenceUserManager {
       );
     }
 
-    const key = _persistenceKeyName(userKey, auth.config.apiKey, auth.name);
+    // Use the first persistence that supports a full read-write roundtrip (._isAvailable()).
+    let chosenPersistence: PersistenceInternal | null = null;
```
samhorlbeck, line 117: Instead of setting this to `| null`, just set it to inMemory here as the default. `_getInstance()` has almost no overhead

packages-exp/auth-exp/test/integration/webdriver/persistence.test.ts
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { UserCredential } from '@firebase/auth-exp';
+import { expect } from 'chai';
+import { API_KEY } from '../../helpers/integration/settings';
+import { AnonFunction, PersistenceFunction } from './util/functions';
+import { browserDescribe } from './util/test_runner';
+
+browserDescribe('WebDriver persistence test', driver => {
```
samhorlbeck, line 25: Side note: we should write some tests for the persistence events and cross-tab/cross-browser synchronization (in a different PR) b/182567813

packages-exp/auth-exp/test/integration/webdriver/static/index.js
```diff
@@ -18,12 +18,14 @@
...
 import * as anonymous from './anonymous';
 import * as core from './core';
+import * as persistence from './persistence';
 import { initializeApp } from '@firebase/app-exp';
 import { getAuth, useAuthEmulator } from '@firebase/auth-exp';
 
 window.core = core;
 window.anonymous = anonymous;
 window.redirect = redirect;
+window.persistence = persistence;
```
samhorlbeck, line 28: I haven't merged it yet but I've changed all of these so that they no longer set the whole module and instead set an object (so that the individual modules can namespace):

`window.persistence = {...persistence};`

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
yuchenshi: done

packages-exp/auth-exp/test/integration/webdriver/static/index.js
yuchenshi: Done. Will clean this up in a follow up PR

samhorlbeck (approved):

PR #4627, opened 2021-03-12
Enable strict mode and remove module prop hacks.

yuchenshi:
I've also updated the error handling so you get meaningful errors if you try those things.


packages-exp/auth-exp/test/integration/webdriver/util/auth_driver.ts
```diff
@@ -31,6 +31,15 @@ import { authTestServer } from './test_server';
 const START_FUNCTION = 'startAuth';
 const PASSED_ARGS = '...Array.prototype.slice.call(arguments, 0, -1)';
 
+type DriverCallResult =
+  | { type: 'success'; value: string /* JSON stringified */ }
+  | {
```
samhorlbeck, line 36: nice

samhorlbeck (approved):

PR #4630, opened 2021-03-13
tree shake WriteTree and WriteTreeRef

Feiyang1:


yarn.lock
```diff
@@ -1219,6 +1219,13 @@
     faye-websocket "0.11.3"
     tslib "^1.11.1"
 
+"@firebase/util@0.3.4":
```
schmidt-sebastian, line 1222: Is this intentional?

packages/database/src/core/WriteTree.ts
```diff
@@ -59,575 +59,745 @@ export interface WriteRecord {
...
+    return [];
   }
+}
 
+/**
+ * WriteTree tracks all pending user-initiated writes and has methods to calculate the result of merging them
+ * with underlying server data (to create "event cache" data).  Pending writes are added with addOverwrite()
+ * and addMerge(), and removed with removeWrite().
+ */
+export class WriteTree {
```
schmidt-sebastian, line 647: This could also be an interface. It is only created in one place (but outside this module - so this change would leak some internals unless you create a `newWriteTree()` function).

packages/database/src/core/WriteTree.ts
```diff
@@ -59,575 +59,745 @@ export interface WriteRecord {
...
+  writeTreeRef: WriteTreeRef,
+  childName: string
+): WriteTreeRef {
+  return new WriteTreeRef(
+    pathChild(writeTreeRef.treePath_, childName),
+    writeTreeRef.writeTree_
+  );
 }
 
 /**
```
schmidt-sebastian, line 803: Even if you keep `WriteTree` as a class, `WriteTreeRef` should probably still be converted to an interface since it is only constructed in this module.

schmidt-sebastian (approved):

packages/database/src/core/WriteTree.ts
Feiyang1: Changed to an interface.

packages/database/src/core/WriteTree.ts
Feiyang1: Changed to an interface

hsubox76 (approved):

Feiyang1:
Blocked by https://github.com/firebase/firebase-js-sdk/pull/4638

PR #4645, opened 2021-03-17
Add tests for setPersistence and fix migrations.

yuchenshi:


packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -54,12 +49,18 @@ export class Auth
 
     // Note this is slightly different behavior: in this case, the stored
     // persistence is checked *first* rather than last. This is because we want
-    // the fallback (if no user is found) to be the stored persistence type
-    const storedPersistence = _getPersistenceFromRedirect(this.auth);
-    const persistences = storedPersistence ? [storedPersistence] : [];
-    persistences.push(exp.indexedDBLocalPersistence);
+    // to prefer stored persistence type in the hierarchy.
```
yuchenshi, line 52: @samhorlbeck please audit the old and new code closely since I'm not sure if I fully understood the logic.

packages-exp/auth-compat-exp/src/persistence.ts
```diff
@@ -113,16 +113,14 @@ export function _getPersistenceFromRedirect(
   const persistence = win.sessionStorage.getItem(key);
 
   switch (persistence) {
-    case exp.inMemoryPersistence.type:
-      return exp.inMemoryPersistence;
-    case exp.indexedDBLocalPersistence.type:
-      return exp.indexedDBLocalPersistence;
-    case exp.browserSessionPersistence.type:
-      return exp.browserSessionPersistence;
-    case exp.browserLocalPersistence.type:
```
yuchenshi, line 122: Previously this case can never be reached since `exp.indexedDBLocalPersistence.type` is also `LOCAL`. I've changed this switch statement to use `Persistence.xxx` to make it more obvious.

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -199,28 +200,29 @@ export class Auth
...
-        case Persistence.NONE:
-          return exp.inMemoryPersistence;
-        default:
-          return exp._fail(exp.AuthErrorCode.ARGUMENT_ERROR, {
-            appName: auth.name
-          });
-      }
+    _validatePersistenceArgument(this.auth, persistence);
+    let converted;
+    switch (persistence) {
```
yuchenshi, line 205: Based on my testing, legacy SDK only accepts `local`, `session`, or `none` as lowercase, not UPPERCASE. Did we unintentionally introduce a breaking change here?

samhorlbeck: No we should fix that :P

packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -199,28 +200,29 @@ export class Auth
...
+      case Persistence.SESSION:
+        converted = exp.browserSessionPersistence;
+        break;
+      case Persistence.LOCAL:
+        // Not using isIndexedDBAvailable() since it only checks if indexedDB is defined.
+        const isIndexedDBFullySupported = await (exp.indexedDBLocalPersistence as exp.PersistenceInternal)._isAvailable();
```
samhorlbeck, line 211: You'll need to run the prettier. From the root of the directory (after a commit on a clean client):
`node tools/gitHooks/precommit.js`

packages-exp/auth-compat-exp/src/persistence.ts
samhorlbeck: Makes sense

packages-exp/auth-compat-exp/src/auth.ts
samhorlbeck: This makes sense to me

samhorlbeck (approved):

packages-exp/auth-compat-exp/src/auth.ts
yuchenshi: Prettier insists on formatting it like this -- I tried wrapping the lines myself and `node tools/gitHooks/precommit.js` would undo it and make it all in one line.

packages-exp/auth-compat-exp/src/auth.ts
yuchenshi: Changed compat to lowercase while leaving `auth-exp` unchanged for now. This will probably also help us tell if we mistaken one for another.

PR #4647, opened 2021-03-17
Use generics for some firestore types

### LINKED ISSUE #4646: FR: DocumentData as generic type

  * Operating System version: macOS 10.14.6
  * Browser version: Chrome latest
  * Firebase SDK version: 8.2.10
  * Firebase Product: firestore

Using Firestore with TypeScript I've found its very useful to type the documents that come back from either `db.collection` or `db.doc`. So far I have found the only way to do this is using type assertion, but it seems like it would be very useful to allow `DocumentData` to be a generic type which would offer some flexibility to us developers on how we write our code. This is especially useful for me because of wrapping database calls in hooks such as `useCollection` and `useDocument`. 

So right now the types for these functions look like this:

```typescript
    collection(collectionPath: string): CollectionReference<DocumentData>;
    doc(documentPath: string): DocumentReference<DocumentData>;
    collectionGroup(collectionId: string): Query<DocumentData>;
```

So then I have to do things like this to type the values that come back from the queries:

```typescript
const data = snap.data() as Product;

// or

db.collection('products').doc('12345').get() as firebase.firestore.DocumentSnapshot<Product>;
```

I think it would be really useful if we updated the types to be something like this:

```typescript
    collection<T = DocumentData>(collectionPath: string): CollectionReference<T>;
    doc<T = DocumentData>(documentPath: string): DocumentReference<T>;
    collectionGroup<T = DocumentData>(collectionId: string): Query<T>;
```

And then I can write:

```typescript
const p = await database.doc<Product>('products/123').get().then(snap => snap.data());
// p is already a Product!
```

I notice that some other classes allow generics but not these which are most commonly used in the Firebase docs.

jsakas:
All use of generic types on Firebase functions `collection`, `doc`, and `collectionGroup`.

Closes #4646

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -8168,7 +8168,9 @@ declare namespace firebase.firestore {
      * @param collectionPath A slash-separated path to a collection.
      * @return The `CollectionReference` instance.
      */
-    collection(collectionPath: string): CollectionReference<DocumentData>;
+    collection<T = DocumentData>(
+      collectionPath: string
+    ): CollectionReference<T>;
 
     /**
      * Gets a `DocumentReference` instance that refers to the document at the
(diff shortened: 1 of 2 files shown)


jsakas:
@googlebot I signed it!

jsakas:
Having some trouble getting the tests to run locally (even on master branch with project set up) so hard to test it locally but I will look into it further.

PR #4648, opened 2021-03-17
Add compat packages to firebase-exp

Feiyang1:


packages-exp/auth-compat-exp/index.ts
```diff
@@ -140,5 +139,4 @@ function registerAuthCompat(instance: _FirebaseNamespace): void {
   instance.registerVersion('auth', version);
 }
 
-impl.registerAuth(_getClientPlatform());
```
hsubox76, line 143: What happened with this, did getClientPlatform turn out to be unnecessary?

packages-exp/firebase-exp/compat/auth/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
hsubox76, line 3: 2020 -> 2021 and elsewhere

packages-exp/firebase-exp/compat/empty-import.d.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2018 Google LLC
```
hsubox76, line 3: 2018 -> 2021

packages-exp/firebase-exp/compat/index.ts
```diff
@@ -39,15 +39,15 @@ import 'firebase/<PACKAGE>';
 import firebase from './app';
 import { name, version } from '../package.json';
 
-// import './auth';
+import './auth';
 // import './database';
 // import './firestore';
```
hsubox76, line 44: firestore/storage being added in other PRs?

packages-exp/auth-compat-exp/index.ts
Feiyang1: Auth-exp will register itself, so it's not necessary to do it in auth-compat

packages-exp/firebase-exp/compat/index.ts
Feiyang1: Good catch. added them here.

packages-exp/firebase-exp/compat/auth/index.ts
Feiyang1: Updated and updated the license script.

hsubox76 (approved):

PR #4656, opened 2021-03-19
Export the compat implementation for FieldPath and FieldValue

Feiyang1:
Changes:

- export compat FieldValue and FieldPath from firestore-compat
- export `Compat` from exp, so `instanceof` works correctly across packages
  - This is a stopgap solution till we implement the interop proposal for all packages.


thebrianchen (approved):
lgtm

PR #4662, opened 2021-03-19
Require _delegate in compat classes

hsubox76:


packages-exp/auth-compat-exp/src/auth.ts
```diff
@@ -41,16 +41,16 @@ export class Auth
   implements compat.FirebaseAuth, Wrapper<exp.Auth>, _FirebaseService {
   // private readonly auth: impl.AuthImpl;
 
-  constructor(readonly app: FirebaseApp, private readonly auth: exp.AuthImpl) {
+  constructor(readonly app: FirebaseApp, readonly _delegate: exp.AuthImpl) {
```
samhorlbeck, line 44: We also have "user.ts" which has an underlying exp class as well. Should that also follow this pattern? `updateProfile(user, {})` vs `user.updateProfile({})` for example

samhorlbeck: Nevermind, that seems irrelevant for this PR since this is specifically for `_FirebaseService`

packages-exp/auth-compat-exp/src/phone_auth_provider.ts
```diff
@@ -49,15 +49,15 @@ export class PhoneAuthProvider
       | compat.PhoneMultiFactorSignInInfoOptions,
     applicationVerifier: compat.ApplicationVerifier
   ): Promise<string> {
-    return this.phoneProvider.verifyPhoneNumber(
+    return this._delegate.verifyPhoneNumber(
       // The implementation matches but the types are subtly incompatible
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       phoneInfoOptions as any,
-      unwrap(applicationVerifier)
+      applicationVerifier
```
samhorlbeck, line 56: Isn't this a type error since we haven't updated the signatures yet?

samhorlbeck (approved):
LGTM for Auth

packages-exp/auth-compat-exp/src/phone_auth_provider.ts
Feiyang1: I didn't get any error. I think they might be duck type compatible.

packages/firestore/src/api/database.ts
```diff
@@ -1306,8 +1286,5 @@ export class CollectionReference<T = PublicDocumentData>
 function castReference<T>(
   documentRef: PublicDocumentReference<T>
 ): ExpDocumentReference<T> {
-  if (documentRef instanceof Compat) {
```
schmidt-sebastian, line 1309: Why is this not needed anymore?

Feiyang1: IIUC, `cast` does the same thing, so it's redundant:
```
  if ('_delegate' in obj) {
    // Unwrap Compat types
    obj = (obj as any)._delegate;
  }
```

schmidt-sebastian (approved):

Feiyang1 (approved):

PR #4665, opened 2021-03-22
Add async `isSupported` to Modular FM SDK

zwu52:
This PR does 2 things
- export async `isSupported` from `index.ts` and `index.sw.ts` (API Approved: go/js-messaging-is-supported)
- separate `isWindowSupported` and `isSwSupported` checks in its own context (`window` and `sw` respectively)  

Deprecating #4446 in favor of this PR


packages-exp/messaging-exp/src/index.sw.ts
```diff
@@ -15,16 +15,29 @@
...
   interface NameServiceMapping {
     'messaging-exp': FirebaseMessaging;
   }
 }
 
+// Top-level 'await' requires 'module' option set to 'esnext' or 'system', and 'target' option
+// set to 'es2017' or higher. For compatibility, use async expression here.
+void (async () => {
+  if (!(await isSwSupported())) {
+    throw ERROR_FACTORY.create(ErrorCode.UNSUPPORTED_BROWSER);
```
Feiyang1, line 39: I think we should check `isSupported()` in the factory function like you do originally, so that developers have a chance to call `isSupported()` themselves and to not initialize messaging in unsupported environments without getting an exception.

That means we need 2 factories, one for the client and one for the sw.

Feiyang1 (changes requested):

packages-exp/messaging-exp/src/index.sw.ts
zwu52: make sense. The solution here is problematic because devs can't handle before the error is thrown.

packages-exp/messaging-exp/src/api/isSupported.ts
```diff
@@ -15,20 +15,7 @@
...
-    // Assume we are in the window context.
-    return isWindowControllerSupported();
-  }
-}
-
-/**
- * Checks to see if the required APIs exist.
- */
-function isWindowControllerSupported(): boolean {
+export async function isWindowSupported(): Promise<boolean> {
```
Feiyang1, line 18: Why is it async? It seems all checks are sync.

packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -22,26 +22,42 @@ import {
...
     container.getProvider('installations-exp-internal').getImmediate(),
     container.getProvider('analytics-internal')
   );
 };
 
-export function registerMessaging(): void {
+export function registerWindowMessaging(): void {
+  // Top-level 'await' requires 'module' option set to 'esnext' or 'system', and 'target' option
+  // set to 'es2017' or higher. For compatibility, use async expression here.
+  void (async () => {
```
Feiyang1, line 43: This unfortunately still doesn't work because `registerMessaging()` gets executed automatically on import. We should move the check into `messagingFactory`

zwu52: couple reasons:
to match the other isSuppoorted API (auth?)
and be able to do idb.openDb check

packages-exp/messaging-exp/src/api/isSupported.ts
zwu52: ATM yes.
It's make to async firstly to make the other `isSupported` style. but more importmently to be able to check for idb.open

packages-exp/messaging-exp/src/helpers/register.ts
zwu52: copy. done

packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -41,8 +45,32 @@ const messagingFactory: InstanceFactory<'messaging-exp'> = (
   );
 };
 
-export function registerMessaging(): void {
+const SwMessagingFactory: InstanceFactory<'messaging-exp'> = (
+  container: ComponentContainer
+) => {
+  // Top-level 'await' requires 'module' option set to 'esnext' or 'system', and 'target' option
+  // set to 'es2017' or higher. For compatibility, use async expression here.
```
Feiyang1, line 52: Please add the comment since it's not at top level now. We still want to use the iife pattern here because the factory needs to be sync.

packages-exp/messaging-exp/src/util/errors.ts
```diff
@@ -24,6 +24,7 @@ export const enum ErrorCode {
   PERMISSION_DEFAULT = 'permission-default',
   PERMISSION_BLOCKED = 'permission-blocked',
   UNSUPPORTED_BROWSER = 'unsupported-browser',
+  INDEXED_DB_UNSUPPORTED = 'index-db-unsupported',
```
hsubox76, line 27: in the lowercase string, `index` => `indexed`

packages-exp/messaging-exp/src/util/errors.ts
```diff
@@ -50,6 +51,8 @@ export const ERROR_MAP: ErrorMap<ErrorCode> = {
     'The notification permission was not granted and blocked instead.',
   [ErrorCode.UNSUPPORTED_BROWSER]:
     "This browser doesn't support the API's required to use the firebase SDK.",
+  [ErrorCode.INDEXED_DB_UNSUPPORTED]:
+    "This browser context doesn't support the idb.open (ex. Safari iFrame, Firefox Private Browsing, etc)",
```
hsubox76, line 55: Suggest "doesn't support indexedDB.open()" (no "the") or "doesn't support the indexedDB.open() method"

packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -22,17 +22,27 @@ import {
...
+  // special handling.
+  isWindowSupported()
+    .then(isSupported => {
+      if (!isSupported) {
+        throw ERROR_FACTORY.create(ErrorCode.UNSUPPORTED_BROWSER);
+      }
+    })
+    .catch(_ => {
+      throw ERROR_FACTORY.create(ErrorCode.INDEXED_DB_UNSUPPORTED);
+    });
```
Feiyang1, line 45: This seems this code will never hit because `isWindowSupported()` & `isSwSupported ()` never throws.

zwu52: good catch!

Feiyang1 (approved):

PR #4669, opened 2021-03-22
[Auth] Update compat layer credential interception; add webdriver tests

samhorlbeck:
The compat layer uses the existing webdriver tests here; we swap out the JS source served to the browser to use the compat layer (keeping the same function names)

### DIFF
--- a/packages-exp/auth-compat-exp/package.json
+++ b/packages-exp/auth-compat-exp/package.json
@@ -25,7 +25,8 @@
     "test:browser:integration": "karma start --single-run --integration",
     "test:node": "ts-node -O '{\"module\": \"commonjs\", \"target\": \"es6\"}' scripts/run_node_tests.ts",
     "test:node:integration": "ts-node -O '{\"module\": \"commonjs\", \"target\": \"es6\"}' scripts/run_node_tests.ts --integration",
-    "test:integration": "run-s test:browser:integration test:node:integration"
+    "test:webdriver": "rollup -c test/integration/webdriver/static/rollup.config.js && ts-node -O '{\"module\": \"commonjs\", \"target\": \"es6\"}' scripts/run_node_tests.ts --webdriver",
+    "test:integration": "run-s test:browser:integration test:node:integration test:webdriver"
   },
   "peerDependencies": {
     "@firebase/app-compat": "0.x"
@@ -41,8 +42,8 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@firebase/app-compat": "0.x",
-    "rollup": "2.35.1",
     "@rollup/plugin-json": "4.1.0",
+    "rollup": "^2.35.1",
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-typescript2": "0.29.0",
     "typescript": "4.2.2"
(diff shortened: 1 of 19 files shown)


packages-exp/auth-compat-exp/package.json
```diff
@@ -41,8 +42,8 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@firebase/app-compat": "0.x",
-    "rollup": "2.35.1",
     "@rollup/plugin-json": "4.1.0",
+    "rollup": "^2.35.1",
```
yuchenshi, line 46: revert this

packages-exp/auth-compat-exp/src/user_credential.ts
```diff
@@ -17,35 +17,91 @@
...
+}
 
 function credentialFromResponse(
   userCredential: exp.UserCredentialInternal
 ): exp.AuthCredential | null {
-  const { providerId, _tokenResponse } = userCredential;
+  return credentialFromObject(userCredential);
+}
+
+function modifyError(auth: exp.Auth, e: FirebaseError): void {
```
yuchenshi, line 36: nit: this can use a more descriptive name such as addExtraErrorFields

packages-exp/auth-compat-exp/src/user_credential.ts
```diff
@@ -17,35 +17,91 @@
...
   // Email and password is not supported as there is no situation where the
   // server would return the password to the client.
   if (!providerId || providerId === exp.ProviderId.PASSWORD) {
     return null;
   }
 
+  // We know for a fact that the function will match the value type
+  // (based on the declaration of fn). We will therefore cast object to a
+  // meaningless type to bypass the type system
+  const castObject = object as exp.UserCredential & FirebaseError;
```
yuchenshi, line 94: This is some serious type hacking here. I think the best way forward is NOT to use string-typing at all. Let's drop `fn` and `CredentialFrom` and all that and do something like this:

```js
function credentialForProvider(provider: typeof BaseOAuthProvider, object: exp.UserCredential | FirebaseError) {
  return object instanceof FirebaseError ? provider.credentialFromError(object) : provider.credentialFromResult(object);
}
```

TypeScript should figure this out correctly.

packages-exp/auth-compat-exp/test/integration/webdriver/static/persistence.js
```diff
@@ -0,0 +1,143 @@
...
+      addEventListener(event, listener) {
+        if (event === 'error') {
+          void Promise.resolve({}).then(listener);
+        }
+      }
+    };
+  };
+}
+
+function dumpStorage(storage) {
```
yuchenshi, line 103: I don't like duplicating the test helper functions themselves. Maybe we should split test helpers into their own files and only make code under test different. We can think about this later though.

packages-exp/auth-compat-exp/src/user_credential.ts
yuchenshi: With a tiny bit of restructuring, you can even inline this by putting `let provider: typeof BaseOAuthProvider` in front of the switch and then pick it up later

packages-exp/auth-compat-exp/package.json
samhorlbeck: I believe we need this now that there's a separate rollup config for the webdriver tests. The main package has it as well

packages-exp/auth-compat-exp/test/integration/webdriver/static/persistence.js
samhorlbeck: Yeah that's fair. We'll need to separate the parts that invoke the SDK in that case

yuchenshi (approved):

PR #4672, opened 2021-03-24
Split listeners registration flow for `window` and `sw`

zwu52:
Specifically:
- index.ts (window entry) registers message listeners
- index.sw.ts registers the onPush, onSubChange, onClick listeners.

Note that some listeners registration requires an instance  of the firebase messaging. Therefore, the listeners registration process needs to be done once firebase sdk registration completes.


packages-exp/messaging-exp/src/index.sw.ts
```diff
@@ -15,16 +15,50 @@
...
+const messaging = getMessaging(getApp());
+self.addEventListener('push', e => {
+  e.waitUntil(onPush(e, messaging as MessagingService));
+});
+self.addEventListener('pushsubscriptionchange', e => {
+  e.waitUntil(onSubChange(e, messaging as MessagingService));
+});
+self.addEventListener('notificationclick', e => {
+  e.waitUntil(onNotificationClick(e));
+});
```
Feiyang1, line 64: I don't think we want to auto-initialize messaging on user's behalf. It also doesn't work well if user creates multiple app/messaging instances. Can we move the listeners registration to the `getMessaging()` function, so they are registered when developer initializes messaging?

It also means we will have 2 versions of `getMessaging()`, one for the client and one of the sw.

Feiyang1 (changes requested):
The PR seems to also contain changes from https://github.com/firebase/firebase-js-sdk/pull/4665

Do you mean to combine the two? If not, can you rebase this PR, so it only contains registration changes?

zwu52 (commented):
> The PR seems to also contain changes from #4665
> 
> Do you mean to combine the two? If not, can you rebase this PR, so it only contains registration changes?

not sure how to do it correctly. Tried rebase to the `isSupported` branch and pulled in a bunch of irrelevant files

Can you approve the isSupported PR and I'll rebase from origin

Feiyang1 (approved):

PR #4686, opened 2021-03-29
Compat and @exp class for DataSnapshot

schmidt-sebastian:
This PR creates the Compat class for DataSnapshot, and also adds the `@exp` class at the same time (mostly by moving code). 

There are a bunch of seemingly unrelated changes, but most of them are meant to remove internal references to the old legacy DataSnapshot from `/core`.

The next step would be to do this same change for Reference, since I only converted the code paths that I needed for the conversion.

One extra thing: The current SDK accepts numbers and string in some places for child keys, but this is not documented. I dropped accepting numbers for now.

### DIFF
--- a/packages/database/src/exp/DataSnapshot.ts
+++ b/packages/database/src/exp/DataSnapshot.ts
@@ -17,7 +17,7 @@
 
 import { child, Reference } from './Reference';
 import { Node } from '../core/snap/Node';
-import { Index } from '../core/snap/indexes';
+import { Index } from '../core/snap/indexes/Index';
 import { Path } from '../core/util/Path';
 import { PRIORITY_INDEX } from '../core/snap/indexes/PriorityIndex';
 import { ChildrenNode } from '../core/snap/ChildrenNode';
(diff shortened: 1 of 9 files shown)


packages/database/src/exp/DataSnapshot.ts
```diff
@@ -15,54 +15,90 @@
  * limitations under the License.
  */
 
-import { Reference } from './Reference';
+import { child, Reference } from './Reference';
+import { Node } from '../core/snap/Node';
+import { Index } from '../core/snap/indexes';
```
hsubox76, line 20: I think this is breaking the build because I think it wants `indexes/Index` and I guess this shortcut works if a file is named `index` but I think it might be case sensitive. Didn't go through the rest of the PR yet, just wanted to put in this comment first because I think it's what's breaking the build.

hsubox76 (commented):
LG other than the build issue (see previous comment).

packages/database/src/exp/DataSnapshot.ts
schmidt-sebastian: Thanks. Fixed this and the tests are passing. I can't come up with a better name for this file, otherwise I would have changed it long ago.

hsubox76 (approved):

PR #4688, opened 2021-03-30
[Auth] Fix an issue with the UI interop

samhorlbeck:
The Closure SDK makes sure the user belongs to the project by checking the API key on the user and the auth object. We don't store the API key on the user in the new SDK so we were checking this using the app name. But it's a perfectly valid use case to pass users between apps on the same project, so this check was wrong.

This PR also removes an unused rollup plugin that was breaking the compat demo build.

### DIFF
--- a/packages-exp/auth-compat-exp/demo/rollup.config.js
+++ b/packages-exp/auth-compat-exp/demo/rollup.config.js
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import commonjs from '@rollup/plugin-commonjs';
 import json from '@rollup/plugin-json';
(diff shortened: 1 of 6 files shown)


yuchenshi:
As of the latest persistence compatibility changes, we do store API keys on `user.toJSON()` so it is indeed possible to compare API keys. Shall we do that?

samhorlbeck:
Done per your suggestion. I've also fixed the `_clone()` method to (correctly) assign a new Auth instance

Feiyang1 (approved):

yuchenshi (approved):

PR #4696, opened 2021-03-31
Create typings for the interop API

Feiyang1:


scripts/exp/create-overloads.ts
```diff
@@ -0,0 +1,321 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as yargs from 'yargs';
+import * as ts from 'typescript';
+import * as fs from 'fs';
+
+/**
+ * TODO:
```
hsubox76, line 23: This seems to be done.

common/api-review/auth-exp.api.md
```diff
@@ -184,6 +185,16 @@ export const debugErrorMap: AuthErrorMap;
 // @public
 export function deleteUser(user: User): Promise<void>;
 
+// @public (undocumented)
```
hsubox76, line 188: Needs documentation now that it's publicly exposed?

scripts/exp/create-overloads.ts
```diff
@@ -0,0 +1,321 @@
...
+        ts.setTextRange(node, {
+          pos: node.getStart(sourceFile),
+          end: node.getEnd()
+        });
+
+        overloads.push(
+          ts.updateFunctionDeclaration(
+            node,
+            node.decorators,
+            [],
```
hsubox76, line 182: Deliberately removing any modifiers? What are modifiers? "public|private|protected"?

hsubox76 (approved):
LG, just a couple small things and a question for my benefit.

common/api-review/auth-exp.api.md
Feiyang1: Yes, we should do a sweep for all packages at some point and add any missing documentation. I will leave this PR as is for now.

scripts/exp/create-overloads.ts
Feiyang1: Yes, to remove `export` and `declare` which are not needed in the `declare module` block

PR #4710, opened 2021-04-01
Fix indexDB putObject concurrency issue and cache.

yuchenshi:
WANT_LGTM=all

### DIFF
--- a/packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
+++ b/packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
@@ -373,7 +373,7 @@ class IndexedDBLocalPersistence implements InternalPersistence {
       }
     }
     for (const localKey of Object.keys(this.localCache)) {
-      if (!keysInResult.has(localKey)) {
+      if (this.localCache[localKey] && !keysInResult.has(localKey)) {
         // Deleted
         this.notifyListeners(localKey, null);
         keys.push(localKey);
@@ -387,8 +387,11 @@ class IndexedDBLocalPersistence implements InternalPersistence {
     newValue: PersistenceValue | null
   ): void {
     this.localCache[key] = newValue;
-    for (const listener of Array.from(this.listeners[key])) {
-      listener(newValue);
+    const listeners = this.listeners[key];
+    if (listeners) {
+      for (const listener of Array.from(listeners)) {
+        listener(newValue);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages-exp/auth-exp/src/platform_browser/persistence/local_storage.ts
```diff
@@ -209,7 +206,6 @@ class BrowserLocalPersistence
   }
 
   _addListener(key: string, listener: StorageEventListener): void {
-    this.localCache[key] = this.storage.getItem(key);
```
avolkovi, line 212: i think you still need this, it warms up the cache from existing data when we register a listener. without this you could have a situation like:
- add value
- attach listener
- delete value
and the listener wouldn't fire

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
```diff
@@ -127,20 +127,11 @@ export async function _putObject(
   key: string,
   value: PersistenceValue | string
 ): Promise<void> {
-  const getRequest = getObjectStore(db, false).get(key);
-  const data = await new DBPromise<DBObject | null>(getRequest).toPromise();
-  if (data) {
```
avolkovi, line 132: i think you still need this branch, my understanding is it's required for updating vs creating objects

packages-exp/auth-exp/src/platform_browser/persistence/local_storage.ts
avolkovi: nvm, you'll have written to the cache regardless during the initial write

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
avolkovi: discussed IRL

avolkovi (approved):

packages-exp/auth-exp/src/platform_browser/persistence/indexed_db.ts
yuchenshi: `The put() method of the IDBObjectStore interface updates a given record in a database, or inserts a new record if the given item does not already exist.` -- [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/put)

packages-exp/auth-exp/src/platform_browser/persistence/local_storage.ts
yuchenshi: Cache is always updated on any data changes (add/put/remove) regardless of # listeners so there is no need for this step. It only complicates things.

samhorlbeck (approved):
Looks good pending tests passing

packages-exp/auth-exp/src/platform_browser/persistence/local_storage.ts
yuchenshi: Actually, some proactive populating helps reduce spuriously triggering and fixes some of the tests. I've added a conditioned clause back and to indexedDb to stay consistent.

PR #4719, opened 2021-04-03
Compat class for Reference

schmidt-sebastian:
One breaking change: The `onComplete` callback of a transaction is no longer invoked synchronously when transactions are aborted. I think this is such a subtle change that it won't break our users - they would have to rely on the timing of a write that takes place outside of the `runTransaction` function and the transaction callback.

OnDisconnect is still missing and is next.

### DIFF
--- a/packages/database/src/api/Reference.ts
+++ b/packages/database/src/api/Reference.ts
@@ -24,32 +24,11 @@ import {
   validateContextObject
 } from '@firebase/util';
 
-import {
-  repoServerTime,
-  repoSetWithPriority,
-  repoStartTransaction,
-  repoUpdate
-} from '../core/Repo';
-import { PRIORITY_INDEX } from '../core/snap/indexes/PriorityIndex';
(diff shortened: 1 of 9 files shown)


packages/database/src/api/Reference.ts
```diff
@@ -570,95 +558,61 @@ export class Reference extends Query implements Compat<ReferenceImpl> {
...
-    if (pathIsEmpty(this._delegate._path)) {
-      return null;
-    } else {
-      return pathGetBack(this._delegate._path);
-    }
+    return this._delegate.key;
   }
 
-  child(pathString: string | Path): Reference {
+  child(pathString: string): Reference {
```
Feiyang1, line 574: Is it a breaking change?

Feiyang1 (approved):

packages/database/src/api/Reference.ts
schmidt-sebastian: Path is an internal class, so it should be fine.

PR #4720, opened 2021-04-03
Database exp

schmidt-sebastian:
This merges all the PRs for database-exp-master.

All build/test failures seem unrelated but are a bit scary.


.changeset/lets-go-travel.md
```diff
@@ -0,0 +1,7 @@
+---
+"firebase": minor
```
Feiyang1: Why is it minor?

Feiyang1 (approved):

PR #4738, opened 2021-04-06
Add database@exp API docs

schmidt-sebastian:


packages/database/src/exp/OnDisconnect.ts
```diff
@@ -46,6 +76,12 @@ export class OnDisconnect {
     return deferred.promise;
   }
 
+  /**
+   * Ensures the data at this location is deleted when the client is disconnected
+   * (due to closing the browser, navigating to a new page, or network issues).
+   *
+   * @return Resolves when synchronization to the server is complete.
```
hsubox76, line 83: I think I remember API documenter/extractor insists on `@returns` instead of `@return` but if you don't get an error when you run it, never mind.

hsubox76 (approved):

packages/database/src/exp/OnDisconnect.ts
schmidt-sebastian: Fixed. There is no API report yet.

PR #4740, opened 2021-04-06
Fix platform logging for exp packages

hsubox76:
Update files as needed for platform logging to work with exp packages.

Part of this requires keeping track of the versions of firestore, storage, and database exp in their respective /exp/package.json files. These files aren't used in publishing the release so we need to add something to the release script to make sure they stay updated the same way as any other subpackage once we move out of alpha.


Feiyang1 (commented):
It doesn't seem you included the release script changes you mentioned. On second thought, I think we can continue to read the version from the root packages json for Firestore, Storage and RTDB because all exp packages have the same version number.

Feiyang1 (approved):

PR #4748, opened 2021-04-07
Fill documentation gaps in analytics and functions exp

hsubox76:
I decided not to document every property of analytics param types like Item, I'm sure Analytics users can find those in the general Google Analytics docs, and that documentation will be better maintained.


Feiyang1 (approved):
Thanks!

PR #4752, opened 2021-04-07
Add storage.useEmulator to exp global typings

hsubox76:
Added to firebase/index.d.ts but forgot to add to firebase-exp/index.d.ts. Also the signature was wrong.


Feiyang1 (commented):
Need a changeset?

Feiyang1 (approved):

PR #4753, opened 2021-04-08
Package database compat

Feiyang1:


packages/database/exp/index.node.ts
```diff
@@ -17,61 +17,6 @@
...
-  startAt,
-  update
-} from '../src/exp/Reference_impl';
-export { increment, serverTimestamp } from '../src/exp/ServerValue';
-export {
-  runTransaction,
-  TransactionOptions,
-  TransactionResult
-} from '../src/exp/Transaction';
+export * from './api';
```
schmidt-sebastian, line 20: Yay :)

packages/database/exp/api.ts
```diff
@@ -0,0 +1,77 @@
...
+  remove,
+  set,
+  setPriority,
+  setWithPriority,
+  startAfter,
+  startAt,
+  update,
+  ReferenceImpl as _ReferenceImpl,
+  QueryImpl as _QueryImpl,
+  child as _child
```
schmidt-sebastian, line 70: `child` should be part of the public API. We can just export it as child (see go/rtdb-next)

packages/database/compat/index.ts
```diff
@@ -0,0 +1,97 @@
...
+          ServerValue,
+          TEST_ACCESS
+        }
+      )
+      .setMultipleInstances(true)
+  );
+
+  instance.registerVersion(name, version);
+
+  if (isNodeSdk()) {
```
hsubox76, line 78: I was wondering why this is here if a Node app should be using index.node.ts instead anyway, but I guess this is copied code and has been here for 4 years so this probably isn't the time to worry about it.

packages/database/compat/index.node.ts
```diff
@@ -0,0 +1,153 @@
+/**
```
hsubox76, line 1: Seems like index.node.ts and index.ts can share some code but don't know if it's worth worrying about that at this time.

hsubox76 (approved):

packages/database/exp/api.ts
Feiyang1: Thanks. Updated.

packages/database/compat/index.ts
Feiyang1: Good point. Removed.

packages/database/compat/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "@firebase/database-compat",
+    "version": "0.0.900",
+    "description": "The Cloud Storage component of the Firebase JS SDK.",
```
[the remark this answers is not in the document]
Feiyang1: Changed.

PR #4760, opened 2021-04-08
Fix infinite recursion in FirebaseStorageError message getter

mitchellwills:
This was originally broken in the refactor in #3945

### DIFF
--- a/.changeset/long-forks-bow.md
+++ b/.changeset/long-forks-bow.md
@@ -1,0 +1,6 @@
+---
+'firebase': patch
+'@firebase/storage': patch
+---
+
+Fix infinite recursion in FirebaseStorageError message getter
--- a/packages/storage/src/implementation/error.ts
+++ b/packages/storage/src/implementation/error.ts
@@ -27,16 +27,15 @@ export class FirebaseStorageError extends FirebaseError {
    */
   customData: { serverResponse: string | null } = { serverResponse: null };
 
+  private readonly _baseMessage: string;
   /**
    * @param code - A StorageErrorCode string to be prefixed with 'storage/' and
    *  added to the end of the message.
    * @param message  - Error message.
    */
   constructor(code: StorageErrorCode, message: string) {
-    super(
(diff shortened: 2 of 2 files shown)


packages/storage/src/implementation/error.ts
```diff
@@ -27,16 +27,15 @@ export class FirebaseStorageError extends FirebaseError {
    */
   customData: { serverResponse: string | null } = { serverResponse: null };
 
+  private readonly baseMessage: string;
```
hsubox76, line 30: `_baseMessage`

.changeset/long-forks-bow.md
```diff
@@ -0,0 +1,6 @@
+---
+'firebase': patch
+'@firebase/storage': patch
+---
+
+Fix infinite recursion in FirebaseStorageError message getter
```
hsubox76, line 6: Is there an associated Github issue? Or an example of when the problem comes up in usage? So I can better explain to users in the release notes.

packages/storage/src/implementation/error.ts
mitchellwills: Done

.changeset/long-forks-bow.md
mitchellwills: No there isn't an existing issue (beyond this PR). Basically any time the `message` getter is accessed you would get a `Maximum call stack size exceeded` Error raised.

hsubox76 (approved):

hsubox76 (changes requested):
Sorry, I thought the test failures were flaky and missed the actual problem. `super()` needs to be called with the message or else when the error is thrown it doesn't actually throw with the message. I don't think the property can be overridden with a getter. I think we should keep `_baseMessage`, get rid of `get message()`, and add some code to `set serverResponse` to either set `this.message` to `_baseMessage + serverResponse` or `_baseMessage` depending on the arg.

mitchellwills:
Superseded by #4807

PR #4767, opened 2021-04-09
Add exports field with subpaths to firebase-exp/package.json

hsubox76:
Background for exports: https://nodejs.org/api/packages.html#packages_conditional_exports
It's also recognized by webpack and other bundlers so we can't assume it's only consumed by Node applications. Requested [here](https://github.com/firebase/firebase-js-sdk/discussions/4532).

Only adding exports to the main package.json for now. We will do the other packages later if there's a need.

### DIFF
--- a/packages-exp/analytics-compat/package.json
+++ b/packages-exp/analytics-compat/package.json
@@ -5,9 +5,8 @@
   "private": true,
   "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
   "main": "dist/index.cjs.js",
-  "browser": "dist/index.esm.js",
-  "module": "dist/index.esm.js",
-  "esm2017": "dist/index.esm2017.js",
+  "browser": "dist/index.esm2017.js",
+  "module": "dist/index.esm2017.js",
   "files": [
     "dist"
   ],
@@ -56,5 +55,14 @@
       ".ts"
     ],
     "reportDir": "./coverage/node"
-  }
+  },
+  "exports": {
+    "main": "./dist/index.cjs.js",
+    "esm5": "./dist/index.esm.js",
+    "browser": "./dist/index.esm2017.js",
+    "module": "./dist/index.esm2017.js",
+    "node": "./dist/index.cjs.js",
+    "default": "./dist/index.esm2017.js"
+  },
+  "esm5": "dist/index.esm.js"
 }
--- a/packages-exp/analytics-compat/package.json
+++ b/packages-exp/analytics-compat/package.json
@@ -5,8 +5,9 @@
   "private": true,
   "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
   "main": "dist/index.cjs.js",
-  "browser": "dist/index.esm2017.js",
-  "module": "dist/index.esm2017.js",
+  "browser": "dist/index.esm.js",
+  "module": "dist/index.esm.js",
+  "esm2017": "dist/index.esm2017.js",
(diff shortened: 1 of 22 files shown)


packages-exp/auth-compat-exp/package.json
```diff
@@ -5,9 +5,8 @@
   "description": "FirebaseAuth compatibility package that uses API style compatible with Firebase@7 and prior versions",
   "author": "Firebase <firebase-support@google.com> (https://firebase.google.com/)",
   "main": "dist/index.node.cjs.js",
-  "browser": "dist/index.esm.js",
-  "module": "dist/index.esm.js",
-  "esm2017": "dist/index.esm2017.js",
+  "browser": "dist/index.esm2017.js",
+  "module": "dist/index.esm2017.js",
```
Feiyang1, line 9: Wonder if we don't want to change compat packages, but I guess it's fine since it's going to be a major release

packages-exp/analytics-exp/package.json
```diff
@@ -63,5 +62,14 @@
...
-  }
-}
+  },
+  "exports": {
+    "main": "./dist/index.cjs.js",
+    "esm5": "./dist/index.esm.js",
+    "browser": "./dist/index.esm2017.js",
+    "module": "./dist/index.esm2017.js",
+    "node": "./dist/index.cjs.js",
+    "default": "./dist/index.esm2017.js"
```
Feiyang1, line 72: Should we assign a browser build to default? IIUC, `default` is used by Nodejs if nothing else matches, so should it be a node build? or should we just leave it since there is a `node` field?

scripts/exp/prepare-database-for-exp-release.ts
```diff
@@ -45,9 +45,25 @@ export async function prepare() {
...
 
   packageJson.main = expPackageJson.main.replace('../', '');
-  packageJson.module = expPackageJson.module.replace('../', '');
-  packageJson.browser = expPackageJson.browser.replace('../', '');
-  packageJson.esm2017 = expPackageJson.esm2017.replace('../', '');
+
+  // Set esm2017 as default.
+  // Delete esm2017 field.
+  // Add esm5 field.
+  packageJson.module = expPackageJson.esm2017.replace('../', '');
```
Feiyang1, line 52: Should we just update the package.json?

packages-exp/analytics-exp/package.json
hsubox76: The node examples always put a browser bundle in the default field. I think they assume any Node app will pick up what's in the Node field.

packages-exp/analytics-compat/package.json
```diff
@@ -56,5 +55,14 @@
...
-}
+  },
+  "exports": {
+    "main": "./dist/index.cjs.js",
+    "esm5": "./dist/index.esm.js",
+    "browser": "./dist/index.esm2017.js",
+    "module": "./dist/index.esm2017.js",
+    "node": "./dist/index.cjs.js",
+    "default": "./dist/index.esm2017.js"
+  },
```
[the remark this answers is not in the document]
hsubox76: I'm not sure, webpack seems to look at a lot of possible fields https://webpack.js.org/guides/package-exports and we don't distinguish between browser/module/react-native/etc in the minimal firebase-exp entry points so if you had an app that cared about the difference it would only find them in the individual `@firebase` packages entry points.

Feiyang1 (approved):

PR #4769, opened 2021-04-11
Adding comments for FirebaseOptions

egilmorez:
Adding comments for FirebaseOptions so that the reference documentation will be populated.

Thanks!


packages-exp/app-exp/src/public-types.ts
```diff
@@ -70,16 +70,48 @@ export interface FirebaseApp {
 /**
  * @public
  *
- * Firebase configuration object
+ * Firebase configuration object. Contains a set of parameters required by
+ * services in order to successfully communicate with Firebase server APIs
+ * and to associate client data with your Firebase project and
+ * Firebase application.
```
hsubox76, line 76: Should we say somewhere that they can get this entire config from the console? Or something so nobody makes the mistake of thinking they have to think about and populate each field individually.

packages-exp/app-exp/src/public-types.ts
```diff
@@ -70,16 +70,48 @@ export interface FirebaseApp {
...
+   * Unique numerical value used to identify each sender that can send
+   * Firebse Cloud Messaging messages to client apps.
+   */
   messagingSenderId?: string;
+  /**
+   * Unique identifier for the app.
+   */
   appId?: string;
+  /**
+   * An ID automatically created when you enable Analytics in your
```
hsubox76, line 112: Should we mention that measurementId is optional in v7.20.0 and up?

packages-exp/app-exp/src/public-types.ts
egilmorez: Good point, done.

packages-exp/app-exp/src/public-types.ts
egilmorez: Done.

hsubox76 (approved):

Feiyang1 (approved):

PR #4773, opened 2021-04-12
Allow FirstPartyAuth to specify a token factory func.

richieforeman:
This adds `authTokenFactory` to `FirstPartyCredentialsSettings`, which avoids a hard dependency on GAPI.JS when specified.


packages/firestore/src/api/credentials.ts
```diff
@@ -294,19 +297,39 @@ export class FirstPartyToken implements Token {
...
+      hardAssert(
+        !!(
+          typeof this.gapi === 'object' &&
+          this.gapi !== null &&
+          this.gapi['auth'] &&
+          this.gapi['auth']['getAuthHeaderValueForFirstParty']
+        ),
+        'unexpected gapi interface'
+      );
+      return this.gapi['auth']['getAuthHeaderValueForFirstParty']([]);
```
schmidt-sebastian, line 321: Could we pass this block in as the default value for authTokenFactory()? We can then just invoke this.authTokenFactory for all callsites.

schmidt-sebastian (approved):
LGTM, but one suggestion for a potential simplification

PR #4775, opened 2021-04-13
Fixing link format for vNext reference.

egilmorez:
Looks like this will work where markdown did not.

Thanks!


Feiyang1:
It seems you are reusing the branch that was merged, so there are some conflicts. Can you create a brand new branch and make a PR from there?

egilmorez:
Oh, sorry.  I find it so unintuitive when git status tells me that my new branch is clean, and then this happens.  Will do.

Issue #4780, opened 2021-04-13
UNKNOWN: 14 UNAVAILABLE: No connection established


nbransby:
This code assumes it will receive a FirestoreError without asserting such and can end up throwing an unwrapped exception from grpc

https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages/firestore/src/remote/datastore.ts#L126-L131

dconeybe:
Hi @nbransby. Thank you for reporting this. I'll get a fix out in the next couple of days.

PR #4788, opened 2021-04-14
Ensure errors are wrapped in FirestoreError in DatastoreImpl methods

dconeybe:
Fixes: #4780


packages/firestore/test/unit/remote/datastore.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  newDatastore,
+  invokeCommitRpc,
+  invokeBatchGetDocumentsRpc
+} from '../../../src/remote/datastore';
+import { JsonProtoSerializer } from '../../../src/remote/serializer';
+import { Code, FirestoreError } from '../../../src/util/error';
+
+use(chaiAsPromised);
+
+// TODO: Improve the coverage of these tests.
```
schmidt-sebastian, line 35: Needs a bug number or a user name (e.g. `wilhuff`)

packages/firestore/test/unit/remote/datastore.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      .and.have.property('code', Code.FAILED_PRECONDITION);
+  });
+
+  it('DatastoreImpl.invokeStreamingRPC() rethrows a FirestoreError', async () => {
+    const connection = new MockConnection();
+    connection.invokeStreamingRPC = () =>
+      Promise.reject(new FirestoreError(Code.ABORTED, 'zzyzx'));
+    const credentials = new MockCredentialsProvider();
+    const datastore = newDatastore(credentials, connection, serializer);
+    await expect(invokeDatastoreImplInvokeStreamingRPC(datastore))
```
schmidt-sebastian, line 169: This doesn't seem to verify that we return a FirestoreError. I think you can do this as such:
 
```
await expect(invokeDatastoreImplInvokeStreamingRPC(datastore)).to.be.eventually.rejectedWIth(new FirestoreError(Code.ABORTED, 'zzyzx'))
``` 

If that doesn't work, you can use a try/catch here.

This also applies to other tests such as"DatastoreImpl.invokeRPC() wraps unknown exceptions in a FirestoreError". Since GRPC errors also have codes, I suspect that these tests may have passed even without your change.

schmidt-sebastian (approved):

PR #4792, opened 2021-04-15
Add mockUserToken support for database emulator.

yuchenshi:
POC for new API with RTDB first.

### Open questions

With this implementation, it doesn't look too challenging to support changing the mock token through repeated calls of `useEmulator(host, port, {mockUserToken})` (provided that `host` and `port` stays the same). However, there are further implications for the underlying requests (e.g. pending requests or retries). I'd prefer we have a discussion before proceeding with a separate PR.

### Testing

Unit testing in `@firebase/util` done, manually tested end to end for `@firebase/database`

### API Changes

Internally approved. See go/firebase-rules-unit-testing-next-api (Googlers only)


packages/util/src/emulator.ts
```diff
@@ -0,0 +1,140 @@
...
+  [claim: string]: any;
+
+  uid?: never; // Try to catch a common mistake of "uid" (should be "sub" instead).
+}
+
+export function createMockUserToken(
+  token: Partial<FirebaseIdToken>,
+  projectId?: string
+): string {
+  if (token.uid) {
```
yuchenshi, line 97: Made this an error since we're technically in a new major version and we may as well introduce this breaking change.

packages/util/test/emulator.test.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+import { expect } from 'chai';
+import { base64 } from '../src/crypt';
+import { createMockUserToken, FirebaseIdToken } from '../src/emulator';
+
+// Firebase Auth tokens contain snake_case claims following the JWT standard / convention.
+/* eslint-disable camelcase */
+
+describe('createMockUserToken()', () => {
+  it('creates a well-formed JWT', () => {
```
yuchenshi, line 25: Migrated from an old rules unit testing library test

packages/util/src/emulator.ts
```diff
@@ -0,0 +1,140 @@
...
+      'Invalid Firebase token field "uid". Did you mean "sub" (for Firebase Auth User ID)?'
+    );
+  }
+  // Unsecured JWTs use "none" as the algorithm.
+  const header = {
+    alg: 'none',
+    type: 'JWT'
+  };
+
+  const project = projectId || 'fake-project';
```
samtstern, line 108: Let's make the default projectId something that starts with `demo-` to align with what we're planning for the emulators.

samtstern (approved):

packages/database/src/api/Database.ts
```diff
@@ -58,9 +58,16 @@ export class Database implements FirebaseService, Compat<ExpDatabase> {
    *
    * @param host - the emulator host (ex: localhost)
    * @param port - the emulator port (ex: 8080)
+   * @param options.mockUserToken - Optional: The mock token to use (for unit testing Security Rules)
```
schmidt-sebastian, line 61: No need to mention that this is optional since this should be apparent from the typings.

packages/database/src/exp/Database.ts
```diff
@@ -286,11 +291,15 @@ export function getDatabase(
  * @param db - The instance to modify.
  * @param host - The emulator host (ex: localhost)
  * @param port - The emulator port (ex: 8080)
+ * @param options.mockUserToken - Optional: The mock token to use (for unit testing Security Rules)
```
schmidt-sebastian, line 294: Same comment as above.

packages/database/src/exp/Database.ts
```diff
@@ -299,8 +308,26 @@ export function useDatabaseEmulator(
       'Cannot call useEmulator() after instance has already been initialized.'
     );
   }
+
+  const repo = db._repo;
+  let tokenProvider: EmulatorTokenProvider | undefined = undefined;
+  if (repo.repoInfo_.nodeAdmin) {
+    if (options.mockUserToken) {
+      fatal(
+        'mockUserToken is not supported on the Admin SDK. For client access with mock users, please use the "firebase" package instead of "firebase-admin".'
```
schmidt-sebastian, line 317:
```suggestion
        'mockUserToken is not supported by the Admin SDK. For client access with mock users, please use the "firebase" package instead of "firebase-admin".'
```

packages/util/test/emulator.test.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
schmidt-sebastian, line 3:
```suggestion
 * Copyright 2021 Google LLC
```

schmidt-sebastian (approved):

packages/util/src/emulator.ts
yuchenshi: Done.

packages/util/src/emulator.ts
samtstern: I think we used `uid` in all of the docs we have ever published about this 😢 
https://firebase.google.com/docs/rules/unit-tests

samtstern (approved):

packages/util/src/emulator.ts
yuchenshi: As discussed offline, I've updated the error message to provide more context. Resolving.

packages/util/src/emulator.ts
```diff
@@ -0,0 +1,143 @@
...
+    sign_in_provider: FirebaseSignInProvider;
+
+    // A map of providers to the user's list of unique identifiers from
+    // each provider
+    identities?: { [provider in FirebaseSignInProvider]?: string[] };
+  };
+
+  // Custom claims set by the developer
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [claim: string]: any;
```
Feiyang1, line 88: Can you use `unknown` ?

packages/util/src/emulator.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  uid?: never; // Try to catch a common mistake of "uid" (should be "sub" instead).
+}
+
+export type EmulatorMockTokenOptions = ({ user_id: string } | { sub: string }) &
+  Partial<FirebaseIdToken>;
+
+export function createMockUserToken(
```
Feiyang1, line 96: Is it expected to be used by developers? `@firebase/util` is currently an internal package, and we ask developer to not use it directly.

yuchenshi: No, this is a share utility used by `database` (and `firestore` #4837 and `storage` later). Developers should only use the `useXXXEmulator` public API.

packages/util/src/emulator.ts
yuchenshi: Done.

Feiyang1 (approved):

PR #4796, opened 2021-04-15
`get()` server data must be indexed

IanWyszynski:
If we're getting data from the server, it should be indexed with the query param index.

For the value event registration path, I think the query index is applied using `IndexFilter::updateFullNode` on the filter attached to the view that the listener sets up front. 

Experimentally, even when the query results are already cached (even if they don't have the same path index), as in:
```
    const things = db.ref().child("things");

    await things.remove();

    things.push().set({ "thing": 1 });
    things.push().set({ "thing": 2 });


    await new Promise((resolve, _) => {
        things.on("value", function (results) {
            resolve();
        })
    });

    things.orderByChild("thing").get().then(function (results) {
        results.forEach(function (node) {
            console.log(JSON.stringify(node));
        });
    }).catch(function (err) {
        console.log("There's an error: " + err);
    })
```
`get()` will add the new path index to the `CacheNode`'s index map. It's only in the case where data comes in from the server for the first time that we'd see this issue.

Side note: Looking at `repoOnDataUpdate` I also noticed that we should probably be calling `repoRerunTransactions` when server data comes in for the first time (though I'm not certain here).


packages/database/src/core/Repo.ts
```diff
@@ -458,13 +458,19 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
+      );
       const events = syncTreeApplyServerOverwrite(
         repo.serverSyncTree_,
         query._path,
         node
       );
-      eventQueueRaiseEventsAtPath(repo.eventQueue_, query._path, events);
+      let affectedPath = query._path;
+      if (events.length > 0) {
+        affectedPath = repoRerunTransactions(repo, query._path);
```
schmidt-sebastian, line 471: I think we already call this:

https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/core/PersistentConnection.ts#L204

which invokes
https://github.com/Firebase/firebase-js-sdk/blob/master/packages/database/src/core/Repo.ts#L396

packages/database/src/core/Repo.ts
```diff
@@ -458,13 +458,19 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
   }
   return repo.server_.get(query).then(
     payload => {
-      const node = nodeFromJSON(payload as string);
+      const node = nodeFromJSON(payload as string).withIndex(
```
schmidt-sebastian, line 461: Can we add a test for this? I assume this would be mostly a copy&paste from the code snippet.

schmidt-sebastian (commented):
Please also add a changelog

packages/database/src/core/Repo.ts
IanWyszynski: Discussed offline. This test requires adding an `.indexOn` to the rules and fitting this into the existing query test suite is sort of awkward rn.

.changeset/rotten-wombats-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/database': patch
+---
+
+Index server data retrieved by get()
```
schmidt-sebastian, line 5: Can you try to word this in a way that is more obvious to our customers?

schmidt-sebastian (approved):

Issue #4812, opened 2021-04-21
onSnapshot() results in `unknown` snapshot data type


samtstern:
### [REQUIRED] Describe your environment

  * Operating System version: macOS
  * Browser version: Chrome latest
  * Firebase SDK version: 9.0.0-beta.1
  * Firebase Product: firestore

#### Relevant Code:

**index.ts**
```ts
import { collection, getFirestore, onSnapshot } from "firebase/firestore";

function main() {
  const firestore = getFirestore();
  const colRef = "col";

  onSnapshot(
    collection(firestore, colRef),
    (snap) => {
      if (snap.empty) {
        return;
      }
      const documents = snap.docs.map((doc) => ({
        id: doc.id,
        ...doc.data(), // error: Spread types may only be created from object types
      }));
    },
    (error) => {
      console.warn(error);
    }
  );
}
```

In this code `snap` is `QuerySnapshot<unknown>` which makes it hard to use `.data()`.  This can be fixed by either:

  * Casting `doc.data() as any`
  * Using `onSnapshot<any>`

Is this the intended behavior in the new SDK?

samtstern:
cc @puchesjr

schmidt-sebastian:
TypeScript does not retain the type of the generic class unless the type is part of the API surface. Unfortunately, `CollectionReference<T>` does not refer to `T` in any of its members. This means that we have to change the API to accommodate this. I will prepare a PR, but we need to obtain approval for this API change internally.

PR #4817, opened 2021-04-21
Make Converter explicit

schmidt-sebastian:
Fixes: https://github.com/firebase/firebase-js-sdk/issues/4812

The current API does not retain the type of CollectionReference<T> and Query<T> since T is not referenced within these classes.

We also want to expose a type property on FirebaseFirestore. This is similar to the type property of FirebaseDatabase@exp  and helps TypeScript tell apart instances of Firestore, FirestorLite and RTDB.


thebrianchen (approved):
optional: include the changeset file.

schmidt-sebastian:
We don't yet use Changesets for firebase@exp. Thanks for the review

Issue #4811, opened 2021-04-21
Firebase warning connecting to database emulator


Joebayld:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 11
  * Browser version: Chrome 89
  * Firebase SDK version: 8.4.1
  * Firebase Product: database emulator

### [REQUIRED] Describe the problem
I was on firebase v8.2 and just made some updates to the latest v8.4.1, but now I can no longer connect my app to the local database emulator. All other emulators work - just not the database one. 

I can confirm the emulator is working as I can access it from the dashboard, and if I run version 8.2 it's also fine. When I go to my site, I get the following warning: 

`@firebase/database: FIREBASE WARNING: Provided authentication credentials for the app named "[DEFAULT]" are invalid. This usually indicates your app was not initialized correctly. Make sure the "apiKey" and "databaseURL" properties provided to initializeApp() match the values provided for your app at https://console.firebase.google.com/. `

Note that I've changing nothing other than the firebase version. 

#### Steps to reproduce:
```typescript 
import firebase from 'firebase/app'
import 'firebase/functions'
import 'firebase/firestore'
import 'firebase/analytics'
import 'firebase/auth'
import 'firebase/storage'
import 'firebase/database'
import { isDevelopment, isStaging } from '@/utility/env'

firebase.initializeApp(config) // hiding my config from here

const firebaseAuth = firebase.auth()
const firestore = firebase.firestore()
const functions = firebase.functions()
const storage = firebase.storage()
const analytics = firebase.analytics()
const database = firebase.database()

if (isDevelopment && !isStaging) {
  localStorage.clear()
  functions.useEmulator(window.location.hostname, 5001)
  // @ts-ignore // disableWarnings flag is still broken in typescript :(
  firebaseAuth.useEmulator('http://' + window.location.hostname + ':9099', { disableWarnings: true })
  firestore.useEmulator(window.location.hostname, 8079)
  database.useEmulator(window.location.hostname, 9000)
}
```

Any ideas?

looptheloop88:
Thanks for the report @Joebayld, however I wasn't able to replicate this issue. I used the SDK version 8.4.1 and I was able to use the emulator for Realtime Database without any issue.

Try to delete the `node_modules` directory and `package-lock.json` file, then run `npm install` again. If the issue persists, please share a sample project that I can run locally to replicate the issue.

Hubiwan:
[...]

PR #4828, opened 2021-04-23
Fix db.useEmulator

schmidt-sebastian:
We need to invoke repoManagerApplyEmulatorSettings with db._repoInternal as db._repo starts the WebSocket connection if it hasn't yet been started and applies the production host.

Also random unrelated cleanup.

Fixes https://github.com/firebase/firebase-js-sdk/issues/4811

### DIFF
--- a/.changeset/sweet-monkeys-warn.md
+++ b/.changeset/sweet-monkeys-warn.md
@@ -1,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):

yuchenshi:
Confirmed locally that this fixes a race condition

jsdt (approved):

PR #4829, opened 2021-04-23
Add `isSupported` for `messaging-compat`

zwu52:
Sticky situation for providing legacy `isSupported`  method with the modular SDK construct. Please let me know if there is a better way to just include and export the legacy API code in the `messaging-compat`. The rational for doing so is as followed: 

- Modular `isSupported` became async. It's not compatible w/ the legacy sync version. 
- Modular SDK has two entry points `window` and `sw`, but export `isWindowSupported` and `isSwSupported` as `isSupported` depends on the context. Not sure if there is a way to import `isWindowSupported` and `isSwSupported` separately from `@firebase/messaging-exp` to `messaging-compat`.

### DIFF
--- a/packages-exp/messaging-compat/src/messaging-compat.ts
+++ b/packages-exp/messaging-compat/src/messaging-compat.ts
@@ -24,8 +24,8 @@ import {
   MessagePayload,
   deleteToken,
   getToken,
-  onMessage,
-  onBackgroundMessage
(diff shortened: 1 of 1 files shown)


hsubox76 (approved):
I think this seems reasonable since we can't introduce breaking changes in compat.

PR #4841, opened 2021-04-28
Add Node debug token support

hsubox76:
Add ability for App Check to use debug token in Node by using `globalThis`.

In `app-check/src/util.ts`, `self` is also used in `self.grecaptcha` and I left that alone because Node can't use `grecaptcha` anyway.

See question about headers.

### DIFF
--- a/packages/database/src/realtime/WebSocketConnection.ts
+++ b/packages/database/src/realtime/WebSocketConnection.ts
@@ -166,10 +166,16 @@ export class WebSocketConnection implements Transport {
           }
         };
 
-        // Adding a comment to ask questions
+        // If using Node with admin creds, AppCheck-related checks are unnecessary.
+        // It will send the authorization token.
         if (this.nodeAdmin) {
           options.headers['Authorization'] = this.authToken || '';
         } else {
+          // If using Node without admin creds (which includes all uses of the
+          // client-side Node SDK), it will send an AppCheck token if available.
+          // Any other auth credentials will eventually be sent after the connection
+          // is established, but aren't needed here as they don't effect the initial
+          // request to establish a connection.
           options.headers['X-Firebase-AppCheck'] = this.appCheckToken || '';
         }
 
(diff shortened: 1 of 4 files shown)


packages/database/src/realtime/WebSocketConnection.ts
```diff
@@ -166,6 +166,7 @@ export class WebSocketConnection implements Transport {
           }
         };
 
+        // Adding a comment to ask questions
```
hsubox76, line 169: Added this comment to place a discussion about this section. The next 5 lines were added in https://github.com/FirebasePrivate/firebase-js-sdk/pull/239/files#diff-ddbaa9e762a6010e5a0aa0c8048d48deea15594f14caa3d51e68b283f36e489fR171

I was wondering if I could get some more info about why only the Authorization header is sent if it's admin node, and only the AppCheck header is sent if it's client node? @schmidt-sebastian ?

To help answer the questions raised in the discussion started here: https://github.com/firebase/firebase-js-sdk/pull/4833#discussion_r620817682

jsdt: On node, we want to send the `Authorization` header if the user has admin credentials. I probably wrote this without knowing that there is a node client SDK. Assuming the node client SDK can use admin creds, this should be checking if the `authToken` looks like an admin credential.

For more context, if a client has admin credentials, we skip any appcheck-related checks on the server side, which is why we send admin creds in the header (and no appcheck token). If the client has any other type of credentials, they will eventually be sent after the connection is established, but they won't effect whether the initial request to establish a connection is allowed (only the appcheck header matters there).

hsubox76: Thanks. As far as I know, the Node client SDK should never have admin creds so looks like we'll always want to send the AppCheck token.

(Which would probably only be the dummy token or the debug token because Node can't do reCAPTCHA. Or possibly a token from a custom provider.)

Feiyang1 (approved):
LGTM. Can we add what is discussed here to the comment?

hsubox76:
> LGTM. Can we add what is discussed here to the comment?

Yes, added.

PR #4842, opened 2021-04-28
Log correct version for v9 Firestore and RTDB

Feiyang1:


packages/firestore/lite/register.ts
```diff
@@ -29,6 +34,7 @@ declare module '@firebase/component' {
 }
 
 export function registerFirestore(): void {
+  setSDKVersion(SDK_VERSION);
```
schmidt-sebastian, line 37: Can we differentiate Firestore lite from Firestore? I wonder if we should prefix or postfix this with `lite`.

schmidt-sebastian (approved):
Please fix lint.

packages/firestore/lite/register.ts
Feiyang1: like `9.0.0-beta.1_lite`?

Feiyang1:
I swear I ran lint locally and everything passed!

schmidt-sebastian (approved):

packages/firestore/lite/register.ts
hsubox76: Don't we normally put variant info in the product string and not in the version number string? It's already in the product string.

hsubox76: The product string is in line 55, I think that should be enough to distinguish for platform logging, unless this is for Firestore-only backend logging? I think it might make platform logging queries a little harder to have one version not like the others.

Feiyang1: This version is logged to Firestore backend directly. We will have to do some data massage to join Firestore data with platform logging data.

hsubox76 (approved):

PR #4845, opened 2021-04-28
Defer check whether auth is available

schmidt-sebastian:
Adds support for calling `getFirestore` before calling `getAuth`.  

Tested manually (our integration tests do not use auth).

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -23,8 +23,10 @@ import { Provider } from '@firebase/component';
 
 import { User } from '../auth/user';
 import { hardAssert, debugAssert } from '../util/assert';
+import { AsyncQueue } from '../util/async_queue';
 import { Code, FirestoreError } from '../util/error';
 import { logDebug } from '../util/log';
+import { Deferred } from '../util/promise';
 
 // TODO(mikelehen): This should be split into multiple files and probably
 // moved to an auth/ folder to match other platforms.
@@ -78,7 +80,7 @@ export class OAuthToken implements Token {
  * token and may need to invalidate other state if the current user has also
  * changed.
  */
-export type CredentialChangeListener = (user: User) => void;
+export type CredentialChangeListener = (user: User) => Promise<void>;
 
 /**
  * Provides methods for getting the uid and token for the current user and
@@ -98,8 +100,13 @@ export interface CredentialsProvider {
    * Specifies a listener to be notified of credential changes
    * (sign-in / sign-out, token changes). It is immediately called once with the
    * initial user.
+   *
+   * The change listener is invoked on the provided AsyncQueue.
    */
-  setChangeListener(changeListener: CredentialChangeListener): void;
+  setChangeListener(
+    asyncQueue: AsyncQueue,
+    changeListener: CredentialChangeListener
+  ): void;
 
   /** Removes the previously-set change listener. */
   removeChangeListener(): void;
@@ -120,14 +127,20 @@ export class EmptyCredentialsProvider implements CredentialsProvider {
 
   invalidateToken(): void {}
 
-  setChangeListener(changeListener: CredentialChangeListener): void {
+  setChangeListener(
+    asyncQueue: AsyncQueue,
+    changeListener: CredentialChangeListener
+  ): void {
     debugAssert(
       !this.changeListener,
       'Can only call setChangeListener() once.'
     );
     this.changeListener = changeListener;
     // Fire with initial user.
-    changeListener(User.UNAUTHENTICATED);
+    asyncQueue.enqueueRetryable(() => {
+      changeListener(User.FIRST_PARTY);
+      return Promise.resolve();
+    });
   }
 
   removeChangeListener(): void {
@@ -175,11 +188,25 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
    * The auth token listener registered with FirebaseApp, retained here so we
    * can unregister it.
    */
-  private tokenListener: ((token: string | null) => void) | null = null;
+  private tokenListener: (() => void) | null = null;
 
   /** Tracks the current User. */
   private currentUser: User = User.UNAUTHENTICATED;
-  private receivedInitialUser: boolean = false;
+
+  /**
+   * Promise that allows blocking on the next `tokenChange` event. The Promise
+   * is re-assgined in `awaitTokenAndRaiseInitialEvent()` to allow blocking on
+   * an a lazily loaded Auth instance. In this case, `this.receivedUser`
+   * resolves once when the SDK first detects that there is no synchronous
+   * Auth, and then gets re-created and resolves again once Auth is loaded.
+   */
+  private receivedUser = new Deferred();
+
+  /**
+   * Whether the initial token event has been raised. This can go back to
+   * `false` if Firestore first starts without Auth and Auth is loaded later.
+   */
+  private initialEventRaised: boolean = false;
 
   /**
    * Counter used to detect if the token changed while a getToken request was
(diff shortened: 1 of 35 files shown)


Feiyang1 (commented):
I got an internal assertion failure `unexpected state` if auth is never initialized, or initialized too late. And requests seem to be queued indefinitely until Auth is initialized.

schmidt-sebastian:
The assertion should be gone as of 1e75d78

Feiyang1:
Thanks. The error went away. Now I found a use case where Firestore still doesn't wait for auth initialization:
```
const firestore = getFirestore(firebaseApp);
setLogLevel("debug");

// this getDoc() is important for repro. If we don't make a getDoc here, the getDoc in setTimeout will succeed.
getDoc(doc(firestore, 'coll/New York')).then(v => {
    console.log('sync firestore get', v.data());
});


setTimeout(() => {
    // assume there is a valid user
    getAuth(); 
    
    // getDoc fails with permission denied, but it should succeed since it's called after getAuth().
    getDoc(doc(firestore, 'coll/New York')).then(v => {
        console.log('Async firestore get', v.data());
    });
}, 500);
```

For repro, you can clone https://github.com/Feiyang1/playground/tree/fs-auth-init and use the project in the `rollup` directory.

You mentioned we may need to queue up work that uses the user, or even the auth, is that what we need to do to make it work?

schmidt-sebastian:
I think this will be solved by your "callback" PR that replaces the Promise in the initialization. Do you still have a link to that? I can apply this here or we can merge that PR before we test here further.

Feiyang1:
Yes! https://github.com/firebase/firebase-js-sdk/commits/fei-firestore-wait

You may need to resolve some merge conflicts. Thanks!

Feiyang1:
I took the liberty to merge the component change into your PR, but It still doesn't work.

schmidt-sebastian:
> I took the liberty to merge the component change into your PR, but It still doesn't work.

I will take over from here :)

schmidt-sebastian:
Updated. Tested with no getDoc initially, with getDoc initially and with no getAuth.

schmidt-sebastian:
This is now completely rewritten. The SDK now blocks on the availability of the use in the Credential Provider itself, which allows the SDK to block twice: Once when we check whether Auth is available at startup time and then once again if Auth is loaded later. We also synchronously invoke the user change handler (right as we release the block), which ensures that the user change gets executed before everything else.

packages/firestore/src/api/credentials.ts
```diff
@@ -188,44 +215,62 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
       this.currentUser = this.getUser();
-      this.receivedInitialUser = true;
-      if (this.changeListener) {
-        this.changeListener(this.currentUser);
+      if (this.initialEventRaised && this.asyncQueue) {
+        // We only invoke the change listener here if the initial event has been
+        // raised. The initial event itself is invoked synchronously in
+        // `awaitTokenAndRaiseInitialEvent()`.
+        this.asyncQueue.enqueueRetryable(() => {
+          this.changeListener(this.currentUser);
```
Feiyang1, line 236: In the `registerAuth()`, we do 
```
        this.awaitTokenAndRaiseInitialEvent();
        this.auth.addAuthTokenListener(this.tokenListener);
```

We already invoke `this.changeListener` in the asyncQueue in `awaitTokenAndRaiseInitialEvent`, should we not do it again in `this.tokenListener`?

packages/firestore/src/api/credentials.ts
```diff
@@ -175,11 +188,25 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
+   * resolves once when the SDK first detects that there is no synchronous
+   * Auth, and then gets re-created and resolves again once Auth is loaded.
+   */
+  private receivedUser = new Deferred();
+
+  /**
+   * Whether the initial token event has been raised. This can go back to
+   * `false` if Firestore first starts without Auth and Auth is loaded later.
+   */
+  private initialEventRaised: boolean = false;
```
Feiyang1, line 209: This variable doesn't seem to do anything, because it's only set in `awaitTokenAndRaiseInitialEvent` which is synchronous. It's not used in `awaitTokenAndRaiseInitialEvent()` itself and is always set to true at the end of the function, so it seems the outside world will always see this variable as `true`.

Feiyang1 (commented):
I think I understand most parts, but left a couple of questions.

I think we also need to do something similar in Database.

packages/firestore/src/api/credentials.ts
```diff
@@ -306,6 +336,33 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
+   * This function also invokes the change listener synchronously once a token
+   * is available.
+   */
+  private awaitTokenAndRaiseInitialEvent(): void {
+    this.invokeChangeListener = false; // Prevent double-firing of the listener
+    if (this.asyncQueue) {
+      // Create a new deferred Promise that gets resolved when we receive the
+      // next token. Ensure that all previous Promises also get resolved.
+      const awaitToken = new Deferred<void>();
+      void awaitToken.promise.then(() => awaitToken.resolve());
```
Feiyang1, line 356: `then` only fires if `awaitToken` resolves, why do we resolve it again in `then`? The comment suggests it ensures all previous Promises also get resolved? But I don't get how that works.

packages/firestore/src/api/credentials.ts
```diff
@@ -306,6 +336,33 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
   }
+
+  /**
+   * Blocks the AsyncQueue until the next user is available. This is invoked
+   * on SDK start to wait for the first user token (or `null` if Auth is not yet
+   * loaded). If Auth is loaded after Firestore,
+   * `awaitTokenAndRaiseInitialEvent()` is also used to block Firestore until
+   * Auth is fully initialized.
+   *
+   * This function also invokes the change listener synchronously once a token
```
Feiyang1, line 347: Do you mean asynchronously, since it happens in the asyncQueue?

packages/firestore/src/api/credentials.ts
```diff
@@ -188,44 +210,57 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
+          // token
+          logDebug('FirebaseCredentialsProvider', 'Auth not yet detected');
+          this.tokenListener();
         }
-      );
-    }
+      }
+    }, 0);
+
+    this.awaitTokenAndRaiseInitialEvent();
```
Feiyang1, line 263: Is `this.asyncQueue` set at this point? 

My understanding is it is not, so `awaitTokenAndRaiseInitialEvent` doesn't do anything except for setting `this.invokeChangeListener = false;` which will lead to wrong behavior if auth is not available - the following code in  this.tokenListener` will not execute:
```
      if (this.invokeChangeListener) {
        this.asyncQueue!.enqueueRetryable(() =>
          this.changeListener(this.currentUser)
        );
      }
```

Feiyang1: This seems to work in code though. Firestore will not receive a null user when auth is not initialized, and nothing seems to break.

schmidt-sebastian: I changed the logic for `invokeChangeListener` to only set it to true if the AsyncQueue is available. This should address this concern.

packages/firestore/src/api/credentials.ts
schmidt-sebastian: I meant "synchronously with the code that resolves the token". I have reworded it.

packages/firestore/src/api/credentials.ts
schmidt-sebastian: This was meant to resolve the existing this.receivedUser Promise if another one was created later on. I changed how the listener gets called if auth is not used, and am no longer using the Promise twice. This allowed me to remove this logic.

schmidt-sebastian:
This should be ready for review again.

packages/firestore/src/api/credentials.ts
```diff
@@ -188,44 +203,55 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
...
+    // Our users can initialize Auth right after Firestore, so we give it
+    // a chance to register itself with the component framework before we
+    // determine whether to start up in unauthenticated mode.
+    setTimeout(() => {
+      if (!this.auth) {
+        const auth = authProvider.getImmediate({ optional: true });
+        if (auth) {
+          registerAuth(auth);
+        } else if (this.invokeChangeListener) {
+          // If auth is still not available, invoke tokenListener once with null
```
Feiyang1, line 246: Comment needs update.

Feiyang1 (approved):

Feiyang1:
Tested the following manually, and they all worked. Yay!
- `initializeAuth` before `getFirestore`
- `initializeAuth` after `getFirestore` synchronously
- `initializeAuth` after `getFirestore` asynchronously
- `initializeAuth` never called

packages/firestore/src/api/credentials.ts
schmidt-sebastian: Fixed

PR #4847, opened 2021-04-28
Implement App Check auto refresh timing and opt out flag

hsubox76:
- Implement global opt-out. API proposal is here: https://docs.google.com/document/d/17RZDbKMRtgHoRGcyAR_KIAHPStr2aFnJmFsffO50pRI/edit?hl=en&forcehl=1#

- Implement auto-refresh timing. Formula is `issuedAtTime + (50% * TTL) + 5 minutes` from here: https://docs.google.com/document/d/1fi4xUsARCs3Uc8c46KvNiIf2rQeWnjxvODW-h0coSmo/edit?resourcekey=0-IK3tvAt8BxEmzoOHbbRfcg&disco=AAAAIUhr8kE

The second required an additional `AppCheckTokenInternal` type to be created that extends `AppCheckToken` with a `issuedAtTimeMillis` field, which is needed to calculate the auto refresh time. This info can also be extracted from the token, but I am following Android's example of adding this data to an internal token type using the local client's clock to avoid clock skew (the issued-at-time in the token is server time).

### DIFF
--- a/packages/app-check/src/internal-api.ts
+++ b/packages/app-check/src/internal-api.ts
@@ -264,11 +264,18 @@ function createTokenRefresher(
 
       if (state.token) {
         // issuedAtTime + (50% * total TTL) + 5 minutes
-        const nextRefreshTimeMillis =
+        let nextRefreshTimeMillis =
           state.token.issuedAtTimeMillis +
           (state.token.expireTimeMillis - state.token.issuedAtTimeMillis) *
             0.5 +
           5 * 60 * 1000;
+        // Do not allow refresh time to be past (expireTime - 5 minutes)
+        const latestAllowableRefresh =
+          state.token.expireTimeMillis - 5 * 60 * 1000;
+        nextRefreshTimeMillis = Math.min(
+          nextRefreshTimeMillis,
+          latestAllowableRefresh
+        );
         return Math.max(0, nextRefreshTimeMillis - Date.now());
       } else {
         return 0;
--- a/packages/app-check/src/internal-api.ts
+++ b/packages/app-check/src/internal-api.ts
@@ -183,7 +183,12 @@ export function addTokenListener(
       newState.tokenRefresher = tokenRefresher;
     }
 
-    if (!newState.tokenRefresher.isRunning()) {
+    // Create the refresher but don't start it if `isTokenAutoRefreshEnabled`
+    // is not true.
+    if (
+      !newState.tokenRefresher.isRunning() &&
+      state.isTokenAutoRefreshEnabled === true
+    ) {
       newState.tokenRefresher.start();
     }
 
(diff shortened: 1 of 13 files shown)


packages/app-check/src/internal-api.ts
```diff
@@ -108,7 +112,8 @@ export async function getToken(
    */
   try {
     if (state.customProvider) {
-      token = await state.customProvider.getToken();
+      const customToken = await state.customProvider.getToken();
+      token = { ...customToken, issuedAtTimeMillis: Date.now() };
```
Feiyang1, line 116: the customToken might be issued earlier than the current timestamp. It might be okay to assume it to be the current timestamp for the purpose of determining the refresh time, but is it the case? We should add comments to clarify.

packages/app-check/src/internal-api.ts
```diff
@@ -245,12 +255,13 @@ function createTokenRefresher(app: FirebaseApp): Refresher {
...
-            Date.now() -
-            TOKEN_REFRESH_TIME.OFFSET_DURATION
-        );
+        // issuedAtTime + (50% * total TTL) + 5 minutes
+        const nextRefreshTimeMillis =
+          state.token.issuedAtTimeMillis +
+          (state.token.expireTimeMillis - state.token.issuedAtTimeMillis) *
+            0.5 +
+          5 * 60 * 1000;
+        return Math.max(0, nextRefreshTimeMillis - Date.now());
```
Feiyang1, line 264: This could put the next refresh time after the token has become expired if `state.token.expireTimeMillis - state.token.issuedAtTimeMillis < 5min`, which is wrong.

We should use whatever is smaller between `issuedAtTime + (50% * total TTL) + 5 minutes` and `expireTime - 5mins`

packages/app-check/src/internal-api.ts
hsubox76: I might be misunderstanding but I think that if it was issued earlier, it would be in the cache and this function would have returned in line 105? Once it's fetched here it'll be written to storage in line 144. Unless the custom `getToken()` method has its own code to read/write from its own separate cache, where someone might have wiped Firebase AppCheck's cache but didn't wipe out the user's custom cache??, which, should we worry about that?

packages/app-check/src/internal-api.ts
hsubox76: Done.

packages/app-check/src/internal-api.ts
Feiyang1: It could happen when `customProvider.getToken()` executes for the first time before cache comes into play. Since `customProvider` is implemented by developers, it can return a token created 3 hours ago or 2 days ago.

Feiyang1 (approved):

PR #4857, opened 2021-04-30
Enable platform logging for AppCheck

hsubox76:
AppCheck is ready for beta, enable platform logging.

### DIFF
--- a/packages/app-check/src/index.ts
+++ b/packages/app-check/src/index.ts
@@ -50,7 +50,8 @@ function registerAppCheck(firebase: _FirebaseNamespace): void {
       container => {
         // getImmediate for FirebaseApp will always succeed
         const app = container.getProvider('app').getImmediate();
-        return internalFactory(app);
+        const platformLoggerProvider = container.getProvider('platform-logger');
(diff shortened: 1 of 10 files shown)


packages/app-check/src/index.ts
```diff
@@ -55,7 +56,7 @@ function registerAppCheck(firebase: _FirebaseNamespace): void {
     )
   );
 
-  // TODO: register AppCheck version with firebase.registerVersion() before BETA. We don't want to report version in EAP
+  firebase.registerVersion(name, version);
```
Feiyang1, line 59: importing json file changes output folder structure of typings file. can you please double check that the typing field and other fields are pointing to valid files?

Feiyang1 (approved):

packages/app-check/src/index.ts
hsubox76: The directory structure didn't change because `client.ts` was already importing package.json but the typings path was wrong anyway (maybe became wrong when `client.ts` originally was modified to import package.json) so I fixed it.

Feiyang1:
LG. Needs a changeset for `@firebase/app`

PR #4861, opened 2021-05-04
Firestore lite overview

egilmorez:
Adding a simple overview to explain what Firestore Lite is.  

Later if we have richer/more detailed information in the guides, we can link to that as well.


Feiyang1 (approved):

packages/firestore/lite/index.ts
```diff
@@ -1,3 +1,13 @@
+/**
+ * Firestore Lite
+ *
+ * @remarks This package includes the same feature set as Firestore, but
+ * without latency compensation or snapshot listeners (and with
```
schmidt-sebastian, line 5: Do users know what these features are? What about something like this:

Firestore Lite is a small online-only SDK that allows read and write access to your Firestore database. All operations talk directly to the backend and `onSnapshot()` APIs are not supported.

schmidt-sebastian (approved):
LGTM with rewrite suggestion :)

packages/firestore/lite/index.ts
schmidt-sebastian: I can certainly take this to completion. Do you have rewording suggestions or should we take my suggestion as is? I would be disappointed if you didn't have an improvement in the back of your head :)

egilmorez: Hehe, all I got is a comma and a substitution for "talk to," which the style guide says to avoid.  How about:

"Firestore Lite is a small online-only SDK that allows read and write access to your Firestore database. All operations connect directly to the backend, and `onSnapshot()` APIs are not supported."

I tried re-running the CI jobs on Feiyang's advice, but it looks like I still have several tests failing.  If you could take this Sebastian, that would be awesome.

schmidt-sebastian: Committed directly to this PR. Fingers crossed.

markarndt (approved):

PR #4864, opened 2021-05-04
Database once() should only fire once

schmidt-sebastian:
The way `once()` was implemented led to some level of recursion if the user updated the value inside the `once` callback. This update would get executed synchronously, which could trigger new invocations of the `once()` callback before it was unregistered.  

Repo: https://jsbin.com/xofinop/4/edit?html,js,console

### DIFF
--- a/.changeset/many-turtles-try.md
+++ b/.changeset/many-turtles-try.md
@@ -1,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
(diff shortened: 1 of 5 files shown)


jsdt (approved):

schmidt-sebastian:
@hsubox76 Do you have time to take a look at this?

hsubox76 (approved):

PR #4866, opened 2021-05-04
Support onInit callback to provider

Feiyang1:
onInit callbacks are invoked synchronously when `Provider.initialize()` executes. It allows a listener to react to provider initialization event synchronously to perform timing sensitive operations.


packages/component/src/provider.ts
```diff
@@ -250,9 +252,44 @@ export class Provider<T extends Name> {
...
   }
 
+  /**
+   *
+   * @param callback - a function that will be invoked  after the provider has been initialized by calling provider.initialize().
+   * The function is invoked SYNCHRONOUSLY, so it should not execute any longrunning tasks in order to not block the program.
+   *
+   * @returns a function to unregister the callback
+   */
+  onInit(callback: (instance: NameServiceMapping[T]) => void): () => void {
```
hsubox76, line 268: `callback` is of type `OnInitCallback<T>` right? Can that just be put here instead of the function signature? I notice `OnInitCallback` has an `identifier` param this one doesn't have, is that intentional?

Feiyang1: Good catch. Updated to use the type alias

hsubox76 (approved):

PR #4870, opened 2021-05-05
Fix database.useEmulator typing.

yuchenshi:
Adds correct typing for new argument introduced in #4792.

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1006,6 +1006,81 @@ declare namespace firebase {
     uid: string;
   }
 
+  type FirebaseSignInProvider =
+    | 'custom'
+    | 'email'
+    | 'password'
+    | 'phone'
+    | 'anonymous'
+    | 'google.com'
+    | 'facebook.com'
+    | 'github.com'
+    | 'twitter.com'
+    | 'microsoft.com'
+    | 'apple.com';
+
+  interface FirebaseIdToken {
+    /** Always set to https://securetoken.google.com/PROJECT_ID */
+    iss: string;
+
+    /** Always set to PROJECT_ID */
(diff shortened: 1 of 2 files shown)


packages/firebase/index.d.ts
```diff
@@ -7054,6 +7136,8 @@ declare namespace firebase.database {
     logger?: boolean | ((a: string) => any),
     persistent?: boolean
   ): any;
+
+  export type EmulatorMockTokenOptions = firebase.EmulatorMockTokenOptions;
```
Feiyang1, line 7140: I think we can just use `firebase.EmulatorMockTokenOptions` without needing the alias

yuchenshi: We already released a version with `firebase.firestore.EmulatorMockTokenOptions` and I think we should probably match that for `database` for consistency. WDYT?

Feiyang1: hmm, good point. Ok, let do that.

Feiyang1 (approved):

egilmorez (approved):

packages/firebase/index.d.ts
```diff
@@ -5711,8 +5786,15 @@ declare namespace firebase.database {
      *
      * @param host the emulator host (ex: localhost)
      * @param port the emulator port (ex: 8080)
+     * @param options.mockUserToken - the mock auth token to use for unit testing Security Rules
```
hsubox76, line 5789: The jsdoc format read by the "old" docgen tool doesn't like dashes (`-`).
The dashes are required in the exp doc comments though. Sorry, confusing.

hsubox76 (approved):
Approved pending one thing (see comment)

PR #4871, opened 2021-05-05
Update grpc-js to 1.3

schmidt-sebastian:
Updates minimum GRPC version to 1.3.

Fixes https://github.com/firebase/firebase-js-sdk/issues/4851

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -71,7 +71,7 @@
     "@firebase/logger": "0.2.6",
     "@firebase/util": "1.1.0",
     "@firebase/webchannel-wrapper": "0.4.1",
-    "@grpc/grpc-js": "^1.0.0",
+    "@grpc/grpc-js": "^1.3.2",
     "@grpc/proto-loader": "^0.5.0",
     "node-fetch": "2.6.1",
     "tslib": "^2.1.0"
@@ -85,7 +85,7 @@
     "@rollup/plugin-alias": "3.1.1",
     "@rollup/plugin-json": "4.1.0",
     "@rollup/plugin-node-resolve": "11.2.0",
-    "@types/eslint": "7.2.6",
+    "@types/eslint": "7.2.10",
     "@types/json-stable-stringify": "1.0.32",
     "json-stable-stringify": "1.0.1",
     "protobufjs": "6.10.2",
--- a/repo-scripts/size-analysis/package.json
+++ b/repo-scripts/size-analysis/package.json
@@ -27,7 +27,7 @@
     "rollup-plugin-typescript2": "0.29.0",
     "@rollup/plugin-virtual": "2.0.3",
     "webpack": "4.44.2",
-    "@types/webpack": "4.41.25",
+    "@types/webpack": "5.28.0",
     "webpack-virtual-modules": "0.4.1",
     "child-process-promise": "2.2.1",
     "memfs": "3.2.0",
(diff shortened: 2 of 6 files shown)


Feiyang1:
Please add a changeset

packages/firestore/package.json
```diff
@@ -71,7 +71,7 @@
     "@firebase/logger": "0.2.6",
     "@firebase/util": "1.0.0",
     "@firebase/webchannel-wrapper": "0.4.1",
-    "@grpc/grpc-js": "^1.0.0",
+    "@grpc/grpc-js": "1.3.0",
```
Feiyang1, line 74: Do you want to use `^` version ?

Feiyang1:
I think so. the base is a commit before appcheck was merged into master. The diff should be much smaller if you rebase the PR.

packages/firestore/package.json
schmidt-sebastian: Makes sense. I didn't realize we only use fixed versions for our own dependencies.

schmidt-sebastian:
Hm, I am not able to run the yarn update right now ("error An unexpected error occurred: "expected workspace package to exist for \"inquirer\"""). I will ping you once I updated the yarn.lock

schmidt-sebastian:
@Feiyang1 CI finally passed. I had to update some random types after removing yarn.lock

repo-scripts/size-analysis/package.json
```diff
@@ -27,7 +27,7 @@
     "rollup-plugin-typescript2": "0.29.0",
     "@rollup/plugin-virtual": "2.0.3",
     "webpack": "4.44.2",
-    "@types/webpack": "4.41.25",
+    "@types/webpack": "5.28.0",
```
Feiyang1, line 30: Is it required to pass CI? This is a wrong change because we are not using webpack@5. I will be okay with it if it's required to pass CI for some unknown reason. I don't want us to waste time on it and we should be updating to `webpack@5` soon.

schmidt-sebastian: I am not sure if 5.x is required, but the version that was installed without the yarn.lock file caused compilation errors. I can try an older 4.x version, but we should be safe since the code itself did not change.

schmidt-sebastian: The problem is here (from the original yarn.lock):

```
"@types/webpack@4.41.25":
  version "4.41.25"
  resolved "https://registry.npmjs.org/@types/webpack/-/webpack-4.41.25.tgz#4d3b5aecc4e44117b376280fbfd2dc36697968c4"
  integrity sha512-cr6kZ+4m9lp86ytQc1jPOJXgINQyz3kLLunZ57jznW+WIAL0JqZbGubQk4GlD42MuQL5JGOABrxdpqqWeovlVQ==
  dependencies:
    "@types/anymatch" "*"
    "@types/node" "*"
    "@types/tapable" "*"
    "@types/uglify-js" "*"
    "@types/webpack-sources" "*"
    source-map "^0.6.0"
```

Webpack 4.x does not compile against tapable 2.x. The old version was 1.06 (before removing the lock filed).

These are some of the errors:

```
../../node_modules/@types/webpack/index.d.ts:1106:32 - error TS2314: Generic type 'HookMap<H>' requires 1 type argument(s).

1106                 callAnyMember: HookMap;
                                    ~~~~~~~
```

Feiyang1 (approved):

PR #4875, opened 2021-05-06
Fix documentation

hsubox76:
I forgot to update the index.d.ts with the new AutoTokenRefresh param/method. Also added more comments.

### DIFF
--- a/packages/app-check-types/index.d.ts
+++ b/packages/app-check-types/index.d.ts
@@ -28,17 +28,35 @@ export interface FirebaseAppCheck {
     siteKeyOrProvider: string | AppCheckProvider,
     isTokenAutoRefreshEnabled?: boolean
   ): void;
+
+  /**
+   *
+   * @param isTokenAutoRefreshEnabled - If set to true, enables SDK to
+   * automatically refresh AppCheck token as needed. If false, disables this
+   * functionality. This overrides any value set during `activate()`.
+   */
   setTokenAutoRefreshEnabled(isTokenAutoRefreshEnabled: boolean): void;
 }
 
+/**
+ * An AppCheck provider. This may be either the built-in reCAPTCHA provider
+ * or a custom provider. For more on custom providers, see
+ * https://firebase.google.com/docs/app-check/web-custom-provider
+ */
 interface AppCheckProvider {
   /**
-   * returns an AppCheck token
+   * Returns an AppCheck token.
    */
   getToken(): Promise<AppCheckToken>;
 }
 
+/**
+ * The token returned from an `AppCheckProvider`.
+ */
 interface AppCheckToken {
+  /**
+   * The token string in JWT format.
+   */
   readonly token: string;
   /**
    * The local timestamp after which the token will expire.
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1469,19 +1469,45 @@ declare namespace firebase.appCheck {
   export interface AppCheck {
     /**
      * Activate AppCheck
-     * @param siteKeyOrOrovider - reCAPTCHA sitekey or custom token provider
+     * @param siteKeyOrProvider - reCAPTCHA sitekey or custom token provider
+     * @param isTokenAutoRefreshEnabled - If true, enables SDK to automatically
+     * refresh AppCheck token as needed. If undefined, the value will default
+     * to the value of `app.automaticDataCollectionEnabled`. That property
+     * defaults to false and can be set in the app config.
      */
-    activate(siteKeyOrProvider: string | AppCheckProvider): void;
+    activate(
+      siteKeyOrProvider: string | AppCheckProvider,
+      isTokenAutoRefreshEnabled?: boolean
+    ): void;
+
+    /**
+     *
+     * @param isTokenAutoRefreshEnabled - If set to true, enables SDK to
+     * automatically refresh AppCheck token as needed. If false, disables this
+     * functionality. This overrides any value set during `activate()`.
+     */
+    setTokenAutoRefreshEnabled(isTokenAutoRefreshEnabled: boolean): void;
   }
 
+  /**
+   * An AppCheck provider. This may be either the built-in reCAPTCHA provider
+   * or a custom provider. For more on custom providers, see
+   * https://firebase.google.com/docs/app-check/web-custom-provider
+   */
   interface AppCheckProvider {
     /**
-     * returns an AppCheck token
+     * Returns an AppCheck token.
      */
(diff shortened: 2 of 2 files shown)


packages/firebase/index.d.ts
```diff
@@ -1469,19 +1469,45 @@ declare namespace firebase.appCheck {
   export interface AppCheck {
     /**
      * Activate AppCheck
-     * @param siteKeyOrOrovider - reCAPTCHA sitekey or custom token provider
+     * @param siteKeyOrProvider - reCAPTCHA sitekey or custom token provider
```
Feiyang1, line 1472: this is tsdoc format (using `-`), but the doc generator uses jsdoc. I'm not sure if it will parse it correctly. We can try running the generator to verify, or just change to jsdoc format which seems easier.

packages/app-check-types/index.d.ts
```diff
@@ -28,17 +28,35 @@ export interface FirebaseAppCheck {
     siteKeyOrProvider: string | AppCheckProvider,
     isTokenAutoRefreshEnabled?: boolean
   ): void;
+
+  /**
+   *
+   * @param isTokenAutoRefreshEnabled - If set to true, enables SDK to
+   * automatically refresh AppCheck token as needed. If false, disables this
```
egilmorez, line 35: Is this a literal?  If not, maybe just App Check here and below (but is App Check actually publicly released right now?)

packages/app-check-types/index.d.ts
```diff
@@ -28,17 +28,35 @@ export interface FirebaseAppCheck {
...
+   * functionality. This overrides any value set during `activate()`.
+   */
   setTokenAutoRefreshEnabled(isTokenAutoRefreshEnabled: boolean): void;
 }
 
+/**
+ * An AppCheck provider. This may be either the built-in reCAPTCHA provider
+ * or a custom provider. For more on custom providers, see
+ * https://firebase.google.com/docs/app-check/web-custom-provider
```
egilmorez, line 44: Looks like this will 404 until 5/13:

https://critique-ng.corp.google.com/cl/371400287

packages/app-check-types/index.d.ts
```diff
@@ -28,17 +28,35 @@ export interface FirebaseAppCheck {
     siteKeyOrProvider: string | AppCheckProvider,
     isTokenAutoRefreshEnabled?: boolean
   ): void;
+
+  /**
+   *
+   * @param isTokenAutoRefreshEnabled - If set to true, enables SDK to
+   * automatically refresh AppCheck token as needed. If false, disables this
+   * functionality. This overrides any value set during `activate()`.
```
kevinthecheung, line 36:
```suggestion
   * @param isTokenAutoRefreshEnabled - If true, the SDK automatically
   * refreshes App Check tokens as needed. This overrides any value set
   * during `activate()`.
```

egilmorez (commented):
Thanks for sending!  Couple of things to look at.

packages/app-check-types/index.d.ts
kevinthecheung, line 44:
```suggestion
 * An App Check provider. This can be either the built-in reCAPTCHA provider
 * or a custom provider. For more on custom providers, see
 * https://firebase.google.com/docs/app-check/web-custom-provider
```

hsubox76: We can hold off on publishing our reference docs until 5/13 also, then. You can make a CL dependent on another CL, right?

packages/app-check-types/index.d.ts
hsubox76: No, it's not a literal, can change to App Check.

We are in a code freeze right now so the release will not happen until 5/11. The documentation can be published along with the rest on 5/13, see other comment.

packages/firebase/index.d.ts
```diff
@@ -1469,19 +1469,45 @@ declare namespace firebase.appCheck {
   export interface AppCheck {
     /**
      * Activate AppCheck
-     * @param siteKeyOrOrovider - reCAPTCHA sitekey or custom token provider
+     * @param siteKeyOrProvider reCAPTCHA sitekey or custom token provider
+     * @param isTokenAutoRefreshEnabled If true, enables SDK to automatically
+     * refresh AppCheck token as needed. If undefined, the value will default
+     * to the value of `app.automaticDataCollectionEnabled`. That property
+     * defaults to false and can be set in the app config.
```
kevinthecheung, line 1476:
```suggestion
     * @param siteKeyOrProvider reCAPTCHA v3 site key (public key) or
     * custom token provider.
     * @param isTokenAutoRefreshEnabled If true, the SDK automatically
     * refreshes App Check tokens as needed. If undefined, defaults to the
     * value of `app.automaticDataCollectionEnabled`, which defaults to
     * false and can be set in the app config.
```

packages/firebase/index.d.ts
```diff
@@ -1469,19 +1469,45 @@ declare namespace firebase.appCheck {
...
+    activate(
+      siteKeyOrProvider: string | AppCheckProvider,
+      isTokenAutoRefreshEnabled?: boolean
+    ): void;
+
+    /**
+     *
+     * @param isTokenAutoRefreshEnabled If set to true, enables SDK to
+     * automatically refresh AppCheck token as needed. If false, disables this
+     * functionality. This overrides any value set during `activate()`.
```
kevinthecheung, line 1487:
```suggestion
     * @param isTokenAutoRefreshEnabled If true, the SDK automatically
     * refreshes App Check tokens as needed. This overrides any value set
     * during `activate()`.
```

packages/firebase/index.d.ts
```diff
@@ -1469,19 +1469,45 @@ declare namespace firebase.appCheck {
...
      */
-    activate(siteKeyOrProvider: string | AppCheckProvider): void;
+    setTokenAutoRefreshEnabled(isTokenAutoRefreshEnabled: boolean): void;
   }
 
+  /**
+   * An AppCheck provider. This may be either the built-in reCAPTCHA provider
+   * or a custom provider. For more on custom providers, see
+   * https://firebase.google.com/docs/app-check/web-custom-provider
```
kevinthecheung, line 1495:
```suggestion
   * An App Check provider. This can be either the built-in reCAPTCHA
   * provider or a custom provider. For more on custom providers, see
   * https://firebase.google.com/docs/app-check/web-custom-provider
```

kevinthecheung (approved):

packages/app-check-types/index.d.ts
egilmorez: Makes sense, thanks!

PR #4878, opened 2021-05-06
Change techwriter alias to individual account

hsubox76:
It doesn't seem like the techwriters alias is working for triggering a suggested reviewer or assignee, change to individual account for now (@egilmorez) and see if it works better.


egilmorez (commented):
Is part of the problem that now doc strings come from many many other sources?

For instance, packages/app-check-types/index.d.ts :)

hsubox76:
> Is part of the problem that now doc strings come from many many other sources?
> 
> For instance, packages/app-check-types/index.d.ts :)

No, that file doesn't affect doc generation, as v8 still only uses the single file. We will have to find a way to deal with that problem in exp packages though.

Feiyang1 (approved):

Issue #4236, opened 2020-12-23
[Firestore] Registering an unload listener and best practices


dbstratta:
### Describe your environment

  * Operating System version: **Ubuntu 20.10**
  * Browser version: **Chrome 87.0.4280.88 (Official Build) (64-bit)**
  * Firebase SDK version: **8.2.1**
  * Firebase Product: **firestore**

### Describe the problem

Lighthouse 6.4.0 on Chrome 87 warns that the firebase SDK registers an `unload` event listener and it can prevent browser optimizations like the Back-Forward Cache.

![Screenshot from 2020-12-23 20-19-52](https://user-images.githubusercontent.com/18295954/103044054-d3a4cc80-455d-11eb-8ee9-b8ee23ccd8c9.png)

#### Steps to reproduce:

Enable the firestore local persistance.

#### Relevant Code:

```javascript
import firebase from 'firebase/app';
import 'firebase/firestore';

firebase.firestore();
firebase.firestore().enablePersistence({ synchronizeTabs: true }); // Happens with `synchronizeTabs: false` too.
```

wu-hui:
Thanks for the report! I believe this is something we should address, I can provide an update after the holidays, in the meantime, I'll keep the issue open.

johanchouquet:
Hi @wu-hui,

I have the same issue. It brings our score in lighthouse to be not as high as it could be.

Did you plan to address that point in the end ?

PR #4886, opened 2021-05-10
Use 'pagehide' for page termination by default.

wu-hui:
Fix: https://github.com/firebase/firebase-js-sdk/issues/4236

See: https://developers.google.com/web/updates/2018/07/page-lifecycle-api#legacy-lifecycle-apis-to-avoid

### DIFF
--- a/packages/firestore/src/util/types.ts
+++ b/packages/firestore/src/util/types.ts
@@ -59,14 +59,6 @@ export interface WindowLike {
   removeEventListener(type: string, listener: EventListener): void;
 }
 
-/**
- * Returns the page termination event to listen to. 'pagehide' is recommended
- * if available, it falls back to the less reliable 'unload'.
- */
-export function terminationEvent(window: WindowLike | null): string {
-  return 'onpagehide' in window! ? 'pagehide' : 'unload';
-}
-
 /** The subset of the browser's Document interface used by the SDK. */
 export interface DocumentLike {
   readonly visibilityState: VisibilityState;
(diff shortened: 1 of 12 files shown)


packages/firestore/src/util/types.ts
```diff
@@ -59,6 +59,14 @@ export interface WindowLike {
   removeEventListener(type: string, listener: EventListener): void;
 }
 
+/**
+ * Returns the page termination event to listen to. 'pagehide' is recommended
+ * if available, it falls back to the less reliable 'unload'.
+ */
+export function terminationEvent(window: WindowLike | null): string {
+  return 'onpagehide' in window! ? 'pagehide' : 'unload';
```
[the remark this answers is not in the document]
wu-hui: Good point, done.

schmidt-sebastian (commented):
Please add changeset.

packages/firestore/test/util/test_platform.ts
```diff
@@ -58,7 +58,7 @@ export class FakeWindow implements WindowLike {
       case 'storage':
         this.storageListeners.push(listener);
         break;
-      case 'unload':
+      case 'pagehide':
       case 'visibilitychange':
         // The spec tests currently do not rely on `unload`/`visibilitychange`
```
0xdevalias, line 63: [nit] Looks like this comment refers to the old ‘unload’ rather than ‘pagehide’

schmidt-sebastian: Thanks for catching this. WIll fix soon.

PR #4888, opened 2021-05-10
Create app-check-exp

hsubox76:
Create modular version of appcheck (`app-check-exp`).

Future PRs will add:
- app-check-compat
- functions-exp integration
- storage-exp integration
- database-exp integration


packages-exp/app-check-exp/src/api.ts
```diff
@@ -0,0 +1,132 @@
...
+  // `isTokenAutoRefreshEnabled` param was not provided by user.
+  newState.isTokenAutoRefreshEnabled =
+    isTokenAutoRefreshEnabled === undefined
+      ? app.automaticDataCollectionEnabled
+      : isTokenAutoRefreshEnabled;
+
+  setState(app, newState);
+
+  // initialize reCAPTCHA if siteKey is provided
+  if (newState.provider instanceof ReCaptchaV3Provider) {
```
Feiyang1, line 99: We can not use instanceof to check the provider type, and do different things based on that. This breaks tree shaking and will include code for all provider implementations.

Instead, we should define an interface with lifecycle methods that all providers implement. For example, they should all implement a `initialize()` method that we can call here. so here we just call `newState.provider.initialize()` and for recaptchav3, the method should call `initializeRecaptcha()`.

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -0,0 +1,311 @@
...
+      // has no `iat` field or value is out of bounds.
+      const issuedAtTimeMillis =
+        issuedAtTimeSeconds !== null &&
+        issuedAtTimeSeconds < Date.now() &&
+        issuedAtTimeSeconds > 0
+          ? issuedAtTimeSeconds * 1000
+          : Date.now();
+
+      token = { ...customToken, issuedAtTimeMillis };
+    }
```
Feiyang1, line 127: We need to encapsulate the special handling for custom tokens in the customProvider, so we don't need to do instanceof check which breaks tree shaking.

We will need to wrap the user provided customProvider in our own class which can do the special handling. We probably need to ask developers to create a customProvider using a class, or a function we provide, something like:

```
import { initializeAppCheck, CustomProvider } from 'firebase/app-check';

initializeAppCheck(app, new CustomProvider({
    getToken: () => {}
    ...
}));

```

packages-exp/app-check-exp/src/providers.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  /**
+   * Returns an AppCheck token.
+   */
+  getToken(): Promise<AppCheckToken> {
+    return this._delegate.getToken();
+  }
+}
+
+export class ReCaptchaV3ProviderImpl implements AppCheckProvider {
```
Feiyang1, line 52: Can we just implement things in `ReCaptchaV3Provider` instead of using `_delegate` with another class?

Feiyang1 (changes requested):

packages-exp/app-check-exp/src/api.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      const fakeCustomTokenProvider = getFakeCustomTokenProvider();
+      const initReCAPTCHAStub = stub(reCAPTCHA, 'initialize');
+      initializeAppCheck(app, {
+        provider: fakeCustomTokenProvider
+      });
+      expect(getState(app).provider).to.equal(fakeCustomTokenProvider);
+      expect(initReCAPTCHAStub).to.have.not.been.called;
+    });
+  });
+  describe('activate()', () => {
```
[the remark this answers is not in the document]
hsubox76: Done.

packages-exp/app-check-exp/src/api.ts
hsubox76: Done.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: Done.

packages-exp/app-check-exp/src/providers.ts
hsubox76: Done.

packages-exp/app-check-exp/src/public-types.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Options for App Check initialization.
+ * @public
+ */
+export interface AppCheckOptions {
+  /**
+   * reCAPTCHA provider or custom provider.
+   */
+  provider: CustomProvider | ReCaptchaV3Provider;
```
hsubox76, line 58: Just wanted to make sure, importing these classes won't affect tree shaking, right? It will be smart enough to only use them as types?

Feiyang1: Correct.

packages-exp/app-check-exp/src/api.ts
```diff
@@ -0,0 +1,117 @@
...
+declare module '@firebase/component' {
+  interface NameServiceMapping {
+    'app-check-exp': AppCheckService;
+  }
+}
+
+export { ReCaptchaV3Provider, CustomProvider } from './providers';
+
+/**
+ * Activate AppCheck for the given app. Can be called only once per app.
```
egilmorez, line 35: Is "AppCheck" the product name, or is it App Check?

If we want to refer to an actual literal AppCheck object or entity, then we should use backticks.  However, I think all of the instances in this comment are just references to the product, where we should use the product name.

This would be something to check throughout the PR.

packages-exp/app-check-exp/src/client.ts
```diff
@@ -0,0 +1,134 @@
...
+  EXCHANGE_DEBUG_TOKEN_METHOD,
+  EXCHANGE_RECAPTCHA_TOKEN_METHOD
+} from './constants';
+import { FirebaseApp } from '@firebase/app-exp';
+import { ERROR_FACTORY, AppCheckError } from './errors';
+import { Provider } from '@firebase/component';
+import { AppCheckTokenInternal } from './types';
+
+/**
+ * Response JSON returned from AppCheck server endpoint.
```
egilmorez, line 29: I think we would normally say "JSON response," correct?

packages-exp/app-check-exp/src/errors.ts
```diff
@@ -0,0 +1,71 @@
...
+  RECAPTCHA_ERROR = 'recaptcha-error'
+}
+
+const ERRORS: ErrorMap<AppCheckError> = {
+  [AppCheckError.ALREADY_INITIALIZED]:
+    'You have already called initializeAppCheck() for FirebaseApp {$appName}, ' +
+    'initializeAppCheck() can only be called once.',
+  [AppCheckError.USE_BEFORE_ACTIVATION]:
+    'AppCheck is being used before initializeAppCheck() is called for FirebaseApp {$appName}. ' +
+    'Please make sure you call initializeAppCheck() before instantiating other Firebase services.',
```
egilmorez, line 38: We try to avoid "please."

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -0,0 +1,293 @@
...
+  tokenErrorData: Record<string, string>
+): string {
+  return base64.encodeString(
+    JSON.stringify(tokenErrorData),
+    /* webSafe= */ false
+  );
+}
+
+/**
+ * This function will always resolve.
```
egilmorez, line 54: Is this a public-facing comment?

We should use present tense instead of "will" if so.  It's a good habit to get into.

packages-exp/app-check-exp/src/providers.ts
```diff
@@ -0,0 +1,123 @@
...
+import { CustomProviderOptions } from './public-types';
+import {
+  getToken as getReCAPTCHAToken,
+  initialize as initializeRecaptcha
+} from './recaptcha';
+import { AppCheckProvider, AppCheckTokenInternal } from './types';
+
+/**
+ * App Check provider that can obtain a reCAPTCHA V3 token and exchange it
+ * for an AppCheck token.
```
egilmorez, line 32: As mentioned above, is there a reason to use literal-looking "AppCheck" instead of "App Check?"

packages-exp/app-check-exp/src/public-types.ts
```diff
@@ -0,0 +1,75 @@
...
+import { CustomProvider, ReCaptchaV3Provider } from './providers';
+
+/**
+ * The Firebase AppCheck service interface.
+ *
+ * @public
+ */
+export interface AppCheck {
+  /**
+   * The FirebaseApp this Functions instance is associated with.
```
egilmorez, line 28: Should this be "`FirebaseApp` object" (a literal) or just Firebase app?

Also, recommend "Cloud Functions instance" unless, again, it is a literal (which would be backticked).

packages-exp/app-check-exp/src/public-types.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Options for App Check initialization.
+ * @public
+ */
+export interface AppCheckOptions {
+  /**
+   * reCAPTCHA provider or custom provider.
+   */
+  provider: CustomProvider | ReCaptchaV3Provider;
+  /**
+   * If set to true, enables automatic background refresh of app check token.
```
egilmorez, line 60: I think this should be capped "App Check"

packages-exp/app-check-exp/src/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+  readDebugTokenFromIndexedDB,
+  readTokenFromIndexedDB,
+  writeDebugTokenToIndexedDB,
+  writeTokenToIndexedDB
+} from './indexeddb';
+import { logger } from './logger';
+import { AppCheckTokenInternal } from './types';
+
+/**
+ * Always resolves. In case of an error reading from indexeddb, resolve with undefined
```
egilmorez, line 31: Are these public-facing comments in this file?

packages-exp/app-check-exp/src/types.ts
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { FirebaseApp } from '@firebase/app-exp';
+import { AppCheckToken } from './public-types';
+
+export interface FirebaseAppCheckInternal {
+  // Get the current AttestationToken. Attaches to the most recent in-flight request if one
```
egilmorez, line 22: If these are public-facing comments, there are several places to consider: is it a literal or just a thing, like "attestation token."

egilmorez (commented):
Thanks for sending Christina!

Please add kevincheung@ to this review, as he's the App Check writer.

packages-exp/app-check-exp/src/api.ts
hsubox76: Went through manually and got most of them I think.

packages-exp/app-check-exp/src/client.ts
hsubox76: I think this makes sense for us and it's not public.

packages-exp/app-check-exp/src/errors.ts
hsubox76: Fixed.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: No, nothing in this file should be public. Fixed though.

packages-exp/app-check-exp/src/providers.ts
hsubox76: Changed to product name.

packages-exp/app-check-exp/src/public-types.ts
hsubox76: Thank you, that was copied and pasted from the Functions repo and should be AppCheck! Also made them both literals.

packages-exp/app-check-exp/src/public-types.ts
hsubox76: Fixed.

packages-exp/app-check-exp/src/storage.ts
hsubox76: No, the only public strings here are the logger.warn and console.log messages.

packages-exp/app-check-exp/src/types.ts
hsubox76: This file is internal only. Maybe we should come up with a comment header to indicate files where none of the comments are public?

packages-exp/app-check-exp/src/client.ts
egilmorez: SG, thanks!

packages-exp/app-check-exp/src/internal-api.ts
egilmorez: That's great, because if we write a decent amount of internal docs/comments, habits will form!  Thanks.

egilmorez (approved):
Thanks Christina!

packages-exp/app-check-exp/src/api.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+      });
+      expect(() =>
+        initializeAppCheck(app, {
+          provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+        })
+      ).to.throw(/appCheck\/already-initialized/);
+    });
+
+    it('initialize reCAPTCHA when a sitekey is provided', () => {
```
Feiyang1, line 55: The text needs to be updated.

packages-exp/app-check-exp/src/api.ts
```diff
@@ -0,0 +1,117 @@
...
+ * Activate App Check
+ * @param app - Firebase app to activate App Check for.
+ * @param provider - reCAPTCHA v3 provider or
+ * custom token provider.
+ * @param isTokenAutoRefreshEnabled - If true, the SDK automatically
+ * refreshes App Check tokens as needed. If undefined, defaults to the
+ * value of `app.automaticDataCollectionEnabled`, which defaults to
+ * false and can be set in the app config.
+ *
+ * @internal
```
Feiyang1, line 69: No need to add the annotation if the function is not exported from the package.

packages-exp/app-check-exp/src/internal-api.test.ts
```diff
@@ -0,0 +1,443 @@
...
+
+      expect(customProviderSpy).to.be.called;
+      expect(token).to.deep.equal({
+        token: 'fake-custom-app-check-token'
+      });
+
+      clock.restore();
+    });
+
+    it('uses reCAPTCHA token to exchange for AppCheck token if no customTokenProvider is provided', async () => {
```
Feiyang1, line 90: The test name is no longer appropriate for the Provider approach we are taking. Perhaps simplify it to something like: `uses RecaptchaV3Provider to get an AppCheck token` similar to the test name for `customTokenProvider`.

packages-exp/app-check-exp/src/internal-api.test.ts
```diff
@@ -0,0 +1,443 @@
...
+
+    it('force to get new token when forceRefresh is true', async () => {
+      initializeAppCheck(app, {
+        provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+      });
+      setState(app, { ...getState(app), token: fakeRecaptchaAppCheckToken });
+
+      stub(reCAPTCHA, 'getToken').returns(Promise.resolve(fakeRecaptchaToken));
+      stub(client, 'exchangeToken').returns(
+        Promise.resolve(fakeRecaptchaAppCheckToken)
```
Feiyang1, line 285: We can't tell whether we got a new token or not if it's the same one as the cached one. Can we?

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -0,0 +1,293 @@
...
+    return {
+      token: token.token
+    };
+  }
+
+  /**
+   * request a new token
+   */
+  try {
+    if (state.provider) {
```
Feiyang1, line 107: I don't think it's possible to not have a provider, so can we use ! to assert it is available, and add a comment to clarify?

packages-exp/app-check-exp/src/types.ts
Feiyang1: Public APIs/comments are marked with `@public`, so we can rely on it.

packages-exp/app-check-exp/src/api.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      initializeAppCheck(app, {
+        provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+      });
+      expect(initReCAPTCHAStub).to.have.been.calledWithExactly(
+        app,
+        FAKE_SITE_KEY
+      );
+    });
+
+    it('does NOT initialize reCAPTCHA when a custom token provider is provided', () => {
```
Feiyang1, line 68: This test seems unnecessary since you have to select a provider explicitly.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: Done.

packages-exp/app-check-exp/src/api.test.ts
hsubox76: Removed.

packages-exp/app-check-exp/src/api.test.ts
hsubox76: Fixed.

packages-exp/app-check-exp/src/api.ts
hsubox76: Removed.

packages-exp/app-check-exp/src/internal-api.test.ts
hsubox76: Done.

packages-exp/app-check-exp/src/internal-api.test.ts
hsubox76: Changed the contents of the stubbed return promise.

packages-exp/app-check-exp/README.md
```diff
@@ -0,0 +1,5 @@
+# @firebase/app-check
+
+This is the Firebase AppCheck component of the Firebase JS SDK.
```
kevinthecheung, line 3:
```suggestion
This is the Firebase App Check component of the Firebase JS SDK.
```

packages-exp/app-check-exp/README.md
```diff
@@ -0,0 +1,5 @@
...
+
+**This package is not intended for direct usage, and should only be used via the officially supported [firebase](https://www.npmjs.com/package/firebase) package.**
```
kevinthecheung, line 5:
```suggestion
**This package is not intended for direct usage, and should only be used via the officially supported [`firebase`](https://www.npmjs.com/package/firebase) package.**
```

packages-exp/app-check-exp/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@firebase/app-check-exp",
+  "version": "0.0.900",
+  "private": true,
+  "description": "An AppCheck package for new firebase packages",
```
kevinthecheung, line 5:
```suggestion
  "description": "An App Check package for new firebase packages",
```

packages-exp/app-check-exp/src/api.ts
```diff
@@ -0,0 +1,117 @@
...
+    'app-check-exp': AppCheckService;
+  }
+}
+
+export { ReCaptchaV3Provider, CustomProvider } from './providers';
+
+/**
+ * Activate App Check for the given app. Can be called only once per app.
+ * @param app - the FirebaseApp to activate App Check for
+ * @param options - Provider implementation that can return an App Check token
```
kevinthecheung, line 37: does this description match the param?

packages-exp/app-check-exp/src/errors.ts
```diff
@@ -0,0 +1,71 @@
...
+  RECAPTCHA_ERROR = 'recaptcha-error'
+}
+
+const ERRORS: ErrorMap<AppCheckError> = {
+  [AppCheckError.ALREADY_INITIALIZED]:
+    'You have already called initializeAppCheck() for FirebaseApp {$appName}, ' +
+    'initializeAppCheck() can only be called once.',
+  [AppCheckError.USE_BEFORE_ACTIVATION]:
+    'App Check is being used before initializeAppCheck() is called for FirebaseApp {$appName}. ' +
+    'Make sure you call initializeAppCheck() before instantiating other Firebase services.',
```
kevinthecheung, line 38:
```suggestion
    'Call initializeAppCheck() before instantiating other Firebase services.',
```

packages-exp/app-check-exp/src/providers.ts
```diff
@@ -0,0 +1,123 @@
...
+import { CustomProviderOptions } from './public-types';
+import {
+  getToken as getReCAPTCHAToken,
+  initialize as initializeRecaptcha
+} from './recaptcha';
+import { AppCheckProvider, AppCheckTokenInternal } from './types';
+
+/**
+ * App Check provider that can obtain a reCAPTCHA V3 token and exchange it
+ * for an AppC heck token.
```
kevinthecheung, line 32:
```suggestion
 * for an App Check token.
```

packages-exp/app-check-exp/src/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function readTokenFromStorage(
+  app: FirebaseApp
+): Promise<AppCheckTokenInternal | undefined> {
+  if (isIndexedDBAvailable()) {
+    let token = undefined;
+    try {
+      token = await readTokenFromIndexedDB(app);
+    } catch (e) {
+      // swallow the error and return undefined
+      logger.warn(`Failed to read token from indexeddb. Error: ${e}`);
```
kevinthecheung, line 42:
```suggestion
      logger.warn(`Failed to read token from IndexedDB. Error: ${e}`);
```

packages-exp/app-check-exp/src/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Always resolves. In case of an error writing to indexeddb, print a warning and resolve the promise
+ */
+export function writeTokenToStorage(
+  app: FirebaseApp,
+  token: AppCheckTokenInternal
+): Promise<void> {
+  if (isIndexedDBAvailable()) {
+    return writeTokenToIndexedDB(app, token).catch(e => {
+      // swallow the error and resolve the promise
+      logger.warn(`Failed to write token to indexeddb. Error: ${e}`);
```
kevinthecheung, line 60:
```suggestion
      logger.warn(`Failed to write token to IndexedDB. Error: ${e}`);
```

packages-exp/app-check-exp/src/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!existingDebugToken) {
+    // create a new debug token
+    const newToken = uuidv4();
+    // We don't need to block on writing to indexeddb
+    // In case persistence failed, a new debug token will be generated everytime the page is refreshed.
+    // It renders the debug token useless because you have to manually register(whitelist) the new token in the firebase console again and again.
+    // If you see this error trying to use debug token, it probably means you are using a browser that doesn't support indexeddb.
+    // You should switch to a different browser that supports indexeddb
+    writeDebugTokenToIndexedDB(newToken).catch(e =>
+      logger.warn(`Failed to persist debug token to indexeddb. Error: ${e}`)
```
kevinthecheung, line 88:
```suggestion
      logger.warn(`Failed to persist debug token to IndexedDB. Error: ${e}`)
```

packages-exp/app-check-exp/src/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+    );
+    // Not using logger because I don't think we ever want this accidentally hidden?
+    console.log(
+      `App Check debug token: ${newToken}. You will need to whitelist it in the Firebase console for it to work`
```
kevinthecheung, line 92:
```suggestion
      `App Check debug token: ${newToken}. You will need to add it to your app's App Check settings in the Firebase console for it to work`
```

kevinthecheung (approved):

packages-exp/app-check-exp/src/api.ts
hsubox76: Oops, thanks, updated.

Feiyang1 (approved):

PR #4897, opened 2021-05-12
Make App Check initialization explicit

hsubox76:
Make AppCheck initialization explicit, so users not intending to use app-check won't get errors in storage, functions, etc. Modify _getService to initialize component if it hasn't been.

### DIFF
--- a/packages/app/src/firebaseApp.ts
+++ b/packages/app/src/firebaseApp.ts
@@ -29,7 +29,8 @@ import {
   ComponentContainer,
   Component,
   ComponentType,
-  Name
+  Name,
+  InstantiationMode
 } from '@firebase/component';
 import { AppError, ERROR_FACTORY } from './errors';
 import { DEFAULT_ENTRY_NAME } from './constants';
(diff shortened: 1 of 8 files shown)


packages/app/src/firebaseApp.ts
```diff
@@ -122,8 +123,17 @@ export class FirebaseAppImpl implements FirebaseApp {
...
 
+    // Initialize instance if InstatiationMode is `EXPLICIT`.
+    const provider = this.container.getProvider(name as Name);
+    if (
+      !provider.isInitialized() &&
+      provider.getComponent()?.instantiationMode === InstantiationMode.EXPLICIT
+    ) {
+      provider.initialize();
+    }
+
```
Feiyang1, line 134: Can you please add the same logic to app-compat? Also let's add a test case for this.

hsubox76: Added.

Feiyang1 (approved):

PR #4899, opened 2021-05-12
WIP: Add isActivated() to app check

hsubox76:
WIP. Need to submit as an API proposal.


hsubox76:
This won't be necessary when initializeAppCheck is made idempotent as is planned.

PR #4902, opened 2021-05-13
Change appcheck activate() to use provider pattern

hsubox76:
Create provider classes that users can use in `activate()`. To avoid breaking changes, `activate` still takes a string site key or our previous custom provider pattern, which is any object with a `getToken` property.

I also changed `addTokenListener` to check for existing tokens in both memory and storage because otherwise it will only check storage if `getToken()` is called, which only happens if the refresher starts, which only happens if automatic token refresh is enabled. I don't think fetching a token from local storage should be gated on that.

Note: API changes should also be propagated to v9 compat after this.

Edit: For docs review - only need to review `packages/firebase/index.d.ts`

### DIFF
--- a/packages/app-check-types/index.d.ts
+++ b/packages/app-check-types/index.d.ts
@@ -91,7 +91,7 @@ interface AppCheckProvider {
   getToken(): Promise<AppCheckToken>;
 }
 
-export class ReCAPTCHAV3Provider {
+export class ReCaptchaV3Provider {
   /**
    * @param siteKey - ReCAPTCHA v3 site key (public key).
    */
--- a/packages/app-check-types/index.d.ts
+++ b/packages/app-check-types/index.d.ts
@@ -90,6 +90,13 @@ interface AppCheckProvider {
   getToken(): Promise<AppCheckToken>;
 }
 
+interface ReCAPTCHAV3Provider {
+  /**
+   * @param siteKey - ReCAPTCHA v3 site key (public key).
+   */
+  constructor(siteKey: string): void;
+}
+
 /**
  * The token returned from an `AppCheckProvider`.
  */
--- a/packages/app-check-types/index.d.ts
+++ b/packages/app-check-types/index.d.ts
@@ -91,20 +91,20 @@ interface AppCheckProvider {
   getToken(): Promise<AppCheckToken>;
 }
 
-interface ReCAPTCHAV3Provider {
+export class ReCAPTCHAV3Provider {
   /**
    * @param siteKey - ReCAPTCHA v3 site key (public key).
    */
-  constructor(siteKey: string): void;
+  constructor(siteKey: string);
 }
 /*
  * Custom token provider.
  */
-interface CustomProvider {
+export class CustomProvider {
   /**
    * @param options - Options for creating the custom provider.
    */
-  constructor(options: CustomProviderOptions): void;
+  constructor(options: CustomProviderOptions);
 }
 interface CustomProviderOptions {
   /**
--- a/packages/app-check-types/index.d.ts
+++ b/packages/app-check-types/index.d.ts
@@ -17,6 +17,7 @@
 
 import { PartialObserver, Unsubscribe } from '@firebase/util';
 import { FirebaseApp } from '@firebase/app-types';
+import { Provider } from '@firebase/component';
 
 export interface FirebaseAppCheck {
   /** The `FirebaseApp` associated with this instance. */
@@ -95,6 +96,22 @@ interface ReCAPTCHAV3Provider {
    * @param siteKey - ReCAPTCHA v3 site key (public key).
    */
   constructor(siteKey: string): void;
+}
+/*
+ * Custom token provider.
+ */
+interface CustomProvider {
+  /**
+   * @param options - Options for creating the custom provider.
+   */
+  constructor(options: CustomProviderOptions): void;
+}
+interface CustomProviderOptions {
+  /**
+   * Function to get an App Check token through a custom provider
+   * service.
+   */
+  getToken: () => Promise<AppCheckToken>;
 }
 
 /**
--- a/packages/app-check/src/api.ts
+++ b/packages/app-check/src/api.ts
@@ -34,7 +34,7 @@ import { CustomProvider, ReCaptchaV3Provider } from './providers';
 /**
  *
  * @param app
- * @param provider - optional custom attestation provider
+ * @param siteKeyOrProvider - optional custom attestation provider
  * or reCAPTCHA provider
  * @param isTokenAutoRefreshEnabled - if true, enables auto refresh
  * of appCheck token.
--- a/packages/app-check/src/api.ts
+++ b/packages/app-check/src/api.ts
@@ -64,7 +64,7 @@ export function activate(
   // Read cached token from storage if it exists and store it in memory.
   newState.cachedTokenPromise = readTokenFromStorage(app).then(cachedToken => {
     if (cachedToken && isValid(cachedToken)) {
-      newState.token = cachedToken;
+      setState(app, { ...getState(app), token: cachedToken });
     }
     return cachedToken;
   });
--- a/packages/app-check/src/api.ts
+++ b/packages/app-check/src/api.ts
@@ -20,7 +20,9 @@ import {
   AppCheckTokenResult
 } from '@firebase/app-check-types';
 import { FirebaseApp } from '@firebase/app-types';
+import { Provider } from '@firebase/component';
 import { ERROR_FACTORY, AppCheckError } from './errors';
+import { ReCAPTCHAProvider, ReCAPTCHAProviderInternal } from './providers';
 import { initialize as initializeRecaptcha } from './recaptcha';
 import { getState, setState, AppCheckState, ListenerType } from './state';
 import {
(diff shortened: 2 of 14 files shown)


packages/app-check/src/internal-api.ts
```diff
@@ -209,14 +167,23 @@ export function addTokenListener(
...
     const validToken = state.token;
     Promise.resolve()
       .then(() => listener({ token: validToken.token }))
       .catch(() => {
         /** Ignore errors in listeners. */
       });
+  } else {
+    // Also try storage. Otherwise isTokenAutoRefreshEnabled == false
+    // will prevent reading existing token from storage.
+    void readTokenFromStorage(app).then(cachedToken => {
```
hsubox76, line 182: I realized that without this, listeners won't read tokens in storage if automatic token refresh is disabled, and I don't think that's expected behavior. If the refresher doesn't start, getToken isn't called, and that's the only place that reads from storage. I'm not sure if this is the best place to change it or if we need to rethink where we block token refresh.

packages/app-check/src/index.ts
```diff
@@ -46,6 +47,10 @@ function registerAppCheck(firebase: _FirebaseNamespace): void {
       },
       ComponentType.PUBLIC
     )
+      .setServiceProps({
```
Feiyang1, line 50: Need to add them to the `declare module` block at the end of the file.

```
  interface FirebaseNamespace {
    appCheck(app?: FirebaseApp): FirebaseAppCheck;
    ReCaptchaV3Provider: typeof ReCAPTCHAV3Provider;
    CustomProvider: typeof CustomProvider;
  }
```

Feiyang1 (changes requested):

packages/app-check/src/providers.ts
```diff
@@ -0,0 +1,140 @@
...
+    app: FirebaseApp,
+    platformLoggerProvider: Provider<'platform-logger'>
+  ): void;
+}
+
+/**
+ * App Check provider that can obtain a reCAPTCHA V3 token and exchange it
+ * for an App Check token.
+ *
+ * @public
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/app-check-types/index.d.ts
```diff
@@ -90,6 +91,29 @@ interface AppCheckProvider {
   getToken(): Promise<AppCheckToken>;
 }
 
+interface ReCAPTCHAV3Provider {
+  /**
+   * @param siteKey - ReCAPTCHA v3 site key (public key).
+   */
+  constructor(siteKey: string): void;
```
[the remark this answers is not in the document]
hsubox76: Actually are these needed at all? If index.ts exports the types based on the imports of the actual classes from provider.ts, and firebase/index.d.ts has these types on the appCheck namespace, I don't think anything uses them from the app-check-types-package?

packages/app-check/src/api.ts
```diff
@@ -30,18 +29,25 @@ import {
...
 import { Provider } from '@firebase/component';
 import { ErrorFn, NextFn, PartialObserver, Unsubscribe } from '@firebase/util';
+import { CustomProvider, ReCaptchaV3Provider } from './providers';
 
 /**
  *
  * @param app
- * @param siteKeyOrProvider - optional custom attestation provider
- * or reCAPTCHA siteKey
+ * @param provider - optional custom attestation provider
```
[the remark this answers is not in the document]
hsubox76: Oops. yes.

packages/app-check/src/internal-api.ts
hsubox76: This is a little complicated, that means either:
1. activate() needs to be made async, which I don't think is an option
2. the in-memory token needs to always be stored as a promise (maybe?)
3. leave it a hanging promise, if it didn't manage to get the token from storage in time for the next getToken() call, forget it and try the exchange endpoint. Probably not ok?

packages/app-check/src/providers.ts
```diff
@@ -0,0 +1,140 @@
...
+      // We don't have an appName to provide if so.
+      // This should already be caught in the top level `getToken()` function.
+      throw ERROR_FACTORY.create(AppCheckError.USE_BEFORE_ACTIVATION, {
+        appName: ''
+      });
+    }
+    const attestedClaimsToken = await getReCAPTCHAToken(this._app).catch(_e => {
+      // reCaptcha.execute() throws null which is not very descriptive.
+      throw ERROR_FACTORY.create(AppCheckError.RECAPTCHA_ERROR);
+    });
```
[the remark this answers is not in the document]
hsubox76: Done.

packages/app-check/src/index.ts
hsubox76: Done.

packages/app-check-types/index.d.ts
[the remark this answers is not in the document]
Feiyang1: True, but we have tried to keep it consistent in types packages, so people can import types using scope packages.

packages/app-check/src/index.ts
```diff
@@ -94,6 +99,8 @@ initializeDebugMode();
 declare module '@firebase/app-types' {
   interface FirebaseNamespace {
     appCheck(app?: FirebaseApp): FirebaseAppCheck;
+    ReCaptchaV3Provider: typeof ReCaptchaV3Provider;
+    CustomProvider: typeof CustomProvider;
```
Feiyang1, line 103: Use the classes exported from the types package, otherwise people may get type errors.

packages/app-check/src/internal-api.ts
Feiyang1: > 1. activate() needs to be made async, which I don't think is an option
> 2. the in-memory token needs to always be stored as a promise (maybe?)
> 3. leave it a hanging promise, if it didn't manage to get the token from storage in time for the next getToken() call, forget it and try the exchange endpoint. Probably not ok?

Yeah, we will need to store a promise for reading the cached token in the state. Here is the workflow when we need a token:
- if `state.token` is defined and valid, return it
- if `state.token` is defined and invalid, try the exchange endpoint
- If `state.token` is undefined, wait for this promise to resolve
- if the promise resolves to null (no cached token), or an invalid token, try the exchange endpoint

The benefits of doing it are that `readTokenFromStorage` is only called in a single place and it won't be called the second time unnecessarily, which is possible with the current implementation. The downside is that we need to manage one additional variable in the state. I will leave it up to you to decide.

packages/app-check/src/index.ts
hsubox76: Done, also made casing consistent.

packages/app-check/src/index.ts
```diff
@@ -48,8 +53,8 @@ function registerAppCheck(firebase: _FirebaseNamespace): void {
       ComponentType.PUBLIC
     )
       .setServiceProps({
-        ReCaptchaV3Provider,
-        CustomProvider
+        ReCaptchaV3Provider: ReCaptchaV3ProviderImpl,
+        CustomProvider: CustomProviderImpl
```
Feiyang1, line 57: Should use `typeof` because we want the constructor type, not the instance type.

Feiyang1 (changes requested):

packages/app-check/src/index.ts
hsubox76: Did you mean to add `typeof` here or in the `declare` block below? I added it to the `declare` block because it's not a type here but let me know if I'm wrong.

kevinthecheung (approved):

packages/app-check/src/internal-api.ts
hsubox76: I missed this comment in the last review. Addressed now.

packages/app-check/src/api.ts
```diff
@@ -52,10 +60,29 @@ export function activate(
   }
 
   const newState: AppCheckState = { ...state, activated: true };
+
+  // Read cached token from storage if it exists and store it in memory.
+  newState.cachedTokenPromise = readTokenFromStorage(app).then(cachedToken => {
+    if (cachedToken && isValid(cachedToken)) {
+      newState.token = cachedToken;
+    }
+    return cachedToken;
```
Feiyang1, line 69: You should use `getState()` and `setState()` since it's async, the `newState` here could already be outdated.

packages/app-check/src/index.ts
Feiyang1: Sorry, wrong place. I meant in the `declare` block.

packages/app-check/src/internal-api.ts
```diff
@@ -65,7 +65,7 @@ export async function getToken(
    */
   if (!token) {
     // readTokenFromStorage() always resolves. In case of an error, it resolves with `undefined`.
```
Feiyang1, line 67: comment needs updating.

Feiyang1 (changes requested):

packages/app-check/src/api.ts
hsubox76: Done.

packages/app-check/src/internal-api.ts
hsubox76: Done.

Feiyang1 (approved):

PR #4905, opened 2021-05-13
Added es5 builds for firestore v9 and firestore lite

Feiyang1:


Feiyang1:
@hsubox76 FYI, I didn't run into out of memory issues.

packages/firestore/rollup.config.exp.js
```diff
@@ -126,6 +125,22 @@ const allBuilds = [
...
+  {
+    input: path.resolve('./exp', pkg['browser']),
+    output: [
+      {
+        file: path.resolve('./exp', pkg['esm5']),
+        format: 'es',
+        sourcemap: true
+      }
+    ],
+    plugins: util.es2017ToEs5Plugins(/* mangled= */ false),
```
schmidt-sebastian, line 138: What is this build used by? If this is a browser build, then we should turn on mangling.

Feiyang1: By people who need to support IE and don't want to compile library code.
The es2017 build which is the source is mangled, so we don't need to mangle the second time.

schmidt-sebastian: I think you will get size savings if you turn on mangling here. It runs terser again. I am no longer sure what the exact numbers for the savings were, but all other builds follow this same pattern (mangle=true for both browser builds and off for Node). If you don't see any difference in size then you can merge as is.

schmidt-sebastian (approved):
LGTM, but might want to do another size test for mangling (I can also run it later today)

packages/firestore/rollup.config.exp.js
Feiyang1: We don't mangle es2017 builds, but mangle es5 builds for the prod SDK. - https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/rollup.config.browser.js#L55-L75

Feiyang1: no mangle -> mangle 
main: 855343 -> 749475
lite: 262207 -> 231834

They does get smaller.

PR #4908, opened 2021-05-14
Increase v8 heap size to 4GB for health metric tests.

yifanyang:


yifanyang:
@Feiyang1 The first commit sets the heap size to 4GB, all the other commits are empty commits pushed just to trigger CI tests. All health metric test runs for this PR passed: https://github.com/firebase/firebase-js-sdk/actions/workflows/health-metrics-test.yml. This should be able to unblock us for now.

Feiyang1 (approved):

PR #4911, opened 2021-05-14
Integration tests for v9 & v9 compat interop

Feiyang1:


Feiyang1:
@avolkovi As promised

jposuna (approved):
LGTM for perf. Thanks for making these changes!

sam-gc (approved):
LGTM for Auth

packages/component/src/provider.ts
```diff
@@ -99,11 +99,16 @@ export class Provider<T extends Name> {
...
     optional?: boolean;
   }): NameServiceMapping[T] | null {
-    const { identifier, optional } = {
+    let { identifier, optional } = {
       identifier: DEFAULT_ENTRY_NAME,
       optional: false,
       ...options
     };
+
+    if (!identifier) {
```
hsubox76, line 108: Is line 102 not working here, does an undefined/null identifier in the `...options` spread overwrite the default in 103? If so, maybe it's better to be explicit just using this `if` block here for identifier, and a similar block for `optional`, instead of the spread extend magic? Up to you though.

hsubox76 (approved):

packages/component/src/provider.ts
Feiyang1: updated with the suggestion.

zwu52:
LGTM. Thanks for making 👍

zwu52 (approved):

PR #4915, opened 2021-05-15
Fix auth popup positioning

dmitri-gb:
The goal is to center the popup on the screen, but due to a mix-up of Math.min vs Math.max it always ends up at (0, 0) instead.


dmitri-gb:
@googlebot I signed it!

sam-gc (approved):
Looks good, thanks!

dmitri-gb:
Fixed in https://github.com/firebase/firebase-js-sdk/pull/5281

PR #4918, opened 2021-05-17
Migrates Listeners Registration from `getMessaging` to Messaging's Factory Methods 🏭

zwu52:
`messaging-compat` is a compatibility layer that bridges the v8 and v9(modular) JS API. It's created through wrapping the v9 SDK to recreate the v8 API surface. The objective is to allow developer to gradually migrate to v9 without having to migrate the entire codebase. 

Now the v9 SDK registers its listeners (`onPush`, `onSubChange`, `onMessage`) in `getMessagingInWindow` and `getMessagaingInSw` which is not called in the `messaging-compat`. Without these listeners, the compat SDK can only `getToken` and not get callbacks in `onMessage` and `onBackgroundMessage` 

- PR moves the  listeners registration to factory methods.    
- PR also removes unused variables in `messaging-compat`.
- PR registers 2 components since they will both be called in `messaging-compat` as `messaging-exp` and `messaging-sw-exp`

b/188418727 for tracking

### DIFF
--- a/common/api-review/messaging-exp.api.md
+++ b/common/api-review/messaging-exp.api.md
@@ -59,11 +59,6 @@ export interface NotificationPayload {
 
 export { Observer }
 
-// Warning: (ae-internal-missing-underscore) The name "onBackgroundMessage" should be prefixed with an underscore because the declaration is marked as @internal
-//
-// @internal
-export function onBackgroundMessage(messaging: FirebaseMessaging, nextOrObserver: NextFn<MessagePayload> | Observer<MessagePayload>): Unsubscribe;
-
 // @public
 export function onMessage(messaging: FirebaseMessaging, nextOrObserver: NextFn<MessagePayload> | Observer<MessagePayload>): Unsubscribe;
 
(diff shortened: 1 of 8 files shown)


common/api-review/messaging-exp.api.md
```diff
@@ -67,6 +67,9 @@ export function onBackgroundMessage(messaging: FirebaseMessaging, nextOrObserver
 // @public
 export function onMessage(messaging: FirebaseMessaging, nextOrObserver: NextFn<MessagePayload> | Observer<MessagePayload>): Unsubscribe;
 
+// @internal
+export function _registerListeners(messaging: FirebaseMessaging): void;
```
Feiyang1, line 71: 👍

Feiyang1: On second thought, can we register the listeners in the factories, so we don't need to export this special function?

Feiyang1 (changes requested):
In `@firebase/messaging-compat`, we need to check the environment, and get the correct provider (`messaging-exp` or `messaging-sw-exp`) depending on the environment. 

https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/messaging-compat/src/registerMessagingCompat.ts#L38

zwu52:
right. Good catch!

Feiyang1 (commented):
Sorry! requesting one more change:
let's remove `onBackgroundMessage()` from the main entry point https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/messaging-exp/src/index.ts#L34,

and remove the `@internal` annotation here - https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/messaging-exp/src/api.ts#L176

Feiyang1 (approved):

PR #4925, opened 2021-05-19
Add warning when developer overrides host without specifying merge

thebrianchen:


packages/firestore/src/api/database.ts
```diff
@@ -216,14 +216,26 @@ export class Firestore
   }
 
   settings(settingsLiteral: PublicSettings): void {
+    const currentSettings = this._delegate._getSettings();
+    if (
+      !settingsLiteral.merge &&
+      currentSettings.host !== settingsLiteral.host
+    ) {
+      logWarn(
+        'The original host is being overridden. If you did not intend ' +
```
schmidt-sebastian, line 225: Maybe change to "You are overriding the original host." to make it easier for us foreigners.

thebrianchen: Done.

PR #4933, opened 2021-05-22
Simplify ObjectValue

schmidt-sebastian:
This is one of those "port of port" PRs. When I implemented iOS (https://github.com/firebase/firebase-ios-sdk/pull/8124) it became apparent that we do not need overlays anymore to implement mutable ObjectValues. There are only needed on Android, since Protobuf manipulations on Android are fairly expensive. On iOS and Web, we can directly edit the Protobuf object and can use a simpler version of ObjectValue.

### DIFF
--- a/packages/firestore/src/model/object_value.ts
+++ b/packages/firestore/src/model/object_value.ts
@@ -36,14 +36,11 @@ export interface JsonObject<T> {
  * ability to add and remove fields (via the ObjectValueBuilder).
  */
 export class ObjectValue {
-  private value: { mapValue: ProtoMapValue };
-
-  constructor(proto: { mapValue: ProtoMapValue }) {
+  constructor(private value: { mapValue: ProtoMapValue }) {
     debugAssert(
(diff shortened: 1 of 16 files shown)


packages/firestore/src/model/object_value.ts
```diff
@@ -125,138 +126,58 @@ export class ObjectValue {
...
-    return modified ? { mapValue: { fields: resultAtPath } } : null;
+    return current.mapValue!.fields!;
   }
 
   /**
-   * Builds the Protobuf that backs this ObjectValue.
-   *
-   * This method applies any outstanding modifications and memoizes the result.
-   * Further invocations are based on this memoized result.
+   * Modifies `parent_map` by adding, replacing or deleting the specified
```
thebrianchen, line 163: s/parent_map/fieldsMap?

schmidt-sebastian: Done

Issue #4235, opened 2020-12-22
Firebase.SDK_VERSION stuck at 8.0.1


lupas:
* Operating System version: Mac OS 11.1
  * Browser version: not applicable
  * Firebase SDK version: 8.2.1
  * Firebase Product: all

####  Describe the problem

```js
const firebase = require('firebase/app')
console.log(firebase.SDK_VERSION)
// Result: "8.0.1"
// Expected: "8.2.1"
```
Also see all the `.js` files in `@firebase/app/dist` contain the following:
```js
const version$1 = "8.0.1";
```
#### Steps to reproduce:

See the two lines above.

#### Relevant Code:

See the two lines above.

#### Background

We use the `SDK_VERSION` in the Firebase module for NuxtJS (https://firebase.nuxtjs.org) to load the appropriate Firebase scripts in the created service-workers (e.g. via `'https://www.gstatic.com/firebasejs/8.0.1/firebase-app.js'`)

hsubox76:
Thanks for reporting this, it's on our list to fix. It seems like we need to have `@firebase/app` get the version string dynamically somehow instead of being bundled into its code at compile time, since the `@firebase/app` package isn't necessarily published with every release.

PR #4941, opened 2021-05-26
Always release @firebase/app

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/4235. 

This is a complete fix by publishing `@firebase/app` in every release. It's a better solution than https://github.com/firebase/firebase-js-sdk/pull/4283 because it also works if people use scoped packages directly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Install Dependencies
         run: yarn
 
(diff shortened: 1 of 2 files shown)


Feiyang1:
Tested locally. Is it possible to test the GHA before merging? @hsubox76

hsubox76 (approved):

PR #4943, opened 2021-05-27
[WIP] Make Auth side effects free

Feiyang1:


packages-exp/auth-exp/index.cordova.ts
```diff
@@ -61,4 +61,13 @@ export function getAuth(app: FirebaseApp): Auth {
   });
 }
 
-registerAuth(ClientPlatform.CORDOVA);
+export function initializeAuth(app: FirebaseApp, deps?: Dependencies): Auth {
+  const provider = _getProvider(app, 'auth-exp');
```
sam-gc, line 65: The body of these `initializeAuth()` functions are all the same except for the platform. Would it be worth moving this to `_initializeAuth` with extra parameters?

i.e.:

```js
return _initializeAuth(app, deps, provider, ClientPlatform.CORDOVA)
```

Feiyang1: Good idea. Updated.

sam-gc (approved):

PR #4963, opened 2021-06-01
[Auth] Make anonymous sign in wait for auth to finish initializing

sam-gc:
This fixes b/189879344


Feiyang1 (approved):
Nice! my test project have thousands of anonymous user because of it.

Issue #4953, opened 2021-05-28
Batched writes failing in 8.6.2


stucker71:
### [REQUIRED] Describe your environment

Firebase hosting
Node v14.15.1

### [REQUIRED] Describe the problem
A simple batch write, like that shown in your online example below will hang, no error message.

Our app was working fine in all versions prior to installing 8.6.2

#### Relevant Code:

```
// Get a new write batch
var batch = firebase.firestore().batch();

// Set the value of 'NYC'
var nycRef = firebase.firestore().collection("cities").doc("NYC");
batch.set(nycRef, {name: "New York City"});



// Commit the batch
batch.commit().then(() => {
    // ...
});
```

looptheloop88:
Hi @stucker71, I wasn't able to replicate the issue. I tried all 8.6.+ versions and worked fine. The field was added to the document and the success callback in `batch.commit()` was called.

Would you mind sharing a [minimal, but complete](https://stackoverflow.com/help/minimal-reproducible-example) sample of a project that I can run locally to replicate the issue?

stucker71:
It's possible this issue is related to https://github.com/firebase/firebase-js-sdk/issues/4932

If you execute the code:

` firebase.auth().setPersistence(firebase.auth.Auth.Persistence.LOCAL)
      .then(() => {
        firebase.auth().signInAnonymously()
          .then(() => {

            console.info("batch started");

            // Get a new write batch
            var batch = firebase.firestore().batch();

            // Set the value of 'NYC'
            var nycRef = firebase.firestore().collection("cities").doc("NYC");
            batch.set(nycRef, { name: "New York City" });

            // Commit the batch
            batch.commit().then(() => {
              console.info("batch done");
            });

          });
      });`

It will hang.

stucker71:
Please try the following code, it will fail with 8.6.3. It will return "batch started" but not "batch done" :

```
firebase.auth().signOut()
      .then(() => {

        firebase.auth().signInAnonymously()
          .then(() => {

            console.info("batch started");

            // Get a new write batch
            var batch = firebase.firestore().batch();

            // Set the value of 'NYC'
            var nycRef = firebase.firestore().collection("cities").doc("NYC");
            batch.set(nycRef, { name: "New York City" });

            // Commit the batch
            batch.commit().then(() => {
              console.info("batch done");
            });
          });
      });
```

I'm using the following basic rules:

```
rules_version = '2';
service cloud.firestore {
[...]

PR #4971, opened 2021-06-03
Fire onInit callback if instance is already available

### LINKED ISSUE #4932: 8.6.2 breaks auth/firestore rules

  * Operating System version: Mac OS X
  * Browser version: Chrome
  * Firebase SDK version: 8.6.2
  * Firebase Product: auth

After upgrading from firebase 8.6.1 → 8.6.2, I get the "ERROR FirebaseError: Missing or insufficient permissions." error when trying to access protected docs/collections. Rolling back to 8.6.1 fixes the issue.

#### Steps to reproduce:

Upgrade from Firebase 8.6.1 → 8.6.2
My app is running on Angular 12

schmidt-sebastian:
This changes Firestore's initialization to detect Auth if Auth is available during client startup. Right now, we only detect once the `setTimeout` call fires.

Fixes https://github.com/firebase/firebase-js-sdk/issues/3218
Fixes https://github.com/firebase/firebase-js-sdk/issues/4953
Fixes #4932


packages/component/src/provider.ts
```diff
@@ -263,9 +263,15 @@ export class Provider<T extends Name> {
    *
    * @returns a function to unregister the callback
    */
-  onInit(callback: OnInitCallBack<T>): () => void {
+  onInit(callback: OnInitCallBack<T>, identifier?: string): () => void {
```
Feiyang1, line 266: Can you please add a comment for the argument `identifier`?

Feiyang1 (approved):

packages/component/src/provider.ts
Feiyang1: Should we trigger the callback for all instances that have already been created? That's what would happen if those were created after the onInit callback is registered. So that we have a consistent behavior no matter whether the callback is registered before or after initialization (it will be invoked for every instances created).

Feiyang1 (changes requested):

packages/component/src/provider.ts
schmidt-sebastian: As discussed offline, I think the current behavior already achieves this. All callbacks that are registered for an existing component will be invoked once when registered (new logic). All callbacks that are registered before a component is available will be invoked once the component is available (existing logic). 

Let me know if there are any more questions.

Feiyang1: Sorry I wasn't clear. What I meant is we don't need the `identifier` argument, and a callback can be invoked multiple times (once per identifier) when registered if the component support multiple instances and there are already multiple instances initialized.

packages/component/src/provider.ts
```diff
@@ -261,11 +261,18 @@ export class Provider<T extends Name> {
...
    * The function is invoked SYNCHRONOUSLY, so it should not execute any longrunning tasks in order to not block the program.
    *
+   * @param identifier An optional instance identifier
    * @returns a function to unregister the callback
    */
-  onInit(callback: OnInitCallBack<T>): () => void {
+  onInit(callback: OnInitCallBack<T>, identifier?: string): () => void {
     this.onInitCallbacks.add(callback);
 
+    const instance = this.getImmediate({ identifier, optional: true });
```
Feiyang1, line 270: Can you please add a comment that OnInit() is only intended to be used for `EXPLICIT` components, so getImmediate() will never initialize a new instance?

packages/component/src/provider.ts
Feiyang1: as per discussion offline, we will change the callback to be associated with an identifier.

packages/component/src/provider.ts
```diff
@@ -261,13 +261,24 @@ export class Provider<T extends Name> {
...
    */
-  onInit(callback: OnInitCallBack<T>): () => void {
-    this.onInitCallbacks.add(callback);
+  onInit(callback: OnInitCallBack<T>, identifier?: string): () => void {
+    const normalizedIdentifier = this.normalizeInstanceIdentifier(identifier);
+    const existingInstances =
+      this.onInitCallbacks.get(normalizedIdentifier) ??
+      new Set<OnInitCallBack<T>>();
+    existingInstances.add(callback);
+    this.onInitCallbacks.set(normalizedIdentifier, existingInstances);
```
Feiyang1, line 273:
```suggestion
    const existingCallbacks =
      this.onInitCallbacks.get(normalizedIdentifier) ??
      new Set<OnInitCallBack<T>>();
    existingCallbacks.add(callback);
    this.onInitCallbacks.set(normalizedIdentifier, existingCallbacks);
```

packages/component/src/provider.ts
```diff
@@ -261,13 +261,24 @@ export class Provider<T extends Name> {
...
+
+    const existingInstance = this.instances.has(normalizedIdentifier);
+    if (existingInstance) {
+      callback(existingInstance, normalizedIdentifier);
+    }
 
     return () => {
-      this.onInitCallbacks.delete(callback);
+      this.onInitCallbacks.get(normalizedIdentifier)!.delete(callback);
```
Feiyang1, line 281: How about 
```suggestion
      existingCallbacks.delete(callback);
```

packages/component/.run/All Tests.run.xml
```diff
@@ -0,0 +1,19 @@
+<component name="ProjectRunConfigurationManager">
```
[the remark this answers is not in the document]
schmidt-sebastian: Okay - these are pretty helpful for me though :)

packages/component/src/provider.test.ts
```diff
@@ -181,6 +181,46 @@ describe('Provider', () => {
       expect(callback2).to.have.been.calledOnce;
     });
 
+    it('invokes callback for existing component', () => {
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

packages/component/src/provider.ts
schmidt-sebastian: Uhm, clever :) Thanks.

Feiyang1 (approved):
pending CI.

PR #4974, opened 2021-06-03
create rollup typings for messaging

Feiyang1:
We need rollup typings to generate overloads for interop in messaging-compat.

### DIFF
--- a/packages-exp/messaging-exp/package.json
+++ b/packages-exp/messaging-exp/package.json
@@ -7,7 +7,7 @@
   "main": "dist/index.cjs.js",
   "browser": "dist/index.esm2017.js",
   "module": "dist/index.esm2017.js",
-  "typings": "./dist/index-public.d.ts",
+  "typings": "./dist/index.d.ts",
   "sw": "dist/index.sw.esm2017.js",
   "files": [
     "dist",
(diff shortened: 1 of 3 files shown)


packages-exp/messaging-exp/package.json
```diff
@@ -18,15 +18,18 @@
...
     "dev": "rollup -c -w",
     "test": "run-p test:karma type-check lint ",
     "test:integration": "run-p test:karma type-check lint && cd ../../integration/messaging && npm run-script test",
     "test:ci": "node ../../scripts/run_tests_in_ci.js",
     "test:karma": "karma start --single-run",
     "test:debug": "karma start --browsers=Chrome --auto-watch",
-    "api-report": "api-extractor run --local --verbose",
-    "type-check": "tsc --noEmit"
+    "api-report": "yarn api-report:rollup && yarn api-report:api-json",
+    "api-report:rollup": "TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' ts-node ../../repo-scripts/prune-dts/extract-public-api.ts --package messaging-exp --packageRoot . --typescriptDts ./dist/index.d.ts --rollupDts ./dist/private.d.ts --untrimmedRollupDts ./dist/internal.d.ts --publicDts ./dist/index-public.d.ts",
```
hsubox76, line 29: Just out of curiosity, when do we use ts-node and ts-node-script?

hsubox76 (approved):

zwu52 (approved):

packages-exp/messaging-exp/package.json
Feiyang1: `ts-node-script` reads the tsconfig from the source directory where the script lives, while `ts-node` loads tsconfig from the cwd. This behavior has changed in [v10](https://github.com/TypeStrong/ts-node/releases/tag/v10.0.0) where `ts-node` behaves the same way as `ts-node-script` by default. (we haven't updated to v10 yet)

 In most cases, you'd want to use `ts-node-script` because you want to use the tsconfig of the script.

Here, we should really use ts-node-script, but it won't matter once we upgrade to v10, so I will keeping it as is for now.

Feiyang1: Well, ended up updating to `ts-node-script`

PR #5002, opened 2021-06-09
Move from manual JWT creation to jsonwebtoken

abeisgoat:
In rules-unit-testing, our naive implementation of unsigned JWT implementation wasn't spec compliant so it was breaking the storage emulator


packages/rules-unit-testing/package.json
```diff
@@ -20,10 +20,11 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "firebase": "8.6.5",
     "@firebase/component": "0.5.1",
     "@firebase/logger": "0.2.6",
     "@firebase/util": "1.1.0",
+    "firebase": "8.6.5",
+    "jsonwebtoken": "^8.5.1",
```
avolkovi, line 27: can you use a jose library instead? something like node-jose

abeisgoat: Any reason why? We use `jsonwebtoken` throughout the emulator suite and it's much more used than `node-jose` (200k download/mo vs 6m)

avolkovi: I guess for emulator that's ok, the node-jose library also implements JWK which is helpful when you actually want to verify the signature, but you don't need that in this case

avolkovi (approved):

samtstern:
@abeisgoat looks like the rules have begun failing, which probably means the token is not encoded correctly:

```
INFO: operation failed: 
Property custom_claim is undefined on object. for 'update' @ L3
com.google.cloud.datastore.core.exception.DatastoreException: 
Property custom_claim is undefined on object. for 'update' @ L3
	at com.google.cloud.datastore.core.exception.DatastoreException$Builder.build(DatastoreException.java:111)```

**Comment by alejo4373** [comment]
Hi All! Any updates on this? I'm running into [firebase-tools/issues/3442](https://github.com/firebase/firebase-tools/issues/3442)

**Review by yuchenshi — CHANGES_REQUESTED** [review]
BLOCKING: We also need to update https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/emulator.ts#L138

I suggest we just trim the padding instead of introducing another dependency since we don't want to bloat @firebase/util which is used in almost all firebase packages. Plus it's a much smaller change and easier to reason about, especially given the broken decoding right now.

File: packages/rules-unit-testing/src/api/index.ts [hunk 2]
```diff
@@ -197,11 +197,7 @@ function createUnsecuredJwt(token: TokenOptions, projectId?: string): string {
 
   // Unsecured JWTs use the empty string as a signature.
   const signature = '';
-  return [
-    base64.encodeString(JSON.stringify(header), /*webSafe=*/ false),
```
yuchenshi, line 201: This needs use base64url instead of base64. https://github.com/firebase/firebase-js-sdk/blob/5d31e2192d0ea68a768bc7826ad5aa830c2bc36c/packages/util/src/crypt.ts#L339 is the utility for that. I think we still need to trim padding ourselves though.

abeisgoat:
@yuchenshi plz take a look and I'm not sure why things are failing after the `package.json` merge, any idea what that token error is?

packages/rules-unit-testing/src/api/index.ts
```diff
@@ -157,6 +158,10 @@ export type FirebaseEmulatorOptions = {
   };
 };
 
+function trimmedBase64Encode(val: string): string {
+  return base64Encode(val).split("=")[0];
```
yuchenshi, line 162: Our utility library uses `.` as the padding character for URL-safe alphabet, causing the error you saw. Suggested fix:

```suggestion
  // Use base64url encoding and remove padding in the end (dot characters).
  return base64Encode(val).replace(/\./g,"");
```

yuchenshi (approved):
Let's get this in first and then in a later PR I'll port the fix to https://github.com/firebase/firebase-js-sdk/blob/master/packages/util/src/emulator.ts#L138

yuchenshi:
Fixes https://github.com/firebase/firebase-tools/issues/3442.

This will be included in the next release.

PR #5008, opened 2021-06-10
Block on Auth synchronously

schmidt-sebastian:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -22,7 +22,7 @@ import {
 import { Provider } from '@firebase/component';
 
 import { User } from '../auth/user';
-import { hardAssert, debugAssert } from '../util/assert';
+import { debugAssert, hardAssert } from '../util/assert';
 import { AsyncQueue } from '../util/async_queue';
 import { Code, FirestoreError } from '../util/error';
 import { logDebug } from '../util/log';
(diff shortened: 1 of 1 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -217,18 +215,17 @@ export class FirebaseCredentialsProvider implements CredentialsProvider {
     this.tokenListener = () => {
       this.tokenCounter++;
       this.currentUser = this.getUser();
-      this.receivedInitialUser.resolve();
-      if (this.invokeChangeListener) {
+      this.authDeferred.resolve();
+      if (this.changeListener) {
         this.asyncQueue!.enqueueRetryable(() =>
-          this.changeListener(this.currentUser)
+          this.changeListener!(this.currentUser)
```
Feiyang1, line 221: `changeListener` is pushed to the `AsyncQueue` when it's registered. So for the first user, it seems it will be called twice. Is it a problem? Perhaps we still need the `invokeChangeListener` flag.

schmidt-sebastian: I don't think this should happen: `setCredentialListener` sets the `changeListener` but only once the `authDeferred` promise resolves. In the initial token callback the `changeListener` will be undefined

Feiyang1 (approved):

Issue #5012, opened 2021-06-10
Stop using const enums in the public API


Feiyang1:
Some build tools (e.g. esbuild, vite) don't work with const enums because they compile Typescript in the `isolatedModules` mode. They compile files individually without type information, so they can't inline const enums, which eventually cause runtime errors.

Plus Javascript users can't use const enums.

Auth exports some const enums, e.g. `ProviderId`, `SignInMethod`. We should probably replace them with just plain object maps.

BayanBennett:
Also, NextJS requires `isolatedModules` due to the way that they have `babel` configured.

Thank you for this change.

PR #5022, opened 2021-06-14
Remove all public const enums; export maps instead (with typings to strings)

sam-gc:
This PR removes public `const enum`s (which are inaccessible to JS anyway) and moves any (public) reference to those types to simply `string`. Instead, we export maps that can be used to reference acceptable values (but these are not used internally to aid tree shaking)

Fixes https://github.com/firebase/firebase-js-sdk/issues/5012

### DIFF
--- a/common/api-review/auth-exp.api.md
+++ b/common/api-review/auth-exp.api.md
@@ -25,12 +25,12 @@ export interface ActionCodeInfo {
 
 // @public
 export const ActionCodeOperation: {
-    EMAIL_SIGNIN: string;
-    PASSWORD_RESET: string;
-    RECOVER_EMAIL: string;
-    REVERT_SECOND_FACTOR_ADDITION: string;
-    VERIFY_AND_CHANGE_EMAIL: string;
-    VERIFY_EMAIL: string;
+    readonly EMAIL_SIGNIN: "EMAIL_SIGNIN";
+    readonly PASSWORD_RESET: "PASSWORD_RESET";
+    readonly RECOVER_EMAIL: "RECOVER_EMAIL";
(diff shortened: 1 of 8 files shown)


packages-exp/auth-exp/src/model/enum_maps.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * An enum of factors that may be used for multifactor authentication.
+ *
+ * @public
+ */
+export const FactorId = {
+  /** Phone as second factor */
+  PHONE: 'phone'
+};
```
yuchenshi, line 26: Is it worth declaring `{...} as const`? This will make `PHONE` typed as `"PHONE"` (string literal) instead of string, which may be a bit more specific. Ditto for below.

yuchenshi (approved):

packages-exp/auth-exp/src/model/enum_maps.ts
sam-gc: Good idea, done

yuchenshi (approved):

common/api-review/auth-exp.api.md
```diff
@@ -20,18 +20,18 @@ export interface ActionCodeInfo {
         multiFactorInfo?: MultiFactorInfo | null;
         previousEmail?: string | null;
     };
-    operation: ActionCodeOperation;
+    operation: string;
```
Feiyang1, line 23: We can use string literal unions to make it more type safe, using something like 
```
typeof ActionCodeOperation[keyof typeof ActionCodeOperation]
```

We can apply the same to other places.

sam-gc: Done (where appropriate)

`ProviderId` and `SignInMethod` should remain `string`. Updated the others.

Feiyang1 (approved):

Issue #4937, opened 2021-05-25
modular SDK throws an error in rxjs


uny:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 10.15.7 (19H1030)
  * Browser version: Chrome 90.0.4430.212 (Official Build) (x86_64)
  * Firebase SDK version: 9.0.0-beta.2
  * Firebase Product: auth

#### Steps to reproduce:

With rxjs@7.1.0, I get a runtime error:
```
Uncaught (in promise) TypeError: Cannot read property 'isStopped' of undefined
    at Subscriber.next (Subscriber.js?4701:29)
    at newNext (index.esm2017.js?5994:823)
    at eval (index-02c0b8e8.js?43e4:2203)
```

#### Relevant Code:

moduler:
```javascript
import { getApp, getApps, initializeApp } from "firebase/app";
import { getAuth, onAuthStateChanged } from "firebase/auth";
import { Observable } from "rxjs";

const firebase = (() => {
  if (getApps().length === 0) {
    return initializeApp({/* OPTIONS */});
  } else {
    return getApp();
  }
})();
new Observable(
  (observer) => {
    const unsubscribe = onAuthStateChanged(getAuth(firebase), observer);
    return { unsubscribe };
  }
).subscribe(console.log);
```

compat:
```javascript
import firebase from "firebase/compat/app";
import "firebase/compat/auth";
import { getApps } from "firebase/app";
import { Observable } from "rxjs";

if (getApps().length === 0) {
  firebase.initializeApp({/* OPTIONS */});
}

new Observable(
  (observer) => {
    const unsubscribe = firebase.auth().onAuthStateChanged(observer);
    return { unsubscribe };
  }
).subscribe(console.log);
```

Both of them throw the error.

#### Additional Information
If I change the code like below, the error will be solved.
```
const unsubscribe = onAuthStateChanged(getAuth(firebase), (user) => {
  observer.next(user);
});
```

If I use firebase@8.6.2 and rxjs@7.1.0, the error also does not occur.

jamesdaniels:
FWIW we ran into this in RxFire, here's the work-around we used in `@exp` https://github.com/FirebaseExtended/rxfire/blob/exp/auth/index.ts#L34

Feiyang1:
@jamesdaniels Have you had a chance to debug? Any chance you already know the root cause?

uny:
In fact, I am using rxfire@4.0.0 and found the problem.
[...]

PR #5033, opened 2021-06-16
Add public token API to AppCheck

hsubox76:
Create public `getToken()` and `onTokenChanged()` methods.

Proposal (internal link):
go/app-check-sdk-get-token-3p-api

For docs reviewers:
The only public doc comments are in `packages/firebase/index.d.ts`

### DIFF
--- a/packages/app-check-interop-types/index.d.ts
+++ b/packages/app-check-interop-types/index.d.ts
@@ -24,10 +24,10 @@ export interface FirebaseAppCheckInternal {
   // registered at the same time for one or more FirebaseAppAttestation instances. The
   // listeners call back on the UI thread whenever the current token associated with this
   // FirebaseAppAttestation changes.
-  addTokenListener(listener: (token: AppCheckTokenResult) => void): void;
+  addTokenListener(listener: AppCheckTokenListener): void;
 
   // Unregisters a listener to changes in the token state.
-  removeTokenListener(listener: (token: AppCheckTokenResult) => void): void;
+  removeTokenListener(listener: AppCheckTokenListener): void;
 }
 
 type AppCheckTokenListener = (token: AppCheckTokenResult) => void;
--- a/packages/app-check-interop-types/index.d.ts
+++ b/packages/app-check-interop-types/index.d.ts
@@ -24,13 +24,16 @@ export interface FirebaseAppCheckInternal {
   // registered at the same time for one or more FirebaseAppAttestation instances. The
   // listeners call back on the UI thread whenever the current token associated with this
   // FirebaseAppAttestation changes.
-  addTokenListener(listener: AppCheckTokenListener): void;
+  addTokenListener(listener: (token: AppCheckTokenResult) => void): void;
 
   // Unregisters a listener to changes in the token state.
-  removeTokenListener(listener: AppCheckTokenListener): void;
+  removeTokenListener(listener: (token: AppCheckTokenResult) => void): void;
 }
 
-type AppCheckTokenListener = (token: AppCheckTokenResult) => void;
+interface AppCheckTokenListener {
+  listener: (token: AppCheckTokenResult) => void;
+  onError?: (error: Error) => void;
+}
 
 // If the error field is defined, the token field will be populated with a dummy token
 interface AppCheckTokenResult {
--- a/packages/app-check-interop-types/index.d.ts
+++ b/packages/app-check-interop-types/index.d.ts
@@ -30,10 +30,7 @@ export interface FirebaseAppCheckInternal {
   removeTokenListener(listener: (token: AppCheckTokenResult) => void): void;
 }
 
-interface AppCheckTokenListener {
-  listener: (token: AppCheckTokenResult) => void;
-  onError?: (error: Error) => void;
-}
+type AppCheckTokenListener = (token: AppCheckTokenResult) => void;
 
 // If the error field is defined, the token field will be populated with a dummy token
 interface AppCheckTokenResult {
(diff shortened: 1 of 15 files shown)


packages/app-check-types/index.d.ts
```diff
@@ -36,6 +38,40 @@ export interface FirebaseAppCheck {
    * during `activate()`.
    */
   setTokenAutoRefreshEnabled(isTokenAutoRefreshEnabled: boolean): void;
+
+  /**
+   * Get the current App Check token. Attaches to the most recent
+   * in-flight request if one is present. Returns null if no token
+   * is present and no token requests are in-flight.
```
kevinthecheung, line 45:
```suggestion
   * is present and no token requests are in flight.
```

kevinthecheung (approved):

packages/app-check/src/api.ts
```diff
@@ -82,3 +92,72 @@ export function setTokenAutoRefreshEnabled(
...
+   * It is added only for API consistency with the observer pattern, which
+   * we follow in JS APIs.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onCompletion?: () => void
+): Unsubscribe {
+  let nextFn: NextFn<AppCheckTokenResult> = () => {};
+  let errorFn: ErrorFn = () => {};
+  if ((onNextOrObserver as PartialObserver<AppCheckTokenResult>).next != null) {
+    nextFn = (onNextOrObserver as PartialObserver<AppCheckTokenResult>).next!;
```
Feiyang1, line 150: We need to bind the functions (next, error) to `onNextOrObserver`, otherwise we will trigger errors with observers created from rxjs.
see https://github.com/firebase/firebase-js-sdk/issues/4937

Feiyang1: Is the `!` necessary? Typscript should know it's not null inside the if block.

packages/app-check/src/internal-api.ts
```diff
@@ -306,9 +326,13 @@ function notifyTokenListeners(
 
   for (const listener of listeners) {
     try {
-      listener(token);
+      listener.listener(token);
```
Feiyang1, line 329: This feels weird with 2 `listener`s. Can you rename the field name to something different.  Perhaps we should just store observers in `state. tokenListeners` instead of `AppCheckTokenListenerInternal`s.

packages/app-check/src/api.test.ts
```diff
@@ -86,4 +96,137 @@ describe('api', () => {
...
+      );
+
+      expect(getState(app).tokenListeners.length).to.equal(2);
+
+      await getToken(app, fakePlatformLoggingProvider);
+
+      expect(listener2).to.be.calledWith({
+        token: fakeRecaptchaAppCheckToken.token
+      });
+      expect(errorFn1).to.be.calledOnce;
```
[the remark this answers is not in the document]
hsubox76: This should be fixed. The error callback will be called if getToken fails, with the error field from getToken.

packages/app-check-interop-types/index.d.ts
```diff
@@ -24,10 +24,10 @@ export interface FirebaseAppCheckInternal {
   // registered at the same time for one or more FirebaseAppAttestation instances. The
   // listeners call back on the UI thread whenever the current token associated with this
   // FirebaseAppAttestation changes.
-  addTokenListener(listener: AppCheckTokenListener): void;
+  addTokenListener(listener: (token: AppCheckTokenResult) => void): void;
 
   // Unregisters a listener to changes in the token state.
-  removeTokenListener(listener: AppCheckTokenListener): void;
+  removeTokenListener(listener: (token: AppCheckTokenResult) => void): void;
```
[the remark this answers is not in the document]
hsubox76: Leftover from another approach, fixed.

packages/app-check/src/api.ts
hsubox76: 1) Binding: done.
2) I don't understand why Typescript complains without the `!` but it does.

packages/app-check/src/internal-api.ts
hsubox76: I converted these to observers.

packages/app-check/src/internal-api.ts
```diff
@@ -264,7 +277,11 @@ function createTokenRefresher(
 
       // getToken() always resolves. In case the result has an error field defined, it means the operation failed, and we should retry.
       if (result.error) {
-        throw result.error;
+        if (onError) {
+          onError(result.error);
```
Feiyang1, line 281: This is redundant. Notifying listeners are already handled in `getToken()` which calls `notifyTokenListeners()`.

Here we should always throw to trigger retries. (Just revert the changes)

Feiyang1 (changes requested):

packages/app-check/src/api.test.ts
```diff
@@ -86,4 +97,171 @@ describe('api', () => {
...
+        app,
+        fakePlatformLoggingProvider,
+        true
+      );
+    });
+    it('getToken() throws errors returned with token', async () => {
+      const app = getFakeApp({ automaticDataCollectionEnabled: true });
+      const fakePlatformLoggingProvider = getFakePlatformLoggingProvider();
+      stub(internalApi, 'getToken').resolves({
+        token: 'a-token-string',
```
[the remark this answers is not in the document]
hsubox76: Clarified.

packages/app-check/src/internal-api.ts
hsubox76: Reverted.

Feiyang1 (approved):

Issue #4993, opened 2021-06-07
Firestore get() returns "Could not reach Cloud Firestore backend" error


hsubox76:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS Big Sur 11.3.1
  * Browser version: Chrome, Firefox
  * Firebase SDK version: 9.0.0-beta.3
  * Firebase Product: firestore

Starting in beta.3, Firestore `get()` has a timeout error:
```
@firebase/firestore: Firestore (9.0.0-beta.3): Could not reach Cloud Firestore backend. Backend didn't respond within 10 seconds.
This typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.
```

Sample code to reproduce (this is compat, but using standard exp functions has the same result):
```
import firebase from "firebase/compat/app";
import "firebase/compat/firestore";

async function main() {
  const app = firebase.initializeApp({ ... });
  const firestore = firebase.firestore();
  const response = await firestore.collection("testCollection").doc("trueDoc").get();
  console.log(response.data());
}
```

Firestore logs:
https://paste.googleplex.com/5208332383027200

This doesn't happen in beta.2. It may have been caused by replacing XHR with fetch in https://github.com/firebase/firebase-js-sdk/pull/4939.

etiennestuder:
I've noticed the same behvior (not working with beta3, working with beta2).

atanaskanchev:
Seconded. I have the same issue with v9b3

vabatta:
@etiennestuder so also beta2 has the issue?
I noticed it with beta3 and thought of downgrade.

etiennestuder:
@vabatta works with beta2

steinn:
I still have the same issue using the v9 beta3 and beta4 versions. I've tried both compat and exp api. 

```typescript
const app = initializeApp({...});

async function main() {
  const docRef = doc(getFirestore(), "Test", "test");
  const data = await getDoc(docRef);
  console.log(data.data());
}
```

```
@firebase/firestore: Firestore (9.0.0-beta.4): Could not reach Cloud Firestore backend. Backend didn't respond within 10 seconds.
This typically indicates that your device does not have a healthy Internet connection at the moment. The client will operate in offline mode until it is able to successfully connect to the backend.
```

TrustyTechSG:
My project having the same issue with v3 and v4, everything works fine on v2.

PR #4994, opened 2021-06-08
Fix FetchXMLHTTPFactory

schmidt-sebastian:
Upgrade closure library to fix networking in firestore v9

Fixes https://github.com/firebase/firebase-js-sdk/issues/4993

PR #5046, opened 2021-06-18
Revert to use Xhr in Firestore

Feiyang1:
[The fix](https://github.com/firebase/firebase-js-sdk/pull/4994) doesn't seem to fix https://github.com/firebase/firebase-js-sdk/issues/4993


schmidt-sebastian:
Hm. I can debug more next week but we should merge this for now then.

schmidt-sebastian (approved):

Issue #5052, opened 2021-06-20
Firebase app check 429 errors


febg11:
I am trying to use firebase app-check but I constantly receive error 429. I can't see how I am exhausting the server. If i load it and refresh the page once they show up. 

The networks requests are calling the following api 

```
https://content-firebaseappcheck.googleapis.com/v1beta/projects
```

(I have removed some project specific get params)

There seems to be other people experiencing this:

https://stackoverflow.com/questions/67982348/firebase-appcheck-web-app-not-working-in-debug-mode

steurt:
I experience the same issue. In my case the `exchangeDebugToken` is requested multiple times in close succession.
`https://content-firebaseappcheck.googleapis.com/v1beta/projects/{projectId}/apps/{appId}:exchangeDebugToken?key={apiKey}`

Resulting in the following error:
```
FirebaseError: AppCheck: Fetch server returned an HTTP error status. HTTP status: 429. (appCheck/fetch-status-error)
```

When the request does succeed, I get the response containing the `attestationToken` and the `ttl` of `3600s`. Seeing that the request is executed multiple it seems the ttl might not be considered correctly.

hsubox76:
Can you describe how you are using App Check? Are you using debug mode like the Stack Overflow question or something else?

febg11:
I am using the debug mode yes. I am also using angular and I am initialising it like the chosen answer in this stackoverflow question

https://stackoverflow.com/questions/67908846/how-to-initialise-firebase-app-check-in-angular

I am using it with Functions, Storage and Realtime Database and they all throw the errors. Occassionally some requests get through and say a storage getDownloadLnk requests works but 99% of the time there will be at least 1 429 error when the page is loaded. 

I have added the debug token generated by the library (also that was hard to find as it only seems to print once then stops logging it as its in local storage.. had to debug the minified code to get it) 

Any help is really appreciated as this is blocking my local development.

hsubox76:
I think this may be caused because if you are in debug mode, the App Check token is not cached, and instead it hits the App Check exchange endpoint on every request. That would be a bug on our part and I'll look into fixing it.

febg11:
Thanks a lot 😄

febg11:
Hi is there any update with this?

hsubox76:
This has just been merged. It missed the cutoff for this week's release but should be released next week.

PR #5055, opened 2021-06-21
Cache App Check debug token

hsubox76:
Cache the debug token to memory and indexedDB in a separate place from the regular app check token. Otherwise it will hit the exchange endpoint on every `getToken()` call and cause 429 (too many requests) errors.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5052

### DIFF
--- a/packages/app-check/src/api.test.ts
+++ b/packages/app-check/src/api.test.ts
@@ -28,6 +28,7 @@ import {
   getFakeApp,
   getFakeCustomTokenProvider,
   getFakePlatformLoggingProvider,
+  getFakeGreCAPTCHA,
   removegreCAPTCHAScriptsOnPage
 } from '../test/util';
 import { clearState, getState } from './state';
@@ -37,8 +38,12 @@ import * as internalApi from './internal-api';
 import * as client from './client';
 import * as storage from './storage';
 import * as logger from './logger';
+import * as util from './util';
 
 describe('api', () => {
+  beforeEach(() => {
+    stub(util, 'getRecaptcha').returns(getFakeGreCAPTCHA());
+  });
   describe('activate()', () => {
     let app: FirebaseApp;
 
--- a/packages/app-check/src/api.test.ts
+++ b/packages/app-check/src/api.test.ts
@@ -16,7 +16,7 @@
  */
 import '../test/setup';
 import { expect } from 'chai';
-import { stub, spy, SinonStub } from 'sinon';
+import { stub, spy } from 'sinon';
 import {
   activate,
   setTokenAutoRefreshEnabled,
@@ -131,8 +131,6 @@ describe('api', () => {
     });
   });
   describe('onTokenChanged()', () => {
-    let storageReadStub: SinonStub;
-    let storageWriteStub: SinonStub;
     const fakePlatformLoggingProvider = getFakePlatformLoggingProvider();
     const fakeRecaptchaToken = 'fake-recaptcha-token';
     const fakeRecaptchaAppCheckToken = {
@@ -142,14 +140,10 @@ describe('api', () => {
     };
 
     beforeEach(() => {
-      storageReadStub = stub(storage, 'readTokenFromStorage').resolves(
-        undefined
-      );
-      storageWriteStub = stub(storage, 'writeTokenToStorage');
+      stub(storage, 'readTokenFromStorage').resolves(undefined);
+      stub(storage, 'writeTokenToStorage');
     });
     afterEach(() => {
-      storageReadStub.restore();
-      storageWriteStub.restore();
       clearState();
       removegreCAPTCHAScriptsOnPage();
     });
--- a/packages/app-check/src/api.test.ts
+++ b/packages/app-check/src/api.test.ts
@@ -16,7 +16,7 @@
  */
 import '../test/setup';
 import { expect } from 'chai';
-import { stub, spy } from 'sinon';
+import { stub, spy, SinonStub } from 'sinon';
 import {
   activate,
   setTokenAutoRefreshEnabled,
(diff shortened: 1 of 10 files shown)


packages/app-check/src/internal-api.test.ts
```diff
@@ -284,6 +284,67 @@ describe('internal api', () => {
       );
       expect(token).to.deep.equal({ token: fakeRecaptchaAppCheckToken.token });
     });
+
+    it('reads any memory-cached debug token if in debug mode', async () => {
+      const clock = useFakeTimers();
```
Feiyang1, line 289: Are faketimers needed in these tests?

packages/app-check/src/internal-api.test.ts
```diff
@@ -284,6 +284,67 @@ describe('internal api', () => {
...
+      const clientStub = stub(client, 'exchangeToken');
+      const readStub = stub(storage, 'readTokenFromStorage');
+      const debugState = getDebugState();
+      debugState.enabled = true;
+      debugState.token = new Deferred();
+      debugState.token.resolve('my-debug-token');
+      debugState.exchangeToken = fakeCachedAppCheckToken;
```
Feiyang1, line 296: `exchangeToken` doesn't sound obvious to me what it is. And having another field `token` confuses me even more. 
Should we change them to `appChekcToken` and `debugToken` respectively?

packages/app-check/src/internal-api.ts
```diff
@@ -76,10 +76,27 @@ export async function getToken(
...
+    if (debugState.exchangeToken && isValid(debugState.exchangeToken)) {
+      return { token: debugState.exchangeToken.token };
+    }
+    // Check for a indexedDB cached debug token from the exchange endpoint.
+    const cachedDebugToken = await readTokenFromStorage(app, 'debug');
+    if (cachedDebugToken && isValid(cachedDebugToken)) {
+      // Write to debug state.
+      debugState.exchangeToken = cachedDebugToken;
+      return { token: cachedDebugToken.token };
+    }
```
Feiyang1, line 90: The naming is very confusing. We are using "debug token" to mean different things here: 1. the debug token used to exchange for an appcheck token 2. the app check token exchanged from a debug token.
 But "debug token" should only mean 1.

Now thinking about it again after seeing the code,  I don't think we need to cache the appcheck token exchanged from a debug token differently. It's just an appcheck token and the backend doesn't care about how it was created as long as it's valid.

That means we just cache the appcheck token normally in the debug mode the same as in the prod mode.

Feiyang1 (changes requested):

packages/app-check/src/internal-api.ts
hsubox76: Done.

packages/app-check/src/internal-api.test.ts
hsubox76: Avoided separating them per comment below.

packages/app-check/src/internal-api.test.ts
hsubox76: I went down quite a rabbit hole of fixing flaky tests and I think hopefully I've gotten it? This is good because app check tests have been flaky for a while and I think it's important to try to get to the bottom of it.

- I removed unnecessary fake timers that only turned out to be needed because it made isValid() work with the really low `expireTimeMillis` I had (123). I set all expireTimeMillis to Date.now() + 60000.
- There were hanging promises caused by grecaptcha.ready() that caused the test suite to end early sometimes because I think karma interpreted resolved promises from a previous test as a done() of the currently running async tests. I stubbed grecaptcha when we weren't testing for it specifically
- Writing to real storage could cause inconsistent states between tests, stubbed the reads and writes by default

packages/app-check/src/internal-api.test.ts
```diff
@@ -197,35 +196,48 @@ describe('internal api', () => {
...
+      const listener2 = spy();
+
+      const errorFn1 = spy();
+
+      addTokenListener(app, fakePlatformLoggingProvider, listener1, errorFn1);
+      addTokenListener(app, fakePlatformLoggingProvider, listener2);
+
+      await getToken(app, fakePlatformLoggingProvider);
+
+      expect(errorFn1).not.to.be.called;
```
Feiyang1, line 215: Probably worth checking that listener1 is called.

packages/app-check/src/internal-api.test.ts
```diff
@@ -197,35 +196,48 @@ describe('internal api', () => {
...
 
-      stub(storage, 'readTokenFromStorage').returns(Promise.resolve(undefined));
-      stub(reCAPTCHA, 'getToken').returns(Promise.resolve(fakeRecaptchaToken));
-      stub(client, 'exchangeToken').returns(
-        Promise.resolve(fakeRecaptchaAppCheckToken)
-      );
-      const storageWriteStub = stub(storage, 'writeTokenToStorage');
+      stub(reCAPTCHA, 'getToken').resolves(fakeRecaptchaToken);
+      stub(client, 'exchangeToken').resolves(fakeRecaptchaAppCheckToken);
+      storageWriteStub.resetHistory();
```
Feiyang1, line 240: what does resetHistory do?

packages/app-check/src/internal-api.test.ts
```diff
@@ -281,11 +289,70 @@ describe('internal api', () => {
       );
       expect(token).to.deep.equal({ token: fakeRecaptchaAppCheckToken.token });
     });
+
+    it('reads any memory-cached debug token if in debug mode', async () => {
```
Feiyang1, line 293: I'm unclear what it is testing. Do you mean the memory cached appcheck token exchanged from a debug token?

I don't think we need a special test case for reading memory cached appcheck token in the debug mode since we don't treat it differently than in prod mode.

Feiyang1: Ditto for the other 2 new tests below.

packages/app-check/src/internal-api.test.ts
```diff
@@ -294,75 +361,75 @@ describe('internal api', () => {
...
       debugState.enabled = true;
       debugState.token = new Deferred();
       debugState.token.resolve('my-debug-token');
 
-      activate(app, FAKE_SITE_KEY, true);
-      addTokenListener(app, fakePlatformLoggingProvider, fakeListener);
+      activate(app, FAKE_SITE_KEY, false);
+      addTokenListener(app, fakePlatformLoggingProvider, listener);
+      await clock.runAllAsync();
+      expect(listener).to.be.calledWith({ token: 'my-debug-token' });
```
Feiyang1, line 431: The listener should get `fakeRecaptchaAppCheckToken` instead of `my-debug-token`. The [logic](https://github.com/firebase/firebase-js-sdk/blob/master/packages/app-check/src/internal-api.ts#L189-L198) in `addTokenListener` is wrong. (I think initially we thought the debug token can be used directly without exchanging for an appcheck token)

There need not to be special handling for the debug mode, and we should always create a token refresher.

Feiyang1 (changes requested):

packages/app-check/src/api.test.ts
```diff
@@ -126,25 +131,35 @@ describe('api', () => {
...
+
+    beforeEach(() => {
+      storageReadStub = stub(storage, 'readTokenFromStorage').resolves(
+        undefined
+      );
+      storageWriteStub = stub(storage, 'writeTokenToStorage');
+    });
     afterEach(() => {
+      storageReadStub.restore();
+      storageWriteStub.restore();
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/app-check/src/internal-api.test.ts
```diff
@@ -38,20 +39,28 @@ import * as reCAPTCHA from './recaptcha';
...
     app = getFakeApp();
+    storageReadStub = stub(storage, 'readTokenFromStorage').resolves(undefined);
+    storageWriteStub = stub(storage, 'writeTokenToStorage');
+    stub(util, 'getRecaptcha').returns(getFakeGreCAPTCHA());
   });
 
-  afterEach(() => {
+  afterEach(async () => {
+    storageReadStub.restore();
+    storageWriteStub.restore();
```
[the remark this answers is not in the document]
hsubox76: Removed.

packages/app-check/src/internal-api.test.ts
hsubox76: If the stub persists from test to test, it clears the call history so "calledOnce" is true even if it's called every time, but considering the stubs are all cleared and re-made after each test it's not needed here. Removed.

packages/app-check/src/internal-api.test.ts
hsubox76: Added.

packages/app-check/src/internal-api.test.ts
hsubox76: Ok, I guess it was just to make sure everything acts as expected and we don't regress if someone changes the order of the `if` checks in the future, but I guess we probably won't do that.

packages/app-check/src/internal-api.test.ts
hsubox76: Oh oops, thanks for the catch. Fixed `addTokenListener()` and removed these cases.

Feiyang1 (approved):

PR #5059, opened 2021-06-22
Auto label PRs that change docs, then auto assign tech writers

Feiyang1:


scripts/exp/detect-doc-changes.ts
```diff
@@ -0,0 +1,89 @@
...
+  await git.stash();
+
+  // generate reference docs using the headSha
+  await git.checkout(headSha);
+  await spawn('yarn', [], { cwd: tmpDir, stdio: 'inherit' });
+
+  // stage the docs generated by baseSha, so we can use git diff to tell if any docs are different in the headSha
+  await git.stash(['pop']);
+  await git.add('*');
+  await spawn('yarn', ['docgen:exp'], { cwd: tmpDir, stdio: 'inherit' });
```
hsubox76, line 57: Is there a difference between `yarn docgen:exp` and `yarn docgen:exp devsite` (line 40 above, the command used for the `baseSha` docgen)

scripts/exp/detect-doc-changes.ts
```diff
@@ -0,0 +1,89 @@
...
+    '--directory',
+    '--exclude-standard'
+  ]);
+
+  return !!diff || !!untracked;
+}
+
+// this function is supposed to run in the Github action.
+async function runInCI() {
+  const headSha = process.env.GITHUB_PULL_REQUEST_HEAD_SHA;
```
hsubox76, line 74: This way seems like it might be more robust than how check_changeset currently works (it does a `git diff --name-only origin/master...HEAD`). Maybe I should change check_changeset to this way?

scripts/exp/detect-doc-changes.ts
Feiyang1: Good catch. We should run `yarn docgen:exp devsite` here too.

scripts/exp/remove-exp.js
```diff
@@ -32,7 +32,6 @@ const {
 // can be used in command line
 if (argv._[0]) {
   const dirOrFile = path.resolve(argv._[0]);
-
```
hsubox76, line 35: Doesn't seem necessary?

.github/workflows/assign-tech-writers.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  assign_tech_writers:
+    if: ${{github.event.label.name == 'doc-changes'}}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: assign techwriters to PR
+        uses: kentaro-m/auto-assign-action@v1.2.0
+        with:
+          configuration-path: ".github/auto_assign.yml"
```
hsubox76, line 15: Do we need a task to remove tech writers if the label is removed or will we just remember to do that? We can probably remember, it's probably not very common.

hsubox76 (approved):

.github/workflows/assign-tech-writers.yml
Feiyang1: It's nice to have, but is probably rarely useful. So it's probably better to keep the workflow simple and not support removing people.

scripts/exp/remove-exp.js
Feiyang1: nop, but I don't want to trigger CI again. :)

PR #5062, opened 2021-06-23
[Auth] Fix Firefox Indexed DB workaround issue for v9. Update test to workaround same issue in v8

sam-gc:


packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
```diff
@@ -31,7 +31,7 @@ import { AuthDriver } from './auth_driver';
  */
 
 interface TempSuite {
-  generator: (driver: AuthDriver) => void;
+  generator: (driver: AuthDriver, browser?: string) => void;
```
yuchenshi, line 34:
```suggestion
  generator: (driver: AuthDriver, browser: string) => void;
```

The question mark makes the argument optional for the _caller_. The implementation can always omit trailing parameters -- there is no need to do anything. `(a) => foo` is always assignable to `(a: AuthDriver, b: string) => void`.

packages-exp/auth-exp/test/integration/webdriver/util/test_runner.ts
```diff
@@ -45,7 +45,7 @@ const SUITES: TempSuite[] = [];
 /** Main entry point for all WebDriver tests */
 export function browserDescribe(
   title: string,
-  generator: (driver: AuthDriver) => void
+  generator: (driver: AuthDriver, browser?: string) => void
```
yuchenshi, line 48:
```suggestion
  generator: (driver: AuthDriver, browser: string) => void
```

Same

yuchenshi (approved):

PR #5069, opened 2021-06-24
Port 3P token API to app-check-exp

hsubox76:
Porting 3P token APIs that were added to v8 (non-modular) API in https://github.com/firebase/firebase-js-sdk/pull/5033

Also added some test reliability improvements.

### DIFF
--- a/packages-exp/app-check-exp/src/api.test.ts
+++ b/packages-exp/app-check-exp/src/api.test.ts
@@ -16,7 +16,7 @@
  */
 import '../test/setup';
 import { expect } from 'chai';
-import { match, spy, stub } from 'sinon';
+import { spy, stub } from 'sinon';
 import {
   setTokenAutoRefreshEnabled,
   initializeAppCheck,
@@ -29,7 +29,6 @@ import {
   getFakeApp,
   getFakeGreCAPTCHA,
   getFakeAppCheck,
-  getFakePlatformLoggingProvider,
   removegreCAPTCHAScriptsOnPage
 } from '../test/util';
 import { clearState, getState } from './state';
@@ -41,6 +40,7 @@ import * as storage from './storage';
 import * as internalApi from './internal-api';
 import { deleteApp, FirebaseApp } from '@firebase/app-exp';
 import { ReCaptchaV3Provider } from './providers';
+import { AppCheckService } from './factory';
 
 describe('api', () => {
   let app: FirebaseApp;
@@ -122,11 +122,7 @@ describe('api', () => {
         token: 'a-token-string'
       });
       await getToken(appCheck, true);
-      expect(internalGetToken).to.be.calledWith(
-        appCheck.app,
-        match.any, // platformLoggerProvider
-        true
-      );
+      expect(internalGetToken).to.be.calledWith(appCheck, true);
     });
     it('getToken() throws errors returned with token', async () => {
       const app = getFakeApp({ automaticDataCollectionEnabled: true });
@@ -160,9 +156,7 @@ describe('api', () => {
       );
       stub(storage, 'writeTokenToStorage').returns(Promise.resolve(undefined));
 
-      const listener1 = (): void => {
-        throw new Error();
-      };
+      const listener1 = stub().throws(new Error());
       const listener2 = spy();
 
       const errorFn1 = spy();
@@ -173,11 +167,9 @@ describe('api', () => {
 
       expect(getState(app).tokenObservers.length).to.equal(2);
 
-      await internalApi.getToken(
-        appCheck.app,
-        getFakePlatformLoggingProvider()
-      );
+      await internalApi.getToken(appCheck as AppCheckService);
 
+      expect(listener1).to.be.called;
       expect(listener2).to.be.calledWith({
         token: fakeRecaptchaAppCheckToken.token
       });
@@ -194,7 +186,6 @@ describe('api', () => {
         provider: new ReCaptchaV3Provider(FAKE_SITE_KEY),
         isTokenAutoRefreshEnabled: true
       });
-      const fakePlatformLoggingProvider = getFakePlatformLoggingProvider();
       const fakeRecaptchaToken = 'fake-recaptcha-token';
       const fakeRecaptchaAppCheckToken = {
         token: 'fake-recaptcha-app-check-token',
@@ -207,9 +198,7 @@ describe('api', () => {
       );
       stub(storage, 'writeTokenToStorage').returns(Promise.resolve(undefined));
 
-      const listener1 = (): void => {
-        throw new Error();
-      };
+      const listener1 = stub().throws(new Error());
       const listener2 = spy();
 
       const errorFn1 = spy();
@@ -230,8 +219,9 @@ describe('api', () => {
 
       expect(getState(app).tokenObservers.length).to.equal(2);
 
-      await internalApi.getToken(appCheck.app, fakePlatformLoggingProvider);
+      await internalApi.getToken(appCheck as AppCheckService);
 
+      expect(listener1).to.be.called;
       expect(listener2).to.be.calledWith({
         token: fakeRecaptchaAppCheckToken.token
       });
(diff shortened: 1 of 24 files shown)


packages-exp/app-check-exp/src/api.test.ts
```diff
@@ -88,8 +109,167 @@ describe('api', () => {
...
+      const unsubscribe2 = onTokenChanged(appCheck, listener2, errorFn2);
+
+      expect(getState(app).tokenObservers.length).to.equal(2);
+
+      await internalApi.getToken(
+        appCheck.app,
+        getFakePlatformLoggingProvider()
+      );
+
+      expect(listener2).to.be.calledWith({
```
Feiyang1, line 181: please check that listener1 is also called.

packages-exp/app-check-exp/src/api.test.ts
```diff
@@ -88,8 +109,167 @@ describe('api', () => {
...
+      const unsubscribe1 = onTokenChanged(appCheck, {
+        next: listener1,
+        error: errorFn1
+      });
+
+      expect(getState(app).tokenObservers.length).to.equal(2);
+
+      await internalApi.getToken(appCheck.app, fakePlatformLoggingProvider);
+
+      expect(listener2).to.be.calledWith({
```
Feiyang1, line 235: please check that listener1 is also called.

packages-exp/app-check-exp/src/factory.ts
```diff
@@ -29,7 +29,11 @@ import { Provider } from '@firebase/component';
  * AppCheck Service class.
  */
 export class AppCheckService implements AppCheck, _FirebaseService {
-  constructor(public app: FirebaseApp) {}
+  platformLoggerProvider: Provider<'platform-logger'>;
+
+  constructor(public app: FirebaseApp) {
+    this.platformLoggerProvider = _getProvider(app, 'platform-logger');
```
Feiyang1, line 35: Providers should be passed in to the service class constructor in the component factory, so that all dependencies are tracked in the component factory.

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -136,12 +137,17 @@ export async function getToken(
 export function addTokenListener(
```
Feiyang1, line 137: Can `getToken()` and `addTokenListener()` take an appcheck instance instead of an app? then `platformLoggerProvider` doesn't need to be passed explicitly.

It's unclear why we are passing `platformLoggerProvider` unless you dive into the details. Using an appcheck instance sidesteps the problem and makes the code a bit cleaner IMO.

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -271,11 +279,22 @@ function notifyTokenListeners(
...
+        // from successes.
+        if (token.error) {
+          observer.error(token.error);
+        } else {
+          observer.next(token);
+        }
+      } else {
+        // Otherwise return the token, whether or not it has an error field.
+        observer.next(token);
+      }
```
Feiyang1, line 297: It is not correct for 3p listeners - If there is no error handler, the error should just be ignored. 

Do we handle both 2p and 3p listeners in this function? If so, they need to be handled differently, where 2p listeners always get the token regardless of the error state, and 3p listeners get a real token in the success handler, and errors in error handler and errors should be ignored if there is no error handler.

That also means we need a way to tell 2p and 3p listeners apart.

Feiyang1: I probably missed it in the PR for v8, but we need to do the same in v8 too.

Feiyang1 (changes requested):

packages-exp/app-check-exp/src/api.test.ts
hsubox76: Done.

packages-exp/app-check-exp/src/api.test.ts
hsubox76: Done.

packages-exp/app-check-exp/src/factory.ts
hsubox76: Done.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: Should be fixed.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: I added a `type` property to the observers. Let me know if it works, and if the names are ok.

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -292,8 +293,12 @@ function notifyTokenListeners(
           observer.next(token);
         }
       } else {
-        // Otherwise return the token, whether or not it has an error field.
-        observer.next(token);
+        // If this is a 2P listener, return the token, whether or not it
```
Feiyang1, line 296: Doesn't 3P observer always have an error callback because we will add one if it's not supplied. I think it will be clearer if we handle 3p and 2p separately in different blocks guarded by `observer.type`.

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -292,8 +293,12 @@ function notifyTokenListeners(
...
+        // has an error field. If it is a 3P listener with no error handler,
+        // ignore the error.
+        if (observer.type === ListenerType['2P']) {
```
Feiyang1, line 299: Can it be just `ListenerType.2P`?

packages-exp/app-check-exp/src/types.ts
```diff
@@ -38,6 +38,12 @@ export interface AppCheckTokenObserver
   extends PartialObserver<AppCheckTokenResult> {
   // required
   next: AppCheckTokenListener;
+  type: ListenerType;
+}
+
+export enum ListenerType {
```
Feiyang1, line 44: Can be made const enum.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: No, syntax error if it starts with a number and isn't in a string.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: I made the code a little more concise, hopefully the comments make it clear.

packages-exp/app-check-exp/src/types.ts
hsubox76: Had to turn on `preserveConstEnums` in tsconfig to enable other files to import this as a const enum. Did this cause problems with `isolatedModules` in the past? It doesn't seem like it should, I checked the generated code and everything seems to be converted to literals, and we're not exporting this type publicly.

Feiyang1: I don't think it's needed for importing from other files in the same project. Auth doesn't turn on it. There is no point of changing to const enum if we have to turn on `preserveConstEnums`.

packages-exp/app-check-exp/src/internal-api.ts
Feiyang1: Can you rename the enum members to start with a letter instead, perhaps `Internal` vs `External`? The indexing style is inconsistent with our code.

hsubox76: Renamed to `INTERNAL`/`EXTERNAL`.

packages-exp/app-check-exp/src/types.ts
hsubox76: Reverted tsconfig change. Rollup previously wouldn't build, but after moving the const enum to another file and then back, it built. I guess we can see in CI if it successfully builds from scratch.

Feiyang1 (approved):

PR #5070, opened 2021-06-24
Add firebase-specific screen_view params to Analytics

hsubox76:
Add `firebase_screen` and `firebase_screen_class` to `EventParams` type and put them in the `logEvent('screen_view', ...)` overload. This should only affect autocomplete and type checking. Adding this to the "guides" in the docs site is also in progress.

This change is for both exp and non-modular.

API proposal: go/firebase-js-screen-view

### DIFF
--- a/packages-exp/analytics-exp/src/public-types.ts
+++ b/packages-exp/analytics-exp/src/public-types.ts
@@ -280,11 +280,11 @@ export interface EventParams {
   promotions?: Promotion[];
   screen_name?: string;
   /**
-   * Firebase-specific. Use to log a screen_name to Firebase Analytics.
+   * Firebase-specific. Use to log a `screen_name` to Firebase Analytics.
    */
   firebase_screen?: string;
   /**
-   * Firebase-specific. Use to log a screen_class to Firebase Analytics.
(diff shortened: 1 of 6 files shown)


Feiyang1 (approved):
Approved for code changes.

packages-exp/analytics-exp/src/public-types.ts
```diff
@@ -279,6 +279,14 @@ export interface EventParams {
   number?: string;
   promotions?: Promotion[];
   screen_name?: string;
+  /**
+   * Firebase-specific. Use to log a screen_name to Firebase Analytics.
```
egilmorez, line 283: Are this and screen_class literals?

egilmorez (commented):
Only one nit for you: if screen_name and screen_class are literals, we should backtick them.

Thanks!

packages-exp/analytics-exp/src/public-types.ts
hsubox76: Yes, changed this in all 4 places.

egilmorez (approved):
Thanks Christina!

PR #5071, opened 2021-06-25
auto-init auth-internal after initializing auth

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/4932

Since Firestore doesn't call `authProvider.get()` to work with asynchronously loaded Auth anymore (changed to use `authProvider.onInit()` in v9 which is shared with v8), `auth-interal` needs to be proactively initialized after Auth has been initialized.

Note that it only affects v8, `auth-internal` is already initialized proactively in v9.

### DIFF
--- a/.changeset/sixty-masks-wash.md
+++ b/.changeset/sixty-masks-wash.md
@@ -1,0 +1,6 @@
+---
+"@firebase/auth": patch
+"@firebase/component": patch
+---
+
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
FYI @sam-gc I am probably going to rely on you for the review since this is not my area of expertise.

sam-gc (approved):

hsubox76 (approved):

PR #5082, opened 2021-06-30
Always send our auth token on Node, and always send an appcheck token if we have one.

jsdt:
`this.nodeAdmin` does not always appear to be set when it should be, which is breaking the admin SDK for users with appcheck enforced. This should be more robust, since we always send whatever credentials the client has.

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/.changeset/silent-seals-approve.md
+++ b/.changeset/silent-seals-approve.md
@@ -2,4 +2,4 @@
 '@firebase/database': patch
 ---
 
-On node, always send auth and appcheck tokens when they are available.
+On Node, always send Auth and AppCheck tokens when they are available.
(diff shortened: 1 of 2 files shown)


schmidt-sebastian (commented):
Please add a changeset, otherwise this will not get released.

.changeset/silent-seals-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/database': patch
+---
+
+On node, always send auth and appcheck tokens when they are available.
```
schmidt-sebastian, line 5:
```suggestion
On Node, always send Auth and AppCheck tokens when they are available.
```

schmidt-sebastian (approved):

PR #5084, opened 2021-06-30
Update appcheck to distinguish between 3P/2P listeners

hsubox76:
Fix a problem discovered while updating the exp version: https://github.com/firebase/firebase-js-sdk/pull/5069

App Check should treat listeners added by 3P vs 2P calls differently. 3P listeners should always call the supplied error handler if there's a token error. 2P listeners should return the token with the error field attached to the token.

Porting it here to the v8 version.

### DIFF
--- a/packages/app-check/src/state.ts
+++ b/packages/app-check/src/state.ts
@@ -34,6 +34,12 @@ export interface AppCheckTokenObserver
   extends PartialObserver<AppCheckTokenResult> {
   // required
   next: AppCheckTokenListener;
+  type: ListenerType;
+}
+
+export const enum ListenerType {
+  '2P' = '2P',
+  '3P' = '3P'
 }
 
 export interface AppCheckState {
--- a/packages/app-check/src/state.ts
+++ b/packages/app-check/src/state.ts
@@ -38,8 +38,8 @@ export interface AppCheckTokenObserver
 }
 
 export const enum ListenerType {
-  '2P' = '2P',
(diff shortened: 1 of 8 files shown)


packages/app-check/src/state.ts
```diff
@@ -34,6 +34,12 @@ export interface AppCheckTokenObserver
   extends PartialObserver<AppCheckTokenResult> {
   // required
   next: AppCheckTokenListener;
+  type: ListenerType;
+}
+
+export const enum ListenerType {
+  '2P' = '2P',
+  '3P' = '3P'
```
Feiyang1, line 42: Change to `INTERNAL` and `EXTERNAL` to be consistent with v9 implementation

packages/app-check/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "../../config/tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "dist"
+    "outDir": "dist",
+    "preserveConstEnums": true
```
Feiyang1, line 5:
```suggestion
    "outDir": "dist"
```

hsubox76: Reverted.

packages/app-check/src/state.ts
hsubox76: Changed.

Feiyang1 (approved):

PR #5089, opened 2021-07-01
Fix app-check-compat initialization

hsubox76:
Try to fix app-check-compat's activate() incorrectly throwing an error that app check has already been initialized. This happens because app-check-compat's factory initializes app-check-exp, which means when the compat activate() calls exp's initializeAppCheck() it is too late, it has already been initialized.

This approach avoids initializing app-check-exp until activate() is called. This means the AppCheck compat service's _delegate is going to be empty at initialization. That should be fine because all the other app check compat methods can throw an error asking the user to call activate() first. I think app check compat should have its own error anyway. If they hit the error in the called exp function, it asks you to "call initializeAppCheck() first" which doesn't exist in compat.

### DIFF
--- a/packages-exp/app-check-compat/src/service.ts
+++ b/packages-exp/app-check-compat/src/service.ts
@@ -31,12 +31,14 @@ import {
   onTokenChanged as onTokenChangedExp
 } from '@firebase/app-check-exp';
 import { PartialObserver, Unsubscribe } from '@firebase/util';
+import { ERROR_FACTORY, AppCheckError } from './errors';
 
-export class AppCheckService implements FirebaseAppCheck, _FirebaseService {
-  constructor(
-    public app: FirebaseApp,
-    readonly _delegate: AppCheckServiceExp
-  ) {}
+export class AppCheckService
+  implements FirebaseAppCheck, Omit<_FirebaseService, '_delegate'>
+{
+  _delegate?: AppCheckServiceExp;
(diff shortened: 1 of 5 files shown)


Feiyang1 (commented):
We should set the instantiation mode of components `app-check-exp` and `app-check-internal` to explicit by chaining `.setInstantiationMode(InstantiationMode.EXPLICIT)` when creating the components.

packages-exp/app-check-compat/src/service.ts
```diff
@@ -47,24 +49,42 @@ export class AppCheckService implements FirebaseAppCheck, _FirebaseService {
     } else {
       provider = new CustomProvider({ getToken: siteKeyOrProvider.getToken });
     }
-    initializeAppCheck(this.app, {
+    this._delegate = initializeAppCheck(this.app, {
```
Feiyang1, line 52: We should check if AppCheck is already initialized first. If it is, we will just get it through `provider.getImmediate()`; This can happen if both app-check and app-check-compat are included in an app and app-check is initialized first.

Feiyang1: hm, that doesn't seem to be a valid use case because people need to provide a provider to `activate`, so if they are using app-check-compat, they should use `activate` and never call `initializeAppCheck()` from the modular package.

Feiyang1 (approved):

PR #5098, opened 2021-07-01
[Auth] Break down event ID digits to a max Math.random() digit count

sam-gc:
This addresses b/184782003

### DIFF
--- a/packages-exp/auth-exp/src/core/util/event_id.ts
+++ b/packages-exp/auth-exp/src/core/util/event_id.ts
@@ -15,6 +15,16 @@
  * limitations under the License.
  */
 
-export function _generateEventId(prefix?: string): string {
-  return `${prefix ? prefix : ''}${Math.floor(Math.random() * 1000000000)}`;
+const MAX_RANDOM_DIGITS = 15;
+
+export function _generateEventId(prefix = '', digits = 10): string {
+  // This array breaks down digits into a list of numbers, maxed at 15, that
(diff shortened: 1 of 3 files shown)


packages-exp/auth-exp/src/core/util/event_id.ts
```diff
@@ -15,6 +15,24 @@
...
+  // This array breaks down digits into a list of numbers, maxed at 15, that
+  // sums to digits. For example, 10 becomes [10] and 16 becomes [15, 1].
+  const digitBreakdown = Array(Math.floor(digits / MAX_RANDOM_DIGITS)).fill(
+    MAX_RANDOM_DIGITS
+  );
+  if (digits % MAX_RANDOM_DIGITS) {
+    digitBreakdown.push(digits % MAX_RANDOM_DIGITS);
+  }
+
+  const random = digitBreakdown
```
avolkovi, line 30: wouldn't it be simpler to just do something like:
```js
(new Array(digits)).map(_ => Math.floor(Math.random() * 10)).join("")
```
?

packages-exp/auth-exp/src/core/util/event_id.ts
```diff
@@ -15,6 +15,24 @@
...
+    .map(digits =>
```
avolkovi, line 31: this should have a different name, it's shadowing the digits parameter

yuchenshi (approved):

PR #5110, opened 2021-07-02
Add debug token caching to app-check-exp

hsubox76:
Ported from v8: https://github.com/firebase/firebase-js-sdk/pull/5055
Also includes some of the test reliability fixes in 5055 that I missed porting over in an earlier PR.

### DIFF
--- a/packages-exp/app-check-exp/src/internal-api.test.ts
+++ b/packages-exp/app-check-exp/src/internal-api.test.ts
@@ -113,10 +113,6 @@ describe('internal api', () => {
         fakeRecaptchaToken
       );
       expect(token).to.deep.equal({ token: fakeRecaptchaAppCheckToken.token });
-      // TODO: Permanently fix.
-      // Small delay to prevent common test flakiness where this test runs
-      // into afterEach() sometimes
-      await new Promise(resolve => setTimeout(resolve, 50));
     });
 
     it('resolves with a dummy token and an error if failed to get a token', async () => {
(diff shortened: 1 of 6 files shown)


packages-exp/app-check-exp/src/internal-api.test.ts
```diff
@@ -239,6 +236,7 @@ describe('internal api', () => {
     });
 
     it('ignores listeners that throw', async () => {
+      stub(console, 'error');
```
Feiyang1, line 239: Is it to just suppress error messages during testing? If so, can you please add comments to clarify?

packages-exp/app-check-exp/src/internal-api.ts
```diff
@@ -152,50 +161,31 @@ export function addTokenListener(
...
-      !newState.tokenRefresher.isRunning() &&
-      state.isTokenAutoRefreshEnabled === true
-    ) {
-      newState.tokenRefresher.start();
-    }
+  // Create the refresher but don't start it if `isTokenAutoRefreshEnabled`
+  // is not true.
+  if (
+    !newState.tokenRefresher.isRunning() &&
+    state.isTokenAutoRefreshEnabled === true
```
Feiyang1, line 176:
```suggestion
    state.isTokenAutoRefreshEnabled
```
We generally don't use `===` explicitly for booleans.

packages-exp/app-check-exp/src/internal-api.test.ts
hsubox76: Yes, done.

packages-exp/app-check-exp/src/internal-api.ts
hsubox76: Fixed.

Feiyang1: It doesn't appear to be fixed.

Feiyang1 (approved):

PR #5128, opened 2021-07-09
Add cordova and react-native entry points to firebase/auth

Feiyang1:


sam-gc (approved):

hsubox76 (commented):
Some package.json seems to be out of sync and has an old version of the rollup dep, causing the yarn.lock changes here. I don't see anything in this PR's files so it might be out of scope of this PR but I think you should probably at least not add yarn.lock to this PR.

hsubox76 (approved):

PR #5135, opened 2021-07-09
Update Firestore type names

Feiyang1:
go/js-v9-types-naming


common/api-review/firestore-lite.api.md
```diff
@@ -4,6 +4,7 @@
 
 ```ts
 
+import { EmulatorMockTokenOptions } from '@firebase/util';
```
schmidt-sebastian, line 7: Can you rename `Settings` to FirestoreSettings as well (for Lite)?

schmidt-sebastian (approved):

common/api-review/firestore-lite.api.md
```diff
@@ -116,9 +118,11 @@ export abstract class FieldValue {
 }
 
 // @public
-export class FirebaseFirestore {
+export class Firestore {
     get app(): FirebaseApp;
     toJSON(): object;
+    // (undocumented)
```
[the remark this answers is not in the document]
Feiyang1: Done.

common/api-review/firestore-lite.api.md
Feiyang1: Done

packages/firestore/src/lite/database.ts
```diff
@@ -59,14 +59,17 @@ declare module '@firebase/component' {
  *
  * Do not call this constructor directly. Instead, use {@link getFirestore}.
  */
-export class FirebaseFirestore implements FirestoreService {
+export class Firestore implements FirestoreService {
+  /**
+   * whether it's a firestore or firestore-lite instance
```
Feiyang1, line 64: @egilmorez Does it look good to you?

egilmorez: If it's public, we should cap "Whether" and either:

-- backtick the types of instances as literals
-- just use the capitalized product names.

packages/firestore/src/exp/database.ts
```diff
@@ -79,7 +79,10 @@ export const CACHE_SIZE_UNLIMITED = LRU_COLLECTION_DISABLED;
  *
  * Do not call this constructor directly. Instead, use {@link getFirestore}.
  */
-export class FirebaseFirestore extends LiteFirestore {
+export class Firestore extends LiteFirestore {
+  /**
+   * whether it's a firestore or firestore-lite instance
```
egilmorez, line 84: Same comment here as below regarding caps/backticks.

But does this "database.ts" imply RTDB, and is this comment in the right place if so?

egilmorez (commented):
Thanks for putting me on the review Feiyang!

One meta question: is there a logical way to separate files with publicly-viewable comments from internal-only files?  If so, that might be a good way to narrow down the size/scope of these kinds of PRs.

packages/firestore/src/lite/database.ts
Feiyang1: They share the same type name, so I changed to use the capitalized product names.

Feiyang1:
> Thanks for putting me on the review Feiyang!
> 
> One meta question: is there a logical way to separate files with publicly-viewable comments from internal-only files? If so, that might be a good way to narrow down the size/scope of these kinds of PRs.

In simpler packages like app, analytics, remote-config, all public APIs are defined in `api.ts` and `public-types.ts`, but it's difficult to enforce the same convention for auth, firestore, database that are more complex. 

The best way to identify public docs is looking for the `@public` annotation. I wonder if we can make a GHA that highlights these locations in PRs. I have seen similar things for highlighting lint errors, so it should be doable. I will investigate.

packages/firestore/index.node.ts
```diff
@@ -20,7 +20,7 @@ import { FirebaseNamespace } from '@firebase/app-types';
 import { name, version } from './package.json';
 import { Firestore, IndexedDbPersistenceProvider } from './src/api/database';
 import { configureForFirebase } from './src/config';
-import { FirebaseFirestore as ExpFirebaseFirestore } from './src/exp/database';
+import { Firestore as ExpFirebaseFirestore } from './src/exp/database';
```
hsubox76, line 23: I know it's just internal so it doesn't matter that much but for consistency in readability, should remove the `Firebase` from the `ExpFirebaseFirestore` too, right? Saw at least one other occurrence in another file.

hsubox76 (approved):

packages/firestore/index.node.ts
Feiyang1: updated

packages/firestore/src/exp/database.ts
```diff
@@ -79,7 +79,10 @@ export const CACHE_SIZE_UNLIMITED = LRU_COLLECTION_DISABLED;
  *
  * Do not call this constructor directly. Instead, use {@link getFirestore}.
  */
-export class FirebaseFirestore extends LiteFirestore {
+export class Firestore extends LiteFirestore {
+  /**
+   * Whether it's a Firestore or Firestore Lite instance
```
egilmorez, line 84: Missing a period here.

packages/firestore/src/lite/database.ts
```diff
@@ -59,14 +59,17 @@ declare module '@firebase/component' {
  *
  * Do not call this constructor directly. Instead, use {@link getFirestore}.
  */
-export class FirebaseFirestore implements FirestoreService {
+export class Firestore implements FirestoreService {
+  /**
+   * Whether it's a Firestore or Firestore Lite instance
```
egilmorez, line 64: Missing a period.

egilmorez (approved):
LG with nits, thanks!

PR #5137, opened 2021-07-10
Update type names in v9 API

Feiyang1:


common/api-review/performance-exp.api.md
```diff
@@ -7,16 +7,16 @@
...
-}
+export function getPerformance(app?: FirebaseApp): PerformanceMonitoring;
 
 // @public
-export function getPerformance(app?: FirebaseApp): FirebasePerformance;
+export function initializePerformance(app: FirebaseApp, settings?: PerformanceSettings): PerformanceMonitoring;
 
 // @public
-export function initializePerformance(app: FirebaseApp, settings?: PerformanceSettings): FirebasePerformance;
+export interface PerformanceMonitoring {
```
jposuna, line 16: Just curious about why we have this below. Wouldn't we want to "introduce it" before the function that uses it?

jposuna (approved):
LGTM for Fireperf

common/api-review/performance-exp.api.md
Feiyang1: This file is auto generated by api-extractor to describe APIs in human readable format. You can see that it's a md file, not actual code where the order also doesn't matter. Types are compile time artifacts, and are available to things before defined them in the file.

Feiyang1: BTW, I reverted Perf related changes because we decided to keep using `FirebasePerformance` as the name.

jposuna: Sounds good, thanks!

hsubox76 (approved):

zwu52 (approved):

egilmorez (approved):
LG.  We'll need to change a bunch of links from the guides and I can take that action as soon as reference docs are generated from this source.  Thanks!

PR #5139, opened 2021-07-10
Create a logging service with flag`setDeliveryMetricsExportedToBigQueryEnabled` to enable/disable

zwu52:
**SUBMIT ONLY After ariane/4082391 APPROVAL**

Add a logging mechanism and a flag to Log client event. 
- These data will be piped to [FM BigQuery Export](https://firebase.google.com/docs/projects/bigquery-export).
- Developers should only turn this on once acquired end user permissions. 
- Logs are batch sent every 24 hours with 3 max retries.


packages-exp/messaging-exp/src/interfaces/logging-types.ts
```diff
@@ -0,0 +1,65 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/* eslint-disable camelcase */
+
+// Describes a Firebase Messaging event on a client. Defined as in
+// firebase_messaging/src/proto/messaging_event.proto
+export interface FcmEvent {
+  project_number: string; //number
```
Feiyang1, line 21: what does the comment `number` mean?

packages-exp/messaging-exp/src/interfaces/logging-types.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface LogResponse {
+  nextRequestWaitMillis: number;
+  logResponseDetails: LogResponseDetails[];
+}
+
+interface LogResponseDetails {
+  responseAction: UserResponse;
+}
+
+export enum UserResponse {
```
Feiyang1, line 61: const enum?

packages-exp/messaging-exp/src/testing/fakes/logging-object.ts
```diff
@@ -0,0 +1,65 @@
...
+    package_name: 'fake-package-name',
+    collapse_key: 'fake-collapse-key',
+    event: EVENT_MESSAGE_DELIVERED.toString(),
+    analytics_label: 'fake-analytics-label'
+    /* eslint-enable camelcase */
+  };
+}
+
+export function getSuccessResponse(): LogResponse {
+  return Object.create({
```
Feiyang1, line 54: It's worth testing responses with non 200 code and that the retry uses the `nextRequestWaitMillis` in the response.

packages-exp/messaging-exp/src/helpers/logToFirelog.ts
```diff
@@ -0,0 +1,215 @@
...
+      let delayInMs: number;
+      try {
+        delayInMs = Number(
+          ((await response.json()) as LogResponse).nextRequestWaitMillis
+        );
+      } catch (e) {
+        delayInMs = DEFAULT_BACKOFF_TIME_MS;
+      }
+
+      await new Promise(resolve => setTimeout(resolve, delayInMs));
```
Feiyang1, line 118: In remote config, we add random fuzz to the retry delays to avoid waves of retries at exact the same time. Maybe worth applying the same here. 

https://github.com/firebase/firebase-js-sdk/blob/7028c1159b9153f14c22a81fdd09df68ffcfcfb6/packages/util/src/exponential_backoff.ts#L63

packages-exp/messaging-exp/src/api.ts
```diff
@@ -141,3 +142,22 @@ export function onBackgroundMessage(
...
+ * default, message delivery metrics are not exported to BigQuery. Use this method to enable or
+ * disable the export at runtime.
+ *
+ * @param messaging - The `FirebaseMessaging` instance.
+ * @param enable - Whether Firebase Cloud Messaging should export message delivery metrics to
+ * BigQuery.
+ *
+ * @public
+ */
+export function setDeliveryMetricsExportedToBigQueryEnabled(
```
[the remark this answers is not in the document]
hsubox76: So that linked flag above is set at app-level. I also see that Android, for example, has [one set for in-app messaging at the messaging service level](https://firebase.google.com/docs/in-app-messaging/modify-message-behavior?platform=ios#enable_opt-out_message_delivery), that I assume falls back to the value of an app-level flag? Should this fall back to the value of the app-level flag if not set explicitly? Also I think messaging on web may need to have its own service-level flag, similar to the one in Android, which stops the use of messaging entirely if the user opts out.

packages-exp/messaging-exp/src/helpers/logToFirelog.test.ts
```diff
@@ -154,8 +176,9 @@ describe('logToFirelog', () => {
       }, 1000);
     });
 
-    it('clears log events if no user logging permission', done => {
+    it('send log events if user logging permission is granted', done => {
```
Feiyang1, line 179:
```suggestion
    it('sends log events if user logging permission is granted', done => {
```

packages-exp/messaging-exp/src/helpers/logToFirelog.ts
```diff
@@ -98,10 +103,21 @@ export async function _dispatchLogEvents(
...
+        if (response.ok) {
+          // existing the do-while interactive retry logic because a 200 response code has been
+          // returned.
+          break;
+        }
+
+        if (!response.ok) {
+          throw new Error(
+            'Non-200 code is returned in fetch to Firelog endpoint.'
+          );
```
Feiyang1, line 115: You probably don't want to retry unrecoverable error codes, such as 400 and 401. Remote config only retries for certain error codes - https://github.com/firebase/firebase-js-sdk/blob/master/packages/remote-config/src/client/retrying_client.ts#L68-L81

Feiyang1:
Thanks for addressing most of my comments. However, some of them are resolved without any comments, nor did I see any related code changes, so I'm not sure if there are pending changes you haven't committed, or you decided to not accept them, which is totally fine.

Can you please reply to those comments to clarify?

packages-exp/messaging-exp/src/helpers/logToFirelog.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (!messaging.deliveryMetricsExportedToBigQueryEnabled) {
+      // flush events and terminate logging service
+      messaging.logEvents = [];
+      messaging.isLogServiceStarted = false;
+
+      return;
+    }
+
+    if (!messaging.logEvents.length) {
+      return _processQueue(messaging, LOG_INTERVAL_IN_MS);
```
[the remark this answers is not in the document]
Feiyang1: I meant to ask if it should use the argument `offsetInMs` instead of using this constant.

packages-exp/messaging-exp/src/helpers/logToFirelog.ts
Feiyang1: yeah, it sounds unlikely. I think fuzzing might reduce the chance even further, but it's probably fine as is.

Feiyang1 (approved):
Approved pending CI and one question

PR #5144, opened 2021-07-12
Add @firebase/logger as dependency to app-types

DennisSimon:
Hey everyone, 

i ran into an issue with the `app-types` module where the typescript compilation of a module that defined `firebase-admin` as a dependency failed with the following error:
```
.yarn/cache/@firebase-app-types-npm-0.6.2-936c84268d-243af1c06f.zip/node_modules/@firebase/app-types/index.d.ts:17:57 - error TS2307: Cannot find module '@firebase/logger' or its corresponding type declarations.
``` 
The project uses yarn2 with a 'strict' PnP mode. 

I have created a small codesandbox that demonstrates this issue: https://codesandbox.io/s/gallant-satoshi-fycm1 
Run `yarn compile` to see the issue in action. 

Looking forward to your review!

### DIFF
--- a/packages/app-types/package.json
+++ b/packages/app-types/package.json
@@ -20,10 +20,10 @@
   "bugs": {
     "url": "https://github.com/firebase/firebase-js-sdk/issues"
   },
+  "dependency": {
+    "@firebase/logger": "0.2.6"
+  },
   "devDependencies": {
     "typescript": "4.2.2"
-  },
-  "peerDependency": {
-    "@firebase/logger": "0.x"
   }
 }
--- a/packages/app-types/package.json
+++ b/packages/app-types/package.json
@@ -21,6 +21,7 @@
     "url": "https://github.com/firebase/firebase-js-sdk/issues"
   },
   "devDependencies": {
-    "typescript": "4.2.2"
+    "typescript": "4.2.2",
+    "@firebase/logger": "0.x"
   }
 }
--- a/packages/app-types/package.json
+++ b/packages/app-types/package.json
@@ -21,7 +21,9 @@
     "url": "https://github.com/firebase/firebase-js-sdk/issues"
   },
   "devDependencies": {
-    "typescript": "4.2.2",
+    "typescript": "4.2.2"
+  },
+  "peerDependency": {
     "@firebase/logger": "0.x"
   }
 }
--- a/.changeset/short-lies-sing.md
+++ b/.changeset/short-lies-sing.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app-types': patch
+---
+
+Add @firebase/logger as a dependency to @firebase/app-types to ensure that it can be resolved when compiling the package in a strict yarn PnP environment.


DennisSimon:
@googlebot I signed it!

packages/app-types/package.json
```diff
@@ -21,6 +21,7 @@
     "url": "https://github.com/firebase/firebase-js-sdk/issues"
   },
   "devDependencies": {
-    "typescript": "4.2.2"
+    "typescript": "4.2.2",
+    "@firebase/logger": "0.x"
```
Feiyang1, line 25: Thanks for the contribution! It should be a `peerDependencies` instead. Can you please make the change?

DennisSimon: Hi @Feiyang1 , thanks for the comment!
I have one question regarding this change: 
So from my understanding, marking this as a `peerDependency` tells node that it should look for the module somewhere else in the dependency-tree, as it should already be installed. 
Yet in my case `@firebase/logger` is not included anywhere in the dependency-tree of `firebase-admin` and will therefore not be found even when adding it as a peer dependency.

In this case, would i still add `@firebase/logger` as a `peerDependency` here and should i then add `@firebase/logger` as a `dependency` in `firebase-admin`?

Feiyang1: `@firebase/logger` is a dependency of `@firebase/database`, so it should be installed.
However, yarn pnp mode has stricter rules that prevent packages to require dependencies that they don't list explicitly, so I think that's why you are seeing the error.

Is there an easy way we can test it locally?

DennisSimon: So i tested it by adding the following entry to my `.yarnrc.yaml` file: 
```yaml
packageExtensions:
  "@firebase/app-types@*":
    peerDependencies:
      "@firebase/logger": "0.x"
```
After running `yarn install`, i got the following message: 
```
➤ YN0002: │ @firebase/database-types@npm:0.7.2 doesn't provide @firebase/logger (pbcca2), requested by @firebase/app-types
```
So it looks like yarn is searching for peer-dependencies of `@firebase/app-types` only one level up? 

After extending my `.yarnrc.yaml` and declaring `@firebase/logger` as a dependency of `@firebase/database-types`, the peer dependency is resolved and the compilation works.

```yaml
packageExtensions:
  "@firebase/database-types@*":
    dependencies:
      "@firebase/logger": "0.x"
  "@firebase/app-types@*":
    peerDependencies:
      "@firebase/logger": "0.x"
```
I'll look into it more tomorrow but at this point i think it is also possible that this might also be a bug of yarn.

I will change the dependency type to a `peerDependency`.

Feiyang1: > So it looks like yarn is searching for peer-dependencies of @firebase/app-types only one level up?
It kinda makes sense. The peerDependencies of package A should be provided by packages that directly depend on A.

So I think we may need to make `@firebase/logger` a dependency of `@firebase/app-types`, and let's use `0.2.6` instead of `0.x`.

DennisSimon: Looking at the code of yarn [here](https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/package-linker.js#L595) it seems like it tries to follow the tree up. 
I think that would also make sense for projects like `firebase` that exist (for the most of it) in a monorepo. 
Well, i can check with yarn if this is expected behaviour or if its a bug but for now i agree that we can add `@firebase/logger` as a dependency of `@firebase/app-types`.

Feiyang1: SG. Can you please add a changeset to this PR via the link in https://github.com/firebase/firebase-js-sdk/pull/5144#issuecomment-878448010 ?

DennisSimon: Sure thing, added the changeset.

DennisSimon:
Hi @Feiyang1, quick ping as a reminder that the changeset is added

Feiyang1 (approved):

Issue #5141, opened 2021-07-10
Non error object is passed to Promise.reject


Monchi:
I can submit PR if this should be fixed.

### [REQUIRED] Describe your environment

  * Operating System version: macOS Big Sur
  * Browser version: Google Chrome 91.0.4472.106
  * Firebase SDK version: 9.0.0-beta.6
  * Firebase Product: auth, remote-config

#### Steps to reproduce:

firebase-js-sdk passes non-error object to Promise.reject. This causes https://github.com/getsentry/sentry-javascript/issues/2546.

I have no idea to reproduce this, but Sentry receive unhandled promise rejection of non-error object. 

#### Relevant Code:

`onerror` event is passed to `reject`. But `onerror` event is not an `Error` object. We need to wrap the event with `Error` class I think.

https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages-exp/auth-exp/src/platform_browser/load_js.ts#L28

PR #5145, opened 2021-07-12
[Auth] Fix promise rejection for failing when loading JS

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5141
Now the promise will reject with an error instead of the result of `onerror`.

### DIFF
--- a/packages-exp/auth-exp/src/platform_browser/load_js.ts
+++ b/packages-exp/auth-exp/src/platform_browser/load_js.ts
@@ -25,9 +25,12 @@ function getScriptParentElement(): HTMLDocument | HTMLHeadElement {
 export function _loadJS(url: string): Promise<Event> {
   // TODO: consider adding timeout support & cancellation
   return new Promise((resolve, reject) => {
-    function onError(e: Event|string): void {
+    function onError(e: Event | string): void {
       const errorEvent = e as ErrorEvent;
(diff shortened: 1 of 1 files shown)


packages-exp/auth-exp/src/platform_browser/load_js.ts
```diff
@@ -25,7 +28,7 @@ export function _loadJS(url: string): Promise<Event> {
     const el = document.createElement('script');
     el.setAttribute('src', url);
     el.onload = resolve;
-    el.onerror = reject;
+    el.onerror = () => reject(_createError(AuthErrorCode.INTERNAL_ERROR));
```
Feiyang1, line 31: Should we pass the original error along?

sam-gc: Done; we pass the original error if the field is populated, otherwise we create an error

Feiyang1: I think it's fine to always create a new error, but what I meant is - can we include the original error whether it's an error or a string as a field in the `INTERNAL_ERROR` object?

sam-gc: Ah okay -- I just set the error event object as the customData field. That way we don't lose any info in the conversion

Feiyang1 (approved):

PR #5147, opened 2021-07-14
Don't normalize Proto data during deepClone

schmidt-sebastian:
The SDK normalizes all Proto data into ProtobufJS before operating on them. This means that internally we use `{ seconds, nanos }`. When we send data via WebChannel, we need to use Proto3 JSON. Unfortunately, in one of the code paths we needlessly normalize the Date data and storing it in a document.

### DIFF
--- a/packages/firestore/src/exp/database.ts
+++ b/packages/firestore/src/exp/database.ts
@@ -15,12 +15,12 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
 import {
   _getProvider,
   _removeServiceInstance,
   FirebaseApp,
   getApp
-  // eslint-disable-next-line import/no-extraneous-dependencies
 } from '@firebase/app-exp';
 import { FirebaseAuthInternalName } from '@firebase/auth-interop-types';
 import { Provider } from '@firebase/component';
--- a/packages/firestore/src/model/values.ts
+++ b/packages/firestore/src/model/values.ts
@@ -559,8 +559,11 @@ export function isMapValue(
 export function deepClone(source: Value): Value {
   if (source.geoPointValue) {
     return { geoPointValue: { ...source.geoPointValue } };
-  } else if (source.timestampValue) {
-    return { timestampValue: { ...normalizeTimestamp(source.timestampValue) } };
+  } else if (
+    source.timestampValue &&
+    typeof source.timestampValue === 'object'
+  ) {
+    return { timestampValue: { ...source.timestampValue } };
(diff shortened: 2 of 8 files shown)


.changeset/thirty-suits-switch.md
```diff
@@ -0,0 +1,6 @@
+---
+"firebase": patch
+"@firebase/firestore": patch
+---
+
+Fixes an issue that prevented Timestamps to be used via `update()` when connected to the Emulator
```
thebrianchen, line 6: nit: s/Fixes/Fixed, s/to be/from being

packages/firestore/src/exp/database.ts
```diff
@@ -15,12 +15,12 @@
  * limitations under the License.
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
thebrianchen, line 18: any reason to move this?

packages/firestore/src/model/values.ts
```diff
@@ -559,8 +559,11 @@ export function isMapValue(
 export function deepClone(source: Value): Value {
   if (source.geoPointValue) {
     return { geoPointValue: { ...source.geoPointValue } };
-  } else if (source.timestampValue) {
-    return { timestampValue: { ...normalizeTimestamp(source.timestampValue) } };
+  } else if (
+    source.timestampValue &&
+    typeof source.timestampValue === 'object'
```
thebrianchen, line 564: I don't follow why we need the `typeof` check in the `else if` block. In what conditions does `timestampValue` not have type `object`? Do we need an separate condition to handle that, or does returning `{...source}` in the `else` statement work?

thebrianchen (approved):
lgtm, since I'll be OOO. I see a failing transaction JSON test, which could be related to this change.

packages/firestore/src/exp/database.ts
schmidt-sebastian: It's now causing a Lint failure (some tools have been updated)

.changeset/thirty-suits-switch.md
schmidt-sebastian: Done

packages/firestore/src/model/values.ts
schmidt-sebastian: The timestampValue can be a string, in which case `{...}` does not work. It would create an array of characters.

schmidt-sebastian (commented):
Thanks for the review. The failing test is from RTDB, but thanks for checking.

PR #5149, opened 2021-07-14
Add NodeJS support to Storage V8

schmidt-sebastian:
This adds NodeJS support to the current Storage release.


Feiyang1 (approved):

Feiyang1:
addresses https://github.com/firebase/firebase-js-sdk/issues/18

Issue #5140, opened 2021-07-10
v9 export issues with Vite


abdo643-HULK:
[REQUIRED] Describe your environment
Operating System version: macOS 11
Browser version: Safari 14
Firebase SDK version: 9
Firebase Product: all

### [REQUIRED] Describe the problem

I am working with SvelteKit which uses Vite for bundling and I am not sure if it's the firebase "export * from '@firebase/app';" or if it's Vite. But when importing firebase like this "import { initializeApp } from 'firebase/app';" Vite says it's a commonjs package. If it's a dev depencency and trying to build it reports as followed:
```
> Named export 'initializeApp' not found. The requested module 'firebase/app' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'firebase/app';
const { initializeApp } = pkg;
```
For the dev server it works fine. When installing firebase as a normal dependency and running build everything works fine but when trying to run the dev server this error pops up: 
```
12:45:52 [vite] Error when evaluating SSR module /node_modules/firebase/app/dist/index.cjs.js:
ReferenceError: exports is not defined
    at eval (/node_modules/firebase/app/dist/index.cjs.js:5:23)
    at instantiateModule (/Users/abdo/Downloads/svelte-firebase-error-master/node_modules/vite/dist/node/chunks/dep-cc49d7be.js:72639:166)
exports is not defined
ReferenceError: exports is not defined
    at eval (/node_modules/firebase/app/dist/index.cjs.js:5:23)
    at instantiateModule (/Users/abdo/Downloads/svelte-firebase-error-master/node_modules/vite/dist/node/chunks/dep-cc49d7be.js:72639:166)
```

Like I said I am not sure if it's because of the all export or if it's Vite but changing the "import { initializeApp } from 'firebase/app';" to "import { initializeApp } from '@firebase/app';" doesn't throw an error in both cases.

sodatea:
I was just about to raise a similar issue 😅

It's because firebase v9 uses nested `package.json`s and puts the `browser` fields into them.
It's an unconventional way to use the `browser` field.

Vite only reads `browser` from the root `package.json`. Webpack does recognize the nearest `package.json`. But as it's undocumented behavior, I doubt this approach would continue to work in the long run.

sodatea:
Never mind. The previous comment is off-topic.

[...]

PR #5167, opened 2021-07-16
add esm entry points for node

Feiyang1:
Addresses https://github.com/firebase/firebase-js-sdk/issues/5140

### DIFF
--- a/packages-exp/firebase-exp/package.json
+++ b/packages-exp/firebase-exp/package.json
@@ -26,182 +26,182 @@
   "exports": {
     "./analytics": {
       "node": {
-        "require": "./analytics/dist/index.cjs.js",
+        "require": "./analytics/dist/index.cjs",
         "import": "./analytics/dist/index.esm.js"
       },
       "default": "./analytics/dist/index.esm.js"
     },
     "./app": {
       "node": {
-        "require": "./app/dist/index.cjs.js",
+        "require": "./app/dist/index.cjs",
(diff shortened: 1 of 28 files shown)


packages-exp/firebase-exp/package.json
```diff
@@ -25,107 +25,185 @@
   ],
   "exports": {
     "./analytics": {
-      "node": "./analytics/dist/index.cjs.js",
+      "node": {
+        "require": "./analytics/dist/index.cjs.js",
+        "import": "./analytics/dist/index.esm.js"
```
sodatea, line 30: The file extension should be `.mjs`, otherwise Node would fail to parse it.

sodatea (changes requested):

packages-exp/firebase-exp/package.json
Feiyang1: Indeed, good catch! I fixed it in a different way by declaring `"type": "module"` in the `package.json`s, and changing the extension of cjs files from js to cjs, so they still work with `require()`s. This way, I only need to generate 2 files per subpath, an esm.js file and a cjs file instead of 3(esm.js, esm.mjs and cjs.js).

Feiyang1: I tested it locally, and both import and require syntax work with this setup.

hsubox76 (approved):

PR #5176, opened 2021-07-20
Fix changeset formatting

hsubox76:
This was caught by the changeset checker in the PR https://github.com/firebase/firebase-js-sdk/pull/5149 but I think it got lost in the noise of the warnings about firebase-compat-typings-test that I think we can't get rid of until v9 is GA.

### DIFF
--- a/.changeset/curly-shrimps-wait.md
+++ b/.changeset/curly-shrimps-wait.md
@@ -1,5 +1,6 @@
 ---
 '@firebase/storage': minor
+'firebase': minor
 ---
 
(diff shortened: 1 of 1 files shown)


.changeset/curly-shrimps-wait.md
```diff
@@ -1,5 +1,5 @@
 ---
-@firebase/storage: minor
+'@firebase/storage': minor
```
Feiyang1, line 2: We should also add `'firebase`: minor`

Feiyang1 (approved):

.changeset/curly-shrimps-wait.md
hsubox76: Done.

Issue #5178, opened 2021-07-20
Export error codes from Auth


Feiyang1:
Export error codes as an object map from Auth to make error code check easier. Addresses https://github.com/firebase/firebase-js-sdk/issues/4551
To enable use case like the following:
```javascript
// ErrorCodes is a map of all possible Auth error codes
import { signInAnonymously, ErrorCodes } from 'firebase/auth';

async function signIn() {
    try {
    const credential = signInAnonymously(auth);
  } catch (e) {
    // instead of e.code === 'auth/network-request-failed'
    if (e.code === ErrorCodes.NetworkRequestFailed) {
      // do stuff
    }
  }
}
```

Object maps are tree shakable, so they incur no size cost if they are not imported.

We may want to do the same in other SDKs as well.

PR #5182, opened 2021-07-21
[Auth] Add exported const map for error codes

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5178


packages-exp/auth-exp/src/core/errors.ts
```diff
@@ -433,3 +431,112 @@ export const _DEFAULT_AUTH_ERROR_FACTORY = new ErrorFactory<
...
+ * thrown by the SDK.
+ *
+ * @remarks
+ * Please note that you can't tree-shake individual keys
+ * in the map, so by using the map you might substantially increase your
+ * bundle size.
+ *
+ * @public
+ */
+export const AUTH_ERROR_CODES_MAP_DO_NOT_USE_INTERNALLY = {
```
Feiyang1, line 446: I think we want to use `as const` to prevent type widening, so we show the actual literal string values in the reference docs, and people get better intellisense (the literal string vs string) in IDE.

sam-gc: Done.

packages-exp/auth-exp/src/core/errors.ts
```diff
@@ -433,3 +431,112 @@ export const _DEFAULT_AUTH_ERROR_FACTORY = new ErrorFactory<
   AuthErrorCode,
   AuthErrorParams
 >('auth', 'Firebase', _prodErrorMap());
+
+/**
+ * A map of potential Auth error codes, for easier comparison with errors
+ * thrown by the SDK.
+ *
+ * @remarks
+ * Please note that you can't tree-shake individual keys
```
egilmorez, line 440: The style guide is not fond of "please."  Just "Note" is helpful enough I think.

egilmorez (approved):
LG with a nit, thanks!

Feiyang1 (approved):

packages-exp/auth-exp/src/core/errors.ts
Feiyang1: @sam-gc Do you mind fixing it, so we can get it released in the next release?

yuchenshi:
```suggestion
 * Note that you can't tree-shake individual keys
```

PR #5204, opened 2021-07-27
Add E2E workflow and tests

hsubox76:
Tests are compiled with karma-typescript. There is a webpack config file, this is only used with the sample-apps, which maybe shouldn't be part of this directory (see below)?

Add:
- e2e Github workflow file with 1 job
  - test - runs compat and exp tests on `firebase@latest`
- e2e test dir: in root and not `/integration` so lerna won't try to yarn install it
  - tests
    - compat
    - modular
  - sample-apps (ready-to-go apps for debugging and custom tests on local machine)
    - compat
    - modular
  - context-template.html
    - injects APP_CHECK_DEBUG_TOKEN into browser context before firebase loads - test-setup.js injects the actual token value from the env variable.


e2e/build/index.html
```diff
@@ -0,0 +1,6 @@
+<body>
+  <script>
+    // self.FIREBASE_APPCHECK_DEBUG_TOKEN = 'APP_CHECK_DEBUG_TOKEN';
```
Feiyang1, line 3: Should it be removed?

e2e/karma.conf.js
```diff
@@ -0,0 +1,91 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+function getTestFiles(argv) {
+  const files = [];
+  if (argv.includes('--compat')) {
+    files.push('./tests/compat.test.ts');
+  } else if (argv.includes('--exp')) {
```
Feiyang1, line 22: `exp` is outdated now, and should be replaced throughout the PR.

e2e/karma.conf.js
```diff
@@ -0,0 +1,91 @@
...
+      bundlerOptions: {
+        resolve: {
+          directories: ['./node_modules'],
+          alias: {
+            '@firebase/messaging/sw':
+              'node_modules/@firebase/messaging/dist/index.sw.esm2017.js'
+          }
+        },
+        transforms: [
+          require('karma-typescript-es6-transform')({
```
Feiyang1, line 66: Why is it needed? Wouldn't `karma-tyescript` already transpile the tests?

e2e/karma.conf.js
```diff
@@ -0,0 +1,91 @@
...
+
+function getTestFiles(argv) {
+  const files = [];
+  if (argv.includes('--compat')) {
+    files.push('./tests/compat.test.ts');
+  } else if (argv.includes('--exp')) {
+    files.push('./tests/exp.test.ts');
+  } else if (argv.includes('--v8')) files.push('./tests/v8.test.ts');
+  {
+  }
```
Feiyang1, line 26: empty block

e2e/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "firebase-smoke-test",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "setup": "node test-setup.js",
+    "test": "yarn setup && karma start --compat --exp",
```
Feiyang1, line 8: no test command for v8?

e2e/sample-apps/compat.js
```diff
@@ -0,0 +1,270 @@
...
+ * Functions smoke test.
+ *
+ * Call a deployed function.
+ * This cloud function must be deployed in this project first. It can be
+ * found in this repo's /functions folder.
+ */
+async function callFunctions() {
+  console.log('[FUNCTIONS] start');
+  const functions = firebase.functions();
+  const callTest = functions.httpsCallable('callTest');
```
Feiyang1, line 90: Do you need to run `yarn test:setup` at the root of the repo to make sure the function is deployed? I don't see it being called anywhere.

e2e/sample-apps/compat.js
```diff
@@ -0,0 +1,270 @@
...
+  console.log('[DATABASE] updating (expect to see snapshot data change)');
+  await ref.update({ number: 987 });
+  console.log("[DATABASE] deleting (expect to see snapshot doesn't exist)");
+  await ref.remove();
+  ref.off();
+}
+
+/**
+ * Messaging smoke test.
+ * Call getToken(), it won't work on localhost, just a minimal test to make
```
Feiyang1, line 190: what wouldn't it work on localhost?

e2e/tests/compat.test.ts
```diff
@@ -0,0 +1,206 @@
...
+        testAccount.email,
+        testAccount.password
+      );
+      console.log('Logged in with test account', cred.user.email);
+      expect(cred.user.email).to.equal(testAccount.email);
+    });
+  });
+
+  describe('APP CHECK', async () => {
+    // @ts-ignore
```
Feiyang1, line 60: Is `ts-ignore` needed?

e2e/package.json
Feiyang1, line 8: Maybe worth running tests in Nodejs as well

e2e/build/index.html
hsubox76: Removed.

e2e/karma.conf.js
hsubox76: It doesn't handle ES6 import and exports. It probably should but it doesn't. https://github.com/monounity/karma-typescript/issues/87

e2e/karma.conf.js
hsubox76: Done.

e2e/karma.conf.js
hsubox76: Removed.

e2e/package.json
hsubox76: I don't know if it's worth maintaining v8 e2e tests for the occasional GCIP update?

e2e/tests/compat.test.ts
hsubox76: Oh that was temporary when AppCheck typings were broken. Removed.

e2e/sample-apps/compat.js
hsubox76: I don't want to add too much to this project and I don't want to put this function in config/functions to keep this separate from main repo testing. I put the source code for the function itself in sample-apps/functions/index.js and it's up to the user to deploy it to their project. It's already deployed to the jscore-sandbox project.

Changed the comment to explain this.

e2e/sample-apps/compat.js
hsubox76: Also added this same function to the modular sample app.

e2e/package.json
hsubox76: Seems like a good idea, probably should be an additional PR.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: echo "NODE_OPTIONS=--max_old_space_size=4096" >> $GITHUB_ENV
+    - name: Write project config
+      env:
+        PROJECT_CONFIG: ${{ secrets.TEST_PROJECT_CONFIG }}
+        TEST_ACCOUNT: ${{ secrets.TEST_ACCOUNT }}
+      run: |
+        echo "export const config = $PROJECT_CONFIG; export const testAccount = $TEST_ACCOUNT" > firebase-config.js
+    - name: Yarn install
+      run: |
+        yarn add firebase@beta
```
Feiyang1, line 35: Remove `@beta`

e2e/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "start:modular": "webpack serve --config-name modular",
+    "start:compat": "webpack serve --config-name compat"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "firebase": "9.0.2"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.14.6",
```
Feiyang1, line 22: use pinned version for devdeps

e2e/sample-apps/compat.js
Feiyang1: We should have a script to automate the deployment, so it's easy to set up again in case something changes in jscore-sandbox. Ideally we should always run this script as a setup step before running the tests without assuming the functions are available.

.github/workflows/e2e-test.yml
hsubox76: Argh, bad merge maybe.

e2e/package.json
hsubox76: Done.

e2e/sample-apps/compat.js
hsubox76: Done. We don't do this for normal CI tests by the way. We might want to but it might be a little more complicated since the top level test setup script has interactive prompts and uses the firebase-cli module which seems to be a little buggy. Also the top level script also deploys database and firestore rules.

hsubox76: Ok, I added a README.md explaining setup.

Feiyang1: I'm good as long as we have an easy way to deploy the function.

Feiyang1 (approved):

Issue #5195, opened 2021-07-23
V9 Realtime Database listener stops listening after transaction


jhuleatt:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Big Sur
  * Browser version: Chrome `91.0.4472.164`
  * Firebase SDK version: `9.0.0-beta.6`
  * Firebase Product: database

### [REQUIRED] Describe the problem

The `onValue` listener stops listening for new values on a reference after a transaction.

#### Steps to reproduce:

Runnable repro is available in codesandbox: https://codesandbox.io/s/rtdb-transaction-update-repro-6u0qt?file=/src/index.ts

1. Click the **"SET" Increment** button a few times. The number next to **Current value from RTDB is:** will change as expected.
2. Click the **"TRANSACTION" Increment** button a few times. The number next to **Current value from RTDB is:** will stop updating ⚠️
4. Check the console logs. You'll see that every transaction completed successfully.
3. Click the **"SET" Increment** button again. The number next to **Current value from RTDB is:** still won't update.
4. Refresh the page. The number next to **Current value from RTDB is:** will be significantly higher from all of your previous clicks, proving that the transactions worked.

schmidt-sebastian:
Thanks for the repro. I started looking at this and will report back in a bit.

schmidt-sebastian:
Fix pending. Thanks for catching this. This was a bug in our listener unregistration.

PR #5205, opened 2021-07-27
Fix transaction interference in v9

schmidt-sebastian:
This fixes a bug in listener unregistration, which exists in both the user facing `unsubscribe` API as well as runTransaction() (since it uses listeners internally).

When we unregister a listener, we go through all existing listeners to find the one that matches. We do this by comparing the callbacks for equality.

In v8, we wrap all listeners, and hence, we have to unwrap the listener to get back to the original user function before checking for equality. Our listener looks as such:

```
const listener = () => { userCallback(); /*...*/ )
listener.userCallback = userCallback;
```

In v9, we have:

```
const listener = userCallback(); 
```

Unfortunately, the comparison function considered all v9 listeners the same since it either compared `listener` (which only matches for the exact listener) or `listener.userCallback` (which matched every listener since the property was never set).

Fixes https://github.com/firebase/firebase-js-sdk/issues/5195

### DIFF
--- a/packages/database/src/core/view/EventRegistration.ts
+++ b/packages/database/src/core/view/EventRegistration.ts
@@ -71,8 +71,10 @@ export class CallbackContext {
   matches(other: CallbackContext): boolean {
     return (
       this.snapshotCallback === other.snapshotCallback ||
-      (this.snapshotCallback.userCallback ===
-        other.snapshotCallback.userCallback &&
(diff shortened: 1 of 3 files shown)


egilmorez (approved):
When there are no changes to comments you don't really need any TW review, but I do approve, thanks!  :)

Feiyang1 (approved):

Issue #5116, opened 2021-07-03
Update links in comments to @link format in Analytics


Feiyang1:
Update links (e.g. https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/analytics-exp/src/public-types.ts#L30)  in Analytics to `@link` format. 

`@link` example: https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/app-exp/src/api.ts#L54

PR #5206, opened 2021-07-27
Fix links and update event typings

hsubox76:
Fixed some link formatting in comments, addressing https://github.com/firebase/firebase-js-sdk/issues/5116.

While doing that, discovered some of the links were for UA and not for GA4, meaning some of the event/config params were incorrect. Corrected them.

The link for screenviews (https://firebase.google.com/docs/analytics/screenviews) doesn't have the JS snippet yet, it can't be submitted until the typing changes in this PR have been released. It should be a short lag time though.

### DIFF
--- a/packages-exp/analytics-exp/src/api.ts
+++ b/packages-exp/analytics-exp/src/api.ts
@@ -221,8 +221,8 @@ export function setAnalyticsCollectionEnabled(
  * app instance on this device.
  * @public
  * List of recommended event parameters can be found in
- * {@link https://developers.google.com/gtagjs/reference/event
- * | the gtag.js reference documentation}.
+ * {@link https://developers.google.com/gtagjs/reference/ga4-events
+ * | the GA4 reference documentation}.
  */
 export function logEvent(
   analyticsInstance: Analytics,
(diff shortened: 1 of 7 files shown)


egilmorez (approved):
LG, thanks!

packages-exp/analytics-exp/src/api.ts
```diff
@@ -221,8 +221,8 @@ export function setAnalyticsCollectionEnabled(
  * app instance on this device.
  * @public
  * List of recommended event parameters can be found in
- * {@link https://developers.google.com/gtagjs/reference/event
- * | the gtag.js reference documentation}.
+ * {@link https://developers.google.com/gtagjs/reference/ga4-events
```
Feiyang1, line 224: What is ga4?

hsubox76: Google Analytics 4, the new version of Google Analytics with App+Web properties and measurement IDs that start with "GA" instead of "UA" for Universal Analytics. The other link was for the older UA reference. All Firebase Analytics properties are GA4.

Feiyang1 (approved):

PR #5207, opened 2021-07-28
Make initializeApp idempotent

Feiyang1:
`initializeApp` now can be called more than once and returns the same instance as long as using the same options and config. It should make the devX more friendly with HMR and other use cases.

No need to wrap `initializeApp()` in the if block anymore:
```javascript
// the if condition can be safely removed now
// if(getApps().length === 0) {
    initializeApp({...})
// }
```

### DIFF
--- a/packages-exp/app-compat/test/firebaseAppCompat.test.ts
+++ b/packages-exp/app-compat/test/firebaseAppCompat.test.ts
@@ -346,7 +346,7 @@ function firebaseAppTests(
       expect(firebase.apps.length).to.eq(2);
     });
 
-    it('initilizeApp can be called more than once and returns the same instance if the options and config are the same', () => {
+    it('initializeApp can be called more than once and returns the same instance if the options and config are the same', () => {
       const app = firebase.initializeApp(
         {
           apiKey: 'test1'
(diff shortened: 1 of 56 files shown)


packages-exp/app-compat/test/firebaseAppCompat.test.ts
```diff
@@ -346,15 +346,61 @@ function firebaseAppTests(
       expect(firebase.apps.length).to.eq(2);
     });
 
-    it('duplicate DEFAULT initialize is an error.', () => {
-      firebase.initializeApp({});
-      expect(() => firebase.initializeApp({})).throws(/\[DEFAULT\].*exists/i);
+    it('initilizeApp can be called more than once and returns the same instance if the options and config are the same', () => {
```
hsubox76, line 349: typo in `initializeApp`

packages-exp/app-exp/src/errors.ts
```diff
@@ -31,7 +31,8 @@ const ERRORS: ErrorMap<AppError> = {
     "No Firebase App '{$appName}' has been created - " +
     'call Firebase App.initializeApp()',
   [AppError.BAD_APP_NAME]: "Illegal App name: '{$appName}",
-  [AppError.DUPLICATE_APP]: "Firebase App named '{$appName}' already exists",
+  [AppError.DUPLICATE_APP]:
+    "Firebase App named '{$appName}' already exists with different options",
```
hsubox76, line 35: "different options or config"? or are we just using "options" as a general term to mean like params?

hsubox76 (approved):

Issue #5086, opened 2021-06-30
Firebase's "Build Unit Tests" storage documentation is incomplete and incorrect


lovelle-cardoso:
### [REQUIRED] Describe your environment

- Operating System version: Windows 10 Home Version 20H2, Build 19042.1052
- Browser version: N/A (using mocha and @firebase/rules-unit-testing testing library)
- Firebase SDK version: @firebase/rules-unit-testing 1.3.8
- Firebase Product: storage, storage emulator, testing sdk

### [REQUIRED] Describe the problem

If you attempt to write storage unit tests in the way the official documentation describes, you will encounter several errors that prevent the tests from running as intended.

#### Steps to reproduce:

**Step 1:** Read the documentation for setting up cloud storage unit tests at: https://firebase.google.com/docs/rules/unit-tests#storage

**Step 2:** Attempt to write unit tests following the examples in the documentation

#### Expected Behavior:

By following along with the documentation, you are able to write unit tests that test your storage security rules.

#### Actual Behavior:

You will encounter several issues which you must work around and are not stated in the documentation.

---

**Issue 1:** If you are writing your tests in a nodejs environment, you will encounter a "XMLHttpRequest is not defined" error when attempting to use the storage module. You must install and import xhr2 to resolve this issue.

---

**Issue 2:** The documentation says you may pass an auth object when initializing your testing app. But if you attempt to pass this auth object and test storage security rules, you will encounter an error that prevents your tests from running. This bug was reported a month ago here: https://github.com/firebase/firebase-tools/issues/3442 but is still being actively worked on here: https://github.com/firebase/firebase-js-sdk/pull/5002. (I'd appreciate if this bug was noted in the documentation, considering how critical authentication is to testing security)

---

**Issue 3:** There is no documentation that describes how to test uploading and downloading a file. The documentation only shows examples for ref.getMetadata(). It's strange that the documentation doesn't include any examples of what most users would be interested in testing for storage (file uploads and downloads) and only includes examples for getting a file's metadata (an action which is not normally that crucial to security)

---

**Issue 4:** If you try to write tests using firebase.assertFails for reading, uploading, and downloading files, the getDownloadUrl(), and put() methods will throw an error that the rules-unit-testing library does not recognize. This results in you seeing oddly contradictory errors like this:

[...]

PR #5209, opened 2021-07-28
fix: assertFails for storage read/write requests (#5086)

lovelle-cardoso:
Fixes Issue 4 of #5086

### Discussion

https://github.com/firebase/firebase-js-sdk/issues/5086

This PR makes assertFails work properly with the storage emulator. 

Most storage permission errors throw errors with the message "(storage/unauthorized)", but assertFails only checks for the string "permission denied". This means that assertFails won't work properly for most storage unit tests. 

This fix is a small one that just adds another check if the message contains the string "unauthorized". A better fix would be to ensure that all firebase products throw consistent error message codes when permission is denied, but this solution will at least fix this bug in the unit testing library for storage for the time being.

### Testing

  * Added unit test to packages/rules-unit-testing/test/database.test.ts called `'assertFails() if message contains unauthorized'`
  
NOTE: It would be better if the rules-unit-testing library also contained tests for real-world usage (like actually spinning up the emulators and checking assertFails and assertSucceeds for basic actions like reading/writing from firestore, realtiime database, or storage). I think having just a few basic tests for usage would have caught this issue much quicker.

### DIFF
--- a/packages/rules-unit-testing/src/api/index.ts
+++ b/packages/rules-unit-testing/src/api/index.ts
@@ -698,7 +698,9 @@ export function assertFails(pr: Promise<any>): any {
         errCode === 'permission-denied' ||
         errCode === 'permission_denied' ||
         errMessage.indexOf('permission_denied') >= 0 ||
-        errMessage.indexOf('permission denied') >= 0;
+        errMessage.indexOf('permission denied') >= 0 ||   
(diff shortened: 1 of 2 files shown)


avolkovi (approved):
Looks good, thanks for the fix and detailed test!

yuchenshi (approved):

PR #5214, opened 2021-07-29
Update README

hsubox76:
- Added products that didn't exist at the time.
- Put individual product imports as the main method and all-in-one imports as the alternative.
- Added `require('firebase/app').default` per our ESM module changes in 8.0.0
- Commented out build badge since we don't use Travis anymore. Reminder to replace later with a new service if possible.

### DIFF
--- a/packages/firebase/README.md
+++ b/packages/firebase/README.md
@@ -1,4 +1,4 @@
-<!-- [![Build Status](https://travis-ci.org/firebase/firebase-js-sdk.svg?branch=master)](https://travis-ci.org/firebase/firebase-js-sdk) -->
+<!-- TODO: Build/Test badges when available. -->
 
 # Firebase - App success made simple
 
--- a/packages/firebase/README.md
+++ b/packages/firebase/README.md
@@ -133,6 +133,23 @@ _The type information from the import statement will include all of the SDKs,
 not just the ones you have `required`, so you could get a runtime error if you
 reference a non-required service._
 
+#### Alternative - all-in-one import
+
+>This brings in all Firebase features. We recommend the method above to
+>minimize download size by only including the scripts you need.
+
+```js
+// This import loads all Firebase services, whether used in your code or not.
+import firebase from 'firebase';
+```
+
+Or with `require()`:
+
+```js
+// This import loads all Firebase services, whether used in your code or not.
+var firebase = require('firebase').default;
+```
+
 ## Get the code (Node.js - server and command line)
 
 ### NPM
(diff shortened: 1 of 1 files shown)


packages/firebase/README.md
```diff
@@ -36,16 +36,31 @@ you should use the
...
 
-### Script include
->This brings in all Firebase features. See
->["Include only the features you need"](#include-only-the-features-you-need)
->below for
->how to minimize download size by only including the scripts you need.
+We recommend only installing the features you need. The individually installable services are:
 
-Include Firebase in your web application via a `<script>` tag:
+- `firebase-app` - The core `firebase` client (required).
```
Feiyang1, line 41: misssing app-check

packages/firebase/README.md
```diff
@@ -154,10 +115,41 @@ import 'firebase/auth';
 import 'firebase/database';
 ```
 
+Or if using `require()`:
+
+_We recommend the `.default` import from `firebase/app` in order for
```
Feiyang1, line 120:
```suggestion
_Use the `.default` import from `firebase/app` in order for
```

Feiyang1: I think you have to use the default import.

packages/firebase/README.md
```diff
@@ -177,7 +169,9 @@ $ npm install --save firebase
 In your code, you can access Firebase using:
 
 ```js
-var firebase = require('firebase');
+var firebase = require('firebase/app').default;
```
hsubox76, line 172: I wanted to double check that Node needs the .default too? I vaguely remember maybe not?

packages/firebase/README.md
hsubox76: Oops, added.

packages/firebase/README.md
Feiyang1: yes, if you want to get typings. Can you also update `var`s to `const`s?

packages/firebase/README.md
Feiyang1: Did you push it? I still don't see appcheck

hsubox76: Missed error on push because I didn't pull the branch after committing the suggestion below.

packages/firebase/README.md
hsubox76: Updated.

packages/firebase/README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/firebase/firebase-js-sdk.svg?branch=master)](https://travis-ci.org/firebase/firebase-js-sdk)
+<!-- [![Build Status](https://travis-ci.org/firebase/firebase-js-sdk.svg?branch=master)](https://travis-ci.org/firebase/firebase-js-sdk) -->
```
Feiyang1, line 1: Should it be deleted?

Feiyang1 (approved):

packages/firebase/README.md
hsubox76: I thought that leaving the comment would be a good reminder to put a new badge there if we go through the README in the future and we have a new service by then. I guess it's better to explicitly put a TODO instead.

PR #5215, opened 2021-07-29
Add storage to Node bundle and Node docs

hsubox76:
We support Node in Storage for both v9 and v8 now (https://github.com/firebase/firebase-js-sdk/pull/5149 and https://github.com/firebase/firebase-js-sdk/pull/4985) so it should be included in Node docs (remove `@webonly` tags) and in the Node all-in-one bundle.


Feiyang1 (approved):

Feiyang1:
Need a changeset for the index.node.ts change

Issue #5171, opened 2021-07-17
V9 Modular Authentication  - browserSessionPersistence log out when refresh


daveteu:
### [REQUIRED] Describe your environment

  * Operating System version: macos / ubuntu / window
  * Browser version: chrome / safari
  * Firebase SDK version: 9 beta 6
  * Firebase Product: auth

### [REQUIRED] Describe the problem
I believe browserSessionPersistence is supposed to persist during the time window/tab is open. However, the moment I refresh my webpage, I get logged out. This behavior did not exist in SDK v8.

With the same codes, browserLocalPersistence works like how it should in Version 8, so I believe browserSessionPersistence is not behaving the way it should.

#### Steps to reproduce:

```
const auth = getAuth()

await setPersistence(auth, browserSessionPersistence)

//.... login codes to different providers
```

looptheloop88:
Hi @daveteu, thanks for the report. I tried replicating the behavior you've reported, but I wasn't able to. I used the following code snippets and the user auth state persists even when the page is refreshed.

Import Auth modules:
```
import { signOut, onAuthStateChanged, signInWithPopup, GoogleAuthProvider, initializeAuth, 
browserSessionPersistence, browserPopupRedirectResolver } from "firebase/auth";
```

Initialize Auth and listen to Auth state:
```
firebaseAuth = initializeAuth(firebaseApp, {
    persistence: browserSessionPersistence,
    popupRedirectResolver: browserPopupRedirectResolver,
});

onAuthStateChanged(firebaseAuth, (user) => {
    if (user) {
        console.log("User is signed in.");
    } else {
        console.log("User is signed out.");
    }
});
```

Sign in the user:
```
const gogleAuthProvider = new GoogleAuthProvider();
signInWithPopup(firebaseAuth, gogleAuthProvider)
    .then((result) => {
        console.log(result);
    })
    .catch((error) => {
        console.log(error);
    });
```

Sign out the user:

```
signOut(firebaseAuth).then(() => {
    console.log("User successfully signed out.");
}).catch((error) => {
    console.log("Signed out error.");
});
```

If I can replicate the issue, I can have a better look into it. Please share a minimal, but complete sample of a project that I can run locally to replicate the issue.

porfirioribeiro:
I'm having the same problem. 
It's hard to debug, 
when you are out of the page and get back, and it refreshes the auth on background, looks like `onAuthStateChanged` is called twice with null and then with the user object

[...]

PR #5239, opened 2021-08-03
[Auth] Fix persistence issue when manually setting/using session storage

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5171

`browserSession` persistence is now added to the escalation path of `getAuth()` to match v8. We also now only "upgrade" the persistence if the persistence deems itself "upgradeable" (which only applies to `indexedDB` and `localStorage`)

### DIFF
--- a/packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
+++ b/packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
@@ -112,36 +112,42 @@ export class PersistenceUserManager {
         userKey
       );
     }
+    
+    // Eliminate any persistences that are not available
+    const persistences = (await Promise.all(persistenceHierarchy.map(async persistence => {
+      if (await persistence._isAvailable()) {
+        return persistence;
+      }
+      return undefined;
+    }))).filter(persistence => persistence) as PersistenceInternal[];
 
-    // Use the first persistence that supports a full read-write roundtrip (or fallback to memory).
-    let chosenPersistence = _getInstance<PersistenceInternal>(
+    // Fall back to the first persistence listed, or in memory if none available
+    let chosenPersistence = persistences[0] || _getInstance<PersistenceInternal>(
       inMemoryPersistence
     );
-    for (const persistence of persistenceHierarchy) {
-      if (await persistence._isAvailable()) {
-        chosenPersistence = persistence;
-        break;
-      }
-    }
 
-    // However, attempt to migrate users stored in other persistences (in the hierarchy order).
-    let userToMigrate: UserInternal | null = null;
     const key = _persistenceKeyName(userKey, auth.config.apiKey, auth.name);
-    for (const persistence of persistenceHierarchy) {
-      // We attempt to call _get without checking _isAvailable since here we don't care if the full
-      // round-trip (read+write) is supported. We'll take the first one that we can read or give up.
+
+    let userToMigrate: UserInternal | null = null;
+    for (const persistence of persistences) {
       try {
-        const blob = await persistence._get<PersistedBlob>(key); // throws if unsupported
+        const blob = await persistence._get<PersistedBlob>(key);
         if (blob) {
           const user = UserImpl._fromJSON(auth, blob); // throws for unparsable blob (wrong format)
-          if (persistence !== chosenPersistence) {
-            userToMigrate = user;
-          }
+          chosenPersistence = persistence;
+          userToMigrate = user;
           break;
         }
       } catch {}
     }
 
+    if (!chosenPersistence._shouldAllowMigration) {
+      return new PersistenceUserManager(chosenPersistence, auth, userKey);
+    }
+
+    // Migrate up the chain
+    const migrationHierarchy = persistences.filter(p => p._shouldAllowMigration);
+    chosenPersistence = migrationHierarchy[0];
     if (userToMigrate) {
       // This normally shouldn't throw since chosenPersistence.isAvailable() is true, but if it does
       // we'll just let it bubble to surface the error.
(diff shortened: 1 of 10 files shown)


packages-exp/auth-exp/test/integration/webdriver/persistence.test.ts
```diff
@@ -182,34 +182,8 @@ browserDescribe('WebDriver persistence test', (driver, browser) => {
       );
     });
 
-    it('migrate stored user to localStorage if indexedDB is readonly', async () => {
```
lisajian, line 185: Clarifying q (trying to understand more than anything!): Not catching why this test would be impacted by the fix to session persistence? This test concerns localStorage and indexDB only, right?

lisajian (commented):
LGTM - will defer to @yuchenshi / @avolkovi who have more context on the change! Thanks for the fix!

packages-exp/auth-exp/test/integration/webdriver/persistence.test.ts
sam-gc: This change also makes it so we only look at "available" persistence layers before initializing the SDK. If `indexedDB` is readonly, it would be filtered out early on, so this behavior (migrating from readonly indexedDB to localStorage) doesn't happen anymore.

lisajian: Ohh got it, thanks for the clarification!

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -113,35 +113,53 @@ export class PersistenceUserManager {
...
+    if (!chosenPersistence._shouldAllowMigration) {
+      return new PersistenceUserManager(chosenPersistence, auth, userKey);
+    }
+
+    // If we find the user in a persistence that does support migration, use
+    // that migration path (of only persistences that support migration)
+    const migrationHierarchy = persistences.filter(
+      p => p._shouldAllowMigration
+    );
+    chosenPersistence = migrationHierarchy[0];
```
yuchenshi, line 162: What if this is undefined?

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -113,35 +113,53 @@ export class PersistenceUserManager {
...
-      inMemoryPersistence
-    );
-    for (const persistence of persistenceHierarchy) {
-      if (await persistence._isAvailable()) {
-        chosenPersistence = persistence;
-        break;
-      }
-    }
+    // Eliminate any persistences that are not available
+    const persistences = (
```
yuchenshi, line 117: nit: rename to `availablePresistences`

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -113,35 +113,53 @@ export class PersistenceUserManager {
...
+    // persistence, we will (but only if that persistence supports migration).
     for (const persistence of persistenceHierarchy) {
-      // We attempt to call _get without checking _isAvailable since here we don't care if the full
-      // round-trip (read+write) is supported. We'll take the first one that we can read or give up.
       try {
-        const blob = await persistence._get<PersistedBlob>(key); // throws if unsupported
+        const blob = await persistence._get<PersistedBlob>(key);
         if (blob) {
           const user = UserImpl._fromJSON(auth, blob); // throws for unparsable blob (wrong format)
-          if (persistence !== chosenPersistence) {
```
yuchenshi, line 137: Please revert this -- the if-statement is intentional to avoid triggering dubious change events with the migration `._set()` below

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
```diff
@@ -113,35 +113,53 @@ export class PersistenceUserManager {
...
-            userToMigrate = user;
-          }
+          chosenPersistence = persistence;
```
yuchenshi, line 145: I'd suggest not assigning to `chosenPersistence` here since it is not the chosen persistence (but rather the persistence to migrate from). It is very confusing that `chosenPersistence` is then overwritten in L162 below. I'd suggest inlining `_shouldAllowMigration` check from below. Or just use a different variable name if you must.

yuchenshi (approved):
A few comments inline. I totally understand you're trying to improve the readability here -- and in that spirit I think we should split it into different helper functions.

Step 1: Choosing the preferred persistence (the first one that is available)
Step 2: Attempting to locate the user from all persistences (return user and source persistence)
Step 3 (can stay in this function): Migration logic (a. if user is in a persistence that doesn't support migration, use that instead and skip migration, or b. migrate if different)

With that being said, I don't feel like I have the authority for TypeScript readability to block this PR any longer. Approved with comments inline -- feel free to file a bug for later cleanup.

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
sam-gc: Fixed by just returning the manager if it is

packages-exp/auth-exp/src/core/persistence/persistence_user_manager.ts
sam-gc: Renamed to "selectedPersistence" as that accurately reflects what this code is doing

sam-gc:
After some tinkering I think refactoring into helper functions would make this code even more complex.

yuchenshi (approved):

PR #5245, opened 2021-08-04
Add back browser CJS build as ./dist/index.browser.cjs.js.

yuchenshi:
This adds back a browser CJS build, once removed by #5149 but still needed by Emulator UI. No changes to exports and package.json.

### DIFF
--- a/.changeset/gorgeous-balloons-confess.md
+++ b/.changeset/gorgeous-balloons-confess.md
@@ -2,4 +2,4 @@
 "@firebase/storage": patch
 ---
 
-Add back browser CJS build as ./dist/index.browser.cjs.js.
+Adds a browser CJS build as ./dist/index.browser.cjs.js.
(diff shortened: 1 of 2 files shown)


.changeset/gorgeous-balloons-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/storage": patch
+---
+
+Add back browser CJS build as ./dist/index.browser.cjs.js.
```
schmidt-sebastian, line 5:
```suggestion
Adds a browser CJS build as ./dist/index.browser.cjs.js.
```

schmidt-sebastian (approved):

Feiyang1 (approved):
Would you still require a cjs browser build when v9 comes out?

yuchenshi:
> Would you still require a cjs browser build when v9 comes out?

We haven't attempted upgrading Emulator UI to v9 and when we do we'd probably start with compat first. I'd assume we'd still need a CJS browser build though unless proven otherwise during the migration.

Feiyang1:
Okay. We can add the build if you need it.

Issue #5258, opened 2021-08-09
ReCaptchaV3Provider Typescript


zifahm:
Version 8.9.0

```

firebase
  .appCheck()
  .activate(
    new firebase.appCheck.ReCaptchaV3Provider(
      process.env.REACT_APP_RECAPTCHA_KEY
    ),
    true
  );

ts error: Argument of type 'ReCaptchaV3Provider' is not assignable to parameter of type 'AppCheckProvider'.
  Property 'getToken' is missing in type 'ReCaptchaV3Provider' but required in type 'AppCheckProvider'.
```

BTW the new version does not reflect the docs specified

Docs previous to version 8.9.0
```
firebase.initializeApp({
  // Your firebase configuration object
});

const appCheck = firebase.appCheck();
// Pass your reCAPTCHA v3 site key (public key) to activate(). Make sure this
// key is the counterpart to the secret key you set in the Firebase console.
appCheck.activate(
  'abcdefghijklmnopqrstuvwxy-1234567890abcd',

  // Optional argument. If true, the SDK automatically refreshes App Check
  // tokens as needed.
  true);
  ```

kroikie:
FYI: this is also showing up in our [snippets repo](https://github.com/firebase/snippets-web/pull/216/checks?check_run_id=3281393210#step:7:20) which was bumped to `8.9.0`

hsubox76:
There is a typings error in the main index.d.ts file. Fix is pending in https://github.com/firebase/firebase-js-sdk/pull/5261.

Both the provider argument and the string sitekey argument should be accepted (when the SDK is working correctly). Until the fix is released, you may be able to get around this temporarily with a `//@ts-ignore`.

PR #5261, opened 2021-08-09
Fix AppCheck typings error

hsubox76:
Incorrect typing of `activate()` params in `packages/firebase/index.d.ts` is causing users to get Typescript errors when using string or provider arguments.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5258

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -1580,14 +1580,18 @@ declare namespace firebase.appCheck {
   export interface AppCheck {
     /**
      * Activate AppCheck
-     * @param provider reCAPTCHA or custom token provider.
+     * @param provider reCAPTCHA provider, custom token provider, or reCAPTCHA site key.
      * @param isTokenAutoRefreshEnabled If true, the SDK automatically
      * refreshes App Check tokens as needed. If undefined, defaults to the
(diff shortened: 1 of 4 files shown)


packages/firebase/index.d.ts
```diff
@@ -1580,14 +1580,18 @@ declare namespace firebase.appCheck {
...
      * refreshes App Check tokens as needed. If undefined, defaults to the
      * value of `app.automaticDataCollectionEnabled`, which defaults to
      * false and can be set in the app config.
      */
     activate(
-      provider: AppCheckProvider,
+      provider:
+        | ReCaptchaV3Provider
+        | CustomProvider
+        | AppCheckProvider
```
Feiyang1, line 1593: There is no need to list the base type if we are enumerating all specific providers. That means we can just remove AppCheckProvider from the public API.

hsubox76: We were trying to make activate() completely backwards compatible, and the old API was that you can supply a siteKey (string) or for a custom provider you can supply an `AppCheckProvider` that only needs to be an object with a `getToken()` method, so I think that type has to stay. The new CustomProvider doesn't have the same API.

Feiyang1 (approved):

PR #5270, opened 2021-08-10
Fixing lint on packages/firestore

thebrianchen:
Not sure what happened, but the running the linter against the master branch results in a bunch of lint fixes.


schmidt-sebastian (approved):
This happens regularly when Prettier gets an update. Thanks for doing this.

PR #5271, opened 2021-08-10
Split storage and storage-compat

Feiyang1:


common/api-review/storage.api.md
```diff
@@ -18,6 +18,11 @@ import { Unsubscribe } from '@firebase/util';
 // @public
 export function connectStorageEmulator(storage: FirebaseStorage, host: string, port: number): void;
 
+// Warning: (ae-forgotten-export) The symbol "StringData" needs to be exported by the entry point index.d.ts
```
hsubox76, line 21: Do we need to fix these warnings?

packages/storage-compat/src/reference.ts
```diff
@@ -26,24 +26,27 @@ import {
...
+  SettableMetadata,
   _UploadTask,
   _getChild,
   _Reference,
-  _FbsBlob
-} from '../exp/api'; // import from the exp public API
+  _FbsBlob,
+  _dataFromString,
+  _invalidRootOperation
+} from '@firebase/storage'; // import from the exp public API
```
hsubox76, line 38: Comment seems outdated now.

packages/storage/src/implementation/observer.ts
```diff
@@ -81,8 +81,8 @@ export class Observer<T> implements StorageObserver<T> {
       isFunction(nextOrObserver) || error != null || complete != null;
     if (asFunctions) {
       this.next = nextOrObserver as NextFn<T>;
-      this.error = error;
-      this.complete = complete;
+      this.error = error || undefined;
```
hsubox76, line 84: Can it have a falsy value that isn't undefined? Isn't the param type `error?: ErrorFn`?

packages/storage/src/implementation/taskenums.ts
```diff
@@ -59,8 +59,9 @@ export const enum InternalTaskState {
 /**
  * Represents the current state of a running upload.
  */
-export type TaskState = string;
+export type TaskState = typeof TaskState[keyof typeof TaskState];
```
hsubox76, line 62: Oh that's better, thanks.

common/api-review/storage.api.md
Feiyang1: no, because these are from internal APIs.

packages/storage-compat/src/reference.ts
Feiyang1: removed

packages/storage/src/implementation/observer.ts
Feiyang1: our public API accepts `null` and `undefined`, but internal api accepts only `undefined`.
I don't think it can have other falsy value because it only function, null and undefined. But to be extra safe, updated to use `??`.

schmidt-sebastian:
Since @hsubox76 already started reviewing this, I will leave the rest to her. Ping me if you want additional eyes.

hsubox76 (approved):

packages/storage-compat/test/unit/service.test.ts
```diff
@@ -0,0 +1,233 @@
...
+        modularStorage,
+        'connectStorageEmulator'
+      ).callsFake(() => {});
+      const service = makeTestCompatStorage(fakeApp, fakeStorage);
+      service.useEmulator('test.host.org', 1234);
+
+      expect(connectStorageEmulatorStub).to.have.been.calledWithExactly(
+        fakeStorage,
+        'test.host.org',
+        1234
```
hsubox76, line 46:
```suggestion
        1234,
        {}
```
I ran into a test failure here while messing with this branch and I think the test needs this to pass because there's an `options` param with a default arg of `{}`.

packages/storage-compat/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "test:node": "TS_NODE_FILES=true TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file src/index.ts --config ../../config/mocharc.node.js",
+    "test:debug": "karma start --browser=Chrome",
+    "prettier": "prettier --write 'src/**/*.ts' 'test/**/*.ts'",
+    "add-compat-overloads": "ts-node-script ../../scripts/exp/create-overloads.ts -i ../storage/dist/storage-public.d.ts -o dist/src/index.d.ts -a -r FirebaseStorage:types.FirebaseStorage -r StorageReference:types.Reference -r FirebaseApp:FirebaseAppCompat --moduleToEnhance @firebase/storage"
+  },
+  "peerDependencies": {
+    "@firebase/app-compat": "0.x"
+  },
+  "dependencies": {
+    "@firebase/storage": "0.6.1",
```
hsubox76, line 32: Make sure this matches `packages/storage/pacakge.json` (currently 0.6.2, will change again Thursday 8/19)

hsubox76, line 32:
```suggestion
    "@firebase/storage": "0.6.2",
```

packages/storage-compat/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "test:node": "TS_NODE_FILES=true TS_NODE_CACHE=NO TS_NODE_COMPILER_OPTIONS='{\"module\":\"commonjs\"}' nyc --reporter lcovonly -- mocha 'test/{,!(browser)/**/}*.test.ts' --file src/index.ts --config ../../config/mocharc.node.js",
+    "test:debug": "karma start --browser=Chrome",
+    "prettier": "prettier --write 'src/**/*.ts' 'test/**/*.ts'",
+    "add-compat-overloads": "ts-node-script ../../scripts/exp/create-overloads.ts -i ../storage/dist/storage-public.d.ts -o dist/src/index.d.ts -a -r FirebaseStorage:types.FirebaseStorage -r StorageReference:types.Reference -r FirebaseApp:FirebaseAppCompat --moduleToEnhance @firebase/storage"
+  },
+  "peerDependencies": {
+    "@firebase/app-compat": "0.x"
+  },
+  "dependencies": {
+    "@firebase/storage": "0.6.2",
```
hsubox76, line 32:
```suggestion
    "@firebase/storage": "0.7.0",
```

PR #5273, opened 2021-08-11
Add support for signInWithCustomToken without refresh token

lisajian:
Adds support for signInWithCustomToken when the response does not have a refresh token. The call to GetAccountInfo is skipped and only the user's subject claim is used. This parallels the change made in cl/386333432. 

Some changes to highlight:
- A `isPassthroughMode` field was added to StsTokenManager to diverge behavior in `getToken()`. The `refreshToken` is used in `getTokenAndRefreshIfNeeded()` (i.e. the pre-existing code) as part of a token expired assert - `UserImpl.getIdToken()` uses errors from this call to logout the user for invalid states. To support both passthrough mode, which allows null refresh tokens, and to maintain pre-existing behavior, added the `isPassthroughMode` bit. A couple alternatives to this were:
  * Considered removing the `|| this.refreshToken` from the assert. However, consider the case in which there is an access token and a refresh token - this currently would not error, but with the change it would. Similarly, removing the assert altogether would impact the case when there is an access token but no refresh token - it currently errors, but removing the assert would not. 
   * One idea explored was using the callback that would be introduced in a later PR as part of the assert statement (e.g. `_assert((!this.accessToken || this.refreshToken) && !callback, ...)`); however, this conflicts with bullet point #2 in go/custom-auth-passthrough-sdk#heading=h.yjbziy33py3d
- Removed a redundant `_assert()` call in token_manager.ts
- Removed unused `ParsedIdToken`

**Corresponding internal bug: b/192699639**

### DIFF
--- a/packages-exp/auth-exp/src/core/user/id_token_result.ts
+++ b/packages-exp/auth-exp/src/core/user/id_token_result.ts
@@ -84,7 +84,6 @@ export async function getIdTokenResult(
     )!,
     signInProvider: signInProvider || null,
     signInSecondFactor: firebase?.['sign_in_second_factor'] || null
-    // TODO(lisajian): Add estimatedLocalExpirationTime
   };
 }
 
--- a/packages-exp/auth-exp/src/core/user/id_token_result.ts
+++ b/packages-exp/auth-exp/src/core/user/id_token_result.ts
@@ -84,6 +84,7 @@ export async function getIdTokenResult(
     )!,
     signInProvider: signInProvider || null,
     signInSecondFactor: firebase?.['sign_in_second_factor'] || null
+    // TODO(lisajian): Add estimatedLocalExpirationTime
   };
 }
 
--- a/packages-exp/auth-exp/src/core/user/reload.test.ts
+++ b/packages-exp/auth-exp/src/core/user/reload.test.ts
@@ -171,6 +171,7 @@ describe('core/user/reload', () => {
   it('nulls out properties in passthrough mode', async () => {
     const user = testUser(auth, 'abc', '');
     user.stsTokenManager.isPassthroughMode = true;
+    const getIdTokenSpy = sinon.spy(user, 'getIdToken');
 
     await _reloadWithoutSaving(user);
 
@@ -186,6 +187,7 @@ describe('core/user/reload', () => {
       createdAt: undefined,
       lastLoginAt: undefined
     });
+    expect(getIdTokenSpy).not.to.have.been.called;
   });
 
   context('anonymous carryover', () => {
--- a/packages-exp/auth-exp/src/core/user/reload.test.ts
+++ b/packages-exp/auth-exp/src/core/user/reload.test.ts
@@ -168,6 +168,26 @@ describe('core/user/reload', () => {
     expect(auth.persistenceLayer.lastObjectSet).to.eql(user.toJSON());
   });
 
+  it('nulls out properties in passthrough mode', async () => {
+    const user = testUser(auth, 'abc', '');
+    user.stsTokenManager.isPassthroughMode = true;
+
+    await _reloadWithoutSaving(user);
+
+    expect(user.uid).to.eq('abc');
+    expect(user.displayName).to.be.null;
+    expect(user.photoURL).to.be.null;
+    expect(user.email).to.be.null;
+    expect(user.emailVerified).to.be.false;
+    expect(user.phoneNumber).to.be.null;
+    expect(user.tenantId).to.be.null;
+    expect(user.providerData).to.eql([]);
+    expect((user.metadata as UserMetadata).toJSON()).to.eql({
+      createdAt: undefined,
+      lastLoginAt: undefined
+    });
+  });
+
   context('anonymous carryover', () => {
     let user: UserInternal;
     beforeEach(() => {
@@ -180,7 +200,7 @@ describe('core/user/reload', () => {
       providerData: Array<{ providerId: string }>
     ): void {
       // Get around readonly property
-      const mutUser = (user as unknown) as Record<string, unknown>;
+      const mutUser = user as unknown as Record<string, unknown>;
       mutUser.isAnonymous = isAnonStart;
       mutUser.email = emailStart;
 
--- a/packages-exp/auth-exp/src/core/user/reload.ts
+++ b/packages-exp/auth-exp/src/core/user/reload.ts
@@ -18,6 +18,7 @@
 import { User, UserInfo } from '../../model/public_types';
 
 import {
+  APIUserInfo,
   getAccountInfo,
   ProviderUserInfo
 } from '../../api/account_management/account';
@@ -30,15 +31,22 @@ import { getModularInstance } from '@firebase/util';
 
 export async function _reloadWithoutSaving(user: UserInternal): Promise<void> {
   const auth = user.auth;
-  const idToken = await user.getIdToken();
-  const response = await _logoutIfInvalidated(
-    user,
-    getAccountInfo(auth, { idToken })
-  );
+  let coreAccount: APIUserInfo = {};
 
-  _assert(response?.users.length, auth, AuthErrorCode.INTERNAL_ERROR);
+  if (user.stsTokenManager.isPassthroughMode) {
+    // The uid === localId === id token's `sub` claim
+    coreAccount.localId = user.uid;
+  } else {
+    const idToken = await user.getIdToken();
+    const response = await _logoutIfInvalidated(
+      user,
+      getAccountInfo(auth, { idToken })
+    );
 
-  const coreAccount = response.users[0];
+    _assert(response?.users.length, auth, AuthErrorCode.INTERNAL_ERROR);
+
+    coreAccount = response.users[0];
+  }
 
   user._notifyReloadListener(coreAccount);
 
@@ -46,7 +54,9 @@ export async function _reloadWithoutSaving(user: UserInternal): Promise<void> {
     ? extractProviderData(coreAccount.providerUserInfo)
     : [];
 
-  const providerData = mergeProviderData(user.providerData, newProviderData);
+  const providerData = !user.stsTokenManager.isPassthroughMode
+    ? mergeProviderData(user.providerData, newProviderData)
+    : [];
 
   // Preserves the non-nonymous status of the stored user, even if no more
   // credentials (federated or email/password) are linked to the user. If
--- a/packages-exp/auth-exp/src/core/user/reload.ts
+++ b/packages-exp/auth-exp/src/core/user/reload.ts
@@ -54,9 +54,9 @@ export async function _reloadWithoutSaving(user: UserInternal): Promise<void> {
     ? extractProviderData(coreAccount.providerUserInfo)
     : [];
 
-  const providerData = !user.stsTokenManager.isPassthroughMode
-    ? mergeProviderData(user.providerData, newProviderData)
-    : [];
+  const providerData = user.stsTokenManager.isPassthroughMode
+    ? []
+    : mergeProviderData(user.providerData, newProviderData);
 
   // Preserves the non-nonymous status of the stored user, even if no more
   // credentials (federated or email/password) are linked to the user. If
--- a/packages-exp/auth-exp/src/core/user/token_manager.ts
+++ b/packages-exp/auth-exp/src/core/user/token_manager.ts
@@ -71,7 +71,7 @@ export class StsTokenManager {
     forceRefresh = false
   ): Promise<string | null> {
     if (this.isPassthroughMode) {
-      return this.getTokenAndTriggerCallback(/* auth, forceRefresh */);
+      return this.getTokenAndTriggerCallback(auth, forceRefresh);
     } else {
       return this.getTokenAndRefreshIfNeeded(auth, forceRefresh);
     }
@@ -100,7 +100,10 @@ export class StsTokenManager {
   }
 
   // TODO(lisajian): Fill in once refresh listener is added
-  private async getTokenAndTriggerCallback(): Promise<string | null> {
+  private async getTokenAndTriggerCallback(
+    auth: AuthInternal,
+    forceRefresh = false
+  ): Promise<string | null> {
     return null;
   }
 
--- a/packages-exp/auth-exp/src/core/user/token_manager.ts
+++ b/packages-exp/auth-exp/src/core/user/token_manager.ts
@@ -101,8 +101,8 @@ export class StsTokenManager {
 
   // TODO(lisajian): Fill in once refresh listener is added
   private async getTokenAndTriggerCallback(
-    auth: AuthInternal,
-    forceRefresh = false
+    _auth: AuthInternal,
+    _forceRefresh = false
   ): Promise<string | null> {
     return null;
   }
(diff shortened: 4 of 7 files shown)


packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -54,14 +55,6 @@ export class StsTokenManager {
     response: IdTokenResponse | FinalizeMfaResponse
   ): void {
     _assert(response.idToken, AuthErrorCode.INTERNAL_ERROR);
-    _assert(
```
avolkovi, line 57: maybe keep these assertions and additionally assert that if things are undefined then we are in pass through mode?

lisajian: Good q - so regarding the first part (and please correct me if I'm wrong!), the set of idTokens that would fail the first assertion (L56) is a superset of the set of idTokens that would fail the second assertion (L57-L60), so the second assertion can be removed safely. For the second part of the question and regarding the third assertion (L61-L64), we know that we're in passthrough mode when there is no refresh token returned and to account for that newly valid case, we need to remove the assert - not sure if there's another way of additionally asserting we're in passthrough mode you're referring to?

sam-gc: @lisajian I think your assessment is right here. The second assertion is definitely redundant and there's no other indicator for being in passthrough mode as far as I can tell

sam-gc (approved):
Looks good!

packages-exp/auth-exp/src/core/user/id_token_result.ts
```diff
@@ -84,6 +84,7 @@ export async function getIdTokenResult(
     )!,
     signInProvider: signInProvider || null,
     signInSecondFactor: firebase?.['sign_in_second_factor'] || null
+    // TODO(lisajian): Add estimatedLocalExpirationTime
```
rosalyntan, line 87: You can remove this TODO -- we opted not to add this field.

lisajian: Sick, done

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -76,6 +69,17 @@ export class StsTokenManager {
...
+      return this.getTokenAndTriggerCallback(/* auth, forceRefresh */);
+    }
+  }
+
+  private async getTokenAndRefreshIfNeeded(
+    auth: AuthInternal,
+    forceRefresh = false
   ): Promise<string | null> {
     _assert(
       !this.accessToken || this.refreshToken,
```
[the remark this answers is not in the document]
lisajian: Hmm good q good q, taking a look at where `getToken()` is called - it's only called in `UserImpl.getIdToken()` and the latter case would still be caught by an assert (see [this line](https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/auth-exp/src/core/user/user_impl.ts#L99)), though the user would not be logged out. I don't have enough context around if this is the expected behavior for when both the access token and refresh token are not present though - perhaps @avolkovi / @sam-gc can weigh in?

As for the impact on how the `refreshToken` check for passthrough happens, could you expand a bit more on this? The only case in which we'd hit the incorrect branching logic in `getToken()` is if a call to `getToken()` happens prior to a call to `updateFromServerResponse()`, but at least for `signInWithCustomToken()`, this would not be the case. But please let me know if there are any blindspots I'm missing!

packages-exp/auth-exp/src/core/user/reload.ts
```diff
@@ -30,23 +31,32 @@ import { getModularInstance } from '@firebase/util';
...
+  }
 
   user._notifyReloadListener(coreAccount);
 
   const newProviderData = coreAccount.providerUserInfo?.length
     ? extractProviderData(coreAccount.providerUserInfo)
     : [];
 
-  const providerData = mergeProviderData(user.providerData, newProviderData);
+  const providerData = !user.stsTokenManager.isPassthroughMode
```
avolkovi, line 57: nit: i would remove the `!` and just switch the order of the if/else:

```
const providerData = user.stsTokenManager.isPassthroughMode ? [] : mergeProviderData(user.providerData, newProviderData);
```
a little easier to reason about

packages-exp/auth-exp/src/core/user/reload.test.ts
```diff
@@ -168,6 +168,26 @@ describe('core/user/reload', () => {
     expect(auth.persistenceLayer.lastObjectSet).to.eql(user.toJSON());
   });
 
+  it('nulls out properties in passthrough mode', async () => {
```
avolkovi, line 171: can you assert getIdToken/logout not called?

packages-exp/auth-exp/src/core/user/token_manager.ts
avolkovi: ack

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -76,6 +69,17 @@ export class StsTokenManager {
   async getToken(
     auth: AuthInternal,
     forceRefresh = false
+  ): Promise<string | null> {
+    if (!this.isPassthroughMode) {
```
avolkovi, line 73: remove the `!` and switch the order of the if/else

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -76,6 +69,17 @@ export class StsTokenManager {
...
+      return this.getTokenAndRefreshIfNeeded(auth, forceRefresh);
+    } else {
+      return this.getTokenAndTriggerCallback(/* auth, forceRefresh */);
```
avolkovi, line 76: why the comment?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -42,6 +42,7 @@ export class StsTokenManager {
   refreshToken: string | null = null;
   accessToken: string | null = null;
   expirationTime: number | null = null;
+  isPassthroughMode: boolean = false;
```
avolkovi, line 45: since this state is equivalent to `this.refreshToken == null` do we need the extra field?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -147,21 +158,33 @@ export class StsTokenManager {
       );
       manager.expirationTime = expirationTime;
     }
+    if (isPassthroughMode) {
```
avolkovi, line 161: is this necessary or can we pull the state from checking refreshToken?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -147,21 +158,33 @@ export class StsTokenManager {
...
+        {
+          appName
+        }
+      );
+      manager.isPassthroughMode = isPassthroughMode;
+    }
     return manager;
   }
 
   toJSON(): object {
```
avolkovi, line 174: tests for to/from JSON?

packages-exp/auth-exp/src/model/id_token.ts
```diff
@@ -23,27 +23,6 @@ import { PhoneOrOauthTokenResponse } from '../api/authentication/mfa';
  */
 export type IdToken = string;
 
-/**
- * Raw parsed JWT
- *
- */
-export interface ParsedIdToken {
```
avolkovi, line 30: was this not used?

avolkovi (approved):

packages-exp/auth-exp/src/model/id_token.ts
lisajian: Yep! Wasn't used

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Yeah, this goes back to the question of "is the absence of a refresh token a sufficient indicator that we're in passthrough mode" and unfortunately, it isn't. Ties into one of the discussion threads from an earlier commit

packages-exp/auth-exp/src/core/user/token_manager.ts
sam-gc: I believe as a reminder for the next PR which will introduce these parameters. (@lisajian correct me if I'm wrong)

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Similar to other discussion threads - it's possible that `refreshToken == null` even when we're not in passthrough mode

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Yeah! Sorry that it was unclear - the getTokenAndTriggerCallback will take in these args, but I was getting compiler errors when I left them as part of the method signature since they're currently unused/unimplemented. See future PR for when these are used

avolkovi: you can put a `_` before the variable name to signify it is unused https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals

packages-exp/auth-exp/src/core/user/token_manager.ts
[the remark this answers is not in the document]
sam-gc: This code is mirroring this bit in the original SDK:
https://github.com/firebase/firebase-js-sdk/blob/a69220b519185239c364f7a5b3d5983e9c87d7c1/packages/auth/src/token.js#L279-L299

I'm not entirely sure _why_ we need to check for the presence of `accessToken` as well, but we just copied that bit over. In this case though it's fine, as this function is the one that gets called when we're _not_ in passthrough mode. In this code, `getTokenAndRefreshIfNeeded()` is called for non-passthrough, while `getTokenAndTriggerCallback()` is the equivalent (unimplemented so far) for passthrough mode. (@lisajian please correct me if I'm wrong)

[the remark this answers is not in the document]
lisajian: Cool, thanks for the link to the OG expiration check! And yep ditto^^, that all checks out on my end

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Good catch, added / updated tests in token_manager.test.ts

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Done!

packages-exp/auth-exp/src/core/user/reload.test.ts
lisajian: You got it! Did so for `getIdToken()`

packages-exp/auth-exp/src/core/user/reload.ts
lisajian: Fair, done!

packages-exp/auth-exp/src/core/user/reload.ts
```diff
@@ -30,23 +31,32 @@ import { getModularInstance } from '@firebase/util';
...
 
   user._notifyReloadListener(coreAccount);
 
   const newProviderData = coreAccount.providerUserInfo?.length
     ? extractProviderData(coreAccount.providerUserInfo)
     : [];
 
-  const providerData = mergeProviderData(user.providerData, newProviderData);
+  const providerData = user.stsTokenManager.isPassthroughMode
+    ? []
```
yuchenshi, line 58: No `custom` provider?

packages-exp/auth-exp/src/core/user/token_manager.ts
yuchenshi: +1 and the warning should be addressed within `getTokenAndTriggerCallback`, not here

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -95,6 +99,11 @@ export class StsTokenManager {
     return null;
   }
 
+  // TODO(lisajian): Fill in once refresh listener is added
+  private async getTokenAndTriggerCallback(): Promise<string | null> {
```
yuchenshi, line 103: This should have the right signature (and then ignore unused warnings).

packages-exp/auth-exp/src/core/user/token_manager.ts
yuchenshi: This if-clause will not handle `false`. Is this expected? If not, try `isPassthroughMode != null`

packages/rules-unit-testing/src/api/index.ts
```diff
@@ -160,7 +160,7 @@ export type FirebaseEmulatorOptions = {
 
 function trimmedBase64Encode(val: string): string {
   // Use base64url encoding and remove padding in the end (dot characters).
-  return base64Encode(val).replace(/\./g, "");
+  return base64Encode(val).replace(/\./g, '');
```
yuchenshi, line 163: Okay, somehow this file is not properly formatted. Can we address this in a different PR though? We're not actually changing RUT here.

yuchenshi (approved):

packages-exp/auth-exp/src/core/user/token_manager.ts
rosalyntan: Out of curiosity -- when will the `refreshToken` be missing _without_ being in passthrough mode? I'm not sure which earlier thread you're referencing (unless it's the one about the `TOKEN_EXPIRED` error). On Android, at least, I've been treating a missing refresh token as an indicator of passthrough mode.

rosalyntan: Ah, found this line in the old SDK: https://github.com/firebase/firebase-js-sdk/blob/a69220b519185239c364f7a5b3d5983e9c87d7c1/packages/auth/src/token.js#L272

packages-exp/auth-exp/src/core/user/reload.ts
lisajian: That I understand, the providerData is something that we would get from `getAccountInfo`, which we're trying to circumvent in passthrough mode. I think you might be referring to the `CustomTokenProvider` (correct me if I'm wrong!) but I think the provider data and the custom token provider are only similar by name (overloading on "provider"). The `CustomTokenProvider` will be added in a subsequent PR!

rosalyntan (approved):

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: SG, added method signature back and noted for future!

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: @yuchenshi - Yep! Should be fine since isPassthroughMode is false by default

@rosalyntan - Yeah, thanks for linking!

packages/rules-unit-testing/src/api/index.ts
lisajian: Ah, the works of an overly aggressive VSCode formatting, undoing and will send out a separate PR!

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Sick, done!

PR #5276, opened 2021-08-11
Split database and database-compat

Feiyang1:


Feiyang1:
Moved some tests to `@firebase/database-compat`, because they use the compat API for testing. Most of them should be probably refactored and moved back to `@firebase/database` because they are really tests for the underlying functionalities, like transactions. And tests in `@firebase/database-compat` should only tests for the compat layer that it delegates correctly to the modular API.

It seems to be a non-trivial amount of work, and I don't think it's urgent right now, so maybe we can do it after GA? wdyt
@schmidt-sebastian

schmidt-sebastian:
The idea behind testing the compat layer was that it allowed us to not rewrite the tests, while ensuring that both the compat and the -exp layer get tested thoroughly. It gets a bit awkward now that the packages are split, so it is worth reevaluating.

For now, we can leave it as is as long as the `database-compat` tests are triggered for every `database` change.

packages/database-compat/karma.conf.js
```diff
@@ -0,0 +1,36 @@
...
+const karma = require('karma');
+const path = require('path');
+const karmaBase = require('../../config/karma.base');
+
+const files = [`test/**/*.test.ts`];
+
+module.exports = function (config) {
+  const karmaConfig = Object.assign({}, karmaBase, {
+    // files to load into karma
+    files: files,
```
schmidt-sebastian, line 27: Nit: This can be simplified.

packages/database/rollup.config.js
```diff
@@ -20,9 +20,10 @@ import typescriptPlugin from 'rollup-plugin-typescript2';
 import typescript from 'typescript';
 import pkg from './package.json';
 
-const deps = Object.keys(
-  Object.assign({}, pkg.peerDependencies, pkg.dependencies)
-);
+const deps = [
+  ...Object.keys(Object.assign({}, pkg.peerDependencies, pkg.dependencies)),
```
schmidt-sebastian, line 24:
```
...Object.keys({...pkg.peerDependencies, ...pkg.dependencies}),
```

schmidt-sebastian (approved):
LGTM (based on trust and some spot checking). Please fix remaining lint and build problems.

PR #5279, opened 2021-08-11
Make initializeAuth() idempotent

hsubox76:
Making all product initialize() functions idempotent.
See https://github.com/firebase/firebase-js-sdk/pull/5272

### DIFF
--- a/packages-exp/auth-exp/src/core/auth/initialize.ts
+++ b/packages-exp/auth-exp/src/core/auth/initialize.ts
@@ -16,6 +16,7 @@
  */
 
 import { _getProvider, FirebaseApp } from '@firebase/app-exp';
+import { deepEqual } from '@firebase/util';
 import { Auth, Dependencies } from '../../model/public_types';
 
 import { AuthErrorCode } from '../errors';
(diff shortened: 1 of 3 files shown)


sam-gc (approved):

packages-exp/auth-exp/src/core/auth/initialize.ts
```diff
@@ -54,7 +55,12 @@ export function initializeAuth(app: FirebaseApp, deps?: Dependencies): Auth {
 
   if (provider.isInitialized()) {
     const auth = provider.getImmediate() as AuthImpl;
-    _fail(auth, AuthErrorCode.ALREADY_INITIALIZED);
+    const initialOptions = provider.getOptions() as Dependencies;
+    if (deepEqual(initialOptions, deps ?? {})) {
```
Feiyang1, line 59: I thought we are going to use reference equality for comparing builtin objects like `popupRedirectResolver`s and `persistence`s. Using deepEqual feels unnecessary.

hsubox76: I was mainly using deepEqual to avoid extra lines of code, but also, while 2 of the 3 options (`popupResolver` and `errorMap`) can be compared using a simple `===`, `persistence` can be an array, and on top of that, an array where order matters.

Feiyang1 (approved):

PR #5286, opened 2021-08-12
Fix deprecation documentation

zwu52:
Deprecated method doc should start with "Use ...".

This will fix: b/170682742

### DIFF
--- a/packages/firebase/index.d.ts
+++ b/packages/firebase/index.d.ts
@@ -7419,7 +7419,7 @@ declare namespace firebase.messaging {
      *
      * @param callback The function to handle the push message.
      *
-     * @deprecated onBackgroundMessage(nextOrObserver: firebase.NextFn<MessagePayload>|
+     * @deprecated Use onBackgroundMessage(nextOrObserver: firebase.NextFn<MessagePayload>|
      * firebase.Observer<MessagePayload>, error?: firebase.ErrorFn,completed?: firebase.CompleteFn):
      * firebase.Unsubscribe.
      */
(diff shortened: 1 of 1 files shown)


egilmorez (approved):
Should this be backticked, or will it be rendered as code font without them?

packages/firebase/index.d.ts
```diff
@@ -7419,7 +7419,7 @@ declare namespace firebase.messaging {
      *
      * @param callback The function to handle the push message.
      *
-     * @deprecated onBackgroundMessage(nextOrObserver: firebase.NextFn<MessagePayload>|
+     * @deprecated Use onBackgroundMessage(nextOrObserver: firebase.NextFn<MessagePayload>|
```
egilmorez, line 7422: Also, will this fix the way this item gets conflated with the following Observer-Unsubscribe item?

Ideally we'd have a way to gen the docs and eyeball that.

hsubox76: Those all belong together, the Observer and error are part of the signature and the Unsubscribe is the return type. There could be some spaces before the `|` and after the comma after `errorFn,`. But actually, do we need the whole signature here, can we just say "Use onBackgroundMessage()" or "Use {@link firebase.messaging.Messaging.onBackgroundMessage}"?

Docs can be generated locally by running
```
yarn
yarn docgen:js
```

 at the root of the repo.
 
 After running that, the affected html file will be found at `scripts/docgen/html/js/firebase.messaging.Messaging.html`

zwu52: added some spaces.

Don't know if should use the full method signature of `{@link firebase.messaging.Messaging.onBackgroundMessage}`. If later, need to change other ones.

Issue #4725, opened 2021-04-04
Error in alpha during building in node environment: Directory import is not supported resolving ES modules


Evertt:
* Operating System version: macOS Big Sur
  * Browser version: irrelevant
  * Firebase SDK version: 0.9.x
  * Firebase Product: all of them

### [REQUIRED] Describe the problem
[I encountered this issue while trying to build a SvelteKit project that uses firebase.](https://github.com/sveltejs/kit/issues/612)

During the build process I get the following error:

```
> Directory import '/Users/evert/Sites/svind/node_modules/firebase/app' is not supported resolving ES modules imported from /Users/evert/Sites/svind/.svelte/output/server/app.js
Did you mean to import firebase@0.900.20/node_modules/firebase/app/dist/index.cjs.js?
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/evert/Sites/svind/node_modules/firebase/app' is not supported resolving ES modules imported from /Users/evert/Sites/svind/.svelte/output/server/app.js
Did you mean to import firebase@0.900.20/node_modules/firebase/app/dist/index.cjs.js?
    at new NodeError (node:internal/errors:329:5)
    at finalizeResolution (node:internal/modules/esm/resolve:319:17)
    at moduleResolve (node:internal/modules/esm/resolve:758:10)
    at Loader.defaultResolve [as _resolve] (node:internal/modules/esm/resolve:869:11)
    at Loader.resolve (node:internal/modules/esm/loader:86:40)
    at Loader.getModuleJob (node:internal/modules/esm/loader:230:28)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:57:40)
    at link (node:internal/modules/esm/module_job:56:36)
```

#### Steps to reproduce:
I don't know any other way to reproduce it, other than just by creating a svelte-kit project, importing firebase and trying to build it.

### Solution

I already know the solution though. You see, RxJS had the same problem and they recently fixed it. It's quite easy. Just add the following lines to the root `package.json`:

```
  "exports": {
    "./app": {
      "node": "./app/dist/index.cjs.js",
      "default": "./app/dist/index.esm.js"
    },
    "./auth": {
      "node": "./auth/dist/index.cjs.js",
      "default": "./auth/dist/index.esm.js"
    },
    "./firestore": {
      "node": "./firestore/dist/index.cjs.js",
      "default": "./firestore/dist/index.esm.js"
    },
    "./functions": {
[...]

PR #5296, opened 2021-08-13
Update package.json

jacoblukewood:
Required for Vite as per https://github.com/vitejs/vite/issues/1505

#4725 added the exports as needed but this one was missing and did not allow projects such as those using SvelteKit to build.

### DIFF
--- a/packages-exp/firebase-exp/package.json
+++ b/packages-exp/firebase-exp/package.json
@@ -24,6 +24,10 @@
     "compat/index.d.ts"
   ],
   "exports": {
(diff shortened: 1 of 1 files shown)


Feiyang1:
Thanks for the contribution! I don't think it's needed though, because you will never use `firebase` as an import in v9. Instead, you will only use submodule imports like `firebase/app`, `firebase/firestore`, etc.

I have tested the v9 SDK with sveltekit and it worked correctly. If it is not the case, can you help me understand your use case?

PR #5300, opened 2021-08-13
Add setCustomTokenProvider and clearCustomTokenProvider for passthrough mode

lisajian:
Added API changes for passthrough mode. The `customTokenProvider` is invoked in place of requesting a new refresh token. A word on how `getTokenAndRefreshIfNeeded` and `getTokenAndTriggerCallback` relate to each other to illustrate that they parallel each other but the former performs a refresh where the latter invokes the callback:

In `getTokenAndRefreshIfNeeded`, the refresh endpoint is hit when `this.refreshToken && (forceRefresh || !this.accessToken || this.isExpired)`. Breaking this down a bit...

1. `forceRefresh` - Callback is called in `getTokenAndTriggerCallback` so long as it is present. This is done in the first `if` block in `getTokenAndTriggerCallback`
2. `!this.accessToken` - This happens when there is no `accessToken` to begin with and indicates the client has not triggered the first call to `signInWithCustomToken`. This would be effectively the same as returning `null` like in `getTokenAndRefreshIfNeeded`
3. `this.isExpired` - Captured in the last `if` block of `getTokenAndTriggerCallback`. The logic diverges a bit here to account for cases when the time is within the buffer window and there is / isn't a callback present.

These changes parallel this change in the Android SDK: cl/388558525

### Testing

* Ran `yarn test` in `packages-exp/auth-exp` and passed
* Ran `yarn test` in `packages-exp/auth-compat-exp` and passed

### API Changes

The changes below are in accordance with [go/custom-auth-passthrough-sdk#heading=h.t4hz4i7fojar](http://go/custom-auth-passthrough-sdk#heading=h.t4hz4i7fojar)
  * Added `setCustomTokenProvider` and `clearCustomTokenProvider` to `auth-exp/src/core/strategies/custom_token`
  * Added `setCustomTokenProvider` and `clearCustomTokenProvider` to `auth-compat-exp/src/auth`

### DIFF
--- a/common/api-review/auth-exp.api.md
+++ b/common/api-review/auth-exp.api.md
@@ -286,6 +286,11 @@ export function createUserWithEmailAndPassword(auth: Auth, email: string, passwo
 export type CustomParameters = Record<string, string>;
 
 // @public
+export interface CustomTokenProvider {
+    getCustomToken(): Promise<string>;
+}
+
+// @public
 export const debugErrorMap: AuthErrorMap;
 
 // @public
@@ -694,9 +699,7 @@ export function sendPasswordResetEmail(auth: Auth, email: string, actionCodeSett
 export function sendSignInLinkToEmail(auth: Auth, email: string, actionCodeSettings: ActionCodeSettings): Promise<void>;
 
 // @public
-export function setCustomTokenProvider(auth: Auth, provider: {
-    getCustomToken(): Promise<string>;
-}): void;
+export function setCustomTokenProvider(auth: Auth, provider: CustomTokenProvider): void;
 
 // @public
 export function setPersistence(auth: Auth, persistence: Persistence): Promise<void>;
--- a/common/api-review/auth-exp.api.md
+++ b/common/api-review/auth-exp.api.md
@@ -250,6 +250,9 @@ export const browserSessionPersistence: Persistence;
 // @public
 export function checkActionCode(auth: Auth, oobCode: string): Promise<ActionCodeInfo>;
 
+// @public
+export function clearCustomTokenProvider(auth: Auth): void;
+
 export { CompleteFn }
 
 // @public
@@ -689,6 +692,11 @@ export function sendPasswordResetEmail(auth: Auth, email: string, actionCodeSett
 
 // @public
 export function sendSignInLinkToEmail(auth: Auth, email: string, actionCodeSettings: ActionCodeSettings): Promise<void>;
+
+// @public
+export function setCustomTokenProvider(auth: Auth, provider: {
+    getCustomToken(): Promise<string>;
+}): void;
 
 // @public
 export function setPersistence(auth: Auth, persistence: Persistence): Promise<void>;
--- a/packages-exp/auth-exp/src/core/auth/auth_impl.ts
+++ b/packages-exp/auth-exp/src/core/auth/auth_impl.ts
@@ -29,7 +29,8 @@ import {
   CompleteFn,
   ErrorFn,
   NextFn,
-  Unsubscribe
+  Unsubscribe,
+  CustomTokenProvider
 } from '../../model/public_types';
 import {
   createSubscribe,
@@ -100,7 +101,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   tenantId: string | null = null;
   settings: AuthSettings = { appVerificationDisabledForTesting: false };
 
-  _customTokenProvider: { getCustomToken(): Promise<string> } | null = null;
+  _customTokenProvider: CustomTokenProvider | null = null;
 
   constructor(
     public readonly app: FirebaseApp,
--- a/packages-exp/auth-exp/src/core/auth/auth_impl.ts
+++ b/packages-exp/auth-exp/src/core/auth/auth_impl.ts
@@ -29,8 +29,7 @@ import {
   CompleteFn,
   ErrorFn,
   NextFn,
-  Unsubscribe,
-  CustomTokenProvider
+  Unsubscribe
 } from '../../model/public_types';
 import {
   createSubscribe,
@@ -40,7 +39,11 @@ import {
   Subscribe
 } from '@firebase/util';
 
-import { AuthInternal, ConfigInternal } from '../../model/auth';
+import {
+  AuthInternal,
+  ConfigInternal,
+  RefreshWithCustomToken
+} from '../../model/auth';
 import { PopupRedirectResolverInternal } from '../../model/popup_redirect';
 import { UserInternal } from '../../model/user';
 import {
@@ -101,7 +104,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   tenantId: string | null = null;
   settings: AuthSettings = { appVerificationDisabledForTesting: false };
 
-  _customTokenProvider: CustomTokenProvider | null = null;
+  _refreshWithCustomTokenProvider: RefreshWithCustomToken | null = null;
 
   constructor(
     public readonly app: FirebaseApp,
--- a/packages-exp/auth-exp/src/core/auth/auth_impl.ts
+++ b/packages-exp/auth-exp/src/core/auth/auth_impl.ts
@@ -100,6 +100,8 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   tenantId: string | null = null;
   settings: AuthSettings = { appVerificationDisabledForTesting: false };
 
+  _customTokenProvider: { getCustomToken(): Promise<string> } | null = null;
+
   constructor(
     public readonly app: FirebaseApp,
     public readonly config: ConfigInternal
--- a/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
+++ b/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
@@ -27,7 +27,11 @@ import { Endpoint } from '../../api';
 import { APIUserInfo } from '../../api/account_management/account';
 import { IdTokenResponse, IdTokenResponseKind } from '../../model/id_token';
 import { UserCredentialInternal } from '../../model/user';
-import { signInWithCustomToken } from './custom_token';
+import {
+  clearCustomTokenProvider,
+  setCustomTokenProvider,
+  signInWithCustomToken
+} from './custom_token';
 
 use(chaiAsPromised);
 
@@ -67,31 +71,59 @@ describe('core/strategies/signInWithCustomToken', () => {
   });
   afterEach(mockFetch.tearDown);
 
-  it('should return a valid user credential', async () => {
-    const {
-      user,
-      operationType,
-      _tokenResponse
-    } = (await signInWithCustomToken(
-      auth,
-      'look-at-me-im-a-jwt'
-    )) as UserCredentialInternal;
-    expect(_tokenResponse).to.eql(idTokenResponse);
-    expect(user.uid).to.eq('local-id');
-    expect(user.displayName).to.eq('display-name');
-    expect(operationType).to.eq(OperationType.SIGN_IN);
-  });
+  describe('#signInWithCustomToken', () => {
+    it('should return a valid user credential', async () => {
+      const { user, operationType, _tokenResponse } =
+        (await signInWithCustomToken(
+          auth,
+          'look-at-me-im-a-jwt'
+        )) as UserCredentialInternal;
+      expect(_tokenResponse).to.eql(idTokenResponse);
+      expect(user.uid).to.eq('local-id');
+      expect(user.displayName).to.eq('display-name');
+      expect(operationType).to.eq(OperationType.SIGN_IN);
+    });
 
-  it('should send with a valid request', async () => {
-    await signInWithCustomToken(auth, 'j.w.t');
-    expect(signInRoute.calls[0].request).to.eql({
-      token: 'j.w.t',
-      returnSecureToken: true
+    it('should send with a valid request', async () => {
+      await signInWithCustomToken(auth, 'j.w.t');
+      expect(signInRoute.calls[0].request).to.eql({
+        token: 'j.w.t',
+        returnSecureToken: true
+      });
+    });
+
+    it('should update the current user', async () => {
+      const { user } = await signInWithCustomToken(auth, 'oh.no');
+      expect(auth.currentUser).to.eq(user);
     });
   });
 
-  it('should update the current user', async () => {
-    const { user } = await signInWithCustomToken(auth, 'oh.no');
-    expect(auth.currentUser).to.eq(user);
+  describe('#setCustomTokenProvider', () => {
+    it('sets a custom token provider', () => {
+      const provider = {
+        async getCustomToken() {
+          return Promise.resolve('');
+        }
+      };
+
+      setCustomTokenProvider(auth, provider);
+
+      expect(auth._customTokenProvider).to.eql(provider);
+    });
+  });
+
+  describe('#clearCustomTokenProvider', () => {
+    it('clears the custom token provider', () => {
+      const provider = {
+        async getCustomToken() {
+          return Promise.resolve('');
+        }
+      };
+      auth._customTokenProvider = provider;
+
+      clearCustomTokenProvider(auth);
+
+      expect(auth._customTokenProvider).to.be.null;
+    });
   });
 });
--- a/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
+++ b/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
@@ -101,22 +101,22 @@ describe('core/strategies/signInWithCustomToken', () => {
   describe('#setCustomTokenProvider', () => {
     it('sets a custom token provider', () => {
       const provider = {
-        async getCustomToken() {
-          return Promise.resolve('');
+        async getCustomToken(): Promise<string> {
+          return '';
         }
       };
 
       setCustomTokenProvider(auth, provider);
 
-      expect(auth._customTokenProvider).to.eql(provider);
+      expect(auth._customTokenProvider).to.eq(provider);
     });
   });
 
   describe('#clearCustomTokenProvider', () => {
     it('clears the custom token provider', () => {
       const provider = {
-        async getCustomToken() {
-          return Promise.resolve('');
+        async getCustomToken(): Promise<string> {
+          return '';
         }
       };
       setCustomTokenProvider(auth, provider);
--- a/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
+++ b/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
@@ -138,7 +138,7 @@ describe('core/strategies/signInWithCustomToken', () => {
       ).to.have.been.calledWith(idTokenResponse);
     });
 
-    it('_refreshWithCustomTokenProvider should error when uid does not match localId', () => {
+    it('_refreshWithCustomTokenProvider should error when uid does not match localId', async () => {
       const provider = {
         async getCustomToken(): Promise<string> {
           return 'custom-token';
@@ -149,9 +149,9 @@ describe('core/strategies/signInWithCustomToken', () => {
       spy(user.stsTokenManager, 'updateFromServerResponse');
 
       setCustomTokenProvider(auth, provider);
-      expect(auth._refreshWithCustomTokenProvider!()).to.be.rejectedWith(
+      await expect(auth._refreshWithCustomTokenProvider!()).to.be.rejectedWith(
         FirebaseError,
-        'auth/user-token-expired'
+        'Firebase: Error (auth/internal-error).'
       );
     });
   });
--- a/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
+++ b/packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
@@ -21,7 +21,7 @@ import * as chaiAsPromised from 'chai-as-promised';
 import { OperationType } from '../../model/enums';
 
 import { mockEndpoint } from '../../../test/helpers/api/helper';
-import { testAuth, TestAuth } from '../../../test/helpers/mock_auth';
+import { testAuth, TestAuth, testUser } from '../../../test/helpers/mock_auth';
 import * as mockFetch from '../../../test/helpers/mock_fetch';
 import { Endpoint } from '../../api';
 import { APIUserInfo } from '../../api/account_management/account';
@@ -32,7 +32,11 @@ import {
   setCustomTokenProvider,
   signInWithCustomToken
 } from './custom_token';
+import { spy } from 'sinon';
+import * as sinonChai from 'sinon-chai';
+import { FirebaseError } from '@firebase/util';
 
+use(sinonChai);
 use(chaiAsPromised);
 
 describe('core/strategies/signInWithCustomToken', () => {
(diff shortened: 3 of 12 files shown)


common/api-review/auth-exp.api.md
```diff
@@ -250,6 +250,9 @@ export const browserSessionPersistence: Persistence;
 // @public
 export function checkActionCode(auth: Auth, oobCode: string): Promise<ActionCodeInfo>;
 
+// @public
+export function clearCustomTokenProvider(auth: Auth): void;
```
lisajian, line 254: Should this bump the changeset? Not sure what the process is for doing so

packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -103,6 +101,8 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   tenantId: string | null = null;
   settings: AuthSettings = { appVerificationDisabledForTesting: false };
 
+  _customTokenProvider: { getCustomToken(): Promise<string> } | null = null;
```
avolkovi, line 104: nit: define a type for this, something like `export interface CustomTokenProvider { getCustomToken(): Promise<string> }` in https://github.com/firebase/firebase-js-sdk/blob/master/packages-exp/auth-exp/src/model/public_types.ts

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -67,31 +71,59 @@ describe('core/strategies/signInWithCustomToken', () => {
...
-    const { user } = await signInWithCustomToken(auth, 'oh.no');
-    expect(auth.currentUser).to.eq(user);
+  describe('#clearCustomTokenProvider', () => {
+    it('clears the custom token provider', () => {
+      const provider = {
+        async getCustomToken() {
+          return Promise.resolve('');
+        }
+      };
+      auth._customTokenProvider = provider;
```
avolkovi, line 122: nit: go ahead and call setCustomTokenProvider in your setup here, that's one less private accessor that your test will need to depend on

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -67,31 +71,59 @@ describe('core/strategies/signInWithCustomToken', () => {
...
-      _tokenResponse
-    } = (await signInWithCustomToken(
-      auth,
-      'look-at-me-im-a-jwt'
-    )) as UserCredentialInternal;
-    expect(_tokenResponse).to.eql(idTokenResponse);
-    expect(user.uid).to.eq('local-id');
-    expect(user.displayName).to.eq('display-name');
-    expect(operationType).to.eq(OperationType.SIGN_IN);
+  describe('#signInWithCustomToken', () => {
```
avolkovi, line 74: if I set a customTokenProvider and then call signInWithCustomToken, do I still have to pass a token? Is there not a version of signInWithCustomToken that will use my provider?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -99,18 +100,47 @@ export class StsTokenManager {
...
+
+    // isExpired includes a buffer window, during which the access token will still be returned even
+    // if a customTokenProvider is not set yet.
+    if (this.isExpired) {
+      _assert(
+        this.expirationTime && Date.now() < this.expirationTime!,
+        auth,
+        AuthErrorCode.TOKEN_EXPIRED
+      );
+      if (!auth._customTokenProvider) {
```
avolkovi, line 121: these few lines may be simpler as `return auth._customTokenProvider ? this.refreshWithCustomToken(auth) : this.accessToken`

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -99,18 +100,47 @@ export class StsTokenManager {
...
-    return null;
+    if (forceRefresh) {
+      // TODO(lisajian): Add and use client error code - ERROR_TOKEN_REFRESH_UNAVAILABLE
+      _assert(auth._customTokenProvider, auth, AuthErrorCode.INTERNAL_ERROR);
+      return this.refreshWithCustomToken(auth);
+    }
+
+    // isExpired includes a buffer window, during which the access token will still be returned even
+    // if a customTokenProvider is not set yet.
+    if (this.isExpired) {
```
avolkovi, line 115: you can have a lot less indentation if you swap things around here a bit and change this if to `if(!this.isExpired) { return this.accessToken; }`

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -99,18 +100,47 @@ export class StsTokenManager {
...
 
   clearRefreshToken(): void {
     this.refreshToken = null;
   }
 
+  private async refreshWithCustomToken(
+    auth: AuthInternal
+  ): Promise<string | null> {
+    const token = await auth._customTokenProvider!.getCustomToken();
+    // signInWithCustomToken creates a new User and subsequently new StsTokenManager, so
```
avolkovi, line 138: this feels like a memory leak in the making, can we not have it do that in pass through mode?

common/api-review/auth-exp.api.md
sam-gc: No, though we will want to assign @egilmorez to review the doc comments

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
rosalyntan: Yeah, that's correct. There is not a version of `signInWithCustomToken` that doesn't take in a parameter. We could add one in the future, but that would need to go through another API review.

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +60,30 @@ export async function signInWithCustomToken(
   await authInternal._updateCurrentUser(cred.user);
   return cred;
 }
+
+/**
+ * Sets a custom token provider, which is invoked when a new Firebase ID token is requested and no
+ * refresh token is present (i.e., in passthrough mode). This callback should be implemented to
```
sam-gc, line 66: Break the new sentence into a new paragraph, separated from the top line by `@remarks`

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +60,30 @@ export async function signInWithCustomToken(
...
+ * obtain a custom token to exchange for a new Firebase ID token.
+ *
+ * @param auth - The Auth instance.
```
sam-gc, line 69: Could you add a short snippet example here? For example:
https://github.com/firebase/firebase-js-sdk/blob/f825b1d83228747e404aaefd9f3948d12257d0fc/packages-exp/auth-exp/src/core/auth/initialize.ts#L44-L48

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -67,31 +71,59 @@ describe('core/strategies/signInWithCustomToken', () => {
...
   });
 
-  it('should update the current user', async () => {
-    const { user } = await signInWithCustomToken(auth, 'oh.no');
-    expect(auth.currentUser).to.eq(user);
+  describe('#clearCustomTokenProvider', () => {
+    it('clears the custom token provider', () => {
+      const provider = {
+        async getCustomToken() {
+          return Promise.resolve('');
```
sam-gc, line 119: No need to `Promise.resolve()` here... the `async` keyword on the function makes that implied. You can just write:
```ts
async getCustomToken(): Promise<string> {
  return '';
}
```

Also, the getCustomToken() function should have a return type defined. I'm surprised the linter isn't complaining

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +193,137 @@ describe('core/user/token_manager', () => {
...
+      };
+      const idTokenResponse: IdTokenResponse = {
+        idToken: 'new-access-token',
+        expiresIn: '3600',
+        localId: serverUser.localId!,
+        kind: IdTokenResponseKind.CreateAuthUri
+      };
+      const fakeCustomToken = 'fake-custom-token';
+      const provider = {
+        async getCustomToken() {
```
sam-gc, line 216: Same comments here

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +193,137 @@ describe('core/user/token_manager', () => {
...
+      it('throws an error if the token is expired beyond the buffer window', async () => {
+        Object.assign(stsTokenManager, {
+          accessToken: 'old-access-token',
+          expirationTime: now - 1,
+          isPassthroughMode: true
+        });
+
+        await expect(stsTokenManager.getToken(auth)).to.be.rejectedWith(
+          FirebaseError,
+          "Firebase: The user's credential is no longer valid. The user must sign in again. (auth/user-token-expired)"
```
sam-gc, line 311: nit: just use the error code substring here:
```ts
to.be.rejectedWith(FirebaseError, 'auth/user-token-expired')
```

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -99,18 +100,46 @@ export class StsTokenManager {
...
+    }
+
+    if (!this.isExpired) {
+      return this.accessToken;
+    }
+
+    // isExpired includes a buffer window, during which the access token will still be returned even
+    // if a customTokenProvider is not set yet.
+    _assert(
+      this.expirationTime && Date.now() < this.expirationTime!,
```
sam-gc, line 120: You shouldn't need the `!` on `this.expirationTime!` since the left-hand branch of the conditional checks for its presence

packages-exp/auth-exp/src/core/user/token_manager.ts
sam-gc: I have an idea about how to do this... let's discuss in our meeting

packages/auth-types/index.d.ts
```diff
@@ -398,6 +398,10 @@ export interface EmulatorConfig {
   };
 }
 
+export interface CustomTokenProvider {
```
sam-gc, line 401: Hold off on the changes to this file.. I'll explain at our meeting

packages-exp/auth-exp/src/core/strategies/custom_token.ts
lisajian: Cool, added the one [go/firebase-auth-passthrough-sdk-api#heading=h.4ej4y6o7kzi5](http://go/firebase-auth-passthrough-sdk-api#heading=h.4ej4y6o7kzi5)

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Done!

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Ooh, got it, copied and pasted from another test in the file, will update that test too!

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Nice catch, done!

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Chatted with Sam offline - there's going to be some churn on this PR, heads up to any folks that are taking a look right now, please hold off until the next revision (i.e. when there are 4 commits on this PR), thank you!

lisajian: Updated in most recent revision - moved refresh logic to custom_token and called API directly rather than signInWithCustomToken. Thanks for walking me through it Sam, and thank you Alex for highlighting the issue!

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -21,14 +21,22 @@ import * as chaiAsPromised from 'chai-as-promised';
...
 import { APIUserInfo } from '../../api/account_management/account';
 import { IdTokenResponse, IdTokenResponseKind } from '../../model/id_token';
 import { UserCredentialInternal } from '../../model/user';
-import { signInWithCustomToken } from './custom_token';
+import {
+  clearCustomTokenProvider,
+  setCustomTokenProvider,
+  signInWithCustomToken
+} from './custom_token';
+import { spy } from 'sinon';
```
sam-gc, line 35: small nit-- for consistency with other tests, please `import * as sinon from 'sinon'` and use `sinon.spy()` in tests

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -67,31 +75,99 @@ describe('core/strategies/signInWithCustomToken', () => {
   });
   afterEach(mockFetch.tearDown);
```
sam-gc, line 76: Now that you're using `sinon.spy()`, we'll need to add a `sinon.restore()` call to the `afterEach` hook

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +63,54 @@ export async function signInWithCustomToken(
...
+ * @param provider - The callback that is invoked when a new Firebase ID token is requested and no
+ * refresh token is present.
+ */
+export function setCustomTokenProvider(
+  auth: Auth,
+  provider: CustomTokenProvider
+): void {
+  const authInternal = _castAuth(auth);
```
sam-gc, line 91: Nice, this looks great and is fully tree-shakeable. @rosalyntan this is the new implementation we chatted about yesterday

sam-gc: Please call `getModularInstance` on the auth variable first, then call `_castAuth` with the result. For example:
https://github.com/firebase/firebase-js-sdk/blob/2901ed05e946d7010443fd3fa48b4aa3395e5463/packages-exp/auth-exp/src/core/index.ts#L57-L62

Users can pass the compat layer auth instance to the modular functions (for migrations) so the first step is to unwrap if necessary

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +63,54 @@ export async function signInWithCustomToken(
...
+  };
+}
+
+/**
+ * Removes the current custom token provider that was set by {@link setCustomTokenProvider}.
+ *
+ * @param auth - The Auth instance.
+ */
+export function clearCustomTokenProvider(auth: Auth): void {
+  const authInternal = _castAuth(auth);
```
sam-gc, line 114: Same here

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +189,97 @@ describe('core/user/token_manager', () => {
...
+          isPassthroughMode: true
+        });
+        setCustomTokenProvider(auth, provider);
+        stub(auth, '_refreshWithCustomTokenProvider').returns(
+          Promise.resolve('new-access-token')
+        );
+
+        const tokens = (await stsTokenManager.getToken(auth, true))!;
+
+        expect(tokens).to.eq('new-access-token');
```
lisajian, line 214: Nit: Noticed that I'm using eq() and eql() interchangeably for string comparisons. I don't think it matters in this case but I'll update the tests for token_manager to be more consistent in the next revision

sam-gc: Yeah it doesn't really matter for strings, don't worry too much if it's not always consistent

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
lisajian: Ooh, nice, done!

packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
avolkovi: ack

packages-exp/auth-exp/src/model/auth.ts
```diff
@@ -48,6 +48,8 @@ export interface ConfigInternal extends Config {
   clientPlatform: ClientPlatform;
 }
 
+export type RefreshWithCustomToken = () => Promise<string | null>;
```
avolkovi, line 51: CustomTokenProvider?

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -28,6 +28,8 @@ import { IdTokenResponse } from '../../model/id_token';
 import { StsTokenManager, Buffer } from './token_manager';
 import { FinalizeMfaResponse } from '../../api/authentication/mfa';
 import { makeJWT } from '../../../test/helpers/jwt';
+import { setCustomTokenProvider } from '../strategies/custom_token';
+import { stub } from 'sinon';
```
sam-gc, line 32: same here re. `sinon.stub()`

avolkovi (approved):

packages-exp/auth-exp/src/model/public_types.ts
```diff
@@ -1190,3 +1190,14 @@ export interface Dependencies {
    */
   errorMap?: AuthErrorMap;
 }
+
+/**
+ * Custom token provider, which is invoked when a new Firebase ID token is requested and no refresh
+ * token is present (i.e., in passthrough mode).
```
sam-gc, line 1196: Mark this as `@public`:

```suggestion
 * token is present (i.e., in passthrough mode).
 *
 * @public
```

sam-gc (approved):

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Whoops, missed that, thanks!

packages-exp/auth-exp/src/model/auth.ts
lisajian: Yeah, good q! So I wanted to distinguish it from the CustomTokenProvider itself - this refers to the wrapper around the CustomTokenProvider that also calls the signIn endpoint and updates the stsTokenManager

packages-exp/auth-exp/src/model/public_types.ts
lisajian: Sick, thanks!

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +63,54 @@ export async function signInWithCustomToken(
   await authInternal._updateCurrentUser(cred.user);
   return cred;
 }
+
+/**
+ * Sets a custom token provider.
```
kevinthecheung, line 68: Maybe something like, "Sets the custom token provider to be used by the given Auth instance."

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +63,54 @@ export async function signInWithCustomToken(
...
+ *
+ * @remarks
+ * The custom token provider is invoked when a new Firebase ID token is requested and no
+ * refresh token is present (i.e., in passthrough mode). This callback should be implemented to
+ * obtain a custom token to exchange for a new Firebase ID token. For instance,
```
kevinthecheung, line 73:
```suggestion
 * refresh token is present (such as in passthrough mode). The provider object must implement a
 * `getCustomToken()` callback, which obtains a custom token to exchange for a new Firebase ID 
 * token. For instance:
```

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -56,3 +63,54 @@ export async function signInWithCustomToken(
...
+ * setCustomTokenProvider(auth, {
+ *   async getCustomToken(): Promise<string> {
+ *     return requestNewCustomToken();
+ *   }
+ * });
+ * ```
+ *
+ * @param auth - The Auth instance.
+ * @param provider - The callback that is invoked when a new Firebase ID token is requested and no
+ * refresh token is present.
```
kevinthecheung, line 85:
```suggestion
 * @param provider - A `CustomTokenProvider` object, which implements a callback that is invoked 
 * when a new Firebase ID token is requested and no refresh token is present.
```

packages-exp/auth-exp/src/model/public_types.ts
```diff
@@ -1190,3 +1190,16 @@ export interface Dependencies {
...
+ *
```
kevinthecheung, line 1197:
```suggestion
 * Interface representing a custom token provider, which implements a callback that is invoked when
 * a new Firebase ID token is requested and no refresh token is present (such as in passthrough 
 * mode).
 *
```

packages-exp/auth-exp/src/core/strategies/custom_token.ts
lisajian: Cool, done!

packages-exp/auth-exp/src/model/public_types.ts
lisajian: Thank you!! Updated

kevinthecheung (approved):

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +188,97 @@ describe('core/user/token_manager', () => {
...
+        sinon
+          .stub(auth, '_refreshWithCustomTokenProvider')
+          .returns(Promise.resolve('new-access-token'));
+
+        const tokens = (await stsTokenManager.getToken(auth))!;
+
+        expect(tokens).to.eq('new-access-token');
+      });
+
+      it('throws an error if the token is expired beyond the buffer window', async () => {
```
rosalyntan, line 257: Can you also add a test for when the token is expired, but there is a `CustomTokenProvider` set (unless you already have this test case and I just missed it)?

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +188,97 @@ describe('core/user/token_manager', () => {
...
+        Object.assign(stsTokenManager, {
+          accessToken: 'old-access-token',
+          expirationTime: now - 1,
+          isPassthroughMode: true
+        });
+
+        await expect(stsTokenManager.getToken(auth)).to.be.rejectedWith(
+          FirebaseError,
+          'auth/user-token-expired'
```
rosalyntan, line 266: Just wanted to note that this should be `token-refresh-unavailable` -- I know you haven't implemented the new error codes in this PR yet, but just wanted to flag this!

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +188,97 @@ describe('core/user/token_manager', () => {
...
+        );
+      });
+
+      it('returns access token when not expired, not refreshing', async () => {
```
rosalyntan, line 270: nit: Maybe also a test for when the token is not expired, and there is a `CustomTokenProvider`, to verify that no new token is requested?

packages-exp/auth-exp/src/core/user/token_manager.ts
```diff
@@ -99,12 +99,34 @@ export class StsTokenManager {
...
+
+    // isExpired includes a buffer window, during which the access token will still be returned even
+    // if a customTokenProvider is not set yet.
+    _assert(
+      this.expirationTime && Date.now() < this.expirationTime,
+      auth,
+      AuthErrorCode.TOKEN_EXPIRED
+    );
+    return auth._refreshWithCustomTokenProvider
+      ? auth._refreshWithCustomTokenProvider()
```
rosalyntan, line 128: Does there need to be an `_assert` here too to check if `_refreshWithCustomTokenProvider` is set when the access token is expired?

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Oh sick, thanks! I'll leave this thread open and resolve it once I fix #5323

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Sick, added!

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Makes sense, added!

packages-exp/auth-exp/src/core/user/token_manager.ts
lisajian: Ah, so this is the case when we're in a buffer window (token is not yet expired but will expire in `Buffer.TOKEN_REFRESH`) - if `_refreshWithCustomTokenProvider` is not set, we just return `this.accessToken`

packages-exp/auth-exp/src/core/user/token_manager.test.ts
```diff
@@ -187,6 +188,130 @@ describe('core/user/token_manager', () => {
...
+          isPassthroughMode: true
+        });
+
+        await expect(stsTokenManager.getToken(auth)).to.be.rejectedWith(
+          FirebaseError,
+          'auth/user-token-expired'
+        );
+      });
+
+      it('throws an error if the token is expired beyond the buffer window, even if provider is set', async () => {
```
rosalyntan, line 270: Ah, so if the token is expired and the provider is set, it should use the provider to obtain a new token.

(i.e., the custom token provider should be used to "refresh" any time the old refresh flow would have occurred)

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Updated as of https://github.com/firebase/firebase-js-sdk/pull/5323/commits/9fa20b135f57e3e3be30ae347a45fe195215e42a

packages-exp/auth-exp/src/core/user/token_manager.test.ts
lisajian: Ah shoot, you're right, I misinterpreted the DD, updated the logic a bit in token_manager and adjusted tests!

rosalyntan (approved):

Issue #5304, opened 2021-08-15
Uncaught (in promise) FirebaseError: Firebase: Error (auth/account-exists-with-different-credential).


DominicTobias:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS
  * Browser version: Chrome 92
  * Firebase SDK version: 9.0.0-beta.7
  * Firebase Product: auth

### [REQUIRED] Describe the problem

After a redirect login to a social provider (`signInWithRedirect(auth, provider)`), if the user already has another provider linked we will get the error `auth/account-exists-with-different-credential`.

```ts
const auth = getAuth(firebaseApp)
getRedirectResult(auth).catch(err => {
  console.error(err)
})
```

We can handle that error to link accounts.

However internally Firebase throws the error anyway and breaks the app:

```
errors.ts:91 Uncaught (in promise) FirebaseError: Firebase: Error (auth/account-exists-with-different-credential).
    at createErrorInternal (assert.ts:95)
    at _createError (assert.ts:81)
    at _makeTaggedError (index.ts:257)
    at _performFetchWithErrorHandling (index.ts:148)
    at async _performSignInRequest (index.ts:187)
    at async _signInWithCredential (credential.ts:41)
    at async RedirectAction.onAuthEvent (abstract_popup_redirect_operation.ts:102)
```

![Screenshot 2021-08-15 at 14 10 41](https://user-images.githubusercontent.com/760314/129479829-227a9963-f527-4407-a737-d2c8d7d40e20.png)

#### Steps to reproduce:

- Try to connect to an oAuth account when there is already one on the account. Upon redirecting back Firebase will throw an error that I haven't managed to handle (tried ErrorBoundaries, try/catches etc). In development mode frameworks like Create React App and NextJS will display unhandled errors in the browser, blocking the UI.

#### Solution

Firebase shouldn't throw an unhandled error that we can't intercept handle ourselves. Gracefully handling `auth/account-exists-with-different-credential` is a legitimate use case to prevent unhandled mayhem.

Thanks!

DominicTobias:
On a side note we are expected to link accounts by getting existing providers `fetchSignInMethodsForEmail(auth, email)` and using `linkWithCredential(result.user, credential)` after authenticating with one of them. However Firebase doesn't provide an idiomatic way to retrieve an email, only to retrieve a credential:

```ts
const credential = OAuthProvider.credentialFromError(err)
[...]

PR #5308, opened 2021-08-16
[Auth] Remove background error throwing in initialization

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5304

We were holding on to any redirect error and throwing it after initialization completed. Unfortunately since init is totally async, there was no way this was actionable for developers. The error can instead be gotten by calling `getRedirectResult()`, while initialization simply swallows any errors in the redirect flow. This matches v8 behavior.

### DIFF
--- a/packages-exp/auth-exp/src/core/auth/auth_impl.ts
+++ b/packages-exp/auth-exp/src/core/auth/auth_impl.ts
@@ -79,7 +79,6 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   private idTokenSubscription = new Subscription<User>(this);
   private redirectUser: UserInternal | null = null;
   private isProactiveRefreshEnabled = false;
-  private redirectInitializerError: Error | null = null;
 
   // Any network calls will set this to true and prevent subsequent emulator
   // initialization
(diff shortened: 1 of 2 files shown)


packages-exp/auth-exp/src/core/auth/auth_impl.ts
```diff
@@ -267,7 +260,8 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
         true
       );
     } catch (e) {
-      this.redirectInitializerError = e;
+      // Swallow any errors here; the code can retrieve them in
+      // getRedirectResult().
       await this._setRedirectUser(null);
```
avolkovi, line 265: do you have a test for the catch?

avolkovi (approved):

yuchenshi (approved):

packages-exp/auth-exp/src/core/auth/auth_impl.ts
sam-gc: Ya there is a test for this code but the error was never caught since the test uses the initialization function that leaves the promise hanging. I've updated this test so that it would catch an error being thrown :+1:

avolkovi: ty!

PR #5318, opened 2021-08-17
Make node and browser entry points identical

sam-gc:
This change uses `@rollup/plugin-alias` to overwrite the browser-specific imports to point to a set of no-ops and errors. I also had to change a bunch of the testing machinery to make this work.


avolkovi (approved):

yuchenshi (approved):

packages-exp/auth-exp/index.ts
```diff
@@ -47,64 +36,65 @@ export {
 // Core functionality shared by all clients
 export * from './src';
 
-// Additional DOM dependend functionality
+// Additional DOM dependend functionality; we need to import and then
+// export separately so that the rollup alias will work (for aliasing these
```
Feiyang1, line 40: interesting. Is it WAI, or a bug in the plugin? If it's a bug, it may be worth linking to the issue.

packages-exp/auth-exp/rollup.config.shared.js
```diff
@@ -37,6 +38,21 @@ const commonPlugins = [
   })
 ];
 
+/**
+ * Node has the same entry point as browser, but browser-specific exports need
```
Feiyang1, line 42:
```suggestion
 * Node has the same entry point as browser, but browser-specific exports
```

Feiyang1 (approved):

packages-exp/auth-exp/index.ts
sam-gc: It's WAI I think. It uses `resolveId` which is import statements [according to rollup](https://rollupjs.org/guide/en/#resolveid). See: https://github.com/rollup/rollup-plugin-alias/blob/master/src/index.js

PR #5335, opened 2021-08-20
Allow publishing from any branch

hsubox76:
This works in tandem with a change to the internal `release-cli` script (that branch is currently named `custombranch`).

The change to that script is, if publishing from `master`, the release branch is named `release`. If publishing from any other branch, the release branch is named `[branch name]-releasebranch`. I was going to make the suffix `-release` but I did a search and it's not uncommon to end dev branches with `-release` if we're working on the release process. So trying to make something more unique.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,10 @@
 name: Release
 
-on:
-  push:
-    branches:
-      - release
+on: push
 
 jobs:
   release:
+    if: github.ref == 'release' || contains(github.ref, '-release')
     name: Release
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,12 +1,10 @@
...
 
-on:
-  push:
-    branches:
-      - release
+on: push
 
 jobs:
   release:
+    if: github.ref == 'refs/heads/release' || endsWith(github.ref, '-releasebranch')
```
Feiyang1, line 7: Nice. Didn't know that GHA provides convenient util functions.

Feiyang1 (approved):

PR #5336, opened 2021-08-20
Create FCM Interop for v9 Function

zwu52:
b/197336448


packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -64,6 +55,38 @@ const WindowMessagingFactory: InstanceFactory<'messaging-exp'> = (
   return messaging;
 };
 
+const WindowMessagingInternalFactory: InstanceFactory<'messaging-exp-internal'> =
+  (container: ComponentContainer) => {
+    maybeThrowWindowError();
+
+    const messaging = _getProvider(
```
hsubox76, line 62: I think this can just be
```
container.getProvider('messaging-exp').getImmediate() as MessagingService
```
similar to https://github.com/firebase/firebase-js-sdk/blob/4d747b158021766b21f37879054c45ddd17051ff/packages-exp/app-check-exp/src/index.ts#L74

I'm not sure if installations has a reason for doing it the other way.

packages-exp/messaging-exp/src/interfaces/internal-dependencies.ts
```diff
@@ -27,3 +27,5 @@ export interface FirebaseInternalDependencies {
   installations: _FirebaseInstallationsInternal;
   analyticsProvider: Provider<FirebaseAnalyticsInternalName>;
 }
+
+export interface MessagingInternal {}
```
hsubox76, line 31: Shouldn't it have the `getToken()` method?

packages/messaging-interop-types/READNE.md
```diff
@@ -0,0 +1,4 @@
+# @firebase/messaging-interop-types
```
hsubox76, line 1: The filename is misspelled (N -> M)

packages/messaging-interop-types/index.d.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
hsubox76, line 3: 2021

packages/messaging-interop-types/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export interface MessagingInternal {
+  getToken(messaging: Messaging, options?: GetTokenOptions): Promise<string>;
```
hsubox76, line 19: Messaging instance shouldn't be a param.

packages/messaging-interop-types/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+interface Messaging {}
```
hsubox76, line 22: Once removed from above, this shouldn't be necessary.

packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -64,6 +55,38 @@ const WindowMessagingFactory: InstanceFactory<'messaging-exp'> = (
   return messaging;
 };
 
+const WindowMessagingInternalFactory: InstanceFactory<'messaging-exp-internal'> =
```
hsubox76, line 58: Looks like the name is `messaging-internal` in other places.

zwu52: good catch. The latest committed was pushed 😬

packages/messaging-interop-types/READNE.md
zwu52: good catch ✋

packages-exp/messaging-exp/src/interfaces/internal-dependencies.ts
zwu52: 👍

packages-exp/messaging-exp/src/helpers/register.ts
zwu52: good call.

packages-exp/messaging-exp/src/helpers/register.ts
```diff
@@ -29,27 +29,18 @@ import {
   onSubChange
 } from '../listeners/sw-listeners';
 
+import { GetTokenOptions } from '../interfaces/public-types';
+import { MessagingInternal } from '../interfaces/internal-dependencies';
```
hsubox76, line 33: I think the CI error is because the source code never imports `@firebase/messaging-interop-types`, which is where the `declare` block that adds `messaging-internal` to `NameServiceMapping` is. I think a good place to import it is here, import the `MessagingInternal` type from that package instead of from `internal-dependencies`, and maybe it can be removed from `internal-dependencies`.

zwu52: got you. You are right. I am able to build now

hsubox76 (approved):

packages/messaging-interop-types/package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "@firebase/messaging-interop-types",
+    "version": "0.1.0",
```
Feiyang1, line 3:
```suggestion
    "version": "0.0.1",
```

we will have a changeset to bump minor, so it will be released as `0.1.0`

Feiyang1 (approved):

PR #5343, opened 2021-08-21
Implement most of RUTv2 features.

yuchenshi:


packages/rules-unit-testing/src/impl/test_environment.ts
```diff
@@ -0,0 +1,257 @@
...
+    this.app = undefined;
+  }
+
+  firestore(settings?: FirestoreSettings): Firestore {
+    assertEmulatorRunning(this.emulators, 'firestore');
+    const firestoreCompat = this.getApp().firestore!();
+    if (settings) {
+      firestoreCompat.settings(settings);
+    }
+    const firestore = firestoreCompat as unknown as Firestore;
```
sam-gc, line 179: This shouldn't be necessary :/

packages/rules-unit-testing/src/util.ts
```diff
@@ -70,7 +70,34 @@ export async function withFunctionTriggersDisabled<TResult>(
...
 export function assertFails(pr: Promise<any>): Promise<any> {
-  throw new Error('unimplemented');
+  return pr.then(
+    () => {
+      return Promise.reject(
+        new Error('Expected request to fail, but it succeeded.')
+      );
+    },
+    (err: any) => {
+      const errCode = (err && err.code && err.code.toLowerCase()) || '';
```
sam-gc, line 80: Can just do `err?.code?.toLowerCase() || ''`

packages/rules-unit-testing/src/util.ts
```diff
@@ -70,7 +70,34 @@ export async function withFunctionTriggersDisabled<TResult>(
...
+      const errMessage =
+        (err && err.message && err.message.toLowerCase()) || '';
```
sam-gc, line 82: Same here

sam-gc (approved):

packages/rules-unit-testing/src/impl/test_environment.ts
yuchenshi: Fixed after rebasing v9 changes in

packages/rules-unit-testing/src/util.ts
yuchenshi: Done

packages/rules-unit-testing/src/util.ts
yuchenshi: Done

Issue #5346, opened 2021-08-22
[Types][v9] Missing [api].app typings for Auth, Performance, and Remote Config 


sujishpatel:
### [REQUIRED] Describe your environment

  * Operating System version: _____
  * Browser version: _____
  * Firebase SDK version: _9.0.0-2021720181311_
  * Firebase Product: _auth, performance, remote-config_

### [REQUIRED] Describe the problem

I was wondering if there is a plan to include `[api].app`  typings for all v9 apis?

Currently auth, performance, and remote-config don’t have it available in their typings, but they do implement it. 

Normally I would assume a typing not being included means it is intended to be private, but it doesn’t follow your typical convention of using an underscore (`_`) to denote a private property/method.

#### Relevant Code:

```javascript
console.log({ auth: getAuth() });
console.log({ performance: getPerformance() });
console.log({ remoteconfig: getRemoteConfig() });
```
![Screen Shot 2021-08-22 at 2 34 30 AM](https://user-images.githubusercontent.com/2984350/130345111-443889fd-02a4-45b5-80e8-6e1060ec71e7.png)
```js
// Typescript Errors
getAuth().app // Property 'app' does not exist on type 'Auth'.ts(2339)
getPerformance().app // Property 'app' does not exist on type 'FirebasePerformance'.ts(2339)
getRemoteConfig().app // Property 'app' does not exist on type 'RemoteConfig'.ts(2339)
```

hsubox76:
Thanks, I think this is an omission and it should be added to our typings files.

PR #5351, opened 2021-08-23
Add `app` property to public typings of all services

hsubox76:
Every service has an `app` property publicly exposed so we should probably have the typings for that. Went through and added the typing to products that did not have it.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5346


jposuna (approved):
LGTM for perf. Thanks for making the changes!

packages/auth/src/model/public_types.ts
```diff
@@ -170,12 +171,14 @@ export interface AuthSettings {
...
-  /** The name of the app associated with the Auth service instance. */
+  /** The {@link @firebase/app#FirebaseApp} associated with the `Auth` service instance. */
+  readonly app: FirebaseApp;
+  /** The name of the app associated with the `Auth` service instance. */
   readonly name: string;
   /** The {@link Config} used to initialize this instance. */
   readonly config: Config;
   /**
-   * Changes the type of persistence on the {@link Auth} instance.
+   * Changes the type of persistence on the `Auth` instance.
```
Feiyang1, line 181: revert?

packages/storage/src/public-types.ts
```diff
@@ -31,7 +31,7 @@ import {
  */
 export interface FirebaseStorage extends _FirebaseService {
   /**
-   * FirebaseApp associated with this StorageService instance.
+   * The `FirebaseApp` associated with this `FirebaseStorage` instance.
```
Feiyang1, line 34:
```suggestion
   * The {@link @firebase/app#FirebaseApp} associated with this `FirebaseStorage` instance.
```

packages/auth/src/model/public_types.ts
hsubox76: This is inside the `Auth` interface so it would just link to itself which would be weird navigation.

Feiyang1 (approved):

PR #5357, opened 2021-08-23
Move Messaging env checks to public methods

hsubox76:
Ensure messaging env checks are only called when the user explicitly initializes messaging. If messaging is initialized by an internal API (Functions specifically) in an unsupported environment, it will leave ugly warnings that aren't catchable because the check is async and leaves a hanging promise. Since the user did not call messaging, the user can't do anything about it, and the warning is not helpful.

Also in Node 16, it actually exits and blocks further execution.

### DIFF
--- a/packages/messaging/src/api.ts
+++ b/packages/messaging/src/api.ts
@@ -31,9 +31,11 @@ import {
 import { MessagingService } from './messaging-service';
 import { deleteToken as _deleteToken } from './api/deleteToken';
 import { getToken as _getToken } from './api/getToken';
+import { isSwSupported, isWindowSupported } from './api/isSupported';
 import { onBackgroundMessage as _onBackgroundMessage } from './api/onBackgroundMessage';
 import { onMessage as _onMessage } from './api/onMessage';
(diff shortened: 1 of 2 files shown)


packages/messaging/src/api.ts
```diff
@@ -43,6 +45,22 @@ import { _setDeliveryMetricsExportedToBigQueryEnabled } from './api/setDeliveryM
...
+      if (!isSupported) {
+        throw ERROR_FACTORY.create(ErrorCode.UNSUPPORTED_BROWSER);
+      }
+    })
+    .catch((e: Error) => {
+      if (e.message.includes(ErrorCode.UNSUPPORTED_BROWSER)) {
+        throw e;
+      }
+      throw ERROR_FACTORY.create(ErrorCode.INDEXED_DB_UNSUPPORTED);
+    });
```
Feiyang1, line 63: We can just use the error callback of `then`, which is much simpler.

hsubox76: Done.

Feiyang1 (approved):

PR #5359, opened 2021-08-23
Storage typing updates

Feiyang1:


common/api-review/storage.api.md
```diff
@@ -133,13 +128,13 @@ export function getMetadata(ref: StorageReference): Promise<FullMetadata>;
 // @public
 export function getStorage(app?: FirebaseApp, bucketUrl?: string): FirebaseStorage;
 
-// Warning: (ae-forgotten-export) The symbol "FirebaseStorageError" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "StorageError" needs to be exported by the entry point index.d.ts
```
hsubox76, line 131: I guess it should be exported? `src/index.dts` only exports one thing from public-types, I'm not sure why. I think some of the types in public-types were previously exported from the code itself (I remember there was a duplicate of some of the interfaces that was removed during the refactor), maybe that's why

Feiyang1: We have the public interface `StorageError` exported. The class itself is only used in internal APIs, so we don't have to export it.

hsubox76 (approved):

PR #5360, opened 2021-08-23
Implement rest of RUTv2 features.

yuchenshi:
This completes all rules unit testing v2 features.


packages/rules-unit-testing/src/impl/test_environment.ts
```diff
@@ -179,6 +202,7 @@ class RulesTestContextImpl implements RulesTestContext {
   }
   storage(bucketUrl?: string): firebase.storage.Storage {
```
sam-gc, line 203:
```suggestion
  storage(bucketUrl = `gs://${this.projectId}`): firebase.storage.Storage {
```

(I double checked this works)

sam-gc (approved):

packages/rules-unit-testing/src/impl/test_environment.ts
yuchenshi: Done

PR #5361, opened 2021-08-24
Add integration tests, fix reload and setCustomTokenProvider

lisajian:
Added integration tests for passthrough, namely,
- for the initial call to signInWithCustomToken
- for a force refresh with a customTokenProvider

Noticed a couple bugs in the process and addressed it here too:
- `user.uid` is not set until a call to `_reloadWithoutSaving()`. The previous logic in `_reloadWithoutSaving()` did not correctly set the `uid` and is now updated to parse the access token for the `sub` claim
- The REST API for [`signInWithCustomToken`](https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/signInWithCustomToken) does not include the `localId` in the response, so the assertion in `setCustomTokenProvider()` was also updated to parse the id token 

`xit` is used in place of `it` since passthrough emulator changes have not been merged into main, but will be updated to the latter once it's merged

### Testing

  * `yarn test` passes
  * With the passthrough emulator changes (i.e. checking out https://github.com/firebase/firebase-tools/commits/lj-passthrough and running `npm run build` in firebase-tools) and changing `xit`'s to `it`'s, `firebase emulators:exec --project p --only auth "yarn test:browser:integration:local"` passes

### API Changes

N/A

### DIFF
--- a/packages/auth/src/core/strategies/custom_token.test.ts
+++ b/packages/auth/src/core/strategies/custom_token.test.ts
@@ -35,6 +35,7 @@ import {
   signInWithCustomToken
 } from './custom_token';
 import { FirebaseError } from '@firebase/util';
+import { makeJWT } from '../../../test/helpers/jwt';
 
 use(sinonChai);
 use(chaiAsPromised);
@@ -52,7 +53,7 @@ describe('core/strategies/signInWithCustomToken', () => {
   };
 
   const idTokenResponse: IdTokenResponse = {
-    idToken: 'my-id-token',
+    idToken: makeJWT({ sub: 'local-id' }),
     refreshToken: 'my-refresh-token',
     expiresIn: '1234',
     localId: serverUser.localId!,
@@ -135,7 +136,7 @@ describe('core/strategies/signInWithCustomToken', () => {
         token: 'custom-token',
         returnSecureToken: true
       });
-      expect(token).to.eq('my-id-token');
+      expect(token).to.eq(idTokenResponse.idToken);
       expect(
         user.stsTokenManager.updateFromServerResponse
       ).to.have.been.calledWith(idTokenResponse);
--- a/packages/auth/src/core/strategies/custom_token.ts
+++ b/packages/auth/src/core/strategies/custom_token.ts
@@ -29,6 +29,7 @@ import { OperationType } from '../../model/enums';
 import { _assert } from '../util/assert';
 import { AuthErrorCode } from '../errors';
 import { UserInternal } from '../../model/user';
+import { _parseToken } from '../user/id_token_result';
 
 /**
  * Asynchronously signs in using a custom token.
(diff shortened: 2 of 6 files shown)


packages-exp/auth-exp/src/core/strategies/custom_token.test.ts
```diff
@@ -52,7 +53,7 @@ describe('core/strategies/signInWithCustomToken', () => {
   };
 
   const idTokenResponse: IdTokenResponse = {
-    idToken: 'my-id-token',
+    idToken: makeJWT({ sub: 'local-id' }),
     refreshToken: 'my-refresh-token',
     expiresIn: '1234',
     localId: serverUser.localId!,
```
lisajian, line 59: Noticed that `localId` is non-optional in `IdTokenResponse` (https://github.com/firebase/firebase-js-sdk/blob/78181760aea6c9b747807c4be07b65e801cd3a45/packages-exp/auth-exp/src/model/id_token.ts#L52) but it seems like not all responses include this field (e.g. [signInWithCustomToken](https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/signInWithCustomToken)) - wanted to get more context on why this field is required @avolkovi / @sam-gc ? Thanks!

sam-gc: Honestly it's probably an oversight

packages-exp/auth-exp/src/core/strategies/custom_token.ts
```diff
@@ -96,8 +97,9 @@ export function setCustomTokenProvider(
       token,
       returnSecureToken: true
     });
+    const responseToken = _parseToken(response.idToken!);
     _assert(
-      response.localId === authInternal.currentUser?.uid,
+      responseToken && responseToken.sub === authInternal.currentUser?.uid,
```
sam-gc, line 102: You can do `responseToken?.sub === authInternal.currentUser?.uid`

sam-gc (approved):
Looks great, thanks!

packages-exp/auth-exp/test/integration/flows/custom.local.test.ts
```diff
@@ -146,6 +150,62 @@ describe('Integration test: custom auth', () => {
...
+      expect(additionalUserInfo.isNewUser).to.be.false;
+    });
+
+    xit('token can be refreshed in passthrough mode', async () => {
+      setCustomTokenProvider(auth, {
+        async getCustomToken(): Promise<string> {
+          return JSON.stringify({
+            uid,
+            claims: {
+              customClaim: 'some-claim'
```
sam-gc, line 197: Could you change this claim here to some other string and then assert it gets updated after the refresh (to ensure this code gets called)

packages-exp/auth-exp/src/core/strategies/custom_token.ts
lisajian: Oof, good call, done!

packages-exp/auth-exp/test/integration/flows/custom.local.test.ts
lisajian: Done!

rosalyntan (approved):

PR #5396, opened 2021-08-27
Fail more gracefully when opening IndexedDb fails in Firefox private browsing

### LINKED ISSUE #5377: FIrefox Private Browsing: A mutation operation was attempted on a database that did not allow mutations
* Operating System version: _____macOS Big Sur 11.5.2
  * Browser version: _____ Firefox Developer Edition 92.0b6 (64bit)
  * Firebase SDK version: _____8.3.1
  * Firebase Product: _____ auth, firestore

#### Steps to reproduce:

I'm seeing the error "A mutation operation was attempted on a database that did not allow mutations" when starting my app in Firefox' private browsing mode. This was reported in #4069 and fixed in #4146 (8.1.2) so could this be a regression?

Exact error message:
`Uncaught (in promise) IndexedDbTransactionError: IndexedDB transaction 'getHighestListenSequenceNumber' failed: InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.`

dconeybe:
Improve the obscure failure that occurs when attempting to open an IndexedDb connection in a Firefox private browsing session (which is a known limitation of Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1639542).

Before this fix, the error that occurred in this scenario in v8.10.0 would be vague:

> Uncaught (in promise) IndexedDbTransactionError: IndexedDB transaction 'getHighestListenSequenceNumber' failed: InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.

In v9.0.0 it was even worse because it mislead readers to suggest that it was related to multi-tab:

> Uncaught (in promise) FirebaseError: Failed to obtain exclusive access to the persistence layer. To allow shared access, multi-tab synchronization has to be enabled in all tabs. If you are using `experimentalForceOwningTab:true`, make sure that only one tab has persistence enabled at any given time.

With this fix, the error message becomes more helpful by specifically calling out Firefox private browsing as a potential culprit:

> Uncaught (in promise) FirebaseError: Unable to open an IndexedDB connection. This could be due to running in a private browsing session on a browser whose private browsing sessions do not support IndexedDB: InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.

Both before and after this fix, Firestore would continue to be functional, just without persistence.

Fixes #5377


schmidt-sebastian (approved):
Thank you!  Can you remove the Firefox mention from the error message since we don't know where else this occurs? You can leave it in the changelog though as it addresses a specific issue.

dconeybe:
> Thank you! Can you remove the Firefox mention from the error message since we don't know where else this occurs? You can leave it in the changelog though as it addresses a specific issue.

The thing is, IndexedDb works in Chrome incognito mode. So removing Firefox from the message IMO incorrectly implies that it is broken on Chrome too. 

Also, the hyperlink the message is to a bug report against Firefox for IndexedDb support in private browsing mode. I'm not sure what the best wording is but I'm open to suggestions.

schmidt-sebastian:
I would try to be more vague and remove the link as well. You can also list it as an example of why it could be broken - but it is likely not the only reason. It would confuse users that see this error that are not on Firefox. Browsers that support private browsing with IndexedDB wouldn't show this error.

dconeybe:
> I would try to be more vague and remove the link as well. You can also list it as an example of why it could be broken - but it is likely not the only reason. It would confuse users that see this error that are not on Firefox. Browsers that support private browsing with IndexedDB wouldn't show this error.

Done. I've changed the error message to the following:

`Unable to open an IndexedDb connection. This could be due to running in a private browsing session on a browser whose private browsing sessions do not support IndexedDb. (InvalidStateError: A mutation operation was attempted on a database that did not allow mutations.)`

packages/firestore/src/local/simple_db.ts
```diff
@@ -326,10 +326,11 @@ export class SimpleDb {
             reject(
               new FirestoreError(
                 Code.FAILED_PRECONDITION,
-                'Unable to open an IndexedDb connection; this may be due to running in a ' +
-                  'Firefox private browsing session, in which IndexedDb is not supported ' +
-                  '(https://goo.gle/3ymSNyf): ' +
-                  error
+                'Unable to open an IndexedDb connection. This could be due to running in a ' +
```
schmidt-sebastian, line 329: The official spelling seems to be IndexedDB.

packages/firestore/src/local/simple_db.ts
```diff
@@ -322,6 +322,17 @@ export class SimpleDb {
                   'data for your app to start fresh.'
               )
             );
+          } else if (error.name === 'InvalidStateError') {
+            reject(
+              new FirestoreError(
+                Code.FAILED_PRECONDITION,
+                'Unable to open an IndexedDb connection. This could be due to running in a ' +
+                  'private browsing session on a browser whose private browsing sessions do not ' +
+                  'support IndexedDb. (' +
```
schmidt-sebastian, line 331: Nit: I think you should move the period after the parenthesis or use `:` without parenthesis.

schmidt-sebastian (approved):

packages/firestore/src/local/simple_db.ts
dconeybe: Done.

packages/firestore/src/local/simple_db.ts
dconeybe: Done (changed to ":").

Issue #5392, opened 2021-08-27
`OAuthProvider` is missing constructor


AgainPsychoX:
### Environment

  * Firebase SDK version: `9.0.0`
  * Firebase Product: `auth`

### Problem

Firebase Auth API v9 does not have `OAuthProvider` constructor, or at least in TypeScript typings.

API v9: https://firebase.google.com/docs/reference/js/auth.oauthprovider (no constructor or a way to get new instance `OAuthProvider`)

API v8: https://firebase.google.com/docs/reference/js/v8/firebase.auth.OAuthCredential (it's fine at v8)

Even comment docs suggest it should be there (:
![a](https://user-images.githubusercontent.com/23557839/131069746-ad3cd82c-4a5b-46e7-836d-f6744a2cc964.png)

I need it as I want to use OIDC (https://cloud.google.com/identity-platform/docs/web/oidc), but no constructor, no `addScope`.

cungminh2710:
Yup `new OAuthProvider('apple.com')` won't work. [These workarounds](https://github.com/firebase/firebase-js-sdk/issues/5392#issuecomment-906938939) can be used in the meantime

jpike88:
Just ran into this issue on Cordova:

```
const auth = getAuth(Environment.firebaseInstance);
			const provider = new GoogleAuthProvider();
			try {
				await signInWithRedirect(auth, provider, cordovaPopupRedirectResolver);
			} catch (error) {
				// swallow this, just return.
				return;
			}
			const result = await getRedirectResult(auth);
			const credential = GoogleAuthProvider.credentialFromResult(result);
			await this.handleSSOAuthResult(
				AuthProviders.Google,
				result.user.email,
				credential.idToken
			);
```

This should work but errors. Error is being caused here:
https://github.com/firebase/firebase-js-sdk/blame/509c18fbc1a0d4f85a0ca8c7cfef8e3b0f3183ec/packages/auth/src/platform_browser/strategies/redirect.ts#L95

jpike88:
Also none of the above workarounds work for me, that assert still trips even when the console seems to be indicating that the OAuthProvider is a subclass of FederatedProvider...

This is broken and a critical issue for us

<img width="1192" alt="Screenshot 2021-08-30 at 3 24 02 PM" src="https://user-images.githubusercontent.com/9585787/131309455-afbc6865-f401-4d56-a9c3-e50c95712028.png">

jpike88:
More investigating...

[...]

PR #5397, opened 2021-08-27
use raw api-report rollup d.ts

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5392

`repo-scripts/prune-dts/prune-dts.ts` removes inherited constructors, which makes the typings of classes incorrect.
This is a temporary fix, and will expose some internal types to developers. We should fix the `prune-dts.ts` to handle constructors correctly.

@schmidt-sebastian fyi

### DIFF
--- a/.changeset/dirty-balloons-unite.md
+++ b/.changeset/dirty-balloons-unite.md
@@ -1,0 +1,5 @@
+---
+"@firebase/auth": patch
+---
+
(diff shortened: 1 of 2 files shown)


jpike88:
@Feiyang1 are my comments on https://github.com/firebase/firebase-js-sdk/issues/5392 applicable to this PR?

PR #5400, opened 2021-08-28
Fix nodejs builds

Feiyang1:
Fixes https://github.com/angular/angularfire/issues/2903

- Fix cjs build of app-compat, so it works with mjs import in Nodejs/bundlers
- Expose firestore/lite in Nodejs by defining the exports field


Feiyang1:
@jamesdaniels fyi

hsubox76 (approved):

Issue #5386, opened 2021-08-26
[v9/auth] Blocking Function Errors in JS SDK do not provide full error message/info from blocking function


tannerlinsley:
### [REQUIRED] Describe your environment

  * Operating System version: any
  * Browser version: any
  * Firebase SDK version: 9.0.0-beta.8
  * Firebase Product: auth

### [REQUIRED] Describe the problem

When using v9 and when following the documentation for blocker functions [here](https://cloud.google.com/identity-platform/docs/blocking-functions#blocking_registration_or_sign-in) to send a custom error message to the Firebase Auth SDK, error messages are malformed and do not contain the data that the docs claim they should.

Here is the blocker function we are running:

```js
const gcipCloudFunctions = require('gcip-cloud-functions');
const authClient = new gcipCloudFunctions.Auth();

exports.beforeCreate = authClient.functions().beforeCreateHandler((user, context) => {
  throw new gcipCloudFunctions.https.HttpsError('invalid-argument', `this is a custom error message`);
})
```

This is the network error that is received in the browser after the blocker function errors:

```json
{
  "error": {
    "code": 400,
    "message": "BLOCKING_FUNCTION_ERROR_RESPONSE : HTTP Cloud Function returned an error: {\"error\":{\"code\":400,\"message\":\"this is a custom error message\",\"status\":\"INVALID_ARGUMENT\"}}",
    "errors": [
      {
        "message": "BLOCKING_FUNCTION_ERROR_RESPONSE : HTTP Cloud Function returned an error: {\"error\":{\"code\":400,\"message\":\"this is a custom error message\",\"status\":\"INVALID_ARGUMENT\"}}",
        "domain": "global",
        "reason": "invalid"
      }
    ]
  }
}

```

This is the error (serialized) that the client SDK receives after the failed attempt to create a user:

```js
react_devtools_backend.js:2850 {
  "code": "auth/blocking-function-error-response",
  "customData": {},
  "name": "FirebaseError"
}
```

As you can see, the SDK error does not contain the same information as is visible in the network tab.

cc: @derekperkins

xil222:
Hey @Feiyang1 , @sam-gc could anyone of you cc me as well once the internal bug is filed? I am also working on blocking function, and receiving a similar issue for blocking function in [link](https://github.com/firebase/firebaseui-web/issues/866).
I might be able to help when I have time, thanks!

PR #5423, opened 2021-08-31
[Auth] Pass through server error messages to the error object when available

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5386

In keeping with the new error format, the message is in `error.customData.message` instead of `error.message`.


lisajian (approved):
LGTM, do we need to update the example in https://cloud.google.com/identity-platform/docs/blocking-functions#blocking_registration_or_sign-in since the error is in `customData`? I think the only bit that needs to be updated is `error.message.indexOf('Cloud Function') !== -1` should be `error.customData.message.indexOf('Cloud Function') !== -1`

yuchenshi (approved):

sam-gc:
In this case we shouldn't update the Cloud Identity Platform docs since the v9 SDK remains in beta for the Cloud Identity Platform features. We will need to update the compatibility layer to put the fields on the error instead of error.customData, but that's a larger PR that I'll do a bit later

Issue #5420, opened 2021-08-31
9.0.0 Auth doesn't work: GoogleAuthProvider is failing FederatedAuthProvider and BaseOAuthProvider instanceof assertions


jpike88:
* Operating System version: macOS
  * Browser version: Chrome
  * Firebase SDK version: 9.0.0-beta.7
  * Firebase Product: auth
  * Environment: Cordova + iOS 15, same issue on Cordova + Android 11 Chromium

(I use fuse-box as a bundler, and my code is being output to ES2015 code, and I have countless other libraries in my codebase that run fine.)

Things were working with Firebase 8.x, but after upgrading to 9 and migrating to the required code:

```

import { GoogleAuthProvider, FacebookAuthProvider } from '@firebase/auth';
import {
	getRedirectResult,
	signInWithRedirect,
	getAuth,
} from '@firebase/auth/cordova';

const auth = getAuth(Environment.firebaseInstance);
			const provider = new GoogleAuthProvider();
			try {
				await signInWithRedirect(auth, provider);
			} catch (error) {
				// swallow this, just return.
				return;
			}
			const result = await getRedirectResult(auth);
			const credential = GoogleAuthProvider.credentialFromResult(result);
			await this.handleSSOAuthResult(
				AuthProviders.Google,
				result.user.email,
				credential.idToken
			);
```

This code doesn't work, an assertion fails here:

https://github.com/firebase/firebase-js-sdk/blame/509c18fbc1a0d4f85a0ca8c7cfef8e3b0f3183ec/packages/auth/src/platform_browser/strategies/redirect.ts#L95

See attached image for what inspecting breakpointed variables looks like:

<img width="1192" alt="131309455-afbc6865-f401-4d56-a9c3-e50c95712028" src="https://user-images.githubusercontent.com/9585787/131432183-ceebebce-fe41-4805-a25b-d94b28cf2722.png">

[see this comment for detailed further investigation](https://github.com/firebase/firebase-js-sdk/issues/5392#issuecomment-908168109)

Current workaround, on postinstall do a find+sed on the @firebase folder, and replace all instances of ```provider instanceof FederatedAuthProvider``` and ```provider instanceof BaseOAuthProvider``` with ```true``` in order to disable the asserts, and everything works properly.

jbalidiong:
Hi @jpike88, thanks for reporting this behavior. If I can replicate the issue, I can have a better look into it. Please share a minimal, but complete sample of a project that I can run locally.

[...]

PR #5427, opened 2021-09-01
[Auth] Add custom error factory for custom messages; add special handling for instanceof checks

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5420

We now provide better messaging in cases where auth is imported from multiple entry points, leading to mismatched types. Also small update for https://github.com/firebase/firebase-js-sdk/pull/5423; we now incorporate the message onto the error itself (as it should have been originally, thanks @lisajian for pointing that out)

### DIFF
--- a/packages/auth/src/core/errors.ts
+++ b/packages/auth/src/core/errors.ts
@@ -359,7 +359,7 @@ export interface ErrorMapRetriever extends AuthErrorMap {
   (): ErrorMap<AuthErrorCode>;
 }
 
-export function _prodErrorMap(): ErrorMap<AuthErrorCode> {
+function _prodErrorMap(): ErrorMap<AuthErrorCode> {
   // We will include this one message in the prod error map since by the very
   // nature of this error, developers will never be able to see the message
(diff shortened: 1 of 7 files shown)


packages/auth/src/core/errors.ts
```diff
@@ -359,7 +359,7 @@ export interface ErrorMapRetriever extends AuthErrorMap {
   (): ErrorMap<AuthErrorCode>;
 }
 
-function _prodErrorMap(): ErrorMap<AuthErrorCode> {
+export function _prodErrorMap(): ErrorMap<AuthErrorCode> {
```
lisajian, line 362: Nit: I think this gets exported as `prodErrorMap` in the same file?

lisajian (approved):

packages/auth/src/core/errors.ts
sam-gc: Good catch, thanks!

sam-gc:
@yuchenshi I still need your review since Lisa's not yet a CODEOWNER

yuchenshi (approved):

PR #5453, opened 2021-09-07
Add protocol as a separate property for FirebaseStorage service

hsubox76:
I tried to avoid adding another property to the service in the initial emulator PR by having some logic in `makeUrl` to deal with the fact that `host` could contain a protocol or not. Unfortunately `Location.refFromUrl` doesn't use `makeUrl` and this caused a bug where there were duplicate protocol strings in the regex. Trying to keep "host" in a consistent format now to avoid future mistakes, separating "protocol" into a separate field.

Fixes https://github.com/firebase/firebase-tools/issues/3715

### DIFF
--- a/packages/storage/src/api.ts
+++ b/packages/storage/src/api.ts
@@ -51,6 +51,7 @@ import {
 } from './reference';
 import { STORAGE_TYPE } from './constants';
 import { EmulatorMockTokenOptions, getModularInstance } from '@firebase/util';
+import { StringFormat } from './implementation/string';
 
 export { EmulatorMockTokenOptions } from '@firebase/util';
 
@@ -109,7 +110,7 @@ export function uploadBytes(
 export function uploadString(
   ref: StorageReference,
   value: string,
-  format?: string,
+  format?: StringFormat,
   metadata?: UploadMetadata
 ): Promise<UploadResult> {
   ref = getModularInstance(ref);
--- a/packages/storage/src/api.ts
+++ b/packages/storage/src/api.ts
@@ -79,7 +79,7 @@ export {
  * Uploads data to this object's location.
  * The upload is not resumable.
  * @public
- * @param ref - StorageReference where data should be uploaded.
+ * @param ref - {@link StorageReference} where data should be uploaded.
  * @param data - The data to upload.
  * @param metadata - Metadata for the data to upload.
  * @returns A Promise containing an UploadResult
@@ -101,7 +101,7 @@ export function uploadBytes(
  * Uploads a string to this object's location.
  * The upload is not resumable.
  * @public
- * @param ref - StorageReference where string should be uploaded.
+ * @param ref - {@link StorageReference} where string should be uploaded.
  * @param value - The string to upload.
  * @param format - The format of the string to upload.
  * @param metadata - Metadata for the string to upload.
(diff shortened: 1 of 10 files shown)


packages/storage/test/unit/location.test.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2017 Google LLC
```
schmidt-sebastian, line 3: 2021

packages/storage/src/api.ts
```diff
@@ -109,7 +110,7 @@ export function uploadBytes(
 export function uploadString(
   ref: StorageReference,
   value: string,
-  format?: string,
+  format?: StringFormat,
```
schmidt-sebastian, line 113: Thanks for fixing this. I ran into this in my getBytes() PR as well.

packages/storage/test/unit/location.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import { DEFAULT_HOST } from '../../src/implementation/constants';
+import { Location } from '../../src/implementation/location';
+
+describe('Firebase Storage > Location', () => {
+  it('makeFromUrl handles an emulator url correctly', () => {
+    const loc = Location.makeFromUrl(
+      'http://localhost:3001/v0/b/abcdefg.appspot.com/o/abcde.txt',
+      'localhost:3001'
+    );
+    expect(loc.bucket).to.equal('abcdefg.appspot.com');
```
schmidt-sebastian, line 27: Should we test the protocol here as well?

packages/storage/src/service.ts
```diff
@@ -158,9 +159,9 @@ export class FirebaseStorageImpl implements FirebaseStorage {
    * This string can be in the formats:
    * - host
    * - host:port
-   * - protocol://host:port
    */
   private _host: string = DEFAULT_HOST;
+  protocol: string = 'https';
```
schmidt-sebastian, line 164: Does this need to be exposed via the API?

hsubox76: This whole class is internal so it shouldn't be exposed publicly. But it can't be private since `connectStorageEmulator` needs to be able to set it. I can add an underscore so it's clear that it's not meant to be public?

packages/storage/test/unit/location.test.ts
hsubox76: Fixed.

packages/storage/test/unit/location.test.ts
hsubox76: Location doesn't seem to store the host or the protocol, makeFromUrl just uses them to do regex matches and then throws them away. It only seems to have bucket, and path (just the part after the bucket).

packages/storage/src/service.ts
schmidt-sebastian: The rest of the class does use underscores. I merely asked because of the change in the API report as it is not obvious that this is not exposed.

packages/storage/test/unit/location.test.ts
schmidt-sebastian: Okay, got it.

schmidt-sebastian (approved):

packages/storage/src/service.ts
hsubox76: Added an underscore.

Feiyang1 (approved):

Feiyang1:
please add changeset

Issue #5451, opened 2021-09-07
Bug still exists: Stop using const enums in the public API #5012


Obiwarn:
I installed firebase 9.0.1.

I use Vite.

On build I still get these errors in node_modules/@firebase/auth/dist/auth-public.d.ts:704:6 which should have been fixed by #5012 

```
node_modules/@firebase/auth/dist/auth-public.d.ts:710:6 - error TS2748: Cannot access ambient const enums when the '--isol
atedModules' flag is provided.
```

Am I missing something?

PR #5460, opened 2021-09-08
Remove Auth internal types from the public typings

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5451

Mark some internal types as `@internal` to prevent `const enum`s from accidentally showing up in the public typing file.


Feiyang1:
@sam-gc FYI

sam-gc (approved):
Thanks for getting to the bottom of this Fei

hsubox76 (approved):

Issue #5384, opened 2021-08-26
error in prod build for 9.0.0 firestore


mikeover:
### Describe your environment

  * Operating System version: MacOS
  * Browser version: Chrome 91
  * Firebase SDK version: 9.0.0
  * Firebase Product: firestore

### Describe the problem

Works fine in dev but when building for production, I get a blank screen and this error in the console:

<img width="607" alt="Screen Shot 2021-08-26 at 2 40 19 PM" src="https://user-images.githubusercontent.com/5045642/131018137-7b8295bc-0d62-41c4-a02e-17aade41a36a.png">

#### Steps to reproduce:

This runs fine with `yarn start` but fails after building for production with `yarn start:dist` (see package.json scripts below).

#### Relevant Code:

index.jsx:
```javascript
import React from 'react'
import ReactDOM from 'react-dom'
import config from 'config'
import { version } from '../package.json'
import App from './App'
import { init as initErrorHandling } from './utils/errorHandler'
import { FirebaseAppProvider, SuspenseWithPerf } from 'reactfire'
import LoadingSpinner from 'components/LoadingSpinner'
import './index.css'

window.version = version
window.config = config

initErrorHandling()

ReactDOM.render(
  <SuspenseWithPerf traceId={'firebase-user-wait'} fallback={<LoadingSpinner />}>
    <FirebaseAppProvider firebaseConfig={config.firebase} suspense={true}>
      <App />
    </FirebaseAppProvider>
  </SuspenseWithPerf>,
  document.getElementById('root')
)
```

App.js:
```javascript
import React from 'react'
import { FirestoreProvider, useInitFirestore } from 'reactfire'
import { initializeFirestore } from 'firebase/firestore'

function App() {
  function FirestoreWrapper({ children }) {
    const { data: firestoreInstance } = useInitFirestore(async (firebaseApp) => {
      const db = initializeFirestore(firebaseApp, {})
      return db
    })
    return <FirestoreProvider sdk={firestoreInstance}>{children}</FirestoreProvider>
  }
  return (
    <FirestoreWrapper>
      <div>my app</div>
    </FirestoreWrapper>
  )
}

export default App
```

package.json scripts:
```
  "scripts": {
    "build": "craco build",
    "eject": "craco eject",
    "setup": "echo \"Generating client config...\" && env-cmd bin/generate-firebase-sdk-config.js --file ./config/default.json",
[...]

PR #5461, opened 2021-09-08
Temporary hack to fix terser output

hsubox76:
Mangled terser code is causing some problems seen in Create React App apps. This is a temporary fix that will avoid mangling the specific variable in question. Long term fix is to see if terser will provide an option to mangle properties without mangling anything else (currently not available), and/or find out what downstream build tool is causing this bug and get them to fix it.

Details:
The `initializeFirestore()` function, when mangled by terser, has two variables with the same name in different scopes (one is an import, one is a `const` in an `if` block). This seems to confuse one of the tools in the Create React App production default build pipeline (Babel?). The code is converted to ES5, which makes the `const` a `var`, which puts both variables sharing a name into the same scope. The build tool gets confused and does not convert the import to the correct symbol, thinking it is the same as the declared `var`. We should try to pinpoint what step of the build this happens in and report a bug to the appropriate library.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5384

### DIFF
--- a/packages/firestore/rollup.shared.js
+++ b/packages/firestore/rollup.shared.js
@@ -145,6 +145,12 @@ const manglePrivatePropertiesOptions = {
     beautify: true
   },
   mangle: {
+    // Temporary hack fix for an issue where mangled code causes some downstream
+    // bundlers (Babel?) to confuse the same variable name in different scopes.
+    // This can be removed if the problem in the downstream library is fixed
+    // or if terser's mangler provides an option to avoid mangling everything
+    // that isn't a property.
+    reserved: ['_getProvider'],
     properties: {
       regex: /^__PRIVATE_/,
       // All JS Keywords are reserved. Although this should be taken cared of by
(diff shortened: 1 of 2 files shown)


packages/firestore/rollup.shared.js
```diff
@@ -145,6 +145,12 @@ const manglePrivatePropertiesOptions = {
     beautify: true
   },
   mangle: {
+    // Temporary hack fix for an issue where mangled code causes some downstream
+    // bundlers (Babel?) to confuse the same variable name in different scopes.
+    // This can be removed if the problem in the downstream library is fixed
+    // or if terser's mangler provides an option to avoid mangling everything
+    // that isn't a property.
```
Feiyang1, line 152: probably worth adding the issue link for future reference.

Feiyang1 (approved):

packages/firestore/rollup.shared.js
hsubox76: Added.

schmidt-sebastian: We have two lists of reserved properties now (see right below - line 158). Do we know that this will work?

PR #5469, opened 2021-09-09
Remove ConnectionPool

schmidt-sebastian:
This PR is meant to reduce the diff for https://github.com/firebase/firebase-js-sdk/pull/4973

The PR removes the old ConnectionPool, which was never used to share connections but rather just allowed tests to inject a different connection type. Since #4973 introduces multiple connection types (string, blob, stream) and since we only want to include the connection types that are actually used, this PR now directly specifies the specific connection factory at the callsite (which for now is just a text-based connection).

### DIFF
--- a/.changeset/dry-toes-hammer.md
+++ b/.changeset/dry-toes-hammer.md
@@ -1,6 +1,6 @@
 ---
 "@firebase/auth-compat": minor
-"@firebase/auth": minor,
+"@firebase/auth": minor
 "firebase": minor
 ---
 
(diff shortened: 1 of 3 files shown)


packages/storage/src/platform/browser/connection.ts
```diff
@@ -22,6 +22,9 @@ import {
 } from '../../implementation/connection';
 import { internalError } from '../../implementation/error';
 
+/** An override for the text-based Connection. Used in tests. */
```
egilmorez, line 25: Should Connection be backticked as a literal?  Looks like there are other instances of the same in other files in this PR.

egilmorez (approved):
LG, one nit.

PR #5486, opened 2021-09-13
Fix workflow that automatically publishes @firebase/app

hsubox76:
`@firebase/app` should be published on every release. The script to automatically do this was failing because `yarn changesets status` couldn't find the `master` branch locally, because actions/checkout doesn't check it out locally. After reproducing locally, it does seem to be able to access `origin/master`, so changed it to that.

### DIFF
--- a/.changeset/bump-sdk-version.md
+++ b/.changeset/bump-sdk-version.md
@@ -1,6 +1,0 @@
-
----
-'@firebase/app': patch
----
-
-Update SDK_VERSION.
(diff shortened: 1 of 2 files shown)


Feiyang1:
LG. Thanks. Can we run the workflow in Github to make sure it actually works?

hsubox76:
I don't think we can test the workflow without merging it because reproducing the exact git state means it has to be on a push to the release branch, and that won't trigger if it's not merged to master. I reproduced it as much as possible locally by manually doing the same weird git checkout steps that actions/checkout does.

Feiyang1 (approved):

PR #5488, opened 2021-09-13
Log release PR number

hsubox76:
Triggers on PR to `release` branch.

- Gets PR payload data through Github provided env variable.
- Check if title is 'Version Packages'.
- If so, extract release version from PR description.
- Call release tracker functions endpoint with release version and PR number.

### DIFF
--- a/.github/workflows/release-log.yml
+++ b/.github/workflows/release-log.yml
@@ -19,4 +19,4 @@ jobs:
           node-version: 14.x
 
       - name: Get PR number and send to tracker.
-        run: npx ts-node-script scripts/ci/log-changesets.ts
+        run: npx ts-node scripts/ci/log-changesets.ts
--- a/.github/workflows/release-log.yml
+++ b/.github/workflows/release-log.yml
@@ -18,8 +18,5 @@ jobs:
         with:
           node-version: 14.x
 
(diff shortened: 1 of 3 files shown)


.github/workflows/release-log.yml
```diff
@@ -0,0 +1,22 @@
+name: Log Release PR
+
+on:
+  pull_request:
+    branches:
+      - release
```
Feiyang1, line 6: Should we add the v8 release branch to capture v8 releases?

scripts/ci/log-changesets.js
```diff
@@ -0,0 +1,77 @@
...
+
+const https = require('https');
+
+async function logChangesets() {
+  if (!process.env.GITHUB_EVENT_PATH) {
+    console.log(`Couldn't find PR event payload.`);
+    return;
+  }
+
+  const prPayload = require(process.env.GITHUB_EVENT_PATH);
```
Feiyang1, line 26: wow, didn't know you can do this.

.github/workflows/release-log.yml
hsubox76: Added.

Feiyang1 (approved):

PR #5489, opened 2021-09-13
Move MemoryPersistenceProvider

schmidt-sebastian:
The MemoryPersistenceProvider is only used in the Console build, as firestore-compat uses the IndexedDB build by default.


thebrianchen (approved):
lgtm

PR #5490, opened 2021-09-14
do not run test:minified for @firebase/firestore in CI

Feiyang1:
test:minified runs in a separate workflow


schmidt-sebastian (commented):
Does this still fail PR submissions for Firestore if these tests fail? These tests are very important, as they are the only tests that can discover problems with minification. We never run them locally...

Feiyang1:
@schmidt-sebastian yup, `firestore-integration` is a required workflow. https://github.com/firebase/firebase-js-sdk/blob/master/.github/workflows/test-changed-firestore-integration.yml

schmidt-sebastian (approved):

PR #5505, opened 2021-09-16
Clean up compat docgen

hsubox76:
Small path fixes and add 2 pages to the TOC that were missing. These additions were before the v9 release but we had not updated the TOC. Will also update and re-run the g3 script after this is checked in.


egilmorez (approved):
Thanks!

Feiyang1 (approved):

PR #5506, opened 2021-09-16
isSupported fixes

jamesdaniels:
`isSupported()` had bugs in messaging, performance, and analytics.

* in messaging, isSupported was not checking for the existence of window or indexedDB before validating that indexedDB is openable
* in performance and analytics, navigator was not being checked for correctly `!navigator` should be `typeof navigator !== 'undefined` since it's a global. This led to runtime failures.
* further I dry'd up performance and messaging to use the same `areCookiesEnabled` test

Further, the indexedDb check was not working correctly in AppCheck due to checking for it on `self` (which isn't available in all environments). I addressed this and dry'd up our `typeof indexedDb` checks to all use `isIndexedDBAvailable()`.

### DIFF
--- a/.changeset/angry-moons-change.md
+++ b/.changeset/angry-moons-change.md
@@ -1,0 +1,6 @@
+---
+'@firebase/app-check': patch
+'@firebase/app-check-compat': patch
+---
+
+AppCheck could encounter runtime errors when initialized in Node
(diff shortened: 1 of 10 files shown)


jposuna (approved):
LGTM for performance. Thanks for making this change!

hsubox76 (approved):
Looks good, thanks for being so thorough!

egilmorez (approved):
Looks like no new doc strings, lmk if I'm wrong about that.  Thanks!

Feiyang1 (approved):

PR #5512, opened 2021-09-17
Initialize App Check debug mode in initializeAppCheck

hsubox76:
Initialize debug mode for App Check (i.e., reading the global var) in the first call to `initializeAppCheck()` instead of when the app-check component is registered. This makes it easier for developers to set the variable in code.

Additionally, it logs a `console.warn` on subsequent calls to initializeAppCheck notifying the user they are still in debug mode, and logging the token (if they don't reload/quit before that token promise resolves). This may help HMR users who lose the initial token message.

### DIFF
--- a/packages/app-check/src/api.test.ts
+++ b/packages/app-check/src/api.test.ts
@@ -121,24 +121,29 @@ describe('api', () => {
       ).to.equal(appCheckInstance);
     });
     it('starts debug mode on first call', async () => {
-      const fakeWrite = (): Promise<void> => Promise.resolve();
-      const writeToIndexedDBStub = stub(
+      let token: string = '';
+      const fakeWrite = (tokenToWrite: string): Promise<void> => {
+        token = tokenToWrite;
+        return Promise.resolve();
+      };
+      stub(
         indexeddb,
         'writeDebugTokenToIndexedDB'
       ).callsFake(fakeWrite);
-      stub(indexeddb, 'readDebugTokenFromIndexedDB').callsFake(() =>
-        Promise.resolve(undefined)
-      );
+      stub(indexeddb, 'readDebugTokenFromIndexedDB').resolves(token);
       const logStub = stub(logger.logger, 'warn');
       const consoleStub = stub(console, 'log');
       self.FIREBASE_APPCHECK_DEBUG_TOKEN = true;
       initializeAppCheck(app, {
         provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
       });
-      await fakeWrite();
-      const token = writeToIndexedDBStub.args[0];
-      expect(logStub).to.not.be.called;
+      // Ensure getDebugToken() call inside `initializeAppCheck()`
+      // has time to resolve, and double check its value matches that
+      // written to indexedDB.
+      expect(await getDebugToken()).to.equal(token);
       expect(consoleStub.args[0][0]).to.include(token);
+      expect(logStub).to.be.calledWith(match('is in debug mode'));
+      expect(logStub).to.be.calledWith(match(token));
       self.FIREBASE_APPCHECK_DEBUG_TOKEN = undefined;
     });
     it('warns about debug mode on second call', async () => {
--- a/packages/app-check/src/api.test.ts
+++ b/packages/app-check/src/api.test.ts
@@ -126,10 +126,7 @@ describe('api', () => {
         token = tokenToWrite;
         return Promise.resolve();
       };
-      stub(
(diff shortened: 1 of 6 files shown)


packages/app-check/src/api.ts
```diff
@@ -73,6 +90,9 @@ export function initializeAppCheck(
     }
   }
 
+  // Only read global variable on first call to `initializeAppCheck()`.
+  initializeDebugMode();
```
Feiyang1, line 94: It should be called only once for all app instances, not per app instance.

packages/app-check/src/api.ts
```diff
@@ -65,6 +67,21 @@ export function initializeAppCheck(
...
+        logger.warn(
+          `App Check is in debug mode. To turn off debug mode, unset ` +
+            `the global variable FIREBASE_APPCHECK_DEBUG_TOKEN and ` +
+            `restart the app.`
+        );
+        // Make this a separate console statement so user will at least have the
+        // first message if the token promise doesn't resolve in time.
+        void getDebugToken().then(token =>
+          logger.warn(`Debug token is ${token}.`)
+        );
```
Feiyang1, line 83: We should print the debug token every time when `initializeAppCheck` is called if it's in the debug mode, not just the second time. For example, print the token after the page is refreshed.

Feiyang1 (changes requested):

packages/app-check/src/api.ts
hsubox76: Ok, can do, just FYI, if it's the first call to initializeApp(), the user will see
`App Check debug token: ${newToken}. You will need to add it to your app's App Check settings in the Firebase console for it to work`
And then they will also see these 2 messages right after that. Is that a bit clunky?

I guess the exception is if they have 2 FirebaseApps, then they'll only see all 3 messages when calling initializeApp for the first FirebaseApp.

Feiyang1: `logger.warn(`Debug token is ${token}.`)` is not necessary, if we already print `App Check debug token: ${newToken}. You will need to add it to your app's App Check settings in the Firebase console for it to work`.

Feiyang1: so no matter whether it's the first or nth time, always show the same message.

rickstaa:
@hsubox76 Thanks a lot for your work on this feature. Just a small heads-up. I noticed that in the old version, the Debug token is not printed to the console in Firefox 92.0 (64-bit) on Linux. It is easily found in the `firebase-app-check-db` though. On chrome, everything works fine.

rickstaa:
@hsubox76 I removed my previous comment since it was caused by some settings in Firefox.

packages/app-check/src/api.ts
hsubox76: Ok, done, see if that looks acceptable now.

packages/app-check/src/api.ts
hsubox76: Changed.

packages/app-check/src/api.ts
```diff
@@ -67,14 +66,13 @@ export function initializeAppCheck(
   // Log a warning when `initializeAppCheck()` is called in debug mode,
```
Feiyang1, line 66: Comment needs update.

packages/app-check/src/api.test.ts
```diff
@@ -118,6 +120,40 @@ describe('api', () => {
...
+        provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+      });
+      // Ensure getDebugToken() call inside `initializeAppCheck()`
+      // has time to resolve, and double check its value matches that
+      // written to indexedDB.
+      expect(await getDebugToken()).to.equal(token);
+      expect(consoleStub.args[0][0]).to.include(token);
+      self.FIREBASE_APPCHECK_DEBUG_TOKEN = undefined;
+    });
+    it('warns about debug mode on second call', async () => {
```
Feiyang1, line 143: What does it test? 
I think we need a test to verify `initializeDebugMode` is called only once.

packages/app-check/src/api.ts
hsubox76: Updated.

packages/app-check/src/api.test.ts
hsubox76: Oh wow, the test caught the fact that I was not setting the `initialize` property... thanks.

Feiyang1 (approved):

Feiyang1:
Please add a changeset. @hsubox76

Issue #5475, opened 2021-09-10
[v9.0.2 auth] INTERNAL ASSERTION FAILED: Expected a class definition


bvanneerven:
In an effort to reduce the dependencies of the authentication code I use, I tried the following:

```ts
// Works
import { getAuth } from 'firebase/auth';
const auth = getAuth(app);

// ERROR!
import { browserLocalPersistence, initializeAuth } from 'firebase/auth';
const auth = initializeAuth(app, { persistence: browserLocalPersistence });
```

<details>
  <summary>Error output</summary>

```
Error: INTERNAL ASSERTION FAILED: Expected a class definition
    at debugFail ({***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:535:11)
    at debugAssert ({***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:546:9)
    at _getInstance ({***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:568:5)
    at Array.map (<anonymous>)
    at _initializeAuthInstance ({***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:637:80)
    at {***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:6979:13
    at Component.instanceFactory ({***path***}/node_modules/.pnpm/@firebase+auth@0.17.2_@firebase+app@0.7.0/node_modules/@firebase/auth/dist/node/index-012195c6.js:6981:11)
    at Provider.getOrInitializeService ({***path***}/node_modules/.pnpm/@firebase+component@0.5.6/node_modules/@firebase/component/dist/index.cjs.js:342:39)
    at Provider.initialize ({***path***}/node_modules/.pnpm/@firebase+component@0.5.6/node_modules/@firebase/component/dist/index.cjs.js:264:29)
[...]

PR #5527, opened 2021-09-21
[Auth] Make the persistences default to inMemory for node

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5475

In node, all persistence types now map to inMemory instead of the not-available error

### DIFF
--- a/.changeset/swift-shirts-appear.md
+++ b/.changeset/swift-shirts-appear.md
@@ -1,5 +1,5 @@
 ---
-"@firebase/auth": patch
+"@firebase/auth": minor
 ---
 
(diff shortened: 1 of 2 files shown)


yuchenshi (approved):

lisajian (approved):

.changeset/swift-shirts-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/auth": patch
```
yuchenshi, line 2: nit: Is this a patch or a minor?

sam-gc: Good call, it should probably be minor

Feiyang1 (approved):

PR #5531, opened 2021-09-22
export types from @firebase/database-compat for admin SDK

Feiyang1:


hiranya911 (approved):
LGTM. We only need the `Database` type from the v8 `@firebase/database` package. But exporting everything from here seems consistent and more future-proof.

Feiyang1:
I see `ServerValue` being used as a value as well.

hiranya911:
> I see ServerValue being used as a value as well.

Yes! Our `database/index` file does so. You're right.

Issue #5499, opened 2021-09-15
Firebase v9.0.1 with Sveltekit node adapter – build failure


kenkunz:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 11.5.2
  * Browser version: n/a
  * Firebase SDK version: 9.0.1 (and newer)
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

I am using Firebase (`auth` and `firestore`) in a Sveltekit project. It has been building and running successfully until I recently tried to upgrade. I was able to isolate the error to `9.0.1` (more specifically: `firestore 3.0.1`).

I created a minimal app to identify what dependencies/versions are required in order to reproduce the issue. This occurs when building any recent version of Sveltekit with the `node` adapter. Using an older version of Sveltekit or the `node` adapter does not resolve the problem. Only by reverting to Firebase `9.0.0` (`firestore 3.0.0`) was it resolved, which is why I believe the issue is with Firebase / Firestore.

#### Steps to reproduce:

1. Create a skeleton [Sveltekit](https://kit.svelte.dev) app:
    ```
    $ npm init svelte@next sveltekit-firebase
    > Skeleton project
    > No Typescript
    > No ESLint
    > No Prettier
    ``` 
2. Install additional dependencies (`firebase` and `@sveltejs/adapter-node`)
    ```
    $ cd sveltekit-firebase
    $ npm i -P firebase@9.0.1
    $ npm i -D @sveltejs/adapter-node@next
    ```
3. Edit `svelte.config.js` and `src/routes/index.svelte` to import/use the two dependencies (see below)
4. Build the project
    ```
    $ npm run build
    ```
5. Note the build errors (see example error output below)
6. Revert `firebase` to `9.0.0` and build again
    ```
    $ npm i -P firebase@9.0.0
    $ npm run build
    ```
7. Note the build succeeds

##### `./svelte.config.js`
```javascript
/** @type {import('@sveltejs/kit').Config} */
import nodeAdapter from "@sveltejs/adapter-node";

const config = {
  kit: {
    adapter: nodeAdapter(),
    // hydrate the <div id="svelte"> element in src/app.html
    target: "#svelte",
  },
};

export default config;
```

##### `./src/routes/index.svelte`
```html
<script>
  import { getFirestore } from "firebase/firestore";
</script>

<h1>Welcome to SvelteKit</h1>
[...]

PR #5532, opened 2021-09-22
Fix Firestore exports for Node

hsubox76:
Create `require` and `import` subfields for `node` exports where the latter points to `mjs` ESM builds for Node. Rename `main-esm` fields so rollup builds ESM node bundles with `mjs` extension.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5499

### DIFF
--- a/packages/firestore/package.json
+++ b/packages/firestore/package.json
@@ -49,16 +49,22 @@
   },
   "exports": {
     ".": {
-      "node": "./dist/index.node.cjs.js",
+      "node": {
+        "require": "./dist/index.node.cjs.js",
+        "import": "./dist/index.node.mjs"
+      },
       "default": "./dist/index.esm2017.js"
     },
     "./lite": {
-      "node": "./dist/lite/index.node.cjs.js",
+      "node": {
+        "require": "./dist/lite/index.node.cjs.js",
+        "import": "./dist/lite/index.node.esm2017.js"
+      },
       "default": "./dist/lite/index.browser.esm2017.js"
     }
   },
   "main": "dist/index.node.cjs.js",
-  "main-esm": "dist/index.node.cjs.esm2017.js",
+  "main-esm": "dist/index.node.mjs",
   "react-native": "dist/index.rn.js",
   "browser": "dist/index.esm2017.js",
   "module": "dist/index.esm2017.js",
--- a/packages/firestore/lite/package.json
+++ b/packages/firestore/lite/package.json
@@ -2,7 +2,7 @@
   "name": "@firebase/firestore-lite",
   "description": "A lite version of the Firestore SDK",
   "main": "../dist/lite/index.node.cjs.js",
-  "main-esm": "../dist/lite/index.node.esm2017.js",
+  "main-esm": "../dist/lite/index.node.mjs",
   "module": "../dist/lite/index.browser.esm2017.js",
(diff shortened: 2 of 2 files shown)


packages/firestore/package.json
```diff
@@ -49,16 +49,22 @@
...
+        "require": "./dist/index.node.cjs.js",
+        "import": "./dist/index.node.mjs"
+      },
       "default": "./dist/index.esm2017.js"
     },
     "./lite": {
-      "node": "./dist/lite/index.node.cjs.js",
+      "node": {
+        "require": "./dist/lite/index.node.cjs.js",
+        "import": "./dist/lite/index.node.esm2017.js"
```
Feiyang1, line 61: import needs to point to a file with `mjs` extension.

packages/firestore/package.json
```diff
@@ -49,16 +49,22 @@
   },
   "exports": {
     ".": {
-      "node": "./dist/index.node.cjs.js",
+      "node": {
+        "require": "./dist/index.node.cjs.js",
+        "import": "./dist/index.node.mjs"
```
Feiyang1, line 54: Where is the build rule for creating this file?

Feiyang1 (changes requested):

packages/firestore/package.json
hsubox76: It gets the name from whatever is in the `main-esm` field in package.json

hsubox76: So here is the build config https://github.com/firebase/firebase-js-sdk/blob/2ad6ce8137c44bfc42cd4ea307e7c26c12fc065b/packages/firestore/rollup.config.js#L84

packages/firestore/package.json
hsubox76: Oops, fixed.

packages/firestore/package.json
Feiyang1: I see. Did you test it in Nodejs - running `import {} from @firebase/firestore` ?

hsubox76: I've added the workaround in Option 2 here: https://www.stefanjudis.com/snippets/how-to-import-json-files-in-es-modules-node-js/#option-2%3A-leverage-the-commonjs-%60require%60-function-to-load-json-files

It required some changes to tsconfig.json to avoid some TS compiler complaints but it compiles correctly and works in both a Node and a Sveltekit test project. It also works in both projects if I don't change tsconfig and I just `tsignore` the offending lines (`import module from 'module'` and `const require = module.createRequire(import.meta.url);`) so any thoughts on which way is better?

Feiyang1: I'd use `ts-ignore` with comments here because we don't want these changes to affect browser code which shares the same tsconfig with Node code.

Feiyang1:
LGTM. Does cjs build work correctly ?

hsubox76:
@schmidt-sebastian This is the PR that would require dropping Node 8 support (so that we can use `import.meta` for the `require` workaround - see earlier comments). I believe you suggested that would be ok with a minor bump.

schmidt-sebastian:
We do have precedent for that. I believe we ran some numbers last time that basically showed that there were no users on Node 6 that used an updated SDK version. I will check tomorrow if I can find some details.

Feiyang1:
@benmccann Yes, we are planning to do that.

PR #5539, opened 2021-09-23
use esm2017 builds by default

Feiyang1:
Esm2017 is the default build for v9 libraries. Changing util libraries to use esm2017 by default.

### DIFF
--- a/packages/firestore/src/util/input_validation.ts
+++ b/packages/firestore/src/util/input_validation.ts
@@ -132,14 +132,10 @@ export function valueDescription(input: unknown): string {
   }
 }
 
-/** Hacky method to try to get the constructor name for an object. */
+/** try to get the constructor name for an object. */
 export function tryGetCustomObjectType(input: object): string | null {
   if (input.constructor) {
-    const funcNameRegex = /function\s+([^\s(]+)\s*\(/;
-    const results = funcNameRegex.exec(input.constructor.toString());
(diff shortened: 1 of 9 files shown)


hsubox76 (approved):

packages/firestore/src/util/input_validation.ts
```diff
@@ -132,14 +132,10 @@ export function valueDescription(input: unknown): string {
...
-/** Hacky method to try to get the constructor name for an object. */
+/** try to get the constructor name for an object. */
 export function tryGetCustomObjectType(input: object): string | null {
   if (input.constructor) {
-    const funcNameRegex = /function\s+([^\s(]+)\s*\(/;
-    const results = funcNameRegex.exec(input.constructor.toString());
-    if (results && results.length > 1) {
-      return results[1];
-    }
+    return input.constructor.name;
```
Feiyang1, line 138: @schmidt-sebastian  I changed the compilation target to es2017 for testing, which preserves class definitions and revealed that this function doesn't work with classes. `constructor.name` is available everywhere except IE, which we don't need to worry about anymore. - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#browser_compatibility

schmidt-sebastian:
Can we get rid of "es2017ToEs5Plugins" in rollup.shared.js and remove the remaining callsites?

Issue #5541, opened 2021-09-24
sendSignInLinkToEmail sends incorrect link on iOS after updating to Firebase 9


tfang17:
### [REQUIRED] Describe your environment

  * Operating System version: N/A
  * Browser version: N/A
  * Firebase SDK version: 9.0.2
  * Firebase Product: auth

### [REQUIRED] Describe the problem

After updating from Firebase 8 to Firebase 9, dynamic link on iOS is broken:

Previous link sent in email: https://eureka.page.link/?link=https://eurekasurveys.com/email...
New link sent in email: https://eurekasurveys.com/email...

Missing the eurkea.page.link prefix being sent by sendSignInLinkToEmail.

#### Steps to reproduce:

Send Signin Link to iOS device on Firebase 9.

#### Relevant Code:

```javascript
    await firebase.auth().sendSignInLinkToEmail(email, {
        handleCodeInApp: true,
        iOS: {
            bundleId: '[REDACTED]',
        },
        url: 'https://www.eurekasurveys.com/surveys/'
    })
```

sam-gc:
Thanks for filing this issue! This is indeed a bug, I've been able to reproduce it. I'm working on a fix, stay tuned!

PR #5542, opened 2021-09-24
[Auth] Fix incorrectly-cased parameter in OOB request

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5541

We were sending `iosBundleId` instead of `iOSBundleId`, which was causing incorrect behavior.


lisajian (approved):

Feiyang1:
Green CI looks so refreshing

yuchenshi (approved):

PR #5557, opened 2021-09-28
WIP: Notify JSCore chat when smoke tests complete

hsubox76:


.github/workflows/e2e-test.yml
```diff
@@ -51,3 +51,15 @@ jobs:
       env:
         APP_CHECK_DEBUG_TOKEN: ${{ secrets.APP_CHECK_DEBUG_TOKEN }}
       run: xvfb-run yarn test:compat
+    - name: Tests succeeded
+      if: success()
+      run: |
+        curl -d "{\"text\": \"E2E Tests Succeded - https://github.com/${{github.repository}}/actions/runs/${{github.run_id}} \"}" -H "Content-Type: application/json" $WEBHOOK_URL
```
Feiyang1, line 57: Can we include the commit info and a link to the corresponding PR and release info if available directly in the message?

.github/workflows/e2e-test.yml
```diff
@@ -1,6 +1,6 @@
 name: E2E Smoke Tests
 
-on: workflow_dispatch
+on: [push, workflow_dispatch]
```
Feiyang1, line 3: Can we trigger it on every canary and staging release?

.github/workflows/e2e-test.yml
hsubox76: Need to separate this out into a node script to parse more data, testing now.

hsubox76: Ok, it's in a Node script now which allows more future flexibility I think.

.github/workflows/e2e-test.yml
```diff
@@ -32,7 +39,7 @@ jobs:
         echo "export const config = $PROJECT_CONFIG; export const testAccount = $TEST_ACCOUNT" > firebase-config.js
     - name: Yarn install
       run: |
-        yarn add firebase
+        yarn add firebase@next
```
Feiyang1, line 42: Instead of hardcoding it to the `next` tag, can the version be passed in from the caller?

hsubox76: Ok, this now takes the tag from the input (defaults to 'next'). I have a parallel change in release-cli ready to send the version and tag inputs, which I can merge when this is merged.

.github/workflows/e2e-test.yml
```diff
@@ -32,7 +43,8 @@ jobs:
         echo "export const config = $PROJECT_CONFIG; export const testAccount = $TEST_ACCOUNT" > firebase-config.js
     - name: Yarn install
       run: |
-        yarn add firebase
+        echo "Installing firebase@${{ github.event.inputs.tag }}"
+        yarn add firebase@${{ github.event.inputs.tag }}
```
Feiyang1, line 47: Can we just make version number a required field and use the version number? Perhaps it should accept both version number and tag name, and be renamed to `versionOrTag`.

PR #5562, opened 2021-09-30
[Auth] Switch compat UserCredential function to use promise chaining instead of async/await

sam-gc:
Hopefully to address b/198623196


lisajian (approved):

yuchenshi (approved):

sam-gc:
@eliot this appears to be fixed in 9.1.2. Regardless, try upgrading to 9.1.3 and seeing if you still see the issue. If you are, please file a new issue in the repo.

PR #5563, opened 2021-09-30
Disable Origin Trial code in WebChannel to reduce JS bundle size.

sampajano:
The new `ALLOW_ORIGIN_TRIAL_FEATURES` option was added to Closure in [this commit](https://github.com/google/closure-library/commit/23314e6077b4b7fb25bd9600a66480529e0ead1f).

Having this flag will save ~1.74 KB in the webchannel wrapper JS bundle compared with not having it (with the next Closure release).


sampajano:
@schmidt-sebastian PTAL :)

schmidt-sebastian (approved):

Issue #5569, opened 2021-10-01
App Check typing errors (incorrect implementations of interface)


Temez1:
### Environment

  * Firebase SDK version: 9.1.1
  * Reactfire version: 4.2.0
  

### Problem

I'm trying to compile my project & it throws 2 errors

```bash
node_modules/@firebase/app-check/dist/app-check-public.d.ts:87:22 - error TS2420: Class 'CustomProvider' incorrectly implements interface 'AppCheckProvider'.
  Type 'CustomProvider' is missing the following properties from type 'AppCheckProvider': getToken, initialize

node_modules/@firebase/app-check/dist/app-check-public.d.ts:171:22 - error TS2420: Class 'ReCaptchaV3Provider' incorrectly implements interface 'AppCheckProvider'.
  Type 'ReCaptchaV3Provider' is missing the following properties from type 'AppCheckProvider': getToken, initialize
```

The source code has the following comments, but the compiler doesn't really care about those.

```js
    /* Excluded from this release type: getToken */
    /* Excluded from this release type: initialize */
    /* Excluded from this release type: isEqual */
```

#### Steps to reproduce:

I started using firebase with Reactfire. I'm using just auth. I'm not using App Check.

#### Relevant Code:

TS config

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "useDefineForClassFields": true,
    "lib": ["DOM", "DOM.Iterable", "ESNext"],
    "allowJs": false,
    "skipLibCheck": false,
    "esModuleInterop": false,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "module": "ESNext",
    "moduleResolution": "Node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": true,
    "jsx": "react-jsx",
    "types": ["vite/client"]
  },
  "include": ["./src"]
}

```

hsubox76:
Thanks, I will look into fixing the typings errors.

In the meantime, so that you are unblocked, can I ask how you are importing Firebase? When directly using Firebase, there shouldn't be any app check code in the bundle if you didn't import app check, and you only imported auth, but I don't know if Reactfire does anything different.

Temez1:
Asked reactifire to give their input. I guess the TS checks the whole firebase library even if I'm not importing it.

PR #5570, opened 2021-10-01
Fix App Check typings

hsubox76:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5569

The two public facing providers (Custom and RecaptchaV3) implement an internal interface `AppCheckProvider` but they hide some of their public methods, causing them to be inconsistent with the internal interface.

If I tag `AppCheckProvider` `@internal`, the "implements AppCheckProvider" is still there on the two public types and there's an error because it can't find it.

I don't want to get rid of the "implements" because it is a helpful check in the internal code.


Feiyang1 (approved):

Feiyang1:
@hsubox76 please add a changeset

PR #5573, opened 2021-10-04
Fix app-check-compat providers

hsubox76:
Not sure how this was overlooked but the special provider classes were not exported on the firebase.appcheck object and `activate` was not set up to process them.

### DIFF
--- a/packages/app-check-compat/src/service.test.ts
+++ b/packages/app-check-compat/src/service.test.ts
@@ -79,7 +79,7 @@ describe('Firebase App Check > Service', () => {
   );
 
   it(
-    'activate(CustomProvider) calls modular initializeAppCheck() with' +
+    'activate({getToken: () => token}) calls modular initializeAppCheck() with' +
     ' a CustomProvider',
     () => {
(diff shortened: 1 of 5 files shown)


packages/app-check-compat/src/service.test.ts
```diff
@@ -65,7 +65,7 @@ describe('Firebase App Check > Service', () => {
 
   it(
     'activate("string") calls modular initializeAppCheck() with a ' +
-    'ReCaptchaV3Provider',
+      'ReCaptchaV3Provider',
```
Feiyang1, line 68: is it correctly formatted?

Feiyang1 (commented):
please add a changeset

packages/app-check-compat/src/service.test.ts
hsubox76: Yes, this seems to be prettier's rule. It's the same through the whole file.

hsubox76:
Changeset added.

Feiyang1 (approved):

Issue #5372, opened 2021-08-25
Unhandled network error rejection when using storage API


VeryCrazyDog:
### Describe your environment

  * Operating System version: Windows 10
  * Browser version: N/A (Node.js 14)
  * Firebase SDK version: 8.10.0
  * Firebase Product: storage

#### Steps to reproduce:
1. Simulate network error such as disconnect from network, blocking DNS server, blocking outgoing traffic, etc.
2. Execute storage API such as `getMetadata()` or `getDownloadURL()`.
3. Unhandled promise rejection is thrown. A sample console log is at below. My expected result is throwing `FirebaseError` with code `storage/retry-limit-exceeded`.

```
C:\Users\VCD\Developer\playground>node --unhandled-rejections=throw fbstorage-bug.js

C:\Users\VCD\Developer\playground\node_modules\node-fetch\lib\index.js:1461
                        reject(new FetchError(`request to ${request.url} failed, reason: ${err.message}`, 'system', err));
                               ^
FetchError: request to https://firebasestorage.googleapis.com/v0/b/xxxxxxxxxxxxxxxxx.appspot.com/o/image.png failed, reason: getaddrinfo ENOTFOUND firebasestorage.googleapis.com
    at ClientRequest.<anonymous> (C:\Users\VCD\Developer\playground\node_modules\node-fetch\lib\index.js:1461:11)
    at ClientRequest.emit (events.js:315:20)
    at TLSSocket.socketErrorListener (_http_client.js:469:9)
    at TLSSocket.emit (events.js:315:20)
    at emitErrorNT (internal/streams/destroy.js:106:8)
    at emitErrorCloseNT (internal/streams/destroy.js:74:3)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  type: 'system',
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND'
}
```

#### Relevant code:

```javascript
'use strict'

const firebase = require('firebase/app')
require('firebase/storage')

const app = firebase.initializeApp({
  apiKey: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
  authDomain: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
  projectId: 'xxxxxxxxxxxxxxxxx',
  storageBucket: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxx',
[...]

PR #5578, opened 2021-10-05
Do not throw from connection.send in Node.js

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5372
We should not throw from `connection.send` to enable retries.

### DIFF
--- a/packages/storage/test/browser/connection.test.ts
+++ b/packages/storage/test/browser/connection.test.ts
@@ -1,0 +1,17 @@
+import { expect } from "chai";
+import { SinonFakeXMLHttpRequest, useFakeXMLHttpRequest } from "sinon";
+import { ErrorCode } from "../../src/implementation/connection";
+import { XhrConnection } from "../../src/platform/browser/connection";
+
+describe('Connections', () => {
+    it('XhrConnection.send() should not reject on network errors', async () => {
+        const fakeXHR = useFakeXMLHttpRequest();
+        const connection = new XhrConnection();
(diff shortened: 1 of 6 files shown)


schmidt-sebastian (approved):

schmidt-sebastian (commented):
FWIW, I rewrote this here - https://github.com/firebase/firebase-js-sdk/pull/4973/files#diff-12013303701640021d24263089ec56052d40f6408d413dc10d78610499706fe2R56

Feiyang1:
@schmidt-sebastian Nice! Definitely agree with using async await here. Since you have started the refactoring in the other PR, I will leave this PR as is and let your PR change it.

Also added a couple of test cases if you'd like to review.

packages/storage/test/browser/connection.test.ts
```diff
@@ -0,0 +1,17 @@
...
+        const fakeXHR = useFakeXMLHttpRequest();
+        const connection = new XhrConnection();
+        const sendPromise = connection.send('testurl', 'GET');
+        // simulate an network error
+        ((connection as any).xhr_ as SinonFakeXMLHttpRequest).error();
+        await sendPromise;
+        expect(connection.getErrorCode()).to.equal(ErrorCode.NETWORK_ERROR);
+        fakeXHR.restore();
+    });
+});
```
schmidt-sebastian, line 17: Nit: empty line.

schmidt-sebastian (approved):
Test cases LGTM as well.

PR #5600, opened 2021-10-10
Fix AuthError incoherent structure.

### LINKED ISSUE #5599: Incoherent data structure on `AuthError`

- Operating System version: Windows 20H2 (Windows 10 10.0.19042)
- Browser version: Chrome 92 (Chrome: 92.0.4515.159)
- Firebase SDK version: 9.1.2 ("firebase": "^9.1.2")
- Firebase Product: auth

Data structure different in production environment than in [official doc][official-auth-error] and [source code][source-code].

- [ ] Also [this part](https://firebase.google.com/docs/auth/web/github-auth?authuser=0#expandable-1) seems to be outdated(v8 SDK yet shown up in a v9 page)
  like `auth.signInWithPopup(OAuthprovider)` should be `signInWithPopup(authInstance,OAuthProvider)`, so should `auth.signInWithEmailAndPassword(email, password)` (first arg: auth instance)... etc.
- [ ] Please confirm if I did everything right: I filled in two forms with screenshot and details but the feedback system seems not to be user-friendly. I cannot get and feedback(of update) of my feedback. The two feedback forms are about last `<li>` and the [official doc][official-auth-error] page

#### Steps to reproduce:

1. Create an account in GitHub with a email `userEmail`.
2. Create an account in Firebase Auth with same email `userEmail`.
3. Setup Firebase OAuth with GitHub
4. Login with OAuth. Then you'll get an Error `authErr` with error code `AuthErrorCodes.NEED_CONFIRMATION === "auth/account-exists-with-different-credential"`
5. This `authErr` should be with type `AuthError` but doesn't have the properties (like `email`) in [official doc][official-auth-error]
6. In the productive module `node_modules\@firebase\auth\dist\auth-public.d.ts`, this type is defined same as [official doc][official-auth-error],

#### Relevant Code:

These following two points are the reason why I think `AuthError` Class needs a change.
- Current master branch code for `AuthError`, extending `FirebaseError`
https://github.com/firebase/firebase-js-sdk/blob/53a9263903396a6ff666c48a75027c2db79748cf/packages/auth/src/model/public_types.ts#L123-L139

- Current `FirebaseError` type definition, with optional `customData` but no `email`,`phoneNumber`, etc.
https://github.com/firebase/firebase-js-sdk/blob/53a9263903396a6ff666c48a75027c2db79748cf/packages/util/src/errors.ts#L74-L94

- Code for reproduce error, step 4 and step 5
```typescript
try {
  const ghAccessToken = await signInWithPopup(auth, ghaProvider);
} catch (err: any) {
  if (err.code === FireAuthErrorCode.NEED_CONFIRMATION) {
    for (let k of Object.getOwnPropertyNames(err)) {
      console.log("error", k, err[k]);
      /*  Here it shows clearly that the err doesn't have the properties of `AuthError`.
      Instead, it has a customData field with same properties extended in `AuthError`  */
    }
  } else {
    console.error("uncaught Error");
    throw err;
  }
}
```

#### I wrote this feedback on official doc

- detail
When I use "firebase": "^9.1.2" with TypeScript, the `err`, thrown from signInWithPopup with error code `AuthErrorCodes.NEED_CONFIRMATION === "auth/account-exists-with-different-credential"`, has the same structure as Firebase error with fields {code,name,message,stack}. The extended part is a `customData` property. `customData` is an Object with properties: {appName, email, _tokenResponse}. My `currUser` doesn't have phoneNumber or tenantId(because I'm using popup not redirect).

- what I think it should be
- Class `AuthError` extends `FirebaseError` with Properties:
  `customData`, with type `AuthErrorData`(Object)
- Class `AuthErrorData` extends `Object` with properties in this paged

[official-auth-error]: https://firebase.google.com/docs/reference/js/auth.autherror
[source-code]: https://github.com/firebase/firebase-js-sdk/blob/53a9263903396a6ff666c48a75027c2db79748cf/packages/auth/src/model/public_types.ts#L123-L139

PabloLION:
Add a AuthErrorData interface for customData in `FirebaseError`
Change tenantid to tenantId, other tenantId are formatted like this, better than smallcase since ID is a "word"

### Discussion

this closes #5599  

### More TODOs

- [ ] Update [this offical document page](https://firebase.google.com/docs/reference/js/auth.autherror)

### Testing

I don't know how to test for this specific issue.

### API Changes

NONE

### DIFF
--- a/packages/auth/src/model/public_types.ts
+++ b/packages/auth/src/model/public_types.ts
@@ -117,7 +117,8 @@ export type NextOrObserver<T> = NextFn<T | null> | Observer<T | null>;
 
 /**
  * Interface for an `Auth` error.
- * 
+ *
+ *
  * @public
  */
 export declare interface AuthError extends FirebaseError {
@@ -138,7 +139,8 @@ export declare interface AuthError extends FirebaseError {
      * after redirection.
      */
     readonly tenantId?: string;
-  };}
+  };
+}
 
 /**
  * Interface representing an {@link Auth} instance's settings.


Pablion:
@sam-gc I tried to fix the MFA error last night and found it quite complex. I'll do what you said now.

PR #5602, opened 2021-10-11
Add GHA workflow for submitting tags to metric service.

yifanyang:
Add a GitHub Actions workflow file for submitting the tags information (name, commit sha1) to the SDK metric service.

- [x] ~The custom action `yifanyang/github-actions/health-metrics/release-diffing` is defined [here](https://github.com/yifanyang/github-actions/pull/1/files). I intend to have it reused across all Android/Apple/Web SDK repositories. For now, I'm putting it in my personal repository `yifanyang/github-actions`, but will figure out later where the best location for the code is.~ 
 The action is open sourced at https://github.com/FirebaseExtended/github-actions/tree/master/health-metrics/release-diffing.
- [x] ~At the moment, on the metric service end, only the logic of persisting tag information is implemented. Upon submission of this PR, we will be start to see tags stored in the database. I will send out the bundle definitions for you folks to review soon, so that we can associate releases/tags with meaningful metric data.~
 The bundle definitions are to be added in #5706.

### DIFF
--- a/.github/workflows/health-metrics-presubmit.yml
+++ b/.github/workflows/health-metrics-presubmit.yml
@@ -3,6 +3,10 @@ name: Health Metrics
 on: [push, pull_request]
 
 env:
+<<<<<<< HEAD:.github/workflows/health-metrics-test.yml
+=======
+  METRICS_SERVICE_URL: "https://api.firebase-sdk-health-metrics.com"
+>>>>>>> 7b37f3376 (Add GHA workflow for submitting tags to metric service.):.github/workflows/health-metrics-presubmit.yml
   GITHUB_PULL_REQUEST_NUMBER: ${{ github.event.pull_request.number }}
   # TODO(yifany): parse from git commit history directly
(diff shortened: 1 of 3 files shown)


.github/workflows/health-metrics-presubmit.yml
```diff
@@ -3,7 +3,7 @@ name: Health Metrics
 on: [push, pull_request]
 
 env:
-  METRICS_SERVICE_URL: ${{ secrets.METRICS_SERVICE_URL }}
+  METRICS_SERVICE_URL: "https://api.firebase-sdk-health-metrics.com"
```
Feiyang1, line 6: I think you removed it in the other PR?

Feiyang1 (approved):

.github/workflows/health-metrics-presubmit.yml
yifanyang: Yes. Once #5706  is merged, I'll don't know whether I'll need to address the merge conflict here or GitHub is smart enough to handle it. Let's see.

PR #5623, opened 2021-10-15
Log E2E Test Result and URL to release tracker

hsubox76:
Send the E2E test result to the release tracker. Put the release tracker functions URL in a secret.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -19,57 +19,59 @@ jobs:
         working-directory: './e2e'
 
     steps:
-    - name: Checkout Repo
-      uses: actions/checkout@master
-    - name: Set up Node (12)
-      uses: actions/setup-node@v2
-      with:
-        node-version: 12.x
-    - name: install Chrome stable
-      run: |
-        sudo apt-get update
-        sudo apt-get install google-chrome-stable
-    - name: Bump Node memory limit
-      run: echo "NODE_OPTIONS=--max_old_space_size=4096" >> $GITHUB_ENV
-    - name: Write project config
-      env:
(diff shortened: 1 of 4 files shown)


scripts/ci/notify-test-result.js
```diff
@@ -50,34 +50,65 @@ async function notifyTestResults() {
...
+    req.write(
+      JSON.stringify({
+        text: message
+      })
+    );
+    req.end();
+  });
+
+  const logPromise = new Promise((resolve, reject) => {
+    const testStatus = status === 'succeeded' ? 'pass' : 'fail';
```
Feiyang1, line 84: where is `status` defined?

hsubox76: line 32

.github/workflows/e2e-test.yml
```diff
@@ -19,57 +19,59 @@ jobs:
         working-directory: './e2e'
 
     steps:
-    - name: Checkout Repo
```
allspain, line 22: can you point out what actually changed? as A newcomer to this it's a little hard to follow

allspain: Actually it looks like it's just the release tracker url?

allspain (changes requested):
If you can just answer the questions that would be helpful to finish the review

Feiyang1 (approved):

.github/workflows/e2e-test.yml
hsubox76: Sorry, everything is in the diff because of applying auto formatting that changed all the indents. I think it is just adding the release tracker url. Most of the other changes are in the script itself.

allspain (approved):

PR #5643, opened 2021-10-20
Disable Fetch for ReactNative

schmidt-sebastian:
We have to use XHR instead of Fetch for ReactNative since `response.blob` is not supported (see https://github.com/facebook/react-native/issues/27741)

Addresses https://github.com/firebase/firebase-js-sdk/issues/1674

### DIFF
--- a/packages/firestore/test/integration/remote/remote.test.ts
+++ b/packages/firestore/test/integration/remote/remote.test.ts
@@ -53,7 +53,7 @@ describe('Remote Storage', () => {
     });
   });
 
-  it.only('can read deleted documents', () => {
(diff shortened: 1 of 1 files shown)


johnnyoshika:
> We have to use Fetch instead of XHR for ReactNative since `response.blob` is not supported (see [facebook/react-native#27741](https://github.com/facebook/react-native/issues/27741))
> 
> Addresses #1674

Is the comment incorrect? Should it say `use XHR instead of Fetch` for ReactNative?

schmidt-sebastian:
Yeah, it should be the other way around. Let me fix the discription.

PR #5676, opened 2021-10-28
Add retry logic to app check

hsubox76:
- Block exchange requests for certain periods of time on certain error codes
  - 400, 404: 1 day
  - Other error codes: exponential backoff
- Fix bugs causing duplicate calls to listeners
- Do not make multiple requests to exchange endpoint if a request is already in flight
- Start a token listener when App Check is initialized to avoid extra wait time on first getToken() call


packages/app-check/src/api.ts
```diff
@@ -97,6 +98,12 @@ export function initializeAppCheck(
 
   const appCheck = provider.initialize({ options });
   _activate(app, options.provider, options.isTokenAutoRefreshEnabled);
+  // Adding a listener will start the refresher and fetch a token if needed.
+  // This gets a token ready and prevents a delay when an internal library
+  // requests the token.
+  // Listener function does not need to do anything, its base functionality
+  // of calling getToken() already fetches token and writes it to memory/storage.
+  addTokenListener(appCheck, ListenerType.INTERNAL, () => {});
```
Feiyang1, line 106: Does it mean the refresher will never stop because there is always a listener? or will there be a case where the refresher is stopped while there is a listener?

packages/app-check/src/providers.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      new FirebaseError(AppCheckError.FETCH_STATUS_ERROR, 'some-message', {
+        httpStatus: 503
+      })
+    );
+    provider.initialize(app);
+    await expect(provider.getToken()).to.be.rejectedWith('503');
+    expect(exchangeTokenStub).to.be.called;
+    exchangeTokenStub.resetHistory();
+    // Try again immediately, should be rejected.
+    await expect(provider.getToken()).to.be.rejectedWith('503');
```
Feiyang1, line 87: Does each rejected call  increase the backoff? you are using timeouts 1.5s -> 10s -> 10s,  but it's not obvious to me they are exponentially increased.

packages/app-check/src/providers.ts
```diff
@@ -52,6 +63,16 @@ export class ReCaptchaV3Provider implements AppCheckProvider {
...
+    throwIfThrottled(this._throttleData);
+
+    if (!this._app || !this._platformLoggerProvider) {
+      // This should only occur if user has not called initializeAppCheck().
+      // We don't have an appName to provide if so.
+      // This should already be caught in the top level `getToken()` function.
+      throw ERROR_FACTORY.create(AppCheckError.USE_BEFORE_ACTIVATION, {
+        appName: ''
+      });
+    }
```
Feiyang1, line 75: I think we discussed it in another PR that we don't need this check and can assume the 2 properties are available.

packages/app-check/src/providers.ts
```diff
@@ -200,3 +242,56 @@ export class CustomProvider implements AppCheckProvider {
     }
   }
 }
+
+/**
+ * Set throttle data to block requests until after a certain time
+ * depending on the failed request's status code.
+ * @param httpStatus - Status code of failed request.
```
Feiyang1, line 249: missing `@param throttleData`

packages/app-check/src/api.ts
hsubox76: I don't know if I understand the question, when should the refresher be stopped? The refresher will always run at the appointed time but the REST call may be blocked by some of the new changes.

However this reminds me I missed something really important, which is that this shouldn't start if `isTokenRefreshEnabled` is false, and if the user later sets it to false, the listener should be removed.

Feiyang1: I don't think you need to remove this dummy listener in case `isTokenRefreshEnabled` is later set to false. It doesn't make any differences functionally, or does it?

hsubox76: Ah, that's right,`setTokenAutoRefreshEnabled(appCheck, false)` already stops the refresher.

hsubox76:
Also rebased and added recaptcha enterprise changes.

packages/app-check/src/internal-api.test.ts
```diff
@@ -385,6 +387,60 @@ describe('internal api', () => {
       );
       expect(token).to.deep.equal({ token: fakeRecaptchaAppCheckToken.token });
     });
+
+    it('throttles exponentially on 503', async () => {
```
Feiyang1, line 391: I don't think the test tests the `exponentially` part? You should either test the exponential algorithm or update the description to accurately reflect what is being tested.

Feiyang1 (commented):
If possible, please don't rebase PRs under review, use merge instead. Rebasing removes the ability for reviewers to only review the changes since their last review.

packages/app-check/src/internal-api.test.ts
hsubox76: Updated description and added a check just to make sure it's not 1 day, which would indicate it's taking the wrong path. (Because of randomization, can't check the exact time in the message, annoying to do math on the time string, provider tests do more detailed testing of the exact backoff times).

Feiyang1 (approved):

kevinthecheung (approved):

Issue #5685, opened 2021-10-29
onAuthStateChange called on first user token refresh and it should not happen


acanimal:
Hi,
I started as a tech question in [stackoverflow](https://stackoverflow.com/questions/69756283/firebase-auth-onauthstatechanged-triggered-also-on-token-refresh) but it seems it is a bug

### [REQUIRED] Describe your environment

  * Operating System version: macosx
  * Browser version: chrome 95
  * Firebase SDK version: 9.2.0, 9.1.3
  * Firebase Product: auth

### [REQUIRED] Describe the problem

As doc explains `onAuthStateChange` should be called only on signin and signout actions. It seems the listener is also called on the first user token refresh.

#### Steps to reproduce:
```html
<html>
  <script src="https://www.gstatic.com/firebasejs/9.1.3/firebase-app-compat.js"></script>
  <script src="https://www.gstatic.com/firebasejs/9.1.3/firebase-auth-compat.js"></script>
  <script>
    // Initialize Firebase
    const config = {
      apiKey: '',
      authDomain: '',
      databaseURL: '',
      projectId: '',
      storageBucket: '',
      messagingSenderId: '',
      appId: '',
    };
    const email = "mail12345@mail.com";
    const pass = "123456abc";
    const app = firebase.initializeApp(config);
    const auth = app.auth();

    const example = async () => {
      auth.onAuthStateChanged(async (user) => {
        console.log('onAuthStateChanged user uid', user ? user.uid : 'null');

        if (!user) {
          try {
            await auth.createUserWithEmailAndPassword(email, pass);
          } catch (e) {
            console.log('user already created');
          }

          console.log('signin')
          await auth.signInWithEmailAndPassword(email, pass);
        }
      })

      setInterval(() => {
        console.log('force token renewal');
        auth.currentUser.getIdToken(true);
      }, 5000)
    };

    example();
  </script>
</html>
```

#### The output

If you run for the first time or if there is no session you get the output:

```
onAuthStateChanged user uid null   <-- becuase there is no session stored
user already created
signin
onAuthStateChanged user uid mqLHdtQhwJPaXqIEbDqFBwFoxLz1 <-- this is because the signin
force token renewal <-- run each 5 secs
```

Having a session stored if you refresh the page you find the next output:

```
onAuthStateChanged user uid mqLHdtQhwJPaXqIEbDqFBwFoxLz1 <-- recover the session
force token renewal
[...]

PR #5686, opened 2021-10-29
[Auth] Fix double-onAuthStateChanged call bug

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5685

### DIFF
--- a/packages/auth/src/platform_browser/auth.test.ts
+++ b/packages/auth/src/platform_browser/auth.test.ts
@@ -319,6 +319,22 @@ describe('core/auth/initializeAuth', () => {
       });
     });
 
+    it('initialization sets the callback UID correctly', async () => {
+      const stub = sinon.stub(
+        _getInstance<PersistenceInternal>(inMemoryPersistence)
+      );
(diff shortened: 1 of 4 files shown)


packages/auth/src/platform_browser/auth.test.ts
```diff
@@ -319,6 +319,22 @@ describe('core/auth/initializeAuth', () => {
...
+      stub._get.returns(Promise.resolve(testUser(oldAuth, 'uid').toJSON()));
+      let authStateChangeCalls = 0;
+
+      const auth = await initAndWait(inMemoryPersistence) as AuthInternal;
+      auth.onAuthStateChanged(() => {
+        authStateChangeCalls++;
+      });
+
+      await auth._updateCurrentUser(testUser(auth, 'uid'));
+      expect(authStateChangeCalls).to.eq(1);
```
yuchenshi, line 335: nit: Please add a small wait (~200ms) before this to make sure `onAuthStateChanged` is not called again during that time frame

yuchenshi (approved):

PR #5698, opened 2021-11-03
Add regression tests to support primitive values in converters

thebrianchen:
Converters should allow `setDoc()` to be called with primitive types, as long as the provided converter converts the primitive to an object that matches `DocumentData`.


schmidt-sebastian (approved):
Thanks!

Issue #5631, opened 2021-10-18
Lighthouse Report Issue with auth iframe


StefanBratanov:
### [REQUIRED] Describe your environment

  * Operating System version: 10
  * Browser version: Chrome 94.0.4606.81
  * Firebase SDK version: 9.1.1
  * Firebase Product: auth

### [REQUIRED] Describe the problem
Hi, i am running into a Lighthouse issue stating that iframe element does not have a title. Please refer to the attached image. Can you please advise? I am using facebook, google, twitter as social sign ins which open a popup so it may be issue with that.

![image](https://user-images.githubusercontent.com/14827647/137748372-917d2640-68d8-4b23-a1d2-b37879fde0e0.png)

#### Steps to reproduce:
Run a Lighthouse Report on website using Firebase auth and facebook, google, twitter as social sign ins.

StefanBratanov:
Hi, i upgraded to latest version 9.2.0 and wanted to let you know i get a different lighthouse issue:

![image](https://user-images.githubusercontent.com/14827647/139527895-5497b0ed-d14e-4325-a4b9-90ae730fc17e.png)

sam-gc:
There's nothing in that iframe that will ever need to be interacted with. I'm not sure how to indicate to lighthouse that this is intentional. I will do some digging

StefanBratanov:
> There's nothing in that iframe that will ever need to be interacted with. I'm not sure how to indicate to lighthouse that this is intentional. I will do some digging

Thank you! Are you able to replicate the lighthouse issue in your environment? It may be an issue on my side.

sam-gc:
I am able to replicate, a fix is on the way!

PR #5700, opened 2021-11-04
[Auth] Add tabindex=-1 to iframe to fix issue with aria-hidden

sam-gc:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5631

This is a follow-up to https://github.com/firebase/firebase-js-sdk/pull/5633/files. There was still an issue with lighthouse until we add tabindex=-1 to the iframe as well

### DIFF
--- a/.changeset/witty-kids-buy.md
+++ b/.changeset/witty-kids-buy.md
@@ -1,0 +1,5 @@
+---
+"@firebase/auth": patch
+---
+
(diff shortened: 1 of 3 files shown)


yuchenshi (approved):
Thanks again for caring about accessibility!

lisajian (approved):

Issue #5692, opened 2021-11-02
Firebase app never terminates, even after deleteApp()


lcrocker:
### [REQUIRED] Describe your environment

Ubuntu 20.04 LTS, kernel 5.11.0-38-generic
NodeJS 14.16.0
Firebase 9.2.0
TypeScript 4.4.4

### [REQUIRED] Describe the problem

App never terminates. Regardless of whether deleteApp() is called or not, node is left waiting for something and hangs.
This may be the a duplicate of #4987 , which is closed and locked so I can't add this as a comment there.

#### Relevant Code:

```javascript
import { CloudFunctionsServiceClient } from "@google-cloud/functions";
import { initializeApp, deleteApp } from "firebase/app";
import { getFunctions, httpsCallable } from "firebase/functions";

import minimist from "minimist";
import fs from "fs";

import { JsonMap } from "./functions/src/generic/lib";

function delay(ms: number): Promise<void> {
        return new Promise((resolve) => {
                setTimeout(() => { resolve(); }, ms);
        });
}

class App {
        private client = new CloudFunctionsServiceClient();
        private fbApp = initializeApp({
                projectId: "xxxx-xxxx",
                apiKey: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
                authDomain: "xxxx-xxxx.firebaseapp.com",
                databaseURL: "https://xxxx-xxxx-default-rtdb.firebaseio.com",
                storageBucket: "xxxx-xxxx.appspot.com",
                messagingSenderId: "xxxxxxxxxxxx",
                appId: "1:xxxxxxxxxxxx:web:xxxxxxxxxxxxxxxxxxxxxx",
        });
        private functions = getFunctions(this.fbApp);

        private deviceId = "";
        private app: JsonMap = {};
        private verbose = false;
        private command = "";
        private filenames: string[] = [];

        constructor() {
                const argv = minimist(process.argv.slice(2), {
                        boolean: [ "v" ],
                        string: [ "d", "c", "_" ],
                });
                if (argv["v"]) this.verbose = true;
                if (argv["c"]) this.command = argv["c"];
                if (argv["d"]) this.deviceId = argv["d"];
                this.filenames = argv["_"];

                if (0 === this.filenames.length && "" === this.command) {
                        throw new Error("No operation specified");
                }
                if (! this.deviceId) {
                        throw new Error("No deviceId specified");
                }
[...]

PR #5701, opened 2021-11-04
clear pending timeout after promise.race

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5692

### DIFF
--- a/packages/functions/src/service.ts
+++ b/packages/functions/src/service.ts
@@ -53,18 +53,34 @@ export interface HttpResponseBody {
   };
 }
 
+interface CancellablePromise<T> {
+  promise: Promise<T>
+  cancel: () => void
+}
+
 /**
  * Returns a Promise that will be rejected after the given duration.
  * The error will be of type FunctionsError.
  *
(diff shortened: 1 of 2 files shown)


packages/functions/src/service.ts
```diff
@@ -247,12 +263,16 @@ async function call(
...
 
+  const failAfterHandle = failAfter(timeout);
   const response = await Promise.race([
     postJSON(url, body, headers, functionsInstance.fetchImpl),
-    failAfter(timeout),
+    failAfterHandle.promise,
     functionsInstance.cancelAllRequests
   ]);
 
+  // Always clear the failAfter timeout
```
hsubox76, line 273: Oh I was going to suggest putting this in the _delete method but I see that that _delete will resolve cancelAllRequests and bring us here, so that's fine.

hsubox76 (approved):

packages/functions/src/service.ts
Feiyang1: Yup

Issue #5628, opened 2021-10-16
Firebase Storage getDownloadURL doesn't finish running


ulbqb:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS 11.6
  * Browser version: No Browser
  * Firebase SDK version: 9.1.3
  * Firebase Product:Storage
  * Nodejs version: 14.18.0

### [REQUIRED] Describe the problem
When I try to get the Storage DownloadURL with Nodejs, the getDownloadURL function works fine, but Nodejs doesn't finish running.
firebase-js-sdk should support the Nodejs environment.
https://firebase.google.com/docs/web/environments-js-sdk

#### Steps to reproduce:
Run the following code in Nodejs (14.18.0).
In my environment, Download URL is output normally, but Nodejs execution does not finish.
#### Relevant Code:

```javascript
'use strict'
const { initializeApp } = require('firebase/app')
const { getStorage, ref, getDownloadURL } = require('firebase/storage')

initializeApp(
  {
    apiKey: '',
    authDomain: '',
    projectId: '',
    storageBucket: '',
    messagingSenderId: '',
    appId: '',
    measurementId: ''
  }
)

!(async () => {
  const storage = getStorage()
  const pathReference = ref(storage, 'sample.txt')
  const url = await getDownloadURL(pathReference)
  console.log(url)
})()
```

jbalidiong:
Hi @ulbqb, thanks for reporting this behavior. I tried following the steps you've provided to replicate the issue, however, I was able to get the download url. If I can replicate the issue, I can have a better look into it. Please share a minimal, but complete sample of a project that I can run locally.

LeMimit:
Hello ! 

I see the same problem with the functions of storage library called with mocha to test the security rules (`@firebase/rules-unit-testing`).

I tested the function `listAll` and `deleteObject`. Same problem with both.
I have to wait around 2 minutes after the test finishes to see the process finish.

PR #5703, opened 2021-11-05
clear the global timeout once an operation is done in the Storage SDK

Feiyang1:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5628


packages/storage/src/implementation/backoff.ts
```diff
@@ -58,22 +60,31 @@ export function start(
...
-    timeoutId = setTimeout(() => {
-      timeoutId = null;
+    retryTimeoutId = setTimeout(() => {
+      retryTimeoutId = null;
       f(handler, canceled());
     }, millis);
   }
 
+  function clearGlobalTimeout() {
+    if(globalTimeoutId){
```
schmidt-sebastian, line 70: Please fix.

This code also needs formatting.

schmidt-sebastian (approved):

packages/storage/src/implementation/backoff.ts
Feiyang1: Done.

PR #5706, opened 2021-11-05
Add bundle definitions and its measurement script.

yifanyang:


hsubox76 (commented):
Looks good, pending addition of `ReCaptchaEnterpriseProvider` to app-check.

repo-scripts/size-analysis/bundle-definitions/app-check.json
```diff
@@ -0,0 +1,92 @@
...
+        ]
+      },
+      {
+        "packageName": "firebase",
+        "versionOrTag": "exp",
+        "imports": [
+          {
+            "path": "app-check",
+            "imports": [
+              "initializeAppCheck",
```
hsubox76, line 24: Did these used to include `getToken`? I thought I remembered seeing them. It makes sense because if app check is being used, getToken() will eventually be called, if not by the user then by another library (storage, functions, etc).

yifanyang: Yes, it was there before. I got confused this morning when reading the [code sample](https://firebase.google.com/docs/app-check/web/recaptcha-enterprise-provider#initialize) for App Check Enterprise, which doesn't reference `getToken`, and I wondered why it was there. I forgot that it was actually Feiyang who added `getToken` for me in the bundle doc last week. I was actually about to ask you, but got stuck in some changeset warnings in this pull request.

Anyway, I have added it back both in the bundle definition doc and the json files in this pull request. Please take a look. Thanks!

egilmorez (commented):
iiuc this PR doesn't have any direct doc impact.  If that's wrong, LMK and I'll take a closer look.

Thanks!

repo-scripts/size-analysis/analyze-all-bundles.ts
```diff
@@ -0,0 +1,85 @@
...
+  fs.writeFileSync(`${temp}/${bundle}`, JSON.stringify(definitions, null, 2), {
+    encoding: 'utf-8'
+  });
+}
+
+function parseAnalysisOutput(product: string, output: string) {
+  const analyses = JSON.parse(fs.readFileSync(output, { encoding: 'utf-8' }));
+  const results: Report[] = [];
+  for (const analysis of analyses) {
+    const sdk = 'bundle';
```
Feiyang1, line 79: What does bundle mean?

repo-scripts/size-analysis/bundle-definitions/analytics.json
```diff
@@ -0,0 +1,32 @@
+[
+  {
+    "name": "logEvent",
+    "dependencies": [
+      {
+        "packageName": "firebase",
+        "versionOrTag": "exp",
```
Feiyang1, line 7: s/exp/latest/g

repo-scripts/size-analysis/bundle-definitions/auth.json
```diff
@@ -0,0 +1,202 @@
+[
```
Feiyang1, line 1: @sam-gc Can you please take a look at the Auth use cases to see if you want to add anything, or if anything is inaccurate?

For context, each object in the array is a bundle definition for a specific use case. It's basically a list of imports that customers need to import to use a certain feature from our products.

We are setting up a dashboard to continuously track the bundle size for them across versions.

repo-scripts/size-analysis/bundle-definitions/database.json
```diff
@@ -0,0 +1,256 @@
+[
+  {
+    "name": "Read data once",
```
Feiyang1, line 3: @schmidt-sebastian  Can you please take a look at the Database/Firestore/Storage use cases to see if you want to add anything, or if anything is inaccurate?

For context, each object in the array is a bundle definition for a specific use case. It's basically a list of imports that customers need to import to use a certain feature from our products.

We are setting up a dashboard to continuously track the bundle size for them across versions.

repo-scripts/size-analysis/cli.ts
```diff
@@ -89,4 +89,4 @@ yargs
     argv => runBundleAnalysis(argv as any)
   )
   .help()
-  .parseSync();
+  .parse();
```
Feiyang1, line 92: I think this is going to cause an unhandled promise error from Node. Why do you change it?

scripts/size_report/report_binary_size.ts
```diff
@@ -157,7 +158,8 @@ async function collectBinarySize(pkg: string): Promise<Report[]> {
 async function generateSizeReport(): Promise<BinarySizeRequestBody> {
   const reports: Report[] = [
     ...generateReportForCDNScripts(),
-    ...(await generateReportForNPMPackages())
+    ...(await generateReportForNPMPackages()),
+    ...(await generateReportForBundles())
```
Feiyang1, line 162: How long does this task take to finish? If it adds significant latency, we may want to set up a separate workflow for it.

Feiyang1:
Thanks for setting up the bundle size tracking! It looks good. It would be great if you can add some comments in `repo-scripts/size-analysis/analyze-all-bundles.ts` to explain the code and the function arguments. E.g. how does the`version` argument change the behavior of the code, who should set it, etc.

repo-scripts/size-analysis/analyze-all-bundles.ts
yifanyang: The name we put here will be appear as the name of a new section under "affected sdks" in a PR report. For example, there is another section named "bundle" in the report above in this PR [1]. We can put whatever name we want here.

[1] https://github.com/firebase/firebase-js-sdk/pull/5706#issuecomment-961607326

I suppose it is a limitation of the metric backend at the moment. When the API for receiving size measurements was first designed, the backend only recognizes http requests with a json body strictly in this format:

```
[
  {
    sdk: <some-string>,
    type: <some-string>,
    value: <some-integer>
  },
  ......
]
```

Writing a new API in the backend and deploying a new version of the service to GCP can be some work, therefore I'm reusing the API for bundle measurements here, although the semantic of the API doesn't make any sense in the context of "bundle-analysis". What do you think?

repo-scripts/size-analysis/bundle-definitions/analytics.json
yifanyang: Done.

repo-scripts/size-analysis/cli.ts
yifanyang: Reverted back.

scripts/size_report/report_binary_size.ts
yifanyang: The GHA runs for this pull request: 
- [push](https://github.com/firebase/firebase-js-sdk/runs/4170287745?check_suite_focus=true) (overall: 24m 49s, yarn build: 20m 59s, size test: 1m 47s)
- [pull_request](https://github.com/firebase/firebase-js-sdk/runs/4170288216?check_suite_focus=true) (overall: 19m 54s, yarn build: 16m 36s, size test: 1m 49s)

Sounds to me not a big concern. What do you think?

yifanyang:
> Thanks for setting up the bundle size tracking! It looks good. It would be great if you can add some comments in `repo-scripts/size-analysis/analyze-all-bundles.ts` to explain the code, hard coded values and the arguments. E.g. how does the`version` argument change the behavior of the code, who should set it, etc.

I have added some jsdoc on the method. Please take a look.

repo-scripts/size-analysis/analyze-all-bundles.ts
Feiyang1: I think it's fine. It would be great if you can copy your comment here to the code.

repo-scripts/size-analysis/analyze-all-bundles.ts
```diff
@@ -0,0 +1,104 @@
...
+ * 2. when omitted (in this case, `version` = null), this method measures the bundle size by
+ *    building test bundles with dependencies from local artifacts (e.g. produced by `yarn build`)
+ *
+ * #1 is intended only for manual runs for the purpose of back-filling historical size data. #2 is
+ * intended for CI runs that measure size for the current commit.
+ *
+ * More details on how a test bundle is built can be found in `bundle-analysis.ts`.
+ *
+ * @param {string} [version] - If present, the SDK version to run measurement against
+ * @returns {Promise<Report[]>} A list of bundle size measurements
```
Feiyang1, line 42: We use [tsdoc](https://tsdoc.org/) instead of jsdoc. Please update. You can take a look at any of our packages for reference.

scripts/size_report/report_binary_size.ts
Feiyang1: Yeah, looks fine to me too, the bottleneck is the build.

repo-scripts/size-analysis/analyze-all-bundles.ts
yifanyang: Is that just a standard of doc format or something I should use to generate the doc?

repo-scripts/size-analysis/analyze-all-bundles.ts
yifanyang: Added some comment in the code to explain the rationale here.

repo-scripts/size-analysis/analyze-all-bundles.ts
yifanyang: Done. Changed to tsdoc format.

Feiyang1 (approved):

PR #5708, opened 2021-11-05
Update build scripts to work with exports

Feiyang1:
Fixes https://github.com/firebase/firebase-admin-node/issues/1487

### DIFF
--- a/packages/firebase/compat/rollup.config.js
+++ b/packages/firebase/compat/rollup.config.js
@@ -243,7 +243,7 @@ const completeBuilds = [
     plugins: [
       sourcemaps(),
       resolveModule({
-        mainFields: ['lite-esm5', 'esm5', 'module']
+        exportConditions: ['liteesm5', 'esm5']
       }),
       typescriptPluginCDN,
       json(),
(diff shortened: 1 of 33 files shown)


packages/firebase/compat/rollup.config.js
```diff
@@ -243,7 +243,7 @@ const completeBuilds = [
     plugins: [
       sourcemaps(),
       resolveModule({
-        mainFields: ['lite-esm5', 'esm5', 'module']
+        exportConditions: ['liteesm5', 'esm5']
```
yuchenshi, line 246: Is `liteesm5` missing a dash here?

yuchenshi (approved):

allspain (approved):

allspain:
Approved but Yuchen's comment should be responded to prior to merging

packages/firebase/compat/rollup.config.js
Feiyang1: no, it's intentional to not include a dash here because it seems rollup doesn't like it.

PR #5725, opened 2021-11-11
Added instructions for running auth demo against emulator

lisajian:
### Discussion

As the title implies^^, a quick addition to run against the Auth Emulator

### Testing

`yarn run demo:emulator`

### API Changes

N/A


sam-gc (approved):
Thanks!

PR #5731, opened 2021-11-12
Fix ts-node build, add snapshot tests

schmidt-sebastian:


packages/firestore/src/platform/node/grpc_connection.ts
```diff
@@ -17,14 +17,14 @@
...
 // @ts-ignore To avoid using `allowSyntheticDefaultImports` flag.
-import module from 'module';
 
 import {
   Metadata,
   GrpcObject,
   credentials as GrpcCredentials,
   ServiceError
 } from '@grpc/grpc-js';
+import * as pgkJson from '@grpc/grpc-js/package.json';
```
hsubox76, line 27: I think we need to use a rollup plugin to hardcode this at build time using search/replace, importing from package.json directly was previously causing an error for some users after switching to using the module.exports field. Feiyang was alternatively looking into if there was some rollup config that would cause the module.createRequire to compile into something usable in Node if you want to check with him.

Issue #5687, opened 2021-10-30
Upgrade to 9.2.0 gives error in jest unit tests


LanderBeeuwsaert:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10
  * Browser version: ---
  * Firebase SDK version: 9.1.3 upgrade to 9.2.0
  * Firebase Product: ----

### [REQUIRED] Describe the problem

When upgrading to firebase 9.2.0 from 9.1.3 our jest unit tests start to fail with the following message:

    TypeError: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received 'http://localhost/index.node.cjs.js'

      at Function.createRequire (node_modules/jest-runtime/build/index.js:1418:23)
      at Object.<anonymous> (node_modules/@firebase/firestore/src/platform/node/grpc_connection.ts:43:24)
      at Object.<anonymous> (node_modules/@firebase/firestore-compat/dist/index.node.cjs.js:6:17)

We do use AngularFire (7.1.1) as well.
Running our Angular application itself seems to be ok. So it seems to be related with the upgrade + our jest configuration maybe?

Adding the full test log below:

 PASS  spec/tests/app/import-service/validators/country.validator.spec.ts
 PASS  spec/tests/app/patchers/organisation/1_0.patcher.spec.ts
 PASS  spec/tests/app/import-service/validators/firestore-doc.validator.spec.ts
 PASS  spec/tests/app/functions/country-service-backend.spec.ts
 PASS  spec/tests/app/scoring/calculate-partial-score.spec.ts
 PASS  spec/tests/app/import-service/validators/unique-column-value.validator.spec.ts
 PASS  spec/tests/app/version-patcher-service.spec.ts
 PASS  spec/tests/app/import-service/validators/column-exclusive-required.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/array-match.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/vat.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/positive-integer.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/inbetween-integer.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/length.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/firestore-key.validator.spec.ts
 PASS  spec/tests/app/import-service/validators/trimmed.validator.spec.ts
[...]

PR #5738, opened 2021-11-16
better meta url polyfill

Feiyang1:
Generate polyfill for Nodejs only.
Partially fixes https://github.com/firebase/firebase-js-sdk/issues/5687 (or rather just hide the problem that Jest will load Nodejs bundles as if it is testing in browser)

### DIFF
--- a/packages/firestore/rollup.config.js
+++ b/packages/firestore/rollup.config.js
@@ -38,7 +38,7 @@ function importMetaUrlPolyfillPlugin() {
     resolveImportMeta(property, { moduleId }) {
       if (property === 'url') {
         // copied from rollup output
-        return `new (require('u' + 'rl').URL)('file:' + __filename).href`;
+        return `new (require('url').URL)('file:' + __filename).href`;
       }
       return null;
     }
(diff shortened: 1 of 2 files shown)


packages/firestore/rollup.config.js
```diff
@@ -29,6 +29,22 @@ import { generateBuildTargetReplaceConfig } from '../../scripts/build/rollup_rep
...
+// Customize how import.meta.url is polyfilled in cjs nodejs build. We use it to be able to use require() in esm.
+// It only generates the nodejs version of the polyfill, as opposed to the default polyfill which
+// supports both browser and nodejs. The browser support is unnecessary and doesn't work well with Jest. See https://github.com/firebase/firebase-js-sdk/issues/5687
+function importMetaUrlPolyfillPlugin() {
+  return {
+    name: 'import-meta-url-current-module',
+    resolveImportMeta(property, { moduleId }) {
+      if (property === 'url') {
+        // copied from rollup output
+        return `new (require('u' + 'rl').URL)('file:' + __filename).href`;
```
hsubox76, line 41: what does separating the string do?

schmidt-sebastian (commented):
FYI I don't think I have enough context to provide a meaningful review. Please ping me if you need input.

packages/firestore/rollup.config.js
Feiyang1: It's what rollup outputs. I don't think it matters though. Updated

hsubox76 (approved):

PR #5739, opened 2021-11-16
Maybe fix deprecation warning

gaearon:
Related to #1847. I don't know if this works.

Note this would still warn if you _don't_ pass the persistence layer yourself. But at least it shouldn't warn if you do.

Sorry for lack of detail, @hsubox76 has context on the conversation

### DIFF
--- a/packages/auth/index.rn.ts
+++ b/packages/auth/index.rn.ts
@@ -22,7 +22,7 @@
  * just use index.ts
  */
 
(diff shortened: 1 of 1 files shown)


sam-gc (approved):

hsubox76:
Oops, sorry I forgot to ask until now but can you add a changeset? Just `yarn changeset` and follow prompts.

Testing checks out so I think that's all I need to merge.

PR #5741, opened 2021-11-17
fix firestore node tests

Feiyang1:
Use babel to compile tests for Node testing, so that `import.meta.url` can be compiled correctly.


Feiyang1:
fyi, @hsubox76

schmidt-sebastian (commented):
I tried to use this via the IntelliJ test runner but it failed with the following error:

```

✖ ERROR: /Users/mrschmidt/GitHub/firebase/firebase-js-sdk/packages/firestore/test/register.ts:18
import firebase from '@firebase/app-compat';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

The diff for the change:

https://paste.googleplex.com/6380219926577152

We really need to find a way to make Firestore run via ts-node, as this allows us to debug in the IDE.

Feiyang1:
I can debug tests fine with the new setup in IDE.  Are you not able to run tests at all?

schmidt-sebastian:
No, I am not able to. I will play around with this some more (but not sure if I can get to it today)

schmidt-sebastian (approved):

hsubox76 (approved):

PR #5743, opened 2021-11-18
Exclude NodeJS.ReadableStream from compat overloads

hsubox76:
Add NodeJS.ReadableStream to excluded builtin types in the compat overloads script. Maybe there's a more robust way to address this, or can look into it later.

Also fix some type errors in storage.

### DIFF
--- a/packages/storage/src/reference.ts
+++ b/packages/storage/src/reference.ts
@@ -167,7 +167,7 @@ export function getBytesInternal(
   );
   return ref.storage
     .makeRequestWithTokens(requestInfo, newBytesConnection)
-    .then((bytes) =>
+    .then(bytes =>
       maxDownloadSizeBytes !== undefined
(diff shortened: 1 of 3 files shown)


packages/storage/src/reference.ts
```diff
@@ -236,7 +236,7 @@ export function getStreamInternal(
 
   ref.storage
     .makeRequestWithTokens(requestInfo, newStreamConnection)
-    .then(stream => stream.pipe(result))
+    .then(stream => (stream as NodeJS.ReadableStream).pipe(result))
```
Feiyang1, line 239: Changes in this file seem unrelated?

hsubox76: This is "also fix some type errors in storage", storage was spitting out some errors on build that I was concerned about.

Feiyang1 (approved):
LGTM. Please fix CI errors.

Issue #5752, opened 2021-11-22
[9.5.0] Error: Cannot find module '@grpc/grpc-js/package.json' in electron


AviVahl:
### [REQUIRED] Describe your environment

  * Operating System version: Fedora 35
  * Browser version: Firefox 64
  * Firebase SDK version: 9.5.0
  * Firebase Product: firestore

#### Steps to reproduce:

Using firebase inside an electron app. I'm trying to upgrade from v8 to v9, and am encountering the following error:
```
Error: Cannot find module '@grpc/grpc-js/package.json'
Require stack:
- /home/avi/projects/<our-project-name>/node_modules/electron/dist/resources/electron.asar/renderer/init.js
    at Module._resolveFilename (:3000/internal/modules/cjs/loader.js:887)
    at Function.o._resolveFilename (:3000/electron/js2c/renderer_init.js:33)
    at Module._load (:3000/internal/modules/cjs/loader.js:732)
    at Function.f._load (:3000/electron/js2c/asar_bundle.js:5)
    at Function.o._load (:3000/electron/js2c/renderer_init.js:33)
    at Module.require (:3000/internal/modules/cjs/loader.js:959)
    at require (:3000/internal/modules/cjs/helpers.js:88)
    at Object.../../node_modules/@firebase/firestore/dist/index.node.cjs.js (:3000/vendors-node_modules_firebase_compat_firestore_dist_index_cjs_js.electron-renderer.js:15278)
    at __webpack_require__ (main.electron-renderer.js:7855)
    at Object.../../node_modules/@firebase/firestore-compat/dist/index.node.cjs.js (:3000/vendors-node_modules_firebase_compat_firestore_dist_index_cjs_js.electron-renderer.js:15)
```

Our setup bundles the library using webpack with `electron-renderer` target. I guess the attempt to dynamically create require breaks out of the bundling process.

#### Relevant Code:

The code that fails is:
```javascript
// This is a hack fix for Node ES modules to use `require`.
// @ts-ignore To avoid using `--module es2020` flag.
const require$1 = module__default["default"].createRequire(new (require('url').URL)('file:' + __filename).href);
// eslint-disable-next-line @typescript-eslint/no-require-imports
const { version: grpcVersion } = require$1('@grpc/grpc-js/package.json');
```

Now, I realize this piece of code already came up in https://github.com/firebase/firebase-js-sdk/issues/5687
The chosen solution doesn't work for our use case, I believe.

[...]

PR #5771, opened 2021-11-30
Hardcode grpc-js version

hsubox76:
Hardcode the grpc-js version used for logging. This isn't ideal, as getting the version at runtime provides the most accurate version for metrics and debugging, but working around problems with import.meta.url in ESM Node builds while ensuring CJS Node builds still work is causing a lot of problems. Make a TODO to change back to getting runtime version when there is a better way to support it across CJS and ESM Node builds.

See issues:
https://github.com/firebase/firebase-js-sdk/issues/5752
https://github.com/firebase/firebase-js-sdk/issues/5687

And PR:
https://github.com/firebase/firebase-js-sdk/pull/5532

### DIFF
--- a/packages/firestore/src/platform/node/grpc_connection.ts
+++ b/packages/firestore/src/platform/node/grpc_connection.ts
@@ -35,7 +35,7 @@ import { NodeCallback, nodePromise } from '../../util/node_api';
 import { Deferred } from '../../util/promise';
 
 // TODO: Fetch runtime version from grpc-js/package.json instead
-// when it becomes possible to use require() in Node ESM without hacks.
+// when there's a cleaner way to dynamic require in both Node ESM and CJS
 const grpcVersion = '__GRPC_VERSION__';
 
 const LOG_TAG = 'Connection';
--- a/packages/firestore/src/platform/node/grpc_connection.ts
+++ b/packages/firestore/src/platform/node/grpc_connection.ts
@@ -35,7 +35,7 @@ import { NodeCallback, nodePromise } from '../../util/node_api';
 import { Deferred } from '../../util/promise';
 
 // TODO: Fetch runtime version from grpc-js/package.json instead
-// when there's a cleaner way to dynamic require in both Node ESM and CJS
+// when there's a cleaner way to dynamic require JSON in both Node ESM and CJS
 const grpcVersion = '__GRPC_VERSION__';
 
 const LOG_TAG = 'Connection';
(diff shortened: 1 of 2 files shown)


Feiyang1:
`import.meta` is also used in https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/platform/node/load_protos.ts#L23, so this is not sufficient to solve the problem.

hsubox76:
> 

It looks like import.meta itself is not the problem, the createRequire is the problem, in that it creates a non-native `require` that doesn't work in certain Node-like environments (Electron, NextJS). This gets rid of that part. I can try to do an Electron test to make sure it solves the problem.

packages/firestore/src/platform/node/grpc_connection.ts
```diff
@@ -38,11 +34,9 @@ import { logError, logDebug, logWarn } from '../../util/log';
 import { NodeCallback, nodePromise } from '../../util/node_api';
 import { Deferred } from '../../util/promise';
 
-// This is a hack fix for Node ES modules to use `require`.
-// @ts-ignore To avoid using `--module es2020` flag.
-const require = module.createRequire(import.meta.url);
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-const { version: grpcVersion } = require('@grpc/grpc-js/package.json');
+// TODO: Fetch runtime version from grpc-js/package.json instead
```
allspain, line 37: IIRC in order to be able to merge this PR, we are waiting on changes from Firestore to surface the GRPC that is why this is not complete? or am I missing something?

hsubox76: No, this is a stopgap measure so that users can use the SDK without errors for now. In the future if and when the grpc-js team (outside of Firebase) is able to provide us with an export, the plan is we'll revert this and use that instead.

allspain: I see so we are not actually intending on passing any sort of version at the moment, but rather a placeholder string and in some future we will pass along the version. Makes sense, just wasn't clear if this TODO was for now or later, thank you for clarifying

allspain (approved):

packages/firestore/src/platform/node/grpc_connection.ts
hsubox76: Oh no, we will pass a version. The version will be grabbed by Rollup at build time based on the version of grpc-js installed when we build the Firebase SDK. Ideally we want the version installed by the user in their node_modules dir, which may differ, but since we specify the version in `@firebase/firestore`'s package.json dependencies, that should be the version they have for the most part.

PR #5772, opened 2021-12-01
[bundle-size][database] Refine bundle definition

yifanyang:


repo-scripts/size-analysis/bundle-definitions/database.json
```diff
@@ -151,7 +151,8 @@
               "ref",
               "onChildAdded",
               "onChildChanged",
-              "onChildRemoved"
+              "onChildRemoved",
+              "off"
```
schmidt-sebastian, line 155: Would it make sense to add another definition that just uses getDatabase+ref+onValue+off? `off` is probably not used by most users, so I don't want to make it the default.

yifanyang: Done. Made it a separate test case now. 

**p.s.** I had to force push the new change which is rebased onto a more recent commit on `master` branch. This is to let the diff report to pick the right comparison target commit. See [here](https://github.com/firebase/firebase-js-sdk/blob/master/.github/workflows/health-metrics-pull-request.yml#L7-L14) for more detail.

schmidt-sebastian (approved):

Feiyang1 (approved):

PR #5784, opened 2021-12-03
Add passthrough compat integration tests

lisajian:
### Discussion

Copied over integration passthrough integration tests (from https://github.com/firebase/firebase-js-sdk/pull/5361) over to compat layer. Additional changes include:

* Adding TODO's to enable tests once passthrough launches
* Remove unneeded timeout from all integ tests

### Testing

To test `auth-compat/`:
* Built emulator suite locally `npm run build` from `firebase-tools/`
* Change `xit` to `it`
* From `packages/auth-compat/` ran `node /Users/lisajian/dev/firebase-tools/lib/bin/firebase emulators:exec --project demo-emulatedproject --only auth "yarn run-s test:browser:integration test:node:integration test:webdriver"` (was having a little trouble with `yarn test:integration` b/c it wasn't picking up my local emulator build)

To test `auth/`:
* Same first two steps as above
* From `packages/auth/` ran `firebase emulators:exec --project p --only auth "yarn test:browser:integration:local"`

### API Changes

N/A

### DIFF
--- a/packages/auth-compat/test/integration/flows/custom.test.ts
+++ b/packages/auth-compat/test/integration/flows/custom.test.ts
@@ -27,6 +27,9 @@ import {
   initializeTestInstance,
   randomEmail
 } from '../../helpers/helpers';
+import { updateEmulatorProjectConfig } from '../../../../auth/test/helpers/integration/emulator_rest_helpers';
+
+declare const xit: typeof it;
 
 use(chaiAsPromised);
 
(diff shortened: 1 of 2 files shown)


packages/auth-compat/test/integration/flows/custom.test.ts
```diff
@@ -137,6 +140,65 @@ describe('Integration test: custom auth', () => {
...
+
+    afterEach(async () => {
+      await updateEmulatorProjectConfig(
+        JSON.stringify({
+          usageMode: 'DEFAULT'
+        })
+      );
+    });
+
+    xit('signs in with custom token in passthrough mode', async () => {
```
sam-gc, line 164: Would you mind adding a comment with a TODO(bug#) so we don't forget to enable these tests after launch?

packages/auth-compat/test/integration/flows/custom.test.ts
```diff
@@ -137,6 +140,65 @@ describe('Integration test: custom auth', () => {
...
+      const cred = await firebase.auth().signInWithCustomToken(customToken);
+      expect(firebase.auth().currentUser).to.eq(cred.user);
+      expect(cred.operationType).to.eq('signIn');
+
+      const { user } = cred;
```
sam-gc, line 169: nit: do this instead to avoid having to assert user every time on the lines below

```ts
const user = cred.user!;
```

packages/auth-compat/test/integration/flows/custom.test.ts
```diff
@@ -137,6 +140,65 @@ describe('Integration test: custom auth', () => {
...
+            uid,
+            claims: {
+              customClaim: 'other-claim'
+            }
+          });
+        }
+      });
+      const { user } = await firebase.auth().signInWithCustomToken(customToken);
+      const origToken = await user!.getIdToken();
+      await new Promise(resolve => setTimeout(resolve, 1000));
```
sam-gc, line 194: Please add a comment explaining the need for the delay

sam-gc (approved):

packages/auth-compat/test/integration/flows/custom.test.ts
lisajian: Ah, I think I copied this over from the non-passthrough refresh `'token can be refreshed'` test but it's not needed here. Removed in both auth/ and auth-compat/ tests!

PR #5799, opened 2021-12-07
[Auth] Add gmpid header

sam-gc:


lisajian (approved):
LGTM - Is there an internal bug and/or issue associated with this change?

sam-gc:
b/205577736

lisajian:
Thanks!

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -577,8 +578,15 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   _getFrameworks(): readonly string[] {
     return this.frameworks;
   }
-  _getSdkClientVersion(): string {
-    return this.clientVersion;
+  async _getAdditionalHeaders(): Promise<Record<string, string>> {
+    // Additional headers on every request
+    const headers: Record<string, string> = {
+      [HttpHeader.X_CLIENT_VERSION]: this.clientVersion,
```
yuchenshi, line 584: Is `this.clientVersion` always non-empty and non-null? If unsure, please add a check. Empty header values may break emulators)

yuchenshi (approved):

packages/auth/src/core/auth/auth_impl.ts
sam-gc: This is just moved code (the prior revision also always sends it). It's also always populated

PR #5816, opened 2021-12-15
Add double to bit encoding for OrderedCode

schmidt-sebastian:
This ports the double encoding from https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/index/OrderedCodeWriter.java

Note that Android does not have tests for this code since it is just a copy of the backend code.


packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+  return leadingZeros;
+}
+
+/**
+ * Returns the number of bytes required to store "value". Leading zero bytes
+ * are skipped.
+ */
+function unsignedNumLength(value: Uint8Array): number {
```
ehsannas, line 83: I think it's important write a comment as well as an assertion that `value` must be an array of size 8.

When I first read this, I thought if `value` is an array of size 1 with number `1` in it, then `numberOfLeadingZeros` would be `7`, and `numBits` would be `64 - 7`. In reality for number `1`, the `value` would be an array of size 8, with the least significant byte being `0b00000001`. All usages of `unsignedNumLength` get their `value` from `doubleToLongBits` which guarantees this.

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+  // This is just the number of bytes for the unsigned representation of the number.
+  const numBits = LONG_SIZE - numberOfLeadingZeros(value);
+  return Math.ceil(numBits / BYTE_SIZE);
```
ehsannas, line 86: For `value` being `0`, this function returns `0`.  I see this is also the case in the Java code. I'm curious if that's intended.

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+ * OrderedCodeWriter is a minimal-allocation implementation of the writing
+ * behavior defined by the backend.
+ *
+ * The code is ported from its Java counterpart.
+ */
+export class OrderedCodeWriter {
+  buffer = new Uint8Array(DEFAULT_BUFFER_SIZE);
+  position = 0;
+
+  writeNumberAscending(val: number): void {
```
ehsannas, line 99: I think it'd be good to write a comment here.
The Java code has:
```
    // Values are encoded with a single byte length prefix, followed
    // by the actual value in big-endian format with leading 0 bytes
    // dropped.
```

ehsannas: The Java code uses big-endian format. but IIUC this code uses little-endian. Is that intentional?

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+    const value = this.toOrderedBits(val);
+    const len = unsignedNumLength(value);
+    this.ensureAvailable(1 + len);
+    this.buffer[this.position++] = len & 0xff; // Write the length
+    for (let i = value.length - len; i < value.length; ++i) {
+      this.buffer[this.position++] = value[i] & 0xff;
+    }
+  }
+
+  writeNumberDescending(val: number): void {
```
ehsannas, line 109: same

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Encodes `val` into an encoding so that the order matches the IEEE 754
+   * floating-point comparison results with the following exceptions:
+   *   -0.0 < 0.0
+   *   all non-NaN < NaN
+   *   NaN = NaN
+   */
+  private toOrderedBits(val: number): Uint8Array {
+    const value = doubleToLongBits(val);
+    const isNegative = (value[0] & 0x80) !== 0;
```
ehsannas, line 128: I'm curious why `const isNegative = val < 0` is not used? Is it because of the three exceptions listed above?

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+      this.buffer[this.position++] = ~(value[i] & 0xff);
+    }
+  }
+
+  /**
+   * Encodes `val` into an encoding so that the order matches the IEEE 754
+   * floating-point comparison results with the following exceptions:
+   *   -0.0 < 0.0
+   *   all non-NaN < NaN
+   *   NaN = NaN
```
ehsannas, line 124: It's not immediately clear to me how this function satisfies the above 3 cases. Can you explain please?

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+   */
+  private toOrderedBits(val: number): Uint8Array {
```
ehsannas, line 126: Would it be legal for `val : number` to be `NaN` ?

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+    const value = doubleToLongBits(val);
```
ehsannas, line 127: What happens if `NaN` is passed to `doublleToLongBits`? (if that's even a possibility)

ehsannas (commented):
Added some comments and some questions that a JS newbie doesn't know...

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,162 @@
...
+ */
+const DEFAULT_BUFFER_SIZE = 1024;
+
+/**
+ * Converts a JavaScript number to a byte array (using little endian
+ * encoding).
+ */
+function doubleToLongBits(value: number): Uint8Array {
+  const dv = new DataView(new ArrayBuffer(8));
+  dv.setFloat64(0, value, false);
```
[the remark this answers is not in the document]
schmidt-sebastian: That's correct, we should be using big endian. Fixed and added comment for what the argument means.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: Added an assert here, as well as in `numberOfLeadingZerosInByte`. These asserts are removed in the final build but do add clarity.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: Yes, 0 can be represented using 0 bytes given the run-length encoding that we use.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: The format should be the same here. The backend code writes right to left (as it uses bit shifting to efficiently access the last 8 bytes). The bytes here can be accessed directly so we can write left to right.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: Done

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: Yeah, this is one of the test cases.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: It returns one of the possible NaN representations. There is no special-casing needed.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: We are not actually looking at a number that is encoded as a long. Instead, we are looking at the first element of the byte array that is encoded as a number from 0 to 255.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: All this magic is hidden by `DataView.setFloat64`. I don't have much more in terms of details :/

packages/firestore/src/index/ordered_code_writer.ts
ehsannas: that's my question.. wouldn't it be easier to detect negativeness by checking the number (`val`) rather than checking the byte array (`value`)?

No action is needed though.

ehsannas (approved):

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: We are checking whether the bit representation is "negative" not the number itself. Maybe the naming could use some improvement, but this is what the original implementation uses.

The tests fail if we just check if `val` is negative.

PR #5826, opened 2021-12-20
Port IndexValueWriter

schmidt-sebastian:
This ports:

https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/index/DirectionalIndexByteEncoder.java
https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/index/FirestoreIndexValueWriter.java
https://github.com/firebase/firebase-android-sdk/blob/master/firebase-firestore/src/main/java/com/google/firebase/firestore/index/IndexByteEncoder.java

### DIFF
--- a/packages/firestore/src/index/firestore_index_value_writer.ts
+++ b/packages/firestore/src/index/firestore_index_value_writer.ts
@@ -174,11 +174,10 @@ export class FirestoreIndexValueWriter {
   ): void {
     this.writeValueTypeLabel(encoder, INDEX_TYPE_REFERENCE);
     const path = DocumentKey.fromName(referenceValue).path;
-    for (let i = 0; i < path.length; ++i) {
-      const segment = path.get(i);
+    path.forEach(segment => {
       this.writeValueTypeLabel(encoder, INDEX_TYPE_REFERENCE_SEGMENT);
       this.writeUnlabeledIndexString(segment, encoder);
-    }
+    });
   }
 
   private writeValueTypeLabel(
--- a/packages/firestore/src/index/firestore_index_value_writer.ts
+++ b/packages/firestore/src/index/firestore_index_value_writer.ts
@@ -1,0 +1,197 @@
+/**
+ * @license
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { DocumentKey } from '../model/document_key';
+import { normalizeByteString, normalizeNumber } from '../model/normalize';
+import { isMaxValue } from '../model/values';
+import { ArrayValue, MapValue, Value } from '../protos/firestore_proto_api';
+import { fail } from '../util/assert';
+import { isNegativeZero } from '../util/types';
+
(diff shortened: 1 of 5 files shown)


packages/firestore/src/index/firestore_index_value_writer.ts
```diff
@@ -0,0 +1,197 @@
...
+    // Write separator to split index values
+    // (see go/firestore-storage-format#encodings).
+    encoder.writeInfinity();
+  }
+
+  private writeIndexValueAux(
+    indexValue: Value,
+    encoder: DirectionalIndexByteEncoder
+  ): void {
+    if ('nullValue' in indexValue) {
```
thebrianchen, line 74: optional: use `switch` here, since every if-condition is a single statement.

packages/firestore/src/index/firestore_index_value_writer.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private writeIndexEntityRef(
+    referenceValue: string,
+    encoder: DirectionalIndexByteEncoder
+  ): void {
+    this.writeValueTypeLabel(encoder, INDEX_TYPE_REFERENCE);
+    const path = DocumentKey.fromName(referenceValue).path;
+    for (let i = 0; i < path.length; ++i) {
```
thebrianchen, line 177: optional:
```
DocumentKey.fromName(referenceValue).path.forEach(segment => {
      this.writeValueTypeLabel(encoder, INDEX_TYPE_REFERENCE_SEGMENT);
      this.writeUnlabeledIndexString(segment, encoder); 
    });
```

packages/firestore/src/index/ordered_code_writer.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law | agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES | CONDITIONS OF ANY KIND, either express | implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { fail } from '../util/assert';
+import { ByteString } from '../util/byte_string';
+
+export class OrderedCodeWriter {
```
thebrianchen, line 20: optional: copy comment over and add a TODO that this section will be implemented in the future, rather than intentionally not implemented.

packages/firestore/src/model/values.ts
```diff
@@ -73,6 +81,10 @@ export function typeOrder(value: Value): TypeOrder {
 
 /** Tests `left` and `right` for equality based on the backend semantics. */
 export function valueEquals(left: Value, right: Value): boolean {
+  if (left === right) {
```
thebrianchen, line 84: I don't fully follow here, were we always able to do this direct comparison, or does this PR now allow us to do this direct comparison?

thebrianchen (approved):
lgtm with some questions and optional suggestions

packages/firestore/src/index/firestore_index_value_writer.ts
schmidt-sebastian: Unfortunately, switch does not work here since we are not doing an equals comparison on a value.

packages/firestore/src/index/firestore_index_value_writer.ts
schmidt-sebastian: Kept "path" but used forEach.

packages/firestore/src/index/ordered_code_writer.ts
schmidt-sebastian: This class is already in review: https://github.com/firebase/firebase-js-sdk/pull/5821

packages/firestore/src/model/values.ts
schmidt-sebastian: We were always able to do this. I added this now to make it more performant to do an equals check on MAX_VALUE.

schmidt-sebastian (commented):
Thanks!

Issue #5823, opened 2021-12-17
[Error] Firebase version compatibility problem in vercel (between 9.2.* -> 9.3.*)


YongwoonJang:
Hello, Here is my bug report

  * Operating System version: macOS Big Sur 11.6.1
  * Browser version: Chrome Version 96.0.4664.110 (Official Build) (x86_64)
  * Firebase SDK version: 9.4.1
  * Firebase Product: storage

### [REQUIRED] Describe the problem
Actually I use vercel and when I use "firebase@9.3.* or @9.4.*" below error happens 
But firebase@9.2.* is good to operate. 
below is error log

`
2021-12-17T16:30:40.729Z	7971a748-dd88-4c6e-973a-0d5c122617fc	ERROR	[2021-12-17T16:30:40.729Z]  @firebase/firestore: Firestore (9.4.0): INTERNAL UNHANDLED ERROR:  Error: ENOENT: no such file or directory, open '/var/task/node_modules/@firebase/firestore/dist/src/protos/google/firestore/v1/firestore.proto'
    at Object.openSync (fs.js:497:3)
    at Object.readFileSync (fs.js:393:35)
    at fetch (/var/task/node_modules/protobufjs/src/root.js:172:34)
    at Root.load (/var/task/node_modules/protobufjs/src/root.js:206:13)
    at Root.loadSync (/var/task/node_modules/protobufjs/src/root.js:247:17)
    at Object.loadProtosWithOptionsSync (/var/task/node_modules/@grpc/proto-loader/build/src/util.js:66:29)
    at Object.loadSync (/var/task/node_modules/@grpc/proto-loader/build/src/index.js:194:31)
    at loadProtos (/var/task/node_modules/@firebase/firestore/dist/index.node.cjs.js:14437:43)
    at newConnection (/var/task/node_modules/@firebase/firestore/dist/index.node.cjs.js:14459:20)
    at OnlineComponentProvider.createDatastore (/var/task/node_modules/@firebase/firestore/dist/index.node.cjs.js:18361:28)
[...]

PR #5830, opened 2021-12-21
Fix Vercel build

schmidt-sebastian:
This changes the manual replacement code to produce the exact output I got in 9.2:

```
url.fileURLToPath((typeof document === 'undefined' ? new (require('url').URL)('file:' + __filename).href : (document.currentScript && document.currentScript.src || new URL('index.node.cjs.js', document.baseURI).href)));
```

Fixes https://github.com/firebase/firebase-js-sdk/issues/5823

### DIFF
--- a/packages/firestore/rollup.config.js
+++ b/packages/firestore/rollup.config.js
@@ -43,10 +43,12 @@ function importMetaUrlPolyfillPlugin(filename) {
     resolveImportMeta(property, { moduleId }) {
       if (property === 'url') {
         // copied from rollup output
-        return "(typeof document === 'undefined' ? new (require('url').URL)"
-          + "('file:' + __filename).href : (document.currentScript && "
-          + `document.currentScript.src || new URL('${filename}', `
-          + "document.baseURI).href))";
+        return (
+          "(typeof document === 'undefined' ? new (require('url').URL)" +
+          "('file:' + __filename).href : (document.currentScript && " +
+          `document.currentScript.src || new URL('${filename}', ` +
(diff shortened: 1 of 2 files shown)


packages/firestore/rollup.config.js
```diff
@@ -31,16 +32,21 @@ import pkg from './package.json';
...
 
-// Customize how import.meta.url is polyfilled in cjs nodejs build. We use it to be able to use require() in esm.
-// It only generates the nodejs version of the polyfill, as opposed to the default polyfill which
-// supports both browser and nodejs. The browser support is unnecessary and doesn't work well with Jest. See https://github.com/firebase/firebase-js-sdk/issues/5687
-function importMetaUrlPolyfillPlugin() {
+// Customize how import.meta.url is polyfilled in cjs nodejs build. We use it to
+// be able to use require() in esm. It only generates the nodejs version of the
+// polyfill, as opposed to the default polyfill which supports both browser and
+// nodejs. The browser support is unnecessary and doesn't work well with Jest.
+// See https://github.com/firebase/firebase-js-sdk/issues/5687
```
schmidt-sebastian, line 39: Re-wrapped this comment

hsubox76: If you check the link, the ternary was originally removed because the document check was coming up true for Jest which caused it to take the second option which broke Jest somehow, so I think this will break Jest again. I am not sure why Vercel is using the Node bundle in what seems to be a non-Node environment but I'll try to look into that right away.

hsubox76 (approved):

Issue #5754, opened 2021-11-23
Confusing error types


laurentpayot:
### [REQUIRED] Describe your environment

  * Operating System version: Ubuntu 20.10
  * Browser version: Chrome 96.0.4664.45 64b
  * Firebase SDK version: 9.5.0
  * Firebase Product: firestore, auth, storage

### [REQUIRED] Describe the problem

To handle Firebase errors in a generic way I had to do dangerous type assertions to calm TypeScript down.

Firestore is a part of Firebase but, _according to TypeScript_,  `FirestoreError` is _not_ a `FirebaseError`.
But `AuthError` _is_ a `FirebaseError`, just like `StorageError` (looks like at least).
But `StorageError` has a special attribute `serverResponse`, not nested into `customData` like `AuthError` does…
Wat?

Here are the incriminated TypeScript interfaces:

```typescript
/** An error returned by a Firestore operation. */
export declare class FirestoreError extends Error {
    /**
     * The backend error code associated with this error.
     */
    readonly code: FirestoreErrorCode;
    /**
     * A custom error description.
     */
    readonly message: string;
    /** The custom name for all FirestoreErrors. */
    readonly name: string;
    /** The stack of the error. */
    readonly stack?: string;
    private constructor();
}
```

```typescript
/**
 * Interface for an `Auth` error.
 *
 * @public
 */
export declare interface AuthError extends FirebaseError {
    /** Details about the Firebase Auth error.  */
    readonly customData: {
        /** The name of the Firebase App which triggered this error.  */
        readonly appName: string;
        /** The email address of the user's account, used for sign-in and linking. */
        readonly email?: string;
        /** The phone number of the user's account, used for sign-in and linking. */
        readonly phoneNumber?: string;
        /**
         * The tenant ID being used for sign-in and linking.
         *
         * @remarks
         * If you use {@link signInWithRedirect} to sign in,
         * you have to set the tenant ID on the {@link Auth} instance again as the tenant ID is not persisted
         * after redirection.
         */
        readonly tenantId?: string;
    };
}
```

```typescript
/**
 * An error returned by the Firebase Storage SDK.
 * @public
 */
export declare interface StorageError {
    /**
     * A server response message for the error, if applicable.
     */
    serverResponse: string | null;
    code: string;
    customData?: Record<string, unknown>;
    name: 'FirebaseError';
    message: string;
    stack?: string;
}
```

```typescript
export declare class FirebaseError extends Error {
    readonly code: string;
[...]

PR #5831, opened 2021-12-21
Change FirestoreError to extend FirebaseError

schmidt-sebastian:
This changes FirestoreError to extend FirebaseError and updates the "Prune DTS" script to not remove this change again.

As part of this, I added a new test to "Prune DTS" and added a test runner for IntelliJ/WebStorm.

Fixes https://github.com/firebase/firebase-js-sdk/issues/5754

### DIFF
--- a/packages/storage/src/implementation/error.ts
+++ b/packages/storage/src/implementation/error.ts
@@ -1,4 +1,5 @@
-import { FirebaseError } from '@firebase/util';
+import { FirebaseError } from '@firebase/app';
+
 /**
  * @license
  * Copyright 2017 Google LLC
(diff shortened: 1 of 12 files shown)


packages/storage/src/implementation/error.ts
```diff
@@ -15,6 +14,9 @@ import { FirebaseError } from '@firebase/util';
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+import { FirebaseError } from '@firebase/util';
```
hsubox76, line 18: Did the placement of this affect the build?

hsubox76 (approved):

packages/storage/src/implementation/error.ts
schmidt-sebastian: No, I just cleaned it up.

PR #5835, opened 2021-12-23
Clear the lastLimboFreeSnapshot version on existence filter mismatch

schmidt-sebastian:
This fixes a problem where the SDK receives an existence filter mismatch and clears the mapping of documents to targets, but does not receive another snapshot that also clears the snapshot version (i.e. the device went in the background and did was not able to actually process a re-run of the query).

The change that this PR makes is that we always reset the lastLimboFreeSnapshot version to 0 if we get a target mismatch.

### DIFF
--- a/.changeset/unlucky-files-smoke.md
+++ b/.changeset/unlucky-files-smoke.md
@@ -2,4 +2,4 @@
 "@firebase/firestore": patch
 ---
 
-Fixed an issue that may have caused the SDK to raise incomplete snapshots when backgrounded during query execution.
+Fixed an issue that can cause incomplete Query snapshots when the SDK is backgrounded during query execution.
--- a/.changeset/unlucky-files-smoke.md
+++ b/.changeset/unlucky-files-smoke.md
@@ -1,0 +1,5 @@
+---
+"@firebase/firestore": patch
+---
+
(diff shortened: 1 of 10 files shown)


.changeset/unlucky-files-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/firestore": patch
+---
+
+Fixed an issue that may have caused the SDK to raise incomplete snapshots when backgrounded during query execution.
```
dconeybe, line 5: "may have caused" sounds to me like you are unsure that this is the root cause; however, it seems that the root cause is definitively known. Would it make sense to re-word this to imply that the buggy behavior is intermittent?

e.g. Fixed an issue that causes the SDK to sometimes raise incomplete snapshots...

(this comment was copied from the corresponding change to the Android SDK: https://github.com/firebase/firebase-android-sdk/pull/3267#discussion_r778347911)

dconeybe (changes requested):

.changeset/unlucky-files-smoke.md
schmidt-sebastian: Done

dconeybe (approved):

Issue #5839, opened 2021-12-27
Improper file extensions used in packaging


benmccann:
### [REQUIRED] Describe your environment

  * Operating System version: all
  * Browser version: any
  * Firebase SDK version: 9
  * Firebase Product: messaging. probably others

### [REQUIRED] Describe the problem

@Feiyang1 thanks for your efforts to improve packaging by adding `exports` to all `package.json` files (https://github.com/firebase/firebase-js-sdk/pull/5646). Unfortunately, I didn't catch this when reviewing that PR, but I'm afraid there's still an issue.

The file extensions are incompatible with Node. You must use `.mjs` for the ESM files (or set `"type": "module"` and use `.cjs` for the CJS files): https://nodejs.org/api/packages.html#determining-module-system

This bug was originally filed in the SvelteKit project (https://github.com/sveltejs/kit/issues/3083), but it's really an issue with the way Firebase is packaged. Vite will try to load Firebase directly using the Node APIs and will fail. It may work if you use a bundler like Webpack because the bundlers will transform the code to be valid before trying to run it, but Vite will try to run the code directly during SSR and will fail as a result.

#### Steps to reproduce:

Try to load `@firebase/messaging` from a Node app without bundling and it will fail

Run `npm install` followed by `npm run build` in this project: https://github.com/sacrosanctic/test-static-messaging

You will get the error:

> SyntaxError: Named export 'onBackgroundMessage' not found. The requested module '@firebase/messaging/sw' is a CommonJS module, which may not support all module.exports as named exports.

`@firebase/messaging/sw` is not CommonJS, but rather it's ESM. However, Node doesn't know that since there is neither `"type": "module"` nor an `.mjs` extension.

#### Relevant Code:

```javascript
import "firebase/compat/messaging"
```

argzdev:
Thanks for the report with the detailed explanation, @benmccann. We'll investigate this and check with feiyang or other engineers to see what we can do about this.

chliangGoogle:
@argzdev can you first see if you can reproduce this and assign back to me or feiyang to investigate?

argzdev:
Hi @chliangGoogle, I was able to repro the issue with the given GitHub [repro](https://github.com/sacrosanctic/test-static-messaging). I'll assign this case to you now, thanks!

jacobbowdoin:
https://github.com/jacobbowdoin/sveltefirets/issues/4 may be caused by this issue so I'll keep an eye on this.

[...]

PR #5857, opened 2022-01-05
DO NOT MERGE: Upgrade idb version from 3.2.0 to 7.0.0

hsubox76:
We had locked our [idb](https://github.com/jakearchibald/idb) dependency version to 3.2.0 because it was the last version to support IE. Since we no longer support IE, we can now upgrade to the most recent version which is now 7.0.0. In addition to being good practice in general, this allows us to provide ES Module versions of installations and messaging which may solve some other packaging problems for users of various frameworks.

The API has changed slightly between 3.2.0 and 7.0.0 and I've made the corresponding changes to installations/messaging code.

Also removed idb from the Renovate ignore list so we will now get Renovate PRs whenever idb is updated.

Problem about incompatibility with Node ESM mentioned in https://github.com/firebase/firebase-js-sdk/pull/5646 should be resolved in 7.0.0 (see https://github.com/jakearchibald/idb/pull/235#issuecomment-957589722). This should enable us to offer a Node ESM export in messaging, which will address https://github.com/firebase/firebase-js-sdk/issues/5839.

### DIFF
--- a/.changeset/quiet-years-sleep.md
+++ b/.changeset/quiet-years-sleep.md
@@ -1,0 +1,7 @@
+---
+'@firebase/installations': patch
+'@firebase/installations-compat': patch
+'@firebase/messaging': patch
+---
+
+Update `idb` dependency from 3.0.2 to 7.0.0.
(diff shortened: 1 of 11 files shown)


allspain (approved):

zwu52 (approved):
LGTM. thanks for the upgrade & work

hsubox76:
Hold off on merging pending discussion of dropping IE support.

hsubox76:
Obsolete, we have already upgraded.

PR #5858, opened 2022-01-05
Add jest exception

hsubox76:
Hack to fix for Vercel while not re-breaking for Jest.

### DIFF
--- a/packages/firestore/rollup.config.js
+++ b/packages/firestore/rollup.config.js
@@ -35,16 +35,22 @@ const util = require('./rollup.shared');
 // Customize how import.meta.url is polyfilled in cjs nodejs build. We use it to
 // be able to use require() in esm. It only generates the nodejs version of the
 // polyfill, as opposed to the default polyfill which supports both browser and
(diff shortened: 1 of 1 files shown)


packages/firestore/rollup.config.js
```diff
@@ -35,16 +35,22 @@ const util = require('./rollup.shared');
...
   return {
     name: 'import-meta-url-current-module',
     resolveImportMeta(property, { moduleId }) {
       if (property === 'url') {
-        // copied from rollup output
+        // Copied from rollup output
+        // Added a check for Jest (see issue 5687 linked above)
         return (
-          "(typeof document === 'undefined' ? new (require('url').URL)" +
+          "((typeof document === 'undefined' || process.env.JEST_WORKER_ID " +
```
schmidt-sebastian, line 51: Do we need to check whether `process` and `process.env` are defined?

hsubox76: Yes, that seems important, added.

schmidt-sebastian (approved):

PR #5860, opened 2022-01-05
Update mismatching UID error code to USER_MISMATCH

lisajian:
### Discussion

As the title implies - USER_MISMATCH is a more descriptive error code than INTERNAL_ERROR. This will also be reflected in iOS and Android.

### Testing

* `yarn test:browser:unit` passes


sam-gc (approved):

rosalyntan (approved):
Thanks, Lisa!

PR #5883, opened 2022-01-12
Load Protos from JSON

schmidt-sebastian:
This changes the Firestore SDK to load protos from JSON, which gets bundled directly into the source file. This should fix a bunch of problems with JS environments that use Node sources + Bundlers.


schmidt-sebastian:
Verified this via a Node sample app

hsubox76 (approved):

Issue #5854, opened 2022-01-04
Firebase v9 and Nuxt 2.15.8 - SyntaxError: Cannot use import statement outside a module


simeon9696:
I'm migrating my codebase to use the firebase version 9 SDK and started to get this error once I started using v9 firestore. I've already migrated the auth part of my project and that works perfectly.

I should note, this only happens when I'm statically generating my website (```npm run generate```) but not when building the site for SSR (```npm run build```). During development (```npm run dev```) I get no errors. The error turns up for every page that has to be generated. It seems to be referencing the messaging module but I'm not using messaging on every page.  I've tried using `"type": "module"` in my package.json but that hasn't helped. See an example of the error below:

```
 ERROR   /faqs

C:\Users\simeon.ramjit\Documents\projects\covid19-tt\node_modules\@firebase\messaging\dist\index.sw.esm2017.js:1
import '@firebase/installations';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)    
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)     
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (C:\Users\simeon.ramjit\Documents\projects\covid19-tt\node_modules\@firebase\messaging-compat\dist\index.cjs.js:8:10)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)    
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)     
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (C:\Users\simeon.ramjit\Documents\projects\covid19-tt\node_modules\firebase\compat\messaging\dist\index.cjs.js:3:1)
```

```plugins/firebase.js```

```
import firebase from 'firebase/compat/app'
[...]

PR #5884, opened 2022-01-13
Add CJS build for messaging/sw

hsubox76:
Although the messaging SW shouldn't technically have a Node version as SW's don't run in Node, some SSR applications generate pages in Node to be later loaded in the browser, and these builds break when they can't find a CJS version. This can also be fixed by adding an ESM Node version, but that can't currently be done until we update the `idb` dependency (see https://github.com/firebase/firebase-js-sdk/issues/5839). Adding a CJS build for now.

See https://github.com/firebase/firebase-js-sdk/issues/5854

### DIFF
--- a/.changeset/slow-bobcats-clap.md
+++ b/.changeset/slow-bobcats-clap.md
@@ -1,0 +1,5 @@
+---
+'@firebase/messaging': patch
+---
+
+Add a CJS bundle for messaging/sw. This enables some SSR frameworks to run their Node.js pipelines without erroring.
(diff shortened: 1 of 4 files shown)


zwu52:
LG. 1 q: `This can also be fixed by adding an ESM Node version, but that can't currently be done until we update the idb dependency`: does this mean that this change wants to be reverted after the idb upgrade ? If so, might be worth to add a TODO to remove

zwu52 (approved):

PR #5897, opened 2022-01-17
Fix Firestore failing to return empty results from the local cache

### LINKED ISSUE #5873: Collection snapshot queries with no documents don't store anything in cache. Always requires server snapshot.

  * Operating System version: MacOS 12.1
  * Browser version: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36
  * Firebase SDK version: 8.10.0
  * Firebase Product: firestore 

If a collection snapshot query doesn't have any documents, nothing is stored in the cache.

Each time we try to wait for an empty collection it has to wait for a server snapshot.

That should be a quick fix.  

It's either not from cache or fromCache in the metadata is wrong.  

Here's our code:

https://gist.github.com/burtonator/b9535237fd1c726960fb0fc77430160a

Maybe what's happening is that since no documents at ALL are found, the code assumes nothing is in the cache.

You should store a negative cache entry for that use case.

This is SUPER important for us because the latency is high when waiting for a server snapshot and this breaks our offline support.  

#### Relevant Code:

Pseudocode is here:

https://gist.github.com/burtonator/b9535237fd1c726960fb0fc77430160a

dconeybe:
Fix a bug where Firestore fails to return results from the local cache if the result set was empty.

This bug is due to the following logic in `event_manager.ts`:

https://github.com/firebase/firebase-js-sdk/blob/4ddf5566ca52b1f9ef776dd647c7d514e3b3f9f5/packages/firestore/src/core/event_manager.ts#L374-L375 where it 

which assumes that if the result set from the local cache is empty that there is no cached result; however, if the result set of the query is indeed empty then it should be returning that empty result set from the cache.

This fix improves the logic to use the presence of a resume token for the query to indicate that the empty result set is cached data and should be raised to the client.

Fixed #5873

### DIFF
--- a/packages/firestore/src/core/event_manager.ts
+++ b/packages/firestore/src/core/event_manager.ts
@@ -307,7 +307,8 @@ export class QueryListener {
         snap.mutatedKeys,
         snap.fromCache,
         snap.syncStateChanged,
-        /* excludesMetadataChanges= */ true
+        /* excludesMetadataChanges= */ true,
+        snap.resumeToken
       );
     }
     let raisedEvent = false;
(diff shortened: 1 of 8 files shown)


dconeybe:
Closing since this PR was obsoleted by https://github.com/firebase/firebase-js-sdk/pull/6624

PR #5903, opened 2022-01-19
Fix for Fireperf counter value might be null

jfcong:
Fix for Fireperf counter value might be null

### DIFF
--- a/packages/performance/src/resources/trace.ts
+++ b/packages/performance/src/resources/trace.ts
@@ -148,12 +148,7 @@ export class Trace implements PerformanceTrace {
     this.durationUs = Math.floor(duration * 1000);
     this.startTimeUs = Math.floor(startTime * 1000);
     if (options && options.attributes) {
-      for (const attributeName of Object.keys(options.attributes)) {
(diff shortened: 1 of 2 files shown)


jfcong:
changeset add

packages/performance/src/resources/trace.ts
```diff
@@ -183,7 +185,10 @@ export class Trace implements PerformanceTrace {
    */
   putMetric(counter: string, numAsInteger: number): void {
     if (isValidMetricName(counter, this.name)) {
-      this.counters[counter] = convertMetricValueToInteger(numAsInteger);
+      this.counters[counter] =
+        numAsInteger === undefined
+          ? 0
+          : convertMetricValueToInteger(numAsInteger);
```
jposuna, line 191: I think you can simplify this to:
```
this.counters[counter] = convertMetricValueToInteger(numAsInteger ?? 0);
```

jposuna (approved):

packages/performance/src/resources/trace.ts
jfcong: updated

PR #5906, opened 2022-01-20
Propagated redirectUrl by signInWithRedirect

atsuo1203:
I created this PR because I want to redirect to any page after authentication using twitter


sam-gc:
Hi @atsuo1203, thanks for submitting this PR. We won't be merging this PR for security reasons, as it would open up the potential for open redirects after sign-in. As a work-around, I'd suggest calling [`getRedirectResult()`](https://firebase.google.com/docs/reference/js/auth#getredirectresult) on page load, then changing page state depending on the result.

atsuo1203:
@sam-gc 
thank you for your answer. 
I understand it!!!
So I will implement it that way.

PR #5911, opened 2022-01-21
Fix e2e notify script again

hsubox76:
Send version and tag.

### DIFF
--- a/scripts/ci/notify-test-result.js
+++ b/scripts/ci/notify-test-result.js
@@ -38,12 +38,14 @@ async function notifyTestResults() {
   }
 
   let message = `E2E Tests ${status}`;
+  let versionOrTag;
 
   // Add version if it can find it in the workflow_dispatch event data.
   if (process.env.GITHUB_EVENT_PATH) {
     const wrPayload = require(process.env.GITHUB_EVENT_PATH);
     if (wrPayload.inputs && wrPayload.inputs.versionOrTag) {
       message += ` for release ${wrPayload.inputs.versionOrTag}.`;
+      versionOrTag = wrPayload.inputs.versionOrTag;
     } else {
       console.log(`Couldn't find versionOrTag in event payload.`);
     }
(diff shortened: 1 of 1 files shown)


scripts/ci/notify-test-result.js
```diff
@@ -102,11 +104,25 @@ async function notifyTestResults() {
 
     req.on('error', error => reject(error));
 
+    const data = {
+      testStatus,
+      testUrl: workflowUrl
+    };
+
+    if (versionOrTag) {
```
allspain, line 112: is there a risk that versionOrTag is a truthy non-string? something like `1`

scripts/ci/notify-test-result.js
```diff
@@ -102,11 +104,25 @@ async function notifyTestResults() {
...
+      // Matches a staging version tag pattern.
+      const match = versionOrTag.match(/^(\d+.\d+.\d+)-\d+$/);
+      if (match) {
```
allspain, line 115: Feels like we should do a length check in case match has fewer than expected matches. Accessing 1 directly without validating that it's a valid key seems dangerous

allspain (changes requested):

scripts/ci/notify-test-result.js
hsubox76: I don't think so because the workflow_dispatch input type[ defaults to `string`](https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/) if you don't specify a type and I haven't here:
https://github.com/firebase/firebase-js-sdk/blob/d612d6f6e4d3113d45427b7df68459c0a3e31a1f/.github/workflows/e2e-test.yml#L7

scripts/ci/notify-test-result.js
hsubox76: I'm not sure that outcome is possible, if the capturing group doesn't exist in the string, it won't match at all and match will be null. I think the only outcomes are null or an array of length 2. I tried a couple of test strings in the console and can't think of a case that would return non-null but also not have a captured group, let me know if I haven't thought of a case though, regex isn't my strength.

allspain: Sounds good I'm ok approving if you are confident

allspain (approved):

PR #5927, opened 2022-01-25
Add changeset for PR #5739

hsubox76:
Changeset for https://github.com/firebase/firebase-js-sdk/pull/5739

### DIFF
--- a/.changeset/yellow-suns-clean.md
+++ b/.changeset/yellow-suns-clean.md
@@ -2,4 +2,4 @@
 '@firebase/auth': patch
 ---
 
-Prevent React Native logging a warning about deprecation of `AsyncStorage` if user has provided the non-deprecated version.
+Prevent React Native from logging a warning about deprecation of `AsyncStorage` if the developer has provided the non-deprecated version.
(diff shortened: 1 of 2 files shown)


.changeset/yellow-suns-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/auth': patch
+---
+
+Prevent React Native logging a warning about deprecation of `AsyncStorage` if user has provided the non-deprecated version.
```
egilmorez, line 5: The preposition could clarify a little: "Prevent React Native from logging..."

egilmorez, line 5: Is "user" definitely the right term?  Or could it be "the developer" or even "the package.json file" or wherever this is defined?

I clearly do not know :)  But if we go with this, I'd suggest "if the user"

egilmorez (commented):
Couple of nits for you to look at, thanks!

.changeset/yellow-suns-clean.md
hsubox76: "developer" makes sense.

hsubox76: Done.

PR #5940, opened 2022-01-27
Limit size of heartbeat header

hsubox76:
Limit size of heartbeat header sent (based on the stringified and base64 encoded size).

If the total size of the heartbeats array is more than header size limit, try to send as many heartbeat objects as fit under the limit and keep the rest in cache to send in a future request.

If even the first heartbeat is too large to send, split it into two objects, which each have a piece of the dates array, and send the oldest.


hsubox76:
Closed in favor of https://github.com/firebase/firebase-js-sdk/pull/5945

PR #5942, opened 2022-01-27
Update backend apis

renkelvin:
This PR updates the 4 email auth apis (sign in, sign up, password reset, email link sign in) to support recaptcha enterprise.

### Development progress
Base apis and types update
- Backend API updates - https://github.com/firebase/firebase-js-sdk/pull/5942
- Public API and internal support types
- New backend error types

Core logic
- Implement getRecaptchaConfig API
- Implement recaptcha enterprise verifier
- Wire recaptcha enterprise verifier to email auth flows

PR #5943, opened 2022-01-27
Public api

renkelvin:
This PR updates the 4 email auth apis (sign in, sign up, password reset, email link sign in) to support recaptcha enterprise.

### Development progress
Base apis and types update
- https://github.com/firebase/firebase-js-sdk/pull/5942
- This PR - https://github.com/firebase/firebase-js-sdk/pull/5943
- New backend error types

Core logic
- Implement getRecaptchaConfig API
- Implement recaptcha enterprise verifier
- Wire recaptcha enterprise verifier to email auth flows


packages/auth/src/model/public_types.ts
```diff
@@ -199,6 +199,7 @@ export interface Auth {
    * @param persistence - The {@link Persistence} to use.
    */
   setPersistence(persistence: Persistence): Promise<void>;
+  setRecaptchaConfig(config: RecaptchaConfig): void;
```
sam-gc, line 202: Please either add a jsdoc comment with a small snippet, or add a `TODO` with a bug number here to add appropriate documentation when it's available from the techwriters.

packages/auth/src/model/public_types.ts
```diff
@@ -313,6 +314,18 @@ export interface Persistence {
...
 }
 
+/**
+ * An interface of the recaptcha configuration.
+ *
+ * @public
+ */
+export interface RecaptchaConfig {
+  /**
+   * The recaptcha enablement status for email password provider for current tenant.
```
sam-gc, line 324: These comments will end up in the reference docs, so please use long-form comments (i.e. include "the" in front of "email password provider" and "current tenant".

Also consider adding an `{@link EmailAuthProvider}`: "`status for the {@link EmailAuthProvider}`".

Don't worry too much about the comments, we can loop in @kevinthecheung towards the end to double-check/finesse the docs when we're ready for launch.

packages/auth/src/core/index.ts
```diff
@@ -60,6 +61,11 @@ export function setPersistence(
 ): Promise<void> {
   return getModularInstance(auth).setPersistence(persistence);
 }
+
+export function setRecaptchaConfig(auth: Auth, config: RecaptchaConfig): void {
```
sam-gc, line 65: Same here re. doc comment or `TODO` to add docs

packages/auth/src/model/auth.ts
```diff
@@ -60,6 +61,7 @@ export interface ConfigInternal extends Config {
 export interface AuthInternal extends Auth {
   currentUser: User | null;
   emulatorConfig: EmulatorConfig | null;
+  recaptchaConfig: RecaptchaConfig | null;
```
sam-gc, line 64: Since this isn't meant to be exposed to developers, please prefix it with `_` (this is the convention we follow for exported/shared fields that are not "public" to users of the library)

sam-gc (changes requested):

packages/auth/src/model/public_types.ts
renkelvin: Good point 👍

packages/auth/src/model/auth.ts
renkelvin: Will this be exposed to developers even `AuthInternal` is marked as `@internal`?

sam-gc: Not from a typescript point of view, but if you introspect the object in JS, the field will be visible. Since it's a public field, it won't be obfuscated by most minifiers.

renkelvin: Got it, thanks!

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -73,6 +74,7 @@ export const enum DefaultConfig {
 export class AuthImpl implements AuthInternal, _FirebaseService {
   currentUser: User | null = null;
   emulatorConfig: EmulatorConfig | null = null;
+  recaptchaConfig: RecaptchaConfig | null = null;
```
sam-gc, line 77: You'll also need to update this field to have the `_` prefix

sam-gc (approved):
Approval pending last comment

PR #5945, opened 2022-01-27
Cache heartbeats one per date

hsubox76:
Heartbeats need to be sent in a format where they are grouped by user agent, but some client-side functionality (such as removing records older than 30 days and limiting the header size) is easier to implement when the heartbeats are grouped by date. This change will group the records by date in client-side memory and storage, and convert it to the standard by-user-agent format when creating a header string to send to the backend.


packages/app/src/heartbeatService.ts
```diff
@@ -30,9 +30,14 @@ import { FirebaseApp } from './public-types';
 import {
   HeartbeatsByUserAgent,
   HeartbeatService,
-  HeartbeatStorage
+  HeartbeatStorage,
+  SingleDateHeartbeat
 } from './types';
 
+const MAX_HEADER_BYTES = 1000;
```
sam-gc, line 37: Does it matter whether or not this is exactly 1kb?

packages/app/src/heartbeatService.ts
```diff
@@ -82,28 +89,28 @@ export class HeartbeatServiceImpl implements HeartbeatService {
...
+      return;
     } else {
-      // There is no entry for this Firebase user agent. Create one.
-      heartbeatsEntry = {
-        userAgent,
-        dates: [date]
-      };
-      this._heartbeatsCache!.push(heartbeatsEntry);
+      // There is no entry for this date. Create one.
+      this._heartbeatsCache!.push({ date, userAgent });
```
sam-gc, line 103: Can you get rid of the `!` assertion now that it's explicitly set on line 92?

packages/app/src/heartbeatService.ts
```diff
@@ -120,19 +127,80 @@ export class HeartbeatServiceImpl implements HeartbeatService {
...
+      void this._storage.deleteAll();
+    }
     return headerString;
   }
 }
 
 function getUTCDateString(): string {
   const today = new Date();
-  return today.toISOString().substring(0,10);
+  return today.toISOString().substring(0, 10);
```
sam-gc, line 155: (small nit) maybe add a comment saying what part of the date string is retained after the substring operation?

sam-gc (approved):

packages/app/src/heartbeatService.ts
hsubox76: Thanks, done.

packages/app/src/heartbeatService.ts
hsubox76: Good idea, done.

hsubox76 (commented):
Removed the more rigorous byte count function that covers all unicode possibilities because we're doing the count on the final base64 string, which will always be 1 byte per character because allowed characters are restricted to A-Z,a-z,0-9,/+=

allspain (approved):

PR #5950, opened 2022-01-28
Fix compilation for MemoryRemoteDocumentCache

schmidt-sebastian:
... why does this not break the build!? No one knows.


wu-hui (approved):
Wooh, weird.

PR #5954, opened 2022-01-31
Add Index Schema

schmidt-sebastian:
Port of https://github.com/firebase/firebase-ios-sdk/pull/9268

### DIFF
--- a/packages/firestore/src/local/indexeddb_schema.ts
+++ b/packages/firestore/src/local/indexeddb_schema.ts
@@ -16,6 +16,7 @@
  */
 
 import { BatchId, ListenSequenceNumber, TargetId } from '../core/types';
+import { IndexKind } from '../model/field_index';
 import { ResourcePath } from '../model/path';
 import { BundledQuery } from '../protos/firestore_bundle_proto';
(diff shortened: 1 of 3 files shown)


packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -49,8 +53,11 @@ import {
  *     an auto-incrementing ID. This is required for Index-Free queries.
  * 10. Rewrite the canonical IDs to the explicit Protobuf-based format.
  * 11. Add bundles and named_queries for bundle support.
+ * 12. Add indexing support.
  */
-export const SCHEMA_VERSION = 11;
+export const SCHEMA_VERSION = INDEXING_ENABLED ? 12 : 11;
```
wu-hui, line 58: Ultranit: use the const defined below?

wu-hui (approved):

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: Updated

PR #5955, opened 2022-01-31
Enable abortOnError for Firestore

schmidt-sebastian:


hsubox76 (approved):
I think this might have been put in place while we had v8 and exp packages simultaneously, to ignore some of the unavoidable errors caused by that, so maybe we can start removing this across the board.

PR #5960, opened 2022-02-01
Port IndexManager APIs

schmidt-sebastian:
This PR ports the new APIs for IndexManager.

### DIFF
--- a/packages/firestore/src/local/index_manager.ts
+++ b/packages/firestore/src/local/index_manager.ts
@@ -15,6 +15,9 @@
  * limitations under the License.
  */
 
+import { Target } from '../core/target';
+import { DocumentKeySet, DocumentMap } from '../model/collections';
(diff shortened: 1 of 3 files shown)


packages/firestore/src/local/index_manager.ts
```diff
@@ -50,4 +53,86 @@ export interface IndexManager {
     transaction: PersistenceTransaction,
     collectionId: string
   ): PersistencePromise<ResourcePath[]>;
+
+  /**
+   * Adds a field path index.
+   *
+   * Values for this index are persisted asynchronously. The index will only be
```
wu-hui, line 60: This is referring to backfill right? If so, we should probably give a high level explanation of backfilling, instead of saying `persisted async`.

schmidt-sebastian: Improved comment.

PR #5964, opened 2022-02-02
Add recaptcha enterprise verifier

renkelvin:
This PR implements the new recaptcha enterprise verifier.
Not that it resolves the conflicts with existing recaptcha v2 and recaptcha v3 & enterprise from App check.

### Development progress
Base apis and types update
- https://github.com/firebase/firebase-js-sdk/pull/5942
- https://github.com/firebase/firebase-js-sdk/pull/5943
- New backend error types

Core logic
- Implement getRecaptchaConfig API
- This PR - https://github.com/firebase/firebase-js-sdk/pull/5964
- Wire recaptcha enterprise verifier to email auth flows

### DIFF
--- a/packages/auth/src/platform_browser/recaptcha/recaptcha.ts
+++ b/packages/auth/src/platform_browser/recaptcha/recaptcha.ts
@@ -17,9 +17,38 @@
 
 import { RecaptchaParameters } from '../../model/public_types';
 
+// reCAPTCHA v2 interface
 export interface Recaptcha {
   render: (container: HTMLElement, parameters: RecaptchaParameters) => number;
   getResponse: (id: number) => string;
   execute: (id: number) => unknown;
   reset: (id: number) => unknown;
+}
+
+export function isV2(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): grecaptcha is Recaptcha {
+  return grecaptcha !== undefined && (grecaptcha as Recaptcha).getResponse !== undefined;
+}
+
+// reCAPTCHA Enterprise & v3 shared interface
+ export interface GreCAPTCHATopLevel extends GreCAPTCHA {
+   enterprise: GreCAPTCHA;
+ }
+ 
+ // reCAPTCHA Enterprise interface
+ export interface GreCAPTCHA {
+   ready: (callback: () => void) => void;
+   execute: (siteKey: string, options: { action: string }) => Promise<string>;
+   render: (
+     container: string | HTMLElement,
+     parameters: GreCAPTCHARenderOption
+   ) => string;
+ }
+ 
+ export interface GreCAPTCHARenderOption {
+   sitekey: string;
+   size: 'invisible';
+ }
+
+ export function isEnterprise(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): grecaptcha is GreCAPTCHATopLevel {
+  return grecaptcha !== undefined && (grecaptcha as GreCAPTCHATopLevel).enterprise !== undefined;
 }
--- a/packages/auth/src/platform_browser/recaptcha/recaptcha.ts
+++ b/packages/auth/src/platform_browser/recaptcha/recaptcha.ts
@@ -25,7 +25,7 @@ export interface Recaptcha {
   reset: (id: number) => unknown;
 }
 
-export function isV2(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): grecaptcha is Recaptcha {
+export function isV2(grecaptcha: Recaptcha | GreCAPTCHA | undefined): grecaptcha is Recaptcha {
   return grecaptcha !== undefined && (grecaptcha as Recaptcha).getResponse !== undefined;
 }
 
@@ -49,6 +49,6 @@ export function isV2(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): gr
    size: 'invisible';
  }
 
- export function isEnterprise(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): grecaptcha is GreCAPTCHATopLevel {
+ export function isEnterprise(grecaptcha: Recaptcha | GreCAPTCHA | undefined): grecaptcha is GreCAPTCHATopLevel {
   return grecaptcha !== undefined && (grecaptcha as GreCAPTCHATopLevel).enterprise !== undefined;
 }
--- a/packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.test.ts
+++ b/packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.test.ts
@@ -1,0 +1,61 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ import { expect, use } from 'chai';
+ import chaiAsPromised from 'chai-as-promised';
+ import * as sinon from 'sinon';
+ import sinonChai from 'sinon-chai';
+  
+ import { testAuth, TestAuth } from '../../../test/helpers/mock_auth';
+ import * as fetch from '../../../test/helpers/mock_fetch';
+ import { _window } from '../auth_window';
+
+ import { MockGreCAPTCHATopLevel } from './recaptcha_mock';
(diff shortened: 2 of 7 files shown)


packages/auth/src/platform_browser/recaptcha/recaptcha.ts
```diff
@@ -17,9 +17,38 @@
...
+}
+
+// reCAPTCHA Enterprise & v3 shared interface
+ export interface GreCAPTCHATopLevel extends GreCAPTCHA {
+   enterprise: GreCAPTCHA;
+ }
+ 
+ // reCAPTCHA Enterprise interface
+ export interface GreCAPTCHA {
+   ready: (callback: () => void) => void;
```
sam-gc, line 39: Would you mind following the function format on these instead?
I.e.
```ts
export interface GreCAPTCHA {
  function ready(callback: () => void): void;
  // etc
}
```

Doesn't really matter a whole lot but that's the format we use for the most part in the rest of the SDK

packages/auth/src/platform_browser/recaptcha/recaptcha.ts
```diff
@@ -17,9 +17,38 @@
...
 
+// reCAPTCHA v2 interface
 export interface Recaptcha {
   render: (container: HTMLElement, parameters: RecaptchaParameters) => number;
   getResponse: (id: number) => string;
   execute: (id: number) => unknown;
   reset: (id: number) => unknown;
 }
+
+export function isV2(grecaptcha: Recaptcha | GreCAPTCHATopLevel | undefined): grecaptcha is Recaptcha {
```
sam-gc, line 28: Please add a test or 2 for the `isV2` and `isEnterprise` functions

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+ export class RecaptchaEnterpriseVerifier {
+   /**
+    * Identifies the type of application verifier (e.g. "recaptcha-enterprise").
+    */
+   readonly type = RECAPTCHA_ENTERPRISE_VERIFIER_TYPE;
+ 
+   /**
+    * Stores the recaptcha site key per tenant.
+    */
+   static siteKeys: [string: string];
```
sam-gc, line 43: Nit: please make this a `Record<string, string>` type (for consistency)

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+   }
+ 
+   /**
+    * Executes the verification process.
+    *
+    * @returns A Promise for a token that can be used to assert the validity of a request.
+    */
+   async verify(): Promise<string> {
+     // TODO(b/217382327): load and manage recaptcha config
+     const siteKey = '6LfyHwgeAAAAAKhL1Ujy7K1fBgX-l6bYdJpYJb_K';
```
sam-gc, line 65: Is this meant to be included here? Should we leave it empty until the config loading is done?

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+ 
+     const grecaptcha = _window().grecaptcha;
+     if (isEnterprise(grecaptcha)) {
+        grecaptcha.enterprise.ready(() => {
+            grecaptcha.enterprise.execute(siteKey, { action: 'login' })
+            .then((token) => {
```
sam-gc, line 71: The .then and .catch look redundant?

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+    *
+    * @returns A Promise for a token that can be used to assert the validity of a request.
+    */
+   async verify(): Promise<string> {
+     // TODO(b/217382327): load and manage recaptcha config
+     const siteKey = '6LfyHwgeAAAAAKhL1Ujy7K1fBgX-l6bYdJpYJb_K';
+ 
+     const grecaptcha = _window().grecaptcha;
+     if (isEnterprise(grecaptcha)) {
+        grecaptcha.enterprise.ready(() => {
```
sam-gc, line 69: Won't this still need to be wrapped in a promise block like the one below?

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+          });
+     }
+     
+     return new Promise<string>((resolve, reject) => {
+       loadReCAPTCHAEnterpriseScript(siteKey, () => {
+        const grecaptcha = _window().grecaptcha as GreCAPTCHATopLevel;
+         if (!grecaptcha) {
+           throw new Error('no recaptcha script loaded');
+         } else {
+           grecaptcha.enterprise.ready(() => {
```
sam-gc, line 86: Since this code is the same as the code above, consider moving it to a nested function or something that takes care of the promise and can be reused both here and above. That will also cut down on bundle growth as well

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -0,0 +1,106 @@
...
+               reject(error);
+             });
+           });
+         }
+       });
+     });
+   }
+ }
+ 
+ function loadReCAPTCHAEnterpriseScript(siteKey: string, onload: () => void): void {
```
sam-gc, line 101: Please use the _loadJS function instead: https://github.com/firebase/firebase-js-sdk/blob/00916e7192e745dbe693565f40200002a2acea61/packages/auth/src/platform_browser/load_js.ts#L25.

You'll have to rework the calling site a bit to accept a promise instead of a callback

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.test.ts
```diff
@@ -0,0 +1,61 @@
...
+     fetch.tearDown();
+   });
+ 
+   context('#verify', () => {
+     let recaptcha: MockGreCAPTCHATopLevel;
+     beforeEach(() => {
+        recaptcha = new MockGreCAPTCHATopLevel();
+        _window().grecaptcha = recaptcha;
+    });
+
```
sam-gc, line 54: Please also add a test for the failure case

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.test.ts
```diff
@@ -0,0 +1,61 @@
...
+ 
+ use(chaiAsPromised);
+ use(sinonChai);
+ 
+ describe('platform_browser/recaptcha/recaptcha_enterprise_verifier', () => {
+   let auth: TestAuth;
+   let verifier: RecaptchaEnterpriseVerifier;
+ 
+   beforeEach(async () => {
+     fetch.setUp();
```
sam-gc, line 38: I don't think you actually need mockFetch for this test suite

sam-gc (changes requested):

packages/auth/src/platform_browser/recaptcha/recaptcha.ts
renkelvin: Thanks for the suggestion. I prefer keep it for now as it's directly duplicated from App Check. I'm looking for a way to share these interfaces between Firebase Auth and App Check and we can make these changes there.

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
renkelvin: Yeah, good idea. I don't find a better way to reuse the code other than passing in the resolve and reject function pointers. Please let me know if you have any ideas!

packages/auth/src/platform_browser/recaptcha/recaptcha.ts
sam-gc: I didn't realize this implementation would be shared. We should probably put the implementation in the `@firebase/util` package or a new package altogether instead. @hsubox76 what do you think?

sam-gc: Depending on timelines we can put it entirely in auth for now then pull it into a separate package later for sharing

sam-gc (approved):

packages/auth/src/platform_browser/recaptcha/recaptcha.ts
renkelvin: Yeah, let's put it in auth first. @hsubox76 I think making the interface and implementation shared makes it less possible to conflict with App Check. Let me know if you have any ideas.

renkelvin:
Hey @sam-gc, do you have any idea about the failures CI here `Timeout of 20000ms exceeded`? It happens to Functions before and not Firestore without code changes. Should we ignore it?

sam-gc:
Yep, in general you can ignore failures if they're in different packages. The main thing is that the **Test Modified Packages** and **Test Auth** pass

PR #5965, opened 2022-02-02
Fix documentation

hsubox76:
Errors caught in review when generating new docs pages. See internal cl/425742150

api-documenter doesn't allow referencing members of an enum, causing there to just be a blank space in the documentation text where `{@link FactorId.PHONE}` should be, so I moved the `.PHONE` out of the link.


egilmorez (approved):
LG, thanks!

lisajian (approved):

PR #5966, opened 2022-02-02
Installations heartbeat implementation

hsubox76:
Replace sending v1 platform logging header with sending new heartbeats header.

Additionally:
- Sending the header to the "createInstallation" endpoint, which happens when an installations ID is registered. Previously it was only being sent to the "authTokens:generate" endpoint, which only happens on the auth token refresh, starting 7 days after registration.
- Adding app ID param to the body sent to the authTokens:generate endpoint.

### DIFF
--- a/.changeset/proud-otters-tap.md
+++ b/.changeset/proud-otters-tap.md
@@ -1,0 +1,5 @@
+---
+'@firebase/installations': patch
+---
+
+Update platform logging code to send to new endpoint.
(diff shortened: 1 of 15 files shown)


sam-gc (approved):

sam-gc (approved):

egilmorez (approved):
I believe there's only one public comment string here, and it looks fine.  Thanks!

PR #5967, opened 2022-02-02
App Check heartbeat implementation

hsubox76:
Replace sending v1 platform logging header with sending new heartbeats header.

Will point at master and add changeset after heartbeat controller PR is merged: https://github.com/firebase/firebase-js-sdk/pull/5723

### DIFF
--- a/packages/app-check/src/client.ts
+++ b/packages/app-check/src/client.ts
@@ -47,7 +47,7 @@ export async function exchangeToken(
   const headers: HeadersInit = {
     'Content-Type': 'application/json'
   };
-  // If platform logger exists, add the platform info string to the header.
+  // If heartbeat service exists, add heartbeat header string to the header.
   const heartbeatService = heartbeatServiceProvider.getImmediate({
     optional: true
   });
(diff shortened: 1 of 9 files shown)


packages/app-check/src/client.ts
```diff
@@ -42,17 +42,18 @@ interface AppCheckRequest {
...
   { url, body }: AppCheckRequest,
-  platformLoggerProvider: Provider<'platform-logger'>
+  heartbeatServiceProvider: Provider<'heartbeat'>
 ): Promise<AppCheckTokenInternal> {
   const headers: HeadersInit = {
     'Content-Type': 'application/json'
   };
   // If platform logger exists, add the platform info string to the header.
-  const platformLogger = platformLoggerProvider.getImmediate({
+  const heartbeatService = heartbeatServiceProvider.getImmediate({
```
allspain, line 51: probably should update the comment above since it reference platform logger which was renamed

allspain (approved):
left minor feedback

kevinthecheung (approved):

PR #5979, opened 2022-02-08
Update console.build.js to set a value on window

c4mb0t:
This change updates the console.build.js script that is used internally at Google to build the Firestore SDK.
Use of `eval` is removed and instead the code is exported using the global `window`.

### DIFF
--- a/packages/firestore-compat/tools/console.build.js
+++ b/packages/firestore-compat/tools/console.build.js
@@ -57,11 +57,7 @@ const es5OutputOptions = {
   format: 'iife'
 };
 
-const PREFIX = `
-goog.module('firestore');
-exports = eval(`;
-
-const POSTFIX = ` + '${EXPORTNAME};');`;
+const POSTFIX = `window['${EXPORTNAME}']=${EXPORTNAME};`;
 
 async function build() {
   const es5Bundle = await rollup.rollup(es5InputOptions);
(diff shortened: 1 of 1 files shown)


hsubox76:
Hi, can you provide some context for this change? Is this a request from the Console team?

c4mb0t:
Hello, for context please see the email thread "Eval in firestore_web_sdk".

packages/firestore-compat/tools/console.build.js
```diff
@@ -57,19 +57,15 @@ const es5OutputOptions = {
...
+const POSTFIX = `window['${EXPORTNAME}']=${EXPORTNAME};`;
 
 async function build() {
   const es5Bundle = await rollup.rollup(es5InputOptions);
   const {
     output: [{ code }]
   } = await es5Bundle.generate(es5OutputOptions);
 
-  const output = `${PREFIX}${JSON.stringify(String(code))}${POSTFIX}`;
+  const output = `${String(code)}${POSTFIX}`;
```
hsubox76, line 68: I think that you need a newline or at least whitespace before `${POSTFIX}` in order to match the code change in the CL. I guess the `code` always ends with a `;` so it's probably okay either way, but just in case.

c4mb0t: Through experimentation, I am seeing that `code` ends with both a semicolon and newline. Do you think we should add another one?

hsubox76: It's probably fine, I guess there was no newline in the previous version because it was JSON.stringified.

hsubox76 (approved):

PR #5980, opened 2022-02-08
Send target to backend before local processing

schmidt-sebastian:


wu-hui (approved):

dconeybe:
Note: This optimization is effectively reverted by https://github.com/firebase/firebase-js-sdk/pull/7229. This is because the number of documents that match the query in the local cache is now part of the request sent to the server, and, therefore, must be calculated before the request is sent to the server.

PR #5991, opened 2022-02-10
Integrate get recaptcha config api

renkelvin:


renkelvin:
> Looks like lint is passing?

Yeah, it's weird.  Just to confirm, we use ESLint right?

Tests fixed. It turns out the `mockEndPoint` method doesn't work for HTTP Get, which adds additional params to the url. I extend the method to `mockEndPointWithParams`.

packages/auth/src/api/authentication/recaptcha.ts
```diff
@@ -33,3 +33,24 @@ export async function getRecaptchaParams(auth: Auth): Promise<string> {
...
+  tenantId?: string;
+  clientType?: RecaptchaClientType;
+  version?: RecaptchaVersion;
+}
+
+interface GetRecaptchaConfigResponse {
+  recaptchaKey?: string;
+}
+
+export async function getRecaptchaConfig(auth: Auth,
```
sam-gc, line 47: Will this be an internal function or is it meant to be exposed to developers? If internal, please prefix with `_`

sam-gc (approved):

packages/auth/src/api/authentication/recaptcha.ts
renkelvin: It should be an internal function while it seems those similar api wrapper under `auth/src/api/` are not prefixed with `_`? https://github.com/firebase/firebase-js-sdk/blob/50de44c2e24ce6c200a638e0a684f476549ac3cf/packages/auth/src/api/authentication/sign_up.ts#L39

sam-gc: Oh good point, I forgot this was in the API section. We need to eventually go through and update those but for now yeah it's fine to leave as is :+1:

renkelvin: Sounds good 👍

PR #5993, opened 2022-02-10
Do not invoke the App Check token listener for the same token string.

ehsannas:


packages/firestore/src/api/credentials.ts
```diff
@@ -461,10 +461,9 @@ export class FirebaseAppCheckTokenProvider
    * we can unregister it.
    */
   private tokenListener!: AppCheckTokenListener;
-
   private forceRefresh = false;
-
   private appCheck: FirebaseAppCheckInternal | null = null;
+  private latestToken: string | null = null;
```
schmidt-sebastian, line 466: Please include appCheck in name. We also have an auth token.

packages/firestore/src/api/credentials.ts
```diff
@@ -482,7 +481,15 @@ export class FirebaseAppCheckTokenProvider
...
         }
-        return changeListener(tokenResult.token);
+        if (tokenResult.token !== this.latestToken) {
+          logDebug('FirebaseAppCheckTokenProvider', 'Received a new token.');
+          return changeListener(tokenResult.token);
+        }
+        logDebug(
+          'FirebaseAppCheckTokenProvider',
+          'Received a token that is the same as the existing token.'
+        );
```
schmidt-sebastian, line 491: Can we combine the log lines into one? You could just have one return statement, which might the code easier to follow (but that is just personal style).

schmidt-sebastian:
Please add changelog as well

packages/firestore/src/api/credentials.ts
schmidt-sebastian: I just want this to be a bit shorter and have less strings, so that our size does not go up by all that much:

```
        const tokenUpdated = tokenResult.token !== this.latestAppCheckToken;
        this.latestAppCheckToken = tokenResult.token;
        logDebug(
          'FirebaseAppCheckTokenProvider',
          `Received ${tokenUpdated ? 'new' : 'existing'} token.`
        );
        return tokenUpdated
          ? changeListener(tokenResult.token)
          : Promise.resolve();
```

Our log lines make up a significant percentage of our SDK size.

ehsannas: Our log lines make up a significant percentage of our SDK size.

> Ah! I did not know this. I'll keep that in mind.

schmidt-sebastian (approved):

Issue #5983, opened 2022-02-09
'The requested module '@grpc/grpc-js' is expected to be of type CommonJS' Getting issue while deploying next js application by gcp CICD pipeline.


krishna-fenixwork:
Hey, I'm building the next js application. Recently, while deploying my project through the GCP CICD pipeline, I'm getting this error - 

`
info  - Checking validity of types...
warn  - No ESLint configuration detected. Run next lint to begin setup
info  - Creating an optimized production build...
info  - Compiled successfully
info  - Collecting page data...

> Build error occurred
file:///workspace/node_modules/@firebase/firestore/dist/index.node.mjs:7
import { credentials, loadPackageDefinition, Metadata } from '@grpc/grpc-js';
                                             ^^^^^^^^
SyntaxError: The requested module '@grpc/grpc-js' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
For example:
import pkg from '@grpc/grpc-js';
const { credentials, loadPackageDefinition, Metadata } = pkg;
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:98:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:137:5)
    at async Loader.import (internal/modules/esm/loader.js:165:24) {
  type: 'SyntaxError'
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
`

I couldn't understand what causing the issue. I tried downgrading the version of firebase and firebase-functions too, but It didn't seem to work. Can anyone help me with that?

jbalidiong:
Hi @krishna-fenixwork, thanks for reaching out. Would you mind sharing more specific details:

- Version numbers of Firebase/Nextjs/etc
- A small code snippet demonstrating the issue

schmidt-sebastian:
@hsubox76 Christina - should we fix our imports?

hsubox76:
Looking around, it seems like if we aren't able to change grpc-js then we should probably follow the pattern given in the error message.

krishna-fenixwork:
Hey, @jbalidiong These are the packages that I'm using - 

`
"next": "^12.0.7",
"firebase": "^9.4.1",
"firebase-admin": "^10.0.0",
"firebase-functions": "^3.16.0",
`

I've also tried to upgrade them to their latest version but the issue still persists so I downgrade them again to these versions. I've my firebase configuration file where I initialized firebase and firestore like this - 

`
[...]

PR #6002, opened 2022-02-11
Use package import for GRPC

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-js-sdk/issues/5983


hsubox76 (approved):
I guess a 0.5KB increase isn't a big deal.

schmidt-sebastian:
The increase is also only for Node.

PR #6008, opened 2022-02-14
Implement Index Scanning

schmidt-sebastian:
This adds all index evaluation logic to the Web SDK and ports all tests from Android. The query logic is mostly the same as Android, with the exception of notIn/notEquals queries which are executed by doing scans across split ranges.

Example:

```
Query q = coll.where("a", ">", 2).where("a", "!=", 3);

Android: SELECT * FROM (SELECT * FROM coll WHERE a > 2) WHERE a != 3

Web: Scan [a>2, a<3] && [a>3, ∞]
```

### DIFF
--- a/packages/firestore/src/local/index_manager.ts
+++ b/packages/firestore/src/local/index_manager.ts
@@ -103,13 +103,12 @@ export interface IndexManager {
 
   /**
    * Returns the documents that match the given target based on the provided
-   * index.
+   * index or `null` if the target does not have a matching index.
    */
   getDocumentsMatchingTarget(
     transaction: PersistenceTransaction,
-    fieldIndex: FieldIndex,
     target: Target
-  ): PersistencePromise<DocumentKeySet>;
+  ): PersistencePromise<DocumentKeySet | null>;
 
   /**
    * Returns the next collection group to update. Returns `null` if no group
--- a/packages/firestore/src/local/indexeddb_index_manager.ts
+++ b/packages/firestore/src/local/indexeddb_index_manager.ts
@@ -847,82 +847,52 @@ export class IndexedDbIndexManager implements IndexManager {
       return indexRanges;
     }
 
-    const result: IDBKeyRange[] = [];
-    for (const indexRange of indexRanges) {
-      // Remove notIn values that are not applicable to this index range
-      const lowerBound = new Uint8Array(indexRange.lower[3]);
-      const upperBound = new Uint8Array(indexRange.upper[3]);
-      const filteredRanges = notInValues.filter(
-        v =>
-          compareByteArrays(v, lowerBound) >= 0 &&
-          compareByteArrays(v, upperBound) <= 0
-      );
-
-      if (filteredRanges.length === 0) {
-        result.push(indexRange);
-      } else {
-        // Use the existing bounds and interleave the notIn values. This means
-        // that we would split an existing range into multiple ranges that exclude
-        // the values from any notIn filter.
-        result.push(...this.interleaveRanges(indexRange, filteredRanges));
-      }
-    }
-    return result;
-  }
-
-  /**
-   * Splits up the range defined by `indexRange` and removes any values
-   * contained in `barrier`. As an example, if the original range is [1,4] and
-   * the barrier is [2,3], then this method would return [1,2), (2,3), (3,4].
-   */
-  private interleaveRanges(
-    indexRange: IDBKeyRange,
-    barriers: Uint8Array[]
-  ): IDBKeyRange[] {
     // The values need to be sorted so that we can return a sorted set of
     // non-overlapping ranges.
-    barriers.sort((l, r) => compareByteArrays(l, r));
+    notInValues.sort((l, r) => compareByteArrays(l, r));
 
     const ranges: IDBKeyRange[] = [];
 
-    // The first index range starts with the lower bound and ends before the
-    // first barrier.
-    ranges.push(
-      IDBKeyRange.bound(
-        indexRange.lower,
-        this.generateNotInBound(indexRange.lower, barriers[0]),
-        indexRange.lowerOpen,
-        /* upperOpen= */ true
-      )
-    );
+    // Use the existing bounds and interleave the notIn values. This means
+    // that we would split an existing range into multiple ranges that exclude
+    // the values from any notIn filter.
+    for (const indexRange of indexRanges) {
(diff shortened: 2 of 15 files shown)


packages/firestore/src/local/index_manager.ts
```diff
@@ -103,13 +103,12 @@ export interface IndexManager {
 
   /**
    * Returns the documents that match the given target based on the provided
-   * index.
+   * index. Returns `null` if the target does not have a matching index.
```
schmidt-sebastian, line 106: This change is pending review on Android: https://github.com/firebase/firebase-android-sdk/pull/3444

packages/firestore/src/platform/node/base64.ts
```diff
@@ -16,23 +16,16 @@
...
-  return new Buffer(encoded, 'base64').toString('binary');
+  // Note: We used to validate the base64 string here via a regular expression.
+  // This was removed to improve the performance of indexing.
+  return Buffer.from(encoded, 'base64').toString('binary');
 }
 
 /** Converts a binary string to a Base64 encoded string. */
 export function encodeBase64(raw: string): string {
-  return new Buffer(raw, 'binary').toString('base64');
+  return Buffer.from(raw, 'binary').toString('base64');
```
schmidt-sebastian, line 28: This creates a deprecation warning in Node.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -516,6 +833,103 @@ export class IndexedDbIndexManager implements IndexManager {
...
+    // non-overlapping ranges.
+    notInValues.sort((l, r) => compareByteArrays(l, r));
+
+    const notInRanges: IDBKeyRange[] = [];
+    for (const indexRange of indexRanges) {
+      // Use the existing bounds and interleave the notIn values. This means
+      // that we would split an existing range into multiple ranges that exclude
+      // the values from any notIn filter.
+
+      // The first index range starts with the lower bound and ends at the
```
wu-hui, line 860: I don't think I understand this well..what if `notInValues[0]` is not covered by the first range? What if it is < indexRange.lower?

schmidt-sebastian: That was indeed completely broken, and also did not work correctly on Android. I ported the Android fix https://github.com/firebase/firebase-android-sdk/pull/3454 and adopted it to work with the Web port.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -516,6 +833,126 @@ export class IndexedDbIndexManager implements IndexManager {
...
+    barriers: Uint8Array[]
+  ): IDBKeyRange[] {
+    const ranges: IDBKeyRange[] = [];
+
+    // The first index range starts with the lower bound and ends before the
+    // first barrier.
+    ranges.push(
+      IDBKeyRange.bound(
+        indexRange.lower,
+        this.generateNotInBound(indexRange.lower, barriers[0]),
```
wu-hui, line 891: What if `barriers[0]` is larger than `indexRange.upper`, is that not possible?

I think we could do something like:

```
// Assume barriers is sorted, it is right?
let runningLower = indexRange.lower
let runningOpen = indexRange.lowerOpen
for(const barrier: barriers) {
  if(barrier <= indexRange.lower) {continue;}
  if(barrier >= indexRange.upper) {break;}
  ranges.push([runningLower /*need to handle lowerOpen as well*/, barrier]);
  runningLower = barrier
  lowerOpen = true
}

// Handles last
ranges.push([runningLower, indexRange.upper])
```

You could even do handle this by taking `IDBKeyRange[]`, sort both `indexRanges` and `barriers` together..then process from left to right. But maybe let's not go there, it should be rare in real world scenarios.

schmidt-sebastian: I updated this code to work on browsers and Node. The browsers seem to do more validations, so I had to change some of the logic here. I also added a bunch of test code. Note that I myself don't fully grasp what is going on here anymore, but the tests are all passing and I think that I covered all corner cases.

schmidt-sebastian: Still broken

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -88,6 +105,15 @@ export class IndexedDbIndexManager implements IndexManager {
 
   private uid: string;
 
+  /**
+   * Maps from a target to its equivalent list of sub-targets. Each sub-target
+   * contains only one term from the target's disjunctive normal form (DNF).
+   */
+  private targetToDnfSubTargets = new ObjectMap<Target, Target[]>(
```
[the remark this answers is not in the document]
wu-hui: My point is maybe we can rename the term sub-target..but we can deal with this later.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -196,19 +222,231 @@ export class IndexedDbIndexManager implements IndexManager {
...
+  }
+
+  /**
+   * Constructs a key range query on `DbIndexEntry.store` that unions all
+   * bounds.
+   */
+  private generateIndexRanges(
+    indexId: number,
+    arrayValues: ProtoValue[] | null,
+    lowerBounds: Uint8Array[] | null,
```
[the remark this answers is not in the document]
wu-hui: I see.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -516,6 +849,77 @@ export class IndexedDbIndexManager implements IndexManager {
...
+    notInValues = notInValues
+      .sort((l, r) => indexEntryComparator(l, r))
+      .filter(
+        (el, i, values) => !i || indexEntryComparator(el, values[i - 1]) !== 0
+      );
+
+    const bounds: IndexEntry[] = [];
+    bounds.push(lower);
+    for (const notInValue of notInValues) {
+      const cpmToLower = indexEntryComparator(notInValue, lower);
```
wu-hui, line 876: typo

wu-hui (approved):

packages/firestore/src/local/indexeddb_index_manager.ts
schmidt-sebastian: Fixed

PR #6039, opened 2022-03-01
Add a last sent date to heartbeat storage

hsubox76:
Persist date of last sent heartbeat header, to prevent adding any more entries for that date. This is done by adding a `lastSentHeartbeatDate` to the heartbeat object stored in IndexedDB (and retained in app memory as `this._heartbeatsCache`).

### DIFF
--- a/packages/app/src/heartbeatService.ts
+++ b/packages/app/src/heartbeatService.ts
@@ -119,7 +119,8 @@ export class HeartbeatServiceImpl implements HeartbeatService {
    * Returns a base64 encoded string which can be attached to the heartbeat-specific header directly.
    * It also clears all heartbeats from memory as well as in IndexedDB.
    *
-   * NOTE: It will read heartbeats from the heartbeatsCache, instead of from indexedDB to reduce latency
+   * NOTE: Consuming product SDKs should not send the header if this method
+   * returns an empty string.
    */
   async getHeartbeatsHeader(): Promise<string> {
     if (this._heartbeatsCache === null) {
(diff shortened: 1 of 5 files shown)


egilmorez:
iiuc, this has a lot of internally-viewable comments with "//" but no 3P-visible comments that would get parsed into reference docs.  LMK if that's not true Christina, thanks!

packages/app/src/heartbeatService.ts
```diff
@@ -117,34 +119,41 @@ export class HeartbeatServiceImpl implements HeartbeatService {
...
   async getHeartbeatsHeader(): Promise<string> {
     if (this._heartbeatsCache === null) {
       await this._heartbeatsCachePromise;
     }
-    // If it's still null, it's been cleared and has not been repopulated.
-    if (this._heartbeatsCache === null) {
+    // If it's still null or the array is empty, there is no data to send.
+    if (
+      this._heartbeatsCache === null ||
+      this._heartbeatsCache.heartbeats.length === 0
```
allspain, line 132: is it possible for `this._heartbeatsCache.heartbeats` to be undefined while `this._heartbeatsCache` is not null.

Might be worth an additional check on `heartbeats`

allspain: Seeing below it's not super obvious how it's even possible

allspain (approved):

packages/app/src/heartbeatService.ts
hsubox76: Typescript should catch if there's ever an attempt to assign an object to this._heartbeatsCache that doesn't have a heartbeats property.

PR #6048, opened 2022-03-03
hasVisibleClients issue at chrome extension with manifest v3

embbnux:
hasVisibleClients can't work at chrome extension with manifest v3. 
Issue can be reproduced with [this project](https://github.com/embbnux/chrome-extension-manifest-v3-with-fcm/tree/firebase-issue-at-v3)

### DIFF
--- a/packages/messaging/src/listeners/sw-listeners.ts
+++ b/packages/messaging/src/listeners/sw-listeners.ts
@@ -210,12 +210,12 @@ async function getWindowClient(url: URL): Promise<WindowClient | null> {
 /**
  * @return If runtime is chrome extension manifest v3.
  */
- function isExtensionManifestV3(): boolean {
-   if (typeof self.chrome === 'undefined') {
-     return false;
-   }
+function isExtensionManifestV3(): boolean {
+  if (typeof self.chrome === 'undefined') {
+    return false;
+  }
   if (typeof self.chrome.runtime.getManifest === 'undefined') {
-      return false;
+    return false;
   }
   const manifest = self.chrome.runtime.getManifest();
   return manifest.manifest_version === 3;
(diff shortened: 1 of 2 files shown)


zwu52 (approved):
LGTM. thanks for the PR

packages/messaging/src/listeners/sw-listeners.ts
```diff
@@ -207,11 +207,29 @@ async function getWindowClient(url: URL): Promise<WindowClient | null> {
   return null;
 }
 
+/**
+ * @return If runtime is chrome extension manifest v3.
```
egilmorez, line 211: Should this be something more like, "If runtime is Chrome, use extension manifest v3?"

I feel like it needs a verb (especially if this is a public string parsed into reference docs).

egilmorez (commented):
Thanks for sending!

All looks good but I do have one suggestion on a comment string.

packages/messaging/src/listeners/sw-listeners.ts
zwu52: sg. updated.

egilmorez (approved):
LG, thanks!

zwu52:
hmm. Looks like `chrome` object isn't an object of the `ServiceWorkerGlobalScope`.  What's the error thrown when running chrome extension with manifest v3?

embbnux:
> hmm. Looks like `chrome` object isn't an object of the `ServiceWorkerGlobalScope`. What's the error thrown when running chrome extension with manifest v3?

`chrome` is an object of `ServiceWorkerGlobalScope` at extension manifest v3:

image

For this PR, it fix `hasVisibleClients` issue at chrome extension manifest v3. It always be false for chrome extension manifest v3.

PR #6068, opened 2022-03-14
Auth middleware (beforeAuthStateChanged)

hsubox76:
Create `beforeAuthStateChanged()` method per internal design doc [go/firebase-auth-middleware](https://goto.google.com/firebase-auth-middleware).

TODO in a separate PR: implement `onAbort()`.

### DIFF
--- a/common/api-review/auth.api.md
+++ b/common/api-review/auth.api.md
@@ -81,6 +81,7 @@ export function applyActionCode(auth: Auth, oobCode: string): Promise<void>;
 // @public
 export interface Auth {
     readonly app: FirebaseApp;
+    beforeAuthStateChanged(callback: (user: User | null) => void | Promise<void>): Unsubscribe;
     readonly config: Config;
     readonly currentUser: User | null;
     readonly emulatorConfig: EmulatorConfig | null;
--- a/packages/auth/src/core/auth/auth_impl.ts
+++ b/packages/auth/src/core/auth/auth_impl.ts
@@ -78,6 +78,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   private redirectPersistenceManager?: PersistenceUserManager;
   private authStateSubscription = new Subscription<User>(this);
   private idTokenSubscription = new Subscription<User>(this);
+  private beforeStateSubscription = new Subscription<User>(this);
   private redirectUser: UserInternal | null = null;
   private isProactiveRefreshEnabled = false;
 
@@ -366,6 +367,17 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
     return this.registerStateListener(
       this.authStateSubscription,
       nextOrObserver,
+      error,
+      completed
+    );
+  }
+
+  beforeAuthStateChanged(
+    next: NextFn<User | null>,
+  ): Unsubscribe {
+    return this.registerStateListener(
+      this.beforeStateSubscription,
+      next,
       error,
       completed
     );
--- a/packages/auth/src/core/auth/auth_impl.ts
+++ b/packages/auth/src/core/auth/auth_impl.ts
@@ -78,7 +78,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   private redirectPersistenceManager?: PersistenceUserManager;
   private authStateSubscription = new Subscription<User>(this);
   private idTokenSubscription = new Subscription<User>(this);
-  private beforeStateSubscription = new Subscription<User>(this);
+  private beforeStateQueue: Array<(user: User | null) => Promise<void>> = [];
   private redirectUser: UserInternal | null = null;
   private isProactiveRefreshEnabled = false;
 
@@ -325,6 +325,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
         AuthErrorCode.TENANT_ID_MISMATCH
       );
     }
+    await this._runBeforeStateCallbacks(user);
 
     return this.queue(async () => {
       await this.directlySetCurrentUser(user as UserInternal | null);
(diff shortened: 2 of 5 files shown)


packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -324,14 +325,26 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
...
+      for (const beforeStateCallback of this.beforeStateQueue) {
+        await beforeStateCallback(user);
+      }
+    } catch (e) {
+      throw this._errorFactory.create(AuthErrorCode.LOGIN_BLOCKED, { message: e.message });
+    }
+  }
+
   async signOut(): Promise<void> {
+    await this._runBeforeStateCallbacks(null);
```
sam-gc, line 347: This will run twice since it's also in _updateCurrentUser

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -324,14 +325,26 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
         AuthErrorCode.TENANT_ID_MISMATCH
       );
     }
+    await this._runBeforeStateCallbacks(user);
```
sam-gc, line 328: `_updateCurrentUser` is also called by the storage onevent callbacks, which means a persistence change in another tab could trigger a stop here and the tabs would get out of sync. I'm not sure what the right behavior should be

packages/auth/src/core/auth/auth_impl.ts
hsubox76: It seems like it needs to run before `_setRedirectUser()` so it can block it, so I added an optional arg to `_updateCurrentUser()` to skip running the callbacks inside it.

packages/auth/src/core/auth/auth_impl.ts
hsubox76: We discussed this, and we want the callbacks to not run if called by _onStorageEvent(), so I made use of the optional `skipBeforeStateCallbacks` argument I added to _updateCurrentUser when called by _onStorageEvent().

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -325,19 +327,38 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
...
+    }
+
     return this.queue(async () => {
       await this.directlySetCurrentUser(user as UserInternal | null);
       this.notifyAuthListeners();
     });
   }
 
+  async _runBeforeStateCallbacks(user: User | null): Promise<void> {
+    try {
```
sam-gc, line 341: This method should probably check first whether or not the user is actually changing (like `directlySetCurrentUser` does).

Right now with this code, if you call auth.signOut() twice in a row, it will call the middleware twice in a row

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -528,6 +575,7 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
    * should only be called from within a queued callback. This is necessary
    * because the queue shouldn't rely on another queued callback.
    */
+  // TODO: Find where this is called and see if we can run the middleware before it
```
sam-gc, line 578: This is only used during initialization and then in a few select cases within this file. I need to do some thinking about the appropriate places in the init phase to use middleware. Off the top of my head, the only place init should run middleware is when signing in a user from redirect (i.e. first page load after a `signInWithRedirect()` call)

packages/auth/src/core/auth/auth_impl.ts
hsubox76: Added a check.

packages/auth/src/core/auth/auth_impl.ts
hsubox76: Inserted it in what I think is the correct place in `initializeCurrentUser()`.

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -223,6 +225,10 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
     _assert(this._popupRedirectResolver, this, AuthErrorCode.ARGUMENT_ERROR);
     await this.getOrInitRedirectPersistenceManager();
 
+    // At this point in the flow, this is a redirect user. Run blocking
+    // middleware callbacks before setting the user.
+    await this._runBeforeStateCallbacks(storedUser);
```
sam-gc, line 230: I think here we should wrap it in a try/catch. Initialization swallows errors. Ideally we would surface the error in `getRedirectResult` but that will require some extra thinking

common/api-review/auth.api.md
```diff
@@ -81,6 +81,7 @@ export function applyActionCode(auth: Auth, oobCode: string): Promise<void>;
 // @public
 export interface Auth {
     readonly app: FirebaseApp;
+    beforeAuthStateChanged(callback: (user: User | null) => void | Promise<void>): Unsubscribe;
```
jamesdaniels, line 84: Did we want to tackle the second argument, the optional `onAbort` callback in this PR? 

That is just a `() => void` that is called useful if `beforeAuthStateChanged` is called multiple times and a subsequent callback rejects the state change. This gives developers a chance to walk back any side effects. 

Useful to me as I integrate with the web frameworks work, for authenticated server-context. onAbort would allow me the chance to destroy the just minted cookie, if the developer is using beforeAuthStateChanged too and rejected the sign in.

hsubox76:
Merging into `auth-middleware` feature branch where work will continue with ongoing PRs.

PR #6069, opened 2022-03-14
Add Firestore Debug Builds

schmidt-sebastian:


hsubox76 (commented):
If this is intended to be consumed by `import { getFirestore } from 'firebase/firestore/debug'` we need to add an entry in packages/firebase in the same pattern as lite here:
https://github.com/firebase/firebase-js-sdk/blob/1c37b5e965978d796c46ff6b9f52051cf6070751/packages/firebase/package.json#L82

And in packages/firestore/package.json here:
https://github.com/firebase/firebase-js-sdk/blob/1c37b5e965978d796c46ff6b9f52051cf6070751/packages/firestore/package.json#L57

And a folder in packages/firebase with an index.ts and package.json like lite here: https://github.com/firebase/firebase-js-sdk/tree/master/packages/firebase/firestore/lite

hsubox76 (commented):
Still need to create new `packages/firebase/firestore/debug` folder with an index.ts and package.json like lite here: https://github.com/firebase/firebase-js-sdk/tree/master/packages/firebase/firestore/lite

PR #5988, opened 2022-02-09
RemoteDocumentCache APIs for Indexing

schmidt-sebastian:
This changes the schema of the Remote Document Cache to rewrite the primary keys. The keys are now:

[ [prefix path ], collection group, read time, document ]

This allows for efficient scanning of indexed and non-indexed queries using `getAll()`. The PR also adds `getDocumentsFromCollectionGroup`, which means that all functionality required by Indexing is now available.

TODO:
- Benchmark the migration
- Re-write multi tab to use the collectionGroup+ReadTime index

PR #6073, opened 2022-03-16
Use collectionGroupReadTimeIndex for multi-tab synchronization

schmidt-sebastian:
This drops one index created by https://github.com/firebase/firebase-js-sdk/pull/5988 and changes the multi-tab synchronization to use the collection group read time index. This works pretty well for normal query execution, but needs some adjusting to bundle loading as we don't know the collection groups of the bundles. As such, this PR adds functionality to the bundle loader to persist the names of the affected collection groups in LocalStoragel

### DIFF
--- a/packages/firestore/test/unit/local/remote_document_cache.test.ts
+++ b/packages/firestore/test/unit/local/remote_document_cache.test.ts
@@ -154,7 +154,7 @@ describe('IndexedDbRemoteDocumentCache', () => {
     assertMatches([doc('a/2', 2, DOC_DATA), doc('a/3', 3, DOC_DATA)], results);
   });
 
-  it('cen get next documents from collection group with document key offset', async () => {
+  it('can get next documents from collection group with document key offset', async () => {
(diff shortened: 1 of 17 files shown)


packages/firestore/test/unit/local/remote_document_cache.test.ts
```diff
@@ -109,79 +106,76 @@ describe('IndexedDbRemoteDocumentCache', () => {
...
+    const results = await cache.getAllFromCollectionGroup(
+      'a',
+      IndexOffset.min(),
+      Number.MAX_SAFE_INTEGER
+    );
+    assertMatches([doc('a/1', 1, DOC_DATA), doc('a/2', 2, DOC_DATA)], results);
   });
 
-  it('can get changes', async () => {
+  it('cen get next documents from collection group with limit', async () => {
```
wu-hui, line 124: `can`, and below

wu-hui (approved):

PR #6075, opened 2022-03-16
Support descending queries

schmidt-sebastian:
This adds code to order all query results by document key, which is needed to return the correct result for limit queries.

Port of https://github.com/firebase/firebase-android-sdk/pull/3499

### DIFF
--- a/packages/firestore/src/local/indexeddb_index_manager.ts
+++ b/packages/firestore/src/local/indexeddb_index_manager.ts
@@ -295,8 +295,8 @@ export class IndexedDbIndexManager implements IndexManager {
                 .loadFirst(indexRange, target.limit)
                 .next(entries => {
                   entries.forEach(entry => {
-                    const documentKey = new DocumentKey(
-                      decodeResourcePath(entry.documentKey)
+                    const documentKey = DocumentKey.fromSegments(
+                      entry.documentKey
                     );
                     if (!existingKeys.has(documentKey)) {
                       existingKeys = existingKeys.add(documentKey);
@@ -737,7 +737,7 @@ export class IndexedDbIndexManager implements IndexManager {
       arrayValue: indexEntry.arrayValue,
       directionalValue: indexEntry.directionalValue,
       orderedDocumentKey: this.encodeDirectionalKey(fieldIndex, document.key),
-      documentKey: encodeResourcePath(document.key.path)
+      documentKey: document.key.path.toArray()
     });
   }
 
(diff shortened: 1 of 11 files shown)


packages/firestore/src/model/field_index.ts
```diff
@@ -79,6 +79,18 @@ export function fieldIndexGetDirectionalSegments(
   return fieldIndex.fields.filter(s => s.kind !== IndexKind.CONTAINS);
 }
 
+/**
+ * Returns the order of the document key component.
```
ehsannas, line 83: Returns the order of the document key component for the given index

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -906,15 +935,17 @@ export class IndexedDbIndexManager implements IndexManager {
             this.uid,
             bounds[i].arrayValue,
             bounds[i].directionalValue,
+            new Uint8Array(),
```
ehsannas, line 938: I'm guessing the encoding of an empty string is an empty byte array

packages/firestore/src/local/indexeddb_schema.ts
```diff
@@ -510,6 +514,11 @@ export interface DbIndexEntry {
   arrayValue: Uint8Array;
   /** The encoded directional value for equality and inequality filters. */
   directionalValue: Uint8Array;
+  /**
+   * The document key this entry points to. This entry is encoded by an ordered
+   * encoder to match the key order of the index.
+   */
+  orderedDocumentKey: Uint8Array;
```
ehsannas, line 521: Is it true that if the order was always ASC, then using `EncodedResourcePath` below would have sufficed? in order words: is the only reason we need this is because of DESC indexes?

ehsannas (commented):
There are some index manager test failures.

packages/firestore/src/local/indexeddb_schema.ts
schmidt-sebastian: FWIW, I changed the `documentKey` entry to no longer use `EncodedResourcePath` since we no longer rely on any ordering.

packages/firestore/src/model/field_index.ts
schmidt-sebastian: Done

ehsannas (commented):
Still seeing these two test failures:

  1) Target Bounds
       less than query:

  2) Target Bounds
       greater than or equals query:

schmidt-sebastian:
I fixed the test failures. Sorry for not catching them earlier. I think we should write better test helpers to reduce the repetitiveness of the bounds tests, but I can't get myself to do that on my current setup.

ehsannas (approved):

PR #6088, opened 2022-03-21
Exclude browser-only types from @firebase/util public types

hsubox76:
Use api-extractor's dtsRollup functionality to exclude browser-only native IndexedDB types from util's public typings. These typings are used in methods needed by other firebase packages for internal development, but end users don't need them.

- `common/api-review/util.api.md` is autogenerated by the `api-extractor` process. Since `util` isn't fully compliant yet, and we don't generate docs for it, it can be ignored for now.
- Warnings are disabled in `api-extractor.json` for util, since we don't care about any of the end products except the public typings file. We may fix it to be tsdoc compliant later for better auditing of util but it's low priority since there's no public facing docs or API.
- `docgen.ts` has been modified to prevent it generating devsite docs for util. It was also generating extraneous docs for `app-compat`, so that's been excluded too.


egilmorez (approved):
Looks like no public-facing strings here.  LGTM.

allspain (approved):
LGTM. Took a second to understand how api-extractor works, but this seems like the right level of fix for this.

PR #6090, opened 2022-03-22
Updated typings for database methods

### LINKED ISSUE #6071: rtdb: Reference#transaction return type

  * Operating System version: macOS
  * Browser version: N/A
  * Firebase SDK version: firebase-admin@10
  * Firebase Product: database

The `transaction` method on a reference is typed to return `Promise<any>` as opposed to `Promise<{ committed: boolean, snapshot: DataSnapshot }>`.

https://github.com/firebase/firebase-js-sdk/blob/a7f4a2eb6ed08596dffe75825bca1a2034bfcd2e/packages/database-types/index.d.ts#L140-L144

This is the same problem in the [docs](https://firebase.google.com/docs/reference/js/v8/firebase.database.Reference#transaction).

I'm happy to create a PR to update the above file, but not sure if that's the correct place to do this?

#### Steps to reproduce:

```js
const { getDatabase } = require('firebase-admin/database')

const result = await getDatabase().ref('blah').transaction(data => data)
```

Result is `any`

#### Relevant Code:

See above.

rhodgkins:
Fixes #6071. I've also update the return type for the `onComplete` handler that the methods also accept.

I've copied the comments for the `TransactionResult` interface from the corresponding `@firebase/database` package:

https://github.com/firebase/firebase-js-sdk/blob/14ccbadd034e1a702f7d91e70ec14856d445764f/packages/database/src/api/Transaction.ts#L42-L49

### DIFF
--- a/.changeset/eighty-coats-hide.md
+++ b/.changeset/eighty-coats-hide.md
@@ -1,0 +1,5 @@
+---
+'@firebase/database-types': patch
+---
+
+Updated typings for Reference methods
--- a/packages/database-compat/src/api/Reference.ts
+++ b/packages/database-compat/src/api/Reference.ts
@@ -601,7 +601,7 @@ export class Reference extends Query implements Compat<ModularReference> {
   set(
     newVal: unknown,
     onComplete?: (error: Error | null) => void
-  ): Promise<unknown> {
+  ): Promise<void> {
     validateArgCount('Reference.set', 1, 2, arguments.length);
     validateCallback('Reference.set', 'onComplete', onComplete, true);
     const result = set(this._delegate, newVal);
@@ -617,7 +617,7 @@ export class Reference extends Query implements Compat<ModularReference> {
   update(
     values: object,
     onComplete?: (a: Error | null) => void
-  ): Promise<unknown> {
+  ): Promise<void> {
     validateArgCount('Reference.update', 1, 2, arguments.length);
 
     if (Array.isArray(values)) {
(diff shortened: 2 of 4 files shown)


hsubox76 (commented):
Looks good, in addition to the small nit about `readonly` (see comments), can you also add a changeset? To do this, run `yarn changeset` in the root of the repo and follow the prompts (select `database-types` as it suggests, then skip the "major" and "minor" options to let it default to "patch", then write a brief change note like "Updated typings for Reference methods"

rhodgkins:
@hsubox76 thanks for the review, `readonly`s removed and change set added 👍

hsubox76:
It seems like the `readonly` has been removed from `firebase/compat/index.d.ts` but not from `database-types/index.d.ts`, is that the case or is it a Github refresh issue on my end?

rhodgkins:
> It seems like the `readonly` has been removed from `firebase/compat/index.d.ts` but not from `database-types/index.d.ts`, is that the case or is it a Github refresh issue on my end?

@hsubox76 ah sorry, totally missed that 🙈 


> `database-compat/src/api/Reference.ts` should be changed so the methods in conflict (update, remove, getPriority, etc) return `Promise<void>`

I'll have a go at changing those and see if that works :)

rhodgkins:
@hsubox76 I've updated that file but there's still CI failures but its in a file I've not touched (I think anyway) - [`ERROR in /home/runner/work/firebase-js-sdk/firebase-js-sdk/packages/database/src/api/Database.ts`](https://github.com/firebase/firebase-js-sdk/runs/5698213257?check_suite_focus=true#step:8:407), is something else needed?

hsubox76:
> @hsubox76 I've updated that file but there's still CI failures but its in a file I've not touched (I think anyway) - [`ERROR in /home/runner/work/firebase-js-sdk/firebase-js-sdk/packages/database/src/api/Database.ts`](https://github.com/firebase/firebase-js-sdk/runs/5698213257?check_suite_focus=true#step:8:407), is something else needed?

It was a flaky test. I re-ran it and it passed. Everything looks good, will merge!

rhodgkins:
Ah great, thanks 👍

PR #6122, opened 2022-04-06
Specify readonly as the default indexeddb transaction mode

### LINKED ISSUE #6121: "Error thrown when reading from storage" app/storage-get on IE11
### Environment

  * Operating System version: Windows 10
  * Browser version: IE11
  * Firebase SDK version: 9.6.10
  * Firebase Product: auth, storage, firestore

### Describe the problem

When loading an app using the 9.6.10 SDK in IE11, an unhandled error is immediately thrown: `Error thrown when reading from storage. Original error: InvalidAccessError. (app/storage-get)`

#### Steps to reproduce:

Load any application using the 9.6.10 SDK in IE11.

#### Relevant Code:

This happens immediately on initializing the application.

```typescript
initializeApp({
    ...myAppConfig 
});
```

tmrudick:
While the mode parameter is optional, the wrapper method will pass `undefined` into the native method if it is not passed into the wrapper. IE11 interprets this as a mode instead of ignoring it and throws an InvalidAccessError.

Specifying `readonly` as the default (which is the default if the parameter is omitted) prevents IE11 from crashing during initialization.

This PR fixes #6121.

### DIFF
--- a/.changeset/hungry-owls-hide.md
+++ b/.changeset/hungry-owls-hide.md
@@ -1,0 +1,5 @@
+---
+'@firebase/util': patch
+---
+
+Default indexeddb transaction mode to readonly for IE11 compatibility
--- a/packages/util/src/indexeddb.ts
+++ b/packages/util/src/indexeddb.ts
@@ -42,7 +42,7 @@ export class DBWrapper {
   }
   transaction(
     storeNames: string[] | string,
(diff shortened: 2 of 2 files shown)


hsubox76 (approved):
Thanks for the fix, will merge when required CI passes.

PR #6126, opened 2022-04-06
Fix websocketonly flag

maneesht:
Fixes an issue where the webSocketOnly field was incorrectly set to undefined when using a websocket database URL.

When initializing the RepoInfo instance, the `nodeAdmin` value was being passed in as the `webSocketOnly` value, leading webSocketOnly to be undefined even when using a websocket url.


.changeset/twenty-shoes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fix websocketonly field to be set to true when using websocket db url.
```
jsdt, line 5: Could you explain what the websocket url is in parens (ws or wss instead of https)? Then someone can figure out how to use it by reading the release notes.

jsdt (approved):

schmidt-sebastian (approved):

jsdt (approved):

hsubox76 (approved):

Issue #6110, opened 2022-03-31
IndexedDB exception on Firestore batch.commit with latest Firebase SDK 9.6.10


olivierkrener:
### [REQUIRED] Describe your environment

  * Operating System version: Windows 10, Ubuntu 20.04.3
  * Browser version: Brave V1.36.122, Chrome 99.0.4844.84
  * Firebase SDK version: 9.6.10
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

After upgrading from v9.6.9 to v9.6.10 with persistence enabled, some writes result in the following IndexedDB exception:

> Uncaught DataError: Failed to execute 'continue' on 'IDBCursor': The parameter is less than or equal to this cursor's
> DOMException: Failed to execute 'continue' on 'IDBCursor': The parameter is less than or equal to this cursor's position.
> Uncaught DOMException: Failed to execute 'continue' on 'IDBCursor': The parameter is less than or equal to this cursor's position.
> 
> @firebase/firestore: Firestore (9.6.10): AsyncQueue Failed to persist write: IndexedDbTransactionError: [code=unavailable]: IndexedDB transaction 'Locally write mutations' failed: AbortError: Uncaught exception in event handler.

And the writes are not propagated to the backend.

Rolling back to v9.6.9 solves the issue.

I would guess this issue is due to the modifications on the IndexedDB data type coming with v9.6.10 as mentioned in the release notes:

> Changed the format of some of the IndexedDB data stored by the Cloud Firestore SDK. This increases the performance of document lookups after an initial migration. If you do not want to migrate data, you can call clearIndexedDbPersistence() before invoking enableIndexedDbPersistence().

Note that the issue also occurs in incognito mode so it should not be due to any pre-existing data in IndexedDB or migration issue.

I cannot give a simple reproduction as the case when the write fails is a fairly complex batch operation with multiple writes (delete, set, update) and I have not tested thoroughly v9.6.10 to see if there are simpler cases leading to the same issue. But it is deterministic, in my app I can reproduce it 100% of the time, doing one specific action.

schmidt-sebastian:
@olivierkrener Thank you for reporting. I will try to figure out where this is going on.

schmidt-sebastian:
[...]

PR #6128, opened 2022-04-07
Fix the ordering of the index-based lookup in getAll(keys)

schmidt-sebastian:
This changes the ordering we user for multi-key lookups to match that of IndexedDB.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6110

### DIFF
--- a/packages/firestore/src/local/indexeddb_remote_document_cache.ts
+++ b/packages/firestore/src/local/indexeddb_remote_document_cache.ts
@@ -611,12 +611,29 @@ function dbCollectionGroupKey(
     /* document id */ path.length > 0 ? path[path.length - 1] : ''
   ];
 }
+
 /**
  * Comparator that compares document keys according to the primary key sorting
- * used by the `DbRemoteDocumentDocument` store (by collection path and then
- * document ID).
+ * used by the `DbRemoteDocumentDocument` store (by prefix path, collection id
+ * and then document ID).
+ *
+ * Visible for testing.
  */
-function dbKeyComparator(l: DocumentKey, r: DocumentKey): number {
-  const cmp = l.path.length - r.path.length;
-  return cmp !== 0 ? cmp : DocumentKey.comparator(l, r);
+export function dbKeyComparator(l: DocumentKey, r: DocumentKey): number {
+  const left = l.path.toArray();
+  const right = r.path.toArray();
+
+  let cmp = 0;
+  for (let i = 0; i < left.length - 2 && i < right.length - 2; ++i) {
+    cmp = primitiveComparator(left[i], right[i]);
+    if (cmp) {
+      return cmp;
+    }
+  }
+
+  return (
+    primitiveComparator(left.length, right.length) ||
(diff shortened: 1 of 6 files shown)


packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -611,12 +611,29 @@ function dbCollectionGroupKey(
...
 }
+
 /**
  * Comparator that compares document keys according to the primary key sorting
- * used by the `DbRemoteDocumentDocument` store (by collection path and then
- * document ID).
+ * used by the `DbRemoteDocumentDocument` store (by prefix path, collection id
+ * and then document ID).
+ *
+ * Visible for testing.
```
wu-hui, line 620: Have you been able to find a spec on IDB ordering? If yes, we can add it here.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
```diff
@@ -611,12 +611,29 @@ function dbCollectionGroupKey(
...
+
+  let cmp = 0;
+  for (let i = 0; i < left.length - 2 && i < right.length - 2; ++i) {
+    cmp = primitiveComparator(left[i], right[i]);
+    if (cmp) {
+      return cmp;
+    }
+  }
+
+  return (
```
wu-hui, line 634: `primitiveComparator` returns number, not boolean..how this works is not obvious to me..

Can we make it more explicit?

wu-hui (changes requested):

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I have not and my contact has left the IndexedDb team. Let me spend some more time on research here though.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: JS compares by "thruhiness" and any non-null number is considered true. We could have something like:

```
let cmp = compare()
if (cmp) return cmp;

cmp = compare()
if (cmp) return cmp;

cmp = compare()
if (cmp) return cmp;
```

That was my original code but it is pretty bloated. Let me know what you prefer.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
schmidt-sebastian: I found https://chromium.googlesource.com/chromium/blink/+/refs/heads/main/Source/modules/indexeddb/IDBKey.cpp#66

So we are doing the right thing. I will add the link once we decide on the second comment thread in the PR.

packages/firestore/src/local/indexeddb_remote_document_cache.ts
wu-hui: I prefer the bloated way..

PR #6131, opened 2022-04-08
Overlay migration

ehsannas:


packages/firestore/src/local/indexeddb_schema_converter.ts
```diff
@@ -240,9 +249,9 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
     }
 
     if (fromVersion < 14 && toVersion >= 14) {
-      p = p.next(() => {
-        createFieldIndex(db);
-      });
+      p = p
+        .next(() => createFieldIndex(db))
+        .next(() => this.runOverlayMigration(db, simpleDbTransaction));
```
schmidt-sebastian, line 254: Same comment as in other PR - you should run this migration as version 14 and move indexing to 15.

packages/firestore/src/local/indexeddb_schema_converter.ts
```diff
@@ -455,6 +464,95 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+                          allDocumentKeysForUser.add(key))
+                    );
+                });
+              })
+              .next(() => {
+                // NOTE: The index manager and the reference delegate are
+                // irrelevant for the purpose of recalculating and saving
+                // overlays. We can therefore simply use the memory
+                // implementation.
+                const memoryPersistence = new MemoryPersistence(
```
schmidt-sebastian, line 529: I am surprised this does not add more to our code size, but it make sense since we always load MemoryPersistence by default.

schmidt-sebastian (commented):
Hm, I thought I would not like this approach but it is actually cleaner that your other proposal. I will let you pick one and will do my next round of reviews when we have a winner + some tests :)

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1044,6 +1049,214 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+        localWriteTimeMs: 1337,
+        baseMutations: undefined,
+        mutations: [testWritePending]
+      },
+      {
+        userId: 'user1',
+        batchId: 5,
+        localWriteTimeMs: 1337,
+        baseMutations: undefined,
+        mutations: [testWritePending]
```
schmidt-sebastian, line 1113: Should this be user3 with no mutations?

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1044,6 +1049,214 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+        fields: {}
+      }
+    };
+    const testWriteBaz = {
+      update: {
+        name: 'projects/test-project/databases/(default)/documents/docs/baz',
+        fields: {}
+      }
+    };
+    const testWritePending = {
```
schmidt-sebastian, line 1073: Foo/Bar and Baz are also pending, aren't they?

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1044,6 +1049,214 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+                      indexKey,
+                      DbDocumentMutationPlaceholder
+                    );
+                  }
+                );
+              });
+            }
+          ).next(() =>
+            // Populate the mutation queues' metadata
+            PersistencePromise.waitFor([
```
schmidt-sebastian, line 1158: You can probably use the loop above to also add this data and then just simply call `.put({userId: user, lastAcknowledgedBatchId: -1, lastStreamToken: ''})` at each iteration. That way you only have one source of truth for your data.

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1044,6 +1049,214 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+          );
+        }
+      );
+    }).then(() =>
+      withDb(14, (db, version) => {
+        expect(version).to.be.equal(14);
+
+        return db.runTransaction(
+          this.test!.fullTitle(),
+          'readwrite',
```
schmidt-sebastian, line 1184: readonly

packages/firestore/test/unit/local/indexeddb_persistence.test.ts
```diff
@@ -1044,6 +1049,214 @@ describe('IndexedDbSchema: createOrUpgradeDb', () => {
...
+            });
+            p = p.next(() => {
+              const key = toDbDocumentOverlayKey(
+                'user1',
+                DocumentKey.fromPath('docs/foo')
+              );
+              return documentOverlayStore.get(key).next(overlay => {
+                expect(overlay).to.not.be.null;
+                expect(overlay!.overlayMutation).to.deep.equal(testWriteFoo);
+              });
```
schmidt-sebastian, line 1207: Can you add a helper?

schmidt-sebastian (approved):
LG. Please make sure to run at least one manual test for this upgrade.

PR #6132, opened 2022-04-08
idea: use standalone functions. wip.

ehsannas:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/firestore/src/local/indexeddb_mutation_queue.ts
+++ b/packages/firestore/src/local/indexeddb_mutation_queue.ts
@@ -352,50 +352,6 @@ export class IndexedDbMutationQueue implements MutationQueue {
       .next(() => results);
   }
 
-  getAllMutationBatchesAffectingDocumentKeys(
-    transaction: PersistenceTransaction,
-    documentKeys: SortedMap<DocumentKey, unknown>
-  ): PersistencePromise<MutationBatch[]> {
-    let uniqueBatchIDs = new SortedSet<BatchId>(primitiveComparator);
-
-    const promises: Array<PersistencePromise<void>> = [];
-    documentKeys.forEach(documentKey => {
-      const indexStart = newDbDocumentMutationPrefixForPath(
-        this.userId,
-        documentKey.path
-      );
-      const range = IDBKeyRange.lowerBound(indexStart);
-
-      const promise = documentMutationsStore(transaction).iterate(
-        { range },
-        (indexKey, _, control) => {
-          const [userID, encodedPath, batchID] = indexKey;
-
-          // Only consider rows matching exactly the specific key of
-          // interest. Note that because we order by path first, and we
-          // order terminators before path separators, we'll encounter all
-          // the index rows for documentKey contiguously. In particular, all
-          // the rows for documentKey will occur before any rows for
-          // documents nested in a subcollection beneath documentKey so we
-          // can stop as soon as we hit any such row.
-          const path = decodeResourcePath(encodedPath);
-          if (userID !== this.userId || !documentKey.path.isEqual(path)) {
-            control.done();
-            return;
-          }
-
-          uniqueBatchIDs = uniqueBatchIDs.add(batchID);
-        }
-      );
-
-      promises.push(promise);
-    });
-
-    return PersistencePromise.waitFor(promises).next(() =>
-      this.lookupMutationBatches(transaction, uniqueBatchIDs)
-    );
-  }
-
   getAllMutationBatchesAffectingQuery(
     transaction: PersistenceTransaction,
     query: Query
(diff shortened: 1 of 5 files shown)


ehsannas:
Odd: I get `IDBKeyRange is not defined` errors in this implementation

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -352,50 +352,6 @@ export class IndexedDbMutationQueue implements MutationQueue {
       .next(() => results);
   }
 
-  getAllMutationBatchesAffectingDocumentKeys(
```
schmidt-sebastian, line 355: I don't think you can do remove this method as it is used by the LocalDocumentsView and needs to be implemented by both the IndexedDbMutationQueue and the MemoryMutationQueue. You can have a standalone function in this file that you can also call from this method.

packages/firestore/src/local/indexeddb_mutation_queue.ts
```diff
@@ -660,3 +593,80 @@ function mutationQueuesStore(
     DbMutationQueueStore
   );
 }
+
+export function getAllMutationBatchesAffectingDocumentKeys(
```
schmidt-sebastian, line 597: This would need a method comment since it is expored.

packages/firestore/src/local/indexeddb_schema_converter.ts
```diff
@@ -107,16 +112,24 @@ import {
...
 } from './local_serializer';
-import { MemoryCollectionParentIndex } from './memory_index_manager';
+import {
+  MemoryCollectionParentIndex
+} from './memory_index_manager';
 import { PersistencePromise } from './persistence_promise';
 import { SimpleDbSchemaConverter, SimpleDbTransaction } from './simple_db';
 
+
+
```
schmidt-sebastian, line 132: Please revert.

packages/firestore/src/local/indexeddb_schema_converter.ts
```diff
@@ -240,9 +253,9 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
     }
 
     if (fromVersion < 14 && toVersion >= 14) {
-      p = p.next(() => {
-        createFieldIndex(db);
-      });
+      p = p
+        .next(() => createFieldIndex(db))
+        .next(() => this.runOverlayMigration(db, simpleDbTransaction));
```
schmidt-sebastian, line 258: Your overlay migration should run as version 14 and then indexing should be version 15. That way, we can enable overlays soon.

packages/firestore/src/local/indexeddb_schema_converter.ts
```diff
@@ -455,6 +468,79 @@ export class SchemaConverter implements SimpleDbSchemaConverter {
...
+
+    return queuesStore
+      .loadAll()
+      .next(queues => {
+        for (const queue of queues) {
+          userIds = userIds.add(queue.userId);
+        }
+      })
+      .next(() => {
+        userIds.forEach(userId => {
```
schmidt-sebastian, line 496: We don't really need the intermediate state (the userIds set). We can directly run the migration in the loop at 491.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -51,7 +51,8 @@ import { SortedMap } from '../util/sorted_map';
 
 import { DocumentOverlayCache } from './document_overlay_cache';
 import { IndexManager } from './index_manager';
-import { MutationQueue } from './mutation_queue';
+import { getAllMutationBatchesAffectingDocumentKeys } from './indexeddb_mutation_queue';
```
schmidt-sebastian, line 54: As stated above, we should not import from IndexedDB or Memory persistence here. We should only use their shared APIs.

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -431,3 +375,94 @@ export class LocalDocumentsView {
...
+ * Recalculates overlays by reading the documents from remote document cache
+ * first, and saves them after they are calculated.
+ */
+export function recalculateAndSaveOverlaysForDocumentKeys(
+  transaction: PersistenceTransaction,
+  documentKeys: DocumentKeySet,
+  remoteDocumentCache: RemoteDocumentCache,
+  documentOverlayCache: DocumentOverlayCache,
+  serializer: LocalSerializer,
+  userId: string
```
schmidt-sebastian, line 455: FWIW it is worth, I think we should consider re-ordering the arguments in the methods. I find it easier to read if we first pass the state that is shared (all caches) and then pass the call specific state. To me, this is a better order:

  remoteDocumentCache: RemoteDocumentCache,
  documentOverlayCache: DocumentOverlayCache,
  serializer: LocalSerializer,
  userId: string
  documentKeys: DocumentKeySet,

Obviously, this is just a gut feeling which can be ignored (especially since I can't actually explain the reasoning well)

schmidt-sebastian (commented):
I haven't looked at the other PR yet, but this PR is roughly what I expected this to look like.

PR #6136, opened 2022-04-11
Update App Check to v1 from v1beta

dwyfrequency:
For both reCAPTCHA v3 and reCAPTCHA Enterprise:

1. Updated URL to exchange for an App Check token
2. Updated the response JSON object containing the FAC token

### DIFF
--- a/.changeset/new-gorillas-film.md
+++ b/.changeset/new-gorillas-film.md
@@ -2,4 +2,4 @@
 '@firebase/app-check': patch
 ---
 
-Update App Check to v1 from v1beta for both reCAPTCHA v3 and reCAPTCHA Enterprise
+Update App Check to use v1 endpoint instead of v1beta endpoint for both reCAPTCHA v3 and reCAPTCHA Enterprise
--- a/.changeset/new-gorillas-film.md
+++ b/.changeset/new-gorillas-film.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app-check': patch
+---
+
+Update App Check to v1 from v1beta for both reCAPTCHA v3 and reCAPTCHA Enterprise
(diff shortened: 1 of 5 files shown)


hsubox76 (commented):
Looks good, last step is to add a changeset. `yarn changeset`, then follow the prompts. It should automatically choose the `@firebase/app-check` package, just confirm. Say no to "major" and "minor" and it will accept it as a "patch" bump. Then just write a short description of the change (for the changelog, doesn't have to be too deep).

dwyfrequency:
Added the changeset for this PR

.changeset/new-gorillas-film.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/app-check': patch
+---
+
+Update App Check to v1 from v1beta for both reCAPTCHA v3 and reCAPTCHA Enterprise
```
hsubox76, line 5: Can you change to "Update App Check to use v1 endpoint instead of v1beta endpoint..."? So it doesn't sound like we're updating the version of the App Check SDK itself.

dwyfrequency: Yes, I just updated the wording

hckhanh:
Hi @dwyfrequency, I receive the error after upgrade to Firebase 9.7.0

```json
{
  "error": {
    "code": 404,
    "message": "Method not found.",
    "status": "NOT_FOUND"
  }
}
```

hckhanh:
Hi @dwyfrequency, I receive the error after upgrade to Firebase 9.7.0

https://content-firebaseappcheck.googleapis.com/v1/projects/<project_id>/apps/<app_id>:exchangeRecaptchaV3Token?key=<api_key>

```json
{
  "error": {
    "code": 404,
    "message": "Method not found.",
    "status": "NOT_FOUND"
  }
}
```

PR #6143, opened 2022-04-13
Support plumbing for partial index execution

schmidt-sebastian:
Port of https://github.com/firebase/firebase-android-sdk/pull/3583. The Query engine part of the PR is not ported as we do not yet have the Query Engine changes that require this change.

### DIFF
--- a/packages/firestore/src/local/indexeddb_index_manager.ts
+++ b/packages/firestore/src/local/indexeddb_index_manager.ts
@@ -27,6 +27,7 @@ import {
   targetGetArrayValues,
   targetGetLowerBound,
   targetGetNotInValues,
+  targetGetSegmentCount,
   targetGetUpperBound
 } from '../core/target';
 import { FirestoreIndexValueWriter } from '../index/firestore_index_value_writer';
@@ -59,7 +60,7 @@ import {
   decodeResourcePath,
   encodeResourcePath
 } from './encoded_resource_path';
-import { IndexManager } from './index_manager';
+import { IndexManager, IndexType } from './index_manager';
 import {
   DbCollectionParent,
   DbIndexConfiguration,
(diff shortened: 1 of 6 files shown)


packages/firestore/test/unit/local/index_manager.test.ts
```diff
@@ -1416,6 +1419,141 @@ describe('IndexedDbIndexManager', async () => {
...
+      queryWithAddedFilter(query('coll'), filter('a', '==', 1)),
+      filter('b', '==', 1)
+    );
+    await validateIsPartialIndex(query5);
+
+    const query6 = queryWithAddedOrderBy(
+      queryWithAddedFilter(query('coll'), filter('a', '==', 1)),
+      orderBy('b')
+    );
+    await validateIsPartialIndex(query6);
```
ehsannas, line 1466: looks like this is failing on CI? can't tell why by skimming the PR

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -449,13 +450,33 @@ export class IndexedDbIndexManager implements IndexManager {
...
-      matchingIndexes.sort((l, r) => r.fields.length - l.fields.length);
-      return matchingIndexes.length > 0 ? matchingIndexes[0] : null;
+  getIndexType(
+    transaction: PersistenceTransaction,
+    target: Target
+  ): PersistencePromise<IndexType> {
+    // TODO(orqueries): We should look at the subtargets here
+    return this.getFieldIndex(transaction, target).next(index => {
+      if (!index) {
+        return IndexType.NONE as IndexType;
```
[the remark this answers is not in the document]
schmidt-sebastian: Yes, since IndexType.NONE is a const enum, the actual type is a number here if we don't cast.

packages/firestore/test/unit/local/index_manager.test.ts
schmidt-sebastian: :/ My local dev setup isn't working so I was hoping the test would just pass. Will investigate.

schmidt-sebastian: Fixed

ehsannas (approved):

PR #6147, opened 2022-04-13
Update babel and changeset packages.

dwyfrequency:
Updating babel and changeset packages as mentioned in [pull request 5566](https://github.com/firebase/firebase-js-sdk/pull/5566/files#) for all non-major dependencies.


hsubox76 (approved):

package.json
```diff
@@ -62,14 +62,14 @@
     "repo-scripts/*"
   ],
   "devDependencies": {
-    "@babel/core": "7.16.0",
-    "@babel/plugin-transform-modules-commonjs": "7.16.0",
-    "@babel/preset-env": "7.16.0",
-    "@babel/register": "7.16.0",
-    "@babel/preset-typescript": "7.16.0",
+    "@babel/core": "7.17.9",
```
hsubox76, line 65: Should the babel updates be here? I noticed they're also in a separate PR: https://github.com/firebase/firebase-js-sdk/pull/6146

hsubox76: Do you mean that the `@changesets` bumps don't work without the `@babel` bumps?

dwyfrequency: No, I mean that I was just adding each dependency back one by one. If there were no failures, the last PR would include all of the dependency updates. @changesets is working with @babel

dwyfrequency: It was largely just a way that I could work on multiple dependency updates while waiting for the prior ones' tests to run.

hsubox76: Ok, as long as it's working, I think it's ok to do PRs in any grouping, but I don't think there should be overlap between PRs, so I think either combine this with https://github.com/firebase/firebase-js-sdk/pull/6146 into one PR, or remove the `@babel` upgrades from this PR so it doesn't overlap with https://github.com/firebase/firebase-js-sdk/pull/6146.

dwyfrequency: Closed https://github.com/firebase/firebase-js-sdk/pull/6146 and updated title & description to include `@babel` and `@changesets` updates

hsubox76: I think there's 2 deps in https://github.com/firebase/firebase-js-sdk/pull/6146 that aren't in this one:
- "babel-plugin-transform-import-meta": "2.1.1",
- "babel-loader": "8.2.4",

dwyfrequency: Good catch! Updated those deps

dwyfrequency: Should be good now

hsubox76 (approved):

PR #6152, opened 2022-04-14
Various improvements from bug bash

renkelvin:
1. Store recaptcha config per tenant (it was global for all tenants)
2. Load recaptcha scripts when recaptcha config is enabled L359-L362 of `auth_impl.ts` (so the risk score is more accurate)
3. Add set recaptcha config to the test app

### DIFF
--- a/packages/auth/src/core/auth/auth_impl.ts
+++ b/packages/auth/src/core/auth/auth_impl.ts
@@ -91,7 +91,8 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   _popupRedirectResolver: PopupRedirectResolverInternal | null = null;
   _errorFactory: ErrorFactory<AuthErrorCode, AuthErrorParams> =
     _DEFAULT_AUTH_ERROR_FACTORY;
-  _recaptchaConfig: RecaptchaConfig | null = null;
+  _agentRecaptchaConfig: RecaptchaConfig | null = null;
+  _tenantRecaptchaConfigs: Record<string, RecaptchaConfig> = {};
   readonly name: string;
 
   // Tracks the last notified UID for state change listeners to prevent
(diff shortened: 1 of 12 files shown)


packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -349,7 +351,23 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
   }
 
   setRecaptchaConfig(config: RecaptchaConfig): void {
-    this._recaptchaConfig = config;
+    if (this.tenantId == null) {
+      this._agentRecaptchaConfig = config;
+    } else {
+      this._tenantRecaptchaConfigs[this.tenantId] = config;
+    }
+    if (config.emailPasswordEnabled === true) {
```
sam-gc, line 359: nit: `=== true` is redundant

packages/auth/src/core/auth/auth_impl.ts
```diff
@@ -349,7 +351,23 @@ export class AuthImpl implements AuthInternal, _FirebaseService {
...
+      const verifier = new RecaptchaEnterpriseVerifier(this);
+      void verifier.verify();
```
sam-gc, line 361: Is it necessary to void out the return value here for the linter? (I can't remember if that's one of the lint checks we have setup)

packages/auth/src/platform_browser/recaptcha/recaptcha_enterprise_verifier.ts
```diff
@@ -75,8 +75,8 @@ export class RecaptchaEnterpriseVerifier {
         if (auth.tenantId == null && RecaptchaEnterpriseVerifier.agentSiteKey != null) {
           return RecaptchaEnterpriseVerifier.agentSiteKey;
         }
-        if (auth.tenantId != null && RecaptchaEnterpriseVerifier.agentSiteKey !== undefined) {
-          return RecaptchaEnterpriseVerifier.siteKeys[auth.tenantId];
+        if (auth.tenantId != null && RecaptchaEnterpriseVerifier.tenantSiteKeys[auth.tenantId] !== undefined) {
```
sam-gc, line 78: Run `yarn format` from the main package page (that will automatically wrap this line)

sam-gc (approved):

sam-gc (commented):
Could you also add a couple tests that switch the auth tenant ID and ensure that the value of `_getRecaptchaConfig()` changes appropriately?

packages/auth/src/core/auth/auth_impl.ts
renkelvin: Yeah, otherwise it will complain not handling the response.

renkelvin:
> Could you also add a couple tests that switch the auth tenant ID and ensure that the value of `_getRecaptchaConfig()` changes appropriately?

Sure, good point, done.

PR #6153, opened 2022-04-14
Port QueryEngine

schmidt-sebastian:
This ports the query engine changes for indexed query execution from Android.

### DIFF
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -19,7 +19,6 @@ import { compareDocumentsByField, Document } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { FieldPath, ResourcePath } from '../model/path';
 import { debugAssert, debugCast, fail } from '../util/assert';
-import { isNullOrUndefined } from '../util/types';
 
 import {
   Bound,
@@ -165,14 +164,6 @@ export function queryMatchesAllDocuments(query: Query): boolean {
       (query.explicitOrderBy.length === 1 &&
         query.explicitOrderBy[0].field.isKeyField()))
   );
-}
-
-export function hasLimitToFirst(query: Query): boolean {
-  return !isNullOrUndefined(query.limit) && query.limitType === LimitType.First;
-}
-
-export function hasLimitToLast(query: Query): boolean {
-  return !isNullOrUndefined(query.limit) && query.limitType === LimitType.Last;
 }
 
 export function getFirstOrderByField(query: Query): FieldPath | null {
--- a/packages/firestore/src/core/query.ts
+++ b/packages/firestore/src/core/query.ts
@@ -155,7 +155,7 @@ export function asCollectionQueryAtPath(
  * Returns true if this query does not specify any query constraints that
  * could remove results.
  */
-export function matchesAllDocuments(query: Query): boolean {
+export function queryMatchesAllDocuments(query: Query): boolean {
   return (
     query.filters.length === 0 &&
     query.limit === null &&
@@ -393,7 +393,7 @@ export function queryWithAddedOrderBy(query: Query, orderBy: OrderBy): Query {
 
 export function queryWithLimit(
   query: Query,
-  limit: number,
+  limit: number | null,
   limitType: LimitType
 ): Query {
   return new QueryImpl(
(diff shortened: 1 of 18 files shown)


packages/firestore/src/core/query.ts
```diff
@@ -155,7 +155,7 @@ export function asCollectionQueryAtPath(
  * Returns true if this query does not specify any query constraints that
  * could remove results.
  */
-export function matchesAllDocuments(query: Query): boolean {
+export function queryMatchesAllDocuments(query: Query): boolean {
```
schmidt-sebastian, line 158: Changed to match naming convection: Classs+MethodName.

packages/firestore/test/unit/local/query_engine.test.ts
```diff
@@ -307,8 +387,9 @@ describe('QueryEngine', () => {
 
     // Add a query mapping for a document that matches, but that sorts below
     // another document due to a pending write.
-    await addDocument(PENDING_MATCHING_DOC_A);
-    await persistQueryMapping(PENDING_MATCHING_DOC_A.key);
+    await addDocument(MATCHING_DOC_A);
+    await addMutation(patchMutation('coll/a', { order: 1 }));
```
schmidt-sebastian, line 391: These changes are to support IndexedDb persistence. We cannot store a document with mutations in indexeddb so we instead have to store a document and its mutation.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -972,6 +972,16 @@ export class IndexedDbIndexManager implements IndexManager {
     }
     return ranges;
   }
+
+  getMinOffset(
+    transaction: PersistenceTransaction,
+    target: Target
+  ): PersistencePromise<IndexOffset> {
+    // TODO(orqueries): Get the minimum offset for all subqueries
+    return this.getFieldIndex(transaction, target).next(index =>
```
wu-hui, line 981: So this is required by or query, and left to implement when we port or queries? Just to clarify.

packages/firestore/src/local/query_engine.ts
```diff
@@ -74,15 +110,122 @@ export class QueryEngine {
...
+    query: Query
+  ): PersistencePromise<DocumentMap | null> {
+    if (!INDEXING_ENABLED) {
+      return PersistencePromise.resolve<DocumentMap | null>(null);
+    }
+
+    if (queryMatchesAllDocuments(query)) {
+      // Don't use indexes for queries that can be executed by scanning the
+      // collection.
+      return PersistencePromise.resolve<DocumentMap | null>(null);
```
wu-hui, line 146: // No need to use indexes if the query requires full collection scan anyways.

packages/firestore/src/local/query_engine.ts
```diff
@@ -74,15 +110,122 @@ export class QueryEngine {
...
+                    if (
+                      (hasLimitToFirst(query) || hasLimitToLast(query)) &&
+                      this.needsRefill(
+                        query.limitType,
+                        previousResults,
+                        sortedKeys,
+                        offset.readTime
+                      )
+                    ) {
+                      return PersistencePromise.resolve<DocumentMap | null>(
```
wu-hui, line 198: Can we add a comment to explain why we are returning null here?

packages/firestore/test/unit/local/query_engine.test.ts
```diff
@@ -85,11 +106,47 @@ class TestLocalDocumentsView extends LocalDocumentsView {
   }
 }
 
-describe('QueryEngine', () => {
+describe('MemoryQueryManager', async () => {
```
wu-hui, line 109: QueryEngine ?

Same below.

packages/firestore/test/unit/local/query_engine.test.ts
```diff
@@ -403,7 +485,85 @@ describe('QueryEngine', () => {
...
+      3,
+      LimitType.First
+    );
+    const results = await expectOptimizedCollectionQuery(() =>
+      runQuery(q, SnapshotVersion.min())
+    );
+
+    verifyResult(results, [doc2]);
+  });
+}
```
wu-hui, line 566: We should also add a test for refilling, unless there is already one. It's late for me so I did not bother to check.

packages/firestore/src/local/query_engine.ts
wu-hui: Thanks!

wu-hui (approved):

PR #6061, opened 2022-03-09
Add a stopgap idb replacement

hsubox76:
Replace `idb` dependency with a temporary set of methods in `@firebase/util`. This is to address an increasing amount of mostly Node ESM module related problems caused by depending on an old version of `idb`. Upgrading to idb 7+ may also fix many of these problems but would require dropping IE support.

In the interest of reducing the amount of surface area for error, I have only filled in the minimum idb methods and properties currently being used by our consuming libraries. We are planning on extracting a more fully featured idb promise wrapper library from Firestore in the near future.

Although these methods won't work in Node, I am exporting them from the `@firebase/util` Node entry point also, otherwise if an app compiles in Node mode (like an SSR app), it'll break on build when the packages try to import these methods. Errors can be avoided by using isSupported() to prevent initialization of relevant products.

PR #6154, opened 2022-04-14
Use idb 7 for products using shared idb library

hsubox76:
Context: Until recently, we were using the external `idb` dependency for app, installations, and messaging. (This also effectively made it a dependency of analytics, performance, and remote-config, as they depend on the installations package.) However, because we were required to support IE, we had to pin the `idb` version to 3.0.2, as subsequent versions of `idb` no longer supported IE. Since `idb` is currently on version 7+, this has caused a number of issues, most recently with lack of named exports causing problems with Angular.

We made a temporary fix to replace `idb` entirely with our own stopgap code in https://github.com/firebase/firebase-js-sdk/pull/6061

Since we are planning to drop IE support at I/O, we can now put the `idb` dependency back and bump it to 7+, which should have named exports and the Node `exports` field and support ESM.


packages/installations/src/helpers/idb-manager.ts
```diff
@@ -25,18 +25,27 @@ const DATABASE_NAME = 'firebase-installations-database';
 const DATABASE_VERSION = 1;
 const OBJECT_STORE_NAME = 'firebase-installations-store';
 
-let dbPromise: Promise<DBWrapper> | null = null;
-function getDbPromise(): Promise<DBWrapper> {
+interface InstallationsDB extends DBSchema {
+  'firebase-installations-store': {
+    key: string;
+    value: InstallationEntry | undefined;
```
dwyfrequency, line 31: Do we prefer this style of optionals instead of `value?: InstallationEntry;`?

dwyfrequency (approved):
LGTM

packages/installations/src/helpers/idb-manager.ts
hsubox76: It would be better to do `value?: InstallationEntry` but unfortunately we're extending a type from the external `idb` library and that type doesn't allow `value` to be an optional property (but does allow `| undefined`):

https://github.com/jakearchibald/idb/blob/231942c8a6b425e799e22885afb88751df4432ca/src/entry.ts#L124

edi:
image

9.8.0 also introduces this issue (not present in 9.7.0) in React Native builds.

Regards

kristof-kovacs:
I am finding the same error as @edi

edi:
Someone opened the same issue here.

https://github.com/firebase/firebase-js-sdk/issues/6253

PR #6162, opened 2022-04-16
Add httpsCallableFromURL

inlined:
Sorry for a late-coming urgent change; AFAICT nobody on the SDK team could staff these changes, so I'm trying to write the change for the top 3 SDKs before I/O freeze.

For context, see [go/cf3v2-crash-landing](http://go/cf3v2-crash-landing).
For API approval, see [go/callable-functions-custom-urls](http://go/callable-functions-custom-urls)


.changeset/new-bugs-think.md
```diff
@@ -0,0 +1,7 @@
+---
+"@firebase/functions-compat": patch
+"@firebase/functions": patch
+"firebase-size-analysis": patch
```
hsubox76, line 4: `firebase-size-analysis` can be removed, it's in the ignored changesets list. Actually, the change to that file seems unneeded (see below) so also for that reason.

e2e/sample-apps/modular.js
```diff
@@ -118,11 +118,28 @@ async function authLogout(app) {
  */
 async function callFunctions(app) {
   console.log('[FUNCTIONS] start');
-  const functions = getFunctions(app);
-  const callTest = httpsCallable(functions, 'callTest');
+  let functions = getFunctions(app);
```
hsubox76, line 121: Is `functions` reassigned?

packages/functions/src/api.ts
```diff
@@ -90,3 +91,23 @@ export function httpsCallable<RequestData = unknown, ResponseData = unknown>(
     options
   );
 }
+
+/**
+ * Returns a reference to the callable HTTPS trigger with the given name.
```
hsubox76, line 96: "given name" => "specified URL"?

repo-scripts/size-analysis/bundle-definitions/functions.json
```diff
@@ -22,11 +22,12 @@
             "path": "functions",
             "imports": [
               "getFunctions",
-              "httpsCallable"
+              "httpsCallable",
+              "httpsCallableFromURL"
```
hsubox76, line 26: Doesn't need to be added here, this represents a typical bundle of imports for a use case and it's not a common case that a developer would use httpsCallable and httpsCallableFromURL in the same app. In any case, since httpsCallable wraps the core code of httpsCallableFromURL, the bundle size should be about the same.

hsubox76 (commented):
Thanks for the PR, just a few small changes.

repo-scripts/size-analysis/bundle-definitions/functions.json
```diff
@@ -29,4 +29,4 @@
       }
     ]
   }
-]
+]
```
hsubox76, line 32: It still thinks there's a whitespace change.

e2e/sample-apps/modular.js
```diff
@@ -122,7 +122,24 @@ async function callFunctions(app) {
   const callTest = httpsCallable(functions, 'callTest');
```
hsubox76, line 122: `functions` isn't reassigned but it looks like `callTest` is, and does still need the `let`.

repo-scripts/size-analysis/bundle-definitions/functions.json
inlined: Dunno why my formatter keeps on insisting on this. Edited in vim to avoid stripping the newline.

e2e/sample-apps/modular.js
inlined: lol. duh

repo-scripts/size-analysis/bundle-definitions/functions.json
```diff
@@ -30,3 +30,4 @@
     ]
   }
 ]
+
```
hsubox76, line 33: This file still seems to show up in the diff, maybe just `git checkout master repo-scripts/size-analysis/bundle-definitions/functions.json`?

inlined: SGTM. Easy enough approach.

hsubox76 (approved):
LG, thanks!

PR #6166, opened 2022-04-19
Fix the cursor expanding result set bug

wu-hui:


packages/firestore/test/unit/local/index_manager.test.ts
```diff
@@ -886,6 +886,23 @@ describe('IndexedDbIndexManager', async () => {
...
+    const testingQuery = queryWithStartAt(
+      queryWithAddedOrderBy(
+        queryWithAddedFilter(query('coll'), filter('c', '>', 2)),
+        orderBy('c', 'asc')
+      ),
+      bound([2], /* inclusive= */ true)
+    );
+    await verifyResults(testingQuery, 'coll/val1');
+  });
+
```
ehsannas, line 905: can you also add another test (or just add to the existing test) for the other direction?  e.g. `c < 2 order by c DESC` with bound 2 inclusive.

wu-hui: And that uncovers one more bug.

ehsannas (approved):

Issue #6134, opened 2022-04-11
Firestore limitToLast with endBefore query contains 1 less document


xxdd13:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 12.3
  * Browser version: Version 100.0.4896.75 (Official Build) (arm64) 
  * Firebase SDK version: 9.6.10
  * Firebase Product: firestore

### [REQUIRED] Describe the problem
When using endBefore(cursor) and limitToLast(n), the snapshot contains n-1 documents

#### Steps to reproduce:

1. Add some data to a collection with timestamp
2. query orderBy timestamp
3. load next page
4. load prev page with `endBefore` and  `limitToLast`
5. snapshot contains 1 less doc

#### Relevant Code:

```javascript

const app = initializeApp({
  //
});
const db = getFirestore(app);
const ref = collection(db, 'tests');

// add test data
const batch = writeBatch(db);
[
  '2022-04-01',
  '2022-04-02',
  '2022-04-03',
  '2022-04-04',
  '2022-04-05',
  '2022-04-06',
].forEach((date) => {
  batch.set(doc(ref, date), {
    timestamp: new Date(date),
    date,
  });
});
await batch.commit();

const page1 = await getDocs(query(ref, orderBy('timestamp'), limit(2))).then(
  (snap) => {
    console.log(
      'page 1',
      snap.docs.map((doc) => doc.data().date)
    );
    return snap.docs;
  }
);

// get next page
const page2 = await getDocs(
  query(
    ref,
    orderBy('timestamp'),
    limit(2),
    startAfter(page1[page1.length - 1])
  )
).then((snap) => {
  console.log(
    'page 2',
    snap.docs.map((doc) => doc.data().date)
  );
  return snap.docs;
});

// get prev page
const shouldBePage1 = await getDocs(
  query(ref, orderBy('timestamp'), limitToLast(2), endBefore(page2[0]))
).then((snap) => {
  console.log(
    'should be page 1',
    snap.docs.map((doc) => doc.data().date)
  );
  return snap.docs;
});
/*
output
page 1          ['2022-04-01', '2022-04-02', '2022-04-03']
page 2          ['2022-04-04', '2022-04-05', '2022-04-06']
should be page1 ['2022-04-02', '2022-04-03']

*/

```

if I use the regular limit, the size of snapshot is correct
```javascript
query(ref, orderBy('timestamp'), limit(2), endBefore(page2[0]))
```

[...]

PR #6168, opened 2022-04-20
Fix a bug with limitToLast and cursors

wu-hui:
Fix: https://github.com/firebase/firebase-js-sdk/issues/6134


cherylEnkidu (approved):
Thank you for your fix! Could you please add one more test to check 
`query(collection, orderBy('sort'), startAfter(0), limitToLast(3))` ?

Issue #6087, opened 2022-03-20
Error with Firestore db persistence on Vue3


Shanzid01:
### Describe your environment

  * Operating System version: MacOS 12.2.1 (M1 chip)
  * Browser version: Chrome 99.0.4844.74
  * Firebase SDK version: 9.6.4
  * Firebase Product: firestore
  * NodeJS: v16.14.0

### Problem
Adding firestore persistence to Vue3 web app (followed instructions [here](https://firebase.google.com/docs/firestore/manage-data/enable-offline#configure_offline_persistence)) results in console errors when setting up snapshot listeners in firestore:

```
[2022-03-19T13:02:21.945Z]  @firebase/firestore: Firestore (9.6.4): INTERNAL UNHANDLED ERROR:  Error: Failed to execute 'put' on 'IDBObjectStore': #<Object> could not be cloned.
    at ti.put (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:5969:24)
    at Proxy.ae (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:7166:22)
    at Proxy.updateTargetData (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:7133:21)
    at eval (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:9146:30)
    at Map.forEach (<anonymous>)
    at eval (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:9105:25)
    at eval (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:8508:20)
    at eval (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:5659:49)
    at zs.wrapUserFunction (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:5652:23)
    at zs.wrapSuccess (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:5659:25)

index.esm2017.js?0829:5883 Uncaught (in promise) DOMException: Failed to execute 'put' on 'IDBObjectStore': #<Object> could not be cloned.
    at ti.put (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:5969:24)
    at Proxy.ae (webpack-internal:///./node_modules/@firebase/firestore/dist/index.esm2017.js:7166:22)
[...]

PR #6173, opened 2022-04-22
[patch] Pre-process variables before adding into indexedDB

Shanzid01:
Serialize and deserialize key-value pairs before adding them into indexedDB. This prevents invalid values (like functions) from being added into indexedDB, and fixes [Issue 6087](https://github.com/firebase/firebase-js-sdk/issues/6087).

You can test these changes live [here](https://firestore-db-persistence-error.pages.dev/) [[see code](https://github.com/Shanzid01/firestore-db-persistence-error)].

### DIFF
--- a/packages/firestore/src/local/simple_db.ts
+++ b/packages/firestore/src/local/simple_db.ts
@@ -589,6 +589,25 @@ export class SimpleDbStore<
     keyOrValue: KeyType | ValueType,
     value?: ValueType
   ): PersistencePromise<void> {
+    try {
+      keyOrValue =
+        keyOrValue != null
+          ? JSON.parse(JSON.stringify(keyOrValue))
+          : keyOrValue;
+    } catch {
+      console.warn(
+        'Could not process value before adding into indexedDB: ',
+        keyOrValue
+      );
+    }
+    try {
+      value = value != null ? JSON.parse(JSON.stringify(value)) : value;
+    } catch {
+      console.warn(
+        'Could not process value before adding into indexedDB:  ',
+        value
+      );
+    }
     let request;
     if (value !== undefined) {
       logDebug(LOG_TAG, 'PUT', this.store.name, keyOrValue, value);


packages/firestore/src/local/simple_db.ts
```diff
@@ -589,6 +589,25 @@ export class SimpleDbStore<
     keyOrValue: KeyType | ValueType,
     value?: ValueType
   ): PersistencePromise<void> {
+    try {
+      keyOrValue =
+        keyOrValue != null
+          ? JSON.parse(JSON.stringify(keyOrValue))
```
ehsannas, line 595: This would mean we do this for every transaction we do. We need to think about implications (performance etc).

ehsannas (commented):
Thanks for your contribution @Shanzid01 . It's very much appreciated.

We'll need to take some time to understand the underlying issue and whether there's a way to address it without affecting all transactions.

packages/firestore/src/local/simple_db.ts
Shanzid01: Thanks for the review! You're correct, I'm concerned about the performance impact as well. And this definitely doesn't fix the underlying issue (functions being added into the indexedDB), it's only a band-aid for those who desperately need a solution.

The firebase team should definitely consider a better solution for the final patch!

wu-hui: Can you explain why a function is being written to indexed db or why it only happens to Vue, if you know the answers?

Shanzid01: I couldn't determine *why* functions were being added, and why that was happening only in Vue3. But I was able to see *what* functions/Objects were being added:
db-error-reason
The culprit here seems to be the document->fields->**Proxy** Object. Firestore is attempting to insert this Proxy object into the IndexedDB and that's causing the error.

ehsannas: I _think_ this is not added by our codebase ([reference](https://cs.opensource.google/search?q=Proxy&sq=&ss=firebase-sdk%2Ffirebase-js-sdk:packages%2Ffirestore%2F)).

Shanzid01: That's very likely. Would also explain why this is only happening on Vue3. Possibly in one of the shared libraries for reactivity. Will investigate more and see if I can find something.

PR #6177, opened 2022-04-25
Change functions bump to minor

hsubox76:
Amend https://github.com/firebase/firebase-js-sdk/pull/6162 to change version bumps to minor since we're adding a new method and changing the API. My bad for not noticing this when PR was approved.


dwyfrequency (approved):
LGTM

PR #6178, opened 2022-04-25
[Auth] Add onAbort and refactor middleware

sam-gc:
I've refactored the middleware code to its own class since it got sufficiently complicated with the `onAbort()` callbacks.

### DIFF
--- a/packages/auth/src/core/auth/middleware.test.ts
+++ b/packages/auth/src/core/auth/middleware.test.ts
@@ -1,0 +1,132 @@
+import { expect, use } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import * as sinon from 'sinon';
+import sinonChai from 'sinon-chai';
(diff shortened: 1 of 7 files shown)


hsubox76 (approved):

packages/auth/src/core/auth/middleware.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    middlewareQueue.pushCallback(() => {
+      throw new Error('no');
+    }, secondOnAbort);
+
+    await expect(middlewareQueue.runMiddleware(user)).to.be.rejectedWith('auth/login-blocked');
+    expect(firstOnAbort).to.have.been.called;
+    expect(secondOnAbort).not.to.have.been.called;
+  });
+
+  it('calls onAbort in order', async () => {
```
yuchenshi, line 97:
```suggestion
  it('calls onAbort in reverse order', async () => {
```

yuchenshi (approved):

PR #6187, opened 2022-04-27
Remove git-rev-sync dependency.

dwyfrequency:
Removing git-rev-sync package as it is no longer used in the codebase


hsubox76 (commented):
You know what, it looks like we don't even use this dependency anymore. Try removing it?

hsubox76 (approved):
Thanks! Good to clean out some deps once in a while.

PR #6191, opened 2022-04-27
Add minor bump to main package

hsubox76:
Forgot firebase minor and major version bumps need to be added manually if a dependent package has a minor/major bump (in this case, functions: https://github.com/firebase/firebase-js-sdk/pull/6177)


dwyfrequency (approved):
LGTM

PR #6214, opened 2022-05-03
Update firebase-functions dependency

dwyfrequency:
Updating firebase-functions package as mentioned in [pull request 5566](https://github.com/firebase/firebase-js-sdk/pull/5566/files#) for all non-major dependencies.

### DIFF
--- a/config/functions/package.json
+++ b/config/functions/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "cors": "2.8.5",
     "firebase-admin": "10.1.0",
-    "firebase-functions": "3.20.0"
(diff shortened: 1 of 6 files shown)


hsubox76 (approved):
I notice you also updated this 8 days ago: https://github.com/firebase/firebase-js-sdk/pull/6183 Maybe to save yourself some work going forward, if it's something that was recently updated with no issues you can save it up and let it be done as part of the all-in-one renovate PR once the problem deps have been fixed (webpack etc), or make your own big PR with a bunch of deps bundled together that you're pretty sure are unlikely to cause any problems.

PR #6215, opened 2022-05-03
Update firebase dependency

dwyfrequency:
Updating firebase package as mentioned in [pull request 5566](https://github.com/firebase/firebase-js-sdk/pull/5566/files#) for all non-major dependencies.

### DIFF
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -16,7 +16,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "firebase": "9.0.2"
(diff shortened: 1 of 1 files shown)


hsubox76 (approved):

dwyfrequency:
> Does renovate have some kind of config file where you can specify packages to ignore?

Yes, they have docs on adding a config file https://docs.renovatebot.com/configuration-options/ . We could use something like https://docs.renovatebot.com/configuration-options/#ignorepaths to ignore the e2e directories

PR #6217, opened 2022-05-03
Add a workflow to merge release into master

hsubox76:
Manually triggerable workflow that merges release into master. Won't be able to test it until after next release.

### DIFF
--- a/.github/workflows/merge-release-branch.yml
+++ b/.github/workflows/merge-release-branch.yml
@@ -1,0 +1,22 @@
+name: Merge Release Into Master
+
+on: repository_dispatch
+
+jobs:
+  merge_to_master:
+    runs-on: ubuntu-latest
+    # Allow GITHUB_TOKEN to have write permissions
+    permissions:
+      contents: write
+    steps:
+      - name: Merge to master
+        uses: actions/github-script@0.9.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+              github.repos.merge({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                base: 'master',
+                head: 'release'
+              })
--- a/.github/workflows/merge-release-branch.yml
+++ b/.github/workflows/merge-release-branch.yml
@@ -10,13 +10,13 @@ jobs:
       contents: write
     steps:
       - name: Merge to master
-        uses: actions/github-script@0.9.0
+        uses: actions/github-script@v6
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-release-branch.yml
```diff
@@ -0,0 +1,22 @@
...
+        uses: actions/github-script@0.9.0
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+              github.repos.merge({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                base: 'master',
+                head: 'release'
+              })
```
lahirumaramba, line 22: nit: new line

.github/workflows/merge-release-branch.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  merge_to_master:
+    runs-on: ubuntu-latest
+    # Allow GITHUB_TOKEN to have write permissions
+    permissions:
+      contents: write
+    steps:
+      - name: Merge to master
+        uses: actions/github-script@0.9.0
```
lahirumaramba, line 13: Looks like the latest version is v6 now (includes breaking changes). Worth checking if we can use one of the recent versions here.

.github/workflows/merge-release-branch.yml
```diff
@@ -0,0 +1,22 @@
+name: Merge Release Into Master
+
+on: repository_dispatch
+
+jobs:
+  merge_to_master:
+    runs-on: ubuntu-latest
+    # Allow GITHUB_TOKEN to have write permissions
+    permissions:
+      contents: write
```
lahirumaramba, line 10: Is the goal here to grant `GITHUB_TOKEN` with default permissions (without write access) and add write access in the workflow?

.github/workflows/merge-release-branch.yml
```diff
@@ -0,0 +1,22 @@
+name: Merge Release Into Master
+
+on: repository_dispatch
```
lahirumaramba, line 3: nit feel free to ignore:
We can set` types:` here if we want to restrict the manual trigger even further with a specific `event_type` value in the webhook payload.

```
on:
  repository_dispatch:
    types: [sdk-release-merge]
```

lahirumaramba (approved):
LGTM! Left a few comments

.github/workflows/merge-release-branch.yml
hsubox76: Will add this when/if we trigger from a webhook. We'll use the button in the Github UI for now.

Issue #6218, opened 2022-05-04
No "identities" property in ParseToken interface


kyun:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Monterey 12.1
  * Browser version: Google Chrome 100.0.4896.127
  * Firebase SDK version: 9.6.10
  * Firebase Product: auth

#### Steps to reproduce:

I've been using firebase v8, and yesterday I migrated v8 to v9. 

With firebase v8, I used to get the userId below code.

```js
user.getIdTokenResult().then(({ token, signInProvider, claims })=>{
  const info = claims.firebase;
  const [id] = info?.identities[signInProvider || ''];
  consoel.log('Here is my id', id);
});
```

but after upgrading to v9, I get a type error

> Property 'identities' does not exist on type '{ sign_in_provider?: string | undefined; sign_in_second_factor?: string | undefined; }'.

For testing, I signed up using Twitter, and I got a `ParsedToken`. 

and I printed that token on the console.

```javascript
{
  aud: "####",
  auth_time: 123456,
  exp: 123456,
  firebase: {
    identities: { twitter.com: ["ID"] },
    sign_in_provider: "twitter.com",
  },
  iat: 123456,
  ...
}
````
`identities` is existed.

but there are no `identities` props in `@firebase/auth/dist/auth-public.d.ts`

```js
/**
 * Interface representing a parsed ID token.
 *
 * @privateRemarks TODO(avolkovi): consolidate with parsed_token in implementation.
 *
 * @public
 */
export declare interface ParsedToken {
    /** Expiration time of the token. */
    'exp'?: string;
    /** UID of the user. */
    'sub'?: string;
    /** Time at which authentication was performed. */
    'auth_time'?: string;
    /** Issuance time of the token. */
    'iat'?: string;
    /** Firebase specific claims, containing the provider(s) used to authenticate the user. */
    'firebase'?: {
        'sign_in_provider'?: string;
        'sign_in_second_factor'?: string;
    };
    /** Map of any additional custom claims. */
    [key: string]: string | object | undefined;
}
```

I think it should be updated like this.

```js
export declare interface ParsedToken {
    /* ... */
    'iat'?: string;
    /** Firebase specific claims, containing the provider(s) used to authenticate the user. */
    'firebase'?: {
        identities: {
          [key: string]: Array<IdTypeOfSomething>
        };
        'sign_in_provider'?: string;
        'sign_in_second_factor'?: string;
    };
[...]

PR #6219, opened 2022-05-04
add "identities" type on ParsedToken interface

kyun:
### Discussion
https://github.com/firebase/firebase-js-sdk/issues/6218


kyun:
https://github.com/firebase/firebase-js-sdk/pull/6234

PR #6223, opened 2022-05-05
Update API Extractor line endings to "lf" and regenerate app check documentation with lf line endings 

dwyfrequency:
Update API Extractor config file to use POSIX-style newlines (lf) to remove inconsistencies on Github. We were having issues where git thought every new line in the md file was different instead of giving a useful diff report. As most users will be using lf line endings, we have changed the config to use this style. 

To demonstrate this working, we regenerated the app-check API documentation. After all of the API documentation is regenerated, we should have useful diffs in github.


hsubox76:
I think there are a number of other api-report files with crlf endings other than app-check, right?

dwyfrequency:
> I think there are a number of other api-report files with crlf endings other than app-check, right?

Yes, do you want me to generate all of these in this PR.

hsubox76:
> > I think there are a number of other api-report files with crlf endings other than app-check, right?
> 
> Yes, do you want me to generate all of these in this PR?

Yes, thanks.

PR #6225, opened 2022-05-05
Update @types/sinon-chai package

dwyfrequency:
Updating @types/sinon-chai package as mentioned in [pull request 5566](https://github.com/firebase/firebase-js-sdk/pull/5566/files#) for all non-major dependencies.


dwyfrequency:
https://github.com/firebase/firebase-js-sdk/pull/6230 will integrate this change

PR #6247, opened 2022-05-06
Fix API call, add commit message

hsubox76:
- github-script v6 needs REST APIs to be called with `github.rest` instead of `github`.
- Add an input field so we can put the version into the merge commit message (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)

### DIFF
--- a/.github/workflows/merge-release-branch.yml
+++ b/.github/workflows/merge-release-branch.yml
@@ -1,6 +1,11 @@
 name: Merge Release Into Master
 
-on: workflow_dispatch
+on: 
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version'     
+        required: true
 
 jobs:
   merge_to_master:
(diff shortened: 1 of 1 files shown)


.github/workflows/merge-release-branch.yml
```diff
@@ -1,6 +1,11 @@
 name: Merge Release Into Master
 
-on: workflow_dispatch
+on: 
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version'     
+        required: true
```
lahirumaramba, line 8: Can you set the version from the Github UI?
If we would like I think we can extract the version from the package.json as well (could be an improvement for later)
see: https://github.com/firebase/firebase-admin-node/blob/a32195daa9848b261fe892d9f606152a40ff2915/.github/scripts/publish_preflight_check.sh#L61

lahirumaramba (approved):
LGTM! Left one comment.

.github/workflows/merge-release-branch.yml
hsubox76: Thanks, I added this, if you want to take another look. I tested it out with the merge step commented out and it seems to work up to that point.

lahirumaramba (approved):
LGTM! Thanks!

Issue #5433, opened 2021-09-02
Feature Request: An equivalent to `firebase.database.INTERNAL.forceWebSockets()` in v9


buu700:
See title. I see that there's an internal class called `RepoInfo` with property `webSocketOnly`, but it's unclear how this can be used in a client application.

billygerhard:
This is required if you have a strict CSP. The CSP I have will not allow external JS files to be loaded, or anything similar to `eval` so the long polling throws a CSP error and will not allow the page to load at all. I am forced to use `forceWebSockets()` on my current app on v8.

buu700:
Hmm, so based on a review of the code, I see that `webSocketOnly` is actually set to true if a scheme of `ws` or `wss` is used. This has been the behavior for a pretty long time, but I ended up using `firebase.database.INTERNAL.forceWebSockets()` back in November as a workaround for a regression that caused the behavior you described. If that has been fixed in this release, it may not be a hard blocker.

Even so, I would prefer to have this as insurance against such incidents in the future. Exporting `BrowserPollConnection` would be sufficient, as that would allow calling `BrowserPollConnection.forceDisallow()` to restore the previous behavior.

billygerhard:
This is causing production issues because its trying to load external javascript when doing long polling due to strict CSP issues. Is there any way to remove long polling completely like there was in V8. I don't want to redo my entire project and downgrade to V8 because of this.

buu700:
@billygerhard I'm not thrilled about it, but this has been working at least as a temporary workaround for me:

```
[...]

PR #6252, opened 2022-05-09
Updated documentation for compat to include new database APIs

maneesht:
Included INTERNAL properties in type definitions.

### DIFF
--- a/packages/firebase/compat/index.d.ts
+++ b/packages/firebase/compat/index.d.ts
@@ -6019,21 +6019,16 @@ declare namespace firebase.database {
      */
     app: firebase.app.App;
     /**
-     * INTERNAL methods for internal-use only (tests, etc.).
+     * Additional methods for debugging and special cases.
      *
-     * @const
      */
     INTERNAL: {
       /**
-       * Deletes a database
-       */
-      delete: () => Promise<void>;
-      /**
-       * Force the use of websockets instead of longPolling.
+       * Force the use of WebSockets instead of long polling.
        */
       forceWebSockets: () => void;
       /**
-       * Force the use of longPolling instead of websockets. This will be ignored if websocket protocol is used in databaseURL.
+       * Force the use of long polling instead of WebSockets. This will be ignored if the WebSocket protocol is used in `databaseURL`.
        */
       forceLongPolling: () => void;
     };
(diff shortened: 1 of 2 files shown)


jsdt (approved):

packages/firebase/compat/index.d.ts
```diff
@@ -6018,6 +6018,25 @@ declare namespace firebase.database {
      * ```
      */
     app: firebase.app.App;
+    /**
+     * INTERNAL methods for internal-use only (tests, etc.).
```
hsubox76, line 6022: Is this accurate? I thought `forceWebSockets` and `forceLongPolling` are for end users. Maybe "additional methods for debugging and special cases"?

packages/firebase/compat/index.d.ts
```diff
@@ -6018,6 +6018,25 @@ declare namespace firebase.database {
...
+     *
+     * @const
```
hsubox76, line 6024: What does this tag do?

packages/firebase/compat/index.d.ts
```diff
@@ -6018,6 +6018,25 @@ declare namespace firebase.database {
...
+      /**
+       * Deletes a database
+       */
+      delete: () => Promise<void>;
+      /**
+       * Force the use of websockets instead of longPolling.
+       */
+      forceWebSockets: () => void;
+      /**
+       * Force the use of longPolling instead of websockets. This will be ignored if websocket protocol is used in databaseURL.
```
hsubox76, line 6036: longPolling => long polling
websockets => WebSockets
websocket protocol => the WebSocket protocol
add backticks around databaseURL

packages/firebase/compat/index.d.ts
Slapbox: We need `forceWebSockets` back _for end users_ or we need to abandon Firebase - which has otherwise served us well.

See: https://github.com/firebase/firebase-js-sdk/issues/5433#issuecomment-1124260695

packages/firebase/compat/index.d.ts
maneesht: Defines a constant, but it isn't supported by TypeScript

packages/firebase/compat/index.d.ts
maneesht: Changed.

hsubox76 (approved):

egilmorez (approved):
LG, thanks!

PR #6261, opened 2022-05-10
Implement Firestore IndexBackfiller

tom-andersen:
Implement index backfiller


packages/firestore/src/local/local_documents_view.ts
```diff
@@ -56,6 +57,13 @@ import { PersistencePromise } from './persistence_promise';
 import { PersistenceTransaction } from './persistence_transaction';
 import { RemoteDocumentCache } from './remote_document_cache';
 
+export class LocalDocumentsResult {
```
wu-hui, line 60: Comment.

packages/firestore/src/local/index_manager.ts
```diff
@@ -138,4 +138,18 @@ export interface IndexManager {
...
+  /**
+   * Iterates over all field indexes that are used to serve the given target,
+   * and returns the minimum offset of them all.
+   */
+  getMinOffset(
+    transaction: PersistenceTransaction,
+    target: Target
+  ): PersistencePromise<IndexOffset>;
+
+  getMinOffsetFromCollectionGroup(
```
ehsannas, line 151: Add comment

packages/firestore/src/local/persistence_promise.ts
```diff
@@ -243,4 +243,31 @@ export class PersistencePromise<T> {
...
+   * return `true`, the process repeats, until `condition` returns false.
+   * @param condition
+   * @param action
+   */
+  static loopUntil(
+    condition: () => boolean,
+    action: () => PersistencePromise<void>
+  ): PersistencePromise<void> {
+    return new PersistencePromise<void>((resolve, reject) => {
+      const process = (): void => {
```
ehsannas, line 261: I think the preferred style is to remove `:void`
```js
const process = () => {
  // ...
}
```

ehsannas (commented):
I didn't realize that this was assigned to Andy until I had read a few of the files. So, it's just a drive-by review, not a thorough one. Please refer to Andy for approval.

packages/firestore/src/local/lru_garbage_collector_impl.ts
```diff
@@ -110,18 +110,18 @@ class RollingSequenceNumberBuffer {
...
-export class LruScheduler implements GarbageCollectionScheduler {
-  private hasRun: boolean = false;
+export class LruScheduler implements Scheduler {
   private gcTask: DelayedOperation<void> | null;
 
   constructor(
     private readonly garbageCollector: LruGarbageCollector,
-    private readonly asyncQueue: AsyncQueue
+    private readonly asyncQueue: AsyncQueue,
+    private readonly localStore: LocalStore
```
ehsannas, line 119: I'm also curious why LocalStore became a member variable. Is it diverging from the other SDKs?

packages/firestore/src/local/local_documents_view.ts
ehsannas: So the Web SDK has an interface called `LocalWriteResult`. [Ref](https://cs.opensource.google/firebase-sdk/firebase-js-sdk/+/master:packages/firestore/src/local/local_store_impl.ts;l=100-103?q=localwriteresult&ss=firebase-sdk), which is the same thing I believe.

The iOS SDK has a similar thing. [Ref](https://cs.opensource.google/firebase-sdk/firebase-ios-sdk/+/master:Firestore/core/src/local/local_write_result.h;l=32?q=localwriteresult&ss=firebase-sdk).

The Web SDK so far has not used `LocalDocumentResult` so far, and the tests also use `LocalWriteResult`.

I think we should use `LocalWriteResult` in this PR.

Potentially, _if we think LocalWriteResult is the better way to go_, we should make a separate PR that swaps all usecases of `LocalWriteResult` with `LocalDocumentResult`. Alternatively, we can update Android to match Web and iOS.

thoughts?

ehsannas: either way we must make sure all of them use the same name

ehsannas:
@tom-andersen There are a number of tests in `SQLiteLocalStoreTest.java` that are missing in the Web SDK. They should also be ported as a part of this PR. I think the reason they haven't been ported so far was that some of them rely on the `backfillIndexes()` API which is being added in this PR.

repo-scripts/prune-dts/.run/AllTests.run.xml
```diff
@@ -2,7 +2,7 @@
   <configuration default="false" name="All Tests" type="mocha-javascript-test-runner">
     <node-interpreter>project</node-interpreter>
     <node-options />
-    <mocha-package>$PROJECT_DIR$/../../node_modules/mocha</mocha-package>
```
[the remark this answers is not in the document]
tom-andersen: Reverted

packages/firestore/src/local/local_documents_view.ts
tom-andersen: Changed to `LocalWriteResult`

packages/firestore/src/local/local_documents_view.ts
```diff
@@ -319,6 +327,61 @@ export class LocalDocumentsView {
...
+                count - originalDocs.size
+              )
+            : PersistencePromise.resolve(newOverlayMap());
+        let largestBatchId = INITIAL_LARGEST_BATCH_ID;
+        let modifiedDocs = originalDocs;
+        return overlaysPromise.next(overlays => {
+          return PersistencePromise.forEach(
+            overlays,
+            (key: DocumentKey, overlay: Overlay) => {
+              if (largestBatchId < overlay.largestBatchId) {
```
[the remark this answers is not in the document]
tom-andersen: Comment copied

packages/firestore/src/local/persistence_promise.ts
ehsannas: huh! TIL

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -1062,3 +1031,23 @@ function indexStateStore(
...
+
+function getMinOffsetFromFieldIndexes(fieldIndexes: FieldIndex[]): IndexOffset {
+  hardAssert(
+    fieldIndexes.length !== 0,
+    'Found empty index group when looking for least recent index offset.'
+  );
+
+  let minOffset: IndexOffset = fieldIndexes[0].indexState.offset;
+  let maxBatchId: number = minOffset.largestBatchId;
+  for (let i = 1; i < fieldIndexes.length; i++) {
```
ehsannas, line 1043: I'm curious why you switched from `for (const fieldIndex of fieldIndexes) {` ?
I _think_ that's the preferred way in JS if you don't really care about the loop index value

ehsannas (commented):
you can run `yarn prettier && yarn lint:fix` (in `packages/firestore` directory) before you run `git push` so you can catch format/lint issues locally.

ehsannas (commented):
Thanks for addressing my comments. Pass to Andy for thorough review and approval.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -951,6 +969,56 @@ export class IndexedDbIndexManager implements IndexManager {
...
+      minOffset.readTime,
+      minOffset.documentKey,
+      maxBatchId
+    );
+  }
+
+  getMinOffset(
+    transaction: PersistenceTransaction,
+    target: Target
+  ): PersistencePromise<IndexOffset> {
```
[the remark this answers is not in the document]
tom-andersen: Did option #1

packages/firestore/src/local/indexeddb_index_manager.ts
ehsannas: but now you're assuming the array has at least 2 elements, right? `fieldIndexes[1]` shouldn't be accessed if we are trying to find the minOffset of an array of size 1.

packages/firestore/src/local/index_backfiller.ts
```diff
@@ -0,0 +1,216 @@
...
+        await this.schedule(REGULAR_BACKFILL_DELAY_MS);
+      }
+    );
+  }
+}
+
+/** Implements the steps for backfilling indexes. */
+export class IndexBackfiller {
+  constructor(
+    private readonly localStore: LocalStore,
```
wu-hui, line 112: Add a comment to explain why we need localstore here.

packages/firestore/src/local/local_store.ts
```diff
@@ -18,11 +18,22 @@
...
 import { logDebug } from '../util/log';
 
+import { IndexManager } from './index_manager';
+import { LocalDocumentsView } from './local_documents_view';
 import { LruGarbageCollector, LruResults } from './lru_garbage_collector';
 import { PRIMARY_LEASE_LOST_ERROR_MSG } from './persistence_transaction';
 
 export interface LocalStore {
   collectGarbage(garbageCollector: LruGarbageCollector): Promise<LruResults>;
+
```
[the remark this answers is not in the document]
wu-hui: As discussed offline, this change makes user change better to handle, even though it derails the convention. We will go forward with this, and change Android to match this.

packages/firestore/src/local/index_backfiller.ts
tom-andersen: Comment added

packages/firestore/src/model/collections.ts
```diff
@@ -54,20 +55,24 @@ export function documentMap(...docs: Document[]): DocumentMap {
   return map;
 }
 
-export type OverlayMap = ObjectMap<DocumentKey, Overlay>;
-export function newOverlayMap(): OverlayMap {
-  return new ObjectMap<DocumentKey, Overlay>(
-    key => key.toString(),
-    (l, r) => l.isEqual(r)
+export function convertDocumentKeyMapWithOverlayedDocumentToDocumentMap(
```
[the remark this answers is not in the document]
tom-andersen: Introduced type OverlayedDocumentMap (following existing pattern), and renamed method accordingly.

packages/firestore/src/local/indexeddb_index_manager.ts
```diff
@@ -951,26 +966,26 @@ export class IndexedDbIndexManager implements IndexManager {
...
-            indexOffsetComparator(index.indexState.offset, offset) < 0
-          ) {
-            offset = index.indexState.offset;
-          }
-        });
-      }
-    ).next(() => offset!);
+      (subTarget: Target) =>
+        this.getFieldIndex(transaction, subTarget).next(index =>
+          !index ? fail('Target cannot be served from index') : index
```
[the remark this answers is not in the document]
wu-hui: Sounds good.

PR #6269, opened 2022-05-11
Deprecate analytics function setCurrentScreen

dwyfrequency:
Mark setCurrentScreen function within analytics as deprecated. Users should utilize logEvent with eventName as 'screen_view' and add the relevant eventParams. For more information, see https://firebase.google.com/docs/analytics/screenviews#web-version-9

### DIFF
--- a/packages/analytics/src/functions.ts
+++ b/packages/analytics/src/functions.ts
@@ -53,6 +53,8 @@ export async function logEvent(
 /**
  * Set screen_name parameter for this Google Analytics ID.
  *
+ * @deprecated Use logEvent with eventName as 'screen_view' and add relevant eventParams.
+ *
  * @param gtagFunction Wrapped gtag function that waits for fid to be set before sending an event
  * @param screenName Screen name string to set.
  */
--- a/packages/firebase/compat/index.d.ts
+++ b/packages/firebase/compat/index.d.ts
@@ -5215,6 +5215,8 @@ declare namespace firebase.analytics {
 
     /**
      * Use gtag 'config' command to set 'screen_name'.
+     *
+     * @deprecated Use logEvent with eventName as 'screen_view' and add relevant eventParams.
      */
     setCurrentScreen(
       screenName: string,
--- a/common/api-review/analytics.api.md
+++ b/common/api-review/analytics.api.md
@@ -388,7 +388,7 @@ export interface Promotion {
 // @public
 export function setAnalyticsCollectionEnabled(analyticsInstance: Analytics, enabled: boolean): void;
 
-// @public
+// @public @deprecated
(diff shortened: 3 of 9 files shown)


packages/analytics/src/functions.ts
```diff
@@ -53,6 +53,8 @@ export async function logEvent(
 /**
  * Set screen_name parameter for this Google Analytics ID.
  *
+ * @deprecated Use logEvent with eventName as 'screen_view' and add relevant eventParams.
```
hsubox76, line 56: Can you add backticks to function and variable names (logEvent, eventName, eventParams)? I think screen_view makes sense with single quotes since it's a string. Actually I think you can link directly to them if they're documented with `{@link logEvent}`, see https://github.com/firebase/firebase-js-sdk/blob/master/packages/analytics/src/api.ts#L85 You can generate the docs and double check the links work.

I think only logEvent and eventParams warrant a link.

packages/firebase/compat/index.d.ts
```diff
@@ -5215,6 +5215,8 @@ declare namespace firebase.analytics {
 
     /**
      * Use gtag 'config' command to set 'screen_name'.
+     *
+     * @deprecated Use logEvent with eventName as 'screen_view' and add relevant eventParams.
```
hsubox76, line 5219: Same as above. Unfortunately links work a little differently in this file as you have to specify the namespace and you're allowed to customize the link text I guess. Search for "@link" in this file for examples, one example is {@link firebase.analytics.Analytics `Analytics`} Again, generate docs and make sure the link looks right.

packages/analytics/src/functions.ts
dwyfrequency: Added the `{@links logEvent}` and added backticks where necessary

packages/firebase/compat/index.d.ts
dwyfrequency: Added the `{@links logEvent}` and added backticks where necessary. The `logEvent` link here auto resolves to the compat version `firebase.analytics.Analytics.logEvent` in vscode and in the resulting html files post docgen:compat

packages/analytics/src/functions.ts
hsubox76: Sorry, slipped my mind we can actually link guide docs.  Can you add a second line
```
 * See {@link https://firebase.google.com/docs/analytics/screenviews
 * | Track Screenviews}.
```

That's from the doc comment for the logEvent overload for 'screen_view' but since we can't link to that overload might as well put it right here.

dwyfrequency: Added the documentation link for Track Screenviews

hsubox76 (approved):

PR #4408, opened 2021-02-03
Apply range-filters in get when probing active-listener cache

IanWyszynski:

PR #6273, opened 2022-05-13
Fix for repoGet request

maneesht:
Fix issue where the wrong cache was being updated by get. First noticed [here](https://github.com/firebase/firebase-ios-sdk/issues/8286)

### DIFF
--- a/packages/database/test/helpers/util.ts
+++ b/packages/database/test/helpers/util.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-
 import { Database, ref } from '../../src';
 import { ConnectionTarget } from '../../src/api/test_access';
 


packages/database/src/core/Repo.ts
```diff
@@ -466,15 +469,50 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
-      );
-      eventQueueRaiseEventsAtPath(repo.eventQueue_, query._path, events);
+      // if this is not a filtered query, then overwrite at path
+      if (query._queryParams.loadsAllData()) {
+        const events = syncTreeApplyServerOverwrite(
+          repo.serverSyncTree_,
+          query._path,
+          node
+        );
+        eventQueueRaiseEventsAtPath(repo.eventQueue_, query._path, events);
```
IanWyszynski, line 482: Looking at this again, I suspect we may need to add a call to `repoRerunTranactions` call here. 

We do so in the data update handling code for a listener: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/core/Repo.ts#L400-L406.

IanWyszynski: (same thing for L504, after `applyTaggedQueryMerge`)

packages/database/src/core/Repo.ts
```diff
@@ -466,15 +469,50 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
+          syncPoint,
+          query,
+          writesCache,
+          serverCache,
+          serverCacheComplete
+        );
+        if (!syncPoint.views.has(query._queryIdentifier)) {
+          syncPoint.views.set(query._queryIdentifier, view);
+        }
+        const tag = syncTreeTagForQuery_(repo.serverSyncTree_, query);
```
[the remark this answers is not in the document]
IanWyszynski: (another reason I'd prefer to move this synctree setup logic out of the result callback)

packages/database/test/exp/integration.test.ts
```diff
@@ -97,6 +102,21 @@ describe('Database@exp Tests', () => {
...
 
+  // Tests to make sure onValue's data does not get mutated after calling get
+  it('calls onValue only once after get request', async () => {
+    const db = getDatabase(defaultApp);
+    const testRef = ref(db, 'foo');
+    const initial = [{ name: 'child1' }, { name: 'child2' }];
+    await set(testRef, initial);
+    let count = 0;
+    onValue(testRef, () => {
+      count++;
```
IanWyszynski, line 113: Perhaps it would be worth asserting that the snapshot is correct.

packages/database/src/core/Repo.ts
maneesht: I don't think we need to do this as `repoOnDataUpdate` does this for the listeners

packages/database/src/core/SyncTree.ts
```diff
@@ -482,16 +501,10 @@ export function syncTreeApplyTaggedQueryMerge(
...
 }
 
-/**
- * Add an event callback for the specified query.
- *
- * @returns Events to raise.
- */
-export function syncTreeAddEventRegistration(
-  syncTree: SyncTree,
+export function syncTreeRegisterSyncPoint(
```
IanWyszynski, line 504: Another comment here to explain the rational for splitting this out from syncTreeAddEventRegistration. Can be short.

packages/database/src/core/SyncTree.ts
```diff
@@ -550,6 +563,36 @@ export function syncTreeAddEventRegistration(
     syncTree.tagToQueryMap.set(tag, queryKey);
   }
   const writesCache = writeTreeChildWrites(syncTree.pendingWriteTree_, path);
+  // TODO: break this down so you do the minimal amount
```
IanWyszynski, line 566: It looks like you added this TODO? Could you either clarify this a bit more or get rid of ti?

packages/database/test/exp/integration.test.ts
```diff
@@ -97,6 +105,39 @@ describe('Database@exp Tests', () => {
...
 
+  // Tests to make sure onValue's data does not get mutated after calling get
+  it('calls onValue only once after get request with a non-default query', async () => {
+    const db = getDatabase(defaultApp);
+    const testRef = ref(db, 'foo');
+    const initial = [{ name: 'child1' }, { name: 'child2' }];
+
+    let count = 0;
+    await set(testRef, initial);
+    onValue(testRef, snapshot => {
```
IanWyszynski, line 116: I think these tests could be refactored into table-tests over a spec like this:

```
{
  "initialData": [{ name: 'child1' }, { name: 'child2' }],
  "initialLocation": <some path>,
  "listenerFunc": either onChild{Added, Changed, Removed} or onValue,
  "listenerQuery": <the query to listen to>,
  "expectedListenResults": [<array of snaps>] # fail if exactly these aren't received"
  "getQuery": <the query to get>,
   "expectedGetResults": <snap>, # fail if exactly this one isn't received.
  
}
```

Then we could easily add similar tests to the ones added here and check for various combinations of listens and gets (listen above get, below get, at get, etc.).

However, as I'm wrapping up this comment I suspect this refactor deserves it's own PR so feel free to ignore.

IanWyszynski (approved):
LGTM after some docs are added. Ignore the note about testing refactor (or take it as a suggestion for a followup PR).

IanWyszynski:
We'll want to generate a changeset here: docs: https://github.com/firebase/firebase-js-sdk/pull/4408/files,  example: https://github.com/firebase/firebase-js-sdk/pull/4408/files.

.changeset/rotten-tables-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fixed issue where `get()` updated incorrect cache.
```
IanWyszynski, line 5: nit: Fixed issue where `get()` updated cache incorrectly for non-default queries.

packages/database/src/core/Repo.ts
```diff
@@ -464,17 +465,37 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
-        node
-      );
-      eventQueueRaiseEventsAtPath(repo.eventQueue_, query._path, events);
+      // if this is a filtered query, then overwrite at path
+      if (query._queryParams.loadsAllData()) {
+        syncTreeApplyServerOverwrite(repo.serverSyncTree_, query._path, node);
+      } else {
+        // Simulate `syncTreeAddEventRegistration` without events/listener setup.
+        // We do this (along with the syncTreeRemoveEventRegistration` below) so that
+        // `repoGetValue` results have the same cache effects as initial listeners
```
IanWyszynski, line 480: nit: s/listeners/listener

packages/database/src/core/Repo.ts
```diff
@@ -464,17 +465,37 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
+          node,
+          tag
+        );
+        // Call `syncTreeRemoveEventRegistration` with a null event registration, since there is none.
+        const cancels = syncTreeRemoveEventRegistration(
+          repo.serverSyncTree_,
+          query,
+          null
+        );
+        assert(
```
IanWyszynski, line 494: We should never get cancel events here, but maybe we should turn this into a warning log just in case.

```
if (cancels.length > 0) {
  repoLog(repo, 'ignoring unexpected cancel events in get for query ' + stringify(query));
}
```

packages/database/src/core/SyncTree.ts
```diff
@@ -423,6 +423,38 @@ export function syncTreeRemoveEventRegistration(
...
+ * needed for listeners without installing an event registration.
+ * If `query` is not `loadsAllData`, it will also provision a tag for
+ * the query so that query results can be merged into the sync
+ * tree using existing logic for tagged listener queries.
+ *
+ * @param syncTree - Synctree to add the query to.
+ * @param query - Query to register
+ * @returns tag as a string if query is not a default query, null if query is not.
+ */
+export function syncTreeRegisterQuery(syncTree: SyncTree, query: QueryContext) {
```
IanWyszynski, line 439: The asymmetry of `syncTreeRegisterQuery` without `syncTreeUnregisterQuery` might raise some eyebrows. I would either add:

```
export function syncTreeUnregisterQuery(syncTree: SyncTree, query: QueryContext) {
  const cancels = syncTreeRemoveEventRegistration(
    repo.serverSyncTree_,
    query,
    null
   );
   if (cancels.length > 0) {
     repoLog(repo, 'ignoring unexpected cancel events in get for query ' + stringify(query));
   }
}
```

or note the reason we don't have `syncTreeUnregisterQuery` in the comments. I'm fine with either.

maneesht: I think this additional function makes sense

maneesht: Ended up just keeping it in `repoGetValue` since we aren't logging in `SyncTree` and it didn't seem like the right place to put the function.

packages/database/test/exp/integration.test.ts
```diff
@@ -97,6 +105,39 @@ describe('Database@exp Tests', () => {
     expect(snap2).to.equal('b');
   });
 
+  // Tests to make sure onValue's data does not get mutated after calling get
```
IanWyszynski, line 108: Lets also add tests where the onValue and get are done on paths above and below each other.

Also, maybe a test that checks that nothing strange happens when a listener is removed

packages/database/src/core/SyncTree.ts
```diff
@@ -423,6 +423,38 @@ export function syncTreeRemoveEventRegistration(
...
+  if (!query._queryParams.loadsAllData()) {
```
IanWyszynski, line 440: I'm not sure if this is incorrect, but it's worth pointing out that the `syncTreeAddEventRegistration` for a default query will do the logic of L441 to L452.

I'm not sure why the [assertion](https://github.com/firebase/firebase-js-sdk/blob/cdada6c68f9740d13dd6674bcb658e28e68253b6/packages/database/src/core/SyncTree.ts#L857-L858) in `syncTreeApplyTaggedQueryOverwrite` doesn't fail in tests for default queries, shouldn't it be set in `syncTreeRegisterSyncPoint`?

IanWyszynski: Will the tests still pass if you push this branch down to:
```
if (!query._queryParams.loadsAllData()) {
  return syncTreeTagForQuery_(syncTree, query);
} else {
  return null;
}
```

yarn.lock
```diff
@@ -17099,7 +17099,7 @@ uuid@^3.0.0, uuid@^3.3.2, uuid@^3.3.3:
   resolved "https://registry.npmjs.org/uuid/-/uuid-3.4.0.tgz#b23e4358afa8a202fe7a100af1f5f883f02007ee"
   integrity sha512-HjSDRw6gZE5JMggctHBcjVak08+KEVhSIiDzFnT9S9aegmp85S/bReBVTb4QTFaRNptJ9kuYaNhnbNEOkbKb/A==
 
-uuid@^8.3.2:
+uuid@^8.0.0, uuid@^8.3.2:
```
IanWyszynski, line 17102: Is this intentional? Seems like we shouldn't need to pull in an older version.

packages/database/test/exp/integration.test.ts
```diff
@@ -95,14 +105,113 @@ describe('Database@exp Tests', () => {
...
+    const db = getDatabase(defaultApp);
+    const { ref: testRef } = getUniqueRef(db);
+    const initial = [{ name: 'child1' }, { name: 'child2' }];
+    const ec = EventAccumulatorFactory.waitsForExactCount(1);
+
+    await set(testRef, initial);
+    const unsubscribe = onValue(testRef, snapshot => {
+      ec.addEvent(snapshot.val());
+    });
+    await get(query(testRef, limitToFirst(1)));
```
IanWyszynski, line 122: Could you turn this into a table test and apply this test case to a few more interesting queries?

```
const queries = [
  query(testRef, limitToFirst(1)),
  query(testRef, startAt("child2")),
  ... etc
]

queries foreach { q => runTest(q) }
```

Some inspiration from the query integration tests: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database-compat/test/query.test.ts#L38.

maneesht: Will do.

yarn.lock
maneesht: Good catch. I assumed yarn would've resolved to the existing one installed in the repo

packages/database/src/core/SyncTree.ts
maneesht: Don't default queries not have tags?

IanWyszynski: Corrected.

yarn.lock
IanWyszynski: Looks like this is still here?

IanWyszynski:
Latest commit w new tests LGTM. @jsdt could you have a look for code-owner approval?

IanWyszynski (approved):

jsdt (approved):

packages/database/package.json
```diff
@@ -56,7 +56,8 @@
     "@firebase/app": "0.7.25",
     "rollup": "2.72.1",
     "rollup-plugin-typescript2": "0.31.2",
-    "typescript": "4.2.2"
+    "typescript": "4.2.2",
+    "uuid": "^8.3.2"
```
hsubox76, line 60: Can we avoid bringing in an extra dep here? I see in app-check we just implemented a function ourselves, is this good enough? https://github.com/firebase/firebase-js-sdk/blob/6f0049e66064809ae990a2d9461e28b2d6d08d19/packages/app-check/src/util.ts#L43

If it is, maybe we can extract it into `@firebase/util` and share it between app-check and database (maybe in a later PR, just have a duplicate implementation in this one?)

maneesht: Yeah, let me just reimplement it here and add a TODO

hsubox76 (approved):

PR #6279, opened 2022-05-16
Create staging release workflow

hsubox76:
Moving as many release tasks as possible out of local script and into Github Actions.

Try creating a workflow for staging first. If all works well, make one for production in a future PR.

- Merge master into release
- Check out release branch
- yarn install
- Add changeset for @firebase/app to ensure that package is always published (this keeps the SDK_VERSION variable current)
- Create release PR (the "Version Changes" PR - use changesets Github action to automatically do this)
- Publish to NPM using wombat-dressing-room tokens
- Launch E2E tests workflow (changed e2e-test.yml workflow to use a `repository_dispatch` trigger to make it easier to call from here)
- Log staging info to Firebase release tracker (call a cloud function) - we should probably add some kind of auth check before launching this

CDN steps will be run separately and manually in google3 after the NPM publish is done.

### DIFF
--- a/.github/workflows/release-staging.yml
+++ b/.github/workflows/release-staging.yml
@@ -1,0 +1,131 @@
+name: Staging Release
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  deploy:
+    name: Staging Release
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Set up Node (14)
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14.x
+    - name: Merge master into release
+      uses: actions/github-script@v6
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        script: |
+          github.repos.merge({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            base: 'release',
+            head: 'master'
+          })
+    - name: Checkout release branch (with history)
+      uses: actions/checkout@master
+      with:
+        # Release script requires git history and tags.
+        fetch-depth: 0
+        ref: release
+    - name: Yarn install
+      run: yarn
+    # Ensures a new @firebase/app is published with every release.
+    # This keeps the SDK_VERSION variable up to date.
+    - name: Add a changeset for @firebase/app
+      # pull master so changeset can diff against it
+      run: |
+        git pull -f --no-rebase origin master:master
+        yarn ts-node-script scripts/ci/add_changeset.ts
+    - name: Create Release Pull Request
+      uses: changesets/action@v1
+      env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-staging.yml
```diff
@@ -0,0 +1,131 @@
...
+        NPM_TOKEN_PERFORMANCE_COMPAT: ${{ secrets.NPM_TOKEN_PERFORMANCE_COMPAT }}
+        NPM_TOKEN_REMOTE_CONFIG_COMPAT: ${{ secrets.NPM_TOKEN_REMOTE_CONFIG_COMPAT }}
+        NPM_TOKEN_DATABASE_COMPAT: ${{ secrets.NPM_TOKEN_DATABASE_COMPAT }}
+        NPM_TOKEN_FIRESTORE_COMPAT: ${{ secrets.NPM_TOKEN_FIRESTORE_COMPAT }}
+        NPM_TOKEN_STORAGE_COMPAT: ${{ secrets.NPM_TOKEN_STORAGE_COMPAT }}
+        NPM_TOKEN_APP_CHECK_COMPAT: ${{ secrets.NPM_TOKEN_APP_CHECK_COMPAT }}
+        NPM_TOKEN_API_DOCUMENTER: ${{ secrets.NPM_TOKEN_API_DOCUMENTER }}
+        CI: true
+    - name: Launch E2E tests workflow
+      uses: peter-evans/repository-dispatch@v2
```
lahirumaramba, line 116: This is a minor concern, but if we think there is a risk of passing tokens to a third party GHA, we could probably write our own action and host it in the same repo.

.github/workflows/release-staging.yml
```diff
@@ -0,0 +1,131 @@
...
+    - name: Log to release tracker
+      # Sends release information to cloud functions endpoint of release tracker.
+      run: |
+        DATE=$(date +'%m/%d/%Y')
+        BASE_VERSION=${{ steps.get-version.outputs.BASE_VERSION }}
+        STAGING_VERSION=${{ steps.get-version.outputs.STAGING_VERSION }}
+        OPERATOR=${{ github.actor }}
+        curl -X POST -H "Content-Type:application/json" \
+        -d "{\"version\":\"$BASE_VERSION\",\"tag\":\"$STAGING_VERSION\",\"date\":\"$DATE\",\"operator\":\"$OPERATOR\"}" \
+        https://us-central1-feature-tracker-8ca2b.cloudfunctions.net/logStaging
```
lahirumaramba, line 131: hmm should we store the endpoint in a secret?

.github/workflows/release-staging.yml
hsubox76: Since it's pretty simple I changed it to just a curl.

dwyfrequency (approved):
LGTM

PR #6283, opened 2022-05-18
Firestore overlays

ehsannas:


ehsannas:
@wu-hui This is to merge 4 PRs that were each reviewed in the `firestore-overlays` branch into `master`. 

I just had to add 1 commit that should be reviewed which resolves merge conflicts into master (see above).

Thanks

packages/firestore/test/unit/local/query_engine.test.ts
```diff
@@ -508,7 +516,117 @@ function genericQueryEngineTest(
...
+
+  if (!durable) {
+    return;
+  }
+
+  if (!INDEXING_ENABLED) {
+    return;
+  }
+
+  it('combines indexed with non-indexed results', async () => {
```
wu-hui, line 528: Should we port this to Android? I could not find it.

wu-hui (approved):

packages/firestore/test/unit/local/query_engine.test.ts
ehsannas: I don't know why GitHub shows this as a new addition in the diff.

These tests exist in the tip-of-tree in the Web SDK. They also exist on Android. Sebastian ported these from Android to the Web via https://github.com/firebase/firebase-js-sdk/pull/6153

PR #6289, opened 2022-05-23
Propagate customData in FirebaseError when the user is disabled. 

prameshj:
Internal bug - b/229774279

### DIFF
--- a/packages/auth/src/api/authentication/idp.test.ts
+++ b/packages/auth/src/api/authentication/idp.test.ts
@@ -87,4 +87,34 @@ describe('api/authentication/signInWithIdp', () => {
     );
     expect(mock.calls[0].request).to.eql(request);
   });
+  it('should handle user disabled errors and propagate email info', async () => {
+    const response = {
+        email: "test123@gmail.com",
+        error: {
+          code: 400,
+          message: ServerError.USER_DISABLED,
+          errors: [
+            {
+              message: ServerError.USER_DISABLED
+            }
+          ]
+        }
+    };
+
+    const mock = mockEndpoint(
+      Endpoint.SIGN_IN_WITH_IDP,
+      response,
+      400
+    );
+
+    await expect(signInWithIdp(auth, request))
+    .to.be.rejectedWith(FirebaseError, 'auth/user-disabled')
+    .eventually.with.deep.property('customData', {
+      appName: 'test-app',
+      _tokenResponse: response,
+      email: "test123@gmail.com"
+    });
+
+    expect(mock.calls[0].request).to.eql(request);
+  });
 });
--- a/.changeset/late-toys-film.md
+++ b/.changeset/late-toys-film.md
@@ -1,0 +1,5 @@
+---
+'@firebase/auth': patch
+---
+
+Propagate customData in FirebaseError when the user is disabled.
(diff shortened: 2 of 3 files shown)


packages/auth/demo/src/index.js
```diff
@@ -277,7 +277,7 @@ function onAuthError(error) {
     // Handle second factor sign-in.
     handleMultiFactorSignIn(getMultiFactorResolver(auth, error));
   } else {
-    alertError('Error: ' + error.code);
+    alertError('Error: ' + error.code + 'user email : ' + error.customData.email);
```
sam-gc, line 280: nit, put a space in front of "user email" so that it doesn't print right next to the error code

lisajian, line 280: Not every error will have a customData.email field - maybe we just leave it as the error code for now?

prameshj: I added this, so we can identify other error paths where the email is missing but should be propagated, during testing with the demo app.

packages/auth/src/api/authentication/idp.test.ts
```diff
@@ -87,4 +87,35 @@ describe('api/authentication/signInWithIdp', () => {
...
+    let error: FirebaseError | null = null;
+    try {
+    await signInWithIdp(auth, request);
+    } catch (e) {
+      error = e;
+    }
+    expect(error!.customData!.email).to.eq("test123@gmail.com");
+    expect(error!.code).to.eq(
+      'auth/user-disabled'
+    );
```
sam-gc, line 117: You can use chai-as-promised to put all of this error testing logic into one assertion phrase. From ap/index.test.ts for example:
https://github.com/firebase/firebase-js-sdk/blob/b6196d28ce6117e17dbe87cfd554d568012ef31d/packages/auth/src/api/index.test.ts#L178-L183

i.e.

```js
await expect(signInWithIdp(auth, request)).to.be.rejectedWith('auth/user-disabled')
    .eventually.with.deep.property('customData', {email: 'test123@gmail.com'});
```

packages/auth/demo/src/index.js
lisajian: Right, what I mean is not every user that is signed up with Firebase Auth will have an email attached to them. For finding error paths where the email is missing, maybe we can leave that as a local change while we're developing

lisajian (commented):
Could you attach a changeset and add an internal bug number or link to a Github issue to the PR description? Overall LGTM pending nits

egilmorez (approved):
Text string LG, thanks!

packages/auth/demo/src/index.js
prameshj: Sounds good, done.

packages/auth/src/api/authentication/idp.test.ts
prameshj: Done.

packages/auth/demo/src/index.js
prameshj: Removed this line.

prameshj:
> Could you attach a changeset and add an internal bug number or link to a Github issue to the PR description? Overall LGTM pending nits

Thanks! I am having trouble reproing the fix in the demo app now. I will submit the changeset once I resolve that.

sam-gc (approved):

lisajian:
> 

SG, thanks! Giving a preemptive LGTM in the meantime. Thanks for the fix!

> > Could you attach a changeset and add an internal bug number or link to a Github issue to the PR description? Overall LGTM pending nits
> 
> Thanks! I am having trouble reproing the fix in the demo app now. I will submit the changeset once I resolve that.

lisajian (approved):

PR #6294, opened 2022-05-24
Update release script with skipPrompts option

hsubox76:
In order to run the release script in CI, it needs to skip the interactive prompts. Added a `skipPrompts` flag to the command to allow this.

Also the package.json isn't updated until after this script has run, so moved the step that gets the release version from the package.json to after the release script step.

### DIFF
--- a/.github/workflows/release-staging.yml
+++ b/.github/workflows/release-staging.yml
@@ -45,19 +45,6 @@ jobs:
       uses: changesets/action@v1
       env:
         GITHUB_TOKEN: ${{ secrets.OSS_BOT_GITHUB_TOKEN }}
-    - name: Get release version
-      id: get-version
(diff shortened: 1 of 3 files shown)


.github/workflows/release-staging.yml
```diff
@@ -116,6 +103,19 @@ jobs:
         NPM_TOKEN_APP_CHECK_COMPAT: ${{ secrets.NPM_TOKEN_APP_CHECK_COMPAT }}
         NPM_TOKEN_API_DOCUMENTER: ${{ secrets.NPM_TOKEN_API_DOCUMENTER }}
         CI: true
+    - name: Get release version
+      id: get-version
+      # STAGING_VERSION = version with staging hash, e.g. 1.2.3-20430523
```
dwyfrequency, line 108: Do you want to keep these comments in here?

dwyfrequency (approved):
LGTM

PR #6296, opened 2022-05-24
Fix staging workflow

hsubox76:
I actually tested this during development by calling the REST endpoint on this branch after each change, so it should all be working now.

- E2E test uses latest Firebase CLI which just recently stopped supporting Node 12
- Update release script so that if it is staging, and in CI, it will publish using the `publishInCI()` function, which runs the `npm publish` command on each package one by one, which lets us provide each package's specific wombat-dressing-room token. The simpler `publish()` function uses `changeset publish` which is more convenient but doesn't allow us to use wombat-dressing-room.
- Delete `release.yml`. This used to be triggered on push to the release branch. It's redundant now as the same step is part of the `release-staging.yml` workflow.

TODO: Production publish script will have to find another way to grab the names of all the packages to publish since it doesn't run `bumpVersionForStaging`.


dwyfrequency (approved):
LGTM

PR #6304, opened 2022-05-26
Create production release workflow

hsubox76:
This is the production version of the staging release workflow, added in:
https://github.com/firebase/firebase-js-sdk/pull/6279
https://github.com/firebase/firebase-js-sdk/pull/6291
https://github.com/firebase/firebase-js-sdk/pull/6294
https://github.com/firebase/firebase-js-sdk/pull/6296

It differs from the staging workflow in that
- it does not create a changeset PR
- it calls the release script with the `--Production` flag which:
  - writes tags to the Github repository afterwards
  - does not bump versions manually since the correct versions should already be merged to the release branch
- it does not run E2E tests
- when done, it notifies the logProduction endpoint on the release tracker instead of logStaging

I've also modified `publishToCI()` so that it checks if a version has already been published to NPM and skips publishing it if so.

### DIFF
--- a/.github/workflows/release-prod.yml
+++ b/.github/workflows/release-prod.yml
@@ -84,37 +84,20 @@ jobs:
         CI: true
     - name: Get release version
       id: get-version
-      # STAGING_VERSION = version with staging hash, e.g. 1.2.3-20430523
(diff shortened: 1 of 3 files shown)


lahirumaramba (approved):
LGTM!

PR #6305, opened 2022-05-27
Prod workflow: Fix logic for detecting published packages

hsubox76:
- If production release, pass all packages to publishToCI(), which will sort out which needs to be published.
- Fix logic in publishToCI() to detect this.

Was able to test run successfully with some modifications (used --dryRun flag, skipped release tracker step)


lahirumaramba (approved):

dwyfrequency (approved):
LGTM

PR #6317, opened 2022-05-31
Restore v9 TOC menu sub-items

hsubox76:
Revert part of https://github.com/firebase/firebase-js-sdk/pull/5333 so that we go back to generating TOC sub-items for each product.


dwyfrequency (approved):
LGTM

PR #6319, opened 2022-05-31
Remove app-compat from peerDependencies in database-compat

hsubox76:
This prevents admin-node users from seeing a warning message during npm/yarn install:
https://github.com/firebase/firebase-admin-node/issues/1656

We're taking the same approach as https://github.com/firebase/firebase-js-sdk/pull/2082 in that admin-node users don't need that dep at all, and client JS SDK users of the database-compat package ought to be installing the `firebase` umbrella package and should never need to worry about `@firebase/app-compat` not being installed.

### DIFF
--- a/.changeset/green-turtles-camp.md
+++ b/.changeset/green-turtles-camp.md
@@ -1,0 +1,5 @@
+---
+'@firebase/database-compat': patch
+---
+
+Remove app-compat from peerDependencies to avoid npm install warning in firebase-admin.


maneesht:
Looks like in `packages/database-compat/src/index.node.ts` we're using `@firebase/app-compat`, is that from the symlinked package then?

hsubox76:
Yes, it's used by both index.node.ts and index.ts. However I'm proposing that it's safe (if a little incorrect) to leave it out of peerDependencies because anyone using builds from index.node.ts or index.ts will be a client-side JS SDK consumer, and they'll already have it installed because npm installing `firebase` installs everything, including app-compat. Or if they're using it from the CDN, they'll have to have added the app-compat.js script before adding any other products.

Same rationale as outlined in https://github.com/firebase/firebase-js-sdk/pull/2082 (where we did the same thing, pre v9).

I've now added an eslint-disable comment in index.node.ts (same as was already in index.ts) to avoid the eslint error from this.

maneesht (approved):

lahirumaramba (approved):
Thank you, Christina! LGTM!

PR #6320, opened 2022-06-01
Update error handling to account for TS v4.4 auto unknown variable in…

dwyfrequency:
Updating error handling in select packages to account for a future TypeScript package upgrade. For more information, see the below blog comment about TypeScript v4.4 errors.

"""
In JavaScript, any type of value can be thrown with throw and caught in a catch clause. Because of this, TypeScript historically typed catch clause variables as any, and would not allow any other type annotation:

Once TypeScript added the unknown type, it became clear that unknown was a better choice than any in catch clause variables for users who want the highest degree of correctness and type-safety, since it narrows better and forces us to test against arbitrary values. Eventually TypeScript 4.0 allowed users to specify an explicit type annotation of unknown (or any) on each catch clause variable so that we could opt into stricter types on a case-by-case basis; however, for some, manually specifying : unknown on every catch clause was a chore.
"""
per https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

### DIFF
--- a/packages/util/test/errors.test.ts
+++ b/packages/util/test/errors.test.ts
@@ -88,9 +88,9 @@ describe('FirebaseError', () => {
     try {
       throw e;
     } catch (error) {
-      assert.isDefined(error.stack);
+      assert.isDefined((error as Error).stack);
       // Multi-line match trick - .* does not match \n
-      assert.match(error.stack, /FirebaseError[\s\S]/);
+      assert.match((error as Error).stack!, /FirebaseError[\s\S]/);
     }
   });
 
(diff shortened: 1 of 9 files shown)


packages/util/test/errors.test.ts
```diff
@@ -88,9 +88,9 @@ describe('FirebaseError', () => {
     try {
```
hsubox76, line 88: Is this the only file in util that had errors? There are a few try/catch blocks in util/src so I'm kind of surprised.

hsubox76 (approved):

packages/util/test/errors.test.ts
dwyfrequency: There are more files to change. This is just an initial attempt

PR #6323, opened 2022-06-01
Fix: set exists:false precondition for deleted docs

ehsannas:


ehsannas:
well... that didn't go as i expected heh

packages/firestore/src/core/transaction.ts
```diff
@@ -142,12 +142,18 @@ export class Transaction {
 
   /**
    * Returns the version of this document when it was read in this transaction,
-   * as a precondition, or no precondition if it was not read.
+   * as a precondition, or an existence precondition if the document did not
```
egilmorez, line 145: This kind of list would normally only have one "or,"
so we'd have:

". . . as a precondition, an existence precondition if the document did not exist, or . . . ."

egilmorez (commented):
Sorry to miss this.  Have one suggestion, thanks!

ehsannas:
closing this in favor of https://github.com/firebase/firebase-js-sdk/pull/6550

Issue #6343, opened 2022-06-08
Error: getStream() is only supported by NodeJS builds


dimmetrius:
### [REQUIRED] Describe your environment

  * Operating System version: macOS Monterey 12.4
  * Node version: v16.15.0
  * Firebase SDK version: 9.8.2
  * Firebase Product: storage

#### Steps to reproduce:

can't use storage getStream function in module

![image](https://user-images.githubusercontent.com/2039160/172678344-e492e60c-15a0-4d32-9de8-866b8eafe0fc.png)

#### Relevant Code:

```javascript
//index.js
import {getStream} from "firebase/storage"
getStream()
```

dimmetrius:
https://github.com/dimmetrius/getStreamDemo

PR #6345, opened 2022-06-08
Fix Node ESM bundle to build from Node entry point.

hsubox76:
The Storage Node ESM bundle was incorrectly building from entry point `index.ts` instead of `index.node.ts`

Ran a build and the `index.node.esm.js` compiled bundle no longer contains "getStream() is only supported by NodeJS builds" error string.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6343.

### DIFF
--- a/.changeset/mean-seals-poke.md
+++ b/.changeset/mean-seals-poke.md
@@ -1,0 +1,5 @@
+---
+'@firebase/storage': patch
+---
+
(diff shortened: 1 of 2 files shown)


maneesht (approved):

dwyfrequency (approved):
LGTM

dimmetrius:
wohoo! ESM bundle fixed 🚀

PR #6346, opened 2022-06-09
Use workflow inputs for release branch

hsubox76:
Running this script in the actual release branch causes some issues in that it will use whatever version of this script is currently in the release branch, which might not be the most current. Use workflow inputs instead to get the "to" and "from" branches for this release.


dwyfrequency (approved):
LGTM

PR #6348, opened 2022-06-09
Fix E2E reporting to release tracker

hsubox76:
E2E tests are running fine but there's an issue with the REST call to the release tracker to log the result.

The E2E workflow was recently changed from a `workflow_dispatch` trigger to a `repository_dispatch` trigger. This changes how it receives input from the calling REST request, so the old github env variables no longer work. The correct variable is now passed in to the Node script through env variables to the "Tests succeeded" and "Tests failed" steps.


dwyfrequency (approved):
LGTM

Issue #6080, opened 2022-03-17
v9 `collection()` not working with @firebase/rules-unit-testing "Expected first argument to collection() to be a CollectionReference, a DocumentReference or FirebaseFirestore"


mikob:
* Operating System version: Linux Mint 20.3
  * Browser version: node 16.14.0
  * Firebase SDK version: 
  * "firebase": "^9.6.9",
  * "@firebase/rules-unit-testing": "^2.0.2",
  * Firebase Product: firestore

I am trying to update my unit tests to work with firebase v9. For firebase v8 I had used mocks and dependency injection, but this is harder with v9 as now every method (e.g. collection, getDoc, setDoc) is separated. In comes this guide: https://firebase.google.com/docs/rules/unit-tests I have it partially working with this new rules unit testing library, but it seems to break with a firebase v9 call to firestore `collection()` with this error:

```
  Rejected promise returned by test. Reason:

  FirebaseError {
    code: 'invalid-argument',
    customData: undefined,
    toString: Function {},
    message: 'Expected first argument to collection() to be a CollectionReference, a DocumentReference or FirebaseFirestore',
  }
```

#### Relevant Code:

```typescript
import fs from "fs";
import {
  assertSucceeds,
  initializeTestEnvironment,
  RulesTestEnvironment,
} from "@firebase/rules-unit-testing";
import { collection } from "firebase/firestore";

const EMAILS_BY_PLAN = {
    0: "test@lipsurf.com"
};
const fsTestEnv = await initializeTestEnvironment({
  projectId: "demo-lipsurf",
  firestore: {
    host: "0.0.0.0",
    port: 8080,
    rules: fs.readFileSync("../backend/firebase/firestore.rules", "utf8"),
  },
});
await fsTestEnv.withSecurityRulesDisabled(async (context) => {
  const db = context.firestore();
  for (const plan in EMAILS_BY_PLAN) {
    const email = EMAILS_BY_PLAN[plan];
    const doc = db.collection("users").doc(email);
    await assertSucceeds(
      doc.set({
        plan: +plan,
      })
    );
  }
});

const db = fsTestEnv
  .authenticatedContext(EMAILS_BY_PLAN[FREE_PLAN])
  .firestore();

// This is the line that throws the error
collection(db, "users")
// this works
// db.collection("users")
```

[...]

PR #6360, opened 2022-06-14
Add a Node ESM build to RUT

hsubox76:
Rules-unit-testing only had a CJS build for Node. This causes it to load the CJS build of Firestore, while the user may load methods from an Node ESM build of Firestore, if their project uses ES modules. This causes there to be two copies of Firestore instances that don't match.

Tested on repro: https://github.com/MarkDuckworth/firebase-js-sdk-issues-6080 and it fixes the issue.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6080

### DIFF
--- a/.changeset/nervous-starfishes-crash.md
+++ b/.changeset/nervous-starfishes-crash.md
@@ -1,0 +1,5 @@
+---
+'@firebase/rules-unit-testing': patch
+---
+
+Add Node ESM build to rules-unit-testing.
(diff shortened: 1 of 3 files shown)


dwyfrequency (approved):
LGTM

Jasperhino:
Thanks a lot for the fix!

laurentpayot:
So [my proposed diagnostic/solution](https://github.com/firebase/firebase-js-sdk/issues/6080#issuecomment-1155058810) was right? Thanks for the credit :wink:

PR #6367, opened 2022-06-16
Add function setDefaultEventParameters() to set data that will be logged on every analytics SDK event 

dwyfrequency:
Add function to set data that will be set on every event logged from the SDK, including automatic one.

### DIFF
--- a/.changeset/silly-panthers-mix.md
+++ b/.changeset/silly-panthers-mix.md
@@ -2,4 +2,4 @@
 '@firebase/analytics': minor
 ---
 
-Add function setDefaultEventParameters() to set data that will be logged on every analytics SDK event
+Add function `setDefaultEventParameters()` to set data that will be logged on every analytics SDK event
--- a/.changeset/silly-panthers-mix.md
+++ b/.changeset/silly-panthers-mix.md
@@ -2,4 +2,4 @@
 '@firebase/analytics': minor
 ---
 
-Add function `setDefaultEventParameters()` to set data that will be logged on every "Analytics" SDK event
+Add function `setDefaultEventParameters()` to set data that will be logged on every Analytics SDK event
--- a/.changeset/silly-panthers-mix.md
+++ b/.changeset/silly-panthers-mix.md
@@ -2,4 +2,4 @@
 '@firebase/analytics': minor
 ---
 
-Add function `setDefaultEventParameters()` to set data that will be logged on every analytics SDK event
+Add function `setDefaultEventParameters()` to set data that will be logged on every "Analytics" SDK event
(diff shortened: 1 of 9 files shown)


.changeset/silly-panthers-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/analytics': minor
+---
+
+Add function setDefaultEventParameters() to set data that will be logged on every analytics SDK event
```
hsubox76, line 5: Backticks on `setDefaultEventParameters()`, "Analytics" capitalized.

packages/analytics/src/functions.ts
```diff
@@ -142,3 +148,22 @@ export async function setAnalyticsCollectionEnabled(
...
+ * Sets the variable {@link defaultEventParametersForInit} for use in the initialization of
+ * analytics.
+ *
+ * @param customParams Any custom params the user may pass to gtag.js.
+ */
+export function _setDefaultEventParametersForInit(
+  customParams: CustomParams
+): void {
+  if (defaultEventParametersForInit) {
+    defaultEventParametersForInit = {
```
hsubox76, line 162: I think because this is called "set", developers will expect it to overwrite and not extend any previous object, so we don't need to extend here.

packages/analytics/src/api.ts
```diff
@@ -224,6 +225,23 @@ export function setAnalyticsCollectionEnabled(
     enabled
   ).catch(e => logger.error(e));
 }
+
+/**
+ * Adds data that will be set on every event logged from the SDK, including automatic one.
```
hsubox76, line 230: @egilmorez Can you review the text here? This comment here should be the only text that will end up in the reference docs.

From me:
- one => ones
- In "gtag's set command" the "set" should be in quotes or backticks, I think quotes because it's a string not a var

hsubox76 (commented):
Looks good, just a few comments. Have @egilmorez review reference doc comment text (I tagged him in a comment at the exact spot).

hsubox76 (approved):

packages/analytics/src/api.ts
```diff
@@ -224,6 +225,23 @@ export function setAnalyticsCollectionEnabled(
     enabled
   ).catch(e => logger.error(e));
 }
+
+/**
+ * Adds data that will be set on every event logged from the SDK, including automatic ones.
+ * By using gtag's "set" command, the values passed persist on the current page and are passed with
```
egilmorez, line 231: Strictly speaking the subject of "using" can't really be "the values" -- a correct form would be more like "By using . . . you can ensure that the values passed persist . . ."

I'd suggest "With gtag's set command, the values . . ."

egilmorez (approved):
Just one suggestion, LG, thanks!

PR #6370, opened 2022-06-20
Fix types for setUserId

elliotdickison:
The types for setUserId in the analytics service module do not allow for passing
null, which is the recommended way for clearing out the user id:
https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics#public-void-setuserid-string-id

I updated the types and related documentation to allow null to be passed
and added a test. The underlying setUserId function in the analytics functions
module already allows for null, so this shouldn't result in any
behavioral changes.


elliotdickison:
I believe this change is still valid...

PR #6376, opened 2022-06-22
Add function setConsent() to set end user consent state for web apps in Firebase Analytics

dwyfrequency:
Sets the applicable end user consent state for this web app across all gtag references once Firebase Analytics is initialized.

### DIFF
--- a/packages/analytics/src/api.ts
+++ b/packages/analytics/src/api.ts
@@ -233,7 +233,7 @@ export function setAnalyticsCollectionEnabled(
  * With gtag's "set" command, the values passed persist on the current page and are passed with
  * all subsequent events.
  * @public
- * @param customParams Any custom params the user may pass to gtag.js.
+ * @param customParams - Any custom params the user may pass to gtag.js.
  */
 export function setDefaultEventParameters(customParams: CustomParams): void {
   // Check if reference to existing gtag function on window object exists
@@ -744,7 +744,7 @@ export type CustomEventName<T> = T extends EventNameString ? never : T;
  * Use the {@link ConsentSettings} to specify individual consent type values. By default consent
  * types are set to "granted".
  * @public
- * @param consentSettings Maps the applicable end user consent state for gtag.js.
+ * @param consentSettings - Maps the applicable end user consent state for gtag.js.
  */
 export function setConsent(consentSettings: ConsentSettings): void {
   // Check if reference to existing gtag function on window object exists


packages/analytics/src/functions.ts
```diff
@@ -142,3 +143,27 @@ export async function setAnalyticsCollectionEnabled(
...
+ */
+export let defaultConsentSettingsForInit: ConsentSettings | undefined;
+
+/**
+ * Sets the variable {@link defaultConsentSettingsForInit} for use in the initialization of
+ * analytics.
+ *
+ * @param consentSettings Maps the applicable end user consent state for gtag.js.
+ */
+export function _setConsentDefaultForInit(
```
dwyfrequency, line 158: I'm assuming you'll want me to change this to just overwriting the existing consent details like in https://github.com/firebase/firebase-js-sdk/blob/16b481619e851c2f1577fbc0b3e7d280f3b0087d/packages/analytics/src/functions.ts#L158

packages/analytics/src/initialize-analytics.ts
```diff
@@ -118,6 +119,11 @@ export async function _initializeAnalytics(
     insertScriptTag(dataLayerName, dynamicConfig.measurementId);
   }
 
+  // Detects if there are consent settings that need to be configured.
+  if (defaultConsentSettingsForInit) {
+    gtagCore(GtagCommand.CONSENT, 'default', defaultConsentSettingsForInit);
```
dwyfrequency, line 124: null out defaultConsentSettingsForInit after the command is called

packages/analytics/src/public-types.ts
```diff
@@ -288,4 +288,14 @@ export interface EventParams {
   page_path?: string;
   [key: string]: unknown;
 }
+
+/** Maps the applicable end user consent state. */
+export interface ConsentSettings {
```
dwyfrequency, line 293: There are additional options listed in https://developers.google.com/tag-platform/devguides/consent#tracking_consent_state that are not covered in mobile should I add a generic `[key: string]: string` to accommodate or make it match mobile

hsubox76: Maybe the best way is to add `[key:string]: string`, so it more or less looks like mobile (as opposed to actually adding the additional types) while not locking out developers adding more if they need.

packages/analytics/src/functions.ts
hsubox76: Yes, thanks.

packages/analytics/src/api.ts
```diff
@@ -716,3 +718,21 @@ export function logEvent(
...
 export type CustomEventName<T> = T extends EventNameString ? never : T;
+
+/**
+ * Sets the applicable end user consent state for this web app across all gtag references once
+ * Firebase Analytics is initialized.
+ *
+ * Use the {@link ConsentSettings} to specify individual consent type values. By default consent
+ * types are set to "granted".
+ *
+ * @param consentSettings Maps the applicable end user consent state for gtag.js.
```
hsubox76, line 729: Add a `@public` tag.

packages/analytics/src/public-types.ts
```diff
@@ -288,4 +288,14 @@ export interface EventParams {
   page_path?: string;
   [key: string]: unknown;
 }
+
+/** Maps the applicable end user consent state. */
```
hsubox76, line 292: Suggest: "Consent status settings for each consent type." or similar, @egilmorez may have suggestions. Also add `@public` tag.

packages/analytics/src/public-types.ts
```diff
@@ -288,4 +288,14 @@ export interface EventParams {
...
+export interface ConsentSettings {
+  ad_storage?: ConsentStatusString;
```
hsubox76, line 294: Add documentation above each option, can steal it from here https://developers.google.com/tag-platform/tag-manager/templates/consent-apis#consent_state_and_consent_types

packages/analytics/src/public-types.ts
dwyfrequency: added it as unknown because I see examples in the guide with Arrays, numbers, and strings. Presumably there are quite of few different allowable options

packages/analytics/src/public-types.ts
dwyfrequency: Added external doc link

egilmorez (approved):
Doc strings LGTM, thanks!

packages/analytics/src/helpers.test.ts
```diff
@@ -213,7 +214,11 @@ describe('Gtag wrapping functions', () => {
       expect((window['dataLayer'] as DataLayer).length).to.equal(1);
     });
 
-    it('new window.gtag function does not wait when sending "set" calls', async () => {
+    it('new window.gtag function does not wait when sending "consent" calls', async () => {
```
hsubox76, line 217: Sorry if I'm just misreading this because of the Github diff UI but is this replacing the "set" test? Should it be a copy?

packages/analytics/src/public-types.ts
```diff
@@ -288,4 +288,33 @@ export interface EventParams {
...
+ * {@link https://developers.google.com/tag-platform/tag-manager/templates/consent-apis
+ * | the GA4 reference documentation for consent state and consent types}.
+ * @public
+ */
+export interface ConsentSettings {
+  // Enables storage, such as cookies, related to advertising
+  ad_storage?: ConsentStatusString;
+  // Enables storage, such as cookies, related to analytics (for example, visit duration)
+  analytics_storage?: ConsentStatusString;
+  // Enables storage that supports the functionality of the website or app such as language settings
```
hsubox76, line 304: Looking at the generated analytics.api.md, it looks like api-documenter still thinks these are undocumented. I think api-documenter may need comments to be in the /** */ format.

packages/analytics/src/api.ts
```diff
@@ -734,3 +736,21 @@ export function logEvent(
...
 export type CustomEventName<T> = T extends EventNameString ? never : T;
+
+/**
+ * Sets the applicable end user consent state for this web app across all gtag references once
+ * Firebase Analytics is initialized.
+ *
+ * Use the {@link ConsentSettings} to specify individual consent type values. By default consent
+ * types are set to "granted".
+ * @public
+ * @param consentSettings Maps the applicable end user consent state for gtag.js.
```
hsubox76, line 747: Oops, I just ran `build` on this branch and noticed warnings from api-documenter that `consentSettings` needs to be followed by a hyphen here, and also customParams on line 236, which I think was from a previous PR.

hsubox76 (approved):
LGTM, just address the param tsdoc comment before merging.

PR #6380, opened 2022-06-23
Upload release tags on behalf of google-oss-bot.

yifanyang:
There is another workflow [1] that listens to tag push events.

However, this workflow is not being triggered by new tags
currently, due to the fact that tag push commands
(`git push <tag-name>`) initiated by GitHub Actions bot do not
trigger other workflows [2].

Switching to a personal GitHub account (in our case the
`google-oss-bot`) should fix the problem.

[1] https://github.com/firebase/firebase-js-sdk/blob/master/.github/workflows/health-metrics-release.yml
[2] https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

### DIFF
--- a/.github/workflows/health-metrics-pull-request.yml
+++ b/.github/workflows/health-metrics-pull-request.yml
@@ -1,6 +1,9 @@
 name: Health Metrics
 
-on: [push, pull_request]
+on:
+  push:
+    branches: ['**']
+  pull_request:
 
 env:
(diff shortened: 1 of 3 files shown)


hsubox76 (approved):

dwyfrequency (approved):
LGTM

PR #6385, opened 2022-06-24
Add composite filters in support of OR queries

MarkDuckworth:
Add composite filters. Port of https://github.com/firebase/firebase-android-sdk/pull/3290


packages/firestore/src/lite-api/query.ts
```diff
@@ -858,3 +989,15 @@ function validateOrderByAndInequalityMatch(
     );
   }
 }
+
+export function validateQueryFilterConstraint(
```
ehsannas, line 993: As you mentioned last time, it's not great that we need to do this validation, and points to room for improvement in the API.

packages/firestore/src/lite-api/query.ts
```diff
@@ -193,6 +201,92 @@ export function where(
...
+ * of all the provided `queryConstraints`
+ *
+ * @param queryConstraints - Optional. The {@link queryConstraints} for AND operation. These must be
+ * created with calls to {@link where}, {@link or}, or {@link and}.
+ * @returns The created {@link QueryConstraint}.
+ */
+export function and(...queryConstraints: QueryConstraint[]): QueryConstraint {
+  // Only support QueryFilterConstraints
+  queryConstraints.forEach(queryConstraint =>
+    validateQueryFilterConstraint('andQuery', queryConstraint)
```
ehsannas, line 281: did you figure out the deal with `and` vs `andQuery` ?

packages/firestore/src/lite-api/query.ts
```diff
@@ -193,6 +201,92 @@ export function where(
   return new QueryFilterConstraint(field, op, value);
 }
 
+class QueryCompositeFilterConstraint extends QueryConstraint {
+  constructor(
+    readonly type: 'or' | 'and',
```
ehsannas, line 206: can we use `CompositeOperator` ?

packages/firestore/src/lite-api/query.ts
```diff
@@ -193,6 +201,92 @@ export function where(
...
+
+/**
+ * Creates a {@link QueryConstraint} that performs a logical OR
+ * of all the provided `queryConstraints`
+ *
+ * @param queryConstraints - Optional. The {@link queryConstraints} for OR operation. These must be
+ * created with calls to {@link where}, {@link or}, or {@link and}.
+ * @returns The created {@link QueryConstraint}.
+ */
+export function or(...queryConstraints: QueryConstraint[]): QueryConstraint {
```
ehsannas, line 258: is there a way to _**not**_ expose `and()` and `or()` publicly yet (and still be able to invoke them in the tests) ?

Android makes that possible using `@RestrictTo(RestrictTo.Scope.LIBRARY)`. But not sure if JS has a way to do this.

ehsannas (changes requested):
This is a great start, and generally looks good to me. Some nits/comments outlined below.

packages/firestore/src/core/target.ts
```diff
@@ -685,21 +696,124 @@ export class FieldFilter extends Filter {
...
+      // For conjunctions, all filters must match, so return false if any filter doesn't match.
+      return this.filters.find(filter => !filter.matches(doc)) === undefined;
+    } else {
+      // For disjunctions, at least one filter should match.
+      return this.filters.find(filter => filter.matches(doc)) !== undefined;
+    }
+  }
+
+  getFlattenedFilters(): FieldFilter[] {
+    // TODO(orquery): memoize this result if this method is used more than once
```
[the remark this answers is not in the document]
MarkDuckworth: @ehsannas, I just looked at the implementation for mutability in the Android SDK. Are you not concerned that the `filters` collection is mutable?

packages/firestore/src/core/target.ts
```diff
@@ -685,21 +696,124 @@ export class FieldFilter extends Filter {
...
+      return result.concat(subfilter.getFlattenedFilters());
+    }, result);
+    return result;
+  }
+
+  getFilters(): Filter[] {
+    return this.filters;
+  }
+
+  getOperator(): CompositeOperator {
```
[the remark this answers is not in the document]
MarkDuckworth: done

packages/firestore/src/core/target.ts
```diff
@@ -685,21 +696,124 @@ export class FieldFilter extends Filter {
...
+      filter.field.canonicalString() +
+      filter.op.toString() +
+      canonicalId(filter.value)
+    );
+  } else {
+    // filter instanceof CompositeFilter
+    const canonicalIdsString = filter.filters
+      .map(filter => canonifyFilter(filter))
+      .join(',');
+    const opString = filter.isConjunction() ? 'and' : 'or';
```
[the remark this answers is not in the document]
MarkDuckworth: Good call on using `${filter.op}` directly.

IMHO, it's better to compute the `cannoicalIdsString` in the previous line. That kind of coding improves debuggability. But I'll use your full suggestion if you feel strongly about it.

packages/firestore/src/core/target.ts
```diff
@@ -536,6 +536,12 @@ export function targetGetSegmentCount(target: Target): number {
 
 export abstract class Filter {
   abstract matches(doc: Document): boolean;
+
+  abstract getFlattenedFilters(): FieldFilter[];
```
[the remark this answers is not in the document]
MarkDuckworth: Yes! This would improved the readability/navigation of the code. However, I want to do this refactoring in another PR with no other changes. I'm adding a TODO for this refactoring.

packages/firestore/src/lite-api/query.ts
MarkDuckworth: I'm going to leave this in until we get approval on a different API that doesn't need this. However, I realized that I need to test this validation.

packages/firestore/src/lite-api/query.ts
MarkDuckworth: I was either using the tools incorrectly or missing an export somewhere. But I'm now able to use `and` as the design specified.

packages/firestore/src/lite-api/query.ts
MarkDuckworth: In QueryConstraint, `type` is defined as `abstract readonly type: QueryConstraintType;`.

It looks like I could use `CompositeOperator`, but I don't think mixing the types there is a best practice. Further, the current code is consistent with other QueryConstraint subtypes, such as [QueryLimitConstraint](https://github.com/firebase/firebase-js-sdk/blob/7cad614ec2d2a34b40a3c24443c4f35571e3e68c/packages/firestore/src/lite-api/query.ts#L283).

packages/firestore/src/lite-api/query.ts
MarkDuckworth: My understanding is by keeping it out of [api.ts](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api.ts), it's still internal, but can be used for testing. Please correct me if you see a gap in my understanding.

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
ehsannas: 👍

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
ehsannas: We don't do any mutations on these filters, but yeah using an `ImmutableList<Filter>` would have been better. I'll post a PR for that.

packages/firestore/src/lite-api/query.ts
ehsannas: Ah, yes. Good call.

packages/firestore/src/lite-api/query.ts
ehsannas: Yep; just looked into it a bit more. (as a side, we also have ["lite" APIs](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/lite/index.ts) for the ["Lite Web SDK"](https://firebase.google.com/docs/firestore/solutions/firestore-lite)).

packages/firestore/src/lite-api/query.ts
ehsannas: so I'm guessing you'll replace `andQuery` with `and` here?

packages/firestore/src/lite-api/query.ts
ehsannas: 👍

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
MarkDuckworth: Cool. I'm changing the filters collection to a `ReadonlyArray<Filter>` here in the Web SDK.

packages/firestore/src/lite-api/query.ts
ehsannas: Given that `or` and `and` are currently not public, and that we expect this validateQueryFilterConstraint function to hopefully go away with the new API design, I think you can skip adding tests for it right now.

packages/firestore/src/core/target.ts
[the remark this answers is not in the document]
ehsannas: Assigned a task to you for this.

ehsannas (approved):

common/api-review/firestore-lite.api.md
```diff
@@ -230,7 +230,7 @@ export abstract class QueryConstraint {
 }
 
 // @public
-export type QueryConstraintType = 'where' | 'orderBy' | 'limit' | 'limitToLast' | 'startAt' | 'startAfter' | 'endAt' | 'endBefore';
+export type QueryConstraintType = 'where' | 'orderBy' | 'limit' | 'limitToLast' | 'startAt' | 'startAfter' | 'endAt' | 'endBefore' | 'and' | 'or';
```
ehsannas, line 233: given that we are not exposing 'and' and 'or' publicly, we should remove this update to the .md file

common/api-review/firestore.api.md
```diff
@@ -359,7 +359,7 @@ export abstract class QueryConstraint {
 }
 
 // @public
-export type QueryConstraintType = 'where' | 'orderBy' | 'limit' | 'limitToLast' | 'startAt' | 'startAfter' | 'endAt' | 'endBefore';
+export type QueryConstraintType = 'where' | 'orderBy' | 'limit' | 'limitToLast' | 'startAt' | 'startAfter' | 'endAt' | 'endBefore' | 'and' | 'or';
```
ehsannas, line 362: given that we are not exposing 'and' and 'or' publicly, we should remove this update to the .md file

ehsannas (commented):
Sorry I missed this in my first review. Given that we do want to use 'and' and 'or', but we don't want to expose them publicly, we shouldn't be updating the .md files.

common/api-review/firestore-lite.api.md
ehsannas: oops.. I thought if something is not exported in [api.ts](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/api.ts) then it's not considered public API.

packages/firestore/src/lite-api/query.ts
ehsannas: @MarkDuckworth , IIUC based on @hsubox76 's comment, this comment still stands

common/api-review/firestore-lite.api.md
ehsannas: Ah! Got it. Thanks

MarkDuckworth: Thanks for the feedback, but I don't think there's any value in trying to address this. Although this addition of 'and' and 'or' in `QueryConstraintType` will be public, it does not make this feature usable. In fact, I'm not sure why `QueryConstraintType` is publicly exposed at all, but given that it is, I'm not going to modify that.

So, @hsubox76, @ehsannas, I'm suggesting to move forward with this PR as is.

common/api-review/firestore.api.md
MarkDuckworth: Merging to a feature branch so that we can avoid exposing any public API changes.

common/api-review/firestore-lite.api.md
MarkDuckworth: Merging to a feature branch so we can avoid any public API changes until the feature is released.

PR #6401, opened 2022-06-29
Add debug output for release tags pushing.

yifanyang:
For some reason, the last release still did not trigger the workflow for uploading tags ([history](https://github.com/firebase/firebase-js-sdk/actions/workflows/health-metrics-release.yml)), even after #6380.

The [log](https://github.com/firebase/firebase-js-sdk/runs/7031148329) of the last release job ended at creating tags locally. I would like to add more debugging output to understand what happened while pushing release tags.

### DIFF
--- a/scripts/release/utils/git.ts
+++ b/scripts/release/utils/git.ts
@@ -45,17 +45,29 @@ export async function hasDiff() {
   return !!diff;
 }
 
+// TODO(yifany): remove console log output that is for debugging purpose
 export async function pushReleaseTagsToGithub() {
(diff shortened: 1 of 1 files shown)


yifanyang:
@hsubox76 

It seems that even after #6380, the workflow for uploading tags ([history](https://github.com/firebase/firebase-js-sdk/actions/workflows/health-metrics-release.yml)) was still not triggered during the release last Friday.

I'm adding some log output for debugging purpose. Let's wait for the next release and I'll continue my investigation then.

PR #6404, opened 2022-06-30
Add installations CDN build and "firebase" entry point

hsubox76:
These were not added in v9 because were under the impression that installations should not be a publicly exposed API but guide docs seem to indicate otherwise: https://firebase.google.com/docs/projects/manage-installations

This allows:
- importing `firebase/installations` and `firebase/compat/installations` - previously users could import `@firebase/installations` and `@firebase/installations-compat` directly, but they were not available on the `firebase` entry point (no `@`)
- usage of firebase-installations.js and firebase-installations-compat.js from the gstatic CDN

### DIFF
--- a/.changeset/late-ducks-invite.md
+++ b/.changeset/late-ducks-invite.md
@@ -1,0 +1,5 @@
+---
+'firebase': patch
+---
+
+Add installations CDN build and entry point.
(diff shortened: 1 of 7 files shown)


russellwheatley:
This is great news! We're currently using CDN scripts for Flutterfire, and this is the last piece in the puzzle for us as you can read [here](https://github.com/firebase/flutterfire/pull/9007#issuecomment-1173473475). Do you have any idea on timeframe when this will be released? 😄 Thanks.

hsubox76:
We have a planned release this Thursday, if I am able to get it merged in time, it should go out then.

dwyfrequency (approved):
LGTM

PR #6409, opened 2022-07-03
Potentially fix token error in GitHub Actions

maneesht:
Created a test file to create a pull request from a forked repo.

### DIFF
--- a/test.md
+++ b/test.md
@@ -1,0 +1,3 @@
+# TEST
+
+This is a test to check pipelines.


.github/workflows/label-doc-changes.yml
```diff
@@ -33,4 +33,4 @@ jobs:
       if: ${{steps.check-doc-changes.outputs.DOC_CHANGED == 'true'}}
       with:
         labels: doc-changes
-        github_token: ${{ secrets.OSS_BOT_GITHUB_TOKEN }}
\ No newline at end of file
+        github_token: ${{ secrets.OSS_BOT_GITHUB_TOKEN || github.token }}
```
hsubox76, line 36: The reason we use the OSS_BOT_GITHUB_TOKEN here is that if you use the built-in github token, the result of this workflow will not trigger any other workflows. (https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) This workflow is intended to trigger [assign-tech-writers.yml](https://github.com/firebase/firebase-js-sdk/blob/e5a46540165f73d843c8d0e484e720cdd48ed2fa/.github/workflows/assign-tech-writers.yml) which is why Fei ended up switching to this token.

This isn't the most critical workflow and we can probably risk the tech writers not automatically being added in some cases if it's not frequent. If this (a fork) is the standard workflow then we probably need to think of another solution.

Another possibility is to not assign tech writers in a separate workflow but as an additional step at the end of this one. I'm not sure why Fei originally didn't do that, but my guess is that he wanted the label to always trigger assigning tech writers, even if the label is added manually. I figure if you are consciously assigning the label manually, you probably will remember to add the tech writer. I think this might be the way to go?

hsubox76 (approved):
Thanks!

dwyfrequency (approved):
LGTM

Issue #6133, opened 2022-04-10
RecaptchaVerifier error when used together with AppCheck


faizanabidnaqvi:
### [REQUIRED] Describe your environment

  * Operating System version: MacOS Monterey 12.3
  * Browser version: Chrome version 100.0.4896.75
  * Firebase SDK version: 9.6.10
  * Firebase Product: auth

### [REQUIRED] Describe the problem
When I use the RecaptchaVerifier on my site with AppCheck enabled, I get an error when I call the verify method on the RecaptchaVerifier. The error is as follows:
`
TypeError: this.getAssertedRecaptcha(...).render is not a function
`
This only happens when AppCheck is enabled; without AppCheck it behaves normally, that is, the verification goes through. Also, it doesn't happen in development or production when running at localhost with the emulator. Only occurs in the deployed code in firebase hosting.

<img width="587" alt="Screenshot 2022-04-10 at 6 29 34 PM" src="https://user-images.githubusercontent.com/5749036/162631164-7ab44038-3b32-43c8-bd99-8fd4dd6e380c.png">

#### Steps to reproduce:
1. Create a react app with firebase app check enabled
2. Create a simple page with an invisible RecaptchaVerifier and a button that when clicked calls 'RecaptchaVerifier.verify()'
3. Build the app and deploy to hosting
4. Notice the error occurs when you press the button that calls the verify function
#### Relevant Code:

In my root App.tsx file, appcheck is initialized this way
```typescript
if(!getApps().length){
  const app = initializeApp(firebaseConfig)
  // console.log(`Initializing app check with ${import.meta.env.VITE_RECAPTCHA_KEY}`)
  //if localhost set debug app check token
  //https://firebase.google.com/docs/app-check/web/debug-provider?authuser=0&hl=en
  if (window.location.hostname === "localhost"){
    //this must be set to true to generate a new token and once generated and registered, we need to set the token value in env
    //(self as any).FIREBASE_APPCHECK_DEBUG_TOKEN = true
    (window.self as any).FIREBASE_APPCHECK_DEBUG_TOKEN = import.meta.env.VITE_APP_CHECK_DEBUG_TOKEN
  }

  initializeAppCheck(app, {
    provider: new ReCaptchaEnterpriseProvider(import.meta.env.VITE_RECAPTCHA_KEY),
    isTokenAutoRefreshEnabled: true
  })
}
```

The RecaptchaVerifier component is as follows
```typescript
[...]

PR #6421, opened 2022-07-07
Check for non-Enterprise recaptcha object

hsubox76:
Auth's `ReCaptchaLoaderImpl.load()` method checks for if the ReCAPTCHA script tag has already been loaded by itself or another library, and does not load it again if so. It does this by checking for the global object `window.grecaptcha`. Unfortunately, this object will exist if the ReCAPTCHA Enterprise version of the script has been loaded (which may be done by App Check if the user is using ReCAPTCHA Enterprise with App Check), BUT it will only contain the `enterprise` property, which is a nested version of the regular ReCAPTCHA object.

This change has the `load()` method check instead for the existence of `window.grecaptcha.render`. If it is not found, it will go ahead and load the non-Enterprise script tag, which seems to be able to co-exist with Enterprise and just adds the render/execute/etc methods onto the top level of the object without removing the `enterprise` property.

In the future, a more robust fix could be to extract a shared recaptcha loader to be used by Auth, App Check, and any other library that needs it.

One interesting note: the condition in `shouldResolveImmediately()` is kind of odd, in that it will always resolve if `window.grecaptcha.render` exists, regardless of the other 2 conditions, since `this.librarySeparatelyLoaded` is just also `!!window.grecaptcha.render`. But in the interest of a quick fix and breaking as little as possible, I will leave it be for now.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6133

### DIFF
--- a/packages/auth/src/platform_browser/recaptcha/recaptcha_loader.ts
+++ b/packages/auth/src/platform_browser/recaptcha/recaptcha_loader.ts
@@ -49,7 +49,12 @@ export interface ReCaptchaLoader {
 export class ReCaptchaLoaderImpl implements ReCaptchaLoader {
   private hostLanguage = '';
   private counter = 0;
-  private readonly librarySeparatelyLoaded = !!_window().grecaptcha;
+  /**
+   * Check for `render()` method. `window.grecaptcha` will exist if the Enterprise
+   * version of the ReCAPTCHA script was loaded by someone else (e.g. App Check) but
+   * `window.grecaptcha.render()` will not. Another load will add it.
+   */
+  private readonly librarySeparatelyLoaded = !!_window().grecaptcha?.render;
 
   load(auth: AuthInternal, hl = ''): Promise<Recaptcha> {
     _assert(isHostLanguageValid(hl), auth, AuthErrorCode.ARGUMENT_ERROR);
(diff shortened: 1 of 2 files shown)


sam-gc (approved):
Thanks!

packages/auth/src/platform_browser/recaptcha/recaptcha_loader.ts
```diff
@@ -112,7 +117,7 @@ export class ReCaptchaLoaderImpl implements ReCaptchaLoader {
     // In cases (2) and (3), we _can't_ reload as it would break the recaptchas
     // that are already in the page
     return (
-      !!_window().grecaptcha &&
+      !!_window().grecaptcha?.render &&
```
prameshj, line 120: Sorry for the late comment - `!!_window().grecaptcha?.render` can be replaced with `librarySeparatelyLoaded`, so this becomes:

return (this.librarySeparatelyLoaded && (hl === this.hostLanguage || this.counter > 0 || this.librarySeparatelyLoaded)

Can we just rewrite this as:

`return (hl === this.hostLanguage || this.counter > 0 || this.librarySeparatelyLoaded)`?

hsubox76: See PR description, these 2 expressions were previously redundant as well before this code change, so I thought about fixing it but I didn't know the original intent, and if one of them was originally intended to be a different condition that also needed to be checked. I wanted to leave it up to someone more familiar with the code.

PR #6423, opened 2022-07-07
Forgot to add firebase version bumps for analytics

hsubox76:
Forgot to add `firebase` to minor version bumps for Analytics. We need to manually add it because changesets won't automatically do a minor bump for the main package when dependent packages get a minor bump.


dwyfrequency (approved):
lgtm

Issue #6373, opened 2022-06-22
[AppCheck] Firestore stops working after leaving the website idle overnight - AppCheck HTTP error 403 (appCheck/fetch-status-error) + infinite loop / high CPU usage


anisabboud:
### AppCheck problem

AppCheck works fine and enforced.
However, after leaving the webapp idle for a while (e.g., if I leave the tab open and go to sleep and put the laptop to sleep, then come to work the next day), Firestore stops working (as if it gets disconnected and doesn't reconnect or refresh AppCheck token).

Console error observed in production: `@firebase/app-check: FirebaseError: AppCheck: Fetch server returned an HTTP error status. HTTP status: 403. (appCheck/fetch-status-error).`

Console error observed on localhost: `zone.js:1061 Unhandled Promise rejection: cancelled ; Zone: <root> ; Task: Promise.then ; Value: cancelled undefined`

### Environment

  * Angular version: 14.0.2 (latest)
  * Firebase SDK version: 9.8.3 (latest)
  * AngularFire version: 7.4.1 (latest)
  * Firebase Product: AppCheck

dconeybe:
@anisabboud Would it be possible for you to provide a minimal app to reproduce this issue, that we could use for debugging?

anisabboud:
Hi @dconeybe, I don't currently have time to spin up a test app, but you could reproduce this issue on your own apps using Firestore listeners and AppCheck enforced. Just leave the tab open, put the computer to sleep, and check the console the next day. Try that a few days in a row.

The problem is that it's cumbersome to reproduce - I need to wait for the next day every time...
Today for instance, I encountered a different but related error when doing this - `appCheck/fetch-network-error`: \
`@firebase/app-check: FirebaseError: AppCheck: Fetch failed to connect to a network. Check Internet connection. Original error: Failed to fetch. (appCheck/fetch-network-error).`

The app continued to function, but the console was outputting the message above in an infinite loop (~10,000 times within 5 minutes).

Attaching a partial log showing the infinite loop:
![image](https://user-images.githubusercontent.com/5672108/175889288-f168821f-599a-46d7-8ac8-8aa2524cac58.png)

dconeybe:
I've re-assigned this issue to my colleague @hsubox76 who has far more expertise in the realm of AppCheck. If this turns out to be a Firestore issue, feel free to assign it back to me.

hsubox76:
[...]

PR #6439, opened 2022-07-15
Fix throttling for App Check to prevent unnecessary requests to backend 

dwyfrequency:
Switch check for FirebaseError code to use includes https://github.com/firebase/firebase-js-sdk/issues/6373

### DIFF
--- a/.changeset/cyan-sheep-battle.md
+++ b/.changeset/cyan-sheep-battle.md
@@ -1,5 +1,5 @@
 ---
-'@firebase/app-check': minor
+'@firebase/app-check': patch
 ---
 
 Fix logic to trigger app check throttling
--- a/.changeset/cyan-sheep-battle.md
+++ b/.changeset/cyan-sheep-battle.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app-check': minor
+---
+
+Fix logic to trigger app check throttling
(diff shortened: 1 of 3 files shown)


.changeset/cyan-sheep-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/app-check': minor
```
hsubox76, line 2: I think it should be a patch?

packages/app-check/src/providers.ts
```diff
@@ -80,7 +80,10 @@ export class ReCaptchaV3Provider implements AppCheckProvider {
         this._heartbeatServiceProvider!
       );
     } catch (e) {
-      if ((e as FirebaseError).code === AppCheckError.FETCH_STATUS_ERROR) {
+      console.log({ coder: (e as FirebaseError).code });
```
hsubox76, line 83: Is this left over from debugging?

packages/app-check/src/internal-api.test.ts
```diff
@@ -396,11 +396,9 @@ describe('internal api', () => {
       const warnStub = stub(logger, 'warn');
       stub(client, 'exchangeToken').returns(
         Promise.reject(
-          new FirebaseError(
-            AppCheckError.FETCH_STATUS_ERROR,
-            'test error msg',
-            { httpStatus: 503 }
-          )
+          ERROR_FACTORY.create(AppCheckError.FETCH_STATUS_ERROR, {
```
hsubox76, line 399: Thanks for fixing the test!

hsubox76 (approved):

PR #6440, opened 2022-07-15
Swap out node-fetch with axios in storage sdk

Yuangwang:

### Discussion

Linked issue https://github.com/firebase/firebase-js-sdk/issues/6441
### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/storage/package.json
+++ b/packages/storage/package.json
@@ -45,7 +45,8 @@
     "@firebase/util": "1.6.3",
     "@firebase/component": "0.5.17",
     "node-fetch": "2.6.7",
-    "tslib": "^2.1.0"
+    "tslib": "^2.1.0",
+    "axios": "^0.27.2"
   },
   "peerDependencies": {
     "@firebase/app": "0.x"
--- a/packages/storage/src/platform/node/connection.ts
+++ b/packages/storage/src/platform/node/connection.ts
@@ -71,12 +71,15 @@ abstract class FetchConnection<T extends ConnectionType>
       this.errorCode_ = ErrorCode.NO_ERROR;
       this.body_ = response.data;
     } catch (e) {
-      if (e instanceof AxiosError && (e as AxiosError)?.response?.status === 404) {
-        const axiosError = (e as AxiosError);
+      if (
+        e instanceof AxiosError &&
+        (e as AxiosError)?.response?.status === 404
+      ) {
+        const axiosError = e as AxiosError;
         this.headers_ = axiosError?.response?.headers;
-        this.statusCode_ = (axiosError)?.response?.status;
+        this.statusCode_ = axiosError?.response?.status;
         this.errorCode_ = ErrorCode.NO_ERROR;
-        this.body_ = (axiosError?.response?.data as ArrayBuffer);
+        this.body_ = axiosError?.response?.data as ArrayBuffer;
         return;
       }
       this.errorText_ = (e as Error)?.message;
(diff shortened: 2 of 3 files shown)


packages/storage/src/platform/node/connection.ts
```diff
@@ -59,16 +59,29 @@ abstract class FetchConnection<T extends ConnectionType>
...
+      this.body_ = response.data;
     } catch (e) {
+      if (
+        e instanceof AxiosError &&
+        (e as AxiosError)?.response?.status === 404
+      ) {
+        const axiosError = e as AxiosError;
+        this.headers_ = axiosError?.response?.headers;
+        this.statusCode_ = axiosError?.response?.status;
+        this.errorCode_ = ErrorCode.NO_ERROR;
```
maneesht, line 81: If there was an error here, why are we setting the `errorCode_` to `ErrorCode.NO_ERROR`?

packages/storage/src/platform/node/connection.ts
```diff
@@ -21,7 +21,7 @@ import {
   ErrorCode
 } from '../../implementation/connection';
 import { internalError } from '../../implementation/error';
-import nodeFetch, { Headers } from 'node-fetch';
+import axios, { AxiosResponseHeaders, AxiosError } from 'axios';
```
maneesht, line 24: It looks like you're just using request, can you replace this with `import { AxiosResponseHeaders, AxiosError, request } from 'axios'`? Otherwise we'll be bringing in the entire package it seems.

packages/storage/src/platform/node/connection.ts
```diff
@@ -161,16 +174,30 @@ export class FetchStreamConnection extends FetchConnection<NodeJS.ReadableStream
...
+      this.stream_ = response.data;
     } catch (e) {
+      if (
+        e instanceof AxiosError &&
+        (e as AxiosError)?.response?.status === 404
+      ) {
+        const axiosError = e as AxiosError;
+        this.headers_ = axiosError?.response?.headers;
+        this.statusCode_ = axiosError?.response?.status;
+        this.errorCode_ = ErrorCode.NO_ERROR;
```
maneesht, line 196: Same question here.

packages/storage/src/platform/node/connection.ts
maneesht: Disregard this. It looks like `request` is only available via default import. This is fine.

packages/storage/src/platform/node/connection.ts
Yuangwang: Same as above

packages/storage/src/platform/node/connection.ts
Yuangwang: This specific if statement here is needed to match the behavior of node-fetch with axios - In the case of 404 in node-fetch there was no error thrown and the 404 is just preserved in the response data and errorCode is set to NO_ERROR. Axios throws an error on 404 and some tests were breaking when errorCode is not set to NO_ERROR.

yuchenshi (approved):

packages/storage/package.json
```diff
@@ -45,7 +45,8 @@
     "@firebase/util": "1.6.3",
     "@firebase/component": "0.5.17",
     "node-fetch": "2.6.7",
```
hsubox76, line 47: Seems like this can be removed?

hsubox76:
I wonder if this might be related to https://github.com/firebase/firebase-js-sdk/issues/5848. Although the environment is different (it's Expo/React Native and it's using the browser bundle of Storage instead of the Node bundle), it's also crashing when it tries to send what I suspect is an unusual Blob format. It's sending it through XHR instead of fetch or node-fetch, though. Just curious if using axios instead of XHR would do anything there.

PR #6449, opened 2022-07-19
Update api-documenter to work with new devsite changes

hsubox76:
Devsite has had some changes and requires some different headers and tags for our reference docs as demonstrated in this CL (internal): https://critique.corp.google.com/cl/461499214

Updating api-documenter to provide this for us, Admin, and Functions. When it is approved, I will publish it to NPM so that Admin and Functions can consume the updates. The update will require an additional command line argument, `--project`, which will determine the path string pointing at the `_project.yaml` file.

Also updated JS SDK internal docgen script to call the api-documenter with `--project js`, plus insert the "do not edit" message in the right place given the new changes.


dwyfrequency (approved):
LGTM

taeold (approved):
Tested the change for functions SDK - works great! Thank you.

PR #6450, opened 2022-07-19
Updating pointer to new _project.yaml file from internal changelist 4…

egilmorez:
This is a quick update to the page templates to point to the new _project.yaml file from internal changelist 461499214.


egilmorez:
I still don't understand why rebasing pulls all these changes into a simple PR.  Abandoning.

PR #6451, opened 2022-07-19
Updating pointer to new _project.yaml file from internal changelist 4…

egilmorez:
This is a simple update to point our layout to a new _project.yaml file from internal changelist 461499214.


hsubox76 (approved):

dwyfrequency (approved):
LGTM

PR #6452, opened 2022-07-19
Add firefox cross browser test to select packages

dwyfrequency:
Add Firefox browser to karma and create github action to run tests across Chrome and Firefox in select packages.

### DIFF
--- a/.github/workflows/cross-browser-test.yml
+++ b/.github/workflows/cross-browser-test.yml
@@ -5,13 +5,14 @@ on:
     branches:
       - master
 
+env:
+  # make chromedriver detect installed Chrome version and download the corresponding driver
+  DETECT_CHROMEDRIVER_VERSION: true
+
 jobs:
   cross-browser-test:
-    name: Cross-Browser (Saucelabs) Tests
+    name: Cross-Browser Tests
     runs-on: ubuntu-latest
-    env:
-      SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
-      SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
 
     steps:
       # install Chrome first, so the correct version of webdriver can be installed by chromedriver when setting up the repo
--- a/.github/workflows/cross-browser-test.yml
+++ b/.github/workflows/cross-browser-test.yml
@@ -2,8 +2,7 @@ name: Cross-Browser Test Flow
 
 on:
   push:
-    branches:
-      - master
+    branches: ['**']
 
 env:
   # make chromedriver detect installed Chrome version and download the corresponding driver
--- a/.github/workflows/cross-browser-test.yml
+++ b/.github/workflows/cross-browser-test.yml
@@ -1,0 +1,40 @@
+name: Cross-Browser Test Flow
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  cross-browser-test:
+    name: Cross-Browser (Saucelabs) Tests
+    runs-on: ubuntu-latest
+    env:
+      SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
+      SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
+
+    steps:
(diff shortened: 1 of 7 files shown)


packages/analytics/package.json
```diff
@@ -30,6 +30,7 @@
     "test:all": "run-p test:browser test:integration",
     "test:ci": "node ../../scripts/run_tests_in_ci.js -s test:all",
     "test:browser": "karma start --single-run --nocache",
+    "test:cross-browser": "karma start ./karma.browser.conf.js --single-run --nocache",
```
dwyfrequency, line 33: Can probably remove this because the test script is in the base package

package.json
```diff
@@ -38,6 +38,7 @@
     "test:changed": "ts-node-script scripts/ci-test/test_changed.ts",
     "test:setup": "node tools/config.js",
     "test:saucelabs": "node scripts/run_saucelabs.js",
+    "test:cross-browser": "node scripts/run_cross_browser.js",
```
hsubox76, line 41: If you use `ts-node` instead of `node`, you can make this script a Typescript file and get extra type checking on it. See `format` in line 54 for an example.

.github/workflows/cross-browser-test.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Bump Node memory limit
+        run: echo "NODE_OPTIONS=--max_old_space_size=4096" >> $GITHUB_ENV
+      - name: Test setup and yarn install
+        run: |
+          cp config/ci.config.json config/project.json
+          yarn
+      - name: yarn build
+        run: yarn build
+      - name: Run Cross Browser tests
+        run: xvfb-run yarn test:cross-browser
```
hsubox76, line 40: Does this command run Firefox and Chrome in parallel? If not, maybe it's worth making 2 separate jobs?

scripts/run_cross_browser.js
```diff
@@ -0,0 +1,145 @@
...
+  .help().argv;
+
+// Get all karma.browser.conf.js files that need to be run.
+// runNextTest() pulls filenames one-by-one from this queue.
+const testFiles = configFiles.length
+  ? configFiles
+  : glob.sync(`packages/*/karma.browser.conf.js`);
+
+// Get CI build number or generate one if running locally.
+const buildNumber =
```
hsubox76, line 43: I don't think we need this, this was for Saucelabs, I think.

packages/analytics/karma.browser.conf.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * @license
+ * Copyright 2019 Google LLC
```
hsubox76, line 3: Bump these to current year (also in other files)

packages/app-check/karma.browser.conf.js
```diff
@@ -0,0 +1,33 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const karmaBase = require('../../config/karma.base');
+
+const files = [`**/*.test.ts`];
+
+module.exports = function (config) {
+  config.set({
+    ...karmaBase,
+    browsers: ['Firefox'],
```
hsubox76, line 26: No problem for this iteration, but start thinking about how to dynamically insert a browser here (env variable from the workflow maybe) for when we're able to do Safari, Edge, etc.

dwyfrequency: Inserting the variable seems easy enough with `browsers: process.env?.BROWSERS?.split(',') ?? ['Firefox'],` 

To me the tricky part is installing the corresponding browser in the vm for github, you would need a browser to installation map.

scripts/run_cross_browser.ts
```diff
@@ -0,0 +1,139 @@
...
+    default: [],
+    describe: 'specify individual karma.browser.conf.js files to run on',
+    type: 'array'
+  })
+  .version(false)
+  .help().argv;
+
+// Get all karma.browser.conf.js files that need to be run.
+// runNextTest() pulls filenames one-by-one from this queue.
+const testFiles = configFiles.length
```
dwyfrequency, line 38: Look for package specific, if none exists default to the the karma base

PR #6454, opened 2022-07-19
Update firebase-functions version

hsubox76:
`firebase-functions` needs to be bumped to 3.22.0 to be compatible with `firebase-admin` version 11.0.0. Also might as well bump our testing functions to use Node 16 while we are at it.


dwyfrequency (approved):
LGTM

PR #6469, opened 2022-07-22
Exclude unwanted packages in docgen script

hsubox76:
Add a `toc` command to the docgen script for finer control of calling `api-documenter toc`, so we can temporarily remove `api.json` files for packages we don't want to get into the TOC, including app-compat, util, and rules-unit-testing.

Also added code to remove the actual markdown documentation files of those excluded packages as a post-process step when running regular `api-documenter devsite`. Packages are interdependent on each other in some ways when api-extractor is running, so excluding them before that step might cause unexpected results.

Also the toc.yaml got accidentally checked in at some point. Removed it and added it to .gitignore.

When this is merged we should update https://source.corp.google.com/piper///depot/google3/firebase/jscore/tools/regenerate-reference-latest;l=26 to simply use `yarn docgen toc` instead of calling api-documenter-devsite directly, and we don't need the manual cleanup steps anymore.


egilmorez (approved):
LGTM, thanks!

dwyfrequency (approved):
LGTM

PR #6470, opened 2022-07-22
Allow selecting a custom release branch

hsubox76:
Last TODO in the new release process: allow the release-prod.yml workflow to take a custom release branch name. This should only be needed in very rare cases but we need to add this functionality for the migration to officially be complete.


dwyfrequency (approved):
LGTM

Issue #6471, opened 2022-07-23
[App Check] wrong urls in gstatic file


wliumelb:
https://www.gstatic.com/firebasejs/9.9.0/firebase-app-check.js

line 1408: `const logger = new Logger('https://www.gstatic.com/firebasejs/9.9.0/firebase-app.js-check');`
line 1844: `const name = "https://www.gstatic.com/firebasejs/9.9.0/firebase-app.js-check";`

I got this error message when using app check for web. The wrong url traces back to the above file.
```
https://www.gstatic.com/firebasejs/9.9.0/firebase-app.js-check: FirebaseError: AppCheck: Fetch server returned an HTTP error status. HTTP status: 403. (appCheck/fetch-status-error).
    at exchangeToken (client.ts:75:25)
    at async ReCaptchaV3Provider.getToken (providers.ts:78:16)
```

hsubox76:
I think it's a lack of precision in the search and replace here https://github.com/firebase/firebase-js-sdk/blob/d4af3e2707f99f10981da068b05cdb5696832607/packages/firebase/gulpfile.js#L36 Will make a PR.

PR #6474, opened 2022-07-25
Fix search/replace for CDN bundles

hsubox76:
Make sure it's looking for just the string `@firebase/app` by putting quote capture groups on either side.
Fixes https://github.com/firebase/firebase-js-sdk/issues/6471


dwyfrequency (approved):
LGTM

PR #6475, opened 2022-07-25
Generate v8 docs with correct project paths

hsubox76:
Changed v8 docgen script so that depending on if "node" or "js" is specified as the script is called, the script will modify theme/layouts/default.hbs to point to the correct _project.yaml.

Sample of the generated files:

First 7 lines of firebase.app.App.html for js/v8:
```
<!DOCTYPE html>
<html devsite class="default no-js">
<head>
	<meta charset="utf-8">
	<meta http-equiv="X-UA-Compatible" content="IE=edge">
	<meta name="hide_page_heading" value="true" />
	<meta name="project_path" value="/docs/reference/js/v8/_project.yaml" />
```
First 7 lines of firebase.app.App.html for node:
```
<!DOCTYPE html>
<html devsite class="default no-js">
<head>
	<meta charset="utf-8">
	<meta http-equiv="X-UA-Compatible" content="IE=edge">
	<meta name="hide_page_heading" value="true" />
	<meta name="project_path" value="/docs/reference/node/_project.yaml" />
```

### DIFF
--- a/scripts/docgen-compat/theme/layouts/default.hbs
+++ b/scripts/docgen-compat/theme/layouts/default.hbs
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="hide_page_heading" value="true" />
-    <meta name="project_path" value="/docs/reference/js/v8/_project.yaml" />
+    <meta name="project_path" value="/docs/reference/node/_project.yaml" />
     <meta name="book_path" value="/docs/reference/_book.yaml" />
     <meta name="translation" value="disabled" />
     <meta name="page_type" value="reference" />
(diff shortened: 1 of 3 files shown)


scripts/docgen-compat/theme/layouts/default.hbs
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="hide_page_heading" value="true" />
-    <meta name="project_path" value="/docs/reference/js/_project.yaml" />
+    <meta name="project_path" value="/docs/reference/node/_project.yaml" />
```
hsubox76, line 7: It doesn't matter what this "value" is set to before the script runs, the docgen script will swap it to the correct one when it runs.

egilmorez (approved):
LGTM

I'll go forward and send a couple of CLs to add those project files, thanks!

dwyfrequency (approved):
LGTM

Issue #6456, opened 2022-07-20
Android mobile terminal H5 error...


HChen318:
Error 1
FirebaseError
Firebase: Error thrown when reading from storage. Original error: Firebase: Error thrown when opening storage. Original error: Version change transaction was aborted in upgradeneeded event handler.. (app/storage-open).. (app/storage-get).

error code:  

throw ERROR_FACTORY.create(AppError.STORAGE_GET, {
    originalErrorMessage: (e as Error)?.message
});

Error 2
FirebaseError
Installations: Could not process request. Application offline. (installations/app-offline).

error code: 
const registrationPromiseWithError = Promise.reject(
        ERROR_FACTORY.create(ErrorCode.APP_OFFLINE)
);

Error 3
FirebaseError
Installations: Could not process request. Application offline. (installations/app-offline).

Error 4
ReferenceError
fetch is not defined

error code: 
requiredApisAvailable(): boolean {
if (!fetch || !Promise || !areCookiesEnabled()) {
  consoleLogger.info(
    'Firebase Performance cannot start if browser does not support fetch and Promise or cookie is disabled.'
  );
  return false;
}

rizafran:
Hi @HChen318, please follow the issue template including the needed information so we can have a better understanding of the issue. Thanks.

HChen318:
> Hi @HChen318, please follow the issue template including the needed information so we can have a better understanding of the issue. Thanks.

[REQUIRED] Describe your environment
Operating System version: [Android 6.1.1]
Browser version: Google Chrome 103.0.5060.114
Firebase SDK version: 9.1.1
Firebase Product: firebase Performance

[REQUIRED] Describe the problem
We use firebase to report performance on the H5 page, and some Android models will report errors on the sentry monitoring platform

hsubox76:
- Are you able to reproduce any of these errors locally on your own Android devices?
- Two of the errors say that the device is offline. Probably the device is offline? That seems normal enough?
- One of the errors says that `fetch` is not defined. It's probably being used on a browser where `fetch` is not defined, maybe an old browser. You listed one Google Chrome version above, is it the same version all the errors are being logged from?
- I'm not sure what the first error is, how frequent is it? If it only occurs once it might just be something unusual happening in one case.

HChen318:
> * Are you able to reproduce any of these errors locally on your own Android devices?
> * Two of the errors say that the device is offline. Probably the device is offline? That seems normal enough?
[...]

PR #6480, opened 2022-07-27
Warn instead of throw for idb errors in core app

hsubox76:
IndexedDB is used in firebase/app for platform logging purposes. If there are any issues reading and writing to IDB, they should not throw and crash the rest of the app. Changed all caught failures to warn instead of throw.

Also renaming the error codes because they can be misleading and make it looks like the errors are related to Firebase Storage, the product.

Addresses "Error 1" listed in https://github.com/firebase/firebase-js-sdk/issues/6456

### DIFF
--- a/.changeset/calm-pugs-leave.md
+++ b/.changeset/calm-pugs-leave.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app': patch
+---
+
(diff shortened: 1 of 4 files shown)


dwyfrequency (approved):
LGTM

PR #6491, opened 2022-07-29
compile protos and upload the generated json file

milaGGL:
compiled the proto files added/created and add the proto.json file that is generated.

### DIFF
--- a/packages/firestore/src/protos/protos.json
+++ b/packages/firestore/src/protos/protos.json
@@ -919,6 +919,15 @@
                 "ruby_package": "Google::Cloud::Firestore::V1"
               },
               "nested": {
+                "AggregationResult": {
+                  "fields": {
+                    "aggregateFields": {
+                      "keyType": "string",
+                      "type": "Value",
+                      "id": 2
+                    }
+                  }
+                },
                 "DocumentMask": {
                   "fields": {
                     "fieldPaths": {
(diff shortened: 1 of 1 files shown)


packages/firestore/src/protos/protos.json
```diff
@@ -1760,6 +1792,63 @@
...
+                        "readTime"
+                      ]
+                    }
+                  },
+                  "fields": {
+                    "parent": {
+                      "type": "string",
+                      "id": 1,
+                      "options": {
+                        "(google.api.field_behavior)": "REQUIRED"
```
milaGGL, line 1815: proto3 deprecated the "required". This should be a tag only, wouldn't have any side effects, right? 🤔

milaGGL:
wonder why this file doesn't have compiled proto files in this PR: https://github.com/firebase/firebase-android-sdk/pull/3847/files

CodeLab said it should generate java files when compile protos: https://g3doc.corp.google.com/codelab/protocolbuffers/g3doc/java.md?cl=head#compiler-output:~:text=with%20the%20option%20java_multiple_files%20%3D%20true%2C%20the%20compiler%20generates%20a%20.java%20file%20for%20each%20message%20in%20each%20.

packages/firestore/src/protos/protos.json
dconeybe: I don't know what effects it has, to be honest. But it is _definitely_ different from proto2's "required" keyword. You can just assume it's correct since it's used elsewhere in the protos.

dconeybe:
> wonder why this PR doesn't have compiled proto files in it: https://github.com/firebase/firebase-android-sdk/pull/3847/files

That is because the Android build system will compile the .proto files into .java files at build time. Our TypeScript build setup does not have that feature. Therefore, we need to compile the protos manually and check them into source control.

dconeybe (approved):

PR #6495, opened 2022-08-02
Allow FirstPartyAuth to specify a token factory func.

wu-hui:
Based upon (#4773)

### DIFF
--- a/packages/firestore/src/api/credentials.ts
+++ b/packages/firestore/src/api/credentials.ts
@@ -37,12 +37,15 @@ import { Deferred } from '../util/promise';
 // TODO(mikelehen): This should be split into multiple files and probably
 // moved to an auth/ folder to match other platforms.
 
+export type AuthTokenFactory = () => string;
+
 export interface FirstPartyCredentialsSettings {
   // These are external types. Prevent minification.
   ['type']: 'gapi';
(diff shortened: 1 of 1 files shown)


packages/firestore/src/api/credentials.ts
```diff
@@ -395,17 +398,46 @@ interface Gapi {
 export class FirstPartyToken implements Token {
   type = 'FirstParty' as TokenType;
   user = User.FIRST_PARTY;
-  headers = new Map();
+  private _headers = new Map();
+
+  constructor(
+    private readonly gapi: Gapi,
```
richieforeman, line 404: Could this potentially be nullable?  I think part of the intent here is to reduce the dep on `gapi.js` for first party usage.

(I recognize that making this type nullable may cause a slight ripple effect in this file).

richieforeman (changes requested):

packages/firestore/src/api/credentials.ts
wu-hui: I made it nullable, but I don't think it is of much difference..unless I misunderstood you.

MarkDuckworth (approved):
Looks good.

hsubox76 (approved):

Issue #6397, opened 2022-06-28
unresolved get cause onValue to not be called on child path


christophe-g:
### Describe your environment

  * Operating System version: linux
  * Browser version: chrome latest 
  * Firebase SDK version: JS SDK v9.8.4
  * Firebase Product: 

### Describe the problem

`onValue` callback function on node for which a get on a parent has not yet resoved will never be called. This is pretty bad as `get` seems to have unpredictable side-effect. 

get should not have ANY side-effect. 

Possibly related to https://github.com/firebase/firebase-js-sdk/issues/6038. Took me ages to identify the reason for some part of my app failing while transitionning to v9. 

Work around for now: avoid any use of get, replace by a thenable `onValue ... {onlyOnce: true}`

#### Steps to reproduce:
[...]

PR #6497, opened 2022-08-02
Fix filtered get() cache

maneesht:
Fixed issue where get() would return incorrect results if you had a filtered `get()` and an event listener `onValue()` on the same path, and get() would override the data received by the push received by `onValue()` yielding incorrect results.

The fix below replicates the logic behind `onValue` `onlyOnce` by adding the new event to the SyncTree with the new node, propagating events, and then removing the event from the tree.


maneesht:
This should also fix #6397

.changeset/smart-crabs-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fix filtered get() cache and therefore fixing https://github.com/firebase/firebase-js-sdk/issues/6397
```
IanWyszynski, line 5: I think this text gets put into public release notes and I haven't seen github issues referenced from those in the past.

maneesht: You're right. Updated.

.changeset/smart-crabs-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@firebase/database": patch
+---
+
+Fix filtered get() cache and fixed issue where events weren't getting propagated to listeners.
```
IanWyszynski, line 5: Lets have one line per issue:
```
Fix issue with how get results for filtered queries are added to cache.
Fix issue with events not getting propagated to listeners by get.
```

packages/database/src/core/Repo.ts
```diff
@@ -470,32 +476,42 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
-        // Call `syncTreeRemoveEventRegistration` with a null event registration, since there is none.
-        // Note: The below code essentially unregisters the query and cleans up any views/syncpoints temporarily created above.
-      }
-      const cancels = syncTreeRemoveEventRegistration(
+      /**
+       * Below we simulate the actions of an `onlyOnce` `onValue()` event where:
+       * we add an event registration,
+       * update data at the path,
+       * trigger the events,
+       * and then cleanup the SyncTree
```
IanWyszynski, line 484: nit:
```
       * - Add an event registration
       * - Update data at the path
       * - Raise any events
       * - Cleanup the SyncTree
```

packages/database/src/core/Repo.ts
```diff
@@ -470,32 +476,42 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
+       * We need to raise events in the scenario where `get()` is called at a parent path, and
+       * while the `get()` is pending, `onValue` is called at a child location. While get() is waiting
+       * for the data, `onValue` will register a new event. Then, get() will come back, and update the syncTree
+       * and its corresponding serverCache, including the child location where `onValue` is called. Then,
+       * `onValue` will receive the event from the server, but look at the syncTree and see that the data received
+       * from the server is already at the SyncPoint, and so the `onValue` callback will never get fired.
+       * Calling `eventQueueRaiseEventsForChangedPath()` is the correct way to propagate the events and
+       * ensure the corresponding child events will get fired.
+       */
+      eventQueueRaiseEventsForChangedPath(
```
IanWyszynski, line 502: Why are you calling `eventQueueRaiseEventsForChangedPath` and not ` eventQueueRaiseEventsAtPath`?

It seems like those two functions are the same, do you know why they both exist?

packages/database/src/core/SyncTree.ts
```diff
@@ -347,72 +350,76 @@ export function syncTreeRemoveEventRegistration(
...
-      if (!subtree.isEmpty()) {
-        // We need to fold over our subtree and collect the listeners to send
-        const newViews = syncTreeCollectDistinctViewsForSubTree_(subtree);
+    if (!skipListenerDedup) {
+      // We may have just removed one of many listeners and can short-circuit this whole process
+      // We may also not have removed a default listener, in which case all of the descendant listeners should already be
+      // properly set up.
+      //
+      // Since indexed queries can shadow if they don't have other query constraints, check for loadsAllData(), instead of
+      // queryId === 'default'
```
IanWyszynski, line 363: Let's use a consistent comment style.

packages/database/src/core/SyncTree.ts
```diff
@@ -1049,3 +998,43 @@ function syncTreeSetupListener_(
...
+/**
+ * Added to support non-listener data queries (Used by `get()`).
+ * Simulates `onValue` `onlyOnce` where it adds the event registration,
+ * and then applies the changes to the tree.
+ * @param repo - repo reference to the overall Repo.
+ * @param query - query to be added.
+ * @param node - Data needed to be added to the tree
+ * @param eventRegistration - a dummy event registration to add to the SyncTree
+ */
+export function syncTreeAddGetRegistration(
```
IanWyszynski, line 1011: I think this function is too similarly named to syncTreeAddEventRegistration, which only adds an event registration whereas this one also updates data in the synctree itself. 

IMO we should either rename this to `syncTreeRegisterGetResults` or inline it in `repoGetValue`.

packages/database/src/core/SyncTree.ts
```diff
@@ -863,7 +811,7 @@ function syncTreeCreateListenerForView_(
 /**
  * Return the tag associated with the given query.
  */
-function syncTreeTagForQuery_(
+export function syncTreeTagForQuery_(
```
IanWyszynski, line 814: `syncTreeTagForQuery(` since it's exported now?

packages/database/src/core/Repo.ts
```diff
@@ -470,32 +476,42 @@ export function repoGetValue(repo: Repo, query: QueryContext): Promise<Node> {
...
+        repo.eventQueue_,
+        query._path,
+        events
+      );
+      syncTreeRemoveEventRegistration(
+        // syncTreeAddGetRegistration calls syncTreeAddEventRegistration. We need to clean that up here.
```
IanWyszynski, line 508: I think we should inline syncTreeAddGetRegistration, then we can remove this comment.

packages/database/test/exp/integration.test.ts
```diff
@@ -114,21 +115,20 @@ describe('Database@exp Tests', () => {
...
+      query(readerRef, limitToFirst(2))
     ];
     await Promise.all(
       queries.map(async q => {
         const initial = [{ name: 'child1' }, { name: 'child2' }];
         const ec = EventAccumulatorFactory.waitsForExactCount(1);
-
-        await set(testRef, initial);
-        const unsubscribe = onValue(testRef, snapshot => {
+        const writerPath = getFreshRepo(readerRef._path);
```
IanWyszynski, line 129: Why don't we just:
```
const { readerRef, writerRef } = getRWRefs(getDatabase(defaultApp));
```
above?

packages/database/test/exp/integration.test.ts
```diff
@@ -140,56 +140,82 @@ describe('Database@exp Tests', () => {
...
+      foo1: {
+        a: 1
+      },
+      foo2: {
+        b: 1
+      }
+    });
+    let resolved = false;
+    get(readerRef).then(() => (resolved = true));
+    const childPath = ref(db, readerRef._path + '/foo1');
```
IanWyszynski, line 157: Maybe:
```
const childRef = child(readerRef, "foo1");
```

packages/database/src/core/Repo.ts
maneesht: Good catch. `changedPath` seems to `synchronously raises all events (including previously queued ones) for
  locations related to the specified change path (i.e. all ancestors and descendants)`

maneesht: I think we should keep using `ChangedPath` as we use that in `repoOnDataUpdate`

packages/database/src/core/SyncTree.ts
maneesht: For a two-lined comment, a block comment seems too much. I'll update the comment above though.

packages/database/src/core/SyncTree.ts
maneesht: Inlined instead.

packages/database/test/exp/integration.test.ts
maneesht: I just wanted a fresh writerRef each time in this case, just to be safe.

packages/database/test/exp/integration.test.ts
maneesht: Good catch.

packages/database/test/exp/integration.test.ts
IanWyszynski: Ah I see.

packages/database/test/exp/integration.test.ts
```diff
@@ -114,109 +118,134 @@ describe('Database@exp Tests', () => {
...
+        a: 1
+      },
+      foo2: {
+        b: 1
+      }
+    });
+    let resolved = false;
+    get(readerRef).then(() => (resolved = true));
+    const childPath = child(readerRef, 'foo1');
+    expect(resolved).to.be.false;
```
IanWyszynski, line 161: Lets add a comment explaining why it's difficult to test the behavior we're trying to test and then convert this into a smoke test (just run this in a loop a few times).

IanWyszynski: But we should get rid of this assertion because it can't always pass).

maneesht: SGTM

packages/database/test/exp/integration.test.ts
```diff
@@ -114,109 +118,150 @@ describe('Database@exp Tests', () => {
...
+      },
+      foo2: {
+        b: 1
+      }
+    });
+    await waitUntil(() => {
+      // Because this is a test reliant on network latency, it can be difficult to reproduce. There are situations when get() resolves immediately, and the above behavior is not observed.
+      let resolved = false;
+      get(readerRef).then(() => (resolved = true));
+      return !resolved;
```
IanWyszynski, line 175: Maybe call resolved:
```
      let pending = true;
      get(readerRef).then(() => (pending = false));
      return pending;
```

packages/database/test/exp/integration.test.ts
```diff
@@ -292,6 +337,116 @@ describe('Database@exp Tests', () => {
...
+      readerRef,
+      orderByKey(),
+      startAt('child2')
+    );
+    const expected = {
+      child2: 'test2',
+      child3: 'test3'
+    };
+    const snapshot = await get(otherChildrenQuery);
+    expect(snapshot.val()).to.deep.eq(expected);
```
IanWyszynski, line 408: Did you mean to have another query here?
```
    const firstChildrenQuery = query(
      readerRef,
      orderByKey(),
      endAt('child1')
    );
   onValue(firstChildrenQuery, ...)
```

maneesht: Yes. Nice catch.

packages/database/test/exp/integration.test.ts
```diff
@@ -292,6 +325,116 @@ describe('Database@exp Tests', () => {
...
+  });
+
+  it('should test get with limitTo listener only fires once', async () => {
```
IanWyszynski, line 406: s/should//

packages/database/test/exp/integration.test.ts
```diff
@@ -292,6 +325,116 @@ describe('Database@exp Tests', () => {
...
+      ec.addEvent(snapshot);
+    });
+    const events = await ec.promise;
+    const expected = { child1: 'test1' };
+    expect(events[0].val()).to.deep.eq(expected);
+    const snapshot = await get(readerRef);
+    expect(snapshot.val()).to.deep.eq(toWrite);
+  });
+
+  it('should test startAt get with listener only fires once', async () => {
```
IanWyszynski, line 426: Where is the listener in this test? It seems like we are just testing that get returns the right thing?

IanWyszynski: also s/should//

IanWyszynski (approved):
LGTM. Thanks!

packages/database/test/exp/integration.test.ts
maneesht: Removing this as the test on line 376 should cover it.

packages/database/test/exp/integration.test.ts
maneesht: We should keep this as it blocks should read as sentences

hsubox76 (approved):

jsdt (approved):

PR #6506, opened 2022-08-04
create sample count test cases

milaGGL:
Test cases are failing cause the current COUNT API is only returning dummy count 42.
Screen Shot 2022-08-04 at 10 09 58 AM


dconeybe (approved):

dconeybe:
Looks great. Go ahead and merge. Then... make those tests pass!

PR #6510, opened 2022-08-05
Make index beta

wu-hui:


packages/firestore/src/api/index_configuration.ts
```diff
@@ -35,11 +35,10 @@ export {
   EmulatorMockTokenOptions
 } from '../lite-api/database';
 
-// TODO(indexing): Remove "@internal" from the API.
-
 /**
  * A single field element in an index configuration.
  */
+/** @beta */
```
hsubox76, line 41: I think these should be in the same comment block e.g.

```
/**
 * A single field element in an index configuration.
 * @beta
 */
 ```

Same for other occurrences.

wu-hui: Fixed

hsubox76 (approved):

PR #6522, opened 2022-08-09
Add Firefox testing to Auth

dwyfrequency:


.github/workflows/test-changed-auth.yml
```diff
@@ -7,32 +7,62 @@ env:
...
+      - name: Bump Node memory limit
+        run: echo "NODE_OPTIONS=--max_old_space_size=4096" >> $GITHUB_ENV
+      - name: Test setup and yarn install
+        run: |
+          cp config/ci.config.json config/project.json
+          yarn
+      - name: build
+        run: yarn build:changed auth
+      - name: Run tests on changed packages
+        run: xvfb-run yarn --cwd packages/auth test:browser:unit
```
hsubox76, line 66: It looks like test:changed auth would have run tests in both auth and auth-compat so in order to match, auth-compat should probably be run as well.

packages/auth/src/platform_browser/persistence/session_storage.test.ts
```diff
@@ -63,13 +63,15 @@ describe('platform_browser/persistence/session_storage', () => {
     });
 
     describe('#isAvailable', () => {
+      afterEach(() => sinon.restore());
+
       it('should emit false if sessionStorage setItem throws', async () => {
-        sinon.stub(sessionStorage, 'setItem').throws(new Error('nope'));
-        expect(await persistence._isAvailable()).to.be.false;
+      sinon.stub(Storage.prototype, 'setItem').throws(new Error('nope'));
```
hsubox76, line 69: Glad this worked!

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
```diff
@@ -73,28 +70,28 @@ describe('platform_cordova/popup_redirect/events', () => {
 
   describe('_savePartialEvent', () => {
     it('sets the event', async () => {
+      const spy = sinon.spy(Storage.prototype, 'setItem');
```
hsubox76, line 73: Can you explain why there is a change from stub to spy?

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
```diff
@@ -73,28 +70,28 @@ describe('platform_cordova/popup_redirect/events', () => {
...
       const event = _generateNewEvent(auth, AuthEventType.REAUTH_VIA_REDIRECT);
       await _savePartialEvent(auth, event);
-      expect(storageStub.setItem).to.have.been.calledWith(
-        'firebase:authEvent:test-api-key:test-app',
-        JSON.stringify(event)
-      );
+      spy.calledWith('firebase:authEvent:test-api-key:test-app',
```
hsubox76, line 76: Does this not need to be wrapped in expect()?

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
```diff
@@ -73,28 +70,28 @@ describe('platform_cordova/popup_redirect/events', () => {
...
+      JSON.stringify(event));
```
hsubox76, line 77: What is this for?

hsubox76 (commented):
Can you add a description? That's a good place to explain the plan overall plus any unusual changes (like spy vs stub?)

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
dwyfrequency: I was having issues with the stub at one point and the test only required verifying if the method was called so stubbing out a return value was not necessary for the test itself.

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
dwyfrequency: per https://blog.logrocket.com/storing-retrieving-javascript-objects-localstorage/, `In the code block, we used the JSON.stringify() method to convert our JavaScript object into a string first because we can only store strings in the window.localStorage object.` I didn't add the json stringify part. It was part of the original test.

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
dwyfrequency: Yes, adding it now

.github/workflows/test-changed-auth.yml
dwyfrequency: Updated to accommodate compat

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
hsubox76: Oh, sorry! I didn't see this was a line wrapping, weird indent.

hsubox76: Actually when I pasted it in and ran prettier inside VSCode, it indented it (the current version):

```
      expect(storageStub.calledWith('firebase:authEvent:test-api-key:test-app',
        JSON.stringify(event))).to.be.true;
```

Weird the formatter check didn't catch this? I ran yarn format and it didn't work but it does if I run prettier inside VSCode.

hsubox76 (approved):

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
```diff
@@ -73,30 +70,30 @@ describe('platform_cordova/popup_redirect/events', () => {
...
   describe('_savePartialEvent', () => {
     it('sets the event', async () => {
+      const spy = sinon.spy(Storage.prototype, 'setItem');
       const event = _generateNewEvent(auth, AuthEventType.REAUTH_VIA_REDIRECT);
       await _savePartialEvent(auth, event);
-      expect(storageStub.setItem).to.have.been.calledWith(
-        'firebase:authEvent:test-api-key:test-app',
-        JSON.stringify(event)
-      );
+      expect(spy.calledWith('firebase:authEvent:test-api-key:test-app',
```
sam-gc, line 76: Why is this not using sinonChai anymore?
i.e.
```js
expect(spy).to.have.been.calledWith(...)
```

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
```diff
@@ -73,30 +70,30 @@ describe('platform_cordova/popup_redirect/events', () => {
...
       );
-      storageStub.getItem.returns(event);
+      sinon.stub(Storage.prototype, 'getItem').returns(event);
+      const spy = sinon.spy(Storage.prototype, 'removeItem');
       expect(await _getAndRemoveEvent(auth)).to.eql(JSON.parse(event));
-      expect(storageStub.removeItem).to.have.been.calledWith(
+      expect(spy.calledWith(
         'firebase:authEvent:test-api-key:test-app'
-      );
+      )).to.be.true;
```
sam-gc, line 96: Same here

sam-gc (commented):
We also have webdriver tests, but those are explicitly disabled in Firefox due to weird timing issues that we couldn't resolve. I don't imagine that's fixed now but just wanted to highlight

sam-gc (approved):
Approved pending comments

packages/auth/src/platform_cordova/popup_redirect/events.test.ts
dwyfrequency: Updated to the syntax you mentioned

PR #6525, opened 2022-08-11
Minify CDN builds

hsubox76:
The compat CDN builds are minified but the modular ones are not, and it may have been an oversight. Using terser to minify the firebase-*.js files results in files about 1/4 the size they were before minification. We publish sourcemaps so this shouldn't interfere with developers debugging.

The license text is also embedded in the sourcemaps which should hopefully be enough, because otherwise each file contains many repeated versions of the license text which impacts file size considerably. I couldn't find an easy way to dedupe it. (I tried to use the gulp step, which is used to swap in the gstatic firebase-app.js url for "@firebase/app", to also append a license header, but this throws off the source map considerably and makes it unusable.)


dwyfrequency (approved):
LGTM

PR #6528, opened 2022-08-12
Mila/count implement count query

milaGGL:
# task
task flow: https://taskflow.corp.google.com/workspaces/483058/iterations/2541152?edit=b%2F240311268

# TODO:
- [x] implement getAggregateFromServerDirect function that provides count result
- [x] implement invokeRunAggregationQueryRpc to wire up aggregation query rpc requests
- [x] make the integration test pass for count query without filter

--added limit, filter and their integration test
-- tested out the converter impact on AggregateQuery type as well


packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,115 @@
...
+   * The query on which you called {@link countQuery} in order to get this
+   * `AggregateQuery`.
+   * Query type is set to unknown to avoid error caused by query type converter.
+   * might change it back to T after testing if the error do exist or not
+   */
+  readonly query: Query<unknown>;
 
   /** @hideconstructor */
-  constructor(query: Query<T>) {
+  constructor(query: Query<unknown>) {
```
milaGGL, line 44: use Query<unknown> as proposed.
using<T> might conflict with custom convertor created by user, will test this down the road.

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,115 @@
...
+ * code that does so.
+ */
 export class AggregateQuerySnapshot {
   readonly type = 'AggregateQuerySnapshot';
   readonly query: AggregateQuery;
+  readonly count: number | null;
 
   /** @hideconstructor */
-  constructor(query: AggregateQuery, readonly _count: number) {
+  constructor(query: AggregateQuery, count: number | null) {
```
milaGGL, line 67: make count nullable to be consistent with count query in andriod

packages/firestore/src/remote/datastore.ts
```diff
@@ -232,6 +237,29 @@ export async function invokeRunQueryRpc(
...
+export async function invokeRunAggregationQueryRpc(
+  datastore: Datastore,
+  aggregateQuery: AggregateQuery
+): Promise<ProtoValue[]> {
+  const datastoreImpl = debugCast(datastore, DatastoreImpl);
+  const request = toRunAggregationQueryRequest(
+    datastoreImpl.serializer,
+    queryToTarget(aggregateQuery.query._query)
+  );
+  const response = await datastoreImpl.invokeStreamingRPC<
```
milaGGL, line 249: use streaming RPC as stated in the proto

packages/firestore/src/remote/serializer.ts
```diff
@@ -852,6 +853,72 @@ export function toQueryTarget(
...
 }
 
+export function toRunAggregationQueryRequest(
+  serializer: JsonProtoSerializer,
+  target: Target
+): ProtoRunAggregationQueryRequest {
+  // Dissect the path into parent, collectionId, and optional key filter.
+  const result: ProtoRunAggregationQueryRequest = {
+    structuredAggregationQuery: {
+      aggregations: [
```
milaGGL, line 863: hard code the count field here for now. should be updated down the road when we added aggregateFields to dynamically run aggregate query (count, sum, min....)

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,107 @@
...
+/**
+ * Creates an {@link AggregateQuery} counting the number of documents matching this query.
+ *
+ * @return An {@link AggregateQuery} object that can be used to count the number of documents in
+ * the result set of this query.
+ */
+export function countQuery(query: Query<unknown>): AggregateQuery {
+  /**
+   * TODO(mila): add the "count" aggregateField to the params after the AggregateQuery is updated.
+   */
```
dconeybe, line 84: Remove this comment about "Query type is set to unknown..."

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,107 @@
...
+      /**
+       * while getting aggregation fields from server direct, it should have only
+       * one RunAggregationQueryResponse returned.
+       * But we used streaming rpc here, so we will have an array of
+       * (one, or possibly more) RunAggregationQueryResponse. For this specific
+       * function, we get the first RunAggregationQueryResponse only.
+       */
+      for (const [key, value] of Object.entries(result[0])) {
+        aggregationFields.set(key, userDataWriter.convertValue(value));
+      }
```
dconeybe, line 107: Since you know what you're looking for, just access `result[0].count_alias`. Use `hardAssert()` to fail if it is `=== undefined`. Then, use `hardAssert()` to fail if the value's type is not numeric. Finally, return the resolved promise with the count result.

packages/firestore/src/remote/datastore.ts
```diff
@@ -232,6 +237,29 @@ export async function invokeRunQueryRpc(
...
+    ProtoRunAggregationQueryRequest,
+    ProtoRunAggregationQueryResponse
+  >('RunAggregationQuery', request.parent!, {
+    structuredAggregationQuery: request.structuredAggregationQuery
+  });
+  return (
+    response
+      // Omit RunQueryResponses that only contain readTimes.
```
dconeybe, line 257: Fix this comment to mention the correct type (it currently mentions RunQueryResponses)

packages/firestore/src/remote/datastore.ts
```diff
@@ -232,6 +237,29 @@ export async function invokeRunQueryRpc(
...
+      .filter(proto => !!proto.result && !!proto.result.aggregateFields)
```
dconeybe, line 258: Did you find that you need to check `!!proto.result.aggregateFields`? If it is not strictly necessary, please remove it.

packages/firestore/src/remote/serializer.ts
```diff
@@ -82,7 +82,8 @@ import {
   TargetChangeTargetChangeType as ProtoTargetChangeTargetChangeType,
   Timestamp as ProtoTimestamp,
   Write as ProtoWrite,
-  WriteResult as ProtoWriteResult
+  WriteResult as ProtoWriteResult,
+  RunAggregationQueryRequest as ProtoRunAggregationQueryRequest
```
dconeybe, line 86: nit: Please put this new line in its alphabetically-sorted place.

packages/firestore/src/remote/serializer.ts
```diff
@@ -852,6 +853,68 @@ export function toQueryTarget(
...
 
+export function toRunAggregationQueryRequest(
+  serializer: JsonProtoSerializer,
+  target: Target
+): ProtoRunAggregationQueryRequest {
+  // Dissect the path into parent, collectionId, and optional key filter.
+  const result: ProtoRunAggregationQueryRequest = {
+    structuredAggregationQuery: {
+      aggregations: [
+        //TODO: dynamically add the aggregate fields in future development
```
dconeybe, line 864: Remove this "TODO". It should be "obvious" in the future and is just a bit confusing at this point.

packages/firestore/src/remote/serializer.ts
```diff
@@ -852,6 +853,68 @@ export function toQueryTarget(
...
+        {
+          count: {
+          },
+          alias:"count_alias"
+        },
+      ],
+      structuredQuery: {}
+    }
+  };
+  const path = target.path;
```
dconeybe, line 874: Instead of duplicating all of this logic below, which is quite error-prone for the future, just call `toQueryTarget()` and "extract" the structured query and parent from its return value.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
     });
   });
 });
+
+describe('countQuery()', () => {
```
dconeybe, line 2049: Can you write a simpler converter? Or perhaps re-use an existing converter defined somewhere else? Since the converter is never actually used for anything, keeping it as simple as possible improves readability by removing distractions.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+    { key: 'b', value: 'Bob' },
+    { key: 'c', value: 'Chris' },
+    { key: 'a', value: 'Anna' },
+    { key: 'b', value: 'Bill' },
+    { key: 'c', value: 'Cooper' }
+  ];
+
+  it('empty collection count equals to 0', () => {
+    return withTestCollection(async coll => {
+      const countQuery_ = countQuery(query(coll));
```
dconeybe, line 2074: Remove this test for .type. It is not relevant to this test case.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+    return withTestCollectionAndInitialData(testDocs, async collection => {
+      const countQuery_ = countQuery(query(collection));
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(6);
+    });
+  });
+
+  it('test collection count with filter', () => {
+    return withTestCollectionAndInitialData(testDocs, async collection => {
+      const query_ = query(collection, where('key', '==', 'a'));
```
dconeybe, line 2092: nit: remove this blank line (and other similar blank lines below).

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+
+      const countQuery_ = countQuery(query_);
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(2);
+    });
+  });
+
```
dconeybe, line 2099: "effected" -> "and a". Also on line 2111.

dconeybe, line 2099: Please remove this comment; it is not relevant to the test case.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+    return withTestCollectionAndInitialData(testDocs, async collection => {
+      const query_ = query(collection, where('key', '==', 'a'), limit(3));
+
+      const countQuery_ = countQuery(query_);
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(2);
+    });
+  });
+
+  it('test collection count with converter on query', () => {
```
dconeybe, line 2122: Please remove this comment too, as it is just repeating what the test description already conveys. Also on line 2136.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+      const query_ = query(collection, where('key', '==', 'a')).withConverter(
+        converter
+      );
+
+      const countQuery_ = countQuery(query_);
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(2);
+    });
+  });
+
```
dconeybe, line 2134: This test is unnecessary; it tests the same code paths as the previous test. Please delete this test.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+      const query_2 = query(collection, where('key', '!=', 'a'));
+
+      const countQuery_1 = countQuery(query_1);
+      const countQuery_2 = countQuery(query_2);
+      expect(aggregateQueryEqual(countQuery_1, countQuery_2)).to.be.false;
+    });
+  });
+
+  it('aggregateQuerySnapshotEqual returns true on same queries', () => {
+    return withTestCollectionAndInitialData(testDocs, async collection => {
```
dconeybe, line 2170: Instead of using the naming "original" and "copy", can you use numbers and/or letters? This will help with vertical alignment and improve readability. e.g. name the variables query1A, query1B, query2, etc.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
...
+      const query_1 = query(collection, where('key', '==', 'a'));
+      const query_2 = query(collection, where('key', '!=', 'a'));
+
+      const countQuery_1 = countQuery(query_1);
+      const countQuery_2 = countQuery(query_2);
+      expect(aggregateQueryEqual(countQuery_1, countQuery_2)).to.be.false;
+    });
+  });
+
```
dconeybe, line 2168: Add a test case where `aggregateQuerySnapshotEqual` returns `false`.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,142 @@
     });
   });
 });
+
```
dconeybe, line 2048: Please add these two tests:
1. AggregateQuery.query === the original query
2. AggregateQuerySnapshot.query === the aggregate query

dconeybe (changes requested):
Great work! I've left some feedback.

packages/firestore/src/lite-api/aggregate.ts
milaGGL: With Value type, I cannot read result[0].count_alias directly, so I used a loop to project the result[0] into a map object, then read the count_alias field.
Screen Shot 2022-08-23 at 9 30 41 AM

packages/firestore/src/remote/datastore.ts
milaGGL: Without checking `!!proto.result.aggregateFields`, the return type would be` (Value | undefined)[]`, and I need to map through the returned value in later process to filter `undefined` in the returned array.

Currently, I filtered out the ones doesn't have aggregateFields, so the returned value is either `value[]` or an empty array, no need to worry about undefined in later process.

packages/firestore/test/lite/integration.test.ts
milaGGL: not sure what this means

packages/firestore/src/remote/datastore.ts
```diff
@@ -18,14 +18,18 @@
...
 import { User } from '../auth/user';
 import { Query, queryToTarget } from '../core/query';
+import { AggregateQuery } from '../lite-api/aggregate';
 import { Document } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { Mutation } from '../model/mutation';
 import {
   BatchGetDocumentsRequest as ProtoBatchGetDocumentsRequest,
   BatchGetDocumentsResponse as ProtoBatchGetDocumentsResponse,
+  RunAggregationQueryRequest as ProtoRunAggregationQueryRequest,
```
dconeybe, line 28: This is a `comment`

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -94,25 +83,24 @@ export function getAggregateFromServerDirect(
...
-       */
-      for (const [key, value] of Object.entries(result[0])) {
-        aggregationFields.set(key, userDataWriter.convertValue(value));
-      }
+      hardAssert(
+        result[0] !== undefined,
+        'Aggregation fields are missing from result.'
+      );
+
+      const countField = (result[0] as any).count_alias;
```
dconeybe, line 91: Our linter is complaining about the use of `any` here. What do you think about something like this alternative:

```ts
const counts = Object.entries(result[0])
  .filter(entry => entry[0] == 'count_alias')
  .map(entry => userDataWriter.convertValue(entry[1] as Value));
hardAssert(
  counts.length > 0,
  'count_alias property is missing from result.'
);
const count = counts[0];
```

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,95 @@
...
   readonly query: AggregateQuery;
 
   /** @hideconstructor */
-  constructor(query: AggregateQuery, readonly _count: number) {
+  constructor(query: AggregateQuery, private readonly _count: number) {
     this.query = query;
   }
 
+  /**
+   * @return The result of a document count aggregation. Returns null if no count aggregation is
```
dconeybe, line 59: Can you double-check that @return is a valid documentation string? Take a look at some of our other classes and see how they document the return values.

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,95 @@
...
+ * @return An `AggregateQuery` object that can be used to count the number of documents in
+ * the result set of this query.
+ */
+export function countQuery(query: Query<unknown>): AggregateQuery {
   return new AggregateQuery(query);
 }
 
 export function getAggregateFromServerDirect(
-  query: AggregateQuery
+  aggregateQuery: AggregateQuery
```
dconeybe, line 78: Please rename `aggregateQuery` back to `query`, to be consistent with the API proposal

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,137 @@
...
+  it('AggregateQuery and AggregateQuerySnapshot inherits the original query', () => {
+    return withTestCollection(async coll => {
+      const query_ = query(coll);
+
+      const countQuery_ = countQuery(query_);
+      expect(countQuery_.type).to.equal('AggregateQuery');
+      expect(countQuery_.query).to.equal(query_);
+
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
```
dconeybe, line 2066: Add a check that the AggregateQuerySnapshot object's `query` property equals `countQuery`.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,137 @@
...
+      expect(snapshot.query.query).to.equal(query_);
+    });
+  });
+
+  it('empty test collection count', () => {
+    return withTestCollection(async coll => {
+      const countQuery_ = countQuery(query(coll));
```
dconeybe, line 2073: nit: please remove this empty line

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,137 @@
...
+
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(0);
+    });
+  });
+
```
dconeybe, line 2079: nit: Replace "sample docs" with the number of documents. Also, remove the trailing space character in this string.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2038,3 +2045,137 @@
...
+    { author: 'authorB', title: 'titleC' },
+    { author: 'authorB', title: 'titleD' },
+    { author: 'authorB', title: 'titleE' }
+  ];
+
+  it('AggregateQuery and AggregateQuerySnapshot inherits the original query', () => {
+    return withTestCollection(async coll => {
+      const query_ = query(coll);
+
+      const countQuery_ = countQuery(query_);
```
dconeybe, line 2062: Please remove the `expect` that checks the `type` property; this is an implementation detail that doesn't need to be tested.

dconeybe (changes requested):
Just a few minor cleanups.

packages/firestore/src/lite-api/aggregate.ts
milaGGL: in web SDK,some files used @return, some used @returns. 
based on JSDoc, @returns is the official tag, and @return is a valid synonym.

i will change this to @returns to follow the majority usage in web SDK

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -67,43 +68,42 @@ export class AggregateQuerySnapshot {
...
-        typeof countAggregateResult === 'number',
-        'Count aggeragte field is not a number: ' + countAggregateResult
-      );
-      return Promise.resolve(
-        new AggregateQuerySnapshot(aggregateQuery, countAggregateResult)
-      );
-    }
-  );
+    const count = counts[0];
+    hardAssert(
```
dconeybe, line 96: Can we get rid of this specific hardAssert for `count !== undefined`? It seems to be covered by the hardAssert below for `typeof count === 'number'`.

packages/firestore/src/lite-api/aggregate.ts
```diff
@@ -15,40 +15,95 @@
...
+/**
+ * An `AggregateQuery` computes some aggregation statistics from the result set of
+ * a base `Query`.
+ */
+export class AggregateQuery {
   readonly type = 'AggregateQuery';
-  readonly query: Query<T>;
+  /**
+   * The query on which you called `countQuery` in order to get this `AggregateQuery`.
+   * Query type is set to unknown to avoid error caused by query type converter.
```
dconeybe, line 36: Please remove "Query type is set to unknown to avoid error caused by query type converter..."

dconeybe (changes requested):

packages/firestore/src/lite-api/aggregate.ts
milaGGL: did I keep missing this comment? I thought i removed it.

dconeybe (approved):
Approved. Yay! :tada:

PR #6529, opened 2022-08-12
Run prettier on Auth

hsubox76:
`packages/auth` was included in the `.prettierignore` file (should be the first file in this PR) ever since this SDK was ported to Github. I think it was because Auth was written in Closure at the time. Now that it's in Typescript like everything else, we should probably not be excluding it from prettier checks.

Obviously there is a bit of a backlog after not running it for so long, so there are 102 changed files (all formatting changes - whitespace, indents, splitting params onto multiple lines, etc.)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,4 @@
 # This file is pre-built and need not be formatted
-packages/auth
 packages/firebase/firebase*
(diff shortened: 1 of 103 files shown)


dwyfrequency (approved):
LGTM

PR #6531, opened 2022-08-15
Have CI deploy rules/functions to test project as needed

hsubox76:
Infrequently, someone may change one of the project config files in the `config/` directory (database rules, firestore rules, or cloud functions) which needs to be pushed to the backend.

`yarn test:setup` can be used to push any changes to the test project the developer is using during development, but the CI tests run specifically against the jscore-sandbox project, and the same rules/functions need to be pushed to that project for CI tests to pass.

This detects if any of those files have changed and tries to deploy them efficiently to the backend using the `only` option (so if only firestore rules have changed, it only deploys firestore rules). If `firebase.json` changes, that is considered to require deploying everything.

This needs to complete before any tests run, but since the `build` step of any test workflow often takes a very long time, it should finish before any tests begin.

### DIFF
--- a/scripts/ci-test/deploy-if-needed.ts
+++ b/scripts/ci-test/deploy-if-needed.ts
@@ -43,7 +43,6 @@ const projectConfigGroups = [
   { file: 'config/functions/index.js', flag: 'functions' }
 ];
 
-
 /**
  * Deploy Firebase project config files (functions, rules) to CI
  * test project if there have been any changes to them.
--- a/scripts/ci-test/deploy-if-needed.ts
+++ b/scripts/ci-test/deploy-if-needed.ts
@@ -1,0 +1,96 @@
+/**
+ * @license
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


scripts/ci-test/deploy-if-needed.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Deploy Firebase project config files (functions, rules) to CI
+ * test project if there have been any changes to them.
```
dwyfrequency, line 20: Maybe move this comment to the top of `deployIfNeeded()`

dwyfrequency (approved):
LGTM

scripts/ci-test/deploy-if-needed.ts
hsubox76: Moved.

PR #6538, opened 2022-08-16
Add browser unit tests for firefox in firestore

dwyfrequency:
Add browser unit tests for firefox in firestore
Update build command for firefox in auth

### DIFF
--- a/packages/firestore/test/unit/util/bundle.test.ts
+++ b/packages/firestore/test/unit/util/bundle.test.ts
@@ -241,7 +241,7 @@ function genericBundleReadingTests(bytesPerRead: number): void {
 
       await expect(
         generateBundleAndParse('{metadata: "no length prefix"}', bytesPerRead)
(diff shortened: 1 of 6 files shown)


packages/firestore/test/unit/util/bundle.test.ts
```diff
@@ -241,7 +241,7 @@ function genericBundleReadingTests(bytesPerRead: number): void {
 
       await expect(
         generateBundleAndParse('{metadata: "no length prefix"}', bytesPerRead)
-      ).to.be.rejectedWith('Unexpected end of JSON input');
+      ).to.be.rejectedWith(/(Unexpected end of )(?=.*\JSON\b).*/gi);
```
hsubox76, line 244: I don't think that you need the `\` before the "J", I plugged it into regexr.com and I think that just escapes the "J" which isn't needed.

hsubox76 (approved):
Looks good! Thanks for doing this.

Issue #6368, opened 2022-06-17
rtdb(types): DataSnapshot#forEach - action parameter has null key


rhodgkins:
### [REQUIRED] Describe your environment

  * Operating System version: N/A
  * Browser version: N/A
  * Firebase SDK version: 9.8
  * Firebase Product: database

### [REQUIRED] Describe the problem

When using `DataSnapshot#forEach` the parameter available to the action closure is a `DataSnapshot` which is correct, but the `key` property of this value has the type `string | null`, but as its a child of the iterating snapshot it can't be the root node so its key can never be `null`.

#### Steps to reproduce:

```js
const node = await getDatabase().ref().get()
node.forEach(child => {
   const key = child.key
   // key is possibly null so have to cast etc.
})
```

#### Relevant Code:

https://github.com/firebase/firebase-js-sdk/blob/efe2000fc499e2c85c4e5e0fef6741ff3bad2eb0/packages/database-types/index.d.ts#L25

https://github.com/firebase/firebase-js-sdk/blob/efe2000fc499e2c85c4e5e0fef6741ff3bad2eb0/packages/firebase/compat/index.d.ts#L5821-L5823

https://github.com/firebase/firebase-js-sdk/blob/efe2000fc499e2c85c4e5e0fef6741ff3bad2eb0/packages/database-compat/src/api/Reference.ts#L167

#### Fix

Happy to do a PR for this - I just remember it being a bit complicated [last time](https://github.com/firebase/firebase-js-sdk/issues/6071#issuecomment-1074261971) I did this. Is the info in that issue still correct - the function signature in the above 3 places just needs to be changed?

Would changing the function signature to `forEach(action: (a: DataSnapshot & { key: string }) => boolean | void): boolean;` be OK for a fix?

maneesht:
That syntax doesn't quite match v9, but your suggestions should work. For v9, you still need to update https://github.com/firebase/firebase-js-sdk/blob/d87d3a8b8cd68e757be6628a72538bfd303e78d1/packages/database/src/api/Reference_impl.ts#L397

maneesht:
If you could create a PR addressing this, that would be great

maneesht:
If you're looking at the root, then:
```typescript
import { getDatabase } from 'firebase/database';
const rootSnapshot = await get(ref(getDatabase()));
rootSnapshot.forEach(node => {
  console.log(node.key);
});
```
[...]

PR #6541, opened 2022-08-17
Updated type of action parameter for DataSnapshot#forEach

maneesht:
Restores original change that required the `key` parameter in the `forEach` callback.

As this is a potentially breaking change for extensions, we will wait for a major release for this to be available to users.


maneesht:
Should fix #6368

IanWyszynski (approved):

egilmorez (approved):
LG docs-wise, thanks!

hsubox76 (approved):

Issue #6503, opened 2022-08-04
upgrade @firebase/polyfill core-js


MHDuke:
npm WARN deprecated core-js@3.6.5: 
`core-js@<3.23.3` is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. 

Please, upgrade your dependencies to `>=3.23.3` of core-js.

g-30:
second this

dorji-tshering:
Same issue here. Just installed firebase right now. I hope it will be resolved soon!

MarwanSdeek:
same issue while installing the latest firebase@9.9.2.

playtinum:
Bump

hsubox76:
So the good news is that `core-js` code isn't used in the code for any Firebase SDK products installed through NPM. It was included in `@firebase/polyfill` package which is currently only bundled into the compat CDN builds (where Firebase is included using script tags, and applying only to urls that end with `-compat.js`). We forgot to remove it from the dependencies in package.json (which is why it's installing during yarn/npm install) but the code isn't imported into any Firebase code.

NPM users and users of the v9 CDN scripts shouldn't need to worry. I'll remove `@firebase/polyfill` from the dependencies and you should stop getting that warning after the next release.

As of v9 we've stopped updating `@firebase/polyfill` and expect users to provide their own so I'll also remove the polyfill import from the compat CDNs.

PR #6543, opened 2022-08-18
Remove all references to `@firebase/polyfill` from repo

hsubox76:
We previously provided a `@firebase/polyfill` library for polyfills needed to get the SDK working in older browsers. We stopped providing this upon the v9 release, as official policy on v9 release was to ask users to provide their own polyfills if they need them. It's still included as a dependency in `packages/firebase/package.json` and it's imported in the compat CDN builds. I'm not sure if that was intentional, but since we don't maintain the package any more (it's been deleted from the repo) and can't update it, I don't think we should depend on it.

Removed import in compat CDN and dependency from firebase/package.json. Also updated the version of `firebase` in `e2e/` even though it doesn't matter (it's reinstalled every time the e2e workflow runs), just so `@firebase/polyfill` gets removed from its `yarn.lock` as well.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6503

### DIFF
--- a/.changeset/lovely-apricots-begin.md
+++ b/.changeset/lovely-apricots-begin.md
@@ -1,0 +1,5 @@
+---
+'firebase': patch
+---
+
+Removed all references to `@firebase/polyfill`.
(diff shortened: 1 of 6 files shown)


dwyfrequency (approved):
LGTM

PR #6544, opened 2022-08-18
Fix proactive refresh logic, add some tests

sam-gc:
We noticed the logic in the proactive refresh code was not quite correct (it wasn't turning on for newly-signed-in users). Turns out we also never tested this. D'oh! I've added some tests and walked through the code with Maneesh and I think this puts us in a better spot.


maneesht (approved):
LGTM. Thanks for doing this!

yuchenshi (approved):

PR #6546, opened 2022-08-18
Update Auth Firefox cross browser action to only run on changed auth files

dwyfrequency:
Update Auth Firefox cross browser action to only run on changed auth files. In the previous iteration, we ran the unit tests irrespective of changed auth files which is incongruent with the purpose of the github action

### DIFF
--- a/scripts/run_tests_in_ci.js
+++ b/scripts/run_tests_in_ci.js
@@ -21,6 +21,8 @@ const { spawn } = require('child-process-promise');
 const { writeFileSync } = require('fs');
 
 const LOGDIR = process.env.CI ? process.env.HOME : '/tmp';
+// Maps the packages that have standalone cross browser testing to the necessary test script.
+// Note, the standalone tests are not covered by `test:changed core`
 const crossBrowserPackages = { 'packages/auth': 'test:browser:unit' };
 
 function writeLogs(status, name, logText) {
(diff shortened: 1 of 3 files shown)


scripts/run_tests_in_ci.js
```diff
@@ -21,6 +21,7 @@ const { spawn } = require('child-process-promise');
 const { writeFileSync } = require('fs');
 
 const LOGDIR = process.env.CI ? process.env.HOME : '/tmp';
+const crossBrowserPackages = { 'packages/auth': 'test:browser:unit' };
```
hsubox76, line 24: Add a comment explaining what this object is for.

hsubox76 (commented):
Thanks for doing this! Just add an explaining comment and should be good to go.

scripts/run_tests_in_ci.js
```diff
@@ -21,6 +21,9 @@ const { spawn } = require('child-process-promise');
 const { writeFileSync } = require('fs');
 
 const LOGDIR = process.env.CI ? process.env.HOME : '/tmp';
+// Maps the packages that have standalone cross browser testing to the necessary test script.
```
hsubox76, line 24: It happens to be that we need these for Firestore and Auth, and also that they happen to not be part of test:changed core, but that's more of a coincidence. The distinguishing feature is that these are packages where we should not run `test:all` (or whatever argument is passed to the script using the `-s` flag in the package's version of `test:ci`) for cross-browser tests. Storage basically has the same problem but we solved it by editing karma.conf.js instead. We may actually want to change storage (in the future, in another PR) to use this same method, for consistency.

hsubox76 (approved):

PR #6548, opened 2022-08-20
fix: prevent build error if not using firebase/functions package

piotrostr:
* add a check to prevent the build error for when using Bun version 0.1.10


piotrostr:
Valid point, that could lead to breaking a feature for other Node runtimes. Seems that this is fixed in bun version 0.1.11, anyway, so will close. Thank you, Christina!

PR #6566, opened 2022-08-25
Mila/count add tests

milaGGL:
# tasks:

- [x] add test case for orderBy
- [x] add test case for startAt, startAfter
- [x] add test case for endAt, endBefore
- [x] add test case for collection group
- [x] add test case for running count query on terminated firestore instance

### DIFF
--- a/packages/firestore/test/lite/integration.test.ts
+++ b/packages/firestore/test/lite/integration.test.ts
@@ -2047,11 +2047,13 @@ describe('withConverter() support', () => {
 
 describe('countQuery()', () => {
   const testDocs = [
-    { author: 'authorA', title: 'titleA' },
-    { author: 'authorA', title: 'titleB' },
-    { author: 'authorB', title: 'titleC' },
-    { author: 'authorB', title: 'titleD' },
-    { author: 'authorB', title: 'titleE' }
+    { id: 3 ,author: 'authorA', title: 'titleA'},
+    { id: 4, author: 'authorA', title: 'titleB'},
+    { id: 2 ,author: 'authorB', title: 'titleC'},
+    { id: null,author: 'authorB', title: 'titleD'},
+    { id: 1, author: 'authorB'},
+    { id: 5, author: null, title: 'titleE'},
+
   ];
 
   it('AggregateQuery and AggregateQuerySnapshot inherits the original query', () => {
@@ -2073,11 +2075,11 @@ describe('countQuery()', () => {
     });
   });
 
-  it('test collection count with 5 docs', () => {
+  it('test collection count with 7 docs', () => {
     return withTestCollectionAndInitialData(testDocs, async collection => {
       const countQuery_ = countQuery(query(collection));
       const snapshot = await getAggregateFromServerDirect(countQuery_);
-      expect(snapshot.getCount()).to.equal(5);
+      expect(snapshot.getCount()).to.equal(7);
     });
   });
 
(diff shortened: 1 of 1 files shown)


packages/firestore/test/lite/integration.test.ts
```diff
@@ -2116,6 +2117,45 @@ describe('countQuery()', () => {
     });
   });
 
+  it('count with filter and order by', () => {
```
milaGGL, line 2120: document without the "orderby-ed" field is excluded from from count

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2047,11 +2047,12 @@ describe('withConverter() support', () => {
 
 describe('countQuery()', () => {
   const testDocs = [
-    { author: 'authorA', title: 'titleA' },
-    { author: 'authorA', title: 'titleB' },
-    { author: 'authorB', title: 'titleC' },
-    { author: 'authorB', title: 'titleD' },
-    { author: 'authorB', title: 'titleE' }
+    { id: 3, author: 'authorA', title: 'titleA' },
```
dconeybe, line 2050: I think it's time to get rid of this `testDocs` variable and re-define it in each test case.

Here is the reason: the structure of the documents in this variable have been carefully crafted to suit several different test cases. Minor changes can break one or more test cases inadvertently.

Unlike "normal" code, test code has some different rules. One different rule is that of duplication. In normal code the DRY principle usually applies: "don't repeat yourself". Duplicated logic is best to be factored out into helpers method or shared variables. The same principle does _not_, however, apply for testing code. Test code must be optimized for readability, and code duplication is acceptable when it improves readability. When you have to scroll up a few pages to find the definition of `testDocs` to understand the test, this reduces readability. In this case, it is better to repeat `testDocs` in each test method so that the data is immediately available to the reader. It also localizes any special tweaking to only the relevant test cases.

I like to think of it this way: the correctness of your production code is verified by the tests, but the correctness of the tests is verified by humans via code inspection. Therefore, test code needs to be optimized for human readability to maximize the probability of it being correct.

Take a look at go/unit-test-readability-legacy if you have some time. Not everything in that presentation is relevant, but the guiding principles definitely are.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2116,6 +2117,45 @@ describe('countQuery()', () => {
     });
   });
 
+  it('count with filter and order by', () => {
+    return withTestCollectionAndInitialData(testDocs, async collection => {
+      const query1 = query(
```
dconeybe, line 2122: nit: rename `query1` to `query_`. When the suffix is a number it implies that there is a `query2`, and maybe a `query3` later on; however, in this case, that is not true, which leads to confusion. By simply naming it `query_` the reader can easily, and correctly, assume that it is the only query in scope.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2116,6 +2117,45 @@ describe('countQuery()', () => {
...
+        collection,
+        where('author', '==', 'authorB'),
```
dconeybe, line 2124: Please remove the filter from this test case. The test should be focused on order-by and adding a filter is not relevant and distracts from what the test is actually testing.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2116,6 +2117,45 @@ describe('countQuery()', () => {
...
+        orderBy('title')
+      );
+      const countQuery_ = countQuery(query1);
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(2);
+    });
+  });
+
+  it('count with order by and startAt, startAfter', () => {
```
dconeybe, line 2133: Please separate the tests for startAt and startAfter. Unit tests should just test one thing at a time. Same for endAt and endAfter.

go/unit-testing-practices#behavior-testing-examples

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2128,6 +2168,28 @@ describe('countQuery()', () => {
     });
   });
 
+  it('count query with collection groups', () => {
+    return withTestDb(async db => {
+      const collectionGroupId = doc(collection(db, 'countTest')).id;
+      const docPaths = [
+        `${collectionGroupId}/cg-doc1`,
+        `abc/123/${collectionGroupId}/cg-doc2`,
+        `invalid${collectionGroupId}/cg-doc3`,
```
dconeybe, line 2177: nit: replace "invalid" with something that is obviously filler, like "zzzz". There is nothing "invalid" about having "invalid" in a collection name, so using the word "invalid" incorrectly implies that something is "invalid", which detracts from readabiltiy.

dconeybe (changes requested):

dconeybe (approved):

PR #6572, opened 2022-08-30
Fix MFA endpoint names to be mfa provider-agnostic.

prameshj:
We will have other MFA providers than phone in the near future.


sam-gc (approved):
Thanks!

yuchenshi (approved):

prameshj:
cc @bhparijat

prameshj:
Ignoring firestore test failures above, since those are connectivity-related and this PR does not change any firestore code.

PR #6575, opened 2022-08-31
Mila/count export aggregate to api

milaGGL:
- [x] rename getAggregateFromServerDirect to getAggregate in lite-api
- [x] export aggregate query from lite-api to api
- [x] export and test getAggregateFromServerDirect in api

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -507,7 +507,7 @@ export function firestoreClientRunAggregationQuery(
   client: FirestoreClient,
   query: AggregateQuery
 ): Promise<AggregateQuerySnapshot> {
-  return client.asyncQueue.enqueue(async () => {
+  return client.asyncQueue.enqueue(() => {
     return getAggregate(query);
   });
 }
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -508,7 +508,7 @@ export function firestoreClientRunAggregationQuery(
   query: AggregateQuery
 ): Promise<AggregateQuerySnapshot> {
   return client.asyncQueue.enqueue(async () => {
-    return await getAggregate(query);
+    return getAggregate(query);
   });
 }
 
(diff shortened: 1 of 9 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -501,6 +503,15 @@ export function firestoreClientTransaction<T>(
   return deferred.promise;
 }
 
+export function firestoreClientRunAggregationQuery(
+  client: FirestoreClient,
+  query: AggregateQuery
+): Promise<AggregateQuerySnapshot> {
+  return client.asyncQueue.enqueue(async () => {
```
dconeybe, line 510: Can you remove `async` from the lambda definition? I'm not sure if it's strictly required, so if it's not, let's remove it.

packages/firestore/src/remote/datastore.ts
```diff
@@ -246,12 +246,17 @@ export async function invokeRunAggregationQueryRpc(
     datastoreImpl.serializer,
     queryToTarget(aggregateQuery.query._query)
   );
+
+  //while invoking streaming RPC, grpc_connection requires parent in the request
```
dconeybe, line 250: Please re-write this comment as a "command" to the program (e.g. "Include the parent in the request for grpc_connection...".

Wording comments this way leads to more readable inline comments.

Also, write it as a complete, valid English sentence. So start the sentence with an uppercase letter and end it with a period.

Finally, Add a space between `//` and the first letter of the comment.

dconeybe (changes requested):

packages/firestore/src/remote/datastore.ts
```diff
@@ -247,8 +247,10 @@ export async function invokeRunAggregationQueryRpc(
     queryToTarget(aggregateQuery.query._query)
   );
 
-  //while invoking streaming RPC, grpc_connection requires parent in the request
-  //while rest_connection excludes it
+  /**
+   * Include the parent in the request for grpc_connection.
```
dconeybe, line 251: I think at this point you can simply remove the comment. It is just repeating what the code below is doing. Generally, you only use inline comments to explain the _why_ of the code, not the _what_.

dconeybe (changes requested):

dconeybe (approved):

PR #6577, opened 2022-08-31
change int32 to int64

milaGGL:
change int32 to int64 for up_to

### DIFF
--- a/packages/firestore/src/protos/google/firestore/v1/query.proto
+++ b/packages/firestore/src/protos/google/firestore/v1/query.proto
@@ -309,7 +309,7 @@ message StructuredAggregationQuery {
       // Requires:
       //
       // * Must be greater than zero when present.
-      int64 up_to = 1;
+      google.protobuf.Int64Value up_to = 1;
     }
 
     // The type of aggregation to perform, required.
--- a/packages/firestore/src/protos/google/firestore/v1/query.proto
+++ b/packages/firestore/src/protos/google/firestore/v1/query.proto
@@ -309,7 +309,7 @@ message StructuredAggregationQuery {
       // Requires:
       //
       // * Must be greater than zero when present.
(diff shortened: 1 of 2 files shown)


packages/firestore/src/protos/google/firestore/v1/query.proto
```diff
@@ -309,7 +309,7 @@ message StructuredAggregationQuery {
       // Requires:
       //
       // * Must be greater than zero when present.
-      int32 up_to = 1;
+      int64 up_to = 1;
```
dconeybe, line 312: This has to be `google.protobuf.Int64Value`.

See https://github.com/firebase/firebase-android-sdk/pull/4038

dconeybe (changes requested):
I think you also need to run that command from update.sh to regenerate the code from the changed protos.

packages/firestore/src/protos/google/firestore/v1/query.proto
milaGGL: do we have google.protobuf.Int64Value in web sdk? 
web sdk used google.protobuf.Int32Value, but didn't see any line uses google.protobuf.Int64Value. 🤔

dconeybe: Yes, I believe we do have `google.protobuf.Int64Value` in the web sdk. Even if we don't it's fine because we never use it. But it's important that the proto definition is correct because if we ever do use it in the future then we want it to work correctly. Are you encountering errors when you change it to Int64Value?

dconeybe (approved):

packages/firestore/src/protos/google/firestore/v1/query.proto
milaGGL: nope, no errors

PR #6579, opened 2022-09-01
check client offline state and test

milaGGL:
- [x] throw error if user is offline and calling getAggregatefromServerDirect
- [x] add a test case

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -512,11 +512,6 @@ export function firestoreClientRunAggregationQuery(
   client.asyncQueue.enqueueAndForget(async () => {
     const remoteStore = await getRemoteStore(client);
     if (!canUseNetwork(remoteStore)) {
-      logDebug(
-        LOG_TAG,
-        'The network is disabled. The task returned by ' +
-          "'getAggregateFromServerDirect()' will not complete until the network is enabled."
-      );
       deferred.reject(
         new FirestoreError(
           Code.UNAVAILABLE,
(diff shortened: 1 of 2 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -507,9 +508,31 @@ export function firestoreClientRunAggregationQuery(
...
 ): Promise<AggregateQuerySnapshot> {
-  return client.asyncQueue.enqueue(() => {
-    return getAggregate(query);
+  const deferred = new Deferred<AggregateQuerySnapshot>();
+  client.asyncQueue.enqueueAndForget(async () => {
+    const remoteStore = await getRemoteStore(client);
+    if (!canUseNetwork(remoteStore)) {
+      logDebug(
+        LOG_TAG,
+        'The network is disabled. The task returned by ' +
```
dconeybe, line 517: This debug message is somewhat misleading. First, the `getAggregateFromServerDirect()` function is going to be renamed. Second, it implies that the promise will be automatically resolved once network connectivity is restored. What do you think about just removing this log message altogether?

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -38,4 +39,23 @@ apiDescribe('Aggregation query', (persistence: boolean) => {
       expect(snapshot.getCount()).to.equal(3);
     });
   });
+
+  it('getAggregateFromServerDirect fails if user is offline async', () => {
```
dconeybe, line 43: What does "async" mean in this test name? Can we just remove that word?

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -38,4 +39,23 @@ apiDescribe('Aggregation query', (persistence: boolean) => {
...
+    const testDocs = {
```
dconeybe, line 44: Can we just use an empty collection here? I don't think it matters how many documents are present in the collection, and, if that is the case, then populating the collection somewhat implies that the contents of the collection are meaningful.

dconeybe (changes requested):

dconeybe (approved):

PR #6581, opened 2022-09-02
Mila/count non lite api tests

milaGGL:
add tests for : 

- [x] Make sure that the query property of AggregateQuery is === to the query specified to countQuery().
- [x] Make sure that the query property of AggregateQuerySnapshot is === to the query specified to getAggregateFromServerDirect().
- [x] count query supports collection group query 
- [x]  count query supports query with a converter
- [x] can't run count query on a terminated Firestore.
- [x] terminate Firestore won't crash while there is flying count query

### DIFF
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -512,11 +512,6 @@ export function firestoreClientRunAggregationQuery(
   client.asyncQueue.enqueueAndForget(async () => {
     const remoteStore = await getRemoteStore(client);
     if (!canUseNetwork(remoteStore)) {
-      logDebug(
-        LOG_TAG,
-        'The network is disabled. The task returned by ' +
-          "'getAggregateFromServerDirect()' will not complete until the network is enabled."
-      );
       deferred.reject(
         new FirestoreError(
           Code.UNAVAILABLE,
--- a/packages/firestore/src/core/firestore_client.ts
+++ b/packages/firestore/src/core/firestore_client.ts
@@ -40,6 +40,7 @@ import { toByteStreamReader } from '../platform/byte_stream_reader';
 import { newSerializer, newTextEncoder } from '../platform/serializer';
 import { Datastore } from '../remote/datastore';
 import {
+  canUseNetwork,
   RemoteStore,
   remoteStoreDisableNetwork,
   remoteStoreEnableNetwork,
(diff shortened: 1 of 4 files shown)


packages/firestore/src/core/firestore_client.ts
```diff
@@ -507,9 +508,31 @@ export function firestoreClientRunAggregationQuery(
   client: FirestoreClient,
   query: AggregateQuery
 ): Promise<AggregateQuerySnapshot> {
-  return client.asyncQueue.enqueue(() => {
```
milaGGL, line 510: same function updated in the PR: https://github.com/firebase/firebase-js-sdk/pull/6579/files . copied here to make sure tests will pass

dconeybe: Please make sure to merge the other PR first then.

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -18,24 +18,125 @@
...
+    };
+    return withTestCollection(persistence, testDocs, async collection => {
+      const countQuery_ = countQuery(collection);
+      const snapshot = await getAggregateFromServerDirect(countQuery_);
+      expect(snapshot.getCount()).to.equal(2);
+    });
+  });
+
+  it('aggregateQuery.query equals to original query', () => {
+    const testDocs = {
```
dconeybe, line 52: This is also a case where the actual contents of the collection are not relevant to the test case. For test cases where the contents of the collection don't matter, please just use an empty collection. By using a non-empty collection it distracts the reader from the purpose of the test case. I believe there is even a withEmptyTestCollection() helper method you could use. Please make the same change in other test methods below.

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -18,24 +18,125 @@
...
+        await terminate(firestore);
+        const countQuery_ = countQuery(collection);
+        expect(() => getAggregateFromServerDirect(countQuery_)).to.throw(
+          'The client has already been terminated.'
+        );
+      }
+    );
+  });
+
+  it("terminate doesn't crash when there is flying aggregate query", () => {
```
dconeybe, line 128: nit: "flying" -> "in flight"

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -18,24 +18,125 @@
...
+    const testDocs = {
+      a: { author: 'authorA', title: 'titleA' }
+    };
+    return withTestCollection(
+      persistence,
+      testDocs,
+      async (collection, firestore) => {
+        const countQuery_ = countQuery(collection);
+        getAggregateFromServerDirect(countQuery_).then();
```
dconeybe, line 137: Can you try puting `void` at the beginning of this line, and removing `.then()`? That appears to be the "correct" way to ignore a promise.

packages/firestore/test/lite/integration.test.ts
```diff
@@ -2047,8 +2047,8 @@ describe('withConverter() support', () => {
 
 describe('countQuery()', () => {
   it('AggregateQuery and AggregateQuerySnapshot inherits the original query', () => {
-    return withTestCollection(async coll => {
-      const query_ = query(coll);
+    return withTestCollection(async collection => {
```
dconeybe, line 2050: Avoid naming variables `collection` because that is also the name of a top-level function in the Firestore SDK. That's why `coll` gets used a lot.

dconeybe (changes requested):

packages/firestore/test/integration/api/aggregation.test.ts
milaGGL: Lite-api test helpers has withTestCollection, which is pointed to an empty collection, but api test helpers doesn't have one similar. So created a simple withEmptyTestCollection helper in this PR.

packages/firestore/test/lite/integration.test.ts
milaGGL: I see 🤔

packages/firestore/test/integration/util/helpers.ts
```diff
@@ -255,6 +255,14 @@ export function withTestCollection(
   return withTestCollectionSettings(persistence, DEFAULT_SETTINGS, docs, fn);
 }
 
+export function withEmptyTestCollection(
```
dconeybe, line 258: Can `withEmptyTestCollection()` be written in terms of `withTestCollection()` like this:

```
return withTestCollection(persistence, {}, fn);
```

dconeybe (approved):

PR #6585, opened 2022-09-06
Fix index backfilling frequency.

wu-hui:


packages/firestore/src/local/index_backfiller.ts
```diff
@@ -35,10 +35,10 @@ import { isIndexedDbTransactionError } from './simple_db';
 const LOG_TAG = 'IndexBackiller';
 
 /** How long we wait to try running index backfill after SDK initialization. */
-const INITIAL_BACKFILL_DELAY_MS = 15;
+const INITIAL_BACKFILL_DELAY_MS = 15 * 1000;
```
tom-andersen, line 38: Nit: I think '15_000' is a cleaner way to write this.

tom-andersen (approved):
One nit, that is a personal preference.

ehsannas:
This warrants a changelog entry; it has already been released though.

PR #6594, opened 2022-09-12
Include a reference to AuthInternal in MultiFactorSessionImpl.

prameshj:
This is needed for TOTP support where the function to generate TOTP Secret (by invoking startEnrollment API) needs the Auth reference, but rather than pass in a parameter, we can derive it from the multiFactorSession that is already passed in as a param. This simplifies the TOTP enrollment API for the developer, by requiring one less parameter.

The usage for auth can be seen in this WIP change - https://github.com/firebase/firebase-js-sdk/compare/mfa-totp...mfa-totp-prameshj#diff-e9ae122839df9e758f95dd35db60444dd51b73a319b02a9ef36b747dc9889fbaR81

### Discussion

### Testing

Added new unit tests.

### DIFF
--- a/packages/auth/src/mfa/mfa_session.ts
+++ b/packages/auth/src/mfa/mfa_session.ts
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import { AuthInternal } from '../model/auth';
 import { MultiFactorSession } from '../model/public_types';
 
 export const enum MultiFactorSessionType {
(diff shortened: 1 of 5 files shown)


sam-gc (approved):

packages/auth/src/mfa/mfa_session.ts
```diff
@@ -31,11 +32,19 @@ interface SerializedMultiFactorSession {
 export class MultiFactorSessionImpl implements MultiFactorSession {
   private constructor(
     readonly type: MultiFactorSessionType,
-    readonly credential: string
+    readonly credential: string,
+    readonly auth?: AuthInternal
```
bhparijat, line 36: Wondering if this change need to be documented somewhere or any proto files need to be changed as we added a new param

prameshj: This is an implementation detail, a developer will not directly interact with this field, hence it is not documented.

bhparijat: What about internal customers?

PR #6596, opened 2022-09-13
add DocumentFieldValue

milaGGL:
# Task link
https://taskflow.corp.google.com/workspaces/483058/work_item/b/246381650?origin=iterations%2F2236092&edit=b%2F246381650

### DIFF
--- a/common/api-review/firestore-lite.api.md
+++ b/common/api-review/firestore-lite.api.md
@@ -80,8 +80,11 @@ export function doc(reference: DocumentReference<unknown>, path: string, ...path
 
 // @public
 export interface DocumentData {
-    [field: string]: any;
+    [field: string]: DocumentFieldValue;
 }
+
+// @public
+export type DocumentFieldValue = any;
 
 // @public
 export function documentId(): FieldPath;
--- a/packages/firestore-types/index.d.ts
+++ b/packages/firestore-types/index.d.ts
@@ -17,7 +17,10 @@
 
 import { EmulatorMockTokenOptions } from '@firebase/util';
 
-export type DocumentData = { [field: string]: any };
+/** Alias dynamic document field value types to any */
+export type DocumentFieldValue = any;
+
+export type DocumentData = { [field: string]: DocumentFieldValue };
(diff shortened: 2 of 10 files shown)


packages/firestore-types/index.d.ts
```diff
@@ -17,7 +17,10 @@
 
 import { EmulatorMockTokenOptions } from '@firebase/util';
 
-export type DocumentData = { [field: string]: any };
+/** Alias dynamic document field value types to any */
```
dconeybe, line 20: Remove this comment from here since nothing else is commented.

common/api-review/firestore-lite.api.md
```diff
@@ -80,9 +80,12 @@ export function doc(reference: DocumentReference<unknown>, path: string, ...path
 
 // @public
 export interface DocumentData {
-    [field: string]: any;
+    [field: string]: DocumentFieldValue;
 }
 
+// @public
+export type DocumentFieldValue = any;
```
dconeybe, line 87: Please move the declaration of `DocumentFieldValue` to immediately above `DocumentData`. Make this change here and elsewhere in this PR.

dconeybe (changes requested):
Overall, looks great. A few non-coding things:

1. Remove the TaskFlow link from the PR's description. TaskFlow is a Google internal tool and nobody outside of Google can access it. We generally avoid including links to Google internal things since it's annoying for those outside of Google who cannot access them. Sometimes we'll include b/123123123 links or go/some-cool-doc but they are only for additional context.
2. Re-write the PR title to briefly explain what you are doing. I often like to prefix the PR title with "Firestore:" to make it clear that it's related to Firestore, and not one of the other Firebase products.
3. Create a "change set" so that this change will cause a version bump and a release notes entry. I don't recall the exact steps to do this, but the instructions are in one of the auto-added comments on the PR.

dconeybe:
Ugh, I'm an idiot. We don't need this for COUNT. Please close this PR. Sorry for making you do that extra work.

PR #6604, opened 2022-09-15
Bump main package to match firestore bump

hsubox76:
Amend changeset in https://github.com/firebase/firebase-js-sdk/pull/6107 to make sure main `fiirebase` package gets the same minor bump.


dwyfrequency (approved):
LGTM

PR #6617, opened 2022-09-20
Fix App Check timer issues

hsubox76:
Fixes https://github.com/firebase/firebase-js-sdk/issues/6373

## Causes
Found 2 causes of the problems:
1) `getToken()` needs to return an error result (a dummy token with an `error` property) whenever there is an error. It was assuming that if there was an error there would be no token. This is not always the case. If the token is about to expire, the proactive refresh attempt may run into an error at the exchange endpoint. Then there will be an existing valid token, but also en error.
2) When `getToken()` makes a request to the exchange endpoint, it stores the promise in `state.exchangeTokenPromise` and doesn't make another request if the promise is in flight, instead, any calls during that period await the existing promise. Once the promise is resolved, it clears `state.exchangeTokenPromise`, allowing future exchange requests to be made when needed. The mistake was that it only cleared this when the promise was resolved, and not when the promise was rejected, so that once an exchange request promise failed, another one wouldn't be made until memory was cleared. (This may be why users had to refresh the page.)

## Solutions
1) We need `getToken()` to return a token with a special error property, `internalError`. The reason this needs to be different from `error`, is that there's 2 different places that look for the `error` property on the result returned by `getToken()`. The first is the Refresher `operation` callback (defined inside `createTokenRefresher`), where we need it to find the `error` property (indicating there was an exchange request error) and throw, which will cause the Refresher to go into backoff retry mode. However, the other place that looks at `error` is `notifyTokenListeners` which calls the error callback for any 3P listeners if it finds `error`. If we have a valid token still, and it's just the proactive refresh attempt that's failing, we should just keep sending the token to 3P listeners as usual, it isn't an error from their point of view. That's why we use `internalError` for this case, so `notifyTokenListeners` can ignore it and the Refresher's operation callback can use it and throw, just as it does for `error`.
2) Replaced `then` with `finally`. The `return token` part of the `then` callback is no longer needed as the original promise already returns the token.

## Testing
Tested this using a 30 minute token and taking the tab offline using devtools, after initial token fetch. While offline, the app attempted to make requests to Firestore, and later App Check, after the App Check token had expired, but no more than 1 request per minute (and I think most were triggered by Firestore, which needs to get an App Check token when it tries to connect to the backend). After returning tab to online state, it immediately got a new token, and Firestore was able to write.

### DIFF
--- a/.changeset/selfish-worms-glow.md
+++ b/.changeset/selfish-worms-glow.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app-check': patch
+---
+
+Fix timer issues in App Check that caused the token to fail to refresh after the token expired, or caused rapid repeated requests attempting to do so.
--- a/packages/app-check/src/indexeddb.ts
+++ b/packages/app-check/src/indexeddb.ts
@@ -80,7 +80,7 @@ export function readTokenFromIndexedDB(
 
 export function writeTokenToIndexedDB(
   app: FirebaseApp,
-  token: AppCheckTokenInternal
+  token?: AppCheckTokenInternal
 ): Promise<void> {
   return write(computeKey(app), token);
 }
--- a/packages/app-check/src/internal-api.test.ts
+++ b/packages/app-check/src/internal-api.test.ts
@@ -373,7 +373,8 @@ describe('internal api', () => {
         provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
       });
       setState(app, {
-        ...getState(app), token: {
+        ...getState(app),
+        token: {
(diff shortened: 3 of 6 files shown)


dwyfrequency (approved):
Great work!! Everything looks good to me here

digibake:
Hi @hsubox76 

Do you know if this fix has been published yet? I'm currently running into mega infinite loop App Check problems (we're getting 90K invalid requests constantly every single hour for the last 7 days) and I think this may go a long way to fix them. We're really keen to update and try this fix as soon as possible.

Thanks for all the hard work

Issue #6511, opened 2022-08-06
Mutating data with Firestore multi-tab-persistence from a browser tab which is not the leader tab causes stale data to emit


jorroll:
### [REQUIRED] Describe your environment

  * Operating System version: macOS 12.4 (though the reproduction is running inside a devcontainer)
  * Browser version: Google Chrome 103.0.5060.134
  * Firebase SDK version: 9.9.2
  * Firebase Product: Firestore

### [REQUIRED] Describe the problem

When using the Firestore SDK with `enableMultiTabIndexedDbPersistence()`, if you open two tabs to the app and you perform a mutation in the second tab (i.e. the tab that was opened second) Firestore will emit an update in response to the change which contains the expected data, then it will emit stale data, then it will emit the correct data again. This happens quickly, I think in less than 100ms (possibly _a lot less_ than 100ms). Meanwhile, the first tab will only emit the correct data (as expected). If you perform the same mutation in the first tab, you do not encounter this error (though the second tab still seems to emit stale data briefly).

My theory is that the bug is triggered when enableMultiTabIndexedDbPersistence is turned on and a client tab which is not the Firestore internal "leader tab" attempts to make a change.

Receiving correct data, then stale data, then correct data again is proving very problematic because there doesn't appear to be a good way for the app to tell that the data is stale (other than, perhaps, debouncing Firestore emissions which feels potentially brittle and will slow the UI down). It's possible I'll need to disable `enableMultiTabIndexedDbPersistence` as a work around. ~~I'll also note that, in attempting to diagnose this issue, I _may_ have been able to recreate it when persistence was turned off (still using two tabs), but I'm not positive and, regardless, this reproduction only focuses on the scenerio where multi-tab-persistence is enabled.~~

_Note: I ran into this bug in production, though the included repository (below) reproduces the bug using the emulators._

#### Steps to reproduce:

See this video showing the bug
- https://www.loom.com/share/65b4a560af5348c38f44e82ce4dd9044

Here is a repository that can be cloned to reproduce the bug. Note it has an included `.devcontainer` configuration for VSCode if you're into that sort of thing. 
- https://github.com/jorroll/firestore-issue-2

#### Relevant Code:

See the included video and repository.

dconeybe:
Hi @jorroll. Thank you for taking the time make a thorough bug report. I'll take a look.

dconeybe:
[...]

PR #6619, opened 2022-09-21
Fix document change leads to time travel across multiple tabs

wu-hui:
Fix: https://github.com/firebase/firebase-js-sdk/issues/6511


dconeybe (approved):
LGTM. Nice fix! I don't think I would have ever found this root cause.

PR #6623, opened 2022-09-22
Add a flag allowing api-documenter to sort functions by first param

hsubox76:
Attempting to clean up pages with large groups of functions such as https://firebase.google.com/docs/reference/js/auth by grouping them by first param.

Will gate this behind a flag, `--sort-functions` so that the change will not be forced on other consumers of our api-documenter fork (functions and admin teams).

### DIFF
--- a/repo-scripts/api-documenter/src/cli/MarkdownAction.ts
+++ b/repo-scripts/api-documenter/src/cli/MarkdownAction.ts
@@ -21,10 +21,10 @@
 import { ApiDocumenterCommandLine } from './ApiDocumenterCommandLine';
 import { BaseAction } from './BaseAction';
 import { MarkdownDocumenter } from '../documenters/MarkdownDocumenter';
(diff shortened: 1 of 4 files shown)


repo-scripts/api-documenter/src/cli/MarkdownAction.ts
```diff
@@ -33,10 +35,24 @@ export class MarkdownAction extends BaseAction {
...
 
+  protected onDefineParameters(): void {
+    super.onDefineParameters();
+
+    this._sortFunctions = this.defineStringParameter({
+      parameterLongName: '--sort-functions',
+      argumentName: 'PRIORITY_PARAMS',
+      description:
+        `Sorts functions tables and listings by first parameter.` +
+        ` Provide comma-separated strings for preferred params to be ` +
```
dwyfrequency, line 46: not required but I would remove the space before here and add it onto the end of the line above.

dwyfrequency (approved):
LGTM

PR #6624, opened 2022-09-22
Fix Firestore failing to return empty results from the local cache

### LINKED ISSUE #5873: Collection snapshot queries with no documents don't store anything in cache. Always requires server snapshot.

milaGGL:
Fix a bug where Firestore fails to return results from the local cache if the result set was empty.

This bug is due to the following logic in `event_manager.ts`:

https://github.com/firebase/firebase-js-sdk/blob/4ddf5566ca52b1f9ef776dd647c7d514e3b3f9f5/packages/firestore/src/core/event_manager.ts#L374-L375 

which assumes that if the result set from the local cache is empty that there is no cached result; however, if the result set of the query is indeed empty then it should be returning that empty result set from the cache.

This fix improves the logic to use the presence of a hasCachedResults flag to indicate that the empty result set is cached data and should be raised to the client.

Fixed #5873

### DIFF
--- a/.changeset/cool-grapes-attend.md
+++ b/.changeset/cool-grapes-attend.md
@@ -1,0 +1,5 @@
+---
+'@firebase/firestore': patch
+---
+
+Fix Firestore failing to raise initial snapshot from empty local cache result
--- a/.changeset/cool-grapes-attend.md
+++ b/.changeset/cool-grapes-attend.md
@@ -1,5 +1,6 @@
 ---
 '@firebase/firestore': patch
+'firebase': patch
 ---
 
 Fix Firestore failing to raise initial snapshot from empty local cache result
--- a/packages/firestore/src/core/view.ts
+++ b/packages/firestore/src/core/view.ts
@@ -26,7 +26,6 @@ import { DocumentKey } from '../model/document_key';
 import { DocumentSet } from '../model/document_set';
 import { TargetChange } from '../remote/remote_event';
 import { debugAssert, fail } from '../util/assert';
-import { ByteString } from '../util/byte_string';
 
 import { LimitType, newQueryComparator, Query, queryMatches } from './query';
 import { OnlineState } from './types';
@@ -73,7 +72,7 @@ export interface ViewChange {
  */
 export class View {
   private syncState: SyncState | null = null;
-  private resumeToken: ByteString | null = null;
+  private hasCachedResults: boolean = false;
   /**
    * A flag whether the view is current with the backend. A view is considered
    * current after it has seen the current flag from the backend and did not
@@ -322,7 +321,8 @@ export class View {
         newSyncState === SyncState.Local,
         syncStateChanged,
         /* excludesMetadataChanges= */ false,
-        targetChange?.resumeToken ?? ByteString.EMPTY_BYTE_STRING
+        targetChange !== undefined &&
+          targetChange.resumeToken.approximateByteSize() > 0
       );
       return {
         snapshot: snap,
@@ -472,7 +472,7 @@ export class View {
       this.documentSet,
       this.mutatedKeys,
       this.syncState === SyncState.Local,
-      this.resumeToken ?? ByteString.EMPTY_BYTE_STRING
+      this.hasCachedResults
     );
   }
 }
--- a/packages/firestore/src/core/view.ts
+++ b/packages/firestore/src/core/view.ts
@@ -321,8 +321,9 @@ export class View {
         newSyncState === SyncState.Local,
         syncStateChanged,
         /* excludesMetadataChanges= */ false,
-        targetChange !== undefined &&
-          targetChange.resumeToken.approximateByteSize() > 0
+        targetChange
+          ? targetChange.resumeToken.approximateByteSize() > 0
+          : false
       );
       return {
         snapshot: snap,
--- a/packages/firestore/src/core/view.ts
+++ b/packages/firestore/src/core/view.ts
@@ -26,6 +26,7 @@ import { DocumentKey } from '../model/document_key';
 import { DocumentSet } from '../model/document_set';
 import { TargetChange } from '../remote/remote_event';
 import { debugAssert, fail } from '../util/assert';
+import { ByteString } from '../util/byte_string';
 
 import { LimitType, newQueryComparator, Query, queryMatches } from './query';
 import { OnlineState } from './types';
@@ -72,6 +73,7 @@ export interface ViewChange {
  */
 export class View {
   private syncState: SyncState | null = null;
+  private resumeToken: ByteString | null = null;
   /**
    * A flag whether the view is current with the backend. A view is considered
    * current after it has seen the current flag from the backend and did not
@@ -319,7 +321,8 @@ export class View {
         docChanges.mutatedKeys,
         newSyncState === SyncState.Local,
         syncStateChanged,
-        /* excludesMetadataChanges= */ false
+        /* excludesMetadataChanges= */ false,
+        targetChange?.resumeToken ?? ByteString.EMPTY_BYTE_STRING
       );
       return {
         snapshot: snap,
@@ -468,7 +471,8 @@ export class View {
       this.query,
       this.documentSet,
       this.mutatedKeys,
-      this.syncState === SyncState.Local
+      this.syncState === SyncState.Local,
+      this.resumeToken ?? ByteString.EMPTY_BYTE_STRING
     );
   }
 }
--- a/packages/firestore/src/core/view_snapshot.ts
+++ b/packages/firestore/src/core/view_snapshot.ts
@@ -20,7 +20,6 @@ import { Document } from '../model/document';
 import { DocumentKey } from '../model/document_key';
 import { DocumentSet } from '../model/document_set';
 import { fail } from '../util/assert';
-import { ByteString } from '../util/byte_string';
 import { SortedMap } from '../util/sorted_map';
 
 import { Query, queryEquals } from './query';
@@ -148,7 +147,7 @@ export class ViewSnapshot {
     readonly fromCache: boolean,
     readonly syncStateChanged: boolean,
     readonly excludesMetadataChanges: boolean,
-    readonly resumeToken: ByteString
+    readonly hasCachedResults: boolean
   ) {}
 
   /** Returns a view snapshot as if all documents in the snapshot were added. */
@@ -157,7 +156,7 @@ export class ViewSnapshot {
     documents: DocumentSet,
     mutatedKeys: DocumentKeySet,
     fromCache: boolean,
-    resumeToken: ByteString
+    hasCachedResults: boolean
   ): ViewSnapshot {
     const changes: DocumentViewChange[] = [];
     documents.forEach(doc => {
@@ -173,7 +172,7 @@ export class ViewSnapshot {
       fromCache,
       /* syncStateChanged= */ true,
       /* excludesMetadataChanges= */ false,
-      resumeToken
+      hasCachedResults
     );
   }
 
--- a/packages/firestore/src/core/view_snapshot.ts
+++ b/packages/firestore/src/core/view_snapshot.ts
@@ -188,8 +188,7 @@ export class ViewSnapshot {
       !this.mutatedKeys.isEqual(other.mutatedKeys) ||
       !queryEquals(this.query, other.query) ||
       !this.docs.isEqual(other.docs) ||
-      !this.oldDocs.isEqual(other.oldDocs) ||
-      this.resumeToken !== other.resumeToken
+      !this.oldDocs.isEqual(other.oldDocs)
     ) {
       return false;
     }
--- a/packages/firestore/src/core/view_snapshot.ts
+++ b/packages/firestore/src/core/view_snapshot.ts
@@ -183,6 +183,7 @@ export class ViewSnapshot {
   isEqual(other: ViewSnapshot): boolean {
     if (
       this.fromCache !== other.fromCache ||
+      this.hasCachedResults !== other.hasCachedResults ||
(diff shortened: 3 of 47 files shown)


dconeybe:
Please add a changeset so an entry will be included in the release notes.

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -1288,6 +1288,93 @@ apiDescribe('Queries', (persistence: boolean) => {
       expect(toDataArray(snapshot)).to.deep.equal([{ map: { nested: 'foo' } }]);
     });
   });
+
+  // Reproduces https://github.com/firebase/firebase-js-sdk/issues/5873
+  // eslint-disable-next-line no-restricted-properties
+  (persistence ? describe : describe.skip)('Caching empty results ', () => {
```
dconeybe, line 1294: nit: Remove the trailing space character in `'Caching empty results '`

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -1288,6 +1288,93 @@ apiDescribe('Queries', (persistence: boolean) => {
...
+    it('can empty cached collection and raise snapshot from it', () => {
+      const testDocs = {
+        a: { key: 'a' }
+      };
+      return withTestCollection(persistence, testDocs, async coll => {
+        // Populate the cache
+        const snapshot1 = await getDocs(coll);
+        expect(snapshot1.metadata.fromCache).to.be.false;
+        expect(toDataArray(snapshot1)).to.deep.equal([{ key: 'a' }]);
+        //empty the collection
```
dconeybe, line 1319: nit: `//empty` -> `// Empty` (add a space and uppercase the "e" in "empty")

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -1288,6 +1288,93 @@ apiDescribe('Queries', (persistence: boolean) => {
       expect(toDataArray(snapshot)).to.deep.equal([{ map: { nested: 'foo' } }]);
     });
   });
+
+  // Reproduces https://github.com/firebase/firebase-js-sdk/issues/5873
+  // eslint-disable-next-line no-restricted-properties
+  (persistence ? describe : describe.skip)('Caching empty results ', () => {
+    it('can cache empty query results', () => {
```
dconeybe, line 1295: Please rename this test to something like "Raises initial snapshot from cache, even if it is empty"

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -1288,6 +1288,93 @@ apiDescribe('Queries', (persistence: boolean) => {
...
+        // Add a snapshot listener whose first event should be raised from cache.
+        const storeEvent = new EventsAccumulator<QuerySnapshot>();
+        onSnapshot(coll, storeEvent.storeEvent);
+        const snapshot2 = await storeEvent.awaitEvent();
+        expect(snapshot2.metadata.fromCache).to.be.true;
+        expect(toDataArray(snapshot2)).to.deep.equal([]);
+      });
+    });
+
+    it('can empty cached collection and raise snapshot from it', () => {
```
dconeybe, line 1310: Please rename this test to something like "Raises initial snapshot from cache, even if it has become empty"

packages/firestore/test/integration/api/query.test.ts
```diff
@@ -1288,6 +1288,93 @@ apiDescribe('Queries', (persistence: boolean) => {
...
+        // Add a snapshot listener whose first event should be raised from cache.
+        const storeEvent = new EventsAccumulator<QuerySnapshot>();
+        onSnapshot(coll, storeEvent.storeEvent);
+        const snapshot2 = await storeEvent.awaitEvent();
+        expect(snapshot2.metadata.fromCache).to.be.true;
+        expect(toDataArray(snapshot2)).to.deep.equal([]);
+      });
+    });
+
+    it('can raise snapshot from a cached collection which was emptied offline', () => {
```
dconeybe, line 1331: I think you can just delete this and the next tests, that verify the offline behavior. They seem superfluous.

packages/firestore/test/util/api_helpers.ts
```diff
@@ -152,7 +154,8 @@ export function querySnapshot(
     mutatedKeys,
     fromCache,
     syncStateChanged,
-    false
+    false,
+    resumeToken
```
dconeybe, line 158: Just specify `ByteString.EMPTY_BYTE_STRING` directly here, instead of assigning it to a local variable first. This is a bit more readable. Just use that `*/ resumeToken=*/` comment style to document what it is.

dconeybe (changes requested):
A few small comments. Also, I added a couple of spec tests. I don't think you need to learn how to write spec tests at this point (you'll definitely have chances later!).

packages/firestore/src/core/view_snapshot.ts
```diff
@@ -184,7 +188,8 @@ export class ViewSnapshot {
       !this.mutatedKeys.isEqual(other.mutatedKeys) ||
       !queryEquals(this.query, other.query) ||
       !this.docs.isEqual(other.docs) ||
-      !this.oldDocs.isEqual(other.oldDocs)
+      !this.oldDocs.isEqual(other.oldDocs) ||
+      this.resumeToken !== other.resumeToken
```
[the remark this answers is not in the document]
dconeybe: I'm only concerned with the publicly-facing DocumentSnapshot class and its equals() method. I don't know what the expected behavior is for our internal ViewSnapshot class. But definitely two DocumentSnapshot instances should be considered "equal" if the only difference is the resume token.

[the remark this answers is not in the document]
milaGGL: snapshotEqual can compare 2 DocumentSnapshots or 2 QuerySnapshots.
If they are DocumentSnapshots, resume token shouldn't effect the result, as it does not include ViewSnapshot related info.
But if it is QuerySnapshots, it requires equality in _snapshot property (ViewSnapshot). so resume token would effect the equality. 
Let's remove resume token from ViewSnaphot.isEqual()

dconeybe:
I just realized that I had forgotten to push up my changes that add two spec tests. I've now pushed them up in https://github.com/firebase/firebase-js-sdk/pull/6624/commits/f06eb5f83dff2190842aff3708c346a8a3e415c2

dconeybe:
Please add a changeset so an entry will be included in the release notes.

dconeybe (changes requested):

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1695,4 +1698,121 @@ describeSpec('Listens:', [], () => {
...
+          // Listen to the query again and verify that the empty snapshot is
+          // raised from cache, even though the write is not yet acknowledged.
+          .userListens(query1, { resumeToken: 'resume-token-1000' })
+          .expectEvents(query1, { fromCache: true })
+      );
+    }
+  );
+
+  specTest(
+    'Empty initial snapshot is raised from cache in multiple tabs',
```
milaGGL, line 1752: having multiple tabs:
1. cache empty result in tab1
2. re-listen to the same query in tab2, it should raise initial snapshot from cache

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1695,4 +1698,121 @@ describeSpec('Listens:', [], () => {
...
+          .watchAcksFull(query1, 2000)
+          .client(1)
+          .expectEvents(query1, { fromCache: false })
+      );
+    }
+  );
+  specTest(
+    'Empty-due-to-delete initial snapshot is raised from cache in multiple tabs',
+    ['multi-client'],
+    () => {
```
milaGGL, line 1787: having multiple tabs:
1. cache result with one doc in tab1
2. unlisten and delete the document
3. re-listen to the result in tab2, it should raise initial snapshot from cache with local mutation (local delete in step2)
4. update the snapshot after the mutation is reflected in server

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1695,4 +1698,121 @@ describeSpec('Listens:', [], () => {
...
+          .client(1)
+          .userListens(query1)
+          .expectEvents(query1, { fromCache: true })
+          // Should get events once stream is caught up.
+          .client(0)
+          .expectListen(query1, { resumeToken: 'resume-token-1000' })
+          .writeAcks('collection/a', 2000)
+          .watchAcksFull(query1, 2000, doc1Deleted)
+          .client(1)
+          .expectListen(query1, { resumeToken: 'resume-token-2000' })
```
milaGGL, line 1813: this test is always passing regardless of the resume token in this line(1813). why?
ie: replace line 1813 with any of this: 
```
.expectListen(query1)
.expectListen(query1, { resumeToken: 'resume-token-1000' })
.expectListen(query1, { resumeToken: 'resume-token-2000' })
.expectListen(query1, { resumeToken: 'resume-token-3000' })
```

packages/firestore/test/unit/api/database.test.ts
```diff
@@ -179,6 +181,48 @@ describe('QuerySnapshot', () => {
     ).to.be.false;
   });
 
+  it('resume token should not effect querySnapshot equality', () => {
```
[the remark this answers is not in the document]
dconeybe: Bump.

dconeybe:
> Please add a changeset so an entry will be included in the release notes.

Bump

dconeybe (changes requested):
I've asked @wu-hui to look at the spec tests tomorrow. Other than that, mostly LGTM (other than my 2 other comments).

packages/firestore/test/unit/specs/listen_spec.test.ts
dconeybe: We figured out that you can remove the expectListen() altogether (it shouldn't be there). It looks like a bug in the spec test framework since it's effectively ignoring this call. But, oh well.

.changeset/cool-grapes-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/firestore': patch
```
dconeybe, line 2: Add `'firebase': patch` as well

wu-hui (commented):
The multi-tab spec tests look good. Also this seems to be a clever fix!

wu-hui:
> The multi-tab spec tests look good. Also this seems to be a clever fix!

I do wonder if we should avoid passing `resumeToken` all the way to snapshots/views however. It is an underlying detail that should be no concern of snapshots/views. Maybe we can replace it with a boolean, something like `hasCachedResults`?

packages/firestore/test/unit/specs/listen_spec.test.ts
```diff
@@ -1695,4 +1698,115 @@ describeSpec('Listens:', [], () => {
...
+          // Populate the cache with the empty query results.
+          .userListens(query1)
+          .watchAcksFull(query1, 1000)
+          .expectEvents(query1, { fromCache: false })
+          .userUnlistens(query1)
+          .watchRemoves(query1)
+          .client(1)
+          // Re-listen to the query in second client and verify that the empty
+          // snapshot is raised from cache.
+          .userListens(query1, { resumeToken: 'resume-token-1000' })
```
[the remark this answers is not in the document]
milaGGL: u r right, it doesn't require resume token in userListens

packages/firestore/src/core/view.ts
```diff
@@ -322,7 +321,8 @@ export class View {
         newSyncState === SyncState.Local,
         syncStateChanged,
         /* excludesMetadataChanges= */ false,
-        targetChange?.resumeToken ?? ByteString.EMPTY_BYTE_STRING
+        targetChange !== undefined &&
```
dconeybe, line 324: Could this be more succinctly written as something like this:

```
targetChange ? targetChange.resumeToken.approximateByteSize() > 0 : false
```

packages/firestore/test/unit/api/database.test.ts
```diff
@@ -179,48 +177,51 @@ describe('QuerySnapshot', () => {
...
-      {},
-      { a: { a: 1 } },
-      keys(),
-      false,
-      false,
-      resumeToken2
-    );
-    expect(snapshotEqual(snapshot1, snapshot1Copy)).to.be.true;
-    expect(snapshotEqual(snapshot1, snapshot2)).to.be.true;
+    // hasCachedResults should effect querySnapshot equality
```
dconeybe, line 180: nit: "effect" -> "affect"

dconeybe (changes requested):

dconeybe (approved):
Awesome job! Thanks!

PR #6625, opened 2022-09-23
WIP: Fix COUNT for the Console team

dconeybe:


dconeybe:
Please just use the https://github.com/firebase/firebase-js-sdk/tree/mila/count-export-count-quries branch directly, as an equivalent fix has been incorporated there.

PR #6629, opened 2022-09-26
Mfa totp demoapp

prameshj:
Exported TOTP objects and updated the demo app to support TOTP MFA enrollment.

Modified the demo app package.json to pick up auth changes from HEAD.

### DIFF
--- a/common/api-review/auth.api.md
+++ b/common/api-review/auth.api.md
@@ -751,8 +751,34 @@ export interface TotpMultiFactorAssertion extends MultiFactorAssertion {
 }
 
 // @public
+export class TotpMultiFactorGenerator {
+    static assertionForEnrollment(secret: TotpSecret, oneTimePassword: string): TotpMultiFactorAssertion;
+    static assertionForSignIn(enrollmentId: string, oneTimePassword: string): TotpMultiFactorAssertion;
+    // Warning: (ae-forgotten-export) The symbol "FactorId" needs to be exported by the entry point index.d.ts
+    static FACTOR_ID: FactorId_2;
+    static generateSecret(session: MultiFactorSession): Promise<TotpSecret>;
+}
+
+// @public
 export interface TotpMultiFactorInfo extends MultiFactorInfo {
 }
+
+// @public
+export class TotpSecret {
+    readonly codeIntervalSeconds: number;
+    readonly codeLength: number;
+    // Warning: (ae-forgotten-export) The symbol "StartTotpMfaEnrollmentResponse" needs to be exported by the entry point index.d.ts
+    //
+    // @internal (undocumented)
+    static _fromStartTotpMfaEnrollmentResponse(response: StartTotpMfaEnrollmentResponse, auth: AuthInternal): TotpSecret;
+    generateQrCodeUrl(accountName?: string, issuer?: string): string;
+    readonly hashingAlgorithm: string;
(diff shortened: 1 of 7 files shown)


prameshj:
cc @sam-gc @bhparijat @renkelvin

common/api-review/auth.api.md
```diff
@@ -750,10 +750,40 @@ export function signOut(auth: Auth): Promise<void>;
...
+    static generateSecret(session: MultiFactorSession): Promise<TotpSecret>;
+}
+
 // @public
 export interface TotpMultiFactorInfo extends MultiFactorInfo {
 }
 
+// @public
+export class TotpSecret {
+    // (undocumented)
```
sam-gc, line 768: I'm not sure why this got included now (maybe because the files were added to the index?) but we should make sure anything that's undocumented is marked internal, or is documented

common/api-review/auth.api.md
```diff
@@ -750,10 +750,40 @@ export function signOut(auth: Auth): Promise<void>;
 export interface TotpMultiFactorAssertion extends MultiFactorAssertion {
 }
 
+// @public
+export class TotpMultiFactorGenerator {
+    static assertionForEnrollment(secret: TotpSecret, oneTimePassword: string): TotpMultiFactorAssertion;
+    static assertionForSignIn(enrollmentId: string, oneTimePassword: string): TotpMultiFactorAssertion;
+    // Warning: (ae-forgotten-export) The symbol "FactorId" needs to be exported by the entry point index.d.ts
+    static FACTOR_ID: FactorId_2;
```
sam-gc, line 758: `FactorId_2` is fishy, the public type might be using the wrong enum

sam-gc (approved):

common/api-review/auth.api.md
prameshj: This uses FactorId from public_types - https://github.com/firebase/firebase-js-sdk/blob/mfa-totp/packages/auth/src/mfa/assertions/totp.ts#L21, which is using the enum type - https://github.com/firebase/firebase-js-sdk/blob/fe42bacddbd3586b12fdbebccaddd65bf639086c/packages/auth/src/model/public_types.ts#L546

This looks right to me.. maybe we are running into something like - https://github.com/microsoft/rushstack/issues/3581


The other option is to define FACTOR_ID as string in https://github.com/firebase/firebase-js-sdk/blob/fe42bacddbd3586b12fdbebccaddd65bf639086c/packages/auth/src/mfa/assertions/totp.ts#L107, similar to what phone mfa does - https://github.com/firebase/firebase-js-sdk/blob/mfa-totp/packages/auth/src/platform_browser/mfa/assertions/phone.ts#L98

common/api-review/auth.api.md
prameshj: These are public fields and were documented in the constructor - https://github.com/firebase/firebase-js-sdk/blob/mfa-totp/packages/auth/src/mfa/assertions/totp.ts#L175-L178

Looks like I would need to declare each member with a doc, rather than implicitly define them in the constructor. 

Fixed it.

common/api-review/auth.api.md
prameshj: Synced offline that we will leave this as-is for now and switch to String if we see the need for it.

PR #6633, opened 2022-09-27
Update prune-dts to allow beta APIs

hsubox76:
Update prune-dts (only used by Firestore I think) to allow beta APIs to get through (this seems to be the default behavior when using api-documenter). I think this is because api-documenter uses the *.api.json files and prune-dts filters out beta APIs before they even get into the json, whereas api-extractor puts everything in the json, and only filters what it puts into the public *.d.ts file.

Also an overload was recently added to `getFirestore()` which causes `{@link getFirestore}` to no longer work, leaving only a blank space in its place in reference docs. Fixed this by specifying the overload.

Also regenerated the firestore.api.md file after these changes.

### DIFF
--- a/common/api-review/firestore.api.md
+++ b/common/api-review/firestore.api.md
@@ -236,6 +236,30 @@ export function getFirestore(app: FirebaseApp): Firestore;
 // @public
 export function increment(n: number): FieldValue;
 
+// @beta
+export interface Index {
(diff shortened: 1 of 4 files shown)


wu-hui (approved):

dwyfrequency (approved):
LGTM

PR #6637, opened 2022-09-28
Add leading slash to toc path

hsubox76:
Add a leading slash to keep the toc contents consistent with what's already on devsite.


dwyfrequency (approved):
LGTM

PR #6626, opened 2022-09-23
sign-in flow for totp

bhparijat:
Support sign-in with MFA TOTP.

Builds on top of the changes in https://github.com/firebase/firebase-js-sdk/pull/6629

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.

PR #6640, opened 2022-09-29
Unit test cases for sign-in flow

bhparijat:
This PR contains code implementation for adding unit test cases for sigin-flow

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/auth/src/mfa/assertions/totp.test.ts
+++ b/packages/auth/src/mfa/assertions/totp.test.ts
@@ -24,7 +24,10 @@ import * as mockFetch from '../../../test/helpers/mock_fetch';
 import { Endpoint } from '../../api';
 import { MultiFactorSessionImpl } from '../../mfa/mfa_session';
 import { StartTotpMfaEnrollmentResponse } from '../../api/account_management/mfa';
-import { FinalizeMfaResponse } from '../../api/authentication/mfa';
+import {
+  FinalizeMfaResponse,
+  StartTotpMfaSignInResponse
+} from '../../api/authentication/mfa';
 import {
   TotpMultiFactorAssertionImpl,
   TotpMultiFactorGenerator,
@@ -34,6 +37,8 @@ import { FactorId } from '../../model/public_types';
 import { AuthErrorCode } from '../../core/errors';
 import { AppName } from '../../model/auth';
 import { _castAuth } from '../../core/auth/auth_impl';
+import { MultiFactorAssertionImpl } from '../mfa_assertion';
+import { FirebaseError } from '@firebase/util';
 
 use(chaiAsPromised);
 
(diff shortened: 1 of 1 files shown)


packages/auth/src/mfa/assertions/totp.test.ts
```diff
@@ -208,6 +219,91 @@ describe('core/mfa/totp/assertions/TotpMultiFactorAssertionImpl', () => {
...
+      '123456'
+    ) as any;
+
+    const mock = mockEndpoint(
+      Endpoint.FINALIZE_MFA_SIGN_IN,
+      totpSignInResponse
+    );
+    //console.log("response: ",await assertion._process(auth,session));
+
+    try {
```
bhparijat, line 276: For some reason, I was not able to do this with the chaiAsPromised format of using `await expect(foo()).to.eventually.be.eql(something)` and so resolved to default try, catch

packages/auth/src/mfa/assertions/totp.test.ts
```diff
@@ -208,6 +213,78 @@ describe('core/mfa/totp/assertions/TotpMultiFactorAssertionImpl', () => {
...
+  it('should finalize mfa signin for totp', async () => {
+    totpSignInResponse = { 'verificationCode': '123456' } as any;
+    assertion = TotpMultiFactorGenerator.assertionForSignIn(
+      'enrollment-id',
+      '123456'
+    ) as any;
+
+    const mock = mockEndpoint(
+      Endpoint.FINALIZE_MFA_SIGN_IN,
+      totpSignInResponse
```
prameshj, line 239: Finalize needs to return id and refresh tokens - https://cloud.google.com/identity-platform/docs/reference/rest/v2/accounts.mfaSignIn/finalize, https://github.com/firebase/firebase-js-sdk/blob/04dcdbb0f5b19e51082a846463d84aadb77e98d4/packages/auth/src/api/authentication/mfa.test.ts#L115

 Let's use that same format in the mocks.

packages/auth/src/mfa/assertions/totp.test.ts
prameshj: This is expected.. you need to wrap await inside try/catch to catch errors.

prameshj (commented):
You could combine this with https://github.com/firebase/firebase-js-sdk/pull/6626 since the unit test changes aren't that big.

packages/auth/src/mfa/assertions/totp.test.ts
```diff
@@ -208,6 +213,91 @@ describe('core/mfa/totp/assertions/TotpMultiFactorAssertionImpl', () => {
...
+  });
+  afterEach(mockFetch.tearDown);
+
+  it('should finalize mfa signin for totp', async () => {
+    
+    totpSignInResponse = { 
+      verificationCode: '123456',
+      idToken: 'final-id-token',
+      refreshToken: 'refresh-token'
+     } as any;
```
sam-gc, line 236: The use of `any` here will probably fail the linter

packages/auth/src/mfa/assertions/totp.test.ts
sam-gc: I'd recommend using the `.to.be.rejectedWith()` construction we use elsewhere in the SDK. It's a bit cleaner than using try/catch, plus as the test is currently written if the assertion._process() _doesn't_ error as expected, the assertions in the catch block will never run. Here's an example:

https://github.com/firebase/firebase-js-sdk/blob/48a127b2f42efac369735f83e2597db08b24a0f9/packages/auth/src/api/index.test.ts#L161-L170

So in your case here, it would be:
```ts
await expect(assertion._process(auth, session)).to.be.rejectedWith('auth/argument-error');
```

If that's not working we should try to figure out why it's not

bhparijat: I have tried this again and it does not work.

bhparijat: The difference between the given example and this test is: just throwing error and throwing an error in an aync function. The latter is same as rejecting a promise. I noticed that if I use
```return new Promise(()=>{
     throw new Error('test')
     })
 ```
 
 ```await expect(assertion._process(auth, session)).to.be.rejectedWith('test');``` works

packages/auth/src/mfa/assertions/totp.test.ts
bhparijat: It does not perhaps because it's a test file

packages/auth/src/mfa/assertions/totp.test.ts
sam-gc: See my comment at https://github.com/firebase/firebase-js-sdk/pull/6626/files#r993768335 and try the promise-based method again, that should fix the issue

sam-gc (approved):

renkelvin (approved):

PR #6643, opened 2022-09-30
Make COUNT queries available to the Firestore Console

dconeybe:
There are two notable changes in this PR:
1. Add a call to `XhrIo.setWithCredentials(true)` when invoking the `RunAggregationQuery` RPC.
2. Add a hook in firestore-compat to call `getCountFromServer()` in the `INTERNAL` object.

The `XhrIo.setWithCredentials(true)` change is required so that the Firestore Console's authentication cookie will be included in the HTTP request headers. Without it, the requests fail with HTTP 401 errors. This problem is unique to the Firestore Console because they use 1st-party authentication. This change should have no effect on any other users. Moreover, the "Listen" and "Write" streams have been invoking `XhrIo.setWithCredentials(true)` indirectly since the dawn of time by specifying `supportsCrossDomainXhr: true` in the `WebChannelOptions`: https://github.com/firebase/firebase-js-sdk/blob/34ad43cc2a9863f7ac326c314d9539fcbc1f0913/packages/firestore/src/platform/browser/webchannel_connection.ts#L179 which trickles down to call `XhrIo.setWithCredentials(true)` here: https://github.com/google/closure-library/blob/e439bfd5ff09e0efd85693fc5adf23462d5a39d4/closure/goog/labs/net/webchannel/webchannelbase.js#L2601

The change to the firestore-compat library is done because the Firestore Console is stuck using the compat SDK for now, and we don't want to add a new public API surface to the firestore-compat library.

Googlers see b/250662155 for more info.

### DIFF
--- a/.changeset/metal-goats-unite.md
+++ b/.changeset/metal-goats-unite.md
@@ -1,0 +1,8 @@
+---
+'@firebase/firestore': minor
+'@firebase/firestore-compat': minor
+'@firebase/webchannel-wrapper': minor
+'firebase': minor
+---
+
+Set withCredentials=true when making requests via non-streaming RPCs, like is done for streaming RPCs.
--- a/packages/firestore-compat/src/index.console.ts
+++ b/packages/firestore-compat/src/index.console.ts
@@ -22,8 +22,11 @@ import {
   Firestore as FirestoreExp,
   FirestoreError,
   _EmptyAuthCredentialsProvider,
-  _EmptyAppCheckTokenProvider
+  _EmptyAppCheckTokenProvider,
+  Query as ExpQuery,
+  getCountFromServer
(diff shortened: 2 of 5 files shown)


dconeybe:
@egilmorez PTAL as a required reviewer. Thanks!

markarndt (approved):

egilmorez (approved):
Wondering why TW review is required when there are no public doc changes, but here you go :)

milaGGL (approved):
LGMT

Thaina:
Pardon me but what is this function? It seem like it is a very significant feature yet I have not seen any documents about it

dconeybe:
We will be talking about this feature more in the coming weeks. Stay tuned!

PR #6644, opened 2022-09-30
Adding Count to be accessible to the console.

dannhauser:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/firestore-compat/src/index.console.ts
+++ b/packages/firestore-compat/src/index.console.ts
@@ -22,8 +22,11 @@ import {
   Firestore as FirestoreExp,
   FirestoreError,
   _EmptyAuthCredentialsProvider,
-  _EmptyAppCheckTokenProvider
+  _EmptyAppCheckTokenProvider,
+  Query as ExpQuery,
+  getCountFromServer
 } from '@firebase/firestore';
+import { Compat } from '@firebase/util';
 
 import {
   Firestore as FirestoreCompat,
@@ -98,6 +101,15 @@ export class Firestore extends FirestoreCompat {
       new MemoryPersistenceProvider()
     );
   }
+
+  INTERNAL = {
+    delete: () => this.terminate(),
+    count: (query: Compat<ExpQuery<unknown>>) => {
+      return getCountFromServer(query._delegate).then(response => {
+        return response.data().count;
+      });
+    }
+  };
 }
 
 function databaseIdFromFirestoreDatabase(


packages/firestore-compat/src/api/database.ts
```diff
@@ -989,6 +989,10 @@ export class Query<T = PublicDocumentData>
     this._userDataWriter = new UserDataWriter(firestore);
   }
 
+  get _query() {
+    return this._delegate._query;
+  }
+
```
ehsannas, line 995: pardon my ignorance. Why this is needed?

dannhauser: As part of the code path for getting the count, it assumes that it has the _query object cause I think that is how the api has evolved? I am guessing this value used to be on the delegate pre v9 but I don't know for sure? This just exposes it so that the code here will not throw:

https://github.com/firebase/firebase-js-sdk/blob/ee871fc0b157e1a186c2895b4290d70c8d1b986f/packages/firestore/src/core/count_query_runner.ts#L39

ehsannas: Ah! Got it.

packages/firestore-compat/src/api/database.ts
```diff
@@ -989,6 +989,10 @@ export class Query<T = PublicDocumentData>
     this._userDataWriter = new UserDataWriter(firestore);
   }
 
+  get _query() {
```
dconeybe, line 992: As lint suggests, please add a return type to this function.

packages/firestore-compat/src/index.console.ts
```diff
@@ -98,6 +102,15 @@ export class Firestore extends FirestoreCompat {
       new MemoryPersistenceProvider()
     );
   }
+
+  INTERNAL = {
+    delete: () => this.terminate(),
+    count: (query: ExpQuery<unknown>) => {
+      return getCountFromServer(query).then((response: AggregateQuerySnapshot<AggregateSpec>) => {
```
dconeybe, line 109: I think the Promise returned from getCountFromServer() needs a .catch() too. See what happens if the promise returned from getCountFromServer() is rejected. I don't think the promises are chained properly.

dconeybe, line 109: Do you really need to provide the type of the `response: AggregateQuerySnapshot<AggregateSpec>` argument? Can't you just omit the type and let the compiler infer it?

dconeybe (changes requested):

packages/firestore-compat/src/index.console.ts
dannhauser: you shouldn't need it cause the consumer should have the catch to handle the error state in the promise chain, but just in case, added the catch to rethrow the error for the consumer to handle.

dannhauser: removed

dconeybe: Ahh I think you're right. Please remove the unnecessary catch(). Sorry for the false alarm... TypeScript is not my strongest language.

dannhauser: removed it 😄

dconeybe (approved):

PR #6650, opened 2022-10-03
Updated readme instructions for storage

maneesht:
Updated README to include CORS setup information to run Storage tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,26 @@ database. When prompted to select the set of initial security rules, select
 any option (e.g. "Start in Production Mode") since these permission settings
 will be overwritten below.
 
+#### Storage Setup
+
+Visit the "Storage" section of the console and create a storage bucket. In 
+order to run the tests, you will need to update your bucket's CORS rules.
+
+1. Create a new file called `cors.json` with the contents:
+```json
+[
+    {
+        "origin": ["http://localhost:8089"],
+        "method": ["GET"],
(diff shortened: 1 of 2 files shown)


packages/storage/.gitignore
```diff
@@ -0,0 +1 @@
+cors.json
```
tonyjhuang, line 1: remove?

tonyjhuang (approved):
LGTM except one comment

packages/storage/.gitignore
maneesht: Done.

hsubox76 (approved):

README.md
```diff
@@ -113,6 +113,26 @@ database. When prompted to select the set of initial security rules, select
 any option (e.g. "Start in Production Mode") since these permission settings
 will be overwritten below.
 
+#### Storage Setup
+
+Visit the "Storage" section of the console and create a storage bucket. In
```
dwyfrequency, line 118: The double quotes look good to me but it could qualify for backticks. @hsubox76 what do you think here?

maneesht: I'm happy to change it, but the rest of the file uses double quotes, so if we change one, we should probably change the rest

dwyfrequency (approved):
LGTM

README.md
dwyfrequency: in that case, double quotes is good

PR #6651, opened 2022-10-03
Remove excluded packages from doc index.md

hsubox76:
Removes the 3 packages from the v9 docs index page that shouldn't have a public link (they're already removed from the TOC and the generated md files are also removed)

This is the page: https://firebase.google.com/docs/reference/js


dwyfrequency (approved):
LGTM

PR #6653, opened 2022-10-04
Implemented exponential backoff with query

maneesht:
When upload requests respond with a 503, there is a situation where the bytes were uploaded, yet finalize did not occur. This can result in an incorrect offset being sent to the server. We should make a request that checks the status of the upload, and updating our offset based on that.

This PR implements that, along with a backoff feature, exponentially increasing the time between different requests being issued.

### DIFF
--- a/packages/storage/src/implementation/backoff.ts
+++ b/packages/storage/src/implementation/backoff.ts
@@ -24,15 +24,24 @@ type id = (p1: boolean) => void;
 export { id };
 
 /**
- * @param f May be invoked
- *     before the function returns.
- * @param callback Get all the arguments passed to the function
- *     passed to f, including the initial boolean.
+ * Accepts a callback for an action to perform (`doRequest`),
+ * and then a callback for when the backoff has completed (`backoffCompleteCb`).
+ * The callback sent to start requires an argument to call (`onRequestComplete`)
+ * When `start` calls `doRequest`, it passes a callback for when the request has
+ * completed, `onRequestComplete`. Based on this, the backoff continues, with
+ * another call to `doRequest` and the above loop continues until the timeout
+ * is hit, or a successful response occurs.
+ * @description
+ * @param doRequest Callback to perform request
+ * @param backoffCompleteCb Callback to call when backoff has been completed
  */
 export function start(
-  f: (p1: (success: boolean) => void, canceled: boolean) => void,
+  doRequest: (
+    onRequestComplete: (success: boolean) => void,
+    canceled: boolean
+  ) => void,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  callback: (...args: any[]) => unknown,
+  backoffCompleteCb: (...args: any[]) => unknown,
   timeout: number
 ): id {
   // TODO(andysoto): make this code cleaner (probably refactor into an actual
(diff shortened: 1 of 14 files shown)


packages/storage/src/task.ts
```diff
@@ -92,6 +94,15 @@ export class UploadTask {
   private _reject?: (p1: StorageError) => void = undefined;
   private _promise: Promise<UploadTaskSnapshot>;
 
+  // TODO(mtewani): Update these to use predefined constants
```
tonyjhuang, line 97: Any reason to not do this in this PR?

packages/storage/src/task.ts
```diff
@@ -344,7 +369,8 @@ export class UploadTask {
         requestInfo,
         newTextConnection,
         authToken,
-        appCheckToken
+        appCheckToken,
+        false
```
tonyjhuang, line 373: why false?

tonyjhuang (changes requested):

packages/storage/src/task.ts
maneesht: I added a comment above, but basically we don't want to retry any of the uploads with backoff immediately since we want to make a query request first before retrying

packages/storage/src/task.ts
maneesht: Done.

packages/storage/src/implementation/constants.ts
```diff
@@ -42,6 +42,11 @@ export const DEFAULT_MAX_OPERATION_RETRY_TIME = 2 * 60 * 1000;
  */
 export const DEFAULT_MAX_UPLOAD_RETRY_TIME = 10 * 60 * 1000;
 
+/**
+ * 1 second
+ */
+export const DEFAULT_MIN_SLEEP_TIME = 1000;
```
tonyjhuang, line 48: I know we don't currently do this but it would be nice to suffix time-related constants with their units. RETRY_MILLIS is a lot more parseable than RETRY_TIME

packages/storage/src/task.ts
maneesht: Actually, nevermind. It should be true. Nice catch

tonyjhuang (approved):
LGTM except one (optional) comment

packages/storage/src/implementation/constants.ts
maneesht: Ended up adding `_MILLIS` to the end of this, but I think we should wait to rename the rest in a different PR

Yuangwang (approved):

Yuangwang:
Thanks for the changes!

packages/storage/src/implementation/backoff.ts
```diff
@@ -24,15 +24,24 @@ type id = (p1: boolean) => void;
 export { id };
 
 /**
- * @param f May be invoked
- *     before the function returns.
- * @param callback Get all the arguments passed to the function
- *     passed to f, including the initial boolean.
+ * Accepts a callback for an action to perform (`doRequest`),
+ * and then a callback for when the backoff has completed (`backoffCompleteCb`).
+ * The callback sent to start requires an argument to call (`onRequestComplete`)
```
markarndt, line 29: Missing end punctuation line 29.

markarndt (approved):

egilmorez (approved):
LG based on Mark's review, thanks!

hsubox76 (approved):

PR #6656, opened 2022-10-05
Expose only necessary recaptcha errors

renkelvin:


packages/auth/src/core/credentials/email.ts
```diff
@@ -124,7 +125,7 @@ export class EmailAuthCredential extends AuthCredential {
           return signInWithPassword(auth, requestWithRecaptcha);
         } else {
           return signInWithPassword(auth, request).catch(async (error) => {
-            if (error.code === `auth/${AuthErrorCode.MISSING_RECAPTCHA_TOKEN}`) {
+            if (error.code === `auth/${ServerError.MISSING_RECAPTCHA_TOKEN}`) {
```
prameshj, line 128: Based on this - https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/src/api/index.ts#L169-L175, in the else case, we will only see INTERNAL_ERROR , if , say, the recaptcha version was invalid. So we rely on the error message from the backend to provide a more verbose description of the error?

renkelvin: Yes, that's correct.

prameshj: Thanks! It would be great if you can verify what error description is returned for invalid_recaptcha_version or similar error. I am fine with this approach as long as we get a verbose description, for debugging.

renkelvin: The error for invalid_recaptcha_version looks like this 
```
FirebaseError: Firebase: Error (auth/invalid-value-at-'version'-(type.googleapis.com/google.cloud.identitytoolkit.v2.getrecaptchaconfigrequest.recaptchaversion),-"recaptcha-enterprise-invalid").
```

prameshj: Thanks for checking! LGTM

sam-gc (approved):

Issue #6660, opened 2022-10-07
Warning: Critical dependency: the request of a dependency is an expression


anisabboud:
Hi, I just updated Firebase SDK to the latest version released yesterday (9.11.0 - 2022-10-06), and now getting a console warning when building the project (with Angular ng build / ng serve):

![image](https://user-images.githubusercontent.com/5672108/194479939-c2153df5-0ee8-4102-aca7-b661131c6ef0.png)

```bash
./node_modules/@firebase/util/dist/index.esm2017.js:741:19-44 - Warning: Critical dependency: the request of a dependency is an expression
```

Specifically, the warning comes from this line:
![image](https://user-images.githubusercontent.com/5672108/194480409-32a62f96-2036-47ae-84ab-bfc742eabc00.png)

which seems to be in this file: **firebase-js-sdk/packages/util/src/deferred.ts**
https://github.com/firebase/firebase-js-sdk/blob/5ad7ff2ae955c297556223e6cb3ad9d4b897f664/packages/util/src/deferred.ts#L39

huzaifa3165:
I'm also getting the same warning.

hsubox76:
This is puzzling because that file hasn't been updated in 2 years. Did you upgrade directly from 9.10.0 to 9.11.0 or did you start from a version earlier than 9.10.0? Did you update any other dependencies as part of that update?

This is the first time `@firebase/util` has had a version bump since July so it's possible some commit to util between then and now caused it also.

Which products are you using (analytics, app-check, database, firestore, etc.)? It might also be that something changed in the way the product code calls `Deferred`.

hsubox76:
Actually I looked up the error and it seems it's a webpack warning about using `require()` on a dynamic variable. In that case it's most likely caused by this line, which was added recently:

https://github.com/firebase/firebase-js-sdk/blob/0a112bd2aee2e709d6733b1a36876e6fae1e347f/packages/util/src/defaults.ts#L78

and it looks like the source map is a little bit off so it's pointing at a different location in the util bundle. I'll see what I can do to address this.

nandorojo:
Seeing the same thing on Next.js. Thanks for looking into it @hsubox76.

tomasconnor:
+1

Amansg4:
Just had this problem. +1. Don't know a fix though

nandorojo:
Downgrading to 9.10 fixed it for me.

hsubox76:
[...]

PR #6663, opened 2022-10-07
Fix modular autoinit breakage

### LINKED ISSUE #6662: Uncaught TypeError: can't access property "__FIREBASE_DEFAULTS__", process.env is undefined
### Describe your environment

  * Operating System version:  Arch Linux, Kernel 5.19.13
  * Browser version: Firefox Developer Edition 106.0b8
  * Firebase SDK version: 9.11.0
  * Firebase Product: the issue is in `util`

#### Steps to reproduce:
1. Use the Firebase JS client SDK in an environment where `process` is defined, but `process.env` isn't, like Nuxt 3 in the browser.
2. Observe the following error in the console:

```
Uncaught TypeError: can't access property "__FIREBASE_DEFAULTS__", process.env is undefined
```

bcspragu:
I ran into this issue updating dependencies for a Nuxt 3 app. The problem is that `process.{client,server,dev}` are defined even on the client, so the client ends up broken with errors like:

```
Uncaught TypeError: can't access property "__FIREBASE_DEFAULTS__", process.env is undefined
```

Since the code only uses `process.env`, this updates the check to look explicitly for `process.env`.

Fixes #6662


hsubox76:
Thank you! Can you add a changeset for `@firebase/util`? Patch should be fine.

bcspragu:
Should be all set, updated `@firebase/util` to `1.7.1`

hsubox76:
Thanks! Just fyi, we use `changesets` so instead of altering the package.json directly, you'd need to follow the directions in the "No Changeset Found" comment above.

But it turns out I also have to make other fixes in that file due to https://github.com/firebase/firebase-js-sdk/issues/6660 so I can just put this fix in the same PR so we won't run into any merge conflicts or anything and you don't have to worry about the changeset tool.

hsubox76:
Rolling this fix into https://github.com/firebase/firebase-js-sdk/pull/6665

bcspragu:
Sounds good to me, and sorry for glossing over the (very clearly laid out) instructions!

PR #6526, opened 2022-08-12
SSR/SSG autoconfig

jamesdaniels:
Hey there! So you want to contribute to a Firebase SDK? 
Before you file this pull request, please read these guidelines:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.

PR #6665, opened 2022-10-07
Remove json path and fix process check

hsubox76:
Fix bugs introduced in https://github.com/firebase/firebase-js-sdk/pull/6526:

1) Remove `__FIREBASE_DEFAULTS_PATH__` option to point to a JSON file path for modular autoinit defaults. This causes webpack warnings about dynamic `require()`s although the code is only meant to take that path in a Node environment. (See https://github.com/firebase/firebase-js-sdk/issues/6660). If we can figure out a way around this, we can add the functionality back.

2) Fix the `process` check to be more robust and handle environments that have `process` but do not have `process.env`. This incorporates the fix in PR https://github.com/firebase/firebase-js-sdk/pull/6663

Fixes https://github.com/firebase/firebase-js-sdk/issues/6660
Fixes https://github.com/firebase/firebase-js-sdk/issues/6662

### DIFF
--- a/.changeset/strong-squids-dress.md
+++ b/.changeset/strong-squids-dress.md
@@ -1,0 +1,5 @@
+---
+'@firebase/util': patch
+---
+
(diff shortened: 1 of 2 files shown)


dwyfrequency (approved):
LGTM

PR #6667, opened 2022-10-08
[Storage] Cleared timeouts upon pause/resume of uploads

maneesht:
Made sure that timers are cleaned up when in the middle of exponential backoff.


packages/storage/src/task.ts
```diff
@@ -94,6 +94,7 @@ export class UploadTask {
   private _metadataErrorHandler: (p1: StorageError) => void;
   private _resolve?: (p1: UploadTaskSnapshot) => void = undefined;
   private _reject?: (p1: StorageError) => void = undefined;
+  private pendingTimeout: ReturnType<typeof setTimeout> | null = null;
```
tonyjhuang, line 97: can replace this with `private pendingTimeout?: ReturnType<typeof setTimeout>`

tonyjhuang (approved):

markarndt (approved):

PR #6673, opened 2022-10-10
Handle IPv6 addresses in emulator autoinit.

yuchenshi:
Fixes IPv6 addresses not parsed correctly in #6526. See tests added

### DIFF
--- a/packages/util/test/defaults.test.ts
+++ b/packages/util/test/defaults.test.ts
@@ -28,7 +28,7 @@ describe('getDefaultEmulatorHost', () => {
 
   context('with no config', () => {
     it('returns undefined', () => {
-      expect(getDefaultEmulatorHost('firestore')).to.equal(undefined);
+      expect(getDefaultEmulatorHost('firestore')).to.be.undefined;
     });
   });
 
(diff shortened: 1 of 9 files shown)


packages/util/test/defaults.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    before(() => {
+      getGlobal().__FIREBASE_DEFAULTS__ = {
+        emulatorHosts: {
+          firestore: '127.0.0.1:8080'
+        }
+      };
+    });
+
+    it('returns hostname and port splitted', () => {
+      expect(getDefaultEmulatorHostnameAndPort('firestore')).to.equal([
```
hsubox76, line 119: This test and the one below failed locally for me unless I changed it to `to.deep.equal`. Is it passing for you?

hsubox76 (commented):
Thanks for fixing this! Just one comment on the tests.

packages/util/test/defaults.test.ts
yuchenshi: Thanks for catching this! Fixed.

hsubox76 (approved):

PR #6675, opened 2022-10-11
fix: add missing 'apple.com' `ProviderId` to enum

mesqueeb:
I have the provider ID typed in my spec of my database like so:

```ts
import { ProviderId } from 'firebase/auth'

export type PROVIDER_ID = typeof ProviderId[keyof typeof ProviderId]
```
only to my surprise, it seems you forgot about apple.com. Let's get it typed correctly with this PR! ;)

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/.changeset/odd-cherries-try.md
+++ b/.changeset/odd-cherries-try.md
@@ -1,0 +1,5 @@
+---
+'@firebase/auth-types': patch
+---
+
+fix: add missing ProviderId type
--- a/packages/auth/src/model/enums.ts
+++ b/packages/auth/src/model/enums.ts
@@ -22,6 +22,8 @@
 export const enum ProviderId {
   /** @internal */
   ANONYMOUS = 'anonymous',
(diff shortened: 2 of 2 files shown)


sam-gc:
Hi @mesqueeb, thanks for opening this PR. The reason we don't list `apple.com` in that enum (and note that that enum is only used internally to the SDK -- it's not exported since it's a `const enum` and just gets compiled to a string in JS) is that we don't have a dedicated `AppleAuthProvider` the way we have dedicated `FacebookAuthProvider`, `GoogleAuthProvider`, etc. For example in the docs, you create an Apple provider by creating an `OAuthProvider`: https://firebase.google.com/docs/auth/web/apple

Going forward, we want to treat new OAuth providers this (generic) way instead of adding new dedicated named providers. So once again, thanks for opening this PR but we won't be merging it for now.

Issue #6668, opened 2022-10-08
FR: rule-unit-testing's withSecurityRulesDisabled method JSDoc incorrect


palmerusaf:
### [REQUIRED] Describe your environment

  * Operating System version: Pop OS 22.04 LTS
  * Browser version: Firefox 105.0.1
  * Firebase SDK version: 9.10.0
  * Firebase Product: rules unit testing (auth, database, storage, etc)

#### Steps to reproduce:

The method withSecurityRulesDisabled has no documentation. After digging, it turns out it does have documentation but it is missing a * to make it a JS doc, resulting in the doc being a comment. This means the api extractor never picks it up during build.
#### Relevant Code:
The relevant line is [here](https://github.com/firebase/firebase-js-sdk/blob/29d034072c20af394ce384e42aa10a37d5dfcb18/packages/rules-unit-testing/src/public_types/index.ts#L205).

I would like to submit a PR to correct the issue. Also, I would like to include an example in the JS Doc to add clarity, since this method differs from the other context methods. I plan on including the following example.
```javascript
const addTestData = async ({ testEnv, collectionName, testData }) => {
  let docId;
  await testEnv.withSecurityRulesDisabled(async (noRulesContext) => {
    const db = await noRulesContext.firestore();
    const dbCollection = collection(db, collectionName);
    docId = await (await addDoc(dbCollection, testData)).id;
  });
  return docId;
};
```
I ran the script ```yarn docgen devsite``` and everything looks ok. Let me know if I'm missing anything or if I should use a different example.
[Here's](https://github.com/firebase/firebase-js-sdk/compare/master...palmerusaf:firebase-js-sdk:rut_doc_improvement) a link to my proposed change.

hsubox76:
Thanks, feel free to make a PR. I was going to add a comment that there needs to be a `-` after `@param callback` - that line is outside of your change but your fix made this error visible.

palmerusaf:
@hsubox76 Good catch. I've updated the commit to include your correction and submitted a [PR](https://github.com/firebase/firebase-js-sdk/pull/6676).

PR #6676, opened 2022-10-11
Fix/Improve jsdoc for withSecurityRulesDisabled

palmerusaf:
fix comment to jsdoc so it gets picked up by the api extractor change description wording slightly to add clarity since method differs from the other context methods add example of use case
add dash after @param

This PR is related to issue #6668. [Link](https://github.com/firebase/firebase-js-sdk/issues/6668#issue-1401911871)


hsubox76 (approved):

hsubox76:
FYI, after this is merged, someone will have to manually run the docgen script and publish it to devsite for the changes to take place, will try to keep it on the todo list.

hsubox76:
Oops, you probably want to run `yarn format` and push again. Looks like just a small whitespace difference.

palmerusaf:
@hsubox76 yarn formatted and re-pushed.

BuyMyBeard:
This never was merged. 2 years later, still a problem.

PR #6682, opened 2022-10-11
Upgrade TypeScript to 4.7.4

hsubox76:
We need to upgrade our TypeScript version, it's still at 4.2.2 and the latest is 4.8.4. We are going to try for a little behind the latest version, because some of our other deps have dependencies on TypeScript and may lag a little on which version they support. So we'll try for 4.7.4, the last 4.7.* release.

Most of the necessary updates (e.g. adding types to caught errors) have already been done by @dwyfrequency in previous PRs.

The most notable fix here is that TypeScript 4.3 introduced the ability for TS to recognize `{@link ` statements in comments, which caused it to break up comments into multiple nodes which caused some assumptions in `prune-dts` to break.

Also updated deprecated methods in `prune-dts` (Firestore-only script that intercepts api-extractor and does some extra things) and `create-overloads` (adds typings to compat d.ts files to enable users to use compat and modular instances in the same code) scripts.


repo-scripts/prune-dts/prune-dts.ts
```diff
@@ -426,7 +439,7 @@ function extractExportedSymbol(
   // symbol with a less restrictive type.
   const localSymbol = typeChecker.getSymbolAtLocation(typeName);
   if (localSymbol) {
```
dwyfrequency, line 441: `if (localSymbol && localSymbol?.declarations)` should remove the need for `localSymbol!.declarations`

hsubox76: Removed all `!`s and added actual checks.

dwyfrequency (approved):
LGTM

Issue #6677, opened 2022-10-11
Error: NotYetImplemented at Document2.exports2.nyi 


Santoshah:
The warning shouldn't affect any code functionality but you can downgrade to 9.10.0 if you want to avoid having a warning (or if I'm wrong and it is affecting your code). I have a fix (https://github.com/firebase/firebase-js-sdk/pull/6665) which should be released next week.

_Originally posted by @hsubox76 in https://github.com/firebase/firebase-js-sdk/issues/6660#issuecomment-1272100219_

Santoshah:
This does affect build of SSR version.

![image](https://user-images.githubusercontent.com/3863771/195037127-c23d0288-417e-42c5-a8d9-f8a8f70ed295.png)

hsubox76:
Thanks for the info, we'll be releasing a fix today or if not today, tomorrow.

hsubox76:
The fix has been released in 9.12.0.

hsubox76:
I don't think it's possible that you could get the same error in 9.8.1 because this method wasn't introduced until 9.11.0, possibly there was something left over in the node_modules or package-lock.json or yarn.lock.

Anyway it seems the problem is in https://github.com/firebase/firebase-js-sdk/blob/3e082cfa009b323b37f95efd4370eb501a562030/packages/util/src/defaults.ts#L73

Your SSR environment has a `document` object but it doesn't have a `document.cookie` or errors when you try to call it. Based on the error it looks like you're using Angular Universal, let me know if that's not the case. I'll do some research into the behavior of the `document` object in Angular Universal and see if I can figure it out. Also if this is a catchable error I'll look into wrapping the entire `getDefaults` function in a try/catch so that whatever odd environment errors happen, it won't block the rest of the app.

Santoshah:
Yes, its an Angular Universal project. 

I first notice this on Oct 10. while the code was uploding via aws codepipeline. This happens very sudden because before that my code was running fine. My firebase version was 9.8.1. (dam sure.) My deployment was success but the site throws 502 error. I login to server and checked that no command was running. Then I manually run the command `npm run serve:ssr` but it throws the same error `Error NotYetImplemented`. Then I checked the logs where I could see the same error after build was made.
<img width="551" alt="image" src="https://user-images.githubusercontent.com/3863771/195392928-3da47f02-cb5d-40c4-98e6-8faa5c3060a9.png">

[...]

PR #6686, opened 2022-10-12
Catch getDefaults() errors and do not throw

hsubox76:
Due to various frameworks and environments, we have been encountering a lot of special cases not accounted for in `getDefaults()` (such as `process` existing but `process.env` not existing in Nuxt 3, or `document` existing but `document.cookie` throwing in Angular Universal).

Adding an overall try/catch block to `getDefaults()` so that if it runs into any errors we haven't thought of, it won't block execution of the rest of Firebase. It will log a console.info so we can find out about these cases and account for them.

Also added a separate catch specifically for the Angular Universal case. Yes, this would be caught by the overall try/catch already, but I'd like to put in the specific cases as we find them so we know the specifics of the various environments we're dealing with, as this knowledge will help us with SSR support going forward.

Fixes https://github.com/firebase/firebase-js-sdk/issues/6677

### DIFF
--- a/.changeset/grumpy-suits-pump.md
+++ b/.changeset/grumpy-suits-pump.md
@@ -1,0 +1,5 @@
+---
+'@firebase/util': patch
+---
+
+Catch errors when the SDK checks for `__FIREBASE_DEFAULTS__` and do not block other app functionality.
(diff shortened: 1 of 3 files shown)


dwyfrequency (approved):
LGTM

PR #6692, opened 2022-10-13
Add a workflow to send release tweet

hsubox76:
This is manually triggered for now. You have to enter the version number. It will poll devsite 10 times every 15 minutes until it finds the release notes have been published, and then should send the tweet. I haven't put the secrets into the repo yet, I'd like to test everything else works up to that step before doing so. This PR needs to be merged to test it though.


dwyfrequency (approved):
LGTM

PR #6693, opened 2022-10-13
Remove mentions of Bower from repo and docs

hsubox76:
We have stopped publishing Bower versions of Firebase. Removing mentions of it from the repo.


dwyfrequency (approved):
LGTM

PR #6694, opened 2022-10-13
Markduckworth/or queries

MarkDuckworth:
This is the implementation of Firestore OR Queries without the public API. The public API will be released later.

Several minor API changes are included in this release: Functions in the Firestore package that return QueryConstraints (for example: where(...), limit(...), and orderBy(...)) now return a more specific type, which extends QueryConstraint. This is a non-breaking change that will support OR Queries. 

Minor version bumps are included for Firebase and @firebase/firestore.


MarkDuckworth:
@ehsannas, @hsubox76, Here's a large PR that has all of the client side changes needed to support OR queries. It was tested against the emulator. @ehsannas has reviewed all incremental changes as they were merged into this feature branch, so I'm not asking for full review of the changes. Areas that might be worth another review are: .md files, (lack of) changeset file, and package.json.

packages/firestore/src/lite-api/query.ts
```diff
@@ -87,56 +86,137 @@ export type QueryConstraintType =
...
  * @param queryConstraints - The list of {@link QueryConstraint}s to apply.
  * @throws if any of the provided query constraints cannot be combined with the
  * existing or new constraints.
  */
 export function query<T>(
   query: Query<T>,
   ...queryConstraints: QueryConstraint[]
+): Query<T>;
+
+export function query<T>(
```
ehsannas, line 156: This is not publicly exposed, is it?

ehsannas (approved):

packages/firestore/src/lite-api/query.ts
MarkDuckworth: It's not. Thanks for double checking. I also reviewed the api-report.md and index.d.ts, and only the original overload is exposed. For additional info on the topic, see https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads.

ehsannas: Thanks! 🚀

ehsannas (approved):

hsubox76 (approved):

PR #6696, opened 2022-10-13
Use API to bold

hsubox76:
API documenter's API doesn't seem to let you add arbitrary HTML or formatting so added bold using its API. Can't seem to do much more than that for now.

Also remove "section: []" from toc. Apparently devsite doesn't like it? But it used to?


dwyfrequency (approved):
LGTM

PR #6701, opened 2022-10-17
Fix version bump for api-documenter

hsubox76:
`@firebase/api-documenter` is excluded from changesets so I guess we update it manually. There is a minor bump in #6623 and a patch bump in https://github.com/firebase/firebase-js-sdk/pull/6696 so that comes out to one minor combined.


dwyfrequency (approved):
LGTM

PR #6705, opened 2022-10-18
[Storage] Removed use of native Node 18 blobs

maneesht:
Removed use of native Node 18 blobs.

### DIFF
--- a/packages/storage/src/implementation/type.ts
+++ b/packages/storage/src/implementation/type.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { isNodeSdk } from '@firebase/util';
+import { isNode } from '@firebase/util';
 import { invalidArgument } from './error';
 
 export function isJustDef<T>(p: T | null | undefined): p is T | null {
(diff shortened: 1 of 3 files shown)


packages/storage/src/implementation/type.ts
```diff
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isNodeSdk } from '@firebase/util';
```
hsubox76, line 18: Just wondering how aggressive you want to be here in the Node environment detection. `isNodeSdk()` only checks if the Node bundle is loaded whereas `isNode()` does a runtime check to see if we are in a Node environment, regardless of what bundle was loaded. A lot of SSR cases might involve the browser bundle being loaded but trying to execute in a Node environment. I do notice this is used in the RTDB codebase and wonder if it might be worth looking at there as well.

packages/storage/src/implementation/type.ts
```diff
@@ -39,7 +40,9 @@ export function isNativeBlob(p: unknown): p is Blob {
 }
 
 export function isNativeBlobDefined(): boolean {
-  return typeof Blob !== 'undefined';
+  // Note: The `isNodeSdk()` check can be removed when `ts-node` adds native Blob support
```
hsubox76, line 43: Nit: the comment doesn't match now.

PR #6706, opened 2022-10-18
Update REST and wire protocol query params for `startAfter`, `endBefore`

tohhsinpei:
We now support `startAfter` and `endBefore` on our REST endpoints ([b/207844859](http://b/207844859)). We've also added the fields `sin` ("start is inclusive") and `ein` ("end is inclusive") to the wire protocol payload (https://github.com/firebase/firebase-server/pull/4006).

These are the corresponding client-side updates. Note that we don't need to compute the successor or predecessor value anymore.

Note that CI checks will fail until the new NSS version is released.


packages/database/src/core/view/QueryParams.ts
```diff
@@ -191,10 +194,12 @@ export class QueryParams {
     copy.limitSet_ = this.limitSet_;
     copy.limit_ = this.limit_;
     copy.startSet_ = this.startSet_;
```
IanWyszynski, line 196: Let's remove `hasStartAfter()` and `hasEndBefore()`, they aren't used.

packages/database/src/core/view/QueryParams.ts
```diff
@@ -53,8 +54,10 @@ const enum REST_QUERY_CONSTANTS {
   PRIORITY_INDEX = '$priority',
   VALUE_INDEX = '$value',
   KEY_INDEX = '$key',
+  START_AFTER = 'startAfter',
   START_AT = 'startAt',
   END_AT = 'endAt',
+  END_BEFORE = 'endBefore',
   LIMIT_TO_FIRST = 'limitToFirst',
   LIMIT_TO_LAST = 'limitToLast'
 }
```
IanWyszynski, line 63: It might be worth adding a field-level comment that `startAfterSet_` is only true if `startAtSet_` is true. It has implications for synctree logic, per: https://github.com/firebase/firebase-js-sdk/blob/master/packages/database/src/core/view/QueryParams.ts#L181-L183.

IanWyszynski (commented):
Thanks for this great cleanup!

packages/database/src/core/view/QueryParams.ts
tohhsinpei: Done.

packages/database/src/core/view/QueryParams.ts
tohhsinpei: Done. I'll do a follow up PR to remove `NextPushId.successor` et al.

scripts/emulator-testing/emulators/database-emulator.ts
```diff
@@ -26,11 +26,11 @@ export class DatabaseEmulator extends Emulator {
 
   constructor(port = 8088, namespace = 'test-emulator') {
     super(
-      'firebase-database-emulator-v4.9.0.jar',
+      'firebase-database-emulator-v4.11.0.jar',
```
maneesht, line 29: Can you extract this version out to a template?

tohhsinpei: Done.

IanWyszynski (approved):

.changeset/empty-doors-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@firebase/database': patch
```
maneesht, line 2: Need to include `@firebase/database-compat1` here

tohhsinpei: Done.

packages/database/test/helpers/syncpoint-util.ts
```diff
@@ -366,10 +368,14 @@ export class SyncPointTestParser {
...
         } else if (paramName === 'limitToLast') {
           q = query(q, limitToLast(paramValue));
+        } else if (paramName === 'startAfter') {
+          q = query(q, startAfter(paramValue.index, paramValue.name));
         } else if (paramName === 'startAt') {
           q = query(q, startAt(paramValue.index, paramValue.name));
         } else if (paramName === 'endAt') {
           q = query(q, endAt(paramValue.index, paramValue.name));
+        } else if (paramName === 'endBefore') {
+          q = query(q, endBefore(paramValue.index, paramValue.name));
```
maneesht, line 378: It looks like we don't have any tests that use the `paramValue.name`, could you add some?

tohhsinpei: Done.

packages/database/test/helpers/syncPointSpec.json
```diff
@@ -4179,6 +4179,249 @@
...
+            "type": "child_added",
+            "name": "b",
+            "prevName": null,
+            "data": { "index": 2, "value": "b" }
+          },
+          {
+            "path": "",
+            "type": "child_added",
+            "name": "c",
+            "prevName": null,
```
maneesht, line 4219: This should be `b`. Your test should be failing locally. Can you re-merge my branch into yours?

packages/database/test/helpers/syncPointSpec.json
```diff
@@ -4179,6 +4179,249 @@
...
+            "data": { ".value": "c" }
+          },
+          {
+            "path": "",
+            "type": "child_added",
+            "name": "d",
+            "prevName": null,
```
maneesht, line 4410: This should be `c`

packages/database/src/core/view/filter/LimitedFilter.ts
```diff
@@ -119,19 +125,17 @@ export class LimitedFilter implements NodeFilter {
...
-            inRange =
-              this.index_.compare(next, this.rangedFilter_.getEndPost()) <= 0;
-          }
+          const inRange =
+            this.withinDirectionalStart(next) &&
+            this.withinDirectionalEnd(next);
           if (inRange) {
             filtered = filtered.updateImmediateChild(next.name, next.node);
             count++;
+          } else if (!this.withinDirectionalStart(next)) {
```
maneesht, line 134: More of a question than anything else. Why did this need to get added? And it looks like we're adding an additional check on line 130

packages/database/src/core/view/filter/LimitedFilter.ts
```diff
@@ -300,4 +292,26 @@ export class LimitedFilter implements NodeFilter {
...
   }
+
+  private withinDirectionalStart = (node: NamedNode) =>
+    this.reverse_ ? this.withinEndPost(node) : this.withinStartPost(node);
+
+  private withinDirectionalEnd = (node: NamedNode) =>
+    this.reverse_ ? this.withinStartPost(node) : this.withinEndPost(node);
+
+  private withinStartPost = (node: NamedNode) => {
+    const compareRes = this.index_.getCompare()(
```
maneesht, line 303: nit: I don't like having two parentheses side by side, could we just use compare directly here?

packages/database/src/core/view/filter/LimitedFilter.ts
```diff
@@ -300,4 +292,26 @@ export class LimitedFilter implements NodeFilter {
...
+      this.rangedFilter_.getStartPost(),
+      node
+    );
+    return this.startIsInclusive_ ? compareRes <= 0 : compareRes < 0;
+  };
+
+  private withinEndPost = (node: NamedNode) => {
+    const compareRes = this.index_.getCompare()(
```
maneesht, line 311: Same here.

packages/database/test/helpers/syncPointSpec.json
IanWyszynski: The test is actually passing locally on https://github.com/firebase/firebase-js-sdk/pull/6706/commits/331085549fd42a8b679f149c16c65b83be25e4df. It, and the next one, starts failing once I pull https://github.com/firebase/firebase-js-sdk/pull/6786/commits/888a98df4f043f957bf61b0057f89b6233985c4e. I think the previous spec tests are somehow missing events.

IanWyszynski: Perhaps we were accidentally only validating the last one: https://github.com/firebase/firebase-js-sdk/pull/6786/files#diff-872d46e57e4152b3a7bd18b86102a8428117be4d55ee65e287439de27b7ee0efL319.

packages/database/src/core/view/filter/LimitedFilter.ts
IanWyszynski: I think this could be simplified to:
```
        while (iterator.hasNext() && count < this.limit_) {
          const next = iterator.getNext();
          if (!this.withinDirectionalStart(next)) {
            continue;
          }
          if (!this.withinDirectionalEnd(next)) {
            break;
          }
          filtered = filtered.updateImmediateChild(next.name, next.node);
          count++;
        }
```

Currently, it's mimicing [LimitedRangeView](https://github.com/firebase/firebase-server/blob/master/firebase-server/firebase-core/src/main/scala/com/firebase/core/namespace/view/LimitedRangeView.scala#L305-L312) on the backend. 

We considered teaching [`iteratorFrom`](https://github.com/firebase/firebase-server/blob/42585d4169124e76582bd07a7bf06e91bfaada5f/firebase-server/firebase-util/src/main/scala/com/firebase/util/data/Query.scala#L142) and `reverseIteratorFrom` about inclusive/exclusive bounds, but there are a few different implementations on the backend and it seemed like more trouble than it's worth.

packages/database/test/helpers/syncPointSpec.json
maneesht: Sorry, I mean when you pull in my branch, it should fail

tohhsinpei: Oop, yeah, it just started failing. Will fix.

packages/database/test/helpers/syncPointSpec.json
tohhsinpei: Done.

packages/database/src/core/view/filter/LimitedFilter.ts
tohhsinpei: Done.

packages/database/src/core/view/filter/LimitedFilter.ts
tohhsinpei: Done.

packages/database/src/core/view/filter/LimitedFilter.ts
tohhsinpei: Done.

maneesht (approved):
LGTM, thanks!

hsubox76 (approved):

Issue #6613, opened 2022-09-18
Uncaught Error in snapshot listener: {"code":"failed-precondition","name":"FirebaseError"}


MubashirWaheed:
### [REQUIRED] Describe your environment

  * Operating System version: Ubuntu 20.04.4 LTS
  * Browser version: 104.0 (64-bit) Mozilla Firefox for Ubuntu canonical - 1.0
  * Firebase SDK version: firebase: 9.9.4
  * Firebase Product: firestore 

#### Steps to reproduce:

I am not getting the link to add an index for the compound queries in my console. and instead, I am getting this error in the console 
`[2022-09-18T05:53:50.590Z]  @firebase/firestore: Firestore (9.9.4): Uncaught Error in snapshot listener: {"code":"failed-precondition","name":"FirebaseError"}` 
After going through stackoverfollow I manually created the composite indexes in the firebase console under the firestore> indexes. 
1) index Descending, createdBy Descending
2) index Descending, createdBy Arrays

for both indexes 
- collection ID is `posts`
- Query scope `collection`
- status `enabled`

When I remove the orderBy there is no error. 

#### Relevant Code:
`following` is an array with string values and the `index` field has number as its value 
ps: I am using react. 
```javascript
// TODO(you): code here to reproduce the problem
    useEffect(()=>{
        const q = query(
        collection(db, "posts"),
        orderBy("index", "desc"),
        where("createdBy", "in", following)
      );
        const unsubscribe = onSnapshot(q, (querySnapshot) => {
        const data = [];
        querySnapshot.forEach((doc) => {
          data.push(doc.data());
        });
        setPosts(data);
      });
}, [])
```

IvoBiaus:
Im getting the same error after deleting my `yarn.lock` and running `yarn install` which in my case, among other things, it updates `@firebase/firestore` from 3.4.0 to 3.5.0. After doing that, if i try to login in my app and request data from firebase all my rules fail because the `request.auth` is sending null all of the sudden, and i get the error: 
`@firebase/firestore: Firestore (9.10.0): Uncaught Error in snapshot listener: {"code":"permission-denied","name":"FirebaseError"}`

Screenshot from Firebase Emulator Suit:
![image](https://user-images.githubusercontent.com/43274372/191028931-2f5395d7-c770-48fc-842e-53d4de8d64e8.png)

MarkDuckworth:
[...]

PR #6712, opened 2022-10-20
Firestore: Fix "missing index" error message

milaGGL:
Fix a bug where missing index error message is not displayed as expected. 

#### Expected error message
It should display "The query requires an index. Please create it here: http://.....". By clicking the link, user will be directed to console where they can create the required index.

#### Current error message: 
- Full SDK: "Server responded with status 400"
- Lite SDK:"Bad request" 

#### Cause
when server respond to the request with missing index, error object is wrapped inside an array. By trying to read the error message field directly from the array, it always fails and get "undefined".
So, extract the error object from the array before reading the error message field.

Fixes [#6613](https://github.com/firebase/firebase-js-sdk/issues/6613)


packages/firestore/src/platform/browser_lite/fetch_connection.ts
```diff
@@ -69,9 +69,11 @@ export class FetchConnection extends RestConnection {
     }
 
     if (!response.ok) {
+      const responseBody = await response.json();
       throw new FirestoreError(
         mapCodeFromHttpStatus(response.status),
-        'Request failed with error: ' + response.statusText
+        //response body is an array with error object
+        'Request failed with error: ' + responseBody[0].error.message
       );
```
milaGGL, line 77: statusText here is only "Bad request". 
Backend rejected the request, but in this file, the this.fetchImpl didn't fail. It passed through the try catch block, and error message send by backend is stored inside the resposne.json.

Wonder
1.  if there is a way to make the fetchImp fail if the backend rejected the request, so that we can get the error message inside the catch block.
2. if we can't , can we get the error message in some other ways, not from response.json(). Does the Response type include fields sth like error, or statusMessage 🤔

dconeybe: This is basically exactly what webchannel_connection.ts does, so I think your solution is quite reasonable.

https://github.com/firebase/firebase-js-sdk/blob/941ad29660a386059e5618b514d43cc5d409e642/packages/firestore/src/platform/browser/webchannel_connection.ts#L90-L131

packages/firestore/src/platform/browser_lite/fetch_connection.ts
```diff
@@ -69,9 +69,23 @@ export class FetchConnection extends RestConnection {
     }
 
     if (!response.ok) {
+      let errorMessage;
+      try {
+        // Try to get any error message text from server response.
+        //response.json returns an error object inside an array.
+        const [errorResponse] = (await response.json()) as [
```
dconeybe, line 76: Reduce the try/catch block to _only_ contain the call to `await response.json()`.

packages/firestore/src/platform/browser_lite/fetch_connection.ts
```diff
@@ -69,9 +69,23 @@ export class FetchConnection extends RestConnection {
     }
 
     if (!response.ok) {
+      let errorMessage;
+      try {
+        // Try to get any error message text from server response.
+        //response.json returns an error object inside an array.
```
dconeybe, line 75: Remove these comments since they are just explaining what the code is clearly doing. Also on line 87.

packages/firestore/src/platform/browser_lite/fetch_connection.ts
```diff
@@ -69,9 +69,23 @@ export class FetchConnection extends RestConnection {
...
+        ];
+        if (errorResponse.error?.message) {
+          errorMessage = errorResponse.error.message;
+        }
+      } catch (_ignored) {}
       throw new FirestoreError(
         mapCodeFromHttpStatus(response.status),
-        'Request failed with error: ' + response.statusText
+        //response body is an array with error object
+        'Request failed with error: ' + errorMessage ?? response.statusText
```
dconeybe, line 88: Use string interpolation here to improve readability.

i.e. ``` `Request failed with error: ${errorMessage ?? response.statusText}` ```

dconeybe (changes requested):
This fix looks perfect to me. Nice work! Just a few minor comments.

packages/firestore/src/platform/browser_lite/fetch_connection.ts
milaGGL: As I am not 100% confident that the response.json() will always return [ {error?: { message?: string };}];, would it be better to keep it in the try/catch block?  If it succeed, display errorMessage. If it fail, then display statusText as minimum

dconeybe: It would be better to guard it with runtime checks than rely on exceptions because it's more explicit about what you are doing, and, therefore, more readable.

Maybe do something like this:
```js
let errorResponse;
try {
  errorResponse = await response.json();
} catch (_ignored) {
}

let errorMessage;
if (errorResponse) {
  errorMessage = Array.from(errorResponse).shift()?.error?.message
}
```

packages/firestore/src/platform/browser/webchannel_connection.ts
```diff
@@ -97,8 +97,11 @@ export class WebChannelConnection extends RestConnection {
                 xhr.getResponseText()
               );
               if (status > 0) {
```
milaGGL, line 99: extract the error object if json response is an array

packages/firestore/src/util/async_observer.ts
```diff
@@ -45,7 +45,7 @@ export class AsyncObserver<T> implements Observer<T> {
     if (this.observer.error) {
       this.scheduleEvent(this.observer.error, error);
     } else {
-      logError('Uncaught Error in snapshot listener:', error);
+      logError('Uncaught Error in snapshot listener:', error.toString());
```
milaGGL, line 48: fixes [#6613](https://github.com/firebase/firebase-js-sdk/issues/6613)

packages/firestore/test/integration/api/aggregation.test.ts
```diff
@@ -118,4 +119,24 @@ apiDescribe('Count quries', (persistence: boolean) => {
       );
     });
   });
+
+  // Only verify the error message for missing indexes when running against
+  // production, since the Firestore Emulator does not require index creation
+  // and will, therefore, never fail in this situation.
+  // eslint-disable-next-line no-restricted-properties
```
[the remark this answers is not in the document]
milaGGL: I remember it complained last time when I use it.skip. I will try this one out.

.changeset/little-rats-relax.md
```diff
@@ -0,0 +1,6 @@
+---
+'@firebase/firestore': patch
+'firebase': patch
+---
+
+Fix "missing index" error message to include the create composite index link.
```
dconeybe, line 6: Slight re-word: "...include the link to create the composite index"

dconeybe (changes requested):
Just one tiny change to the changelog entry. Otherwise LGTM.

dconeybe (approved):

Issue #6466, opened 2022-07-22
Notification does not show icon.


NorbertBodziony:
### [REQUIRED] Describe your environment

  * Operating System version: macOs 12.3
  * Browser version: None
  * Firebase SDK version: v10.2.0
  * Firebase Product: FCM (auth, database, storage, etc)

#### Steps to reproduce:
Icon passed via `notification` does not show on device. Notification itself works. 
Notification triggered via test firebase UI works and shows correct image.

#### Relevant Code:

```javascript
await messaging.sendToDevice(token, {
      data: data,
      notification: { ...notification, icon: 'https://synthetify.io/icons/sny.png' },
    })
```

jbalidiong:
Hi @NorbertBodziony, thanks for reaching out. I was able to reproduce the behavior. Let me check what we can do for this issue or bring someone here that can provide more context about it. I’ll update this thread if I have any information to share.

I've added this to our internal tracker (b/241885017).

PR #6722, opened 2022-10-24
Expose 'icon' field from the Firebase Messaging SDK

zwu52:
Fixes for b/241885017.

Would patch https://github.com/firebase/firebase-js-sdk/issues/6466

### DIFF
--- a/packages/messaging/src/helpers/externalizePayload.ts
+++ b/packages/messaging/src/helpers/externalizePayload.ts
@@ -60,6 +60,11 @@ function propagateNotificationPayload(
   if (!!image) {
     payload.notification!.image = image;
   }
+
+  const icon = messagePayloadInternal.notification!.icon;
+  if (!!icon) {
+    payload.notification!.icon = icon;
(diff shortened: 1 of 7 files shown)


packages/messaging/src/helpers/externalizePayload.ts
```diff
@@ -60,6 +60,11 @@ function propagateNotificationPayload(
   if (!!image) {
     payload.notification!.image = image;
   }
+
+  const icon = messagePayloadInternal.notification!.icon;
```
dwyfrequency, line 64: I see this is just matching the style from above. In the future, I think using the `?.` instead of `!.` will be better

dwyfrequency:
Can you also run the formatter locally? Seems to be an issue there

packages/messaging/src/helpers/externalizePayload.ts
zwu52: sg. will make a pr to batch fix

Issue #6657, opened 2022-10-06
Cannot read property 'controller' of undefined


nilptr:
### [REQUIRED] Describe your environment

  * Operating System version: Mac OS X 10.15.7
  * Browser version: Chrome 98.0.4758
  * Firebase SDK version: 9.6.1
  * Firebase Product: performance

Actually, this error occurs on more than one specific operating system and browser version. Here are browser and OS versions we can see in Sentry.

Browsers:
- Chrome 98.0.4758
- Chrome 99.0.4844
- Chrome 96.0.4664
- Chrome 105.0.0
- Firefox 95.0
- Chrome 103.0.0
- Chrome 106.0.0
- Chrome 96.0.4606

OS:
- Mac OS X 10.15.7
- Windows 10
- Mac OS X 10.14.6
- Mac OS X 11.6
- Mac OS X 12.2.1
- Mac OS X 12.6
- Mac OS X 10.15
- Mac OS X 10.14.3
- Mac OS X 10.14.4
- Mac OS X 10.15.6

### [REQUIRED] Describe the problem

We got  `Cannot read property 'controller' of undefined` errors in Sentry recently.
<img width="1570" alt="image" src="https://user-images.githubusercontent.com/16560775/194201257-4ebf362c-2fe1-46b2-954d-7f748715b2fa.png">

The stack trace shows it comes from here.
https://github.com/firebase/firebase-js-sdk/blob/cdada6c68f9740d13dd6674bcb658e28e68253b6/packages/performance/src/utils/attributes_utils.ts#L60-L71
Although it looks already defensive enough, it doesn't catch the case that `'serviceWorker' in navigator` but `navigator.serviceWorker === undefined`.

#### Steps to reproduce:

N/A
#### Relevant Code:

https://github.com/firebase/firebase-js-sdk/blob/cdada6c68f9740d13dd6674bcb658e28e68253b6/packages/performance/src/utils/attributes_utils.ts#L60-L71

DellaBitta:
Hi @nilptr,

Is this still an issue after the PR was released? Thanks!

nilptr:
Hi @DellaBitta, after the release, we never saw this error again.

Thanks!

PR #6723, opened 2022-10-24
Account for a `navigator` that has a key of `serviceWorker` but an undefined value

dwyfrequency:
https://github.com/firebase/firebase-js-sdk/issues/6657 Cannot read property 'controller' of undefined


jposuna (approved):

egilmorez (approved):
LG, thanks!

PR #6728, opened 2022-10-26
Add 'icon' to '/messaging-types' 

zwu52:
This follows https://github.com/firebase/firebase-js-sdk/pull/6722 to supplement 'icon' also to the '/messaging-types'  dir.

ref: b/241885017

### DIFF
--- a/packages/messaging-types/index.d.ts
+++ b/packages/messaging-types/index.d.ts
@@ -31,6 +31,7 @@ export interface NotificationPayload {
   title?: string;
   body?: string;
   image?: string;
+  icon?: string;
 }
 
 export interface FcmOptions {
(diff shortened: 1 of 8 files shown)


egilmorez (approved):
LGTM, thanks!

packages/messaging-types/index.d.ts
```diff
@@ -31,6 +31,7 @@ export interface NotificationPayload {
   title?: string;
   body?: string;
   image?: string;
+  icon?: string;
```
hsubox76, line 34: Just out of curiosity I looked to see if this package (`@firebase/messaging-types`) is being used anywhere and it doesn't seem to be? I see it as a dependency in packages/messaging/package.json but I don't see it imported anywhere in the source code. The other instances of [product]-types packages we have are legacy typings that are only used by [product]-compat but messaging-compat doesn't seem to use anything from it, and regular v9 messaging also doesn't. Maybe we could remove it? (in a later PR)

hsubox76 (approved):

Issue #6659, opened 2022-10-06
transaction.set(...) fails the transaction without retry when document created by another app


MarkDuckworth:
### [REQUIRED] Describe your environment

  * Firebase SDK version: 9.10.0
  * Firebase Product: firestore

### [REQUIRED] Describe the problem

`transaction.get(ref)` followed by `transaction.set(ref, {...})` is causing the transaction to fail with error 'Document already exists' if the document was created by another app between the get and set.

Expected behavior is for the transaction to fail the first attempt because the document changed (was created) between the read and write. But it's also expected that the transaction will automatically retry. 

I see the expected behavior in v9.6.11, and the error in v9.10.0. There appears to be a regression somewhere in between these versions.

#### Steps to reproduce:
1. Before repro, execute `clean()` to delete the document.
2. Execute method `transact()`. 
3. During the sleep in `transact()`, execute method `write()`. This might be done by a second browser.
4. Observe that the transaction from step 2 fails without retry. But the transaction retries and is successful in 9.6.11.

#### Relevant Code:

```javascript

async function sleep(ms) {
    return new Promise((resolve) => {
        setTimeout(resolve, ms);
    });
}

async function clean() {
    console.log("starting clean")
    await deleteDoc(ref);
    console.log("finished clean")
}

async function write() {
    await setDoc(ref, {created: (new Date()).toUTCString()})
    console.log("Doc created")
}

async function transact() {
    await runTransaction(db, async (tx) => {
        const snapshot = await tx.get(ref);
        console.log("doc read")

        if (snapshot.exists()) {
            console.log("document exists - updating last write")
            tx.update(ref, {lastWrite: (new Date()).toUTCString()})
            return;
        }

        console.log("sleeping")
        await sleep(10000);
        console.log("wake from sleep")

        tx.set(ref, {created: (new Date()).toUTCString()});
    });

    console.log("Transaction successful");
}

async function reproduce()
{
    await clean();
    setTimeout(write, 2000);
    transact();
}

reproduce();
```

tom-andersen:
Thank you very much for the example code! This made reproduction of problem very easy.

We have a fix planned. I will let you know when it becomes available.

milaGGL:
Hi @MarkDuckworth, the bug fix has been merged and will be included in the next release.

dconeybe:
FYI v9.14.0 released on November 10, 2022 contains the fix
https://firebase.google.com/support/release-notes/js#version_9140_-_november_10_2022

PR #6729, opened 2022-10-26
Fix transaction.set() failure without retry on "already-exists" error

milaGGL:
#### Problem
Transaction.get(ref) followed by transaction.set(ref, {...}) is causing the transaction to fail with error 'Document already exists' if the document was created by another app between the get and set.

#### Cause
'already-exists' is considered to be a permanent error for transactions, therefore will not be retried.
https://github.com/firebase/firebase-js-sdk/blob/f86f7de090ae7f61b795d0dbbe9b3765590e0286/packages/firestore/src/core/transaction_runner.ts#L115

Fix Issue: https://github.com/firebase/firebase-js-sdk/issues/6659

### DIFF
--- a/packages/firestore/test/integration/api/transactions.test.ts
+++ b/packages/firestore/test/integration/api/transactions.test.ts
@@ -629,6 +629,30 @@ apiDescribe('Database transactions', (persistence: boolean) => {
     });
   });
 
+  it('retries when document already exists', () => {
+    return withTestDb(persistence, async db => {
+      let retryCounter = 0;
+      const docRef = doc(collection(db, 'nonexistent'));
+
+      await runTransaction(db, async transaction => {
+        ++retryCounter;
+        const snap1 = await transaction.get(docRef);
+
+        if (retryCounter === 1) {
+          expect(snap1.exists()).to.be.false;
+          // On the first attemp, create a doc before transaction.set(), so that
+          // the transaction fails with "already-exists" error, and retries.
+          await setDoc(docRef, { count: 1 });
+        }
+
+        transaction.set(docRef, { count: 2 });
+      });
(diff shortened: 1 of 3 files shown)


packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -629,6 +629,30 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
+      let retryCounter = 0;
+      const docRef = doc(collection(db, 'nonexistent'));
+
+      await runTransaction(db, async transaction => {
+        ++retryCounter;
+        const snap1 = await transaction.get(docRef);
+
+        if (retryCounter === 1) {
+          expect(snap1.exists()).to.be.false;
+          // On the first attemp, create a doc before transaction.set(), so that
```
dconeybe, line 643: nit: "attemp" -> "attempt"

packages/firestore/test/integration/api/transactions.test.ts
```diff
@@ -629,6 +629,30 @@ apiDescribe('Database transactions', (persistence: boolean) => {
...
+          // the transaction fails with "already-exists" error, and retries.
+          await setDoc(docRef, { count: 1 });
+        }
+
+        transaction.set(docRef, { count: 2 });
+      });
+      expect(retryCounter).to.equal(2);
+      const result = await getDoc(docRef);
```
dconeybe, line 651: nit: Please rename `result` to `snapshot`.

dconeybe (changes requested):
LGTM. Just two very minor comments!

dconeybe (approved):

tom-andersen (approved):
Nice

PR #6735, opened 2022-10-28
fix(AppCheck): Fixed getToken promise not being cleared

### LINKED ISSUE #6734: Sometimes old AppCheck tokens continue to be returned.

  * Operating System version: _____
  * Browser version: _____
  * Firebase SDK version: 9.12.0
  * Firebase Product: appcheck

If certain conditions are met, even if `getToken` is executed with forceRefresh, `getToken` will only return the same token, and if more than 1 hour, the token's expiration time, elapses, the application will crash due to the token expiration.
It is a small project, but according to sentry it appears to be occurring several times a day.

Debugging the internal code showed that the `exchangeTokenPromise` doesn't cleared and kept returning the `exchangeTokenPromise` return value.

related pull request: https://github.com/firebase/firebase-js-sdk/pull/6617

Translated with www.DeepL.com/Translator (free version)

#### Relevant Code:

```javascript
// TODO(you): code here to reproduce the problem
```

ishowta:
Fix #6734

Immutable assignment by `setState` failed to clear the `exchangeTokenPromise` by assigning undefined to it.
https://github.com/firebase/firebase-js-sdk/blob/e2a90bf678eb6fa505f8b2f627e03cff622607b5/packages/app-check/src/internal-api.ts#L126

There may be a better way, but I wrapped the `exchangeTokenPromise` in an object to prevent it from deep copy due to spread assignments.

### DIFF
--- a/packages/app-check/src/internal-api.test.ts
+++ b/packages/app-check/src/internal-api.test.ts
@@ -368,6 +368,103 @@ describe('internal api', () => {
       });
     });
 
+    it('no dangling exchangeToken promise internal', async () => {
+      const appCheck = initializeAppCheck(app, {
+        provider: new ReCaptchaV3Provider(FAKE_SITE_KEY)
+      });
+
+      setState(app, {
+        ...getState(app),
+        token: fakeRecaptchaAppCheckToken,
+        cachedTokenPromise: undefined
+      });
+
+      stub(reCAPTCHA, 'getToken').returns(Promise.resolve(fakeRecaptchaToken));
+      stub(client, 'exchangeToken').returns(
+        Promise.resolve({
+          token: 'new-recaptcha-app-check-token',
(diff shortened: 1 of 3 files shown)


hsubox76:
Just wanted to double check, were you able to put this code into the buggy app and observe that it fixed the bug you were seeing?

ishowta:
I'm sorry, but we do not do that.

hsubox76:
So thanks for pointing out that state setting is inconsistently done in this codebase, I've decided to do a more systematic fix of it with https://github.com/firebase/firebase-js-sdk/pull/6740 since I think it would be a bad idea to mix in more mutation. I added your tests into that PR and it passes, so if the test is a good replication of your situation it should fix the bug. Feel free to take a look at that PR and let me know what you think.

PR #6740, opened 2022-10-31
Fix App Check state setting and promise clearing

hsubox76:
Fixes https://github.com/firebase/firebase-js-sdk/issues/6734 (probably)?

Based on solution suggested in https://github.com/firebase/firebase-js-sdk/pull/6735

`setState()` and `state` usage is already pretty inconsistent throughout the App Check codebase (sometimes immutable, sometimes mutated) and I don't think the best answer is mixing in more mutation. Additionally, the way `setState()` is used depends on extending a temporary version of `state` that, due to a lot of async code, may be stale by the time it `setState()` is called.

Added a method called `setStateProperty()` which starts with a fresh `getState()` clone, directly sets a property on the clone (which ensures `undefined` will override any previous set values, unlike an assign/extend), then uses setState to put the entire new state object back in the `APP_CHECK_STATES` Map.

Replaced `setState` with this new method in the problem areas causing this bug, as well as everywhere else it is appropriate.

### DIFF
--- a/.changeset/spicy-bags-sparkle.md
+++ b/.changeset/spicy-bags-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@firebase/app-check': patch
+---
+
+Clear App Check exchange promise correctly after request succeeds.
--- a/packages/app-check/src/api.ts
+++ b/packages/app-check/src/api.ts
@@ -23,7 +23,7 @@ import {
   PartialObserver
 } from './public-types';
 import { ERROR_FACTORY, AppCheckError } from './errors';
-import { getState, setState, AppCheckState, getDebugState } from './state';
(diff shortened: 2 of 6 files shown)


ishowta:
I think it is better than my code, but it seems to be prone to bugs because the information rewritten in the setter is not immediately reflected in the local state.

Performance may be somewhat degraded, but I thought it would be safer and more readable to stop using setter and getter and use mutex to execute the entire function exclusively.

hsubox76:
That's true, we would have to use `getState()` before every reference to `state` to be sure. I still think whichever solution we go with should be global and not a mix of mutable and immutable state so perhaps state should just be rewritten to be directly mutable across the board, I'll take a look at doing that.

dwyfrequency (approved):
LGTM

PR #6745, opened 2022-11-01
Have Renovate ignore auth demo projects

hsubox76:
Use https://docs.renovatebot.com/configuration-options/#ignorepaths to have Renovate ignore auth demo projects. If these need to be regularly updated, we should add a process to do it on release. If they only need to be updated whenever someone decides to use them, they can be updated then. We don't do Renovate updates on a regular schedule so it's a solution that doesn't really fit either case.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -21,6 +21,11 @@
     "typedoc",
     "@microsoft/tsdoc"
   ],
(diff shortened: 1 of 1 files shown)


dwyfrequency (approved):
LGTM

PR #6746, opened 2022-11-01
Updated base64 error to be more descriptive

maneesht:
When users of Expo and other similar environments don't support `atob` or `btoa`, polyfills are required, but the SDK doesn't provide this information to the user. 

This PR checks whether `atob` is available and throws an error if it is not.

### DIFF
--- a/packages/storage/src/implementation/error.ts
+++ b/packages/storage/src/implementation/error.ts
@@ -261,6 +261,13 @@ export function noDownloadURL(): StorageError {
   );
 }
 
+export function missingPolyFill(polyFill: string): StorageError {
+  return new StorageError(
+    StorageErrorCode.UNSUPPORTED_ENVIRONMENT,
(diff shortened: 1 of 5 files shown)


tonyjhuang (approved):

packages/storage/src/implementation/error.ts
```diff
@@ -261,6 +261,13 @@ export function noDownloadURL(): StorageError {
   );
 }
 
+export function missingPolyFill(polyFill: string): StorageError {
+  return new StorageError(
+    StorageErrorCode.UNSUPPORTED_ENVIRONMENT,
+    `${polyFill} is missing. Please install the right polyfills. Visit https://firebase.google.com/docs/web/environments-js-sdk#polyfills for more information.`
```
egilmorez, line 267: Suggest "`${polyFill} is missing. Make sure to install the required polyfills. See..."

egilmorez (approved):
A nit on the wording, but LG, thanks!

PR #6748, opened 2022-11-02
chore: use .mjs extension for modules

### LINKED ISSUE #6747: Jest does not work with Firebase out of the box
Currently there [are ".js" files generated](https://github.com/firebase/firebase-js-sdk/blob/9e9ee7ba3383237e0b92ad8183dd0f12640f4f3a/packages/firebase/package.json#L88) which are "es modules". Since there is no `"type": "module"` in the [package.json](https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/package.json), this is [simply illegal](https://github.com/facebook/jest/issues/11809#issuecomment-1048704932) as far as I can tell. It creates issues when trying to load firebase with `jest` and `jsdom`.

Either we have to make this a module or always use `.mjs` for all modules.

levino:
fix #6747


levino:
Will the checks run on a draft?

Issue #6736, opened 2022-10-28
@firebase/auth includes selenium-webdriver and tslib as dependencies


jekh:
### [REQUIRED] Describe your environment
  * Firebase SDK version: 9.13.0
  * Firebase Product: auth

### [REQUIRED] Describe the problem
`@firebase/auth` includes `selenium-webdriver` and `tslib` as runtime `dependencies`. This causes them to get pulled into production dependency bundles unnecessarily. These should probably be `devDependencies` (I'm assuming they're used at build and test time?), unless there is something very strange going on that requires `@firebase/auth` to parse typescript or use selenium at runtime.  If there is a runtime dependency on `selenium-webdriver` or `tslib`, that is probably worth investigating too.

https://github.com/firebase/firebase-js-sdk/blob/master/packages/auth/package.json#L104
![image](https://user-images.githubusercontent.com/9272923/198700919-7c14976c-71bc-4b28-9149-8e256a21dae9.png)

#### Steps to reproduce:
1. Install `firebase`
2. Run `pnpm why selenium-webdriver` (or yarn/npm equivalent)
3. See that `selenium-webdriver` is included in the dependency (not `devDependency`) tree
```
dependencies:
firebase 9.13.0
├─┬ @firebase/auth 0.20.11
│ └── selenium-webdriver 4.5.0
└─┬ @firebase/auth-compat 0.2.24
  ├─┬ @firebase/auth 0.20.11
  │ └── selenium-webdriver 4.5.0
  └── selenium-webdriver 4.5.0
```

Simply moving `tslib` and `selenium-webdriver` to `devDependencies` should fix this.

hsubox76:
I think we can move `selenium-webdriver` but `tslib` is actually called at runtime in the bundles (if you check the bundle in the dist folder you'll find it). If your bundler does tree shaking I think it should be able to only bring in the necessary modules from tslib, but not 100% sure.

I don't think it should be bringing in any selenium-webdriver code at all? What bundler are you using? Does it automatically bundle all dependencies into production code? I thought putting unnecessary packages into `dependencies` would only impact npm installation, making you download more than necessary, but wouldn't end up in the production bundle unless actually referenced in the source code.

But I think we should move `selenium-webdriver` for clarity and to avoid excessive npm downloads by our users.

jekh:
Thanks for looking into this! I do see that tslib is used for downleveling, so that does make sense, my apologies.

[...]

PR #6762, opened 2022-11-07
Add selenium-webdriver to devDependencies

prameshj:
Fixes https://github.com/firebase/firebase-js-sdk/issues/6736


sam-gc (approved):
Does compat have the same issue?

prameshj:
> Does compat have the same issue?

good qn, i fixed compat as well.

hsubox76 (approved):

PR #5390, opened 2021-08-27
Clean up things from v9 beta

Feiyang1:
- remove scripts that are not needed anymore
- remove `-exp` references
- clean up scripts folder

PR #6772, opened 2022-11-09
Update parent path for check_changeset script

hsubox76:
This path seems to have not been updated when the script was moved inside a subdirectory in https://github.com/firebase/firebase-js-sdk/pull/5390


dwyfrequency (approved):
LGTM

PR #6781, opened 2022-11-10
Add createTime to Document.

ehsannas:


packages/firestore/src/model/document.ts
```diff
@@ -101,6 +101,13 @@ export interface Document {
    */
   readonly readTime: SnapshotVersion;
 
+  /**
+   * The timestamp at which the document was created. This value increases
```
wu-hui, line 105: Is this comment copied from proto? Looks pretty thorough.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1958,6 +1958,33 @@ function genericLocalStoreTests(
...
+        .toReturnChanged(
+          doc('col/doc1', 13, { foo: 'newBar' }, 5).setHasCommittedMutations()
+        )
+        .toContain(
+          doc('col/doc1', 13, { foo: 'newBar' }, 5).setHasCommittedMutations()
+        )
+        .finish()
+    );
+  });
+
```
wu-hui, line 1987: I wonder what happens when:
1. we create a document with set locally.

2. that set gets accepted first, leading us to having a document with createTime=0 in the cache. This is problematic for count because it will "discount" the document. We should probably special case for SetMutation acknowledgement: if the remote doc has no createTime (or there is no createTime), we should use the document's version as createTime when we update remote document cache.

3. Then will the following remote event have a correct createTime for the document? will we save it if they do? We should be good because watch stream is separate, and it will continue to send the new document in the remote event. And we are saving createTime now. 

Let's work on this together today!

packages/firestore/src/model/document.ts
ehsannas: Yes, copied from the proto.

packages/firestore/test/unit/local/local_store.test.ts
```diff
@@ -1958,6 +1997,203 @@ function genericLocalStoreTests(
...
+        )
+        .toContain(
+          doc('col/doc1', 13, { foo: 'newBar' }, 5).setHasCommittedMutations(),
+          compareDocsWithCreateTime
+        )
+        .finish()
+    );
+  });
+
+  it('saves updateTime as createTime when receives ack for creating a new doc', () => {
```
wu-hui, line 2037: Can we add one more variant on top of this:

after the set mutation ack, then comes a remote event with createTime = 5. This should reset the time to 5.

Also, what if the remote event comes first, then comes the ack, I'd expect the createTime is still 5 in this case (i think your PR also does this), but let's add one more to verify.

wu-hui (approved):
Two more tests requested, but LGTM so you can merge after adding them.

packages/firestore/test/unit/local/local_store.test.ts
ehsannas: Good idea. Done. Also merged with tip-of-tree (which should hopefully make the Firefox CI bot happy).

PR #6784, opened 2022-11-12
 integration tests for totp

bhparijat:

### Discussion

### Testing

### API Changes

    can discuss it together.

### DIFF
--- a/packages/auth/test/helpers/integration/helpers.ts
+++ b/packages/auth/test/helpers/integration/helpers.ts
@@ -22,11 +22,7 @@ import { Auth, User } from '../../../src/model/public_types';
 import { getAuth, connectAuthEmulator } from '../../../'; // Use browser OR node dist entrypoint depending on test env.
 import { _generateEventId } from '../../../src/core/util/event_id';
 import { getAppConfig, getEmulatorUrl } from './settings';
-import {
-  getOobCodes,
-  OobCodeSession,
-  resetEmulator
-} from './emulator_rest_helpers';
+import { resetEmulator } from './emulator_rest_helpers';
 import totp from 'totp-generator';
 interface IntegrationTestAuth extends Auth {
   cleanUp(): Promise<void>;
@@ -82,15 +78,9 @@ export function getTestInstance(requireEmulator = false): Auth {
   return auth;
 }
 
-export async function cleanUpTestInstance(
-  auth: Auth,
-  tests?: string
-): Promise<void> {
+export async function cleanUpTestInstance(auth: Auth): Promise<void> {
   await auth.signOut();
-
-  if (typeof tests === 'undefined') {
-    await (auth as IntegrationTestAuth).cleanUp();
-  }
+  await (auth as IntegrationTestAuth).cleanUp();
 }
 
 function stubConsoleToSilenceEmulatorWarnings(): sinon.SinonStub {
@@ -104,11 +94,6 @@ function stubConsoleToSilenceEmulatorWarnings(): sinon.SinonStub {
       originalConsoleInfo(...args);
     }
   });
-}
-
-export async function code(toEmail: string): Promise<OobCodeSession> {
-  const codes = await getOobCodes();
-  return codes.reverse().find(({ email }) => email === toEmail)!;
 }
 
 export function getTotpCode(
--- a/packages/auth/test/helpers/integration/helpers.ts
+++ b/packages/auth/test/helpers/integration/helpers.ts
@@ -23,7 +23,12 @@ import { getAuth, connectAuthEmulator } from '../../../'; // Use browser OR node
 import { _generateEventId } from '../../../src/core/util/event_id';
 import { getAppConfig, getEmulatorUrl } from './settings';
 import { resetEmulator } from './emulator_rest_helpers';
+import { StartTotpMfaEnrollmentResponse } from '../../../src/api/account_management/mfa';
 
+//import * as otpauth from "https://deno.land/x/otpauth@v9.0.1/dist/otpauth.esm.js";
+//// <reference path="helpers.d.ts"/>
+const totp = require('totp-generator');
+//import * as totp from 'otpauth';
 interface IntegrationTestAuth extends Auth {
   cleanUp(): Promise<void>;
 }
@@ -92,4 +97,19 @@ function stubConsoleToSilenceEmulatorWarnings(): sinon.SinonStub {
       originalConsoleInfo(...args);
     }
   });
+}
+
+export async function mockTotp(sharedSecretKey: string, periodSec: number, verificationCodeLength: number){
+  console.log("**** starting to mock totp");
+
+  let digits = 9;
+  let period = 30;
+  let secret = "private";
+  const headers = new Headers();
+  
+  let token = totp(sharedSecretKey, { period: periodSec, digits: verificationCodeLength });
+  console.log("***"+token);
+  
+return token 
+ 
 }
--- a/packages/auth/test/helpers/integration/helpers.ts
+++ b/packages/auth/test/helpers/integration/helpers.ts
@@ -22,8 +22,8 @@ import { Auth, User } from '../../../src/model/public_types';
 import { getAuth, connectAuthEmulator } from '../../../'; // Use browser OR node dist entrypoint depending on test env.
 import { _generateEventId } from '../../../src/core/util/event_id';
 import { getAppConfig, getEmulatorUrl } from './settings';
-import { resetEmulator } from './emulator_rest_helpers';
-import { StartTotpMfaEnrollmentResponse } from '../../../src/api/account_management/mfa';
+import { getOobCodes, OobCodeSession, resetEmulator } from './emulator_rest_helpers';
+import * as config from '../../../../../config/project.json';
 
 //import * as otpauth from "https://deno.land/x/otpauth@v9.0.1/dist/otpauth.esm.js";
 //// <reference path="helpers.d.ts"/>
@@ -62,6 +62,8 @@ export function getTestInstance(requireEmulator = false): Auth {
 
   auth.cleanUp = async () => {
     // If we're in an emulated environment, the emulator will clean up for us
+
+    console.log('Auth cleanup should not be called');
     if (emulatorUrl) {
       await resetEmulator();
     } else {
@@ -81,9 +83,13 @@ export function getTestInstance(requireEmulator = false): Auth {
   return auth;
 }
 
-export async function cleanUpTestInstance(auth: Auth): Promise<void> {
+export async function cleanUpTestInstance(auth: Auth, tests? : string): Promise<void> {
   await auth.signOut();
-  await (auth as IntegrationTestAuth).cleanUp();
+
+  if(typeof tests === 'undefined') {
+    await (auth as IntegrationTestAuth).cleanUp();
+  }
+  
 }
 
 function stubConsoleToSilenceEmulatorWarnings(): sinon.SinonStub {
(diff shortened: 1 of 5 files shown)


packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+  cleanUpTestInstance,
+  getTestInstance,
+  mockTotp,
+  randomEmail
+} from '../../helpers/integration/helpers';
+import { MultiFactorAssertionImpl } from '../../../src/mfa/mfa_assertion';
+
+import { MultiFactorSessionImpl } from '../../../src/mfa/mfa_session';
+import { TotpMultiFactorAssertionImpl, TotpMultiFactorGenerator, TotpSecret } from '../../../src/mfa/assertions/totp';
+import * as MFA from '../../../src/api/account_management/mfa';
```
prameshj, line 36: Can we import just the methods we want?

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import { FirebaseError } from '@firebase/util';
+
+
+
+use(chaiAsPromised);
+use(sinonChai);
+
+const TOTP_COMB_A = {
```
prameshj, line 44: Since this integration test will call the real backend to generate the TOTP response, we don't need these mock values.

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    verificationCodeLength: 30,
+    hashingAlgorithm: 'sha1',
+    periodSec: 30,
+    sessionInfo: 'testsSessionInfo',
+    finalizeEnrollmentTime:  Date.now()
+    },
+    code: '...'
+  };
+
+describe(' Integration tests: Mfa TOTP', () => {
```
prameshj, line 68: I suggest that we have the following test cases inside this:

1) Create user with email/password + enroll in TOTP + login with correct code - success case end-to-end
   In this test, we make sure that enrollment succeeds, then complete login and check that the error 'auth/multi-factor-auth-required' is thrown and supply the otp code - similar to this sample - https://firebase.google.com/docs/auth/web/multi-factor#signing_users_in_with_a_second_factor, but with totp.
2) Create user with email/password + enroll in TOTP +login with incorrect code - fails login. We can reuse the same email/password in test case 1 for this.

3) Unenroll from TOTP for the test case 1 email and ensure that sign in succeeds without asking for OTP code i.e no 'auth/multi-factor-auth-required' error.

You can refer to this test - https://github.com/firebase/firebase-js-sdk/blob/e9fb4037f3908d2a05070de9219aa8a94d287f14/packages/auth/test/integration/flows/email.test.ts#L85, which reuses the same email/password user to test more than one case.


4) Create user with email/password + enroll in TOTP with wrong OTP code - enrollment fails.

prameshj, line 68: It appears that firebase integration tests use the project "jscore-sandbox-141b5".  https://github.com/firebase/firebase-js-sdk/blob/5a9a14473674d7f03de3b3b4cdc1e66e653f5790/config/ci.config.json#L5

We need to enable MFA on this project, need to figure out if we can reuse this same project or need a different one.

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    let auth: Auth;
+    let idToken: string;
+    let signUpCred: UserCredential;
+    let email: string;
+    let assertion: MultiFactorAssertionImpl;
+    let _request: MFA.StartTotpMfaEnrollmentRequest;
+    let startMfaResponse: MFA.StartTotpMfaEnrollmentResponse;
+    let displayName: string;
+  beforeEach(async () => {
+    auth = getTestInstance();
```
prameshj, line 78: We don't need a new email/user per test case. might be enough to initialize just auth and email in the beforeEach block here.

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      
+
+     
+      const mfaUser =  multiFactor(cr.user);
+      sinon.spy(MultiFactorSessionImpl, '_fromIdtoken');
+ 
+      sinon.stub(mfaUser, 'getSession').returns(
+       Promise.resolve(MultiFactorSessionImpl._fromIdtoken(idToken, auth as any)));
+        
+      sinon.stub(MFA, 'startEnrollTotpMfa').callsFake((_auth,_request)=>{
```
prameshj, line 108: We won't need to stub any calls, we will invoke apis to the actual backend (or emulator).

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      expect(MultiFactorSessionImpl._fromIdtoken).to.have.been.calledOnce;
+      //expect(TotpSecret._fromStartTotpMfaEnrollmentResponse).to.have.been.calledOnce;
+      expect(MFA.startEnrollTotpMfa).to.have.been.calledOnce;
+
+      expect(await MFA.startEnrollTotpMfa(auth as any, _request)).to.eql(startMfaResponse)
+      
+      expect(totpSecret.secretKey).to.eql(startMfaResponse.totpSessionInfo.sharedSecretKey)
+      expect(totpSecret.codeLength).to.eql(startMfaResponse.totpSessionInfo.verificationCodeLength)
+
+      const totpVerificationCode = await mockTotp(totpSecret.secretKey, totpSecret.codeLength, totpSecret.codeIntervalSeconds);
```
prameshj, line 135: nit: instead of mockTotp, we can call it "getTotpCode" since it generates a real code for the given params.

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      const multiFactorAssertion = TotpMultiFactorGenerator.assertionForEnrollment(
+        totpSecret,
+        totpVerificationCode
+      );
+      console.log(totpVerificationCode);
+      // auth/invalid-idToken
+      await expect(mfaUser.enroll(multiFactorAssertion, displayName)).to.be.rejectedWith('auth/invalid-user-token')
```
prameshj, line 143: This error will change once we use actual tokens.

prameshj (commented):
Thanks for working on this!

packages/auth/test/integration/flows/totp.test.ts
bhparijat: ACK

packages/auth/test/helpers/integration/helpers.ts
```diff
@@ -22,8 +22,12 @@ import { Auth, User } from '../../../src/model/public_types';
...
 import { _generateEventId } from '../../../src/core/util/event_id';
 import { getAppConfig, getEmulatorUrl } from './settings';
-import { resetEmulator } from './emulator_rest_helpers';
-
+import {
+  getOobCodes,
+  OobCodeSession,
+  resetEmulator
+} from './emulator_rest_helpers';
+const totp = require('totp-generator');
```
sam-gc, line 30: Can this be an `import` instead?

packages/auth/test/helpers/integration/helpers.ts
```diff
@@ -76,9 +80,15 @@ export function getTestInstance(requireEmulator = false): Auth {
...
 
-export async function cleanUpTestInstance(auth: Auth): Promise<void> {
+export async function cleanUpTestInstance(
+  auth: Auth,
+  tests?: string
+): Promise<void> {
   await auth.signOut();
-  await (auth as IntegrationTestAuth).cleanUp();
+
+  if (typeof tests === 'undefined') {
```
sam-gc, line 89: Why does this "tests" parameter determine whether or not we call `cleanUp()`? (Also, why shouldn't the TOTP tests call `cleanUp()`?)

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      TotpMultiFactorGenerator.assertionForEnrollment(
+        totpSecret,
+        totpVerificationCode + '0'
+      );
+
+    await expect(
+      mfaUser.enroll(multiFactorAssertion, displayName)
+    ).to.be.rejectedWith('auth/invalid-verification-code');
+  });
+  it('should enroll using correct otp', async () => {
```
sam-gc, line 78: Nit: space above this line

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,166 @@
...
+        totpVerificationCode
+      );
+    await expect(mfaUser.enroll(multiFactorAssertion, displayName)).to.be
+      .fulfilled;
+  });
+
+  it('should not allow sign-in with incorrect totp', async () => {
+    let resolver;
+    // Added a delay so that getTotpCode() actually generates a new totp code
+    await delay(30 * 1000);
```
sam-gc, line 105: These 30 second delays are going to cause a much longer presubmit :(

We should investigate getting these tests into the emulator ASAP

prameshj: I don't think a timeout is necessary here.. we can just pass in a known invalid code like "abcde" for this test case.. valid codes will be numbers.

packages/auth/test/helpers/integration/helpers.ts
```diff
@@ -76,9 +80,15 @@ export function getTestInstance(requireEmulator = false): Auth {
...
+    await (auth as IntegrationTestAuth).cleanUp();
```
prameshj, line 90: If we are doing with skipping cleanup of hard-coded email, I prefer modifying https://github.com/firebase/firebase-js-sdk/blob/2a12a01c5113a3b0476d1775f4bd84697a5eb339/packages/auth/test/helpers/integration/helpers.ts#L70 to skip deleting users that have the string "donotdelete" in them. We use this same string to construct our hardcoded email.

sam-gc: +1, this feels clearer

sam-gc (approved):

packages/auth/test/helpers/integration/helpers.ts
```diff
@@ -93,3 +95,24 @@ function stubConsoleToSilenceEmulatorWarnings(): sinon.SinonStub {
...
+
+  return token;
+}
+
+export function delay(dt: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, dt));
+}
+
+export const email = 'totpuser-donotdelete@test.com';
+export const incorrectTotpCode = '1000000';
```
prameshj, line 118: nit: can we make the incorrect code obvious like "abcd"? Or add a comment like "1000000" is always incorrect since it has 7 digits and we expect 6.

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,156 @@
+/**
+ * @license
+ * Copyright 2020 Google LLC
```
prameshj, line 3: nit: 2022

packages/auth/test/integration/flows/totp.test.ts
```diff
@@ -0,0 +1,156 @@
...
+        'auth/invalid-verification-code'
+      );
+    }
+  });
+
+  it('should allow sign-in with for correct totp and unenroll successfully', async () => {
+    let resolver;
+
+    await delay(30 * 1000);
+    // Added a delay so that getTotpCode() actually generates a new totp code
```
prameshj, line 127: add a "// TODO(bhparijat) generate the otp code for the next time window by passing the appropriate timestamp to avoid the 30s delay. The delay is needed because the otp code used for enrollment cannot be reused for signing in.

prameshj (commented):
3 small nits, lgtm otherwise. thanks for adding this!

prameshj (approved):

PR #6785, opened 2022-11-12
OR queries more validation

MarkDuckworth:
OR queries more validation


packages/firestore/src/lite-api/query.ts
```diff
@@ -168,10 +168,12 @@ export function query<T>(
 
   if (nonFilters !== undefined) {
     queryConstraints = queryConstraints.concat(
-      nonFilters as QueryNonFilterConstraint[]
+      nonFilters as AppliableConstraint[]
```
ehsannas, line 171: I realize it's not a change made in this PR, but I just noticed it:
it looks like there's a bug on line 166 above (the return value of concat is ignored)?

ehsannas (approved):

packages/firestore/src/lite-api/query.ts
```diff
@@ -154,24 +155,22 @@ export function query<T>(
...
 export function query<T>(
   query: Query<T>,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  filterOrQueryConstraints: any,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ...nonFilters: any
+  filterOrQueryConstraint:
+    | QueryCompositeFilterConstraint
+    | QueryConstraint
+    | undefined,
```
ehsannas, line 161: why do we need `| undefined` here?

MarkDuckworth: `filterOrQueryConstraint` can be undefined in this case `query(collection)`. The code would also work if the parameter is declared optional `filterOrQueryConstraint?`, but it also works if the parameter is not declared optional or `| undefined`. Out of the three, I wanted to be explicit, but could also go with making it optional...

ehsannas: 👍

ehsannas (approved):

PR #6786, opened 2022-11-12
Fixed event comparison issue in Syncpointspec tests

maneesht:
Fixes problem where comparisons between expected and actual events weren't happening correctly. 

Fixing this caused other tests to fail because we expected specific events to be raised before other events. However, this is not how the SDKs work now. They order events based on where the event is in the synctree, not based on type of the event.


packages/database/test/helpers/syncpoint-util.ts
```diff
@@ -229,22 +229,7 @@ export class SyncPointTestParser {
       }
     };
 
-    const EVENT_ORDERING = [
-      'child_removed',
-      'child_added',
-      'child_moved',
-      'child_changed',
-      'value'
-    ];
```
IanWyszynski, line 238: @puf It seems that at one point we were enforcing this order. Just FYI.

IanWyszynski (approved):

egilmorez (approved):
Approving as formality (no doc impact).  Thanks!

PR #6796, opened 2022-11-15
TypeScript 4.7.4 upgrade for Firestore

dconeybe:


hsubox76:
> Note that grpc-js was pinned at 1.6.12 because 1.6.0 and 1.7.3 all caused seemingly-unrelated timeout errors in the tests (e.g. [1](https://github.com/firebase/firebase-js-sdk/actions/runs/3491296337/jobs/5843761473), [2](https://github.com/firebase/firebase-js-sdk/actions/runs/3485480218/jobs/5831056934)). The root cause is unknown, so I just accepted that v1.6.12 is good. I needed to upgrade grpc-js to fix some typing issues that resulted from upgrading TypeScript.

Is the grpc-js pin comment out of date? Saw the latest commit bumps it to ~1.7.0

MarkDuckworth (approved):
LGTM. Thank you for taking this one off my plate. I'm curious about the other comment about grpc version.

dconeybe:
> Is the grpc-js pin comment out of date? Saw the latest commit bumps it to ~1.7.0

Yes. Thanks for pointing that out. I've removed the description.

.changeset/breezy-wombats-care.md
```diff
@@ -0,0 +1,8 @@
+---
+'@firebase/database': minor
```
hsubox76, line 2: Why database?

hsubox76 (approved):
Thanks for doing this, I'll approve so as not to hold you up, just remove `database` from the changelog if it was an oversight or leave it be if there's a reason.

.changeset/breezy-wombats-care.md
dconeybe: I had mentioned `database` because the upgrade of `grps-js` I think indirectly affected its version as well. But looking more closely, that probably isn't the case. I'll remove it.

PR #6799, opened 2022-11-15
Upgrade individual packages to TS 4.7.4

hsubox76:
Previously only the main package.json had its typescript version bumped.


packages/performance/src/utils/attributes_utils.ts
```diff
@@ -44,12 +44,12 @@ const enum EffectiveConnectionType {
  *
  * ref: https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation
  */
-interface NetworkInformation {
+interface NetworkInformationWithEffectiveType extends NetworkInformation {
```
lahirumaramba, line 47: Just curious: what made these types to rename to `NetworkInformationWithEffectiveType`? Could this be breaking?

lahirumaramba (approved):

packages/performance/src/utils/attributes_utils.ts
hsubox76: I don't think so, it's not publicly exposed and is only used here. I will add a comment about what the deal is:

```
  // `effectiveType` is an experimental property and not included in
  // TypeScript's typings for the native NetworkInformation interface
  ```

File: integration/compat-interop/package.json [hunk 2]
```diff
@@ -24,6 +24,6 @@
     "@firebase/remote-config-compat": "0.1.16"
   },
   "devDependencies": {
-    "typescript": "4.2.2"
+    "typescript": "4.7.4"
   }
 }
```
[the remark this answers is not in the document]
hsubox76: Sure, while we're in this file. Fixed.

Issue #3707, opened 2020-08-30
Error: @firebase/database tried to access @firebase/app, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.


ScottPierce:
### [REQUIRED] Describe your environment

  * Operating System version: _____
  * Browser version: _____
  * Firebase SDK version: _____
  * Firebase Product: _____ (auth, database, storage, etc)

### [REQUIRED] Describe the problem
I get an error when bundling with Yarn 2 and firebase:
```
(node:82642) [MODULE_NOT_FOUND] Error: @firebase/database tried to access @firebase/app, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

#### Steps to reproduce:

I'm building a create-react-app application with yarn2 with pnp that depends on firebase, and it outputs the following error:
```
(node:82642) [MODULE_NOT_FOUND] Error: @firebase/database tried to access @firebase/app, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

Feiyang1:
Thanks for the report! Are you able to get it to work using `loose` mode (I tried it, but it still didn't work)? Otherwise I'm afraid firebase is not compatible with `yarn 2` pnp at the moment. 

As the message states, `@firebase/app` is not declared as a dependency for `@firebase/database`. This is for working around the `unmet peerDependency` warning for `firebase-admin` users (see this [PR](https://github.com/firebase/firebase-js-sdk/pull/2082)). I admit it is not an ideal solution and we probably should create a separate npm package to be used in `firebase-admin`, so we can list dependencies correctly.

I will keep it open as a feature request, but it is probably a low priority for us right now.
@hiranya911 FYI

PutziSan:
This problem is still there.

Currently workaround is adding this to `.yarnrc.yml`

```yaml
packageExtensions:
  # https://github.com/firebase/firebase-js-sdk/issues/3707
  "@firebase/database@*":
    peerDependencies:
      "@firebase/app": "*"
      "@firebase/app-types": "*"
```

Zloka:
> Thanks for the report! Are you able to get it to work using `loose` mode (I tried it, but it still didn't work)? Otherwise I'm afraid firebase is not compatible with `yarn 2` pnp at the moment.
> 
[...]

PR #6818, opened 2022-11-21
updating database package dependency to add required peer dependency

john-arccos:
Hi! I noticed this warning when using firebase-admin in our app.

I was going through the test set up and was almost complete, but then it asked me to switch to the "pay as you go" plan. I'm not sure I feel comfortable signing up for that as I'm pretty unfamiliar with firebase and mostly just want to clean up our dependency tree. I don't really want to pay money just to run this test suite. Please advise on how to continue.

This is the warning we got when running `yarn`:

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ @firebase/database@npm:0.13.10 doesn't provide @firebase/app-types (pc007b), requested by @firebase/auth-interop-types
```

and when I use `yarn link` to link this version of `@firebase/database` that warning goes away.

### DIFF
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -45,6 +45,7 @@
   "license": "Apache-2.0",
   "peerDependencies": {},
   "dependencies": {
+    "@firebase/app-types": "0.x",
     "@firebase/logger": "0.3.4",
     "@firebase/util": "1.7.3",
     "@firebase/component": "0.5.21",


hsubox76:
We're looking into fixing this. The source of this issue is somewhat complex and has to do with the fact that the `@firebase/database` package is used both in the JS SDK and as a dependency of the Admin Node SDK, and that each use case really requires a different package.json (the Admin Node version should not depend on app or app-types at all), but we've been hacking solutions so far by trying to make a package.json that somehow works for both.

It seems to me with increasing Yarn2 and PNPM problems that this isn't sustainable anymore and we may need to go ahead and make a separate NPM package for the database package consumed by Admin SDK.

Related ongoing issues:
https://github.com/firebase/firebase-js-sdk/issues/3707
https://github.com/firebase/firebase-admin-node/issues/973

PR #6826, opened 2022-11-22
Put firebase/* entry point ESM files in their own subdirs with "type: module" package.json 

hsubox76:
See https://github.com/firebase/firebase-js-sdk/issues/6747

More build tooling these days is requiring all ESM bundles conform to Node ESM rules, even if the bundles are mainly used in the browser (for example, many test frameworks run browser tests in Node and require all bundles to meet Node ESM rules when the tests are run). This may also affect SSR/SSG.

The rules are essentially that the extension must be `.mjs` OR the nearest package.json file must have a line that says `type: "module"`.

We avoid putting an mjs extension on browser bundles because some older build tools and environments don't know how to handle that extension. Instead, in the `@firebase/*` packages, we have put ESM files in their own subdirectory in `dist/` (`dist/esm/`) and used a rollup plugin put a one-line package.json file in there with them that says "type: module". This build step had not been applied to the `firebase/*` packages (the entry point packages that just redirect to the equivalent `@firebase/*` package).

It's possible that some tooling is happy enough with the end package (`@firebase/*`) having this setup (the way it is now), while other tooling requires the Node ESM indicators to be there at every step of resolution (explaining issues still being reported by some users).

I've added this step to the entry point `firebase/*` packages in this PR hoping this will fix it for those users.

### DIFF
--- a/.changeset/eighty-ants-do.md
+++ b/.changeset/eighty-ants-do.md
@@ -1,0 +1,5 @@
+---
+'firebase': patch
+---
+
+Modify entry point ESM bundles to conform to Node ESM specs.
--- a/packages/firebase/analytics/package.json
+++ b/packages/firebase/analytics/package.json
@@ -1,7 +1,7 @@
 {
   "name": "firebase/analytics",
   "main": "dist/index.cjs.js",
-  "browser": "dist/index.esm.js",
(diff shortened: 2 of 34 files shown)


dwyfrequency (approved):
LGTM

PR #6829, opened 2022-11-24
Revert "TypeScript 4.7.4 upgrade for Firestore"

dconeybe:
This reverts #6796 which upgraded TypeScript to 4.7.4 for Firestore. The revert is being done because the upgrade somehow broke the browser tests in GitHub Actions. I'm going to investigate and re-push the upgrade once the root cause of the browser test failures is found and fixed.


dconeybe:
Hmm, this didn't fix the github actions failures at all. Closing.

PR #6833, opened 2022-11-28
Fix Firefox CI issue

hsubox76:
Github's latest image for ubuntu is ubuntu-22.04 which seems to come with a version of Firefox that causes the karma launcher to hang and eventually give up when trying to launch Firefox. Other users have reported the Firefox bin isn't there at all, which is not exactly the same issue, but could be related. Locking it to ubuntu-20.04 until we can figure out why (hopefully it isn't that current and future versions of Firefox don't work with karma-firefox-launcher).


dwyfrequency (approved):
LGTM

dconeybe:
Thansk @hsubox76!

PR #6839, opened 2022-11-30
Add bloom filter to existence filter and watchFilters spec builder

milaGGL:
- [x] Add nullable bloom filter to ExistenceFilter class
- [x] Update watchFilters to optionally accept a bloom filter without exploding existing tests

Note: The logic for using bloom filter and skipping full re-query is not implemented yet.

### DIFF
--- a/packages/firestore/src/protos/firestore_proto_api.ts
+++ b/packages/firestore/src/protos/firestore_proto_api.ts
@@ -456,8 +456,7 @@ export declare type BeginTransactionRequest =
   firestoreV1ApiClientInterfaces.BeginTransactionRequest;
 export declare type BeginTransactionResponse =
   firestoreV1ApiClientInterfaces.BeginTransactionResponse;
-  export declare type BloomFilter =
-  firestoreV1ApiClientInterfaces.BloomFilter;
+export declare type BloomFilter = firestoreV1ApiClientInterfaces.BloomFilter;
 export declare type CollectionSelector =
   firestoreV1ApiClientInterfaces.CollectionSelector;
 export declare type CommitRequest =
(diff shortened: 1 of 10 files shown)


packages/firestore/test/unit/specs/spec_builder.ts
```diff
@@ -769,18 +770,19 @@ export class SpecBuilder {
     return this;
   }
 
-  watchFilters(queries: Query[], ...docs: DocumentKey[]): this {
```
milaGGL, line 772: Original watchFilters accept an array of targetIds (limited to only 1 targetId), and spread operator to pass in unlimited number of docs(only to get count later on).
updated the watchFilters to accept array of targetIds, array of docs(default to empty array), and an optional bloom filter.

packages/firestore/src/protos/firestore_proto_api.ts
```diff
@@ -456,6 +456,8 @@ export declare type BeginTransactionRequest =
   firestoreV1ApiClientInterfaces.BeginTransactionRequest;
 export declare type BeginTransactionResponse =
   firestoreV1ApiClientInterfaces.BeginTransactionResponse;
+  export declare type BloomFilter =
```
dconeybe, line 459: ultranit: I think this line should be un-indented to be consistent with the surrounding lines.

packages/firestore/src/remote/existence_filter.ts
```diff
@@ -15,7 +15,9 @@
  * limitations under the License.
  */
 
+import { BloomFilter as ProtoBloomFilter } from '../protos/firestore_proto_api';
+
 export class ExistenceFilter {
   // TODO(b/33078163): just use simplest form of existence filter for now
-  constructor(public count: number) {}
+  constructor(public count: number, public unchangedNames?: ProtoBloomFilter) {}
```
dconeybe, line 22: You can actually remove the comment on the line above this one. The linked bug is about adding bloom filters, which you are doing!

dconeybe (changes requested):
Looks great! Just two extremely minor comments.

dconeybe (approved):

Issue #6801, opened 2022-11-16
singInWithCustomToken not working in Firefox with third party cookies disabled


mwisniewski2000:
### [REQUIRED] Describe your environment

  * Operating System version: macOs Monterey 12.4
  * Browser version: Firefox 107.0 (64-bit)
  * Firebase SDK version: 9.14.0
  * Firebase Product: auth

#### Steps to reproduce:
When user blocks all third party cookies on Firefox browser and tries to login with `signInWithCustomToken`  this error shows: `FirebaseError: Firebase: Error (auth/network-request-failed).`, while when we block third party cookies on chrome everything works perfectly.
Is there some way to make it work, or is it just impossible with third party cookies blocked in firefox?
<img width="516" alt="Screenshot 2022-11-15 at 19 51 16" src="https://user-images.githubusercontent.com/108661625/202186840-69549311-5a41-4603-93d6-151e5bc0da81.png">
<img width="722" alt="Screenshot 2022-11-15 at 19 50 45" src="https://user-images.githubusercontent.com/108661625/202186991-f37d2953-e086-428b-8543-43e78e9b2706.png">
#### Relevant Code:

```javascript
await signInWithCustomToken(auth, token);
```

sam-gc:
This is odd, can you check the network panel to see if Firefox is directly blocking the relevant API call? `signInWithCredential` translates to a straightforward REST call.

mwisniewski2000:
> This is odd, can you check the network panel to see if Firefox is directly blocking the relevant API call? `signInWithCredential` translates to a straightforward REST call.

In Firefox network tab I can't see any request at all.  In chrome with disabled third party cookies I can see this `accounts:signInWithCustomToken` request.

prameshj:
> In Firefox network tab I can't see any request at all. In chrome with disabled third party cookies I can see this accounts:signInWithCustomToken request.

Interesting, that looks like Firefox might be blocking the network request from happening. Any console logs that hint what might be going on?

mwisniewski2000:
> > In Firefox network tab I can't see any request at all. In chrome with disabled third party cookies I can see this accounts:signInWithCustomToken request.
> 
> Interesting, that looks like Firefox might be blocking the network request from happening. Any console logs that hint what might be going on?

Just this error I attached in description :(

abilicz:
Its basically this part of code throwing operation insecure:
` this._heartbeatsCachePromise = this._storage.read().then(result => {
[...]

PR #6841, opened 2022-11-30
6801 catch insecure indexedDB access

abilicz:
Related issue:
https://github.com/firebase/firebase-js-sdk/issues/6801

### DIFF
--- a/.changeset/three-glasses-judge.md
+++ b/.changeset/three-glasses-judge.md
@@ -1,0 +1,5 @@
+---
+'@firebase/util': patch
+---
+
+Fix for third party window content that cannot access IndexedDB if the browser is set to never accept third party cookies on Firefox.
--- a/packages/util/src/environment.ts
+++ b/packages/util/src/environment.ts
@@ -142,9 +142,8 @@ export function isSafari(): boolean {
 export function isIndexedDBAvailable(): boolean {
   try {
     return typeof indexedDB === 'object';
-  }
-  catch (e) {
-      return false;
+  } catch (e) {
+    return false;
   }
 }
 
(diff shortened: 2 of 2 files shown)


hsubox76:
I'm trying to reproduce the error you showed here https://github.com/firebase/firebase-js-sdk/issues/6801#issuecomment-1332711754

I set Firefox to block all third-party cookies and I can't seem to reproduce it. When I type in "indexedDB" it says `IDBFactory {  }`. I'm using Firefox 107.0.1 on Mac. I went to Settings > Privacy & Security, set Browser Privacy to Custom, and set Cookies to "All third-party cookies (may cause websites to break)".

abilicz:
I think the key part here is:

> Third party window content (e.g. [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) content) cannot access IndexedDB if the browser is set to [never accept third party cookies](https://support.mozilla.org/en-US/kb/third-party-cookies-firefox-tracking-protection?redirectslug=disable-third-party-cookies&redirectlocale=en-US)

So if you will open for example this page https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe , open console and type `indexedDB` it will be fine, but if you change evaluation context to some iframe inside it will throw that error 
Screen Shot 2022-12-01 at 8 19 36 PM

For us this is an issue because we use Firebase from inside an iframe hosted by us, but iframe is inside Jira/Azure context , so top domain is different. Looks like some companies like to have this cookie policy set on some browsers. Unfortunately  Firefox throws some exception.

hsubox76 (approved):

hsubox76:
Thanks! Got it. In that case the change looks good. Will merge when required checks pass.

hsubox76:
Oops, looks like it failed the formatting check. Can you run `yarn format` on it?

